PR #9, opened 2023-11-28
Add agent controllers

fzdarsky:
Adds two controllers that report TPM boot measurements (PCR values) and the status of user-selectable systemd units. The systemd controller should eventually also contain the logic to reload or restart systemd units based on config file changes.


mangelajo:
Oh damn, this probably should have gone first #9 now it's on conflict after #10

fzdarsky:
> Just a little rebase I guess

[x] done

oglok:
Tested and fully works. Some strange behavior I will document in an issue.

PR #11, opened 2023-12-01
Allow using a multi-arch image for postgresql

mangelajo:
Without this you can't locally deploy on an arm64 mac book. The caveat is that this image is behind login and you will need to run:
```
podman login registry.redhat.io
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,14 @@ make build
 
 ## Running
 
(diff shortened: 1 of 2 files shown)


fzdarsky:
@mangelajo could you please add the need for this extra step to the README?

mangelajo:
Oh, it breaks testing because the github runners don't have credentials for registry.redhat.io, hmmm

akoserwal:
It works getting up the flightctl server and db via podman.

PR #12, opened 2023-12-01
Enrollment QR

mangelajo:


mangelajo:
This breaks the device simulator. I need to handle this:

`E1201 19:37:47.919999 2831071 main.go:85] device-0000 : requestAndWaitForEnrollment: writeEnrollmentBanner: writeQRBanner: open /run/flightctl-banner: permission denied`

fzdarsky:
The new URLs point to the UI endpoint, not API, right? Better to make it configurable sooner than later, as we cannot assume the host is the same as that of the API.

To deal with the deviceSimulator, I'd suggest adding a field "bannerUpdater" to the agent struct that can be set via a setter to point to the "writeBanner()" function or kept nil.

Rest LGTM.

mangelajo:
> The new URLs point to the UI endpoint, not API, right? Better to make it configurable sooner than later, as we cannot assume the host is the same as that of the API.
> 
> To deal with the deviceSimulator, I'd suggest adding a field "bannerUpdater" to the agent struct that can be set via a setter to point to the "writeBanner()" function or kept nil.
> 
> Rest LGTM.

Ok, I made it simpler. Without enrollmentUiUrl, the code will just skip the banner creation/print. So the simulator is not impacted by it.

mangelajo:
This has also been rebased so we all tests pass we could merge it.

PR #14, opened 2023-12-05
Fix create-or-update

fzdarsky:


cmd/flightctl/main.go
```diff
@@ -338,8 +338,8 @@ func applyFromReader(client *client.ClientWithResponses, filename string, r io.R
 
 		switch kind {
 		case "Device":
-			device := &api.Device{}
-			err := yaml.Unmarshal(buf, device)
+			var device api.Device
+			err := json.Unmarshal(buf, &device)
```
oglok, line 342: yaml.Unmarshal() for consistency

oglok (commented):
Just a small comment. Otherwise, lgtm

cmd/flightctl/main.go
fzdarsky: [x] fixed

PR #20, opened 2023-12-12
Add fleet logic

fzdarsky:
This PR updates the API to add configuration providers to Fleet resources and to update which fields are mandatory (and thus cannot be nil) and optional (can be nil).
It also wraps ArgoCD's git repo provider and caching layer packages, so they can be reused for fetching OS config data, too.


mangelajo:
I rebases and handled the go.sum conflict. Merging once it passes CI, and then continuing on separate PR.

PR #21, opened 2023-12-12
Add RPM spec and systemd unit

oglok:


mangelajo:
It looks good for an initial version. Later on we can handle the versioning etc. But this will serve us already.

PR #24, opened 2023-12-13
Add config struct for agent to read from file

oglok:


mangelajo:
LGTM, just the conflict

internal/config/config.go
```diff
@@ -65,6 +75,13 @@ func NewDefault() *Config {
 			CertStore: CertificateDir(),
 			BaseUrl:   "http://localhost:3333/api",
 		},
+		Agent: &agentConfig{
```
mangelajo, line 78: At some point I agree we may want separate configs, but it's good for the POC

mangelajo (commented):
It looks good to me . With the comment of us needing to split the config files in a future.

PR #30, opened 2023-12-19
Support altnames in flightctl-server config

mangelajo:
Support altNames in server cert:
```
            X509v3 Subject Alternative Name:
                DNS:my-host.abc.com, DNS:localhost
```


oglok:
/lgtm

:-)

PR #31, opened 2023-12-19
EnrollmentRequestApproval schema

mangelajo:


api/v1alpha1/openapi.yaml
```diff
@@ -829,6 +835,9 @@ components:
     DeviceSpec:
       type: object
       properties:
+        fleetName:
```
fzdarsky, line 838: Making this part of the device spec implies the user explicitly specifies the fleet, which implies that fleet must exist and it inverts the logic of fleet label-selecting devices. Unless we want to abandon label-selection, I'd suggest moving this out of spec, either into a metadata field that's set by the fleet controller (maybe "ownerRef" or an annotation).

avishayt: How does the LabelSelector work - what if a device's labels match two fleet specs?  Does it make sense to reimage a device based on a label?

api/v1alpha1/openapi.yaml
```diff
@@ -940,6 +949,31 @@ components:
           type: string
           description: The active state of the systemd unit.
       description: The status of the systemd unit.
+    EnrollmentRequestApproval:
+      type: object
+      properties:
+        fleetName:
```
avishayt, line 955: Is the enrollment per fleet?  If I want to move the device to another fleet, I need a new enrollment?

mangelajo: We could later on change the fleetName field of a device, if we want to allow that.

api/v1alpha1/openapi.yaml
mangelajo: I think label selection is a bad idea, because we can have a device being selected by multiple fleets.

I believe it's a safer route to do explicit assignments.

mangelajo: I thought of using ownerRef's ... it's an option, I think it's better than label selections. Still at the database level we need to extract the fleet name, and have an index of that... otherwise there is no easy way to pull al devices from a fleet in a query.

avishayt: I don't think it's an owner - it's a device property

fzdarsky: Yeah, I agree to the problem of devices being selected by multiple fleets. That can also happen in K8s and there it's seen as "user problem". Then again, mistakes are less costly to fix in K8s.

In principle, I'm ok with inverting, but remember we don't want users to have to manipulate device resources but to work at the fleet resource level instead. Think about what you'd store in Git to fully describe the system. If we need to start storing/updating thousands of device manifests in Git as a result that'd not be a good design.

Also, the device's spec is a copy / rendering of the fleet's device template, so I don't think the fleet reference should be there.

api/v1alpha1/openapi.yaml
fzdarsky: We're enrolling devices into the system. Reassigning to another fleet shouldn't require a new enrolment, nor should migrating to another service instance in another region.

mangelajo: We could remove the fleetName from the approval, so then the enrollment approval stops being related to a specific fleet.

fzdarsky:
@mangelajo LGTM, except it would have been nice to have had a separate PR for adding the labels… and then doing so as part of the Resource struct and making the labels available to the other resource kinds, too.

Was there a specific reason you didn’t put it into the Resource struct?

mangelajo:
> @mangelajo LGTM, except it would have been nice to have had a separate PR for adding the labels… and then doing so as part of the Resource struct and making the labels available to the other resource kinds, too.
> 
> Was there a specific reason you didn’t put it into the Resource struct?

As we talked in slack, making the indexing of the different resources generic enough was taking me too much thinking. I will handle in follow up.

mangelajo:
fixed a missing change in the cmdline https://github.com/flightctl/flightctl/compare/74184aa..0a8a7c7084a0ae656ee2fd6449318a1f534b24e9

Merging as per slack conversation

PR #34, opened 2023-12-20
Creation of the flightctl agent image

oglok:


packaging/Containerfile.fedora
```diff
@@ -0,0 +1,19 @@
...
+    chmod 0600 /usr/etc-system/root.keys
+VOLUME /var/roothome
+
+ADD packaging/flightctl-public-service/config.yaml /etc/flightctl/
+ADD packaging/flightctl-public-service/ca.crt /etc/flightctl
+ADD packaging/flightctl-public-service/client-enrollment.* /etc/flightctl/
+
+RUN rpm-ostree install -y /tmp/flightctl-agent-0.0.1-1.el9.x86_64.rpm
+RUN ln -s /usr/lib/systemd/system/podman.socket /usr/lib/systemd/system/multi-user.target.wants/
+RUN ln -s /usr/lib/systemd/system/flightctl-agent.service /usr/lib/systemd/system/multi-user.target.wants/
```
fzdarsky, line 18: Shouldn't these symlinks be created from the RPM spec's `%install` section?
Also, are we missing a symlink from the bannerfile in `/run/flightctl-banner` into `%{buildroot}%{_sysconfdir}/issue.d/flightctl.issue`?

README.md
```diff
@@ -65,3 +65,23 @@ podman-compose -f deploy/podman/observability.yaml up
 ```
 
 The Grafana and Prometheus web UIs are then accessible on `http://localhost:3000` and `http://localhost:9090`, respectively.
+
+
+## Public Service Enrollment
+
+A public version of FlightCtl is running on a OpenShift cluster managed by Red Hat. This provide the capability of having a public service for enrolling your own devices and showcase FlightCtl.
```
fzdarsky, line 72: That service is for internal development/demos, so not accessible publicly. How about we keep this section generic, describing the general process how to build and use a bootc image with the agent and configuring it for a given service?

packaging/Containerfile.fedora
```diff
@@ -0,0 +1,19 @@
+FROM quay.io/centos-bootc/fedora-bootc:eln
+
+COPY rpmbuild/RPMS/x86_64/flightctl-agent-0.0.1-1.el9.x86_64.rpm /tmp/
```
fzdarsky, line 3: Would it make sense to use a COPR repo to host this RPM instead?

fzdarsky (changes requested):

packaging/Containerfile.fedora
oglok: Having flightctl-agent service enabled by default (via rpm) is a decision to be made, of course. If we all agree that's the way to go, I can change it. No problem!

Regarding the banner, #33 just took care of that, by creating the QR code in there.

README.md
oglok: ack

packaging/Containerfile.fedora
oglok: is the ration work vs benefit worth it? I think having a Makefile target doing the rpm build is enough for now.

mangelajo: I think currently a local build would be enough, and helps with easy-quick development cycle.

packaging/Containerfile.fedora
fzdarsky: If we write the file directly to `/etc` now, why still keep the `/run` one?

oglok: right, I don't know why @mangelajo chose that location. If it was an arbitrary one, we can certainly remove it.

fzdarsky: Probably because `transmission` used a symlink, which it did because `zezere` did it back then. I don't recall the reason anymore, really. If there was a good reason, I could imagine it was around SELinux (re)labeling or simply that /etc is for config and processes shouldn't be writing to it really.

packaging/systemd/flightctl-agent.service
```diff
@@ -2,8 +2,10 @@
 [Unit]
 Description=Flightctl Agent
 After=network.target
+Before=getty@tty1.service
 
 [Service]
+ExecStartPre=/bin/systemctl start podman.socket
```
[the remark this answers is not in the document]
oglok: Now podman is enabled in the Containerfile as part of the image. Let's remove this then.

README.md
```diff
@@ -65,3 +65,73 @@ podman-compose -f deploy/podman/observability.yaml up
...
+The certificates will be stored probably in `$HOME/.flightctl/certs/`.
+
+The config.yaml should look like this:
+
+```
+agent:
+  server: https://localhost:3333
+  enrollementUi: https://localhost:3333
+  statusUpdateInterval: 1m0s
+  fetchSpecInterval: 1m0s
```
mangelajo, line 121:
```suggestion
agent:
  server: https://localhost
  enrollementUi: https://localhost
  statusUpdateInterval: 1m0s
  fetchSpecInterval: 1m0s
```

With the helm files we are providing it just goes to plain 443

README.md
```diff
@@ -65,3 +65,73 @@ podman-compose -f deploy/podman/observability.yaml up
...
+```
+
+Change localhost by the IP of the machine where the service is exposed. Take into account that these services must be exposed to where the agent is going to be deployed.
```
mangelajo, line 124:
```suggestion
Change localhost by the IP/hostname and port where the service is exposed. Take into account that these services must be exposed to where the agent is going to be deployed.
```

What do you mean by "Take into account that these services must be exposed to where the agent is going to be deployed." ?

README.md
```diff
@@ -65,3 +65,73 @@ podman-compose -f deploy/podman/observability.yaml up
...
+```
+podman build -t quay.io/$MYQUAYUSER/flightctl-agent:latest -f packaging/Containerfile.fedora ./
+podman push quay.io/$MYQUAYUSER/flightctl-agent:latest
+```
+
+In order to make the deployment of this image easy, you can convert it to QCOW2 format and use it as a raw disk in Libvirt. The following command will create a file called disk.qcow2 within the output folder:
+
+```
+mkdir output && sudo podman run     --rm     -it     --privileged     --pull=newer     --security-opt label=type:unconfined_t     -v $(pwd)/output:/output     quay.io/centos-bootc/bootc-image-builder:latest   quay.io/$MYQUAYUSER/flightctl-agent:latest
+```
```
mangelajo, line 137:
```suggestion
mkdir output && sudo podman run --rm -it --privileged --pull=newer \
                                                    --security-opt label=type:unconfined_t \
                                                    -v $(pwd)/output:/output \
                                                    quay.io/centos-bootc/bootc-image-builder:latest \
                                                    quay.io/$MYQUAYUSER/flightctl-agent:latest
```

since I am in nit picking mode, feel free to ignore ':D

mangelajo: welll it ended weird.. mehh

packaging/Containerfile.fedora
mangelajo: I don't know :D

README.md
oglok: Thanks, actually it had the indentation from the terminal separated in different lines, but I copied the wrong output!

README.md
oglok: The helm charts are only for K8s deployments. Did that change for podman compose local deployments too? I'd like to put it as generic as possible.

README.md
mangelajo: May be... take into account these services must be accessible from where the agent is going to be deployed ?

oglok: Hey! yes, brain switched off yesterday night... xD

I'll rephrase it, I just wanted to say that services must be reachable by the agent device.

fzdarsky: Using active voice may be easier to read: "Take into account that agents must be able to connect to the service."

oglok: +1

oglok:
PTAL @fzdarsky @mangelajo

mangelajo (approved):
LGTM

packaging/Containerfile.fedora
fzdarsky: Ok, having the agent write it directly, the SELinux context becomes `unconfined_u:object_r:etc_t:s0` instead of `system_u:object_r:etc_t:s0`, but that should be fine. Let's remove the code that writes the duplicate banner to `/run` then?

oglok: Ok, I'll add that in a separate PR to avoid mixing fixes and have a better tracking history. Is that fine?

oglok: #38

PR #36, opened 2023-12-21
Fix wrong resource names in lists

fzdarsky:
Fixes a bug introduced by the recent refactoring of the data store where resource names were copied by pointer not value, resulting in wrong name being used in lists of resources.


mangelajo:
That wasn't an easy one!, 👍

mangelajo (approved):

PR #45, opened 2023-12-27
Pagination

### LINKED ISSUE #18: Add paging via limit-continue

avishayt:
Resolves #18


cmd/flightctl/main.go
```diff
@@ -209,21 +216,23 @@ func RunGet(kind, name string, labelSelector string, output string) error {
...
 			}
-			fmt.Println(string(marshalled))
 		} else {
-			params := &api.ListDevicesParams{}
+			params := api.ListDevicesParams{}
 			if len(labelSelector) > 0 {
 				params.LabelSelector = &labelSelector
 			}
-			response, err := c.ListDevicesWithResponse(context.Background(), params)
+			if limit > 0 {
```
filanov, line 228: you can make them pointers and avoid checking the value

internal/service/device.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	"github.com/go-openapi/swag"
 	"github.com/google/uuid"
 	"gorm.io/gorm"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
 type DeviceStoreInterface interface {
 	CreateDevice(orgId uuid.UUID, device *api.Device) (*api.Device, error)
-	ListDevices(orgId uuid.UUID, labels map[string]string) (*api.DeviceList, error)
+	ListDevices(orgId uuid.UUID, labels map[string]string, limit int, cont string) (*api.DeviceList, error)
```
filanov, line 16: there will probably be additional parameters in the future, consider using struct parameters
```
type ListDeviceParams struct {
     Limit int
     Cont string
     ...
}
```

cmd/flightctl/main.go
avishayt: Done

internal/service/device.go
avishayt: Done

internal/store/common.go
```diff
@@ -0,0 +1,37 @@
...
+	return query
+}
+
+func FormatResourcesAfterPagination(query *gorm.DB, limit int, lastItemName string) (*string, *int64) {
+	if limit == 0 {
+		return nil, nil
+	}
+
+	var count int64
+	query.Where("name > ?", lastItemName).Count(&count)
```
mangelajo, line 31: This is expensive.

https://dba.stackexchange.com/questions/317626/optimizing-select-count-result-in-postgresql
https://wiki.postgresql.org/wiki/Slow_Counting

Apparently, the DB needs to walk all the rows for every query you make; if you keep iterating as you query this is an O(n!) complexity 

I propose we don't return a count, and we let iteratively discover. Alternatively, I think we may need to cache the count and provide an approximate count to avoid the O(n!) trap.

mangelajo (changes requested):
It looks great, but see what I found about the count. I propose we exclude it for now, later in time if we need we can cache/optimize.

I guess it's necessary for a meaningful pagination.

An alternative could be an API endpoint that returns the count and we just grab it once.

mangelajo:
Please note the count issue; it's only with devices that could have millions.

internal/store/common.go
filanov: We will probably be out of memory before we return millions of results. 
We can maybe try to enforce pagination? AWS is doing that, they limit the result to 1K objects.

avishayt:
> It looks great, but see what I found about the count. I propose we exclude it for now, later in time if we need we can cache/optimize.
> 
> I guess it's necessary for a meaningful pagination.
> 
> An alternative could be an API endpoint that returns the count and we just grab it once.

I agree that it's expensive, but I don't know how else to do it.
I don't agree that it's necessary - why not just return "continue" until there are no more records?  The backend can request limit+1 records to know if more exist.
Does "count" help with anything other than helping a UI know how many pages to expect?

internal/store/common.go
avishayt: I agree that we should limit to 1K objects.  Enforcing that later would be an API breakage.  @mangelajo WDYT?

mangelajo: Yeah, it should be limited. My worry is the count result, we can either cache it for subsequent pagination requests, or as discussed in slack, accept the previous count as the follow up pagination requests.

Copying here your description of the idea which I think is good: 

image

mangelajo:
> > It looks great, but see what I found about the count. I propose we exclude it for now, later in time if we need we can cache/optimize.
> > I guess it's necessary for a meaningful pagination.
> > An alternative could be an API endpoint that returns the count and we just grab it once.
> 
> I agree that it's expensive, but I don't know how else to do it. I don't agree that it's necessary - why not just return "continue" until there are no more records? The backend can request limit+1 records to know if more exist. Does "count" help with anything other than helping a UI know how many pages to expect?

I was thinking that it was only helpful for the UI but @filanov also commented that it can be useful for page statistics (get the count of devices in a fleet, etc..), for graphs, UI details.

filanov:
> > > It looks great, but see what I found about the count. I propose we exclude it for now, later in time if we need we can cache/optimize.
> > > I guess it's necessary for a meaningful pagination.
> > > An alternative could be an API endpoint that returns the count and we just grab it once.
> > 
> > 
> > I agree that it's expensive, but I don't know how else to do it. I don't agree that it's necessary - why not just return "continue" until there are no more records? The backend can request limit+1 records to know if more exist. Does "count" help with anything other than helping a UI know how many pages to expect?
> 
> I was thinking that it was only helpful for the UI but @filanov also commented that it can be useful for page statistics (get the count of devices in a fleet, etc..), for graphs, UI details.

For the overview in the UI i suggest to create a dedicated API, we probably don't want the UI to go over thousands or records and start summarizing them, the DB can do that much faster.

filanov (commented):
LGTM

I liked the query helper functions

PR #46, opened 2024-01-03
Add git source to REST API definition

avishayt:
It supports only HTTPS auth for now.  I will add SSH auth as a later step.


mangelajo:
We could already include an SSH PK as ssh access; probably would not require other changes, right?

including credential details here makes me a bit reluctant, but.. you need to put them somewhere, may be we need to model them in a way that those would be fetched from a K8s secret when running in k8s. (i.e. what we do in the Fleet object with alternative sources of authentication) i.e. https://docs.openshift.com/pipelines/1.13/secure/authenticating-pipelines-using-git-secret.html

mangelajo:
There is a compile error:
```
internal/service/handler.go:67:38: cannot use (*ServiceHandler)(nil) (value of type *ServiceHandler) as server.StrictServerInterface value in variable declaration: *ServiceHandler does not implement server.StrictServerInterface (missing method CreateGitSource)
```

mangelajo (approved):
LGTM :)

PR #49, opened 2024-01-10
Add log pkg

filanov:
This PR is the first step, this logger mechanism can be used for all the components and request id can be used between the service and the clients so a request can be debugged between the components.

Log package will enable standard logger and support request id using middleware.RequestID mechanism
Each package can contain a logger that will be used and adding a package source so it's east to find relevant code later.
It required to change funcitons signatures to pass context that is used to store the request-id and user it in the logs

with logrus without report caller
```
INFO[0003] db.Find(): 10 rows affected, error is <nil>   pkg=store request_id=rdu-infra-edge-03.infra-edge.lab.eng.rdu2.redhat.com/JdyoA8tWDH-000001
```
with logrus with report caller
```
INFO[0005]/root/test/flightctl/internal/store/enrollmentrequest.go:56 github.com/flightctl/flightctl/internal/store.(*EnrollmentRequestStore).ListEnrollmentRequests() db.Find(): 10 rows affected, error is <nil>   pkg=store request_id=rdu-infra-edge-03.infra-edge.lab.eng.rdu2.redhat.com/1kblckPxDu-000001         
```
master logger
```
2024/01/10 08:26:58 db.Find(): 10 rows affected, error is <nil>
```


hexfusion:
I know this is after the fact but in terms of structured loggers I was curious why we would not use ` zap`[1]?

[1] https://github.com/uber-go/zap?tab=readme-ov-file#performance

PR #51, opened 2024-01-11
Repository tester thread

avishayt:
Thread that periodically tests the connectivity to a git repository and
updates the status of the corresponding repository resource accordingly.

### DIFF
--- a/internal/monitors/repotester/repotester.go
+++ b/internal/monitors/repotester/repotester.go
@@ -1,0 +1,96 @@
+package repotester
+
+import (
+	"context"
+
+	api "github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/service"
+	"github.com/flightctl/flightctl/internal/store"
+	"github.com/flightctl/flightctl/internal/store/model"
+	"github.com/flightctl/flightctl/internal/util"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/reqid"
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/config"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+type API interface {
+	Test()
+}
+
+type RepoTester struct {
+	log       logrus.FieldLogger
+	db        *gorm.DB
+	repoStore service.RepositoryStoreInterface
+}
+
+func NewRepoTester(log logrus.FieldLogger, db *gorm.DB, store *store.Store) *RepoTester {
+	return &RepoTester{
+		log:       log,
+		db:        db,
+		repoStore: store.GetRepositoryStore(),
+	}
+}
+
+func (r *RepoTester) TestRepo() {
+	reqid.OverridePrefix("repotester")
+	requestID := reqid.NextRequestID()
+	ctx := context.WithValue(context.Background(), middleware.RequestIDKey, requestID)
+	log := log.WithReqIDFromCtx(ctx, r.log)
+
+	log.Info("Running RepoTester")
+
+	repositories, err := r.repoStore.ListAllRepositoriesInternal()
+	if err != nil {
+		log.Errorf("error fetching repositories: %s", err)
+		return
+	}
+
+	for _, repository := range repositories {
+		remote := git.NewRemote(memory.NewStorage(), &config.RemoteConfig{
+			Name:  *repository.Metadata.Name,
+			URLs:  []string{*repository.Spec.Repo},
+			Fetch: []config.RefSpec{"HEAD"},
+		})
+
+		_, err = remote.List(&git.ListOptions{
+			Auth: &http.BasicAuth{
+				Username: *repository.Spec.Username,
+				Password: *repository.Spec.Password,
+			},
+		})
+
+		r.setAccessCondition(log, repository, err)
+	}
+}
+
+func (r *RepoTester) setAccessCondition(log logrus.FieldLogger, repository model.InternalRepository, err error) {
+	var condition api.RepositoryCondition
+	if err == nil {
+		condition = api.RepositoryCondition{
+			Type:               api.Accessible,
+			LastTransitionTime: util.TimeStampStringPtr(),
+			Status:             api.True,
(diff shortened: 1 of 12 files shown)


mangelajo:
Question: why do we keep an internal version of the repositories in the DB, what's the purpose?

avishayt:
> Question: why do we keep an internal version of the repositories in the DB, what's the purpose?

What do you mean? Where?

internal/monitors/repotester/repotester.go
```diff
@@ -0,0 +1,86 @@
...
+
+func NewRepoTester(log logrus.FieldLogger, db *gorm.DB, store *store.Store) *RepoTester {
+	return &RepoTester{
+		log:       log,
+		db:        db,
+		repoStore: store.GetRepositoryStore(),
+	}
+}
+
+func (r *RepoTester) Test() {
```
filanov, line 37: where is it used? 
The name is a bit confusing because the standard for golang tests is a function starting with Test, but not a big deal

internal/monitors/repotester/repotester.go
```diff
@@ -0,0 +1,86 @@
...
+	reqid.OverridePrefix("repotester")
```
filanov, line 38: Are you sure that it depend on the package that import the `reqid`? wouldn't that affect all the loggers? 
We can change the basic request id to be something else if needed and hostname is not required. to log specific package you can add another field to the logger like `r.log = r.log.WithField("pkg", "repotester")`

internal/monitors/repotester/repotester.go
avishayt: I will change it to TestRepo.  This is the method that the thread calls - set in cmd/flightctl-server/main.go

internal/monitors/repotester/repotester.go
avishayt: I tested this by adding another thread.  The logs showed each thread with its own prefix and reqid.

mangelajo:
> > Question: why do we keep an internal version of the repositories in the DB, what's the purpose?
> 
> What do you mean? Where?

I mean here:
https://github.com/flightctl/flightctl/pull/51/files#diff-c41ede519aa3ece3639825a2e5656e5afa2b552bcf5d5774ac0df485fb3dcd23R78

We have Repository, and InternalRepository

Repository is exposed through the API, but what is the intent of InternalRepository and why do we need both?

avishayt:
> > > Question: why do we keep an internal version of the repositories in the DB, what's the purpose?
> > 
> > 
> > What do you mean? Where?
> 
> I mean here: https://github.com/flightctl/flightctl/pull/51/files#diff-c41ede519aa3ece3639825a2e5656e5afa2b552bcf5d5774ac0df485fb3dcd23R78
> 
> We have Repository, and InternalRepository
> 
> Repository is exposed through the API, but what is the intent of InternalRepository and why do we need both?

Repository, which is returner to users, doesn't contain the spec with credentials. The internal struct has the credentials for internal use of the service.

internal/monitors/repotester/repotester.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (r *RepoTester) TestRepo() {
+	reqid.OverridePrefix("repotester")
+	requestID := reqid.NextRequestID()
+	ctx := context.WithValue(context.Background(), middleware.RequestIDKey, requestID)
+	log := log.WithReqIDFromCtx(ctx, r.log)
+
+	log.Info("Running RepoTester")
+
+	repositories, err := r.repoStore.ListAllRepositoriesInternal()
```
filanov, line 45: it will probably not be a big number but consider using batches

internal/monitors/repotester/repotester.go
```diff
@@ -0,0 +1,86 @@
...
+
+		err := git.TestRepo(*repository.Spec.Repo, argoRepo.GetGitCreds(gitCredsStore), false, false, "")
+
+		var condition api.RepositoryCondition
+		if err == nil {
+			condition = api.RepositoryCondition{
+				Type:               "Accessible",
+				LastTransitionTime: util.TimeStampStringPtr(),
+				Status:             "True",
+				Reason:             util.StrToPtr("Accessible"),
```
filanov, line 68: consider using consts

internal/monitors/repotester/repotester.go
avishayt: Agree, I already put a TODO in the function definition

filanov:
beside the minor comments LGMT

internal/monitors/repotester/repotester.go
avishayt: Done, please take a look

mangelajo:
> beside the minor comments LGMT

See my comment about removing duplicity and just blanking out user/password (or password in the ListRepositories and GetRepository functions here:

https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L96
https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L70

avishayt:
> > beside the minor comments LGMT
> 
> See my comment about removing duplicity and just blanking out user/password (or password in the ListRepositories and GetRepository functions here:
> 
> https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L96 https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L70

What duplicity?  Do you suggest that ToApiResource() have all the details, then we process the return value again to remove those details?  It's inefficient.  Unless I misunderstood.

mangelajo:
> > > beside the minor comments LGMT
> > 
> > 
> > See my comment about removing duplicity and just blanking out user/password (or password in the ListRepositories and GetRepository functions here:
> > https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L96 https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L70
> 
> What duplicity? Do you suggest that ToApiResource() have all the details, then we process the return value again to remove those details? It's inefficient. Unless I misunderstood.

ToApiResource would be a better place, yes.

mangelajo:
> > > > beside the minor comments LGMT
> > > 
> > > 
> > > See my comment about removing duplicity and just blanking out user/password (or password in the ListRepositories and GetRepository functions here:
> > > https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L96 https://github.com/flightctl/flightctl/blob/main/internal/service/repository.go#L70
> > 
> > 
> > What duplicity? Do you suggest that ToApiResource() have all the details, then we process the return value again to remove those details? It's inefficient. Unless I misunderstood.
> 
> ToApiResource would be a better place, yes.

Oh no, but then you could not use the resource internally as is... because you'd miss the password.

PR #55, opened 2024-01-22
Fix lint errors, add unit tests to smoke test

avishayt:


mangelajo:
other than that, it looks good.

fzdarsky:
> @avishayt any idea of why the linter is complaining in CI with:
> 
> ```
> level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: buildir: failed to load package gpgme: could not load export data: no export data for \"github.com/proglottis/gpgme\"\n\n"
> ```
> 
> ?

The reason is that the agent's Podman controller uses container/storage packages with C bindings, so you need to add build tags to skips those. https://github.com/flightctl/flightctl/blob/733a8eed0a28bdc14ae4937742cfc6184e7d5d26/Makefile#L3.
We could add the tags as work-around, but really we should switch to `exec`ing out to Podman (and systemd, bootc, ...) and remove those package requirements instead.

.github/workflows/pr-smoke-testing.yaml
```diff
@@ -16,12 +16,16 @@ jobs:
...
         uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v4
 
       # The `run` key tells the job to execute a command on the runner. In this example, `npm ci` is used to install the npm software packages for the project.
       - name: Install dependencies
         run: |
           sudo apt-get update -y # fix broken repo cache
-          sudo apt install -y make golang-1.20 python3-pip
+          sudo apt install -y make golang-1.20 python3-pip libgpgme-dev
```
avishayt, line 26: Remove libgpgme-dev

mangelajo (approved):
LGTM, let's merge, and remove the podman dependencies so we can run the linter without the need for any build flags...

PR #56, opened 2024-01-23
Add device last checkin timestamp

avishayt:


mangelajo:
LGTM, I was going in my head over all the possibilities here:

1) In metadata
2) In the status conditions
3) Just in the status as you did it

I think 3 is what makes most sense, as ... 
  2 - is not part of a status change
  1 - is not really an update to the device itself, we just want to log the last heartbeat.

I am even wondering if at some point this will need to be moved somewhere else (not the database), for speed, but .. that's an optimization that simply would happen in the future if we needed it.

mangelajo (approved):

avishayt:
> I am even wondering if at some point this will need to be moved somewhere else (not the database), for speed, but .. that's an optimization that simply would happen in the future if we needed it.

What do you mean?  We're already writing the status to the DB, so it's just one more column.

PR #57, opened 2024-01-25
Use exec instead of importing podman and add lint to smoke test

avishayt:


internal/agent/controller/containers.go
```diff
@@ -41,54 +50,48 @@ func (c *ContainerController) FinalizeUpdate(r *api.Device) (bool, error) {
...
 	}
 
-	// Set the socket path based on the user ID.
-	var socketPath string
-	if currentUser.Uid == "0" {
-		socketPath = "unix:///run/podman/podman.sock"
-	} else {
-		socketPath = fmt.Sprintf("unix:///run/user/%s/podman/podman.sock", currentUser.Uid)
+	var containers PodmanList
+	klog.Infof("OUT: %s", out)
```
hexfusion, line 81: nit  :)

PR #60, opened 2024-01-29
Device updater

avishayt:
1. Track fleet generations
2. Device updater thread: Fairly simple thread, for now, that updates the target generation of all devices to the generation of their respective fleet.  The thread can later be extended to handle more complex rollouts.


internal/service/device.go
```diff
@@ -22,6 +23,8 @@ type DeviceStoreInterface interface {
 	UpdateDeviceStatus(ctx context.Context, orgId uuid.UUID, device *api.Device) (*api.Device, error)
 	DeleteDevices(ctx context.Context, orgId uuid.UUID) error
 	DeleteDevice(ctx context.Context, orgId uuid.UUID, name string) error
+	ListAllDevicesInternal(map[string]string) ([]model.Device, error)
```
hexfusion, line 26: nit: in general it feels the interface naming studders since we using a `DeviceStore` and naming it `Interface` is unnecessary. It also feels awkward to have an internal naming designation on a public-facing method of an interface. These things will involve rework in the future and it would be nice to break these patterns?

```
type DeviceStore interface {
  ....
  UpdateStatus
  DeleteAll
  // Get all Devices regardless of ownership. Used internally by the DeviceUpdater.
  ListAll
```

internal/store/fleet.go
```diff
@@ -111,28 +133,77 @@ func (s *FleetStore) CreateOrUpdateFleet(ctx context.Context, orgId uuid.UUID, r
...
-		if result.Error == gorm.ErrRecordNotFound {
-			created = true
+
+	err := s.db.Transaction(func(tx *gorm.DB) error {
+		existingRecord := model.Fleet{Resource: model.Resource{OrgID: fleet.OrgID, Name: fleet.Name}}
+		result := tx.First(&existingRecord)
+		if result.Error != nil {
+			if !errors.Is(result.Error, gorm.ErrRecordNotFound) {
+				return result.Error
+			} else {
```
hexfusion, line 153: nit: the if statement above returns do we need the else statement? Also in general nesting of if conditions should be minimized to improve readability and reduce complexity.

```golang
if result.Error != nil && !errors.Is(result.Error, gorm.ErrRecordNotFound) {
 // something specific
 }
 
if result.Error != nil {
   // general error case
  }

// happy path
 
 ```

File: internal/service/device.go (hunk 1 above)
Comment (avishayt) [reply internal/service/device.go:26]: I agree.  Want to submit a PR? :)
By "Internal", I mean that it's not used for handling API calls, but rather for calls from within the service.  The main difference is the lack of orgId.

File: internal/store/fleet.go (hunk 2 above)
Comment (avishayt) [reply internal/store/fleet.go:153]: True

Comment (avishayt) [reply internal/store/fleet.go]: Fixed

File: internal/service/device.go (hunk 1 above)
Comment (hexfusion) [reply internal/service/device.go:26]: Yeah not pointing fingers here of course :) but yeah I will take some time to outline an alternative path forward in a PR down the road hopefully soon and we can talk about it.

File: internal/store/fleet.go [hunk 3]
```diff
@@ -111,28 +133,77 @@ func (s *FleetStore) CreateOrUpdateFleet(ctx context.Context, orgId uuid.UUID, r
...
-	if result.Error != nil {
-		if result.Error == gorm.ErrRecordNotFound {
+
+	err := s.db.Transaction(func(tx *gorm.DB) error {
+		existingRecord := model.Fleet{Resource: model.Resource{OrgID: fleet.OrgID, Name: fleet.Name}}
+		result := tx.First(&existingRecord)
+		if result.Error != nil {
+			if !errors.Is(result.Error, gorm.ErrRecordNotFound) {
+				return result.Error
+			}
```
hexfusion, line 153:
```suggestion
// note on why not found is returned here
if result.Error != nil && !errors.Is(result.Error, gorm.ErrRecordNotFound) {
    return result.Error
}
if result.Error != nil {
...
```

internal/monitors/device-updater/device_updater.go
```diff
@@ -0,0 +1,83 @@
...
+func NewDeviceUpdater(log logrus.FieldLogger, db *gorm.DB, store *store.Store) *DeviceUpdater {
+	return &DeviceUpdater{
+		log:        log,
+		db:         db,
+		fleetStore: store.GetFleetStore(),
+		devStore:   store.GetDeviceStore(),
+	}
+}
+
+func (d *DeviceUpdater) UpdateDevices() {
```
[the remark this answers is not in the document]
avishayt: I'm sorry, but I don't understand the question.  I create a new ctx with a new request ID each time.  Do you mean we should use the same one each time, or the same one from main.go?

internal/monitors/device-updater/device_updater.go
```diff
@@ -0,0 +1,83 @@
...
+	log.Info("Running DeviceUpdater")
+
+	fleets, err := d.fleetStore.ListAllFleetsInternal()
+	if err != nil {
+		log.Errorf("Failed to list fleets: %w", err)
+		return
+	}
+
+	for _, fleet := range fleets {
+		if fleet.Spec == nil || fleet.Spec.Data.Template.Metadata == nil {
```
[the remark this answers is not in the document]
avishayt: Sure, done

internal/monitors/device-updater/device_updater.go
```diff
@@ -0,0 +1,83 @@
...
+			continue
+		}
+
+		devices, err := d.devStore.ListAllDevicesInternal(fleet.Spec.Data.Selector.MatchLabels)
+		if err != nil {
+			log.Errorf("Failed to list devices for fleet %s: %w", fleet.Name, err)
```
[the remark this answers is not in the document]
avishayt: That's fine.   devices should have 0 length and we would skip the loop below.

internal/service/device.go
avishayt: Great, thanks!

internal/store/device.go
```diff
@@ -102,6 +102,27 @@ func (s *DeviceStore) ListDevices(ctx context.Context, orgId uuid.UUID, listPara
 	return &apiDevicelist, result.Error
 }
 
+// Get all Devices regardless of ownership. Used internally by the DeviceUpdater.
+// TODO: Add pagination, perhaps via gorm scopes.
+func (s *DeviceStore) ListAllDevicesInternal(labels map[string]string) ([]model.Device, error) {
```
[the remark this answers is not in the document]
avishayt: The map is created in the context of a thread and isn't shared

internal/store/fleet.go
avishayt: Done

hexfusion (approved):

PR #62, opened 2024-01-30
Set conditions for device heartbeats

avishayt:


internal/monitors/device-updater/device_updater.go
```diff
@@ -76,8 +85,75 @@ func (d *DeviceUpdater) updateDeviceSpecAccordingToFleetTemplate(log logrus.Fiel
 	device.Generation = &targetGeneration
 	device.Spec = model.MakeJSONField(fleet.Spec.Data.Template.Spec)
 	log.Infof("Updating device %s to generation %d", device.Name, *device.Generation)
-	err := d.devStore.UpdateDeviceInternal(device)
-	if err != nil {
-		log.Errorf("Failed to update target generation for device %s (fleet %s): %w", device.Name, fleet.Name, err)
+}
+
+func (d *DeviceUpdater) updateDeviceHeartbeatConditions(log logrus.FieldLogger, device *model.Device, fleet *model.Fleet) {
+	if device.Status.Data.UpdatedAt == nil {
```
hexfusion, line 91: can we add a note why it's okay to bail here?

internal/util/util.go
```diff
@@ -59,6 +59,11 @@ func TimeStampStringPtr() *string {
 	return StrToPtr(time.Now().Format(time.RFC3339))
 }
 
+func ParseTimeStamp(t string) time.Time {
+	parsed, _ := time.Parse(time.RFC3339, t)
```
hexfusion, line 63: string to time can be fairly error-prone should we return errors so they can be logged?

internal/monitors/device-updater/device_updater.go
avishayt: I actually already check this above in line 55, removing it

internal/util/util.go
avishayt: I'm using it only for internally generated strings, so there shouldn't be any errors.  On the other hand, others might use this method for user inputs in the future.  Maybe I should rename it to something like ParseTimeStampIgnoreErrors?

internal/monitors/device-updater/device_updater.go
avishayt: Sure

internal/monitors/device-updater/device_updater.go
```diff
@@ -77,8 +83,76 @@ func (d *DeviceUpdater) updateDeviceSpecAccordingToFleetTemplate(log logrus.Fiel
...
+		return
+	}
+
+	warningTime := fleet.Spec.Data.DeviceConditions.HeartbeatElapsedTimeWarning
+	errorTime := fleet.Spec.Data.DeviceConditions.HeartbeatElapsedTimeError
+	now := util.TimeStampStringPtr()
+
+	if device.Status.Data.Conditions == nil {
+		device.Status.Data.Conditions = &[]api.DeviceCondition{
```
hexfusion, line 99: just so I understand in terms of ownership how are we handling who can write to status. Like this appears an external monitor is updating the status of a device vs the device itself? Or is the device making a call to update status and this is the result.

hexfusion (approved):

internal/util/util.go
hexfusion: how about  MustParseTime and panic on error?

internal/monitors/device-updater/device_updater.go
avishayt: The device calls the updatestatus API, but some conditions have to be updated by the service logic

internal/util/util.go
avishayt: Panic for the whole service sounds scary.  I'll just return the error :)

avishayt:
Replaced by #290

PR #65, opened 2024-02-02
internal: remove Interface from non generated interface names

hexfusion:
While `oapi-codegen` generates its interfaces with the `Interface` suffix this pattern isn't followed in idiomatic go. This PR renames all non codegen interfaces removing this `Interface` suffix which can be assumed by its type.


avishayt:
LGTM, note that all PRs to this repo must be signed

PR #66, opened 2024-02-05
Remove redundancy from store method names

avishayt:


internal/service/device.go
```diff
@@ -15,22 +15,22 @@ import (
...
-	ListAllDevicesInternal(map[string]string) ([]model.Device, error)
-	UpdateDeviceInternal(device *model.Device) error
+	Create(ctx context.Context, orgId uuid.UUID, device *api.Device) (*api.Device, error)
+	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.DeviceList, error)
+	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.Device, error)
+	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, device *api.Device) (*api.Device, bool, error)
+	UpdateStatus(ctx context.Context, orgId uuid.UUID, device *api.Device) (*api.Device, error)
+	DeleteAll(ctx context.Context, orgId uuid.UUID) error
+	Delete(ctx context.Context, orgId uuid.UUID, name string) error
+	ListIgnoreOrg(map[string]string) ([]model.Device, error)
```
hexfusion, line 25: this still feels like something that should be external to the interface. Can't we unify List to provide a way to expose a way to do something like this.  List(ctx, listParams. WithOrg(false)) maybe sth like

```
List(ctx context.Context, orgId uuid.UUID, listParams ListParams, opts listOpts...) (*api.DeviceList, error)
```

internal/service/enrollmentrequest.go
```diff
@@ -19,13 +19,13 @@ import (
...
-	CreateOrUpdateEnrollmentRequest(ctx context.Context, orgId uuid.UUID, enrollmentrequest *api.EnrollmentRequest) (*api.EnrollmentRequest, bool, error)
-	UpdateEnrollmentRequestStatus(ctx context.Context, orgId uuid.UUID, enrollmentrequest *api.EnrollmentRequest) (*api.EnrollmentRequest, error)
-	DeleteEnrollmentRequests(ctx context.Context, orgId uuid.UUID) error
-	DeleteEnrollmentRequest(ctx context.Context, orgId uuid.UUID, name string) error
+	Create(ctx context.Context, orgId uuid.UUID, req *api.EnrollmentRequest) (*api.EnrollmentRequest, error)
+	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.EnrollmentRequestList, error)
+	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.EnrollmentRequest, error)
+	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, enrollmentrequest *api.EnrollmentRequest) (*api.EnrollmentRequest, bool, error)
+	UpdateStatus(ctx context.Context, orgId uuid.UUID, enrollmentrequest *api.EnrollmentRequest) (*api.EnrollmentRequest, error)
+	DeleteAll(ctx context.Context, orgId uuid.UUID) error
```
hexfusion, line 27: Still not sure DeleteAll should be a dependency of the interface vs an implementation detail

internal/service/fleet.go
```diff
@@ -16,14 +16,14 @@ import (
...
-	DeleteFleet(ctx context.Context, orgId uuid.UUID, name string) error
-	ListAllFleetsInternal() ([]model.Fleet, error)
+	Create(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, error)
+	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.FleetList, error)
+	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.Fleet, error)
+	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, bool, error)
+	UpdateStatus(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, error)
+	DeleteAll(ctx context.Context, orgId uuid.UUID) error
+	Delete(ctx context.Context, orgId uuid.UUID, name string) error
+	ListIgnoreOrg() ([]model.Fleet, error)
```
hexfusion, line 26: same question with regards to unification with List

internal/service/repository.go
```diff
@@ -14,21 +14,21 @@ import (
...
-	DeleteRepository(ctx context.Context, orgId uuid.UUID, name string) error
-	UpdateRepositoryStatusInternal(repository *model.Repository) error
+	Create(ctx context.Context, orgId uuid.UUID, repository *api.Repository) (*api.Repository, error)
+	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.RepositoryList, error)
+	ListIgnoreOrg() ([]model.Repository, error)
+	DeleteAll(ctx context.Context, orgId uuid.UUID) error
+	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.Repository, error)
+	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, repository *api.Repository) (*api.Repository, bool, error)
+	Delete(ctx context.Context, orgId uuid.UUID, name string) error
+	UpdateStatusIgnoreOrg(repository *model.Repository) error
```
hexfusion, line 24: IgnoreOrg should be implementation detail exposed by Update.

internal/service/device.go
avishayt: One main difference is that this returns model.Device and not api.Device

PR #67, opened 2024-02-05
Fix gocyclo issues in CLI

avishayt:


cmd/flightctl/main.go
```diff
@@ -494,6 +337,100 @@ func printSingleResourceResponse(response interface{}, name string) error {
...
+		printDevicesTable(w, response.(*client.ListDevicesResponse))
+	case "enrollmentrequests":
+		printEnrollmentRequestsTable(w, response.(*client.ListEnrollmentRequestsResponse))
+	case "fleets":
+		printFleetsTable(w, response.(*client.ListFleetsResponse))
+	case "repositories":
+		printRepositoriesTable(w, response.(*client.ListRepositoriesResponse))
+	case "resourcesyncs":
+		printResourceSyncsTable(w, response.(*client.ListResourceSyncResponse))
+	}
```
hexfusion, line 372: is it an error if we don't match

hexfusion (commented):
thanks for doing this, just a few notes.

hexfusion (approved):

PR #68, opened 2024-02-07
MGMT-16697: Added ResourceSync monitor thread

sagidayan:
This PR adds a ResourceSync monitor thread:
- checks if repository resource exists (adds condition)
 - checks if git repository is accessible (adds condition)
 - checks if path in repo exists (adds condition)
 - Parse dir/file path to resources (adds condition)
 - if resource is a **fleet** apply (Only fleets are supported at the moment)
 - adds latest commit hash and path to the resourcesync status (if path changed or commit - sync should happen)
 - Added a db transaction to apply all fleets or none

In addition there is a commit that addresses RepoTester monitor. It attaches auth option only if user creds exists.


internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		if fleet.Metadata.Name == nil {
+			return fmt.Errorf("%s: decoding Fleet resource: missing field .metadata.name: %v", rsName, err)
+		} else {
+			fmt.Printf("%s: applying fleet/%s: ", rsName, *fleet.Metadata.Name)
+			_, _, err = r.store.GetFleetStore().CreateOrUpdate(context.Background(), orgID, &fleet)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("%s fleet/%s applyed successfully\n", rsName, *fleet.Metadata.Name)
```
avishayt, line 265:
```suggestion
			fmt.Printf("%s fleet/%s applied successfully\n", rsName, *fleet.Metadata.Name)
```

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+	// We now have a list of generic resources and resourcesync - we can attempt to apply and set conditions
+	for i := range syncLinks { // Check if new commit hash
+		link := syncLinks[i]
+		var err error
+		for _, genResource := range link.resources {
+			err = r.applyResource(genResource, &link.resourceSync)
+			if err != nil {
+				log.Errorf("resourcesync/%s: Failed to apply resource. Error: %s", link.resourceSync.Name, err.Error())
+				break
+			}
```
avishayt, line 154: Why don't we set a condition here?

avishayt (commented):
Great work!

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+func (r *ResourceSync) extractResourcesFromFile(orgId string, mfs billy.Filesystem, path string) ([]genericResource, error) {
+	genericResources := []genericResource{}
+
+	file, err := mfs.Open(path)
+	if err != nil {
+		// Failed to open file....
+		return nil, err
+	}
+	defer file.Close()
+	decoder := yamlutil.NewYAMLOrJSONDecoder(file, 100)
```
hexfusion, line 221: Just curious what was the reason for the 100 size buffer. Probably be good to add a test and or docs around this decoding.

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+func (r *ResourceSync) applyResource(resource genericResource, rs *model.ResourceSync) error {
+	kind, ok := resource["kind"].(string)
+	rsName := fmt.Sprintf("resourcesync/%s", rs.Name)
+	orgID := rs.OrgID
+	if !ok {
+		return fmt.Errorf("%s: skipping resource of unspecified kind: %v", rsName, resource)
+	}
+	buf, err := json.Marshal(resource)
+	if err != nil {
+		return fmt.Errorf("%s: skipping resource of kind %q: %v", rsName, kind, err)
```
hexfusion, line 247: good practice to wrap all errors this can be useful in testing etc.
```suggestion
		return fmt.Errorf("%s: skipping resource of kind %q: %w", rsName, kind, err)
```

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	switch kind {
+	case model.FleetKind:
+		var fleet api.Fleet
+		err := yamlutil.Unmarshal(buf, &fleet)
+		if err != nil {
+			return fmt.Errorf("%s: decoding Fleet resource: %v", rsName, err)
```
hexfusion, line 255: reminder to review all returned errors and add wrapping.

good practice to wrap errors this can be useful in testing etc.
```suggestion
			return fmt.Errorf("%s: decoding Fleet resource: %w", rsName, err)
```

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		if fleet.Metadata.Name == nil {
+			return fmt.Errorf("%s: decoding Fleet resource: missing field .metadata.name: %v", rsName, err)
+		} else {
```
hexfusion, line 259: in general it reads nicer to omit the else statement in cases where it is logically unessisary. IE the if statement returns here so else is assumed.

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+			fmt.Printf("%s: applying fleet/%s: ", rsName, *fleet.Metadata.Name)
+			_, _, err = r.store.GetFleetStore().CreateOrUpdate(context.Background(), orgID, &fleet)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("%s fleet/%s applied successfully\n", rsName, *fleet.Metadata.Name)
```
hexfusion, line 265: reminder to use logger.

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+
+	default:
+		return fmt.Errorf("%s: skipping resource of unkown/unsupported kind %q: %v", rsName, kind, resource)
```
hexfusion, line 269: :+1:

internal/monitors/resourcesync/resourcesync.go
```diff
@@ -0,0 +1,289 @@
...
+	return prevHash == nil || hash != *prevHash
+}
+
+func isValidFile(filename string) bool {
+	ext := ""
+	splits := strings.Split(filename, ".")
+	if len(splits) > 0 {
+		ext = splits[len(splits)-1]
+	}
+	return ext == "yaml" || ext == "yml" || ext == "json"
```
hexfusion, line 288: should ext be an enum to document the supported extensions?

hexfusion (commented):
Looks good mostly questions

internal/monitors/resourcesync/resourcesync.go
sagidayan: Correct. An oversight by me :)

cmd/flightctl-server/main.go
```diff
@@ -139,6 +140,12 @@ func main() {
 	deviceUpdaterThread.Start()
 	defer deviceUpdaterThread.Stop()
 
+	resourceSync := resourcesync.NewResourceSync(log, db, store)
+	resourceSyncThread := thread.New(
+		log.WithField("pkg", "resourcesync"), "Resource Sync", time.Duration(2*float64(time.Minute)), resourceSync.Poll)
```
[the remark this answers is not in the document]
sagidayan: Instead of a const, its a function now. We might want different times for each monitor.

internal/monitors/resourcesync/resourcesync.go
sagidayan: no reason, looked at the cli (cmd), did the same). Any suggestions? Its still WIP, so please do suggest. :)

hexfusion: It would be good to understand the size of files we expect to process. Broken down into 100 byte increments is this efficient? Hard to know without more details.

internal/store/fleet.go
```diff
@@ -217,7 +226,37 @@ func (s *FleetStore) CreateOrUpdate(ctx context.Context, orgId uuid.UUID, resour
 	return &updatedResource, created, nil
 }
 
+func (s *FleetStore) CreateOrUpdateMultiple(ctx context.Context, orgId uuid.UUID, resources ...*api.Fleet) error {
```
hexfusion, line 229: If the implementation is going to return on the first error could we just have a single method on the interface like this?

```
CreateOrUpdate(ctx context.Context, orgId uuid.UUID, fleets ...*api.Fleet) error

File: internal/store/fleet.go [hunk 12]
```diff
@@ -22,7 +22,9 @@ type Fleet interface {
 	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.FleetList, error)
 	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.Fleet, error)
 	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, bool, error)
+	CreateOrUpdateMultiple(ctx context.Context, orgId uuid.UUID, fleets ...*api.Fleet) error
 	UpdateStatus(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, error)
+	UpdateStatusMultiple(ctx context.Context, orgId uuid.UUID, fleets ...*api.Fleet) error
```
hexfusion, line 27: Why does this need to be a method on an interface vs a simple for loop against CreateOrUpdate or UpdateStatus?

sagidayan: This wraps all the DB calls in a transaction. its not a simple loop of a call in each iteration.

internal/monitors/resourcesync/resourcesync.go
sagidayan: Agree, we can discuss this in our next design meeting.

avishayt:
LGTM as long as  @hexfusion is happy as well

hexfusion (approved):
Looks good only thought is should we be adding TXN support to methods vs introducing `Multiple` variants to the interface. But that can be something to revisit. Thanks!

```
 UpdateStatus(ctx context.Context, orgId uuid.UUID, fleet ...*api.Fleet).

PR #69, opened 2024-02-07
Enabling packit for PRs

mangelajo:
Enable packit for PRs, merges and releases:

   commit merges to the main branch will be released in: https://copr.fedorainfracloud.org/coprs/g/redhat-et/flightctl-dev/ and can be enabled by using:

`dnf copr enable @redhat-et/flightctl-dev`

or importing the right repo: i.e.: https://copr.fedorainfracloud.org/coprs/g/redhat-et/flightctl-dev/repo/fedora-eln/group_redhat-et-flightctl-dev-fedora-eln.repo

New releases in github should be released in: https://copr.fedorainfracloud.org/coprs/g/redhat-et/flightctl/ and can be enabled by using:

`dnf copr enable @redhat-et/flightctl`

or importing the right repo: i.e.: https://copr.fedorainfracloud.org/coprs/g/redhat-et/flightctl/repo/fedora-eln/group_redhat-et-flightctl-fedora-eln.repo

Packit takes care of creating transient repositories for the PRs

### DIFF
--- a/packaging/Containerfile.fedora
+++ b/packaging/Containerfile.fedora
@@ -1,6 +1,6 @@
 FROM quay.io/centos-bootc/fedora-bootc:eln
 
-COPY rpmbuild/RPMS/x86_64/flightctl-agent-0.0.1-1.el9.x86_64.rpm /tmp/
+COPY bin/rpm/flightctl-agent-*rpm /tmp/
 
 COPY packaging/flightctl-custom-assets/flightctl_rsa.pub /usr/etc-system/root.keys
 RUN touch /etc/ssh/sshd_config.d/30-auth-system.conf; \
@@ -13,7 +13,7 @@ ADD packaging/flightctl-custom-assets/config.yaml /etc/flightctl/
 ADD packaging/flightctl-custom-assets/ca.crt /etc/flightctl
 ADD packaging/flightctl-custom-assets/client-enrollment.* /etc/flightctl/
 
-RUN rpm-ostree install -y /tmp/flightctl-agent-0.0.1-1.el9.x86_64.rpm
+RUN rpm-ostree install -y /tmp/flightctl-agent-*.rpm
 RUN ln -s /usr/lib/systemd/system/podman.socket /usr/lib/systemd/system/multi-user.target.wants/
 RUN ln -s /usr/lib/systemd/system/flightctl-agent.service /usr/lib/systemd/system/multi-user.target.wants/
 COPY packaging/flightctl-custom-assets/00-fedora.toml /usr/lib/bootc/install/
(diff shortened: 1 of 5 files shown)


packaging/rpm/flightctl-agent.spec
```diff
@@ -30,5 +44,7 @@ cp flightctl-agent.service %{buildroot}/usr/lib/systemd/system
 /usr/lib/systemd/system/flightctl-agent.service
 
 %changelog
+* Wed Feb 7 2024 Miguel Angel Ajo Pelayo <majopela@redhat.com> - 0.0.1-2
+- Initial RPM building via packit for the flightctl agent
```
hexfusion, line 48: :+1:

hexfusion (approved):
Looks great thanks!

packaging/rpm/flightctl-agent.spec
```diff
@@ -2,33 +2,50 @@
...
 %build
 
+# if this is a buggy version of go we need to set GOPROXY as workaround
+# see https://github.com/golang/go/issues/61928
+GOENVFILE=$(go env GOROOT)/go.env
+if [[ ! -f "{$GOENVFILE}" ]]; then
+    export GOPROXY='https://proxy.golang.org,direct'
+fi
+
+make build
```
oglok, line 33: is this a macro from Packit or something like that? because you have removed the build target from the Makefile

mangelajo: I didn't remove the build target, it's still there :-D

https://github.com/flightctl/flightctl/pull/69/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R59

What I did was removing the dependency from the rpm call. Instead what we do, to be able to us builders from source code, is calling make build inside the rpm spec, so the srpm is packaged with the whole sourcecode (not the binaries), and the .spec file... then the builders will simply compile the go for each architecture.

packaging/Containerfile.fedora
```diff
@@ -1,6 +1,6 @@
 FROM quay.io/centos-bootc/fedora-bootc:eln
 
-COPY rpmbuild/RPMS/x86_64/flightctl-agent-0.0.1-1.el9.x86_64.rpm /tmp/
+COPY x86_64/flightctl-agent-*rpm /tmp/
```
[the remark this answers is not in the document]
mangelajo: True, I was removing the rpmbuild/RPMS part... I wonder if I can call $(uname -m) from inside the docker file? I hope..

packaging/rpm/flightctl-agent.spec
oglok: ohhh true! hahah, I saw a red line on build, and boom! my bad... I need more coffee...

PR #70, opened 2024-02-08
Device owner ref

avishayt:
Commit 1:
    Add Owner to ObjectMeta

Commit 2:
    When a change in a fleet's selector or a device's labels causes the
    device to:
    1. Be owned by a different fleet
    2. Be owned by no fleet
    3. Be owned by more than one fleet (error condition)
    
    Also fixed the device updater because although it's correct for it to
    list all fleets regardless of orgID, it should handle devices according
    to the fleet's orgID.

Commit 3:
    When deleting one or all fleets, unset the owner ref of the relevant
    devices.


internal/monitors/device-updater/device_updater.go
```diff
@@ -47,42 +50,119 @@ func (d *DeviceUpdater) UpdateDevices() {
...
+		if fleets[fleetIndex].Spec == nil || fleets[fleetIndex].Spec.Data.Template.Metadata == nil {
 			continue
 		}
 
-		devices, err := d.devStore.ListIgnoreOrg(fleet.Spec.Data.Selector.MatchLabels)
+		listParams := service.ListParams{
+			Labels: fleets[fleetIndex].Spec.Data.Selector.MatchLabels,
+			Limit:  service.MaxRecordsPerListRequest, //TODO: Handle paging
+		}
+		devices, err := d.devStore.List(ctx, orgID, listParams)
```
mangelajo, line 64: I would move this work down to the database level for efficiency, allowing us to benefit from DB transactionality, etc.

1) Validate that we are only moving devices that don't belong to the fleet already or are orphaned:
SELECT device WHERE match the tags and owner!="the-new-owner-name" to identify devices already belonging to other fleets.

2) If (1) went OK

 UPDATE device SET owner="the-new-owner-name" WHERE devices match the tags AND owner=""

mangelajo: i.e.

d.devStore.UpdateOwnerRef(fleets[fleetIndex].Spec.Data.Selector.MatchLabels, fleets[fleetIndex].Name)

this would be updating the OwnerField of the model.


Then, when the model is de/serialized from/to DB, because somebody is using it, we could reconstruct the owner ref field in the metadata.

mangelajo (changes requested):

internal/monitors/device-updater/device_updater.go
```diff
@@ -24,12 +27,12 @@ type DeviceUpdater struct {
 	devStore   service.DeviceStore
 }
 
-func NewDeviceUpdater(log logrus.FieldLogger, db *gorm.DB, store *store.Store) *DeviceUpdater {
+func NewDeviceUpdater(log logrus.FieldLogger, db *gorm.DB, stores *store.Store) *DeviceUpdater {
 	return &DeviceUpdater{
 		log:        log,
 		db:         db,
```
hexfusion, line 33: just a general question but if we are abstracting the db as store do we also need db here? I don't see it used.

internal/monitors/device-updater/device_updater.go
```diff
@@ -24,12 +27,12 @@ type DeviceUpdater struct {
...
-		fleetStore: store.GetFleetStore(),
-		devStore:   store.GetDeviceStore(),
+		fleetStore: stores.GetFleetStore(),
+		devStore:   stores.GetDeviceStore(),
```
hexfusion, line 35: if the store abstracts the DB then should this remain singular?

```suggestion
		fleetStore: store.GetFleet(),
		devStore:   store.GetDevice(),
```

internal/monitors/device-updater/device_updater.go
```diff
@@ -47,42 +50,119 @@ func (d *DeviceUpdater) UpdateDevices() {
 		return
 	}
 
-	for _, fleet := range fleets {
+	for fleetIndex := range fleets {
```
hexfusion, line 53: I think its easier to reason about the fleet directly why use the index directly?

internal/monitors/device-updater/device_updater.go
```diff
@@ -47,42 +50,119 @@ func (d *DeviceUpdater) UpdateDevices() {
...
+			dev := devices.Items[devIndex]
+			if dev.Metadata.Owner == nil || *dev.Metadata.Owner == "" {
+				dev.Metadata.Owner = util.SetResourceOwner(model.FleetKind, fleets[fleetIndex].Name)
+				log.Infof("Setting previously unset fleet of device %s to %s", *dev.Metadata.Name, *dev.Metadata.Owner)
+			} else if *dev.Metadata.Owner != fleets[fleetIndex].Name {
+				// The fleet for this device changed
+				err = d.handleOwningFleetChanged(ctx, log, &dev, &fleets[fleetIndex])
+				if err != nil {
+					log.Errorf("failed to handle fleet update for device %s: %v", *dev.Metadata.Name, err)
+				}
```
hexfusion, line 79: Can you help me understand why UpdateDevices() doesn't return an error. You could aggregate errors but in terms of testability we would need to parse logs to validate failure?

internal/monitors/device-updater/device_updater.go
```diff
@@ -47,42 +50,119 @@ func (d *DeviceUpdater) UpdateDevices() {
...
+check that the current owner fleet doesn't still match the device before we
+update the device's owner, because that means the descriptors of two fleets
+match the device.
+*/
+func (d *DeviceUpdater) handleOwningFleetChanged(ctx context.Context, log logrus.FieldLogger, device *api.Device, fleet *model.Fleet) error {
+	// "fleet" is potentially the new owner of "device" because, but we first need
+	// to make sure that the label selectors of both the current fleet and the new
+	// fleet aren't a match for this device.
+	currentOwningFleet, err := d.fleetStore.Get(ctx, fleet.Resource.OrgID, *device.Metadata.Owner)
+	if err != nil && err != gorm.ErrRecordNotFound {
```
hexfusion, line 112:
```suggestion
if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
```

internal/monitors/device-updater/device_updater.go
```diff
@@ -47,42 +50,119 @@ func (d *DeviceUpdater) UpdateDevices() {
...
+		return err
+	}
+	if currentOwningFleet != nil && util.LabelsMatchLabelSelector(*device.Metadata.Labels, currentOwningFleet.Spec.Selector.MatchLabels) {
+		d.log.Warnf("Device %s matches more than one fleet (%s and %s)", *device.Metadata.Name, *device.Metadata.Owner, fleet.Name)
+		// Oh no, a device with 2 owners! How do we indicate that?
```
hexfusion, line 117: I think this is an error vs warn and should return?

internal/monitors/device-updater/device_updater.go
```diff
@@ -47,42 +50,119 @@ func (d *DeviceUpdater) UpdateDevices() {
...
+		return fmt.Errorf("failed to list devices that no longer belong to fleet %s: %v", fleet.Name, err)
+	}
+
+	for devIndex := range devices.Items {
+		dev := devices.Items[devIndex]
+		d.log.Infof("Unsetting fleet for device %s (was %s)", *dev.Metadata.Name, *dev.Metadata.Owner)
+		dev.Metadata.Owner = util.StrToPtr("")
+		_, _, err := d.devStore.CreateOrUpdate(ctx, fleet.OrgID, &dev)
+		if err != nil {
+			return fmt.Errorf("failed to update device without fleet %s: %v", *dev.Metadata.Name, err)
```
hexfusion, line 148: wrapping errors can help with better testing and fine grained error control.
```suggestion
			return fmt.Errorf("failed to update device without fleet %s: %w", *dev.Metadata.Name, err)
```

internal/monitors/device-updater/device_updater_test.go
```diff
@@ -91,12 +54,12 @@ var _ = Describe("DeviceUpdater", func() {
...
 		It("Update devices good flow", func() {
-			fleet := createFleet(ctx, fleetStore, orgId)
-			fleet, err := fleetStore.Get(ctx, orgId, *fleet.Metadata.Name)
+			store.CreateTestFleets(1, ctx, fleetStore, orgId, "myfleet", true)
+			fleet, err := fleetStore.Get(ctx, orgId, "myfleet-1")
 			Expect(err).ToNot(HaveOccurred())
 			Expect(*fleet.Metadata.Generation).To(Equal(int64(1)))
 			Expect(*fleet.Spec.Template.Metadata.Generation).To(Equal(int64(1)))
-			createDevices(numDevices, ctx, deviceStore, orgId)
+			store.CreateTestDevices(numDevices, ctx, deviceStore, orgId, true)
```
hexfusion, line 62: ideally, this logic would live in a test harness/helper package.

hexfusion (commented):
First pass looks good few small suggestions and questions

internal/monitors/device-updater/device_updater_test.go
hexfusion: This can totally be a TODO

PR #71, opened 2024-02-08
Splitting tests into their own jobs

mangelajo:


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,30 @@
...
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+  pull-requests: read
+jobs:
+  # This line defines a job with the ID `check-links` that is stored within the `jobs` key.
+  unit-tests:
```
fzdarsky, line 14: nit: probably want to update the job name

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,30 @@
+name: "Lint"
```
fzdarsky, line 1: Do we just lint or eventually run more "code quality" checks?

mangelajo: I think it's only linting so far, but we could aggregate those under a common flag eventually :-)

avishayt (approved):

Issue #72, opened 2024-02-08
Cyclical dependencies.


hexfusion:
When trying to break out the server logic for easier testing, I noticed a few things that I would like to track.

internal/server is the location for the oapi-codegen generated files. But since we would like to break out server logic for easier testing out of flightctl-server these generated files should probably find a new home. The result is that there internal/service depends on internal/server.

Also the internal/store depends on internal/service so any reference to store in internal/server would be a cyclical dependency.

Not a top priority but it would be good to resolve sooner than later.

Logically
- Service should not depend on the Server
- Store should not depend on Service
- The Agent should have its own config not shared with Database and Server.

Proposal
- Move all storage-related interfaces to internal/store if we want to keep the interface.
- The generated code should live in a new internal directory perhaps api/
- abstract out Server from cmd/flightctl-server/main.go into internal/server following the existing pattern for monitor and agent etc. The main reason is reusability in tests.
- break out Agent Config.

PR #73, opened 2024-02-08
internal: move generated files into internal/api

hexfusion:
This PR moved generated files from internal/{server,client} as part of #72


mangelajo:
Makes sense!

The linter is complaining, not sure what it means, is it complaining about the import order may be? 
```
internal/service/handler.go:8: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
	"github.com/flightctl/flightctl/internal/crypto"
internal/service/handler.go:9: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
	"github.com/flightctl/flightctl/internal/api/server"
cmd/flightctl-server/main.go:14: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
```

fzdarsky:
> The linter is complaining, not sure what it means, is it complaining about the import order may be?

That's it, yes. Running `gci write --skip-generated .` would fix the order.

mangelajo:
@hexfusion looks great, only two things:
1) Repo is configured to require signing of commits (with -S / gpg...)
2) Your PR is marked as draft. Is it really draft?

mangelajo (approved):
We need GPG signatures of the commits ':D

PR #74, opened 2024-02-08
internal: breakout Server logic from cmd/flightctl-server

hexfusion:
This PR creates a new server object and breaks out the actual server generation logic to assist with testing. 

Depends on #73 
Related to #72


mangelajo:
+1

Again the gci linter comment which we need to look into.

hexfusion:
will rebase and resign in order as they merge thanks for the reviews.

PR #75, opened 2024-02-08
internal: decouple service from store

hexfusion:
This PR attempts to decouple the stores dependency on the service. By having the store depend on the service we expose ourselves to circular dependencies. 

This PR also removes the agents config from the server config.

Logically the service should consume and expose the store.


mangelajo:
Looks good, keep in mind that we were considering to move the agent at some point into it's own module to limit the dependency graph.

hexfusion:
> Looks good, keep in mind that we were considering to move the agent at some point into it's own module to limit the dependency graph.

makes sense

internal/store/enrollmentrequest.go
```diff
@@ -7,23 +7,33 @@ import (
...
+	Delete(ctx context.Context, orgId uuid.UUID, name string) error
+	InitialMigration() error
+}
+
 type EnrollmentRequestStore struct {
 	db  *gorm.DB
 	log logrus.FieldLogger
 }
 
 // Make sure we conform to EnrollmentRequestStore interface
```
hexfusion, line 33: nit

internal/store/repository.go
```diff
@@ -7,23 +7,34 @@ import (
...
+	UpdateStatusIgnoreOrg(repository *model.Repository) error
+	InitialMigration() error
+}
+
 type RepositoryStore struct {
 	db  *gorm.DB
 	log logrus.FieldLogger
 }
 
 // Make sure we conform to RepositoryStore interface
```
hexfusion, line 34: nit

internal/store/resourcesync.go
```diff
@@ -7,23 +7,34 @@ import (
...
+	UpdateStatusIgnoreOrg(resourceSync *model.ResourceSync) error
+	InitialMigration() error
+}
+
 type ResourceSyncStore struct {
 	db  *gorm.DB
 	log logrus.FieldLogger
 }
 
 // Make sure we conform to ResourceSyncStore interface
```
hexfusion, line 34: nit

PR #81, opened 2024-02-12
Add API client config file

fzdarsky:
Adds a config file encapsulates API client information (server URL, CA bundle, user auth), writes that file to `~/.flightctl/client.yaml` when the server starts, updates the `flightctl` CLI client to use that file.

### DIFF
--- a/cmd/flightctl-server/main.go
+++ b/cmd/flightctl-server/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"path/filepath"
 
+	"github.com/flightctl/flightctl/internal/client"
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/crypto"
 	"github.com/flightctl/flightctl/internal/server"
@@ -44,9 +45,15 @@ func main() {
 	if err != nil {
 		log.Fatalf("ensuring server cert: %v", err)
 	}
-	_, _, err = ca.EnsureClientCertificate(certFile(clientBootstrapCertName), keyFile(clientBootstrapCertName), clientBootstrapCertName, clientBootStrapValidityDays)
+	clientCert, _, err := ca.EnsureClientCertificate(certFile(clientBootstrapCertName), keyFile(clientBootstrapCertName), clientBootstrapCertName, clientBootStrapValidityDays)
 	if err != nil {
 		log.Fatalf("ensuring bootstrap client cert: %v", err)
+	}
+
+	// also write out a client config file
+	err = client.WriteConfig(config.ClientConfigFile(), cfg.Service.BaseUrl, "", ca.Config, clientCert)
+	if err != nil {
+		log.Fatalf("writing client config: %v", err)
 	}
 
 	log.Println("Initializing data store")
--- a/cmd/flightctl/main.go
+++ b/cmd/flightctl/main.go
@@ -17,12 +17,12 @@ import (
 	"text/tabwriter"
 
 	api "github.com/flightctl/flightctl/api/v1alpha1"
-	"github.com/flightctl/flightctl/internal/api/client"
-	"github.com/flightctl/flightctl/internal/config"
-	"github.com/flightctl/flightctl/internal/crypto"
+	apiclient "github.com/flightctl/flightctl/internal/api/client"
+	"github.com/flightctl/flightctl/internal/client"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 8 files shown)


internal/agent/config.go
```diff
@@ -7,7 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/flightctl/flightctl/internal/util"
-	"gopkg.in/yaml.v2"
+	"sigs.k8s.io/yaml"
```
hexfusion, line 10: :+1:

internal/api/client/config.go
```diff
@@ -0,0 +1,127 @@
...
+	"net/url"
+	"os"
+
+	"github.com/flightctl/flightctl/internal/crypto"
+	oscrypto "github.com/openshift/library-go/pkg/crypto"
+	"sigs.k8s.io/yaml"
+)
+
+// Config holds the information needed to connect to a FlightCtl API server
+type Config struct {
```
mangelajo, line 17: The structure makes sense to me, we talked about trying to unify the cmdline tool config and agent config, we could do that eventually.

This config could be easily extended (AuthInfo section) to user/address/token (o whatever makes sense...) of a keycloak server for authentication when necessary,

mangelajo (approved):
We only need to figure out how to get the unit tests running properly.

cmd/flightctl-server/main.go
```diff
@@ -44,11 +45,17 @@ func main() {
 	if err != nil {
 		log.Fatalf("ensuring server cert: %v", err)
 	}
-	_, _, err = ca.EnsureClientCertificate(certFile(clientBootstrapCertName), keyFile(clientBootstrapCertName), clientBootstrapCertName, clientBootStrapValidityDays)
+	clientCert, _, err := ca.EnsureClientCertificate(certFile(clientBootstrapCertName), keyFile(clientBootstrapCertName), clientBootstrapCertName, clientBootStrapValidityDays)
```
hexfusion, line 48: Would it make sense to move  keyFile etc funcs into the config? I would expect an explicit path passed by the user to the flag or a generated default using these funcs. Seems like once we are in server, agent etc the paths should be set by config already.

cmd/flightctl/main.go
```diff
@@ -31,20 +31,24 @@ const (
...
 	}
 	resourceNameRegex = regexp.MustCompile(`^[a-zA-Z0-9\-]+$`)
-	serverUrl         = "https://localhost:3333"
 	fileExtensions    = []string{".json", ".yaml", ".yml"}
 	inputExtensions   = append(fileExtensions, "stdin")
 	legalOutputTypes  = []string{yamlFormat}
 )
 
+func init() {
+	clientConfigFile = filepath.Join(homedir.HomeDir(), ".flightctl", "client.yaml")
```
hexfusion, line 49: question: would we want to allow the config path to also be defined by the user?

internal/client/config_test.go
```diff
@@ -0,0 +1,92 @@
...
+	caCertValidityDays          = 365 * 10
+	serverCertValidityDays      = 365 * 1
+	clientBootStrapValidityDays = 365 * 1
+	signerCertName              = "ca"
+	clientBootstrapCertName     = "client-enrollment"
+)
+
+func TestClientConfig(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
```
hexfusion, line 25: nit: just need require here its not common to use both but I see why you did. require has all of the assertion methods that assert does.

hexfusion (approved):
lgtm just a  few notes

internal/client/config_test.go
fzdarsky: Changed that to use `require` only. Reasons I sprinkled in `assert` were that (unlike `require`) it doesn't immediately terminate the test, so provides more useful feedback in the same run, and it can be used in conditionals.

cmd/flightctl/main.go
fzdarsky: We definitely should. Will add this in a follow-up PR.

cmd/flightctl-server/main.go
fzdarsky: Makes sense, yes. I'll create a separate PR that'll allow to override the default config path and makes key and cert paths part of the config object. Would a fixed path relative to the config path be ok or do we need them to be configurable, too?

mangelajo (approved):
Looks great, let's merge and continue improving!

PR #82, opened 2024-02-12
internal/server: ensure CA passed to handler

hexfusion:
Unfortunately, it seems I had a bad rebase and introduced a bug as part of #74. Apologies.

This PR returns the original ordering of events and resolves the missing CA to the NewServiceHandler.

### DIFF
--- a/cmd/flightctl-server/main.go
+++ b/cmd/flightctl-server/main.go
@@ -21,7 +21,6 @@ const (
 
 func main() {
 	log := log.InitLogs()
-
 	log.Println("Starting device management service")
 	defer log.Println("Device management service stopped")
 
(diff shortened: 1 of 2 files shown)


cmd/flightctl-server/main.go
```diff
@@ -21,54 +21,51 @@ const (
...
+	if err != nil {
+		log.Fatalf("ensuring bootstrap client cert: %v", err)
+	}
 
 	log.Println("Initializing data store")
 	db, err := store.InitDB(cfg)
 	if err != nil {
 		log.Fatalf("initializing data store: %v", err)
 	}
-	defer store.CloseDB(db)
```
hexfusion, line 45: will return this in a separate PR for conversation.

PR #84, opened 2024-02-13
Create task infra based on channels

avishayt:
The task infra design:
- A goroutine for each task
- The task is notified when a resource (or part of it) is updated via channel

This is a temporary solution until we decide on a task tool. This task infra does not handle retries and tasks are lost when the server is terminated.

Also translated the "device updater" periodic task to use this infra. The new task is called "fleet rollout", and is triggered whenever a fleet's template is updated. This doesn't handle devices moving from one fleet to another - that will be handled in a separately once the service tracks what fleet each device belongs to.


internal/store/fleet.go
```diff
@@ -18,11 +18,11 @@ import (
...
 
 type Fleet interface {
-	Create(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, error)
+	Create(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet, callback FleetStoreCallback) (*api.Fleet, error)
 	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.FleetList, error)
 	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.Fleet, error)
-	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet) (*api.Fleet, bool, error)
-	CreateOrUpdateMultiple(ctx context.Context, orgId uuid.UUID, fleets ...*api.Fleet) error
+	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet, callback FleetStoreCallback) (*api.Fleet, bool, error)
+	CreateOrUpdateMultiple(ctx context.Context, orgId uuid.UUID, callback FleetStoreCallback, fleets ...*api.Fleet) error
```
hexfusion, line 25: in love with the callbacks here very flexible

internal/store/device.go
```diff
@@ -22,8 +22,7 @@ type Device interface {
 	UpdateStatus(ctx context.Context, orgId uuid.UUID, device *api.Device) (*api.Device, error)
 	DeleteAll(ctx context.Context, orgId uuid.UUID) error
 	Delete(ctx context.Context, orgId uuid.UUID, name string) error
-	ListIgnoreOrg(map[string]string) ([]model.Device, error)
-	UpdateIgnoreOrg(device *model.Device) error
```
hexfusion, line 26: :heart:

internal/server/server.go
```diff
@@ -104,32 +104,10 @@ func (s *Server) Run() error {
...
-		s.log.WithField("pkg", "device-updater"), "Device updater", threadIntervalMinute(2), deviceUpdater.UpdateDevices)
-	deviceUpdaterThread.Start()
-	defer deviceUpdaterThread.Stop()
-
-	resourceSync := resourcesync.NewResourceSync(s.log, s.store)
-	resourceSyncThread := thread.New(
-		s.log.WithField("pkg", "resourcesync"), "ResourceSync", threadIntervalMinute(2), resourceSync.Poll)
-	resourceSyncThread.Start()
-	defer resourceSyncThread.Stop()
-
```
hexfusion, line 124: :tada:

hexfusion (commented):
first pass this is great!

cmd/flightctl-server/main.go
```diff
@@ -62,12 +63,17 @@ func main() {
 		log.Fatalf("running initial migration: %v", err)
 	}
 
+	log.Println("Initializing async jobs")
+	taskManager := tasks.Init(log, store)
+	taskManager.Start()
+	defer taskManager.Stop()
+
```
hexfusion, line 70: nit: would love this to live in Server if possible. It simplifies tests to not need to create the taskManager since server already has its deps.

internal/tasks/manager.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (t TaskManager) SubmitTask(taskName string, resource ResourceReference) {
+	t.channels[taskName] <- resource
+}
+
+func (t TaskManager) GetTask(taskName string) ResourceReference {
+	return <-t.channels[taskName]
+}
+
+func (t TaskManager) FleetTemplateRolloutCallback(orgId uuid.UUID, name *string, templateUpdated bool) {
```
hexfusion, line 81: sweet

internal/tasks/manager.go
```diff
@@ -0,0 +1,87 @@
...
+		t.log.WithField("pkg", "resourcesync"), "ResourceSync", threadIntervalMinute(2), resourceSync.Poll)
+	resourceSyncThread.Start()
+	defer resourceSyncThread.Stop()
+}
+
+func (t TaskManager) Stop() {
+	for c := range t.channels {
+		close(t.channels[c])
+	}
+	t.cancelFunc()
```
hexfusion, line 70: I think you want this order.
- tear down all go routines capable of writing to the channel with cancelFunc()
- close only when you are sure no one can write.

This would panic if stop was called twice so you should doc that or use something like sync.Once https://pkg.go.dev/sync#Once

internal/tasks/manager.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx        context.Context
+	cancelFunc context.CancelFunc
+	channels   map[string]chan ResourceReference
+	store      store.Store
+}
+
+const (
+	FleetTemplateRollout = "fleet-template-rollout"
+
+	ChannelSize = 20
```
hexfusion, line 31: buffered channel is great but remember every op in a txn will be sent to the channel so this might need to be bigger. if the callback can be blocked by write to channel we are blocking the store I believe. TGhis might need to be configurable. depending on scale or a larger default.

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,87 @@
...
+	api "github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/store"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/reqid"
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+)
+
+func FleetRollouts(taskManager TaskManager) {
```
hexfusion, line 15: lets add some docs so we can see when this goroutine closes.

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,87 @@
...
+	reqid.OverridePrefix("fleet-rollout")
+
+	for {
+		select {
+		case fleetRef := <-taskManager.channels[FleetTemplateRollout]:
+			rolloutFleet(taskManager.log, taskManager.store.Fleet(), taskManager.store.Device(), fleetRef)
+		case <-taskManager.ctx.Done():
+			return
```
hexfusion, line 23: nit: add logs telling us why we are returning

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,87 @@
...
+	requestID := reqid.NextRequestID()
+	ctx := context.WithValue(context.Background(), middleware.RequestIDKey, requestID)
+	log := log.WithReqIDFromCtx(ctx, baseLog)
+
+	log.Infof("Rolling out fleet %s/%s", fleetRef.OrgID, fleetRef.Name)
+
+	fleet, err := fleetStore.Get(ctx, fleetRef.OrgID, fleetRef.Name)
+	if err != nil {
+		log.Errorf("failed to get fleet: %v", err)
+		return
```
hexfusion, line 38: lets return the errors and log above.
```
err := rolloutFleet(taskManager.log, taskManager.store.Fleet(), taskManager.store.Device(), fleetRef)
if err != nil {
    log.Errorf("rollout failed: %v", err)
}
```

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,87 @@
...
+			// TODO: Retry when we have a mechanism that allows it
+			return
+		}
+
+		for devIndex := range devices.Items {
+			device := &devices.Items[devIndex]
+			err = updateDeviceSpecAccordingToFleetTemplate(ctx, log, devStore, fleetRef.OrgID, device, fleet)
+			if err != nil {
+				log.Errorf("failed to update target generation for device %s (fleet %s): %v", *device.Metadata.Name, *fleet.Metadata.Name, err)
+				failureCount = failureCount + 1
```
hexfusion, line 61:
```suggestion
				failureCount++
```

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+		}
+
+		if devices.Metadata.Continue == nil {
+			break
+		}
+	}
+
+	if failureCount != 0 {
+		// TODO: Retry when we have a mechanism that allows it
```
hexfusion, line 71: :+1:

internal/tasks/manager.go
avishayt: Done

internal/server/server.go
```diff
@@ -34,11 +31,12 @@ const (
...
-	cfg       *config.Config
-	store     store.Store
-	tlsConfig *tls.Config
-	ca        *crypto.CA
+	log         logrus.FieldLogger
+	cfg         *config.Config
+	store       store.Store
+	tlsConfig   *tls.Config
+	ca          *crypto.CA
+	taskManager tasks.TaskManager
```
hexfusion, line 39: nit: drop

internal/server/server.go
```diff
@@ -65,6 +63,10 @@ func (s *Server) Run() error {
 	s.log.Println("Initializing config providers")
 	_ = git.NewGitConfigProvider(cache, cacheExpirationTime, cacheExpirationTime)
 
+	s.log.Println("Initializing async jobs")
+	s.taskManager = tasks.Init(s.log, s.store)
```
hexfusion, line 67: minor: since its local no need to add to server 
```suggestion
taskManager := tasks.Init(s.log, s.store)
```

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,93 @@
...
+/* Wait to be notified via channel about fleet template updates, exit upon ctx.Done() */
+func FleetRollouts(taskManager TaskManager) {
+	reqid.OverridePrefix("fleet-rollout")
+
+	for {
+		select {
+		case fleetRef := <-taskManager.channels[FleetTemplateRollout]:
+			err := rolloutFleet(taskManager.log, taskManager.store.Fleet(), taskManager.store.Device(), fleetRef)
+			if err != nil {
+				taskManager.log.Errorf("failed rolling out fleet %s/%s: %v", fleetRef.OrgID, fleetRef.Name, err)
```
hexfusion, line 25: :+1:

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,93 @@
...
+	log.Infof("Rolling out fleet %s/%s", fleetRef.OrgID, fleetRef.Name)
+
+	fleet, err := fleetStore.Get(ctx, fleetRef.OrgID, fleetRef.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get fleet: %w", err)
+	}
+
+	// If there is no template set in the fleet, then there is nothing to sync to the devices
+	if fleet.Spec.Template.Metadata == nil {
+		log.Warn("fleet does not have a template")
```
hexfusion, line 48: this is pretty much expected then right? I mean it would be common. Maybe warn is too harsh. debug?

internal/tasks/fleet_rollout.go
```diff
@@ -0,0 +1,93 @@
...
+
+	failureCount := 0
+	for {
+		devices, err := devStore.List(ctx, fleetRef.OrgID, store.ListParams{Labels: fleet.Spec.Selector.MatchLabels, Limit: ItemsPerPage})
+		if err != nil {
+			// TODO: Retry when we have a mechanism that allows it
+			return fmt.Errorf("failed fetching devices: %w", err)
+		}
+
+		for devIndex := range devices.Items {
```
hexfusion, line 60:
```suggestion
for _, device := range devices.Items {
    err = updateDeviceSpecAccordingToFleetTemplate(ctx, log, devStore, fleetRef.OrgID, &device, fleet)
```

internal/tasks/manager.go
```diff
@@ -0,0 +1,95 @@
...
+	resourceSyncThread := thread.New(
+		t.log.WithField("pkg", "resourcesync"), "ResourceSync", threadIntervalMinute(2), resourceSync.Poll)
+	resourceSyncThread.Start()
+
+	t.threads[0] = repoTesterThread
+	t.threads[1] = resourceSyncThread
+}
+
+func (t TaskManager) Stop() {
+	var once sync.Once
```
hexfusion, line 71: hang once sync.Once on the TaskManager and put everything in it. the larger concern is closing channels but I was thinking this.

```go
func (t TaskManager) Stop() {
   t.once.Do(func() {
     for _, thread := range t.threads {
		thread.Stop()
	}
     t.cancelFunc
     for c := range t.channels {
		close(t.channels[c])
     }
   })
   ```

hexfusion (approved):

internal/server/server.go
avishayt: Done

internal/server/server.go
avishayt: Done

internal/tasks/fleet_rollout.go
avishayt: No, it's a weird condition that should never exist.  It means that it was never initialized.

internal/tasks/fleet_rollout.go
avishayt: G601: Implicit memory aliasing in for loop. (gosec)

internal/tasks/manager.go
avishayt: Done

internal/tasks/fleet_rollout.go
hexfusion: interesting

PR #86, opened 2024-02-13
WIP: Add RHEL based bootc Containerfile

oglok:


fzdarsky:
I feel we should move this to the `flightctl-demos` repo, as it's not about FlightCtl itself, but examples how to use it (with different base OSes, with or without MicroShift, etc.).

packaging/bootc/Containerfile.rhel
```diff
@@ -0,0 +1,22 @@
+FROM registry-proxy.engineering.redhat.com/rh-osbs/rhel9-rhel_bootc:rhel-9.4
+ADD etc etc
```
fzdarsky, line 2: Use `dnf copr enable redhat-et/flightctl-dev` instead of manually adding a repo file?

packaging/bootc/Containerfile.rhel
```diff
@@ -0,0 +1,22 @@
...
+RUN rm -rf /opt && \
+    mkdir -p /opt/cni
+RUN dnf install -y microshift
+RUN ln -s microshift.service /usr/lib/systemd/system/default.target.wants
+RUN dnf install -y flightctl-agent
+RUN ln -s /usr/lib/systemd/system/podman.socket /usr/lib/systemd/system/multi-user.target.wants/
```
fzdarsky, line 8: Now that we've switched to `exec`'ing to podman instead of using the lib, do we still need this?

oglok:
I had the same thought. I was hesitating because it might be usefult to have one bootc image for testing directly from the repo, but for that, we could just improve our device simulator. wdyt?

packaging/bootc/Containerfile.rhel
```diff
@@ -0,0 +1,22 @@
+FROM registry-proxy.engineering.redhat.com/rh-osbs/rhel9-rhel_bootc:rhel-9.4
+ADD etc etc
+RUN rm -rf /opt && \
+    mkdir -p /opt/cni
+RUN dnf install -y microshift
```
fzdarsky, line 5: Maybe put all commands under a single `RUN` so we don't create additional layers?

packaging/bootc/Containerfile.rhel
oglok: we need to add the repo file for microshift anyway, and we'd avoid one command extra.

fzdarsky:
> I had the same thought. I was hesitating because it might be usefult to have one bootc image for testing directly from the repo, but for that, we could just improve our device simulator. wdyt?

You're referring to the problem that the simulator would want bootc to be available and working, right? Yeah, I think we'll need to mock that.

The nice thing about having these images in the demos repo is we can start the demo story from the image building and even include how to build these images using GitHub Actions.

packaging/bootc/Containerfile.rhel
oglok: Good one, I totally missed that change.

packaging/bootc/Containerfile.rhel
oglok: +1

oglok:
@fzdarsky I'll move all the bootc stuff to flightctl-demos repo with your suggestions.

Closing this PR then.

PR #90, opened 2024-02-15
Async task for setting device owner by selector

avishayt:
If a device's labels are updated, iterate over all fleets in the org.  The device's owner can remain, change to another fleet, can be removed if none match, or can be set to a special value "error:multiple" if more than one fleet matches.

If a fleet's selector is updated, list all devices that match the new selector.
1. If a device had no owner, set it to this fleet.
2. If a device had a different owner, check if the previous owner's selector matches the device's labels as well. If yes, set the owner to "error:multiple", if not set it to this fleet.
3. If a device had multiple owners ("error:multiple"), iterate over all fleets in the org to potentially find a new owner.
4. Find "orphaned" devices - those that are owned by this fleet by no longer match its selector. We remove the owner ref from these.


fzdarsky:
> If a device's labels are updated, iterate over all fleets in the org. The device's owner can remain, change to another fleet, can be removed if none match, or can be set to a special value "error:multiple" if more than one fleet matches.

I'm not a fan of special values to encode errors. Would mean the schema is not either ownerRef or empty, but ownerRef or empty or ErrorCode.

How about we model this situation as condition in the fleet status instead? That is do something like:
When selection conflict is detected between 2 or more fleets,
- _pause_ these fleets from updating their devices by setting a `.fleet.status.condition["pause"]` to "true" with reason "selectorConflict" or so. As results, the task rolling out fleet updates to devices skips these devices until the conflict is resolved.
- keep the affected devices' owner ref as is. Or set to empty. Important part is we continue to serve the last spec.

fzdarsky:
> If a device's labels are updated, iterate over all fleets [...]
> If a fleet's selector is updated, list all devices that match the new selector [...]

Arguably, the device spec served to agents can be eventually consistent with label/selector updates, i.e. it's ok to still serve the old spec even though the label/selector has already been updated. What's important is that this is consistent by the time the fleet-rollout task needs to make changes to the device spec.

Wondering how we can use this slack to optimize the process? For example, do we need to run this _every time_ a device label gets updated or just at the beginning of a fleet-rollout task?

internal/store/device.go
```diff
@@ -190,6 +240,18 @@ func (s *DeviceStore) Delete(ctx context.Context, orgId uuid.UUID, name string)
 	return result.Error
 }
 
+func (s *DeviceStore) UpdateDeviceOwner(ctx context.Context, orgId uuid.UUID, name string, owner string, callback DeviceStoreCallback) error {
+	device := model.Device{Resource: model.Resource{OrgID: orgId, Name: name}}
+	err := s.db.Model(device).Updates(map[string]interface{}{
+		"generation": 0,
```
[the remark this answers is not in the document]
mangelajo: same question.

internal/store/fleet.go
```diff
@@ -187,14 +187,16 @@ func (s *FleetStore) createOrUpdateTx(tx *gorm.DB, ctx context.Context, orgId uu
 			}
 			fleet.Generation = util.Int64ToPtr(1)
 			fleet.Spec.Data.Template.Metadata.Generation = util.Int64ToPtr(1)
-			templateUpdated = true
+			templateUpdated = false // No point in rolling out until selectors are processed
+			selectorUpdated = true
```
mangelajo, line 203: May be this should be logic in the callback itself?, it's in the end inner knowledge that we first need to process all the devices and update based on selector, then update based on the template content. I'd move it down there.

internal/store/fleet.go
```diff
@@ -208,6 +210,10 @@ func (s *FleetStore) createOrUpdateTx(tx *gorm.DB, ctx context.Context, orgId uu
 				fleet.Generation = existingRecord.Generation
 			}
 
+			if !sameSelector {
+				selectorUpdated = true
```
mangelajo, line 226: some useless comment ahead:
it could be this, but probably less idiomatic. `selectorUpdated = selectorUpdated || !sameSelector`

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+			logic := FleetSelectorMatchingLogic{
+				taskManager: taskManager,
+				ctx:         ctx,
+				log:         log,
+				fleetStore:  taskManager.store.Fleet(),
+				devStore:    taskManager.store.Device(),
+				resourceRef: resourceRef,
+			}
+
+			if resourceRef.Kind == model.FleetKind {
```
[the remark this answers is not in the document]
mangelajo: +1 We could break down to a 2nd sub-function to avoid the excessive indent.

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+			if err != nil {
+				f.log.Errorf("failed to get owner of device %s/%s: %v", f.resourceRef.OrgID, *device.Metadata.Name, err)
+				errors++
+				continue
+			}
+			if ownerType != model.FleetKind {
+				continue
+			}
+			currentOwnerFleetName := ownerName
+			fmt.Printf("CURRENT OWNER: %s , FLEET: %s\n", currentOwnerFleetName, f.resourceRef.Name)
```
mangelajo, line 136: Should we use logging instead?

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+type FleetSelectorMatchingLogic struct {
+	taskManager TaskManager
+	ctx         context.Context
+	log         logrus.FieldLogger
+	fleetStore  store.Fleet
+	devStore    store.Device
+	resourceRef ResourceReference
+}
+
+func (f FleetSelectorMatchingLogic) fleetSelectorUpdated() error {
```
mangelajo, line 69: We would need to architect all this in a way that we could additionally perform dry-run checks, i.e. from a pull request/merge request that we have been notified of via webhook.

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+
+	for {
+		devices, err := f.devStore.List(f.ctx, f.resourceRef.OrgID, listParams)
+		if err != nil {
+			return fmt.Errorf("failed to list devices that no longer belong to fleet: %w", err)
+		}
+
+		for devIndex := range devices.Items {
+			device := devices.Items[devIndex]
+			fmt.Printf("PROCESSING DEVICE %s\n", *device.Metadata.Name)
```
mangelajo, line 102: move to logging?

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+func (f FleetSelectorMatchingLogic) handleOwningFleetChanged(device *api.Device, fleet *api.Fleet, currentOwnerFleetName string) error {
+	// "fleet" is potentially the new owner of "device" because, but we first need
+	// to make sure that the label selectors of both the current fleet and the new
+	// fleet aren't a match for this device.
+	currentOwningFleet, err := f.fleetStore.Get(f.ctx, f.resourceRef.OrgID, currentOwnerFleetName)
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+		return err
+	}
+
+	fmt.Printf("CHECKING IF CURRENT OWNER %s STILL MATCHES: %v\n", currentOwnerFleetName, currentOwningFleet)
```
mangelajo, line 175: move to logging?

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+	newOwnerFleet := *fleet.Metadata.Name
+	if currentOwningFleet != nil && util.LabelsMatchLabelSelector(*device.Metadata.Labels, currentOwningFleet.Spec.Selector.MatchLabels) {
+		fmt.Printf("DEVICE MATCHES CURRENT AND NEW\n")
```
mangelajo, line 178: Move to logging?

internal/store/device.go
```diff
@@ -190,6 +240,18 @@ func (s *DeviceStore) Delete(ctx context.Context, orgId uuid.UUID, name string)
...
+		"owner":      owner,
+	}).Error
```
mangelajo, line 248: I would also insert a status condition update about the owner change for some degree of traceability on the device.

internal/store/fleet.go
```diff
@@ -122,9 +122,21 @@ func (s *FleetStore) ListIgnoreOrg() ([]model.Fleet, error) {
...
+		if result.Error != nil {
+			return result.Error
+		}
+
+		// Unset owner for all devices in org, now that all fleets have been deleted
+		devCondition := model.Device{
+			Resource: model.Resource{OrgID: orgId},
+		}
+		result = tx.Model(devCondition).Select("owner").Updates(map[string]interface{}{"owner": nil})
+		return result.Error
```
mangelajo, line 137: I think we should make this a call to the Device store instead, the generation should be incremented, and probably the change also tracked in the status.condition updates.

internal/store/fleet.go
```diff
@@ -307,9 +319,22 @@ func (s *FleetStore) updateStatusTx(tx *gorm.DB, ctx context.Context, orgId uuid
...
+		fleetCondition := model.Fleet{
+			Resource: model.Resource{OrgID: orgId, Name: name},
+		}
+		result := tx.Unscoped().Delete(&fleetCondition)
+		if result.Error != nil {
+			return result.Error
+		}
+
+		// Unset owner for all devices that this fleet owned
+		devCondition := model.Device{
```
mangelajo, line 332: Same comment as above for the device updates. Let's do it on the device store.

mangelajo (changes requested):
Some comments on a first pass. Nice work in general.

internal/store/device.go
```diff
@@ -135,35 +143,77 @@ func (s *DeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string) (*a
...
-	findDevice.OrgID = orgId
-	result := s.db.First(&findDevice)
-	if result.Error != nil {
-		if result.Error == gorm.ErrRecordNotFound {
+	labelsUpdated := false
+
+	err := s.db.Transaction(func(innerTx *gorm.DB) (err error) {
+		existingRecord := model.Device{Resource: model.Resource{OrgID: device.OrgID, Name: device.Name}}
+		result := innerTx.First(&existingRecord)
+		log.Printf("db.Find(%s): %d rows affected, error is %v", existingRecord, result.RowsAffected, result.Error)
```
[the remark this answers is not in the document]
avishayt: It's all copy-paste; I'll make a task to remove all these and see how to turn on DB logs for debugging.

internal/service/fleet.go
```diff
@@ -115,7 +115,7 @@ func (h *ServiceHandler) ReplaceFleet(ctx context.Context, request server.Replac
 		return server.ReplaceFleet400Response{}, fmt.Errorf("fleet name specified in metadata does not match name in path")
 	}
 
-	result, created, err := h.store.Fleet().CreateOrUpdate(ctx, orgId, request.Body, h.taskManager.FleetTemplateRolloutCallback)
+	result, created, err := h.store.Fleet().CreateOrUpdate(ctx, orgId, request.Body, h.taskManager.FleetUpdatedCallback)
```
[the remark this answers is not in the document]
avishayt: Maybe, need to think about how it would look.

internal/store/device.go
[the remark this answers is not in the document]
avishayt: Currently, the device generation matches the fleet template generation.  So, if you see a device whose generation is lower than the generation of its fleet's template, you know the device should be updated.  The agent needs to look at the pair of owner and generation to know if a device's spec changed.  I know the behavior doesn't match k8s, but I don't want to be too religious about that, and this made sense to me.  I'm open to other suggestions.

internal/store/device.go
avishayt: What kind of condition?  When would it be set/unset?
Also, the agent owns the device status.
I was thinking that we would implement an event mechanism for traceability in general.

internal/store/fleet.go
avishayt: I can make it call the device store.  We need to discuss device generations and conditions, as per the comment above.

internal/store/fleet.go
avishayt: It's just telling the callback what changed, not what to do.  I can leave "templateUpdated=true" to avoid inner knowledge.  It doesn't break anything, just a small optimization.  WDYT?

internal/tasks/fleet_selector.go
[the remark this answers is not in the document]
avishayt: ACK

internal/tasks/fleet_selector.go
avishayt: Not sure I follow, let's discuss on Monday.

internal/tasks/fleet_selector.go
```diff
@@ -0,0 +1,317 @@
...
+	if fleet.Spec.Selector == nil || len(fleet.Spec.Selector.MatchLabels) == 0 {
+		return nil
+	}
+
+	// List the devices that now match the fleet's selector
+	listParams := store.ListParams{
+		Labels: fleet.Spec.Selector.MatchLabels,
+		Limit:  ItemsPerPage,
+	}
+	errors := 0
```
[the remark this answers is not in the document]
avishayt: Correct, though there's not much to do except for print a log at this point.  We'll have better retries once we have a proper task mechanism.

internal/tasks/fleet_selector.go
avishayt: Debug log, need to remove, thanks

internal/tasks/fleet_selector.go
avishayt: Debug log, need to remove

internal/tasks/fleet_selector.go
avishayt: Debug log, need to remove

internal/tasks/fleet_selector.go
avishayt: Debug log, need to remove

avishayt:
> > If a device's labels are updated, iterate over all fleets in the org. The device's owner can remain, change to another fleet, can be removed if none match, or can be set to a special value "error:multiple" if more than one fleet matches.
> 
> I'm not a fan of special values to encode errors. Would mean the schema is not either ownerRef or empty, but ownerRef or empty or ErrorCode.

Yea I don't love it either.

> How about we model this situation as condition in the fleet status instead? That is do something like: When selection conflict is detected between 2 or more fleets,

I'll see how this looks.  Need to think of the scenarios of when the condition is set and unset.

> * _pause_ these fleets from updating their devices by setting a `.fleet.status.condition["pause"]` to "true" with reason "selectorConflict" or so. As results, the task rolling out fleet updates to devices skips these devices until the conflict is resolved.
> * keep the affected devices' owner ref as is. Or set to empty. Important part is we continue to serve the last spec.

This is the effect I was going for, which is what happens in the current PR.  The device is in this "limbo" owner and wouldn't have its spec updated, but its spec is preserved.
I like the suggestion, just need to work through the details.

avishayt:
> > If a device's labels are updated, iterate over all fleets [...]
> > If a fleet's selector is updated, list all devices that match the new selector [...]
> 
> Arguably, the device spec served to agents can be eventually consistent with label/selector updates, i.e. it's ok to still serve the old spec even though the label/selector has already been updated. What's important is that this is consistent by the time the fleet-rollout task needs to make changes to the device spec.
> 
> Wondering how we can use this slack to optimize the process? For example, do we need to run this _every time_ a device label gets updated or just at the beginning of a fleet-rollout task?

I think it's important to communicate to the user which fleet each device belongs to.  If they update a label/selector, they would expect to see the effect, especially if there's some error with overlapping selectors.

fzdarsky:
> I think it's important to communicate to the user which fleet each device belongs to. If they update a label/selector, they would expect to see the effect, especially if there's some error with overlapping selectors.

There'd always be _some_ lag, at least until the fleet-selector task completed, or? But I see your point.

internal/store/fleet.go
sagidayan: @avishayt You should add `.Where()` otherwise you will remove owner from all.
`tx.Model(devCondition).Where(devCondition).Select("owner").Updates(map[string]interface{}{"owner": nil})`

internal/util/util.go
```diff
@@ -170,3 +177,16 @@ func CreateRandomJitterDuration(max int64, duration time.Duration) time.Duration
 	jitteredDuration := duration * time.Duration(n)
 	return jitteredDuration
 }
+
+func LabelsMatchLabelSelector(labels map[string]string, labelSelector map[string]string) bool {
```
hexfusion, line 181: This is expected but I think it would be good to document empty label selector is a match

internal/util/util.go
```diff
@@ -153,9 +157,12 @@ func GetResourceOwner(owner *string) (string, string, error) {
 	if owner == nil {
 		return "", "", fmt.Errorf("owner string is nil")
 	}
+	if strings.HasPrefix(*owner, "error:") {
+		return "error", *owner, nil
+	}
```
hexfusion, line 162: can we document why this is the case?

internal/tasks/manager.go
```diff
@@ -81,16 +101,100 @@ func (t TaskManager) Stop() {
...
 	t.channels[taskName] <- resource
 }
 
 func (t TaskManager) GetTask(taskName string) ResourceReference {
 	return <-t.channels[taskName]
 }
 
-func (t TaskManager) FleetTemplateRolloutCallback(orgId uuid.UUID, name *string, templateUpdated bool) {
-	t.SubmitTask(FleetTemplateRollout, ResourceReference{OrgID: orgId, Name: *name})
+func (t TaskManager) FleetUpdatedCallback(before *model.Fleet, after *model.Fleet) {
```
hexfusion, line 113: what if we changed the format of the callback to something like this.

````
type SomethingCallback struct {
    Before func(fleet *model.Fleet) 
    After  func(fleet *model.Fleet) 
}
```

Then you can place the  before after calls in different sections of the call. Its a common pattern to have the resource (input) as the before and the response and the after. I think it might simplify some of this logic.

internal/store/device.go
```diff
@@ -182,15 +243,77 @@ func (s *DeviceStore) UpdateStatus(ctx context.Context, orgId uuid.UUID, resourc
...
+				Name:       name,
+				Owner:      &owner,
+				Generation: util.Int64ToPtr(0),
+			}}
+		callback(before, after)
+	}
+	return err
+}
+
+func (s *DeviceStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, name string, annotations *map[string]string, deleteKeys *[]string) error {
```
hexfusion, line 288: nit: it's strange to have a param that is a reference to a map. Under the hood it's already a referenced type.  same with `LabelMapToArray` if you want to know if its empty just check len.

```
if len(annotations) == 0 {
  // do something
}
```

hexfusion: Same for slices.

internal/store/device.go
```diff
@@ -135,35 +152,79 @@ func (s *DeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string) (*a
...
+			return result.Error
+		}
+		if result.Error != nil {
 			created = true
+			device.Generation = util.Int64ToPtr(1)
+
+			result = innerTx.Create(device)
+			if result.Error != nil {
+				return result.Error
+			}
```
hexfusion, line 189: can we document why we overwrite result.Error instead of returning it immediately?

hexfusion (commented):
first pass on new code will take another look in morning

internal/store/device.go
avishayt: I made the code clearer to show that the first check of result.Error actually means that the device wasn't found and it's not a real error.

internal/store/device.go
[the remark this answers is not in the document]
avishayt: Added a comment. I'll deal with this more in the upcoming task of rendering the config.

internal/store/device.go
avishayt: Fixed

internal/tasks/manager.go
avishayt: Sorry, I don't follow

internal/util/util.go
avishayt: This code is outdated, from a previous version where I set the owner to "error:multiple" if the device had multiple owners.  Removed.

internal/util/util.go
avishayt: I don't see it this way... I think it's a bug.  A fleet with no label selector should match nothing, not everything.  I changed the code to match this behavior.

internal/store/device.go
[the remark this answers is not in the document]
avishayt: Opened [MGMT-17070](https://issues.redhat.com/browse/MGMT-17070)

internal/store/fleet.go
avishayt: Fixed

internal/store/fleet.go
avishayt: Code was removed

internal/store/device_test.go
```diff
@@ -266,5 +257,38 @@ var _ = Describe("DeviceStore create", func() {
...
+		})
+
+		It("UpdateDeviceAnnotations", func() {
+			firstAnnotations := map[string]string{"key1": "val1", "key2": "val2"}
+			err := store.Device().UpdateAnnotations(ctx, orgId, "mydevice-1", firstAnnotations, nil)
+			Expect(err).ToNot(HaveOccurred())
+			dev, err := store.Device().Get(ctx, orgId, "mydevice-1")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(dev.Metadata.Annotations).ToNot(BeNil())
+			Expect(len(*dev.Metadata.Annotations)).To(Equal(2))
```
mangelajo, line 281: Some of the coolnes of ginkgo would let you do this too:
```
    Expect(*dev.Metadata.Annotations).To(HaveLen(2))
```

mangelajo: and same in other places where we use len

mangelajo (commented):
I am halfway through the review, I need to make a pause.

internal/util/util.go
hexfusion: OK I guess we should decide if we are going to maintain k8s parity or not.

https://github.com/kubernetes/kubernetes/blob/54f9807e1e84981b2053f4daf779f5ed19962144/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types_swagger_doc_generated.go#L166

internal/store/test_common.go
```diff
@@ -0,0 +1,73 @@
+package store
```
hexfusion, line 1: these get compiled into the binary can we put these helpers in test/util/util.go along with any other test helpers.

internal/tasks/common.go
```diff
@@ -1,5 +1,26 @@
...
 
-const (
-	ItemsPerPage = 1000
+import (
+	api "github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/store/model"
+	"github.com/flightctl/flightctl/internal/util"
 )
+
+var ItemsPerPage int
```
hexfusion, line 9: We should only use globals when necessary is this the case here? Global scope can introduce strange bugs and introduce race conditions.

internal/tasks/fleet_rollout.go
```diff
@@ -6,23 +6,46 @@ import (
...
+				}
+			} else if resourceRef.Kind == model.DeviceKind {
+				err := logic.rolloutDevice()
+				if err != nil {
+					taskManager.log.Errorf("failed rolling out device %s/%s: %v", resourceRef.OrgID, resourceRef.Name, err)
+				}
+			} else {
+				taskManager.log.Errorf("FleetRollouts called with incorrect resource kind %s", resourceRef.Kind)
 			}
 		case <-taskManager.ctx.Done():
```
hexfusion, line 50: I think we want to out this case at the top so we fail fast. WDYT?

internal/tasks/fleet_rollout.go
```diff
@@ -6,23 +6,46 @@ import (
...
-		case fleetRef := <-taskManager.channels[FleetTemplateRollout]:
-			err := rolloutFleet(taskManager.log, taskManager.store.Fleet(), taskManager.store.Device(), fleetRef)
-			if err != nil {
-				taskManager.log.Errorf("failed rolling out fleet %s/%s: %v", fleetRef.OrgID, fleetRef.Name, err)
+		case resourceRef := <-taskManager.channels[ChannelFleetTemplateRollout]:
+			requestID := reqid.NextRequestID()
+			ctx := context.WithValue(context.Background(), middleware.RequestIDKey, requestID)
+			log := log.WithReqIDFromCtx(ctx, taskManager.log)
+			logic := FleetRolloutsLogic{
+				ctx:         ctx,
```
hexfusion, line 26: its generally considered an antipattern to attach ctx to a struct can we look at ways to pass it as a param?

internal/tasks/fleet_rollout.go
```diff
@@ -6,23 +6,46 @@ import (
...
+				devStore:    taskManager.store.Device(),
+				resourceRef: resourceRef,
+			}
+
+			if resourceRef.Op != FleetRolloutOpUpdate {
+				taskManager.log.Errorf("received unknown op %s", resourceRef.Op)
+				break
+			}
+			if resourceRef.Kind == model.FleetKind {
+				err := logic.rolloutFleet()
```
hexfusion, line 38:
```
err := logic.rolloutFleet(ctx)
```

internal/tasks/fleet_rollout.go
```diff
@@ -31,43 +54,55 @@ func FleetRollouts(taskManager TaskManager) {
 	}
 }
 
-func rolloutFleet(baseLog logrus.FieldLogger, fleetStore store.Fleet, devStore store.Device, fleetRef ResourceReference) error {
-	requestID := reqid.NextRequestID()
-	ctx := context.WithValue(context.Background(), middleware.RequestIDKey, requestID)
-	log := log.WithReqIDFromCtx(ctx, baseLog)
+type FleetRolloutsLogic struct {
+	ctx         context.Context
```
hexfusion, line 58: reminder ctx should not be stored it should be passed

internal/tasks/fleet_rollout.go
```diff
@@ -31,43 +54,55 @@ func FleetRollouts(taskManager TaskManager) {
...
+	devStore    store.Device
+	resourceRef ResourceReference
+}
 
-	log.Infof("Rolling out fleet %s/%s", fleetRef.OrgID, fleetRef.Name)
+func (f FleetRolloutsLogic) rolloutFleet() error {
+	f.log.Infof("Rolling out fleet %s/%s", f.resourceRef.OrgID, f.resourceRef.Name)
 
-	fleet, err := fleetStore.Get(ctx, fleetRef.OrgID, fleetRef.Name)
+	fleet, err := f.fleetStore.Get(f.ctx, f.resourceRef.OrgID, f.resourceRef.Name)
```
hexfusion, line 68: ctx would be consumed naturally if it is passed as a param

```
func (f FleetRolloutsLogic) rolloutFleet(ctx context.Context) error {
```

internal/tasks/fleet_rollout.go
```diff
@@ -79,7 +114,44 @@ func rolloutFleet(baseLog logrus.FieldLogger, fleetStore store.Fleet, devStore s
...
+func (f FleetRolloutsLogic) rolloutDevice() error {
+	f.log.Infof("Rolling out device %s/%s", f.resourceRef.OrgID, f.resourceRef.Name)
+
+	device, err := f.devStore.Get(f.ctx, f.resourceRef.OrgID, f.resourceRef.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get device: %w", err)
+	}
+
+	if device.Metadata.Owner == nil || len(*device.Metadata.Owner) == 0 {
+		f.log.Info("Device has no owner")
```
hexfusion, line 127: Warn or is this expected? if exected will it be over chatty at scale?

internal/tasks/fleet_rollout.go
```diff
@@ -79,7 +114,44 @@ func rolloutFleet(baseLog logrus.FieldLogger, fleetStore store.Fleet, devStore s
...
+		return nil
+	}
+
+	if device.Metadata.Annotations != nil {
+		multipleOwners, ok := (*device.Metadata.Annotations)[model.DeviceAnnotationMultipleOwners]
+		if ok && len(multipleOwners) > 0 {
+			f.log.Infof("Device has multiple owners, skipping rollout: %s", multipleOwners)
```
hexfusion, line 134: this seems Warn or even Error level?

internal/tasks/fleet_rollout.go
```diff
@@ -79,7 +114,44 @@ func rolloutFleet(baseLog logrus.FieldLogger, fleetStore store.Fleet, devStore s
...
+	if !isFleetOwner {
+		return nil
+	}
+
+	fleet, err := f.fleetStore.Get(f.ctx, f.resourceRef.OrgID, owner)
+	if err != nil {
+		return fmt.Errorf("failed to get fleet %s: %w", owner, err)
+	}
+
+	return f.updateDeviceSpecAccordingToFleetTemplate(device, fleet)
```
hexfusion, line 151: :heart:

internal/tasks/manager.go
```diff
@@ -53,12 +70,15 @@ func Init(log logrus.FieldLogger, store store.Store) TaskManager {
 }
 
 func (t TaskManager) Start() {
+	ItemsPerPage = 1000
```
hexfusion, line 73: const?

internal/tasks/manager.go
hexfusion: we can talk about this more as followup

internal/tasks/manager.go
```diff
@@ -81,16 +101,100 @@ func (t TaskManager) Stop() {
...
-	t.SubmitTask(FleetTemplateRollout, ResourceReference{OrgID: orgId, Name: *name})
+func (t TaskManager) FleetUpdatedCallback(before *model.Fleet, after *model.Fleet) {
+	var templateUpdated bool
+	var selectorUpdated bool
+	var fleet *model.Fleet
+
+	if before == nil && after == nil {
+		// Shouldn't be here, return
+		return
+	} else if before == nil {
```
hexfusion, line 121:
```suggestion
if before == nil && after == nil {
		// Shouldn't be here, return
		return
}

if before == nil {
````

File: internal/tasks/manager.go [hunk 34]
```diff
@@ -81,16 +101,100 @@ func (t TaskManager) Stop() {
...
+}
+
+func (t TaskManager) DeviceUpdatedCallback(before *model.Device, after *model.Device) {
+	var labelsUpdated bool
+	var ownerUpdated bool
+	var device *model.Device
+
+	if before == nil && after == nil {
+		// Shouldn't be here, return
+		return
```
hexfusion, line 166: same as above

internal/store/device.go
```diff
@@ -135,35 +152,87 @@ func (s *DeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string) (*a
...
+			// Update the generation if the spec was updated
+			if !sameSpec {
+				if existingRecord.Owner != nil && len(*existingRecord.Owner) != 0 {
+					return fmt.Errorf("cannot update a device owned by a fleet")
+				}
+
+				if existingRecord.Generation == nil {
+					device.Generation = util.Int64ToPtr(1)
+				} else {
+					device.Generation = util.Int64ToPtr(*existingRecord.Generation + 1)
```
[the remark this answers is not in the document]
avishayt: I like the idea but don't think we would reuse it.

internal/store/device.go
```diff
@@ -182,15 +251,78 @@ func (s *DeviceStore) UpdateStatus(ctx context.Context, orgId uuid.UUID, resourc
...
+
+func (s *DeviceStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, name string, annotations map[string]string, deleteKeys []string) error {
+	err := s.db.Transaction(func(innerTx *gorm.DB) (err error) {
+		existingRecord := model.Device{Resource: model.Resource{OrgID: orgId, Name: name}}
+		result := innerTx.First(&existingRecord)
+		if result.Error != nil {
+			return result.Error
+		}
+		existingAnnotations := util.LabelArrayToMap(existingRecord.Annotations)
+		if annotations != nil {
```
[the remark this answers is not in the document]
avishayt: Actually no need to check for nil

internal/store/device.go
```diff
@@ -182,15 +251,78 @@ func (s *DeviceStore) UpdateStatus(ctx context.Context, orgId uuid.UUID, resourc
...
+			existingAnnotations = util.MergeLabels(existingAnnotations, annotations)
+		}
+
+		if len(deleteKeys) > 0 {
```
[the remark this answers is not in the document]
avishayt: Actually no need for the "if"

internal/store/device_test.go
avishayt: Nice!

internal/store/fleet_test.go
```diff
@@ -317,22 +306,22 @@ var _ = Describe("FleetStore create", func() {
 			Expect(*updatedFleet.Metadata.Owner).To(Equal("test"))
 
 			updatedFleet.Metadata.Owner = util.StrToPtr("test2")
-			updated = false
+			called = false
 			_, _, err = store.Fleet().CreateOrUpdate(ctx, orgId, updatedFleet, callback)
-			Expect(updated).To(BeFalse())
+			Expect(called).To(BeFalse())
 			Expect(err).To(HaveOccurred())
```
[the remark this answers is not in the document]
avishayt: Good point.  I opened MGMT-17082 to tackle it separately.

internal/store/model/resource.go
```diff
@@ -29,6 +29,8 @@ type Resource struct {
 	// that we can perform indexing and queries on them.
 	Labels pq.StringArray `gorm:"type:text[]"`
 
+	Annotations pq.StringArray `gorm:"type:text[]"`
```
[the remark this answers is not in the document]
avishayt: Discussed on slack.

internal/store/test_common.go
avishayt: That was a pain to change without cyclic imports, but did it :)

internal/tasks/fleet_rollout.go
avishayt: Removed

internal/util/util.go
avishayt: You're right, this is more consistent.  Changed.

PR #91, opened 2024-02-16
MGMT-16972 Deploy via kind+helm

mangelajo:
This pull request moves the deployment of flightctl-server and db to use the existing
helm charts which have been tuned to allow conditional deployment in kind, and conditional deployment of the server.

This also adds the capability for the flightctl cmdline to use other servers based on the ~/.fligthctl/client.yaml config

Other services like keycloak, or metrics also need to be deployed with helm, but that will be addressed in a follow up patch as this one is already lengthy enough.

### DIFF
--- a/.github/workflows/pr-smoke-testing.yaml
+++ b/.github/workflows/pr-smoke-testing.yaml
@@ -9,9 +9,10 @@ on:
 permissions:
   contents: read
   pull-requests: read
+
 jobs:
   # This line defines a job with the ID `check-links` that is stored within the `jobs` key.
-  podman-smoke-test:
+  kind-cluster:
     runs-on: "ubuntu-latest"
     steps:
       - name: Checkout
@@ -27,24 +28,18 @@ jobs:
           go install gotest.tools/gotestsum@latest
           go get -u github.com/proglottis/gpgme
 
-      - name: Install podman-compose
-        run: pip3 install podman-compose
-
-      - name: Install podman 4
-        run: |
-          sudo mkdir -p /etc/apt/keyrings
-          curl -fsSL "https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/Release.key" \
-            | gpg --dearmor \
-            | sudo tee /etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg > /dev/null
-          echo \
-            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg]\
-              https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
-            | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list > /dev/null
-          sudo apt-get update -qq
-          sudo apt-get -qq -y install podman
-
+      - name: Create cluster
+        run: make cluster
+      
       - name: Deploy
-        run: make build deploy
+        run: make deploy
+
+      - name: Check
+        run: |
+          kubectl get pods --all-namespaces
+
+      - name: Make the cmdline tool
+        run: make build
 
       - name: Apply device
         run: bin/flightctl apply -f examples/device.yaml
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -26,22 +26,8 @@ jobs:
           go install gotest.tools/gotestsum@latest
           go get -u github.com/proglottis/gpgme
 
-      # used to deploy the database container during unit testing
-      - name: Install podman 4
-        run: |
-          sudo mkdir -p /etc/apt/keyrings
-          curl -fsSL "https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/Release.key" \
-            | gpg --dearmor \
-            | sudo tee /etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg > /dev/null
-          echo \
-            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg]\
-              https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
-            | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list > /dev/null
-          sudo apt-get update -qq
-          sudo apt-get -qq -y install podman
-
-      - name: Install podman-compose
-        run: pip3 install podman-compose
+      - name: Deploy DB with helm and kind
+        run: make deploy-db-helm
 
       - name: Running Unit tests
-        run: make unit-test
+        run: make run-unit-test
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,11 @@ help:
 	@echo "    lint:            run golangci-lint"
 	@echo "    build:           run all builds"
 	@echo "    unit-test:       run unit tests"
-	@echo "    deploy:          deploy flightctl-server and db as containers in podman"
-	@echo "    deploy-db:       deploy only the database as a container in podman"
+	@echo "    deploy:          deploy flightctl-server and db as pods in kind"
+	@echo "    deploy-db:       deploy only the database as a pod in kind"
 	@echo "    clean:           clean up all containers and volumes"
+	@echo "    cluster:         create a kind cluster and load the flightctl-server image"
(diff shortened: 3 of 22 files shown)


hack/install_kind.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+KIND_VERSION=0.22.0
+
+if [ -x "$(command -v kind)" ]; then
+    kind version | grep -q v${KIND_VERSION} && echo "Kind v${KIND_VERSION} already installed"
```
mangelajo, line 5: This doesn't upgrade pre-existing 0.21 version in github ci.

.github/workflows/unit-tests.yaml
```diff
@@ -26,22 +26,5 @@ jobs:
...
-              https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
-            | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list > /dev/null
-          sudo apt-get update -qq
-          sudo apt-get -qq -y install podman
-
-      - name: Install podman-compose
-        run: pip3 install podman-compose
-
       - name: Running Unit tests
         run: make unit-test
```
mangelajo, line 30: This now automatically creates a cluster with only the DB and the admin settings.

deploy/helm/flightctl/values.yaml
```diff
@@ -8,10 +8,12 @@ flightctl:
     user: demouser
     userPassword: demopass
   server:
+    enabled: true
     namespace: flightctl-external
     image: quay.io/flightctl/flightctl-server:latest
     imagePullPolicy: Always
     hostName: api.flightctl.example.com
+    #nodePort: false # if true, it will use the hostPort value instead of a router/ingress
```
mangelajo, line 16: We can remove this altogether.

.github/workflows/pr-smoke-testing.yaml
```diff
@@ -27,24 +28,18 @@ jobs:
...
-          echo \
-            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/devel_kubic_libcontainers_unstable.gpg]\
-              https://download.opensuse.org/repositories/devel:kubic:libcontainers:unstable/xUbuntu_$(lsb_release -rs)/ /" \
-            | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:unstable.list > /dev/null
-          sudo apt-get update -qq
-          sudo apt-get -qq -y install podman
-
-      - name: Deploy
-        run: make build deploy
+          sleep 60
```
mangelajo, line 39: We don't need this sleep anymore because of:
https://github.com/flightctl/flightctl/pull/91/commits/7aafc2f617951f6de1e38e7cda16ff83dc746f13#diff-dd530797092baca20b42189b556574346477aa6f9c99fd3ce82668d3ba4aee19R62

hack/deploy_with_helm.sh
```diff
@@ -0,0 +1,73 @@
...
+helm ${METHOD} --values ./deploy/helm/flightctl/values.kind.yaml ${ONLY_DB} ${NO_AUTH} ${DB_IMG} flightctl ./deploy/helm/flightctl/
+
+"${SCRIPT_DIR}"/wait_for_postgres.sh
+
+# Make sure the database is usable from the unit tests
+DB_POD=$(kubectl get pod -n flightctl-internal -L flightctl.service=flightctl-db --no-headers -o custom-columns=":metadata.name")
+kubectl exec -n flightctl-internal "${DB_POD}" -- psql -c 'ALTER ROLE admin WITH SUPERUSER'
+kubectl exec -n flightctl-internal "${DB_POD}" -- createdb admin 2>/dev/null|| true
+
+mkdir -p  ~/.flightctl/certs
```
mangelajo, line 57: evertything after this line could be moved to it's own script (useful if you need to re-extract credentials from a running server volume).

i.e.
`./hack/get_flightctl_server_creds.sh`

cmd/flightctl/main.go
```diff
@@ -97,6 +96,15 @@ type GetOptions struct {
 	Continue      string
 }
 
+func getConfig() (*Config, error) {
```
mangelajo, line 99: I will remove the config changes for waiting for other patches in flight like: #99 and #81

oglok:
I think you need a rebase

hack/install_kind.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
hexfusion, line 1: minor: 
```suggestion
#!/usr/bin/env bash
```

hack/install_helm.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
hexfusion, line 1: minor: 
```suggestion
#!/usr/bin/env bash
```

hack/deploy_with_helm.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
hexfusion, line 1: minor: 
```suggestion
#!/usr/bin/env bash
```

Makefile
```diff
@@ -48,13 +50,6 @@ build: bin
 flightctl-server-container:
 	podman build -f Containerfile -t flightctl-server:latest
 
-deploy-db:
-	cd deploy/podman && podman-compose up -d flightctl-db
```
hexfusion, line 52: could we leave this as `deploy-db-podman` or allow override via ENV? As a dev I would like to iterate with the least amount of deps possible.

hexfusion (commented):
Looks cool, I would like to see a shellcheck test as a follow-up so we can keep all this bash sane :). WDYT?  Added a few notes.

hack/install_kind.sh
hexfusion: Great additions though thanks for adding these scripts.

Makefile
mangelajo: It is in the deploy.mk file now, you can do deploy-db and it will only deploy the db. I moved all the deployment makefile targets to a separate file.

Is that what you are looking for?

hack/install_kind.sh
mangelajo: I always find it annoying to bring up a devel environment asking for lots of tiny tools ':D

hexfusion (approved):

deploy/deploy.mk
```diff
@@ -0,0 +1,27 @@
...
+
+deploy-helm:
+	kubectl config set-context kind-kind
+	hack/install_helm.sh
+	hack/deploy_with_helm.sh
+
+deploy-db-helm: cluster
+	hack/deploy_with_helm.sh --only-db
+
+deploy-db:
```
hexfusion, line 18: :heart:

deploy/helm/flightctl/values.kind.yaml
```diff
@@ -0,0 +1,12 @@
+flightctl:
+  db:
+    nodePort: 5432 # this is also mappend in /hack/create_cluster.sh as an extraPortMapping
```
avishayt, line 3: s/mappend/mapped/

deploy/helm/flightctl/values.kind.yaml
```diff
@@ -0,0 +1,12 @@
...
+    imagePullPolicy: IfNotPresent
+  server:
+    image: localhost/flightctl-server:latest
+    imagePullPolicy: IfNotPresent
+    hostName: localhost
+    nodePort: 3333 # this is also mappend in /hack/create_cluster.sh as an extraPortMapping
```
avishayt, line 9: s/mappend/mapped/

examples/repository.yaml
```diff
@@ -3,6 +3,4 @@ kind: Repository
 metadata:
   name: defaultRepo
 spec:
-  repo: https://github.com/flightctl/flightctl.git
-  username: myuser
-  password: mypassword
+  repo: https://github.com/flightctl/flightctl-demos.git
```
avishayt, line 6: Nice

Makefile
```diff
@@ -115,26 +109,16 @@ _post_unit_test: $(REPORTS)
 run-unit-test:
 	SKIP_UT_DB=1 $(MAKE) _unit_test TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./...))"
 
-run-db-container:
-	podman rm -f flightctl-db || true
-	podman volume rm podman_flightctl-db || true
-	podman volume create --opt device=tmpfs --opt type=tmpfs --opt o=nodev,noexec podman_flightctl-db
```
mangelajo, line 121: re-add these volume settings

examples/repository.yaml
mangelajo: Oh this is unrelated to this patch. ':D

Makefile
mangelajo: Done in deploy.mk

mangelajo:
Performed the requested changes. I am waiting for CI to pass, then I will merge.

PR #93, opened 2024-02-16
*: refactor agent and add config controller

hexfusion:
Please do not review yet lots of things changing.

Built on #88


internal/agent/configcontroller/enrollment.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	klog.Infof("%swaiting for enrollment to be approved", c.logPrefix)
+	err := wait.ExponentialBackoff(c.enrollmentVerifyBackoff, func() (bool, error) {
+		return c.verifyDeviceEnrollment(ctx, device)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to verify device enrollment: %w", err)
+	}
+	if err := c.writeDeviceEnrollmentBanner(ctx, device); err != nil {
```
fzdarsky, line 34: The banner is what allows enrolment requests to be eventually approved, so needs to happen before waiting for approval.

internal/agent/configcontroller/config_controller.go
```diff
@@ -0,0 +1,262 @@
...
+
+	var conditions []v1alpha1.DeviceCondition
+	// post enrollment update status
+	deviceCondition := v1alpha1.DeviceCondition{
+		Type:   "Enrolled",
+		Status: v1alpha1.True,
+	}
+	conditions = append(conditions, deviceCondition)
+	deviceStatus.Conditions = &conditions
+	_, updateErr := c.managementClient.UpdateDeviceStatus(ctx, *device.Metadata.Name, deviceStatus)
```
fzdarsky, line 122: Is this message sent on every call to sync(), even if the device is already enrolled?

internal/agent/configcontroller/config_controller.go
```diff
@@ -0,0 +1,259 @@
...
+	if updateErr != nil {
+		klog.Errorf("%sfailed to update device status: %v", c.logPrefix, updateErr)
+	}
+
+	// ensure the device is configured
+	if err := c.ensureConfig(ctx, device); err != nil {
+		klog.Errorf("%s configuration did not succeed: %v", c.logPrefix, err)
```
fzdarsky, line 129: If the configuration din't complete for whatever reason, now do we now completely roll back any changes (in a transaction with the OS image update)?

internal/agent/configcontroller/enrollment.go
fzdarsky, line 34: This needs to happen before waiting for approval, as it's pre-requisite for the approval process.

fzdarsky:
Overall, this refactoring looks great!

One question I still have is why introduce parallelism that makes it harder to reason about the agent to ensure determinism and safety? We're like polling for status update at most 1/min while taking ms to complete, so it's not clear why we need a worker queue and async exporters/observers?

internal/agent/configcontroller/config_controller.go
hexfusion: Correct rollback is not yet supported.

internal/agent/configcontroller/config_controller.go
hexfusion: makes sense to batch this.

internal/agent/configcontroller/config_controller.go
```diff
@@ -0,0 +1,259 @@
...
+	Raw  json.RawMessage `json:"inline"`
+	Name string          `json:"name"`
+}
+
+func (c *ConfigController) ensureConfig(_ context.Context, device *v1alpha1.Device) error {
+	if device.Spec.Config == nil {
+		return fmt.Errorf("device config is nil")
+	}
+
+	for _, config := range *device.Spec.Config {
```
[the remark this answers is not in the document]
hexfusion: Right I am going off the example provided WRT inline config so we can adjust this once that flow gets nailed down.

hexfusion:
I need to rebase

PR #99, opened 2024-02-20
MGMT-17000: refactor agent and add config controller

hexfusion:
This PR adds
- test/util: a package for providing reusable testing functionality
- ability to run both ginkgo and go tests
- config controller to manage parsing device.Spec.Config decoding ignition and writing the bytes to disk. This borrows heavily from MCD without adding it as a dep.
- a test providing e2e coverage on enrollment and config file sync.
- refactor agent

Jira:
MGMT-17000
MGMT-16999


internal/util/util.go
```diff
@@ -159,3 +161,27 @@ func GetResourceOwner(owner *string) (string, string, error) {
...
+	n, err := rand.Int(rand.Reader, big.NewInt(max))
+	if err != nil {
+		return 0, err
+	}
+
+	return (duration * time.Duration(n.Int64())), nil
+}
+
+// NewTickerWithJitter creates a new ticker with a random jitter duration in the range [0, max) * duration
+func NewTickerWithJitter(d time.Duration, jitterDuration time.Duration, maxJitter int64) (*time.Ticker, error) {
```
fzdarsky, line 181: This implementation isn't a jitter anymore, i.e. random fluctuations in the interval between ticks that still have a mean interval of d, but a constant interval between ticks that's been chosen once at startup from [d, d + [0, j_max] * j)). One consequence is higher sync and lower staggering of requests, but more importantly it'd be nice to have d as the mean interval to be able to reason about the load of the system independently of jitter (e.g. a system with d=60s and 100k agents should see a load of 1.7k requests per second).

internal/util/util.go
```diff
@@ -1,8 +1,10 @@
 package util
 
 import (
+	"crypto/rand"
```
fzdarsky, line 4: Do we need the more costly randomness vs. `math/rand`?

Makefile
```diff
@@ -103,7 +104,7 @@ clean:
 	- rm -f -r debian
 
 _unit_test: $(REPORTS)
-	gotestsum $(GO_UNITTEST_FLAGS) $(TEST) $(GINKGO_UNITTEST_FLAGS) -timeout $(TIMEOUT) || ($(MAKE) _post_unit_test && /bin/false)
+	gotestsum -- $(GO_UNITTEST_FLAGS) -timeout $(TIMEOUT) ./... || ($(MAKE) _post_unit_test && /bin/false)
```
mangelajo, line 107: Don't we run the ginkgo tests anymore?

mangelajo (commented):
I did a first pass and it looks good, some nits that we could handle in follow up.

The unit test for the device agent doesn't seem to work though, and a question about ginkgo.

internal/agent/device/device.go
```diff
@@ -0,0 +1,162 @@
...
+		return err
+	}
+	defer fetchStatusTicker.Stop()
+	configSyncTicker, err := util.NewTickerWithJitter(a.syncConfigInterval, 100*time.Millisecond, 50)
+	if err != nil {
+		return err
+	}
+	defer configSyncTicker.Stop()
+
+	for {
```
fzdarsky, line 96: Let's review in the near future when / based on which events we sync config or update status. For example:
* We currently periodically poll for spec updates. Maybe in the future we add a server push of spec updates, too. When the spec updates, is there any reason to not immediately call configController.Sync()?
* Which other events require us to run config.Sync()? Does it need to run periodically?
* Which events require us to fetch status? And if status changes, would we immediately want to notify the service?

internal/agent/device/config/controller.go
```diff
@@ -0,0 +1,165 @@
...
+		Cap:      3 * time.Minute,
+		Duration: 10 * time.Second,
+		Factor:   1.5,
+		Steps:    24,
+	}
+
+	c := &Controller{
+		enrollmentClient:        enrollmentClient,
+		enrollmentVerifyBackoff: enrollmentVerifyBackoff,
+		enrollmentEndpoint:      enrollmentEndpoint,
```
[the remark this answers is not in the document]
fzdarsky: The enrollment API endpoint may become different in the future, when we have central enrollment for multiple management service instances, for example.

I think you're more concerned that the UI's endpoint (linked from the QR code) is different from the API's endpoint.

Makefile
hexfusion: this invocation runs both

internal/util/util.go
hexfusion: true this only randomizes the start time, we can move back to jitterbug

internal/agent/device/config/controller.go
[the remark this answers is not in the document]
hexfusion: my understanding is there are potentially two service enrollment and management or just one where they are both the same. Will the UI's endpoint be different from these endpoints potentially?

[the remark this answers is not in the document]
fzdarsky: The QR code (which the agent generates on the console, but could be printed on the device's casing) contains a URL to the UI, specifically the details page of the device's enrollment request. Therefore, it's a different endpoint from the corresponding URL on the API server.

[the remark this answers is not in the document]
hexfusion: so it would actually handled by the managentServiceEndpoint

internal/agent/device/config/bootstrap.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if err := os.WriteFile(filepath.Join(c.managementCertFilePath), []byte(*enrollmentRequest.Status.Certificate), os.FileMode(0600)); err != nil {
+		return false, fmt.Errorf("%swriting signed certificate: %v", c.logPrefix, err)
+	}
+
+	return true, nil
+}
+
+func (c *Controller) writeDeviceEnrollmentBanner(_ context.Context, device *v1alpha1.Device) error {
+	url := fmt.Sprintf("%s/enroll/%s", c.enrollmentEndpoint, *device.Metadata.Name)
```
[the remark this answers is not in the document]
hexfusion: makes sense

Makefile
hexfusion: image

internal/util/util.go
mangelajo: True, I had the same thought but I didn’t consider it of importance, but @fzdarsky you are right, it’s statistically important

internal/agent/device/config/controller.go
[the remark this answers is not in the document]
mangelajo: Correct now we have 2: management api and enrollment UI

in the future we could have management api, enrollment api, and enrollment UI

PR #110, opened 2024-02-22
MGMT-17016: Set ResourceSync 'Generation' on Create API call

sagidayan:


celdrake (approved):
Worked well from the UI.

PR #120, opened 2024-02-26
MGMT-16972 Avoid rebuilding the server container every time we deploy

mangelajo:
We introduce an intermediate empty target that depends on go.mod,
go.sum and the go files of the project.

The rebuild has become more evident during the work to deploy with kind+helm.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,12 @@ lint: tools
 build: bin
 	go build -buildvcs=false $(GO_BUILD_FLAGS) -o $(GOBIN) ./cmd/...
 
-flightctl-server-container:
+# rebuild container only on source changes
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,8 +45,11 @@ lint: tools
 build: bin
 	go build -buildvcs=false $(GO_BUILD_FLAGS) -o $(GOBIN) ./cmd/...
 
-flightctl-server-container:
+bin/.flightctl-server-container: bin Containerfile go.mod go.sum $(shell find ./ -name "*.go" -not -path "./packaging/*")
```
hexfusion, line 48: optional: add a comment `# rebuild container only on source change`

hexfusion (approved):

Issue #121, opened 2024-02-26
Client config adds /api, at the end which breaks client usage


mangelajo:
<img width="1641" alt="image" src="https://github.com/flightctl/flightctl/assets/1176434/36027035-8b63-4df5-a1e5-12f77feb39df">

```
$ cat ~/.flightctl/client.yaml
authentication:
  client-certificate-data: ---
  client-key-data: ---
service:
  certificate-authority-data: ---
  server: https://localhost:3333/api/
``` 

should be
```
$ cat ~/.flightctl/client.yaml
authentication:
  client-certificate-data: ---
  client-key-data: ---
service:
  certificate-authority-data: ---
  server: https://localhost:3333/
```

mangelajo:
I will also verify why the tests don't fail

mangelajo:
@fzdarsky *

PR #122, opened 2024-02-26
Fix client file endpoint

mangelajo:
Fixes-Issue: #121

### DIFF
--- a/cmd/flightctl/main.go
+++ b/cmd/flightctl/main.go
@@ -521,56 +521,66 @@ func applyFromReader(client *apiclient.ClientWithResponses, filename string, r i
 		}
 
 		if dryRun {
-			fmt.Printf("%s: applying device/%s (dry run only)\n", filename, resourceName)
+			fmt.Printf("%s: applying %s/%s (dry run only)\n", strings.ToLower(kind), filename, resourceName)
 			continue
 		}
-		fmt.Printf("%s: applying device/%s: ", filename, resourceName)
+		fmt.Printf("%s: applying %s/%s: ", strings.ToLower(kind), filename, resourceName)
 		buf, err := json.Marshal(resource)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("%s: skipping resource of kind %q: %v", filename, kind, err))
 		}
 
+		var httpResponse *http.Response
+
 		switch kind {
 		case "Device":
-			response, err := client.ReplaceDeviceWithBodyWithResponse(context.Background(), resourceName, "application/json", bytes.NewReader(buf))
-			if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/flightctl/main.go
```diff
@@ -554,23 +558,31 @@ func applyFromReader(client *apiclient.ClientWithResponses, filename string, r i
...
 				continue
 			}
 			fmt.Printf("%s\n", response.HTTPResponse.Status)
+			httpResponse = *response.HTTPResponse
 		default:
 			errs = append(errs, fmt.Errorf("%s: skipping resource of unkown kind %q: %v", filename, kind, resource))
 		}
+
+		// bad HTTP Responses don't generate an error on the OpenAPI client, we need to check the status code manually
+		if httpResponse.StatusCode != http.StatusOK && httpResponse.StatusCode != http.StatusCreated {
```
hexfusion, line 583: if we hit the default case would this panic?

cmd/flightctl/main.go
```diff
@@ -554,23 +558,31 @@ func applyFromReader(client *apiclient.ClientWithResponses, filename string, r i
...
+			errs = append(errs, fmt.Errorf("%s: failed to apply device/%s: %v", filename, resourceName, httpResponse.Status))
```
fzdarsky, line 584:
```suggestion
			errs = append(errs, fmt.Errorf("%s: failed to apply %s/%s: %v", filename, strings.ToLower(kind), resourceName, httpResponse.Status))
```

cmd/flightctl/main.go
hexfusion: no it will just be status 0 I thought it was a ref my bad :)

cmd/flightctl/main.go
fzdarsky: Maybe fixing lines 524 and 527 in the process :)

cmd/flightctl/main.go
fzdarsky, line 583: In the default case of the switch statement, httpResponse.StatusCode would be == 0, so produce two errors.

mangelajo: ah but it would emit an additional error in such case. I'll fix that

mangelajo: each call seems to return a different type, but I will try to factor out as much as I can.

mangelajo: done

cmd/flightctl/main.go
mangelajo: done.

cmd/flightctl/main.go
mangelajo: done.

PR #125, opened 2024-02-27
NO-ISSUE: Makefile: fix override of GO_UNITTEST_FLAGS

hexfusion:
This PR fixes a bug where a test overide to `GO_UNITTEST_FLAGS` with -race flag was committed. It also cleans up the population of the `coverprofile` and  `junitfile` artifacts into the the reports directory. The Ginkgo specific ENV have been removed for now as they are run directly by gotestsum. In the future we can use full ginkgo test invocation for e2e tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,14 +7,13 @@ VERBOSE ?= false
 REPORTS ?= $(ROOT_DIR)/reports
 
 GO_TEST_FORMAT = pkgname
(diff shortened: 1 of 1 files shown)


hexfusion:
Pinning the commit message to Jira/GH feels excessive should we force squash on merge so the merge message always contains the reference from the title? I will update it for now.

mangelajo (approved):

PR #126, opened 2024-02-27
NO-ISSUE: internal/agent: improve cert documentation/naming

hexfusion:
This PR improves documentation and makes the certificate naming more obvious to its usecase.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -25,14 +25,14 @@ import (
 const (
 	// name of the CA bundle file
 	caBundleFile = "ca.crt"
-	// name of the agent's key file
+	// name of the agent's key file, this file is generated if it does not exist
 	agentKeyFile = "agent.key"
 	// name of the enrollment certificate file
 	enrollmentCertFile = "client-enrollment.crt"
 	// name of the enrollment key file
 	enrollmentKeyFile = "client-enrollment.key"
-	// name of the management client certificate file
-	clientCertFile = "client.crt"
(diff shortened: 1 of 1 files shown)


internal/agent/agent.go
```diff
@@ -25,14 +25,14 @@ import (
...
+	// name of the agent's key file, this file is generated if it does not exist
 	agentKeyFile = "agent.key"
 	// name of the enrollment certificate file
 	enrollmentCertFile = "client-enrollment.crt"
 	// name of the enrollment key file
 	enrollmentKeyFile = "client-enrollment.key"
-	// name of the management client certificate file
-	clientCertFile = "client.crt"
+	// name of the management client certificate file, this file is generated during bootstrap
+	managementClientCertFile = "client-management.crt"
```
hexfusion, line 35: this could cause trouble for existing deployments but I assume these deployments to be transient.

mangelajo: I wonder if we would just want to source a client.yaml instead like the flightctl bin

mangelajo: @oglok

hexfusion: I think the agent does this but 100% it can be improved. For example ideally the config validates the existence of the certs and popualtes the path so the full path is what is sent via Config. Then the names are arbirary.

https://github.com/flightctl/flightctl/blob/54ed3534fb5fc842f16166720b8d0ca763a92b7b/cmd/flightctl-agent/main.go#L20

PR #131, opened 2024-02-29
Handle input flags properly

mangelajo:
Otherwise we never get the data-dir right, and defaults are initialized with anything that not necessarily matches the config file.

I am not completely happy with this implementation, I'd propose that we switch instead to just accepting a config file, the config file already contains a certDir.


hexfusion (approved):

avishayt:
@mangelajo, Is this PR still relevant?

mangelajo:
Not anymor, closing it , thanks @avishayt

PR #133, opened 2024-03-01
MGMT-17101: api/v1alpha1: add device conditions

hexfusion:
This PR adds the following device conditions.

### Device: Config (MCD Like)
These are the core conditions that cover the operation of the agent's device config controller

| Name    | Desc |
| -------- | ------- |
| Available  | True if the controller is available for deploying configurations and postAction commands otherwise False  |
| Progressing | True if the controller is in the process of reconciling Spec,otherwise False    |
| Degraded    |  True if the controllers functionality is impaired in some way but not unavailable, otherwise False |

### Device: Workload (Kubelet/Node Like)
These are the core conditions that cover the ability of the device to accept/manage workloads.

| Name    | Desc |
| -------- | ------- |
| Ready  | True if the Device is ready and can accept new containers otherwise False. This is also the only condition which can be set by the server and can be done if healthchecks have not been updated for the device-monitor-grace-period|
| DiskPressure | True if pressure exists on the disk size—that is, if the disk capacity is low; otherwise False |
| PIDPressure | True if pressure exists on the processes—that is, if there are too many processes on the node; otherwise False | 
| MemoryPressure |True if pressure exists on the device memory—that is, if the device memory is low; otherwise False


mangelajo:
It looks reasonable, I would turn the nice PR you wrote into a document living along the openapi.yaml

PR #135, opened 2024-03-02
Further CLI Improvements

fzdarsky:
Uses a common structure for options and how they are completed, validated, and run.
Cleans up description of commands and flags.
Adds support for JSON as output format for get command.

Depends on #136

### DIFF
--- a/internal/cli/apply.go
+++ b/internal/cli/apply.go
@@ -33,9 +33,8 @@ type ApplyOptions struct {
 func NewCmdApply() *cobra.Command {
 	o := &ApplyOptions{Filenames: []string{}, DryRun: false, Recursive: false}
 	cmd := &cobra.Command{
-		Use:                   "apply -f FILENAME",
-		DisableFlagsInUseLine: true,
-		Short:                 "apply a configuration to a resource by file name or stdin",
+		Use:   "apply -f FILENAME",
+		Short: "Apply a configuration to a resource by file name or stdin.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(cmd, args); err != nil {
 				return err
@@ -49,7 +48,7 @@ func NewCmdApply() *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.StringSliceVarP(&o.Filenames, "filename", "f", o.Filenames, "read resources from file or directory")
+	flags.StringSliceVarP(&o.Filenames, "filename", "f", o.Filenames, "The files or directory that contain the resources to apply.")
 	annotations := make([]string, 0, len(fileExtensions))
 	for _, ext := range fileExtensions {
 		annotations = append(annotations, strings.TrimLeft(ext, "."))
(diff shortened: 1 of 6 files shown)


mangelajo (approved):
Looks great frank!

internal/cli/approve.go
```diff
@@ -11,36 +11,54 @@ import (
...
+				return err
+			}
+			return o.Run(cmd.Context(), args)
 		},
 		SilenceUsage: true,
 	}
 
-	cmd.Flags().StringArrayVarP(&approveLabels, "label", "l", []string{}, "labels to add to the device")
-	cmd.Flags().StringVarP(&approveRegion, "region", "r", "default", "region for the device")
+	cmd.Flags().StringArrayVarP(&o.ApproveLabels, "label", "l", []string{}, "Labels to add to the device, as a comma-separated list of key=value.")
```
oglok, line 37: while Use and Short is ok to start with capital letters, I think the standard for flags is lower case. Not strict rule though.

internal/cli/apply.go
```diff
@@ -32,25 +32,23 @@ type ApplyOptions struct {
...
 	cmd := &cobra.Command{
-		Use:                   "apply -f FILENAME",
-		DisableFlagsInUseLine: true,
-		Short:                 "apply a configuration to a resource by file name or stdin",
+		Use:   "apply -f FILENAME",
+		Short: "Apply a configuration to a resource by file name or stdin.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if !cmd.Flags().Lookup("filename").Changed {
-				return fmt.Errorf("must specify -f FILENAME")
+			if err := o.Complete(cmd, args); err != nil {
```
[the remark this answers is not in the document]
hexfusion: I believe the idea is to reuse the pattern on every command. 

<complete>
<verify>
<run>

Just for now some methods will not have any work but probably will in the future?

internal/cli/apply.go
```diff
@@ -59,12 +57,86 @@ func NewCmdApply() *cobra.Command {
...
+		default:
+			expandedFilenames, err := expandIfFilePattern(filename)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			for _, filename := range expandedFilenames {
+				_, err := os.Stat(filename)
+				if os.IsNotExist(err) {
+					errs = append(errs, fmt.Errorf("the path %q does not exist", filename))
```
hexfusion, line 100: My only question is should we fail fast here?

hexfusion (approved):

internal/cli/apply.go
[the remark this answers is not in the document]
fzdarsky: Yes, it's to have a common structure and a place for adding code in the future when we need to fill in options prior to validation. Structure is that also used in K8s.

internal/cli/approve.go
fzdarsky: Do you a pointer to best-practices or style guides? I'm ok either way, but this follows what kubectl standardises on.

internal/cli/apply.go
fzdarsky: IDK. Maybe? I only moved this method up without changing it. Q is whether we want to greedily apply as many resources as possible or to stop on the first error. What's K8s' behavior?

hexfusion: I think this is considered a preflight check, kube bails
```
oc apply -f test.yaml,test.broke
error: the path "test.broke" does not exist
```

hexfusion:
```
oc apply -f test.yaml,test.broke,test.broke2
the path "test.broke" does not exist
the path "test.broke2" does not exist
```

mangelajo:
I say we merge this one, and continue improving.

hexfusion (approved):

PR #136, opened 2024-03-03
MGMT-17019: Separate CLI commands and move to internal/cli package

fzdarsky:
Separates flightctl CLI commands into separate files.


internal/cli/apply.go
```diff
@@ -0,0 +1,255 @@
...
+				if err != nil {
+					errs = append(errs, fmt.Errorf("error walking %q: %v", filename, err))
+				}
+			}
+		}
+	}
+	return errors.Join(errs...)
+}
+
+func expandIfFilePattern(pattern string) ([]string, error) {
```
hexfusion, line 230: since you have pattern and its the only ever returned if we don't have error
```suggestion
func checkFilePattern(pattern string) error
```

internal/cli/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+
+var (
+	defaultClientConfigFile string
+	resourceKinds           = map[string]string{
+		"device":            "",
+		"enrollmentrequest": "",
+		"fleet":             "",
+		"repository":        "",
+		"resourcesync":      "",
+	}
```
hexfusion, line 20: if you dont expect to use the value you can use map[string]struct{} as a cheaper lookup.
```
resourceKinds := map[string]struct{}{
    "device":            {},
    "enrollmentrequest": {},
    "fleet":             {},
    "repository":       {},
    "resourcesync":     {},
}
```

PR #137, opened 2024-03-03
NO-ISSUE: Use consts for resource kinds in CLI

avishayt:


internal/cli/get.go
```diff
@@ -80,152 +80,101 @@ func NewCmdGet() *cobra.Command {
...
-				Continue:      cont,
-			}
-			response, err := c.ListDevicesWithResponse(context.Background(), &params)
-			if err != nil {
-				return fmt.Errorf("listing devices: %v", err)
-			}
-			return printListResourceResponse(response, err, "devices", output)
+	var response interface{}
+
+	switch true {
```
hexfusion, line 104: I love using a switch here very clean true is assumed in your && evaluations

```suggestion
	switch {
```

internal/cli/get.go
```diff
@@ -80,152 +80,101 @@ func NewCmdGet() *cobra.Command {
...
+	cmd.Flags().StringVar(&o.Owner, "owner", o.Owner, "filter by owner")
 	cmd.Flags().StringVarP(&o.LabelSelector, "labelselector", "l", o.LabelSelector, "label selector as a comma-separated list of key=value")
 	cmd.Flags().StringVarP(&o.Output, "output", "o", o.Output, "output format (yaml)")
 	cmd.Flags().Int32Var(&o.Limit, "limit", o.Limit, "the maximum number of results returned in the list response")
 	cmd.Flags().StringVar(&o.Continue, "continue", o.Continue, "query more results starting from the value of the 'continue' field in the previous response")
 	return cmd
 }
 
-func RunGet(kind, name string, labelSelector, fleetName *string, output string, limit *int32, cont *string) error {
+func RunGet(kind, name string, labelSelector, owner *string, output string, limit *int32, cont *string) error {
```
hexfusion, line 96: context is a critical component of graceful termination anytime we are Background()ing context we should ask why,
```suggestion
func RunGet(ctx context.Context, kind, name string, labelSelector, owner *string, output string, limit *int32, cont *string) error {
```

internal/cli/helpers.go
```diff
@@ -3,22 +3,28 @@ package cli
...
 
 	"k8s.io/client-go/util/homedir"
 )
 
+const (
+	DeviceKind            = "device"
+	EnrollmentRequestKind = "enrollmentrequest"
+	FleetKind             = "fleet"
+	RepositoryKind        = "repository"
+	ResourceSyncKind      = "resourcesync"
```
hexfusion, line 16: + 1

hexfusion (commented):
first pass looks good few notes.

internal/cli/get.go
avishayt: I probably have some golang knowledge gap here, but why should we pass ctx if we don't use it?

hexfusion: All of the calls below plumb context.Background() I assume they used it.

hexfusion: Ah I see its cli vs server but lets say you wanted to enforce a timeout. You would want a real context here.

internal/cli/helpers.go
hexfusion: question: are these globally used or is this just cli

avishayt: Just CLI

internal/cli/get.go
avishayt: OK I see now.  I didn't see all the context.Background() below.  Will fix, thanks!

avishayt: Fixed in new commit

internal/cli/delete.go
```diff
@@ -18,87 +18,87 @@ func NewCmdDelete() *cobra.Command {
...
 
 	switch kind {
-	case "device":
+	case DeviceKind:
 		if len(name) > 0 {
-			response, err := c.DeleteDeviceWithResponse(context.Background(), name)
+			response, err := c.DeleteDeviceWithResponse(ctx, name)
 			if err != nil {
-				return fmt.Errorf("deleting device/%s: %v", name, err)
+				return fmt.Errorf("deleting %s/%s: %v", kind, name, err)
```
hexfusion, line 39: minor: wrap errors %w

hexfusion (approved):
nice

PR #138, opened 2024-03-03
MGMT-17054: TemplateVersion CRUD

avishayt:
Create/List/DeleteAll use the path /api/v1/templateversions:
- Create puts the fleet in the spec so that it can be created with just a YAML and no additional parameters.
- List and DeleteAll allow listing/deleting TemplateVersions from multiple fleets in an org.

Get/Delete use the path /api/v1/fleets/{fleet}/templateversions/{name}:
- They must refer to a version of a specific fleet because the name isn't necessarily unique.


api/v1alpha1/openapi.yaml
```diff
@@ -1087,6 +1087,163 @@ paths:
...
+          schema:
+            type: string
+        - name: labelSelector
+          in: query
+          description: A selector to restrict the list of returned objects by their labels. Defaults to everything.
+          schema:
+            type: string
+        - name: limit
+          in: query
+          description: The maximum number of results returned in the list response. The server will set the 'continue' field in the list response if more results exist. The continue value may then be specified as parameter in a subesquent query.
```
hexfusion, line 1110: s/subesquent/subsequent/

api/v1alpha1/openapi.yaml
```diff
@@ -1087,6 +1087,163 @@ paths:
...
+        "401":
+          description: Unauthorized
+          content: {}
+        "409":
+          description: StatusConflict
+          content: {}
+    delete:
+      tags:
+        - templateversion
+      description: delete collection of template versions
```
hexfusion, line 1164:
```suggestion
      description: delete a collection of template versions
```

internal/cli/delete.go
```diff
@@ -18,14 +25,39 @@ func NewCmdDelete() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			return RunDelete(kind, name)
+
+			if cmd.Flags().Lookup("owner").Changed && kind != TemplateVersionKind {
```
hexfusion, line 29: feels like all this logic should be part of a verify method. In general, cobra exposes flags that populate some object then we verify values against defaults. This can be a followup.

internal/service/templateversion.go
```diff
@@ -0,0 +1,127 @@
...
+		return server.ListTemplateVersions400Response{}, fmt.Errorf("limit cannot exceed %d", store.MaxRecordsPerListRequest)
+	}
+
+	result, err := h.store.TemplateVersion().List(ctx, orgId, listParams)
+	switch err {
+	case nil:
+		return server.ListTemplateVersions200JSONResponse(*result), nil
+	default:
+		return nil, err
+	}
```
hexfusion, line 83:
```suggestion
	result, err := h.store.TemplateVersion().List(ctx, orgId, listParams)
        if err != nil{
           return err
        }
        return server.ListTemplateVersions200JSONResponse(*result), nil
}

File: internal/service/templateversion.go [hunk 5]
```diff
@@ -0,0 +1,127 @@
...
+func (h *ServiceHandler) DeleteTemplateVersions(ctx context.Context, request server.DeleteTemplateVersionsRequestObject) (server.DeleteTemplateVersionsResponseObject, error) {
+	orgId := store.NullOrgId
+
+	err := h.store.TemplateVersion().DeleteAll(ctx, orgId, request.Params.Owner)
+	switch err {
+	case nil:
+		return server.DeleteTemplateVersions200JSONResponse{}, nil
+	default:
+		return nil, err
+	}
```
hexfusion, line 96: same as above switch doesn't add value here

internal/store/model/templateversion.go
```diff
@@ -0,0 +1,112 @@
...
+			Owner:       resource.Metadata.Owner,
+			Annotations: util.LabelMapToArray(resource.Metadata.Annotations),
+		},
+		Spec:   MakeJSONField(resource.Spec),
+		Status: MakeJSONField(status),
+	}
+}
+
+func (t *TemplateVersion) ToApiResource() api.TemplateVersion {
+	if t == nil {
```
hexfusion, line 57: can we add a comment on how this can be true

internal/store/model/templateversion.go
```diff
@@ -0,0 +1,112 @@
...
+		Status: &status,
+	}
+}
+
+func (tl TemplateVersionList) ToApiResource(cont *string, numRemaining *int64) api.TemplateVersionList {
+	if tl == nil {
+		return api.TemplateVersionList{
+			ApiVersion: TemplateVersionAPI,
+			Kind:       TemplateVersionListKind,
+			Items:      []api.TemplateVersion{},
```
hexfusion, line 93: minor: since this is not a reference you get it for free so not necessary.

internal/store/model/templateversion.go
```diff
@@ -0,0 +1,112 @@
...
+
+	deviceList := make([]api.TemplateVersion, len(tl))
+	for i, device := range tl {
+		deviceList[i] = device.ToApiResource()
+	}
+	ret := api.TemplateVersionList{
+		ApiVersion: TemplateVersionAPI,
+		Kind:       TemplateVersionListKind,
+		Items:      deviceList,
+		Metadata:   api.ListMeta{},
```
hexfusion, line 105: same

api/v1alpha1/openapi.yaml
```diff
@@ -1743,6 +1976,7 @@ components:
       - DeviceAvailable
       - DeviceProgressing
       - DeviceDegraded
+      - TemplateVersionReady
```
hexfusion, line 1979: I am not openapi expert so i am curious what is the best way to document these. For example it would be nice to understand what `TemplateVersionReady` means.

internal/store/templateversion.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+type TemplateVersion interface {
+	Create(ctx context.Context, orgId uuid.UUID, templateVersion *api.TemplateVersion) (*api.TemplateVersion, error)
+	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.TemplateVersionList, error)
+	DeleteAll(ctx context.Context, orgId uuid.UUID, fleet *string) error
+	Get(ctx context.Context, orgId uuid.UUID, fleet string, name string) (*api.TemplateVersion, error)
+	Delete(ctx context.Context, orgId uuid.UUID, fleet string, name string) error
+	InitialMigration() error
+}
```
hexfusion, line 25: these interfaces feel like copy/paste boilerplate I wonder if we can use go generics here. High level we have 2 objects

- Resource
- ResourceList

So we have a single interface implemented for each resource, this can be a followup but I would like to track it.
```
type Service[T Resource, L ResourceList[T]] interface {
	Create(ctx context.Context, orgId uuid.UUID, resource *T) (*T, error)
	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*L, error)
	DeleteAll(ctx context.Context, orgId uuid.UUID, fleet *string) error
	Get(ctx context.Context, orgId uuid.UUID, fleet string, name string) (*T, error)
	Delete(ctx context.Context, orgId uuid.UUID, fleet string, name string) error
	InitialMigration() error
}
```

internal/service/templateversion.go
```diff
@@ -0,0 +1,127 @@
...
+
+	labelMap, err := labels.ConvertSelectorToLabelsMap(labelSelector)
+	if err != nil {
+		return nil, err
+	}
+
+	cont, err := store.ParseContinueString(request.Params.Continue)
+	if err != nil {
+		return server.ListTemplateVersions400Response{}, fmt.Errorf("failed to parse continue parameter: %s", err)
```
hexfusion, line 61:
```suggestion
		return server.ListTemplateVersions400Response{}, fmt.Errorf("failed to parse continue parameter: %w", err)

File: internal/cli/delete.go [hunk 12]
```diff
@@ -74,6 +106,23 @@ func RunDelete(kind, name string) error {
 			}
 			fmt.Printf("%s\n", response.Status())
 		}
+	case TemplateVersionKind:
+		if len(name) > 0 {
+			response, err := c.DeleteTemplateVersionWithResponse(context.Background(), *fleetName, name)
+			if err != nil {
+				return fmt.Errorf("deleting %s/%s: %v", kind, name, err)
```
hexfusion, line 113:
```suggestion
				return fmt.Errorf("deleting %s/%s: %w", kind, name, err)
```

internal/cli/delete.go
```diff
@@ -74,6 +106,23 @@ func RunDelete(kind, name string) error {
 			}
 			fmt.Printf("%s\n", response.Status())
 		}
+	case TemplateVersionKind:
+		if len(name) > 0 {
+			response, err := c.DeleteTemplateVersionWithResponse(context.Background(), *fleetName, name)
```
hexfusion, line 111: I wish I had a real context  0:)
```suggestion
			response, err := c.DeleteTemplateVersionWithResponse(ctx, *fleetName, name)
```

api/v1alpha1/openapi.yaml
```diff
@@ -1087,6 +1087,163 @@ paths:
         "404":
           description: NotFound
           content: {}
+  /api/v1/templateversions:
```
hexfusion, line 1090: this isn't part of /api/vi/fleet? could you add details of the API in the description and use case please :pray:

avishayt: I put an explanation in the commit message, but added it here as well.

api/v1alpha1/openapi.yaml
avishayt: Changed all 6 occurrences of this copy-paste typo :)

api/v1alpha1/openapi.yaml
avishayt: Fixed all 6 occurrences :)

api/v1alpha1/openapi.yaml
avishayt: I suppose we can add # comments?

internal/cli/delete.go
avishayt: Do you mean to use PreRunE for validations rather than RunE?

internal/cli/delete.go
avishayt: Fixed for all!

internal/cli/delete.go
avishayt: Fixed for all!

internal/service/templateversion.go
avishayt: Fixed for all resources

internal/service/templateversion.go
avishayt: I definitely get what you're saying.  It's the convention in this project for all methods of all resources.  I assume the motivation was that there are some methods that deal with multiple error codes.  I opened MGMT-17112 to tackle this.

internal/service/templateversion.go
avishayt: Will handle in MGMT-17112

internal/store/model/templateversion.go
avishayt: Done

internal/store/templateversion.go
avishayt: I would love to see that happen.  There is a ton of copy-paste in this service for each resource.  Definitely not in the scope of this PR though.  @sagidayan was looking into this.

hexfusion: 100% followup

hexfusion (approved):

hexfusion (approved):
I think this will evolve over time but this is good progress to move us forward.

PR #143, opened 2024-03-04
NO-ISSUE: Return error messages in API

avishayt:


api/v1alpha1/openapi.yaml
```diff
@@ -1925,15 +2194,13 @@ components:
...
+      required:
+        - message
       properties:
-        error:
+        message:
           type: string
-          description: 'Error message.'
-      required:
-        - error
+          description: Error message
```
mangelajo, line 2203: Nice, now all of them can return proper errors.

We should also talk to the UI folks, sometimes UI expects a list of parameters where the error is detected so they can highlight it on the UI when some input is wrong.

I am pinging them for feedback on that (for a follow up improvement)

mangelajo (approved):
Looks good to me

internal/service/enrollmentrequest.go
```diff
@@ -160,8 +159,11 @@ func (h *ServiceHandler) ReadEnrollmentRequest(ctx context.Context, request serv
 // (PUT /api/v1/enrollmentrequests/{name})
 func (h *ServiceHandler) ReplaceEnrollmentRequest(ctx context.Context, request server.ReplaceEnrollmentRequestRequestObject) (server.ReplaceEnrollmentRequestResponseObject, error) {
 	orgId := store.NullOrgId
-	if request.Body.Metadata.Name == nil || request.Name != *request.Body.Metadata.Name {
-		return server.ReplaceEnrollmentRequest400Response{}, nil
+	if request.Body.Metadata.Name == nil {
+		return server.ReplaceEnrollmentRequest400JSONResponse{Message: "resource name not specified in metadata"}, nil
```
rawagner, line 163: I'd prefer to have exact path in the error msg like 
```suggestion
		return server.ReplaceEnrollmentRequest400JSONResponse{Message: "metadata.name is not specified"}, nil
```

I believe this aligns with what k8s does. It is especially handy for more complicated paths.

PR #145, opened 2024-03-05
MGMT-17100: Factor out agent/server/db/client test harness

mangelajo:
This allows reusability of the test harness to provide an easy framework to write more mocked E2E tests.

### DIFF
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -1,87 +1,38 @@
-package agent
+package agent_test
 
 import (
-	"context"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
 	"github.com/flightctl/flightctl/api/v1alpha1"
-	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/util"
-	"github.com/flightctl/flightctl/pkg/log"
-	testutil "github.com/flightctl/flightctl/test/util"
+	"github.com/flightctl/flightctl/test/harness"
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/yaml"
 )
 
 func TestDeviceAgent(t *testing.T) {
-	require := require.New(t)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	require := require.New(t)
 
 	testDirPath := t.TempDir()
 	defer t.Cleanup(func() {
 		_ = os.RemoveAll(testDirPath)
 	})
-	err := makeTestDirs(testDirPath, []string{"/etc/issue.d/"})
-	require.NoError(err)
-
-	serverCfg := *config.NewDefault()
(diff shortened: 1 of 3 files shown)


internal/agent/agent_test.go
```diff
@@ -1,87 +1,35 @@
-package agent
+package agent_test
```
mangelajo, line 1: We need to avoid a circular dependency, we can test the agent without internal knowledge here.

internal/agent/testdata/device.yaml
```diff
@@ -18,7 +18,7 @@ spec:
             - contents:
                 source: >-
                   data:,This%20system%20is%20managed%20by%20flightctl.%0A
-              mode: 600
+              mode: 0600
```
mangelajo, line 21: A small error that I fixed on the way it's actually not relevant to this change. I will remove

test/util/util.go
```diff
@@ -28,66 +25,76 @@ const (
 )
 
 // NewTestServer creates a new test server and returns the server and the listener listening on localhost's next available port.
-func NewTestServer(t testing.TB, log logrus.FieldLogger, cfg *config.Config, store store.Store, ca *crypto.CA, serverCerts *crypto.TLSCertificateConfig) (*server.Server, net.Listener) {
-	t.Helper()
-	require := require.New(t)
+func NewTestServer(log logrus.FieldLogger, cfg *config.Config, store store.Store, ca *crypto.CA, serverCerts *crypto.TLSCertificateConfig) (*server.Server, net.Listener, error) {
```
mangelajo, line 28: I also made them independent of the testing library to allow usability from ginkgo or testify as wanted.

internal/agent/agent_test.go
hexfusion: if we moved the test into test/integration/agent would that help and consume your harness?

mangelajo: That would also work, I find it nicer to keep test and tested code closer while we can. But we could change that later if we think that's better.

internal/agent/agent_test.go
```diff
@@ -1,87 +1,35 @@
...
 )
 
 func TestDeviceAgent(t *testing.T) {
-	require := require.New(t)
-
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
-	testDirPath := t.TempDir()
-	defer t.Cleanup(func() {
```
mangelajo, line 27: Oh this is missing, fixing.

mangelajo: I will be leaving this outside of the test harness because ginkgo provides a TempDir function that provides automatic cleanup at the end of tests.

mangelajo:
fixed a linter error, and re-added temp directory cleanup.

test/util/util.go
hexfusion: That makes sense but I believe Ginkgo can get that interface via `t := GinkgoT()` I guess I don't feel overly strong here but I was trying to ensure they are only ever used in tests.

PR #146, opened 2024-03-05
NO-ISSUE: expose timeout ENV.

hexfusion:
This PR allows the Makefile timeout to be configurable on a per test basis


Makefile
```diff
@@ -2,6 +2,7 @@ GOBASE=$(shell pwd)
 GOBIN=$(GOBASE)/bin
 GO_BUILD_FLAGS := ${GO_BUILD_FLAGS}
 ROOT_DIR := $(or ${ROOT_DIR},$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST)))))
+TIMEOUT ?= 30m
```
hexfusion, line 5: I think we can reduce this :)

mangelajo: hehe may be 10min for now would be enough. We can bump it later if needed.

But I think it's just more important to limit specific tests that wait to not wait more than expectable.

mangelajo (approved):

PR #148, opened 2024-03-06
MGMT-17100 Convert agent tests into Ginkgo

mangelajo:
Refactor the tests into ginkgo before adding more tests.

This makes use of the test harness, and segments the features being tested, although it adds some level of repetition to what we execute on the tests, this helps isolate better why an specific test could be failing.

### DIFF
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -1,113 +1,142 @@
 package agent_test
 
 import (
+	"io/fs"
 	"os"
 	"path/filepath"
 	"testing"
-	"time"
 
 	"github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/test/harness"
-	"github.com/stretchr/testify/require"
-	"k8s.io/apimachinery/pkg/util/wait"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 	"sigs.k8s.io/yaml"
 )
 
-func TestDeviceAgent(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/agent/agent_test.go
```diff
@@ -1,113 +1,139 @@
...
+				er, err := h.Client.ReadEnrollmentRequestWithResponse(h.Context, deviceName)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(er.JSON200.Status.Conditions).ToNot(BeNil())
+
+				Expect(v1alpha1.IsStatusConditionTrue(*er.JSON200.Status.Conditions, "Approved")).To(BeTrue())
+
+			})
+
+			It("should create a device", func() {
+				dev := waitForDevice(h)
```
mangelajo, line 66:
```suggestion
				dev := enrollAndWaitForDevice(h)
```

internal/agent/testdata/device.yaml
```diff
@@ -18,7 +18,7 @@ spec:
             - contents:
                 source: >-
                   data:,This%20system%20is%20managed%20by%20flightctl.%0A
-              mode: 600
+              mode: 0600
```
hexfusion, line 21: pretty sure I copied this from examples so we should update that can be a followup

hexfusion (approved):
looks like and e2e test now :)

internal/agent/testdata/device.yaml
mangelajo: Yes, we have the error all around. I was crawling inside the ignition config specs and docs, it seems like in the json formating octal is not even supported and you need to write the decimal version of that octal (384) for it to work .

At least in yaml we can use 0xxx

PR #149, opened 2024-03-06
MGMT-17137: Add systemd notify to ensure enrollment QR is ready

oglok:
Flightctl-agent service notifies readiness way before the enrollment QR code is written in the issue directory. The systemd service is supposed to get active before getty, but it depends on the intervals from the config file.

### DIFF
--- a/internal/agent/device/config/bootstrap.go
+++ b/internal/agent/device/config/bootstrap.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 
@@ -96,6 +97,7 @@ func (c *Controller) writeDeviceEnrollmentBanner(_ context.Context, device *v1al
 	if err := c.writeQRBanner("\nEnroll your device to flightctl by scanning\nthe above QR code or following this URL:\n%s\n\n", url); err != nil {
 		return fmt.Errorf("failed to write device enrollment banner: %w", err)
 	}
+
 	return nil
 }
 
@@ -133,6 +135,10 @@ func (c *Controller) writeQRBanner(message, url string) error {
 		return fmt.Errorf("failed to write banner to disk: %w", err)
 	}
 
+	if err := SdNotify("READY=1"); err != nil {
+		klog.Warningf("failed to notify systemd: %v", err)
+	}
+
 	// additionally print the banner into the output console
 	klog.Info(buffer.String())
 	return nil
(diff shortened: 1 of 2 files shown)


internal/agent/device/config/bootstrap.go
```diff
@@ -96,6 +97,24 @@ func (c *Controller) writeDeviceEnrollmentBanner(_ context.Context, device *v1al
 	if err := c.writeQRBanner("\nEnroll your device to flightctl by scanning\nthe above QR code or following this URL:\n%s\n\n", url); err != nil {
 		return fmt.Errorf("failed to write device enrollment banner: %w", err)
 	}
+
```
mangelajo, line 100: Please move it to the writeQRBanner instead, otherwise once enrolled we will never notify on next restart, but otherwise we will do it when we write the management banner.

mangelajo: https://github.com/flightctl/flightctl/pull/149/files#diff-7012d62d090ebd6a70117c0f57275b301bc6c86605ae42c51497463265d9b39aR134

mangelajo (changes requested):
see the comments

mangelajo:
not sure why it fails on the motd test, can you bump the timeout I have on agent_test.go from 20s to 40s just in case?

internal/agent/device/config/bootstrap.go
hexfusion: In the future the bootstrap logic will be moved before the controllers start so we will not wait for sync. But just to align on flow it will look like this. So regardless of enrolled or not once bootstrap is complete we notify.

<bootstrap>
<-- bootstrap complete -->
<notify>
<run agent>

Does that sound logical?

oglok:
> not sure why it fails on the motd test, can you bump the timeout I have on agent_test.go from 20s to 40s just in case?
It was a matter of rerunning the job, maybe it was just a slow run, because it works on devstack (oops, sorry, in my laptop)

internal/agent/device/config/bootstrap.go
mangelajo: makes sense, we reduce the initialization time. Also we may want to re-trigger bootstrap in some cases (may be simply by exiting the service and letting systemd restart)

mangelajo (approved):

internal/agent/device/config/bootstrap.go
oglok: Agree, waiting 60 seconds for the machine to give a tty makes no sense... We should add that refactor to JIRA

PR #151, opened 2024-03-06
 MGMT-17098: Ensure that agent will finish enrollment

mangelajo:
Now a device will re-enroll if credentials are removed and remote device is deleted.

This was triggered mostly when testing manually.


internal/agent/agent_test.go
```diff
@@ -113,6 +113,33 @@ var _ = Describe("Device Agent behavior", func() {
...
+
+				// wait for the agent to retrieve the agent certificate from the EnrollmentRequest
+				Eventually(h.AgentDownloadedCertificate, TIMEOUT, POLLING).Should(BeTrue())
+
+				h.StopAgent()
+				_, err := h.Client.DeleteDevice(h.Context, *dev.Metadata.Name)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = os.Remove(filepath.Join(h.TestDirPath, "agent.key"))
+				Expect(err).ToNot(HaveOccurred())
```
hexfusion, line 128: Why would we remove the agent key from disk[1].  Isn't this what generates the CSR? If you delete this do you also change the fingerprint on reboot? 

failing test case

- restart the agent
- ensure fingerprint is not changed

Then in general I feel the agent should not react to unclear signals. In this case we are proposing replacing client side certs because of a possible human error server side. (delete csr) For this to work safely there should be very clear signal in the spec of intent. enrolled: false is an example I don't love it but its clear.

All changes to state during runtime should be explicit and deliberate/auditable. 

[1] https://github.com/flightctl/flightctl/blob/main/internal/agent/agent.go#L82

hexfusion (changes requested):

internal/agent/agent_test.go
```diff
@@ -113,6 +113,33 @@ var _ = Describe("Device Agent behavior", func() {
 			})
 		})
 
+		When("device/enrollmentrequests have been deleted", func() {
```
hexfusion, line 116: I get the intent but I believe we have also already agreed that this flow is not stable or even desired. So why are we adding a test case?  We should have a clear path for this moving forward.

1.  proposed bug/feature
1.  will it be a supported flow
1.  design/implement

internal/agent/agent_test.go
mangelajo: Right, it is a weird situation, and now that I think about it... as soon as we delete the key, it doesn't,t matter if we delete the device/enrollment request or not.  

The agent, at this point, just feels it's not enrolled unless we persist some additional information to disk.

I started to think that we should ideally add another identity fingerprint related to the hardware (processor ID or anything that univocally tracks a device. so if an admin needs it can figure out that the same hardware was previously connected to a different device/enrollment request)

internal/agent/agent_test.go
mangelajo: Yes, I was blindly following the description of MGMT-17100 but it doesn't make much sense to support as a workflow. If key/csr are deleted, the device can't continue.

May be we should test for...

1) csr has been lost (a device looses power mid-enrollment) ... and but we guarantee that on the next attempt it will grab the CSR from the enrollment request.

2) device has csr+key, but the device does not exist anymore (has been deleted), what do we expect in that case, a new enrollment request to happen? (at least until we have more complex decomission flows).

internal/agent/agent_test.go
fzdarsky: Keep in mind we'll eventually use an agent.key generated and sealed in the TPM instead of the agent.key file on disk (though I may want to use that as fallback in case of TPM-less systems).

Then, consider that we *want* that an admin doesn't care about a real hardware identifier. If the agent.key ever needs to be deleted, e.g. as part of decommissioning, then,  yes, the device needs to be reenrolled with the new identity. Whether it's the same hardware shouldn't matter (but we _could_ probably determine that from the TPM chip if necessary).

mangelajo: But well, I can envision cases where the admin cares if it is the same hardware or not. For example, if you want to keep track of specific hardware for any reason (block hardware that could have been compromised from re-joining in any way..., hardware that has anything very special.

mangelajo: Thanks for having an eye @fzdarsky !

api/v1alpha1/openapi.yaml
```diff
@@ -775,6 +775,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/EnrollmentRequest'
+        "208":
```
hexfusion, line 778: 208 looking at the docs it seems specific to a 207 response in WebDav. Is this a non fatal conflict so 409?

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/208

mangelajo: May be 409, right. As long as the agent understands what the server means with that.

It could also simply be 200 (OK) ... because we are not creating anything new (202) and we don't really imply there is an error, and we are returning the object, and not an error-type response.

WDYT?

test/harness/test_harness.go
```diff
@@ -19,10 +19,20 @@ import (
...
 	serverListener net.Listener
 
+	// error handler for go routines
+	goRoutineErrorHandler func(error)
+
+	// internals for client context
+	cancelCtx context.CancelFunc
+
 	// attributes for the test harness
 	Context     context.Context
```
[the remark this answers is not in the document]
mangelajo: Yes, you are probably right for the client context (not the agent context which we use to stop the agent).

My intent was that we can use that context in client calls, in a way that we can kill any pending client call (even if it is under a go-routine or similar ...) when we do the harness Cleanup()

api/v1alpha1/openapi.yaml
hexfusion: anything but 500 works for me :)

hexfusion (approved):

test/harness/test_harness.go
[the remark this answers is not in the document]
hexfusion: we can work this out moving forward just an observation

api/v1alpha1/openapi.yaml
mangelajo: Let's go with that the 208 then which is meaningful for our server<-> agent context, and  I will follow up rethinking how we handle those contexts. 

Thank you!

PR #152, opened 2024-03-07
NO-ISSUE: replace gomock with mock

hexfusion:
This PR replaces the deprecated gomock with ubers mock and adds tooling for generation.

Adding new mocks is simple just add the package to the list and run `make generate`

file format: source=destination
hack/mock.list.txt

```
pkg/executer/executer.go=pkg/executer/mock_executer.go
```


mangelajo (approved):

Makefile
```diff
@@ -34,8 +34,8 @@ help:
 	@echo "    rpm/deb:         generate rpm or debian packages"
 
 generate:
-	find . -name 'mock_*.go' -type f -not -path './vendor/*' -delete
```
hexfusion, line 37: This probably should stay

hexfusion: moved into script

PR #153, opened 2024-03-08
MGMT-17169: Refactor agent cmd and config

hexfusion:
Currently the config does a poor job of evaluating required resources before runtime.  Paths are generated in the agent and fail during runtime.

This PR:
-  improves the flightctl-agent cmd aligning it better with cli. 
- adds preflight validation for required files/flags
- all configurations are now declarative
- flightctl accepts all configurations via config file.


cmd/flightctl-agent/main.go
```diff
@@ -2,74 +2,94 @@ package main
...
-	if *enrollmentServerEndpoint == "" {
-		log.Warningf("flightctl enrollment endpoint is missing, using management endpoint")
-		*enrollmentServerEndpoint = *managementServerEndpoint
+func initConfig(configFile string, log *logrus.Logger) {
+	if configFile != "" {
+		// Use config file from the flag.
+		viper.SetConfigFile(configFile)
+	} else {
+		// Use the default config path
+		viper.SetConfigFile(agent.DefaultConfigFile)
```
hexfusion, line 58: Move this into the flag

internal/agent/config.go
```diff
@@ -7,35 +7,60 @@ import (
...
 )
 
 const (
 	// DefaultFetchSpecInterval is the default interval between two reads of the remote device spec
 	DefaultFetchSpecInterval = time.Second * 60
-
 	// DefaultStatusUpdateInterval is the default interval between two status updates
 	DefaultStatusUpdateInterval = time.Second * 60
+	// DefaultDataDir is he default directory where the device's data is stored
+	DefaultDataDir = "/etc/flightctl/"
```
fzdarsky, line 20: Would `DefaultConfigDir` be more appropriate, since it's `/etc`?
Read-write data would instead go into `/var/lib/flightctl`, as ostree leaves that alone when merging.

cmd/flightctl-agent/main.go
```diff
@@ -2,74 +2,94 @@ package main
...
-	"syscall"
-	"time"
 
 	"github.com/flightctl/flightctl/internal/agent"
-	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
```
fzdarsky, line 13: At one point, we had discussed whether to remove flags (except one for config file location) and only read from config file. Are Viper et al. giving us a lot of benefit in that scenario? How much weight do we add with them?

hexfusion: Good points.

- removing flags feels strange most binaries allow both + ENV? do we have good examples of agents that don't allow flags for configuration? What is the general standard for systemd units to accept  variable params. kubelet and etcd have flags but generally take runtime params via ENV through an ENV file or directly.
- for now I think taking a dep on viper gives us progress in terms of weight we can optimize down the road. IE add the logic directly vs a dep. But is this a priority now?

```
find vendor/github.com/spf13/viper/ -type f -name "*.go" ! -name "*_test.go" -print0 | xargs -0 du -ch
4.0K    vendor/github.com/spf13/viper/file.go
4.0K    vendor/github.com/spf13/viper/file_finder.go
4.0K    vendor/github.com/spf13/viper/flags.go
4.0K    vendor/github.com/spf13/viper/logger.go
8.0K    vendor/github.com/spf13/viper/util.go
64K     vendor/github.com/spf13/viper/viper.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/decoder.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/encoder.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/error.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/dotenv/codec.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/dotenv/map_utils.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/hcl/codec.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/ini/codec.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/ini/map_utils.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/javaproperties/codec.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/javaproperties/map_utils.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/json/codec.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/toml/codec.go
4.0K    vendor/github.com/spf13/viper/internal/encoding/yaml/codec.go
4.0K    vendor/github.com/spf13/viper/internal/features/bind_struct.go
4.0K    vendor/github.com/spf13/viper/internal/features/bind_struct_default.go
148K    total
```

tl;dr 
- we should agree on functionality but perhaps this will bleed until after the demo?
- some binary optimizations will be post MVP. IE seperate module for agent, no dependencies etc

internal/agent/config.go
hexfusion: makes sense to me this was the existing value will update

cmd/flightctl-agent/main.go
fzdarsky: It's probably a matter of taste. I see config that's reasonably complex (more than a few args, esp. when config is better structured) and config that may change at runtime (on config file change or on systemd unit reload) being passed in via config file more often. Command line flags seem to be used more for configuring the systemd context in which an app runs (forking policy, pidfiles, log level, config file path) rather than app config itself.

Anyway, was comparing over-the-wire sizes of the binary (after removing debug symbols and compression), and it "only" increased by 9% or 440kB. Quite a bit, but I expected worse TBH. So, yeah, let's go ahead.

hexfusion: I think your ask is reasonable if I can do it without the dep quickly I will.

internal/agent/config.go
```diff
@@ -7,49 +7,87 @@ import (
...
 	return &Config{
-		ManagementServerEndpoint: "https://localhost:3333",
-		StatusUpdateInterval:     util.Duration(DefaultStatusUpdateInterval),
-		FetchSpecInterval:        util.Duration(DefaultFetchSpecInterval),
+		ManagementEndpoint:   DefaultManagementEndpoint,
+		EnrollmentEndpoint:   DefaultManagementEndpoint,
+		EnrollmentUIEndpoint: DefaultManagementEndpoint,
+		ConfigDir:            DefaultConfigDir,
+		DataDir:              DefaultDataDir,
+		Cacert:               filepath.Join(DefaultDataDir, DefaultCertsDirName, CacertFile),
```
mangelajo, line 84: This one must be on the config dir, not the data dir, since it's provided via image/configuration, not downloaded by the agent.

internal/agent/config.go
```diff
@@ -7,49 +7,87 @@ import (
...
+		Key:                  filepath.Join(DefaultDataDir, DefaultCertsDirName, KeyFile),
+		GeneratedCert:        filepath.Join(DefaultDataDir, DefaultCertsDirName, GeneratedCertFile),
+		EnrollmentCertFile:   filepath.Join(DefaultDataDir, DefaultCertsDirName, EnrollmentCertFile),
```
mangelajo, line 87: Same for this one

internal/agent/config.go
```diff
@@ -7,49 +7,87 @@ import (
...
+		EnrollmentKeyFile:    filepath.Join(DefaultDataDir, DefaultCertsDirName, EnrollmentKeyFile),
```
mangelajo, line 88: And this one.

internal/agent/config.go
```diff
@@ -66,13 +104,68 @@ func (cfg *Config) SetEnrollmentMetricsCallback(cb func(operation string, duract
...
+		name      string
+		checkPath bool
+	}{
+		{cfg.ManagementEndpoint, "management-endpoint", false},
+		{cfg.EnrollmentEndpoint, "enrollment-endpoint", false},
+		{cfg.GeneratedCert, "generated-cert", false},
+		{cfg.ConfigDir, "config-dir", true},
+		{cfg.DataDir, "data-dir", true},
+		{cfg.Cacert, "ca-cert", true},
+		{cfg.Key, "key", true},
```
mangelajo, line 120: must be false, this is generated by the agent.

internal/agent/config.go
```diff
@@ -66,13 +104,68 @@ func (cfg *Config) SetEnrollmentMetricsCallback(cb func(operation string, duract
...
+	requiredFields := []struct {
+		value     string
+		name      string
+		checkPath bool
+	}{
+		{cfg.ManagementEndpoint, "management-endpoint", false},
+		{cfg.EnrollmentEndpoint, "enrollment-endpoint", false},
+		{cfg.GeneratedCert, "generated-cert", false},
+		{cfg.ConfigDir, "config-dir", true},
+		{cfg.DataDir, "data-dir", true},
```
mangelajo, line 118: I would say that we could create data-dir if it does not exist. We may need to do it in .spec %post install otherwise, because /var/lib does not get from the rpm package for ostree... may be there is some inner magic to handle it though.

PR #161, opened 2024-03-12
NO-ISSUE: Do not allow setting certain fields from API

avishayt:


mangelajo:
It's failing here:

https://github.com/flightctl/flightctl/actions/runs/8247357767/job/22555308042?pr=161#step:6:85

Around here? https://github.com/flightctl/flightctl/blob/main/internal/store/enrollmentrequest_test.go#L165

Other than that it looks reasonable.

sagidayan:
LGTM

PR #162, opened 2024-03-13
NO-ISSUE: Remove deviceStore.UpdateTemplateVersionAndOwner()

avishayt:
deviceStore.CreateOrUpdate now has the option to set properties to nil. This requires explicitly selecting those fields in the update query, so we make sure to select every non-nil field plus the fields that should be nilled.


internal/service/device.go
```diff
@@ -5,10 +5,10 @@ import (
 	"fmt"
 
 	"github.com/flightctl/flightctl/internal/api/server"
+	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/store"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/go-openapi/swag"
-	"gorm.io/gorm"
```
hexfusion, line 11: :heart:

sagidayan:
LGTM

PR #163, opened 2024-03-13
NO-ISSUE: Fix default data-dir does not exist

oglok:
```
3T10:41:48Z" level=fatal msg="Error validating config: data-dir: path does not exist: /var/lib/flightctl" func=main.NewAgentCommand file="/builddir/b>
Main process exited, code=exited, status=1/FAILURE
```

Agent throws this error when using default data-dir. This should be fixed in packaging step.

### DIFF
--- a/packaging/rpm/flightctl-agent.spec
+++ b/packaging/rpm/flightctl-agent.spec
@@ -2,7 +2,7 @@
 
 Name: flightctl-agent
 Version: 0.0.1
-Release: 2%{?dist}
+Release: 3%{?dist}
 Summary: Flightctl Agent
 
 License: XXX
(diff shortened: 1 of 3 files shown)


packaging/rpm/flightctl-agent.spec
```diff
@@ -2,7 +2,7 @@
 
 Name: flightctl-agent
 Version: 0.0.1
-Release: 2%{?dist}
+Release: 2.20240313105953187195.fix_data_dir%{?dist}
```
mangelajo, line 5:
```suggestion
Release: 3%{?dist}
```

To match the changelog, I guess the other stuff is from packit.

mangelajo (changes requested):
release number

mangelajo (approved):

PR #164, opened 2024-03-13
NO-ISSUE: Move config assets to config directory

oglok:
Some assets that are not generated during runtime are expected to be under /var/lib/flightctl, which is not conceptually accurate. Changing that to config dir.


internal/agent/config.go
```diff
@@ -117,7 +117,7 @@ func (cfg *Config) Validate() error {
 		{cfg.ConfigDir, "config-dir", true},
 		{cfg.DataDir, "data-dir", true},
 		{cfg.Cacert, "ca-cert", true},
-		{cfg.Key, "key", true},
+		{cfg.Key, "key", false},
```
hexfusion, line 120: This still has me slightly confused. In the production ENV we are getting the key from what again? Today the agent will create a new keypair if it doesn't exist but that should be removed IMO. @fzdarsky has explained it but we should document it.

internal/agent/config.go
```diff
@@ -81,11 +81,11 @@ func NewDefault() *Config {
 		EnrollmentUIEndpoint: DefaultManagementEndpoint,
 		ConfigDir:            DefaultConfigDir,
 		DataDir:              DefaultDataDir,
-		Cacert:               filepath.Join(DefaultDataDir, DefaultCertsDirName, CacertFile),
+		Cacert:               filepath.Join(DefaultConfigDir, DefaultCertsDirName, CacertFile),
```
mangelajo, line 84: 👍

internal/agent/config.go
```diff
@@ -81,11 +81,11 @@ func NewDefault() *Config {
...
 		Key:                  filepath.Join(DefaultDataDir, DefaultCertsDirName, KeyFile),
 		GeneratedCert:        filepath.Join(DefaultDataDir, DefaultCertsDirName, GeneratedCertFile),
-		EnrollmentCertFile:   filepath.Join(DefaultDataDir, DefaultCertsDirName, EnrollmentCertFile),
-		EnrollmentKeyFile:    filepath.Join(DefaultDataDir, DefaultCertsDirName, EnrollmentKeyFile),
+		EnrollmentCertFile:   filepath.Join(DefaultConfigDir, DefaultCertsDirName, EnrollmentCertFile),
+		EnrollmentKeyFile:    filepath.Join(DefaultConfigDir, DefaultCertsDirName, EnrollmentKeyFile),
```
mangelajo, line 88: 👍

internal/agent/config.go
mangelajo: The agent creates the agent.key at boot, eventually requests the TPM to create a key instead..

mangelajo: So it should not exist... it's itself who will create it. I wonder how are the tests passing ':D may be because we don't call validate on the config, and just start the Agent controller from tests?

mangelajo (approved):
Looks good to me

internal/agent/config.go
hexfusion: > The agent creates the agent.key at boot, eventually requests the TPM to create a key instead..
OK maybe we should move this logic into the mock TMP.

hexfusion: > it's itself who will create it.

It makes sense that the TPM will generate it. I wasn't certain on the management of the keys.

hexfusion (approved):

PR #166, opened 2024-03-13
NO-ISSUE: internal/agent: add basic config test

hexfusion:


oglok:
I thought we were using Ginkgo for testing. Is this different for some reason?

hexfusion:
> I thought we were using Ginkgo for testing. Is this different for some reason?

we use both go tests and ginkgo

internal/agent/config_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var yamlConfig = `management-endpoint: https://management.endpoint
+enrollment-endpoint: https://management.endpoint
+enrollment-ui-endpoint: https://ui.management.endpoint
+spec-fetch-interval: 0m10s
+status-update-interval: 0m10s`
```
mangelajo, line 14: I would also check the data-dir , and the expected file paths to be in the data dir to be generated accordingly

internal/agent/config_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// ensure override
+	require.Equal("https://management.endpoint", cfg.ManagementEndpoint)
+	require.Equal("https://management.endpoint", cfg.EnrollmentEndpoint)
+	require.Equal("https://ui.management.endpoint", cfg.EnrollmentUIEndpoint)
+	require.Equal("10s", cfg.SpecFetchInterval.String())
+	require.Equal("10s", cfg.StatusUpdateInterval.String())
+
+	// ensure defaults
+	require.Equal("/etc/flightctl", cfg.ConfigDir)
```
mangelajo, line 36: Should we also test the Validate ?

internal/agent/config_test.go
oglok: the data dir (/var/lib/flightctl) and config dir should be part of the rpm spec IIUC. The cert directory inside configDir for example, could be created by the agent.

hexfusion: OK we should explicitly define what directories the config must ensure

```
EnrollmentCertFile
EnrollmentKeyFile 
GeneratedCert
Cacert
Key
```

oglok: EnrollmentCertFile, EnfollmentKeyFile and Cacert are mandatory for the agent's bootstrap. Key and GeneratedCert are created by the agent at runtime.

hexfusion: what about TPMPath

oglok: Not mandatory so far, but I defer to @fzdarsky

hexfusion: > EnrollmentCertFile, EnfollmentKeyFile and Cacert are mandatory for the agent's bootstrap. Key and GeneratedCert are created by the agent at runtime.

right but the directories must exist do we fail if they don't or do we create them. lets solve this for good :)

oglok: > EnrollmentCertFile, EnfollmentKeyFile and Cacert are mandatory for the agent's bootstrap. Key and GeneratedCert are created by the agent at runtime.

But these are not directories, but assets. Directories are only "ConfigDir and DataDir". right?

hexfusion: IE if I define Key: /some/strange/path is that an error if it doesn't exist or should it always be created. I assume error is worse in this case but then also what is an invalid path?

internal/agent/config_test.go
hexfusion: will add as followup in next PR I have some things that will help testing in spec manager PR

oglok:
LGTM

PR #167, opened 2024-03-13
NO-ISSUE: Gather client cert CN and create separate certs

mangelajo:
From the http server, every time a connection is served make sure that we gather the TLS client certificate details and put the CommonName in the context, so it's accessible later down when serving the API request.

Additionally when creating client certificates, define different common names for a client-admin key/cert, and for general agent bootstrapping certificates in a way that we can differentiate the origin of requests.

### DIFF
--- a/cmd/flightctl-server/main.go
+++ b/cmd/flightctl-server/main.go
@@ -15,9 +15,11 @@ const (
 	caCertValidityDays          = 365 * 10
 	serverCertValidityDays      = 365 * 1
 	clientBootStrapValidityDays = 365 * 1
+	adminCertValidityDays       = 365 * 1
 	signerCertName              = "ca"
 	serverCertName              = "server"
 	clientBootstrapCertName     = "client-enrollment"
+	flightctlAdminCertificate   = "flightctl-admin"
 )
 
 func main() {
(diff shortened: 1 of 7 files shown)


internal/server/server.go
```diff
@@ -92,6 +96,23 @@ func (s *Server) Run() error {
 		Handler:      router,
 		ReadTimeout:  5 * time.Second,
 		WriteTimeout: 10 * time.Second,
+		ConnContext: func(ctx context.Context, c net.Conn) context.Context {
+			tc := c.(*tls.Conn)
+			// We need to ensure TLS handshake is complete before
+			// we try to get anything useful from the ConnectionState
+			// tls delays handshanke until the first Read of Write
```
hexfusion, line 103: minor: 
```suggestion
			// tls delays handshake until the first Read of Write
```

hexfusion (approved):
I didn't test this but it looks sane. Should it have a test or ana issue to add one as followup :)

internal/server/server.go
mangelajo: ooopsies :D fixing thanks

mangelajo:
> I didn't test this, but it looks sane. Should it have a test or an issue to add one as follow-up :)

Yeah, I thought a bit about it, I couldn't figure out a sane way of testing this quickly without tons of mocking.

So far, I have tested it manually. I was hoping to test it functionally by verifying that the admin cert could do anything, but enrollment-cert only enrolls.

I will give it another thought

mangelajo:
@hexfusion I wonder if it would make sense to have some reflection endpoint or server version endpoint that would return the detected client-cert CN in the headers. That would make it very clean to test.

hexfusion:
something that only responds to localhost or such seems reasonable start for testing

cmd/flightctl-server/main.go
```diff
@@ -15,9 +15,11 @@ const (
 	caCertValidityDays          = 365 * 10
 	serverCertValidityDays      = 365 * 1
 	clientBootStrapValidityDays = 365 * 1
+	apiAdminValidityDays        = 365 * 1
 	signerCertName              = "ca"
 	serverCertName              = "server"
 	clientBootstrapCertName     = "client-enrollment"
+	apiAdminCertificate         = "api-admin"
```
fzdarsky, line 22: "api-admin" sounds strange. As if it was the API being administrated or the admin was for the API only. Maybe just "admin" or "flightctl-admin"?

internal/crypto/cert.go
```diff
@@ -276,6 +277,9 @@ func (ca *CA) MakeClientCertificate(certFile, keyFile string, subject string, ex
 		AuthorityKeyId: ca.Config.Certs[0].SubjectKeyId,
 		SubjectKeyId:   clientPublicKeyHash,
 	}
+	if subject != "" {
```
mangelajo, line 280: This is just to allow testing the behavior of the server when there is no subject on the client cert (make sure that we can't be crashed because of that, etc..)

internal/server/server_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/flightctl/flightctl/internal/crypto"
+	"github.com/flightctl/flightctl/internal/server"
+	"github.com/flightctl/flightctl/pkg/log"
+	testutil "github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const TIMEOUT = "20s"
+const POLLING = "250ms"
```
mangelajo, line 21:
```suggestion
```

PR #168, opened 2024-03-13
MGMT-17064: move bootstrap out of config 

hexfusion:
TODO:
- fix test
- add status handling for controllers (followup)
- reflect base conditions (followup)


internal/agent/device/status/status.go
```diff
@@ -10,9 +10,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
+var (
+	DeviceConditionBootstrapReason string = "Bootstrap"
+	DeviceConditionExpectedReason  string = "AsExpected"
+)
+
```
hexfusion, line 17: status still needs work

internal/agent/agent.go
```diff
@@ -100,35 +108,78 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	specManager := spec.NewManager(
+		deviceName,
+		currentSpecFilePath,
+		desiredSpecFilePath,
+		deviceWriter,
+		deviceReader,
+		managementClient,
+		backoff,
+		a.log,
+		a.config.LogPrefix,
```
hexfusion, line 160: todo we need a prefix logger :p

internal/agent/device/fileio/writer.go
```diff
@@ -67,6 +67,7 @@ func (w *Writer) WriteIgnitionFiles(files ...ign3types.File) error {
 		// if err := createOrigFile(file.Path, file.Path); err != nil {
 		// 	return err
 		// }
+		klog.Info("Writing file: ", w.rootDir+file.Path)
```
avishayt, line 70: Debug log?

avishayt (commented):
LGTM

internal/agent/device/bootstrap.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+
+	// write the management banner
+	if err := b.writeManagementBanner(); err != nil {
+		return err
+	}
+
+	// notify
```
hexfusion, line 136: need to add notify logic

internal/agent/agent.go
```diff
@@ -100,35 +108,78 @@ func (a *Agent) Run(ctx context.Context) error {
...
-	controller := config.NewController(
+	// create status manager
+	statusManager := status.NewCollector(tpmChannel, &executer.CommonExecuter{})
+
+	backoff := wait.Backoff{
+		Cap:      3 * time.Minute,
+		Duration: 10 * time.Second,
+		Factor:   1.5,
+		Steps:    24,
+	}
```
hexfusion, line 121: this all needs tuned going to add a TODO for perf eval.

mangelajo:
reviewing

internal/agent/agent.go
```diff
@@ -86,7 +94,7 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	// initialize the TPM
 	var tpmChannel *tpm.TPM
-	if len(a.config.TPMPath) > 0 {
+	if a.config.TPMPath == "" {
```
mangelajo, line 97: Why did we change this condition? shouldn't it stay as before?

mangelajo: This is another topic, but I suspect we should try to default to /dev/tpm0 or the right one... if it fails for no tpm may be the user can disable it.

hexfusion: just a style change

hexfusion: > This is another topic, but I suspect we should try to default to /dev/tpm0 or the right one... if it fails for no tpm may be the user can disable it.

Yeah I feel this is not related to this PR but a good topic.

hexfusion: > Why did we change this condition? shouldn't it stay as before?
good catch I think I was looking at adjust the logic but was not completed.

hexfusion: done

hexfusion: tracked here https://github.com/flightctl/flightctl/issues/170

internal/agent/agent.go
```diff
@@ -100,45 +108,99 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	}
+	return client.NewEnrollment(httpClient), nil
+}
+
+func newManagementClient(cfg *Config) (*client.Management, error) {
+	httpClient, err := client.NewWithResponses(cfg.ManagementEndpoint, cfg.Cacert, cfg.GeneratedCert, cfg.Key)
+	if err != nil {
+		return nil, err
+	}
+	return client.NewManagement(httpClient), nil
```
hexfusion, line 203: TODO these should get moved to client eventually

internal/agent/device/bootstrap.go
```diff
@@ -0,0 +1,293 @@
...
+	if err := b.deviceReader.CheckPathExists(b.managementGeneratedCertFile); err != nil {
+		return fmt.Errorf("generated cert: %q: %w", b.managementGeneratedCertFile, err)
+	}
+
+	// create the management client
+	managementHTTPClient, err := client.NewWithResponses(b.managementEndpoint, b.caFile, b.managementGeneratedCertFile, b.keyFile)
+	if err != nil {
+		return fmt.Errorf("create management client: %w", err)
+	}
+	managementClient := client.NewManagement(managementHTTPClient)
```
hexfusion, line 101: use create helper

mangelajo:
image
It crashes here right away (panic) I guess there is no reader now, as you said, we need to work on the test harness again at this point.

I am ok to merge and help with that tomorrow.

internal/agent/agent.go
```diff
@@ -100,45 +108,99 @@ func (a *Agent) Run(ctx context.Context) error {
...
-			a.log.Fatalf("%s: %v", a.config.LogPrefix, err)
-		}
-	}()
+	return agent.Run(ctx)
+}
 
-	<-ctx.Done()
-	return nil
+func newEnrollmentClient(cfg *Config) (*client.Enrollment, error) {
+	httpClient, err := client.NewWithResponses(cfg.EnrollmentEndpoint, cfg.Cacert, cfg.EnrollmentCertFile, cfg.Key)
```
oglok, line 191: This is going to crash due TLS private key does not match public key.

```suggestion
	httpClient, err := client.NewWithResponses(cfg.EnrollmentEndpoint, cfg.Cacert, cfg.EnrollmentCertFile, cfg.EnrollmentKeyFile)
```

oglok (changes requested):
Running the agent with this PR (after fixing the use of matching key and cert) leads to an enrollment request that cannot be approved by the server:

https://ui.flightctl.edge-devices.net/devicemanagement/enrollmentrequests/5b1e0038a7d7bca46f83201a9ddfedb214db9504855f7004a1b705e562961ad9

If you realize, the Certificate Signing Request is not shown, and approval throws the following error:

`Error 422: - {"message":"Error approving and signing enrollment request: not a valid PEM encoded block"}`

internal/agent/agent.go
hexfusion: good catch

hexfusion:
> image It crashes here right away (panic) I guess there is no reader now, as you said, we need to work on the test harness again at this point.
> I am ok to merge and help with that tomorrow.

good catch late change I didn't vet added coverage if config is not created with NewDefault()

hexfusion:
> Running the agent with this PR (after fixing the use of matching key and cert) leads to an enrollment request that cannot be approved by the server:
> 
> https://ui.flightctl.edge-devices.net/devicemanagement/enrollmentrequests/5b1e0038a7d7bca46f83201a9ddfedb214db9504855f7004a1b705e562961ad9
> 
> If you realize, the Certificate Signing Request is not shown, and approval throws the following error:
> 
> `Error 422: - {"message":"Error approving and signing enrollment request: not a valid PEM encoded block"}`

udates bootstrap to properly plumb csr

hexfusion:
Test are failing because the CSR is not being approved

internal/agent/device/bootstrap.go
```diff
@@ -0,0 +1,295 @@
...
+	if err := b.deviceWriter.WriteFile("/etc/issue.d/flightctl-banner.issue", buffer.Bytes(), os.FileMode(0666)); err != nil {
+		return fmt.Errorf("failed to write banner to disk: %w", err)
+	}
+
+	if err := SdNotify("READY=1"); err != nil {
+		b.log.Warningf("failed to notify systemd: %v", err)
+	}
+
+	// additionally print the banner into the output console
+	b.log.Info(buffer.String())
```
mangelajo, line 244: We need to print this through plain stdout, otherwise logging breaks format:

image

mangelajo:
> Test are failing because the CSR is not being approved the issue appears to be that we are waiting for the test to `enrollAndWaitForDevice` but I believe that process has changed a bit.

The first test passes, device creates enrollment,
but then it hangs on the cleanup:

image


We close the context but the agent continues looping trying to enroll even when shutdown is requested, looking at the code to figure out why.

mangelajo (changes requested):

internal/agent/device/bootstrap.go
```diff
@@ -0,0 +1,295 @@
...
+		if err := b.writeEnrollmentBanner(); err != nil {
+			return err
+		}
+
+		if err := b.enrollmentRequest(ctx); err != nil {
+			return err
+		}
+
+		b.log.Infof("%swaiting for enrollment to be approved", b.logPrefix)
+		err := wait.ExponentialBackoff(b.backoff, func() (bool, error) {
```
mangelajo, line 129: Ah, this is the reason the agent won't exit on shutdown:

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoffWithContext

We need to use ExponentialBackoffWithContext here

mangelajo:
https://github.com/mangelajo/flightctl/commit/dbe6a599d4b0f3677052fbe45e2d2f1eb0f0e9a0
This commit should fix most of the agent testing but one... I guess we need to disable that specific one and ping @avishayt for handling the new template version API in there

image

mangelajo (approved):

mangelajo:
> Running the agent with this PR (after fixing the use of matching key and cert) leads to an enrollment request that cannot be approved by the server:
> 
> https://ui.flightctl.edge-devices.net/devicemanagement/enrollmentrequests/5b1e0038a7d7bca46f83201a9ddfedb214db9504855f7004a1b705e562961ad9
> 
> If you realize, the Certificate Signing Request is not shown, and approval throws the following error:
> 
> `Error 422: - {"message":"Error approving and signing enrollment request: not a valid PEM encoded block"}`

This has been handled 👍 
 @avishayt can you merge on Sunday and look at necessary changes to re-enable the last agent test case?

PR #181, opened 2024-03-18
MGMT-17259: Agent should use specified matchPatterns

avishayt:


hexfusion (commented):
If there were 50 containers that matched these patterns what is the expected status?

api/v1alpha1/openapi.yaml
```diff
@@ -2115,11 +2115,10 @@ components:
         containers:
           type: object
           properties:
-            matchPattern:
+            matchPatterns:
               type: array
               items:
                 type: string
-                additionalProperties: true
```
mangelajo, line 2122: what does additionalProperties mean?

mangelajo: Right I was looking at the code generation now, and it didn't change... so... good cleanup. Thanks @avishayt

internal/agent/device/status/containers_test.go
```diff
@@ -126,14 +126,15 @@ var _ = Describe("containers exporter", func() {
...
 	BeforeEach(func() {
 		ctrl = gomock.NewController(GinkgoT())
-		deviceStatus = v1alpha1.DeviceStatus{}
+		deviceStatus = v1alpha1.DeviceStatus{Conditions: &[]v1alpha1.Condition{}}
 		execMock = executer.NewMockExecuter(ctrl)
 		container = newContainer(execMock)
 	})
 
 	Context("containers controller", func() {
 		It("list podman containers", func() {
```
mangelajo, line 135:
```suggestion
		It("should list podman containers", func() {
```

ginkgo idiomatic nit... I find it very nice when we use the describe/contect/it/should levels idiomatically, the errors are easier to understand.

mangelajo (changes requested):
Some tiny nitting

mangelajo (approved):
this seems good to go

Issue #180, opened 2024-03-18
Bootc image manager design


hexfusion:
### Bootc image manager draft.

This issue describes the high level functionality required for the bootc image manager.  The image manager must be able to query for status from the bootc host and provide the following functionality.

#### bootstrap phase

- check: status.bootcHost if this is set true then all further interactions with the image Manager will use bootc commands. Today we only support bootc so anything else should fail bootstrap.

- First check `spec.image.image` = `status.booted.image.image.image` if yes then we have `reconciled` image on bootc level. 
- Next If `device.current.spec.Os` == `device.desired.spec.Os` is false then we are `progressing` . This means we are in a transition state and since we are seeing this during bootstrap we should have either recently upgraded or recently failed to upgrade.
- If `reconciled` and `progressing` os image has been reconciled. replace current spec with desired.
- If `reconciled` and not `progressing` happy path no image update `bootstrap complete`
- If `reconciled` false the image update failed [retry?] this requires more thought with regards to default behavior.

#### bootstrap phase flow diagram
```mermaid
graph TD;
    A[Start: Query `bootc status`] -->|status.bootcHost set| B[Use bootc commands];
    A -->|status.bootcHost not set| X[Bootstrap Failed];
    B --> C[Verify OS Image Sanity];
    C --> D{Check spec.image.image = status.booted.image.image.image};
    D -->|Yes| E[Reconciled: Upgrade Complete];
    D -->|No| I[Degraded=true,Progressing=true.Available=false Retry?];
    E --> F{Check device.current.spec.Os == device.desired.spec.Os};
    F -->|False| G[Degraded=false,Progressing=true.Available=true];
    F -->|True| J[Degraded=false,Progressing=false.Available=true];
    G --> H[Reconciled and Progressing: Replace current with desired];
    H --> L[Degraded=false,Progressing=false.Available=true];
    L --> K
    J --> K[Reconciled and Not Progressing: Bootstrap Complete];

```

oglok:
There could be the case that `spec.image.image == status.booted.image.image.image` and still not `reconciled` because there is a newer version (upgrade) in staged mode. This state would miss a reboot to the to a reconciled state (if successful).

```
apiVersion: org.containers.bootc/v1alpha1
kind: BootcHost
metadata:
  name: host
spec:
  image:
    image: quay.io/flightctl/flightctl-agent-centos:bootstrap
    transport: registry
status:
  staged:
    image:
      image:
[...]

PR #183, opened 2024-03-18
MGMT-16600: os image controller

hexfusion:
This PR adds the first iteration of the os image controller.

Followup: add bootstrap logic #180

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -19,11 +19,12 @@ import (
 
 // Agent is responsible for managing the workloads, configuration and status of the device.
 type Agent struct {
-	name             string
-	deviceWriter     *fileio.Writer
-	statusManager    status.Manager
-	specManager      *spec.Manager
-	configController *config.Controller
+	name              string
+	deviceWriter      *fileio.Writer
+	statusManager     status.Manager
+	specManager       *spec.Manager
+	configController  *config.Controller
+	osImageController *OSImageController
 
 	fetchSpecInterval   time.Duration
 	fetchStatusInterval time.Duration
@@ -41,6 +42,7 @@ func NewAgent(
 	fetchSpecInterval time.Duration,
 	fetchStatusInterval time.Duration,
 	configController *config.Controller,
+	osImageController *OSImageController,
 	log *logrus.Logger,
 	logPrefix string,
 ) *Agent {
@@ -52,6 +54,7 @@ func NewAgent(
 		fetchSpecInterval:   fetchSpecInterval,
 		fetchStatusInterval: fetchStatusInterval,
 		configController:    configController,
+		osImageController:   osImageController,
 		log:                 log,
 		logPrefix:           logPrefix,
 	}
@@ -72,6 +75,7 @@ func (a *Agent) Run(ctx context.Context) error {
 		case <-fetchSpecTicker.C:
 			klog.V(4).Infof("%s fetching device spec", a.logPrefix)
 			if err := a.ensureDevice(ctx); err != nil {
+				// TODO handle status updates
 				a.log.Errorf("%sfailed to ensure device: %v", a.logPrefix, err)
 			}
 		case <-fetchStatusTicker.C:
@@ -89,6 +93,7 @@ func (a *Agent) Run(ctx context.Context) error {
 }
 
 func (a *Agent) ensureDevice(ctx context.Context) error {
+	// TODO: don't keep the spec constantly in memory
 	current, desired, skipSync, err := a.specManager.GetRendered(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


internal/agent/device/device.go
```diff
@@ -104,7 +109,12 @@ func (a *Agent) ensureDevice(ctx context.Context) error {
 		return fmt.Errorf("reconcile device: %w", err)
 	}
 
-	// desired is now the new current
+	if err := a.osImageController.Sync(ctx, &desired); err != nil {
+		return fmt.Errorf("sync os image: %w", err)
+	}
+
+	// write the rendered spec to disk if there is a difference
```
hexfusion, line 116: desired -> current

internal/agent/device/osimage.go
```diff
@@ -0,0 +1,90 @@
...
+		klog.V(4).Infof("%s device os image is nil", c.logPrefix)
+		return nil
+	}
+
+	host, err := c.bootc.Status(ctx)
+	if err != nil {
+		return err
+	}
+
+	if host.Status.Booted.Image.Image.Image == desired.Os.Image {
```
avishayt, line 67: > Image.Image.Image

:dizzy_face: :smile:

internal/agent/device/osimage.go
```diff
@@ -0,0 +1,90 @@
...
+		klog.V(4).Infof("Host is already running os image %s", desired.Os.Image)
+		return nil
+	}
+
+	image := desired.Os.Image
+	c.log.Infof("Switching to os image %s", image)
+	os := *desired.Os
+	if err := c.bootc.Switch(ctx, os.Image); err != nil {
```
avishayt, line 75:
```suggestion
	if err := c.bootc.Switch(ctx, image); err != nil {
```

internal/agent/device/osimage.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+
+	// Update the status to progressing
+	if err := c.statusManager.UpdateCondition(ctx, v1alpha1.DeviceProgressing, v1alpha1.ConditionStatusTrue, RebootingReason, fmt.Sprintf("Rebooting into new os image: %s", image)); err != nil {
+		return err
+	}
+
+	c.log.Infof("Os image switch complete rebooting into new image")
```
avishayt, line 84:
```suggestion
	c.log.Infof("Os image switch complete - rebooting into new image")
```

internal/agent/device/osimage.go
avishayt, line 75: How long do Switch and Apply take approximately?  Will it look like the agent is down for a long time?

internal/agent/device/osimage.go
```diff
@@ -0,0 +1,90 @@
...
+	if err := c.bootc.Apply(ctx); err != nil {
+		return err
```
avishayt, line 86: Any cleanup needed?

internal/agent/device/osimage.go
hexfusion: swithc takes a bit because it pulls down the image but this happens during runtime. It could in theory take a very long time. Apply is just a reboot I didn't see a long duration on that but it has not been calculated. We should be able to track this with metrics.

internal/agent/device/osimage.go
hexfusion: the previous commands do leave a copy of the previous image. It could be potentially feasabile to clean that up once we are stable but that would be a future optimization. As soon as Apply is triggered the device will reboot.

internal/container/container.go
```diff
@@ -116,3 +116,9 @@ func (b *BootcCmd) UsrOverlay(ctx context.Context) error {
 	}
 	return nil
 }
+
+// IsOsImageDirty returns true if the booted image does not equal the spec image.
+func IsOsImageDirty(host *BootcHost) bool {
```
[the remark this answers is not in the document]
hexfusion: not yet I think we will want these types of helpers vs inline comparisons will add usage

internal/agent/device/osimage.go
hexfusion: put it in a go routine and will log every 30s  for now

hexfusion: This will block status updates so we can consider that in a followup as well

avishayt: OK great

hexfusion:
I am going to add some mock tests then I think it is good progress for next iteration. Bootstrap handling will be a followup.

internal/agent/device/osimage.go
avishayt: I guess garbage is fine as long as (1) it doesn't interfere with subsequent attempts and (2) it eventually gets cleaned up so the filesystem doesn't fill up.  It can be a future task.

hexfusion: the image pull itself could take the entire device workload down so lots of thought needs to happen here.

hexfusion: But we need to start somewhere

hexfusion:
Merging with green Ci has known issues with status but non fatal.

PR #184, opened 2024-03-19
MGMT-17295: Deleting a fleet without selectors crashes the server

avishayt:


internal/tasks/fleet_selector_test.go
```diff
@@ -106,14 +106,29 @@ var _ = Describe("FleetSelector", func() {
 			Expect(device.Metadata.Owner).To(BeNil())
 		})
 
-		It("Fleet selector removed no overlap", func() {
+		It("Nil fleet selector matches all", func() {
```
mangelajo, line 109: We should make those 
```suggestion
		It("should match all for nil fleet selector", func() {
```

internal/tasks/fleet_selector_test.go
```diff
@@ -106,14 +106,29 @@ var _ = Describe("FleetSelector", func() {
...
+			listParams := store.ListParams{Limit: 0}
+			devices, err := deviceStore.List(ctx, orgId, listParams)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(devices.Items)).To(Equal(4))
+			for _, device := range devices.Items {
+				Expect(*device.Metadata.Owner).To(Equal("Fleet/fleet"))
+			}
+
+			callback := store.FleetStoreCallback(func(before *model.Fleet, after *model.Fleet) {})
+			err = fleetStore.Delete(ctx, orgId, callback, "fleet")
```
mangelajo, line 127:
```suggestion
			err = fleetStore.Delete(ctx, orgId, callback, "fleet")
```

I would suggest we move deletion to a separate test

It("should delete a fleet will nil selectors without errors")

so if something goes wrong with deletion, we have  more granular test output

internal/tasks/fleet_selector_test.go
avishayt: Why?  This description is more difficult for me to understand.
How about "Nil fleet selector should match all devices"?

internal/tasks/fleet_selector_test.go
avishayt: Done

internal/tasks/fleet_selector_test.go
avishayt: Nice reference, now I understand what you meant.  I'll do an overhaul in a separate PR.

mangelajo:
As we commented on slack, let's merge, and then we can do a more ginkgo/idiomatic overhaul

PR #186, opened 2024-03-19
NO-ISSUE: Create an agent config we can use

mangelajo:
This exposes the service over the local address of our ethernet/wifi interface, by finding the exit route IP.

Then it prepares this configuration along with the certificates into the bin/agent/ directory for later image embedding.


hexfusion (approved):
didn't test this but the over all idea makes sense ok to merge as progress assuming you tested it :)

PR #187, opened 2024-03-19
MGMT-16600: fix status updates

hexfusion:
Update status does not take a DeviceStatus but a Device with a Status. This gets resolved by using static types vs bytes,

Manually verified tracking new test. https://github.com/flightctl/flightctl/issues/189


mangelajo (commented):
:+1

PR #192, opened 2024-03-20
NO-ISSUE: unit tests for osimage

avishayt:


internal/agent/device/osimage.go
```diff
@@ -44,13 +44,12 @@ func (c *OSImageController) Sync(ctx context.Context, desired *v1alpha1.Rendered
 
 	err := c.ensureImage(ctx, desired)
 	if err != nil {
-		if err := c.statusManager.UpdateConditionError(ctx, OsImageDegradedReason, err); err != nil {
-			klog.Errorf("Failed to update condition: %v", err)
+		if updateErr := c.statusManager.UpdateConditionError(ctx, OsImageDegradedReason, err); updateErr != nil {
+			klog.Errorf("Failed to update condition: %v", updateErr)
```
avishayt, line 48: I made this change because the way it is, we're returning an error only if we fail setting the condition.

internal/agent/device/osimage.go
```diff
@@ -44,13 +44,12 @@ func (c *OSImageController) Sync(ctx context.Context, desired *v1alpha1.Rendered
...
 		}
-		return err
```
hexfusion, line 50: I think this should stay and we return nil in terms of scope

avishayt: Right, fixed

hexfusion (approved):
thanks !

PR #194, opened 2024-03-20
NO-ISSUE: Check booted image on bootstrap and update current spec

avishayt:


internal/agent/device/bootstrap.go
```diff
@@ -110,6 +119,31 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
 	return nil
 }
 
+func (b *Bootstrap) ensureCurrentRenderedSpecUpToDate(ctx context.Context) error {
+	desiredSpec, err := spec.ReadRenderedSpecFromFile(b.deviceReader, b.desiredRenderedFile)
+	if err != nil {
+		if err == spec.ErrMissingRenderedSpec {
```
hexfusion, line 125: lets add an explicit note why this is ok.

```
// during the initial bootstrap it is expected that the spec does not exist yet.  this assumption is validated later.
```

Also ensure error is not wrapped 
```
if errors.Is(err, spec.ErrMissingRenderedSpec) {
```

internal/agent/device/spec/spec.go
```diff
@@ -266,7 +242,29 @@ func EnsureDesiredRenderedSpec(
 	return desired, nil
 }
 
-func writeRenderedSpecToFile(writer *fileio.Writer, rendered *v1alpha1.RenderedDeviceSpec, filePath string) error {
+func ReadRenderedSpecFromFile(
+	reader *fileio.Reader,
+	filePath string,
+) (v1alpha1.RenderedDeviceSpec, error) {
+	var current v1alpha1.RenderedDeviceSpec
+	renderedBytes, err := reader.ReadFile(filePath)
```
hexfusion, line 250: lets add a TODO to use buffered read. these are perf optimizations but we should be tracking them as these files could get fairly large potentially. I know your just moving it but its on my list.

internal/agent/device/bootstrap.go
```diff
@@ -110,6 +119,31 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
...
+	}
+
+	bootcCmd := container.NewBootcCmd(b.executer)
+	bootcHost, err := bootcCmd.Status(ctx)
+	if err != nil {
+		return fmt.Errorf("getting current bootc status: %w", err)
+	}
+
+	if container.IsOsImageReconciled(bootcHost, &desiredSpec) {
+		err = spec.WriteRenderedSpecToFile(b.deviceWriter, &desiredSpec, b.currentRenderedFile)
```
hexfusion, line 138: lets log this as info

internal/agent/device/spec/spec_test.go
```diff
@@ -94,7 +94,7 @@ func TestManager(t *testing.T) {
 				return
 			}
 			if tt.wantErr != nil {
-				require.ErrorAs(err, &tt.wantErr)
+				require.ErrorAs(err, &tests[tIndex].wantErr)
```
hexfusion, line 97: `tt` does this for you

internal/agent/device/spec/spec.go
```diff
@@ -180,15 +161,13 @@ func EnsureCurrentRenderedSpec(
...
 	reader *fileio.Reader,
-	deviceName,
 	filePath string,
 ) (v1alpha1.RenderedDeviceSpec, error) {
-	var current v1alpha1.RenderedDeviceSpec
-	renderedBytes, err := os.ReadFile(filePath)
+	current, err := ReadRenderedSpecFromFile(reader, filePath)
 	if err != nil {
-		if os.IsNotExist(err) {
+		if err == ErrMissingRenderedSpec {
```
hexfusion, line 168: ensure not wrapped errors.Is..

internal/agent/device/bootstrap.go
hexfusion: we want container.IsOsImageReconciled and something like below that signals spec is stale
```
func IsOsImageInTransition(current *v1alpha1.RenderedDeviceSpec, desired * v1alpha1.RenderedDeviceSpec ) bool {
 return current.Spec.Os.Image != desired.Spec.Os.Image
}
```

hexfusion (commented):
great start couple things.
- only read files once
- buffered reads for perf
- ensure docs are clear why order is important.

internal/agent/device/bootstrap.go
hexfusion: lets log a  `Warn`  `if container.IsOsImageReconciled` is not true. this means the device was rebooted before  Apply in which case it will reconcile later. Or worse we failed to upgrade. But right now we have not understanding of what to do so we continue.

internal/agent/device/spec/spec.go
```diff
@@ -266,7 +242,29 @@ func EnsureDesiredRenderedSpec(
...
+
+	// read bytes from file
+	if err := json.Unmarshal(renderedBytes, &current); err != nil {
+		return v1alpha1.RenderedDeviceSpec{}, fmt.Errorf("unmarshal device specification: %w", err)
+	}
+
+	return current, nil
+}
+
+func WriteRenderedSpecToFile(writer *fileio.Writer, rendered *v1alpha1.RenderedDeviceSpec, filePath string) error {
```
hexfusion, line 267: with the perf improvements in #195 lets use writer.WriteFile() here

internal/agent/device/spec/spec_test.go
avishayt: There was a lint warning about passing a pointer to the loop variable

internal/agent/device/bootstrap.go
```diff
@@ -110,6 +120,68 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
 	return nil
 }
 
+func (b *Bootstrap) EnsureCurrentRenderedSpecUpToDate(ctx context.Context) error {
```
hexfusion, line 123: just curious why you want to export these the bootstrap object will only ever call Initialize I would like to keep these local until we have an explicit usecase IE integration tests requires etc. I just don't want to give the impression that bootstrap is portable.

internal/agent/device/bootstrap.go
```diff
@@ -110,6 +120,68 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
...
+	} else {
+		// We rebooted without applying the new OS image - something went wrong
+		b.log.Warnf("%sstarted bootstrap with OS image not equal to desired image", b.logPrefix)
+		condErr := fmt.Errorf("booted image %s, expected %s", container.GetImage(bootcHost), desiredSpec.Os.Image)
+		err = b.statusManager.UpdateConditionError(ctx, BootedWithUnexpectedImage, condErr)
+		if err != nil {
+			b.log.Warnf("%sfailed setting error condition: %v", b.logPrefix, err)
+			return fmt.Errorf("setting error condition: %w", err)
+		}
+		return fmt.Errorf("os image was not reconciled")
```
hexfusion, line 167: would this fail bootstrap? we just need `Warn` via conditions/log but it is a possible condition that can be reconciled.

internal/agent/device/bootstrap.go
```diff
@@ -110,6 +120,68 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
...
+			return fmt.Errorf("writing rendered spec to file: %w", err)
+		}
+	} else {
+		// We rebooted without applying the new OS image - something went wrong
+		b.log.Warnf("%sstarted bootstrap with OS image not equal to desired image", b.logPrefix)
+		condErr := fmt.Errorf("booted image %s, expected %s", container.GetImage(bootcHost), desiredSpec.Os.Image)
+		err = b.statusManager.UpdateConditionError(ctx, BootedWithUnexpectedImage, condErr)
+		if err != nil {
+			b.log.Warnf("%sfailed setting error condition: %v", b.logPrefix, err)
+			return fmt.Errorf("setting error condition: %w", err)
```
hexfusion, line 165: we don't want to fail bootstrap for this just warn so don't return error here

internal/agent/device/spec/spec_test.go
hexfusion: I think we just want ErrorIs so we don't need a ref

internal/agent/device/boostrap_test.go
```diff
@@ -117,6 +128,195 @@ func TestEnsureEnrollment(t *testing.T) {
...
+			err := os.WriteFile(filepath.Join(tmpDir, desiredSpecFilePath), defaultRenderedData, 0600)
+			Expect(err).ToNot(HaveOccurred())
+			err = bootstrap.EnsureCurrentRenderedSpecUpToDate(ctx)
+			Expect(err).ToNot(HaveOccurred())
+		})
+	})
+
+	Context("When the desired spec does not exist", func() {
+		It("should return with no action", func() {
+			err := os.WriteFile(filepath.Join(tmpDir, currentSpecFilePath), defaultRenderedData, 0600)
```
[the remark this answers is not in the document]
avishayt: Done

internal/agent/device/bootstrap.go
avishayt: I wanted the unit tests to be in a separate package, but you're right.  I moved the unit tests back to the same package.

internal/agent/device/bootstrap.go
avishayt: Done

internal/agent/device/bootstrap.go
avishayt: done

internal/agent/device/bootstrap.go
avishayt: done

internal/agent/device/spec/spec_test.go
avishayt: Right, removed the ref.

mangelajo (approved):

mangelajo (approved):

PR #195, opened 2024-03-21
NO-ISSUE: improve fileio writer

hexfusion:
This PR improves the performance of fileio  atomic writer directly instead of converting files to and from ignition.

### DIFF
--- a/internal/agent/device/fileio/writer.go
+++ b/internal/agent/device/fileio/writer.go
@@ -76,9 +76,11 @@ func (w *Writer) WriteIgnitionFiles(files ...ign3types.File) error {
 
 // WriteFile writes the provided data to the file at the path with the provided permissions
 func (w *Writer) WriteFile(name string, data []byte, perm fs.FileMode) error {
-	// TODO: rethink how we are persisting files
-	// convert to ign file so we can use the atomic writer we can do this more directly in future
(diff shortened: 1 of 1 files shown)


internal/agent/device/fileio/reader.go
```diff
@@ -30,7 +32,27 @@ func (r *Reader) PathFor(filePath string) string {
...
+	file, err := os.Open(r.PathFor(filePath))
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	reader := bufio.NewReader(file)
+	var data []byte
+	for {
+		buf := make([]byte, 1024) // TODO: benchmark and optimize buffer size
```
avishayt, line 44: I'm probably missing some information about why this is better.
1. Allocating a small buffer, reading a small chunk, and copying it into a larger buffer can't be faster than reading the whole thing in one shot (more system calls, data copies, etc).
2. In either case, we are allocating a buffer containing the full contents of the file (`data` here), so how does it reduce RSS?

hexfusion: haha yeah I will add benchmarks before re add. creating the buffer in the for loop is a bug anyways.

avishayt (approved):

PR #197, opened 2024-03-24
NO-ISSUE: Decouple agent integration test from unit tests

sagidayan:
The agent tests are not really unit-tests, so moved them to integration.

Added new make target `integration-test`


.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -y # fix broken repo cache
+          sudo apt install -y make golang-1.20 python3-pip
+          go install gotest.tools/gotestsum@latest
+          go get -u github.com/proglottis/gpgme
+
+      - name: Deploy DB with helm and kind
+        run: make deploy-db-helm
```
mangelajo, line 30:
```suggestion
      - name: Setup all dependencies
        uses: ./.github/actions/setup-dependencies
        with:
          setup_podman4: yes
```

I refactored it in another patch.

integration_tests/agent_test.go
```diff
@@ -1,4 +1,4 @@
-package agent_test
```
mangelajo, line 1: Can we move it to /tests/mock-e2e/agent/  in alignment with https://github.com/flightctl/flightctl/pull/204/files if that one makes sense to you?

There, I define two make targets, but so far, we only have one, so it's good to just keep it as integration-test and run-integration test. The dirs would make it easier for me to add the non-mock target.

mangelajo (changes requested):
Some comments to track https://github.com/flightctl/flightctl/pull/204/files

integration_tests/agent_test.go
sagidayan: Sure, Ill wait until #204 is merged

tests/mock-e2e/agent/agent_test.go
```diff
@@ -1,4 +1,4 @@
-package agent_test
+package agent
```
hexfusion, line 1: I don't see this move as productive, this test was created to begin the process of adding more tests. To better understand how the components work together. Moving our main agent test out of agent as the only integration test feels premature.

hexfusion (changes requested):
I would like to see us agree on a reasonable MVP testing methodology with the hopes of evolving over time.

tests/mock-e2e/agent/agent_test.go
mangelajo: Ok, as per https://github.com/flightctl/flightctl/tree/main/test strategy, the final directory would be /test/integration/agent

mangelajo (approved):
Looks good 👍

PR #203, opened 2024-03-25
MGMT-17322: Set condition on fleet when created/updated with non-exis…

avishayt:
…tent git repo

Validate the fleet config before creating a TemplateVersion. The same task will be called in the future when repository resources are updated.


internal/tasks/templateversion.go
```diff
@@ -65,15 +61,113 @@ type TemplateVersionLogic struct {
...
+
+	condition := api.Condition{Type: api.FleetValid}
+	var retErr error
+	if len(invalidConfigs) == 0 {
+		condition.Status = api.ConditionStatusTrue
+		condition.Reason = util.StrToPtr("Valid")
+		retErr = nil
+	} else {
+		condition.Status = api.ConditionStatusFalse
+		condition.Reason = util.StrToPtr("Invalid")
```
mangelajo, line 112: Shouldn't this be just Valid true/false ?

Valid + true
Invalid + false

looks inverting, I am looking at k8s api and will ping back here.

I wonder if it makes more sense to keep:

Valid true
Valid false

or 

Valid true
Invalid true 
?

internal/tasks/templateversion.go
```diff
@@ -65,15 +61,113 @@ type TemplateVersionLogic struct {
...
+		condition.Message = util.StrToPtr(fmt.Sprintf("Fleet has %d invalid configurations: %s", len(invalidConfigs), strings.Join(invalidConfigs, ", ")))
```
mangelajo, line 113: 👍

internal/tasks/templateversion_test.go
```diff
@@ -0,0 +1,215 @@
...
+		badInlineConfig.Inline = badInline
+
+		callback = store.FleetStoreCallback(func(before *model.Fleet, after *model.Fleet) {})
+	})
+
+	AfterEach(func() {
+		store.DeleteTestDB(cfg, storeInst, dbName)
+	})
+
+	Context("When a Fleet has a valid configuration", func() {
```
mangelajo, line 108:
```suggestion
	When("a Fleet has a valid configuration", func() {
```

internal/tasks/templateversion_test.go
```diff
@@ -0,0 +1,215 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(fleet.Status.Conditions).ToNot(BeNil())
+			Expect(*fleet.Status.Conditions).To(HaveLen(1))
+			Expect((*fleet.Status.Conditions)[0].Type).To(Equal(api.FleetValid))
+			Expect((*fleet.Status.Conditions)[0].Status).To(Equal(api.ConditionStatusTrue))
+		})
+	})
+
+	Context("When a Fleet has an invalid git configuration", func() {
```
mangelajo, line 144:
```suggestion
	When("a Fleet has an invalid git configuration", func() {
```

internal/tasks/templateversion_test.go
```diff
@@ -0,0 +1,215 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(fleet.Status.Conditions).ToNot(BeNil())
+			Expect(*fleet.Status.Conditions).To(HaveLen(1))
+			Expect((*fleet.Status.Conditions)[0].Type).To(Equal(api.FleetValid))
+			Expect((*fleet.Status.Conditions)[0].Status).To(Equal(api.ConditionStatusFalse))
+		})
+	})
+
+	Context("When a Fleet has an invalid inline configuration", func() {
```
mangelajo, line 180:
```suggestion
	When("a Fleet has an invalid inline configuration", func() {
```

mangelajo (changes requested):

internal/tasks/templateversion.go
mangelajo: Ignore this comment, I was looking wrong at Reason/Type

your logic here is good and correct.

mangelajo (commented):
Thanks for addressing the changes so quickly

PR #210, opened 2024-03-27
MGMT-17323: Update fleet condition upon repository update

avishayt:
Two commits:
    MGMT-17323: Update fleet condition upon repository update
    
    When a repository is created/updated/deleted, update the Condition of
    any Fleet that references it. If the Fleet is valid, create a new
    TemplateVersion.

    NO-ISSUE: Better organize async tasks
    
    Rename tasks to <resource>-<verb> for consistency.
    Split "templateversion" task into two separate tasks since they were
    essentially unrelated.

### DIFF
--- a/internal/store/repository_test.go
+++ b/internal/store/repository_test.go
@@ -30,7 +30,8 @@ func createRepositories(numRepositories int, ctx context.Context, storeInst stor
 			},
 		}
 
-		_, err := storeInst.Repository().Create(ctx, orgId, &resource)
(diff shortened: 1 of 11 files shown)


internal/store/repository_test.go
sagidayan, line 1: Do we have a test that checks if the `callback` is being called or not? (if there is an error, it should not be called)

mangelajo (approved):

PR #217, opened 2024-04-01
NO-ISSUE: Change repo polling intervals from 2 minutes to 2 seconds

avishayt:
In the future this should be moved into the config.


oglok:
Maybe 10-15 seconds to avoid rate block from GH?

PR #219, opened 2024-04-01
NO-ISSUE: do not pass templateVersion during sync 

hexfusion:
During sync is the expectation that the templateVersion is empty to receive the latest?

cc @avishayt

### DIFF
--- a/internal/service/device.go
+++ b/internal/service/device.go
@@ -195,7 +195,6 @@ func (h *ServiceHandler) ReplaceDeviceStatus(ctx context.Context, request server
 	}
 }
 
-
 // GetRenderedDeviceSpec returns the rendered device spec for the given device
 // name. If the expectation is to receive the latest rendered spec the last
 // known template version param should be nil.


internal/agent/device/spec/spec.go
```diff
@@ -78,7 +80,7 @@ func (s *Manager) GetRendered(ctx context.Context) (v1alpha1.RenderedDeviceSpec,
 		return v1alpha1.RenderedDeviceSpec{}, v1alpha1.RenderedDeviceSpec{}, false, err
 	}
 
-	desired, skipSync, err := s.getDesiredRenderedSpec(ctx, current.Owner, current.TemplateVersion)
+	desired, skipSync, err := s.getDesiredRenderedSpec(ctx, current.Owner, EmptyTemplateVersion)
```
hexfusion, line 83: If I have the current and its empty during bootstrap then it should return the latest templateVersion? Then moving forward we continue each time we will pass the last known template version. So it would be OK?

hexfusion:
This would not make sense as it would be very inefficient to ask for latest every time.

PR #220, opened 2024-04-02
NO-ISSUE: Agent and server log level defined in config

sagidayan:
Defaults to "info"

### DIFF
--- a/cmd/flightctl-agent/main.go
+++ b/cmd/flightctl-agent/main.go
@@ -47,6 +47,12 @@ func NewAgentCommand() *agentCmd {
 		a.log.Fatalf("Error validating config: %v", err)
 	}
 
+	logLvl, err := logrus.ParseLevel(a.config.LogLevel)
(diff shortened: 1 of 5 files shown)


cmd/flightctl-agent/main.go
```diff
@@ -47,6 +47,12 @@ func NewAgentCommand() *agentCmd {
 		a.log.Fatalf("Error validating config: %v", err)
 	}
 
+	logLvl, err := logrus.ParseLevel(a.config.LogLevel)
+	if err != nil {
+		logLvl = logrus.ErrorLevel
```
hexfusion, line 52: Thanks for putting up this PR just curious why not info?

sagidayan: You are right, Info is the default we would want. Updated

mangelajo (approved):
Thank you @sagidayan this looks good to me.

PR #222, opened 2024-04-03
NO-ISSUE: Create doc about async service tasks

avishayt:
Also fixed a bug that deleting a device should trigger the task that updates device ownership (important in case there are overlapping selectors, so the Fleet's Conditions can be updated if fixed).

### DIFF
--- a/docs/architecture/tasks.md
+++ b/docs/architecture/tasks.md
@@ -1,0 +1,14 @@
+|          **Trigger**         |                   **Task**                  |
+|------------------------------|---------------------------------------------|
+| Fleet template updated       | Create new TemplateVersion                  |
+| Fleet created                | Create new TemplateVersion                  |
+| TemplateVersion created      | Populate the TemplateVersion's Status       |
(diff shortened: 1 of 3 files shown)


docs/architecture/tasks.md
```diff
@@ -0,0 +1,14 @@
+|          **Trigger**         |                   **Task**                  |
+|------------------------------|---------------------------------------------|
+| Fleet template updated       | Create new TemplateVersion                  |
```
mangelajo, line 3: I wonder if we could have this in code instead. A table that maps the triggers into the tasks, where tasks repeat sometimes for different triggers.

avishayt:
I'd rather not invest in tests and code for this file because it will be deleted once we use a real task system.

oglok:
maybe we should stop doing self-merges? :-)

mangelajo:
> I'd rather not invest in tests and code for this file because it will be deleted once we use a real task system.

makes sense

PR #228, opened 2024-04-09
MGMT-17326: ResourceSync properties are required

sagidayan:


mangelajo (approved):
It makes sense, those fields are never optional

PR #230, opened 2024-04-11
NO-ISSUE: Adding support for CRIO container status

oglok:
Adding support to report CRIO containers status


internal/agent/device/status/containers.go
```diff
@@ -98,6 +115,83 @@ func (c *Container) Export(ctx context.Context, status *v1alpha1.DeviceStatus) e
...
+	if _, err := exec.LookPath("crictl"); err == nil {
+		err := c.CrioExport(ctx, status)
+		if err != nil {
+			return fmt.Errorf("failed exporting crio status: %w", err)
+		}
+	}
+
+	return nil
+}
+
```
mangelajo, line 194: Looks good, could we extend unit tests in here to verify the crictl output parsing? https://github.com/flightctl/flightctl/blob/main/internal/agent/device/status/containers_test.go

I guess, we could also test this in E2E (later, not speaking fo this patch):

1) We add crio and podman to the image
2) When the agent has booted, we go behind the curtains, and with ssh we spawn a podman container... and verify that it's reported
3) Then we spawn a crio container... and wait it to be reported.

:-D

hexfusion (commented):
generally looks good a few questions.

internal/agent/device/status/containers.go
```diff
@@ -98,6 +115,83 @@ func (c *Container) Export(ctx context.Context, status *v1alpha1.DeviceStatus) e
...
+		err := c.PodmanExport(ctx, status)
+		if err != nil {
+			return fmt.Errorf("failed exporting podman status: %w", err)
+		}
+	}
+
+	if _, err := exec.LookPath("crictl"); err == nil {
+		err := c.CrioExport(ctx, status)
+		if err != nil {
+			return fmt.Errorf("failed exporting crio status: %w", err)
```
hexfusion, line 188: is this an error or should we warn? this would mean we require podman and crictl to report status.

hexfusion:
Another question is since workloads are declarative and thus deterministic. Do we need to report a container as running or do we just need to report when something is wrong? I think we don't need to make that decision now but we should consider it.

internal/agent/device/status/containers.go
```diff
@@ -16,11 +17,14 @@ const (
 	podmanCommand          = "/usr/bin/podman"
 	podmanCommandTimeout   = 2 * time.Minute
 	podmanContainerRunning = "running"
+	crioCommand            = "/usr/bin/crictl"
+	crioCommandTimeout     = 2 * time.Minute
```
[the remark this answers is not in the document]
oglok: I copied the 2 minutes existing timer from podman, but it's true, I'll change both of them.

internal/agent/device/status/containers.go
```diff
@@ -98,6 +115,83 @@ func (c *Container) Export(ctx context.Context, status *v1alpha1.DeviceStatus) e
...
+	runningCondition := v1alpha1.Condition{
+		Type: v1alpha1.DeviceContainersRunning,
+	}
+
+	deviceContainerStatus := make([]v1alpha1.ContainerStatus, len(containers.Containers))
+	for i, c := range containers.Containers {
+		deviceContainerStatus[i].Name = c.Metadata.Name
+		deviceContainerStatus[i].Status = c.State
+		deviceContainerStatus[i].Image = c.Image
+		deviceContainerStatus[i].Id = c.Id
```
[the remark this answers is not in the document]
oglok: so I guess if you want to monitor a running application or container, you might wanna know wether it's running or not, and the actual image it's using. ID could be only worth for troubleshooting. I'm good at reconsidering these params, but that's an API change to agree on.

internal/agent/device/status/containers.go
```diff
@@ -98,6 +115,83 @@ func (c *Container) Export(ctx context.Context, status *v1alpha1.DeviceStatus) e
...
+		runningCondition.Status = v1alpha1.ConditionStatusTrue
+		runningCondition.Reason = util.StrToPtr("Running")
+	} else {
+		runningCondition.Status = v1alpha1.ConditionStatusFalse
+		runningCondition.Reason = util.StrToPtr("NotRunning")
+		containerStr := "container"
+		if notRunning > 1 {
+			containerStr = "containers"
+		}
+		runningCondition.Message = util.StrToPtr(fmt.Sprintf("%d %s not running", notRunning, containerStr))
```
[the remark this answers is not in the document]
oglok: @hexfusion do you mean something like this?

```
		runningCondition.Message = util.StrToPtr(fmt.Sprintf("%d %s not running: %s", notRunning, containerStr, crioErrOut))
```

Not sure what could help for debugging, as we're only reporting the status and not getting into reporting logs of the container engine (or that's my intention xD)

internal/agent/device/status/containers.go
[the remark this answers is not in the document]
hexfusion: Just stuff to consider as all bytes count. We can always revise in the future.

hexfusion (dismissed):

oglok:
Writing unit tests. Please do not merge yet!

internal/agent/device/status/containers.go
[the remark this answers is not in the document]
mangelajo: Yes, changing the API itself would be a different topic worth exploring, but we must do it together with the rest of the team :)

internal/agent/device/status/containers_test.go
```diff
@@ -144,10 +195,41 @@ var _ = Describe("containers exporter", func() {
...
+			Expect((*deviceStatus.Containers)[0].Id).To(Equal("id1"))
+			Expect((*deviceStatus.Containers)[0].Image).To(Equal("docker.io/library/alpine@sha256:c4a262d530f57d1b7b68b52ba8383c2e55fd1a0cb5b4f46b11eed7a2c4e143da"))
+			Expect((*deviceStatus.Containers)[0].Name).To(Equal("alpine"))
+			Expect((*deviceStatus.Containers)[0].Status).To(Equal("CONTAINER_RUNNING"))
+			Expect((*deviceStatus.Containers)[0].Engine).To(Equal("crio"))
+			Expect((*deviceStatus.Containers)[1].Id).To(Equal("id2"))
+			Expect((*deviceStatus.Containers)[1].Image).To(Equal("docker.io/library/busybox@sha256:4be429a5fbb2e71ae7958bfa558bc637cf3a61baf40a708cb8fff532b39e52d0"))
+			Expect((*deviceStatus.Containers)[1].Name).To(Equal("busybox"))
+			Expect((*deviceStatus.Containers)[1].Status).To(Equal("CONTAINER_EXITED"))
+			Expect((*deviceStatus.Containers)[1].Engine).To(Equal("crio"))
```
mangelajo, line 232: I would add another test at least that runs with both podman and crio and verify that we have 4 containers.

This would handle if someone for example modifies your current code order (podman/crio...) where podman creates, and crio adds... which could get broken otherwise.

We can add it as a follow up too. For example we need to see how to get gomock to expect one call then another... same for lookpath.

mangelajo (approved):
I would follow up extending with the new test case if we can

mangelajo:
The lint error should go away with a rebase.

internal/agent/device/status/containers_test.go
oglok: ok, I can do a followup PR with that. I need to learn a bit more about Ginkgo and Gomock to be honest :-)

mangelajo (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -1858,6 +1859,9 @@ components:
         image:
           type: string
           description: Image of the container.
+        engine:
+          type: string
```
hexfusion, line 1863: since this is declared why do we need to define it per container? If we must have an engine type I would expect the actual containers to be inside of it.
```
status:
  workload:
    -  engine: Podman
      <workload status>
```

hexfusion: We can revist down the road I guess.

hexfusion (approved):

api/v1alpha1/openapi.yaml
mangelajo: You could have some running as part of crio, and some running as part of podman. 

We could may be optimize by having two different arrays?

hexfusion: right, that is what I described above an array of engines each with its own containers vs each container describing its engine

oglok:
All commits are now signed. Thanks for reviewing @hexfusion @mangelajo

PR #231, opened 2024-04-11
NO-ISSUE: E2E Full harness (with API) + test for checking TPM status reporting

mangelajo:
This PR extends the basic E2E harness including the admin API
client, and moves the two e2e agent tests to the new harness

As an example where the API is required introduces a new test that:

* Verifies the TPM reporting from within the VM

This also switches the runner for E2E to the ginkgo binary which
produces better reporting console output as well as better junit output.

### DIFF
--- a/test/e2e/agent_test.go
+++ b/test/e2e/agent_test.go
@@ -2,18 +2,18 @@ package agent_test
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 	"testing"
 
-	"github.com/flightctl/flightctl/test/harness/e2e/vm"
-	"github.com/google/uuid"
+	"github.com/flightctl/flightctl/internal/api/client"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	testutil "github.com/flightctl/flightctl/test/util"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
 )
 
-const TIMEOUT = "30s"
+const TIMEOUT = "1m"
 const POLLING = "250ms"
 
 func TestAgent(t *testing.T) {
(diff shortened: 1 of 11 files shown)


test/harness/e2e/vm/domain-template.xml
```diff
@@ -14,10 +14,13 @@
   <on_poweroff>destroy</on_poweroff>
   <on_reboot>restart</on_reboot>
   <on_crash>destroy</on_crash>
-  <os>
+  <os firmware='efi'>
     <type machine='q35'>hvm</type>
     <boot dev='hd'/>
-
+    <firmware>
+      <feature enabled='yes' name='secure-boot'/>
```
mangelajo, line 21: 😎

test/harness/e2e/vm/domain-template.xml
```diff
@@ -14,10 +14,13 @@
...
+      <feature enabled='no' name='enrolled-keys'/>
```
mangelajo, line 22: I wonder if this could be "yes" ... checking

mangelajo: oh it works, nice :)

test/harness/e2e/vm/domain-template.xml
mangelajo: Hmmm not sure if this is making it fail to boot on github CI :-( checking

mangelajo: 4 days later... https://github.com/flightctl/flightctl/pull/231/files#diff-29fa4e9e9dfff8f099fe2fdc77e1e702165851b9a835991d8d4a1f7cd14a71beR17

test/harness/e2e/vm/domain-template.xml
```diff
@@ -26,7 +34,7 @@
       </target>
       <alias name='serial0'/>
     </serial>
-    <console type='pty' tty='/dev/pts/2'>
+    <console type='pty'>
```
mangelajo, line 37: fixing typo, this shouldn't be specific about the pts... libvirt (or qemu) will do auto-allocation

test/test.mk
```diff
@@ -19,16 +18,14 @@ endif
...
 
 _integration_test: $(REPORTS)
 	gotestsum $(GO_TEST_E2E_FLAGS) -- $(GO_INTEGRATIONTEST_FLAGS) -timeout $(TIMEOUT) || ($(MAKE) _collect_junit && /bin/false)
 	$(MAKE) _collect_junit
 
 _e2e_test: $(REPORTS)
 	sudo chown $(shell whoami):$(shell whoami) -R bin/output
-	gotestsum $(GO_TEST_E2E_FLAGS) -- $(GO_E2E_FLAGS) -timeout $(TIMEOUT) || ($(MAKE) _collect_junit && /bin/false)
-	$(MAKE) _collect_junit
+	ginkgo run --timeout 30m --race -vv --junit-report $(REPORTS)/junit_e2e_test.xml --github-output $(GO_E2E_DIRS)
```
mangelajo, line 28: I switched to ginkgo runner because it provides better output of what's going on, and better junit output too, we may consider also switching the integration test output for the same reason.

test/harness/e2e/vm/vm_linux.go
```diff
@@ -299,3 +305,17 @@ func (v *VMInLibvirt) IsRunning() (exists bool, err error) {
 		return false, nil
 	}
 }
+
+func (v *VMInLibvirt) RunAndWaitForSSH() error {
```
mangelajo, line 309: Not explicitly required for this patch. But when I expect that some tests will require running the VM without waiting to SSH.

test/harness/e2e/vm/vm_linux.go
```diff
@@ -113,7 +125,9 @@ func (v *VMInLibvirt) Run() error {
 func (v *VMInLibvirt) ReadConsole() string {
 	v.consoleMutex.Lock()
 	defer v.consoleMutex.Unlock()
-	return v.consoleOutput.String()
+	str := v.consoleOutput.String()
+	v.consoleOutput.Reset()
```
mangelajo, line 129: We need to clear the buffer after reading.

mangelajo:
> https://github.com/flightctl/flightctl/actions/runs/8708012675/job/23884384875?pr=231#step:9:4336

Thanks, yes... 
https://github.com/flightctl/flightctl/pull/231/files#diff-08d1d54287572446320e640945c840e8b94aebdad92c871d5e344270fe2a0d4dR128
I had to clear a buffer after reading ... otherwise depending on timing, the tests waiting for specific console output would get confused

test/e2e/agent_test.go
```diff
@@ -2,18 +2,18 @@ package agent_test
...
+	"github.com/flightctl/flightctl/internal/api/client"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	testutil "github.com/flightctl/flightctl/test/util"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
 )
 
-const TIMEOUT = "30s"
+const TIMEOUT = "1m"
```
hexfusion, line 16: just curious why we need to double the timeout?

test/harness/e2e/harness.go
```diff
@@ -0,0 +1,108 @@
...
+
+	parts := strings.Split(currentWorkDirectory, "/")
+	for i := len(parts) - 1; i >= 0; i-- {
+		if parts[i] == "test" {
+			path := strings.Join(parts[:i], "/")
+			logrus.Debugf("Top-level directory: %s", path)
+			return path
+		}
+	}
+	Expect(false).To(BeTrue(), "could not find top-level directory")
```
hexfusion, line 41: can you return an empty string and then validate the return to maybe not sure if this would just work
```
Fail("Could not find top-level directory")
```

test/e2e/agent_test.go
```diff
@@ -23,47 +23,94 @@ func TestAgent(t *testing.T) {
...
+		pcr := measurements[pcrReg]
+		Expect(pcr).ToNot(BeNil())
+
+		pcrRemove0x := strings.ReplaceAll(pcr, "0", "")
+		Expect(pcrRemove0x).ToNot(BeEmpty(), "PCR %s is empty or all 0's", pcrReg)
+	}
+}
+
+func getDeviceWithStatusSystemInfo(harness *e2e.Harness, enrollmentID string) *client.ReadDeviceResponse {
+	device, err := harness.Client.ReadDeviceWithResponse(harness.Context, enrollmentID)
```
hexfusion, line 110: should we log the original error?

test/e2e/agent_test.go
```diff
@@ -23,47 +23,94 @@ func TestAgent(t *testing.T) {
...
+
+func verifyPCRRegistersNotEmpty(measurements map[string]string) {
+	Expect(measurements).ToNot(BeNil())
+	for i := 1; i < 10; i++ {
+		pcrReg := fmt.Sprintf("pcr%02d", i)
+		pcr := measurements[pcrReg]
+		Expect(pcr).ToNot(BeNil())
+
+		pcrRemove0x := strings.ReplaceAll(pcr, "0", "")
+		Expect(pcrRemove0x).ToNot(BeEmpty(), "PCR %s is empty or all 0's", pcrReg)
```
hexfusion, line 105: can we add a comment on what is happening here why are we testing a mutated copy of PCR?

hexfusion (commented):
looks good few notes

test/e2e/agent_test.go
mangelajo: The TPM test is more complex and takes longer. (Waiting for enrollment, approval, waiting for device report to check again the TPM details)... 

We could bump that case , I am trying to remember and may be some of the other tests also timed out. We could be more specific per test I guess.

test/e2e/agent_test.go
mangelajo: Ack, I will clarify what's going on here. Thanks for bringing it up.

test/e2e/agent_test.go
mangelajo: Hmm, you are right; an error at this point is probably worth logging, at least. At least with the current usage, it should not happen, I believe.

test/harness/e2e/harness.go
mangelajo: Thanks. I remember writing that and thinking... I am sure there is a better way :D

test/e2e/agent_test.go
mangelajo: Ok, I checked, and I decided to go a little bit harder. The test will fail on errors here. As that should not be happening.

hexfusion (approved):

PR #234, opened 2024-04-16
NO-ISSUE: Add missing test for both podman and crio containers

oglok:
It also fixes the Condition status, which was not carrying over the count of notRunning containers among podman and crio engines.


mangelajo (approved):
LGTM

mangelajo (approved):

PR #235, opened 2024-04-17
Optimize flightctl-server-container build

mangelajo:
The go-cache in bin/go-cache creates issues with the linter and other processes. And Copying the whole directory is slow when we have artifacts in bin/

We could use dockerignore and be less explicit but then we cannot copy some artifacts that we need from bin/ for other container builds.


mangelajo:
> I wonder if we should be moving fixtures out of bin/ otherwise this looks sane

That could be an option too...

PR #247, opened 2024-05-08
RFE: API iteration device/fleet

mangelajo:

PR #248, opened 2024-05-08
NO-ISSUE: API-Fleet templating enhancement

mangelajo:


docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+```
+
+* Enable different references for a config source based on labels: 
+```yaml
+- name: microshift-manifests
+  configType: GitConfigProviderSpec
+  gitRef:
+    repository: basic-nginx-demo
+    targetRevision: 1.0.0-{{ .device.labels["region"] }}
+    path: /basic-nginx-demo/configuration
```
hexfusion, line 36: From a UX perspective do we need a dry-run or such ability to validate these templates client side?

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+## Alternatives
+
+### Every configuration combination has it's own fleet
+This is the most obvious alternative, this proposal is not implemented, and administrators
+take care of replicating fleet definitions as necessary to accomodate the different
+configurations.
+
+This approach does not allow customization based on fields like the device name.
+
+### Fleet template inheritance / sub-fleets
```
hexfusion, line 206: If you have a fleet of cars with various manufacture dates/capabilities. Would this require a fleet for every car model/year? It seems that templating solves some but not all of these problems.  An intermediary abstraction might also be required. But I would agree in the happy path templating solves problems sub-fleets do not.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+### Troubleshooting
+
+Troubleshooting can be done by looking at the device status and rendered device spec. Any errors found in templates
+could be reported as part of the fleet status conditions.
+
+## Implementation History
+
+## Drawbacks
+
+## Alternatives
```
hexfusion, line 197: Is another option allowing the user to generate the templated options client side by consuming our API's?

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+#### Story 8: region specific radio bands/regulations
+As a developer, I want to ensure that my devices automatically comply with local regulations when enrolled in specific regions. (i.e. configured with region-specific radio bands for WiFi, 5G, or other radio technologies)
+
+#### Story 9: region specific data sovereignty
+As a developer, I want to ensure that my devices connect to the specific regional endpoints to ensure compliance with data sovereignty.
+
+### Notes/Constraints/Caveats
+
+We would need to identify the design changes relative to templateVersions and
+if the relationship of those continues to be to the fleet or device level.
```
hexfusion, line 175: This would be good to describe the ability to audit these templated variables in more detail. For example it should be possible to regenerate all of the device configurations in the past to validate correctness.

hexfusion (commented):
few questions otherwise good start!

docs/enhancements/fep-001-api-fleet-templating.md
mangelajo: Maybe not in the past, but the existing ones.

That would make sense as a good safety check.

We could also prevent modifying device labels via the device API when the rendering of the existing fleet is going to fail.

docs/enhancements/fep-001-api-fleet-templating.md
mangelajo: I don't understand what you mean :D

I need more details and an example, I will capture it here.

docs/enhancements/fep-001-api-fleet-templating.md
mangelajo: Yes, it would require a separate fleet for each end variant in the end.

The main proposal is better in this aspect.

I will capture this.

docs/enhancements/fep-001-api-fleet-templating.md
hexfusion: Often this type of templating is done client side. Given a file template if the user has access to the fleet/device apis then can consume those apis and perform the templating on the client side.

given the example below a user could generate all the possible configurations locally. I think a good example of this is exploring how terraform generates the state file.  Complex templating is done client side and verified before pushing. In some circumstances where this could be many thousands of configurations this might not be the appropriate approach but in simpler cases I think it would be pretty useful. Also we should potentally look at how terraform handles this at scale?

```
<loop syntax>
- name: contrived-config-{{ .device.name }}
  region: {{ .device.labels["region"] }}
 </loop>
 ```

File: docs/enhancements/fep-001-api-fleet-templating.md (hunk 4 above)
Comment (hexfusion) [reply docs/enhancements/fep-001-api-fleet-templating.md:175]: I can see govt or insurance companies asking for how did this device get this spec on this day. Depending on the data retention policy of the user this should be possible in theory. Just stuff to think about.

File: docs/enhancements/fep-001-api-fleet-templating.md (hunk 2 above)
Comment (hexfusion) [reply docs/enhancements/fep-001-api-fleet-templating.md:206]: OK so then as a user  `"how can I efficiently manage many fleets"` is maybe a good question we can flesh out.

File: docs/enhancements/fep-001-api-fleet-templating.md [hunk 5]
```diff
@@ -0,0 +1,257 @@
...
+```yaml
+- name: microshift-manifests
+  configType: GitConfigProviderSpec
+  gitRef:
+    repository: basic-nginx-demo
+    targetRevision: 1.0.0-{{ .device.labels["region"] }}
+    path: /basic-nginx-demo/configuration
+```
+
+* Enable different folders withing a config source based on labels:
```
oglok, line 39:
```suggestion
* Enable different folders within a config source based on labels: 
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+```yaml
+- name: microshift-manifests
+  configType: GitConfigProviderSpec
+  gitRef:
+    repository: basic-nginx-demo
+    targetRevision: 1.0.0
+    path: /basic-nginx-demo/configuration/{{ .device.labels["region"] }}
+```
+
+* Enable different folders withing a config source based on labels:
```
oglok, line 50:
```suggestion
* Enable different folders within a config source based on labels: 
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Device API
+
+The device API does not change, in the device API we don't allow templating. Any
+device specifications from a fleet should already be rendered into the device
+
+Variable/template expansion is not supported at the device API level, as:
+  * We are operating on a single device, variables aren't really beneficial
+  * Updates to labels will not affect configuration unless the device is part of a fleet.
+ll be considered offline.
```
oglok, line 89: incomplete sentence?

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+
+The exposed values for use in templates is intentionally limited:
+* device.metadata.labels
+* device.metadata.name
+
+Making sure that the render of the device template is deterministic (i.e. not including timestamps or dates, etc..).
+
+#### Expected behavior
+
+When fleet.speec.template.metadata or fleet.spec.template.spec is updated, the Flet deployment controller
```
oglok, line 134:
```suggestion
When fleet.spec.template.metadata or fleet.spec.template.spec is updated, the Fleet deployment controller
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+
+Allowing templating of the device spec in the fleet API we could achieve a more
+flexible management, allowing customers to manage devices in a more granular way.
+
+### Goals
+
+* Define templating mechanism for the device spec template configuration in fleets,
+  based on labels and other metadata.
+
+* Enable different os images based on the device images.
```
celdrake, line 21: Should this read "based on the device labels"?

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,257 @@
...
+metadata:
+  name: autonomous-forklifts
+spec:
+  template:
+    spec:
+      os:
+        image: forklifts-{{ .device.metadata.label[factory] }}:latest
+      config:
+        - name: factory-wifi-access-credentials
+          configType: K8sSecret
```
celdrake, line 105: Since `K8sSecret` are not yet implemented, this example may not be accurate (actually the current API value is `KubernetesSecretProviderSpec`)

docs/enhancements/fep-001-api-fleet-templating.md
mangelajo: I still don't fully understand the example.

I don't know if you mean that templating should be performed client side (the agent), in such case:

* We lose the ability to freeze floating tags into hashes (for images)
* We cannot freeze git references that have templating into plain data.

Or do you mean that some external entity must take the templating and generate the device's data?

Or something else.

I guess my friday-brain is fried :D

Can I ask you to write me a fully-fledged alternative entry?

hexfusion: > Or do you mean that some external entity must take the templating and generate the device's data?


the client being the cli and the user being someone who wants to push something with git ops. I will write something up

hexfusion:
```
Use case: as a developer, I would like to be able to render local template config files before I push them to my GitOps workflow. This is the same process we use when we generate our Teraform state. I would expect an SDK or go client code be available to generate this from the flightctl service apis.
```

hexfusion: This might play into the unmanaged usecase. I see solutions like terraform which already perform this templating efficiently with modules.  Now that IBM owns Hashicorp I think we should be considering how solutions like consol and terraform can work into these designs even if only in theory.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+
+* Define templating mechanism for the device spec template configuration in fleets,
+  based on labels and other metadata.
+
+* Enable different os images based on the device labels.
+```yaml
+  template:
+    spec:
+      os:
+        image: access-control-{{ .device.labels["region"] }}:1.0.0
```
fzdarsky, line 26: Do we need the leading `.` on `.device.labels`?

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+The device API does not change, in the device API we don't allow templating. Any
+device specifications from a fleet should already be rendered into the device
+
+Variable/template expansion is not supported at the device API level, as:
+  * We are operating on a single device, variables aren't really beneficial
+  * Updates to labels will not affect configuration unless the device is part of a fleet.
+
+### Fleet API
+
+The template.spec field of a fleet allows the use of go-template logic to render the device spec.
```
fzdarsky, line 92: Should we document alternatives considered and their pros/cons? And compare to the (IIRC rather limited) use of templating in MCO?

For example, having go-templates here means one cannot process these files with yaml editing tools like `yq` anymore.

An alternative could be a replacement mechanism like in the K8s downward API, but isn't very user-friendly (as in: simple to grok) and I also do not see how we can map the "env vars" back to the fields we want to parameterise.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+
+The exposed values for use in templates is intentionally limited:
+* device.metadata.labels
+* device.metadata.name
+
+Making sure that the render of the device template is deterministic (i.e. not including timestamps or dates, etc..).
+
+#### Expected behavior
+
+When fleet.spec.template.metadata or fleet.spec.template.spec is updated, the Flet deployment controller
```
fzdarsky, line 133:
```suggestion
When `fleet.spec.template` is updated, the Fleet deployment controller
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+starts rolling out that update to devices. This update can take long (depending on the upgrade strategy)
+
+When a template is applied to a specific device via the Device API:
```
fzdarsky, line 136:
```suggestion
When the fleet controller applied a template to a specific device via the Device API:
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+* The template variables get rendered (i.e. the labels at that point in time get used)
```
fzdarsky, line 137:
```suggestion
* The template variables get rendered (i.e. the device's labels at that point in time get used)
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+* Floating tags for external references (os image, git refs, etc..) get converted to stable tags (hashes)
```
fzdarsky, line 138:
```suggestion
* For versioned configuration sources (OS images, git repos, etc.), any floating tags ("latest", "stable", "v1.0", ...) get replaced with stable tags (e.g. commit hashes).
* For non-versioned configuration sources (Vault secrets, K8s config maps, etc.), a snapshot of the configuration will be created and referenced under a (sortable) pseudo-version (e.g. `secret-$namespace-$name-$datetime-$hash`).
* Replacements with stable tags / pseudo-versions will be stored so they apply to all following devices with the same labels, too.
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+* When later in time a device has labels updated, the device controller will queue the device to be re-evaulated and
+  the fleet template re-rendered. (this is the same workflow as when the labels selected by a fleet changes)
```
fzdarsky, line 140: And also when we detect a change in the configuration source, right?

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+1. We ignore it, and expect eventual consistency, on a 2nd round the server will detect that the image
+   reference has been updated, and the devices with the older images will be re-rendered.
+   Additionally, when using free-floating tags, it's probably a risk the customer is assuming in
+   favor of flexilibity and speed of development. This should in general be a practice we don't
+   recommend, as we only supporting it to make development easier while unfavouring determinism.
+
+2. We could mantain a list of all possible values for labels possibilities referenced within a fleet
+   This would help us create a list of references (i.e. an internal database object that holds all references)
+   for a rollout. This would not help with device names, but would help with labels.
+
```
fzdarsky, line 204: 3. After variable substitution, whenever we encounter a given (config_source, floating_tag)-tuple for the first time we cache a snapshot locally and use that same substitution for all following devices with the same tuple. This ensures consistency for all like devices. Devices with another tuple (after substituting in labels) will anyway receive a different configuration. Then we apply the same rules about "expiring" cache entries and re-checking sources as we have to do without variable substitution anyway.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Risks and Mitigations
+
+When rolling out a configuration to a fleet of devices, during the rollout, some devices could turn out to be
+not renderable, for example:
+* due to a mising label
```
fzdarsky, line 210:
```suggestion
* due to a missing label
```

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,295 @@
...
+* because the template rendering is referencing non existing
+
+This could be mitigated by performing a dry-run of the rollout, and only applying the rollout if all devices
```
fzdarsky, line 213: We probably need to differentiate between "fleet controller did not find the label, so failed to do the substitution" and "the referenced resource (after substitution) does not exist / isn't accessible". There is little the service can do about the latter: verifying service-side is costly and also futile, because the resource could be there when the service tried, but no longer there when the agent needs it. At the end of the day, the agent will try during staging, fail, and set its condition to "failed to reconcile".

docs/enhancements/fep-001-api-fleet-templating.md
fzdarsky: At the end of the day, we're implementing look-up tables (LUTs) that map (<path_to_field_in_device_template>, <old_value>, <device_name>, <device_labels>) to <new_value>. The proposal outlined in this doc is generating the LUTs based on purely local information (go-template + device spec). IIUC, you're suggesting users should be able to upload their own LUTs. Alternatively, we could call a lambda functions provided by the users that would take the same params and compute the new values.

One challenge when using GitOps: Users will know all the labels (they'd be stored in git, too), so can include these in the LUTs. However, it's hard to include device.name (the device identity / fingerprint), as that is only known once the device starts for the first time and enrolls. And then you'd need to create a commit to add the new device id to git. Or users would have to query the current set of devices from the API, compute the LUT, then upload the LUT. For potentially a million devices. I don't think that's viable.

docs/enhancements/fep-001-api-fleet-templating.md
fzdarsky: Fleets of fleets ;)

docs/enhancements/fep-001-api-fleet-templating.md
hexfusion: > Or users would have to query the current set of devices from the API, compute the LUT, then upload the LUT. For potentially a million devices. I don't think that's viable.

This rendering would take place on the fleet level though so I would imagine smaller numbers :)

docs/enhancements/fep-001-api-fleet-templating.md
hexfusion: As we spoke about offline. The alternative approach proposed would serve the known device resourse and or provide basic agent templating tokens. Perhaps this is {{ .device.name}} or `__DEVICE_NAME__`  for example. This would not replace a server-side solution but could provide flexibility for files and services looking for this data.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,300 @@
...
+
+```yaml
+kind: Fleet
+metadata:
+  name: autonomous-forklifts
+spec:
+  template:
+    spec:
+      os:
+        switchFromLabel:
```
hexfusion, line 269: wouldn't a mutation of the spec be enough to show intent? It's not clear to me what ` switchFromLabel` is attempting.

hexfusion: I see now so conditional image spec.

hexfusion: Would it make sense to use a labelSelector then template the device spec?

mangelajo: I think the idea was to make the system able to pre-calculate, I think this alternative does not need a lot of development now, as @avishayt had some agreement on the feasibility of the main one now (based on  #247 final proposal).

I am happy to continue developing the alternative if we don't see the main one going forward.

docs/enhancements/fep-001-api-fleet-templating.md
mangelajo: @fzdarsky I have modified the examples in a way that it's valid for gotemplate and at the same time it would allow yq. 

Also I am using the index .map `key` which is the right way for gotemplate, the other one is jinja which won't work here :-/

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,300 @@
...
+
+When the fleet controller applies a `TemplateVersion` to a specific device through the Device API:
+* The template variables get rendered (i.e. the device's labels at that point in time get used)
+ 
+* For versioned configuration sources (OS images, git repos, etc.), any floating tags ("latest",
+  "stable", "v1.0", ...) get replaced with stable tags (e.g. commit hashes), those references
+  were already captured during the device scan phase.
+
+* Replacements with stable tags / pseudo-versions will be stored so they apply to all following
+  devices with the same labels, too.
```
mangelajo, line 151: removing this as it's already covered in #247

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,300 @@
...
+when a fleet.spec.template changes. Still, labels could change while a fleet is being rolled out.
+
+We probably need to differentiate between "fleet controller did not find the label, so
+failed to do the substitution" and "the referenced resource (after substitution) does
+not exist / isn't accessible". There is little the service can do about the latter:
+verifying service-side is costly and also futile, because the resource could be there when
+the service tried, but no longer there when the agent needs it. At the end of the day,
+the agent will try during staging, fail, and set its condition to "failed to reconcile".
+The former case would be similar (failure to reconcile). Then, if errors exceed threshold,
+stop the rollout.
```
mangelajo, line 218: Removing this paragraph in favor of a more specific implementation in sections: 
#### Expected behavior on issues
##### Device cannot be rendered from Template due to missing label
##### Device cannot generate /device/{name}/rendered version due to missing resource

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,303 @@
...
+  evaluation changes) the device controller will queue the device to be re-evaluated and the
+  `TemplateVersion` re-rendered.
+
+#### Expected behavior on issues
+
+##### Device cannot be rendered from Template due to missing label
+When a device cannot be rendered for some reason (e.g. missing label), the fleet-controller
+will mark the device with a label `fleet-controller/failed-to-reconcile=true`, and add an annotation
+`fleet-controller/failed-to-reconcile-reason` with a human-readable reason. The fleet will include
+a status condition `DeviceFailedToReconcile`.
```
hexfusion, line 160: then a list of devices in the message to how do you describe multiple devices in failure state?

docs/enhancements/fep-001-api-fleet-templating.md
fzdarsky: I probably didn’t notice: Is your solution to put quotes around the curly braces (`”{{…}}”`)? If so, an alternative could be to change the delimiters to `${{…}}`, which seems to be valid YAML even without quotes.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,303 @@
...
+* device.metadata.name
+
+Making sure that the render of the device template is deterministic (i.e. not including timestamps or dates, etc..).
+
+#### Expected behavior
+
+When `fleet.spec.template` is updated, the fleet controller
+freezes the template, including non-expanded variables in the TemplateVersion. All existing devices
+are scanned to determine possible git/image/secret sources references by the `TemplateVersion` in the
+fleet. Those source references are captured in/alongside the `TemplateVersion` object. For non-versioned
```
avishayt, line 137: I know I pushed for this initially, but I don't think it's scalable.  If we have 10K devices in a fleet, and some secret has {{device.metadata.name}} in it, we'll have 10K entries in the TemplateVersion and validate 10K secrets before proceeding.  Obviously the example is a bit extreme, but you get the point.  Maybe we should validate the fleet template with best effort, ignoring any config items that have parameters.  Then we might fail to render devices and will see the error message on each relevant device.  Devices won't have downtime, however, because they will still see the latest successfully-rendered config.

avishayt, line 137: Please add the reasons here - (1) validation, (2) freezing the same versions for all devices

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,303 @@
...
+
+* When later in time a device has labels updated (which could potentially result in template
+  evaluation changes) the device controller will queue the device to be re-evaluated and the
+  `TemplateVersion` re-rendered.
+
+#### Expected behavior on issues
+
+##### Device cannot be rendered from Template due to missing label
+When a device cannot be rendered for some reason (e.g. missing label), the fleet-controller
+will mark the device with a label `fleet-controller/failed-to-reconcile=true`, and add an annotation
```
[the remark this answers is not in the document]
avishayt: We already have "Valid" conditions on both the fleet and device.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,303 @@
...
+`fleet-controller/failed-to-reconcile-reason` with a human-readable reason. The fleet will include
+a status condition `DeviceFailedToReconcile`.
+
+Once all devices are corrected and can be reconciled again this status condition will be removed.
+
+#### Device cannot generate /device/{name}/rendered version due to missing resource
+When a device cannot render the template due to missing resources (e.g. git repo not found, secret not found),
+the device-controller will set a status condition on the device `MissingResource` with a human-readable reason.
```
avishayt, line 166: We already catch these "not found" errors when freezing the template, and in that case we set the Valid condition on the fleet to False.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Story 8: region specific radio bands/regulations
+As a developer, I want to ensure that my devices automatically comply with local regulations when enrolled in specific regions. (i.e. configured with region-specific radio bands for WiFi, 5G, or other radio technologies)
+
+#### Story 9: region specific data sovereignty
+As a developer, I want to ensure that my devices connect to the specific regional endpoints to ensure compliance with data sovereignty.
+
+### Notes/Constraints/Caveats
+
+Traceability of device spec changes over time becomes increasingly important with templating,
```
avishayt, line 199: I think this is an important point, but also goes beyond the scope of this enhancement.  I think we'd want a separate enhancement describing an audit mechanism and detailing what events to audit.

docs/enhancements/fep-001-api-fleet-templating.md
```diff
@@ -0,0 +1,303 @@
...
+        image: "access-control-{{ index .device.metadata.labels `region` }}:1.0.0"
+```
+
+* Enable different references for a config source based on labels:
+```yaml
+- name: microshift-manifests
+  configType: GitConfigProviderSpec
+  gitRef:
+    repository: basic-nginx-demo
+    targetRevision: "1.0.0-{{ index .device.metadata.labels `region` }}"
```
avishayt, line 35: Can we limit targetRevision to have a single parameter?  We need to map the parameter to frozen hash, so this will be easier without multiple parameters.

PR #249, opened 2024-05-08
MGMT-17908: agent: refactor bootstrap

hexfusion:
This PR does a minor cleanup and consolidation of bootstrap logic. The reason this work is important is that it is a dependency for proper rollback support and status updates post update.

Note: this work is blocked by #255 today if we were to update into an image without bootc support it would brick the device.

Depends on # https://github.com/flightctl/flightctl/pull/255


hexfusion:
closing for now

PR #255, opened 2024-05-14
MGMT-17762: internal: unify bootc/ostree clients

hexfusion:
This PR is a first pass at providing an interface that provides the best of both bootc and rpm-ostree while we are in transition period as bootc matures.  

Why this is important.
- bootc does not support unstaging updates currently which is important for rollbacks.
- without this if we update into an image without bootc agent will fail today

changes:
-  Add support for a single status that is compatible with bootc or rpm-ostree.
-  Added a new interface bootimage.Manager and implementations for both rpm-ostree and bootc
-  Added support to disable image management if the host does not support bootc or rpm-ostree


internal/bootimage/rpm-ostree.go
```diff
@@ -0,0 +1,134 @@
...
+			ImageDigest: deployment.ContainerImageReferenceDigest,
+		},
+		Ostree: BootEntryOstree{
+			Checksum:     deployment.Checksum,
+			DeploySerial: int(deployment.Serial),
+		},
+	}, nil
+}
+
+// TODO: there is probably a supported way to do this in the containers/image library
```
hexfusion, line 127: https://github.com/coreos/rpmostree-client-go/blob/main/pkg/imgref/imgref.go#L58C6-L58C11

internal/bootimage/rpm-ostree.go
```diff
@@ -0,0 +1,133 @@
+package bootimage
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/containers/image/v5/docker/reference"
+	"github.com/coreos/rpmostree-client-go/pkg/client"
```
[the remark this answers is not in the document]
fzdarsky: You’re right. I remembered MCD execs out to rpm-ostree, but didn’t think of the types.

internal/agent/device/bootstrap.go
```diff
@@ -165,34 +167,8 @@ func (b *Bootstrap) ensureCurrentRenderedSpecUpToDate(ctx context.Context) error
...
-	// create the management client
-	managementHTTPClient, err := client.NewWithResponses(b.managementEndpoint,
-		b.deviceReader.PathFor(b.caFile),
-		b.deviceReader.PathFor(b.managementGeneratedCertFile),
-		b.deviceReader.PathFor(b.keyFile))
-	if err != nil {
-		return fmt.Errorf("create management client: %w", err)
-	}
-	managementClient := client.NewManagement(managementHTTPClient)
-
```
hexfusion, line 194: mgt client is created earlier this is a dupe

internal/bootimage/rpm-ostree.go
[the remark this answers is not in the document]
hexfusion: I will ensure MCD adopts these clients as well so there is no hand rolled execution.

internal/bootimage/rpm-ostree.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (c *OsTreeClient) RemoveStaged(_ context.Context) error {
+	return c.client.RemovePendingDeployment()
+}
+
+func (c *OsTreeClient) IsDisabled() bool {
+	return false
+}
+
+// getOsTreeImageStatus returns ImageStatus given an rpm-ostree Deployment
```
hexfusion, line 108: nit update naming

PR #256, opened 2024-05-16
MGMT-17847: internal/experimental: add support for experimental features

hexfusion:
This PR adds a simple mechanism for the agent and service to enable experimental features based on the existence of `FLIGHTCTL_EXPERIMENTAL_FEATURES_ENABLED` with a non-empty value.. 

usage
```
experimentalFeatures := experimental.NewFeatures()
if experimentalFeatures.IsEnabled() {
  // do potentially unstable things 
}
```

OCP ref. https://github.com/openshift/installer/blob/294318e39b4ba31a16f87fed688134fee51dd6c2/cmd/openshift-install/create.go#L204C15-L204C67

### DIFF
--- a/internal/experimental/experimental.go
+++ b/internal/experimental/experimental.go
@@ -1,0 +1,39 @@
+package experimental
+
+import (
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+const (
+	// ExperimentalFeatureEnvKey is the environment variable key used to enable experimental features.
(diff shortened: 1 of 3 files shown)


mangelajo (approved):
I was hoping that @fzdarsky would chime in, but sounds good to me.

hexfusion:
merging, we can talk about this in more detail when implemented.

PR #263, opened 2024-05-27
NO-ISSUE: Enable ssh repositories and additional auth methods

rawagner:
This allows us to connect to ssh repositories and also mTLS enabled https ones.

ssh repo example
```yaml
apiVersion: v1alpha1
kind: Repository
metadata:
  name: sshRepo
spec:
  repo: git@github.com:foo/private-repo.git
  sshConfig:
    sshPrivateKey: base64-encoded-private-key 
    privateKeyPassphrase: optional-key-passphrase
```

mTLS https example
```yaml
apiVersion: v1alpha1
kind: Repository
metadata:
  name: httpsRepo
spec:
  repo: https://foo/bar
  httpConfig:
    tls.crt: base64-encoded-client-cert
    tls.key: base64-encoded-client-key
    ca.crt: base64-encoded-ca
```


api/v1alpha1/openapi.yaml
```diff
@@ -1622,19 +1622,54 @@ components:
...
-          type: string
-          description: 'The (possibly remote) repository URL to clone from'
         username:
           type: string
           description: 'The username for auth with HTTP transport'
         password:
           type: string
           description: 'The password for auth with HTTP transport'
           format: password
+        tlsClientCertData:
```
hexfusion, line 1635: curious your thoughts on using thed below keys?  Seems like we might need this kind of object again, perhaps we could consider a reusable type either now or as a followup. Is this a secret?

```
  tls.crt
  tls.key
  ca.crt
  ```

File: internal/tasks/git_helpers.go [hunk 2]
```diff
@@ -22,12 +22,11 @@ func CloneGitRepo(repo *model.Repository, revision *string, depth *int) (billy.F
 	if depth != nil {
 		opts.Depth = *depth
 	}
-	if repo.Spec.Data.Username != nil && repo.Spec.Data.Password != nil {
-		opts.Auth = &githttp.BasicAuth{
-			Username: *repo.Spec.Data.Username,
-			Password: *repo.Spec.Data.Password,
-		}
+	auth, err := repotester.GetAuth(repo)
```
hexfusion, line 25: :+1:

internal/tasks/repotester/repotester.go
```diff
@@ -59,6 +73,100 @@ func (r *RepoTester) TestRepositories() {
...
+			if repository.Spec.Data.SshConfig.SkipServerVerification != nil && *repository.Spec.Data.SshConfig.SkipServerVerification {
+				auth.HostKeyCallbackHelper = gitssh.HostKeyCallbackHelper{
+					HostKeyCallback: ssh.InsecureIgnoreHostKey(), //nolint:gosec
+				}
+			}
+
+			return auth, nil
+		}
+	}
+	return nil, nil
```
hexfusion, line 167: Should we document that finding no auth returns a nil AuthMethod? Or is this  an error?

internal/tasks/repotester/repotester.go
```diff
@@ -59,6 +73,100 @@ func (r *RepoTester) TestRepositories() {
...
+		if repository.Spec.Data.HttpConfig.Username != nil && repository.Spec.Data.HttpConfig.Password != nil {
+			auth := &githttp.BasicAuth{
+				Username: *repository.Spec.Data.HttpConfig.Username,
+				Password: *repository.Spec.Data.HttpConfig.Password,
+			}
+			return auth, nil
+		}
+	} else if scpLikeUrlRegExp.MatchString(*repository.Spec.Data.Repo) {
+		if repository.Spec.Data.SshConfig != nil {
+
```
hexfusion, line 143: nit: drop newline

internal/tasks/repotester/repotester.go
```diff
@@ -59,6 +73,100 @@ func (r *RepoTester) TestRepositories() {
...
+			if err != nil {
+				return nil, err
+			}
+
+			repoSubmatch := scpLikeUrlRegExp.FindStringSubmatch(*repository.Spec.Data.Repo)
+			password := ""
+			if repository.Spec.Data.SshConfig.PrivateKeyPassphrase != nil {
+				password = *repository.Spec.Data.SshConfig.PrivateKeyPassphrase
+			}
+			auth, err := gitssh.NewPublicKeys(repoSubmatch[1], sshPrivateKey, password)
```
hexfusion, line 154: should we check the length of repoSubmatch?

internal/tasks/repotester/repotester.go
```diff
@@ -59,6 +73,100 @@ func (r *RepoTester) TestRepositories() {
...
+		&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tlsConfig,
+			},
+		},
+	))
+	return nil
+}
+
+func GetAuth(repository *model.Repository) (transport.AuthMethod, error) {
```
hexfusion, line 125: since this is public can we add some basic docs?

hexfusion (commented):
Thanks for the contribution, added a few notes.

api/v1alpha1/openapi.yaml
mangelajo: This API should be subject to RBAC (i.e. Admins are the only ones to read/write), but we should not depend on secrets, as the idea is to be heavily decoupled of K8s. (Although we know that in most cases this is going to be deployed on top of openshift.)

Eventually we can replace on those objects references to external secret vaults/stores may be.

mangelajo:
Nice contribution, thanks @rawagner , some nits need handling, but other than that...

internal/tasks/repotester/repotester.go
rawagner: I don't think it is needed. The `FindStringSubmatch` always returns an array which size is equal to `scpLikeUrlRegExp` match groups. If the group is not found, then the value in the array is an empty string. 

But I've added the check just in case. Being more defensive here won't hurt :)

internal/tasks/repotester/repotester.go
rawagner: Docs added. Let me know if it's not enough.

internal/tasks/repotester/repotester.go
rawagner: `nil` is not an error in this case. It simply means that repo did not have any ssh/http config. 
It is documented now.

api/v1alpha1/openapi.yaml
rawagner: IIUC what @hexfusion meant is that we would have our own resource, similar to k8s secret.

I think that would be nice as it would also enable users to reuse credentials in multiple places.
In k8s we would replace one reference (our secret resource) for the other (k8s secret)

rawagner: I've updated the keys as suggested by @hexfusion . Much better!

hexfusion (dismissed):

api/v1alpha1/openapi.yaml
```diff
@@ -1622,19 +1622,54 @@ components:
         - metadata
         - spec
       description: Repository represents a git repository
-    RepositorySpec:
+    RepositoryHttpConfig:
```
avishayt, line 1625:
```suggestion
    GitHttpConfig:
```

api/v1alpha1/openapi.yaml
```diff
@@ -1622,19 +1622,54 @@ components:
...
+          type: string
+          description: 'Base64 encoded TLS cert key'
+          format: password
+        ca.crt:
+          type: string
+          description: 'Base64 encoded root CA'
+        skipServerVerification:
+          type: boolean
+          description: 'Skip remote server verification'
+    RepositoryGitConfig:
```
avishayt, line 1649:
```suggestion
    GitSSHConfig:
```

api/v1alpha1/openapi.yaml
```diff
@@ -1622,19 +1622,54 @@ components:
...
+          type: string
+          description: 'The passphrase for sshPrivateKey'
+          format: password
+        skipServerVerification:
+          type: boolean
+          description: 'Skip remote server verification'
+    RepositorySpec:
+      type: object
+      properties:
+        repo:
```
avishayt, line 1666: Let's move this into both GitHttpConfig and GitSSHConfig

api/v1alpha1/openapi.yaml
```diff
@@ -1622,19 +1622,54 @@ components:
...
+        privateKeyPassphrase:
+          type: string
+          description: 'The passphrase for sshPrivateKey'
+          format: password
+        skipServerVerification:
+          type: boolean
+          description: 'Skip remote server verification'
+    RepositorySpec:
+      type: object
+      properties:
```
avishayt, line 1665: Let's use OneOf?

mangelajo: Right, OneOf would make sense here.

api/v1alpha1/openapi.yaml
avishayt: I was wondering what we would put here for a k8s secret?  The name?  Seems weird.

api/v1alpha1/openapi.yaml
rawagner: I've used `OneOf`. The schema looks like 
```yaml
GitHttpConfig:
      type: object
      properties:
        repo:
          type: string
          description: 'The (possibly remote) repository URL to clone from'
        username:
          type: string
          description: 'The username for auth with HTTP transport'
        password:
          type: string
          description: 'The password for auth with HTTP transport'
          format: password
        tls.crt:
          type: string
          description: 'Base64 encoded TLS cert data'
          format: password
        tls.key:
          type: string
          description: 'Base64 encoded TLS cert key'
          format: password
        ca.crt:
          type: string
          description: 'Base64 encoded root CA'
        skipServerVerification:
          type: boolean
          description: 'Skip remote server verification'
      required:
        - repo
    GitHttpConfig:
      type: object
      properties:
        repo:
          type: string
          description: 'The (possibly remote) repository URL to clone from'
        sshPrivateKey:
          type: string
          description: 'The private SSH key'
          format: password
        privateKeyPassphrase:
          type: string
          description: 'The passphrase for sshPrivateKey'
          format: password
        skipServerVerification:
          type: boolean
          description: 'Skip remote server verification'
      required:
        - repo
    RepositorySpec:
      type: object
      oneOf:
        - $ref: "#/components/schemas/GitHttpConfig"
        - $ref: "#/components/schemas/GitSshConfig"
```

However there were couple of issues.
 - the `GitHttpConfig` and `GitSshConfig` objects are clashing - if you try to create a repo you'll get a response error ```Error at "/spec": input matches more than one oneOf schemas```
  - I've added `additionalProperties: false` which solved the error.
  - oapi generator seems to be not handling the `additionalProperties: false` so I had to write our own `GetGitHttpConfig` / `GetGitSshConfig` parsers which are using strict json decoding (disallowing unknown fields)

rawagner: hmm, and now we can't handle the simplest repo example which has only `spec.repo` defined

rawagner: I see two options:
 go back to
```yaml
spec:
  repo: string
  httpConfig/sshConfig: {}

```

or keep using OneOf and add a discriminator field ie `type: 'ssh' | 'http'`

avishayt: Yes, we use a discriminator for device.Spec

fzdarsky: In the device.spec the discriminator is redundant and exposed these really clunky long internal names to users, so I anyway wanted to propose to re-discuss them. What is the issue with the PR's original proposal?

avishayt: The benefit to OneOf is that it's clear at the API level that only one of the properties should be populated at a given time, the API checks it, and the discriminator makes it easy for the code that deals with it to know what kind of resource it is (in this case, is it SSH or HTTP).

rawagner: In the latest changes I've moved the http/ssh config into a separate field `config`.
That way we make use of `oneOf` and `discriminator` is not needed.

WDYT ? @avishayt @fzdarsky

avishayt: I was OK with the discriminator.  I'll let @fzdarsky chime in.

mangelajo: I believe usability is better without discriminator.

api/v1alpha1/openapi.yaml
mangelajo: It would be similar to the git https repo, probably the connection details for the k8s service account and endpoint.

mangelajo (approved):
👍 Thank you :)

avishayt (approved):

PR #264, opened 2024-05-28
Device API phase 1

avishayt:
This PR implements the device API, allowing users to configure devices without fleets.  The device configuration is currently rendered only upon modifying the device spec or fleet template.  Config source tracking will be introduced in a later PR.

6 commits:

- **MGMT-17878: Remove rendered config from TemplateVersion**
This commit removes the rendered config from the TemplateVersion resource because it will be moved to the Device resource. Also, the templateversion_populate task now only "freezes" the config rather than also rendering it.

- **MGMT-17878: Roll out to devices by copying spec**
Rather than setting the templateVersion, copy the spec to each device.

- **MGMT-17916: Render device configuration at the device level**
  - New task called DeviceRender that renders the device spec
    - Calculates and sets the annotation 'device-controller/renderedVersion' which is monotonically increasing number that changes on each render
  - Changes the getRenderedDeviceSpec API
    - The API endpoint is now "rendered" instead of "configuration"
    - Removed the knownOwner and knownTemplateVersion parameters, and instead added the knownRenderedVersion parameter
  - Necessary changes to the agent and CLI to call the modified API

- **MGMT-17917: Update rendered CLI command**
Instead of "bin/flightctl debug devspec <device>", it should be "bin/flightctl get device/xxxxxxx --rendered"

- **MGMT-17919: Remove createTemplateVersion API**
This API is not used and no longer needed

- **MGMT-17922: Move /templateversions endpoint under /fleets**


internal/store/device.go
```diff
@@ -155,6 +155,9 @@ func (s *DeviceStore) CreateOrUpdate(ctx context.Context, orgId uuid.UUID, resou
 	device := model.NewDeviceFromApiResource(resource)
 	device.OrgID = orgId
 
+	// Use the dedicated API to update annotations
+	device.Annotations = nil
+
```
mangelajo, line 162: Why do we clear all annotations here?

avishayt: The user shouldn't set them, so we clear them.  The service uses a dedicated API to set annotations.

internal/cli/get.go
```diff
@@ -61,7 +62,8 @@ func NewCmdGet() *cobra.Command {
 	cmd.Flags().StringVarP(&o.Output, "output", "o", o.Output, fmt.Sprintf("Output format. One of: (%s).", strings.Join(legalOutputTypes, ", ")))
 	cmd.Flags().Int32Var(&o.Limit, "limit", o.Limit, "The maximum number of results returned in the list response.")
 	cmd.Flags().StringVar(&o.Continue, "continue", o.Continue, "Query more results starting from the value of the 'continue' field in the previous response.")
-	cmd.Flags().StringVarP(&o.FleetName, "fleetname", "f", o.FleetName, "Fleet name for accessing individual templateversions.")
+	cmd.Flags().StringVar(&o.FleetName, "fleetname", o.FleetName, "Fleet name for accessing templateversions (templateversions only).")
```
hexfusion, line 65: minor: Could we clarify the description a little around what  `templateversions only` means.

internal/cli/get.go
```diff
@@ -61,7 +62,8 @@ func NewCmdGet() *cobra.Command {
...
+	cmd.Flags().BoolVar(&o.Rendered, "rendered", false, "Return the rendered device configuration that is presented to the device (devices only).")
```
hexfusion, line 66: I guess the same here does this mean these flags only work in certain contexts etc.

internal/store/device.go
```diff
@@ -27,7 +28,8 @@ type Device interface {
 	DeleteAll(ctx context.Context, orgId uuid.UUID, callback DeviceStoreAllDeletedCallback) error
 	Delete(ctx context.Context, orgId uuid.UUID, name string, callback DeviceStoreCallback) error
 	UpdateAnnotations(ctx context.Context, orgId uuid.UUID, name string, annotations map[string]string, deleteKeys []string) error
-	GetRendered(ctx context.Context, orgId uuid.UUID, name string, knownOwner, knownTemplateVersion *string) (*api.RenderedDeviceSpec, error)
+	UpdateRendered(ctx context.Context, orgId uuid.UUID, name string, rendered string) error
+	GetRendered(ctx context.Context, orgId uuid.UUID, name string, knownRenderedVersion *string) (*api.RenderedDeviceSpec, error)
```
hexfusion, line 32: Optional

```suggestion
GetRendered(ctx context.Context, orgId uuid.UUID, name string, knownVersion *string) (*api.RenderedDeviceSpec, error)
```

internal/store/device.go
```diff
@@ -301,64 +303,62 @@ func (s *DeviceStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, na
...
+			currentRenderedVersion, err = strconv.ParseInt(renderedVersionString, 10, 64)
+			if err != nil {
+				return err
+			}
 		}
 
-		if currentTemplateVersion == "" {
-			return flterrors.ErrTemplateVersionIsNil
-		}
+		currentRenderedVersion++
```
hexfusion, line 324: question: Does the store need to be thread safe?

internal/store/device.go
```diff
@@ -301,64 +303,62 @@ func (s *DeviceStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, na
...
+	if result.Error != nil {
+		return nil, flterrors.ErrorFromGormError(result.Error)
 	}
 
-	if templateVersion == nil {
-		return nil, nil
+	annotations := util.LabelArrayToMap(device.Annotations)
+	renderedVersion, ok := annotations[model.DeviceAnnotationRenderedVersion]
+	if !ok {
+		return nil, flterrors.ErrNoRenderedVersion
```
hexfusion, line 349: question: should we wrap this error with some of the context of the call to help debug? Maybe the name and or orgId.

internal/store/templateversion.go
```diff
@@ -18,10 +18,10 @@ import (
 type TemplateVersion interface {
 	Create(ctx context.Context, orgId uuid.UUID, templateVersion *api.TemplateVersion, callback TemplateVersionStoreCallback) (*api.TemplateVersion, error)
 	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.TemplateVersionList, error)
-	DeleteAll(ctx context.Context, orgId uuid.UUID, fleet *string) error
+	DeleteAll(ctx context.Context, orgId uuid.UUID, fleet string) error
 	Get(ctx context.Context, orgId uuid.UUID, fleet string, name string) (*api.TemplateVersion, error)
 	Delete(ctx context.Context, orgId uuid.UUID, fleet string, name string) error
-	UpdateStatusAndConfig(ctx context.Context, orgId uuid.UUID, resource *api.TemplateVersion, valid *bool, config *string, callback TemplateVersionStoreCallback) error
+	UpdateStatus(ctx context.Context, orgId uuid.UUID, resource *api.TemplateVersion, valid *bool, callback TemplateVersionStoreCallback) error
```
hexfusion, line 24: :heart:

internal/tasks/device_render.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/flightctl/flightctl/internal/store"
+	"github.com/flightctl/flightctl/internal/util"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/reqid"
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
+)
+
+func DeviceRender(taskManager TaskManager) {
```
hexfusion, line 20: nit: please add docs to public functions.

internal/tasks/device_render.go
```diff
@@ -0,0 +1,173 @@
...
+	device         *api.Device
+	renderedConfig *config_latest_types.Config
+}
+
+func NewDeviceRenderLogic(taskManager TaskManager, log logrus.FieldLogger, store store.Store, resourceRef ResourceReference) DeviceRenderLogic {
+	return DeviceRenderLogic{taskManager: taskManager, log: log, store: store, resourceRef: resourceRef}
+}
+
+func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
+	t.log.Infof("Rendering device %s/%s", t.resourceRef.OrgID, t.resourceRef.Name)
```
hexfusion, line 57: quesiton: should this be debug how chatty do we expect this to be?

internal/tasks/device_render.go
```diff
@@ -0,0 +1,173 @@
...
+
+	device, err := t.store.Device().Get(ctx, t.resourceRef.OrgID, t.resourceRef.Name)
+	if err != nil {
+		return err
+	}
+	t.device = device
+	emptyIgnitionConfig, _, _ := config_latest.ParseCompatibleVersion([]byte("{\"ignition\": {\"version\": \"3.4.0\"}"))
```
hexfusion, line 64: question: should the supported IGN version be a const somewhere?

hexfusion (commented):
looks good few minor things

internal/store/device.go
avishayt: What's your reasoning?  It's not clear to me why this is better

internal/store/device.go
avishayt: Right now we use annotations only for the service to annotate a resource for its own purposes.  If a user changed one of those annotations, things would break.  I'm happy to discuss changing that, but it's beyond the scope of this PR.

internal/store/device.go
avishayt: Until now I return the raw error from this layer and then add context in the caller

internal/tasks/device_render.go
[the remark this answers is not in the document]
avishayt: Based on your comment below I assume you meant "DeviceRenderer".  It's a better name, but DeviceRender fits the current convention.  I'd rather leave that change to a separate PR, perhaps when moving to the new event mechanism.  WDYT?

internal/tasks/device_render.go
```diff
@@ -0,0 +1,173 @@
...
+					log.Errorf("failed rendering device %s/%s: %v", resourceRef.OrgID, resourceRef.Name, err)
+				}
+			} else {
+				log.Errorf("DeviceRender called with unexpected kind %s and op %s", resourceRef.Kind, resourceRef.Op)
+			}
+		}
+	}
+}
+
+type DeviceRenderLogic struct {
```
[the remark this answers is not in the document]
avishayt: Same as above - I agree, but let's leave it for now?

internal/tasks/device_render.go
```diff
@@ -0,0 +1,173 @@
...
+	taskManager    TaskManager
+	log            logrus.FieldLogger
+	store          store.Store
+	resourceRef    ResourceReference
+	device         *api.Device
+	renderedConfig *config_latest_types.Config
+}
+
+func NewDeviceRenderLogic(taskManager TaskManager, log logrus.FieldLogger, store store.Store, resourceRef ResourceReference) DeviceRenderLogic {
```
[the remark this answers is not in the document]
avishayt: Same as above

internal/tasks/device_render.go
avishayt: Once every time a device spec is rendered :)
I'm honestly not sure how chatty is too chatty... I think it's a useful log message.  Perhaps leave it for now and tune in the future if necessary?

internal/tasks/device_render.go
avishayt: Do you mean what version the agent supports?  Good question.  My hunch is:
1. We need version negotiation for the entire agent-service API or at least some mechanism for the agent to upgrade itself to the latest version.
2. The ignition version we use would be tied to the overall version of the API.

WDYT?

hexfusion: Yes, a protocol version makes sense but for now, we should put the ign version `3.4` as a public const vs copying it around.

internal/store/device.go
hexfusion: From an API perspective, you are creating a getter for `RenderedDeviceSpec`. As a parameter, you are asking for `knownRenderedVersion`. Rendered and known are already inferred. It is known if it is passed as a param?

I would argue that `RenderedDeviceSpec` should have a version instead of `RenderedVersion` because Rendered studders.  You have a `Version` of a `DeviceSpec` which is `Renderd`. Vs a `RenderedVersion` of a `RenderedDeviceSpec` as the version itself is not `Renderd`.

```suggestion
        GetRenderedSpec(ctx context.Context, orgId uuid.UUID, name string, version *string) (*api.RenderedDeviceSpec, error)
```

internal/store/device.go
mangelajo: Let's leave it for now. My worry is that this may need to change i.e. if we exposed devices to external controllers, like a fleet manager but made in a different way..

internal/store/device.go
mangelajo: I think @avishay is just trying to align with the RFE we approved. https://github.com/flightctl/flightctl/pull/247

I agree it stutters, and I am ok to change it, but I would do it as a follow up on the existing RFE+code to keep them aligned.

And by doing a follow up we also let avishay continue with the stream of changes

mangelajo (approved):
I am good, although look at the comments from @hexfusion we may want to do some follow ups.

mangelajo (approved):

internal/store/device.go
hexfusion: OK can we create an issue to track?

PR #265, opened 2024-05-28
NO-ISSUE: Fix E2E VM boot on fedora hosts

mangelajo:
On some fedora hosts the enrollment keys installed through libvirt into the UEFI firmware are incomplete or incorrect and we cannot boot our images.

So we leave secureboot enabled for the attestation part, but we disable enforcing signature checks by not installing secureboot keys.

### DIFF
--- a/test/harness/e2e/vm/domain-template.xml
+++ b/test/harness/e2e/vm/domain-template.xml
@@ -24,7 +24,9 @@
     <boot dev='hd'/>
     <firmware>
       <feature enabled='yes' name='secure-boot'/>
-      <feature enabled='yes' name='enrolled-keys'/>
+      <!-- we enable secureboot to make sure that attestation is happening, but don't enroll the keys to avoid
+	   issues booting vms in fedora -->
+      <feature enabled='no' name='enrolled-keys'/>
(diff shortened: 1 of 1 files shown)


test/harness/e2e/vm/domain-template.xml
```diff
@@ -24,7 +24,9 @@
     <boot dev='hd'/>
     <firmware>
       <feature enabled='yes' name='secure-boot'/>
-      <feature enabled='yes' name='enrolled-keys'/>
+      <!-- we enable secureboot to make sure that attestation is happening, but don't enroll the keys to avoid
+	   issues booting vms in fedora -->
+      <feature enabled='no' name='enrolled-keys'/>
```
hexfusion, line 29: Should we make this conditional and on by default? Could be a follow-up.

hexfusion (approved):
Thanks for fixing this!

test/harness/e2e/vm/domain-template.xml
mangelajo: it's ok like this, we don't really need the signature check for our testing.

PR #267, opened 2024-05-29
NO-ISSUE: RFE for external system integration

mangelajo:
This RFE provides a proposal for enabling device integration to external systems, when external systems can provide transient configuration details into devices. I.e. configuration tokens, onboarding details, network configuration, per-device.

### DIFF
--- a/docs/enhancements/fep-002-config-hooks.md
+++ b/docs/enhancements/fep-002-config-hooks.md
@@ -1,0 +1,204 @@
+# EP: Definition of the Fleet and Device API spec
+<!-- this format is inspired by the K8S KEP format https://raw.githubusercontent.com/kubernetes/enhancements/master/keps/NNNN-kep-template/README.md -->
+## Summary
+
+This RFE defines a mechanism for per-device configuration extensiblity,
+where the device configuration can be instructed to talk to an external
+system or hook to fetch specific configurations for the device. This
+RFE also defines a talk-back mechanism for those hooks to tell flightctl
+that such configuration needs to be updated/revoked/etc.
+
+## Motivation
+
+Integration with external systems when devices may need specific
+per-device tokens, per-device configurations sometimes transient
+is a not-well covered use case in our API. Enabling such thing
+with today's API would not be possible. This is why we propose
+this configuration hooks API.
+
+This could be useful i.e. to integrate with ACM, where a service could
+be running on the ACM cluster, and the fleets configured to have
+the devices onboarded to ACM would point to that service in the fleet
+configuration items to fetch the configuration at specific
+points in time (hook points), anyway this is just an example, the
+mechanism should be generic to enable customers to leverage this.
+
+### Goals
+
+* Define hook points in the device configuration or flightctl
+  to enable the described functionality.
+
+* Allow an external system to be queried to retrieve per-device specific
+  configuration.
+
+* Allow the external system to call-back to flightctl to inform about
+  changes to be made in the previously supplied configuration.
+
+### Non-Goals
+
+* Define a specific product integration, this is a generic mechanism
+  that could be used with any external system.
+
+## Proposal
+
+We define a specific type of configuration that can be used to query an external system.
+
+```yaml
+apiVersion: v1alpha1
+kind: Fleet
+metadata:
+  creationTimestamp: "2024-04-30T14:06:17Z"
+  generation: 1
+  labels: {}
+  name: optical-inspector-production-fleet
+  owner: ResourceSync/basic-nginx-r3sourcesync
+spec:
+  selector:
+    matchLabels:
+      device: optical-inspector
+  template:
+    metadata:
+      generation: 1
+    spec:
+      config:
+...
+...
+      - configType: ExternalConfigProviderSpec
+        externalEndpoint:
+          config-name: my-config
+          url: https://acm-cluster-flightctl-hook.example.com/flightctl-hooks/
+          httpCredentials: my-credentials
+...
+...
+
+      os:
+        image: quay.io/flightctl/flightctl-agent-basic-nginx:latest
+      monitoring:
+        systemd:
+          matchPatterns:
+          - microshift.service
+          - crio.service
+          - flightctl-agent.service
+status:
+  conditions:
+  - lastTransitionTime: "2024-05-09T14:20:21Z"
+    status: "False"
+    type: OverlappingSelectors
+```
+
+### High level format of the hook calls
+#### FlightCTL to external system
+
+##### CALL: POST https://acm-cluster-flightctl-hook.example.com/flightctl-hooks/
+[{"event: "ON_DEVICE_SPEC_UPDATE",
+ "event_uuid": "uuid-for-tracking",
+ "config_-_name": "my-config",
+ "device_metadata": { -json representation of the device metadata on our API, including the owner/fleet -},
+ "callback_endpoint": "https://api.flightctl.example.com/external-provider-callback/",
+ "callback_token": "my-token"}
+ "last_config_hash": "hash-of-the-last-applied-config"
+]
+
+###### RESPONSE: 200 OK
+ [{"event": "ON_DEVICE_SPEC_UPDATE",
+  "event_uuid": "uuid-for-tracking",
+  "device_name": "device-name",
+  "config-name": "my-config",
+  "configuration": { representation of the configuration to be injected on the device i.e. in ignition format },
+ }]
+
+###### RESPONSE: 304 Not Modified
+ [{"event": "ON_DEVICE_SPEC_UPDATE",
+  "event_uuid": "uuid-for-tracking",
+  "device_name": "device-name",
+  "config-name": "my-config",
+ }]
+
+ 
+#### External system to FlightCTL
+
+
+##### CALLS POST to https://api.flightctl.example.com/external-provider-callback/
+
+[{"event: "ON_CONFIG_UPDATE",
+ "event_uuid": "uuid-for-tracking",
+ "config_name": "my-config",
+ "device_names": [`device-name1`, `device-name2`],
+]
+
+
+### The hook events
+
+#### FlightCTL to external system
+* `ON_FLEET_JOIN`: This event is triggered when a device is added to a fleet for
+  the first time. The external system can use this event to provide the initial
+  configuration or to start building any specific configuration that needs to be
+  created.
+
+* `ON_DEVICE_SPEC_UPDATE`: This event is triggered when a device.spec is updated. At this
+  point, if the device spec contains an ExternalConfigProviderSpec, the external system
+  can be called to get the new configuration.
+
+* `ON_DEVICE_METADATA_UPDATE`: This event is triggered when a device.metadata is updated. At this
+  point, if the device spec contains an ExternalConfigProviderSpec, the external system
+  can be called to let it know about metadata changes: labels, etc.
+
+* `ON_DEVICE_POLL`: This event is triggered by flightctl device-controller periodically
+  to check if the device configuration injected from this external system is still valid
+  the periodicty is decided by the external system on previous hook calls.
+
+* `ON_ENROLLMENT_REQUEST`: This is an event that is triggered when a device waiting
+  for enrollment request is targeted to this fleet because of the labels being
+  proposed on the device. This is an early notification that can be used to start
+  constructing credentials or triggering any other action.
+
+#### External system to FlightCTL
+When calling the external system flightctl provides a callback endpoint, as well as the
+necessary authorization (like a token) to call back to flightctl. The external system
+can notify flightctl about the following events:
+
+* `ON_CONFIG_UPDATE`: This event is triggered when the external system wants an specific
+ device configuration to be updated: i.e. new credentials for the network, a token
+ being removed because the application has finished onboarding, etc. This event
+ can point to a single device, or to a list of devices.
+
+
+
+
+### User Stories
+
+#### Story 1
+As a Fleet administrator, I want my Microshift based devices on a specific fleet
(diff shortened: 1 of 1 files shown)


docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+is a not-well covered use case in our API. Enabling such thing
+with today's API would not be possible. This is why we propose
+this configuration hooks API.
+
+This could be useful i.e. to integrate with ACM, where a service could
+be running on the ACM cluster, and the fleets configured to have
+the devices onboarded to ACM would point to that service in the fleet
+configuration items to fetch the configuration at specific
+points in time (hook points), anyway this is just an example, the
+mechanism should be generic to enable customers to leverage this.
```
hexfusion, line 24: > fetch the configuration at specific points in time (hook points)

question: This configuration is obtained by the service then injected into he renderedDeviceSpec?

mangelajo: Yes, correct, we would probably want to emit an audit trail. But that's the idea.

For example,  service on the ACM side, could be listening to flightctl events, and providing klusterlet+token details to devices that need to be on-boarded to ACM, then... once onboarded, the token can be removed., etc..

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+based on device attributes (location, serial number, region..). While this could
+be handled with templating, we need additional flexibility or templating cannot
+
+capture our needs well enough.
+
+### Risks and Mitigations
+
+Fleets consuming this introduce an external dependency that needs to be called for every
+device. This could be mitigated by having a local cache of the configuration on the device
+and only calling the external system when the cache is stale.
```
mangelajo, line 187: We should probably log and emit metrics related to those types of endpoints, so the admin can understand response times and any potential slow-down related to those services.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+      generation: 1
+    spec:
+      config:
+...
+...
+      - configType: ExternalConfigProviderSpec
+        externalEndpoint:
+          config-name: my-config
+          url: https://acm-cluster-flightctl-hook.example.com/flightctl-hooks/
+          httpCredentials: my-credentials
```
oglok, line 70:
```suggestion
          httpCredentials: my-credentials
          hook-step: onFleetJoin
```

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+ "event_uuid": "uuid-for-tracking",
+ "config_name": "my-config",
+ "device_names": [`device-name1`, `device-name2`],
+]
+
+
+### The hook events
+
+#### FlightCTL to external system
+* `ON_FLEET_JOIN`: This event is triggered when a device is added to a fleet for
```
oglok, line 133: My previous suggestion is covered here :-)

This hook event should be specified in the config, isn't it?

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
+# EP: Definition of the Fleet and Device API spec
+<!-- this format is inspired by the K8S KEP format https://raw.githubusercontent.com/kubernetes/enhancements/master/keps/NNNN-kep-template/README.md -->
+## Summary
+
+This RFE defines a mechanism for per-device configuration extensiblity,
+where the device configuration can be instructed to talk to an external
```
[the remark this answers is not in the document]
mangelajo: The service talks to the external system, I will clarify it.

docs/enhancements/fep-002-config-hooks.md
mangelajo: Right, that could make sense, I would provide it as an optional field instead maybe?

```suggestion
          httpCredentials: my-credentials
          hook-events:
            - onFleetJoin
            - onMetadataUpdate
            - onDevicePoll
            
```

In this way, the administrator can optionally be precise on which events to send. If not provided we would post all.

docs/enhancements/fep-002-config-hooks.md
mangelajo: Once you set up the endpoint, my idea is that we would send any events related to the fleet or fleet devices. But also filtering which events you want could make sense.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+This could be useful i.e. to integrate with ACM, where a service could
+be running on the ACM cluster, and the fleets configured to have
+the devices onboarded to ACM would point to that service in the fleet
+configuration items to fetch the configuration at specific
+points in time (hook points), anyway this is just an example, the
+mechanism should be generic to enable customers to leverage this.
+
+### Goals
+
+* Define hook points in the device configuration or flightctl
```
fzdarsky, line 28: I can see a need to emit fleet/device lifecycle events to an external system (say, to trigger a job via AAP's Event-Driven Automation), but not for the purpose of / being limited to configuration. Would that then duplicate/overlap with the proposed configuration provider's event notifications?

docs/enhancements/fep-002-config-hooks.md
fzdarsky, line 70: The httpCredentials would probably be a named reference to a Repository or Secret resource?

fzdarsky: Who would send these events to whom as a result? E.g. it'd be the fleet-controller sending "onFleetJoin/Leave" events and the device-controller the "onMetadataUpdate" events (assuming this refers to device.metadata). And the receiver could well be the Repository task so it could query the source.

What is confusing is this seems to suggest you want to send the events to the external system directly, i.e. sending it via Kafka or via callback URLs to one ore more consumers. However, event notifications are informational; you do not / cannot expect the receiver to be available or take an action and respond (sync or async). Instead, we'd want to _query_ the external system, or? And then we're back to what the Repository task already does.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+ "event_uuid": "uuid-for-tracking",
+ "config_-_name": "my-config",
+ "device_metadata": { -json representation of the device metadata on our API, including the owner/fleet -},
+ "callback_endpoint": "https://api.flightctl.example.com/external-provider-callback/",
+ "callback_token": "my-token"}
+ "last_config_hash": "hash-of-the-last-applied-config"
+]
+
+###### RESPONSE: 200 OK
+ [{"event": "ON_DEVICE_SPEC_UPDATE",
```
fzdarsky, line 103: This shows it's a query (resquest/response), not an event notification. So then we can have this handled by the Repository task.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+  "config-name": "my-config",
+ }]
+
+ 
+#### External system to FlightCTL
+
+
+##### CALLS POST to https://api.flightctl.example.com/external-provider-callback/
+
+[{"event: "ON_CONFIG_UPDATE",
```
fzdarsky, line 123: This could be covered by the Repository callback.

docs/enhancements/fep-002-config-hooks.md
fzdarsky: As said above, emitting lifecycle event notifications (i.e. without response) to external systems makes sense to me, but tying/limiting this to configuration not.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+
+### The hook events
+
+#### FlightCTL to external system
+* `ON_FLEET_JOIN`: This event is triggered when a device is added to a fleet for
+  the first time. The external system can use this event to provide the initial
+  configuration or to start building any specific configuration that needs to be
+  created.
+
+* `ON_DEVICE_SPEC_UPDATE`: This event is triggered when a device.spec is updated. At this
```
fzdarsky, line 138: This would be quite noisy, so we'd need to filter most of this out again. Again the notification could go to the Repository task who'd do the querying.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+    matchLabels:
+      device: optical-inspector
+  template:
+    metadata:
+      generation: 1
+    spec:
+      config:
+...
+...
+      - configType: ExternalConfigProviderSpec
```
fzdarsky, line 66: All config providers (except for the InlineConfig one) are querying external systems. Would "HTTPConfigProvider" make more sense?

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+  point, if the device spec contains an ExternalConfigProviderSpec, the external system
+  can be called to get the new configuration.
+
+* `ON_DEVICE_METADATA_UPDATE`: This event is triggered when a device.metadata is updated. At this
+  point, if the device spec contains an ExternalConfigProviderSpec, the external system
+  can be called to let it know about metadata changes: labels, etc.
+
+* `ON_DEVICE_POLL`: This event is triggered by flightctl device-controller periodically
+  to check if the device configuration injected from this external system is still valid
+  the periodicty is decided by the external system on previous hook calls.
```
fzdarsky, line 148: This is functionally equivalent to the Repository task polling.

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
+# EP: Definition of the Fleet and Device API spec
+<!-- this format is inspired by the K8S KEP format https://raw.githubusercontent.com/kubernetes/enhancements/master/keps/NNNN-kep-template/README.md -->
+## Summary
+
+This RFE defines a mechanism for per-device configuration extensiblity,
+where the device configuration can be instructed to talk to an external
+system or hook to fetch specific configurations for the device. This
+RFE also defines a talk-back mechanism for those hooks to tell flightctl
+that such configuration needs to be updated/revoked/etc.
```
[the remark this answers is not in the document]
mangelajo: Yes, it's an external source that provides configuration for devices, and it's responsible for caching on its side. 

This also defines a webhook on our side that can be called back by the remote system (i.e., a bearer token is generated for that system, and as part of our interactions we provide it with the talk-back endpoint and token). Such call back can be used for updates (removing the need for polling the remote system for changes).

docs/enhancements/fep-002-config-hooks.md
```diff
@@ -0,0 +1,204 @@
...
+per-device tokens, per-device configurations sometimes transient
+is a not-well covered use case in our API. Enabling such thing
+with today's API would not be possible. This is why we propose
+this configuration hooks API.
+
+This could be useful i.e. to integrate with ACM, where a service could
+be running on the ACM cluster, and the fleets configured to have
+the devices onboarded to ACM would point to that service in the fleet
+configuration items to fetch the configuration at specific
+points in time (hook points), anyway this is just an example, the
```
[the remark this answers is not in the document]
mangelajo: ack. Can you propose an alternative name for this?

docs/enhancements/fep-002-config-hooks.md
mangelajo: Are you referring to another RFE or this?

In this case, I am talking about the configuration entry, which would be the source/reason for such events. Of course, we could model this in a different way altogether.

docs/enhancements/fep-002-config-hooks.md
mangelajo: HTTPConfigProvider would make more sense, if for example we are just targeting an URL, may be a field of that URL, or some headers is the device name.

docs/enhancements/fep-002-config-hooks.md
mangelajo: That is a good point, if the system is not there to answer on the other side we may need to either retry, or queue, a messaging system would be more robust in a case like this and would avoid the need to retry on our side.

mangelajo: At the point where this RFE was created it was intentionally open. to be decided...

docs/enhancements/fep-002-config-hooks.md
mangelajo: Yes this would be unavoidably noisy. There is always the chance to optimize by grouping multiple requests.

docs/enhancements/fep-002-config-hooks.md
mangelajo: it is

mangelajo:
Just some comments; please feel free to re-create or edit this RFE as you see fit. @oglok I have granted you permission on this branch on this repo/branch on my git repo.

mangelajo:
I also understood that from the point of view of ACM they may want to do this differently, there is may be more chance for an HTTP Config provider instead of webhooks.

My intent at that time was to make a braindump of my ideas, this doesn't need to be the way we do it. Please change as necessary.

oglok:
Let's start simple. I will close this RFE (if you agree @fzdarsky @mangelajo ) and create a new one for an HTTP Config Provider. This new provider would help with ACM integration, but it would be completely generic. In a follow up RFE, we could discuss about "configuration triggers".

mangelajo:
> Let's start simple. I will close this RFE (if you agree @fzdarsky @mangelajo ) and create a new one for an HTTP Config Provider. This new provider would help with ACM integration, but it would be completely generic. In a follow up RFE, we could discuss about "configuration triggers".

Makes perfect sense to me 👍

PR #268, opened 2024-05-29
MGMT-17918: Add templateVersion annotation to fleet

avishayt:


internal/store/fleet.go
```diff
@@ -391,3 +397,25 @@ func (s *FleetStore) UpdateConditions(ctx context.Context, orgId uuid.UUID, name
 
 	return err
 }
+
+func (s *FleetStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, name string, annotations map[string]string, deleteKeys []string) error {
```
hexfusion, line 401: question: what is the use-case for `deleteKeys` param on this Update method?

internal/store/fleet.go
```diff
@@ -165,6 +168,9 @@ func (s *FleetStore) createOrUpdateTx(tx *gorm.DB, orgId uuid.UUID, resource *ap
 	fleet := model.NewFleetFromApiResource(resource)
 	fleet.OrgID = orgId
 
+	// Use the dedicated API to update annotations
+	fleet.Annotations = nil
```
hexfusion, line 172: Should we log a warn if fleet. Annotations aren't before setting nil or at least document that it will be discarded on the method?

internal/store/fleet.go
avishayt: If the caller needs to unset an annotation

internal/store/fleet.go
avishayt: I think logging a warning isn't warranted because it's fine to have annotations in the resource.  For example, if a user GETs a fleet, modifies it, and POSTs the yaml back.  Same for an internal task.  Anyway, we'll discuss it further because @mangelajo commented on the same thing in my other PR :)
But for now, I'd like to proceed instead of waiting to nail down the semantics of annotations.

https://github.com/flightctl/flightctl/pull/264#discussion_r1617001744

mangelajo (approved):
looks good, agreed on our need to polish the annotations behavior specially when we have external controllers at some point.

internal/store/fleet.go
hexfusion: OK to defer can we create an issue to track?

internal/store/fleet.go
hexfusion: OK so if the annotations map doesn't contain an existing key we don't delete the diff? Feels like an unnessisry step.

avishayt: No, it's not an overwrite API, it's a patch.  We update only the keys that are passed so that we don't accidentally delete keys that a different process is updating.

hexfusion (approved):
Would be nice to see some tracking of followups in Jira otherwise looks fine.

internal/store/fleet.go
hexfusion: maybe we should document the method?

PR #269, opened 2024-05-31
MGMT-17911:  implement new status

hexfusion:
# Note: This is still very much WIP please do not bother reviewing yet.


api/v1alpha1/openapi.yaml
```diff
@@ -2298,6 +2306,34 @@ components:
           type: string
           description: 'Status of the operation. One of: "Success" or "Failure". More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status'
       description: Status is a return value for calls that don't return other objects.
+    ApplicationStatus:
+      type: object
+      properties:
```
mangelajo, line 2311: It's related, but probably to be kept separate: 

https://github.com/flightctl/flightctl/blob/main/docs/enhancements/fep-000-api-device-fleet.md#devicesname-user-facing-method 

We are also proposing that this "app" monitoring goes under a spec field to make it clear that it's containers or systemd services that need monitoring. It looks based on this that may be we want to add an app name to the containers or services that we monitor?

hexfusion: > Based on this, may we want to add an app name to the containers or services that we monitor?

Can you clarify what you mean here are you saying you would like the status to lool like

```
status:
  monitoring:
      systemd:
        - name: mircroshift.service
          restarts: 0
          state: Running
         - name: whatever
       containers:
         - name: nginx
         ....
```

mangelajo: I meant, that, based on the API you were defining, when we create a list of monitored containers or systemd services, we could give them a logical application name. But you're right, and that's just the container name or systemd name to start with.

On the status field, we could put it where it's now or under monitoring, I guess it's not much different, on the device .spec side, I moved it under monitoring to make it clear that we were not deploying the systemd or containers, just... requesting monitoring on them.

api/v1alpha1/openapi.yaml
```diff
@@ -1801,6 +1801,14 @@ components:
           description: 'Current state of systemd units on the device.'
           items:
             $ref: '#/components/schemas/DeviceSystemdUnitStatus'
+        fingerprint:
```
fzdarsky, line 1804: No need to add an extra field; the device's fingerprint == `device.name`.

hexfusion: That makes sense but I was going off the `Data Definition Dictionary`[1] which describes it as a separate part of the status.

image


[1] https://docs.google.com/spreadsheets/d/1xlslENyYVWL-M2Qw4lOwHOEWMXXk9cPN-tCb-IH7t3c/edit#gid=21336438

celdrake:
@hexfusion I think this PR is outdated now, can it be closed?

hexfusion:
closing for now will breakout logic I need in another PR

Issue #266, opened 2024-05-28
prepare_agent_config errors due to mac os doesn't have `ip` command default avaliable


akoserwal:
`make deploy` errors out to generate the client config currently due to `ip` command is not available in mac os

```

At the end I got this error ```hack/prepare_agent_config.sh
hack/prepare_agent_config.sh: line 7: ip: command not found
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
Found ethernet interface IP:

```

PR #270, opened 2024-06-03
OCTOET-412 Deploy local registry for E2E in kind

mangelajo:
1st commit:
Creates an additional helm chart only for the E2E elements necessary for local E2E testing.

Also fixes: https://github.com/flightctl/flightctl/issues/266

2nd commit:
Moves all testing related hack scripts to test/scripts/


hack/get_ext_ip.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+if which ip 2>/dev/null 1>/dev/null; then
+    ip route get 1.1.1.1 | grep -oP 'src \K\S+'
+else
+    # MacOS does not have ip, so we use route and ifconfig instead
+    INTERFACE=$(route get 1.1.1.1 | grep interface | awk '{print $2}')
+    ifconfig | grep $INTERFACE -A 10 | grep "inet " | grep -Fv 127.0.0.1 | awk '{print $2}' | head -n 1
+fi
```
hexfusion, line 8: Would putting these e2e-related scripts in the test/scripts or similar make sense?

deploy/helm/e2e-extras/templates/registry-deployment.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: apps/v1
```
hexfusion, line 1: did you mean to commit the empty .gitignore?

hexfusion (commented):
Thanks for doing this. Do we want to plumb this into CI to ensure it works? Genreally LGTM.

hack/get_ext_ip.sh
mangelajo: I think so, probably most the `hack/*` scripts could go to test/scripts/

I can do that on a separate PR to keep this one clean.

deploy/helm/e2e-extras/templates/registry-deployment.yaml
mangelajo: Yes, we have it on the other module because helm expects that directory AFAIK, I think it was created by the helm initialization of the module.

mangelajo: Just did it now with a 2nd commit :)

hexfusion (approved):

mangelajo:
Thank you @hexfusion

PR #272, opened 2024-06-04
OCTOET-412: E2E registry bootc images

mangelajo:
* Setups an E2E CA and registry service certificates that the agents will trust
* Creates several agent image variants and pushes them all to the registry

### DIFF
--- a/deploy/agent-vm.mk
+++ b/deploy/agent-vm.mk
@@ -4,25 +4,6 @@ VMCPUS ?= 1
 VMDISK = /var/lib/libvirt/images/$(VMNAME).qcow2
 VMWAIT ?= 0
 
-bin/output/qcow2/disk.qcow2: hack/Containerfile.local bin/.rpm
-	sudo podman build -f hack/Containerfile.local -t localhost/local-flightctl-agent:latest .
-	mkdir -p bin/output
-	sudo podman run --rm \
-					-it \
-					--privileged \
-					--pull=newer \
-					--security-opt label=type:unconfined_t \
-	                -v $(shell pwd)/bin/output:/output \
-					-v /var/lib/containers/storage:/var/lib/containers/storage \
-					quay.io/centos-bootc/bootc-image-builder:latest \
-					--type qcow2 \
-					--local localhost/local-flightctl-agent:latest
-
-agent-image: bin/output/qcow2/disk.qcow2
-	@echo "Agent image built at bin/output/qcow2/disk.qcow2"
-	sudo chmod a+rw $(VMDISK) 2>/dev/null || true
-
-.PHONY: agent-image
 
 agent-vm: bin/output/qcow2/disk.qcow2
 	@echo "Booting Agent VM from $(VMDISK)"
--- a/test/scripts/agent-images/README.md
+++ b/test/scripts/agent-images/README.md
@@ -1,0 +1,38 @@
+# E2E Agent images
+
+We generate multiple agent images for testing purposes, each with a different
+services running, but all connected to our flightctl service for management.
+
+This work is performed by the `create_agent_images.sh` script in this
+directory.
+
+And can be triggered from the top-level makefile with: `make e2e-agent-images`
+
+The images are built using the `Containerfile-*` files in this directory, functionality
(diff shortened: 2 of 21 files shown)


hack/agent-image/Containerfile-e2e-v2.local
```diff
@@ -0,0 +1,7 @@
+FROM localhost:5000/flightctl-base:latest
+
```
hexfusion, line 2: could we add a short comment in all of these container images as to the intent?

hack/agent-image/test-e2e-dummy-v3.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Test E2E Dummy V3
```
hexfusion, line 2: what does the v3 refer to?

deploy/agent-vm.mk
```diff
@@ -3,9 +3,27 @@ VMRAM ?= 512
 VMCPUS ?= 1
 VMDISK = /var/lib/libvirt/images/$(VMNAME).qcow2
 VMWAIT ?= 0
+IP=$(shell test/scripts/get_ext_ip.sh)
 
-bin/output/qcow2/disk.qcow2: hack/Containerfile.local bin/.rpm
-	sudo podman build -f hack/Containerfile.local -t localhost/local-flightctl-agent:latest .
+
+out/.agent-bootc-images: out rpm hack/agent-image/Containerfile-e2e-base.local hack/agent-image/Containerfile-e2e-v2.local hack/agent-image/Containerfile-e2e-v3.local
```
hexfusion, line 9: I am getting nervous about the complexity of these make files. Is there anything we can do to make this easier for folks to grok.

Like what does this mean to a developer who is trying to understand a failing test.

hexfusion (commented):
nice addition! few open questions on documentation.

hack/agent-image/Containerfile-e2e-v2.local
mangelajo: Good idea, doing.

hack/agent-image/test-e2e-dummy-v3.service
mangelajo: V3 of the image, may be we can just call it test-e2e-another-dummy.service

deploy/agent-vm.mk
mangelajo: About the makefile complexity... yes, it's getting complex. Maybe we should pack some of those things in simpler sell scripts, as we do for the helm deploy, etc. Also, some shell script would help make those things leaner. I'll apply that idea to this patch more and see how it looks.


To understand a failing test, I think it's good to write idiomatic and very well-documented tests so it's clear what the intent of the test is, what images are involved, and what those images contain. I should probably add a section describing the 3 images we have and what is in each.

hexfusion: Sure but if the test is failing because of test infra or someone is trying to extended e2e they end up here. Thanks for looking into it.

deploy/agent-vm.mk
```diff
@@ -43,6 +24,7 @@ agent-vm-console:
 .PHONY: agent-vm
 
 clean-agent-vm:
+	rm out/.agent-bootc-images || true
```
mangelajo, line 27:
```suggestion
```

test/scripts/agent-images/README.md
```diff
@@ -0,0 +1,38 @@
...
+And can be triggered from the top-level makefile with: `make e2e-agent-images`
+
+The images are built using the `Containerfile-*` files in this directory, functionality
+or service deployment changes should be implemented on those container files, or if
+additional transition images are required we should create and document new Containerfiles.
+
+| Name   | Image                         | Bootc Containers                 |
+|------  |-------------------------------|----------------------------------|
+| base   | `bin/output/qcow2/disk.qcow2` | ${IP}:5000/flightctl-device:base |
+| v2     | N/A                           | $(IP):5000/flightctl-device:v3   |
```
mangelajo, line 18: typo in the bootc container refs v2/v3

hexfusion (approved):

PR #273, opened 2024-06-04
MGMT-18036: Fix fetching git hashes

avishayt:
The code supported cloning branches and tags, but there is no concept in git for cloning a hash. Fixed the code to perform a "checkout" after clone in case a hash is requested.

Note that until now there was an optimization of passing "depth=1" which means pulling with no history, but this means we can't check out hashes. We will optimize later when implementing a cache.

### DIFF
--- a/internal/tasks/git_helpers.go
+++ b/internal/tasks/git_helpers.go
@@ -39,18 +39,35 @@ func CloneGitRepo(repo *model.Repository, revision *string, depth *int) (billy.F
 		return nil, "", err
 	}
 	opts.Auth = auth
+	hash := ""
 	if revision != nil {
-		opts.ReferenceName = gitplumbing.ReferenceName(*revision)
+		referenceIsHash := gitplumbing.IsHash(*revision)
+		if !referenceIsHash {
+			opts.ReferenceName = gitplumbing.ReferenceName(*revision)
+		} else {
+			hash = *revision
+		}
 	}
 	gitRepo, err := git.Clone(storage, mfs, opts)
 	if err != nil {
-		return nil, "", err
+		return nil, "", fmt.Errorf("failed cloning git repo: %w", err)
 	}
-	head, err := gitRepo.Head()
-	if err != nil {
-		return nil, "", err
+	if hash != "" {
+		worktree, err := gitRepo.Worktree()
+		if err != nil {
+			return nil, "", fmt.Errorf("failed getting git repo worktree: %w", err)
+		}
+		err = worktree.Checkout(&git.CheckoutOptions{Hash: gitplumbing.NewHash(hash), Force: true})
+		if err != nil {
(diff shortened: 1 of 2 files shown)


celdrake (approved):
Worked well with the case I was testing before.  The device spec is now synchronizing correctly.

Edit:
Just FYI, that I noted that the rendering of the device via  `bin/flightctl get device/A --rendered > my-device.json`  results in JSON file of ~29MB. It seems to contain all files in the repository (I can see "path" entries for files in the root directory, in the ".github" folder, etc.

internal/tasks/git_helpers.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/memfs"
 	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	gitplumbing "github.com/go-git/go-git/v5/plumbing"
```
hexfusion, line 18: nit these are dupes.

hexfusion (commented):
Would it make sense to add a test to gate regression? Otherwise just one nit.

celdrake:
> Worked well with the case I was testing before. The device spec is now synchronizing correctly.
> 
> Edit: Just FYI, that I noted that the rendering of the device via `bin/flightctl get device/A --rendered > my-device.json` results in JSON file of ~29MB. It seems to contain all files in the repository (I can see "path" entries for files in the root directory, in the ".github" folder, etc.

As pointed by @avishayt, my example fleet spec was missing the `path` for the configs, and that meant that the entire repository content was being included. I fixed it [here](https://github.com/flightctl/flightctl-demos/commit/3f8e03b686222035684be4f3b31cfaed387cea79), and I'll create a PR to fix the demo repository.

I'm wondering if we could have a validation to ensure that the `path` is set.

internal/tasks/git_helpers.go
avishayt: Good call

avishayt:
> > Worked well with the case I was testing before. The device spec is now synchronizing correctly.
> > Edit: Just FYI, that I noted that the rendering of the device via `bin/flightctl get device/A --rendered > my-device.json` results in JSON file of ~29MB. It seems to contain all files in the repository (I can see "path" entries for files in the root directory, in the ".github" folder, etc.
> 
> As pointed by @avishayt, my example fleet spec was missing the `path` for the configs, and that meant that the entire repository content was being included. I fixed it [here](https://github.com/flightctl/flightctl-demos/commit/3f8e03b686222035684be4f3b31cfaed387cea79), and I'll create a PR to fix the demo repository.
> 
> I'm wondering if we could have a validation to ensure that the `path` is set.

We do validate that 'path' is set when you create a fleet via the fleet API, but ResourceSync skips these checks.  I opened MGMT-18048 to track this issue.

celdrake (approved):

PR #275, opened 2024-06-06
MGMT-18009: Enable JSON Patch on device/fleet/repo/resourcesync

rawagner:
The patched object is verified against OpenAPI schema before persisting to DB.

The patch format is using https://jsonpatch.com/ (same as k8s)


mangelajo:
Nice :) I think you need to rebase on top of latest spec.gen.go

internal/service/common.go
```diff
@@ -11,3 +21,63 @@ func NilOutManagedObjectMetaProperties(om *api.ObjectMeta) {
...
+
+	router, _ := gorillamux.NewRouter(swagger)
+	route, pathParams, _ := router.FindRoute(httpReq)
+
+	requestValidationInput := &openapi3filter.RequestValidationInput{
+		Request:    httpReq,
+		PathParams: pathParams,
+		Route:      route,
+	}
+	return openapi3filter.ValidateRequest(context.Background(), requestValidationInput)
```
hexfusion, line 52: please ensure this and friends have a real context, it will help with graceful termination flow.

internal/service/common.go
```diff
@@ -11,3 +21,63 @@ func NilOutManagedObjectMetaProperties(om *api.ObjectMeta) {
...
+}
+
+func ApplyJSONPatch[T any](obj T, newObj T, patchRequest api.PatchRequest, objPath string) error {
```
hexfusion, line 55: generics cute :)

nit: please plumb context into the signature.

internal/service/enrollmentrequest.go
```diff
@@ -28,7 +27,7 @@ func validateAndCompleteEnrollmentRequest(enrollmentRequest *api.EnrollmentReque
 	return nil
 }
 
-func approveAndSignEnrollmentRequest(ca *crypto.CA, enrollmentRequest *api.EnrollmentRequest, approval *v1alpha1.EnrollmentRequestApproval) error {
+func approveAndSignEnrollmentRequest(ca *crypto.CA, enrollmentRequest *api.EnrollmentRequest, approval *api.EnrollmentRequestApproval) error {
```
hexfusion, line 30: I believe we standardized on `v1alpha1` please drop this and if you see `api` usage update. The reason is it is more explicit in the code and updating versions is also more explicit in the diff.  :pray:

internal/service/fleet_test.go
```diff
@@ -0,0 +1,249 @@
+package service
+
+import (
+	"context"
+	"testing"
+
+	api "github.com/flightctl/flightctl/api/v1alpha1"
```
hexfusion, line 7: nit: :)

hexfusion (commented):
First pass, code looks reasonable in general. I think I am missing where the interfaces were updated?

internal/service/common.go
rawagner: great catch. I'm passing a real context now.

internal/service/enrollmentrequest.go
rawagner: makes sense. Updated

internal/service/common.go
rawagner: ctx added

rawagner:
> First pass, code looks reasonable in general. I think I am missing where the interfaces were updated?

what interfaces ? The only ones Im aware are in the server/client.gen.go which are updated.

hexfusion:
> > First pass, code looks reasonable in general. I think I am missing where the interfaces were updated?
> 
> what interfaces ? The only ones Im aware are in the server/client.gen.go which are updated.

I guess you are right I was confusing [Store](https://github.com/flightctl/flightctl/blob/abbf1a54bf9b179563e54522ff0428efc3f89588/internal/store/device.go#L22)

internal/service/common.go
```diff
@@ -1,13 +1,83 @@
...
+		return err
+	}
+	httpReq := &http.Request{
+		Method: "PUT",
+		URL:    url,
+		Body:   io.NopCloser(bytes.NewReader(obj)),
+		Header: http.Header{"Content-Type": []string{"application/json"}},
+	}
+
+	router, _ := gorillamux.NewRouter(swagger)
```
hexfusion, line 44: nit: please return error

internal/service/device.go
```diff
@@ -212,3 +214,57 @@ func (h *ServiceHandler) GetRenderedDeviceSpec(ctx context.Context, request serv
...
+	}
+	if currentObj.ApiVersion != newObj.ApiVersion {
+		return server.PatchDevice400JSONResponse{Message: "apiVersion is immutable"}, nil
+	}
+	if currentObj.Kind != newObj.Kind {
+		return server.PatchDevice400JSONResponse{Message: "kind is immutable"}, nil
+	}
+	if !reflect.DeepEqual(currentObj.Status, newObj.Status) {
+		return server.PatchDevice400JSONResponse{Message: "status is immutable"}, nil
+	}
```
hexfusion, line 253: Should we document these validation requirements?

internal/service/device.go
```diff
@@ -212,3 +214,57 @@ func (h *ServiceHandler) GetRenderedDeviceSpec(ctx context.Context, request serv
...
+	if newObj.Metadata.Name == nil || *currentObj.Metadata.Name != *newObj.Metadata.Name {
+		return server.PatchDevice400JSONResponse{Message: "metadata.name is immutable"}, nil
+	}
+	if currentObj.ApiVersion != newObj.ApiVersion {
+		return server.PatchDevice400JSONResponse{Message: "apiVersion is immutable"}, nil
+	}
+	if currentObj.Kind != newObj.Kind {
+		return server.PatchDevice400JSONResponse{Message: "kind is immutable"}, nil
+	}
+	if !reflect.DeepEqual(currentObj.Status, newObj.Status) {
```
hexfusion, line 251: this is tricky because we have time stamps here, will that cause problems?

internal/service/device.go
```diff
@@ -212,3 +214,57 @@ func (h *ServiceHandler) GetRenderedDeviceSpec(ctx context.Context, request serv
...
 }
+
+// (PATCH /api/v1/devices/{name})
+func (h *ServiceHandler) PatchDevice(ctx context.Context, request server.PatchDeviceRequestObject) (server.PatchDeviceResponseObject, error) {
+	orgId := store.NullOrgId
+
+	currentObj, err := h.store.Device().Get(ctx, orgId, request.Name)
+	if err != nil {
+		switch err {
+		case flterrors.ErrResourceIsNil:
```
hexfusion, line 225: do we need explicit cases here?
```suggestion
		case flterrors.ErrResourceIsNil, flterrors.ErrResourceNameIsNil:
```

internal/service/common.go
rawagner: added. Also for the line below

internal/service/device.go
rawagner: It should not AFAIK, all the types are just string/int or some other primitives so `reflect.DeepEqual` will work for these.

internal/service/device.go
rawagner: We dont actually. Thanks, I've updated the PR!

internal/service/device.go
rawagner: Added a doc which describes that only metadata.labels and spec are patch-able.

mangelajo (approved):

internal/service/device.go
rawagner: it should not though. If it has, it means that patch request changed it and it should be rejected. Or maybe I am missing something :)

PR #277, opened 2024-06-06
MGMT-17920: Add resourceVersion to all resources

avishayt:
For now, we simply base64 encode the UpdatedAt property.


internal/store/model/resource.go
```diff
@@ -56,3 +57,8 @@ func (r *Resource) BeforeCreate(tx *gorm.DB) error {
 	}
 	return nil
 }
+
+func GetResourceVersion(updatedAt time.Time) *string {
+	resourceVersion := base64.StdEncoding.EncodeToString([]byte(updatedAt.String()))
```
hexfusion, line 62: In terms of performance since you already have time.Time object, another option would be using unix time.
```go
resourceVersion := strconv.FormatInt(updatedAt.UnixNano(), 10)
```

api/v1alpha1/openapi.yaml
```diff
@@ -2286,6 +2286,9 @@ components:
           additionalProperties:
             type: string
           description: 'Properties set by the service.'
+        resourceVersion:
+          type: string
+          description: An opaque string that identifies the server's internal version of an object.
```
hexfusion, line 2291: should we document that this is intended to be unique on a fleet level vs globally?

hexfusion (approved):

api/v1alpha1/openapi.yaml
hexfusion: Nmd I guess it's self documenting :)

internal/store/model/resource.go
avishayt: I like.  Done.

hexfusion (approved):

PR #279, opened 2024-06-06
OCTOET-412 Add in-e2e-cluster git service

mangelajo:
Creates a git server container image, with two ssh commands:
* create-repo
* delete-repo

which can be used from E2E tests to create/delete repos on runtime.

It supports authentication via the ssh key stored in `bin/.ssh/id_rsa` as user `user`, and also accepts the "user" password.


test/scripts/agent-images/agent-images.mk
```diff
@@ -8,5 +8,5 @@ e2e-agent-images: bin rpm bin/e2e-certs
 
 clean-e2e-agent-images:
 	sudo rm -f bin/output/qcow2/disk.qcow2
-	rm -f out/.e2e-agent-images
+	rm -f bin/.e2e-agent-images
```
hexfusion, line 11: question/note it would be nice if we generated a build dir or such that contained all the files generated. Having these hidden in bin is not super intuitive,

test/scripts/git-server-cmds/create-repo
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
```
hexfusion, line 2: should we standardize that all bash scripts use

`set -euo pipefail` or at minimum `-e`

test/scripts/git-server-cmds/delete-repo
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# delete a git repo in /home/user/repos/$1.git
+# $1 is the name of the repo
```
hexfusion, line 4: can these two files be a single file we import and wrap these in functions? just trying to reduce bash sprawl

hexfusion (commented):
cool!

test/scripts/prepare_git_server.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
hexfusion, line 1: :+1:  I would love to see us standardize to using this shebang invocation for all bash WDYT?

hexfusion:
Last thing could we add a shellcheck to the linter if we don't have one? Could be a followup.

https://github.com/marketplace/actions/shellcheck

test/scripts/agent-images/agent-images.mk
mangelajo: that is bin so far, we could eventually change the name to something more generic, I agree with that, because we are not only storing binaries now... but all kinds of artifacts. The nice thing of having all in a directory is that full cleanup has a better warranty rm -rf ./bin ... or rm -f ./output

test/scripts/git-server-cmds/delete-repo
mangelajo: In this case we can't, because this is what the git shell expects, one single binary per git command. Those are the bash scripts that we put on the containers to be found by https://github.com/flightctl/flightctl/pull/279/files#diff-17d3cf62a08f372fc25c4dd5a52e747ba04bca4b89e471a094d76332f556b13bR7 `/usr/bin/git-shell`

test/scripts/git-server-cmds/create-repo
mangelajo: Using set -euo pipefail now where possible, on places where we used $1 $2 ... just -eo pipefail

test/README.md
```diff
@@ -131,8 +131,57 @@ with agents on VMs, or connect the server to the local kind k8s cluster.
 We use ginkgo/gomega for these tests, as they are more complex and require
 more setup and teardown than unit tests.
 
-For E2E testing we build several agent images that we exercise and update through
-in the tests, more details can be found here  [Agent Images](./scripts/agent-image/README.md)
+### Local testing side services
```
hexfusion, line 134: :pray:

test/scripts/git-server-cmds/create-repo
hexfusion: cool

hexfusion (approved):
thanks for quick turn around

mangelajo:
> Last thing could we add a shellcheck to the linter if we don't have one? Could be a followup.
> 
> https://github.com/marketplace/actions/shellcheck

That is a good idea : I opened: https://github.com/flightctl/flightctl/issues/280
since it would probably need lots of tweaks on our non-linted bash :D

PR #281, opened 2024-06-07
NO-ISSUE: Client config improvements

fzdarsky:
- Introduces FLIGHTCTL_TEST_ROOT_DIR env var to make the test root dir configurable.
- Allows client config file to contain relative paths to certs.
- Adds Equal() and DeepCopy() methods to client config.

### DIFF
--- a/internal/agent/config.go
+++ b/internal/agent/config.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path"
 	"path/filepath"
 	"time"
@@ -38,6 +39,8 @@ const (
 	EnrollmentCertFile = "client-enrollment.crt"
 	// name of the enrollment key file
 	EnrollmentKeyFile = "client-enrollment.key"
+	// TestRootDirEnvKey is the environment variable key used to set the file system root when testing.
+	TestRootDirEnvKey = "FLIGHTCTL_TEST_ROOT_DIR"
 )
 
 type Config struct {
@@ -83,7 +86,7 @@ type Config struct {
 }
 
 func NewDefault() *Config {
-	return &Config{
+	c := &Config{
 		ManagementEndpoint:   DefaultManagementEndpoint,
 		EnrollmentEndpoint:   DefaultManagementEndpoint,
 		EnrollmentUIEndpoint: DefaultManagementEndpoint,
@@ -99,14 +102,14 @@ func NewDefault() *Config {
 		reader:               fileio.NewReader(),
 		LogLevel:             logrus.InfoLevel.String(),
 	}
-}
 
-func (cfg *Config) SetTestRootDir(rootDir string) {
-	klog.Warning("Setting testRootDir is intended for testing only. Do not use in production.")
-	if cfg.reader != nil {
-		cfg.reader.SetRootdir(rootDir)
+	if value := os.Getenv(TestRootDirEnvKey); value != "" {
+		klog.Warning("Setting testRootDir is intended for testing only. Do not use in production.")
+		c.testRootDir = filepath.Clean(value)
+		c.reader.SetRootdir(c.testRootDir)
 	}
-	cfg.testRootDir = rootDir
+
+	return c
 }
 
 func (cfg *Config) GetTestRootDir() string {
--- a/internal/client/config.go
+++ b/internal/client/config.go
@@ -20,10 +20,18 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+const (
+	// TestRootDirEnvKey is the environment variable key used to set the file system root when testing.
+	TestRootDirEnvKey = "FLIGHTCTL_TEST_ROOT_DIR"
+)
+
 // Config holds the information needed to connect to a FlightCtl API server
 type Config struct {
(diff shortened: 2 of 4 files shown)


mangelajo (approved):
Looks good to me

internal/agent/config.go
```diff
@@ -99,14 +103,15 @@ func NewDefault() *Config {
...
-}
 
-func (cfg *Config) SetTestRootDir(rootDir string) {
-	klog.Warning("Setting testRootDir is intended for testing only. Do not use in production.")
-	if cfg.reader != nil {
-		cfg.reader.SetRootdir(rootDir)
+	if value := os.Getenv(TestRootDirEnvKey); value != "" {
+		klog.Warning("Setting testRootDir is intended for testing only. Do not use in production.")
+		c.reader.SetRootdir(value)
+		c.testRootDir = value
```
hexfusion, line 110: Do we need to mutate the config in the agent vs cmd? Ideally, the config is complete and validated before here.

internal/client/config.go
```diff
@@ -20,10 +21,23 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+var (
+	// TestRootDir is the root directory for test files.
+	testRootDir string
```
hexfusion, line 26: do we need to make this a global or could it be done another way? When I make a new client could the ENV be consumed?

internal/client/config.go
```diff
@@ -56,6 +70,74 @@ type AuthInfo struct {
 	ClientKeyData []byte `json:"client-key-data,omitempty" datapolicy:"security-key"`
 }
 
+func (c *Config) Equal(c2 *Config) bool {
+	if c == c2 {
+		return true
+	}
+	if c == nil || c2 == nil {
```
hexfusion, line 77: the check above would tolerate both objects being nil. Config should always be created with a constructor. Trying to understand what case c would be nil I guess if I did something like below which would deserve a panic.
```
var c *Config
```

internal/client/config.go
```diff
@@ -56,6 +70,74 @@ type AuthInfo struct {
...
+	}
+	return &Config{
+		Service:  *c.Service.DeepCopy(),
+		AuthInfo: *c.AuthInfo.DeepCopy(),
+	}
+}
+
+func (s *Service) DeepCopy() *Service {
+	if s == nil {
+		return nil
```
hexfusion, line 119: If s is nil why not panic this would be a misuse of an object unless I am missing something. If we are going to return a nill object we should also return an error or at least document that it's possible.  But then again would s ever be nil in a sane case?

internal/client/config_test.go
```diff
@@ -22,16 +27,27 @@ const (
 )
 
 func TestValidConfig(t *testing.T) {
-	assert := assert.New(t)
+	require := require.New(t)
+
+	testRootDir, err := os.MkdirTemp("", "flightctl-root")
+	require.NoError(err)
+	defer os.RemoveAll(testRootDir)
```
hexfusion, line 34: optional: t.TempDir() makes a tmp directory and auto-cleans up

hexfusion (commented):
thanks, added a few notes nothing blocking

internal/client/config.go
fzdarsky: If both are `nil`, they are Equal() by definition. I also think the receiver should never be `nil`, but if someone used `cmd.Equal()` on Configs, then this seems to be a case that can happen (https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal).

internal/client/config_test.go
mangelajo: Right I had forgotten about that one

internal/agent/config.go
fzdarsky: The challenged I faced is `Validate()` currently tests whether specified files exist and are readable, so the testRootDir needs to be set before it.

internal/client/config.go
fzdarsky: The idea was to allow recursive copying of structs that may contain optional (thus nullable) fields. Of course, one could test every optional field beforehand, but that'd be quite verbose. Also, the method becomes chain-able when not returning an error.

internal/client/config.go
fzdarsky: I guess we could carry it in the struct like `baseDir`, yes. Would just need to do that in a constructor and was probably shying away from changing so many places where we're using complex literals to create config.

internal/client/config_test.go
fzdarsky: Good point! Will change.

fzdarsky: ✅ done

internal/client/config.go
mangelajo: This would be only an issue if we start running unit tests in parallels. May be we should document this somewhere? I am confident the race detector would catch it as soon as we enabled it though.

mangelajo (approved):

mangelajo:
Nice, I see E2E passes now

internal/client/config.go
hexfusion: I still don't understand how c could be nil but this is non-blocking

hexfusion (approved):

PR #285, opened 2024-06-10
Refresh devices/fleets upon repo update

avishayt:
Two commits:

- MGMT-18098: Track references of fleets/devices to repositories
    Create a join table in SQL and update references from fleets and
    standalone devices to repositories. Later, when a repository is updated
    we can immediately know which fleets and devices are affected.
- MGMT-17945: Refresh devices/fleets upon repo update
    If a repository resource is updated, make sure to:
    1. Create a new TemplateVersion for each fleet that references it
    2. Render each standalone device that references it

### DIFF
--- a/internal/tasks/common.go
+++ b/internal/tasks/common.go
@@ -1,7 +1,7 @@
 package tasks
 
 import (
-	"fmt"
+	"errors"
 
 	api "github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/store/model"
@@ -10,21 +10,7 @@ import (
 
 const ItemsPerPage = 1000
 
-type ErrUnknownConfigName struct {
-	Err error
-}
-
-func (e ErrUnknownConfigName) Error() string {
-	return fmt.Sprintf("failed to find configuration item name: %v", e.Err)
-}
-
-func (e ErrUnknownConfigName) Unwrap() error {
-	return e.Err
-}
-
-func NewUnknownConfigNameError(err error) error {
-	return ErrUnknownConfigName{Err: err}
-}
+var ErrUnknownConfigName = errors.New("failed to find configuration item name")
 
 func getOwnerFleet(device *api.Device) (string, bool, error) {
 	if device.Metadata.Owner == nil {
--- a/internal/tasks/common.go
+++ b/internal/tasks/common.go
@@ -1,12 +1,30 @@
 package tasks
 
 import (
+	"fmt"
+
 	api "github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/store/model"
 	"github.com/flightctl/flightctl/internal/util"
 )
 
 const ItemsPerPage = 1000
+
+type ErrUnknownConfigName struct {
+	Err error
+}
+
+func (e ErrUnknownConfigName) Error() string {
+	return fmt.Sprintf("failed to find configuration item name: %v", e.Err)
+}
+
+func (e ErrUnknownConfigName) Unwrap() error {
+	return e.Err
+}
+
+func NewUnknownConfigNameError(err error) error {
+	return ErrUnknownConfigName{Err: err}
+}
 
 func getOwnerFleet(device *api.Device) (string, bool, error) {
 	if device.Metadata.Owner == nil {
--- a/internal/tasks/device_render.go
+++ b/internal/tasks/device_render.go
@@ -75,8 +75,7 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 			configItem := (*device.Spec.Config)[i]
 			name, err := t.handleConfigItem(ctx, &configItem)
 			if err != nil {
-				var unknownErr ErrUnknownConfigName
-				if errors.As(err, &unknownErr) {
+				if errors.Is(err, ErrUnknownConfigName) {
 					name = "<unknown>"
 				}
 				invalidConfigs = append(invalidConfigs, name)
@@ -116,7 +115,7 @@ func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *ap
 
 	disc, err := configItem.Discriminator()
 	if err != nil {
-		return "", NewUnknownConfigNameError(fmt.Errorf("failed getting discriminator: %w", err))
+		return "", fmt.Errorf("%w: failed getting discriminator: %w", ErrUnknownConfigName, err)
 	}
 
 	switch disc {
@@ -127,14 +126,14 @@ func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *ap
 	case string(api.TemplateDiscriminatorInlineConfig):
 		return t.handleInlineConfig(configItem)
 	default:
-		return "", NewUnknownConfigNameError(fmt.Errorf("unsupported discriminator %s", disc))
+		return "", fmt.Errorf("%w: unsupported discriminator: %s", ErrUnknownConfigName, disc)
 	}
 }
 
 func (t *DeviceRenderLogic) handleGitConfig(ctx context.Context, configItem *api.DeviceSpec_Config_Item) (string, error) {
 	gitSpec, err := configItem.AsGitConfigProviderSpec()
 	if err != nil {
-		return "", NewUnknownConfigNameError(fmt.Errorf("failed getting config item as GitConfigProviderSpec: %w", err))
+		return "", fmt.Errorf("%w: failed getting config item as GitConfigProviderSpec: %w", ErrUnknownConfigName, err)
 	}
 
 	t.repoNames = append(t.repoNames, gitSpec.GitRef.Repository)
@@ -168,7 +167,7 @@ func (t *DeviceRenderLogic) handleGitConfig(ctx context.Context, configItem *api
 func (t *DeviceRenderLogic) handleK8sConfig(configItem *api.DeviceSpec_Config_Item) (string, error) {
 	k8sSpec, err := configItem.AsKubernetesSecretProviderSpec()
 	if err != nil {
-		return "", NewUnknownConfigNameError(fmt.Errorf("failed getting config item as KubernetesSecretProviderSpec: %w", err))
+		return "", fmt.Errorf("%w: failed getting config item as KubernetesSecretProviderSpec: %w", ErrUnknownConfigName, err)
 	}
 
 	return k8sSpec.Name, fmt.Errorf("service does not yet support kubernetes config")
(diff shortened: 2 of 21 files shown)


internal/tasks/device_render.go
```diff
@@ -66,29 +70,54 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
...
+	}
+
 	configjson, err := json.Marshal(t.renderedConfig)
 	if err != nil {
 		return t.setStatus(ctx, err)
 	}
 
+	if device.Metadata.Owner == nil || *device.Metadata.Owner == "" {
+		// The standalone device is valid, set the many-to-may relationship with the repos
+		err = t.store.Device().OverwriteRepositoryRefs(ctx, t.resourceRef.OrgID, *device.Metadata.Name, t.repoNames...)
```
hexfusion, line 106: If Repository is an associated object that is part of Device would it make sense to update the Interface to allow something like this. Instead of bolting Repository methods into device directly. This can be a followup.

```suggestion
		err = t.store.Device(ctx, name).Repositories().Clear(ctx, t.repoNames...)
```

internal/tasks/device_render.go
```diff
@@ -66,29 +70,54 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
...
 
+	invalidConfigs := []string{}
 	if device.Spec != nil && device.Spec.Config != nil {
 		for i := range *device.Spec.Config {
 			configItem := (*device.Spec.Config)[i]
-			err := t.handleConfigItem(ctx, &configItem)
+			name, err := t.handleConfigItem(ctx, &configItem)
 			if err != nil {
-				return t.setStatus(ctx, err)
+				invalidConfigs = append(invalidConfigs, name)
```
hexfusion, line 79: We don't want to lose the original error 
```suggestion
if err != nil {
    if errors.Is(err, ErrUnknownName) {
	invalidConfigs = append(invalidConfigs, unknownName)
   }
   //log/handle original error
 }
```

internal/tasks/device_render.go
```diff
@@ -66,29 +70,54 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
...
 	return t.setStatus(ctx, err)
 }
 
-func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *api.DeviceSpec_Config_Item) error {
+func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *api.DeviceSpec_Config_Item) (string, error) {
+
 	disc, err := configItem.Discriminator()
 	if err != nil {
-		return fmt.Errorf("failed getting discriminator: %w", err)
+		return unknownName, fmt.Errorf("failed getting discriminator: %w", err)
```
hexfusion, line 120: Could we do this another way generally speaking idiomatic go assumes the value returned from a method is not used. If this error means something then wrap the err.

```suggestion
		return fmt.Errorf("failed getting discriminator %w: %w", ErrUnknownName, err)

File: internal/tasks/device_render.go [hunk 4]
```diff
@@ -66,29 +70,54 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
...
+			configurationStr += "s"
+		}
+		err = fmt.Errorf("device has %d invalid %s: %s", len(invalidConfigs), configurationStr, strings.Join(invalidConfigs, ", "))
+		return t.setStatus(ctx, err)
+	}
+
+	if repoRefsErr != nil {
+		return t.setStatus(ctx, repoRefsErr)
+	}
+
```
hexfusion, line 98:
```suggestion
	if err !=  t.store.Device().OverwriteRepositoryRefs(ctx, t.resourceRef.OrgID, *device.Metadata.Name, t.repoNames...); err != nil {
	   return t.setStatus(ctx, repoRefsErr)
       }
 ```

File: internal/tasks/device_render.go [hunk 5]
```diff
@@ -97,77 +126,87 @@ func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *ap
...
+	return gitSpec.Name, nil
 }
 
 // TODO: implement
-func (t *DeviceRenderLogic) handleK8sConfig(configItem *api.DeviceSpec_Config_Item) error {
-	return fmt.Errorf("service does not yet support kubernetes config")
+func (t *DeviceRenderLogic) handleK8sConfig(configItem *api.DeviceSpec_Config_Item) (string, error) {
+	k8sSpec, err := configItem.AsKubernetesSecretProviderSpec()
+	if err != nil {
+		return unknownName, fmt.Errorf("failed getting config item as KubernetesSecretProviderSpec: %w", err)
```
hexfusion, line 172: same here lets wrap error and handle vs use return value on error

internal/tasks/fleet_validate.go
```diff
@@ -56,16 +52,23 @@ type FleetValidateLogic struct {
...
 
 func NewFleetValidateLogic(taskManager TaskManager, log logrus.FieldLogger, store store.Store, resourceRef ResourceReference) FleetValidateLogic {
 	return FleetValidateLogic{taskManager: taskManager, log: log, store: store, resourceRef: resourceRef}
 }
 
 func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Context, fleet *api.Fleet) error {
-	err := t.validateFleetTemplate(ctx, fleet)
-	if err != nil {
-		return fmt.Errorf("validating fleet: %w", err)
+	validationErr := t.validateFleetTemplate(ctx, fleet)
```
hexfusion, line 63: If there is a validation error why do we continue execution and mutate the store or should we return?

internal/tasks/manager.go
```diff
@@ -125,6 +130,10 @@ func (t TaskManager) GetTask(taskName string) ResourceReference {
 	return <-t.channels[taskName]
 }
 
+func (t TaskManager) HasTasks(taskName string) bool {
+	return len(t.channels[taskName]) != 0
```
hexfusion, line 134: Task manager is not thread-safe? Is that documented?

test/util/create_utils.go
```diff
@@ -114,3 +114,29 @@ func CreateTestTemplateVersions(numTemplateVersions int, ctx context.Context, tv
 	}
 	return nil
 }
+
+func CreateRepositories(numRepositories int, ctx context.Context, storeInst store.Store, orgId uuid.UUID) error {
```
hexfusion, line 118: minor: ctx first param :pray:

internal/tasks/manager.go
```diff
@@ -164,17 +173,22 @@ func (t TaskManager) FleetUpdatedCallback(before *model.Fleet, after *model.Flee
 	}
 }
 
+func (t TaskManager) FleetSourceUpdated(orgId uuid.UUID, name string) {
+	ref := ResourceReference{OrgID: orgId, Kind: model.FleetKind, Name: name}
+	t.SubmitTask(ChannelFleetValidate, ref, FleetValidateOpUpdate)
```
hexfusion, line 178: If multiple goroutines SubmitTask concurrently will this panic?

test/util/create_utils.go
avishayt: Yep.  Fixed this one and a few other helper methods.

avishayt (commented):
Thanks for your excellent feedback!

internal/tasks/fleet_validate.go
```diff
@@ -101,9 +107,13 @@ func (t *FleetValidateLogic) validateFleetTemplate(ctx context.Context, fleet *a
 	for i := range *fleet.Spec.Template.Spec.Config {
 		configItem := (*fleet.Spec.Template.Spec.Config)[i]
 		name, err := t.validateConfigItem(ctx, &configItem)
-		t.log.Debugf("Validated config %s from fleet %s/%s: %v", name, t.resourceRef.OrgID, *fleet.Metadata.Name, err)
 		if err != nil {
+			var unknownErr ErrUnknownConfigName
+			if errors.As(err, &unknownErr) {
```
hexfusion, line 112:
```suggestion
			if errors.Is(err, ErrUnknownConfigName) {
```

internal/tasks/common.go
```diff
@@ -1,13 +1,31 @@
...
+	"fmt"
+
 	api "github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/store/model"
 	"github.com/flightctl/flightctl/internal/util"
 )
 
 const ItemsPerPage = 1000
 
+type ErrUnknownConfigName struct {
```
hexfusion, line 13: No need to reimpliment error
```suggestion
var ErrUnknownConfigName = errors.New("failed to find configuration item name")
```

internal/tasks/device_render.go
```diff
@@ -85,10 +112,11 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 	return t.setStatus(ctx, err)
 }
 
-func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *api.DeviceSpec_Config_Item) error {
+func (t *DeviceRenderLogic) handleConfigItem(ctx context.Context, configItem *api.DeviceSpec_Config_Item) (string, error) {
+
 	disc, err := configItem.Discriminator()
 	if err != nil {
-		return fmt.Errorf("failed getting discriminator: %w", err)
+		return "", NewUnknownConfigNameError(fmt.Errorf("failed getting discriminator: %w", err))
```
hexfusion, line 119:
```suggestion
		return "", NewUnknownConfigNameError(fmt.Errorf("%w: failed getting discriminator: %w", ErrUnknownConfigName, err))
```

internal/tasks/repo_update.go
```diff
@@ -0,0 +1,127 @@
...
+	log         logrus.FieldLogger
+	store       store.Store
+	resourceRef ResourceReference
+}
+
+func NewRepositoryUpdateLogic(taskManager TaskManager, log logrus.FieldLogger, store store.Store, resourceRef ResourceReference) RepositoryUpdateLogic {
+	return RepositoryUpdateLogic{taskManager: taskManager, log: log, store: store, resourceRef: resourceRef}
+}
+
+func (t *RepositoryUpdateLogic) HandleRepositoryUpdate(ctx context.Context) {
```
hexfusion, line 52: why not return errors here then allow caller to decide log or handle?

hexfusion (commented):
few small nits then LGTM

internal/tasks/common.go
avishayt: Done

internal/tasks/device_render.go
avishayt: Done

internal/tasks/repo_update.go
avishayt: Done

avishayt:
@hexfusion Fixed your comments - thanks as always for the great feedback!

PR #288, opened 2024-06-12
api: update status per Data Definition Dictionary

hexfusion:
This PR updates the status to reflect the proposal of the Data Dictionary.


api/v1alpha1/README.md
```diff
@@ -4,24 +4,46 @@ This document is intended to provide context to the API definition.
...
 | -------- | ------- |
-| Ready  | True if the Device is ready and can accept new containers otherwise False. This is also the only condition which can be set by the server, otherwise False. |
-| DiskPressure | True if pressure exists on the disk size—that is, if the disk capacity is low; otherwise False. |
-| PIDPressure | True if pressure exists on the processes—that is, if there are too many processes on the node, otherwise False. | 
-| MemoryPressure |True if pressure exists on the device memory—that is, if the device memory is low, otherwise False. |
-| CPUPressure |True if device is experiencing high CPU utilization that might affect the performance of the workloads running on that device, otherwise False. |
-| SystemdUnitsRunning | True is all specified systemd units are running, otherwise False. |
-| ContainersRunning | True is all specified containers are running, otherwise False. |
+| ID | Unique identifier for the application |
+| Name | A human readable name for the application |
```
fzdarsky, line 12: Would the name not be unique and able to serve as identifier like in K8s? Or better asked:
What would be examples of ID and name and where would they come from? Let's say the user deploys a Helm chart with a certain release name, would the application's name be the release name? The name of a deployment?

api/v1alpha1/README.md
```diff
@@ -4,24 +4,46 @@ This document is intended to provide context to the API definition.
...
+| Status | The current lifecycle phase of an application |
+| Ready  | Is the application ready to accept work |
```
fzdarsky, line 14: Is that "true/false" or "x out of y" (i.e. containers in the same pod) as in the DDD?

api/v1alpha1/README.md
```diff
@@ -4,24 +4,46 @@ This document is intended to provide context to the API definition.
...
+| Degraded    | True if the device has reported a transient retryable error or failure during its operations, otherwise False |
+
+#### Applications/Workloads
+These are the core conditions that cover the ability of the Device to accept/manage applications.
+
+| Name    | Desc |
+| -------- | ------- |
+| AppsAvailable  | True if the Device reports no resource pressure and can accept new application; otherwise is False. |
+| AppsDegraded   | True if one or more of the applications has reported error; otherwise False |
+| DiskPressure | True if pressure exists on the disk size—that is, if the disk capacity is low; otherwise False. |
```
fzdarsky, line 47: The DDD describes "warning" and "critical" (= alarm) watermarks.

api/v1alpha1/README.md
```diff
@@ -4,24 +4,46 @@ This document is intended to provide context to the API definition.
...
 | Name    | Desc |
 | -------- | ------- |
-| Available  | True if the controller is available for deploying configurations and postAction commands otherwise False  |
-| Progressing | True if the controller is in the process of reconciling Spec, otherwise False    |
-| Degraded    | True if the controllers functionality is impaired in some way but not unavailable, otherwise False |
\ No newline at end of file
+| Available  | True if the device is available for reconciling changes to the Device Spec otherwise False  |
+| Updating | True if the controller is in the process of reconciling Spec, otherwise False    |
+| Degraded    | True if the device has reported a transient retryable error or failure during its operations, otherwise False |
+
```
fzdarsky, line 39: The problem I see with such conditions is that aggregating them across a set of devices isn't obvious: What does it mean if "20 out of 1000 are updating and 10 out of 1000 are degraded"? Are they degraded because they are updating? Or are these actually different devices from those updating?

api/v1alpha1/README.md
hexfusion: Ready is true or false. There is currently no abstraction for an application with multiple containers or a Pod. So each application type would implement this based on the data it has today.

api/v1alpha1/README.md
hexfusion: that would be implementation details but I will update the description.
`DiskPressure=true, Reason=Warning, Message="details of warning"`
`DiskPressure=true, Reason=Alarm, Message="details of alarm"`

api/v1alpha1/README.md
hexfusion: They are independent Conditions. 

> What does it mean if "20 out of 1000 are updating

20 of 1000 are updating

> 10 out of 1000 are Degraded

10 of 1000 are Degraded

They could be related but the status is on the device level. So the device has experienced an error of warn level and is Updating. If you click on Degraded you should see a list of devices with degraded status and a message to why.

api/v1alpha1/README.md
hexfusion: ID could be internal and not exposed I don't see a problem with that.

api/v1alpha1/README.md
hexfusion: If I click on Updating I get the next level. This is why Available sort of fits better than healthyIMO. Because if the device hasn't checked in should you Say its unhealthy or not available?

| Name | TemplateVersion | Healthy | Updating | Degraded | Since | Message |
|-----------| ----------| ------------|----------| ------------|----------| ------------|
| device-1| 12345678| True | True | True | 1hr | Failed to pull osImage xyz authentication failed |
| device-2| 12345678| True | True | False | 30m | Downloading OsImage 25% complete |
| device-3| 8888888| False | False | False | 30m | Device has not checked in for 2 days |

PR #289, opened 2024-06-14
More api input validation

fzdarsky:
- Changes device fingerprints from hex to base32 encoding, so they stay within the 63 char limits for resource names.
- Adds input validation of labels and annotations using the validators from Kubernetes.
- Adds input validation for many string field types like OCI image reference.
- Moves input validation from OAPI middleware to service, so it's easier to also apply to calls not coming via API.
- When provided label selectors are invalid, return "400 Bad Request" with the actual validation error instead of a 500.
- Fixes CLI to accept label selection without "="


api/v1alpha1/openapi.yaml
```diff
@@ -1577,21 +1577,26 @@ components:
       properties:
```
celdrake, line 1577: (Comment does not belong here, I don't know where it belongs)

When trying to create a new source of type `InlineConfigProviderSpec`, it is created, but we receive an invalid condition but no much info about why. Is there some issue on the request, or any idea why this is happening?

Request:
```
{
  "apiVersion": "v1alpha1",
  "kind": "Fleet",
  "metadata": {
    "name": "name",
    "labels": {}
  },
  "spec": {
    "selector": {
      "matchLabels": {}
    },
    "template": {
      "metadata": {
        "labels": {
          "fleet": "name"
        }
      },
      "spec": {
        "os": {
          "image": "some-image"
        },
        "config": [
          {
            "configType": "InlineConfigProviderSpec",
            "inline": {
              "one": "a",
              "two": "b"
            },
            "name": "my-config"
          }
        ]
      }
    }
  }
}
```

Response:
```
  "status": {
    "conditions": [
      {
        "lastTransitionTime": "2024-06-14T15:06:34.509800963Z",
        "message": "fleet has 1 invalid configuration: my-config",
        "reason": "Invalid",
        "status": "False",
        "type": "Valid"
      }
    ]
  }
```

api/v1alpha1/openapi.yaml
```diff
@@ -1791,6 +1817,8 @@ components:
...
         image:
           type: string
+          minLength: 1
```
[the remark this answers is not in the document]
fzdarsky: We do, yes. ATM, adding `spec.os` is optional, but when added must contain `spec.os.image`. So if a user doesn't want to have the OS updated, would they leave out `spec.os` or make leave the `spec.os.image` empty? I can see arguments for both. Guess it'll depend on whether there'll be other properties below `spec.os`. I'll remove the min constraint for now.

api/v1alpha1/openapi.yaml
```diff
@@ -1723,12 +1736,17 @@ components:
               properties:
                 repository:
                   type: string
+                  minLength: 1
```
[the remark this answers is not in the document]
fzdarsky: Good catch! Added.

api/v1alpha1/openapi.yaml
fzdarsky: The reason is `config[0].inline` must be valid Ignition (see examples/fleet.yaml). I think the message could be a bit more helpful in that respect.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
celdrake: I missed the fact that `spec.os` is optional, sorry.
I think it was correct as you had it, and we should not send `spec.os` when there's no image.

api/v1alpha1/openapi.yaml
```diff
@@ -2335,17 +2389,21 @@ components:
...
           description: The last time the condition transitioned from one status to another.
         message:
           type: string
+          maxLength: 32768
           description: Human readable message indicating details about last transition.
         reason:
           type: string
+          minLength: 1
+          maxLength: 1024
+          pattern: '^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$'
```
avishayt, line 2408: Do we need to be concerned with outputs?  Same for the other validations for Conditions.

api/v1alpha1/validation.go
```diff
@@ -0,0 +1,41 @@
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateLabels validates that a set of labels are correctly defined.
+func ValidateLabels(labels *map[string]string) error {
```
avishayt, line 10: Please either link to JIRA [MGMT-17069](https://issues.redhat.com/browse/MGMT-17069) or close it as a duplicate

api/v1alpha1/openapi.yaml
celdrake: We'll add more validations in the UI for this particular case, but I think @avishayt 's proposal would help better understand other errors.

avishayt: I opened [MGMT-18179](https://issues.redhat.com/browse/MGMT-18179) to track it.  It's out of scope for this PR.

api/v1alpha1/openapi.yaml
```diff
@@ -2264,24 +2311,33 @@ components:
...
         generation:
           type: integer
           format: int64
+          minimum: 0
           description: A sequence number representing a specific generation of the desired state. Populated by the system. Read-only.
         owner:
           type: string
+          minLength: 0
+          maxLength: 127
+          pattern: '^([a-z0-9]([-a-z0-9]*[a-z0-9])?\/)?[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```
celdrake, line 2338: This pattern is not correct. As implied by the description, it prefixes the value with the `Kind`.

I think this relates to previous comments from @avishayt , should we maybe only restrict the input field? The owner is being set by the system,  and should not be possible to overwrite it from the clients.

With the current pattern, any PUT request on a device owned by a fleet will fail.

api/v1alpha1/validation.go
```diff
@@ -0,0 +1,41 @@
...
+	errs := validateLabels(labels, field.NewPath("metadata").Child("labels"))
+	return errors.Reduce(errs.ToAggregate())
+}
+
+// ValidateAnnotations validates that a set of annotations are correctly defined.
+func ValidateAnnotations(labels *map[string]string) error {
+	errs := validateLabels(labels, field.NewPath("metadata").Child("annotations"))
```
celdrake, line 17: Not all annotations are labels. 
When attempting to update a device using PUT method, an error 400 is received. This is a device that is bound to a Resource sync and has the following annotations:

```
    "annotations": {
      "device-controller/renderedVersion": "2",
      "fleet-controller/templateVersion": "2024-06-18T09:59:44Z"
    },
```

avishayt: Annotations shouldn't be settable by the user according to how we defined them

celdrake: Yes, but we're currently using PUT since that's the only request available, and the entire object needs to comply with the spec, which implies sending all mandatory fields, etc. 

So either we can blank them on the UI, or the API controller could ignore them for the validations, since anyway there are blanked out later in `NilOutManagedObjectMetaProperties`.

Also, this would not happen once we have the PATCH request available either.

avishayt: Right, the backend ignores any annotations set by the user, so the UI doesn't need to allow setting them, and the API doesn't need to validate them either I suppose.

rawagner: Annotations are not as strict as labels in k8s. For a correct k8s annotation validation, we should use https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L44

oglok:
@fzdarsky needs a rebase :-(

api/v1alpha1/openapi.yaml
fzdarsky: I've dropped validation of Conditions for now, but think that anything that can be modified via the API needs validation. Will add in a follow-up PR.

api/v1alpha1/validation.go
fzdarsky: Done

api/v1alpha1/openapi.yaml
fzdarsky: I've removed this validation and will add a fixed version in a follow-up PR.

api/v1alpha1/validation.go
fzdarsky: Updated the validation to use the K8s code.

internal/service/enrollmentrequest.go
```diff
@@ -76,6 +76,12 @@ func (h *ServiceHandler) createDeviceFromEnrollmentRequest(ctx context.Context,
 // (POST /api/v1/enrollmentrequests)
 func (h *ServiceHandler) CreateEnrollmentRequest(ctx context.Context, request server.CreateEnrollmentRequestRequestObject) (server.CreateEnrollmentRequestResponseObject, error) {
 	orgId := store.NullOrgId
+	if err := api.ValidateLabels(request.Body.Metadata.Labels); err != nil {
```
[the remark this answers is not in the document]
fzdarsky: Done.

internal/util/validation/formats.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// Validates an OCI image reference.
+func ValidateOciImageReference(s *string, path string) []error {
+	return ValidateString(s, path, 1, OciImageReferenceMaxLength, OciImageReferenceRegexp, OciImageReferenceFmt, "quay.io/flightctl/flightctl:latest")
+}
+
+const (
+	// as per https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags
+	GitRevisionFmt string = `[a-zA-Z]([a-zA-Z0-9\.\-\_\/])*`
```
celdrake, line 45: This regexp does not work well with hashes, eg

```
spec.config[].gitRef.targetRevision: Invalid value: "9fac431b7f4f319ead0195034064012e732bbb0c": Invalidpattern (regex used for validation is '[a-zA-Z]([a-zA-Z0-9\.\-\_\/])*')
```

So a PUT with a device having such a hash would fail.

api/v1alpha1/validation.go
celdrake: Issue fixed with the last changes, thanks

internal/service/enrollmentrequest.go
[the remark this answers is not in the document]
celdrake: Fixed as well

internal/util/validation/formats.go
fzdarsky: Done.

avishayt (approved):
Looks great!

internal/util/validation/formats.go
celdrake: Worked well now :)

celdrake (approved):

PR #291, opened 2024-06-17
MGMT-18183: Add AMQP queues infrastructure

ori-amizur:
Add AMQP queues infrastrcture to pkg.
It includes simplified interface for consuming and producing queues payload

/cc @hexfusion

### DIFF
--- a/pkg/queues/amqp_provider.go
+++ b/pkg/queues/amqp_provider.go
@@ -1,0 +1,182 @@
+package queues
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
+
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/reqid"
+	"github.com/go-chi/chi/v5/middleware"
+	amqp "github.com/rabbitmq/amqp091-go"
+	"github.com/sirupsen/logrus"
+)
+
+type amqpProvider struct {
+	url     string
+	log     logrus.FieldLogger
+	wg      *sync.WaitGroup
+	queues  []*amqpQueue
+	stopped atomic.Bool
+	mu      sync.Mutex
+}
+
+func NewAmqpProvider(url string, log logrus.FieldLogger) Provider {
+	var wg sync.WaitGroup
+	wg.Add(1)
+	return &amqpProvider{
+		url: url,
+		log: log,
+		wg:  &wg,
+	}
+}
+
+func (r *amqpProvider) newQueue(queueName string) (*amqpQueue, error) {
+	var (
+		err        error
+		connection *amqp.Connection
+		channel    *amqp.Channel
+	)
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.stopped.Load() {
+		return nil, errors.New("provider is stopped")
+	}
+	defer func() {
+		if err != nil {
+			if channel != nil {
+				channel.Close()
+			}
+			if connection != nil {
+				connection.Close()
+			}
+		}
+	}()
+	connection, err = amqp.Dial(r.url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create connection: %w", err)
+	}
+	channel, err = connection.Channel()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create channel: %w", err)
+	}
+	_, err = channel.QueueDeclare(queueName,
+		true,  // durable
+		false, // auto delete
+		false, // exclusive
+		false, // no wait
+		nil)   // args
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to declare queue %s: %w", queueName, err)
+	}
+	ret := &amqpQueue{
+		name:       queueName,
+		connection: connection,
+		channel:    channel,
+		log:        r.log,
+		wg:         r.wg,
+	}
+	r.queues = append(r.queues, ret)
+	return ret, nil
+}
+
+func (a *amqpProvider) NewConsumer(queueName string) (Consumer, error) {
+	return a.newQueue(queueName)
+}
+
+func (a *amqpProvider) NewPublisher(queueName string) (Publisher, error) {
+	return a.newQueue(queueName)
+}
+
+func (a *amqpProvider) Stop() {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	if a.stopped.Swap(true) {
+		return
+	}
+	defer a.wg.Done()
+	for _, q := range a.queues {
+		q.Close()
+	}
+}
+
+func (a *amqpProvider) Wait() {
+	a.wg.Wait()
+}
+
+type amqpQueue struct {
+	connection *amqp.Connection
+	channel    *amqp.Channel
+	name       string
+	wg         *sync.WaitGroup
+	log        logrus.FieldLogger
+	closed     atomic.Bool
+}
+
+func (r *amqpQueue) Publish(payload []byte) error {
+	if r.closed.Load() {
+		return errors.New("queue is closed")
+	}
+	return r.channel.Publish("",
+		r.name, // queue name
+		false,  // mandatory
+		false,  // immediate
+		amqp.Publishing{
+			DeliveryMode: amqp.Persistent,
+			ContentType:  "text/plain",
+			Body:         payload,
+		})
+}
+
+func (r *amqpQueue) Consume(ctx context.Context, handler ConsumeHandler) error {
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
+	msgs, err := r.channel.ConsumeWithContext(ctx,
+		r.name,
+		"",    // consumer (exchange)
+		false, // auto ack
+		false, // exclusive
+		false, // no local
+		false, // no wait
+		nil)   // args
+	if err != nil {
+		cancel()
+		return err
+	}
+	r.wg.Add(1)
+	go func() {
+		var err error
+		defer r.wg.Done()
+		defer cancel()
+		for d := range msgs {
+			requestID := reqid.NextRequestID()
+			reqCtx := context.WithValue(ctx, middleware.RequestIDKey, requestID)
+			log := log.WithReqIDFromCtx(reqCtx, r.log)
+			if err = handler(reqCtx, d.Body, log); err != nil {
+				log.WithError(err).Errorf("failed to consume message: %s", string(d.Body))
+			}
+			if err = d.Ack(false); err != nil {
+				log.WithError(err).Errorf("failed to acknowledge message")
+			}
+		}
+		if !r.closed.Load() {
+			r.log.Fatal("channel was closed by AMQP provider")
+		}
+	}()
+	return nil
+}
+
+func (a *amqpQueue) Close() {
+	if a.closed.Swap(true) {
+		return
+	}
+	if a.channel != nil {
+		a.channel.Close()
+	}
+	if a.connection != nil {
+		a.connection.Close()
+	}
+}
--- a/pkg/queues/provider.go
+++ b/pkg/queues/provider.go
@@ -1,0 +1,26 @@
+package queues
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+)
+
+type Provider interface {
+	NewConsumer(queueName string) (Consumer, error)
+	NewPublisher(queueName string) (Publisher, error)
+	Stop()
+	Wait()
+}
+
+type ConsumeHandler func(ctx context.Context, payload []byte, log logrus.FieldLogger) error
+
+type Consumer interface {
+	Consume(ctx context.Context, handler ConsumeHandler) error
+	Close()
+}
+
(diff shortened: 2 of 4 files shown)


pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+func NewAmqpProvider(url string, log logrus.FieldLogger) Provider {
+	var wg sync.WaitGroup
+	ctx, cancel := context.WithCancel(context.Background())
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		<-ctx.Done()
+	}()
+	return &amqpProvider{
+		ctx:    ctx,
```
hexfusion, line 35: question: do we need to embed ctx vs pass as a param?

pkg/queues/provider.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+)
+
+type Provider interface {
+	NewConsumer(queueName string) (Consumer, error)
+	NewPublisher(queueName string) (Publisher, error)
```
hexfusion, line 11: Do the Consumer and Publisher constructors need to be in the interface?

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+
+		}
+	}()
+	return nil
+}
+
+func (a *amqpQueue) close() {
+	a.closed = true
+	if a.channel != nil {
+		a.channel.Close()
```
hexfusion, line 184: question: what happens if close is called more than once is it idempotent?

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func NewAmqpProvider(url string, log logrus.FieldLogger) Provider {
+	var wg sync.WaitGroup
+	ctx, cancel := context.WithCancel(context.Background())
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		<-ctx.Done()
+	}()
```
hexfusion, line 33: feels a little strange for a constructor to start a go routine that just waits for ctx to be cancelled.

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+	defer func() {
+		if err != nil {
+			if channel != nil {
+				channel.Close()
+			}
+			if connection != nil {
+				connection.Close()
+			}
+		}
+	}()
```
hexfusion, line 61: would it be cleaner to declare the defer statement after you instantiate the variables?

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+		ctx:        r.ctx,
+		wg:         r.wg,
+	}
+	r.queues = append(r.queues, ret)
+	return ret, nil
+}
+
+func (a *amqpProvider) NewConsumer(queueName string) (Consumer, error) {
+	q, err := a.newQueue(queueName)
+	return q, err
```
hexfusion, line 94:
```suggestion
	return a.newQueue(queueName)
```

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+		false,  // mandatory
+		false,  // immediate
+		amqp.Publishing{
+			DeliveryMode: amqp.Persistent,
+			ContentType:  "text/plain",
+			Body:         payload,
+		})
+}
+
+func (r *amqpQueue) Consume(handler ConsumeHandler) error {
```
hexfusion, line 138: As noted above lets pass context
```suggestion
func (r *amqpQueue) Consume(ctx context.Context, handler ConsumeHandler) error {
```

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (a *amqpProvider) NewConsumer(queueName string) (Consumer, error) {
+	q, err := a.newQueue(queueName)
+	return q, err
+}
+
+func (a *amqpProvider) NewPublisher(queueName string) (Publisher, error) {
+	q, err := a.newQueue(queueName)
+	return q, err
```
hexfusion, line 99:
```suggestion
	return  a.newQueue(queueName)
```

hexfusion (commented):
Thanks for putting this up!. First pass I have a few questions on the general structure, but you are off to a great start!

pkg/queues/amqp_provider.go
ori-amizur: When the provider is stopped the ctx is canceled.  This should cause the consume loop to stop.  If we pass it as param, the consume loop will not be stopped.

pkg/queues/provider.go
ori-amizur: They rely on the provider since they are using variables that are part of the provider.

pkg/queues/amqp_provider.go
hexfusion: I think the general problem is that context is meant to be passed as a param not internalized in a struct[1].  You can achieve the same thing and only register the cancelFn to the struct.

[1] https://go.dev/blog/context-and-structs

pkg/queues/amqp_provider.go
ori-amizur: Good point.  I believe it should be OK but I will check.

pkg/queues/amqp_provider.go
ori-amizur: The intention was that `Wait` can called on a provider even if there are no consumer loops.  I will try to avoid it.

pkg/queues/amqp_provider.go
ori-amizur: OK

pkg/queues/amqp_provider.go
ori-amizur: This will break the ability to close consumer loops when provider is stopped.

pkg/queues/amqp_provider.go
ori-amizur: OK

pkg/queues/provider.go
hexfusion: Can you decouple those variables? Generally, constructors are functions.

```go
consumer, err := NewConsumer(log )
// handle err

provider.AddQueue(consumer)
```

pkg/queues/amqp_provider.go
ori-amizur: OK

pkg/queues/amqp_provider.go
hexfusion: Only if ctx remains on the struct.

pkg/queues/amqp_provider.go
ori-amizur: Made sure it is called only once

pkg/queues/amqp_provider.go
ori-amizur: How is it the same?  The cancelFn is used to cancel (close) the `ctx.Done()` channel.  The ctx is used to get indication that channel was closed.

ori-amizur: OK.  Maybe I can avoid using context.  Instead I can pass a channel that will be closed when the provider is stopped.  Do you think it is better ?

pkg/queues/provider.go
hexfusion: If the provider holds data that it must share it should expose it as a method.

url := provider.URL()
consumer, err := NewConsumer(log, url)

pkg/queues/amqp_provider.go
hexfusion: Context is useful if you expect the underlying implementation to take a context or if you want to use it for go routine synchronization.  But you want to keep the context external. This assists in graceful termination. But could you just `close(doneCh)`. That is your choice.

```go
// higher level context
ctx, cancelFn :=  context.WithCancel(context.Background())

// give the cancelFn to some actor who controls the lifecycle of the provider?

provider, err  := NewProvider()
// handle err

// here my Run loop which takes a context inside Run I create a child context and pin that cancelFn to the struct.
go provider.Run(ctx)

<-ctx.Done() // Wait?

```

pkg/queues/provider.go
ori-amizur: I thought that provider is a factory for Consumer/Producer

pkg/queues/amqp_provider.go
ori-amizur: OK

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,190 @@
...
+	url     string
+	log     logrus.FieldLogger
+	wg      *sync.WaitGroup
+	queues  []*amqpQueue
+	stopped atomic.Bool
+}
+
+func NewAmqpProvider(url string, log logrus.FieldLogger) Provider {
+	var wg sync.WaitGroup
+	wg.Add(1)
```
hexfusion, line 28: drop

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,190 @@
...
+		false,  // immediate
+		amqp.Publishing{
+			DeliveryMode: amqp.Persistent,
+			ContentType:  "text/plain",
+			Body:         payload,
+		})
+}
+
+func (r *amqpQueue) Consume(ctx context.Context, handler ConsumeHandler) error {
+	ctxWithCancel, cancel := context.WithCancel(ctx)
```
hexfusion, line 135:
```suggestion
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
```

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+	r.wg.Add(1)
+	go func() {
+		var err error
+		defer r.wg.Done()
+		defer cancel()
+		for {
+			select {
+			case <-r.doneCh:
```
hexfusion, line 155: Since you have ctx as long as the parent ctx CancelFn is on the struct you don't need the done channel. When the parent is canceled so will its childern.

```suggestion
			case <-ctx.Done():
```

pkg/queues/amqp_provider.go
```diff
@@ -0,0 +1,190 @@
...
+	msgs, err := r.channel.ConsumeWithContext(ctxWithCancel,
+		r.name,
+		"",    // consumer (exchange)
+		false, // auto ack
+		false, // exclusive
+		false, // no local
+		false, // no wait
+		nil)   // args
+	if err != nil {
+		cancel()
```
hexfusion, line 145: if you defer above you can drop this

hexfusion (commented):
Thanks for quick updates, looking good.

pkg/queues/amqp_provider.go
ori-amizur: I want to be able to wait on provider even if there are no consumer loops yet.  How can I achieve it without it?

pkg/queues/amqp_provider.go
ori-amizur: OK

pkg/queues/amqp_provider.go
ori-amizur: I don't want to call it, unless there is a failure.  Otherwise, cancel will be called from the goroutine.

pkg/queues/amqp_provider.go
ori-amizur: It wouldn't help if the provider is stopped, since now the ctx is received as parameter.

pkg/queues/amqp_provider.go
ori-amizur: For example, I would like the following code to run properly:
```
   provider := queues.NewAmqpProvider(url, log)
   ...
   // Wait until provider is stopped
   provider.Wait() 
```

avishayt:
Please fix lint issues

pkg/queues/amqp_provider.go
hexfusion: So to be clear what you are doing would technically work it is just not proper usage of a wait.Group. Wait group is intended for waiting for goroutines to finish[1]. So if you wanted to just block with no goroutines you have a couple of options. Add back your doneCh or wait on the ctx.

```go
func (a *amqpProvider) Wait() {
        <-a.doneCh
	a.wg.Wait()
}
```
```go
func (a *amqpProvider) Wait(ctx context.Context) {
        <-ctx.Done()
	a.wg.Wait()
}
```

[1]  https://pkg.go.dev/sync#WaitGroup

pkg/queues/amqp_provider.go
hexfusion: gotcha you're right I missed that usage.

pkg/queues/amqp_provider.go
hexfusion: As we discussed offline change here is not necessary.

hexfusion (approved):

hexfusion (approved):

hexfusion (approved):

PR #294, opened 2024-06-18
MGMT-18179: Remove duplication between config validate and render

avishayt:
1. Combined the code that validates a fleet template and the code that renders a device since they are very similar.
2. Add an error message to the condition if validation fails.

### DIFF
--- a/internal/tasks/device_render.go
+++ b/internal/tasks/device_render.go
@@ -15,6 +15,7 @@ import (
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/flightctl/flightctl/pkg/reqid"
 	"github.com/go-chi/chi/v5/middleware"
+	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
 )
@@ -45,13 +46,10 @@ func DeviceRender(taskManager TaskManager) {
 }
 
 type DeviceRenderLogic struct {
-	taskManager    TaskManager
-	log            logrus.FieldLogger
-	store          store.Store
-	resourceRef    ResourceReference
-	device         *api.Device
-	renderedConfig *config_latest_types.Config
-	repoNames      []string
+	taskManager TaskManager
+	log         logrus.FieldLogger
+	store       store.Store
+	resourceRef ResourceReference
 }
 
 func NewDeviceRenderLogic(taskManager TaskManager, log logrus.FieldLogger, store store.Store, resourceRef ResourceReference) DeviceRenderLogic {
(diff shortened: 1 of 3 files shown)


test/integration/tasks/fleet_validate_test.go
```diff
@@ -287,7 +275,7 @@ var _ = Describe("FleetValidate", func() {
 			Expect(*fleet.Status.Conditions).To(HaveLen(1))
 			Expect((*fleet.Status.Conditions)[0].Type).To(Equal(api.FleetValid))
 			Expect((*fleet.Status.Conditions)[0].Status).To(Equal(api.ConditionStatusFalse))
-			Expect(*((*fleet.Status.Conditions)[0].Message)).To(Equal("fleet has 1 invalid configuration: <unknown>"))
+			Expect(*((*fleet.Status.Conditions)[0].Message)).To(Equal("1 invalid configuration: <unknown>. Error: failed to find configuration item name: unsupported discriminator: InvalidProviderSpec"))
```
avishayt, line 278: FYI @celdrake , this is the improvement we discussed

internal/tasks/device_render.go
```diff
@@ -63,129 +61,193 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
...
-		}
+	// If device.Spec or device.Spec.Config are nil, we still want to render an empty ignition config
+	var config *[]api.DeviceSpec_Config_Item
+	if device.Spec != nil {
+		config = device.Spec.Config
 	}
 
+	renderedConfig, repoNames, renderErr := renderConfig(ctx, t.resourceRef.OrgID, t.store, config, false)
+
+	// Set the many-to-may relationship with the repos (we do this even if the render failed so that we will
```
celdrake, line 75:
```suggestion
	// Set the many-to-many relationship with the repos (we do this even if the render failed so that we will
```

internal/tasks/fleet_validate.go
```diff
@@ -53,25 +44,29 @@ type FleetValidateLogic struct {
...
-	validationErr := t.validateFleetTemplate(ctx, fleet)
+func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Context) error {
+	fleet, err := t.store.Fleet().Get(ctx, t.resourceRef.OrgID, t.resourceRef.Name)
+	if err != nil {
+		return fmt.Errorf("failed getting fleet %s/%s: %w", t.resourceRef.OrgID, t.resourceRef.Name, err)
+	}
+
+	_, repoNames, validationErr := renderConfig(ctx, t.resourceRef.OrgID, t.store, fleet.Spec.Template.Spec.Config, true)
 
 	// Set the many-to-may relationship with the repos (we do this even if the validation failed so that we will
```
celdrake, line 61:
```suggestion
	// Set the many-to-many relationship with the repos (we do this even if the validation failed so that we will
```

test/integration/tasks/fleet_validate_test.go
celdrake: Thanks, I checked and I could see the message when for example passing an invalid ignition version.

celdrake (approved):

internal/tasks/device_render.go
avishayt: Fixed

internal/tasks/fleet_validate.go
avishayt: Fixed

mangelajo (approved):

PR #295, opened 2024-06-18
OCTOET-476: Create agent api/client definitions

mangelajo:
The openapi.yaml for the agent references the main one for schemas.

Subsequent commits will remove some of the methods and paths from the general server, and move those into the agent only API.

This PR updates to oapi-gen v2.3.0 in the second commit, since v.1.15 has some broken behavior for references.


api/v1alpha1/agent/openapi.yaml
```diff
@@ -0,0 +1,182 @@
...
+            application/json:
+              schema:
+                $ref: '../openapi.yaml#/components/schemas/Error'
+        "404":
+          description: NotFound
+          content:
+            application/json:
+              schema:
+                $ref: '../openapi.yaml#/components/schemas/Error'
+  /api/v1/devices/{name}/rendered:
```
mangelajo, line 55:
```
  /api/v1/devices/{name}/rendered:
    $ref: '../openapi.yaml#/paths/~1api~1v1~1devices~1{name}~1rendered'
  /api/v1/enrollmentrequests/{name}:
    $ref: '../openapi.yaml#/paths/~1api~1v1~1enrollmentrequests~1{name}'
```

To avoid duplication we need something like this, but it seems to generate buggy code in oapi-gen v1.x , v2 seems to fix it, checking if it's easy to update to v2.

mangelajo:
@fzdarsky @avishayt working on the next PR that will really create the additional endpoint and remove some methods from the existing one.

Let me know if this makes sense.

api/v1alpha1/openapi.yaml
```diff
@@ -858,18 +858,18 @@ paths:
       description: get the full specification for the specified device
       operationId: getRenderedDeviceSpec
       parameters:
-        - name: name
-          in: path
-          description: Name of the device
-          required: true
-          schema:
-            type: string
```
mangelajo, line 866: Oh I must move this back to the original place, I was hopping to un-trigger the openapi bug for the import paths...

api/v1alpha1/agent/openapi.yaml
```diff
@@ -0,0 +1,187 @@
+openapi: 3.0.1
```
fzdarsky, line 1: Should we structure APIs as `apis/agent/v1alpha1` and `apis/service/v1alpha1` immediately, so they can be versioned independently? Or too disruptive?

fzdarsky (approved):

mangelajo:
oh I did not sign all commits. I will update the PR with the final methods to match the current implementation when I finish the 2nd PR

mangelajo:
Updated with the right methods now...

mangelajo:
@fzdarsky github asking for re-review, I changed the put->post method for enrollment request to match the existing codebase/usage.

fzdarsky (approved):

PR #297, opened 2024-06-18
[WIP] adding console insights identity header parser middleware

akoserwal:
Add insights middleware to parse x-rh-identity header passed from console.

Keeping the middleware commented.


internal/util/xrh_header_parser.go
```diff
@@ -0,0 +1,81 @@
...
+	UserId        string `json:"user_id"`
+	OrgId         string `json:"org_id"`
+	AccountNumber string `json:"account_number"`
+	Username      string `json:"username"`
+}
+
+// Parse JWT token without decode key
+func ParseJWTToken(tokenString string) (DecodedToken, error) {
+	str := tokenString
+	str = strings.Split(str, ".")[1]
```
hexfusion, line 43: question: are you 1000% certain every token passed to this function will contain a period or should we validate len and return an error vs panic?

internal/util/xrh_header_parser.go
```diff
@@ -0,0 +1,81 @@
+package util
```
hexfusion, line 1: could we add a few basic unit tests for these functions?

internal/util/xrh_header_parser.go
```diff
@@ -0,0 +1,81 @@
...
+	decodedIdentity, err := base64.StdEncoding.DecodeString(identityHeader)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding Identity: %v", err)
+	}
+
+	err = json.Unmarshal(decodedIdentity, &XRHIdentity)
+
+	if err != nil {
+		logrus.Errorf("x-rh-identity header is not a valid json: %s. Identity: %s", err.Error(), identityHeader)
+		return nil, fmt.Errorf("x-rh-identity header is not a valid json: %s", err.Error())
```
hexfusion, line 26:
```suggestion
		logrus.Errorf("x-rh-identity header is not valid json: %s. Identity: %s", err.Error(), identityHeader)
		return nil, fmt.Errorf("x-rh-identity header is not valid json: %w", err)
```

internal/auth/headers.go
```diff
@@ -0,0 +1,46 @@
...
+		ctx := r.Context()
+		if header == "" {
+			errString := "Missing authentication"
+			w.WriteHeader(http.StatusForbidden)
+			if _, err := w.Write([]byte(errString)); err != nil {
+				logrus.Errorf("Failed to write response: %v", err)
+			}
+			logrus.Errorf("missing the %s header", XRHIDENTITY)
+			return
+		} else {
```
hexfusion, line 31: optional: since the if block returns you technically don't need the else statement.

hexfusion (commented):
Thanks for the PR! Added a few mostly minor notes.

internal/util/xrh_header_parser.go
akoserwal: A valid JWT will have follow <header>.<payload>.<signature> format

I copied most of the code from chrome-service-backend: https://github.com/RedHatInsights/chrome-service-backend/blob/main/rest/util/xrh_header_parser.go

I believe it tried and tested for console apps. This middleware is only used when flightctl get deployed on console dot environment

I will add another ParseJWT, which decodes the token using a key and verifies the signature.

internal/server/server.go
```diff
@@ -81,6 +81,7 @@ func (s *Server) Run() error {
 		middleware.Logger,
 		middleware.Recoverer,
 		oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapiOpts),
+		// auth.ParseHeaders,
```
fzdarsky, line 84: We need to gate the activation of this with a config option for users to explicitly "opt in". This must only be activated when deployed behind an API gateway that removes headers from all incoming requests and sets the headers new itself.

mangelajo: @akoserwal, that would make sense...

internal/util/xrh_header_parser.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/redhatinsights/platform-go-middlewares/identity"
+	"github.com/sirupsen/logrus"
+)
+
+func ParseXRHIdentityHeader(identityHeader string) (*identity.XRHID, error) {
+	var XRHIdentity identity.XRHID
+	decodedIdentity, err := base64.StdEncoding.DecodeString(identityHeader)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding Identity: %v", err)
```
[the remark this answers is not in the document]
mangelajo: For context, "%w" keeps a reference of the nested errors and can be used to distinguish between types of errors programmatically.

internal/util/xrh_header_parser.go
mangelajo: This is something coming from the external world. We should check the length of the array first to avoid a panic. Return error in case of wrong length.

internal/util/xrh_header_parser.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Parse JWT token without decode key
+func ParseJWTToken(tokenString string) (DecodedToken, error) {
+	str := tokenString
+	str = strings.Split(str, ".")[1]
+	str = strings.ReplaceAll(str, "-/", "+")
+	str = strings.ReplaceAll(str, "_/", "/")
+
+	switch len(str) % 4 {
```
mangelajo, line 47: We should add some unit tests to verify the different possibilities (that will help document what the function is doing exactly).

internal/util/xrh_header_parser.go
```diff
@@ -0,0 +1,81 @@
...
+
+	str = string(data)
+	str, err = url.QueryUnescape(str)
+	if err != nil {
+		return DecodedToken{}, err
+	}
+
+	var res DecodedToken
+	err = json.Unmarshal([]byte(str), &res)
+	fmt.Println(res)
```
mangelajo, line 75:
```suggestion
```

internal/util/xrh_header_parser.go
hexfusion: I would consider this blocking merge as it could take down the entire service.

PR #298, opened 2024-06-18
MGMT-17911: api: update status per Data Definition Dictionary

hexfusion:
This PR updates the status as per Data Definition Dictionary. All API's marked deprecated will be removed in a followup PR.

```
status:
  conditions:
  - type: Approved
    status: 'True'
    observedGeneration: 0
    lastTransitionTime: '2024-06-25T15:38:47.913Z'
    message: string
    reason: string
  systemInfo:
    architecture: string
    bootID: string
    operatingSystem: string
  applications:
    data:
      additionalProp1:
        name: string
        ready: string
        restarts: 0
        status: Preparing
      additionalProp2:
        name: string
        ready: string
        restarts: 0
        status: Preparing
      additionalProp3:
        name: string
        ready: string
        restarts: 0
        status: Preparing
    summary:
      status: Healthy
      info: string
  resource:
    cpu: Healthy
    memory: Healthy
    disk: Healthy
  integrity:
    summary:
      status: Passed
      info: string
  config:
    renderedVersion: string
  os:
    image: string
  updated:
    status: UpToDate
    info: string
  summary:
    status: Online
    info: string
  updatedAt: string
 ```

### CODE REVIEW

File: api/v1alpha1/openapi.yaml [hunk 1]
```diff
@@ -2502,6 +2532,231 @@ components:
...
+          description: 'Human readable information about the last update phase transition.'
+    DeviceUpdatePhaseType:
+      type: string
+      enum:
+      - 'Downloading'
+      - 'Verifying'
+      - 'Updating'
+      - 'Rebooting'
+      - 'RollingBack'
+      - 'Failed'
```
mangelajo, line 2674: - 'RolledBack' ? to indicate that the device has been automatically rolled back by the OS,.. and that may require action (you would see the device is booted to the previous image, and understand why that happens because of the 'RolledBack' status.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2532,231 @@ components:
...
+      type: object
+      properties:
+        status:
+          $ref: '#/components/schemas/DeviceApplicationSummary'
+          description: 'Summary status of device applications.'
+        applications:
+          type: array
+          description: 'Status of device applications.'
+          items:
+            $ref: '#/components/schemas/DeviceApplication'
```
mangelajo, line 2545: I think we should not focus on the "Application" statuses yet. This is something we may want to look at when we have the next MVP version.

Because we need to define what an application is, and how do we monitor those.

While for now we were just monitoring simple systemd and container processes.

All the other status bits

api/v1alpha1/openapi.yaml
```diff
@@ -2547,16 +2802,16 @@ components:
...
+      - 'Ready'                # Device (deprecated)
+      - 'DiskPressure'         # Device (deprecated)
+      - 'MemoryPressure'       # Device (deprecated)
+      - 'PIDPressure'          # Device (deprecated)
+      - 'CPUPressure'          # Device (deprecated)
+      - 'Available'            # Device (deprecated)
+      - 'Progressing'          # Device (deprecated)
+      - 'Degraded'             # Device (deprecated)
+      - 'SystemdUnitsRunning'  # Device (deprecated)
+      - 'ContainersRunning'    # Device (deprecated)
```
mangelajo, line 2814: Should we just go ahead and remove them? may be in a follow up PR if those are used from code let's first agree on the API change.

api/v1alpha1/openapi.yaml
hexfusion: @fzdarsky ?

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2532,231 @@ components:
...
+      - 'DeviceRuntimeUnknown'
+    DeviceUpdateStatus:
+      type: object
+      properties:
+        status:
+          $ref: '#/components/schemas/DeviceUpdateSummaryStatus'
+          description: 'Status of the device update.'
+        phase:
+          $ref: '#/components/schemas/DeviceUpdatePhaseStatus'
+          description: 'The passphrase for the private key.'
```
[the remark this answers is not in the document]
hexfusion: yeah this description is wrong good eyes

api/v1alpha1/openapi.yaml
hexfusion: Right, removal will require code.

api/v1alpha1/openapi.yaml
```diff
@@ -2005,6 +2005,36 @@ components:
           description: 'Current state of systemd units on the device.'
           items:
             $ref: '#/components/schemas/DeviceSystemdUnitStatus'
+        applications:
```
fzdarsky, line 2008: Do we still need `containers:` when we have this?

api/v1alpha1/openapi.yaml
```diff
@@ -2005,6 +2005,36 @@ components:
...
+          type: object
+          description: 'Current device integrity status.'
+          items:
+            $ref: '#/components/schemas/DeviceIntegrityStatus'
+        config:
+          type: object
+          description: 'Current device config status.'
+          items:
+            $ref: '#/components/schemas/DeviceConfigStatus'
+        resources:
```
fzdarsky, line 2033: The `runtime:`/`system:` status already factors in the status of the individual resources (in "degraded" and "error" states). I'm wondering whether we should either keep these individual resource statuses under `conditions:` like now (if we can figure out the "warn" and "alarm" watermarks) or move them under `runtime:`/`system:`?
Basically, like `applications:` which has a summary status plus list of constituent applications, we could have a system summary status plus list of constituent system states and resources.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2532,231 @@ components:
...
+      - 'Unknown'
+      - 'Unsupported'
+      x-enum-varnames:
+      - 'DeviceIntegrityPassed'
+      - 'DeviceIntegrityFailed'
+      - 'DeviceIntegrityUnknown'
+      - 'DeviceIntegrityUnsupported'
+    DeviceResourceStatus:
+      type: object
+      properties:
```
fzdarsky, line 2705: Do we need to have different schemas or can we have a list of generic resources (that a user may potentially extend in the future)?

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2532,231 @@ components:
...
+      - 'UpToDate'
+      - 'OutOfDate'
+      - 'Updating'
+      - 'Unknown'
+      x-enum-varnames:
+      - 'DeviceUpdateUpToDate'
+      - 'DeviceUpdateOutOfDate'
+      - 'DeviceUpdateUpdating'
+      - 'DeviceUpdateUnknown'
+    DeviceUpdatePhaseStatus:
```
fzdarsky, line 2657: The `updating` condition with the phase in `reason` was totally fine, I think (as we don't intend to filter by phase, only report in tooltip). The missing bit was the `DeviceUpdateStatus` summary status above that we filter/summarize.

api/v1alpha1/openapi.yaml
hexfusion: I agree it should be removed but would require code change if we can agree on the additions in the PR we can remove deprecated fields in a followup which would include code change. I will mark it deprecated for now?

api/v1alpha1/openapi.yaml
hexfusion: Makes sense can we add this to the DDD please?

api/v1alpha1/openapi.yaml
hexfusion: It would be good if you could update the DDD to reflect this or allow me edit access. Maybe a field that designates status vs condition. Then also sortable vs not to ensure we are all on the same page.

api/v1alpha1/openapi.yaml
hexfusion: Makes sense please update the DDD to reflect or provide edit access.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2532,231 @@ components:
...
+      - 'Healthy'
+      - 'Warning'
+      - 'Critical'
+      - 'Unknown'
+      x-enum-varnames:
+      - 'DeviceResourceHealthy'
+      - 'DeviceResourceWarning'
+      - 'DeviceResourceCritical'
+      - 'DeviceResourceUnknown'
+    DeviceConfigStatus:
```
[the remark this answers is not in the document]
hexfusion: Makes sense can we add that to the DDD doc please.

api/v1alpha1/openapi.yaml
fzdarsky: What’s the diff between “RolledBack” and “Failed”?

BTW, I feel we need a condition with which the agent can aignal that some error occurred when trying to reconcile the spec (validation error, download error, rollback, …) and the agent will not try again until the spec has been updated.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2532,231 @@ components:
...
+      properties:
+        status:
+          $ref: '#/components/schemas/DeviceApplicationSummaryType'
+          description: 'Summary status of applications.'
+        info:
+          type: string
+          description: 'Human readable information indicating details about the application status.'
+    DeviceApplicationSummaryType:
+      enum:
+      - 'Healthy'
```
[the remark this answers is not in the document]
hexfusion: I see what you are saying but are you asking that they be the same or?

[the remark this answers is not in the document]
celdrake: What I meant is that "Running" should be removed from `DeviceApplicationSummaryType` (line 2577).
It should only exist in  `DeviceApplicationStatusType` (as it currently does, in line 2591).

That would make "Running" aligned to the DDD.

api/v1alpha1/openapi.yaml
mangelajo: Ok, as per slack conversation, we want to have a definition of the API now.

api/v1alpha1/openapi.yaml
mangelajo: Failed could be: I could not finish download, there is a permanent problem with the image download.

api/v1alpha1/openapi.yaml
hexfusion: systemd also is deprecated and replaced by application. I have marked both as deprecated.

api/v1alpha1/openapi.yaml
hexfusion: I went back and forth with that, generally speaking, API's should be explicit. Having a generalized list adds the complexity of validation is the enum element valid for the type. But I will update it to be generic.

api/v1alpha1/openapi.yaml
hexfusion: removing

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
hexfusion: this ends up being removed

api/v1alpha1/openapi.yaml
hexfusion: RollingBack is a very transient state "hopefully" and would also be overwritten with Rebooting. Failed is the canonical summary. Rolledback would be a reason.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2514,244 @@ components:
...
+    DeviceWorkloadSummaryType:
+      enum:
+      - 'Healthy'
+      - 'Degraded'
+      - 'Error'
+      - 'Unknown'
+      x-enum-varnames:
+      - 'DeviceWorkloadSummaryHealthy'
+      - 'DeviceWorkloadSummaryDegraded'
+      - 'DeviceWorkloadSummaryRunning'
```
celdrake, line 2572:
```suggestion
      - 'DeviceWorkloadSummaryRunning'
```

Please, remove this line as it causes the next variables to be shifted, and the open-api generator produces wrong types:


```
export enum DeviceWorkloadSummaryType {
  DeviceWorkloadSummaryHealthy = 'Healthy',
  DeviceWorkloadSummaryDegraded = 'Degraded',
  DeviceWorkloadSummaryRunning = 'Error',
  DeviceWorkloadSummaryError = 'Unknown',
}

```

hexfusion: oops, fixed

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2511,255 @@ components:
...
+      x-enum-varnames:
+      - 'DeviceResourceStatusHealthy'
+      - 'DeviceResourceStatusWarning'
+      - 'DeviceResourceStatusCritical'
+      - 'DeviceResourceStatusUnknown'
+    DeviceConfigStatus:
+      type: object
+      required:
+        - osImage
+        - templateVersion
```
hexfusion, line 2751: dropping

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2511,254 @@ components:
...
+    DeviceSystemStatus:
+      type: object
+      required:
+        - architecture
+        - bootID
+        - machineID
+        - operatingSystem
+        - measurements
+        - resource
+        - integrity
```
hexfusion, line 2597: drop resource and integrity

api/v1alpha1/openapi.yaml
```diff
@@ -1993,20 +1993,29 @@ components:
...
           description: Statuses of containers in the device.
           items:
             $ref: '#/components/schemas/ContainerStatus'
-        systemdUnits:
+        systemdUnits: # deprecated
           type: array
           description: 'Current state of systemd units on the device.'
           items:
             $ref: '#/components/schemas/DeviceSystemdUnitStatus'
+        workload:
```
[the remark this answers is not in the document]
hexfusion: I was thinking virtual machines might not fit the application model. Plus other things we might not of thought of. But I'm open to moving it to root level.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2511,254 @@ components:
...
+        memory:
+          $ref: '#/components/schemas/DeviceResourceStatusType'
+          description: 'Status of the device memory resources.'
+        disk:
+          $ref: '#/components/schemas/DeviceResourceStatusType'
+          description: 'Status of the device disk resources.'
+        summary:
+          $ref: '#/components/schemas/DeviceResourceStatusSummary'
+          description: 'Summary status of the device resources.'
+    DeviceResourceStatusSummary:
```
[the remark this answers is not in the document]
hexfusion: Yeah I wanted to continue the pattern and see how it looked but it didn't fit will remove it.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
fzdarsky: Yeah, I agree workloads are broader and cover non-applications, too. system services probably also fall into that bucket. More importantly, though, applications are the a world view from the user down, what performs the tasks they want to get done. Workload is the view from the system up, it’s what puts load onto system resources. VMs can be both: a workload or application (appliance).

To sum up, I agree there’d potentially be other workloads. Would they be summarized under the ApplicationStatusSummary, too?

Somehow I’m wondering whether the schema should be flatter, i.e. whether we need the workload/system/config buckets at the top.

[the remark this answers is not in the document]
hexfusion: If we keep this structure the highest level object would summarize its children. So WorkloadSummary would summarize all workload types.

ApplicationSummary would summarize all applications. High level are all workloads working without error. Then each level provides more granular details.

If every level is on the root. You might get simpler queries but lose the relationship or need to reconstruct it.

I think if we keep the api at most 2 levels it should be fine but we could benchmark to understand the trade offs on read/write?

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2511,254 @@ components:
...
+    DeviceConfigStatus:
+      type: object
+      required:
+        - osImage
+        - summary
+      properties:
+        osImage:
+          type: string
+          description: 'The current device Os image.'
+        templateVersion:
```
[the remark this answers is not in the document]
hexfusion: Yeah this ended up being forced and I agree suboptimal. In general an update is considered like a transaction if any part fails all changes are rolled back.

In that case the only value that needs tracked is renderedVersion right. Since the os etc is declared why does the agent need to reflect that data. The service already has it.

I guess if the service would like to write the field that is another question. Does it concern you that it appears to be a growing number of fields being written to device status by the service? Will that have performance implications long term?

api/v1alpha1/openapi.yaml
```diff
@@ -1993,20 +1993,29 @@ components:
...
           description: 'Current state of systemd units on the device.'
           items:
             $ref: '#/components/schemas/DeviceSystemdUnitStatus'
+        workload:
+          $ref: '#/components/schemas/DeviceWorkloadStatus'
+          description: 'Current status of the device workload.'
+        system:
+          $ref: '#/components/schemas/DeviceSystemStatus'
+          description: 'Current status of the device system.'
+        config:
```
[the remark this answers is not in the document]
hexfusion: This is a good question. In general mirroring is not the case because it's not required. For the reasons I stated about a renderedVersion being like a transaction a single failure means none of the status values will be updated for fields provided by spec.

Also because we have one giant object there is no way for partial sync. For example only add a workload unless that is all that is added.

So what value is there to reflect the values defined by the spec and gated by the renderedVersion? I see little. 

If there is a failure at its logical level it will be summarized. Providing a reason for failure and info to help debug. Also during runtime if a workload fails in the steady state we get that information.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2511,254 @@ components:
...
+          description: 'Number of restarts observed for the application.'
+    DeviceWorkloadSummaryStatus:
+      type: object
+      required:
+        - status
+      properties:
+        status:
+          $ref: '#/components/schemas/DeviceWorkloadSummaryType'
+          description: 'Summary status of workloads.'
+        info:
```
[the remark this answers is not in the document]
hexfusion: info provides context to the status. How that is updated is an implementation detail. For example, if you want this to work like a logger an Error status could be updated on sync if additional or updated context exists. If not it could be updated only once on phase/status change.  There are tradeoffs to both but I don't believe the API declares that directly.

api/v1alpha1/openapi.yaml
```diff
@@ -2502,6 +2511,254 @@ components:
...
+          $ref: '#/components/schemas/DeviceWorkloadSummaryStatus'
+          description: 'Summary status of device workloads.'
+    DeviceApplicationStatus:
+      type: object
+      required:
+        - id
+        - name
+        - status
+      properties:
+        id:
```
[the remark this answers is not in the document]
hexfusion: True but for other resources the name is generated by the service. And even if the service did create this name it would not be a unique instance. For example, a device reboots in terms of auditing when it starts is a second instance of an application. Also if the service or user defines a pod, when it starts that name will be different from the declared name.

It felt like it was needed to track lifecycle phases and would not necessarily be needed to be pushed because its optional. 

I guess I can wrap status in another object during implementation and we can drop it here.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
hexfusion: added renderedVersion to root

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
hexfusion: removed

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
hexfusion: > Somehow I’m wondering whether the schema should be flatter, i.e. whether we need the workload/system/config buckets at the top.

If you can outline the advantage you see for this change it can help me weigh that. My general goal is to provide a simple dependency tree. So the status can be organized logically to ensure future extensibility.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
hexfusion: One area of clarification would be what are the responsibilities with regards to status if the device is unmanaged

oglok:
I believe this change might be confusing to customers and users. Following the K8s analogy, if there is a `DeviceSpec`, there should be a `DeviceStatus`. 

If the goal is to add an aggregate status of applications + system, it would make sense to have something like `device.status.system` and `device.status.applications`. 

That's just my opinion.

hexfusion:
> I believe this change might be confusing to customers and users. Following the K8s analogy, if there is a `DeviceSpec`, there should be a `DeviceStatus`.
> 
> If the goal is to add an aggregate status of applications + system, it would make sense to have something like `device.status.system` and `device.status.applications`.
> 
> That's just my opinion.

Yeah I guess the issue is that the `system` ends up being on top of the device. But life isn't perfect.

hexfusion:
I will mock up the alternative and we can choose which we like better

hexfusion:
```
status:
  conditions:
    additionalProp1:
      type: Approved
      status: 'True'
      observedGeneration: 0
      lastTransitionTime: '2024-06-25T14:07:23.550Z'
      message: string
      reason: string
    additionalProp2:
      type: Approved
      status: 'True'
      observedGeneration: 0
      lastTransitionTime: '2024-06-25T14:07:23.550Z'
      message: string
      reason: string
    additionalProp3:
      type: Approved
      status: 'True'
      observedGeneration: 0
      lastTransitionTime: '2024-06-25T14:07:23.550Z'
      message: string
      reason: string
  systemInfo:
    architecture: string
    bootID: string
    machineID: string
    operatingSystem: string
    measurements:
      additionalProp1: string
      additionalProp2: string
      additionalProp3: string
  containers:
  - name: string
    id: string
    status: string
    image: string
    engine: string
  systemdUnits:
  - name: string
    loadState: string
    activeState: string
  applications:
    data:
      additionalProp1:
        name: string
        ready: string
        restarts: 0
        status: Preparing
      additionalProp2:
        name: string
        ready: string
        restarts: 0
        status: Preparing
      additionalProp3:
        name: string
        ready: string
        restarts: 0
        status: Preparing
    summary:
      status: Healthy
      info: string
  resource:
    cpu: Healthy
    memory: Healthy
    disk: Healthy
  integrity:
    summary:
      status: Passed
      info: string
  config:
    renderedVersion: string
  os:
    image: string
  updated:
    status: UpToDate
    info: string
  summary:
    status: Online
    info: string
  updatedAt: string

```

mangelajo:
Hey, I'm late to the conversation, but I like @oglok 's suggestion. It would maintain the semantics of kubernetes.

oglok:
I like this approach much more. Thanks @hexfusion

internal/agent/device/status/status.go
```diff
@@ -74,7 +75,7 @@ func (m *StatusManager) Get(ctx context.Context) (*v1alpha1.DeviceStatus, error)
 }
 
 func (m *StatusManager) aggregateDeviceStatus(ctx context.Context) (*v1alpha1.DeviceStatus, error) {
-	deviceStatus := v1alpha1.DeviceStatus{Conditions: []v1alpha1.Condition{}}
+	deviceStatus := newDeviceStatus()
```
hexfusion, line 78: this will get fixed in the followup I didn't want to get too into it here.

oglok (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -1994,30 +1994,59 @@ components:
...
+        os:
+          $ref: "#/components/schemas/DeviceOSStatus"
+          description: "Current status of the device OS."
+        updated:
+          $ref: "#/components/schemas/DeviceUpdatedStatus"
+          description: "Current status of the device update."
+        summary:
+          $ref: "#/components/schemas/DeviceSummaryStatus"
+          description: "Summary status of the device."
+        updatedAt:
```
fzdarsky, line 2047: Missing `          format: date-time`

fzdarsky (approved):

PR #299, opened 2024-06-18
NO-ISSUE: add wip testing db benchmark tests

hexfusion:
### Note:  only for testing changes exist that are not planned to be merged

Just for testing purposes to better understand db scalability. very preliminary data but it appears given a set of 100k an returning all 100k (worst case) searching by condition is 28% slower than searching by label.

#### insert 100k devices and return them using 4 label match
```sh
go test -v -benchmem -run=^$ -bench ^BenchmarkLabelSelector$ github.com/flightctl/flightctl/internal/store
goos: linux
goarch: amd64
pkg: github.com/flightctl/flightctl/internal/store

BenchmarkLabelSelector-12              1        4016723456 ns/op        854278064 B/op  12700737 allocs/op
```
#### insert 100k devices and return them using 4 Condition match

```sh
go test -v -benchmem -run=^$ -bench ^BenchmarkConditionSelector$ github.com/flightctl/flightctl/internal/store
goos: linux
goarch: amd64
pkg: github.com/flightctl/flightctl/internal/store

BenchmarkConditionSelector-12                  1        5342411953 ns/op        892666224 B/op  13200759 allocs/op
```


mangelajo:
For labels on the devices we use an index on the DB which may be helping.

https://github.com/flightctl/flightctl/blob/main/internal/store/device.go#L64

We should add indexes to other places we may need. Although that has an impact on insertions, insertions is probably not our most executed query

mangelajo:
Oh I see you were creating a GIN index too for the status.

hexfusion:
this was for testing will put up some more reasonable stuff soonish

PR #301, opened 2024-06-19
MGMT-18103: Modify existing go code to support api and worker servers

ori-amizur:
Note: this PR doesn't contain any changes in Makefile, Dockerfile, and deployment scripts.

The main idea of this PR is to separate the existing server to api, worker, and periodic.  It is based on the flow in TaskManager.  All logic behind the channel listeneres is now handled by the worker server. The api server is the one listening to web interface and executes the API commands.
The periodic server periodically wakes up and executes periodic tasks.

Currently some of the code is duplicated.  Eventually the duplicated old code will
be removed.
As mentioned, the server was split to 3 servers:
- api: It runs the api.  Provides rest api endpoint to the system based on the swagger defined api.
- worker: provides worker(s) handling tasks initiated by any of the components in flightctl.  It is a process that runs thread(s) listening to AMQP queue.
- periodic: runs periodic tasks.  The plan is that it will be invoked by a CronJob.

The cut-off point between api and worker is what is currently implemented in `TaskManager` as channels.  So `TaskManager` was split between the processes.  `TaskManager` will eventually be removed.

Some of the code was partially duplicated. Mainly:
- The initiation of the processes.
- The callbacks in `TaskManager` were copied and modified to a struct which implements only these callbacks.  This struct called `callbackManager`. 
- Currently, each of the threads handling the channels contains handling when a message arrives.  The code was extracted to a different function only handling the message.  Example: 
https://github.com/flightctl/flightctl/blob/58b7cf41fcc565a8996ddf8a5ec2806b0e7b470e/internal/tasks/fleet_rollout.go#L42-L73

And

https://github.com/flightctl/flightctl/blob/58b7cf41fcc565a8996ddf8a5ec2806b0e7b470e/internal/tasks/fleet_rollout.go#L17-L39

Eventually the first will be removed.


hexfusion (commented):
The code seems reasonable on the first pass, I will dig into it more when the tests are up. With regards to the overall design, it makes sense for separation concerns. But I don't see an explanation for why we must have separate processes. From a scalability perspective, there can be some advantages to having processes in separate containers or compute. But for now, it will add a fair amount of complexity for testing/demo.

high-level questions:
- why separate processes vs separate goroutines and what are the tradeoffs?
- why now?

ori-amizur:
> * why separate processes vs separate goroutines and what are the tradeoffs?

In general you can see the reasoning for it here: https://issues.redhat.com/browse/MGMT-17935 .  There two main reasons for it:

- Scalability: handle more load and split it across multiple node (if available).
- Persistence: Using AMQP queues will enable us to avoid messages in flight loss in case that server goes down when there are still messages in flight.

> * why now?

- The reasoning was that now will be easier than later when there will be much more code involved.

@avishayt do you want to add your opinion?

avishayt:
My main motivation was to nail down the system architecture earlier rather than later. @hexfusion, what complexity concerns you?

hexfusion:
> My main motivation was to nail down the system architecture earlier rather than later. @hexfusion, what complexity concerns you?

Just high level I am asking why :), this is a fairly large architectural change without a lot of context. I am probably missing the design doc which would answer these questions. From the top of my head adding separate processes potentially adds.

- health,liveness checks (how do you start this safely)
- graceful termination synchronization concerns (how do you shut this down safely)
- upgrade complexity (version skew of components)
- testing/debug/demo complexity involved with any multi-process system
- additional metrics endpoints

Now this could all be justified it's just hard to review that without more details, differing to @fzdarsky

cmd/flightctl-periodic/main.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		logLvl = logrus.InfoLevel
+	}
+	log.SetLevel(logLvl)
+
+	log.Println("Initializing data store")
+	db, err := store.InitDB(cfg, log)
+	if err != nil {
+		log.Fatalf("initializing data store: %v", err)
+	}
```
mangelajo, line 31:
```suggestion
	log.Println("Initializing data store")
	db, err := store.InitDB(cfg, log)
	if err != nil {
		log.Fatalf("initializing data store: %v", err)
	}
```

Should we call to Init db from multiple places? I am worried about migrations etc, at least no two processes should attempt DB migration/creation. Is there a way to configure it so it will only wait for DB to be ready?

cmd/flightctl-periodic/main.go
```diff
@@ -0,0 +1,44 @@
...
+
+	store := store.NewStore(db, log.WithField("pkg", "store"))
+	defer store.Close()
+
+	if err := store.InitialMigration(); err != nil {
```
mangelajo, line 36:
```suggestion
	if err := store.InitialMigration(); err != nil {
```

See my comments above.

cmd/flightctl-worker/main.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		log.Fatalf("initializing data store: %v", err)
+	}
+
+	store := store.NewStore(db, log.WithField("pkg", "store"))
+	defer store.Close()
+
+	if err := store.InitialMigration(); err != nil {
+		log.Fatalf("running initial migration: %v", err)
+	}
```
mangelajo, line 39: Ditto.

internal/worker_server/server.go
```diff
@@ -0,0 +1,58 @@
+package worker_server
```
hexfusion, line 1: minor: package names should not have underscores

```suggestion
package workerserver
```
ref. https://google.github.io/styleguide/go/decisions.html#package-names

internal/api_server/server.go
```diff
@@ -0,0 +1,115 @@
...
+	router := chi.NewRouter()
+	router.Use(
+		middleware.RequestID,
+		middleware.Logger,
+		middleware.Recoverer,
+		oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapiOpts),
+	)
+
+	h := service.NewServiceHandler(s.store, callbackManager, s.ca, s.log)
+	server.HandlerFromMux(server.NewStrictHandler(h, nil), router)
```
mangelajo, line 91: just for awareness... I am splitting the server in two endpoints over here: https://github.com/flightctl/flightctl/pull/303/files

We will need to eventually merge both PRs

internal/config/config.go
```diff
@@ -40,6 +41,11 @@ type svcConfig struct {
 	LogLevel    string   `json:"logLevel,omitempty"`
 }
 
+type queueConfig struct {
+	AmqpURL       string `json:"amqpUrl,omitempty"`
+	ManagementURL string `json:"managementUrl,omitempty"`
```
mangelajo, line 46: We may need credentials too, right?

hexfusion (commented):
One small nit then code LGTM

mangelajo:
I need to review with more care, just had a diagonal read (Friday afternoon ...) I looks good in general to me. I would like to see this deployed in helm + e2e, containers built etc.

A comment about the CronJobs, the design was made so K8s is an optional comment. How can we make that work without CronJobs?

cmd/flightctl-periodic/main.go
ori-amizur: Init yes, migration no

internal/config/config.go
ori-amizur: Yes

cmd/flightctl-periodic/main.go
ori-amizur: Good point.  Options are:
1. Use kubernetes leader election on one of the components (for example api).  The downside is that it will not work for podman since the solution is bound to kubernetes.
2. Use a database mechanism to lock - might need extra table.

ori-amizur:
> A comment about the CronJobs, the design was made so K8s is an optional comment. How can we make that work without CronJobs?

I will move to regular deployment

internal/worker_server/server.go
ori-amizur: OK

mangelajo:
Based on hexfusion concerns. Could we make this design run inside the flightctl-server for now. We could start them as multiple co-routines (i.e. https://github.com/flightctl/flightctl/pull/303/files#diff-c3691d43252f14347ffb6d535b0a739859e0031a33d3bfefcfe552a7d626e5ccR98) . Then after we have our demo we would be ready to separate in multiple deployments.

ori-amizur:
> Could we make this design run inside the flightctl-server for now

Actually it works for me with multiple processes.  I also wanted to keep the server with old functionality until we decide it is not needed.
Changing the mechanism might introduce other problems that we wouldn't face otherwise.  So my preference is to keep it as is.  We will be able to run both configurations until we decide the the new is ready to be used.

mangelajo:
> > Could we make this design run inside the flightctl-server for now
> 
> Actually it works for me with multiple processes. I also wanted to keep the server with old functionality until we decide it is not needed. Changing the mechanism might introduce other problems that we wouldn't face otherwise. So my preference is to keep it as is. We will be able to run both architectures until we decide the the new is ready to be used.

Ack, I understand what you mean. I am good as long as we create the new helm charts/e2e/bash scripts and remove the old server code to maintain proof in e2e that this continues to work. I am fine if we do that part in a 2nd PR

ori-amizur:
If anyone wants to play with it, here is a branch containing deployment scripts:
https://github.com/ori-amizur/flightctl/tree/testing 
The deployment scripts will be added in the next PR

cmd/flightctl-periodic/main.go
ori-amizur: Added code in **store** package to handle single invocation of auto-migration.

internal/worker_server/server.go
```diff
@@ -0,0 +1,58 @@
...
+		cfg:   cfg,
+		log:   log,
+		store: store,
+	}
+}
+
+func (s *Server) Run() error {
+	s.log.Println("Initializing async jobs")
+	provider := queues.NewAmqpProvider(s.cfg.Queue.AmqpURL, s.log)
+	publisher, err := tasks.FleetQueuePublisher(provider)
```
mangelajo, line 38: So we have a single queue so far, although in that queue we send some other tasks that aren't only fleet related.

Should it be TaskQueue or TaskQueuePublisher instead?

otherwise it is confusing.

mangelajo (changes requested):
Still couldn't finish it at whole. It's a big one. May be I should just try it.

api/v1alpha1/agent/spec.gen.go
```diff
@@ -19,58 +19,59 @@ import (
...
+	"JU3XQF5OjpJRUqk8mSVhanN7ezuhdnsiVTb1Z/X07dnx6bvL0/HLydFkbQpb3xtuckT3vgRB/Acu581b",
+	"2XxxRsaEZvh3mDQlo2QTZJlUwo0LmR+MC1ryZJb81+Ro8sIXM1ZcU1ry6ebF1LVhevoFr7GbBinYMg8i",
+	"4wbsGdHYV1We13HQRcDwoadfBeZb5GYeLsUZS2bJL5gfe+aNzClagLFG+/Guz5pqvBx3bIEW4kzI3o1q",
+	"XdnjrDGa/Ac/HbafU5KuYXiqnyuw9aIna2EveqDDZG9s2CklWgLuvzw68j5tQBjfL+ZettPf/Ze2Db67",
+	"XC4iXWu9nWrjDdrIy6NXkW/IJAmM7EbJq6MXT8aaa+Ii3HwQtDJrm2GYI/rq+Ym+k+ZnWQlP8KfnJ+hG",
+	"C1j75jw87NPMfnDqjfoG1wa8sxmQlpWJzazKnKa916HaHU/i7njhju0NWe5xxnb2PnlKZ7xxwKDNa8m2",
+	"T6YPz+NuP+gjM7tndMM21ZjrvXpCWoMW95oyEl62/iG+fI9TQT2wCu8x1qOkjrqUm5w3Z9y4b8CVjm1v",
+	"1X9SfB6r7tM5yMBfPDcDnaGhlQlzuebHv5f2PFdA2ZZc+Lf+f5jX/WcTWs/P7nNDn+YGa0/UZSelNVYQ",
+	"SWuUxTzxzsRmS1suMlCl4sIMzrifMt09U/Y5yEFCIvpHJYWoYWLX6Z72rVm4Dm6a7G52/x8AAP//Ilx5",
+	"5Qo5AAA=",
```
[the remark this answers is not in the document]
ori-amizur: I don't know why, but I had lint problems that indicated that it needs to be generated again.

internal/periodic_checker/server.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (s *Server) Run() error {
+	provider := queues.NewAmqpProvider(s.cfg.Queue.AmqpURL, s.log)
+	defer provider.Stop()
+	publisher, err := tasks.FleetQueuePublisher(provider)
+	if err != nil {
+		return err
+	}
+	callbackManager := tasks.NewCallbackManager(publisher, s.log)
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/store/postgres_migrator.go
```diff
@@ -0,0 +1,80 @@
...
+		db:      db,
+		migrate: migrate,
+		log:     log,
+	}
+}
+
+func (p *postgresMigrator) performMigrate() error {
+	// This transaction is leaked on purpose in order to keep the shared lock
+	// as long as the process is alive
+	tx := p.db.Begin()
```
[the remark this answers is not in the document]
ori-amizur: we leave the transaction open on purpose to avoid gorm to close it as part of its connection pooling

internal/worker_server/server.go
ori-amizur: I will rename

internal/store/postgres_migrator.go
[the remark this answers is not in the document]
mangelajo: Is there any details about this?

Why closing a transaction would close a connection?

internal/config/config.go
```diff
@@ -72,6 +78,10 @@ func NewDefault() *Config {
 			BaseUrl:   "https://localhost:3333",
 			LogLevel:  "info",
 		},
+		Queue: &queueConfig{
+			AmqpURL:       "http://localhost:5672",
+			ManagementURL: "http://localhost:15672",
```
mangelajo, line 83: Not a change request. I am trying to remember how AMQP worked, I remember the main URL but what was the management URL for?

mangelajo: Is that rabbitmq specific?

internal/tasks/callback_manager.go
```diff
@@ -0,0 +1,218 @@
...
+	FleetSourceUpdated(orgId uuid.UUID, name string)
+	DeviceSourceUpdated(orgId uuid.UUID, name string)
+}
+
+type callbackManager struct {
+	publisher queues.Publisher
+	log       logrus.FieldLogger
+}
+
+func TaskQueuePublisher(provider queues.Provider) (queues.Publisher, error) {
```
mangelajo, line 54: 👍 Thanks :-)

internal/config/config.go
ori-amizur: Maybe.  Actually we don't need now.  I will remove

internal/store/postgres_migrator.go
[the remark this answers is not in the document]
ori-amizur: gorm has connection pooling https://gorm.io/docs/connecting_to_the_database.html#Connection-Pool .
The connection pool may close a connection if it is idle more than the connection lifetime set in the connection pool.  This shouldn't be a problem since every time a query is executed, a connection is taken from the pool or a new one is created.
There is an exception to this.  When using transaction, the same connection must be kept since they are all need to be part of the series of queries that are performed as a whole - meaning that they all need to be performed on the same database connection.
Since the user may want to continue queries on an open transaction, the connection pool cannot close it.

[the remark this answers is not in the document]
mangelajo: But we would be closing the transaction after the .migrate

I still don't understand what you mean

tx: 
   lock()
   migrate
(tx is closed)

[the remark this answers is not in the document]
ori-amizur: The process is the following
```
    tx()
    lock_succeeded = try_lock_exclusive() // Only one should succeed
    lock_shared() // everyone should lock the shared lock
    if not lock_succeeded {
         return
    }
    // If we reached here then the process holds both exclusive lock and shared lock
    migrate()
    unlock_exclusive() // We are left with the shared lock
```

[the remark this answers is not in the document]
ori-amizur: So at the end everyone is left with a shared lock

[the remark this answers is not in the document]
ori-amizur: We decided to remove this part from the PR and possibly do it (or something similar) in another PR

mangelajo (approved):

mangelajo (approved):
It looks reasonable to me, we will need to address in follow ups:

* A global lock to avoid two migrations starting at once
* Deployment and testing
* Removing the monolithic server

Please give a few hours for others to look before merging please

mangelajo:
You need to rebase I believe

mangelajo (approved):

PR #302, opened 2024-06-19
NO-ISSUE: client: make sure to close response body

hexfusion:
This PR ensures response bodies are closed for HTTP client requests. Failure to do so can result in resource leaks and I believe this is a possible root cause of the observed memory leak.

The benchmark was added to validate that the leak was not from status collection. This benchmark can be used in the future to ensure memory allocations and time complexity are kept in check.

### DIFF
--- a/internal/agent/device/status/status_test.go
+++ b/internal/agent/device/status/status_test.go
@@ -1,0 +1,52 @@
+package status
+
+import (
+	"context"
+	"testing"
+
+	"github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/tpm"
+	"github.com/flightctl/flightctl/pkg/executer"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


internal/agent/device/status/status_test.go
```diff
@@ -0,0 +1,52 @@
+package status
```
mangelajo, line 1: I suspect the benchmark sneaked in and is unrelated

mangelajo: But very glad to see benchmarks :)

mangelajo (changes requested):
Just the benchmark, other than that it looks great, thanks for looking at this.

internal/agent/device/status/status_test.go
hexfusion: Yeah it was related in the sense that I used it to validate the leak was not in status collection. We can use the benchmark in the future to ensure allocations and serialization costs are reasonable.

mangelajo (approved):
Ack thanks for the explanation.

PR #303, opened 2024-06-20
OCTOET-476: Split agent facing API endpoint to its own port

mangelajo:
Also ensures that only admin can access to the main endpoint, and
that devices can only update their own status, and pull their own
specification.

### DIFF
--- a/cmd/flightctl-server/main.go
+++ b/cmd/flightctl-server/main.go
@@ -1,12 +1,18 @@
 package main
 
 import (
+	"context"
+	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 
 	"github.com/flightctl/flightctl/internal/client"
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/crypto"
 	"github.com/flightctl/flightctl/internal/server"
+	"github.com/flightctl/flightctl/internal/server/agentserver"
+	"github.com/flightctl/flightctl/internal/server/middleware"
 	"github.com/flightctl/flightctl/internal/store"
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/sirupsen/logrus"
@@ -88,15 +94,34 @@ func main() {
 		log.Fatalf("failed creating TLS config: %v", err)
 	}
 
-	listener, err := server.NewTLSListener(cfg.Service.Address, tlsConfig)
-	if err != nil {
-		log.Fatalf("creating listener: %s", err)
-	}
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT)
+	go func() {
+		listener, err := middleware.NewTLSListener(cfg.Service.Address, tlsConfig)
+		if err != nil {
+			log.Fatalf("creating listener: %s", err)
+		}
 
-	server := server.New(log, cfg, store, ca, listener)
-	if err := server.Run(); err != nil {
-		log.Fatalf("Error running server: %s", err)
-	}
+		server := server.New(log, cfg, store, ca, listener)
+		if err := server.Run(ctx); err != nil {
+			log.Fatalf("Error running server: %s", err)
+		}
+		cancel()
+	}()
+
+	go func() {
+		listener, err := middleware.NewTLSListener(cfg.Service.AgentEndpointAddress, tlsConfig)
+		if err != nil {
+			log.Fatalf("creating listener: %s", err)
+		}
+
+		agentserver := agentserver.New(log, cfg, store, ca, listener)
+		if err := agentserver.Run(ctx); err != nil {
+			log.Fatalf("Error running server: %s", err)
+		}
+		cancel()
+	}()
+
+	<-ctx.Done()
 }
 
 func certFile(name string) string {
--- a/internal/agent/client/client.go
+++ b/internal/agent/client/client.go
@@ -1,0 +1,34 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	client "github.com/flightctl/flightctl/internal/api/client/agent"
+	baseclient "github.com/flightctl/flightctl/internal/client"
+	"github.com/flightctl/flightctl/pkg/reqid"
+	"github.com/go-chi/chi/middleware"
+)
+
+// NewFromConfig returns a new FlightCtl API client from the given config.
+func NewFromConfig(config *baseclient.Config) (*client.ClientWithResponses, error) {
+
+	httpClient, err := baseclient.NewHTTPClientFromConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("NewFromConfig: creating HTTP client %w", err)
+	}
+	ref := client.WithRequestEditorFn(func(ctx context.Context, req *http.Request) error {
+		req.Header.Set(middleware.RequestIDHeader, reqid.GetReqID())
+		return nil
+	})
+	return client.NewClientWithResponses(config.Service.Server, client.WithHTTPClient(httpClient), ref)
+}
+
+type Config = baseclient.Config
+type AuthInfo = baseclient.AuthInfo
+type Service = baseclient.Service
+
+func NewDefault() *Config {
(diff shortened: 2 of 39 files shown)


cmd/devicesimulator/main.go
```diff
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/flightctl/flightctl/internal/agent"
-	"github.com/flightctl/flightctl/internal/client"
+	baseClient "github.com/flightctl/flightctl/internal/client"
```
mangelajo, line 17: ?? this should not be necessary

test/harness/test_harness.go
```diff
@@ -79,15 +79,35 @@ func NewTestHarness(testDirPath string, goRoutineErrorHandler func(error)) (*Tes
...
 		}
+		cancel()
+	}()
+
+	// start agent api server
+	go func() {
+		err := agentServer.Run(ctx)
+		if err != nil {
+			// provide a wrapper to allow require.NoError or ginkgo handling
+			goRoutineErrorHandler(fmt.Errorf("error starting main api server: %w", err))
```
mangelajo, line 108:
```suggestion
			goRoutineErrorHandler(fmt.Errorf("error starting agent-side api server: %w", err))
```

internal/client/enrollment.go
```diff
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/flightctl/flightctl/api/v1alpha1"
-	"github.com/flightctl/flightctl/internal/api/client"
+	client "github.com/flightctl/flightctl/internal/api/client/agent"
```
mangelajo, line 10: oh this one should not be.. this is the base client,

internal/server/server.go
```diff
@@ -58,7 +56,7 @@ func oapiErrorHandler(w http.ResponseWriter, message string, statusCode int) {
 	http.Error(w, fmt.Sprintf("API Error: %s", message), statusCode)
 }
 
-func (s *Server) Run() error {
+func (s *Server) Run(ctx context.Context) error {
```
hexfusion, line 59: :+1:

internal/server/tlscontext/tls.go
```diff
@@ -1,4 +1,4 @@
-package server
+package tlscontext
```
hexfusion, line 1: this feels a little awkward why do we need to change this?

internal/service/agent/handler.go
```diff
@@ -0,0 +1,121 @@
+package service
+
+import (
+	"context"
+	"errors"
+
+	"github.com/flightctl/flightctl/internal/api/server"
+	agentServer "github.com/flightctl/flightctl/internal/api/server/agent"
```
hexfusion, line 8:
```suggestion
	agentserver "github.com/flightctl/flightctl/internal/api/server/agent"
```

internal/service/agent/handler.go
```diff
@@ -0,0 +1,121 @@
...
+	if !ok {
+		log.Warningf("an attempt to access device %q without a CN in tls certificate has been detected", name)
+		return errors.New("no common name in certificate")
+	}
+	if expectedCn, err := crypto.CNFromDeviceName(name); err != nil {
+		return err
+	} else if cn != expectedCn {
+		log.Warningf("an attempt to access device %q with a certificate with CN %q has been detected", name, cn)
+		return errors.New("invalid tls CN for device")
+	}
```
hexfusion, line 36: optional: IMO easier to read
```suggestion
        expectedCn, err := crypto.CNFromDeviceName(name)
	if err != nil {
		return err
	}
	if cn != expectedCn {
		log.Warningf("an attempt to access device %q with a certificate with CN %q has been detected", name, cn)
		return errors.New("invalid tls CN for device")
	}
```

hexfusion (commented):
I assume a lot of this code is copied and pasted on the service end, so I briefly skimmed that. Looking good!

cmd/flightctl-server/main.go
```diff
@@ -88,15 +94,34 @@ func main() {
...
-	if err != nil {
-		log.Fatalf("creating listener: %s", err)
-	}
-
-	server := server.New(log, cfg, store, ca, listener)
-	if err := server.Run(); err != nil {
-		log.Fatalf("Error running server: %s", err)
-	}
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT)
+	go func() {
```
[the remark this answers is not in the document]
mangelajo: no, it's actually blocking on the Run. If you see the original code waited on Run.

But we could change the semantics and make server.Run parallel.

fzdarsky (commented):
Overall this looks good, the only question I have is around the level of duplication of code and the direction of restructuring the service:

IIRC, we wanted to refactor the code to have a single entry point into the service layer (single authorization, input validation, business logic), so no matter how requests come in (from users, task handlers, etc.) they are treated consistently and we minimise the security surface.

At the end of the day, requests from the agent are authenticated (cert vs JWT) and authorized (static vs ReBAC server) differently, the routes are set up differently, but why make the service layer different?

internal/client/enrollment.go
mangelajo: I remember finding references from the cli tools, but I will re-check.

internal/agent/client/client.go
```diff
@@ -0,0 +1,34 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	agentClient "github.com/flightctl/flightctl/internal/api/client/agent"
```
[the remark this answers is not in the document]
mangelajo: Ack, standards are there for something I'll update. Thanks

internal/config/config.go
```diff
@@ -67,10 +69,12 @@ func NewDefault() *Config {
 			Password: "adminpass",
 		},
 		Service: &svcConfig{
-			Address:   ":3333",
-			CertStore: CertificateDir(),
-			BaseUrl:   "https://localhost:3333",
-			LogLevel:  "info",
+			Address:              ":3333",
+			AgentEndpointAddress: ":4444",
```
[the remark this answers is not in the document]
mangelajo: Yeah, I wasn't completely happy with naming here. I think we should eventually refactor that config. 👍

internal/server/tlscontext/tls.go
mangelajo: it was moved from the base server directory to it's own module "tlscontext" so it can be reused from both places, the agent server, and the general API server.

It's just a TLS middleware that injects the CN as context for the HTTP handlers.

mangelajo:
> Overall this looks good, the only question I have is around the level of duplication of code and the direction of restructuring the service:
> 
> IIRC, we wanted to refactor the code to have a single entry point into the service layer (single authorization, input validation, business logic), so no matter how requests come in (from users, task handlers, etc.) they are treated consistently and we minimise the security surface.
> 
> At the end of the day, requests from the agent are authenticated (cert vs JWT) and authorized (static vs ReBAC server) differently, the routes are set up differently, but why make the service layer different?

I am not making the routes differently, what do you mean? :)

I tried to make it very clean on the main API server by just adding a middleware, that middleware can eventually be removed if we don't want/needTLS checking anymore. 
Also avoided duplication of the server logic for the shared functions in a common module.

Let's see it on Monday probably we can check it together in our 1:1 I guess it will be easier for me to understand your concern.

mangelajo:
Thanks for the reviews!, I'll tackle it Monday morning.

internal/server/tlscontext/tls.go
hexfusion: How about naming the package middleware?

cmd/flightctl-server/main.go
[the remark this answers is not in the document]
mangelajo: I am leaving it like this, I started writing it and I noticed that I needed to pass down the cancellation function to each running server. This should go away once we separate this in two independent services.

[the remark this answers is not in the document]
hexfusion: Its fine as long as we need it, as you noted I was looking at the old Run() implementation

fzdarsky (approved):

fzdarsky (approved):

PR #304, opened 2024-06-20
MGMT-17911: implement status updates

hexfusion:
This PR implements the status updates from #298 and introduces a new mechanism for filtering devices by status fields.

`StatusFilter` is type `map[string][]string` where the key is the field path and the value is an array of string values which that are treated as OR operations in the query.

Example Search for DeviceUpdatedStatus.Status with values UpToDate and OutOfDate. The key is the lowercase path of the field we are selecting.  `update.status` maps to `status.update.status` and to search for more than one value(s) requires a comma separator.

resulting SQL
```sql
status -> 'updated' ->> 'status' = ? OR
status -> 'updated' ->> 'status' = ?
```
```
devices?statusFilter=updated.status=UpToDate&statusFilter=updated.status=OutOfDate
```
```go
filterMap := map[string]string{
    "updated.status": {"UpToDate","OutOfDate"},
},
```

```sh
bin/flightctl get device --status-filter=updated.status=UpToDate
```

The key is the field path of DeviceStatus. So for example `updated.status` maps to `status.updated.status`

TODO:

- [x] remove deprecated types
- [x] implement basic status
- [x] add support for statusFilter which adds searching/sorting for status key/values
- [x] ensure proper defaults for DeviceStatus
- [x] add GIN index for status
- [x] add cli support


oglok (approved):

internal/agent/device/osimage_test.go
```diff
@@ -105,7 +104,7 @@ var _ = Describe("Calling osimages Sync", func() {
 			desired := v1alpha1.RenderedDeviceSpec{Os: &v1alpha1.DeviceOSSpec{Image: "mynewimage"}}
 			execMock.EXPECT().ExecuteWithContext(gomock.Any(), container.CmdBootc, "status", "--json").Return(string(hostJson), "", 0)
 			execMock.EXPECT().ExecuteWithContext(gomock.Any(), container.CmdBootc, "switch", "--retain", "mynewimage").Return("", "status error", 1)
-			statusManager.EXPECT().UpdateConditionError(gomock.Any(), device.OsImageDegradedReason, fmt.Errorf("stage image: status error"))
+			// statusManager.EXPECT().Update(gomock.Any(), gomock.Any()).Return(nil, nil)
```
hexfusion, line 107: nit

internal/service/device.go
```diff
@@ -47,19 +47,29 @@ func (h *ServiceHandler) ListDevices(ctx context.Context, request server.ListDev
...
+	if request.Params.FieldSelector != nil {
+		fieldSelector = *request.Params.FieldSelector
+	}
 
 	labelMap, err := labels.ConvertSelectorToLabelsMap(labelSelector)
 	if err != nil {
 		return server.ListDevices400JSONResponse{Message: err.Error()}, nil
 	}
 
+	fieldMap, err := labels.ConvertSelectorToLabelsMap(fieldSelector)
```
rawagner, line 60: this function splits field selector by comma which breaks querying for multiple values.
If I try to do 
```
/devices?fieldSelector=applications.summary.status=Healthy,Unknown
```

we split it into two selectors
`applications.summary.status=Healthy` and `Unknown`.
The second one is invalid

rawagner (changes requested):

internal/service/device.go
hexfusion: Yeah this took a little bit of thought thanks for the report this should be resolved

```
devices?statusFilter=updated.status=UpToDate,Unknown&statusFilter=applications.summary.status=Unknown
```

rawagner: works fine

rawagner (approved):

hexfusion:
These changes unblock the UI, lets start here. Further optimizations/improvements will follow up.

PR #305, opened 2024-06-24
MGMT-18055: Initial Auth handling

rawagner:
k8s auth can be enabled by adding `auth` to `<home>/.flightctl/config.yaml`

```
auth:
  k8sApiUrl: <api_server_url>
```

I added a check on list devices - checking if user has permissions to list `flightctl.io/devices` resource. See also [this slack thread](https://redhat-internal.slack.com/archives/C01KER8QV33/p1718895153458559)


internal/server/server.go
```diff
@@ -75,12 +76,14 @@ func (s *Server) Run() error {
...
 	}
 
+	authMiddleware := auth.CreateAuthMiddleware(s.cfg)
 	router := chi.NewRouter()
 	router.Use(
 		middleware.RequestID,
 		middleware.Logger,
 		middleware.Recoverer,
 		oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapiOpts),
+		authMiddleware.AuthHandler,
```
fzdarsky, line 86: Heads-up for #303 that'll split the middlewares for user- and device-facing APIs.

fzdarsky: Should the Auth middleware be earlier in the chain? E.g. should we even validate before authenticating?

internal/auth/nil_auth.go
```diff
@@ -0,0 +1,19 @@
+package auth
+
+import (
+	"context"
+	"net/http"
+)
+
+type NilAuth struct{}
```
fzdarsky, line 8: I assume when choosing `NilAuth` it means "run without external auth backend", but not "without authentication", or? IOW, we'd still want to generate an admin account (read from a Secret) that a user needs to log in with. Any ideas how you'd plug this into this framework? (Can be a follow-up PR.)

akoserwal: "we'd still want to generate an admin account (read from a Secret) that a user needs to log in"

Doing that will require basic authentication middleware, which can verify the user/password and allow authentication. 
We need to have a separate route for basic auth and a route for token-based authentication

@fzdarsky @mangelajo do you think a separate route is possible?

internal/server/server.go
rawagner: right, we should not validate before auth. I've moved the auth middleware higher.

internal/auth/nil_auth.go
rawagner: I changed this to a basic auth.
The password is generated on runtime and stored in `config.yaml` (user can also override it to define a custom one).
Would this be enough for the OOTB auth ?

rawagner: ...aand back to NilAuth which we would use for the development envs.

deploy/helm/flightctl/templates/flightctl-server-deployment.yaml
```diff
@@ -25,6 +25,8 @@ spec:
           env:
             - name: HOME
               value: "/root"
+            - name: FLIGHTCTL_DISABLE_AUTH
```
rawagner, line 28: i've disabled auth until we have a default (keycloak) auth setup.
Otherwise deployment e2e tests would fail (or any other deployment not using k8s auth)

internal/auth/nil_auth.go
fzdarsky: Hahaha, thank you. Just for the records: We decided not to support basic auth, but deploy Keycloak as default auth server.

internal/auth/auth.go
```diff
@@ -0,0 +1,62 @@
...
+var auth AuthZMiddleware
+
+func GetAuth() AuthZMiddleware {
+	return auth
+}
+
+func CreateAuthMiddleware(cfg *config.Config, log logrus.FieldLogger) (AuthNMiddleware, error) {
+	value, exists := os.LookupEnv(DisableAuthEnvKey)
+	if exists && value != "" {
+		log.Println("Auth disabled")
```
fzdarsky, line 37: Should this be a more prominent warning like https://github.com/flightctl/flightctl/blob/main/internal/agent/config.go#L119?

fzdarsky (approved):

internal/auth/auth.go
rawagner: yes, that would be better. It is now a warning.

internal/auth/authz/k8s.go
```diff
@@ -0,0 +1,66 @@
...
+type K8sAuthZ struct {
+	ApiUrl string
+}
+
+func createSSAR(resource string, verb string) ([]byte, error) {
+	ssar := k8sAuthorizationV1.SelfSubjectAccessReview{
+		Spec: k8sAuthorizationV1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &k8sAuthorizationV1.ResourceAttributes{
+				Verb:     verb,
+				Group:    "flightctl.io",
```
akoserwal, line 22: Can we make this as a configuration? "flightctl.io" is finalized?

rawagner: Just my wild guess.

It seems weird to me to have the group/kind configurable.
@fzdarsky  WDYT ?

fzdarsky: The domain isn't reserved / finalized, no. I'd suggest going with it for now, though.

fzdarsky (approved):

PR #307, opened 2024-06-25
Make Conditions required

fzdarsky:
Our resources have many fields which aren't marked as "required" (likely by omission) but logically should be. This leads to a lot of tests whether fields are not `nil` cluttering the code.

This PR starts with making `status.conditions` required for all resources and all fields of `Conditions` except `observedGeneration` (which we do not use) required.


hexfusion (approved):
thanks!

internal/service/enrollmentrequest.go
```diff
@@ -23,7 +23,7 @@ func validateAndCompleteEnrollmentRequest(enrollmentRequest *v1alpha1.Enrollment
 	if enrollmentRequest.Status == nil {
 		enrollmentRequest.Status = &v1alpha1.EnrollmentRequestStatus{
 			Certificate: nil,
-			Conditions:  &[]v1alpha1.Condition{},
+			Conditions:  []v1alpha1.Condition{},
```
hexfusion, line 26: This is not valid because you required the fields in condition. I believe you need a default set of conditions?

hexfusion: nmd read this wrong

hexfusion (approved):

PR #308, opened 2024-06-26
MGMT-18196: Replace parameters upon fleet rollout

avishayt:
At this point, the only parameters recognized are of the form {{ device.metadata.labels[key] }}, and the service will replace that with the label value corresponding to the specified key for each device.

Parameters in gitConfig and inlineConfig are replaced upon rollout. Errors are currently logged but not reported to users - this will be fixed in a followup.

For gitConfig, parameters in TargetRevision require special handling and will be handled in a followup.

Outline of changes:
1. When validating fleets: do not validate ignition format if inline config contains parameters.
2. When populating templateVersion: remove redundant validations of inline configs.
3. When populating templateVersion: propogate the error when setting the status.
4. When rolling out a templateVersion to a fleet: replace parameters before writing to device.spec

### DIFF
--- a/internal/tasks/config_helpers.go
+++ b/internal/tasks/config_helpers.go
@@ -1,0 +1,65 @@
+package tasks
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+var (
+	paramsRegex *regexp.Regexp = regexp.MustCompile(`(?P<full>{{\w*(?P<param>.*?)\w*}})`)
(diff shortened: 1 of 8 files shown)


fzdarsky:
That's a great start, thank you!

Only comment I'd have is I believe we should switch from search/replace to gotemplates rather sooner than later, because I suspect the change will be harder then.

fzdarsky (approved):

PR #292, opened 2024-06-17
NO-ISSUE: Changes for running flightctl server and agent VM locally

vraiti:
Changed some of the YAML indentation to make helm happy, updated the documentation to reflect that the VM will have a login of user/user, and added CLI option to agent VM to emulate TPM

PR #310, opened 2024-06-26
NO-ISSUE: Update documentation and modify makefiles to make local deployment easier

vraiti:
Along with the updated README, these are changes that were required in order to build the `build`, `deploy`, and `agent-vm` targets on my Fedora 40 machine. The one new file, install_kubectl.sh, tries to roughly mirror install_kind.sh.

This is a redo of #292 since I figured that it would be easier to fix the issues with that PR by just doing the branch from scratch.


README.md
```diff
@@ -7,7 +7,7 @@ Flight Control is a service for declarative, GitOps-driven management of edge de
 ## Building
 
 Prerequisites:
-* `git`, `make`, and `go` (>= 1.20), and `podman-compose`
+* `git`, `make`, `openssl`, `go` (>= 1.20), and `podman-compose`
```
hexfusion, line 10: nit: lets bump go to 1.21 since we are here.

README.md
```diff
@@ -29,6 +29,9 @@ To generate API code and mocks, use `make generate`  This requires installing mo
 
 ## Running
 
+Prerequisites:
+* `virt-install`
```
hexfusion, line 33: This is more of a testing dependency right?

deploy/agent-vm.mk
```diff
@@ -10,14 +10,15 @@ agent-vm: bin/output/qcow2/disk.qcow2
...
+					  --tpm backend.type=emulator,backend.version=2.0,model=tpm-tis \
 					  --vcpus $(VMCPUS) \
 					  --memory $(VMRAM) \
 					  --import --disk $(VMDISK),format=qcow2 \
 					  --os-variant fedora-eln  \
 					  --autoconsole text \
 					  --wait $(VMWAIT) \
 					  --transient || true
-
+	sudo firewall-cmd --zone=libvirt --add-port=7443/tcp
```
hexfusion, line 21: Is firewall-cmd the default for all linux distros?

test/scripts/install_kubectl.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ $arch == "x86_64" ]; then 
+    kube_arch="amd64"
+elif [ $arch == "aarch64" ]; then
+    kube_arch="arm64"
+else
+    echo "Error: no kubectl available for $arch"
+fi
+
+curl -LO "https://dl.k8s.io/release/v1.30.0/bin/linux/$kube_arch/kubectl"
+curl -LO "https://dl.k8s.io/release/v1.30.0/bin/linux/$kube_arch/kubectl.sha256"
```
hexfusion, line 22: nit: can we make the version a var

hexfusion (commented):
Thanks for helping out!

test/scripts/install_kubectl.sh
hexfusion: Should this get added to the kind install script? Maybe check that it exists then install?

README.md
```diff
@@ -58,7 +61,7 @@ this only works on a Linux host.
 ```
 # will create the cluster, and the agent config files in bin/agent which will be embedded in the image
 make deploy
-make agent-vm agent-vm-console # user/password is redhat/redhat
+make agent-vm agent-vm-console # user/password is user/user
```
mangelajo, line 64: Thanks :)

deploy/agent-vm.mk
mangelajo: Also, tweaking firewall rules when the cluster is up will break the network settings and tweaks made by podman on the kernel network stack (because of how firewalld works...). 

I would put any necessary commands to open ports as part of the configuration pre-requisites (documentation). Would that make sense?

test/scripts/install_kubectl.sh
mangelajo: +1 to an env var :)
```suggestion
curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${kube_arch}/kubectl.sha256"
```

test/scripts/install_kubectl.sh
```diff
@@ -0,0 +1,33 @@
...
+arch=$(uname -m)
+if [ $arch == "x86_64" ]; then 
+    kube_arch="amd64"
+elif [ $arch == "aarch64" ]; then
+    kube_arch="arm64"
+else
+    echo "Error: no kubectl available for $arch"
+fi
+
+curl -LO "https://dl.k8s.io/release/v1.30.0/bin/linux/$kube_arch/kubectl"
```
mangelajo, line 21: curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${kube_arch}/kubectl"

test/scripts/install_kubectl.sh
mangelajo: Also, downloading the sha256 and then checking it is not considered very secure (if a bad actor modifies the binary... they can also modify the signature hash). Ideally, you would just store the hashes in this script for the specific version and each architecture... 

Then every time you bumped the version .. you may need to collect all hashes for each arch... a little bit annoying.

README.md
vraiti: Yes

hexfusion: OK so can we move this to the test README as its not actually a dep of the project :pray:

hexfusion: I guess the README steps would require it, my bad.

PR #313, opened 2024-06-27
NO-ISSUE: Provide console access via gRPC streams

mangelajo:
Adds an api request /devices/{name}/console to request a console session with a device, this request returns the gRPC endpoint to be used, as well as a session ID

Anotates the device, and eventually the device will get an update in the form of rendered spec with the same session ID details.


internal/agent/device/device.go
```diff
@@ -150,14 +157,17 @@ func (a *Agent) syncDevice(ctx context.Context) error {
...
 	updateFns := []status.UpdateStatusFn{
-		status.SetOSImage(v1alpha1.DeviceOSStatus{
-			Image: current.Os.Image,
-		}),
 		status.SetConfig(v1alpha1.DeviceConfigStatus{
 			RenderedVersion: current.RenderedVersion,
 		}),
 	}
 
+	if current.Os != nil {
```
mangelajo, line 165: this was crashing sometimes @hexfusion because current.Os was nil

mangelajo:
I shall rebase %)

internal/service/console.go
```diff
@@ -0,0 +1,40 @@
...
+		case flterrors.ErrResourceNotFound:
+			return server.RequestConsole404JSONResponse{}, nil
+		default:
+			return nil, err
+		}
+	}
+
+	sessionId := uuid.New().String()
+
+	annotations := map[string]string{"flightctl.io/console": sessionId}
```
avishayt, line 28: Please declare a const in store/model/device.go
According to the format we've been using, it would be device-controller/console.

internal/service/console.go
```diff
@@ -0,0 +1,40 @@
...
+
+	if err := h.store.Device().UpdateAnnotations(ctx, orgId, request.Name, annotations, []string{}); err != nil {
+		return server.RequestConsole401JSONResponse{Message: "Unable to annotate device for console setup"}, err
```
avishayt, line 31: Why a 401 error?

internal/service/console.go
```diff
@@ -0,0 +1,40 @@
...
+		default:
+			return nil, err
+		}
+	}
+
+	sessionId := uuid.New().String()
+
+	annotations := map[string]string{"flightctl.io/console": sessionId}
+
+	if err := h.store.Device().UpdateAnnotations(ctx, orgId, request.Name, annotations, []string{}); err != nil {
```
avishayt, line 30: How does the user stop the console connection?  We currently don't allow setting/deleting annotations.  We should either add an API to stop the connection or allow modifying annotations.

api/v1alpha1/openapi.yaml
```diff
@@ -892,6 +892,38 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /api/v1/devices/{name}/console:
+    get:
```
avishayt, line 896: Do you think this is a GET, as in "get a console connection", or POST, as in "create a console connection".  The main reason I lean to POST is because we're updating the DB, which we usually don't with GET.  I don't have strong feelings about it though.  However, if we add DELETE to stop the connection, I would definitely move to POST.

api/v1alpha1/openapi.yaml
```diff
@@ -892,6 +892,38 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /api/v1/devices/{name}/console:
```
ori-amizur, line 895: Why is this API endpoint needed? Can't the client issue a gRPC call and get a stream back and start using it without accessing the REST API?

internal/agent/device/console.go
```diff
@@ -0,0 +1,224 @@
...
+					return fmt.Errorf("failed to close console stream: %w", err)
+				}
+			}
+			c.active = false
+		}
+		return nil
+	}
+
+	// do we have a console stream open? we should keep it open
+	if c.active && c.streamClient != nil {
```
ori-amizur, line 56: What if the session id has changed without closing the previous connection? Is such scenario expected to be supported?
May such scenario happen if a new client attempts to create new console connection?

internal/api_server/agentserver/grpc.go
```diff
@@ -0,0 +1,168 @@
...
+			stctx.closed = true
+			s.pendingStreams.Store(sessionId, stctx)
+			return nil
+		} else {
+			actual.(streamCtx).cancel()
+			return err
+		}
+	} else {
+		// the map did not have a value, we are the first client, so we wait for the second
+		s.log.Infof("client %s waiting for peer %s", clientName, sessionId)
```
ori-amizur, line 134: What happens if the other side never connects? Will the first gRPC connection leak?

internal/cli/console.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return fmt.Errorf("creating grpc client: %w", err)
+	}
+	// add key-value pairs of metadata to context
+	ctx = metadata.AppendToOutgoingContext(ctx, agentserver.SessionIDKey, sessionID)
+	ctx = metadata.AppendToOutgoingContext(ctx, agentserver.ClientNameKey, "flightctl-cli")
+
+	stream, err := client.Stream(ctx)
+	if err != nil {
+		return fmt.Errorf("error creating stream: %w", err)
```
ori-amizur, line 125: If there is a failure to connect, shouldn't the server be notified (by REST) to free the session id?

internal/server/agentserver/grpc.go
```diff
@@ -0,0 +1,168 @@
+package agentserver
```
ori-amizur, line 1: Isn't this file a duplicate of internal/api_server/agentserver/grpc.go ?

internal/service/console.go
ori-amizur, line 28: If there is a new request to connect to a console from a new client, and there is an existing console client , will the annotation be overwritten?

internal/service/console.go
mangelajo: hmm, yes, may be an error 500 would be the right thing, this is not about authorization.

internal/service/console.go
mangelajo: Yes, this is a POC now, but I am going to open a RFE in this PR with the next items, this should be one of those.

In the current code, once the device or the client disconnects, the server will prevent new connections through the same session ID. Also, the device does not connect if it's a session ID it connected to and disconnected. But that's just a simple version for now, we need to clear those annotations when a session ends from one side (client disconnected or process on the device exited).

api/v1alpha1/openapi.yaml
mangelajo: May be a post, we are creating something, but it's a bit weird, I don't think the user should actively delete connections, they should naturally go away once the session has been used (client connected and then disconnected, or remote process exited), or after timeout, or we will end with lots of dangling connections.

api/v1alpha1/openapi.yaml
mangelajo: This is an interim solution. In the long term, the plan will be:

1) Make request for console, and we get a link to a web socket URL within the same server
2) The client connects to the websocket URL, 
3) Once the session ends, the session is deleted.

The client should not be connecting to the agent grpc, or another grpc, because that is under the mTLS authentication domain that we want to get rid of eventually. But it's a solution for starting it as a proof of concept now.

internal/agent/device/console.go
mangelajo: Today this will happen:
0) There is an existing console connection (stream A)
1) Client initiates another console connection (stream B)
2) Device gets annotated and gets notified of stream B, drops stream A, connects to B
3) Because stream A is closed, the close is propagated to the client in 0, the client exits
4) Client in 1 was waiting on stream B, gets the console from the connection that happened on 2


So, the first client exits, the new client has a console.


That will drop the device to the new connection. The old connection get's closed client side which is propagated to the client, which closes.

internal/api_server/agentserver/grpc.go
mangelajo: Yes, we will need to add timeouts in here 👍

internal/cli/console.go
mangelajo: I believe we should just rely on timeouts to keep things simple.

internal/server/agentserver/grpc.go
mangelajo: Oh probably, this should have been deleted in rebase, will fix.

internal/service/console.go
mangelajo: Yes. More or less what we were discussing here: https://github.com/flightctl/flightctl/pull/313#discussion_r1671981031

In the final solution we shall account for timeouts, clearing out the annotations.  May be refuse to create a session when an ongoing session is running, and offer a --force method / query flag.

internal/agent/device/console.go
ori-amizur: What will make (2) to happen if in line 56 if the session id is not checked?

api/v1alpha1/openapi.yaml
ori-amizur: Why not websockets all the way?

internal/api_server/agentserver/grpc.go
```diff
@@ -0,0 +1,168 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+
+	sctx := streamCtx{
+		cancel: cancel,
+		stream: stream,
+		closed: false,
+		// TODO: Add expiration fields so we can clean up old streams ver time
+	}
+
+	actual, loaded := s.pendingStreams.LoadOrStore(sessionId, sctx)
```
ori-amizur, line 102: Side note (not for now): This will not always work in case we have more than a single API server since the linkage between client and agent is relying that both connections using the same session id are created in the same process.

api/v1alpha1/openapi.yaml
fzdarsky: Yes, the connection should be deleted when the session ends / times out, but consider the scenario where an admin would like to force-disconnect sessions.

hack/grpcgen.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
+go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2
+
+export PATH=$PATH:$(go env GOPATH)/bin
+protoc --go_out=. --go_opt=paths=source_relative \
```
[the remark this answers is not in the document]
mangelajo: I am not sure, will add to the TO-DO in document.

api/v1alpha1/openapi.yaml
mangelajo: yes, that makes sense too, may be a method like

/ws/devices/{name}/console

which drops you right away on the console?

api/v1alpha1/openapi.yaml
mangelajo: I will leave it as is now, since we want to remove the method altogether in the final solution (see comments in https://github.com/flightctl/flightctl/pull/313/files#r1674031310 )

I will make a note about admin closing console connections.

internal/api_server/agentserver/grpc.go
mangelajo: Yes, this is POC-material. We will need to use exchanges/queues in rabbitmq for a fully distributed system. 

The agent will talk through GRPC to the agent-api,

The API server will talk to the agent-server (which doesn't exist independently yet) via rabbitmq queues/exchanges.

ori-amizur: In general if we use rabbit the user might overload it (i.e if the user issues a command "ls -R /" or similar), which might affect the behavior of the system.

api/v1alpha1/openapi.yaml
ori-amizur: Yes

fzdarsky (approved):

fzdarsky (approved):

fzdarsky (approved):

PR #315, opened 2024-06-28
NO-ISSUE: Add simple validation for different Repo types

oglok:
We are getting an error where the generic repo is being validated, but it does not allow either the sshConfig or httpConfig object.

### DIFF
--- a/api/v1alpha1/validation.go
+++ b/api/v1alpha1/validation.go
@@ -97,12 +97,31 @@ func (r Repository) Validate() []error {
 	allErrs = append(allErrs, validation.ValidateResourceName(r.Metadata.Name)...)
 	allErrs = append(allErrs, validation.ValidateLabels(r.Metadata.Labels)...)
 	allErrs = append(allErrs, validation.ValidateAnnotations(r.Metadata.Annotations)...)
-	repoSpec, err := r.Spec.GetGitGenericRepoSpec()
-	if err != nil {
-		allErrs = append(allErrs, fmt.Errorf("invalid repository type: %s", err))
-	} else {
-		allErrs = append(allErrs, validation.ValidateString(&repoSpec.Repo, "spec.repo", 1, 2048, nil, "")...)
+
+	// Validate GitGenericRepoSpec
+	gitGenericRepoSpec, genericErr := r.Spec.GetGitGenericRepoSpec()
+	if genericErr == nil {
+		allErrs = append(allErrs, validation.ValidateString(&gitGenericRepoSpec.Repo, "spec.repo", 1, 2048, nil, "")...)
 	}
+
+	// Validate GitHttpRepoSpec
+	gitHttpRepoSpec, httpErr := r.Spec.GetGitHttpRepoSpec()
+	if httpErr == nil {
+		allErrs = append(allErrs, validation.ValidateString(&gitHttpRepoSpec.Repo, "spec.repo", 1, 2048, nil, "")...)
+		// TODO: Validate httpRepoSpec
+	}
+
+	// Validate GitSshRepoSpec
+	gitSshRepoSpec, sshErr := r.Spec.GetGitSshRepoSpec()
+	if sshErr == nil {
+		allErrs = append(allErrs, validation.ValidateString(&gitSshRepoSpec.Repo, "spec.repo", 1, 2048, nil, "")...)
+		// TODO: Validate sshRepoSpec
+	}
+
+	if genericErr != nil && httpErr != nil && sshErr != nil {
+		allErrs = append(allErrs, fmt.Errorf("invalid repository type: no valid spec found"))
+	}
+
 	return allErrs
 }
 
--- a/api/v1alpha1/validation.go
+++ b/api/v1alpha1/validation.go
@@ -108,14 +108,14 @@ func (r Repository) Validate() []error {
 	gitHttpRepoSpec, httpErr := r.Spec.GetGitHttpRepoSpec()
 	if httpErr == nil {
 		allErrs = append(allErrs, validation.ValidateString(&gitHttpRepoSpec.Repo, "spec.repo", 1, 2048, nil, "")...)
-		// TODO: Validate httpRepoSpec
+		allErrs = append(allErrs, validateGitHttpConfig(&gitHttpRepoSpec.HttpConfig)...)
 	}
 
 	// Validate GitSshRepoSpec
 	gitSshRepoSpec, sshErr := r.Spec.GetGitSshRepoSpec()
 	if sshErr == nil {
 		allErrs = append(allErrs, validation.ValidateString(&gitSshRepoSpec.Repo, "spec.repo", 1, 2048, nil, "")...)
-		// TODO: Validate sshRepoSpec
+		allErrs = append(allErrs, validateGitSshConfig(&gitSshRepoSpec.SshConfig)...)
 	}
 
 	if genericErr != nil && httpErr != nil && sshErr != nil {
@@ -138,4 +138,53 @@ func (r ResourceSync) Validate() []error {
 
 func (d *DeviceSystemInfo) IsEmpty() bool {
 	return *d == DeviceSystemInfo{}
+}
+
+func validateGitHttpConfig(config *GitHttpConfig) []error {
+	var errs []error
+	if config != nil {
+		if config.CaCrt != nil {
+			if !validation.IsBase64(*config.CaCrt) {
+				errs = append(errs, fmt.Errorf("httpConfig.caCrt must be a valid base64 encoded string"))
+			}
+		}
+
+		if config.TlsCrt != nil {
+			if !validation.IsBase64(*config.TlsCrt) {
+				errs = append(errs, fmt.Errorf("httpConfig.tlsCrt must be a valid base64 encoded string"))
+			}
+		}
+
+		if config.TlsKey != nil {
+			if !validation.IsBase64(*config.TlsKey) {
+				errs = append(errs, fmt.Errorf("httpConfig.tlsKey must be a valid base64 encoded string"))
+			}
+		}
+
+		if config.Username != nil {
+			errs = append(errs, validation.ValidateString(config.Username, "httpConfig.username", 1, 256, nil, "")...)
+		}
+
+		if config.Password != nil {
+			errs = append(errs, validation.ValidateString(config.Password, "httpConfig.password", 1, 256, nil, "")...)
+		}
+	}
+	return errs
+}
+
+func validateGitSshConfig(config *GitSshConfig) []error {
+	var errs []error
+	if config != nil {
+		if config.PrivateKeyPassphrase != nil {
+			errs = append(errs, validation.ValidateString(config.PrivateKeyPassphrase, "sshConfig.privateKeyPassphrase", 1, 256, nil, "")...)
+		}
+
+		if config.SshPrivateKey != nil {
+			if !validation.IsBase64(*config.SshPrivateKey) {
+				errs = append(errs, fmt.Errorf("sshConfig.sshPrivateKey must be a valid base64 encoded string"))
+			}
+		}
+	}
+
+	return errs
 }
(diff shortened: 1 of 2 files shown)


api/v1alpha1/validation.go
```diff
@@ -116,3 +135,52 @@ func (r ResourceSync) Validate() []error {
 	allErrs = append(allErrs, validation.ValidateString(&r.Spec.Path, "spec.path", 0, 2048, nil, "")...)
 	return allErrs
 }
+
+func validateGitHttpConfig(config *GitHttpConfig) []error {
+	var errs []error
+	if config != nil {
+		if config.CaCrt != nil {
```
fzdarsky, line 142: Could we have a generic `ValidateBase64Field(field *string, maxLen int)` in util that in addition to testing for nil and base64 also tests for a max length constraint? That'd make this code more compact.

api/v1alpha1/validation.go
```diff
@@ -116,3 +135,52 @@ func (r ResourceSync) Validate() []error {
...
+		}
+
+		if config.Username != nil {
+			errs = append(errs, validation.ValidateString(config.Username, "httpConfig.username", 1, 256, nil, "")...)
+		}
+
+		if config.Password != nil {
+			errs = append(errs, validation.ValidateString(config.Password, "httpConfig.password", 1, 256, nil, "")...)
+		}
+	}
```
fzdarsky, line 167: Would it make sense to also test for which fields go/don't go together? E.g. user+pass must be given both or nothing, either TLS or user+pass must be provided but not both?

fzdarsky (commented):
Thank you, good catch!

api/v1alpha1/validation.go
rawagner: I think valid options are:

for http config:
 - user + pass
 - tlsCrt + tlsKey
 - user + pass + tlsCrt + tlsKey (yes, you can have combination)
 - other fields are optional

for sshConfig:
 - passphrase cannot be specified without private key

internal/util/validation/validation.go
```diff
@@ -70,6 +71,28 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
 	return asErrors(errs)
 }
 
+func ValidateBase64Field(field *string, maxLenMB int) []error {
```
fzdarsky, line 74: The other validations all measure length in bytes, so having MB here could easily lead to confusion. I'd suggest using bytes here and call the function with `...(config.CaCrt, 20*1024*1024)` above.

internal/util/validation/validation.go
```diff
@@ -70,6 +71,28 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
...
+
+	if field == nil {
+		errs = append(errs, fmt.Errorf("field is nil"))
+	} else {
+		if len(*field) == 0 {
+			errs = append(errs, fmt.Errorf("field is empty"))
+		}
+		maxLen := maxLenMB * 1000000
+		if len(*field) > maxLen {
+			errs = append(errs, fmt.Errorf("field exceeds maximum length of %d", maxLen))
```
fzdarsky, line 85: Using the `field.ErrorList` type as in the other functions (e.g. `errs = append(errs, field.TooLong(fieldPathFor(path), s, maxLen))`) would give more consistent validation output.

internal/util/validation/validation.go
```diff
@@ -70,6 +71,28 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
 	return asErrors(errs)
 }
 
+func ValidateBase64Field(field *string, maxLenMB int) []error {
+	var errs []error // Initialize an empty slice of errors
+
+	if field == nil {
```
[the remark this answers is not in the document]
oglok: but we are calling this function only if it's not nil, right?

internal/util/validation/validation.go
oglok: I was calculating an expected length for a base64 encoded CA or cert/key. It's true that the rest are in bytes, but I thought having a long number there would be too ugly. I could get along with * :-)

internal/util/validation/validation.go
oglok: I didn't use field.ErrorList on purpose because it's from the K8s apimachinery, which might not make sense in this context?

internal/util/validation/validation.go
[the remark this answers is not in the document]
fzdarsky: Sure, the code executes correctly. But the function itself rejects valid values and also behaves differently to all other `ValidateX()` functions, which increases the risks of bugs. And the added bonus would be much more compact code (you could eliminate the majority of lines where you call the function if you first tested all fields unconditionally, the only tested for invalid field combinations).

internal/util/validation/validation.go
fzdarsky: Readability > beauty. You could, of course, define a `const maxBase64CertificateLength = 20*1024*1024` or so, then you'd have both beauty and readability :)

internal/util/validation/validation.go
fzdarsky: The logic is already used by the other validation functions in this package. It provides the helpful information _where_ in the manifest the validation error is and promotes that the same kind of error is reported the same way. And you're not using it _on purpose_?

oglok:
```

[oglok@maxwell flightctl]$ ./bin/flightctl apply -f examples/repository2.yaml
repository: applying examples/repository2.yaml/default-repo-2: 201 Created
[oglok@maxwell flightctl]$ ./bin/flightctl apply -f examples/repository1.yaml
repository: applying examples/repository1.yaml/default-repo-1: 400 Bad Request
{"message":"spec.httpConfig.TlsKey: Invalid value: \"LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tCk1IY0NBUUVFSUl2WlBlakM4U1o2ZDI2dWQ1aU9LOTJNZzMzRHhPOFQ0RDdRdTFGMzZ5TE9vQW9HQ0NxR1NNNDkKQXdFSG9VUURRZ0FFVUFBUCthWTRDR0tpYUxFeGxGcVF0Tnk1RUpEYjRJUGpiQWI1SmZvcHNIQlRodUdNalZseAorK1NxbVduTW9wL3NKRXB0QmhHWG9kTkpjVU11akdKZTFBPT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS\": must be a valid base64 encoded string"}

Error: repository: failed to apply examples/repository1.yaml/default-repo-1: 400 Bad Request
[oglok@maxwell flightctl]$ ./bin/flightctl apply -f examples/repository.yaml
repository: applying examples/repository.yaml/default-repo: 201 Created
[oglok@maxwell flightctl]$ cat examples/repository*
apiVersion: v1alpha1
kind: Repository
metadata:
  name: default-repo-1
spec:
  repo: https://github.com/flightctl/flightctl-demos.git
  httpConfig:
    tls.crt: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJyRENDQVZLZ0F3SUJBZ0lJTENSNjhMZ3JEczR3Q2dZSUtvWkl6ajBFQXdJd0RURUxNQWtHQTFVRUF4TUMKWTJFd0hoY05NalF3TnpBeU1UQXpNVFF5V2hjTk1qVXdOekF5TVRBek1UUXpXakFhTVJnd0ZnWURWUVFERXc5bQpiR2xuYUhSamRHd3RZV1J0YVc0d1dUQVRCZ2NxaGtqT1BRSUJCZ2dxaGtqT1BRTUJCd05DQUFTVDBqSlBBcWpQCndyS1JGcjJ1b1pWcGFqczVCbFgxRkRWcVR3WHJuZlcyalJKa3hJaW15ZDNpL2tDbmhBUWpnZlBsZCtKelljYnkKNVFSZnNUVGZpREdMbzRHT01JR0xNQTRHQTFVZER3RUIvd1FFQXdJRm9EQVRCZ05WSFNVRUREQUtCZ2dyQmdFRgpCUWNEQWpBTUJnTlZIUk1CQWY4RUFqQUFNQ2tHQTFVZERnUWlCQ0EvbUt0a2NTdDV1S2RsUzJmWjNQRzJYWDkzCmZCWnppTEtkdEtNWGZKdXd4REFyQmdOVkhTTUVKREFpZ0NEQXNVb01oWjlGM29iUmcyWVF3cnF2Rmh5UVpDTDUKK1h5QTJhWURuVGk0QXpBS0JnZ3Foa2pPUFFRREFnTklBREJGQWlFQXJMelliVGF5eFhWUlZ0bGxRUXJlV0hYeQpUeGErNHEzK3ZReHlDVCtITTlNQ0lHclp1TkJ1V2xwc2JnSUVYM0pnRm1tMlRvVFBrcE9BeWRvc21yQ1luWUVWCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K
    tls.key: LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tCk1IY0NBUUVFSUl2WlBlakM4U1o2ZDI2dWQ1aU9LOTJNZzMzRHhPOFQ0RDdRdTFGMzZ5TE9vQW9HQ0NxR1NNNDkKQXdFSG9VUURRZ0FFVUFBUCthWTRDR0tpYUxFeGxGcVF0Tnk1RUpEYjRJUGpiQWI1SmZvcHNIQlRodUdNalZseAorK1NxbVduTW9wL3NKRXB0QmhHWG9kTkpjVU11akdKZTFBPT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS
apiVersion: v1alpha1
kind: Repository
metadata:
  name: default-repo-2
spec:
  repo: https://github.com/flightctl/flightctl-demos.git
  sshConfig:
    skipServerVerification: true
    privateKeyPassphrase: lkjlkjlkjlkjkl
    sshPrivateKey: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJyRENDQVZLZ0F3SUJBZ0lJUkRBRHdRWGM2MEV3Q2dZSUtvWkl6ajBFQXdJd0RURUxNQWtHQTFVRUF4TUMKWTJFd0hoY05NalF3TnpBeU1UTXpNRFE0V2hjTk1qVXdOekF5TVRNek1EUTVXakFhTVJnd0ZnWURWUVFERXc5bQpiR2xuYUhSamRHd3RZV1J0YVc0d1dUQVRCZ2NxaGtqT1BRSUJCZ2dxaGtqT1BRTUJCd05DQUFSUUFBLzVwamdJCllxSm9zVEdVV3BDMDNMa1FrTnZnZytOc0J2a2wraW13Y0ZPRzRZeU5XWEg3NUtxWmFjeWluK3drU20wR0VaZWgKMDBseFF5Nk1ZbDdVbzRHT01JR0xNQTRHQTFVZER3RUIvd1FFQXdJRm9EQVRCZ05WSFNVRUREQUtCZ2dyQmdFRgpCUWNEQWpBTUJnTlZIUk1CQWY4RUFqQUFNQ2tHQTFVZERnUWlCQ0NvVkt5dmtlYzJzMHl4Mm1BQVdRWXdueU4wCkVOcmdVT0kzQjhiSEdzZ2RaakFyQmdOVkhTTUVKREFpZ0NCalhlUG5LZVFrTWdDa2ZjTjRvTWFycUwwa21Rd1cKQ3VGRVgvYStkemJ0b0RBS0JnZ3Foa2pPUFFRREFnTklBREJGQWlBcFVOSmdYWVRvaS83dXh4eVI1b0ZmODdCTwo5NmduRi8va01iV3I3T1VMTHdJaEFNMmlFMkdFY2lmSk16Z2w3dHFrRzBFaGhNd080aDN3RXc5N3VwN29tMkluCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K
apiVersion: v1alpha1
kind: Repository
metadata:
  name: default-repo
spec:
  repo: https://github.com/flightctl/flightctl-demos.git
  
  ```
  
It performs validation of the three types now.

celdrake (approved):

PR #316, opened 2024-06-28
MGMT-18104: Modify deployment scripts for switching deployment topology to AMQP queues

ori-amizur:
This PR contains 2 commit:
1. Add deployment scripts for service scale out. The "make deploy" target now deploys api, worker, and rabbitmq. 
2. Code cleanup, the code is cleaned from the remaining server code and TaskManager code.


mangelajo:
Looks good in general, I need to test it. Can you please remove the old go server code duplicated now on the codebase as part of this commit?

deploy/helm/flightctl/templates/flightctl-rabbitmq-statefulset.yaml
```diff
@@ -0,0 +1,46 @@
...
+  replicas: {{ .Values.flightctl.rabbitmq.replicaCount }}
+  selector:
+    matchLabels:
+      flightctl.service: {{ .Values.flightctl.rabbitmq.name }}
+  template:
+    metadata:
+      labels:
+        flightctl.service: {{ .Values.flightctl.rabbitmq.name }}
+    spec:
+      containers:
```
mangelajo, line 20: Thank you for simplifying this to a simple image + deployment files

Containerfile.base
```diff
@@ -0,0 +1,14 @@
...
+COPY ./deploy deploy
+COPY ./hack hack
+COPY ./internal internal
+COPY ./go.* .
+COPY ./pkg pkg
+COPY ./test test
+COPY ./Makefile .
+
+USER 0
+RUN make build
```
mangelajo, line 14: I know this is annoying, and we would replicate this code across all but, can we have all the container files self-contained (build stage + image creation?)

We could modify the "make build" to have single binary targets, which is something that bugged me for long, it makes no sense to build the client in there for example.

The advantage of single container files is that it's easier to audit or build by external people, to build in other downstream systems, quay, etc.

Makefile
```diff
@@ -40,12 +40,31 @@ build: bin
...
+	mkdir -p $${HOME}/go/flightctl-go-cache
+	podman build -f Containerfile.api -v $${HOME}/go/flightctl-go-cache:/opt/app-root/src/go:Z -t flightctl-api:latest
+	touch bin/.flightctl-api-container
+
+bin/.flightctl-worker-container: bin bin/.flightctl-base-container Containerfile.worker go.mod go.sum $(shell find ./ -name "*.go" -not -path "./bin/*" -not -path "./packaging/*")
+	mkdir -p $${HOME}/go/flightctl-go-cache
+	podman build -f Containerfile.worker -v $${HOME}/go/flightctl-go-cache:/opt/app-root/src/go:Z -t flightctl-worker:latest
+	touch bin/.flightctl-worker-container
+
+bin/.flightctl-periodic-container: bin bin/.flightctl-base-container Containerfile.periodic go.mod go.sum $(shell find ./ -name "*.go" -not -path "./bin/*" -not -path "./packaging/*")
```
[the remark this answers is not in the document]
ori-amizur: OK

Containerfile.base
ori-amizur: OK

deploy/helm/flightctl/templates/flightctl-worker-config.yaml
```diff
@@ -0,0 +1,19 @@
+{{ if .Values.flightctl.worker.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: flightctl-config
```
mangelajo, line 5:
```suggestion
  name: flightctl-workers-config
```

I know they live in separate namespaces (the other flightctl-config), but I would want to make sure they work even if we used the same namespace, also make it clear under a list that they have different purposes.

ori-amizur: OK

mangelajo (approved):
It looks good to me

PR #317, opened 2024-06-28
MGMT-17911: agent: add resource manager and monitor implementations

hexfusion:
This PR adds

Resource Manager:  exposes device monitors and ensures proper status updates are sent.

Monitor Interface: Track resource usage and exposes threshold knobs for warnings and alerts.

Implementations:
- CPU
- Disk

The goal is to avoid using cAdvisor, with the intention of eventually removing it as a dependency. All current implementations assume Linux support which feels reasonable for MVP. In the future this support can be extended.

This work is a dependency for rollback support. 

Followups:
- expose config


internal/agent/agent.go
```diff
@@ -101,9 +102,26 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	executer := &executer.CommonExecuter{}
 
+	// TODO: expose through config
```
avishayt, line 105: I would prefer that this go through the fleet/device config (not relevant for this PR of course)

internal/agent/agent.go
```diff
@@ -101,9 +102,26 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	fsAlertThreshold := 10
+	fsWarnThreshold := 20
```
avishayt, line 107: Let's make these more descriptive?  Maybe fsWarnFreeCapacityThreshold?

internal/agent/agent.go
```diff
@@ -101,9 +102,26 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	fsPaths := []string{"/"}
+	fsSyncDuration := time.Minute
+	fsTimeoutDuration := time.Second * 5
```
avishayt, line 110: I think these aren't used?

internal/agent/device/status/resources.go
```diff
@@ -0,0 +1,43 @@
...
+func (r *Resources) Export(ctx context.Context, status *v1alpha1.DeviceStatus) error {
+	usage, err := r.manager.Usage()
+	if err != nil {
+		status.Resources.Disk = v1alpha1.DeviceResourceStatusUnknown
+		return fmt.Errorf("getting resource usage: %w", err)
+	}
+
+	switch {
+	case usage.FsUsage.IsAlert():
+		status.Resources.Disk = v1alpha1.DeviceResourceStatusCritical
```
avishayt, line 32: 1. Do we want to standardize on disk vs file system?
2. Do we have any way to add a message here to say which file system is reaching capacity and how much is left?

internal/agent/device/resource/resource.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (f *Fs) update(usage *FsUsage, append bool) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if append {
+		f.usage.Total += usage.Total
+		f.usage.Free += usage.Free
+		f.usage.Used += usage.Used
```
avishayt, line 173: Why can't we set f.usage.* to 0 in the beginning and then always append?

internal/agent/device/resource/resource.go
```diff
@@ -0,0 +1,202 @@
...
+	ticker := time.NewTicker(f.syncDuration)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if err := f.sync(ctx); err != nil {
+				f.log.Errorf("Error syncing filesystem usage: %v", err)
```
avishayt, line 129: Is there any way to get this information to the user?  For example, if they misconfigured the monitoring?

internal/agent/device/resource/resource.go
hexfusion: I am reusing the object vs reallocate. So each sync we reset the value. I
Will make this more clear.

internal/agent/device/status/resources.go
hexfusion: Data dictionary provided more context in the degraded state. So we can push more data in warn and alert.

hexfusion: Also, it makes sense to add an error state to the enum. Will update.

internal/agent/agent.go
hexfusion: They are passed NewFs to and consumed by NewFs via Manager. Eventually, this should have a config.

internal/agent/device/resource/resource.go
```diff
@@ -0,0 +1,202 @@
...
+func (f *Fs) Usage() *FsUsage {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return &f.usage
+}
+
+func (f *Fs) sync(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, f.syncTimeout)
+	defer cancel()
+
```
hexfusion, line 144: need to hold lock here

internal/agent/device/resource/resource.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (f *Fs) update(usage *FsUsage, append bool) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
hexfusion, line 168: move to sync

internal/agent/device/status/resources.go
```diff
@@ -0,0 +1,52 @@
...
+		manager: manager,
+		log:     log,
+	}
+}
+
+func (r *Resources) Export(ctx context.Context, status *v1alpha1.DeviceStatus) error {
+	resource := r.manager.Usage()
+	diskStatus, err := diskUsageStatus(resource.DiskUsage)
+	if err != nil {
+		// TODO: add error to status
```
hexfusion, line 29: This is MVP and now sets the disk monitor to error status. Will go over options with UI in next design sync.

internal/agent/device/resource/resource.go
hexfusion: wasn't work complexity dropped

avishayt (approved):

PR #318, opened 2024-06-30
Cli improvements

fzdarsky:
This is a start of a series of small improvements to the CLI. This PR contains:

- Add structure for global options.
- Switch reading the client config from the OS's standard config directory (on Linux `$HOME/.config/flightctl/client.yaml`).
- Add a `-c/--context`global option that allows to select between multiple client config files.
- Add new statuses and other fields to the output of `flightctl get`.

### DIFF
--- a/internal/cli/global.go
+++ b/internal/cli/global.go
@@ -17,18 +17,22 @@ const (
 
 type GlobalOptions struct {
 	ConfigFilePath string
+	Context        string
 }
 
 func DefaultGlobalOptions() GlobalOptions {
 	return GlobalOptions{
-		ConfigFilePath: ConfigFilePath(),
+		ConfigFilePath: ConfigFilePath(""),
+		Context:        "",
 	}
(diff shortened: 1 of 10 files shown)


avishayt (approved):

internal/cli/global.go
```diff
@@ -0,0 +1,55 @@
...
+	if len(context) > 0 && context != "default" {
+		return filepath.Join(ConfigDir(), defaultConfigFileName+"_"+context+"."+defaultConfigFileExt)
+	}
+	return filepath.Join(ConfigDir(), defaultConfigFileName+"."+defaultConfigFileExt)
+}
+
+func ConfigDir() string {
+	configRoot, err := os.UserConfigDir()
+	if err != nil {
+		return ""
```
hexfusion, line 52: should we log the error?

hexfusion (approved):

PR #319, opened 2024-06-30

avishayt:
If the user passes in a resourceVersion, compare it with the value currently in the DB to make sure that no updates are overwritten. If the values don't match, return 409 (Conflict).

Main changes:
1. Align the CreateOrUpdate methods of all resources in the store.  They now all start a transaction, fetch the existing record, compare, update, and return the updated object.
2. Return 409 if the resourceVersions don't match.


fzdarsky:
I stead of starting a transaction, could we use optimistic concurrency (as in https://www.2ndquadrant.com/en/blog/postgresql-anti-patterns-read-modify-write-cycles/)?

avishayt:
> I stead of starting a transaction, could we use optimistic concurrency (as in https://www.2ndquadrant.com/en/blog/postgresql-anti-patterns-read-modify-write-cycles/)?

I don't love their approach (adding the version to the WHERE clause) because it doesn't differentiate between the resource not being present and the version being outdated.  My code uses a transaction (which was already there in some cases), but has a clear error message.

mangelajo:
> > I stead of starting a transaction, could we use optimistic concurrency (as in https://www.2ndquadrant.com/en/blog/postgresql-anti-patterns-read-modify-write-cycles/)?
> 
> I don't love their approach (adding the version to the WHERE clause) because it doesn't differentiate between the resource not being present and the version being outdated. My code uses a transaction (which was already there in some cases), but has a clear error message.

I agree with @fzdarsky that this is the way to go; we already have issues with transactions and multiple workers, so we must be really mindful about how we access the database.

Also the use of resourceversion is thought for this...

internal/store/enrollmentrequest.go
```diff
@@ -121,24 +123,73 @@ func (s *EnrollmentRequestStore) CreateOrUpdate(ctx context.Context, orgId uuid.
...
 	created := false
-	findEnrollmentRequest := model.EnrollmentRequest{
-		Resource: model.Resource{OrgID: orgId, Name: *resource.Metadata.Name},
-	}
-	result := s.db.First(&findEnrollmentRequest)
-	if result.Error != nil {
-		if result.Error == gorm.ErrRecordNotFound {
+	var existingRecord *model.EnrollmentRequest
+
+	err := s.db.Transaction(func(innerTx *gorm.DB) (err error) {
```
ori-amizur, line 128: Actually what we are doing here is select, and then insert or update - meaning that we first find if the record exists and if it does not exist then it is created, otherwise it is updated.
Even if it is all done in transaction, if the initial select statement doesn't have the "FOR UPDATE" clause, then the updated record might not be identical to the queried record which is no desirable.
Actually we can avoid doing the whole thing in transaction if we will use upsert (https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/)  which means that the whole operation is done in a single statement.  This is also supported by gorm https://gorm.io/docs/create.html#Upsert-x2F-On-Conflict

PR #323, opened 2024-07-01
adding jwt token auth middleware

akoserwal:


internal/auth/authn/jwtverifer.go
```diff
@@ -0,0 +1,54 @@
...
+type JWTAuth struct {
+	JwksUrl string
+}
+
+func (j JWTAuth) CheckPermission(ctx context.Context, resource string, op string) (bool, error) {
+	//TODO implement me
+	panic("implement me")
+}
+
+func (j JWTAuth) AuthHandler(next http.Handler) http.Handler {
```
rawagner, line 28: Isn't this essentially the same as `K8sAuthN` ? Should it replace the the original implementation ?

akoserwal: I think keep it separate. Retrieving JWKS certs from a kubernetes cluster can be tricky compared to an external SSO provider.

go.sum
```diff
@@ -139,6 +143,18 @@ github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/kr/text v0.2.0 h1:5Nx0Ya0ZqY2ygV366QzturHI13Jq95ApcVaJBhpS+AY=
 github.com/kr/text v0.2.0/go.mod h1:eLer722TekiGuMkidMxC/pM04lWEeraHUUmBw8l2grE=
+github.com/lestrrat-go/blackmagic v1.0.2 h1:Cg2gVSc9h7sz9NOByczrbUvLopQmXrfFx//N+AkAr5k=
+github.com/lestrrat-go/blackmagic v1.0.2/go.mod h1:UrEqBzIR2U6CnzVyUtfM6oZNMt/7O7Vohk2J0OGSAtU=
```
hexfusion, line 147: these deps seem possibly unmaintained what is the justification for using these vs a more standard packages?

akoserwal: - Library: https://github.com/lestrrat-go/jwx is maintained; I can check with regards to dependencies 

I can easily replace it with https://github.com/golang-jwt/jw

akoserwal: List of company/project using "lestrrat-go/jwx": https://github.com/lestrrat-go/jwx/wiki/Users

hexfusion: https://github.com/lestrrat-go/jwx does seem maintained but blackmagic does not.  We are just trying to ensure we use the best options available. I will leave that to you.

rawagner:
@akoserwal heads up - I've changed the AuthN interface in https://github.com/flightctl/flightctl/pull/325

go.sum
akoserwal: I verified the deps are stable even if they aren't updated frequently.

internal/auth/authn/jwtverifer.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+type JWTAuth struct {
+	JwksUrl          string
+	OidcDiscoveryUrl string
+}
+
+func (j JWTAuth) ValidateToken(ctx context.Context, token string) (bool, error) {
```
hexfusion, line 19: question: what value is the bool if its always false on error?

internal/auth/authn/jwtverifer.go
```diff
@@ -0,0 +1,47 @@
...
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (j JWTAuth) GetTokenRequestURL(ctx context.Context) (string, error) {
+	res, err := http.Get(j.OidcDiscoveryUrl)
+	if err != nil {
+		return "", err
```
hexfusion, line 35: would it make sense to add more context on these errors to make it easier to debug?

internal/config/config.go
```diff
@@ -50,6 +50,7 @@ type queueConfig struct {
 
 type authConfig struct {
 	K8sApiUrl string `json:"k8sApiUrl,omitempty"`
+	JwksUrl   string `json:"jwksUrl"`
```
hexfusion, line 53: should this be omitempty as well?

internal/auth/auth.go
```diff
@@ -68,6 +68,9 @@ func CreateAuthMiddleware(cfg *config.Config, log logrus.FieldLogger) (func(http
 		log.Println("k8s auth enabled")
 		authZ = K8sToK8sAuth{K8sAuthZ: authz.K8sAuthZ{ApiUrl: cfg.Auth.K8sApiUrl}}
 		authN = authn.OpenShiftAuthN{OpenShiftApiUrl: cfg.Auth.K8sApiUrl}
+	} else if cfg.Auth != nil && cfg.Auth.JwksUrl != "" {
+		log.Println("odic auth enabled")
+		authN = authn.JWTAuth{JwksUrl: cfg.Auth.JwksUrl}
```
rawagner, line 73: you need to define `authZ` too. Otherwise calls like this https://github.com/flightctl/flightctl/blob/main/internal/service/device.go#L45 will fail.
If we dont have actual permissions checks yet, the authZ can simply return true for everything (same as `NilAuth`)

internal/auth/authn/jwtverifer.go
akoserwal: It must be true only in one case where the JWT token is valid. If there is an error, it should return false, which means authentication failed.

internal/auth/auth.go
akoserwal: Authz should be decoupled from the authn interface. This will make it flexible to use different providers for authn and authz.

* auth: jwt token verifier
* authz: rebac or rbac
or
* authn: openshift 
* authz: opa or rebac

For now, I can add check call to return true

rawagner: I agree, and the interfaces are already decoupled. Or is something missing ?

akoserwal: The interface is decoupled; I should have checked the nil_auth properly after the new updates. That's why I was wondering whether I could use NilAuth as authz and keep authz to jwtauth. 
```
authZ = NilAuth{}
authN = authn.JWTAuth{JwksUrl: cfg.Auth.JwksUrl, OidcDiscoveryUrl: cfg.Auth.OidcDiscoveryUrl}
```

internal/config/config.go
akoserwal: I agree

internal/auth/authn/jwtverifer.go
akoserwal: updated

internal/auth/authn/jwtverifer.go
hexfusion: If there is an error auth fails otherwise it passes, this is not blocking just pointing out that the bool is redundant.

akoserwal: I agree, if you have suggestion to improve it let me know.

akoserwal:
@rawagner @hexfusion @fzdarsky can be merged or do see scope for improvement?

rawagner:
LGTM

somebody from backend needs to approve
@hexfusion @fzdarsky

hexfusion (approved):

PR #325, opened 2024-07-01
MGMT-18056: Support login via CLI

rawagner:
Adds a basic login command for CLI
`flightctl login <API_URL> --token <token>`

A successful login command persists a client.yaml into the ConfigFilePath.
The resulting client.yaml is a combination of `~/.flightctl/client.yaml` + valid token.
Once we get rid of mTLS requirement, we wont need `~/.flightctl/client.yaml`


internal/client/config.go
```diff
@@ -65,7 +76,8 @@ type AuthInfo struct {
 	ClientKey string `json:"client-key,omitempty"`
 	// ClientKeyData contains PEM-encoded data from a client key file for TLS. Overrides ClientKey.
 	// +optional
-	ClientKeyData []byte `json:"client-key-data,omitempty" datapolicy:"security-key"`
+	ClientKeyData []byte   `json:"client-key-data,omitempty" datapolicy:"security-key"`
+	K8sAuth       *K8sAuth `json:"k8sAuth,omitempty"`
```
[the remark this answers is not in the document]
rawagner: ok, so we would add two new endpoints
`/api/v1/token/request` - which would redirect to whatever auth server is configured
`/api/v1/token/validate` - for token validation

does that sound good to you ?

internal/client/config.go
```diff
@@ -232,14 +249,20 @@ func NewFromConfigFile(filename string) (*client.ClientWithResponses, error) {
...
+	config, err := ParseConfigFile(filename)
+	if err != nil {
 		return nil, err
 	}
 	return NewFromConfig(config)
 }
 
 // WriteConfig writes a client config file using the given parameters.
-func WriteConfig(filename string, server string, tlsServerName string, ca *crypto.TLSCertificateConfig, client *crypto.TLSCertificateConfig) error {
+func WriteConfig(filename string, cfg config.Config, tlsServerName string, ca *crypto.TLSCertificateConfig, client *crypto.TLSCertificateConfig) error {
```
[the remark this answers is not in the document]
rawagner: Ok, i've changed this back so that the server config is not required. Now with the other changes in the code, we are passing if auth is disabled or not.

internal/client/config.go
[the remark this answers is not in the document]
rawagner: I've added these new endpoints.

The client now stores only token in the config file. It does not have a specific knowledge of what auth backend is configured for the server. It always just talks to the `/api/v1/token` endpoints

internal/auth/authn/k8s.go
```diff
@@ -2,32 +2,82 @@ package authn
...
 
-		ctx := context.WithValue(r.Context(), K8sTokenKey, authToken[1])
-		next.ServeHTTP(w, r.WithContext(ctx))
+func (k8s K8sAuthN) GetTokenRequestURL(ctx context.Context) (string, error) {
+	res, err := http.Get(fmt.Sprintf("%s/.well-known/oauth-authorization-server", k8s.K8sApiUrl))
+	if err != nil {
+		return "", err
+	}
+	oauthResponse := OauthServerResponse{}
+	bodyBytes, err := io.ReadAll(res.Body)
```
hexfusion, line 75: do we need to close the body after reading?

internal/auth/nil_auth.go
```diff
@@ -2,16 +2,22 @@ package auth
...
-	fn := func(w http.ResponseWriter, r *http.Request) {
-		next.ServeHTTP(w, r)
-	}
-	return http.HandlerFunc(fn)
+func (a NilAuth) AuthHandler(next http.Handler, w http.ResponseWriter, r *http.Request) {
+	next.ServeHTTP(w, r)
+}
+
+func (a NilAuth) ValidateToken(ctx context.Context) (bool, error) {
+	return false, errors.New("auth disabled")
```
hexfusion, line 16: would it be useful to reuse might help for testing as well.

var ErrAuthDisabled = errors.New("auth disabled")

internal/client/config.go
```diff
@@ -260,10 +273,18 @@ func WriteConfig(filename string, server string, tlsServerName string, ca *crypt
...
+}
+
+func (c *Config) Persist(filename string) error {
+	contents, err := yaml.Marshal(c)
 	if err != nil {
 		return fmt.Errorf("encoding config: %v", err)
 	}
+	directory := filename[:strings.LastIndex(filename, "/")]
+	if err := os.MkdirAll(directory, 0700); err != nil {
+		return fmt.Errorf("writing config: %v", err)
```
hexfusion, line 286: minor: please wrap errors

```
		return fmt.Errorf("writing config: %w", err)
```

internal/service/token.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/flightctl/flightctl/internal/api/server"
+	"github.com/flightctl/flightctl/internal/auth"
+)
+
+// (GET /api/v1/token/request)
+func (h *ServiceHandler) TokenRequest(ctx context.Context, request server.TokenRequestRequestObject) (server.TokenRequestResponseObject, error) {
+	authn := auth.GetAuthN()
+	_, ok := authn.(auth.NilAuth)
+	if ok {
+		return server.TokenRequest418Response{}, nil
```
hexfusion, line 15: 418? :) :tea:

internal/client/config.go
rawagner: fixed, i've changed also other errors in this file

internal/auth/nil_auth.go
rawagner: done

internal/auth/authn/k8s.go
rawagner: yes, we do! Fixed

internal/auth/authn/k8s.go
```diff
@@ -2,32 +2,52 @@ package authn
...
+}
 
-type ctxKeyAuthHeader string
+type OauthServerResponse struct {
+	TokenEndpoint string `json:"token_endpoint"`
+}
 
-const K8sTokenKey ctxKeyAuthHeader = "k8s-token"
+func (k8s K8sAuthN) ValidateToken(ctx context.Context, token string) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/apis/user.openshift.io/v1/users/~", k8s.K8sApiUrl), nil)
```
mangelajo, line 20: This would be coupling very much to openshift.

Should we call this authn/openshift.go ? :)

internal/cli/login.go
```diff
@@ -0,0 +1,141 @@
...
+	TokenEndpoint string `json:"token_endpoint"`
+}
+
+func (o *LoginOptions) Run(ctx context.Context, args []string) error {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	configFilePath := filepath.Join(homeDir, ".flightctl", "client.yaml")
+	config, err := client.ParseConfigFile(configFilePath)
```
fzdarsky, line 81:
```suggestion
	config, err := client.ParseConfigFile(o.ConfigFilePath)
```

internal/cli/login.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return err
+	}
+
+	config.Service.Server = args[0]
```
fzdarsky, line 86: Totally fine to leave this for a follow up PR:
When the user needs to log in a second time (because the token expired), the user wouldn't need to specify the server as argument anymore but we could take the server URL from the client.yaml directly, or? In that case, wondering whether the test would belong here or in Validate().

fzdarsky (approved):

fzdarsky (approved):

PR #329, opened 2024-07-02
NO-ISSUE: api: add support for resource monitor configuration

hexfusion:
This PR exposes the configuration for resource monitors. If no monitors are defined they will be disabled. Each resource can define one alert rule per severity which for now is `critical`,  `warning`, and `info` this verification will need to be done by the service.

 ```yaml
resources:
  - monitorType: cpu
    alertRules:
     - severity: Warning
      duration: 10m
      percentage: 50
      description: 'CPU load is above 50% for more than 10 minutes.'
    samplingInterval: 2s
  - monitorType: memory
    alertRules:
     - severity: warning
      duration: 5m
      percentage: 80
      description: 'Memory usage is above 80% for more than 5 minutes'
    samplingInterval: 1s
  - monitorType: disk
    alertRules:
    - severity: warning
      duration: 5h
      percentage: 80
      description: 'Disk usage is above 80% for more than 5 hours.'
    samplingInterval: 2m
    path: /
  - monitorType: battery
    alertRules:
    - severity: Critical
      duration: 1m
      percentage: 5
      description: string
    samplingInterval: 1s
```

### CODE REVIEW

File: api/v1alpha1/openapi.yaml [hunk 1]
```diff
@@ -2529,6 +2632,22 @@ components:
               type: array
               items:
                 type: string
+        resources:
+          type: object
```
fzdarsky, line 2635: Could we make this an array to eventually allow user-defined resource monitors (e.g. battery level)?

api/v1alpha1/openapi.yaml
```diff
@@ -2529,6 +2631,22 @@ components:
               type: array
               items:
                 type: string
+        resources:
```
celdrake, line 2634: Just to confirm, we'd keep the `device.status.resources` with the summary status for each device resource.
And additionally, this new `device.spec.resources` would have the alarm details.

LGTM.

celdrake (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -2529,6 +2631,22 @@ components:
...
+          type: object
+          properties:
+            cpu:
+              $ref: '#/components/schemas/ResourceCPUConfigSpec'
```
fzdarsky, line 2638: The resources and alertRules are essentially the same, right? Could we have one generic ResourceAlertRule schema that applies to all of them (plus user-defined ones)?

api/v1alpha1/openapi.yaml
hexfusion: so like oneOf

 type: array
  description: 'Array of resource monitor configurations.'
  items:
    type: object
    oneOf:
      - $ref: '#/components/schemas/ResourceCPUConfigSpec'
      - $ref: '#/components/schemas/ResourceMemoryConfigSpec'
      - $ref: '#/components/schemas/ResourceDiskConfigSpec'
    discriminator:
      propertyName: type
      mapping:
        cpu: '#/components/schemas/ResourceCPUConfigSpec'
        memory: '#/components/schemas/ResourceMemoryConfigSpec'
        disk: '#/components/schemas/ResourceDiskConfigSpec'

hexfusion: will play with it

api/v1alpha1/openapi.yaml
```diff
@@ -1975,6 +1975,67 @@ components:
...
+          type: string
+        alertRules:
+          type: array
+          items:
+            $ref: '#/components/schemas/ResourceAlertRule'
+          description: "Array of alert rules. Only one alert per severity is allowed." 
+        frequency:
+          type: string
+          pattern: '^\d+[smhd]$'
+          description: "Frequency is the time between disk usage checks. Format: number followed by 's' for seconds, 'm' for minutes, 'h' for hours, 'd' for days."
```
hexfusion, line 1991: nit

api/v1alpha1/openapi.yaml
hexfusion: I guess I don't understand how the user can update the api. You would need something like CRD right?

hexfusion: Any monitor would require agent support so I'm not sure I understand how a generic API would be helpful the type would need to be extended based on support. Maybe you can provide a more concrete example.

hexfusion: per conversation this would be future support

PR #333, opened 2024-07-05
NO-ISSUE: Validate the device template settings of a fleet

celdrake:
Currently, a fleet's device template configuration items are only partially validated.

I completed the validations by adding the OS image and config template validations.
Alternatively, would it be better to validate the entire spec, by invoking the same validation method for a device spec?

Note: The UI does not validate the OS image using the regexp yet, it's tracked separately.

### DIFF
--- a/internal/service/fleet.go
+++ b/internal/service/fleet.go
@@ -41,9 +41,6 @@ func (h *ServiceHandler) CreateFleet(ctx context.Context, request server.CreateF
 	if errs := request.Body.Validate(); len(errs) > 0 {
 		return server.CreateFleet400JSONResponse{Message: errors.Join(errs...).Error()}, nil
 	}
-	if err := validateDiscriminators(request.Body); err != nil {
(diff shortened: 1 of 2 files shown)


internal/service/fleet.go
```diff
@@ -41,9 +41,6 @@ func (h *ServiceHandler) CreateFleet(ctx context.Context, request server.CreateF
 	if errs := request.Body.Validate(); len(errs) > 0 {
 		return server.CreateFleet400JSONResponse{Message: errors.Join(errs...).Error()}, nil
 	}
-	if err := validateDiscriminators(request.Body); err != nil {
```
avishayt, line 44: Should you also remove the other call in this file to the function, and delete the function?

celdrake: Right, `validateDiscriminators` is now unnecessary also for Replace / PUT.

avishayt (approved):
Nice!

PR #339, opened 2024-07-08
NO-ISSUE: Align CLI and UI device fingerprint naming

celdrake:
Currently, the terminology in the CLI and the UI regarding the identifier for a device is different.
The UI refers to this as the "Fingerprint", while the CLI refers to it as the "name".

Since we also define a "display name / name", it seems best that we apply the same terminology everywhere.

### DIFF
--- a/internal/cli/get.go
+++ b/internal/cli/get.go
@@ -300,7 +300,7 @@ func printListResourceResponse(response interface{}, err error, resourceType str
 }
 
 func printDevicesTable(w *tabwriter.Writer, response *apiclient.ListDevicesResponse) {
-	fmt.Fprintln(w, "NAME\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN")
+	fmt.Fprintln(w, "FINGERPRINT\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN")
 	for _, d := range response.JSON200.Items {
 		lastSeen := "<never>"
 		if !d.Status.UpdatedAt.IsZero() {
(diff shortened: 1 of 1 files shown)


internal/cli/get.go
```diff
@@ -318,7 +318,7 @@ func printDevicesTable(w *tabwriter.Writer, response *apiclient.ListDevicesRespo
 }
 
 func printEnrollmentRequestsTable(w *tabwriter.Writer, response *apiclient.ListEnrollmentRequestsResponse) {
-	fmt.Fprintln(w, "NAME\tAPPROVAL\tAPPROVER\tAPPROVED LABELS")
+	fmt.Fprintln(w, "FINGERPRINT\tAPPROVAL\tAPPROVER\tAPPROVED LABELS")
```
hexfusion, line 321: we store this value in the db as `name`  should we align on a naming standard for the product.

celdrake: That's a good point. We should be consistent also at the API / server level.

avishayt: We should align everything.  Do we currently use the "fingerprint" as a name, and later might use something else as a name (e.g., hostname)?  Then maybe we have name and fingerprint, both with the same value currently?  Or do we just have name?  Or just fingerprint?
@fzdarsky

celdrake: FYI, in the UI we have:
- "Fingerprint --> Backend's "name"
- "Name" --> based on the "displayName" label (optional)

fzdarsky: Note that in general, `$resource.metadata.name` is an identifier and immutable. The agent currently sets this to its fingerprint, but that may change (e.g. allowing users to configure the agent to generate its name based on its MAC address or similar instead). So +1 to using "Name" everywhere.

avishayt: I would probably rename "name" to "ID", and then use a label for displayname.

fzdarsky: `$resource.metadata.name` is what we use for K8s-compatibility and tools like kustomize to work on our manifests, so that shouldn't change. Do you mean just in the column header of the CLI/UI?

avishayt: Right, resource.metadata.name is the resource identifier that doesn't change. We can call this ID in the UI, though that might be confusing. Maybe support a displayName label, and we'll have name vs displayName?

celdrake: After further discussion, we agreed that:
- The Backend and CLI will stay as is: the main identifier is called "name", and it currently is populated with the device fingerprint
- The UI columns will change: "Name" will be renamed to "Alias", and  "Fingerprint" will be renamed to "Name".

This allows that everywhere in Flight Control we use the same terminology and are aligned in this matter.

I'll close this MR and create a new one in the UI for the aforementioned changes.

PR #340, opened 2024-07-08
MGMT-18289: Support k8s secrets as a configuration source

ori-amizur:
K8S secrets are defined as config item in either fleet or device.  Such config item contains the following fields:
- name: the secret name
- namespace: the secret namespace
- mountPoint: the location were the secret(s) should be stored. Every secret has a "data" dictionary.  The data itself is stored as key/value pairs.
The aim is to place the values on the filesystem of a target device. So the mountPath is the parent directory, and each of the keys are the file names in the parent directory.
Every K8S secret contains the values encoded with base64.  When placing in the target device this encoding is removed.
To place the secrets on a target device, ignition is used.  Since kubernetes config item is not in ignition format,
"InlineConfigProviderSpec" is used for that purpose.


avishayt:
Really great first pass!  I noticed two things when quickly reviewing:
1. We pass the k8s client throughout the worker.  We may support clients to additional k8s clusters in the future, so let's call this one the "local" k8s client, and maybe also instantiate it only in the relevant tasks when needed?
2. We need to be able to have parameters in the k8s config.  These parameters are replaced in the fleet rollout task.  So, for now, let's not "freeze" the configuration in the templateversion_populate task.  Follow the example of the git config instead.  In templateversion_populate, return an error if there are parameters and simply copy the config to the templateversion as-is.  Then in the device_render task, fetch the contents of the secret and put them in device.spec as inline config (basically move that logic from templateversion_populate to device_render).  Later on we'll implement the logic to properly freeze k8s secrets (we'll need to copy them because there's no versioning) and deal with parameters (we'll need to track secrets per device label value).

ori-amizur:
>  We may support clients to additional k8s clusters in the future

Additional clusters will require additional parameter (i.e kubeconfig) so will the secret spec change to contain kubeconfig or anything else? Anyway if kubeconfig is needed maybe we should hold it as a secret by itself?  If this is the case that we might want to separate the secret reference and what it is used for.

> 2. We need to be able to have parameters in the k8s config. 

What field should be parameterized for secrets ?

avishayt:
> > We may support clients to additional k8s clusters in the future
> 
> Additional clusters will require additional parameter (i.e kubeconfig) so will the secret spec change to contain kubeconfig or anything else? Anyway if kubeconfig is needed maybe we should hold it as a secret by itself? If this is the case that we might want to separate the secret reference and what it is used for.

Let's leave it as is for now and deal with multiple clients if and when we will support it.

> > 2. We need to be able to have parameters in the k8s config.
> 
> What field should be parameterized for secrets ?

Let's check the name and namespace for parameters and return a human-readable error.  Otherwise the code looks great.  We'll need to move things around when we support parameters but the code is correct as is for now.

avishayt (approved):

PR #349, opened 2024-07-10
NO-ISSUE: ensure degraded status on resource alert 

hexfusion:
This PR

- mocks agent http clients for future testing
- ensures that degraded status eventually bubbles up to device.summary for any errors reported during status collection.
- status cleanup/bugfixes

```json
    "resources": {
      "cpu": "Warning",
      "disk": "Healthy",
      "memory": "Healthy"
    },
    "summary": {
      "info": "cpu: alert is firing: Warning",
      "status": "Degraded"
    },
```


avishayt (approved):
The code itself looks good. I think the status can be more user-friendly, but I don't want to block the PR. We should review the wording with UX.
For example, "Warning: CPU usage exceeded 50% for over 10 minutes" instead of "cpu: alert is firing: Warning"

hexfusion:
> The code itself looks good. I think the status can be more user-friendly, but I don't want to block the PR. We should review the wording with UX.
> 
> For example, "Warning: CPU usage exceeded 50% for over 10 minutes" instead of "cpu: alert is firing: Warning"

Yeah there has been some back and forth here. cc @celdrake

hexfusion:
Merging and we can revisit the messages when we add defaults. Using the description directly in status is trivial.

PR #356, opened 2024-07-12
NO-ISSUE: Update readme to include login command.

rawagner:
Remove section about running server locally as it does not work anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,10 @@ make deploy
 Note it stores its generated CA cert, server cert, and client-bootstrap cert in `$HOME/.flightctl/certs`
 and the client configuration in `$HOME/.flightctl/config.yaml`.
 
-Use the `flightctl` CLI to apply, get, or delete resources:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,9 +44,10 @@ make deploy
 Note it stores its generated CA cert, server cert, and client-bootstrap cert in `$HOME/.flightctl/certs`
 and the client configuration in `$HOME/.flightctl/config.yaml`.
 
-Use the `flightctl` CLI to apply, get, or delete resources:
+Use the `flightctl` CLI to login and then apply, get, or delete resources:
 
 ```
+bin/flightctl login <api_server_url>
```
hexfusion, line 50: should we show an example with the port to a simple way to get the server URL

```
 cat ~/.flightctl/client.yaml | grep server | awk '{print $2}'
 ```

rawagner: updated

hexfusion (approved):

PR #363, opened 2024-07-15
NO-ISSUE: Update SSHCommand function in e2e

siserafin:
Updated SSHCommand function with "PubkeyAuthentication=no" to avoid that any local SSH key is used for the authentication.

This will avoid a possible "Too many authentication failures" error when the test runs in local env


mangelajo (approved):

oglok:
rebase plz!

PR #365, opened 2024-07-15
Clean up Keycloak deployment manifests

fzdarsky:
* Removes passwords from values.yaml, replaces with auto-generated passwords.
* Restructure values.yaml and add comments.


deploy/helm/flightctl/charts/keycloak/values.yaml
```diff
@@ -1,37 +1,64 @@
...
-    postgresqlMasterUser: admin
-    postgresqlPassword: CHANGE_ME
-    postgresqlUser: keycloak
-    dbAddr: jdbc:postgresql://keycloak-db/keycloak
-    dbVendor: postgres
+# Copyright Red Hat, Inc. All Rights Reserved.
+# SPDX-License-Identifier: APACHE-2.0
 
+## @param namespace The namespace into which to deploy Keycloak
+namespace: flightctl
```
mangelajo, line 5:
```suggestion
namespace: flightctl-external
```

let's make sure we provide the separation of keeping externally exposed services in another namespace

deploy/helm/flightctl/charts/keycloak/templates/keycloak-admin-secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: keycloak-admin-secret
+  namespace: {{ .Values.namespace }}
+type: Opaque
+data:
+  username: {{ .Values.auth.adminUsername | b64enc | quote }}
+  password: {{ include "keycloak.lookupOrGeneratePassword" (dict "secret" "keycloak-admin-secret" "namespace" .Values.namespace "key" "password" "context" $) }}
```
mangelajo, line 9: ❤️

deploy/helm/flightctl/charts/keycloak/values.yaml
fzdarsky: I was thinking we should change the default to a single namespace for a better first-user experience and to follow the Helm best practice to let the user decide the namespace and whether Helm shall create it in the `helm upgrade --install` command. For the PaaS, we can then use a `values.yaml` override?

PR #366, opened 2024-07-15
EDM-177: ResourceSync overwrites fleet status on sync

avishayt:
1. When the ResourceSync task updates the fleet spec, it should nil out some fields, just like the code in internal/service/fleet.go does
2. For CreateOrUpdate, we should always nil out the status.  This applies to other resources as well.


oglok:
Needs rebase

avishayt:
@oglok Done

oglok (approved):

PR #367, opened 2024-07-15
EDM-178: Fix approve error checking

vraiti:
Currently, `flightctl approve` expects the server to return an HTTP 201 when the API specifies that it returns a HTTP 200. Correct this so that we don't get an error on successful enrollment request approval.


mangelajo (approved):
Thank you!!

PR #369, opened 2024-07-16
EDM-184: CLI should respect output format

avishayt:
Currently the CLI always prints singular items as YAML. Now, the default is tabular format regardless of single/multiple items.  YAML and JSON formats can be passed for both cases as well.


fzdarsky (approved):

fzdarsky:
Nice one, thank you!

PR #370, opened 2024-07-16

ori-amizur:
Compare resource version when performing device updates.  In case of
conflict perform retry when applicable.

### DIFF
--- a/internal/store/device.go
+++ b/internal/store/device.go
@@ -17,6 +17,7 @@ import (
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/google/uuid"
 	"github.com/lib/pq"
+	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
 	"gorm.io/gorm"
 )
@@ -38,6 +39,8 @@ type Device interface {
 	GetRepositoryRefs(ctx context.Context, orgId uuid.UUID, name string) (*api.RepositoryList, error)
 	InitialMigration() error
 }
+
+const retryIterations = 10
 
 type DeviceStore struct {
 	db  *gorm.DB
@@ -107,13 +110,14 @@ func (s *DeviceStore) Create(ctx context.Context, orgId uuid.UUID, resource *api
 	if resource == nil {
 		return nil, flterrors.ErrResourceIsNil
 	}
-	device := model.NewDeviceFromApiResource(resource)
+	device, err := model.NewDeviceFromApiResource(resource)
+	if err != nil {
+		return nil, err
+	}
 	device.OrgID = orgId
-	device.Generation = util.Int64ToPtr(1)
-
-	result := s.db.Create(device)
+	_, err = s.createDevice(device)
 	callback(nil, device)
-	return resource, flterrors.ErrorFromGormError(result.Error)
+	return resource, err
 }
 
 func (s *DeviceStore) List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.DeviceList, error) {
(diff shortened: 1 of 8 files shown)


avishayt:
Please link to [MGMT-18287](https://issues.redhat.com/browse/MGMT-18287) in the commit message and PR

internal/store/model/resource.go
```diff
@@ -32,11 +32,11 @@ type Resource struct {
 
 	Annotations pq.StringArray `gorm:"type:text[]"`
 
-	Generation *int64
-
-	CreatedAt time.Time
-	UpdatedAt time.Time
-	DeletedAt gorm.DeletedAt `gorm:"index"`
+	Generation      *int64
+	ResourceVersion *int64
```
avishayt, line 36: Why a pointer if it always exists?

internal/store/device.go
```diff
@@ -107,9 +110,13 @@ func (s *DeviceStore) Create(ctx context.Context, orgId uuid.UUID, resource *api
 	if resource == nil {
 		return nil, flterrors.ErrResourceIsNil
 	}
-	device := model.NewDeviceFromApiResource(resource)
+	device, err := model.NewDeviceFromApiResource(resource)
+	if err != nil {
+		return nil, err
```
avishayt, line 115: flterrors.ErrIllegalResourceVersionFormat ?

internal/store/device.go
```diff
@@ -107,9 +110,13 @@ func (s *DeviceStore) Create(ctx context.Context, orgId uuid.UUID, resource *api
...
+	}
 	device.OrgID = orgId
 	device.Generation = util.Int64ToPtr(1)
+	device.ResourceVersion = lo.ToPtr[int64](1)
 
 	result := s.db.Create(device)
```
avishayt, line 121: Should you call createDevice()?

internal/store/device.go
```diff
@@ -181,88 +188,105 @@ func (s *DeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string) (*a
...
+	device.ResourceVersion = lo.ToPtr[int64](1)
+	if result := s.db.Create(device); result.Error != nil {
+		err := flterrors.ErrorFromGormError(result.Error)
+		return err == flterrors.ErrDuplicateName, err
+	}
+	return false, nil
+}
+
+func (s *DeviceStore) updateDevice(fromAPI bool, existingRecord, device *model.Device, fieldsToUnset []string) (bool, error) {
+	if device.ResourceVersion != nil && lo.FromPtr(existingRecord.ResourceVersion) != lo.FromPtr(device.ResourceVersion) {
```
avishayt, line 213: Move this check down after the check for changing the spec of an owned device

avishayt (commented):
Overall looks great. Did a quick first pass. Need to do a more in depth review.

internal/store/model/resource.go
ori-amizur: When the existing and the new resourceVersion(s) are compared, if it is nil (when it comes from the API) we ignore the version check.
In addition, in general I prefer to detect cases when it was not set for some reason.

internal/store/device.go
ori-amizur: It is created here https://github.com/flightctl/flightctl/blob/2c89591e575ea3e52de6a46878404b0c940e4cd2/internal/store/model/device.go#L77
This way the function may return other errors (in the future)

internal/store/device.go
ori-amizur: OK

internal/store/device.go
```diff
@@ -181,88 +185,105 @@ func (s *DeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string) (*a
 	return &apiDevice, nil
 }
 
-func (s *DeviceStore) CreateOrUpdate(ctx context.Context, orgId uuid.UUID, resource *api.Device, fieldsToUnset []string, fromAPI bool, callback DeviceStoreCallback) (*api.Device, bool, error) {
+func (s *DeviceStore) getExistingRecord(name string, orgId uuid.UUID) (*model.Device, bool, error) {
```
avishayt, line 188: Nice helper function.  Let's replace the logic in Get() with this?

avishayt (commented):
Looks great, just one comment.  Thank you!

internal/store/device.go
ori-amizur: The semantics is a bit different.  The `getExistingRecord` returns nil error in case the record does not exist.

avishayt (approved):

PR #376, opened 2024-07-17
EDM-202: Publish helm charts to quay.io on merge

mangelajo:


.github/workflows/publish-containers.yaml
```diff
@@ -10,6 +10,32 @@ env:
...
+  publish-helm-charts-containers:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Generate tags
+        id: generate-tags
+        run: |
+          tags=( latest-${GITHUB_SHA} latest )
+          echo "tags=${tags[@]}" >> $GITHUB_OUTPUT
+          echo "tags=${tags[@]}"
```
mangelajo, line 22: we don't need this

fzdarsky (approved):

fzdarsky:
The resulting container ref is a bit clunky, but let’s merge so we can test.

fzdarsky (approved):

PR #377, opened 2024-07-17
EDM-197: Set Condition on dev with multiple owners

avishayt:
Currently we use an annotation, which isn't optimal. Use a Condition instead.


internal/tasks/fleet_rollout.go
```diff
@@ -123,11 +123,8 @@ func (f FleetRolloutsLogic) RolloutDevice(ctx context.Context) error {
 		return nil
 	}
 
-	if device.Metadata.Annotations != nil {
-		multipleOwners, ok := (*device.Metadata.Annotations)[model.DeviceAnnotationMultipleOwners]
-		if ok && len(multipleOwners) > 0 {
-			f.log.Warnf("Device has multiple owners, skipping rollout: %s", multipleOwners)
-		}
+	if api.IsStatusConditionTrue(device.Status.Conditions, api.DeviceMultipleOwners) {
+		f.log.Warnf("Device has multiple owners, skipping rollout")
```
hexfusion, line 127: Wouldn't this be an error or can it be acceptable?

internal/tasks/fleet_selector.go
```diff
@@ -450,19 +456,30 @@ func (f FleetSelectorMatchingLogic) findDeviceOwnerAmongAllFleets(ctx context.Co
...
 
 	err := f.setDeviceOwnerAccordingToMatchingFleets(ctx, device, currentOwnerFleet, matchingFleets)
 	if err != nil {
 		return nil, err
 	}
 
-	if currentAnnotationValue != newAnnotationValue {
-		annotations := map[string]string{
-			model.DeviceAnnotationMultipleOwners: newAnnotationValue,
+	f.log.Infof("HERE %s: %s - %s", *device.Metadata.Name, currentConditionMessage, newConditionMessage)
```
hexfusion, line 471: nit :)

internal/tasks/fleet_selector.go
```diff
@@ -450,19 +456,30 @@ func (f FleetSelectorMatchingLogic) findDeviceOwnerAmongAllFleets(ctx context.Co
...
+	if currentConditionMessage != newConditionMessage {
+		condition := api.Condition{Type: api.DeviceMultipleOwners}
+		if len(matchingFleets) > 1 {
+			condition.Status = api.ConditionStatusTrue
+			condition.Reason = "MultipleOwners"
```
hexfusion, line 476: should the type be a higher level condition just feels extremely specific. For example. 

DeviceVerification
False
MultipleOwners
Msg

celdrake (commented):
Tested in the UI, works well.
https://github.com/flightctl/flightctl-ui/pull/46

internal/tasks/fleet_selector.go
avishayt: I see your point, but I'd like to make it clearer that the condition relates to the resources in the DB and not the actual device.  Also, "Device" in "DeviceVerification" feels redundant.

The device already has a "ValidSpec" condition.  If I generalize both of them, we want to know if the service is offering the latest spec to the agent.  If a device has multiple owners, the fleet logic won't even update the device's spec.  If the device.spec is invalid, the device logic won't offer the spec to the agent.  I can't think of a good name for this, but maybe that should be the scope of the condition.

avishayt: @hexfusion WDYT?

hexfusion: I am fine with it as is just something to think about.

hexfusion (approved):

celdrake (approved):

PR #380, opened 2024-07-18
NO-ISSUE: fix rabbitmq pv path

mparram:
Exclude .erlang.cookie file from PV to fix the startup error:
```
Exception during startup:

error:{badmatch,{error,{{shutdown,{failed_to_start_child,auth,{"Cookie file /var/lib/rabbitmq/.erlang.cookie must be accessible by owner only",
```

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-rabbitmq-statefulset.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-rabbitmq-statefulset.yaml
@@ -35,7 +35,7 @@ spec:
               value: "{{ .Values.rabbitmq.auth.password }}"
           volumeMounts:
             - name: rabbitmq-data
-              mountPath: /var/lib/rabbitmq
+              mountPath: /var/lib/rabbitmq/mnesia
   volumeClaimTemplates:
     - metadata:
         name: rabbitmq-data


ori-amizur:
Isn't the pass we are providing (/var/lib/rabbitmq) an arbitrary path? Why changing the mount path will solve the problem?

mparram:
It's not arbitrary. By adding mnesia to the path, we are extracting the .erlang.cookie from the PV. Therefore, if it restarts, it will generate a new one and won't cause the error I mentioned.
https://github.com/docker-library/rabbitmq/issues/171#issuecomment-316302131

mangelajo:
The cookie is only important if you are going to setup a cluster of rabbitmq's which is not the case. In such chase I believe operators just inject a predefined command cookie via env var. Again, we should not care about that. May be for production deployment we need to figure out if using an operator, same for postgresql

ori-amizur:
The problem is that the permissions on the file.  Please see https://www.rabbitmq.com/docs/clustering#erlang-cookie 
Maybe we just can change the permissions on the file (i.e chmod) ?

mangelajo (approved):
I think we should merge it for now :)

mangelajo (approved):

hexfusion (approved):

hexfusion (approved):

PR #382, opened 2024-07-21

ori-amizur:
Changes for enrollmentrequests and fleets.


avishayt (approved):
Looks great!  I like seeing code moving into common.go!

PR #383, opened 2024-07-21
NO-ISSUE: added deploy-rabbitmq

amir-yogev-gh:
Added the option to deploy local RabbitMQ.
Usage:
make deploy-rabbitmq


deploy/deploy.mk
```diff
@@ -28,6 +28,10 @@ deploy-db:
 	podman exec -it flightctl-db psql -c 'ALTER ROLE admin WITH SUPERUSER'
 	podman exec -it flightctl-db createdb admin || true
 
+deploy-rabbitmq:
+	podman rm -f flightctl-rabbitmq || true
+	cd deploy/podman && podman-compose up -d flightctl-rabbitmq
+
 kill-db:
```
hexfusion, line 35: We also want a way to tear down the message queue. `kill-rabbitmq` WDYT?

hexfusion (commented):
Looks reasonable just one nit, thanks for the PR!

deploy/deploy.mk
```diff
@@ -28,6 +28,10 @@ deploy-db:
 	podman exec -it flightctl-db psql -c 'ALTER ROLE admin WITH SUPERUSER'
 	podman exec -it flightctl-db createdb admin || true
 
+deploy-rabbitmq:
```
avishayt, line 31: Let's call it "deploy-mq".  Rabbitmq is an implementation detail.  Similar to "deploy-db" instead of "deploy-postgres".

avishayt (approved):

deploy/deploy.mk
```diff
@@ -28,6 +28,10 @@ deploy-db:
 	podman exec -it flightctl-db psql -c 'ALTER ROLE admin WITH SUPERUSER'
 	podman exec -it flightctl-db createdb admin || true
 
+deploy-mq:
+	podman rm -f flightctl-mq || true
+	cd deploy/podman && podman-compose up -d flightctl-mq
+
```
avishayt, line 34: Please add kill-mq

PR #389, opened 2024-07-23
NO-ISSUE: agent: remove cadvisor as a dep and add non linux build targets

hexfusion:
This PR removes cadvisor as a dep and gates linux specific agent logic with build targets. Resource collectors are linux opinionated and disabled for all non-linux builds.

### DIFF
--- a/internal/agent/device/status/exporters_linux.go
+++ b/internal/agent/device/status/exporters_linux.go
@@ -1,0 +1,38 @@
+//go:build linux
+// +build linux
+
+package status
+
+import (
+	"os"
+
+	"github.com/flightctl/flightctl/internal/agent/device/resource"
+	"github.com/flightctl/flightctl/pkg/executer"
+	"github.com/flightctl/flightctl/pkg/log"
+)
+
+const (
+	// BootIDPath is the path to the boot ID file.
+	DefaultBootIDPath = "/proc/sys/kernel/random/boot_id"
+)
+
+func newExporters(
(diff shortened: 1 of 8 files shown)


mangelajo:
You need to fix the last commit title, other than that it looks good to me.

internal/agent/device/status/exporters_linux.go
```diff
@@ -0,0 +1,33 @@
+//go:build linux
+// +build linux
```
fzdarsky, line 2: Do we still need the old `// +build` style build constraints?

internal/agent/device/status/exporters_non_linux.go
```diff
@@ -0,0 +1,45 @@
+//go:build darwin || !linux
```
fzdarsky, line 1: Need to rename the file to `exporters_other.go` for the conditional build to work on Darwin.

fzdarsky (changes requested):

internal/agent/device/status/exporters_non_linux.go
hexfusion: wild :) no idea why

internal/agent/device/status/exporters_linux.go
hexfusion: ahh your right old habits I guess go 1.17+ does not

fzdarsky (approved):

PR #391, opened 2024-07-23
MGMT-18170: PreEnrollment Request Endpoint draft

lkatalin:
Drafted some schemas based on https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json but saw it is highly similar to the schemas used by the enrollmentrequests endpoint. Will discuss what if anything needs to be different before making updates. See also the update comment in the referenced issue.


lkatalin:
Closing in favor of #405

PR #392, opened 2024-07-24
MGMT-18491: Add support for config postHooks

hexfusion:
This PR introduces support for device hooks.

### Overview
Configuration hooks provide a mechanism to attach custom actions to detailed filesystem observations. These hooks can monitor various file operations and execute specified actions when those operations occur. This functionality is beneficial for automating tasks that depend on dynamic changes in the filesystem.

### Key Features
- Fine-Grained Monitoring: Hooks can observe specific file operations, such as creation, modification, and deletion.
- Custom Actions: Users can define actions that execute executable files or binaries with parameters when specified events are detected. Actions can be chained.
- Flexible Configuration: Hooks can be configured to monitor specific paths, allowing targeted automation based on the file or directory being observed.

### Supported File Operations
- File Creation
- File Update/Modification
- File Deletion
- Change Owner

### MVP Supported Action Types
More types could be supported directly in the future.

- Systemd
- Executable

### API/Configuration Format
The API for hooks is defined in a structured format, allowing users to specify the `watchPath` to be monitored and the actions to be executed.

### Validation
- WatchPath must be unique multiple actions can be defined per path.

## fleet example
```
apiVersion: v1alpha1
kind: Fleet
metadata:
  name: default
spec:
  selector:
    matchLabels:
      fleet: default
  template:
    metadata:
      labels:
        fleet: default
    spec:
      config:
        postHooks:
          - name: "microshift apply hook"
            description: "A hook to kubectl apply manifest if created in the klusterlet-manifests directory"
            watchPath: "/var/usr/klusterlet-manifests"
            actions:
              - type: "Executable"
                triggerOn:
                  - "Create"
                timeout: "1m"
                executable:
                  command: "/usr/bin/kubectl"
                  args:
                    - "apply"
                    - "-f"
                    - "{{ .FilePath }}"  # This token will be replaced by the actual file observed
                  envVars:
                    - "KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig"
                workDir: "/var/lib/microshift/resources/kubeadmin"
          - name: "systemd example hook"
            description: "A hook to manage systemd units based on update file operations"
            watchPath: "/etc/systemd/system"
            actions:
              - type: "Systemd"
                triggerOn:
                  - "Update"
                timeout: "30s"
                unit:
                  name: ""  # if empty auto-discover from the file name if it is a valid systemd unit
                  operations: 
                    - "DaemonReload"
                    - "Restart"
```        

TODO
- [x] add API's
- [x] update service and fleet
- [x] plumb in new api for service and fleet support
- [x] implement hook api support and test
- [x] implement controller logic
- [x] add examples 
- [x] status


internal/agent/device/action/manager..go
oglok, line 1: manager..go to manager.go?

internal/agent/client/systemd.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+
+	"github.com/flightctl/flightctl/pkg/executer"
+)
+
+const (
+	systemctlCommand = "/usr/bin/systemctl"
+)
+
+func NewSystemd(exec executer.Executer) *Systemd {
```
hexfusion, line 14: todo: use in status

internal/agent/device/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/fsnotify/fsnotify"
+)
+
+type HookActionType string
+
+const (
+	SystemdHookActionType    = "Systemd"
+	ExecutableHookActionType = "Executable"
```
hexfusion, line 14: add docs

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,229 @@
...
+			if !ok {
+				return
+			}
+			m.log.Errorf("error: %v", err)
+		}
+	}
+}
+
+// Update the manager with the new hook if appropriate.
+func (m *hookManager) Update(hook *v1alpha1.DeviceConfigHook) (bool, error) {
```
hexfusion, line 82: add preflight check of all binary paths before we continue

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+func (m *hookManager) handleHookActionExecutable(ctx context.Context, configHook *v1alpha1.ConfigHookActionExecutable) error {
+	dirExists, err := dirExists(configHook.WorkingDirectory)
+	if err == nil {
+		return err
+	}
+
+	// we expect the directory to exist should be created by config if its new.
```
hexfusion, line 127: this isn't true we are going to run hooks after config

internal/agent/device/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+
+	"github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/fsnotify/fsnotify"
+)
+
+type HookActionType string
+
+const (
+	SystemdHookActionType    = "Systemd"
```
hexfusion, line 13: this is optional as we can use executable but I think its worth it as systemd is so common and will be easy to support Ign systemd spec

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+func (m *hookManager) initialize() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	// initialize the handlers map here for testing observability.
+	m.handlers = make(map[string]HookHandler)
+}
+
+func (m *hookManager) ResetDefaults() error {
```
hexfusion, line 313: inject defaults here like resources

internal/agent/device/config/hook_test.go
```diff
@@ -0,0 +1,138 @@
...
+	tmpDir := t.TempDir()
+	cwd, err := os.Getwd()
+	require.NoError(err)
+	execPath := filepath.Join(cwd, "testdata", "executable_script.sh")
+	varDirPath := filepath.Join(tmpDir, "var/lib/stuff")
+	err = os.MkdirAll(varDirPath, 0755)
+	require.NoError(err)
+
+	log := log.NewPrefixLogger("test")
+	log.Infof("tmpDir: %s", tmpDir)
```
hexfusion, line 32: nit

pkg/executer/executer.go
```diff
@@ -30,6 +32,13 @@ func (e *CommonExecuter) ExecuteWithContext(ctx context.Context, command string,
 	return e.execute(cmd)
 }
 
+func (e *CommonExecuter) ExecuteWithContextFromDir(ctx context.Context, workingDir string, command string, args ...string) (stdout string, stderr string, exitCode int) {
+	fmt.Printf("Executing command: %s %v in %s\n", command, args, workingDir)
```
hexfusion, line 36: nit

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,141 @@ components:
...
+        - $ref: '#/components/schemas/ConfigHookActionSpec'
+        - type: object
+          properties:
+            executable:
+              $ref: '#/components/schemas/ConfigHookActionExecutable'
+          discriminator:
+            propertyName: type
+          required:
+            - executable
+    ConfigHookActionExecutable:
```
hexfusion, line 2301: Reuse this for applications so make more generic.

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,141 @@ components:
...
+        - $ref: '#/components/schemas/ConfigHookActionSpec'
+        - type: object
+          properties:
+            unit:
+              $ref: '#/components/schemas/ConfigHookActionSystemdUnit'
+          discriminator:
+            propertyName: type
+          required:
+            - unit
+    ConfigHookActionSystemdUnit:
```
hexfusion, line 2257: resue for applications so make more generic

internal/agent/device/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+// map. This should be more efficient than using go templates for this use-case.
+func replaceTokensInArgs(args []string, tokenMap map[string]string) []string {
+	var result []string
+	for _, arg := range args {
+		trimmedArg := strings.TrimSpace(arg)
+		if strings.HasPrefix(trimmedArg, "{{") && strings.HasSuffix(trimmedArg, "}}") {
+			// remove the {{ and }} from the token and trim any spaces
+			body := strings.TrimSpace(trimmedArg[2 : len(trimmedArg)-2])
+			switch body {
+			case "." + FilePathKey:
```
ori-amizur, line 64: What is the meaning of the tokenMap if only "FilePath" is handled?

internal/agent/device/config/config_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestReplaceTokensInArgs(t *testing.T) {
+	require := require.New(t)
+	tests := []struct {
+		name     string
+		args     []string
+		tokens   map[string]string
+		expected []string
+	}{
+		{
```
ori-amizur, line 17: Can you add tests that tokenMap contains key other than "FilePath" ?

internal/agent/device/config/controller.go
```diff
@@ -30,12 +33,31 @@ func (c *Controller) Sync(desired *v1alpha1.RenderedDeviceSpec) error {
...
+		c.log.Debug("Device resources are nil")
+		// Reset all resource alerts to default
+		if err := c.hookManager.ResetDefaults(); err != nil {
+			return err
+		}
+	} else {
+		// order is important here install new hooks before applying config data
+		// so they can be consumed.
+		hooks := *desired.Config.Hooks
+		if err := c.ensureHooks(&hooks); err != nil {
```
ori-amizur, line 46: Why pass pointer to slice ? you can send the slice itself and then dereference will not be needed.

internal/agent/device/config/controller.go
```diff
@@ -45,6 +67,19 @@ func (c *Controller) Sync(desired *v1alpha1.RenderedDeviceSpec) error {
 	if err != nil {
 		return fmt.Errorf("writing ignition files failed: %w", err)
 	}
+	return nil
+}
 
+func (c *Controller) ensureHooks(hooks *[]v1alpha1.DeviceConfigHookSpec) error {
+	for i := range *hooks {
+		hook := (*hooks)[i]
```
ori-amizur, line 75: What if a hook is deleted? Where is it removed ?

internal/tasks/templateversion_populate.go
```diff
@@ -31,14 +31,14 @@ func templateVersionPopulate(ctx context.Context, resourceRef *ResourceReference
...
-	fleet           *api.Fleet
-	frozenConfig    []api.TemplateVersionStatus_Config_Item
+	callbackManager    CallbackManager
+	log                logrus.FieldLogger
+	store              store.Store
+	k8sClient          k8sclient.K8SClient
+	resourceRef        ResourceReference
+	templateVersion    *api.TemplateVersion
+	fleet              *api.Fleet
+	frozenConfigSource []api.DeviceConfigSourceSpec
```
ori-amizur, line 41: Don't we want to freeze the hooks too ?

api/v1alpha1/openapi.yaml
ori-amizur: Maybe we can add the changed file as an argument too?

internal/agent/device/config/config.go
hexfusion: the idea is that more tokens will be added shortly

internal/agent/device/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+			result = append(result, arg)
+		}
+	}
+
+	return result
+}
+
+func newTokenMap(filePath string) map[string]string {
+	return map[string]string{
+		"FilePath": filePath,
```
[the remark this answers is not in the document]
hexfusion: The core idea with the map is that the agent will have multiple tokens but your right if we only have one a map is not useful.

internal/agent/device/config/controller.go
hexfusion: good catch this is not supported yet

internal/agent/device/config/config.go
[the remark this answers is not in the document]
hexfusion: But yeah we can use the constant in the map

internal/agent/device/config/config_test.go
hexfusion: I can be they are not supported directly. This does need more thought

internal/agent/device/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+		if strings.HasPrefix(trimmedArg, "{{") && strings.HasSuffix(trimmedArg, "}}") {
+			// remove the {{ and }} from the token and trim any spaces
+			body := strings.TrimSpace(trimmedArg[2 : len(trimmedArg)-2])
+			switch body {
+			case "." + FilePathKey:
+				if tokenData, ok := tokenMap[FilePathKey]; ok {
+					result = append(result, tokenData)
+				}
+			default:
+				result = append(result, arg)
```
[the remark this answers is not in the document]
hexfusion: unless it matches the expected token we can not conclude intent so I believe the answer is yes.

internal/agent/device/config/config.go
```diff
@@ -0,0 +1,83 @@
...
+func replaceTokensInArgs(args []string, tokenMap map[string]string) []string {
+	var result []string
+	for _, arg := range args {
+		trimmedArg := strings.TrimSpace(arg)
+		if strings.HasPrefix(trimmedArg, "{{") && strings.HasSuffix(trimmedArg, "}}") {
+			// remove the {{ and }} from the token and trim any spaces
+			body := strings.TrimSpace(trimmedArg[2 : len(trimmedArg)-2])
+			switch body {
+			case "." + FilePathKey:
+				if tokenData, ok := tokenMap[FilePathKey]; ok {
```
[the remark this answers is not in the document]
hexfusion: Correct, if we don't support the token directly we do not attempt to conclude intent. If we wanted to error we should do that at a higher level.

internal/tasks/templateversion_populate.go
hexfusion: Yeah I am not sure I did think about it a bit, what are the criteria for freezing?

internal/agent/device/config/config.go
ori-amizur: Why not create generic code to handle all?

internal/tasks/templateversion_populate.go
ori-amizur: According to my understanding if some configuration is applied to a fleet - and after a month a device joins the fleet, we would like to the same configuration that was applied previously to other devices in the fleet should also be applied to the new device.

internal/agent/device/config/config.go
hexfusion: yeah map will drive this

internal/agent/device/config/controller.go
hexfusion: yeah I am not sure why honestly I was dealing with some loops where this was required in another area. I blame old age. :) fixed.

internal/tasks/templateversion_populate.go
hexfusion: good conversation for monday thanks for bringing it up

internal/agent/device/config/config.go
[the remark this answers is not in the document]
hexfusion: I did make this an error for now after thinking about it more.

api/v1alpha1/openapi.yaml
hexfusion: Yeah I have been playing with how this will look my current approach is in the example.

internal/agent/device/config/config.go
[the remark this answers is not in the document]
hexfusion: changed this to error

internal/agent/device/config/config.go
hexfusion: > Why not create generic code to handle all?
mostly lazy haha

internal/agent/device/config/config_test.go
hexfusion: passes now

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,375 @@
...
+	if exitCode != 0 {
+		return fmt.Errorf("failed to execute command: %s %d: %s", configHook.Executable.Command, exitCode, stderr)
+	}
+
+	return nil
+}
+
+func (m *hookManager) addOrReplaceHookHandler(hook *v1alpha1.DeviceConfigHookSpec) error {
+	// build lookup for file operations
+	opActions := make(map[fsnotify.Op][]v1alpha1.ConfigHookAction)
```
hexfusion, line 260: TODO: all logic to delete

internal/agent/device/config/controller.go
ori-amizur: It seems to me if a hook which is based on a path does not appear anymore in the spec - it will continue to be fully functional.   Unless I am missing something.

hexfusion: added basic tests and ensured watches are torn down.

internal/agent/client/systemd.go
hexfusion: followup

internal/agent/device/config/config.go
hexfusion: followup

internal/tasks/templateversion_populate.go
hexfusion: going to need t be a followup

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,141 @@ components:
...
+    FileOperation:
+      type: string
+      description: 'The type of operation that was observed on the file.'
+      enum:
+        - "Create"
+        - "Remove"
+        - "Update"
+        - "Delete"
+        - "Rename"
+        - "ChangePermissions"
```
avishayt, line 2212: Similar to the above comment - will the agent change permissions or delete+create?

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,141 @@ components:
...
+        - watchPath
+    FileOperation:
+      type: string
+      description: 'The type of operation that was observed on the file.'
+      enum:
+        - "Create"
+        - "Remove"
+        - "Update"
+        - "Delete"
+        - "Rename"
```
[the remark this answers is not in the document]
hexfusion: The watch is generally a directory for performance reasons. So in that context a rename is possible. But I agree we should will be using create/delete model. Will drop and we can bring back if we have demand/ use-case.

api/v1alpha1/openapi.yaml
hexfusion: Makes sense

hexfusion: fixed

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,141 @@ components:
...
+        actions:
+          type: array
+          items:
+            $ref: '#/components/schemas/ConfigHookAction'
+        watchPath:
+          type: string
+          description: 'The path to monitor for changes in configuration files. This path can point to either a specific file or an entire directory.'
+      required:
+        - name
+        - description
```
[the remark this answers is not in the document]
hexfusion: done

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,433 @@
...
+
+	// push errors to status
+	handle.SetError(errors.Join(errs...))
+}
+
+func (m *hookManager) getHandle(eventName string) *HookHandler {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	// check if the event name is a file or directory
+	paths := []string{eventName, filepath.Dir(eventName)}
```
ori-amizur, line 196: What if we have watchers both on the event and the directory.  Is the directory hidden for this event?

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,433 @@
...
+		default:
+			errs = append(errs, fmt.Errorf("unknown hook action type: %s", hookActionType))
+		}
+	}
+
+	// push errors to status
+	handle.SetError(errors.Join(errs...))
+}
+
+func (m *hookManager) getHandle(eventName string) *HookHandler {
```
ori-amizur, line 192: getHandler ?

internal/agent/device/config/hook.go
hexfusion: yeah, we will have 2 separate events I will clean this up and add tests

hexfusion: added

internal/agent/device/config/hook.go
hexfusion: actually this should be on handle because the binaries could be added by config.

internal/agent/device/config/hook.go
hexfusion: cc @ori-amizur

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,136 @@ components:
...
+            propertyName: type
+          required:
+            - executable
+    ConfigHookActionExecutable:
+          type: object
+          properties:
+            command:
+              type: string
+              description: 'The path or name of the executable file to run. This can be the name of a binary located in $PATH, or a full path to the binary.'
+            args:
```
oglok, line 2302: can args be used to add an environment variable to the executable command? In order to apply a set of manifests on MicroShift using the OC client, we should be able to point at the kubeconfig file. Something like this:

`KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig oc apply -f /var/usr/acm-manifests/`

oglok:
The overall design looks fine. From the config hooks proposal document, I understood we were going to implement opinionated "actions" for systemd, firewalld, kubectl, etc. Having a generic executable command could bring some configuration drift we can't control, however, I think for a starting MVP, this design would be very useful without too much extra work.

Thanks for the PR man!

examples/fleet.yaml
```diff
@@ -15,25 +14,40 @@ spec:
...
-              version: 3.4.0
-            storage:
-              files:
-                - contents:
-                    source: >-
-                      data:,This%20system%20is%20managed%20by%20flightctl.%0A
-                  mode: 422
-                  overwrite: true
-                  path: "/etc/motd"
+        source:
```
oglok, line 17: why changing the name from config to source?

hexfusion: It is technically config.source as config now is being extended. I am open to other names but since it now refers to the source of the config via provider on the fleet level I went with it.

api/v1alpha1/openapi.yaml
hexfusion: well we can pass env directly so I will add that good point

hexfusion:
> Having a generic executable command could bring some configuration drift we can't control

I don't disagree but my understanding was this was a requirement. At a minimum, it's a place to start.

examples/fleet.yaml
hexfusion: I believe this feature is config hooks because the hook is into the application of the configuration files to disk.

internal/agent/device/config/hook.go
```diff
@@ -0,0 +1,474 @@
...
+func addOrReplaceHookHandler(fsMonitor FileMonitor, newHook *v1alpha1.DeviceConfigHookSpec, existingHandlers map[string]*HookHandler) error {
+	// build lookup for file operations
+	opActions := make(map[fsnotify.Op][]v1alpha1.ConfigHookAction)
+	for _, action := range newHook.Actions {
+		hookActionType, err := action.Discriminator()
+		if err != nil {
+			return err
+		}
+		switch hookActionType {
+		case SystemdHookActionType:
```
ori-amizur, line 333: The code for `SystemdHookActionType` and  `ExecutableHookActionType` looks exactly the same.  Can they be unified to a single `case` section ?

api/v1alpha1/openapi.yaml
```diff
@@ -2181,6 +2181,141 @@ components:
...
+      type: string
+      description: 'The type of operation that was observed on the file.'
+      enum:
+        - "Create"
+        - "Remove"
+        - "Update"
+        - "Delete"
+      x-enum-varnames:
+        - "FileOperationCreate"
+        - "FileOperationRemove"
```
ori-amizur, line 2212: What is the difference between FileOperationRemove, and FileOperationDelete?

hexfusion: its a dupe and will be removed I am leaning on keeping `Remove`

internal/agent/device/hook/hook.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (h *Handler) SetError(err error) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.err = err
+}
+
+func fsnotifyOpToFileOperation(op fsnotify.Op) (v1alpha1.FileOperation, error) {
+	switch op {
```
ori-amizur, line 118: Please note the fsnotify.Op may be a `bitwise or`  bitmask value: https://pkg.go.dev/github.com/fsnotify/fsnotify#Event

hexfusion: updated eval

internal/agent/device/hook/hook.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (h *Handler) SetError(err error) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.err = err
+}
+
+func fsnotifyOpToFileOperation(event fsnotify.Event) (v1alpha1.FileOperation, error) {
+	switch {
+	case event.Has(fsnotify.Create):
```
ori-amizur, line 119: What if the event contains create and write? Do we ignore the second one?

hexfusion:
This PR was an exploration into the implementation of the API and will probably change going to close this for now until the API is merged.

PR #402, opened 2024-07-25
Testing CI with CSR endpoint WIP

lkatalin:


api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
...
+        status:
+          description: 'True, False, or Unknown'
+          $ref: '#/components/schemas/ConditionStatus'
+        type:
+          description: 'Approved, Denied, and/or Failed'
+          type: string
+          enum:
+            - 'Approved'
+            - 'Denied'
+            - 'Failed'
```
hexfusion, line 3059: question: what state is the request in if it has not been yet been approved?

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
...
+        lastUpdateTime:
+          description: 'Time of the last update to current condition'
+          type: string
+          format: date-time
+        message:
+          description: 'Human readable message with details about the request state'
+          type: string
+        reason:
+          description: 'Brief reason for the request state'
+          type: string #TODO - enum?
```
hexfusion, line 3049: The advantages of a string are that you do not need to extend the API to add/remove options.

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
...
+          description: 'Requested duration of validity for the certificate'
+        extra:
+          additionalProperties:
+            items:
+              type: string
+            type: array
+          description: 'Extra attributes of the user that created the CSR, populated by the API server on creation and immutable'
+          type: object
+        groups:
+          description: 'Group membership of the user that created the CSR, populated by the API server on creation and immutable'
```
hexfusion, line 3120: question: would membership be able to change what is the reason behind storing this here?

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
...
+        signerName:
+          description: 'Indicates the requested signer, and is a qualified name'
+          type: string
+        uid:
+          description: 'UID of the user that created the CSR, populated by the API server on creation and immutable'
+          type: string
+        usage:
+          description: 'Usages (purpose or scope) requested in the issued certificate - this differs from k8s in that we restrict it to several known usages'
+          type: string
+          $ref: '#/components/schemas/CertSigningRequestUsages'
```
hexfusion, line 3137: question: do you foresee ever having more than one usage per request?

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
...
+        username:
+          description: 'Name of the user that created the CSR, populated by the API server on creation and immutable'
```
hexfusion, line 3139: question: if you have the uid do we also need the username and vice versa

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
       - 'TemplateDiscriminatorKubernetesSec'
       - 'TemplateDiscriminatorInlineConfig'
       - 'TemplateDiscriminatorHttpConfig'
+    CertSigningRequestApproval:
```
hexfusion, line 3012: question: Approval sounds a bit presumptuous :) is this CertSigningRequestStatus?

hexfusion (commented):
couple questions looks good!

api/v1alpha1/openapi.yaml
lkatalin: Good thought - it's modeled on https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json and they only have these 3 states, but we may want to include another that is something like "pending." This is part of the description from k8s: 

```
certificate is populated with an issued certificate by the signer after an Approved condition is present. This field is set via the /status subresource. Once populated, this field is immutable.\n\nIf the certificate signing request is denied, a condition of type \"Denied\" is added and this field remains empty. If the signer cannot issue the certificate, a condition of type \"Failed\" is added and this field remains empty"
```

api/v1alpha1/openapi.yaml
lkatalin: True - though since the `message` field is described as human-readable (again this is modeled on k8s), that makes me think this is meant to be machine readable and should only contain expected values. But I am not sure of that.

lkatalin: Looking again, maybe they are both human-readable and it's fine.

api/v1alpha1/openapi.yaml
lkatalin: I assume this ties into access control somehow but I could look more into it.

api/v1alpha1/openapi.yaml
lkatalin: Another good point - so, this is where I changed from the k8s version as they do in fact have an array and multiple usages are possible.

As far as I was aware a couple of days ago while creating this, we only had three valid usages: `enrollment`, `agent`, and `renewal`, and these are all non-overlapping. I am now hearing that we may have other usages as well and so I'd like to check in about this.

api/v1alpha1/openapi.yaml
lkatalin: It could help from a human-readability perspective.

api/v1alpha1/openapi.yaml
lkatalin: Hm, not sure what you mean - we have this schema for [EnrollmentRequests](https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/openapi.yaml#L2478) and I figured it would work the same way. We've (me, Anton, and Frank) also talked about using "approvals" in the CSR policies before a certificate can be signed.

api/v1alpha1/openapi.yaml
lkatalin: Also I think we're actually just going to add a few variants to the `Condition` schema instead of adding this.

lkatalin:
Thank you for taking a look @hexfusion !
I'm closing this in favor of #405 - but I've carried your points over to there.
Apologies for the multiple PR confusion as I was working out what exactly was breaking the CI tests.

PR #404, opened 2024-07-25
NO-ISSUE: Avoid deadloacks when running multiple parallel threads

ori-amizur:
When applying examples/fleet.yaml and
example/repository-flightctl-demos.yaml in a loop, this change avoided deadlocks to occure when the worker ran with 8 threads.


avishayt (approved):
Nice!

PR #405, opened 2024-07-25
MGMT-18170: Add CertSigningRequest schemas

lkatalin:
These are modeled on
https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json


api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
       - 'TemplateDiscriminatorKubernetesSec'
       - 'TemplateDiscriminatorInlineConfig'
       - 'TemplateDiscriminatorHttpConfig'
+    CertSigningRequestApproval:
```
fzdarsky, line 3012: Spell out "Certificate" (here and below)

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
...
+        approvedBy:
+          type: string
+          description: 'approvedBy is the name of the approver.'
+        approvedAt:
+          type: string
+          format: date-time
+          description: 'approvedAt is the time at which the request was approved.'
+      required:
+        - approved
+    CertSigningRequestCondition:
```
fzdarsky, line 3032: Instead of defining a new condition type, would it make sense to reuse the generic `Condition` type and add to the `ConditionType` enum? I see K8s defines a separate type, but IDK the motivation for this.

api/v1alpha1/openapi.yaml
```diff
@@ -3009,3 +3009,157 @@ components:
       - 'TemplateDiscriminatorKubernetesSec'
       - 'TemplateDiscriminatorInlineConfig'
       - 'TemplateDiscriminatorHttpConfig'
+    CertSigningRequestApproval:
+      type: object
+      properties:
+        labels:
```
fzdarsky, line 3015: Unlike with EnrollmentRequests, CSRs shouldn't change labels.

fzdarsky (commented):
We're missing the path definitions still.

api/v1alpha1/openapi.yaml
lkatalin: Hm yes, that makes sense.

lkatalin:
> We're missing the path definitions still.

Yes, these are coming. `make build` needs a lot of functionality to already be stubbed in, ex. in `internal/service`, so I am having to add those when I add the new endpoints. Perhaps that should be a separate PR.

lkatalin:
A few good points from @hexfusion over on https://github.com/flightctl/flightctl/pull/402/ (which I am closing in favor of this):

- Should we include a "pending" state under `Condition` to represent a pre-approval/pre-denial state? Kubernetes doesn't do this, but I do think it would make sense.

- We have a "groups" property on the ` CertSigningRequestSpec` just to follow the kubernetes example - do we want to keep this?

- Will we have other `usage`s beyond `enrollment`, `agent`, and `renewal`? This would impact not just the `... Usage` schema but whether the spec uses an array for that property vs. a string. This also actually gets to a larger point about what types of requests we expect the CSR endpoint to be fielding.

lkatalin:
@fzdarsky This one is ready, let me know if you'd like any changes.

fzdarsky:
> * Should we include a "pending" state under `Condition` to represent a pre-approval/pre-denial state? Kubernetes doesn't do this, but I do think it would make sense.
Isn't this implied by the `certificatesigningrequest.status` having neither `Approved` nor `Denied` conditions?

> * We have a "groups" property on the ` CertSigningRequestSpec` just to follow the kubernetes example - do we want to keep this?
I don't think we've figured out how users are organized into gropus/orgs/..., so I'd leave it out for now. Easy to add / hard to remove later.

> * Will we have other `usage`s beyond `enrollment`, `agent`, and `renewal`? This would impact not just the `... Usage` schema but whether the spec uses an array for that property vs. a string. This also actually gets to a larger point about what types of requests we expect the CSR endpoint to be fielding.

Aren't "usages" the ones defined in the X.509 spec (i.e. "signing", "digital signature", "client auth" etc.)?
To encode the scope/purpose, you'd instead choose a signer (using the `signerName` field). IOW, there'd be an "enrollment-signer" and an "agent-signer" or something like that. I don't think we need a "renewal-signer", because that's simply the "agent-signer" again. Signers would eventually use different signing certs. See https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers.

lkatalin:
> Isn't this implied by the `certificatesigningrequest.status` having neither `Approved` nor `Denied` conditions?

It certainly can be. I'm not sure if it makes a difference. So we can leave it implied to match kubernetes.

> I don't think we've figured out how users are organized into gropus/orgs/..., so I'd leave it out for now. Easy to add / hard to remove later.

Removed!
 
> Aren't "usages" the ones defined in the X.509 spec (i.e. "signing", "digital signature", "client auth" etc.)? 

Okay sure, I see what you are saying. I've changed it.

api/v1alpha1/openapi.yaml
```diff
@@ -2971,7 +2971,12 @@ components:
     ConditionType:
       type: string
       enum:
+      # TODO - Change Approved to EnrollmentRequestApproved, or
+      # have Enrollment use CSR endpoint
       - 'Approved'             # EnrollmentRequest
+      - 'CertificateApproved'  # CertificateSigningRequest
```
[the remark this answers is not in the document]
lkatalin: Oh interesting, thanks. I'm going to update that and then go ahead and merge this.

fzdarsky (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -3016,3 +3022,97 @@ components:
...
+      required:
+        - apiVersion
+        - kind
+        - metadata
+        - items
+      description: 'CertificateSigningRequestList is a list of CertificateSigningRequest'
+    CertificateSigningRequestSpec:
+      description: 'Wrapper around a user-created CSR, modeled on kubernetes io.k8s.api.certificates.v1.CertificateSigningRequestSpec'
+      properties:
+        expirationSeconds:
```
mangelajo, line 3071: expiration seconds is not defining type and the result is an open interface on the struct https://github.com/flightctl/flightctl/pull/405/files#diff-7912c3b33172d43970c1a7662ba6f48b3c7fcc7d0c057f0d3ef84ca25e09b35aR185

mangelajo (changes requested):
I found an issue with one of the fields

api/v1alpha1/openapi.yaml
lkatalin: Good catch, thanks.

lkatalin:
Addressed your comment @mangelajo !

PR #408, opened 2024-07-29
MGMT-18261: Add shell autocompletion script generator command

vraiti:
Adds `flightctl completion` command that writes an autocompletion script for the specified shell to stdout. This will autocomplete nouns declared in cobra.Command.ValidArgs and flags bound in o.Bind.


internal/cli/helpers.go
```diff
@@ -34,6 +34,16 @@ var (
 	}
 )
 
+func getValidResourceKinds() []string {
+	resourceKinds := make([]string, len(pluralKinds))
+	i := 0
+	for _, v := range pluralKinds {
+		resourceKinds[i] = v
+		i++
+	}
```
avishayt, line 43:
```suggestion
	for i, v := range pluralKinds {
		resourceKinds[i] = v
	}
```

avishayt (commented):
It looks great.  I tested it and it works well.  I just had one minor comment.

vraiti:
> It looks great. I tested it and it works well. I just had one minor comment.

Turns out that `range dict` will return (key, value), so declaring the index and incrementing it explicitly is the only way to convert a dict to a slice AFAIK

avishayt (approved):

PR #413, opened 2024-07-31
MGMT-18497: Support device name parameter

avishayt:


internal/tasks/config_helpers.go
```diff
@@ -4,46 +4,52 @@ import (
...
 	for _, match := range matches {
 		paramsToMatches[match[0]] = match[1]
 	}
 
 	for param, match := range paramsToMatches {
-		if !labelRegex.MatchString(param) {
+		switch {
+		case nameRegex.MatchString(param):
+			if objectMeta.Name == nil {
+				return []byte(""), fmt.Errorf("parameter referenced name, but no name found")
```
hexfusion, line 34: optional: it's common to return nil if a function fails to generate a result
```suggestion
				return nil, fmt.Errorf("parameter referenced name, but no name found")
```

hexfusion (approved):

PR #418, opened 2024-07-31
MGMT-18276: Add CLI flag to show raw HTTP requests/responses

vraiti:
Allow user to specify `--verbose-http,-H` to see "raw" http requests and responses.

Since actually intercepting the http requests/responses would require modifying the net package, the printed messages are constructed from info in the http.{Request,Response} objects.

This includes the changes from #411 since both are somewhat dependent on each other.


mangelajo:
It needs rebase now ':)

not sure how versed you are with git, I would do:

git fetch --all
git rebase HEAD~3 --onto remotes/origin/main

then follow the instructions carefully ':D

Then you may need to sign them all too:

git rebase -S HEAD~3 -i

mark them all for edit...
and continue one by one

PR #419, opened 2024-08-01
MGMT-18126: Rename device.status.updatedAt to lastSeen

avishayt:
@celdrake FYI


hexfusion (approved):
reads better but this will break clients, technically fine

PR #430, opened 2024-08-06
EDM-301: Remove mTLS from user facing API

rawagner:


internal/crypto/tls.go
```diff
@@ -7,30 +7,45 @@ import (
...
 		ClientCAs:    caPool,
 		ClientAuth:   tls.RequireAndVerifyClientCert,
 		MinVersion:   tls.VersionTLS13,
-	}, nil
+	}
+
+	grpcTlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		ClientCAs:    caPool,
+		ClientAuth:   tls.VerifyClientCertIfGiven,
```
rawagner, line 44: another attempt to make grpc endpoit work for both users & devices ... until we have a proper user-facing endpoint.

The mTLS cert is optional. Any connection has to present either valid Auth header (users) or mTLS cert (devices).

@mangelajo WDYT ?

mangelajo:
E2e: you need to add the skip insecure certs probably to the login call in the cli_tests.go (tests/e2e)

And I suspect you will need to do a similar thing for the integration testing clients: 
https://github.com/flightctl/flightctl/actions/runs/10265028056/job/28400326907?pr=430#step:4:114

mangelajo:
https://github.com/flightctl/flightctl/actions/runs/10265622273/job/28402206831?pr=430#step:10:217

Now the console command is failing to connect to the grpc agent endpoint, check if it's sending/receiving the auth token

rawagner:
> E2e: you need to add the skip insecure certs probably to the login call in the cli_tests.go (tests/e2e)
> 
> And I suspect you will need to do a similar thing for the integration testing clients: https://github.com/flightctl/flightctl/actions/runs/10265028056/job/28400326907?pr=430#step:4:114

Thanks for the pointers. I've made sure that the CA cert/I is properly added to the client. The tests are passing now :)

rawagner:
Thanks for the pointers. I've made sure that CA cert/InsecureSkipVerify is properly configured for clients. The tests are passing now.

mangelajo:
Thank you @rawagner good work, doing a final pass.

internal/api_server/agentserver/grpc.go
```diff
@@ -43,11 +48,51 @@ func NewAgentGrpcServer(
 		pendingStreams: &sync.Map{},
 	}
 }
+func validateClientTlsCert(ctx context.Context) (context.Context, error) {
```
mangelajo, line 51: Can we as a follow up move validateClientTlsCert, and authFn to the API server middleware (this one in tls, the other one probably another .go file) ?

mangelajo (approved):
Looks good, just asking to reorganize a couple of functions in (opinion) better place.

internal/api_server/agentserver/grpc.go
rawagner: Sure, i will do that in a followup.

rawagner: Followup https://github.com/flightctl/flightctl/pull/435

PR #423, opened 2024-08-05
[WIP] MGMT-17001: add support for spec rollback

hexfusion:
This PR is a sketch of adding support for rolling back device spec. This is still very much a draft and not yet intended for review.

PR #438, opened 2024-08-07
MGMT-17001: add support for greenboot rollback on agent service failure

hexfusion:
The script checks if the flightctl-agent.service is active and running without failure. It waits for the service to become active within a specified timeout (60s) and ensures it remains running for an additional duration (60s). If the service fails or does not become active in time, the script aborts exiting non 0 with an error message.

- On failure 1,2 greenboot will reboot
- On on failure 3 it will rollback if available.

Testing: manually tested

### DIFF
--- a/packaging/greenboot/flightctl-agent-running-check.sh
+++ b/packaging/greenboot/flightctl-agent-running-check.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+
+set -x -eo pipefail
+
+TIMEOUT=60
+SUCCESS_DURATION=60
+
+# This script is used to check if the flightctl-agent service is running. It is
+# used in the greenboot process to ensure that the flightctl-agent service is
+# running before proceeding. The script will wait for the service to be active,
+# and will fail if the service is not active within the timeout.
+
+verify_flightctl_agent_status() {
+  local -r failed_status=$(systemctl is-failed flightctl-agent.service)
+  local -r active_status=$(systemctl is-active flightctl-agent.service)
+
+  # fail fast
+  if [ "${failed_status}" = "failed" ] ; then
+    echo "Critical: flightctl-agent.service has encountered a failure. Aborting..."
+    kill -TERM ${SCRIPT_PID}
+  fi
+
+  if ! [ "${active_status}" = "active" ] ; then
+    return 1
+  fi
+
+  return 0
+}
+
+wait_for_success() {
+  local timeout=$1
+  local success_duration=$2
+  shift 2
+  local command="$@"
+
+  local -r start=$(date +%s)
+  local now
+
+  while : ; do
+    $command & # run in background
+    local command_pid=$!
+
+    while : ; do
+      # verify command still running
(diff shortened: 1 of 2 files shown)


packaging/rpm/flightctl.spec
```diff
@@ -52,7 +56,11 @@ cp packaging/systemd/flightctl-agent.service %{buildroot}/usr/lib/systemd/system
 /usr/bin/flightctl-agent
 /usr/lib/systemd/system/flightctl-agent.service
 %{_sharedstatedir}/flightctl
+/usr/lib/greenboot/check/required.d/20_check_flightctl_agent.sh
+/usr/lib/greenboot/red.d/10_rollback_os.sh
 
 %changelog
 * Wed Mar 13 2024 Ricardo Noriega <rnoriega@redhat.com> - 0.0.1-3
 - New specfile for both CLI and agent packages
+* Wed Aug 7 2024 Sam Batschelet <sbatsche@redhat.com> - 0.0.1-4
```
mangelajo, line 65: We also need to bump Release: 3%{?dist} up in the spec to 4 :)

hexfusion:
probably going to just move this to https://github.com/flightctl/flightctl/pull/423 but will leave it open for now

packaging/greenboot/flightctl-agent-running-check.sh
```diff
@@ -0,0 +1,76 @@
...
+set -x -eo pipefail
+
+TIMEOUT=60
+
+# This script is used to check if the flightctl-agent service is running It is
+# used in the greenboot process to ensure that the flightctl-agent service is
+# running before proceeding The script will wait for the service to be active,
+# and will fail if the service is not active within the timeout
+
+verify_flightctl_agent_status() {
```
avishayt, line 12: This won't catch a flaky agent, right?  (i.e., keeps starting and crashing)
Not a blocker, but if this is true then you can open a continuation JIRA.

hexfusion: I am also unsure how greenboot would handle that as this is a "is boot sane" check failures during runtime might need another way of monitoring I will dig further.

avishayt: I meant something like "make sure it stays up for a minute straight before assuming it's OK".  Of course it can crash again later.

avishayt (approved):

packaging/greenboot/bootc-rollback-os.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -x
+set -eo pipefail
+
+LOGFILE="/var/log/rollback.log"
+TIMESTAMP=$(date "+%Y-%m-%d %H:%M:%S")
+
+rollback() {
```
fzdarsky, line 9: When a script in `/usr/lib/greenboot/check/required.d/` fails, greenboot should _automatically_ reboot twice into the new version, and on the 3rd attempt roll-back.

avishayt (approved):

PR #439, opened 2024-08-08
NO-ISSUE: 3 bug fixes related to resourceVersion

avishayt:


internal/service/device.go
```diff
@@ -241,6 +241,7 @@ func (h *ServiceHandler) PatchDevice(ctx context.Context, request server.PatchDe
 	}
 
 	common.NilOutManagedObjectMetaProperties(&newObj.Metadata)
+	newObj.Metadata.ResourceVersion = nil
```
ori-amizur, line 244: Aren't we missing the point of resource version? Don't we want to avoid updating  if someone else updated the resource ?

internal/tasks/fleet_rollout.go
```diff
@@ -216,3 +219,27 @@ func (f FleetRolloutsLogic) getDeviceConfig(device *api.Device, templateVersion
...
+	for i := 0; i < 10; i++ {
+		if device.Metadata.Owner == nil || *device.Metadata.Owner != f.owner {
+			return fmt.Errorf("device owner changed, skipping rollout")
+		}
+
+		device.Spec = newDeviceSpec
+		_, _, err = f.devStore.CreateOrUpdate(ctx, f.resourceRef.OrgID, device, nil, false, f.callbackManager.DeviceUpdatedCallback)
+		if err != nil {
+			if errors.Is(err, flterrors.ErrResourceVersionConflict) {
+				device, err = f.devStore.Get(ctx, f.resourceRef.OrgID, *device.Metadata.Name)
```
ori-amizur, line 235: Ignore the returned error?

internal/store/device.go
```diff
@@ -185,9 +193,11 @@ func (s *DeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string) (*a
 
 func (s *DeviceStore) createDevice(device *model.Device) (bool, error) {
 	device.Generation = lo.ToPtr[int64](1)
+	originalResourceVersion := device.ResourceVersion
 	device.ResourceVersion = lo.ToPtr[int64](1)
 	if result := s.db.Create(device); result.Error != nil {
 		err := flterrors.ErrorFromGormError(result.Error)
+		device.ResourceVersion = originalResourceVersion
```
ori-amizur, line 200: I don't understand why is this needed.  We create the device from the `api.Device` during every retry iteration.

internal/service/device.go
avishayt: No, as I mentioned in the commit message, this is a PATCH.  The user didn't specify the resourceVersion - they just want to replace some field in the DB, regardless of what's currently there.  From the kubernetes docs:

> Instead of sending a PUT request, the client can send an instruction to the API server to patch an existing resource. A patch is typically appropriate if the change that the client wants to make isn't conditional on the existing data. Clients that need effective detection of lost updates should consider making their request conditional on the existing resourceVersion (either HTTP PUT or HTTP PATCH), and then handle any retries that are needed in case there is a conflict.

internal/tasks/fleet_rollout.go
avishayt: Good call, fixed

internal/service/device.go
ori-amizur: OK

internal/store/device.go
avishayt: You're right.  I was fixing the tests and I guess some failure led to me thinking this was an issue.  Reverted.

ori-amizur (approved):

PR #440, opened 2024-08-08
EDM-327: agent: ensure stale configs are removed on sync

hexfusion:
This is really a bug today we only add new configs we don't purge

TODO:

- [x] add tests

### DIFF
--- a/internal/agent/device/config/controller.go
+++ b/internal/agent/device/config/controller.go
@@ -12,6 +12,7 @@ import (
 // against the device spec.
 type Controller struct {
 	deviceWriter *fileio.Writer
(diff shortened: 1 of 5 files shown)


internal/agent/device/config/controller.go
```diff
@@ -22,11 +23,12 @@ func NewController(
 ) *Controller {
 	return &Controller{
 		deviceWriter: deviceWriter,
+		trackedFiles: make(map[string]struct{}),
```
hexfusion, line 26: Still needs logic to remove tracked

ori-amizur (approved):

PR #441, opened 2024-08-08
EDM-299: add device applications api 

hexfusion:
This PR introduces Applications as a top-level extension of the Device and Fleet API, enabling the declaration of applications in the context of an application package. An application package is defined as a collection of one or more application manifests representing supported types. In this initial iteration, the manifest type supported is podman-compose, and the package type is an OCI image.

A typical example of an application could look like this:

 ```
spec:
   applications:
     - name: "wordpress"
       image: "quay.io/org/wordpress-bundle:v1.0"
       envVars:
         APP_ENV: "production"
         LOG_LEVEL: "info"
         SOME_KEY: "{{ .device.metadata.label[x] }}"
 ```
 
 ref:
[1] [Flightctl: Applications API design](https://docs.google.com/document/d/1t0ymgOUwhkA99DTZNmHISesavagwl3G6OSTctKMndQ4/edit) 
 [2] [Application Package Specification](https://docs.google.com/document/d/1Yw-9mQ_HON3zDJNyLWCdMUMwQLZ8E8xvjBvXzMuRCxM/edit#heading=h.sd418tys091k)


api/v1alpha1/openapi.yaml
```diff
@@ -2580,6 +2580,64 @@ components:
...
+      oneOf: # extend this list with more providers to ensure that only one provider is used
+        - required: ["podmanCompose"]
+        # - required: ["quadlets"]
+    ConfigMounts:
+      type: array
+      items:
+        $ref: '#/components/schemas/ApplicationConfigRef'
+      description: |
+        An array of configuration mounts. 
+        Each item maps a configuration name reference to a mount point inside the application.
```
mangelajo, line 2613: A configuration name? what's that exactly where does it come from?

hexfusion: I will flesh out the docs a bit but

https://github.com/flightctl/flightctl/blob/3e1633c7dc7a647e94f8847c08ee6ed4ae10c238/api/v1alpha1/openapi.yaml#L2493-L2499

```
      config:
        - name: base
          configType: GitConfigProviderSpec
          gitRef:
            repository: flightctl-demos
            targetRevision: main
            path: /demos/basic-nginx-demo/configuration
  ```

Comment (hexfusion) [reply api/v1alpha1/openapi.yaml:2613]: I gues we could use `{ device.config[n].name }`

Comment (hexfusion) [reply api/v1alpha1/openapi.yaml]: I guess technically its a device.respository.name in the db.

Comment (hexfusion) [reply api/v1alpha1/openapi.yaml]: So I guess the question is this a  `ConfigMount` or `RepoMount`

Comment (hexfusion) [reply api/v1alpha1/openapi.yaml]: I will align with config
```
The name of the repository resource to use as the sync source
```

Comment (hexfusion) [reply api/v1alpha1/openapi.yaml]: But when it reaches the agent its a config right? :)

Comment (mangelajo) [reply api/v1alpha1/openapi.yaml]: The agent doesn't really know, it's just a set of files. Which have been mapped through ignition. But the server could also resolve those volume mount paths to device paths

**Comment by mangelajo** [comment]
I think what you had at the start was ok, isn't the ref to the config name enough? :)

**Comment by avishayt** [comment]
I think a design doc would help understand the vision and facilitate discussion

**Comment by hexfusion** [comment]
Since this is not going to make poc we can take the time to do that.

File: api/v1alpha1/openapi.yaml [hunk 2]
```diff
@@ -3386,6 +3501,10 @@ components:
                 type: string
         config:
           type: string
+        applications:
+          type: array
+          items:
+            $ref: '#/components/schemas/ApplicationSpec'
```
fzdarsky, line 3507: RenderedApplicationSpec?

api/v1alpha1/openapi.yaml
```diff
@@ -2549,26 +2521,169 @@ components:
...
+          properties:
+            image:
+              $ref: '#/components/schemas/ApplicationBundleImageRef'
+            data:
+              $ref: '#/components/schemas/ApplicationBundle'
+          oneOf:
+            - required: ["image"]
+            - required: ["data"]
+        envVars:
+          $ref: '#/components/schemas/ApplicationEnvVars'
```
fzdarsky, line 2643: If an application type offers passing in both envVars and envFiles (e.g. Helm `--set` vs `-f` flags), they typically define an order of precedence for duplicate keys that the service wouldn't know about. Converting to envVars may then cause unexpected behavior.

api/v1alpha1/openapi.yaml
```diff
@@ -2549,26 +2521,169 @@ components:
...
+        - bundle
+    ApplicationEnvVars:
+      type: object
+      description: "Environment variable key-value pairs, injected during runtime"
+      additionalProperties:
+        type: string
+    ApplicationBundleImageRef:
+      type: object
+      properties:
+        imageRef:
```
fzdarsky, line 2624: Just `image` (for consistency with `spec.os`)?

api/v1alpha1/openapi.yaml
```diff
@@ -2549,26 +2521,169 @@ components:
...
+            image:
+              $ref: '#/components/schemas/ApplicationBundleImageRef'
+            data:
+              $ref: '#/components/schemas/ApplicationBundle'
+          oneOf:
+            - required: ["image"]
+            - required: ["data"]
+        envVars:
+          $ref: '#/components/schemas/ApplicationEnvVars'
+    ApplicationBundle:
```
fzdarsky, line 2644: Would the ApplicationBundle not simply the content of the corresponding OCI image / ApplicationBundleImage, i.e. directories and files? Why do we need to create an API for this?

hexfusion: Few reasons why
- future proof: in the future bootc will want a secret/configmap like object to apply configs. This object would conform to that pattern.
- metadata: the files themselves don't have metadata such as dependencies and file permissions. If we require an object in the bundle to define this why not have automation to "bundle" this into a standard resource?  That process could also validate the sanity of a bundle.
- file delivery: today we only provide ignition based file delivery we would need a new alternative or use ignition again which feels like overkill

api/v1alpha1/openapi.yaml
hexfusion: ok it would echo but bootc has the same problem I guess

api/v1alpha1/openapi.yaml
hexfusion: I suppose but that can be solved with documentation. Not shipping a file greatly simplifies delivery.

api/v1alpha1/openapi.yaml
hexfusion: going to drop for this iteration of the design. this is a larger problem that requires more thought.

api/v1alpha1/openapi.yaml
hexfusion: I think we can leave this as is and then followup with file delivery

api/v1alpha1/openapi.yaml
```diff
@@ -2549,26 +2521,131 @@ components:
...
+        - bundle
+    ApplicationEnvVars:
+      type: object
+      description: "Environment variable key-value pairs, injected during runtime"
+      additionalProperties:
+        type: string
+    ApplicationBundleImageRef:
+      type: object
+      properties:
+        image:
```
hexfusion, line 2624: I think we could consider moving this back to a string I extended to an object to allow for concerning things like authentication. But I believe bundle can  have `secrets:`  which could provide this auth

avishayt:
Looks good to me.  @fzdarsky ?

api/v1alpha1/openapi.yaml
```diff
@@ -2569,6 +2569,42 @@ components:
               - filePath
           required:
           - httpRef
+    ApplicationSpec:
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/ApplicationEnvVars'
+        - type: object
+          properties:
+            name:
```
celdrake, line 2578: The current definition has `name` as optional. Shouldn't it be a required field?

fzdarsky: I agree the app should have a name (that's then shown in lists of apps in UI and CLI). Could we say that if the name is not specified, we default it to the image's base name?

api/v1alpha1/openapi.yaml
```diff
@@ -2569,6 +2569,42 @@ components:
...
+    ImageApplicationProvider:
+      type: object
+      properties:
+        image:
+          type: string
+          description: "Reference to the container image for the application package"
+    ApplicationEnvVars:
+      type: object
+      properties:
+        envVars:
```
fzdarsky, line 2593: The name "envVars" suggests a specific mechanism how parameters are passed into the templating mechanism, which doesn't apply to Helm charts, for example. Not saying we need to change it, just wanted to point out this potential source of confusion.

fzdarsky (approved):
Just as reminder that after agreeing on the API we should immediately add input validation to the API server.

api/v1alpha1/openapi.yaml
```diff
@@ -2569,6 +2569,42 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/ApplicationEnvVars'
+        - type: object
+          properties:
+            name:
+              type: string
+              description: "The name of the application"
+        - oneOf:
+            - $ref: '#/components/schemas/ImageApplicationProvider'
+            # extend application providers
```
mangelajo, line 2583:
```suggestion
            # extend application providers
```

ok that makes sense, I hope we will be able to add other application providers (like coming from a directory in the device that holds podman-compose, helm charts or other stuff i.e)

api/v1alpha1/openapi.yaml
hexfusion: Will improve the docs.

api/v1alpha1/openapi.yaml
hexfusion: We have been back and forth I agree with Frank that in the image case we can just use the image basse name. But in gitops or inline is this as clear?

api/v1alpha1/openapi.yaml
hexfusion: Right this is MVP for applications and with this we can demo the flow e2e. From there we will branch out.

PR #443, opened 2024-08-09
EDM-330: Wait for keycloak deployment if auth is enabled

rawagner:


celdrake (approved):
Run the target 4 times and all succeded :)

PR #448, opened 2024-08-12
MGMT-18558: Add CSR model to internal store / database

lkatalin:
This is a prereq for being able to add the correct logic to the CSR API paths in [`internal/service`](https://github.com/flightctl/flightctl/blob/main/internal/service/certificatesigningrequest.go). So far it is modeled on the enrollment request.

### DIFF
--- a/internal/store/certificatesigningrequest.go
+++ b/internal/store/certificatesigningrequest.go
@@ -1,0 +1,218 @@
+package store
+
+import (
+	"context"
+	b64 "encoding/base64"
+	"encoding/json"
+	"errors"
+	"reflect"
+
+	api "github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/flterrors"
+	"github.com/flightctl/flightctl/internal/store/model"
+	"github.com/google/uuid"
+	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+type CertificateSigningRequest interface {
+	Create(ctx context.Context, orgId uuid.UUID, req *api.CertificateSigningRequest) (*api.CertificateSigningRequest, error)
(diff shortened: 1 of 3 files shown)


internal/store/certificatesigningrequest.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+type CertificateSigningRequest interface {
+	Create(ctx context.Context, orgId uuid.UUID, req *api.CertificateSigningRequest) (*api.CertificateSigningRequest, error)
+	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.CertificateSigningRequestList, error)
+	Get(ctx context.Context, orgId uuid.UUID, name string) (*api.CertificateSigningRequest, error)
+	CreateOrUpdate(ctx context.Context, orgId uuid.UUID, certificatesigningrequest *api.CertificateSigningRequest) (*api.CertificateSigningRequest, bool, error)
```
amir-yogev-gh, line 23: Why do you need both "Create" & "CreateOrUpdate"?
What is the use case that will use CreateOrUpdate?
Do we have a scenario where we try to update existing request but end up creating a new one?

avishayt (approved):
Looks like good copy-paste :)

internal/store/certificatesigningrequest.go
lkatalin: There is one of these methods for each resource (Fleet, Repository, etc.). In the case of the enrollment request (which this closely mimics), it looks like it's used for the PUT method [here](https://github.com/flightctl/flightctl/blob/b3b9c71465d7130865977606f2c8c4fce1792b49/internal/service/enrollmentrequest.go#L154). I believe this makes sense because PUT "creates a new resource or replaces a representation of the target resource with the request payload".

amir-yogev-gh: Roger that. thx!

PR #449, opened 2024-08-13
MGMT-18492: fixup hooks API

hexfusion:
This PR adds some fixups to the hooks API. Previous to this change the Reboot hooks included the `on` and `path` fields both of which are not usable in the Reboot hooks. `on` has been changed to `onFile`.
```
    hooks:
        beforeUpdating:
          - name: "Before Update Hook"
            description: "Hook to show cool execute before updating"
            path: "/etc/motd"
            onFile: [Create]
            actions:
              - executable:
                  run: "echo 'Before updating' > /tmp/hook-output"
 ```
 minimal
 ```
     hooks:
        beforeUpdating:
          - path: "/etc/motd"
            onFile: [Create]
            actions:
              - executable:
                  run: "echo 'Before updating' > /tmp/hook-output"
 ```

### CODE REVIEW

**Review by avishayt — APPROVED** [review]

**Review by avishayt — APPROVED** [review]

File: api/v1alpha1/openapi.yaml [hunk 1]
```diff
@@ -2703,16 +2721,19 @@ components:
     HookAction:
```
celdrake, line 2721: @hexfusion There seems to be something weird with this type.
Previously, the generated type using the openapi generator was like this:
```
export type HookAction = (Record<string, any> | HookActionSystemdSpec | HookActionExecutableSpec);
```
whereas now it fails to generate it (there is nothing after the equal).

Could we use the same pattern as in the `RepositorySpec` for example? 
```
    RepositorySpec:
      type: object
      oneOf:
        - $ref: "#/components/schemas/GenericRepoSpec"
        - $ref: "#/components/schemas/HttpRepoSpec"
        - $ref: "#/components/schemas/SshRepoSpec
```

That means there is no top level property to identify the type though, but it's more consistent with other parts of the API and would not break the type generation for us.

celdrake: I created a bug to track this issue, as it breaks the UI type generation process.
 https://issues.redhat.com/browse/EDM-449

hexfusion: The goal was to improve UX by not requiring a discriminator field. The downside is it adds complexity on implementation.

api/v1alpha1/util.go
```diff
@@ -0,0 +1,29 @@
...
+	"fmt"
+)
+
+const (
+	systemdActionType    = "systemd"
+	executableActionType = "executable"
+)
+
+// Type returns the type of the action.
+func (t HookAction) Type() (string, error) {
```
hexfusion, line 14: @celdrake this was the approach I took for Golang.

api/v1alpha1/openapi.yaml
celdrake: I don't know that much about go, and IDK what are other alternatives.

Just that the [openapi-generator](https://github.com/OpenAPITools/openapi-generator) is not able to parse this information, and that only happens with this field.

PR #451, opened 2024-08-13
EDM-297: Publish helm charts under quay.io/flightctl/charts

mangelajo:
This makes the charts compatible with ArgoCD usage, since ArgoCD expects a base OCI repo like this.

Also makes the helm charts publishing depending on the publishing of the containers. (Otherwise we would render helm charts that point to a container version that has not been published)

### DIFF
--- a/hack/publish_charts.sh
+++ b/hack/publish_charts.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+QUAY_CHARTS=${QUAY_CHARTS:-quay.io/flightctl/charts}
+FLIGHTCTL_VERSION=$(git describe --long --tags)
+FLIGHTCTL_VERSION=${FLIGHTCTL_VERSION#v} # remove the leading v prefix for version
+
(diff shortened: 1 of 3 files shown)


lkatalin (approved):
I am not a helm expert, but it lgtm.

hack/publish_charts.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+QUAY_CHARTS=${QUAY_CHARTS:-quay.io/flightctl/charts}
+VERSION=$(git describe --long --tags)
```
hexfusion, line 5:
```suggestion
FLIGHTCTL_VERSION=$(git describe --long --tags)
```

hexfusion (approved):

mangelajo:
wait do not merge yet

I noticed this will publish quay.io/flightctl/charts/flightctl-helm

while it should be 

quay.io/flightctl/charts/flightct

PR #453, opened 2024-08-13
MGMT-18558: Add validation to CSR request

lkatalin:
Edit: Text below is outdated.

==============
Some TODOs include:
- Based on the CLI command planning, this may be a request for more than one certificate and we would need to handle that.
- Do we want to simply overwrite any existing resources (seems bad)? In enrollment request we return 208 when the resource already exists, but here we chose not to have a 208 status because k8s doesn't. We could still check for this and return a different response code though ("invalid" or simliar).

Also, could use feedback on what we actually want to allow for the CSR metadata: usages, expiration, etc.


internal/util/validation/validation.go
```diff
@@ -107,6 +108,52 @@ func ValidateBearerToken(token *string, path string) []error {
...
+	errs := field.ErrorList{}
+
+	if *e < 600 {
+		errs = append(errs, field.Invalid(fieldPathFor("spec.expirationSeconds"), e, "expiration minimum is 600 seconds"))
+		return asErrors(errs)
+	}
+	return nil
+}
+
+func ValidateRequest(csr []byte) []error {
```
avishayt, line 146:
```suggestion
func ValidateCSRRequest(csr []byte) []error {
```

internal/util/validation/validation.go
```diff
@@ -107,6 +108,52 @@ func ValidateBearerToken(token *string, path string) []error {
...
+func ValidateSignerName(s string) []error {
+	errs := field.ErrorList{}
+	if s != "ca" {
+		errs = append(errs, field.Invalid(fieldPathFor("spec.signerName"), s, "must specify a valid signer"))
+		return asErrors(errs)
+	}
+	return nil
+}
+
+// TODO - should this just be a warning?
```
avishayt, line 135: Depends - what happens if it's less than 600?

internal/util/validation/validation.go
```diff
@@ -107,6 +108,52 @@ func ValidateBearerToken(token *string, path string) []error {
...
+	for _, a := range *u {
+		if a == "client auth" {
+			return nil
+		}
+	}
+	errs = append(errs, field.Invalid(fieldPathFor("spec.usages"), u, "must contain 'client auth'"))
+	return asErrors(errs)
+}
+
+// signer name from https://github.com/flightctl/flightctl/blob/75cb5bfada1f1dbd995797edb5d8a67ac251b861/cmd/flightctl-api/main.go#L26
```
avishayt, line 125: How does the user know to put "ca" in that field?  Will it ever be something else?  If not, then why have the user provide it at all?

internal/util/validation/validation.go
```diff
@@ -107,6 +108,52 @@ func ValidateBearerToken(token *string, path string) []error {
 	return asErrors(errs)
 }
 
+// usages for a CSR should include 'client auth'
+// TODO - do we want to add more here? or just not let the user populate this?
+func ValidateUsages(u *[]string) []error {
```
avishayt, line 113:
```suggestion
func ValidateCSRUsages(u *[]string) []error {
```

internal/service/common/certificatesigningrequest.go
```diff
@@ -0,0 +1,63 @@
...
+	case nil:
+		return server.CreateCertificateSigningRequest201JSONResponse(*result), nil
+	case flterrors.ErrResourceIsNil, flterrors.ErrIllegalResourceVersionFormat:
+		return server.CreateCertificateSigningRequest400JSONResponse{Message: err.Error()}, nil
+	default:
+		return nil, err
+	}
+}
+
+func ReadCertificateSigningRequest(ctx context.Context, st store.Store, request server.ReadCertificateSigningRequestRequestObject) (server.ReadCertificateSigningRequestResponseObject, error) {
```
avishayt, line 51: Should this be called from internal/service/certificatesigningrequest.go?

internal/service/common/certificatesigningrequest.go
```diff
@@ -0,0 +1,63 @@
+package common
```
avishayt, line 1: I don't foresee reuse here.  Why not move the code to the relevant functions in internal/service/certificatesigningrequest.go?

internal/util/validation/validation.go
kot-begemot-uk: 600 is an arbitrary limit. All of this should be pulled out from hard coded into config at a later date

internal/util/validation/validation.go
kot-begemot-uk: IMHO the user should not provide a signer. It should be chosen by flightctl. 
Again - config and/or configurable policy.
We should check the CSR for some fields though - it should have CA:false in its usage, otherwise anyone with a signed CSR can generate certs which may authenticate correctly vs flightctl as they have a valid ca path.

internal/util/validation/validation.go
```diff
@@ -107,6 +108,52 @@ func ValidateBearerToken(token *string, path string) []error {
...
+	errs := field.ErrorList{}
+
+	for _, a := range *u {
+		if a == "client auth" {
+			return nil
+		}
+	}
```
kot-begemot-uk, line 120: This should be two loops. 
One with fields that must be present setting an interim OK.
One with fields that must not be present bouncing it.

internal/util/validation/validation.go
lkatalin: I've made it a warning for now. We can make it stricter later if it matters.

internal/util/validation/validation.go
lkatalin: Agree, it doesn't make sense to have the user provide it currently since we only have the one signer. I think the long term vision was the "signer" field is how the user would request a certain type of certificate - enrollment cert is signed by the enrollment signer, vs. some other type of cert is some other signer.

I've added the validation as a placeholder, but we can make sure the signer field is filled by FC for now so the user doesn't get an error for no reason.

internal/service/common/certificatesigningrequest.go
lkatalin: I took this from enrollmentrequest, but yep that makes sense.

internal/util/validation/validation.go
kot-begemot-uk: That makes sense. Though in that case we will need to apply filtering/sanity checks to this field.
As far as requesting - that is exactly how it is requested in the order interface in certmanager. You have to specify a signer there.

internal/util/validation/validation.go
```diff
@@ -107,6 +114,64 @@ func ValidateBearerToken(token *string, path string) []error {
...
+		for _, na := range notAllowed {
+			if usage == na {
+				err := fmt.Sprintf("usage not allowed: %s\n", na)
+				errs = append(errs, field.Invalid(fieldPathFor("spec.usages"), u, err))
+			}
+		}
+		for i, r := range requiredAllOf {
+			if usage == r {
+				// remove fulfilled req from list
+				requiredAllOf = append(requiredAllOf[:i], requiredAllOf[i+1])
```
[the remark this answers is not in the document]
lkatalin: Whoops, not sure what happened there. Thanks

lkatalin:
Current status: should pass CI properly once I tweak the base64 en/decoding. The CSR API expects bytes where enrollment API expects a string.

lkatalin:
@avishayt @kot-begemot-uk @ori-amizur This one is ready for review.

examples/csr.yaml
```diff
@@ -1,8 +1,9 @@
 apiVersion: v1alpha1
 kind: CertificateSigningRequest
 metadata:
-  name: myCSR
+  name: mycsr
 spec:
+  request: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURSBSRVFVRVNULS0tLS0KTUlJQkJqQ0JyUUlCQURCTE1Va3dSd1lEVlFRREUwQXhNMlEyTXpKa09EWXpOek5qWVdGbFpHRTJOR1JoTkRBMQpNbVk1TlRkbVlXRTNOREptWXpjd05UQXdNalkzTkRoaVl6YzVNRFkxWXpnNE1UbGtNV0l3TUZrd0V3WUhLb1pJCnpqMENBUVlJS29aSXpqMERBUWNEUWdBRUhCTUMzME5SSlRKM05XT2xlaTlZV2tCTS9Nem9OREd5TVJLdXpFMC8KY0treVhWSmV4SHEweEluMjJZQ05OM1pETnh4d1c2TGxRZzNpakdudlN0UU9iNkFBTUFvR0NDcUdTTTQ5QkFNQwpBMGdBTUVVQ0lRRFNyeGk2a3JkQ0tjTENoZE1CM3c3SXFtRGxaaFJFR2JITUNFU1ZuNnNzUEFJZ1pMK1FySnR3CjhSK3I0cGlhWXZaa3Ryc2szeUNJdXozaEk3bG9ZbVhQM2Q4PQotLS0tLUVORCBDRVJUSUZJQ0FURSBSRVFVRVNULS0tLS0= 
+  signerName: ca
+  usages: ["client auth", "CA:false"]
```
hexfusion, line 8: OpenSSL extendedKeyUsage used camel case IE clientAuth. Would it make sense to align here?

internal/util/validation/validation.go
```diff
@@ -3,15 +3,21 @@ package validation
...
 
+	"github.com/flightctl/flightctl/internal/crypto"
 	k8sapivalidation "k8s.io/apimachinery/pkg/api/validation"
 	k8smetav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	k8sutilvalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
+const (
+	expirationSecMin = 600
```
hexfusion, line 18:
```suggestion
	minExpirationSeconds = 600
```

internal/util/validation/validation.go
```diff
@@ -107,6 +113,68 @@ func ValidateBearerToken(token *string, path string) []error {
 	return asErrors(errs)
 }
 
+func ValidateCSRUsages(u *[]string) []error {
+	errs := field.ErrorList{}
+	requiredAllOf := []string{"client auth", "CA:false"}
+	notAllowed := []string{}
```
hexfusion, line 119: if you made these maps you can reduce lookups to O(1). That should reduce this to one loop and you can just delete() off the map if required.

```
requiredAllOf := map[string]struct{}{
	"clientAuth": {},
	"CA:false":    {},
}

notAllowed := map[string]struct{}{}
```

internal/util/validation/validation.go
```diff
@@ -107,6 +113,68 @@ func ValidateBearerToken(token *string, path string) []error {
...
+	//		return nil
+	//	}
+	//}
+
+	//errs = append(errs, field.Invalid(fieldPathFor("spec.signerName"), s, "must specify a valid signer"))
+	//return asErrors(errs)
+	return nil
+}
+
+func ValidateExpirationSeconds(e *int32) []error {
```
hexfusion, line 160: can we assume e is never nil?

avishayt: No, we cannot, it's optional

kot-begemot-uk:
Mostly there. I second @hexfusion most recent comments. 

In the future, we should make the "allowed/not allowed" fields and values configurable.

lkatalin:
Thanks for the reviews - changes incorporated.

internal/util/validation/validation.go
```diff
@@ -107,6 +113,71 @@ func ValidateBearerToken(token *string, path string) []error {
...
+		return nil
+	}
+
+	errs = append(errs, field.Invalid(fieldPathFor("spec.signerName"), s, "must specify a valid signer"))
+	return asErrors(errs)
+}
+
+func ValidateExpirationSeconds(e *int32) []error {
+	if e != nil {
+		if *e < minExpirationSeconds {
```
avishayt, line 162:
```suggestion
	if e != nil && *e < minExpirationSeconds {
```

internal/util/validation/validation.go
```diff
@@ -107,6 +113,71 @@ func ValidateBearerToken(token *string, path string) []error {
...
+			log.Printf("WARNING: recommended minimum for ExpirationSeconds is %d", minExpirationSeconds)
```
avishayt, line 163: The user won't see this (it will go to the service's log), and it's also not using our configured logger.  I think you should turn this into a TODO comment and open a JIRA ticket for further handling.

mangelajo:
Looks good already, waiting for an update (now out of date and with some nits)

avishayt (approved):

PR #454, opened 2024-08-14
MGMT-18170: CSR CLI plus CSR API changes

avishayt:


api/v1alpha1/openapi.yaml
```diff
@@ -1560,55 +1560,6 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/CertificateSigningRequest'
-        "202":
-          description: Accepted
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/CertificateSigningRequest'
-        "401":
```
lkatalin, line 1569: Will there not be a case where the requestor may be unauthorized to update the CSR? I can see removing 202 but we might need a non-200 response available.

lkatalin: Oh wait, I see - the 401 from below will be part of this path. :)

lkatalin (approved):
This looks good to me, thank you so much @avishayt !

lkatalin (approved):

PR #455, opened 2024-08-14
NO-ISSUE:  Support microshift manifests as default hooks

ori-amizur:
Add default hooks for microshift manifests.  The watched directory is /var/usr/klusterlet-manifests The manifest is applied after create or update, and deleted before update or remove.


hexfusion (commented):
this looks great thanks @ori-amizur going to hold this so we can get validation on requirements.

avishayt (approved):

internal/agent/device/hook/default_hooks.go
```diff
@@ -46,5 +56,15 @@ func defaultBeforeUpdateHooks() []v1alpha1.DeviceUpdateHookSpec {
 					"/var/run/flightctl/compose", "1m"),
 			},
 		},
+		{
+			Name:        util.StrToPtr("microshift manifest down"),
+			Path:        util.StrToPtr("/var/usr/klusterlet-manifests"),
+			Description: util.StrToPtr("Delete the provided microshift manifest"),
+			OnFile:      &[]v1alpha1.FileOperation{v1alpha1.FileOperationRemove},
+			Actions: []v1alpha1.HookAction{
+				marshalExecutable("kubectl delete -f {{ .FileName }}", &[]string{"KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig"},
```
mangelajo, line 65: Are we 100% sure this is the desired behavior? 

Do we want to enable those hooks all by default? 

I think it's ok, although it'd be good to have a disable flag somehow for the defaults (just in case something goes wrong with our defaults)

PR #457, opened 2024-08-14
MGMT-18284: CLI Unit Tests

vraiti:
Add unit tests for Flightctl CLI

Also fixes issues identified by the unit tests (mainly error reporting)

Adds FLIGHTCTL_CLI_NO_EXIT environment variable that `flightctl console` will check for. If set, it will not call `os.Exit(0)` when the gRPC console is closed so that we can actually run unit tests on it.


internal/cli/login.go
```diff
@@ -249,12 +249,11 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
...
 
 	if resp.StatusCode() == http.StatusTeapot {
-		fmt.Println("Auth is disabled")
 		err = config.Persist(o.ConfigFilePath)
 		if err != nil {
-			return fmt.Errorf("persisting client config: %w", err)
+			return fmt.Errorf("auth is disabled, persisting client config: %w", err)
 		}
-		return nil
+		return fmt.Errorf("Auth is disabled")
```
mangelajo, line 256: Auth is disabled is not a fatal error and should not return as such. It's just a warning to users (i.e. this happens in a development environment)

If we want to return as error we should tolerate that on the other tests (smoke and e2e/cli)

lkatalin (approved):

oglok:
Please, rebase this PR.

PR #458, opened 2024-08-15
EDM-118: MGMT-17001: MGMT-17908: add support for spec rollback

hexfusion:
https://github.com/user-attachments/assets/18b70acf-89e2-4e16-ab13-a8bb5afeba08

### Rollback Logic for OS Image and Device Specification
This rollback logic ensures a reliable and consistent rollback process for the OS image and device specification in case of issues during an  OS upgrade. The key components include:

**Spec Files:** Three files—current, desired, and rollback—are used to manage state transitions during updates. Before applying a new update, a copy of the current state is made to a rollback snapshot. If an update fails, the system can revert to this rollback state. The rollback spec only contains the OS Image and the current rendered version.

**Rollback Detection:** During bootstrap the agent uses bootc status and the spec files to determine if the device is in a rollback state. The logic for a rollback is.

1. Bootc status shows no staged OS image. This tells the agent there is no upgrade in progress.
2. The currently booted OS image matches the rollback image. This indicated that the current agent state matches the last revision's current state.
3. The booted image does not match the desired OS image. If the OS upgrade had been completed successfully the agent's booted image would equal the desired.

**SpecManager:** The SpecManager exposes helmer methods that allow for bootstrap logic to understand rollback intent.

Full Flow

1. Upgrade: When an upgrade occurs, the current spec is copied to the rollback spec.
2. Reboot: After reboot, during the bootstrap process, the system checks if the booted OS matches the desired OS.
3. Continue Bootstrap: If the OS matches, the bootstrap continues, and the desired spec is copied to the current spec.
4. Rollback Detection: If the booted OS does not match the desired OS and the boot status does not show a staged image, the system checks if the booted OS matches the rollback OS but not the desired OS. If this is true, a rollback is detected.
5. After a rollback is detected the agent increments the last seen rendered version by one. This ensures that the last failed revision is skipped. The agent will then wait for a new rendered version from the API and continue as normal.

```
time="2024-08-19T01:32:18.947500Z" level=info msg="Management client set"
time="2024-08-19T01:32:19.088256Z" level=warning msg="Booted OS image (192.168.1.40:5000/flightctl-device:base) does not match the desired OS image (quay.io/sbatsche/flightctl-upgrade:latest)" file="agent/device/bootstrap.go:205"
time="2024-08-19T01:32:19.153249Z" level=warning msg="Starting spec rollback" file="agent/device/bootstrap.go:226"
time="2024-08-19T01:32:19.156018Z" level=info msg="Spec rollback complete, resuming bootstrap"
time="2024-08-19T01:32:19.167479Z" level=info msg="Bootstrap complete"
time="2024-08-19T01:32:19.168433Z" level=info msg="Starting resource manager..."
time="2024-08-19T01:32:29.137961Z" level=info msg="Rollback detected, awaiting next rendered version"
time="2024-08-19T01:32:39.123379Z" level=info msg="Rollback detected, awaiting next rendered version"
time="2024-08-19T01:32:49.075742Z" level=info msg="Rollback detected, awaiting next rendered version"
```


internal/agent/device/fileio/writer.go
```diff
@@ -26,48 +25,119 @@ const (
...
+	return newManagedFile(file, w.rootDir)
+}
+
+// RemoveFile removes the file at the provided path
+func (w *FileWriter) RemoveFile(path string) error {
+	return os.Remove(filepath.Join(w.rootDir, path))
+}
+
+func (w *FileWriter) CopyFile(src, dst string) error {
+	uid, gid, err := getUserIdentity()
```
ori-amizur, line 105: In case of rollback, should we use the original user identity?

ori-amizur:
BTW, how can we start an application after reboot, is it a reboot hook? It seems the reboot hooks are not bounded to watch paths - so if not, how can it be done?

hexfusion:
post reboot hooks will need to be read from spec.

hexfusion:
>BTW, how can we start an application after reboot, is it a reboot hook? It seems the reboot hooks are not bounded to watch paths - so if not, how can it be done?

good question I am not going to tackle reboot hooks in this PR.

internal/agent/device/device.go
```diff
@@ -147,34 +159,28 @@ func (a *Agent) syncDevice(ctx context.Context) (bool, error) {
...
+	if err := a.osImageController.Sync(ctx, desired); err != nil {
 		return false, err
 	}
 
 	// set status collector properties based on new desired spec
-	a.statusManager.SetProperties(&desired)
-
-	// we have ensured that the desired spec is applied to the device. if the
-	// rendered versions are the same no more work is needed.
-	if current.RenderedVersion == desired.RenderedVersion {
```
hexfusion, line 171: nit: going to return this with IsUpdating

internal/agent/config.go
```diff
@@ -121,17 +121,18 @@ func NewDefault() *Config {
 		ManagementService:    ManagementService{Config: *client.NewDefault()},
 		StatusUpdateInterval: DefaultStatusUpdateInterval,
 		SpecFetchInterval:    DefaultSpecFetchInterval,
-		reader:               fileio.NewReader(),
+		reader:               fileio.NewReadWriter(),
```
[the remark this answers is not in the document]
hexfusion: The config needs a Reader[1] NewReadWriter() returns a ReadWriter but only passes the required Reader. But not a required change will revert.

https://github.com/flightctl/flightctl/blob/5215f33faaa62cfe49de565a44fb769e20adf12b/internal/agent/config.go#L88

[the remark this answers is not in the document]
hexfusion: updated

internal/agent/agent.go
```diff
@@ -70,23 +71,14 @@ func (a *Agent) Run(ctx context.Context) error {
 	}(ctx)
 
 	// create file io writer and reader
-	deviceWriter, deviceReader := initializeFileIO(a.config)
-
-	currentSpecFilePath := filepath.Join(a.config.DataDir, spec.CurrentFile)
-	desiredSpecFilePath := filepath.Join(a.config.DataDir, spec.DesiredFile)
+	deviceReadWriter := fileio.NewReadWriter(fileio.WithTestRootDir(a.config.testRootDir))
```
tsorya, line 74: I wonder maybe better to allow providing deviceReadWriter to the object ? In that case there will be no need to for fileio.WithTestRootDir here and test will be able to set which path it wants without setEnv?

internal/agent/config.go
```diff
@@ -85,7 +85,7 @@ type Config struct {
 	// DefaultLabels are automatically applied to this device when the agent is enrolled in a service
 	DefaultLabels map[string]string `json:"default-labels,omitempty"`
 
-	reader *fileio.Reader
+	reader fileio.Reader
```
tsorya, line 88: it is reader cause you don't want to allow it to write? Though at the end you create it with NewReadWriter?

internal/agent/agent.go
hexfusion: you mean in the config? that does make sense

internal/agent/config.go
tsorya: Same question actually, why it should be part of config and not part of the object ?

hexfusion: correct it's an interface so as long as the interface fulfills Reader it is technically fine. And yeah the idea is to limit access to Writer where possible. But I did revert the change.

hexfusion: to do what you want we could add a readWriter I limited it more because Write wasn't needed.

tsorya: Yeah but creating object in config is sort of abuse of it , it should be part of object that will actually use it, in anycase it is only IMO

hexfusion: I am not a fan of how this is done so for sure open to ideas

internal/agent/device/fileio/reader.go
```diff
@@ -7,46 +7,45 @@ import (
...
+// FileExists checks if a path exists and returns a boolean indicating existence,
+// and an error only if there was a problem checking the path.
+func (r *reader) FileExists(filePath string) (bool, error) {
 	return checkPathExists(r.PathFor(filePath))
 }
 
-func checkPathExists(filePath string) error {
+func checkPathExists(filePath string) (bool, error) {
 	_, err := os.Stat(filePath)
-	if err != nil && os.IsNotExist(err) {
```
tsorya, line 44: i think this way it will be cleaner? WDYT?
`_, err := os.Stat(path)
    if err == nil { return true, nil }
    if os.IsNotExist(err) { return false, nil }
    return false, fmt.Errorf("error checking path: %w", err)`

internal/agent/device/fileio/writer.go
```diff
@@ -66,6 +66,51 @@ func (w *writer) RemoveFile(file string) error {
 	return nil
 }
 
+func (w *writer) CopyFile(src, dst string) error {
+	return w.copyFile(filepath.Join(w.rootDir, src), filepath.Join(w.rootDir, dst))
+}
+
+func (w *writer) copyFile(src, dst string) error {
```
tsorya, line 73: There is small project github.com/otiai10/copy for copying files, maybe we can use it

hexfusion: we want to reduce deps on the agent but we can consider it.

internal/container/container.go
```diff
@@ -152,9 +159,20 @@ func IsOsImageDirty(host *BootcHost) bool {
 // IsOsImageReconciled returns true if the booted image equals the spec image.
 func IsOsImageReconciled(host *BootcHost, desiredSpec *v1alpha1.RenderedDeviceSpec) bool {
 	// If the booted image equals the desired image, the OS image is reconciled
```
tsorya, line 161: seems like comment doesn't match  new code

internal/agent/device/fileio/reader.go
hexfusion: generally idiomatic to check !=nil first. but I don't love returning true at the end either will update.

hexfusion:
ci will be fixed by #470

avishayt (approved):

mangelajo (approved):

PR #463, opened 2024-08-17
MGMT-18558: Issue certificate when CSR is approved

lkatalin:
Note: I'll drop the first 3 commits when #453 is merged.

### DIFF
--- a/internal/crypto/csr.go
+++ b/internal/crypto/csr.go
@@ -56,7 +56,9 @@ func ParseCSR(csrPEM []byte) (*x509.CertificateRequest, error) {
 	return csr, nil
 }
 
-func (ca *CA) IssueRequestedClientCertificate(csr *x509.CertificateRequest, expiryDays int) ([]byte, error) {
+// IssueRequestedClientCertificate issues a client certificate based on the provided
+// Certificate Signing Request (CSR) and the desired expiration time in seconds.
+func (ca *CA) IssueRequestedClientCertificate(csr *x509.CertificateRequest, expirySeconds int) ([]byte, error) {
 	now := time.Now()
 	template := &x509.Certificate{
 		Subject: csr.Subject,
@@ -70,7 +72,7 @@ func (ca *CA) IssueRequestedClientCertificate(csr *x509.CertificateRequest, expi
 		Issuer: ca.Config.Certs[0].Subject,
 
 		NotBefore:    now.Add(-1 * time.Second),
-		NotAfter:     now.Add(time.Duration(expiryDays) * 24 * time.Hour),
+		NotAfter:     now.Add(time.Duration(expirySeconds) * time.Second),
 		SerialNumber: big.NewInt(1),
 
 		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
--- a/internal/service/certificatesigningrequest.go
+++ b/internal/service/certificatesigningrequest.go
@@ -8,12 +8,64 @@ import (
 
 	api "github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/api/server"
+	"github.com/flightctl/flightctl/internal/crypto"
 	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/service/common"
 	"github.com/flightctl/flightctl/internal/store"
 	"github.com/go-openapi/swag"
+	"github.com/google/uuid"
 	"k8s.io/apimachinery/pkg/labels"
 )
+
+const DefaultEnrollmentCertExpirySeconds int32 = 60 * 60 * 24 * 7 // 7 days
+
+func signApprovedCertificateSigningRequest(ca *crypto.CA, request *api.CertificateSigningRequest) error {
+	if request == nil {
+		return errors.New("signApprovedCertificateSigningRequest: CertificateSigningRequest is nil")
+	}
+
+	// NOTE: this currently only requires one approval; this will change in the future based on policy
+	if api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestDenied) {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request has been denied and cannot be signed")
+	}
+	if !api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestApproved) {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is not approved and cannot be signed")
+	}
+
+	csr, err := crypto.ParseCSR(request.Spec.Request)
+	if err != nil {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: error parsing CSR: %w", err)
+	}
+
+	if err := csr.CheckSignature(); err != nil {
+		return err
+	}
+
+	// the CN will need the enrollment prefix applied;
+	// if a CN is not specified in the CSR, generate a UUID to represent the future device
+	u := csr.Subject.CommonName
+	if u == "" {
+		u = uuid.NewString()
+	}
+	csr.Subject.CommonName, err = crypto.CNFromRequestedName(u)
+	if err != nil {
+		return fmt.Errorf("signApprovedCertificateRequest: error setting CN in CSR: %w", err)
+	}
+
+	expiry := DefaultEnrollmentCertExpirySeconds
+	if request.Spec.ExpirationSeconds != nil {
+		expiry = *request.Spec.ExpirationSeconds
+	}
+
+	certData, err := ca.IssueRequestedClientCertificate(csr, int(expiry))
+	if err != nil {
+		return err
+	}
+
+	request.Status.Certificate = &certData
+
+	return nil
+}
 
 // (DELETE /api/v1/certificatesigningrequests)
 func (h *ServiceHandler) DeleteCertificateSigningRequests(ctx context.Context, request server.DeleteCertificateSigningRequestsRequestObject) (server.DeleteCertificateSigningRequestsResponseObject, error) {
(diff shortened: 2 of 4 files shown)


internal/crypto/csr.go
```diff
@@ -56,7 +56,7 @@ func ParseCSR(csrPEM []byte) (*x509.CertificateRequest, error) {
 	return csr, nil
 }
 
-func (ca *CA) IssueRequestedClientCertificate(csr *x509.CertificateRequest, expiryDays int) ([]byte, error) {
+func (ca *CA) IssueRequestedClientCertificate(csr *x509.CertificateRequest, expirySeconds int) ([]byte, error) {
```
hexfusion, line 59: Would it be worth document these details?
```
// IssueRequestedClientCertificate issues a client certificate based on the provided 
// Certificate Signing Request (CSR) and the desired expiration time in seconds.
```

internal/service/certificatesigningrequest.go
```diff
@@ -231,15 +290,24 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
 		Message: "Approved",
 	}
 	err = h.store.CertificateSigningRequest().UpdateConditions(ctx, orgId, request.Name, []api.Condition{approvedCondition})
+	if err != nil {
+		if errors.Is(err, flterrors.ErrResourceNotFound) {
+			return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
+		} else {
```
hexfusion, line 296: optional: since you return no need for else?

```
           if err != nil {
		} 
		return nil, err
           }

File: internal/service/certificatesigningrequest.go [hunk 3]
```diff
@@ -231,15 +290,24 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
...
-		return server.ApproveCertificateSigningRequest200JSONResponse{}, nil
-	case flterrors.ErrResourceNotFound:
-		return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
-	default:
-		return nil, err
+	err = signApprovedCertificateSigningRequest(h.ca, csr)
+	if err != nil {
+		if errors.Is(err, flterrors.ErrResourceNotFound) {
+			return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
+		} else {
```
hexfusion, line 305: same

internal/service/certificatesigningrequest.go
```diff
@@ -8,13 +8,71 @@ import (
...
+
+	if request.Metadata.Name == nil {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is missing metadata.name")
+	}
+
+	// NOTE: this currently only requires one approval; this will change in the future based on policy
+	// TODO: do we need CertificateSigningRequestApproved instead of the string?
+	approved := false
+	for _, c := range request.Status.Conditions {
+		if c.Type == "Denied" {
```
ori-amizur, line 35: There are constants declaring exactly these values:
`v1alpha1.CertificateSigningRequestDenied`

avishayt, line 35: This just means that the "Denied" condition exists - it doesn't check if the condition is True or False.  So you could be returning an error here for Denied=False.
I suggest using the helper function `api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestDenied)`

(Also note using the type rather than hard-coded string)

internal/service/certificatesigningrequest.go
```diff
@@ -8,13 +8,71 @@ import (
...
+	}
+
+	if request.Metadata.Name == nil {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is missing metadata.name")
+	}
+
+	// NOTE: this currently only requires one approval; this will change in the future based on policy
+	// TODO: do we need CertificateSigningRequestApproved instead of the string?
+	approved := false
+	for _, c := range request.Status.Conditions {
```
ori-amizur, line 34: For conditions, don't we need to check the the value of condition.status to know what is the actual meaning of the condition?
status may be either true or false.

internal/service/certificatesigningrequest.go
```diff
@@ -8,13 +8,71 @@ import (
...
+		}
+		if c.Type == "Approved" {
+			approved = true
+		}
+	}
+	if !approved {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is not approved and cannot be signed")
+	}
+
+	// TODO: maybe remove since we did validation up front when first storing the CSR - but it could detect if something
```
avishayt, line 46: No need to re-validate IMO

internal/service/certificatesigningrequest.go
```diff
@@ -8,13 +8,71 @@ import (
...
+const DefaultEnrollmentCertExpirySeconds int32 = 60 * 60 * 24 * 7 // 7 days
+
+func signApprovedCertificateSigningRequest(ca *crypto.CA, request *api.CertificateSigningRequest) error {
+	if request == nil {
+		return errors.New("signApprovedCertificateSigningRequest: CertificateSigningRequest is nil")
+	}
+
+	if request.Metadata.Name == nil {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is missing metadata.name")
+	}
```
avishayt, line 29: No need to re-validate IMO

internal/util/validation/validation.go
```diff
@@ -107,6 +113,68 @@ func ValidateBearerToken(token *string, path string) []error {
...
+		err := fmt.Sprintf("required usages must be present in request: %s\n", requiredAllOf)
+		errs = append(errs, field.Invalid(fieldPathFor("spec.usages"), u, err))
+	}
+	return asErrors(errs)
+}
+
+// NOTE: ValidateSignerName will not error and is a placeholder for when FC has multiple signers.
+// Currently every request is sent to the only signer, named "ca" and defined in cmd/flightctl-api/main.go
+func ValidateSignerName(s string) []error {
+	//errs := field.ErrorList{}
```
ori-amizur, line 145: Why not verify that the only signer is "ca"?
If it needs to be commented out, why not remove the code?

internal/util/validation/validation.go
```diff
@@ -107,6 +113,68 @@ func ValidateBearerToken(token *string, path string) []error {
 	return asErrors(errs)
 }
 
+func ValidateCSRUsages(u *[]string) []error {
+	errs := field.ErrorList{}
+	requiredAllOf := []string{"client auth", "CA:false"}
+	notAllowed := []string{}
```
ori-amizur, line 119: The notAllowed are currently empty.  Why is it needed?

internal/util/validation/validation.go
```diff
@@ -107,6 +113,68 @@ func ValidateBearerToken(token *string, path string) []error {
...
+			}
+		}
+		for i, r := range requiredAllOf {
+			if usage == r {
+				// remove fulfilled req from list
+				requiredAllOf = append(requiredAllOf[:i], requiredAllOf[i+1:]...)
+			}
+		}
+	}
+	if len(requiredAllOf) > 0 {
```
ori-amizur, line 135: NIT: you can consider finding the unfulfilled requirements by:
`lo.Without(requiredAllOf, (*u)...)`

internal/util/validation/validation.go
lkatalin: It's mainly a placeholder for the future and we do not want to cause an error on this for the user currently, but I take your point. I added this back in so it will verify the signer (see #453 where these commits belong), which is a value that can just be generated internally for the moment if we want.

internal/util/validation/validation.go
lkatalin: Mainly for the future so that we recognize it is an option to add some values that aren't allowed.

internal/util/validation/validation.go
lkatalin: I ended up making the slices into maps (see comments on #453 ) but this is a good library to know about, thanks.

internal/crypto/csr.go
lkatalin: Agree, always worth it to document.

internal/service/certificatesigningrequest.go
lkatalin: The helper function is quite handy, thank you

internal/service/certificatesigningrequest.go
lkatalin: Actually, I missed this earlier but we do need it because it returns the X509 struct.

lkatalin:
This should pass CI once it is rebased on a merged #453

internal/service/certificatesigningrequest.go
```diff
@@ -8,13 +8,58 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: error parsing CSR: %w", err)
+	}
+
+	if err := csr.CheckSignature(); err != nil {
+		return err
+	}
+
+	// as we don't have a specific device yet, generate a UUID to represent the future device
+	uuid := uuid.NewString()
```
mangelajo, line 45: We should use the csr Subject CommonName if already provided as part of the CSR.

internal/service/certificatesigningrequest.go
```diff
@@ -8,13 +8,58 @@ import (
...
+	csr, err := crypto.ParseCSR(request.Spec.Request)
+	if err != nil {
+		return fmt.Errorf("signApprovedCertificateSigningRequest: error parsing CSR: %w", err)
+	}
+
+	if err := csr.CheckSignature(); err != nil {
+		return err
+	}
+
+	// as we don't have a specific device yet, generate a UUID to represent the future device
```
mangelajo, line 44: This API would initially be used to provide certs to image builders that need an enrollment certificate (instead of having to manually download our single one cert/key) more than devices themselves (that would come later I guess through having the enrollment request API use this).

For enrollment certs the tls auth expects this : https://github.com/flightctl/flightctl/blob/0a574fa0a2145e7d06f95cba25c01bad8e4790f9/internal/crypto/cert.go#L20

But I guess we should move to accept:

"client-enrollment: $uuid" or something like that  so we can have different enrollment identities.

lkatalin: I've made it so the CN created from the CSR will be `client-enrollment: $uuid` where the `uuid` is either the user-requested name or (if none) then a randomly generated uuid.

Looking at [`ValidateEnrollmentAccessFromContext`](https://github.com/flightctl/flightctl/blob/main/internal/service/agent/handler.go#L42), that function *only* is called during an EnrollmentRequest - which has its own validation rules - so until we include checks in the EnrollmentRequest process to see that a correct CSR/enrollment cert is present (post-POC IIUC), it seems like we can leave that alone.

hexfusion (approved):
Logically it seems reasonable.

PR #466, opened 2024-08-18
NO-ISSUE: Fix simulator

tsorya:
1. Fixing simutalor configuration and pathes
2. Fixing small code issues
3. Adding auto approve for devices that were created by simulator

### DIFF
--- a/.github/workflows/pr-smoke-testing.yaml
+++ b/.github/workflows/pr-smoke-testing.yaml
@@ -51,3 +51,6 @@ jobs:
 
       - name: Apply resourcesync
         run: bin/flightctl apply -f examples/resourcesync.yaml
+
+      - name: Simulator run
+        run: bin/devicesimulator --config bin/agent/etc/flightctl/config.yaml -count 1 -stop-after 1m
--- a/cmd/devicesimulator/main.go
+++ b/cmd/devicesimulator/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"log"
 	"math/rand"
 	"os"
 	"os/signal"
@@ -13,10 +14,16 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/agent"
+	"github.com/flightctl/flightctl/internal/agent/device"
+	apiClient "github.com/flightctl/flightctl/internal/api/client"
 	"github.com/flightctl/flightctl/internal/client"
 	"github.com/flightctl/flightctl/internal/util"
 	flightlog "github.com/flightctl/flightctl/pkg/log"
+	testutil "github.com/flightctl/flightctl/test/util"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


cmd/devicesimulator/main.go
```diff
@@ -64,37 +80,59 @@ func main() {
...
+	}
+
 	log.Infoln("creating agents")
 	agents := make([]*agent.Agent, *numDevices)
+	agentsFolders := make([]string, *numDevices)
 	for i := 0; i < *numDevices; i++ {
 		agentName := fmt.Sprintf("device-%04d", i)
 		certDir := filepath.Join(agentConfigTemplate.ConfigDir, "certs")
 		agentDir := filepath.Join(*dataDir, agentName)
+		err := os.Setenv(client.TestRootDirEnvKey, agentDir)
```
tsorya, line 95: Added it from test_harness but i believe this os env should be deleted and paths should be provided from outside to each object and not magically set by some env in different places. 
I believe it is worth ticket in backlog for cleanup

internal/client/config.go
```diff
@@ -148,7 +148,7 @@ func (c *Config) HasCredentials() bool {
 }
 
 func (c *Config) GetClientKeyPath() string {
-	return resolvePath(c.AuthInfo.ClientCertificate, c.baseDir)
+	return resolvePath(c.AuthInfo.ClientKey, c.baseDir)
```
tsorya, line 151: Change it as seems like it is a bug as there is GetClientCertificatePath for certificate path

mangelajo:
Nice, thank you!, 

Do you think it could somehow be added to https://github.com/flightctl/flightctl/blob/main/.github/workflows/pr-smoke-testing.yaml so we make sure it doesn't break over time?, it may need to exit 1 / 0 under some conditions (may be a flag to exit after enrolling devices or something..)

But just an idea for a follow up.

mangelajo:
@tsorya one of the changes is influencing the enrollment request ID discovery from e2e tests on the console: https://github.com/flightctl/flightctl/actions/runs/10442439645/job/28917390092?pr=466#step:10:87

tsorya:
> Nice, thank you!,
> 
> Do you think it could somehow be added to https://github.com/flightctl/flightctl/blob/main/.github/workflows/pr-smoke-testing.yaml so we make sure it doesn't break over time?, it may need to exit 1 / 0 under some conditions (may be a flag to exit after enrolling devices or something..)
> 
> But just an idea for a follow up.

Good idea, added an option to run simulator for some amount of time and in case we are able to run it should stop it after given duration and exit 0 and in case there is an issue with agent.Run we should exit 1

tsorya:
e2e failed on 
Waiting for root prompt on device 293gupbavh35npjankertq0tsetqd0ln8usitf7kdjk1d4kvpcng console
don't think it is related to the change

hexfusion (approved):

.github/workflows/pr-smoke-testing.yaml
```diff
@@ -51,3 +51,6 @@ jobs:
 
       - name: Apply resourcesync
         run: bin/flightctl apply -f examples/resourcesync.yaml
+
+      - name: Simulator run
+        run: bin/devicesimulator --config bin/agent/etc/flightctl/config.yaml -count 1 -stop-after 1m
```
mangelajo, line 56: 🚀 

Thanks!

mangelajo:
> /retest

Was it really a flake?

I thought it was due to the changes in "get id from console" changes.

@hexfusion let's document if we see flakes, we should fix them where in our hand.

tsorya:
> Looks great, I was trying to merge, but the commits need the signature
> 
> ```
> Merging is blocked
> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/authentication/managing-commit-signature-verification/about-commit-signature-verification)
> ```
> 
> can you sign the commits?
> 
> Thank you :-)

done

PR #468, opened 2024-08-19
EDM-287: 500 when trying to apply a device yaml without removing the creationTimestamp and resourceVersion

ori-amizur:
Make all API places the return ErrNoRowsUpdated to return 409 status instead of 500.


mangelajo (commented):
:+1

PR #473, opened 2024-08-20
EDM-353: Allow skipping TLS validation or providing CA cert for auth

rawagner:


rawagner:
I still need to check if custom CA option works well if provided as multiline string. Converting to draft

internal/auth/auth.go
```diff
@@ -65,17 +67,27 @@ func CreateAuthMiddleware(cfg *config.Config, log logrus.FieldLogger) (func(http
...
+		tlsConfig := &tls.Config{
+			InsecureSkipVerify: cfg.Auth.InsecureSkipTlsVerify, //nolint:gosec
+		}
+		if cfg.Auth.CaCrt != "" {
+			caCertPool := x509.NewCertPool()
+			caCertPool.AppendCertsFromPEM([]byte(cfg.Auth.CaCrt))
+			tlsConfig.RootCAs = caCertPool
+		}
+		if cfg.Auth != nil && cfg.Auth.OpenShiftApiUrl != "" {
+			log.Println("OpenShift auth enabled")
```
mangelajo, line 80: idea: I would add the URL (since we are at it :-) )

internal/auth/auth.go
```diff
@@ -65,17 +67,27 @@ func CreateAuthMiddleware(cfg *config.Config, log logrus.FieldLogger) (func(http
...
+			authZ = K8sToK8sAuth{K8sAuthZ: authz.K8sAuthZ{ApiUrl: cfg.Auth.OpenShiftApiUrl, ClientTlsConfig: tlsConfig}}
+			authN = authn.OpenShiftAuthN{OpenShiftApiUrl: cfg.Auth.OpenShiftApiUrl, ClientTlsConfig: tlsConfig}
+		} else if cfg.Auth != nil && cfg.Auth.OIDCAuthority != "" {
+			log.Println("OIDC auth enabled")
```
mangelajo, line 84: I would add details here too (if those aren't secret/private) I am not familiar enough.

rawagner:
I tried both new options, in UI and CLI. Worked fine.

internal/config/config.go
```diff
@@ -54,6 +54,8 @@ type authConfig struct {
 	OpenShiftApiUrl       string `json:"openShiftApiUrl,omitempty"`
 	OIDCAuthority         string `json:"oidcAuthority,omitempty"`
 	InternalOIDCAuthority string `json:"internalOidcAuthority,omitempty"`
+	CaCrt                 string `json:"caCrt,omitempty"`
```
hexfusion, line 57: CACert?

rawagner: Sounds good, updated

mangelajo (approved):

PR #474, opened 2024-08-20
EDM-270 Fix localhost/flightctl-db:latest issue

mangelajo:
Depending on the host, kind load docker-image does not work in combination with podman and it needs workaroud.

This commit refactors the workaround to be used for every single instance of kind load


avishayt (commented):
Looks OK other than the small comments

avishayt:
Need a colon in the commit message "EDM-270: ..."

mangelajo:
Done :-)

hexfusion (approved):
bash is the devil :)

mangelajo:
> bash is the devil :)

it is! X'D

PR #475, opened 2024-08-20
EDM-360: improve graceful termination during node reboot

hexfusion:
Currently because bootc upgrade --apply reboots the node and doesn't return we see an unexpected error from os.Exec which gives the impression that something went wrong. This PR improves graceful termination handling ensuring that sigint is only passed to the main processes on termination. `SIGINT` is caught by the agent, allowing us a timeout duration to finish any work. The PR also added `Pdeathsig` to send a `SIGTERM` to subprocesses started by the agent that will allow them to terminate when the main process is killed gracefully.

```
Aug 20 15:54:42 localhost.localdomain flightctl-agent[968]: time="2024-08-20T15:54:42.395902Z" level=info msg="Agent received shutdown signal: interrupt"
Aug 20 15:56:12 localhost.localdomain flightctl-agent[968]: time="2024-08-20T15:56:12.405914Z" level=error msg="Failed to sync device: failed to update os image: apply image: signal: killed" file="agent/device/device.go:93"
Aug 20 15:56:12 localhost.localdomain systemd[1]: flightctl-agent.service: Killing process 968 (flightctl-agent) with signal SIGKILL. 
```


pkg/executer/executer.go
```diff
@@ -44,7 +45,17 @@ func (e *CommonExecuter) execute(cmd *exec.Cmd) (stdout string, stderr string, e
 	var stdoutBytes, stderrBytes bytes.Buffer
 	cmd.Stdout = &stdoutBytes
 	cmd.Stderr = &stderrBytes
+	// Set Pdeathsig to SIGTERM to kill the process and its children when the parent process is killed.
+	// This should prevent orphaned processes and allow for the sub process to gracefully terminate.
+	// ref. https://github.com/golang/go/blob/release-branch.go1.21/src/syscall/exec_linux.go#L91
+	cmd.SysProcAttr = &syscall.SysProcAttr{Pdeathsig: syscall.SIGTERM}
```
mangelajo, line 51: can we add the same thing to console:

https://github.com/flightctl/flightctl/blob/main/internal/agent/device/console.go#L204

So when we are killed and the console is running we don't leave any orphans, etc.

avishayt (approved):

PR #478, opened 2024-08-21
   EDM-368: An application does not restart after a device reboot or a device upgrade

ori-amizur:
Add a restart as file operation and add reboot as file operation in
default hooks.


internal/agent/device/config/controller.go
```diff
@@ -18,6 +19,7 @@ type Controller struct {
 	hookManager  hook.Manager
 	deviceWriter fileio.Writer
 	log          *log.PrefixLogger
+	initialized  atomic.Bool
```
hexfusion, line 22: we shouldn't need atomic bool here the reconciliation loops are single threaded right?

internal/agent/device/config/controller.go
```diff
@@ -33,10 +35,26 @@ func NewController(
...
+func (c *Controller) init(ctx context.Context, current *v1alpha1.RenderedDeviceSpec) {
+	if current == nil {
+		return
+	}
+	currentIgnition, err := parseAndConvertConfig(lo.FromPtr(current.Config))
+	if err != nil {
+		c.log.Warnf("failed to parse current ignition: %+v", err)
+		return
+	}
+	lo.ForEach(currentIgnition.Storage.Files, func(f ignv3types.File, _ int) { c.hookManager.OnAfterReboot(ctx, f.Path) })
```
hexfusion, line 47: I am not a fan of all of this `lo` package usage I understand it is helpful but it hides the actual logic.  Which to me defeats its purpose. Please use go code and if we need helper methods add to pkg/. The agent should not depend on 3rd party packages to perform simple operations.

internal/agent/device/config/controller.go
```diff
@@ -33,10 +35,26 @@ func NewController(
 	}
 }
 
+func (c *Controller) init(ctx context.Context, current *v1alpha1.RenderedDeviceSpec) {
```
hexfusion, line 38: I would expect any initialization to take place during bootstrap related to after reboot to take place in bootstrap.

hexfusion (changes requested):
Overall post reboot hooks should fire in bootstrap before any reconciliation logic is started. With regards to applications. Internal support for application types should ensure they restart without hooks. So I would expect something like. ensureApplications in the bootstrap logic that contains specific logic perhaps from a future applications manager where we perform any necessary actions to ensure applications start as expected.

In the case of docker-compose I would expect (for now) the spec manager should be able to give you a list of compose files to invoke via podman-compose up.

maybe something like this 
```go
func (b *Bootstrap) ensureBootstrap(ctx context.Context) error {
	desiredSpec, err := b.specManager.Read(spec.Desired)
	if err != nil {
		return err
	}

	if err := b.ensureBootedOS(ctx, desiredSpec); err != nil {
		return err
	}

        // fire all after reboot hooks (not 100% sure on ordering here)
	if err := b.hookManager.OnAfterReboot(ctx); err != nil {
		return err
	}
  
        // ensure application dependencies are met
	if err := b.ensureApplications(ctx); err != nil {
		return err
	}

	return nil
}
```

ori-amizur:
> With regards to applications. Internal support for application types should ensure they restart without hooks

This contradicts the current mechanism where we require that in order to launch applications we need hooks.  The mechanism already exists.  Why build a new one.

> In the case of docker-compose I would expect (for now) the spec manager or hooks manager could be extended to give you a list of compose files to invoke via podman-compose up.

Currently the list of existing files is kept in the `current` spec.  Why do we need another list?

Maybe the needed change now is to initially let the hookManager reconcile from the `current` spec and then let the bootstrap do the job.

internal/agent/device/config/controller.go
ori-amizur: OK

internal/agent/device/config/controller.go
ori-amizur: OK. Will change.

internal/agent/device/config/controller.go
ori-amizur: If we move to bootstrap then it is not needed

avishayt (approved):

PR #480, opened 2024-08-21
EDM-359: Return Bad request in case user tries to appove already approved request

tsorya:
EDM-359: Return Bad request in case user tries to appove already approved requestApprove cli changed to print error message and not only error status


internal/service/enrollmentrequest.go
```diff
@@ -233,29 +232,28 @@ func (h *ServiceHandler) ApproveEnrollmentRequest(ctx context.Context, request s
...
 	}
 
+	// if the enrollment request was already approved we should not try to approve it one more time
 	if request.Body.Approved {
-
+		if v1alpha1.IsStatusConditionTrue(enrollmentReq.Status.Conditions, v1alpha1.EnrollmentRequestApproved) {
+			return server.ApproveEnrollmentRequest400JSONResponse{Message: "Enrollment request is already approved"}, nil
+		}
 		if request.Body.ApprovedAt != nil {
 			return server.ApproveEnrollmentRequest422JSONResponse{Message: "ApprovedAt is not allowed to be set when approving enrollment requests"}, nil
```
avishayt, line 241: This is the only API that returns 422 - can you please change all instances to 400 while you're at it?

tsorya: sure

internal/cli/approve.go
```diff
@@ -116,15 +117,22 @@ func (o *ApproveOptions) Run(ctx context.Context, args []string) error {
...
 }
 
-func processApprovalReponse(response interface{}, err error, kind string, name string) error {
+func processApprovalReponse(response *http.Response, err error, kind string, name string) error {
 	errorPrefix := fmt.Sprintf("approving %s/%s", kind, name)
 	if err != nil {
 		return fmt.Errorf("%s: %w", errorPrefix, err)
 	}
-
-	v := reflect.ValueOf(response).Elem()
```
hexfusion, line 125: :+1:

internal/cli/approve.go
```diff
@@ -116,15 +117,22 @@ func (o *ApproveOptions) Run(ctx context.Context, args []string) error {
...
-		return fmt.Errorf(errorPrefix+": %s (%d)", v.FieldByName("Status").String(), v.FieldByName("StatusCode").Int())
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		var responseError api.Error
+		// not handling errors as we are only interested in the message
+		// and in case there will be a problem in reading the body or unmarshalling it
+		// we will print the status like
+		// Error: approving enrollmentrequest/<name>:  (422 Unprocessable Entity)
+		body, _ := io.ReadAll(response.Body)
+		_ = json.Unmarshal(body, &responseError)
```
hexfusion, line 133: why not handle this error?

internal/service/enrollmentrequest_test.go
```diff
@@ -0,0 +1,79 @@
...
+		callbackManager: dummyCallbackManager(),
+	}
+	resp, err := serviceHandler.ApproveEnrollmentRequest(context.Background(), server.ApproveEnrollmentRequestRequestObject{
+		Name: "foo",
+		Body: approval,
+	})
+	require.NoError(err)
+	r, ok := resp.(server.ApproveEnrollmentRequest400JSONResponse)
+	require.True(ok)
+	require.Equal("Enrollment request is already approved", r.Message)
```
hexfusion, line 78: :heart:

internal/service/enrollmentrequest.go
```diff
@@ -248,14 +249,13 @@ func (h *ServiceHandler) ApproveEnrollmentRequest(ctx context.Context, request s
 		}
 
 		if err := approveAndSignEnrollmentRequest(h.ca, enrollmentReq, request.Body); err != nil {
-			return server.ApproveEnrollmentRequest422JSONResponse{Message: fmt.Sprintf("Error approving and signing enrollment request: %v", err.Error())}, nil
+			return server.ApproveEnrollmentRequest400JSONResponse{Message: fmt.Sprintf("Error approving and signing enrollment request: %v", err.Error())}, nil
 		}
 
 		if err := h.createDeviceFromEnrollmentRequest(ctx, orgId, enrollmentReq); err != nil {
-			return server.ApproveEnrollmentRequest422JSONResponse{Message: fmt.Sprintf("Error creating device from enrollment request: %v", err.Error())}, nil
+			return server.ApproveEnrollmentRequest500JSONResponse{Message: fmt.Sprintf("Error creating device from enrollment request: %v", err.Error())}, nil
```
hexfusion, line 256: why is this a 500 now?

hexfusion (commented):
looks good couple questions then LGTM

internal/cli/approve.go
tsorya: I actually thought that it is not relevant for user, I mean that it is better to return something more valuable than I am failing to unmarshal server response

internal/service/enrollmentrequest.go
tsorya: Talked about it with Avishay and this exactly the case for internal server error as we are trying to create a device in db and we will fail cause of our issue and not cause of request

internal/cli/approve.go
hexfusion: fair

internal/service/enrollmentrequest.go
hexfusion: lets add that context into a short comment then LGTM :pray:

hexfusion (approved):

PR #485, opened 2024-08-22
MGMT-18170: Accept enrollment certificates from the CSR API

mangelajo:
During enrollment it should be possible to accept certificates created for such purpose through our CSR API.

### DIFF
--- a/internal/service/agent/handler.go
+++ b/internal/service/agent/handler.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"errors"
+	"strings"
 
 	"github.com/flightctl/flightctl/internal/api/server"
 	agentServer "github.com/flightctl/flightctl/internal/api/server/agent"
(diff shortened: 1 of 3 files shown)


lkatalin:
Thank you @mangelajo looks great!

Is there also a check somewhere (not sure where to look for this) that makes sure the enrollment cert was signed by the expected CA? This may already exist.

It would also make sense to check that the UUID/name is expected (in the case of the prefix + UUID). Is that being handled somewhere currently?

mangelajo:
> Thank you @mangelajo looks great!
> 
> Is there also a check somewhere (not sure where to look for this) that makes sure the enrollment cert was signed by the expected CA? This may already exist.
> 
> It would also make sense to check that the UUID/name is expected (in the case of the prefix + UUID). Is that being handled somewhere currently?

Yes, this is checked before this. We have a strict requirement on mTLS authentication for the agent. If the Cert is not trusted based on the CA it will not handshake TLS.

mangelajo:
> > Thank you @mangelajo looks great!
> > Is there also a check somewhere (not sure where to look for this) that makes sure the enrollment cert was signed by the expected CA? This may already exist.
> > It would also make sense to check that the UUID/name is expected (in the case of the prefix + UUID). Is that being handled somewhere currently?
> 
> Yes, this is checked before this. We have a strict requirement on mTLS authentication for the agent. If the Cert is not trusted based on the CA it will not handshake TLS.

https://github.com/flightctl/flightctl/blob/main/internal/crypto/tls.go#L37

That already applies to toe client-enrollment certs and would apply to the new ones as well.

internal/service/agent/handler.go
```diff
@@ -44,7 +45,8 @@ func ValidateEnrollmentAccessFromContext(ctx context.Context, log logrus.FieldLo
 	if !ok {
 		return errors.New("no common name in certificate")
 	}
-	if cn != crypto.ClientBootstrapCommonName {
+	if cn != crypto.ClientBootstrapCommonName &&
+		!strings.HasPrefix(cn, crypto.ClientBootstrapCommonNamePrefix) {
 		log.Warningf("an attempt to perform enrollment with a certificate with CN %q has been detected", cn)
```
hexfusion, line 50: should this be a log.Error? just looks a little strange that we are returning an error and warning log.

mangelajo:
@hexfusion updated to Errorf

lkatalin (approved):

PR #487, opened 2024-08-22
EDM-365: Report hooks errors to device summary

ori-amizur:


internal/agent/device/status/config_hooks.go
```diff
@@ -25,7 +26,7 @@ func newHooks(log *log.PrefixLogger, manager hook.Manager) *Hooks {
 
 // Export returns the status of the config hooks.
 func (s *Hooks) Export(ctx context.Context, status *v1alpha1.DeviceStatus) error {
-	return nil
+	return errors.Join(s.manager.Errors()...)
```
hexfusion, line 29: question: since this is general status do we need to prefix these errors with any context or will the errors have enough details for a user to understand already?

ori-amizur: Will modify to make sure

internal/agent/device/status/config_hooks.go
```diff
@@ -25,7 +27,11 @@ func newHooks(log *log.PrefixLogger, manager hook.Manager) *Hooks {
 
 // Export returns the status of the config hooks.
 func (s *Hooks) Export(ctx context.Context, status *v1alpha1.DeviceStatus) error {
-	return nil
+	err := errors.Join(s.manager.Errors()...)
+	if err != nil {
+		err = fmt.Errorf("hook manager: %v", err)
+	}
+	return err
```
hexfusion, line 34:
```suggestion
        if err := errors.Join(s.manager.Errors()...); err != nil {
	      return fmt.Errorf("hook manager: %v", err)
	}
	return nil
```

ori-amizur: OK

hexfusion (approved):

PR #489, opened 2024-08-22
EDM-351: Aggregate device status when listing devices

ori-amizur:
Add summary to device list which contains a map of statuses to their respective count, according to the current filter.


hexfusion (commented):
thanks @ori-amizur, please add client usage and response to the PR description then LGTM!

api/v1alpha1/openapi.yaml
```diff
@@ -2731,6 +2731,14 @@ components:
               description: 'The directory in which the executable will be run from if it is left empty it will run from the users home directory.'
           required:
             - run
+    DeviceSummary:
```
avishayt, line 2734: Why not reuse the existing DevicesSummary?

ori-amizur: OK

avishayt (approved):

PR #491, opened 2024-08-23
EDM-370: Bug fix for 500 error when approving certificate

lkatalin:
This PR fixes several bugs, most notably the 500 error that was appearing during certificate approval as well as the (somewhat related) bug of the certificate not appearing in the CSR after approval seems to succeed. The biggest change is that I moved "updating the Condition to Approved" to the very end of this process so that there will not be a scenario where the CSR shows as Approved but actually lacks a certificate, and consolidated where in the code the CSR resource is being updated.

There are some TODO comments inline for changes I wanted to draw attention to / validate - I'll remove these once we agree the changes make sense. For example, both updating the Condition and updating the Certificate are updating the Resource Version as well. Does this make sense? Currently approval + cert issuance is all one operation, however in the future as we have more complex policies around approval we may have several approvals that update the resource version before a certificate is issued. So perhaps the way it currently is does make sense.


lkatalin:
I think the e2e failure is unrelated to the PR as it seems to take place before the e2e tests are run.

tsorya:
I think this requires unitests

internal/store/certificatesigningrequest.go
```diff
@@ -267,3 +268,38 @@ func (s *CertificateSigningRequestStore) UpdateConditions(ctx context.Context, o
...
+		existingRecord.Status = model.MakeJSONField(api.CertificateSigningRequestStatus{})
+	}
+	existingRecord.Status.Data.Certificate = &certificate
+
+	// TODO - this currently updates the resource version *in addition to* the update that occurs when the resource's Condition
+	// is set to approved. these are technically two distinct operations but in reality will generally bundled together. should we
+	// only update the resource version once between them?
+	result = s.db.Model(existingRecord).Where("resource_version = ?", lo.FromPtr(existingRecord.ResourceVersion)).Updates(map[string]interface{}{
+		"status":           existingRecord.Status,
+		"resource_version": gorm.Expr("resource_version + 1"),
```
tsorya, line 289: I wonder about resource version, generally i believe status should not update resource version so why it should be done here?

internal/service/certificatesigningrequest.go
```diff
@@ -272,33 +271,45 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
...
 			return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
 		}
+		if errors.Is(err, flterrors.ErrResourceVersionConflict) {
+			return server.ApproveCertificateSigningRequest409JSONResponse{}, nil
+		}
 		return nil, err
 	}
 
-	err = signApprovedCertificateSigningRequest(h.ca, csr)
+	err = h.store.CertificateSigningRequest().UpdateCertificate(ctx, orgId, request.Name, signedCert)
```
tsorya, line 293: Should not from this point everything be part of transaction? I mean in case https://github.com/flightctl/flightctl/pull/491/files#diff-d5e3988a9944d806d028372b1fc8f031511b4f6495c386eebf33bca63f003e6eR309 will fail should not all the updates done previously be reverted?

internal/store/certificatesigningrequest.go
tsorya: Sorry actually my fault here, it should be updated though this is more k8s functionality for etcd and not sure why it is needed here

lkatalin: I am following the example of enrollmentrequests's [`UpdateStatus`](https://github.com/flightctl/flightctl/blob/main/internal/store/enrollmentrequest.go#L204).

internal/service/certificatesigningrequest.go
lkatalin: Updating the certificate here is what make the operation meaningful - and if this fails, we return an error and don't set condition to "approved". If the certificate has been added but then setting the condition fails (below), we could remove the certificate before returning an error - I can make this update if it seems better.

In the future we will collect _n_ number of approvals (_n_ being set by policy) before issuing the cert, so these will get separated out a bit more than they are here.

lkatalin:
> I think this requires unitests

Agree and they are one of the next things on the todo list - though it seems best to merge this first so that QE can do more testing sooner to find any other problems.

internal/service/certificatesigningrequest.go
```diff
@@ -19,26 +19,27 @@ import (
...
-	if api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestDenied) {
-		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request has been denied and cannot be signed")
-	}
-	if !api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestApproved) {
-		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is not approved and cannot be signed")
+	// TODO - we don't check for Approved status here anymore as it makes more sense to set Approved only after a certiicate
+	// is updated - otherwise a CSR may be marked as Approved but not actually make it through the entire Approval process (ex.
+	// certificate may be missing). this assumes the checks for "should be approved/signed" happen in the calling function
+	if request.Status.Conditions != nil && len(request.Status.Conditions) > 0 {
+		if api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestDenied) {
```
ori-amizur, line 31: If the length of conditions is greater than 0 - does it mean there should always be a `CertificateSigningRequestDenied` condition ?

internal/service/certificatesigningrequest.go
```diff
@@ -19,26 +19,27 @@ import (
...
-	// NOTE: this currently only requires one approval; this will change in the future based on policy
-	if api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestDenied) {
-		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request has been denied and cannot be signed")
-	}
-	if !api.IsStatusConditionTrue(request.Status.Conditions, api.CertificateSigningRequestApproved) {
-		return fmt.Errorf("signApprovedCertificateSigningRequest: certificate signing request is not approved and cannot be signed")
+	// TODO - we don't check for Approved status here anymore as it makes more sense to set Approved only after a certiicate
+	// is updated - otherwise a CSR may be marked as Approved but not actually make it through the entire Approval process (ex.
+	// certificate may be missing). this assumes the checks for "should be approved/signed" happen in the calling function
+	if request.Status.Conditions != nil && len(request.Status.Conditions) > 0 {
```
avishayt, line 30: 1. ApproveCertificateSigningRequest() already checks if the CSR was denied - it looks like we're checking it twice
2. I don't understand why we shouldn't check if the CSR was already approved?

internal/service/certificatesigningrequest.go
```diff
@@ -272,33 +271,45 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
...
-		Reason:  "Approved",
-		Message: "Approved",
-	}
-	err = h.store.CertificateSigningRequest().UpdateConditions(ctx, orgId, request.Name, []api.Condition{approvedCondition})
+	signedCert, err := signApprovedCertificateSigningRequest(h.ca, csr)
 	if err != nil {
 		if errors.Is(err, flterrors.ErrResourceNotFound) {
 			return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
 		}
+		if errors.Is(err, flterrors.ErrResourceVersionConflict) {
```
avishayt, line 287: Where is this returned in signApprovedCertificateSigningRequest()?

internal/service/certificatesigningrequest.go
avishayt: This check isn't needed - request.Status.Conditions can't be nil, and we don't need to check its length.  BTW, we should probably set both the approved and denied conditions to false when creating the CSR, right?

lkatalin: I think it could also be another approval, or something else - we aren't really constraining it AFAIA

internal/service/certificatesigningrequest.go
lkatalin: 1. Yes, that's true.
2. In this version approval is not set until after we know we have a certificate, so it comes later in the logic. We could do this differently, although that would require more time.

In short it may make sense to remove both checks

internal/service/certificatesigningrequest.go
lkatalin: Hm, I think this may be outdated. The error above it also may be outdated. I can remove these.

internal/service/certificatesigningrequest.go
lkatalin: Failing to check the len was resulting in errors

lkatalin: But I think we are removing these checks anyway

lkatalin: Yeah what Avishay said ^

> BTW, we should probably set both the approved and denied conditions to false when creating the CSR, right?
I think this is happening but I will double check

lkatalin: Creation of the CSR sets empty Conditions: https://github.com/flightctl/flightctl/blob/main/internal/store/model/certificatesigningrequest.go#L41 - is that good enough, or do they need to explicitly exist and be marked as "false"?

internal/service/certificatesigningrequest.go
lkatalin: I've changed it to remove the certificate that was previously set, in the case that approval fails.

tsorya: I believe that UpdateCertificate and UpdateConditions should be atomic db operation, right? 
I mean we should not create and than delete but we should rather do both in the same db call.

internal/service/certificatesigningrequest.go
avishayt: Explicitly marked as "false" - otherwise the default is "unknown"

internal/service/certificatesigningrequest.go
avishayt: It seems like I'm missing something.  If someone approves, and then approves again, is that not an error?

internal/service/certificatesigningrequest.go
lkatalin: Currently it is (or should be) atomic to be correct, but they will be separate later.
I can make UpdateCertificate and UpdateConditions into one function and have a todo to separate them out later.

avishayt: UpdateCertificate can just set "approved" to "true".

internal/service/certificatesigningrequest.go
lkatalin: Okay, can do in an upcoming PR

internal/store/certificatesigningrequest.go
```diff
@@ -267,3 +268,35 @@ func (s *CertificateSigningRequestStore) UpdateConditions(ctx context.Context, o
...
+	if result.Error != nil {
+		return false, flterrors.ErrorFromGormError(result.Error)
+	}
+
+	if existingRecord.Status == nil {
+		existingRecord.Status = model.MakeJSONField(api.CertificateSigningRequestStatus{})
+	}
+	existingRecord.Status.Data.Certificate = &certificate
+
+	result = s.db.Model(existingRecord).Where("resource_version = ?", lo.FromPtr(existingRecord.ResourceVersion)).Updates(map[string]interface{}{
```
tsorya, line 284: I wonder if we can create update() function that will get object and push it to db while validating resource version cause it seems like same code is duplicate over different functions like here 
https://github.com/flightctl/flightctl/pull/491/files#diff-ee6d72cabf35aed3a2691f1505a707f49a38e752478c98a66d1dfc5ec6b16e51R252

internal/service/certificatesigningrequest.go
lkatalin: You are right, we don't want to approve twice as this would overwrite the old cert. I'm putting a check in the POST approval ServiceHandler.

internal/store/certificatesigningrequest.go
lkatalin: Are you suggesting to update the Resource Version only once between Approval and setting the Certificate? That would be an easy change

lkatalin: Oh I think you are talking about abstracting over the various updates. I agree that would be a good change, likely out of scope for this PR though

tsorya: Nope, i suggest to have one function that will run 
`s.db.Model(existingRecord).Where("resource_version = ?", lo.FromPtr(existingRecord.ResourceVersion)).Updates(map[string]interface{}{...`
as seems like there is code duplication for the same logic between different functions

lkatalin:
I am attempting now to use the [`Update()`](https://github.com/flightctl/flightctl/blob/main/internal/store/certificatesigningrequest.go#L22) function such that we only update db resources once.

However, when I do this, the updates don't actually go through. (Although the resource version is updated.)

```
// where csr is an *api.CertificateSigningRequest

csr.Status.Certificate = &signedCert
api.SetStatusCondition(&csr.Status.Conditions, approvedCondition)

_, err = h.store.CertificateSigningRequest().Update(ctx, orgId, csr)
```

internal/store/certificatesigningrequest.go
```diff
@@ -267,3 +268,48 @@ func (s *CertificateSigningRequestStore) UpdateConditions(ctx context.Context, o
...
+	existingRecord.Status.Data.Certificate = &certificate
+
+	approvedCondition := api.Condition{
+		Type:    api.CertificateSigningRequestApproved,
+		Status:  api.ConditionStatusTrue,
+		Reason:  "Approved",
+		Message: "Approved",
+	}
+
+	err := s.UpdateConditions(ctx, orgId, name, []api.Condition{approvedCondition})
```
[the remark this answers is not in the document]
lkatalin: I'm removing updateCertificate and just trying to use Update

internal/store/certificatesigningrequest.go
```diff
@@ -267,3 +268,48 @@ func (s *CertificateSigningRequestStore) UpdateConditions(ctx context.Context, o
...
+	if err != nil {
+		return false, err
+	}
```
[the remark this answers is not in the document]
lkatalin: I'm removing updateCertificate and just trying to use Update

lkatalin:
Changed it to use the `Update()` function! All of the trouble was caused by [this one line](https://github.com/flightctl/flightctl/pull/491/files#diff-ee6d72cabf35aed3a2691f1505a707f49a38e752478c98a66d1dfc5ec6b16e51L167).

internal/store/certificatesigningrequest.go
```diff
@@ -164,7 +164,6 @@ func (s *CertificateSigningRequestStore) createOrUpdate(orgId uuid.UUID, resourc
 		return nil, false, false, err
 	}
 	certificatesigningrequest.OrgID = orgId
-	certificatesigningrequest.Status = nil
```
tsorya, line 167: I must say i have no idea why it was here but i hope it doesn't break other things.

lkatalin: This function is based on the enrollmentrequest version of the same functionality, and those requests must use Status in a different way (not sure why they'd want to set it to nil there). Nothing is relying on this code yet for the CSR functionality so I don't think that's an issue.

internal/service/certificatesigningrequest.go
```diff
@@ -268,42 +258,48 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, r
...
+	csr.Status.Conditions = append(csr.Status.Conditions, approvedCondition)
+	_, err = storeCsr.Update(ctx, orgId, csr)
+
+	switch err {
+	case nil:
+		return server.ApproveCertificateSigningRequest200JSONResponse{}, nil
+	case flterrors.ErrResourceNotFound:
+		return server.ApproveCertificateSigningRequest404JSONResponse{Message: "Resource not found"}, nil
+	case flterrors.ErrNoRowsUpdated, flterrors.ErrResourceVersionConflict:
+		return server.ApproveCertificateSigningRequest409JSONResponse{Message: "Version conflict or no rows updated"}, nil
```
ori-amizur, line 299: Maybe you can add the text from the errors themselves? There are 2 possible errors here

internal/store/certificatesigningrequest.go
avishayt: Status was being set to nil here because Create, Update and CreateOrUpdate are user-facing, and the user shouldn't update Status.  That's why we have a separate UpdateStatus method.  I wouldn't mind adding 
a list of conditions as a parameter to UpdateStatus.

internal/service/certificatesigningrequest.go
lkatalin: Sure that's easy

internal/store/certificatesigningrequest.go
lkatalin: I can redo it, please hold

lkatalin: It *looks* like `UpdateStatus()` shouldn't need any more parameters as it already takes the whole CSR resource and uses its status

internal/service/certificatesigningrequest.go
```diff
@@ -268,42 +258,48 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, r
...
-		return nil, err
-	}
 
-	err = signApprovedCertificateSigningRequest(h.ca, csr)
-	if err != nil {
-		if errors.Is(err, flterrors.ErrResourceNotFound) {
-			return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
-		}
+	csr.Status.Certificate = &signedCert
+	csr.Status.Conditions = append(csr.Status.Conditions, approvedCondition)
```
avishayt, line 290:
```suggestion
	api.SetStatusCondition(&csr.Status.Conditions, approvedCondition)
```

lkatalin: It didn't like this last time, but some other code has changed so I'll try again

lkatalin: @avishayt fixed

avishayt (approved):

PR #492, opened 2024-08-26
EDM-206: Allow fetching devices with value-less labelSelector

avishayt:
Previously we modified the CLI to add empty values to the labelSelector, but this doesn't help the UI which uses the API directly. This commit moves the logic from the CLI to the service so that the behavior is consistent for the CLI and API.


celdrake (approved):
Issue is fixed, thanks!

PR #495, opened 2024-08-27
EDM-378: Implement mount path for GitConfigProviderSpec

tsorya:


celdrake:
@avishayt I discussed with Igal, that maybe it's best to add a default to this new field, at least until the PoC final branch is created. We're still fixing bugs in both the API and the UI (mainly for the Terminal), so having a new feature with breaking changes could be problematic.

The UI changes seem straightforward though.

avishayt:
> @avishayt I discussed with Igal, that maybe it's best to add a default to this new field, at least until the PoC final branch is created. We're still fixing bugs in both the API and the UI (mainly for the Terminal), so having a new feature with breaking changes could be problematic.
> 
> The UI changes seem straightforward though.

We can remove the default after the UI and demo repo are fixed.

avishayt:
Looks OK, need to fix lint and tests

internal/tasks/device_render.go
```diff
@@ -257,7 +258,7 @@ func renderGitConfig(ctx context.Context, configItem *api.DeviceSpec_Config_Item
 	}
 
 	// Create an ignition from the git subtree and merge it into the rendered config
-	ignitionConfig, err := ConvertFileSystemToIgnition(mfs, gitSpec.GitRef.Path)
+	ignitionConfig, err := ConvertFileSystemToIgnition(mfs, gitSpec.GitRef.Path, swag.StringValue(gitSpec.GitRef.MountPath))
```
ori-amizur, line 261: The package `lo` has already `ToPtr` and `FromPtr` functions

avishayt (approved):

PR #498, opened 2024-08-27
NO-ISSUE: Add status info to the application status

ori-amizur:
Status info gives additional information to containers


api/v1alpha1/openapi.yaml
```diff
@@ -2849,6 +2849,7 @@ components:
         - ready
         - restarts
         - status
+        - status_info
```
celdrake, line 2852: We're always using camelCase in the API, so this should be renamed.

```suggestion
        - statusInfo
```

hexfusion: I understand the idea of this change, but this breaks a general status pattern and is not described in the data dictionary[1]. Can you provide an example of what this status would include?

[1] https://docs.google.com/spreadsheets/d/1xlslENyYVWL-M2Qw4lOwHOEWMXXk9cPN-tCb-IH7t3c/edit?gid=21336438#gid=21336438

ori-amizur: In case a specific application (container for our case) fails - then we just mark it as error.  There is no place to put exit-code or failure reason.  The  status info should do just that - provide more information about why the application is in its current failure state.
If we have many applications, and some of them are failing, adding the failure reason to some summary field doesn't make sense.  It is better to provide the failure reason to the record representing the application itself.

hexfusion: I understand where you are coming from but the Data Dictionary was an agreed-upon contract between the device and service. So changing this would be an enhancement vs a bug. In general summary status is the same as a condition. So the pattern (for now) is any failures at the application level should be included in the summary status. We can propose your change when everyone gets back. It makes logical sense to have this granularity but we need to consider scalability as well.

ori-amizur: It does not make sense to have a status field in application - but if we need to get more info on it, we need to go device summary which is more high level.
Today as @siserafin mentioned (https://issues.redhat.com/browse/EDM-365?focusedId=25422326&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25422326) , the application status is not clear.

fzdarsky:
Let's hold off from making changes to the API until we've fully converged on where we want to end up.

The application status needs to both _aggregate_ the status of its component pods+volumes+networks as well as _be aggregate-able_ across a device (summary status: "all applications on this device are healthy") and fleet ("2 out of 10 recently updated devices reported an unhealthy application").

PR #499, opened 2024-08-27
EDM-385: Error 400 with wrong validation prevents being able to edit a fleet

ori-amizur:
The fields in the device summary are considered required.  Therefore they must appear when device summary is part of any entity sent in the REST API.
This commit removes these fields as required, so sending these fields as null values will not cause edit to fail.


api/v1alpha1/openapi.yaml
```diff
@@ -3391,10 +3391,6 @@ components:
     DevicesSummary:
       type: object
       description: A summary of the devices in the fleet returned when fetching a single Fleet.
-      required:
-      - total
-      - summaryStatus
-      - updateStatus
```
avishayt, line 3397: Why is this the correct fix?  If DevicesSummary appears, I should be able to assume that it's populated.  The bug shows a PATCH request, so the issue is probably in the returned fleet?

ori-amizur: First - the scenario that was described in the ticket was tested.
Second - If the properties are required, then the API prevents sending PATCH request with missing status since the device summary fields are required.

avishayt: But the PATCH request has a totally different format - an array of {path, value, op}.  There's no status, no spec, or anything else.  That's why I assume the error is on the return.

ori-amizur: Yes, but the error is from the library (not us).
From the ticket:
```
{
  "message": "request body has an error: doesn't match schema #/components/schemas/Fleet: Error at \"/status/devicesSummary/summaryStatus\": Value is not nullable\nSchema:\n  {\n    \"additionalProperties\": {\n      \"type\": \"integer\"\n    },\n    \"description\": \"A breakdown of the devices in the fleet by \\\"summary\\\" status.\",\n    \"type\": \"object\"\n  }\n\nValue:\n  null\n"
} 
```
This is not an error we generate and it complains specifically on these fields.

avishayt: Right, PATCH returns #/components/schemas/Fleet but accepts #/components/schemas/PatchRequest

ori-amizur: How else can the problem be solved ?

ori-amizur: If the problem is in return, why do list and get operations succeed ?

ori-amizur: The PATCH uses CreateOrUpdate which also used by PUT.  Why does the return from PUT succeeds ?

ori-amizur: I tested with a latest version and also added device-summary to the returned fleet.  The same problem still happens.
So it seems that it happens during the request.

ori-amizur: Actually the problem happens here:
https://github.com/flightctl/flightctl/blob/cdc91b98401868d39fa48c2bc4bbea83941f5ad8/internal/service/fleet.go#L235

avishayt: OK, I figured it out.  We usually don't validate what the API returns, so GET and PUT work fine.  But PATCH calls GET, applies the patch, and then validates.  So summaryStatus and updateStatus are indeed nil.  The change is OK then, but we should leave "total" required IMO.

I then tried to test with a fleet that has devices, but GET on the fleet returns this:
    devicesSummary:
      summaryStatus: null
      total: 1
      updateStatus: null

devicesSummary when I GET devices works.  Maybe another bug?

ori-amizur: Isn't this the behavior we decided?
Unless we request `WithSummary` both maps will be nil except the total

avishayt: Right.  So please just return total to be required and we're good to go.

avishayt (approved):

PR #501, opened 2024-08-28
EDM-387: Filter resourcesyncs by repository

avishayt:
This commit adds a filter allowing listing resourcesyncs by repository. This is needed by the UI.

The changes are:
1. Make the current status filter logic generic so that it can also filter specs.
2. Add the repository filter, internally implemented by the above filter logic.
3. Ensure that all specs and statuses are jasonb rather than bytea so that the filters work properly.

### DIFF
--- a/internal/service/common.go
+++ b/internal/service/common.go
@@ -89,8 +89,8 @@ func ApplyJSONPatch[T any](ctx context.Context, obj T, newObj T, patchRequest v1
 	return decoder.Decode(&newObj)
 }
 
-// ConvertStatusFilterParamsToMap converts statusFilter query params to to a validated filterMap map.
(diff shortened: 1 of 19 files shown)


internal/service/common.go
```diff
@@ -136,7 +136,7 @@ func validateFieldKey(key string) (string, error) {
 func validateFieldValue(value string) (string, error) {
 	value = strings.TrimSpace(value)
 	for _, char := range value {
-		if !unicode.IsLetter(char) && !unicode.IsDigit(char) && char != ',' {
+		if !unicode.IsLetter(char) && !unicode.IsDigit(char) && char != ',' && char != '-' && char != '.' {
```
ori-amizur, line 139: Maybe update the comment above?

ori-amizur (approved):

PR #503, opened 2024-08-28
MGMT-18170: Document usage of CSR API for getting started

lkatalin:
Some details may need to change around creating the `config.yaml` for the Containerfile, as I am not currently sure how/whether this happens. See [TODO comment](https://github.com/flightctl/flightctl/pull/503/files#diff-816862240c6f5d4959307b5248230cc3c4cb3df5895b2309c4d8883619cec7d4R323)


lkatalin:
I'm going to hold off on updating this until we have a way to get this `client.yaml` file with https://github.com/flightctl/flightctl/pull/504

docs/user/getting-started.md
```diff
@@ -222,29 +222,116 @@ Browse to `ui.flightctl.MY.DOMAIN` and login with the demouser obtained from the
...
+```console
+$ cat > mycsrresource.yaml <<EOF
+apiVersion: v1alpha1
+kind: CertificateSigningRequest
+metadata:
+  name: mycsr
+spec:
+  request: <add base64-encoded CSR>                                                                                                                                                                                                             signerName: ca
+  usages: ["clientAuth", "CA:false"]
+  expirationSeconds: 600
```
rawagner, line 261: as pointed by @celdrake , this 600 seconds is very low in order to get through all the steps to start device/VM. Maybe in this example (which most will just copy-paste), lets rise the value ?

lkatalin: Good idea, updated

docs/user/getting-started.md
```diff
@@ -259,7 +346,7 @@ RUN dnf -y copr enable @redhat-et/flightctl-dev centos-stream-9-x86_64 && \
     dnf -y clean all; \
     systemctl enable flightctl-agent.service
 
-ADD config.yaml /etc/flightctl/
+ADD agentconfig.yaml /etc/flightctl/
```
celdrake, line 349: I believe the agent is still looking for `config.yaml`, I did something similar myself and the agent failed to find the file.

```suggestion
ADD agentconfig.yaml /etc/flightctl/config.yaml
```

docs/user/getting-started.md
celdrake: Could you also please update the timeout in the `examples/csr.yaml` file?

I think it's convenient for using it as a template to create our own CSRs.

docs/user/getting-started.md
lkatalin: Thank you @celdrake !

docs/user/getting-started.md
lkatalin: Done

lkatalin:
I'm hoping to create a command that can do steps 1 and 2 (and maybe 3) for the user. So if I succeed, I will update the docs in a subsequent PR.

lkatalin:
I need to fix a formatting error

lkatalin:
What do others think - should we remove the manual CSR creation option since it can be automated?

hexfusion:
> What do others think - should we remove the manual CSR creation option since it can be automated?

if manual CSR creation is a small use case that can be solved via documention it might be a better option than maintaining CLI support. High level this is how I would consider the choice.

lkatalin:
> if manual CSR creation is a small use case that can be solved via documention it might be a better option than maintaining CLI support. High level this is how I would consider the choice

I don't think it's a small use case - IIUC it's the main way that users will create this resource config. That being said, I've been surprised before.

.spelling
```diff
@@ -55,9 +58,11 @@ OIDC
 onboard
 onboarding
 OpenShift
+openssl
 OSes
 ostree
 OTA
+outfile
```
avishayt, line 65: Not a word :)

docs/user/getting-started.md
```diff
@@ -222,29 +222,137 @@ Browse to `ui.flightctl.MY.DOMAIN` and login with the demouser obtained from the
...
+
+For more options, including generating a new private key to sign the CSR, see [openssl's documentation](https://docs.openssl.org/master/man1/openssl-req/#options).
+
+#### 2/4 Create the CSR resource configuration file
+
+You can create a CSR resource configuration file wrapping the CSR file above either by generating it with flightctl or manually. Each option is described below. You may choose the CSR name, and desired expiration in seconds.
+
+##### Option 1: Generate the CSR config file
+
+Issue the command below, specifying your `.csr` file and optionally specifying the CSR name, expiration in seconds, and an outfile. The CSR name defaults to `mycsr`, the expiration to 604800 seconds, and the outfile to `mygeneratedcsr.yaml`.
```
avishayt, line 254:
```suggestion
Issue the command below, specifying your `.csr` file and optionally specifying the CSR name, expiration in seconds, and an outfile. The CSR name defaults to `mycsr`, the expiration to 604800 seconds, and the output file to `mygeneratedcsr.yaml`.
```

avishayt: I don't think this example with default values is necessary.  Just show a typical example and let them refer to --help for more information.

avishayt (approved):

PR #504, opened 2024-08-29
EDM-400: Add Enrollment config endpoint

rawagner:
After creating and approving CSR, you can use this new command to generate an enrollment config
 
Run `flightctl enrollmentconfig mycsr -p myeckey.pem` to get the enrollment-config file


lkatalin:
Thank you @rawagner 

I think this may need to grow a `/{name}` field so that the user can get it for a specific device - like we have for the CSRs - since this will basically be a "last step" on top of the CSR steps. So it would match https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/openapi.yaml#L1468 . Though this may imply some shared namespace  or naming convention between this API and the CSR API.

internal/service/enrollmentconfig.go
```diff
@@ -0,0 +1,34 @@
...
+// (GET /api/v1/enrollmentconfig)
+func (h *ServiceHandler) EnrollmentConfig(ctx context.Context, request server.EnrollmentConfigRequestObject) (server.EnrollmentConfigResponseObject, error) {
+	cert, _, err := h.ca.Config.GetPEMBytes()
+	if err != nil {
+		return server.EnrollmentConfig400JSONResponse{Message: err.Error()}, nil
+	}
+	return server.EnrollmentConfig200JSONResponse{
+		EnrollmentService: v1alpha1.EnrollmentService{
+			Authentication: v1alpha1.EnrollmentServiceAuth{
+				ClientCertificateData: "",
```
lkatalin, line 20: something like
```
ClientCertificateData: h.store.CertificateSigningRequest().Get(ctx, store.NullOrgId, request.Name).ToApiResource().Status.Certificate,
```

lkatalin:
Also would change NO-ISSUE to MGMT-18170

lkatalin:
This version is working as expected for me with the CLI command. :rocket:

rawagner:
thanks @lkatalin . I've updated the code as you suggested. The EnrollmentConfig endpoint now expects a name of approved CSR.

lkatalin (approved):

lkatalin (approved):

PR #508, opened 2024-08-30
MGMT-18170: Fix error handling

lkatalin:
Make 500 error verbose with an error message, and add a ~422~ 400 error for instances where the CSR is not processable


internal/service/certificatesigningrequest.go
```diff
@@ -274,9 +274,17 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
...
 	if err != nil {
-		return nil, err
+		switch err {
+		case flterrors.ErrCNLength:
+			msg := fmt.Sprintf("cn: %s does not meet requirement: %s", errctx, err.Error())
+			return server.ApproveCertificateSigningRequest422JSONResponse{Message: msg}, nil
+		case flterrors.ErrInvalidPEMBlock:
+			return server.ApproveCertificateSigningRequest422JSONResponse{Message: err.Error()}, nil
+		default:
+			return server.ApproveCertificateSigningRequest500JSONResponse{Message: err.Error()}, nil
```
rawagner, line 286: lets use `errors.Is`
```suggestion
		switch {
		case errors.Is(err, flterrors.ErrCNLength) || errors.Is(err, flterrors.ErrInvalidPEMBlock):
			return server.ApproveCertificateSigningRequest422JSONResponse{Message: err.Error()}, nil
		default:
			return server.ApproveCertificateSigningRequest500JSONResponse{Message: err.Error()}, nil
		}
```

internal/service/certificatesigningrequest.go
```diff
@@ -19,18 +19,18 @@ import (
 
 const DefaultEnrollmentCertExpirySeconds int32 = 60 * 60 * 24 * 7 // 7 days
 
-func signApprovedCertificateSigningRequest(ca *crypto.CA, request *api.CertificateSigningRequest) ([]byte, error) {
+func signApprovedCertificateSigningRequest(ca *crypto.CA, request *api.CertificateSigningRequest) ([]byte, error, string) {
```
rawagner, line 22:
```suggestion
func signApprovedCertificateSigningRequest(ca *crypto.CA, request *api.CertificateSigningRequest) ([]byte, error) {
```

lets add the msg and wrap it using `fmt.Errorf`. It will make the code cleaner

lkatalin:
Thanks for the suggestions @rawagner , I will see if I can extrapolate from those to return a few more specific ones as well

lkatalin:
These seem to work!

api/v1alpha1/openapi.yaml
```diff
@@ -1712,6 +1712,18 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+        "422":
```
avishayt, line 1715: The convention for this project is to use 400 instead of 422

internal/service/certificatesigningrequest.go
```diff
@@ -264,19 +261,33 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
...
+	signedCert, err := signApprovedCertificateSigningRequest(h.ca, *csr)
 	if err != nil {
-		return nil, err
+		switch {
+		case errors.Is(err, flterrors.ErrCNLength) ||
+			errors.Is(err, flterrors.ErrInvalidPEMBlock) ||
+			errors.Is(err, flterrors.ErrSignature) ||
+			errors.Is(err, flterrors.ErrCSRParse) ||
+			errors.Is(err, flterrors.ErrSignCert) ||
+			errors.Is(err, flterrors.ErrEncodeCert):
```
avishayt, line 286: I think the golang style is to use commas:

```suggestion
		case errors.Is(err, flterrors.ErrCNLength), errors.Is(err, flterrors.ErrInvalidPEMBlock),  errors.Is(err, flterrors.ErrSignature), errors.Is(err, flterrors.ErrCSRParse), errors.Is(err, flterrors.ErrSignCert), errors.Is(err, flterrors.ErrEncodeCert):
```

avishayt (commented):
Nice PR! A couple small comments

internal/service/certificatesigningrequest.go
```diff
@@ -264,19 +261,33 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
...
 	if api.IsStatusConditionTrue(csr.Status.Conditions, api.CertificateSigningRequestApproved) {
 		return server.ApproveCertificateSigningRequest409JSONResponse{Message: "The request has already been approved"}, nil
 	}
 
-	signedCert, err := signApprovedCertificateSigningRequest(h.ca, csr)
+	signedCert, err := signApprovedCertificateSigningRequest(h.ca, *csr)
 	if err != nil {
-		return nil, err
+		switch {
+		case errors.Is(err, flterrors.ErrCNLength) ||
```
tsorya, line 281: I wonder if not better to create SignatureError and return it? and not to verify list of errors?

internal/crypto/cert.go
```diff
@@ -24,7 +25,7 @@ const DeviceCommonNamePrefix = "device:"
 
 func BootrapCNFromName(name string) (string, error) {
```
ori-amizur, line 26:
```suggestion
func BootstrapCNFromName(name string) (string, error) {
```

internal/crypto/csr.go
```diff
@@ -37,9 +38,9 @@ func ParseCSR(csrPEM []byte) (*x509.CertificateRequest, error) {
 	block, rest := pem.Decode(csrPEM)
 	switch {
 	case block == nil:
```
ori-amizur, line 40:
```suggestion
	if block == nil || len(bytes.TrimSpace(rest)) > 0 {
	       return nil, flterrors.ErrInvalidPEMBlock 
	}
```

internal/service/certificatesigningrequest.go
ori-amizur: I am not sure this form is relevant here, since comma separated case values are for values - not for conditions

lkatalin: So should I leave this as-is?

internal/service/certificatesigningrequest.go
lkatalin: The problem I encountered was that there can be many types of errors that occur during `signApprovedCertificateSigningRequest`, and generalizing over the 400 or 422 errors doesn't give as much information to the user as they sometimes need to print different specifics.

That being said - in a near-future PR I could make an error category for all of these so that this reads a little cleaner.

lkatalin:
Made the suggested changes

avishayt (approved):

PR #513, opened 2024-08-30
MGMT-18170: Add CLI command to generate CSR resource config

lkatalin:
This will help users avoid manually having to create the `.yaml` file from their openssl-generated `.csr` file.

### DIFF
--- a/internal/cli/csrconfig.go
+++ b/internal/cli/csrconfig.go
@@ -1,0 +1,123 @@
+package cli
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+type CSRConfigOptions struct {
+	GlobalOptions
+	CertSigningRequestFile string
+	Output                 string
+	Name                   string
+	ExpirationSeconds      string
+	Overwrite              bool
+}
+
+func DefaultCSRConfigOptions() *CSRConfigOptions {
+	return &CSRConfigOptions{
+		GlobalOptions: DefaultGlobalOptions(),
+		Overwrite:     false,
+	}
+}
+
+func NewCmdCSRConfig() *cobra.Command {
+	o := DefaultCSRConfigOptions()
+	cmd := &cobra.Command{
+		Use:   "csr-generate",
+		Short: "generate a CSR resource config .yaml based on a CSR file .csr",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := o.Complete(cmd, args); err != nil {
+				return err
+			}
+			if err := o.Validate(args); err != nil {
+				return err
+			}
+			return o.Run(cmd.Context(), args)
+		},
+		SilenceUsage: true,
+	}
+	o.Bind(cmd.Flags())
+	_ = cmd.MarkFlagRequired("output")
+	return cmd
+}
+
+func (o *CSRConfigOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Output, "output", "o", "", "specify a filename in which to save the generated csr resource config")
+	fs.StringVarP(&o.Name, "name", "n", "mycsr", "specify a name for the csr")
+	fs.StringVarP(&o.ExpirationSeconds, "expiry", "e", "604800", "specify desired certificate expiration in seconds")
+	fs.BoolVarP(&o.Overwrite, "overwrite", "y", false, "setting this flag overwrites the specified output file without prompting the user")
+}
+
+func (o *CSRConfigOptions) Complete(cmd *cobra.Command, args []string) error {
+	return o.GlobalOptions.Complete(cmd, args)
+}
+
+func (o *CSRConfigOptions) Validate(args []string) error {
+	if err := o.GlobalOptions.Validate(args); err != nil {
+		return err
+	}
+
+	if len(args) == 0 {
+		return fmt.Errorf("must specify a CSR file")
+	}
+	_, err := os.ReadFile(args[0])
+	if err != nil {
+		return fmt.Errorf("invalid CSR file specified: %w", err)
+	}
+
+	return nil
+}
+
+func (o *CSRConfigOptions) Run(ctx context.Context, args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("must specify a CSR file")
+	}
+
+	data, err := os.ReadFile(args[0])
+	if err != nil {
+		return fmt.Errorf("invalid CSR file specified: %w", err)
+	}
+
+	encoded := base64.StdEncoding.EncodeToString(data)
+
+	output := fmt.Sprintf(`apiVersion: v1alpha1
+kind: CertificateSigningRequest
+metadata:
+  name: %s
+spec:
+  request: %s
+  signerName: ca
+  usages: ["clientAuth", "CA:false"]
+  expirationSeconds: %s
+`, o.Name, encoded, o.ExpirationSeconds)
+
+	if !o.Overwrite {
+		_, err := os.Stat(o.Output)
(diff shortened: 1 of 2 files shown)


hexfusion (changes requested):
Mostly cosmetic suggestions but we have a few potential panics we need to cover

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	fmt.Printf("config file written to: %s\n", filename)
+
+	return nil
+}
+
+func createFilename(filename string) string {
+	fileroot := filename
+	if filename[len(filename)-5:] == ".yaml" {
```
tsorya, line 122: if filepath.Ext(path) != ".yaml"

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,135 @@
...
+	CertSigningRequestFile string
+	Outfile                string
+	Name                   string
+	ExpirationSeconds      string
+}
+
+func DefaultCSRConfigOptions() *CSRConfigOptions {
+	return &CSRConfigOptions{
+		GlobalOptions:          DefaultGlobalOptions(),
+		CertSigningRequestFile: "",
```
ori-amizur, line 25: No need to add empty values.

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+type CSRConfigOptions struct {
+	GlobalOptions
+	CertSigningRequestFile string
+	Outfile                string
+	Name                   string
+	ExpirationSeconds      string
```
ori-amizur, line 19: Why not integer?

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func createFilename(filename string) string {
+	fileroot := filename
+	if filename[len(filename)-5:] == ".yaml" {
+		fileroot = filename[:len(filename)-5]
+	}
+
+	// if file already exists, don't overwrite it; add suffix
+	if _, err := os.Stat(filename); err == nil {
```
ori-amizur, line 127: Check error type?

internal/cli/csrconfig.go
lkatalin: This processes all the data into strings for the .yaml file - it is never actually treated as an integer here

internal/cli/csrconfig.go
lkatalin: My understanding is that in this case any error will indicate that the file does not exist, so we can safety write a file with that name. But let me know if this is incorrect.

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,135 @@
...
+		u := uuid.NewString()
+		filename = fmt.Sprintf(fileroot + "-" + u[:7] + ".yaml")
+	} else {
+		filename = fmt.Sprintf(fileroot + ".yaml")
+	}
+
+	return filename
```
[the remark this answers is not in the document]
lkatalin: The fmt.Sprintf formatting here is returning a filename that looks like `'somefilename-%!d(string=7ecaace).yaml'` so I'm using `return fileroot + "-" + u[:7] + ".yaml"`. The other suggestions work great

lkatalin:
Excellent suggestions, thanks @hexfusion @ori-amizur @tsorya .

I've made updates.

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,132 @@
...
+		GlobalOptions: DefaultGlobalOptions(),
+	}
+}
+
+func NewCmdCSRConfig() *cobra.Command {
+	o := DefaultCSRConfigOptions()
+	cmd := &cobra.Command{
+		Use:   "csr-generate",
+		Short: "generate a CSR resource config .yaml based on a CSR file .csr",
+		Args:  cobra.ExactArgs(1),
```
hexfusion, line 34: Looking deeper ExactArgs(1) will return an error if no args are passed.

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return nil
+}
+
+func (o *CSRConfigOptions) Validate(args []string) error {
+	if err := o.GlobalOptions.Validate(args); err != nil {
+		return err
+	}
+
+	_, err := os.ReadFile(args[0])
```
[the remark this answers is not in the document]
hexfusion: Looking deeper ExactArgs(1) will return an error if no args are passed. But logically its good practice.

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,132 @@
...
+spec:
+  request: `
+
+	output3 := `
+  signerName: ca
+  usages: ["clientAuth", "CA:false"]
+  expirationSeconds: `
+	filename := createFilename(o.Outfile)
+
+	f, err := os.Create(filename)
```
hexfusion, line 103: we need to close this file or it will leak a file descriptor.

defer f.Close()

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,132 @@
...
+metadata:
+  name: `
+	output2 := `
+spec:
+  request: `
+
+	output3 := `
+  signerName: ca
+  usages: ["clientAuth", "CA:false"]
+  expirationSeconds: `
```
hexfusion, line 100: optional: We can use Sprintf here to make this a bit easier to read?

```
	output := fmt.Sprintf(`apiVersion: v1alpha1
metadata:
  name: %s
spec:
  request: %s
  signerName: ca
  usages: ["clientAuth", "CA:false"]
  expirationSeconds: %s
`, o.Name, encoded, o.ExpirationSeconds)
```

internal/cli/csrconfig.go
lkatalin: good call thanks

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("config file written to: %s\n", filename)
+
+	return nil
+}
+
+func createFilename(filename string) string {
```
avishayt, line 118: I would be surprised if I specified a filename in a command and the command created a different file.  I would overwrite the file.  This is how most of the Linux command line tools I'm familiar with behave.

lkatalin:
WDYT of this version, @avishayt ?

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,126 @@
...
+		SilenceUsage: true,
+	}
+	o.Bind(cmd.Flags())
+	_ = cmd.MarkFlagRequired("output")
+	return cmd
+}
+
+func (o *CSRConfigOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Outfile, "output", "o", "", "specify a filename in which to save the generated csr resource config")
```
avishayt, line 54: Need to validate that this was passed?

internal/cli/csrconfig.go
```diff
@@ -0,0 +1,126 @@
...
+metadata:
+  name: %s
+spec:
+  request: %s
+  signerName: ca
+  usages: ["clientAuth", "CA:false"]
+  expirationSeconds: %s
+`, o.Name, encoded, o.ExpirationSeconds)
+
+	if !o.Overwrite {
```
avishayt, line 103: I'd rather not make the CLI interactive.  I'm OK with the overwrite flag for extra safety, but if the user didn't pass it, just return an error.

internal/cli/csrconfig.go
lkatalin: Apparently pflag validates it for you!

internal/cli/csrconfig.go
lkatalin: Done

avishayt (approved):

PR #516, opened 2024-09-01
NO-ISSUE:  Change template version name to be based on the fleet generation

ori-amizur:
This will avoid possible conflict when 2 template versions are created within the same second.


internal/tasks/fleet_validate.go
```diff
@@ -59,7 +61,7 @@ func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Co
 
 	templateVersion := api.TemplateVersion{
 		Metadata: api.ObjectMeta{
-			Name:  util.TimeStampStringPtr(),
+			Name:  lo.ToPtr(strconv.FormatInt(lo.FromPtr(fleet.Metadata.Generation), 10)),
```
hexfusion, line 64: as an alternative could we make TimeStampStringPtr more granular to avoid this condition?

hexfusion: I guess in retrospect this would still allow a potential conflict.

avishayt: Unfortunately, this won't work because new TemplateVersions can be created while the corresponding fleet resource remains unchanged.  For example, the fleet's template can contain a git reference, and we will create a new TemplateVersion each time something changes in that git reference.
Using a UUID would solve this issue, but it's not very user-friendly. Could the DB provide us with an auto-incrementing number per org+fleet?

ori-amizur: Why not this solution?

ori-amizur: More granular TimeStampStringPtr reduces the risk.  This should eliminate it.

hexfusion: Right it could still cause a conflict unless you added thread safety locking

avishayt: 1. User sets fleet template to point to gitrepo's main branch
2. Service sets fleet.metadata.generation to 5
3. Service creates a TemplateVersion named "5"
4. User pushes new commits to gitrepo's main branch
5. Service creates a TemplateVersion, again named "5" because the fleet resource didn't change

ori-amizur: OK, so the generation is not good here.
Actually, there is no way to know if the TV has to be created from the fleet data.  Therefore, we need to verify that TV can always be created.
Some auto generated value seems best.

ori-amizur: Changed @hexfusion suggestion - use nanosecond strin format

PR #520, opened 2024-09-03
EDM-373: Spec manager test coverage

DakCrowder:
Adds some unit tests for the following methods inside the spec manager:
- NewManager 
- Initialize
- Ensure

This will be the first of a few prs seeking to improve  test coverage for the spec manager.

### DIFF
--- a/internal/agent/device/spec/spec_test.go
+++ b/internal/agent/device/spec/spec_test.go
@@ -3,6 +3,8 @@ package spec
 import (
 	"context"
 	"encoding/json"
+	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/flightctl/flightctl/api/v1alpha1"
@@ -11,6 +13,7 @@ import (
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 func TestBootstrapCheckRollback(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/agent/device/spec/spec_test.go
```diff
@@ -93,6 +95,126 @@ func TestBootstrapCheckRollback(t *testing.T) {
...
+		deviceReadWriter: mockReadWriter,
+	}
+
+	writeErr := errors.New("write failure")
+
+	t.Run("error writing current file", func(t *testing.T) {
+		// current
+		mockReadWriter.EXPECT().WriteFile(gomock.Any(), gomock.Any(), gomock.Any()).Return(writeErr)
+		err := s.Initialize()
+		require.ErrorContains(err, "writing current rendered spec:")
```
hexfusion, line 141: This can be a followup but we should always strive to test for error types. So if we see a common error lets create a an error like

```go
var ErrWritingSpec = errors.New("writing current spec") 
```
Then we can wrap `%w` that error on the return and explicitly check for the type. This is much less brittle for the cases where the error message changes.

```go
return fmt.Errorf("%w: current: %w", ErrWritingSpec, err)
```
Then testing is 

```go
require.ErrorIs(err, ErrWritingSpec)
```

internal/agent/device/spec/spec_test.go
```diff
@@ -93,6 +95,126 @@ func TestBootstrapCheckRollback(t *testing.T) {
...
+	s := &SpecManager{
+		log:              log.NewPrefixLogger("test"),
+		deviceReadWriter: mockReadWriter,
+	}
+
+	t.Run("error checking if file exists", func(t *testing.T) {
+		errMsg := "unable to check if file exists"
+		mockReadWriter.EXPECT().FileExists(gomock.Any()).Return(false, errors.New(errMsg))
+		err := s.Ensure()
+		require.ErrorContains(err, errMsg)
```
hexfusion, line 189: same general principles here we can make this error a var and explicitly return/test it. Then as we begin to fill these in out testing will become stronger.

https://github.com/flightctl/flightctl/blob/73b1a2de0b9461862f8e034f2295b83c2391b16b/internal/agent/device/fileio/reader.go#L50

hexfusion (approved):
Congrats @DakCrowder a wonderful start!! I added a few comments you can consider for your next iterations as followups.

PR #525, opened 2024-09-04
EDM-426: Temporary autoapprove hack for CSRs

kot-begemot-uk:
Quick hack to auto-approve CSRs


hexfusion (changes requested):
In general we don't want these types of changes in the code base. If you want an auto approver why not write it in python or bash using the cli?

also unclear how does this related to https://issues.redhat.com/browse/EDM-426

kot-begemot-uk:
I will close and stash it. Concur - better if we do not use this approach.

PR #528, opened 2024-09-05
Spec manager test enhancements

DakCrowder:
Additional testing for the Spec Manager including:
- Some new error types / handling changes
- Unit tests covering additional currently untested methods

Test coverage % isn't the be all end all but this moves us to > 90% cov for the spec manager


internal/agent/device/spec/spec_test.go
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+
+		rollbackSpec, err := createTestSpec(currentImage)
+		require.NoError(err)
+		mockReadWriter.EXPECT().WriteFile(rollbackPath, rollbackSpec, gomock.Any()).Return(nil)
+
+		err = s.PrepareRollback(ctx)
+		require.NoError(err)
+	})
+
+	t.Run("writes the os image from bootc when the current spec os is nil", func(t *testing.T) {
```
hexfusion, line 616: nice test case

internal/agent/device/spec/spec_test.go
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+	emptySpec, err := createEmptyTestSpec()
+	require.NoError(err)
+
+	specErr := errors.New("unable to use spec")
+
+	t.Run("error reading current spec", func(t *testing.T) {
+		mockReadWriter.EXPECT().ReadFile(currentPath).Return(emptySpec, specErr)
+
+		err = s.PrepareRollback(ctx)
+		require.ErrorIs(err, ErrReadingRenderedSpec)
```
hexfusion, line 585: see how nice this is :)

internal/agent/device/spec/spec_test.go
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+		require.NoError(err)
+		mockReadWriter.EXPECT().ReadFile(rollbackPath).Return(rollbackSpec, nil)
+
+		mockClient.EXPECT().GetRenderedDeviceSpec(ctx, gomock.Any(), gomock.Any()).Return(nil, http.StatusServiceUnavailable, specErr)
+
+		_, err = s.GetDesired(ctx, renderedVersion)
+		require.ErrorIs(err, ErrGettingDeviceSpec)
+	})
+
+	t.Run("desired spec is returned when managemnt api returns no content", func(t *testing.T) {
```
hexfusion, line 807: minor:
```suggestion
	t.Run("desired spec is returned when management api returns no content", func(t *testing.T) {
```

internal/agent/device/spec/spec_test.go
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+		renderedDesiredSpec := createRenderedTestSpec(image)
+		marshaledDesiredSpec, err := json.Marshal(renderedDesiredSpec)
+		require.NoError(err)
+		mockReadWriter.EXPECT().ReadFile(desiredPath).Return(marshaledDesiredSpec, nil)
+
+		rollbackSpec, err := createEmptyTestSpec()
+		require.NoError(err)
+		mockReadWriter.EXPECT().ReadFile(rollbackPath).Return(rollbackSpec, nil)
+
+		// Api is returning a renderd version that is different from the read desired spec
```
hexfusion, line 856: minor:
```suggestion
		// Api is returning a rendered version that is different from the read desired spec

```

internal/agent/device/spec/spec_test.go
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+		apiResponse := &v1alpha1.RenderedDeviceSpec{RenderedVersion: "5"}
+		mockClient.EXPECT().GetRenderedDeviceSpec(ctx, gomock.Any(), gomock.Any()).Return(apiResponse, 200, nil)
+
+		// The difference results in a write call for the desired spec
```
hexfusion, line 860: :+1:  Because the rendered version has changed we persist the desired state to disk then begin to reconcile.

internal/agent/device/spec/spec_test.go
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+		log: log.NewPrefixLogger("test"),
+	}
+
+	testCases := []struct {
+		Name                    string
+		CurrentRenderedVersion  string
+		DesiredRenderedVersion  string
+		RollbackRenderedVersion string
+		ExpectedReturnValue     string
+		ExpectedError           error
```
hexfusion, line 1037: minor:

Generally, this test structure (which I like) uses non-exported keys and for readability use the key name in the case. if expectedError is nil you can just leave it out. `name` in this case is the name of the test so adding a bit more context is helpful when you are trying to debug.

```
func Test_pathFromType(t *testing.T) {
	require := require.New(t)

	s := &SpecManager{
		currentPath:  "test/current.json",
		desiredPath:  "test/desired.json",
		rollbackPath: "test/rollback.json",
	}

	testCases := []struct {
		name          string
		specType      Type
		expectedPath  string
		expectedError error
	}{
		{
			name:"current",
			specType:"current",
			expectedPath:s.currentPath,
		},
		...
	```

File: internal/agent/device/spec/spec_test.go [hunk 7]
```diff
@@ -551,13 +553,525 @@ func TestCheckOsReconciliation(t *testing.T) {
...
+		t.Run(testCase.Name, func(t *testing.T) {
+			renderedVersion, err := s.getRenderedVersion(
+				testCase.CurrentRenderedVersion,
+				testCase.DesiredRenderedVersion,
+				testCase.RollbackRenderedVersion,
+			)
+
+			if testCase.ExpectedError != nil {
+				require.ErrorIs(err, testCase.ExpectedError)
+			} else {
```
hexfusion, line 1058: style: since the above is an error case there won't be any more reasonable tests below so just return. that also gets rid of the else so its cleaner. 
 
 ```
if testCase.ExpectedError != nil {
   require.ErrorIs(err, testCase.ExpectedError)
   return
}
require.NoError(err)
require.Equal(testCase.ExpectedReturnValue, renderedVersion)
```

hexfusion (commented):
Really nice focus on edge cases here I am excited to see the coverage. Thanks :pray: . Few minor nits and style notes. Otherwise LGTM

PR #529, opened 2024-09-06
MGMT-18708: Add single CLI command to obtain enrollment cert

lkatalin:
~Needs some cleanup but comments welcome in the meantime.~
I'm not too opinionated on the name of the command, the flags, etc. - we could change it to `certificate request` - though I thought it could be helpful to introduce the user to the concept that they are doing pre-enrollment or registration.

Update: at least one commit also relates to MGMT-19059 (they are closely tied together).


fzdarsky:
> I'm not too opinionated on the name of the command, the flags, etc. - we could change it to `certificate request` - though I thought it could be helpful to introduce the user to the concept that they are doing pre-enrollment or registration.

The command doesn't register / pre-enroll a device, though, no? I mean, it requests a certificate from the service and returns that, but isn't tied to any device, doesn't tell the service anything about a device, and may even be shared between devices?

lkatalin:
> The command doesn't register / pre-enroll a device, though, no? I mean, it requests a certificate from the service and returns that, but isn't tied to any device, doesn't tell the service anything about a device, and may even be shared between devices?

I think it will make sense to consider "user requests an enrollment cert" as a specific step in the device lifecycle, rather than outside of the lifecycle. And it would almost certainly be the pre-enrollment step, or part of that step.

The way I had understood it, the eventual vision for the "pre-enrollment" step is it should let the FC service know that a device is expected to enroll, without any guaranteed data available about that device (this relates to the UUID we need to add as a primary key, which isn't in this PR). IMO it makes sense to do this when the user requests an enrollment cert because they are essentially declaring that one or more devices will enroll. 

I also wasn't sure whether we will keep supporting sharing the cert between devices long-term, for security reasons, but it should still be possible at the moment until there is a policy-based server-side check on whether the enrollment cert has been used before.

If we want to get this merged ASAP, I can just change the name of the command. But it will be good to sync up on design shortly as I'm trying to also lay the groundwork for the updated pre-enrollment process.

lkatalin:
Okay, as discussed, this command can just get a certificate without registering a device, to support legacy onboarding where the user does not know anything about the devices or how many devices will be enrolled. Later on with registration-based onboarding the user can get this same cert with a registration command when registering a device. I've updated the command name to `certificate request`.

internal/cli/certificate.go
```diff
@@ -0,0 +1,324 @@
...
+
+	switch o.OutputFormat {
+	case "components":
+		cadata, decodeErr := base64.StdEncoding.DecodeString(response.JSON200.EnrollmentService.Service.CertificateAuthorityData)
+		if decodeErr != nil {
+			return fmt.Errorf("error decoding CA data: %w", decodeErr)
+		}
+		err = createOutputComponents(name, currentCsr, priv, cadata)
+	default:
+		err = createOutputConfig(name, c, ctx, priv, response)
```
fzdarsky, line 159: What should the default be? Intuitively, the command is about getting the CA bundle, certificate, and keys first, getting them in the "right" format for an agent second?

cmd/flightctl/main.go
```diff
@@ -34,6 +34,7 @@ func NewFlightCtlCommand() *cobra.Command {
 	cmd.AddCommand(cli.NewConsoleCmd())
 	cmd.AddCommand(cli.NewCmdCompletion())
 	cmd.AddCommand(cli.NewCmdEnrollmentConfig())
+	cmd.AddCommand(cli.NewCmdCertificate())
```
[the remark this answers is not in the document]
lkatalin: I suppose it depends on whether we want to have the `apply -f` command (which mimics kubernetes) available as an alternative to this. If we do want it available, the enrollmentconfig command is the only way to produce the config file that is needed in that flow.

```
# Example 1
flightctl apply -f examples/csr.yaml
flightctl approve csr/mycsr # <-- this will go away with auto-approve
enrollmentconfig mycsr -p myeckey.pem

# Example 2 - equivalent to above
flightctl certificate request 
```

Note that if for some reason the user wants to manually approve an ecert (will we ever want to support that?) or hits `ctrl+C` during the `certificate request` before the CSR is approved and the files are written out, there isn't a good way to get the enrollment config asynchronously without the enrollmentconfig API.

internal/cli/certificate.go
```diff
@@ -0,0 +1,324 @@
...
+		SilenceUsage: true,
+	}
+	o.Bind(cmd.Flags())
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "mycsr", "specify a name for the certificate signing request")
+	fs.StringVarP(&o.Scope, "scope", "s", "enrollment", "specify certificate scope - the only supported scope is currently 'enrollment'")
```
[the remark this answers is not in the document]
lkatalin: That sounds fine. I'll update it.

internal/cli/certificate.go
lkatalin: I suppose as the main purpose for this output would be to get an agent config, I would tend to think of that as the default. But perhaps I am not aware of some other use cases.

cmd/flightctl/main.go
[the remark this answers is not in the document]
lkatalin: And, actually, the enrollmentconfig API requires specifying a key - so if we do want to make it an option to get the info with `enrollmentconfig` after a `certificate request`, then the key should be written out before the CSR is sent so that it is available even if the command doesn't complete. We should also probably provide an option to encrypt the keys.

internal/cli/certificate.go
```diff
@@ -0,0 +1,324 @@
...
+		return err
+	}
+
+	if len(args) == 0 {
+		return fmt.Errorf("one subcommand must be provided to 'certificate': options are: 'request'")
+	}
+
+	if args[0] != "request" {
+		return fmt.Errorf("one subcommand must be provided to 'certificate': options are: 'request'")
+	}
```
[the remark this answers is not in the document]
lkatalin: Great - done

cmd/flightctl/main.go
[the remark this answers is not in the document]
lkatalin: Based on today's sync we agreed there should be some API to fetch this type of data, but the `enrollmentconfig` API may morph into something more generic such as a discovery API. In any case changes to `enrollmentconfig` can be in a later PR.

internal/cli/certificate.go
```diff
@@ -0,0 +1,349 @@
...
+
+	fmt.Println("this operation will generate a new key pair")
+	fmt.Println("creating new key pair...")
+	_, priv, err := fccrypto.NewKeyPair()
+	if err != nil {
+		return fmt.Errorf("error creating new key pair: %w", err)
+	}
+
+	if o.PrivKeyPassword == "" {
+		pw, err := getPassword()
```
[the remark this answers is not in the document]
mangelajo: Ignore this, I saw later, that when generating an agent config you write the non-encrypted version of it.

cmd/flightctl/main.go
```diff
@@ -34,6 +34,7 @@ func NewFlightCtlCommand() *cobra.Command {
 	cmd.AddCommand(cli.NewConsoleCmd())
```
mangelajo, line 34: Additional note, 

If you add a:
   ./bin/flightctl login https://${IP}.nip.io:3443
in
    https://github.com/flightctl/flightctl/blob/main/test/scripts/deploy_with_helm.sh#L157

and then

remove the totality of:
https://github.com/flightctl/flightctl/blob/main/test/scripts/prepare_agent_config.sh
to use your ./bin/flightctl certificate request
command, leaving the agent config in the same place... you would have full e2e validation of your command in CI :)

mangelajo (changes requested):

internal/cli/certificate.go
```diff
@@ -0,0 +1,349 @@
...
+		},
+		SilenceUsage: true,
+	}
+	o.Bind(cmd.Flags())
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "mycsr", "specify a name for the certificate signing request")
```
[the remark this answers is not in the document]
lkatalin: I'm not aware of a way to randomize it in the cobra `StringVarP` itself (but let me know if there is) but the internal logic will handle this and always randomize to `mycsr-382093` or so.

internal/cli/certificate.go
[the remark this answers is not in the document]
lkatalin: Yes, this is true. However you raise an interesting point that I am still thinking through re: how we might want to protect the agent config.

internal/cli/certificate.go
```diff
@@ -0,0 +1,349 @@
...
+	fs.StringVarP(&o.SignerName, "signer", "s", "ca", "specify certificate signer - options currently supported: 'ca'")
+	fs.Int32VarP(&o.ExpirationSeconds, "expiry", "e", 604800, "specify desired certificate expiration in seconds")
+	fs.StringVarP(&o.OutputFormat, "format", "f", "components", "specify desired output format: either 'agent-config' to output values in one config file, or 'components' to output cert, key, and ca values in separate files")
+	fs.StringVarP(&o.PrivKeyPassword, "password", "w", "", "specify a password for newly generated private key, which will be encrypted on local disk")
```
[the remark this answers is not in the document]
lkatalin: If the flag is empty, the user will be prompted for a pw when this runs. If they then hit enter, the key will be written out unencrypted.

cmd/flightctl/main.go
lkatalin: Nice! Thanks, I will try it

lkatalin:
New version incoming that parses expiration in the suggested format. Please hold.

internal/cli/certificate.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	o.Bind(cmd.Flags())
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "mycsr", "specify a name for the certificate signing request")
+	fs.StringVarP(&o.SignerName, "signer", "s", "ca", "specify certificate signer - options currently supported: 'ca'")
+	fs.Int32VarP(&o.ExpirationSeconds, "expiry", "e", 604800, "specify desired certificate expiration in seconds")
```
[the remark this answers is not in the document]
lkatalin: `time.ParseDuration` doesn't support this but I am working on it.

lkatalin:
@mangelajo @fzdarsky I've pushed a new version.

Highlights:
- The user inputs the desired cert expiration in the format `Xd` where X is a number of days
- I realized if the user uses the `enrollmentconfig` API (or whatever it morphs into) to retrieve the agent config later, then the agent config will have the encrypted version of the key embedded and other parts of the system won't know what to do with this - so I fixed that in `crypto/key.go` by checking if the key needs to be decrypted

cmd/flightctl/main.go
lkatalin: I probably didn't implement this correctly because it caused CI to fail, but I can play around with it more separately to this PR.

internal/cli/certificate.go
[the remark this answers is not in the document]
fzdarsky: I think it‘s ok to only append the random suffix later. I’d just drop the “my” from the default name.

One problem I see is we also append the random suffix to the name of saved components. That makes them cumbersome to use in automation. Should we just give these fixed names (e.g. ca-cert-bundle.pem, cert.pem, key.pem)?

internal/cli/certificate.go
```diff
@@ -0,0 +1,367 @@
...
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "mycsr", "specify a name for the certificate signing request")
+	fs.StringVarP(&o.SignerName, "signer", "s", "ca", "specify certificate signer - options currently supported: 'ca'")
+	fs.StringVarP(&o.Expiration, "expiration", "e", "7d", "specify desired certificate expiration in days, example: 7d")
+	fs.StringVarP(&o.OutputFormat, "output", "o", "components", "specify desired output format: either 'agent-config' to output values in one config file, or 'components' to output cert, key, and ca values in separate files")
+	fs.StringVarP(&o.PrivKeyPassword, "password", "w", "", "specify a password for newly generated private key, which will be encrypted on local disk")
```
fzdarsky, line 80: Passing in a password like this is insecure (it'll show in `ps` for example). Instead, we'd probably want to read it either from env var or from stdin (say, if the parameter was `-w -`).

internal/cli/certificate.go
```diff
@@ -0,0 +1,367 @@
...
+		SilenceUsage: true,
+	}
+	o.Bind(cmd.Flags())
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "mycsr", "specify a name for the certificate signing request")
+	fs.StringVarP(&o.SignerName, "signer", "s", "ca", "specify certificate signer - options currently supported: 'ca'")
```
fzdarsky, line 77: Can we replace "ca" here immediately, so it doesn't bite us later? `enrollment`, `enrollmentSigner` or similar?

internal/cli/certificate.go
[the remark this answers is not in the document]
fzdarsky: Both the `agent-config` and `component` outputs may be used to enrol agents and in that case cannot be encrypted. How can users specify "unencrypted" for either from their automation (when you cannot interactively ask for the password)? Should password encryption be opt-in, e.g. with an `--encrypt-key`, or asking for a password `--ask-password`?

internal/cli/certificate.go
```diff
@@ -0,0 +1,367 @@
...
+func (o *CertificateOptions) Run(ctx context.Context, args []string) error {
+	fmt.Printf("the expiration is %s\n", o.Expiration)
+
+	name := createUniqueName(o.Name)
+
+	fmt.Println("this operation will generate a new key pair")
+	fmt.Println("creating new key pair...")
+	_, priv, err := fccrypto.NewKeyPair()
+	if err != nil {
+		return fmt.Errorf("error creating new key pair: %w", err)
```
fzdarsky, line 110: Cobra itself will prefix `Error:`, so you can leave out the `error ` from the messages.

internal/cli/certificate.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+func createOutputConfig(name string, c *apiclient.ClientWithResponses, ctx context.Context, priv crypto.PrivateKey, response *apiclient.EnrollmentConfigResponse) error {
+	filename := "enrollment-config-" + name + ".yaml"
+	f, err := os.Create(filename)
+	fmt.Printf("creating output for enrollment config: %s\n", filename)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
fzdarsky, line 306: Can we just print to `stdout`? Then users can redirect to whatever they want to call the file or to pipe to other processors like `yq`? (Obviously, we'd then need to ensure the log output goes to `stderr` instead.)

internal/cli/certificate.go
mangelajo: https://docs.openssl.org/1.0.2/man1/openssl/#encoding-and-cipher-commands this is what openssl accept, I think it's fine to accept cmdline, then it's up to the user to decide , also the fallback is "asking password", we could add env in addition.

cmd/flightctl/main.go
mangelajo: It is ok if you make it fail here I can help look at it.

internal/cli/certificate.go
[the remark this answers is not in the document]
mangelajo: May be the randomized is only a default. You could generate "mycsr" with a default name during "Bind" method call, then at each boot you'd get a different default.

internal/cli/certificate.go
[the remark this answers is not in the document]
mangelajo: There's where the interactions with TPM could help at some point.

internal/cli/certificate.go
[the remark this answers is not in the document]
mangelajo: The agent config key is always non-encrypted in this code.

We should fail if encryption and agent-config are requested at once in flags.

internal/cli/certificate.go
lkatalin: Yeah, I didn't love this either but wasn't sure whether to completely drop it as an option. I can add an env var option in addition to taking the pw at runtime or as a command line arg. I don't know if cobra will read from stdin? I can look into it...

internal/cli/certificate.go
lkatalin: I was going to leave this to be included in @kot-begemot-uk 's PR but I am fine to change it here.

internal/cli/certificate.go
lkatalin: Sure, I'll update

lkatalin:
@fzdarsky Not sure why I can't reply inline to your comment about the component output encryption so I'll do it here - you have a good point about this. I'd prefer encrypting the key to be opt-out but I can add an option for that.

internal/cli/certificate.go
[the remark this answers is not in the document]
fzdarsky: I'm explicitly talking about `component` here. Remember that users who manage the agent's config.yaml separately from the enrollment credentials do not embed the credentials into the config, but _reference_ the files from the config. Then, it's important the files that `component` outputs are non-encrypted, too.

internal/cli/certificate.go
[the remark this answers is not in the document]
fzdarsky: As long as the `component` files do not contain the random suffix I think that's ok.

internal/cli/certificate.go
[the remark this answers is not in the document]
lkatalin: Hmmm these are good points about usage. There is a security tradeoff here that I am trying to think through but not yet sure what the best solution is.

lkatalin:
This new version removes the `signername` option because there is only one valid signer for this `enrollment` - so we just generate it for the user. It also does not write anything to a file, instead writing to stdout. We can discuss more tomorrow if needed.

internal/cli/certificate.go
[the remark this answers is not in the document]
fzdarsky: It does, thanks!

lkatalin:
Highlights in the new version:
- The output formats are called `reference` and `embedded`.  Both print an agent config to stdout. The `reference` format includes the file paths (can be chosen by the user, default `bin/agent/etc/flightctl/certs/`) to the agent key and cert, while the `embedded` format includes that material in the config itself. 
- The `reference` type also outputs the key, cert, and ca as files. The `embedded` type does not write out to files.
-The `reference` output type can opt in to use encryption for the key saved on disk. The password is read from an env var `FCPASS` or if this env var is not set it will prompt the user.

This is what the output looks like for a default `reference` output type:

```
[lily@fedora flightctl]$ bin/flightctl certificate request -n nameichose
this operation will generate a new ecdsa key pair
creating new key pair...
writing private key to: bin/agent/etc/flightctl/certs/nameichose.key...
submitting CSR to flightctl service...
201 Created: nameichose
checking for certificate...
...
certificate is ready
writing data to bin/agent/etc/flightctl/certs/nameichose.crt...
writing data to bin/agent/etc/flightctl/certs/nameichose-ca-data.crt...
enrollment-service:
  authentication:
    client-certificate-data: bin/agent/etc/flightctl/certs/nameichose.crt
    client-key-data: bin/agent/etc/flightctl/certs/nameichose.key
  enrollment-ui-endpoint: http://localhost:9000
  service:
    certificate-authority-data: LS0tL...
    server: https://agent-api.192.168.0.12.nip.io:7443/
grpc-management-endpoint: grpcs://agent-grpc.192.168.0.12.nip.io:7444
```

Re: opt-in encryption of the key, as a reference kubernetes does the same as stated here https://kubernetes.io/docs/concepts/configuration/secret/ :
```
Kubernetes Secrets are, by default, stored unencrypted in the API server's underlying data store (etcd). Anyone with API access can retrieve or modify a Secret, and so can anyone with access to etcd.
```
However I think it will be good to get some kind of integration with Vault or similar on the roadmap for this purpose.

cmd/flightctl/main.go
lkatalin: I took a closer look - in order to do this, I will need to place the agent config in a specific location in the `embedded` format, is this what we want? @fzdarsky was suggesting `stdout` for the output so that's what I implemented. We can certainly also have some option to put it in the file that CI expects, but as there have been so many changes to this PR it may be good to do it separately so that this can be merged. Thoughts?

lkatalin: Ah, I misunderstood here because my brain is a bit fried. Totally possible and easy to do this so in `prepare_agent_config.sh` I've removed everything but the `cp` commands that we still need and I added the call to `./bin/flightctl/ certificate request > bin/agent/etc/flightctl/certs/config.yaml`. We'll see if it passes CI!

mangelajo: https://github.com/flightctl/flightctl/actions/runs/11153840866/job/31002105193#step:7:1064

For the login failure you need to use the insecure tls flags (we have two, I never remember which one it is), otherwise we need to know the CA before hand, but for the devel env is completely fine to ignore that.

test/scripts/deploy_with_helm.sh
```diff
@@ -155,6 +155,8 @@ if [ -z "$ONLY_DB" ]; then
   chmod og-rwx ~/.flightctl/certs/*.key
 fi
 
+./bin/flightctl login https://api.${IP}.nip.io:3443
```
mangelajo, line 158:
```suggestion
./bin/flightctl login --insecure-skip-tls-verify https://api.${IP}.nip.io:3443
```

test/scripts/prepare_agent_config.sh
```diff
@@ -1,24 +1,6 @@
 #!/usr/bin/env bash
 
-SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
-IP=$("${SCRIPT_DIR}"/get_ext_ip.sh)
-OUTFILE=bin/agent/etc/flightctl/config.yaml
-mkdir -p bin/agent/etc/flightctl/certs
 cp ~/.flightctl/certs/ca.crt bin/agent/etc/flightctl/certs/ca.crt
 cp ~/.flightctl/certs/client-enrollment.{crt,key} bin/agent/etc/flightctl/certs/
```
mangelajo, line 4:
```suggestion
```

The agent config has the CA embedded and the enrollment cert and key as well, so we don't need to copy that anymore I believe

test/scripts/prepare_agent_config.sh
```diff
@@ -1,24 +1,6 @@
...
-  service:
-    certificate-authority: certs/ca.crt
-    server: https://agent-api.${IP}.nip.io:7443
-  enrollment-ui-endpoint: https://ui.${IP}.nip.io:8080
-grpc-management-endpoint: grpcs://agent-grpc.${IP}.nip.io:7444
-spec-fetch-interval: 0m10s
-status-update-interval: 0m10s
-tpm-path: /dev/tpmrm0
-EOF
+./bin/flightctl certificate request > bin/agent/etc/flightctl/config.yaml
```
mangelajo, line 6:
```suggestion
echo Requesting enrollment enrollment certificate/key and config for agent =====
./bin/flightctl certificate request | tee bin/agent/etc/flightctl/config.yaml
```

Just a suggestion, it will help us see what's going on.

internal/cli/certificate.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+func (o *CertificateOptions) Validate(args []string) error {
+	if err := o.GlobalOptions.Validate(args); err != nil {
+		return err
+	}
+
+	if o.OutputFormat != "reference" && o.OutputFormat != "embedded" {
+		return fmt.Errorf("invalid output format specified: %s\noutput format must be either `reference` (agent config printed to stdout references key, cert, CA files written out during runtime) or `embedded` (agent config printed to stdout has key, cert, CA info embedded)", o.OutputFormat)
+	}
```
fzdarsky, line 95: This would mean users can no longer request certificates without agent-config or that are not for enrollment.
We can do that, but should then `certificat request` no longer seems appropriate as command and we should probably pivot back to `enrollmentconfig`.

internal/cli/certificate.go
```diff
@@ -0,0 +1,403 @@
...
+	// write out agent cert
+	certpath := path + name + ".crt"
+	cert := currentCsr.Status.Certificate
+	err := writeCertFile(*cert, certpath)
+	if err != nil {
+		return err
+	}
+
+	// write out ca cert
+	capath := path + name + "-ca-data.crt"
```
fzdarsky, line 331: `-ca-bundle.crt`?

internal/cli/certificate.go
```diff
@@ -0,0 +1,403 @@
...
+		return fmt.Errorf("decoding CA data: %w", err)
+	}
+	err = writeCertFile(cadata, capath)
+	if err != nil {
+		return err
+	}
+
+	// print config to stdout
+	response.JSON200.EnrollmentService.Authentication.ClientKeyData = path + name + ".key"
+	response.JSON200.EnrollmentService.Authentication.ClientCertificateData = certpath
```
fzdarsky, line 343: This must be the path where the _agent_ will find the files _on the device_ (/etc/flightctl/certs).

fzdarsky (changes requested):

internal/cli/certificate.go
```diff
@@ -0,0 +1,403 @@
...
+	o.Bind(cmd.Flags())
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "csr", "specify a name for the certificate signing request")
+	fs.StringVarP(&o.Expiration, "expiration", "x", "7d", "specify desired certificate expiration in days, example: 7d")
+	fs.StringVarP(&o.OutputFormat, "output-format", "o", "reference", "specify desired output format: either 'agent-config' to output values in one config file, or 'components' to output cert, key, and ca values in separate files")
+	fs.StringVarP(&o.OutputDir, "output-dir", "d", "bin/agent/etc/flightctl/certs/", "specify desired output directory for key, cert, and ca files")
```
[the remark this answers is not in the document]
mangelajo: I was about to say that... I was testing and I saw it
```
$ ./bin/flightctl certificate request --output-format embedded -x 365d
this operation will generate a new ecdsa key pair
creating new key pair...
writing private key to: bin/agent/etc/flightctl/certs/csr-109c1717.key...
submitting CSR to flightctl service...
400 Bad Request: csr-109c1717
Error: submitting csr to flightctl service: csr-109c1717: applying CertificateSigningRequest: {"message":"spec.signerName: Invalid value: \"enrollment\": must specify a valid signer"}

[kni@titan91 flightctl]$ ls bin/agent/etc/flightctl/certs/csr-109c1717.key
bin/agent/etc/flightctl/certs/csr-109c1717.key
```

mangelajo:
@lkatalin there is progress!, 

although there is something weird... it gets stuck waiting for the csr:

https://github.com/flightctl/flightctl/actions/runs/11162417791/job/31027018516#step:7:1161

internal/cli/certificate.go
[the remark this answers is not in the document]
lkatalin: Sure, I am updating the default. The 400 bad request is unrelated though - ?

internal/cli/certificate.go
lkatalin: Hm, yes. Okay, I will make it so they can get an enrollment cert or some other cert, and in case of enrollment cert it has output format options of reference or embedded.

lkatalin: Higher-level question: if this CLI command is going to be used for non-enrollment certs, those will not be auto-approved, so should this process block until the cert is ready? That doesn't seem to make sense. Maybe I'll include the blocking and checking for the cert only in the case that it is an enrollment cert request

internal/cli/certificate.go
lkatalin: I've done the following:
If the output type is `reference`, the key, cert, and ca-bundle files are emitted to the user-chosen directory. However, the agent config references `/etc/flightctl/certs/` + the filename.

internal/cli/certificate.go
lkatalin: Actually, looking back at the title of this PR, I think the original goal as I had understood it was to offer a command to get enrollment certs only. This makes sense because these are auto-approved so the command can wait for the certificates to be populated. This wouldn't work with other types of certs unless we want to extend the auto-approval functionality to cover more types - which may not be a good idea.

lkatalin:
An update, the CI keeps failing once I change `prepare_agent_config.sh` to use the new `certificate request` command as the output of `certificate request` uses the `enrollmentconfig` API structs and these do not match the devicesimulator's `Config` and `AuthInfo` structs... for example the API has no field `ClientKey`, only `ClientKeyData`. I have to think through how to get these to agree, in order to make this CI change.

lkatalin:
Looks like this works, please take a look @mangelajo @fzdarsky

lkatalin:
CI was passing until I rebased on main a few times, so I believe the failure is unrelated to this PR.

mangelajo:
> CI was passing until I rebased on main a few times, so I believe the failure is unrelated to this PR.

I confirm it's unrelated to your change, seems to be in the console, but I cannot see the root cause.

test/scripts/agent-images/prepare_agent_config.sh
```diff
@@ -1,25 +1,6 @@
...
-  service:
-    certificate-authority: certs/ca.crt
-    server: ${FLIGHTCTL_API_ENDPOINT}
-  enrollment-ui-endpoint: ${FLIGHTCTL_UI_ENDPOINT}
-grpc-management-endpoint: ${FLIGHTCTL_AGENT_GRPC}
-spec-fetch-interval: 0m10s
-status-update-interval: 0m10s
-tpm-path: /dev/tpmrm0
-EOF
+echo Requesting enrollment enrollment certificate/key and config for agent =====
```
mangelajo, line 5: 🤟

internal/cli/certificate.go
```diff
@@ -0,0 +1,434 @@
...
+	return cmd
+}
+
+func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.Name, "name", "n", "csr", "specify a name for the certificate signing request")
+	fs.StringVarP(&o.Expiration, "expiration", "x", "7d", "specify desired certificate expiration in days, example: 7d")
+	fs.StringVarP(&o.OutputFormat, "output-format", "o", "default", "specify desired output format for an enrollment cert: either 'reference' to have the config file reference key and cert file paths, or 'embedded' to have the key and cert embedded in the config file")
+	fs.StringVarP(&o.OutputDir, "output-dir", "d", "./", "specify desired output directory for key, cert, and ca files (defaults to current directory)")
+	fs.StringVarP(&o.SignerName, "cert-type", "t", "enrollment", "specify the type of certificate requested: 'enrollment' or 'ca' (default 'enrollment')")
```
mangelajo, line 88: I understand it's the signer not the "type", 

I find it weird that we could ask the CA for a CA certificate.

I guess we are asking for a specific signer.

We need to iron this out, but I am good to do this as a follow up, the functionality you are building here would be useful already there: https://github.com/flightctl/flightctl/pull/569/files#diff-c3579ee33fafd5cc39c2bcc363f2867cf84a172ec353e7d53265439170111719

mangelajo (approved):
IMO this is good to merge, then we can follow up on details (like the cert-type = ca detail)

mangelajo:
There is something with the E2E console, could be indirectly related, I am running locally to gather more details.

i.e. may be the TLS cert or config obtained from this does not connect correctly to the console grpc endpoint?

test/scripts/agent-images/prepare_agent_config.sh
```diff
@@ -1,25 +1,6 @@
...
+./bin/flightctl certificate request -n client-enrollment -d bin/agent/etc/flightctl/certs/ | tee bin/agent/etc/flightctl/config.yaml
```
mangelajo, line 6:
```suggestion
./bin/flightctl certificate request -d bin/agent/etc/flightctl/certs/ | tee bin/agent/etc/flightctl/config.yaml

# enforce the agent to fetch the spec and update status every 2 seconds to improve the E2E test speed
cat <<EOF | tee -a  bin/agent/etc/flightctl/config.yaml
spec-fetch-interval: 0m2s
status-update-interval: 0m2s
EOF
```

mangelajo: We were missing a couple of settings which are important for E2E testing, the update intervals, otherwise the tests need to wait too long... this is why the console timeout was materializing, the agent takes too long to realize there is a console to be served... the test gives up.

Also note that I have removed the -n client-enrollment

The reason for that, is that if you run this a second time, it will still work because of the random name, otherwise it fails, and you can't run make e2e-test  twice without throwing again and redeploying or deleting csr/client-enrollment.

test/scripts/deploy_with_helm.sh
```diff
@@ -156,6 +156,8 @@ if [ -z "$ONLY_DB" ]; then
   chmod og-rwx ~/.flightctl/certs/*.key
 fi
 
+./bin/flightctl login --insecure-skip-tls-verify https://api.${IP}.nip.io:3443
```
mangelajo, line 159: # attempt to login, it could take some time for API to be stable
```suggestion
for i in {1..5}; do
  if ./bin/flightctl login --insecure-skip-tls-verify https://api.${IP}.nip.io:3443; then
    break
  fi
  sleep 5
done
```

Locally this worked better, somehow my API needed a few more seconds to be fully functional and responding.

mangelajo (changes requested):
See comments, that should fix the console.

mangelajo (approved):
Let's see if E2E console tests are happy again

test/scripts/agent-images/prepare_agent_config.sh
```diff
@@ -1,25 +1,12 @@
...
-  service:
-    certificate-authority: certs/ca.crt
-    server: ${FLIGHTCTL_API_ENDPOINT}
-  enrollment-ui-endpoint: ${FLIGHTCTL_UI_ENDPOINT}
-grpc-management-endpoint: ${FLIGHTCTL_AGENT_GRPC}
-spec-fetch-interval: 0m10s
-status-update-interval: 0m10s
-tpm-path: /dev/tpmrm0
+echo Requesting enrollment enrollment certificate/key and config for agent =====
+./bin/flightctl certificate request -d bin/agent/etc/flightctl/certs/ | tee bin/agent/etc/flightctl/config.yaml
```
mangelajo, line 6:
```suggestion
# remove any previous CSR with the same name in case it existed
./bin/flightctl delete csr/client-enrollment || true

./bin/flightctl certificate request -n client-enrollment  -d bin/agent/etc/flightctl/certs/ | tee bin/agent/etc/flightctl/config.yaml
```

mangelajo (approved):

internal/cli/certificate.go
fzdarsky: Yeah, it's about having a certificate signed for a specific purpose by a signer specialized on that purpose. Right now, we only have CSRs for enrollment certs, at some point _possibly_ also for the management certs (for cert rotation) or applications.
I believe the "ca" type in here is a leftover from refactoring that we should remove.

internal/cli/certificate.go
```diff
@@ -0,0 +1,434 @@
...
+func (o *CertificateOptions) Complete(cmd *cobra.Command, args []string) error {
+	return o.GlobalOptions.Complete(cmd, args)
+}
+
+func (o *CertificateOptions) Validate(args []string) error {
+	if err := o.GlobalOptions.Validate(args); err != nil {
+		return err
+	}
+
+	errs := validation.ValidateSignerName(o.SignerName)
```
fzdarsky, line 101: I will probably pull in the server-side validation code, including all deps from K8s. We might want to replace that in a later PR.

test/scripts/agent-images/prepare_agent_config.sh
lkatalin: Interesting, these fields were removed deliberately in https://github.com/flightctl/flightctl/pull/522/ 

That sounds good, I had just put `-n client-enrollment` to make it match `prepare_agent_config.sh` as I was not sure if the name is important in terms of where the file is placed.

lkatalin: I see the intervals are needed just for testing.

internal/cli/certificate.go
lkatalin: Yes, we decide/indicate the "type" (enrollment, management, other types) based on the signer, so "type" and "signer" are effectively the same thing.

"ca" is only left in as a type/signer because that is the only other type/signer we have other than "enrollment" and I needed something from which to differentiate "enrollment" in the logic to do auto-approves.

fzdarsky (approved):
There are a couple of smaller things to iron out, but I'd suggest merging this and doing the changes in a follow-up PR.

mangelajo:
🥳

PR #532, opened 2024-09-10
EDM-447: Add template version approval API endpoint

ori-amizur:


fzdarsky:
We'll need a design for EDM-424 first before we can determine whether we need a separate API and how it should look/behave.

PR #534, opened 2024-09-12
EDM-462: adding unit tests for console.go.

amir-yogev-gh:


DakCrowder:
Just a heads up, looks like there is a panic occurring when running in the GH action via `make unit-test`

internal/agent/device/console/console_test.go
```diff
@@ -0,0 +1,149 @@
...
+		})
+
+		It("no desired console with active stream failure", func() {
+			consoleController.active = true
+			consoleController.streamClient = mockStreamClient
+
+			mockStreamClient.EXPECT().CloseSend().Return(errors.New("close send error"))
+
+			err := consoleController.Sync(ctx, &api.RenderedDeviceSpec{})
+			Expect(err).To(HaveOccurred())
```
hexfusion, line 88: The actual CloseSend() errors are discarded so this test case is not testing the implementation logic.

https://github.com/flightctl/flightctl/blob/f1dd705d06150674746b697d91b935a04ffeada4/internal/cli/console.go#L226C4-L226C26

internal/agent/device/console/console.go
```diff
@@ -141,6 +141,10 @@ func (c *ConsoleController) startForwarding(ctx context.Context, stdin io.WriteC
 				c.log.Errorf("stream > bash:error receiving message for stdin: %s", err)
 				return fmt.Errorf("stream > bash:error receiving message for stdin: %w", err)
 			}
+			if msg == nil {
```
hexfusion, line 144: can this happen in practice? my understanding is the only way Recv can return a nil value is if err is non nil.

internal/agent/device/console/console_test.go
```diff
@@ -0,0 +1,149 @@
...
+			consoleController.currentStreamID = sessionId
+
+			mockStreamClient.EXPECT().Recv().Return(nil, nil).AnyTimes()
+			mockStreamClient.EXPECT().Send(gomock.Any()).Return(nil).AnyTimes()
+			mockStreamClient.EXPECT().CloseSend().Return(nil).AnyTimes()
+			mockGrpcClient.EXPECT().Stream(gomock.Any()).Return(mockStreamClient, nil)
+			mockExecutor.EXPECT().CommandContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(testCommand)
+
+			err := consoleController.Sync(ctx, desired)
+			Expect(err).ToNot(HaveOccurred())
```
hexfusion, line 102: how do we know what error has occurred and that it is expected?

internal/agent/device/console/console_test.go
```diff
@@ -0,0 +1,149 @@
...
+			err := consoleController.Sync(ctx, desired)
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("error creating console stream client", func() {
+			mockGrpcClient.EXPECT().Stream(gomock.Any()).Return(nil, errors.New("stream creation error"))
+			mockExecutor.EXPECT().CommandContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(testCommand)
+
+			err := consoleController.Sync(ctx, desired)
+			Expect(err).To(HaveOccurred())
```
hexfusion, line 135: again it would be good to understand what errored and if that is expected. see how spec manager created explicit error types that can be wrapped and validated. That would make sense here as well.

hexfusion (commented):
Thanks for the PR! I think this coverage is a step in the right direction but it does point to the challenges of unit testing gRPC. I wonder if the logic could be broken into functions that would make this easier to test in the future?

internal/agent/device/console/console_test.go
hexfusion: as we spoke about I missed a usage in console.go so disregard this statement.

internal/agent/device/console/console_test.go
hexfusion: A good example of this can be seen in this search for ErrNoContent. It shows how we can explicitly ensure an error without having to match error strings. This can be a future change just something that will help future tests.

```go
// declare
var (
   ErrNoContent         = fmt.Errorf("no content")
 )
 
 // wrap and return
 if noContent {
     return fmt.Errorf("we failed for some reason: %w")
 }
 
 // test
 require.ErrorIs(err, ErrNoContent)
 ```
ref. https://github.com/search?q=repo%3Aflightctl%2Fflightctl%20ErrNoContent&type=code

File: internal/agent/agent.go [hunk 5]
```diff
@@ -194,7 +195,7 @@ func (a *Agent) Run(ctx context.Context) error {
 	)
 
 	// create console controller
-	consoleController := device.NewConsoleController(
+	consoleController := console.NewConsoleController(
```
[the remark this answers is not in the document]
amir-yogev-gh: I only moved the existing code to a new folder.
Sure, I can rename it.

internal/agent/device/console/console_test.go
```diff
@@ -0,0 +1,149 @@
...
+			Expect(consoleController.active).To(BeFalse())
+		})
+
+		It("active console with new desired console", func() {
+			consoleController.active = true
+			consoleController.streamClient = mockStreamClient
+
+			err := consoleController.Sync(ctx, desired)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(consoleController.active).To(BeTrue())
```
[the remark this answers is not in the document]
amir-yogev-gh: It's the old session.
The setup before the call to Sync() is marking the old session as active.

internal/agent/device/console/console.go
amir-yogev-gh: This defensive writing. The logic will panic without it for a nil message.
Such case is considered as an error.
I can adjust the test case and send a non-nil message.

If this can happen, I prefer a graceful error instead of an exception.

internal/agent/device/console/console_test.go
amir-yogev-gh: All error are provided through the mocks. It's built-in for each test case.

internal/agent/device/console/console_test.go
amir-yogev-gh: The provided examples are great, but for these unit tests:

1. The tested logic only distinguish between nil or non-nil errors. The concrete error is irrelevant for the current console implementation.
2. The concrete errors are not duplicated between the test cases.

Even though it's possible to follow your example, I don't see the benefit in these unit tests.

internal/agent/device/console/console_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"context"
+	"errors"
+	"os/exec"
+	"testing"
+
+	"github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/pkg/executer"
+	"github.com/flightctl/flightctl/pkg/log"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
```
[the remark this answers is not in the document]
amir-yogev-gh: Refactored

internal/agent/device/console/console.go
amir-yogev-gh: Dropped this check

mangelajo (approved):
Looks good to me, sets a foundation to start writing more console unit tests.

PR #535, opened 2024-09-17
EDM-458: adding devices summary API

asafbss:


api/v1alpha1/openapi.yaml
```diff
@@ -1013,6 +1013,50 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+
+  /api/v1/devices/summary:
```
fzdarsky, line 1017: This overloads `/api/v1/devices/{name}`. ATM, we only allow device fingerprints for names and those do not conflict, of course. However, we do have the request to support more human-readable names (not aliases) and may solve that through user-defined generator functions (e.g. based on hostname). In that case there is a residual risk of conflict (e.g. a host being named "summary").

api/v1alpha1/openapi.yaml
```diff
@@ -3507,16 +3551,25 @@ components:
       properties:
         total:
           type: integer
+          format: int64
           description: The total number of devices in the fleet.
+        applicationStatus:
```
fzdarsky, line 3556: You may want to make these statuses required, so we do not need to handle nil pointers later.

internal/cli/get.go
```diff
@@ -113,14 +113,14 @@ func (o *GetOptions) Validate(args []string) error {
 	if err != nil {
 		return err
 	}
-	if len(name) > 0 && len(o.LabelSelector) > 0 {
+	if len(name) > 0 && !strings.EqualFold(name, "summary") && len(o.LabelSelector) > 0 {
```
fzdarsky, line 116: This overloads the `flightctl get resource_kind/resource_name` command. Could we find an unambiguos alternative?

Maybe `flightctl get devices --summary`, although the problem with that is `get devices` implies a list of devices that then gets filtered/formatted according to the flags and the summary is a completely different thing.

Maybe instead of enhancing the get/apply/delete commands that manipulate the _API resources_ we model this as a separate, high-level command like `flightctl devices summary` or `flightctl stats devices`. Needs discussion.

avishayt: ListDevices already returns this summary, correct?  The difference is that the summary currently corresponds to the devices that it returns, which may be filtered or paginated.
Maybe we can have a new query param like "deviceSummaryOnly=true" that would ignore all the filters and just return the summary (items would be an empty list).
WDYT?

avishayt (commented):
@ori-amizur The DB code looks good to me but please review as well

internal/cli/get.go
asafbss: Yes, I wouldn't say I liked it myself. Without it, we crash trying to unmarshal something not a device list. 
The challenge with the list devices API is that it fetches the devices' objects up to the limit and uses the []devices to calculate the summary on the API service. We can add another query string summaryOnly to return only the summary of the entire devices, leveraging the DB capabilities to group it without fetching the entire devices' objects. But keep in mind that the summary of the list is not reflected in the CLI, only per device: "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN." So, we need a command to reflect the summary
We should meet and discuss this.

api/v1alpha1/openapi.yaml
asafbss: Yes agree, it makes issues with the CLI as well

internal/cli/get.go
avishayt: `GET /api/v1/devices` currently returns the following properties:
1. metadata (pagination info)
2. items (the list of devices)
3. summary (a tally of how many devices have each summary/update/application status)

The summary corresponds only to the devices returned in "items".

I propose to add a flag that will return "items" as empty.  This way, no pagination is necessary, and "summary" can easily reflect all devices (or any other group of devices, depending on the set filters).

asafbss: It will work, but what about CLI support to reflect the summary/overview?

avishayt: We can skip CLI for this IMO

fzdarsky: That approach makes sense to me. Need to define+document the behaviour if used in connection with `limit`, though.

Re the CLI: I'm thinking we could add a flag `--summary,-s` to print the summary after the table output (the table isn't meant to be machine-parseable anyway). And then `--summaryOnly` could only print that summary?

avishayt: Good points.  Passing any pagination, filters, etc. together with `--summaryOnly` should return 400.  We can add a flag to the CLI to print a table of the summary as you suggested.

internal/store/device.go
```diff
@@ -26,6 +26,7 @@ type Device interface {
 	Create(ctx context.Context, orgId uuid.UUID, device *api.Device, callback DeviceStoreCallback) (*api.Device, error)
 	Update(ctx context.Context, orgId uuid.UUID, device *api.Device, fieldsToUnset []string, fromAPI bool, callback DeviceStoreCallback) (*api.Device, error)
 	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.DeviceList, error)
+	Summary(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.DevicesSummary, error)
```
asafbss, line 29: @avishayt @fzdarsky, I have decided to keep the Device Summary for components that might use it.
The API of ListDevices will handle the case when "SummaryOnly" is specified and call the summary accordingly.
No new API is presented in this case.

internal/cli/get.go
```diff
@@ -120,10 +124,23 @@ func (o *GetOptions) Validate(args []string) error {
 		if kind != DeviceKind && kind != FleetKind {
 			return fmt.Errorf("owner can only be specified when fetching devices and fleets")
 		}
-		if (kind == DeviceKind || kind == FleetKind) && len(name) > 0 {
```
asafbss, line 123: There is no need to check the kind; it's already verified on the previous if condition.

internal/cli/get.go
```diff
@@ -120,10 +124,23 @@ func (o *GetOptions) Validate(args []string) error {
...
+		if len(name) > 0 {
 			return fmt.Errorf("cannot specify owner together with a device or fleet name")
 		}
 	}
+	if o.Summary || o.SummaryOnly {
```
asafbss, line 131: @avishayt @fzdarsky, please share your opinion about the current integration to the CLI.
Keep in mind that "get fleets/{name}" also has DevicesSummary inside the FleetStatus field, which only displays the device count in the chosen fleet.

asafbss:
New PR was created on top of the latest changes #549 
This one is too far behind and causing many conflicts.
@avishayt @ori-amizur @fzdarsky

PR #536, opened 2024-09-17
EDM-476: apply filters when listing enrollment requests

asafbss:


cmd/flightctl-api/main.go
```diff
@@ -39,6 +39,12 @@ func main() {
 	}
 	log.Printf("Using config: %s", cfg)
 
+	cfg.Database.Port = uint(8888)
+	cfg.Service.Address = ":3444"
+	cfg.Service.AgentEndpointAddress = ":7445"
+	cfg.Service.AgentGrpcAddress = ":7446"
+	cfg.Queue.AmqpURL = "amqp://user:password@127.0.0.1:5674"
+
```
fzdarsky, line 47: Accidental commit after debugging?

asafbss: Yes, my bad. I just reverted this file.
Thanks Frank

avishayt (changes requested):
I think this API would be more friendly if the query param was approved=true/false.  The backend can translate that to status.approval.approved.

celdrake:
> I think this API would be more friendly if the query param was approved=true/false. The backend can translate that to status.approval.approved.

The param change LGTM. However there's an issue ATM, that the current filter only works when obtaining the approved ERs. For the pending ones, no results are returned. I believe this may be caused by https://issues.redhat.com/browse/EDM-484

avishayt:
@asafbss Is this PR still relevant, or will it be changed to use the new query language?

asafbss:
@avishayt the new query language will handle it, I'm closing it.

PR #539, opened 2024-09-18
EDM-455: Simplify inline config API

avishayt:
The inline config API is now a native list of files rather than embedding ignition. Only the files portion of the ignition config was supported and it was cumbersome for users.


avishayt:
@fzdarsky , @hexfusion - do we need directories as well?  I'm not sure if they worked with the previous API.

examples/fleet-b.yaml
```diff
@@ -30,16 +30,10 @@ spec:
...
-              version: 3.4.0
-            storage:
-              files:
-                - contents:
-                    source: >-
-                      data:,This%20system%20is%20managed%20by%20flightctl.%0A
-                  mode: 422
-                  overwrite: true
-                  path: "/etc/motd"
+            - path: "/etc/motd"
```
celdrake, line 33: Could we also update the examples in the flightct-demos repo?
Since it's a breaking change, those examples won't work. (The UI won't either, but we will start working on it soon)

avishayt: Yes of course

api/v1alpha1/openapi.yaml
```diff
@@ -2539,10 +2539,42 @@ components:
...
+        contentsEncoding:
+          type: string
+          description: How the contents are encoded.
+          enum:
+          - base64
+          - url
+        mode:
+          type: integer
+          description: |
+            The file’s permission mode. You may specify the more familiar octal with a leading zero (e.g., 0644) or as
```
celdrake, line 2565: According to the description, the decimal value should work, but it returns error 400.
POST Fleet
```
{
  "apiVersion": "v1alpha1",
  "kind": "Fleet",
  "metadata": {
    "name": "last",
    "labels": {}
  },
  "spec": {
    "selector": {
      "matchLabels": {}
    },
    "template": {
      "metadata": {
        "labels": {
          "fleet": "last"
        }
      },
      "spec": {
        "config": [
          {
            "configType": "InlineConfigProviderSpec",
            "inline": [
              {
                "path": "/etc/motd",
                "contents": "whatever",
                "contentsEncoding": "base64",
                "mode": 777,
                "user": "camadorg",
                "group": "33"
              }
            ],
            "name": "ff"
          }
        ]
      }
    }
  }
}
```

Response:
```
Error 400: Bad Request - spec.config[].inline[0].mode: Invalid value: 604: must have 3 octal digits
```

Sending the decimal value seems like the best option for us, as otherwise looks like we'd need to do some conversions for being able to send the octal value as a number.

avishayt: 777 is octal, so it should be specified as 0777.

celdrake: I see, we can either convert to decimal ourselves, or send the value as `0o777` which gets converted to the decimal value. Just "0777" is not a valid value in Typescript. Either way, the API is working well, thanks for checking.

octal

api/v1alpha1/openapi.yaml
```diff
@@ -2539,10 +2539,42 @@ components:
...
+          description: The absolute path to the file on the device. Note that any existing file will be overwritten.
+        contents:
+          type: string
+          description: The URL-encoded or base64-encoded contents of the file.
+        contentsEncoding:
+          type: string
+          description: How the contents are encoded.
+          enum:
+          - base64
+          - url
```
fzdarsky, line 2561: I assume this is still inspired by Ignition, so the `url` would refer to the ability to pull in content from external sources via "http, https, tftp, s3, arn, gs" schemes. Is that the case and, if so, who would pull and would this conflict with the config providers?

Wondering whether we should limit ourselves to base64.

avishayt: No, it just specifies how the contents are encoded (either base64 encoding or URL encoding).  All of our examples currently use URL encoding so I assumed we wanted to support both.

api/v1alpha1/openapi.yaml
```diff
@@ -2539,10 +2539,42 @@ components:
...
+        user:
+          type: string
+          description: The file's owner, specified either as a name or numeric ID. Defaults to "root".
+        group:
+          type: string
+          description: The file's group, specified either as a name or numeric ID.
+      required:
+      - path
+      - contents
+      - contentsEncoding
```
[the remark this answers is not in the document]
avishayt: I started by detecting the encoding, but understood that in some (pathological) cases where this is ambiguous (i.e., there exist strings where both base64 and URL decoding are possible).

api/v1alpha1/openapi.yaml
fzdarsky: Ah, the data URI scheme `data:content/type;base64,` that Ignition prefixes, right? Do we need the ability to express MIME type, character encoding, etc? Could we autodetect that? Default that?

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
fzdarsky: Interesting. I assumed data URIs always have the `,` separator that is invalid bas64.

fzdarsky:
> @fzdarsky , @hexfusion - do we need directories as well? I'm not sure if they worked with the previous API.

Ignition supports them, though I wouldn't be able to come up with an example where this would be needed. Maybe to ensure a directory has the right mode+ownership before writing files to it.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
avishayt: We don't give the full data URL here - it's either URL-encoded text like "hello%20world%21" or base64-encoded like "aGVsbG8gd29ybGQh".

api/v1alpha1/openapi.yaml
avishayt: "Contents" now is just the data, without the data URL format.  Then "ContentsEncoding" tells us how the data is encoded so that it can later be properly decoded.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
fzdarsky: Ok, makes sense. Then we need a contentEncoding, of course. Should we default that, though?

The other Q I had: Do we need URL-encoding? IIRC, we used this to have some form of human-readable plain text and Ignition only allowed the URL-encoding. If we anyway define our own format, should that be plain text instead (in addition to base64)?

api/v1alpha1/openapi.yaml
```diff
@@ -2539,10 +2539,42 @@ components:
...
+                $ref: '#/components/schemas/FileSpec'
           required:
           - inline
+    FileSpec:
+      type: object
+      properties:
+        path:
+          type: string
+          description: The absolute path to the file on the device. Note that any existing file will be overwritten.
+        contents:
```
fzdarsky, line 2553: `content` (singular)?

api/v1alpha1/openapi.yaml
```diff
@@ -2539,10 +2539,42 @@ components:
...
+          type: string
+          description: The URL-encoded or base64-encoded contents of the file.
+        contentsEncoding:
```
fzdarsky, line 2556: `contentEncoding`?

api/v1alpha1/openapi.yaml
```diff
@@ -2539,10 +2539,42 @@ components:
...
+          description: |
+            The file’s permission mode. You may specify the more familiar octal with a leading zero (e.g., 0644) or as
+            a decimal without a leading zero (e.g., 420). Setuid/setgid/sticky bits are supported. If not specified,
+            the permission mode for files defaults to 0644.
+        user:
+          type: string
+          description: The file's owner, specified either as a name or numeric ID. Defaults to "root".
+        group:
+          type: string
+          description: The file's group, specified either as a name or numeric ID.
```
fzdarsky, line 2573: What does this default to?

hexfusion: What providers can populate user/group etc?

api/v1alpha1/openapi.yaml
avishayt: I agree with you, but copied from ignition.  Will change.

api/v1alpha1/openapi.yaml
avishayt: Also defaults to "root", will fix.
@hexfusion what do you mean by providers?

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
avishayt: You're right, it can be plain text with no need for URL encoding.

api/v1alpha1/openapi.yaml
hexfusion: Sorry let me rephrase what ConfigProviders allow setting group, user, perms for ignition/Files?

avishayt: Only this one.  Git/k8s secret/HTTP don't have a way to express ownership or permissions.  Does this suggest it's not needed here?  Or that we need to supplement the others?

hexfusion: Just an observation at this point I don't know the answer.

hexfusion: Is it even realistically possible to add to Git/HTTP? Secrets you could consider labels or annotations possibly

avishayt: Anything is possible :)
I mean you can extend our API, for example:
```
httpRef:
  repository: myrepo
  suffix: myfile.yaml
  filePath: /etc/foo
  owner: avishay
  group: avishay
  permissions: 0o600
```

But for now @fzdarsky suggested that we ignore this difference for now.  We need permissions for inline, and can consider how to add it for other config providers later on.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
avishayt: Removed the URL encoding.  Now it's either base64 or unencoded.

api/v1alpha1/openapi.yaml
avishayt: Fixed

api/v1alpha1/openapi.yaml
avishayt: Fixed

avishayt:
@fzdarsky @hexfusion Ready for another review

fzdarsky (commented):
Looks great. Just a few small things.

api/v1alpha1/openapi.yaml
```diff
@@ -2545,10 +2545,40 @@ components:
...
+          type: string
+          description: The absolute path to the file on the device. Note that any existing file will be overwritten.
+        content:
+          type: string
+          description: The URL-encoded or base64-encoded contents of the file.
+        contentEncoding:
+          type: string
+          description: How the contents are encoded. The content is assumed to be plain text if encoding is not specified.
+          enum:
+          - base64
```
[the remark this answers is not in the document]
avishayt: Fixed

api/v1alpha1/openapi.yaml
```diff
@@ -2545,10 +2545,40 @@ components:
...
+            the permission mode for files defaults to 0644.
+        user:
+          type: string
+          description: The file's owner, specified either as a name or numeric ID. Defaults to "root".
+        group:
+          type: string
+          description: The file's group, specified either as a name or numeric ID. Defaults to "root".
+      required:
+      - path
+      - content
```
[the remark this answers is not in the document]
avishayt: We can always make "content" optional.  I started, then realized the agent didn't support directories in the ignition and got into weeds for corner cases like removing directories that still have files.  We can always do it later.

api/v1alpha1/validation.go
```diff
@@ -100,6 +100,24 @@ func (c KubernetesSecretProviderSpec) Validate() []error {
...
+	for i := range c.Inline {
+		allErrs = append(allErrs, validation.ValidateString(&c.Inline[i].Path, fmt.Sprintf("spec.config[].inline[%d].path", i), 0, 2048, nil, "")...)
+		allErrs = append(allErrs, validation.ValidateLinuxUserGroup(c.Inline[i].User, fmt.Sprintf("spec.config[].inline[%d].user", i))...)
+		allErrs = append(allErrs, validation.ValidateLinuxUserGroup(c.Inline[i].Group, fmt.Sprintf("spec.config[].inline[%d].group", i))...)
+		allErrs = append(allErrs, validation.ValidateLinuxFileMode(c.Inline[i].Mode, fmt.Sprintf("spec.config[].inline[%d].mode", i))...)
+
+		if c.Inline[i].ContentEncoding != nil {
+			if *(c.Inline[i].ContentEncoding) == Base64 {
+				// Contents should be base64 encoded and limited to 1MB (1024*1024=1048576 bytes)
+				allErrs = append(allErrs, validation.ValidateBase64Field(c.Inline[i].Content, fmt.Sprintf("spec.config[].inline[%d].content", i), 1048576)...)
```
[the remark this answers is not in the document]
avishayt: Fixed

internal/util/validation/validation.go
```diff
@@ -92,6 +93,73 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
...
+
+	if isID {
+		// https://systemd.io/UIDS-GIDS/
+		if id >= 4294967295 {
+			errs = append(errs, field.Invalid(fieldPathFor(path), *s, "must be smaller than 4294967295 (invalid user ID)"))
+		}
+		if id == 65535 {
+			errs = append(errs, field.Invalid(fieldPathFor(path), *s, "must not be equal to 65535 (invalid user ID)"))
+		}
+		return asErrors(errs)
```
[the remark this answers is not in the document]
avishayt: Fixed

api/v1alpha1/openapi.yaml
```diff
@@ -2545,10 +2545,40 @@ components:
...
+        content:
+          type: string
+          description: The URL-encoded or base64-encoded contents of the file.
+        contentEncoding:
+          type: string
+          description: How the contents are encoded. The content is assumed to be plain text if encoding is not specified.
+          enum:
+          - base64
+        mode:
+          type: integer
```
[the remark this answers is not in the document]
avishayt: Yes it does

internal/util/validation/validation.go
```diff
@@ -92,6 +93,73 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
...
+
+	octal := strconv.FormatInt(int64(*m), 8)
+	if len(octal) != 3 {
+		errs = append(errs, field.Invalid(fieldPathFor(path), *m, "must have 3 octal digits"))
+	}
+
+	re := regexp.MustCompile(`^[0-7]+$`)
+	if !re.Match([]byte(octal)) {
+		errs = append(errs, field.Invalid(fieldPathFor(path), *m, "is not a valid mode"))
+	}
```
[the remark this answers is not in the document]
avishayt: Good call, fixed

api/v1alpha1/validation.go
```diff
@@ -100,6 +100,24 @@ func (c KubernetesSecretProviderSpec) Validate() []error {
 func (c InlineConfigProviderSpec) Validate() []error {
 	allErrs := []error{}
 	allErrs = append(allErrs, validation.ValidateGenericName(&c.Name, "spec.config[].name")...)
+	for i := range c.Inline {
+		allErrs = append(allErrs, validation.ValidateString(&c.Inline[i].Path, fmt.Sprintf("spec.config[].inline[%d].path", i), 0, 2048, nil, "")...)
```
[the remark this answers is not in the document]
avishayt: Done (added a helper and called it for all file paths)

avishayt:
@fzdarsky Addressed all comments

fzdarsky (approved):

PR #542, opened 2024-09-19
EDM-427: Device image enhancements

DakCrowder:
This PR seeks to:

1.  Fix an issue where passing a tag with image and digest to bootc results in an inability to upgrade

bootc errors when it is passed an image definition with both a tag AND digest specified.  In cases like this we should use the digest as it is the unique identifier and also follows what kubernetes does

2.  Adds an image digest to the device os status

This provides better clarity to users on what exact image is running on a device.

Notes:
`bootc` returns the following when `switch` is used targeting one of the local e2e images when a tag and digest is specified:
```
[user@localhost ~]$ sudo bootc switch 192.168.1.149:5000/flightctl-device:v3@sha256:fb91c3dd1c99c70512c262216770d749ac44d6a30c26aa018ada70c172262d8f
[  866.101616] SELinux:  Context unconfined_u:object_r:invalid_bootcinstall_testlabel_t:s0 is not valid (left unmapped).
ERROR Switching: Pulling: Creating importer: Failed to invoke skopeo proxy method OpenImage: remote error: Docker references with both a tag and digest are currently not supported
```


internal/agent/device/image/image.go
```diff
@@ -0,0 +1,104 @@
...
+
+func SpecToImage(spec *v1alpha1.DeviceOSSpec) *Image {
+	if spec == nil || spec.Image == "" {
+		return nil
+	}
+	return parseImage(spec.Image)
+}
+
+func BootcStatusToImage(bootc *container.BootcHost) *Image {
+	if bootc == nil {
```
hexfusion, line 68: In general, if a function returns nil without a non nil error we should be asking why. This ends up exposing lots of other things to the possibility of nil pointer. Go code expects that nil checks are done before you pass anything to a function that is maintained by a very boring pattern.

```
something, err := myFunc()
if err != nil {
  return err
 }
 ```
 here we can assume if the function return a non nil error that the returned value is not nil. This is repetitive and maybe ugly but it provides us with guardrails and reasonable assumptions.
 
 here I would argue that you should never be able to create bootc that is nil. If you can then that should get fixed.
 
 Of course, there are exceptions when we are dealing with serialization etc I think you get what I am saying :)

File: internal/agent/device/device.go [hunk 2]
```diff
@@ -199,8 +199,13 @@ func (a *Agent) syncDevice(ctx context.Context) (bool, error) {
 	}
 
 	if desired.Os != nil {
+		bootedImage, err := a.osImageController.GetCurrentBootcImage(ctx)
```
hexfusion, line 202: as I noted the controller should not expose logic for reuse. If the device needs access to a bootc client then give it one. The reason is we don't want to create cross dependencies on controllers. A controllers only job is to reconcile the spec that it owns.

internal/agent/device/osimage.go
```diff
@@ -58,24 +59,24 @@ func (c *OSImageController) ensureImage(ctx context.Context, desired *v1alpha1.R
...
+	if reconciled {
 		c.log.Debugf("Host is reconciled to os image %s", desired.Os.Image)
 		return nil
 	}
 
-	image := desired.Os.Image
-	c.log.Infof("Switching to os image: %s", image)
-	if err := c.bootc.Switch(ctx, image); err != nil {
+	c.log.Infof("Switching to os image: %s", desired.Os.Image)
+	target := desiredImage.ToBootcTarget()
```
hexfusion, line 74: My understanding is we are converting the spec to image then back to something bootc can consume. But it really has what it needed before :). I could be missing something.

internal/agent/device/image/image.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/container"
+)
+
+type Image struct {
+	Base   string
+	Tag    string
+	Digest string
+}
```
hexfusion, line 15: Solving problems with types is the way but in this case we should really look at if we need this.

hexfusion (commented):
@DakCrowder thanks for the PR! A few notes to mull over I think the approach was reasonable approach. But maybe we can simplify just a bit?

internal/container/container.go
```diff
@@ -169,10 +165,34 @@ func (b *BootcHost) GetBootedImage() string {
...
+
 func (b *BootcHost) GetStagedImage() string {
 	return b.Status.Staged.Image.Image.Image
 }
 
 func (b *BootcHost) GetRollbackImage() string {
 	return b.Status.Rollback.Image.Image.Image
 }
+
+// Bootc does not accept images with tags AND diegsts specified - in the case when we
```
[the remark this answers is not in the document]
DakCrowder: I did a bit of digging (can definitely still raise an issue) but it looks like the ultimate "root cause" for the failure is mentioned in this issue: https://github.com/containers/image/issues/1736 - interestingly it also looks like a ref with a name only may also fail, will test that scenario
https://github.com/containers/image/blob/d372f0e440d35c6041de39023b0b6eb131fba54b/docker/docker_transport.go#L70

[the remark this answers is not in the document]
hexfusion: The error looks like its skopeo under the hook throwing the error via bootc as well.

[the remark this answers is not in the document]
DakCrowder: Yep, I should've added more context first found this issue which points to the one I posted above - https://github.com/containers/skopeo/issues/2367

api/v1alpha1/openapi.yaml
```diff
@@ -3000,10 +3000,14 @@ components:
...
       required:
         - image
+        - imageDigest
       properties:
         image:
           type: string
           description: "Version of the OS image."
+        imageDigest:
+          type: string
+          description: "The digest of the OS image (e.g. sha256:a0...)"
```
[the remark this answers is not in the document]
hexfusion: @celdrake WDYT? We want to do our best about reporting the digest in cases where there is a static tag like the latest and there could be divergence between the fleet.  This could be data that is show on hover or could just be used for debug,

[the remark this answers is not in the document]
celdrake: I agree that we should display this, as it's an important info important to know.
I'll discuss with Nir whether we want to display it by default or on hover, etc.

internal/agent/device/bootstrap.go
```diff
@@ -201,9 +206,15 @@ func (b *Bootstrap) ensureBootedOS(ctx context.Context, desired *v1alpha1.Render
 		return fmt.Errorf("writing current rendered spec: %w", err)
 	}
 
+	bootcStatus, err := b.bootcClient.Status(ctx)
```
hexfusion, line 209: as a followup in agent.go when we create the bootc client lets fail fast if bootc does not exist.

hexfusion (approved):

hexfusion (approved):

internal/container/container.go
```diff
@@ -90,7 +95,7 @@ func (b *BootcCmd) Status(ctx context.Context) (*BootcHost, error) {
 	}
 
 	var bootcHost BootcHost
-	if err := yaml.Unmarshal([]byte(stdout), &bootcHost); err != nil {
+	if err := json.Unmarshal([]byte(stdout), &bootcHost); err != nil {
```
DakCrowder, line 98: We request the status in json - yaml unmarshal was actually _mostly_ working but failed for certain fields / cases

hexfusion (approved):
:vulcan_salute: nice work, lgtm

PR #544, opened 2024-09-23
EDM-431: Autoapprove CSRs upon creation

kot-begemot-uk:
Presently, CSR creation is restricted to authenticated users. This sets them to autoapproved as creation is already controlled by authentication.

### DIFF
--- a/internal/service/certificatesigningrequest.go
+++ b/internal/service/certificatesigningrequest.go
@@ -110,9 +110,21 @@ func (h *ServiceHandler) ListCertificateSigningRequests(ctx context.Context, req
 func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, request server.CreateCertificateSigningRequestRequestObject) (server.CreateCertificateSigningRequestResponseObject, error) {
 	orgId := store.NullOrgId
 
-	// don't set fields that are managed by the service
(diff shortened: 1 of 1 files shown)


internal/service/certificatesigningrequest.go
```diff
@@ -110,10 +110,22 @@ func (h *ServiceHandler) ListCertificateSigningRequests(ctx context.Context, req
...
 
+	signedCert, err := signApprovedCertificateSigningRequest(h.ca, *request.Body)
+	if err != nil {
+		return nil, err
+	}
+	approvedCondition := api.Condition{
+		Type:    api.CertificateSigningRequestApproved,
+		Status:  api.ConditionStatusTrue,
+		Reason:  "Approved",
+		Message: "Approved",
```
hexfusion, line 123: Would it make sense to provide the signal that this is an `AutoApproved` CSR? Maybe that is not important at this point.

kot-begemot-uk: Yes, but in a different way. 
This should be recorded in "who approved". If we do not have "who approved" our approval data structure needs fixing.
Let me have a look at what do we have there.

lkatalin:
This is superseded by #529

PR #547, opened 2024-09-26
EDM-519: add better handling of obsolete files

hexfusion:
This PR
- fixes a bug in the config controller where the desired config if nil did not result in removal of all current files in config.
- replaced the config controller with the hooks manager for firing after update hooks during bootstrap
- unexported all methods in config controller except Sync.
- added and extended tests for fileio pkg.
- minor refactoring


internal/agent/device/fileio/managed_file.go
```diff
@@ -2,47 +2,49 @@ package fileio
...
-	"path/filepath"
 
 	ign3types "github.com/coreos/ignition/v2/config/v3_4/types"
 )
 
+var ErrPathIsDir = errors.New("provided path is a directory")
+
 type managedFile struct {
 	ign3types.File
-	initialized bool
```
hexfusion, line 14: note: since we now `initialize` on new there is no need to track this anymore

internal/agent/device/fileio/managed_file.go
```diff
@@ -2,47 +2,49 @@ package fileio
...
 	if err != nil {
 		if os.IsNotExist(err) {
-			m.initialized = true
 			return nil
 		}
 		return err
 	}
 	if fileInfo.IsDir() {
-		return fmt.Errorf("provided path %q is a directory", m.Path())
+		return fmt.Errorf("%w: %s", ErrPathIsDir, path)
```
hexfusion, line 44: note: wrapping error types allows for more fine-grained testing.

internal/agent/device/config/controller_test.go
```diff
@@ -80,55 +95,98 @@ func TestSync(t *testing.T) {
...
-			for i, f := range lo.Without(tt.createdFiles, tt.removedFiles...) {
-				mockWriter.EXPECT().CreateManagedFile(gomock.Any()).Return(mockManagedFile)
-				upToDate := i%2 == 0
-				mockManagedFile.EXPECT().IsUpToDate().Return(upToDate, nil)
-				if !upToDate {
-					mockManagedFile.EXPECT().Exists().Return(true, nil)
-					mockManagedFile.EXPECT().Write().Return(nil)
-					mockHookManager.EXPECT().OnBeforeUpdate(gomock.Any(), f)
-					mockHookManager.EXPECT().OnAfterUpdate(gomock.Any(), f)
-				}
```
hexfusion, line 102: note: testing for update hooks was moved to IsUpToDate testing in fileio

internal/agent/device/config/controller.go
```diff
@@ -14,15 +14,9 @@ import (
 	"github.com/flightctl/flightctl/pkg/log"
 )
 
-type Controller interface {
-	Initialize(ctx context.Context, current *v1alpha1.RenderedDeviceSpec)
-	Sync(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error
-	WriteIgnitionFiles(ctx context.Context, files []ignv3types.File) error
```
hexfusion, line 20: note: exposing this meant the bootstrap controller had access to write ign. this was removed to ensure the only writes of files are the spec manager and config controller during Sync.

internal/agent/device/fileio/fileio.go
```diff
@@ -16,11 +15,11 @@ type ManagedFile interface {
 
 type Writer interface {
 	SetRootdir(path string)
-	WriteFileBytes(name string, data []byte, perm os.FileMode) error
```
hexfusion, line 19: note: this was an unused more or less dupe of WriteFile

internal/agent/agent.go
```diff
@@ -146,21 +146,14 @@ func (a *Agent) Run(ctx context.Context) error {
...
-
 	bootstrap := device.NewBootstrap(
 		deviceName,
 		executer,
 		deviceReadWriter,
 		csr,
 		specManager,
 		statusManager,
-		configController,
+		hookManager,
```
hexfusion, line 156: note: shared functionality for hooks should be exposed by the hookManager not by its consumer.

internal/agent/device/config/controller.go
```diff
@@ -126,17 +98,35 @@ func (c *controller) ensureConfigData(ctx context.Context, currentData, desiredD
...
-func (c *controller) WriteIgnitionFiles(ctx context.Context, files []ignv3types.File) error {
+// removeObsoleteFiles removes files that are present in the currentFiles but not in the desiredFiles.
+func (c *Controller) removeObsoleteFiles(ctx context.Context, currentFiles, desiredFiles []ignv3types.File) error {
+	removeFiles := computeRemoval(currentFiles, desiredFiles)
+	for _, file := range removeFiles {
+		c.log.Infof("Deleting file: %s", file)
+		// trigger delete pre hook and wait for it to complete
+		c.hookManager.OnBeforeRemove(ctx, file)
+		if err := c.deviceWriter.RemoveFile(file); err != nil {
+			return fmt.Errorf("deleting files failed: %w", err)
```
DakCrowder, line 117: If the file we want to remove doesn't exist is that worth catching and moving forward with?

hexfusion: Logically what you're saying makes sense, don't block progress for a successful operation.  RemoveFile is fairly smart and filtering those errors.  SO an error here would be pretty unexpected so I think its valid to report the agent as degraded in this case.

https://github.com/flightctl/flightctl/blob/1016e2c165d34c813cd6b0d60a742d283009399c/internal/agent/device/fileio/writer.go#L63

DakCrowder (approved):

PR #548, opened 2024-09-29
EDM-441: A missing device spec file should always cause an error status

ori-amizur:
During bootstrap both desired spec and the current spec are read.  In case there is an error to read the current spec, the error is ignored. The fixes it, so in case there is an error to read the current spec, an error is returned.


hexfusion:
@ori-amizur sorry I missed this, the code has already been updated I think we can close this?

PR #549, opened 2024-09-29
adding devices summary API

asafbss:
This PR supersedes #535 (https://github.com/flightctl/flightctl/pull/535).
The old branch had fallen behind.

PR #558, opened 2024-10-02
EDM-540: Fix validateAgainstSchema causing 400 errors on non-nullable…

asafbss:
Patching a fleet doesn't work because the status fields are marked as Required (merged in #549).

The process fails with the error: "doesn't match schema #/components/schemas/Fleet: Error at \"/status/devicesSummary/applicationStatus\": Value is not nullable."

By the following validation:
https://github.com/flightctl/flightctl/blob/main/internal/service/common.go#L82


asafbss:
@celdrake

avishayt (approved):

avishayt (approved):

PR #562, opened 2024-10-04
EDM-554: e2e images: enable podman.service

hexfusion:
This PR enables the podman service for our e2e images which . This is required for podman-compose to work correctly as the root user. Specifically operations such as podman up -d which should run the application in the background will block, causing unexpected operations. While podman can run daemonless for it to work with root user correctly requires the daemon to start and expose API's via podman.sock.


avishayt (approved):

hexfusion:
> > > Specifically operations such as podman up -d which should run the application in the background will block, causing unexpected operations
> 
> > 
> 
> > Will this block the agent entirely? We should be careful with user inputs that can cause the agent to fail.
> 
> 
> 
> We should detect such issue and avoid it from happening, throwing an error for the user to know. 
> 
> 
> 
> 

This is detectable as a beforeUpdate validation. Will add to applications PR

PR #563, opened 2024-10-04
EDM-555: add resource monitor API validation

hexfusion:
This PR adds `ResourceMonitor` validation rules and fixes a bug where we allow defining time duration in days which is not supported. Instead of adding that support, it makes sense to allow the largest interval to be in hours as days are not percise.

Ref: golang/go#17767


internal/util/validation/validation.go
```diff
@@ -282,3 +282,18 @@ func asErrors(errs field.ErrorList) []error {
 	}
 	return agg.Errors()
 }
+
+// CheckUnique checks for duplicate values in a slice of items.
+func CheckUnique[T any](slice []T, fieldFn func(T) string) error {
```
avishayt, line 287: Where is this used?

avishayt (commented):
Overall looks good.  What was the issue with days?

internal/util/validation/validation.go
hexfusion: Meant to remove

hexfusion:
> Overall looks good.  What was the issue with days?

Days are not constant length in every location.

https://github.com/golang/go/issues/17767

avishayt:
> > Overall looks good.  What was the issue with days?
> 
> Days are not constant length in every location.
> 
> [golang/go#17767](https://github.com/golang/go/issues/17767)

Interesting - thanks!

internal/util/validation/validation.go
hexfusion: Done

avishayt (approved):

PR #564, opened 2024-10-05
Reduce backoff durations and log slow API requests

hexfusion:
We have seen situations where the agent will appear offline because it has not reported status for more than 5 minutes while the agent appears to be working correctly.  One theory  is if the service is not responding the agent will retry for up to an hour. This would block status updates during that time. The updated backoff maxes out under 3 minutes and logs requests taking over a minute.

This PR also removed unneeded locks from status.


avishayt:
Have you been able to reproduce the issue?  Does this fix it?  Is that why it's a draft?

hexfusion:
QE is validating

siserafin:
After 24h of testing the PR with a device created with make agent-vm, the longest sync duration that I see is 447.481481ms.

hexfusion:
We want to merge for further testing but this has looked promising in my local testing as well as QE.

avishayt (approved):

PR #565, opened 2024-10-06
EDM-447: Add fleet policy in openapi

ori-amizur:
Add rollout policy in openapi


api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,82 @@ components:
...
       description: FleetList is a list of Fleets.
+
+    DisruptionAllowance:
+      type: object
+      properties:
+        groupBy:
+          type: array
+          items:
+            type: string
+          description: List of labels to perform grouping for the disruption allowance.
```
avishayt, line 3443: Change to "selector" and use our LabelSelector definition

api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,82 @@ components:
...
+      items:
+        $ref: '#/components/schemas/Batch'
+      description: BatchSequence defines the list of batches to be executed in sequence.
+
+    RolloutPolicy:
+      type: object
+      properties:
+        disruptionAllowance:
+          $ref: '#/components/schemas/DisruptionAllowance'
+        batchSequence:
```
avishayt, line 3494: We said that we'll have a property to allow different strategies in the future, other than batchSequence.

api/v1alpha1/openapi.yaml
ori-amizur: This is group by keys.  It does not contain label values.

api/v1alpha1/openapi.yaml
ori-amizur: Currently it is kind of 'dangling' property since the only strategy is batch.  I think we should have it once we have additional strategy.

api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,88 @@ components:
...
+      properties:
+        deviceSelectionStrategy:
+          type: string
+          enum: ['batch']
+          default: 'batch'
+        disruptionAllowance:
+          $ref: '#/components/schemas/DisruptionAllowance'
+        batchSequence:
+          $ref: '#/components/schemas/BatchSequence'
+        defaultSuccessThreshold:
```
hexfusion, line 3502: a default is something that can't change, what do you think about dropping that prefix?
```suggestion
        successThreshold:
```

api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,88 @@ components:
...
+          type: integer
+          description: The maximum number of unavailable devices allowed during rollout.
+        maxUnavailable:
+          type: integer
+          description: The minimum number of required available devices during rollout.
+      description: DisruptionAllowance defines the level of allowed disruption when rollout is in progress.
+
+    Percentage:
+      type: string
+      format: '^(?:(?:[1-9][0-9]?)|100|0)%$'
```
hexfusion, line 3454: optional: Slightly simpler 
```suggestion
      format: '^(100|[1-9]?[0-9])%$'
```

api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,88 @@ components:
...
+          $ref: '#/components/schemas/Percentage'
+        limit:
+          oneOf:
+            - $ref: '#/components/schemas/Percentage'
+            - type: integer
+              minimum: 1
+      description: Batch is an element in batch sequence.
+    Duration:
+      type: string
+      pattern: '^[1-9]\d*[smhd]$'
```
hexfusion, line 3472: we had a bug here that I introduced time duration doesn't parse to days and most time based tools such as openssl user hours. I would consider dropping `d`  days.

api/v1alpha1/openapi.yaml
ori-amizur: Each batch can have its own success threshold and this is the default in case it is not specified for a batch.

api/v1alpha1/openapi.yaml
ori-amizur: OK

api/v1alpha1/openapi.yaml
ori-amizur: We need days units.  The format is simple, so in case it doesn't work elsewhere, we will implement it.

api/v1alpha1/openapi.yaml
fzdarsky: Avishay's right, We cannot have `batchSequence` at this level, it belongs below the selection strategy. How about
```
rolloutPolicy:
  selectionStrategy:
    type: batchSequence
    sequence:
    - ...
```

api/v1alpha1/openapi.yaml
fzdarsky: I guess it depends on whether we want to allow selecting by group (key) or also specific location. With a selector that could be `store=Madrid` or just `store` (existence op). What would an example `groupBy` be? Would the list items be OR'ed or AND'ed?

api/v1alpha1/openapi.yaml
ori-amizur: We agreed on device selection strategy on the top level of rollout policy.
I added it here: https://github.com/flightctl/flightctl/compare/e74eda546f0ed592cb7f1e0364a7d32c9e3afca5..2c4662d3ad83abae5a0df1bbbf8e64e8d1fc53cd

ori-amizur: Although I can change to what @fzdarsky suggested

api/v1alpha1/openapi.yaml
ori-amizur: This is not a selector.  It is used for disruption allowance.
Its purpose is that we want to define (for example) that we don't want (for any site) that the number of unavailable devices will be greater than the unavailable devices as a result of a rollout.
If we want a specific site to be selected then it needs to be defined in the batch sequence definition.

ori-amizur: Think of it like the `group by` clause in aggregate SQL select statement.  Only field names are defined there.

api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,88 @@ components:
...
+        disruptionAllowance:
+          $ref: '#/components/schemas/DisruptionAllowance'
+        batchSequence:
+          $ref: '#/components/schemas/BatchSequence'
+        defaultSuccessThreshold:
+          $ref: '#/components/schemas/Percentage'
+        defaultBatchTimeout:
+          $ref: '#/components/schemas/Duration'
+        defaultUpdateTimeout:
+          $ref: '#/components/schemas/Duration'
```
fzdarsky, line 3507: I'd vote for that we hold off from adding these until their definition and reason for existence is clearer (see my comments on the design doc).

ori-amizur: We need timeouts for batches.  Otherwise they will never finish.

api/v1alpha1/openapi.yaml
```diff
@@ -3432,11 +3432,97 @@ components:
...
       description: FleetList is a list of Fleets.
+
+    DisruptionAllowance:
+      type: object
+      properties:
+        groupBy:
+          type: array
+          items:
+            type: string
+          description: List of labels to perform grouping for the disruption allowance.
```
avishayt, line 3443:
```suggestion
          description: List of label keys to perform grouping for the disruption allowance.
```

ori-amizur: OK

api/v1alpha1/openapi.yaml
```diff
@@ -3438,11 +3438,97 @@ components:
...
+      type: object
+      properties:
+        disruptionAllowance:
+          $ref: '#/components/schemas/DisruptionAllowance'
+        deviceSelection:
+          $ref: '#/components/schemas/RolloutDeviceSelection'
+        successThreshold:
+          $ref: '#/components/schemas/Percentage'
+        defaultBatchTimeout:
+          $ref: '#/components/schemas/Duration'
```
avishayt, line 3520: Would these go under BatchSequence?  Because they describe the success threshold and timeout for a batch?

api/v1alpha1/openapi.yaml
avishayt: Does this describe the timeout for a batch or an individual device?

api/v1alpha1/openapi.yaml
ori-amizur: OK

api/v1alpha1/openapi.yaml
ori-amizur: Removed the `defaultBatchDuration`

avishayt (approved):

avishayt (approved):

PR #567, opened 2024-10-07
EDM-546: Fix device spec comparisons

avishayt:
There are a couple instances where we need to understand if device specs are being updated by the user.  reflect.DeepEqual doesn't work properly here because oneOf is translated to json.RawMessage, and comparing JSONs isn't reliable because properties aren't always sorted the same way. Instead, we use reflect.DeepEqual where possible, and unmarshal when necessary.

### DIFF
--- a/internal/store/common.go
+++ b/internal/store/common.go
@@ -4,12 +4,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
+	"slices"
 	"strings"
 
+	api "github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/store/model"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/google/uuid"
+	"github.com/samber/lo"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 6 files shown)


internal/store/common.go
```diff
@@ -297,3 +299,158 @@ func retryUpdate(fn func() (bool, error)) error {
...
+
+func configsAreEqual(c1, c2 *[]api.DeviceSpec_Config_Item) bool {
+	if c1 == nil {
+		return c2 == nil
+	}
+	if c2 == nil {
+		return false // c2 is nil, c1 is not
+	}
+
+	if len(*c1) != len(*c2) {
```
ori-amizur, line 311: The first 2 conditions can be skipped and use only this one

internal/store/common.go
```diff
@@ -297,3 +299,158 @@ func retryUpdate(fn func() (bool, error)) error {
 	}
 	return err
 }
+
+func configsAreEqual(c1, c2 *[]api.DeviceSpec_Config_Item) bool {
```
ori-amizur, line 303: Maybe we can use this generic func https://pkg.go.dev/slices#EqualFunc as an envelope for all the functions?

avishayt: Nice.  Done.

internal/store/common.go
avishayt: Fixed.

ori-amizur (approved):

PR #570, opened 2024-10-08
EDM-567: Do not trigger full tests on docs changes

fzdarsky:


lkatalin (approved):
lgtm

mangelajo:
I am not sure if it will work with github expecting some of those runs as required hmmm but we can try.

PR #571, opened 2024-10-08
EDM-531: Docs for ACM registration

oglok:


docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
+# Auto-Registering Devices with MicroShift into ACM
```
fzdarsky, line 1: Can you please link from the README.md to this doc?

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Configuring RHACM Agent Registration
```
fzdarsky, line 3: In general, let's spell out acronyms once "Advanced Cluster Management (ACM)" before using. In the heading, this would be too much, of course.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+          suffix: /agent-registration/manifests/{{ device.metadata.name }}
+        name: acm-import
+      hooks:
+        afterUpdating:
+          - path: "/var/local/crd.yaml"
+            onFile: [Create]
+            actions:
+              - executable:
+                  run: "kubectl apply -f /var/local/crd.yaml"
+                  envVars: ["KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig"]
```
fzdarsky, line 75: I think this needs to be in `key: value` format:
```
envVars:
  KUBECONFIG: /var/lib/microshift/resources/kubeadmin/kubeconfig
```

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+                  workDir: "/usr/bin/"
```
fzdarsky, line 76: This should not be necessary. Can we remove it or is this a workaround for a bug we need to fix?

docs/user/registering-microshift-devices-acm.md
fzdarsky, line 3: The user documentation follows a specific pattern that we should also adopt for this section (but we can update that in a follow-up PR):

1. First give an overview of what the section is about ("If you have a fleet of devices running MicroShift and workloads on MicroShift to be managed through ACM, you can set up the fleet so that MicroShift instances get automatically registered into ACM.") and a general explanation of concepts+processes.
2. Implementing this using the UI.
3. Implementing this using the CLI.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
+# Auto-Registering Devices with MicroShift into ACM
+
+## Configuring RHACM Agent Registration
+
+Auto-registration of devices with MicroShift into ACM relies on a feature called agent registration, allowing clusters to be imported via REST API calls using a CA bundle and token. Follow these [instructions](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.11/html/clusters/cluster_mce_overview#importing-managed-agent) to configure the necessary RBAC policies for your own user. Skip step 7 of the instructions in section 1.6.5.2.3, as this step will be automatically handled by the fleet devices.
```
fzdarsky, line 5: Section 1.6.5.**3**.3. Referencing a section (or even step) by number will frequently break, though. Can we just describe what needs to be skipped ("Skip the step to import the managed cluster to the hub cluster").

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+
+As part of this setup, you will need to collect the following information:
+
+- Agent registration URL
+- Token
+
+## Creating Repositories in Flight Control
```
fzdarsky, line 12: How to use the HTTP Config Provider is already explained in section
https://github.com/flightctl/flightctl/blob/main/docs/user/managing-devices.md#getting-configuration-from-an-http-server. I see it lacks the generic description of how to set up a corresponding Repo.

So here we shouldn't repeat how to set up repo or configure the provider, but just tell the user that these need to be created with X and Y as parameters.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+  labels: {}
+  name: acm-test
+spec:
+  httpConfig:
+    token: $token
+  type: http
+  url: https://$agent_registration_host
+```
+
+Note that this repository will show as not valid, because the root URL does not return a value directly. However, this endpoint will be used in the fleet spec configuration section with a set of suffixes that will return the required Kubernetes manifests for auto-registration.
```
fzdarsky, line 31: Do we have a bug report for that? Can we make this an alert?
```
> [!NOTE] This repository will show ...
```

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+        name: acm-crd
+      - configType: HttpConfigProviderSpec
+        httpRef:
+          filePath: /var/local/import.yaml
+          repository: acm-test
+          suffix: /agent-registration/manifests/{{ device.metadata.name }}
+        name: acm-import
+      hooks:
+        afterUpdating:
+          - path: "/var/local/crd.yaml"
```
fzdarsky, line 70: I'm not sure what the right location for this is. I would have said more under `/run` or `/tmp`, but IDK whether that would work. Anyway, we shouldn't write random files to a shared location like `/var/local` but to a dedicated directory. If we used, say, `/var/local/klusterlet-bootstrap` instead, we could have a single update action that applies the whole dir here.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+                  envVars: ["KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig"]
+                  workDir: "/usr/bin/"
+```
+
+### Fleet Specification Breakdown
+
+The fleet specification is composed of three main sections: OS, config and hooks.
+
+1. **Operating System (OS) Image**: This section defines the target Operating System image required for all devices, in this case an image containing the MicroShift runtime. This is essential for running the Klusterlet, the agent responsible for communication with ACM.
+
```
fzdarsky, line 91: I'd just make this a pre-requisite at the beginning, so you don't need to repeat explanations. Like "Prerequisite: * You have a fleet of devices running an OS image containing MicroShift."

fzdarsky (commented):
There are a few small things we should fix while at it. Any larger adjustments we can leave for a follow-up PR.

docs/user/registering-microshift-devices-acm.md
oglok: +1

docs/user/registering-microshift-devices-acm.md
oglok: it works this way, not with key:value format. I just tested it.

docs/user/registering-microshift-devices-acm.md
oglok: Let's do it in a follow up PR, because the UI is not ready for hooks yet.

docs/user/registering-microshift-devices-acm.md
oglok: +1

docs/user/registering-microshift-devices-acm.md
oglok: Ok, makes sense.

docs/user/registering-microshift-devices-acm.md
oglok: Having a single action could cause problems if order is not guaranteed. If kubectl applies the CR before the CRD is created this will fail.

docs/user/registering-microshift-devices-acm.md
oglok: I will highlight it with **_NOTE:_**

This is something I missed when designing the solution, that if you configure a root URL that doesn't return data it will be shown as not valid. I need to talk to @avishayt about this.

docs/user/registering-microshift-devices-acm.md
oglok: +1

docs/user/registering-microshift-devices-acm.md
fzdarsky: Sigh. Something `kubectl` should be smart enough to handle.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
+# Auto-Registering Devices with MicroShift into ACM
+
+## Pre-requisite
+
+The Operating System image running on our fleet of devices must contain MicroShift. This is essential for running the Klusterlet, the agent responsible for communication with Red Hat Advanced Cluster Management (ACM)
```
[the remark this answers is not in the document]
oglok: This should already be learned here: https://github.com/flightctl/flightctl/blob/main/docs/user/building-images.md
It's just dnf installing an rpm in a Containerfile.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+Auto-registration of devices with MicroShift into ACM relies on a feature called agent registration, allowing clusters to be imported via REST API calls using a CA bundle and token. Follow these [instructions](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.11/html/clusters/cluster_mce_overview#importing-managed-agent) to configure the necessary RBAC policies for your own user. Skip running the command to import the managed clusters, as this step will be automatically handled by the fleet devices.
+
+As part of this setup, you will need to collect the following information:
+
+- Agent registration URL
+- Token
+
+## Creating Repositories in Flight Control
+
+A new repository must be created with the data collected before from ACM's agent registration endpoint. This repository will be used by the fleet configuration using the HTTP config provider described in the following [link](https://github.com/flightctl/flightctl/blob/main/docs/user/managing-devices.md#getting-configuration-from-an-http-server).
```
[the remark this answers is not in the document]
oglok: It's described in the previous section where there is a link to the ACM's documentation. There, the user will have to execute a set of steps once, and as a result, the agent registration endpoint and the token must be stored.

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+            workDir: "/usr/bin/"
+```
+
+Flight Control hooks are explained in the following [link](https://github.com/flightctl/flightctl/blob/main/docs/user/managing-devices.md#using-device-lifecycle-hooks).
+Hooks are triggered at specific moments on the lifecycle of a device such as before/after updating, before/after reboot, etc.
+The hooks definition shown above describe two hooks executed after updating the fleet, one to apply the CRD file and one to apply the rest of the Klusterlet manifests stored in the import.yaml file.
+
+## Device Reconciliation Process
+
+Once devices are enrolled in Flight Control and assigned to a fleet definition like the one above, the agent will begin a reconciliation process as outlined in the project documentation. This process ensures each device is updated with the required OS image, configuration and applications. In this specific workflow, it ensures all steps required to register a MicroShift cluster into ACM. From updating the Operating System image to one containing MicroShift and its dependencies, getting the Klusterlet manifests into disk and apply them.
```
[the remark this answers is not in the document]
oglok: True, it sounds funny! I'll reword it. Thanks!

docs/user/registering-microshift-devices-acm.md
```diff
@@ -0,0 +1,157 @@
...
+
+    This API call to ACM's agent registration endpoint will retrieve a set of Kubernetes manifests to deploy the Klusterlet. Once we have both the Klusterlet CRD and deployment manifests, we will use Flight Control hooks to apply them.
+
+1. **Hooks**: Once the configuration is fetched, the hooks apply the Kubernetes manifests to the device:
+
+```console
+hooks:
+  afterUpdating:
+    - path: "/var/local/crd.yaml"
+      onFile: [Create]
```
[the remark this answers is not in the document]
oglok: Hooks might change, so we will have to update this documentation, but in the meantime, this works. 

MicroShift has the capability to apply manifests at a certain location `/etc/microshift/manifests`, and this will happen everytime Microshift systemd service starts (so at every reboot). Once Klusterlet agent is started in the MicroShift cluster, ACM will be in charge for managing it. This is why this solution didn't work. 

The reboot policy would have the same behavior as the MicroShift applier, so I didn't want to use that, because it would bring up the same questions.

In the demo I'm recording, I basically run devices with a bootstrap image with flightctl-agent and microshift, so image is already reconciled... that would allow hooks to work because microshift binary is already present in the device.

fzdarsky (approved):

PR #573, opened 2024-10-08
MGMT-18492: Update Hooks API

fzdarsky:
This PR updates the Hooks API to radically simplify it and aligning it with the original design in #424:

```
hooks:
  beforeUpdating:
  - run: /usr/local/bin/create-backup.sh  # action
  beforeRebooting:
  - run: /usr/local/bin/shutdown-service.sh
  afterUpdating:
  - if:  # condition
      path: "/etc/systemd/system/"
      op: [create]       
    run: /usr/bin/kubectl apply -f "{{ .FilePath }}" # action
  - if:
      path: "/etc/systemd/system/"
      op: [delete] # defined as "before" deleting by convention
    run: /usr/bin/kubectl remove -f "{{ .FilePath }}" # action
  - if:
      path: "/var/local/import.yaml"
      op: [create]
    run: /usr/bin/kubectl apply -f /var/local/crd.yaml
    envVars:
      KUBECONFIG: /var/lib/microshift/resources/kubeadmin/kubeconfig
    timeout: 10s
    workDir: /
```


api/v1alpha1/openapi.yaml
```diff
@@ -2784,95 +2749,68 @@ components:
...
+              type: string
+              pattern: '^[1-9]\d*[smhd]$'
+              description: |
+                The maximum duration allowed for the action to complete. 
+                The duration should be specified as a positive integer 
+                followed by a time unit. Supported time units are:
+                - 's' for seconds
+                - 'm' for minutes
+                - 'h' for hours
+                - 'd' for days
```
hexfusion, line 2774: We should consider dropping days as they are not consistent measurements of time.

https://github.com/golang/go/issues/17767#issuecomment-258167205

api/v1alpha1/openapi.yaml
```diff
@@ -2784,95 +2749,68 @@ components:
...
-            workDir:
-              type: string
-              description: 'The directory in which the executable will be run from if it is left empty it will run from the users home directory.'
-          required:
-            - run
+          oneOf:
+            - $ref: '#/components/schemas/HookConditionFileOp'
+            # extend hook conditions
+      required:
+        - if
```
fzdarsky, line 2800: Do not require if.

api/v1alpha1/openapi.yaml
lkatalin: If we are not supporting days, this will impact the CSR CLI where we decided to support them to the exclusion of other units.

avishayt: Can we define a day to be 24 hours?

ori-amizur:
Couple of questions:

Does it make sense to have `beforeUpdating` hook (like the example in the description) with any reference to any path?  When will it run? on every file modification?

In the same example, even when there is a hook without a path - the `run` always accesses some file.  Therefore, maybe a hook should always be bound some file?

It seems to me the separation between the `beforeUpdate`, `afterUpdate` etc it somewhat artificial. The hooks are always run with a context of [before, after] and usually also with [create,update,delete, reboot].  So maybe this structure will reflect it:
```
hooks:
  - run: /usr/local/bin/create-backup.sh  # action
    when: [beforeUpdate, beforeRemove]
  - run: /usr/local/bin/shutdown-service.sh
    when: [beforeReboot]
  - if:  # condition
      path: "/etc/systemd/system/"
    when: [afterCreate, afterUpdate, afterReboot]
    run: /usr/bin/kubectl apply -f "{{ .FilePath }}" # action
  - if:
      path: "/etc/systemd/system/"
    when: [beforeRemove]
    run: /usr/bin/kubectl remove -f "{{ .FilePath }}" # action
  - if:
      path: "/var/local/import.yaml"
    when: [afterCreate, afterUpdate]
    run: /usr/bin/kubectl apply -f /var/local/crd.yaml
    envVars:
      KUBECONFIG: /var/lib/microshift/resources/kubeadmin/kubeconfig
    timeout: 10s
    workDir: /

```

ori-amizur:
There are cases that file hooks need to be run also after reboot.  I think the model needs to support it.

> Yes, because it triggers on an event unrelated to any specific path changing, but to the device transitioning into "updating".

So it seems that hooks may run in different contexts.  File-less hooks run during transitions to `update` state and others run on file change.  A bit confusing.

> A hook is a defined point in a control flow where users can insert custom actions to run. It seems you're thinking of "hook" as the (sequence of) actions themselves?

Actually I took the suggestion in the description and put it in another form which seems simpler to me.

fzdarsky:
> There are cases that file hooks need to be run also after reboot. I think the model needs to support it.

Yes. And eventually also "beforeDecommissioning" for example.

> So it seems that hooks may run in different contexts. File-less hooks run during transitions to `update` state and others run on file change. A bit confusing.

Hooks don't run. They are events. That's exactly why they need to be at the top level (compare GH Workflows, for example). What you see is the typical ECA (Event-Condition-Action) pattern. "if $path was created" is a condition and we may need other, non-file-related types of conditions for the non-update hooks.

> Actually I took the suggestion in the description and put it in another form which seems simpler to me.

That suggestion lumps together device-lifecycle events and file change-conditions and that's really confusing (and I'd argue does not address the use case).

mangelajo:
Ok,
I have looked at both suggestions and the original https://github.com/flightctl/flightctl/pull/424,

The original was a bit confusing, and I understood what would have confused @ori-amizur , shame on me for not reviewing that once when it was a PR.

While @ori-amizur is a slightly higher level definition, it does not provide a good separation/or it does not capture when in the lifecycle of the device a hook is triggered, it's more tied to file events.

IMHO [latest frank suggestion](https://github.com/flightctl/flightctl/pull/573#issue-2574015839) is clear enough, I would may be give it the extra twist of calling 'hooks' 'lifecycleHooks' so it's crystal clear that we are talking of events related to the devices and not to any path (which is where PR#424 was confusing)

api/v1alpha1/openapi.yaml
hexfusion: I think that is fine but what is the use case for a timeout that would be over 24 hours? Even in a disconnected environment getting a signal that there is a problem and then retrying feels a bit more reasonable than a timeout lasting multiple days.

hexfusion: > If we are not supporting days, this will impact the CSR CLI where we decided to support them to the exclusion of other units.

I don't see any relationship between the two honestly.

hexfusion:
I've said this before my general concern is with allowing `any` command to be run. I understand that its flexible but is it also dangerous? I think in the future we would want a white/black list commands that can be run or some chroot env. For example `shutdown`, `rm`, `reboot` feel scary. Otherwise this seems reasonable.
```
run: /usr/bin/kubectl remove -f "{{ RemovedFiles }}" # action
```

oglok:
I agree that random custom actions on the device can be dangerous. The original design from the gdoc was about having specific actions (systemd enable, disable, reload, etc... - kubectl apply, create, delete - podman run... ). With this change, only custom scripts or actions will be allowed, especially with scripts, there is no visibility on the status of the device after execution.

Anyway, this could be a good first and flexible step for a more solid hooks framework.

fzdarsky:
> I agree that random custom actions on the device can be dangerous. The original design from the gdoc was about having specific actions (systemd enable, disable, reload, etc... - kubectl apply, create, delete - podman run... ). With this change, only custom scripts or actions will be allowed, especially with scripts, there is no visibility on the status of the device after execution.
> 
> Anyway, this could be a good first and flexible step for a more solid hooks framework.

The original design and later revisions all had both built-in and user-defined actions. There also isn’t a choice: We cannot foresee which services the user installs via RPM and how they pick up configuration changes.

Anyway, the point of this PR is to fix issues with the current API and its implementation (which both differ from the designs) and not to fix that the current implementation allows custom actions (which was the intention).

oglok:
Not sure if there is a disconnect between the changes in this PR from the API and the agent's reconciliation logic, but I can't make it to work with the ACM use case:

Using an image without microshift, and updating to a new one with it, hooks don't work as expected (as we discussed in the past, and I keep saying this will bring issues with users). So instead of using the afterUpdating policy, I use the afterRebooting one:

```
  "hooks": {
    "afterRebooting": [
      {
        "envVars": {
          "KUBECONFIG": "/var/lib/microshift/resources/kubeadmin/kubeconfig"
        },
        "run": "/usr/bin/kubectl apply -f /var/local/crd.yaml",
        "timeout": "10s",
        "workDir": "/"
      },
      {
        "envVars": {
          "KUBECONFIG": "/var/lib/microshift/resources/kubeadmin/kubeconfig"
        },
        "run": "/usr/bin/kubectl apply -f /var/local/import.yaml",
        "timeout": "10s",
        "workDir": "/"
      }
    ]
  },
  "os": {
    "image": "192.168.0.20:5000/flightctl-device:microshift"
  },
  "renderedVersion": "3"
  ```
  
  This is the desired_spec ^^
  
  However, the agent shows:
  
  ```
  Nov 04 14:49:19 localhost.localdomain flightctl-agent[1146]: ▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
Nov 04 14:49:19 localhost.localdomain flightctl-agent[1146]: Your device is enrolled to flightctl,
Nov 04 14:49:19 localhost.localdomain flightctl-agent[1146]: you can manage your device scanning the above QR. or following this URL:
Nov 04 14:49:19 localhost.localdomain flightctl-agent[1146]: http://192.168.0.20.nip.io:9000/manage/ikdqi7pgnfmcn59e2144fjacekt0n7j8nahsjr29obmqjo8nbpe0
Nov 04 14:49:19 localhost.localdomain flightctl-agent[1146]: time="2024-11-04T14:49:19.922817Z" level=info msg="Management client set"
Nov 04 14:49:19 localhost.localdomain flightctl-agent[1146]: time="2024-11-04T14:49:19.926338Z" level=info msg="No OS update in progress"
Nov 04 14:49:20 localhost.localdomain flightctl-agent[1146]: time="2024-11-04T14:49:20.010325Z" level=info msg="Bootstrap complete"

```

And nothing else is executed.

I'm gonna test with an image that already contains microshift and the afterUpdating policy.

oglok:
I don't see it working with afterUpdating either:
```
  "hooks": {
    "afterUpdating": [
      {
        "envVars": {
          "KUBECONFIG": "/var/lib/microshift/resources/kubeadmin/kubeconfig"
        },
        "if": {
          "op": [
            "Create"
          ],
          "path": "/var/local/crd.yaml"
        },
        "run": "/usr/bin/kubectl apply -f /var/local/crd.yaml",
        "timeout": "10s",
        "workDir": "/"
      },
      {
        "envVars": {
          "KUBECONFIG": "/var/lib/microshift/resources/kubeadmin/kubeconfig"
        },
        "if": {
          "op": [
            "Create"
          ],
          "path": "/var/local/import.yaml"
        },
        "run": "/usr/bin/kubectl apply -f /var/local/import.yaml",
        "timeout": "10s",
        "workDir": "/"
      }
    ]
  },
  "os": {
    "image": "192.168.0.20:5000/flightctl-device:microshift"
  },
  "renderedVersion": "2"
}
```

hexfusion (commented):
Overall looks good

internal/agent/device/device.go
```diff
@@ -383,10 +387,16 @@ func (a *Agent) systemdControllerSync(_ context.Context, desired *v1alpha1.Rende
 	return nil
 }
 
-func (a *Agent) afterUpdate(ctx context.Context) error {
+func (a *Agent) afterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
 	a.log.Debug("Executing post actions")
 	defer a.log.Debug("Finished executing post actions")
 
+	rebooted := false
+	if err := a.hookManager.OnAfterUpdating(ctx, current, desired, rebooted); err != nil {
```
hexfusion, line 395: I believe we want this at the end of afterUpdate(). For example if ExecuteActions fails do we want these hooks to fire?

internal/agent/device/hook/actions.go
```diff
@@ -1,248 +1,184 @@
...
-	switch hookActionType {
-	case v1alpha1.ExecutableActionType:
-		actionHook, err = a.createExecutableActionHook(exec, log)
-	case v1alpha1.SystemdActionType:
-		actionHook, err = a.createSystemdActionHook(exec, log)
-	default:
-		return nil, fmt.Errorf("%w: %s", errors.ErrActionTypeNotFound, hookActionType)
+	for _, f := range currentFileList {
+		if content, ok := desiredFileMap[f.Path]; !ok {
+			actionCtx.removedFiles[f.Path] = content
```
hexfusion, line 126: I get why we are doing this but do we want to maintain the entire file in memory?

internal/agent/device/hook/conditions.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func pathEquals(path string, files map[string]ignv3types.File) bool {
+	_, ok := files[path]
+	return ok
+}
+
+func treeFromPathContains(path string, files map[string]ignv3types.File) bool {
+	for file := range files {
+		if strings.HasPrefix(file, path) {
```
hexfusion, line 169: Just something to think about as it might not be an issue here but HasPrefix on a path would match `/etc/foo` and `/etc/foobar`.

internal/agent/device/hook/manager.go
```diff
@@ -2,270 +2,169 @@ package hook
...
-}
-
-func (m *manager) submitBackgroundJob(path string, actions ActionMap) {
-	actionList := m.getActionsForPath(path, actions)
-	if len(actionList) == 0 {
-		return
+	// Overlay actions from the user-writable hooks directory (/etc/flightctl/hooks.d/${hookType}/*.yaml)
+	err = m.loadActions(actionsMap, filepath.Join(UserWritableConfigDir, HooksDropInDirName, strings.ToLower(string(hookType)), "*.yaml"))
+	if err != nil {
+		return []api.HookAction{}, err
```
hexfusion, line 99: in a non nil error the object should never be consumed.
```suggestion
		return nil, err
```

internal/agent/device/hook/manager.go
```diff
@@ -2,270 +2,169 @@ package hook
...
-	}
-	var errs []error
-	for _, actionHook := range actionHooks {
-		if err := actionHook.OnChange(ctx, path); err != nil {
-			m.log.Errorf("error while running hook for path %s: %+v", path, err)
-			errs = append(errs, fmt.Errorf("failed to run hook for path %s: %w", path, err))
+func (m *manager) loadActions(actionsMap map[string][]api.HookAction, actionFilesGlob string) error {
+	actionFiles, _ := filepath.Glob(m.reader.PathFor(actionFilesGlob))
+	for _, f := range actionFiles {
+		contents, err := os.ReadFile(f)
```
hexfusion, line 117: do we need to consider fileio.Reader?

packaging/hooks.d/afterupdating/00-default.yaml
```diff
@@ -0,0 +1,12 @@
...
+    op: [created, updated, removed]
+  run: systemctl daemon-reload
+- if:
+  - path: /etc/NetworkManager/system-connections/
+    op: [created, updated, removed]
+  run: nmcli conn reload
+- if:
+  - path: /etc/firewalld/
+    op: [created, updated, removed]
+  run: firewall-cmd --reload
```
hexfusion, line 12: The opinionated nature of these is ok because we are injecting via RPM.

packaging/rpm/flightctl.spec
```diff
@@ -42,10 +42,12 @@ mkdir -p %{buildroot}/usr/bin
 cp bin/flightctl %{buildroot}/usr/bin
 mkdir -p %{buildroot}/usr/lib/systemd/system
 mkdir -p %{buildroot}/%{_sharedstatedir}/flightctl
+mkdir -p %{buildroot}/usr/lib/flightctl/hooks.d/{afterupdating,beforeupdating,afterrebooting,beforerebooting}
 mkdir -p %{buildroot}/usr/lib/greenboot/check/required.d
 install -m 0755 packaging/greenboot/flightctl-agent-running-check.sh %{buildroot}/usr/lib/greenboot/check/required.d/20_check_flightctl_agent.sh
 cp bin/flightctl-agent %{buildroot}/usr/bin
 cp packaging/must-gather/flightctl-must-gather %{buildroot}/usr/bin
+cp packaging/hooks.d/afterupdating/00-default.yaml %{buildroot}/usr/lib/flightctl/hooks.d/afterupdating
```
hexfusion, line 50: nice

internal/agent/device/hook/manager.go
```diff
@@ -2,270 +2,169 @@ package hook
...
-	if len(actionHooks) == 0 {
-		return
-	}
-	var errs []error
-	for _, actionHook := range actionHooks {
-		if err := actionHook.OnChange(ctx, path); err != nil {
-			m.log.Errorf("error while running hook for path %s: %+v", path, err)
-			errs = append(errs, fmt.Errorf("failed to run hook for path %s: %w", path, err))
+func (m *manager) loadActions(actionsMap map[string][]api.HookAction, actionFilesGlob string) error {
+	actionFiles, _ := filepath.Glob(m.reader.PathFor(actionFilesGlob))
```
hexfusion, line 115: should we at least log this error?

internal/agent/device/hook/manager.go
```diff
@@ -2,270 +2,169 @@ package hook
...
 
-func (m *manager) OnAfterReboot(ctx context.Context, path string) {
-	m.submitBackgroundJob(path, m.onAfterReboot)
-}
+		actionTimeout, err := parseTimeout(action.Timeout)
+		if err != nil {
+			return err
+		}
+		ctx, cancel := context.WithTimeout(ctx, actionTimeout)
+		defer cancel()
```
hexfusion, line 163: defer fires on return since we are in a loop we should manually cancel()

internal/agent/device/hook/conditions.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	lhs, op, rhs := match[1], match[2], match[3]
+	switch lhs {
+	case "rebooted":
+		b, err := strconv.ParseBool(rhs)
+		if err != nil {
+			return false
+		}
+		return checkOp(actionCtx.systemRebooted, op, b)
+	default:
```
hexfusion, line 61: minor: slightly more easier to read if you return false here

internal/agent/device/hook/conditions.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	return false
+}
+
+func checkOp[T comparable](lhs T, op string, rhs T) bool {
```
hexfusion, line 66: :)

internal/agent/device/hook/conditions.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+
+// checkFileOpConditionForDir checks whether a specified operation (create, update, remove) has been performed
+// on any file in the tree rooted at the specified path.
+// As a side-effect, it populates the command line variables of the action context with the corresponding list of files.
+func checkPathOpConditionForDir(cond api.HookConditionPathOp, actionCtx *actionContext) bool {
+	// ensure dir paths end with a trailing slash, so we don't accidentally match a file with a similar prefix
+	if len(cond.Path) < 1 || cond.Path[len(cond.Path)-1] != '/' {
+		cond.Path += "/"
```
hexfusion, line 96: super nitty: totally get the reasoning here but we are mutating the input which we could just copy and handle.

packaging/hooks.d/afterupdating/00-default.yaml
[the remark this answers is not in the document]
hexfusion: nmd this is injected only via RPM

internal/agent/device/hook/manager.go
hexfusion: or maybe even easier put this inside executeAction

internal/agent/device/hook/conditions.go
fzdarsky: The caller ensures paths terminate with `/`, but I agree in terms of defensive programming I should hedge for this here, too. Will add.

internal/agent/device/hook/manager.go
fzdarsky: Tried that originally. Problem is `filepath.Glob()` above already needs the rootDir prefixed, so the list of matching files does, too. Using fileio.Reader would the double-prefix the rootDir.

internal/agent/device/device.go
fzdarsky: Importantly, we need to explicit for each hook about what happens when an action fails (see PR #619). I believe that, yes, also for an afterUpdating hook we should by default roll back when an action fails. Let's say an invalid firewall rule that fails to activate.

There may be other actions that are less critical. In that case, maybe the user should be able to explicitly specify what to do, e.g. adding an `onError:` field to the action that can `rollback` or `ignore`.

internal/agent/device/bootstrap.go
```diff
@@ -171,15 +170,8 @@ func (b *Bootstrap) ensureBootstrap(ctx context.Context) error {
...
-	if err != nil {
-		return fmt.Errorf("parsing current ignition: %w", err)
-	}
-	b.log.Info("Executing after reboot hooks")
-	defer b.log.Info("Finished executing after reboot hooks")
-	for _, f := range currentIgnition.Storage.Files {
-		b.hookManager.OnAfterReboot(ctx, f.Path)
-	}
+	// TODO: call if device has rebooted, rather than just agent restarted
+	// b.hookManager.OnAfterRebooting(ctx)
```
hexfusion, line 174: ack

internal/agent/device/hook/actions.go
```diff
@@ -1,248 +1,187 @@
...
-	return &systemdActionHook{
-		actionTimeout: actionTimeout,
-		operations:    operations,
-		unitName:      unitName,
-		exec:          exec,
-		log:           log,
-	}
-}
+	ctx, cancel := context.WithTimeout(ctx, actionTimeout)
+	defer cancel()
```
hexfusion, line 138: :+1:

hexfusion (approved):

PR #575, opened 2024-10-09
EDM-573: Run helm dependency build in release job

rawagner:


mangelajo:
This way the dependency is bundled inside?, or it's because otherwise it won't work?

rawagner:
Yes, it will be bundled inside. Without it, the helm package step is [failing](https://github.com/flightctl/flightctl/actions/workflows/publish-containers.yaml)

mangelajo:
oops I see. Let's go :D

fzdarsky (approved):

PR #577, opened 2024-10-09
Device sim labels

DakCrowder:
Adds ability to pass labels in a `key=value` format to the device simulator.  All devices spun up by the sim will share the passed labels.


fzdarsky (approved):

amir-yogev-gh (changes requested):
Who asked for it?
I'm the feature owner for the simulator improvements and this was not discussed with me...

mangelajo:
I think that in general we should add this to any devices from the device config, not only the simulator, that specific was discussed in many meetings. In that way, devices can have pre-defined config labels, that then can be used during enrollment phase, or device lifecycle.

mangelajo:
> Who asked for it? I'm the feature owner for the simulator improvements and this was not discussed with me...

This is probably necessary to scale/test devices grouped by fleets.

avishayt:
> This is probably necessary to scale/test devices grouped by fleets.
It might be useful but is not necessary.  It's possible to list ERs/devices and add labels to them.

amir-yogev-gh:
"Probably" - is this the justification for a feature?

I'm complaining about the process, not the feature.
Where does it fit in the overall testing?
What are the alternatives?
Do you have the details to answer these questions?

No, this is unacceptable for me.

mangelajo:
> "Probably" - is this the justification for a feature?
> 
> I'm complaining about the process, not the feature. Where does it fit in the overall testing? What are the alternatives? Do you have the details to answer these questions?
> 
> No, this is unacceptable for me.

Anyway, this is an internal development tool, more than a customer-side tool, right, or are we planning to support a device simulator? :)

fzdarsky:
> Who asked for it? I'm the feature owner for the simulator improvements and this was not discussed with me...

The ask came from Perf&Scale testing (Yuri). It's required so we can run multiple dev sim instances in parallel and automatically label devices so they can be selected into different fleets.

amir-yogev-gh (approved):

PR #580, opened 2024-10-10
NO-ISSUE: Move device/fleet spec comparison methods to api

avishayt:
The new comparison methods compare api objects, so they should be in the api package. Also, fleet rollout still used the reflect.DeepEqual comparison which doesn't work on DeviceSpec.


api/v1alpha1/util.go
```diff
@@ -51,3 +55,113 @@ func getApplicationType(union json.RawMessage) (string, error) {
 
 	return "", fmt.Errorf("unable to determine application provider type: %+v", data)
 }
+
+func configsAreEqual(c1, c2 *[]DeviceSpec_Config_Item) bool {
+	return slices.EqualFunc(lo.FromPtr(c1), lo.FromPtr(c2), func(item1 DeviceSpec_Config_Item, item2 DeviceSpec_Config_Item) bool {
```
hexfusion, line 60: I know you are just moving but we have `util.FromPtr`

hexfusion (approved):

api/v1alpha1/util.go
avishayt: Should we make a concentrated effort to remove lo?  Or specific methods of it?

hexfusion: We have moved a lot of support for PTRs to util. We should go one way or the other. Since most of these are simple functions, moving to util makes sense? But we should be consistent either way.

avishayt: Agreed

avishayt: Opened [EDM-600](https://issues.redhat.com/browse/EDM-600)

PR #581, opened 2024-10-10
K8s selectors for labels fields

asafbss:
### Overview
This PR introduces support for the Kubernetes-style field selector.

### Usage Example:
```
Filtering by multiple fields:
devices?fieldSelector=metadata.name!=mydevice,status.updated.status=UpToDate

Filtering the same field with multiple values:
devices?fieldSelector=metadata.name!=device1,metadata.name!=device2,status.updated.status=UpToDate
```

### Key Changes

- Commit d3d1837: We added a new query parser package to facilitate tokenized, generic, function-based calls. This parser offers robust syntax validation, allows for the definition of relationships, and enforces restrictions, ultimately enabling the construction of secure and valid parameterized SQL queries that GORM will use.

- Commit 7290ce9: We added a field selector implementation that uses the Kubernetes selector parser to parse the selector data. We then tokenize the output to enable effective querying.

- Commit d514f76: The field selector has been integrated throughout the API and storage components.

### Additional Features
- Added support for OrderBy functionality, enhancing the query capabilities and allowing for more sophisticated data retrieval.

### Future Enhancements
This initial implementation lays the groundwork for further enhancements, including:

- Implementing label selectors that will leverage the same tokenization technique.
- Subsequent PRs will refine the field selector functionality, including features such as substring matching.
- Corresponding tests will be integrated in future PRs.


celdrake:
> Looks great just high level could you provide some usage examples in the description, and it is somewhat noticeable that there isn't a single test :),. For reviewers and in general, tests are a great place to step through the logic quickly.

I agree about the examples. Could you please provide some examples showcasing the syntax of more complex use cases, such as:
- Filtering by several fields at the same time
- Filtering the same field with multiple values.

asafbss:
Hey @hexfusion, @celdrake,

I consulted with @avishayt. Since this serves as the baseline for selectors and to avoid overwhelming PR reviews, we decided to create the initial PR and introduce tests in subsequent ones.

The usage follows the current k8s field selector (before enhancement), meaning it only supports =, ==, and != for now. Any model objects tagged with 'selector' can be queried using these operators.

I’ll add the following usage examples to the description:

Example:

Filtering by multiple fields:
devices?fieldSelector=metadata.name!=assaf,status.updated.status=UpToDate

Filtering the same field with multiple values:
devices?fieldSelector=metadata.name!=assaf,metadata.name!=assaf2,status.updated.status=UpToDate

In the next phase, we’ll enhance the k8s field selector package to support additional operators. This enhancement will integrate smoothly with this PR, which uses the native k8s parser implementation.

api/v1alpha1/openapi.yaml
```diff
@@ -67,6 +72,12 @@ paths:
           description: The name of the repository to filter results by.
```
celdrake, line 72: Should the `repository` filter also be dropped in favour of `?fieldSelector=spec.repository=someRepo`?

BTW, If I attempt this at the moment, the error returned is error 500, with the message
```
unable to tokenize input using custom tokenizer: unknown field name "spec.repository"
```

asafbss: Hey, not all the fields were marked with selector. I marked only metadata and status. We can mark the spec at the model level to be a field that can be queried if required.

celdrake: Sure, I understand it's not a supported field yet. 

The comment was also about the fact that the error should be 400 instead of 500, as it's considered a bad request as it's referencing an unsupported filtering field. 

BTW, the same happens when specifying an unsupported sorting field in  "sortBy" query parameter and IMO it should also be error 400.
```
Error 500: unable to resolve field name "spec.template.spec.os"
```

internal/store/selector/field.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/flightctl/flightctl/pkg/queryparser/sql"
+	"gorm.io/gorm/schema"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+type fieldSelector struct {
+	parser        queryparser.Parser
+	fieldResolver *selectorFieldResolver
+	operators     map[selection.Operator]string
```
avishayt, line 22: It looks like this map never changes, so why have it in this struct?

internal/service/common.go
```diff
@@ -121,6 +123,40 @@ func ConvertFieldFilterParamsToMap(params []string) (map[string][]string, error)
 	return fieldMap, nil
 }
 
+// validateSortField validates a sort field and returns a slice of SortField and a boolean indicating success.
+func validateSortField(field string) ([]store.SortField, bool) {
```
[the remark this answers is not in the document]
asafbss: The validation function will apply to the type ListDevicesParams since we're not passing a device kind in the request body. We also make other validations to the requests once they arrive before calling List.
Should we move the function to the validation.go? Should we implement validation for ListDevicesParams?

api/v1alpha1/openapi.yaml
```diff
@@ -67,6 +72,12 @@ paths:
...
           schema:
             type: string
+        - name: sortBy
+          in: query
+          description: Specifies the field(s) to sort by and the sort order (ascending or descending). The format is 'fieldName:asc' or 'fieldName:desc'. You can specify multiple sort fields by separating them with commas, e.g., 'field1:asc,field2:desc'.
```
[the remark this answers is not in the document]
asafbss: @celdrake should we apply multiple sort fields?

internal/store/store.go
```diff
@@ -144,6 +148,18 @@ type Continue struct {
 	Count   int64
 }
 
+type SortOrder string
+
+const (
+	SortAsc  SortOrder = "ASC"
+	SortDesc SortOrder = "DESC"
```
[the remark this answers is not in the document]
asafbss: We need to decide if multiple fields are required to sort by or might be required.
@celdrake

internal/service/certificatesigningrequest.go
```diff
@@ -102,6 +122,9 @@ func (h *ServiceHandler) ListCertificateSigningRequests(ctx context.Context, req
 	case nil:
 		return server.ListCertificateSigningRequests200JSONResponse(*result), nil
 	default:
+		if s, ok := selector.IsSelectorError(err); ok {
```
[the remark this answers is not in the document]
asafbss: Each selector error is a new instance containing the flightctl error that is already allocated and a selector error that contains more information. Since we switch (err) this approach cannot work because we cannot compare.
We can change the switch case to something like switch{ case err ==x: case err==y: , case IsSelectorError(err)}

api/v1alpha1/openapi.yaml
```diff
@@ -55,6 +55,11 @@ paths:
           description: A selector to restrict the list of returned objects by their labels. Defaults to everything.
           schema:
             type: string
+        - name: fieldSelector
+          in: query
+          description: A selector to restrict the list of returned objects by their fields. Defaults to everything.
```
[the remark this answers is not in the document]
asafbss: Once we apply the full set of operators in the next PR (which is almost ready for commit), it should be fully described in the docs. Should I write some draft now?

internal/store/selector/field.go
asafbss: This can be done, but keep in mind that this package will have more selectors (i.e., labels) that will use the same methods and will have their operator map. I wanted to avoid name collisions for this small map with others. 
Will appreciate your advice.

api/v1alpha1/openapi.yaml
```diff
@@ -588,6 +610,11 @@ paths:
...
+          schema:
+            type: string
         - name: statusFilter
```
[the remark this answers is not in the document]
asafbss: @celdrake, please pull the latest changes. They are still working but can be applied via the field selector.
@avishayt @fzdarsky

api/v1alpha1/openapi.yaml
asafbss: Please pull the latest changes and check now

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
celdrake: Not at the moment, and we didn't discuss need for it so far.

api/v1alpha1/openapi.yaml
celdrake: Fixed, thanks. For both filtering and sorting, API now returns error 400 for unknown fields.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
celdrake: Nice, I see they are working again. 
IMO it's fine to keep the old filters and only drop them once the new mechanisms allows all functionality. 
Longer term, I imagine the API should have a single filtering mechanism.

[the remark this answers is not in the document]
avishayt: Agree with @celdrake

asafbss:
Hey guys, I have added a new test file for the field selector that can be found in the enhanced operators branch.
It can help understand better the usage:

[](https://github.com/flightctl/flightctl/blob/k8s-selectors-enhance-operators/internal/store/selector/field_test.go)

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
asafbss: I changed the description and added more clarity

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
asafbss: Done

internal/service/common.go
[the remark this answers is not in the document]
asafbss: Done, It was removed due to the API changes

internal/store/store.go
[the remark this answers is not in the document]
asafbss: Done

internal/store/selector/field.go
asafbss: Done

internal/service/certificatesigningrequest.go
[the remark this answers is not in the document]
asafbss: Done, the approach has changed.

pkg/queryparser/sql/sql.go
```diff
@@ -0,0 +1,654 @@
...
+type parser struct {
+	sqlparser *SQLParser
+}
+
+// Parse constructs a SQL query based on the provided input.
+// This method tokenizes the input, verifies the structure of the tokens,
+// and executes the corresponding SQL functions to generate the final query
+// string along with its parameters.
+func (sp *SQLParser) Parse(ctx context.Context, input any, params ...string) (string, []any, error) {
+	if input == "" {
```
hexfusion, line 279: nit: If input can be any then we should use type assertion to ensure `input.(string)` before we evaluate it as a string.

internal/store/selector/field.go
```diff
@@ -0,0 +1,384 @@
...
+			return "", nil, err
+		}
+		return "", nil, NewSelectorError(flterrors.ErrFieldSelectorParseFailed, err)
+	}
+	return q, args, nil
+}
+
+// Tokenize converts a selector string into a set of queryparser tokens.
+func (fs *fieldSelector) Tokenize(ctx context.Context, input any) (queryparser.TokenSet, error) {
+	if input == "" {
```
hexfusion, line 75: are we sure this is a string?

internal/store/selector/field.go
```diff
@@ -0,0 +1,384 @@
...
+
+		resolvedTokens := queryparser.NewTokenSet()
+		for _, resolvedField := range resolvedFields {
+			fieldToken, err := fs.createFieldToken(resolvedField)
+			if err != nil {
+				return nil, err
+			}
+
+			var valueToken *queryparser.TokenSet
+			if value != "" {
```
hexfusion, line 107: so if value =="" then valueToken is nil? Does createOperatorToken assume it is not?

internal/store/selector/selector.go
```diff
@@ -0,0 +1,256 @@
...
+	"reflect"
+	"strings"
+	"sync"
+
+	"github.com/flightctl/flightctl/internal/flterrors"
+	gormschema "gorm.io/gorm/schema"
+)
+
+var (
+	cacheStore           = &sync.Map{}
```
hexfusion, line 15: nit: are there any drawbacks to having the cache as a global?

pkg/queryparser/sql/sql.go
asafbss: Hey,

I will fix it and compare it to nil.
The input can be anything, for the field selector is a type selector that will be processed by the tokenizer.
In this case, the field selector implements a tokenizer and doesn't use the default.
Therefore, the assertion can be found here: https://github.com/flightctl/flightctl/blob/9e7765f7d44602c5757d1414be45ebc3094a858e/internal/store/selector/field.go#L80

In case when the default tokenizer is being used: https://github.com/flightctl/flightctl/blob/9e7765f7d44602c5757d1414be45ebc3094a858e/pkg/queryparser/queryparser.go#L299

Initially, it was indeed string until more types of inputs were introduced.

internal/store/selector/field.go
asafbss: Same as last comment, I will fix it

internal/store/selector/field.go
asafbss: The createOperatorToken requires a tokenset representing the value token.
It doesn't care if the token set is empty or nil, the tokenset is able to process it:
https://github.com/flightctl/flightctl/blob/9e7765f7d44602c5757d1414be45ebc3094a858e/internal/store/selector/field.go#L201

For example, look at the enhance operators branch that implements the entire set of operators.
Since we can have multiple values (i.e., IN), the createOperatorToken still takes a single token set for all values.
https://github.com/flightctl/flightctl/blob/19198b6c7d1a53a2601924e7ce2065f1295cf23b/internal/store/selector/field.go#L111

internal/store/selector/selector.go
asafbss: Gorm uses quite a similar approach :
https://github.com/go-gorm/gorm/blob/deceebfab8c460cfee229233aded2821ac6b08eb/gorm.go#L173

This is to save some reflection processing for the same model fields.

internal/store/selector/selector.go
```diff
@@ -235,11 +235,22 @@ func (e *SelectorError) Unwrap() error {
 	return e.OriginalError
 }
 
+// AsSelectorError checks if an error is of type SelectorError and assigns it to target.
+func AsSelectorError(err error, target any) bool {
+	if target == nil {
+		return false
+	}
+	// Ensure target is a pointer to the correct type
+	switch t := target.(type) {
```
celdrake, line 244: Hi! I am adding the comment in this line, but it may not the the right place.
The issue is that the API only identifies an invalid field if the first part of the path is incorrect.
eg. "status.summarywhatever.status" is accepted as valid, whereas "statuswhatever.summary.status" is not.

IMO both fields should be identified as invalid fields.

field-validation

siserafin: @celdrake there's an open issue for it https://issues.redhat.com/browse/EDM-605

avishayt (approved):

PR #582, opened 2024-10-10
NO-ISSUE: Add support for match expressions

ori-amizur:
Match expressions provides more capabilities than match labels only by value.


avishayt (commented):
Nice!  @asafbss, please review it and see how it fits your current work.

internal/store/common.go
```diff
@@ -152,6 +156,66 @@ func GetNonNilFieldsFromResource(resource model.Resource) []string {
 	return ret
 }
 
+func matchExpressionQueryAndArgs(matchExpression api.MatchExpression, colName string) (string, []any) {
+	params := strings.Join(lo.Map(lo.FromPtr(matchExpression.Values), func(_ string, _ int) string { return "?" }), ",")
```
asafbss, line 160: We might consider simply this and loop once something like:
```
params := make([]string, len(*matchExpression.Values))
args := make([]any, len(*matchExpression.Values))

for i, val := range *matchExpression.Values {
    params[i] = "?"
    args[i] = matchExpression.Key + "=" + val
}

return fmt.Sprintf("%s && ARRAY[%s]", colName, strings.Join(params, ",")), args
```

internal/store/common.go
```diff
@@ -152,6 +156,66 @@ func GetNonNilFieldsFromResource(resource model.Resource) []string {
...
+		return matchDoesNotExistQuery(query, matchExpression, colName)
+	default:
+		// TODO: Return error for this case and handle it
+		return query
+	}
+}
+
+func matchExpressionsQuery(query *gorm.DB, matchExpressions api.MatchExpressions, colName string) *gorm.DB {
+	for _, me := range matchExpressions {
+		query = matchExpressionQuery(query, me, colName)
```
asafbss, line 205: Would it be better to apply matchExpressionQuery within this loop and eliminate the standalone matchExpressionQuery function? We can conceptualize that even a single expression is part of a matching set

api/v1alpha1/openapi.yaml
```diff
@@ -3833,15 +3833,38 @@ components:
...
+        $ref: '#/components/schemas/MatchExpression'
+
     LabelSelector:
       type: object
       properties:
         matchLabels:
           type: object
           additionalProperties:
             type: string
-      required:
```
asafbss, line 3843: It's a bit misleading functionality-wise; it implies that someone can define a LabelSelector but keep both matchLabels and matchExpressions empty.

asafbss (commented):
Looks good, sharing my comments :)

api/v1alpha1/openapi.yaml
avishayt: I don't know if there's a better way.  Though we should make sure to validate it in the code (api/v1alpha1/validation.go).

asafbss: Can we keep the matchLabels AS IS and only set matchExpressions as optional?

ori-amizur: If we keep it as is then we will be required to always add the matchLabels field even if it is empty.

internal/store/common.go
ori-amizur: Why ?  Is it better ?

internal/store/common.go
```diff
@@ -152,6 +156,66 @@ func GetNonNilFieldsFromResource(resource model.Resource) []string {
...
+		return matchInQuery(query, matchExpression, colName)
+	case api.NotIn:
+		return matchNotInQuery(query, matchExpression, colName)
+	case api.Exists:
+		return matchExistsQuery(query, matchExpression, colName)
+	case api.DoesNotExist:
+		return matchDoesNotExistQuery(query, matchExpression, colName)
+	default:
+		// TODO: Return error for this case and handle it
+		return query
```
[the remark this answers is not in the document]
ori-amizur: OK, but it will mean we will carry the error all the way. Currently we don't.

internal/store/common.go
ori-amizur: OK

asafbss (approved):

asafbss (approved):

PR #584, opened 2024-10-10
MGMT-18170: Update documentation for certificate request

lkatalin:
This updates the docs based on changes in #529 .

As we want to support manually creating the CSR for some advanced users, this flow has been moved to a separate file.

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -243,126 +243,27 @@ Browse to `ui.flightctl.MY.DOMAIN` and use the login "demouser" and the password
 
 ## Building a Bootable Container Image including the Flight Control Agent
 
-Next, we will use [Podman](https://github.com/containers/podman) to build a [bootable container image (bootc)](https://containers.github.io/bootc/) that includes the Flight Control Agent binary and configuration. The configuration contains the connection details and credentials required by the agent to discover the service and send an enrollment request to the service. After deploying the flightctl server, this configuration can be retrieved after a Certificate Signing Request has been completed as outlined in the steps below.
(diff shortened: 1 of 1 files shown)


fzdarsky:
The "getting started" page is more a "quickstart" than a documentation. Means, it's meant to be highly opinionated (no description of alternatives) and concise ("do X to achieve Y", no detailed explanations). So basically, let's simply revert the section on getting the enrollment config to what it was originally (with the embedded certs), please.

More detailed documentation would belong into the building images, device provisioning, and device enrollment sections, respectively. However, even there we don't want to describe multiple options, unless there's strong reason and user demand. Therefore, let me complete these sections first, please, then we can review if something needs adding and where.

lkatalin:
>The "getting started" page is more a "quickstart" than a documentation.

Okay, this makes sense.

> let's simply revert the section on getting the enrollment config to what it was originally

Do you mean the 4 part process using `apply -f`? Isn't that more complicated than what we want the user to do?

> even there we don't want to describe multiple options

Sounds good.

> let me complete these sections first, please, then we can review if something needs adding and where

This is fine as well. Should I then close this PR?

lkatalin:
@fzdarsky Please take a look when you get the chance, I've got it back to your first simpler version plus updating the command flags.

fzdarsky (approved):

lkatalin:
I am ready to merge this, however it seems that it cannot be merged because some tests are disabled.

fzdarsky:
> I am ready to merge this, however it seems that it cannot be merged because some tests are disabled.

I'm afraid this is an issue I introduced with #570 that attempts to avoid running expensive code tests when only docs have changed. I've prepared PR #592 that should fix this and that we need to merge first.

PR #586, opened 2024-10-13
EDM-240: breaking: make application status data an array

hexfusion:
To ensure consistency between the spec and the application status API, this PR proposes changing the application status from a map to an array. This change will align the data structures and maintain spec and status parity. Note this change would be breaking.


hexfusion:
cc @celdrake for awareness we have not agreed on the chance but FYI

celdrake:
> cc @celdrake for awareness we have not agreed on the chance but FYI

Thanks for the heads-up, @hexfusion . Please let me know if it's confirmed, as the changes in the UI are small and I could prepare a PR quickly for adapting to this change.

fzdarsky:
Good idea, +1 on this.

Question for discussion: While at it, should we move `applications.data` up to `applications` and `applications.summary` to `applicationsSummary` for consistency (and as the Service may need to update/compute that field also to separate responsibilities)?

celdrake:
@hexfusion I adapted the UI to the API changes up to the second commit.

https://github.com/flightctl/flightctl-ui/pull/131 can be merged when this PR is.

fzdarsky (approved):

PR #590, opened 2024-10-17
EDM-603: Add prometheus target filters

vraiti:
Filter discovered pods by port to only scrape those that use ports 15690 (API), 15691 (DB), and 15692 (RabbitMQ) and avoid unnecessary unconnected targets.


hexfusion (approved):
```
curl -s http://localhost:9090/api/v1/targets | jq '.data.activeTargets[].labels.instance'
"10.244.0.13:15690"
"10.244.0.13:15691"
"10.244.0.11:15692"
```

PR #591, opened 2024-10-17
EDM-609: EDM-608: ensure unique application names

hexfusion:
This PR enforces unique application names. If a name isn't provided, the image name is used for validation in the case of an image provider.

Additionally, it fixes a bug in the update application flow, ensuring that containers from the previous lifecycle are correctly removed from the status.


celdrake:
@hexfusion I noticed a small problem.
The validation that the name is unique is correctly applied when POSTing the fleet / device template, but not when PATCHing it. So I can't create a spec with two apps with the same name, but later I can edit them and get to the problematic situation.

https://github.com/user-attachments/assets/65039e75-d99f-4677-b58d-1ffe9e681a1b

siserafin:
I also see another issue. Adding twice the same healthy application, with different names, causes a reported device status.summary  of degraded: 
   ```
 summary:
    info: 'unknown application summary status: Unknown'
    status: Degraded
```

The application reported status in the device is:
```
applications: []
  applicationsSummary:
    status: Healthy
```

hexfusion:
Right this is a limitation on compose I believe. I will work on better status reporting. But service names must be unique is what I believe is the issue.

DakCrowder (approved):

celdrake (approved):

PR #592, opened 2024-10-19
EDM-567: fix conditional workflows

fzdarsky:
Fixes a problem introduced by PR #570, which made docs-only PRs skip code-related checks. However, some of these checks are configured as required by branch protection, causing jobs to be stuck with "Waiting for status to be reported". This is a well-known problem in GitHub that does not yet have a solution.

A recommended workaround is to filter for changes and make steps conditional within the job. This is what this PR implements using the @dorny/paths-filter action.

Depens on #593


.github/workflows/lint.yaml
```diff
@@ -1,46 +1,56 @@
...
-
-      - name: Setup all dependencies
-        uses: ./.github/actions/setup-dependencies
-
-      - name: Check that generated files have been updated and that go.mod is tidy
-        run: |
-          make generate
-          go mod tidy -v
-          git diff --exit-code
+      #   with:
```
mangelajo, line 19: I guess it's WIP yet, right? :-)

fzdarsky: Only until I can find out why these checks fail even when I revert them to original. grrrr

fzdarsky: Turns out golangci-lint v1.54 does not like golang 1.23 that we're using in some CI tests, although our artifacts correctly build with 1.21. I've created PR #593 to fix that and rebased this PR on it.

mangelajo (approved):
Looks good.

PR #596, opened 2024-10-21
NO-ISSUE: Remove containers matchPattern from device spec

hexfusion:
Breaking change, this is no longer used with the addition of applications and will only cause confusion.


hexfusion:
cc @celdrake

fzdarsky (approved):

celdrake (commented):
@hexfusion In the UI, we are not using  `containers`.

However, we are using "systemdUnits.matchPatterns". Looks like these are not being dropped, is that correct?

hexfusion:
> @hexfusion In the UI, we are not using `containers`.
> 
> However, we are using "systemdUnits.matchPatterns". Looks like these are not being dropped, is that correct?

that is correct

PR #598, opened 2024-10-22
NO-ISSUE: Move Prometheus components to e2e-extras

vraiti:
Since the Prometheus server is for development purposes, it makes more sense to place it in e2e-extras than the core flightctl helm chart.


mangelajo:
it looks good although now it needs a rebase.

mangelajo (approved):

deploy/helm/flightctl/values.yaml
```diff
@@ -112,5 +112,3 @@ ui:
   enabled: true
   api:
     insecureSkipTlsVerify: true
-prometheus:
-  enabled: false
```
mangelajo, line 116: oh we should have left a default here to 

prometheus:
    enabled: false

otherwise some charts fail when deploying on openshift

PR #606, opened 2024-10-24
NO-ISSUE: ensure agent teardown on failure

hexfusion:
The simulator would not tear down agents on error


DakCrowder (approved):

cmd/devicesimulator/main.go
```diff
@@ -87,31 +89,40 @@ func main() {
 	}()
 
 	log.Infoln("running agents")
-	wg := new(sync.WaitGroup)
-	wg.Add(*numDevices)
```
hexfusion, line 91: note: wg is usually needed in more complex situations, if we properly ensure each goroutine collapses when the context is cancelled waiting on <-ctx.Done() is all we need.

DakCrowder (approved):

PR #610, opened 2024-10-29
K8s selectors enhance operators

asafbss:
# Overview

This PR builds on the enhancements introduced in PR #581, bringing improved Kubernetes (k8s) operators for field selectors.

## Key Changes

- **Integration of Kubernetes Official Package**:
  - Enhanced field selectors with recognized k8s operators:
    - `Exists`: `field`
    - `DoesNotExist`: `!field`
    - `Equals`: `field =`
    - `DoubleEquals`: `field ==`
    - `NotEquals`: `field !=`
    - `In`: `field in (val1, val2)`
    - `NotIn`: `field notin (val1, val2)`
    - `GreaterThan`: `field >`
    - `GreaterThanOrEquals`: `field >=`
    - `LessThan`: `field <`
    - `LessThanOrEquals`: `field <=`

- **Introduction of the Containment Operator (`contains, notcontains`)**:
  - **String**: `field contains val` -> Checks if the field value contains `val`.
  - **Array**: `field contains val` -> Checks if the array field contains the item `val`.
  - **JSONB**: `field contains val` -> Checks if the left JSON field contains the right JSON path/value entries at the top level.

  ### Usage Examples of Containment

```
--field-selector='metadata.name contains a'
--field-selector='metadata.labels contains key1\=value1'
--field-selector='status.conditions contains [{"type":"Accessible"}]'
```

**New Field in Device Object**:
- Introduced a new field named `Alias` to the `Device` object.

These changes are aimed at enhancing the flexibility and capability of field selectors, making them more powerful and aligned with Kubernetes standards.


api/v1alpha1/openapi.yaml
```diff
@@ -2616,7 +2616,7 @@ components:
           type: string
           description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
         metadata:
-          $ref: '#/components/schemas/ObjectMeta'
+          $ref: '#/components/schemas/DeviceMetadata'
```
hexfusion, line 2619: cc @celdrake

api/v1alpha1/openapi.yaml
```diff
@@ -3958,7 +3958,14 @@ components:
           type: string
           description: An opaque string that identifies the server's internal version of an object.
       description: ObjectMeta is metadata that all persisted resources must have, which includes all objects users must create.
-
+    DeviceMetadata:
+      allOf:
+        - $ref: '#/components/schemas/ObjectMeta'
+        - type: object
+          properties:
+            alias:
```
hexfusion, line 3966: do we need to update any docs on this?

pkg/k8sselector/selection/operator.go
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
hexfusion, line 2: Since we will probably have more k8s-related logic would it make sense to have a `pkg/k8s/{selector,whatever}`

internal/store/device.go
```diff
@@ -82,6 +82,28 @@ func (s *DeviceStore) InitialMigration() error {
 		}
 	}
 
+	// Create indexes for device 'Alias' column
```
hexfusion, line 85: :heart:

api/v1alpha1/openapi.yaml
celdrake: We should be aligned to the contents of this PR in https://github.com/flightctl/flightctl-ui/pull/130
I imagine we should probably wait until 0.3.0 is released to merge them both.

api/v1alpha1/openapi.yaml
asafbss: Yes, we do; I'll update you once they are ready for review.
They will be added to this PR.

docs/user/field-selectors.md
amir-yogev-gh, line 1: Adding more examples can be useful, especially for popular cases.

asafbss: Will do

asafbss: 4c8f1cb25411f362bc72044a34e137764cb569ed

asafbss: @fzdarsky @hexfusion @celdrake please share your comments for the committed docs

docs/user/field-selectors.md
```diff
@@ -0,0 +1,149 @@
+# Field Selectors
+
+Field selectors filter a list of Flight Control objects based on specific resource field values.<br>
+They follow the same syntax, principles, and support the same operators as Kubernetes Field and Label selectors.
```
celdrake, line 4: I'd add something along of the lines of the suggestion, given we next describe more operators than only those supported by Kubernetes.

```suggestion
They follow the same syntax, principles, and support the same operators as Kubernetes Field and Label selectors. Some additional operators are also supported to allow for more advanced search use cases.
```

docs/user/field-selectors.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Supported Fields
+
+The following table lists the fields supported for filtering Flight Control objects:
+
+| Field Name                   | Field Type           |
+|------------------------------|----------------------|
+| `metadata.name`              | `String`             |
+| `metadata.owner`             | `String`             |
+| `metadata.labels`            | `StringArray`        |
```
celdrake, line 49: Maybe it's worth nothing that `labels` can be queried more extensively using `labelSelector` instead.

docs/user/field-selectors.md
```diff
@@ -0,0 +1,149 @@
...
+
+Use the `--field-selector` flag to filter devices based on field values. Here are some examples:
+
+#### Example 1: Exclude a Specific Device by Name
+This command filters out a specific device by its name:
+```bash
+flightctl get devices --field-selector 'metadata.name!=c3tkb18x9fw32fzx5l556n0p0dracwbl4uiojxu19g2'
+```
+#### Example 2: Filter by Owner, Labels, and Creation Timestamp
+This command retrieves devices owned by `Fleet/pos-fleet`, located in the `us` region, and created in 2024:
```
celdrake, line 67:
```suggestion
This command retrieves devices owned by `Fleet/pos-fleet`, located in the `us` region, and created before 2025:
```

docs/user/field-selectors.md
```diff
@@ -0,0 +1,149 @@
...
+In this example, the field `text` is not a valid field for filtering. The error message provides a list of supported fields that can be used with `--field-selector` for the `device` object.
+
+You can then use one of the supported fields, as shown below:
+```bash
+flightctl get devices --field-selector 'metadata.alias @> cluster'
+```
+In this command, the `metadata.alias` field is checked with the containment operator `@>` to see if it contains the value `cluster`.
+
+
+### Fields Discovery
```
celdrake, line 97: "Fields Discovery" section is duplicate, with a few minor variations in each case.

docs/user/field-selectors.md
asafbss: but it also has 'metadata.creationTimestamp >= 2024-01-01T00:00:00Z'

docs/user/field-selectors.md
asafbss: Added note: 'While `metadata.labels` can be queried using Field Selectors, for more extensive label querying options, consider using Label Selectors.'

docs/user/field-selectors.md
asafbss: Fixed

docs/user/field-selectors.md
asafbss: Change to: 
They follow the same syntax, principles, and support the same operators as Kubernetes Field and Label selectors, with additional operators available for more advanced search use cases.

docs/user/field-selectors.md
celdrake: Sorry, for some reason I missed the check >= Jan 2024. Resolving comment :)

docs/user/field-selectors.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Field Name                   | Field Type           |
+|------------------------------|----------------------|
+| `metadata.name`              | `String`             |
+| `metadata.owner`             | `String`             |
+| `metadata.labels`            | `StringArray`        |
+| `metadata.annotations`       | `StringArray`        |
+| `metadata.creationTimestamp` | `Timestamp`          |
+| `metadata.deletionTimestamp` | `Timestamp`          |
+| `spec`                       | `JSON`               |
```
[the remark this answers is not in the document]
asafbss: @fzdarsky, we already expose the Status field through StatusFilter, which is currently supported by the API.

Modifying the Spec or Status fields could also disrupt clients that use the Go client package to query these structures, as well as CLI with oyaml and ojson and manually parse them.

These JSONB fields are indexed in the same way, indexing all JSON structures within

docs/user/field-selectors.md
```diff
@@ -0,0 +1,135 @@
...
+JSON fields allow querying by specific JSON values, but certain scenarios may require additional functionality.
+<br>For example:
+
+- Casting is available to treat values as specific data types, such as numbers, enabling comparison operators.
+- The containment operator allows checking if a JSON field or array includes particular values.
+
+These features provide greater flexibility for querying JSON data.
+
+
+#### Casting
```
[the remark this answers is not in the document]
asafbss: @fzdarsky They will lose valuable capabilities; see the example. I believe those who want to delve deeper into advanced functionality would actually appreciate this.

In any case, most clients will likely use the basics, but we shouldn't limit those who want to become Flightctl experts.

docs/user/field-selectors.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### Containment Operator
+
+The containment operator `@>` matches if the value at the specified key path contains the provided JSON value.
+
+This operator can also handle JSON arrays and can be used to check if a JSON array contains specific JSON path/value entries at the top level.
+
+For example:
+```bash
+flightctl get devices --field-selector 'spec.applications @> [{"image":"quay.io/rh_ee_camadorg/oci-app"}]'
```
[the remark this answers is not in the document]
asafbss: @fzdarsky I initially started with =~ (like) for the Alias field. This introduces some challenges: is it a prefix match, suffix match, or both? Additionally, the JSONB containment operator is well-defined. The term 'Contains' is stronger and is ultimately used for JSON, arrays, and strings in a single operator.

api/v1alpha1/openapi.yaml
fzdarsky: Yes, but as you can see in the generated `types.gen.go` this makes DeviceMetadata an independent type, not a type composed of ObjectMeta and the Alias.

docs/user/field-selectors.md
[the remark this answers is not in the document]
fzdarsky: > @fzdarsky, we already expose the Status field through StatusFilter, which is currently supported by the API.

That's fine. I didn't say we shouldn't support anything under `.status`, I'm wondering whether we should start with an allow-list containing a small number of fields that we can expand according to user demand, rather than opening up everything forever (as we won't be able to reduce the list later).

> Modifying the Spec or Status fields could also disrupt clients that use the Go client package to query these structures, as well as CLI with oyaml and ojson that manually parse them.

True. But that's not a good analogy because a) the client-go package can hide changes through defaulting, up/down-conversion, etc. and b) it means we don't place the burden on the user but own it ourselves.

Also, remember that a client-go user would be a software developer familiar with handling API versioning. The user of the field selector would be a Linux SysAdmin.

docs/user/field-selectors.md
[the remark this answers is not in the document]
fzdarsky: The example use case is valuable, I agree. My question was more why we don't simply allow `status.lastSeen < 2024-11-07T00:00:00Z` and own the casting ourselves?

docs/user/field-selectors.md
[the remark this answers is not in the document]
fzdarsky: There isn't a single definition for `=~`. Most commonly it's a regex match (Perl, Bash) and then prefix match or suffix match wouldn't even be an issue, but other places define it as case-insensitive string match. And even if that's a concern, we could simply chose `~` or `~=` or `∈` or whatever symbol we want. Really just something that's _somewhat_ intuitive and memorizable to a _non-developer_ (and non-SQL-expert ;)).

The semantic of that symbol is also up to us to define. We could simply say this is the symbol for `contains`.

docs/user/field-selectors.md
```diff
@@ -0,0 +1,135 @@
...
+| `metadata.labels`            | `StringArray`        |
+| `metadata.annotations`       | `StringArray`        |
+| `metadata.creationTimestamp` | `Timestamp`          |
+| `metadata.deletionTimestamp` | `Timestamp`          |
+| `spec`                       | `JSON`               |
+| `status`                     | `JSON`               |
+
+> **Note:** While `metadata.labels` can be queried using Field Selectors, for more extensive label querying options, consider using Label Selectors.
+
+### Usage Examples
```
fzdarsky, line 58: Fantastic examples ❤️

fzdarsky (approved):

fzdarsky (approved):

avishayt (approved):

PR #613, opened 2024-10-30
EDM-587: Fix dev Kind deployment with auth enabled

rawagner:


celdrake (approved):
Both the deployment flow with auth enabled and disabled worked well, thanks!

PR #616, opened 2024-10-30
EDM-649: Improve checks and error message for enrollment CSRs

lkatalin:
This PR adds a check that a CSR with a specific name does not already exist, and also clarifies the error message when an enrollment request cannot be auto-approved.

Additionally it removes two checks which seem extraneous (but let me know if we do need these):
- checking that the CN for an enrollment certificate is 16 chars long (AFAIU this name is allowed to be anything)
- ~~checking that the `Name` for a CSR resource is valid in kubernetes (AFAIU this name is allowed to be anything)~~

### DIFF
--- a/internal/service/certificatesigningrequest.go
+++ b/internal/service/certificatesigningrequest.go
@@ -39,10 +39,7 @@ func signApprovedCertificateSigningRequest(ca *crypto.CA, request api.Certificat
 	if u == "" {
 		u = uuid.NewString()
 	}
-	csr.Subject.CommonName, err = crypto.BootstrapCNFromName(u)
-	if err != nil {
-		return nil, fmt.Errorf("cn: %s does not meet requirement: %w", u, err)
-	}
+	csr.Subject.CommonName = crypto.BootstrapCNFromName(u)
 
 	expiry := DefaultEnrollmentCertExpirySeconds
 	if request.Spec.ExpirationSeconds != nil {
--- a/internal/service/certificatesigningrequest.go
+++ b/internal/service/certificatesigningrequest.go
@@ -262,7 +262,8 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, r
 			approveResp, _ := h.ApproveCertificateSigningRequest(ctx, approveReq)
 			_, ok := approveResp.(server.ApproveCertificateSigningRequest200JSONResponse)
 			if !ok {
-				return server.ReplaceCertificateSigningRequest400JSONResponse{Message: "CSR created but could not be approved"}, nil
+				msg := fmt.Sprintf("enrollment CSR for %s could not be auto-approved: %s", request.Name, approveResp)
+				return server.ReplaceCertificateSigningRequest400JSONResponse{Message: msg}, nil
 			}
 		}
 		if created {
--- a/internal/crypto/cert.go
+++ b/internal/crypto/cert.go
@@ -12,7 +12,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/flightctl/flightctl/internal/flterrors"
 	oscrypto "github.com/openshift/library-go/pkg/crypto"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
@@ -23,11 +22,8 @@ const ClientBootstrapCommonNamePrefix = "client-enrollment-"
 const AdminCommonName = "flightctl-admin"
 const DeviceCommonNamePrefix = "device:"
 
-func BootstrapCNFromName(name string) (string, error) {
-	if len(name) < 16 {
-		return "", flterrors.ErrCNLength
-	}
-	return ClientBootstrapCommonNamePrefix + name, nil
+func BootstrapCNFromName(name string) string {
+	return ClientBootstrapCommonNamePrefix + name
 }
 
 func CNFromDeviceFingerprint(fingerprint string) (string, error) {


internal/service/certificatesigningrequest.go
```diff
@@ -243,6 +240,13 @@ func (h *ServiceHandler) PatchCertificateSigningRequest(ctx context.Context, req
 func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, request server.ReplaceCertificateSigningRequestRequestObject) (server.ReplaceCertificateSigningRequestResponseObject, error) {
 	orgId := store.NullOrgId
 
+	// check that the CSR does not already exist
+	_, err := h.store.CertificateSigningRequest().Get(ctx, orgId, request.Name)
+	if err != flterrors.ErrResourceNotFound {
```
hexfusion, line 245: technically nil is also true here right?
```suggestion
	if err != nil {
	  if !errors.Is(flterrors.ErrResourceNotFound) {
```

api/v1alpha1/validation.go
```diff
@@ -236,7 +236,6 @@ func (r EnrollmentRequestApproval) Validate() []error {
 
 func (r CertificateSigningRequest) Validate() []error {
 	allErrs := []error{}
-	allErrs = append(allErrs, validation.ValidateResourceName(r.Metadata.Name)...)
```
fzdarsky, line 239: Please leave in. All of our API resources conform to this for consistency and compatibility with K8s.

fzdarsky (changes requested):

internal/service/certificatesigningrequest.go
lkatalin: In this case we want a nil error to evaluate to true because both a nil error and a non-ResourceNotFound error both mean the resource exists. ErrResourceNotFound is the only acceptable result of CertificateSigningRequest().Get(...) to execute the rest of the logic.

api/v1alpha1/validation.go
lkatalin: Okay no problem.

internal/service/certificatesigningrequest.go
hexfusion: Ok then I would still recommend using errors.Is so that the error in the future could be `%w` wrapped and this eval would fail. Your call on that but it's a good practice in general.
```
if !errors.Is(err, flterrors.ErrResourceNotFound)
```

lkatalin: Good suggestion, updated. I'm still getting used to this go syntax.

lkatalin:
@hexfusion @fzdarsky Any other thoughts here? Is this good to merge after a rebase?

hexfusion (approved):

lkatalin:
@fzdarsky Does this look fine to merge?

internal/service/certificatesigningrequest.go
```diff
@@ -244,6 +241,13 @@ func (h *ServiceHandler) PatchCertificateSigningRequest(ctx context.Context, req
 func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, request server.ReplaceCertificateSigningRequestRequestObject) (server.ReplaceCertificateSigningRequestResponseObject, error) {
 	orgId := store.NullOrgId
 
+	// check that the CSR does not already exist
+	_, err := h.store.CertificateSigningRequest().Get(ctx, orgId, request.Name)
+	if !errors.Is(err, flterrors.ErrResourceNotFound) {
+		msg := fmt.Sprintf("could not create CSR resource: CSR with name %s already exists", request.Name)
+		return server.ReplaceCertificateSigningRequest409JSONResponse{Message: msg}, nil
+	}
+
```
fzdarsky, line 250: A PUT verb on a resource has the behaviour "CreateOrUpdate", so it shouldn't be an error if the CSR already exists?

internal/service/certificatesigningrequest.go
```diff
@@ -265,7 +269,8 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, r
 			approveResp, _ := h.ApproveCertificateSigningRequest(ctx, approveReq)
 			_, ok := approveResp.(server.ApproveCertificateSigningRequest200JSONResponse)
 			if !ok {
-				return server.ReplaceCertificateSigningRequest400JSONResponse{Message: "CSR created but could not be approved"}, nil
+				msg := fmt.Sprintf("enrollment CSR for %s could not be auto-approved: %s", request.Name, approveResp)
+				return server.ReplaceCertificateSigningRequest400JSONResponse{Message: msg}, nil
```
fzdarsky, line 273: A HTTP 400 (Bad Request) error is for when the API request isn't valid, not for when there are errors during handling of the request.

internal/service/certificatesigningrequest.go
lkatalin: There are problems with auto-approving the CSR twice, so I will need to revisit this to see if there is another alternative besides throwing the error.

internal/service/certificatesigningrequest.go
lkatalin: We could grow a 500 error here, but the API currently only has 200, 201, 400, 401, 404, and 409.

lkatalin: @fzdarsky This is already a 400 error so that is not a new thing I am introducing in this PR. I am planning to submit a separate PR that updates the error codes where needed. Can this be merged as-is?

internal/service/certificatesigningrequest.go
lkatalin: Update: I am going to make these changes in a separate PR.

lkatalin:
Ping on merging this @fzdarsky @hexfusion :pray:

internal/service/certificatesigningrequest.go
fzdarsky: Sure. The PR removes the length requirement, so that'll be helpful. It's mainly the change here that'll need to be undone in a follow-up.

fzdarsky (approved):

PR #622, opened 2024-10-31
EDM-240: docs: add compose requirements to Containerfile

hexfusion:


celdrake (approved):

docs/user/building-images.md
```diff
@@ -97,6 +97,11 @@ RUN dnf -y copr enable @redhat-et/flightctl centos-stream-9-x86_64 && \
     dnf -y clean all && \
     systemctl enable flightctl-agent.service
 
+# compose application support
```
fzdarsky, line 100: I’d like to ensure users users can distinguish between minimum requirement and options, please.

Could we clarify in the comments something lime “# Optionally, if you want support for podman-compse applications, install podman-compose / uncomment below”?

fzdarsky: Also, should we then copy to getting started.md, too?

hexfusion: updated

PR #628, opened 2024-11-04
NO-ISSUE: device: improve context cancel handling

hexfusion:
This PR improves the handling of context cancellation around image upgrades. When we apply the upgrade a SIGINT is sent to the agent process. This change helps to halt future progress by returning when the ctx error is observed. The net result can be an invalid upgrade.


hexfusion:
combined with #632

PR #629, opened 2024-11-04
Update ui chart to 0.3.0-rc2

mangelajo:


deploy/helm/flightctl/Chart.yaml
```diff
@@ -10,6 +10,6 @@ dependencies:
     version: "0.0.1"
   - name: flightctl-ui
     condition: ui.enabled
-    version: "0.3.0-rc1"
+    version: "0.3.0-rc2"
```
amir-yogev-gh, line 13: Should be 0.3.1.rc2

amir-yogev-gh (changes requested):

deploy/helm/flightctl/Chart.yaml
mangelajo: why? we have not released 0.3.0 yet :)

This is still "I am trying to release 0.3.0", my 2nd candidate. At least for all the Red Hat standards I have been exposed too. I understand this could be different in other companies.

fzdarsky: Yup. It's the second _candidate_ that might be tagged 0.3.0 eventually, i.e. 0.2.0 < 0.3.0-rc1 < 0.3.0-rc2 < or = 0.3.0.

celdrake (approved):

deploy/helm/flightctl/Chart.yaml
oglok: We want to release 0.3.0, and we will be tagging as many release candidates for QE to test as needed, if you change the version, the scope should be different, and we're not introducing features, only bugfixes. So I think it's just 0.3.0-rc2.

mangelajo:
had to update the commit, I had not signed last update. Sorry about that.

PR #630, opened 2024-11-04
EDM-622: Fix local registry deploy targets

oglok:
Without the deploy-e2e-extras the local registry won't be deployed on Kind, so agent-vm scripts will fail.

### DIFF
--- a/deploy/deploy.mk
+++ b/deploy/deploy.mk
@@ -6,7 +6,7 @@ cluster: bin/e2e-certs/ca.pem
 clean-cluster:
 	kind delete cluster
 
-deploy: cluster build deploy-helm prepare-agent-config
+deploy: cluster build deploy-helm deploy-e2e-extras prepare-agent-config
 
 deploy-helm: git-server-container flightctl-api-container flightctl-worker-container flightctl-periodic-container
 	kubectl config set-context kind-kind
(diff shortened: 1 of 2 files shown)


docs/developer/README.md
```diff
@@ -64,7 +64,7 @@ this only works on a Linux host.
 ```
 # will create the cluster, and the agent config files in bin/agent which will be embedded in the image
 make deploy
-make agent-vm agent-vm-console # user/password is user/user
+make deploy-e2e-extras agent-vm agent-vm-console # user/password is user/user
```
hexfusion, line 67: thanks for doing this PR I just wonder if the registry should just be part of deploy. as a dev this is an extra step we need to do fairly often.

oglok: I guess that makes sense. wdyt @mangelajo ?

mangelajo: make it a dependency to deploy? that would make sense I think, I don't remember now if there was any technical reason why I didn't add it to deploy. I know I split it from the deploy script to be able to do it to a target cluster instead (OpenShift),

https://github.com/flightctl/flightctl/blob/main/deploy/deploy.mk#L9

We could add it here right before the prepare-agent

oglok: Updated!

mangelajo (approved):

PR #634, opened 2024-11-05
EDM-668: Change verification to Kubernetes qualified name verification

asafbss:
Validation should use https://pkg.go.dev/k8s.io/apimachinery@v0.31.1/pkg/util/validation#IsQualifiedName instead of  
https://pkg.go.dev/k8s.io/apimachinery@v0.31.1/pkg/api/validation#NameIsDNSSubdomain

Is there any reason why 'NameIsDNSSubdomain' was used in the first place?
IsQualifiedName includes NameIsDNSSubdomain.

### DIFF
--- a/api/v1alpha1/validation.go
+++ b/api/v1alpha1/validation.go
@@ -170,7 +170,7 @@ func (r ResourceAlertRule) Validate(specSampleInterval string) []error {
 func (c GitConfigProviderSpec) Validate() []error {
 	allErrs := []error{}
 	allErrs = append(allErrs, validation.ValidateGenericName(&c.Name, "spec.config[].name")...)
-	allErrs = append(allErrs, validation.ValidateResourceNameReference(&c.GitRef.Repository, "spec.config[].gitRef.repository")...)
+	allErrs = append(allErrs, validation.ValidateResourceDNSNameReference(&c.GitRef.Repository, "spec.config[].gitRef.repository")...)
 	allErrs = append(allErrs, validation.ValidateGitRevision(&c.GitRef.TargetRevision, "spec.config[].gitRef.targetRevision")...)
 	allErrs = append(allErrs, validation.ValidateString(&c.GitRef.Path, "spec.config[].gitRef.path", 0, 2048, nil, "")...)
 	return allErrs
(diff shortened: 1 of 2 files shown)


asafbss:
Can a resource sync name can be that long?

=== Failed
=== FAIL: internal/service TestResourceSyncCreateWithLongNames (0.00s)
    resourcesync_test.go:98: 
        	Error Trace:	/home/runner/work/flightctl/flightctl/internal/service/resourcesync_test.go:98
        	Error:      	Not equal: 
        	            	expected: server.ReplaceResourceSync200JSONResponse(server.ReplaceResourceSync200JSONResponse{ApiVersion:"v1", Kind:"ResourceSync", Metadata:v1alpha1.ObjectMeta{Annotations:(*map[string]string)(nil), CreationTimestamp:<nil>, DeletionTimestamp:<nil>, Generation:(*int64)(nil), Labels:(*map[string]string)(0xc0000263c8), Name:(*string)(0xc0008d1c30), Owner:(*string)(nil), ResourceVersion:(*string)(nil)}, Spec:v1alpha1.ResourceSyncSpec{Path:"/foo", Repository:"01234567890123456789012345678901234567890123456789012345678901234567890123456789", TargetRevision:"main"}, Status:(*v1alpha1.ResourceSyncStatus)(nil)})
        	            	actual  : server.ReplaceResourceSync400JSONResponse(server.ReplaceResourceSync400JSONResponse{Message:"metadata.name: Invalid value: \"01234567890123456789012345678901234567890123456789012345678901234567890123456789\": name part must be no more than [63](https://github.com/flightctl/flightctl/actions/runs/11683563640/job/32533010018?pr=634#step:5:64) characters\nspec.repository: Invalid value: \"01234567890123456789012345678901234567890123456789012345678901234567890123456789\": name part must be no more than 63 characters"})
        	Test:       	TestResourceSyncCreateWithLongNames

DONE 318 tests, 1 failure in 109.870s

celdrake:
@asafbss This PR wouls also require UI adaptations Please don't merge until the UI is aligned. I'll keep an eye on the PR.

fzdarsky:
IDK whether it's a solid reason, but here's what was my thinking behind it:

Yes, there are resource types that allow QualifiedName, but that's resource type specific:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L247-L321

Pods, Nodes, Secrets, etc. are all limited to NameIsSubDomain, because their names are being used in DNS look-ups and thus need to follow RFC1123 (see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

So when picking one, I made the analogy Node <=> Device and chose to err on the conservative side. Also, such that any FC name could become a valid K8s name if needed.

asafbss:
Thanks, @fzdarsky!  
I'm closing this PR. The reason behind it was the `Alias` field, which moved back as a label value and is verified by `ValidateLabels->IsValidLabelValue`, which uses the qualified name regex.

PR #637, opened 2024-11-06
EDM-760: Document device statuses

fzdarsky:
Documents the Device API's `device.status.summay`, `device.status.updated`, and`device.status.applicationsSummary` statuses from the p.o.v. of the service.

Defines possible status values, their meaning and formal definitions. Also adds a state chart for the observable transitions between statuses and states.

Depends on implementation in PR #668.

### DIFF
--- a/docs/user/device-api-statuses.md
+++ b/docs/user/device-api-statuses.md
@@ -1,0 +1,236 @@
+# Device API Statuses
+
+The following sections document the various statuses reported by Flight Control's Device API. These statuses represent the _service's view_ of the managed device, which is the _last reported_ status from the device agent augmented by service's local context and user policies. The _last reported_ status in turn may be outdated relative to the agent's _current status_ due to reporting delays.
+
+## Device Status
+
+The Device Status represents the availability and health of the device's hardware resources and operating system.
+
+The `device.status.summary` field can have the following values:
+
+| Status | Description | Formal Definition<sup>1</sub> |
+| ------ | ----------- | ----------------------------- |
+| `Online` | All hardware resources and operating system services are reported to be healthy. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∈{Healthy}` |
+| `Degraded` | One or more hardware resources or operating system services are reported to be degraded but in a still functional or recovering state. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∉{Error, Critical} && ∃ r∈{CPU, Memory, Disk}, status.resources[r]∈{Degraded}` |
+| `Error` | One or more hardware resources or operating system services are reported to be in error or critical state. | `!deviceIsDisconnected && !deviceIsRebooting && ∃ r∈{CPU, Memory, Disk}, status.resources[r]∈{Error, Critical}` |
+| `Rebooting` | The device is rebooting. | `!deviceIsDisconnected && deviceIsRebooting` |
+| `Offline` | The device is disconnected from the service but may still be running. | `deviceIsDisconnected` |
+
+<sup>1</sup> For the detailed definitions derived from the device specs and statuses, see [Helper Definitions](#helper-definitions).
+
+The following state diagram shows the possible transitions between device statuses, including when the corresponding device lifecycle hooks would be called.
+
+```mermaid
+stateDiagram
+    classDef transientNotReported stroke-dasharray:5 5
+    class ActivatingConfig transientNotReported
+
+    classDef notYetSupported fill:white
+    class PoweringOff notYetSupported
+    class PoweredOff notYetSupported
+    class PoweringOn notYetSupported
+
+    direction LR
+
+    state Unknown {
+        Offline
+    }
+    state Known {
+        state PoweredOn {
+            direction LR
+
+            [*] --> Online
+            Online --> Degraded
+            Online --> Error
+
+            Degraded --> Online
+            Degraded --> Error
+
+            Error --> Online
+            Error --> Degraded
+
+            state beforeRebootingHook <<choice>>
+            state afterRebootingHook <<choice>>
+            Online --> beforeRebootingHook
+            Degraded --> beforeRebootingHook
+            Error --> beforeRebootingHook
+            beforeRebootingHook --> Rebooting: beforeRebooting hook
+            Rebooting --> afterRebootingHook
+            afterRebootingHook --> Online: afterRebooting hook
+        }
+
+        [*] --> PoweredOn
+        PoweredOn --> PoweringOff
+        PoweringOff --> PoweredOff
+        PoweredOff --> PoweringOn
+        PoweringOn --> PoweredOn
+    }
+
+    [*] --> Unknown
+    Unknown --> Known
+    Known --> Unknown
+```
+
+## Device Update Status
+
+The Device Update Status represents whether the device's currently running specification (OS, configuration, applications, etc.) matches the user's intent as expressed via the device spec or the fleet's device template.
+
+The `device.status.updated.status` field can have the following values:
+
+| Status | Description | Formal Definition<sup>1</sub> |
+| ------ | ----------- | ----------------------------- |
+| `UpToDate` | The device is updated to its device spec. If the device is member of a fleet, its device spec is at the same template version as its fleet's device template. | `!deviceIsUpdating && deviceIsUpdatedToDeviceSpec && (deviceIsNotManaged \|\| deviceIsUpdatedToFleetSpec)` |
+| `Updating` | The device is in the process of updating to its device spec. | `deviceIsUpdating` |
+| `OutOfDate` | The device is not updating and either not updated to its device spec or - if it is member of a fleet - its spec is not yet of the same template version as its fleet's device template. | `!deviceIsUpdating && (!deviceIsUpdatedToDeviceSpec \|\| (deviceIsManaged && !deviceIsUpdatedToFleetSpec))` |
+| `Unknown` | The device's agent either never reported status or its last reported status was `Updating` and the device has been disconnected since. | `deviceIsDisconnected && lastStatus == Updating` |
+
+<sup>1</sup> For the detailed definitions derived from the device specs and statuses, see [Helper Definitions](#helper-definitions).
+
+The `device.status.conditions.Updating.Reason` field contains the current state of the update in progress and can take the following values:
+
+| Update State | Description |
+| ------------ | ----------- |
+| `Preparing` | The agent is validating the desired device spec and downloading dependencies. No changes have been made to the device's configuration yet. |
+| `ReadyToUpdate` | The agent has validated the desired spec, downloaded all dependencies, and is ready to update. No changes have been made to the device's configuration yet. |
+| `ApplyingUpdate` | The agent has started the update transaction and is writing the update to disk. |
+| `Rebooting` | The agent initiated a reboot required to activate the new OS image and configuration. |
+| `ActivatingConfig` | (transient, not reported) The agent is activating the new configuration without requiring a reboot. |
+| `RollingBack` | The agent has detected an error and is rolling back to the pre-update OS image and configuration. |
+| `Updated` | The agent has successfully completed the update and the device is conforming to its device spec. Note that the device's update status may still be reported as `OutOfDate` if the device spec is not yet at the same version as the fleet's device template. |
+| `Error` | The agent failed to apply the desired spec and will not retry. The device's OS image and configuration have been rolled back to the pre-update version and have been activated. |
+
+The `device.status.updated.info` field contains a human readable more detailed information about the last state transition.
+
+The following state diagram shows the possible transitions between update statuses and states, including when the corresponding device lifecycle hooks would be called.
+
+```mermaid
+stateDiagram
+    classDef transientNotReported stroke-dasharray:5 5
+    class ActivatingConfig transientNotReported
+
+    classDef notYetSupported fill:white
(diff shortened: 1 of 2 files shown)


docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+
+// a device meets its device spec when the version of the device spec that the agent reports as running
+// equals the version rendered to the device by the service
+deviceMeetsDeviceSpec := device.status.config.renderedVersion == device.metadata.annotations.renderedVersion
+
+// a device spec is up-to-date relative to the fleet's device template if both are of the same template version.
+deviceSpecAtFleetSpec := device.metadata.annotations.templateVersion == fleet[device.metadata.owner].spec.templateVersion
+```
+
+The `device.status.updated.???` field contains the current state of the update in progress and can take the following values:
```
hexfusion, line 47: A clarification here is that the agent writes to a condition as per the data dictionary and the service writes this status. The status as described by this document is different than was is outlined.

ref. https://docs.google.com/spreadsheets/d/1xlslENyYVWL-M2Qw4lOwHOEWMXXk9cPN-tCb-IH7t3c/edit?gid=21336438#gid=21336438

fzdarsky: This document describes what the service returns via the API to the user (as described in the first paragraph), not what the agent sends to the service. So that is correct then, or?

What I'm missing is where in the status do we store the update state?

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+The Device Update Status reflects whether the device's currently running specification (OS, configuration, applications, etc.) matches the user's intent as expressed via the device spec or the fleet's device template.
+
+The `device.status.updated.status` field can have the following values:
+
+| Status | Description | Formal Definition |
+| ------ | ----------- | ----------------- |
+| `UpToDate` | The device is running according to its device spec. If the device is part of a fleet, that device spec is at the same version as the fleet's device template. | `deviceMeetsDeviceSpec && (deviceIsNotFleetManaged \|\| deviceSpecAtFleetSpec)` |
+| `Updating` | The device is in the process of updating to its device spec. | `deviceIsUpdating` |
+| `OutOfDate` | The device is not `UpToDate` and not `Updating`. | |
+| `Unkown` | The device's agent either never reported status or its last reported status was `Updating` and the device has been disconnected since. | |
```
celdrake, line 25:
```suggestion
| `Unknown` | The device's agent either never reported status or its last reported status was `Updating` and the device has been disconnected since. | |
```

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+The following sections document the various statuses reported by Flight Control's Device API. These statuses represent the _service's view_ of the managed device, which is the _last reported_ status from the device agent augmented by service's local context and user policies. The _last reported_ status in turn may be outdated relative to the agent's _current status_ due to reporting delays.
+
+## Device Status
+
+The Device Status reflects the availability and health of the device's hardware and operating system.
+
+The `device.status.summary` field can have the following values:
+
+| Status | Description | Formal Definition |
+| ------ | ----------- | ----------------- |
```
avishayt, line 12: ?

docs/user/device-api-statuses.md
avishayt, line 25: It would be unknown any time it's disconnected, right?  Regardless of the last reported status.

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+| `ReadyToUpdate` | The agent has validated the desired spec, downloaded all dependencies, and is ready to update. No changes have been made to the device's configuration yet. |
+| `ApplyingUpdate` | The agent has started the update transaction and is writing the update to disk. |
+| `Rebooting` | The agent initiated a reboot required to activate the new OS image and configuration. |
+| `ActivatingConfig` | (transient, not reported) The agent is activating the new configuration without requiring a reboot. |
+| `RollingBack` | The agent has detected and error and is rolling back to the pre-update OS image and configuration. |
+| `Updated` | The agent has successfully completed the update and the device is conforming to its device spec. Note that the device's update status may still be reported as `OutOfDate` if the device spec is not yet at the same version as the fleet's device template. |
+| `Cancelled` | The agent has cancelled the update because the desired spec was reverted to the current spec before the update process started. **[TODO: Do we keep downloaded deps?]** |
```
avishayt, line 58: Canceled with one L is more common in American English, while the version with two L's is more common in British English.  Personally I like two L's better, but we use American English generally, so we should use one.

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+// a device spec is up-to-date relative to the fleet's device template if both are of the same template version.
+deviceSpecAtFleetSpec := device.metadata.annotations.templateVersion == fleet[device.metadata.owner].spec.templateVersion
+```
+
+The `device.status.updated.???` field contains the current state of the update in progress and can take the following values:
+
+| Update State | Description |
+| ------------ | ----------- |
+| `Preparing` | The agent is validating the desired device spec and downloading dependencies. No changes have been made to the device's configuration yet. |
+| `ReadyToUpdate` | The agent has validated the desired spec, downloaded all dependencies, and is ready to update. No changes have been made to the device's configuration yet. |
```
avishayt, line 52: I guess the device will be in this state while waiting for its maintenance window?  Otherwise it will immediately move to ApplyingUpdate and this state isn't needed.

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+| `ApplyingUpdate` | The agent has started the update transaction and is writing the update to disk. |
+| `Rebooting` | The agent initiated a reboot required to activate the new OS image and configuration. |
+| `ActivatingConfig` | (transient, not reported) The agent is activating the new configuration without requiring a reboot. |
+| `RollingBack` | The agent has detected and error and is rolling back to the pre-update OS image and configuration. |
```
avishayt, line 56:
```suggestion
| `RollingBack` | The agent has detected an error and is rolling back to the pre-update OS image and configuration. |
```

docs/user/device-api-statuses.md
avishayt, line 47: Why are these statuses separate from what's defined above in device.status.updated.status?  In other words, why can't we just have a single field?  Is it because one is set by the agent and one by the service?  If so, isn't that leaking implementation details?

docs/user/device-api-statuses.md
celdrake: Not according to recent discussions https://redhat-internal.slack.com/archives/C01KER8QV33/p1730996905728849?thread_ts=1730909616.756739&cid=C01KER8QV3

fzdarsky: > It would be unknown any time it's disconnected, right? Regardless of the last reported status.
I believe we can _infer_ its status from the last report: If the agent's last report before getting disconnected doesn't have the "DeviceUpdating" condition set to True, it implies the agent is not aware of any changes to its spec since, which implies it's still in the last reported state.

docs/user/device-api-statuses.md
fzdarsky: Note: Canceled isn't implemented yet, so I'll need to remove it from the docs again until then.

docs/user/device-api-statuses.md
fzdarsky: Yes, that'd be the idea. I'd like to get alignment on the states and behaviour, even if currently that state doesn't exist.

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,126 @@
...
+
+// a device spec is up-to-date relative to the fleet's device template if both are of the same template version.
+deviceSpecAtFleetSpec := device.metadata.annotations.templateVersion == fleet[device.metadata.owner].spec.templateVersion
+```
+
+The `device.status.updated.???` field contains the current state of the update in progress and can take the following values:
+
+| Update State | Description |
+| ------------ | ----------- |
+| `Preparing` | The agent is validating the desired device spec and downloading dependencies. No changes have been made to the device's configuration yet. |
```
fzdarsky, line 51: As downloading can take minutes to hours, it's important the user has _some_ way of knowing a) that the device is downloading and b) the progress of downloading. IIRC, we said this would eventually be in the Info text "Downloading (x %)". IOW, we don't need to split this into Downloading + X, right?

docs/user/device-api-statuses.md
fzdarsky: This is why it's still in "Draft" 😄

docs/user/device-api-statuses.md
avishayt: If the device died or was disconnected for any reason, the state is "unknown".  I agree that it's different than a node that didn't return from reboot, and we can indicate that.

docs/user/device-api-statuses.md
fzdarsky: [x] added

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,237 @@
...
+## Device Status
+
+The Device Status represents the availability and health of the device's hardware resources and operating system.
+
+The `device.status.summary` field can have the following values:
+
+| Status | Description | Formal Definition<sup>1</sub> |
+| ------ | ----------- | ----------------------------- |
+| `Online` | All hardware resources and operating system services are reported to be healthy. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∈{Healthy}` |
+| `Degraded` | One or more hardware resources or operating system services are reported to be degraded but in a still functional or recovering state. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∈{Error, Critical} && ∃ r∈{CPU, Memory, Disk}, status.resources[r]∈{Degraded}` |
```
celdrake, line 14:
```suggestion
| `Degraded` | One or more hardware resources or operating system services are reported to be degraded but in a still functional or recovering state. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∉{Error, Critical} && ∃ r∈{CPU, Memory, Disk}, status.resources[r]∈{Degraded}` |
```

fzdarsky: Very good eye, thanks a lot!
[x] fixed

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,237 @@
...
+
+The Device Status represents the availability and health of the device's hardware resources and operating system.
+
+The `device.status.summary` field can have the following values:
+
+| Status | Description | Formal Definition<sup>1</sub> |
+| ------ | ----------- | ----------------------------- |
+| `Online` | All hardware resources and operating system services are reported to be healthy. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∈{Healthy}` |
+| `Degraded` | One or more hardware resources or operating system services are reported to be degraded but in a still functional or recovering state. | `!deviceIsDisconnected && !deviceIsRebooting && ∀ r∈{CPU, Memory, Disk}, status.resources[r]∉{Error, Critical} && ∃ r∈{CPU, Memory, Disk}, status.resources[r]∈{Degraded}` |
+| `Error` | One or more hardware resources or operating system services are reported to be in error or critical state. | `!deviceIsDisconnected && !deviceIsRebooting && ∃ r∈{CPU, Memory, Disk}, status.resources[r]∈{Error, Critical}` |
```
hexfusion, line 15: today there is no resources status as per the Data Dictionary and I am not sure when that would happen should we remove or?

docs/user/device-api-statuses.md
```diff
@@ -0,0 +1,237 @@
...
+| `ApplyingUpdate` | The agent has started the update transaction and is writing the update to disk. |
+| `Rebooting` | The agent initiated a reboot required to activate the new OS image and configuration. |
+| `ActivatingConfig` | (transient, not reported) The agent is activating the new configuration without requiring a reboot. |
+| `RollingBack` | The agent has detected an error and is rolling back to the pre-update OS image and configuration. |
+| `Updated` | The agent has successfully completed the update and the device is conforming to its device spec. Note that the device's update status may still be reported as `OutOfDate` if the device spec is not yet at the same version as the fleet's device template. |
+| `Canceled` | The agent has canceled the update because the desired spec was reverted to the current spec before the update process started. **[TODO: Do we keep downloaded dependencies?]** |
+| `Error` | The agent failed to apply the desired spec and will not retry. The device's OS image and configuration have been rolled back to the pre-update version and have been activated. **[TODO: Do we keep downloaded dependencies?]** |
```
hexfusion, line 101: In the case that the agent can retry I currently have a `Retry`

fzdarsky: `Retry` doesn't exist in the diagram below and I'm not sure where we'd add this state and transitions. Would it even be a separate state rather than a transition back into the same or a previous state?

`Error` as state is important, because we need to signal to users that the update failed finally (i.e. after max retries) and so this version of the spec will never be reconciled again.

hexfusion: > Could you describe a scenario in which an update would fail but be successful a second time?
An example would networking failure during image prefetch. In this case we have hit a retryable error.

hexfusion: Sure I guess I was thinking the user would like to know that the agent failed for a retryable error. But maybe that is just bug food and the user should not need to know about retry.

hexfusion: we can drop the update in the agent.

avishayt (approved):

PR #638, opened 2024-11-06
EDM-648: Add validationSuffix for HTTP repositories

oglok:
- Adds a validationSuffix to HTTP repos
- Adds a new repotester type
- Refactoring of libraries


oglok:
It requires to write tests for the repotester file, but I can do it in a follow up so we can get this into 0.3.0

oglok:
@celdrake this will require to add a new field to the http repo structure

celdrake (approved):
We'll catch up in the UI soon, but since it's an optional field, there's no harm in merging Backend first.

mangelajo (approved):

celdrake (approved):

PR #641, opened 2024-11-07
EDM-673: Enable exposing APIs via Gateway

rawagner:
I've updated the `make deploy` to also configure Gateway (Contour) on KIND.

`GATEWAY=true make deploy`

The TLS APIs are then exposed via `4443` port and HTTP via `4480` (ie Keycloak).

The real-world deployments will require setting `global.gatewayClass` that referenes cluster-scoped GatewayClass CR like [this one](https://github.com/flightctl/flightctl/pull/641/files#diff-8ce7141bc2838cb52b1b5e61f5752bc7058b7af1491fe498e896210b8a440f6f)


amir-yogev-gh:
Is this related to the API Gateway we need to use for the integration with AAP?
If it does, can you connect the dots and explain?

fzdarsky:
> Is this related to the API Gateway we need to use for the integration with AAP? If it does, can you connect the dots and explain?

It is not related, no. This refers to the Kubernetes Gateway API that'll replace the Kubernetes Ingress and OpenShift Routes API, respectively.

deploy/helm/flightctl/templates/flightctl-api-gateway-route.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: TLSRoute
+metadata:
+  name: flightctl-api-route
+  namespace: {{ .Release.Namespace }}
+spec:
+  parentRefs:
+    - name: flightctl-gateway
+      sectionName: api
+  hostnames:
+    - api.{{ include "flightctl.getBaseDomain" . }}
```
mangelajo, line 12: does this one allow pass-through?

deploy/helm/flightctl/templates/flightctl-gateway.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - name: keycloak
+    hostname: auth.{{ include "flightctl.getBaseDomain" . }}
+    {{- if .Values.global.baseDomainTls.cert }}
+    port: {{ .Values.global.gatewayPorts.tls }}
+    protocol: HTTPS
+    tls:
+      mode: Terminate
+      certificateRefs:
+      - kind: Secret
+        name: gateway-cert
```
mangelajo, line 20: ohhh yes

deploy/helm/flightctl/templates/flightctl-gateway.yaml
```diff
@@ -0,0 +1,48 @@
...
+    {{- else }}
+    port: {{ .Values.global.gatewayPorts.http }}
+    protocol: HTTP
+    {{- end }}
+  {{- end }}
+  {{- if .Values.ui.enabled }}
+  - name: ui
```
mangelajo, line 27: would it make sense to have the UI have it's own gateway?

test/scripts/gateway/install-gateway.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
```
hexfusion, line 2: I general would like to see us standardize on fail fast with bash scripts. WDYT?

```suggestion
set -eo pipefail
```

deploy/helm/flightctl/templates/flightctl-gateway.yaml
rawagner: I tried but If I move the UI specific bits to a separate gateway, it doesn't work and I dont see any error anywhere :/

test/scripts/gateway/install-gateway.sh
rawagner: sure, added

mangelajo (approved):
Looks great thanks

PR #649, opened 2024-11-13
NO-ISSUE: Improve UpdateDeviceWithRetries to handle all errors

mangelajo:
If the API server returns an error (outside of conflict), print the details, and fail the test instead of blindly accepting and continuing.

### DIFF
--- a/test/harness/e2e/harness.go
+++ b/test/harness/e2e/harness.go
@@ -284,10 +284,20 @@ func (h *Harness) UpdateDeviceWithRetries(deviceId string, updateFunction func(*
 			logrus.Warningf("conflict updating device: %s", deviceId)
 			return errors.New("conflict")
 		}
+
 		// if other type of error happens we fail
 		Expect(err).ToNot(HaveOccurred())
+
+		// response code 200 = updated, we are expecting to update... something else is unexpected
+		if resp.StatusCode() != 200 {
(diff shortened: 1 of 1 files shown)


test/harness/e2e/harness.go
```diff
@@ -284,10 +284,20 @@ func (h *Harness) UpdateDeviceWithRetries(deviceId string, updateFunction func(*
...
+		if resp.StatusCode() != 200 {
+			logrus.Errorf("Unexpected http status code received: %d", resp.StatusCode())
+			logrus.Errorf("Unexpected http response: %s", string(resp.Body))
+		}
+		// make the test fail and stop the Eventually loop if at this point we didn't have a 200 response
+		Expect(resp.StatusCode()).Should(Equal(200))
+
 		return nil
-	}, TIMEOUT, "2s").WithArguments(updateFunction).Should(BeNil())
+	}, TIMEOUT, "1s").WithArguments(updateFunction).Should(BeNil())
```
mangelajo, line 300:
```suggestion
	}, TIMEOUT, "1s").WithArguments(updateFunction).Should(BeNil())
```

Updating the retry interval to 1s, i.e. the agent updates every 2s with status...

So we always have at least one cycle in the middle if the agent were to be updating relentlessly

hexfusion (approved):

PR #653, opened 2024-11-15
EDM-424: expose local system time zone via status

hexfusion:
This PR extends systemInfo to include the time zone.

```
[root@localhost ~]# date
Sat Nov 16 15:04:01 UTC 2024
```
```
    systemInfo:
      architecture: amd64
      bootID: |
        0c022745-96af-492f-a1e3-4460f7b1ae0f
      operatingSystem: linux
      timeZone: UTC
```

```
[root@localhost ~]# timedatectl set-timezone EST

[root@localhost ~]# date
Sat Nov 16 10:06:43 EST 2024

[root@localhost ~]# ls -l /etc/localtime
lrwxrwxrwx. 1 root root 25 Nov 16 10:06 /etc/localtime -> ../usr/share/zoneinfo/EST

```

```
  systemInfo:
    architecture: amd64
    bootID: |
      d3650636-304a-4d04-9d88-2660b4365942
    operatingSystem: linux
    timeZone: EST
```


hexfusion:
cc @celdrake fyi

fzdarsky:
See PR #654 on open question re tz identifier vs abbreviation.

Also, is this just informational or something the service would need to perform other functions?

hexfusion:
> Also, is this just informational or something the service would need to perform other functions?

I think the service will want this information to validate TZ in some cases. For example, if an entire fleet of devices should have the same TZ this would provide observability.

But we do not offer an API to adjust the TZ today and I am not sure we should. OpenShift warns against changing the timezone from UTC[1],[2] as it is not supported. The reasoning is fairly reasonable.

[1] https://access.redhat.com/solutions/4994241
[2] https://access.redhat.com/solutions/5487331

hexfusion:
This has resolved via https://github.com/flightctl/flightctl/pull/737

PR #654, opened 2024-11-15
EDM-424: add update policy api and validation

hexfusion:
This PR adds phase one up Update Policy API as described in the design doc[1]

```sh
fleet: applying examples/fleet-policy.yaml/default: 400 Bad Request
{"message":"invalid time zone: EST"}
```

```sh
fleet: applying examples/fleet-policy.yaml/default: 400 Bad Request
{"message":"startGraceDuration exceeds the cron interval between schedule times: 30000h0m0s"}
```

```yaml
spec:
  updatePolicy:
    downloadSchedule: 
      timeZone: America/New_York
      at: "0 0-8,16-23 * * *"
      startGraceDuration: "10m"
    updateSchedule: 
      timeZone: America/New_York
      at: "0 1 * * *" # At 1:00 AM every day
      startGraceDuration: "12h"
```

[1]https://docs.google.com/document/d/1zkCGT7KCea1Ec-Dluu8VAzhmVM6HrnH2arJwJFLdGNo/edit?tab=t.0


api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,56 @@ components:
               description: 'The directory in which the executable will be run from if it is left empty it will run from the users home directory.'
           required:
             - run
+    DeviceUpdatePolicySpec:
+      type: object
+      properties:
+        download:
```
fzdarsky, line 3037: Should this be “downloadSchedule” or “downloadWindow”?

Otherwise, it’d seem that this field is subdividing “updatePolicy” by phases (download, update, rollback, …), meaning we’d put other policies related to “download” besides schedule (rate limits, retention policies, etc) here as well. Same with “update” and potential other phases (rollback?). IMO, that’d just increase nesting and be less extensible.

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,56 @@ components:
...
+          $ref: '#/components/schemas/UpdatePolicy'
```
fzdarsky, line 3038: DownloadPolicy? Assuming that this isn’t just schedule and would need to differ from update now/in the future.

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,56 @@ components:
...
+        update:
```
fzdarsky, line 3039: Should this be “updateSchedule” or “updateWindow”? (see above)

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,56 @@ components:
...
+      properties:
+        schedule:
+          $ref: '#/components/schemas/UpdateSchedule'
+      description: "A policy defining the schedule for downloads or updates."
+    UpdateSchedule:
+      type: object
+      properties:
+        timezone:
+          type: string
+          description: "Timezone for the schedule (e.g., GMT) the defaults to the local system timezone"
```
fzdarsky, line 3053: Do we expect time zones to be specified by the tzdata identifier (e.g. America/New_York) or abbreviation (e.g. EST, EDT)? The latter are ambiguous.

Should we default the local timezone instead of making it optional (avoid pointer)? Should default be an explicit “local” instead of “”?

hexfusion: will align with https://data.iana.org/time-zones/theory.html#naming

api/v1alpha1/validation.go
```diff
@@ -336,7 +373,7 @@ func (r ResourceSync) Validate() []error {
 
 func (l *LabelSelector) Validate() []error {
 	if l != nil && l.MatchExpressions == nil && l.MatchLabels == nil {
-		return []error{errors.New("At least one of [matchLabels,matchExpressions] must appear in a label selector")}
+		return []error{errors.New("at least one of [matchLabels,matchExpressions] must appear in a label selector")}
```
asafbss, line 376: :heart:

api/v1alpha1/validation.go
```diff
@@ -485,3 +522,8 @@ func getAppName(app ApplicationSpec, appType ApplicationProviderType) (string, e
 		return "", fmt.Errorf("unsupported application provider type: %s", appType)
 	}
 }
+
+func validateTimezone(tz string) error {
+	_, err := time.LoadLocation(tz)
+	return err
```
asafbss, line 528: I'm not quite familiar with 'LoadLocation' errors; we might consider wrapping them.

if _, err := time.LoadLocation(tz); err != nil {
		return fmt.Errorf("invalid timezone: %s", tz)
	}
	return nil

internal/util/validation/validation.go
```diff
@@ -94,6 +94,47 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
...
+		return nil
+	}
+
+	// break down the field into its parts
+	parts := strings.Split(field, ",")
+	for _, part := range parts {
+		if strings.Contains(part, "-") { // Validate ranges
+			rangeParts := strings.Split(part, "-")
+			if len(rangeParts) != 2 {
+				return fmt.Errorf("invalid range in %s field: %q", fieldName, field)
```
asafbss, line 108: Could it be better to quote the fieldName
return fmt.Errorf("invalid range in %q field: %s", fieldName, field)

internal/util/validation/validation.go
```diff
@@ -94,6 +94,47 @@ func ValidateString(s *string, path string, minLen int, maxLen int, patternRegex
 	return asErrors(errs)
 }
 
+func ValidateCronField(field string, min, max int, fieldName string) error {
```
asafbss, line 97: Can this be handled more efficiently using a compiled regex with capturing groups?

internal/util/validation/validation_test.go
```diff
@@ -0,0 +1,121 @@
+package validation
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateCronSchedule(t *testing.T) {
```
asafbss, line 9: :rocket:

internal/util/validation/validation.go
hexfusion: yeah copilot malfunction here :)

internal/util/validation/validation.go
hexfusion: Great question, under the hood these string evaluations are highly optimized. In general, unlike some languages like Perl regex is really not the best for performance even when compiled.  

For the test I came up with this monster which I believe doesn't even cover everything.

```
regex := regexp.MustCompile(`^
    (\*|[0-5]?\d|\d+-\d+|[0-5]?\d/\d+) # Minutes
    (\s+) 
    (\*|[0-5]?\d|\d+-\d+|[0-5]?\d/\d+) # Hours
    (\s+) 
    (\*|[1-2]?\d|3[0-1]|\d+-\d+|[1-2]?\d/\d+) # Day of month
    (\s+)
    (\*|1[0-2]|[1-9]|\d+-\d+|1[0-2]/\d+) # Months
    (\s+) 
    (\*|[0-6]|\d+-\d+|[0-6]/\d+)  # Day of week
$`)
```
Then created these benches with a simple test case

func BenchmarkValidateCronSchedule(b *testing.B) {
	require := require.New(b)
	cron := "* 1 * * *"
	for i := 0; i < b.N; i++ {
		err := ValidateCronSchedule(cron)
		require.NoError(err)
	}
}

func BenchmarkValidateCronScheduleWithRegex(b *testing.B) {
	require := require.New(b)
	cron := "* 1 * * *"
	for i := 0; i < b.N; i++ {
		err := ValidateCronScheduleWithRegex(cron)
		require.NoError(err)
	}
}
```

The output shows the regex was quite a bit slower.

```
BenchmarkValidateCronSchedule-12    	  750240	      1771 ns/op	      96 B/op	       2 allocs/op
```

```
BenchmarkValidateCronScheduleWithRegex-12    	   12692	    101958 ns/op	   39783 B/op	     210 allocs/op
```

Comment (asafbss) [reply internal/util/validation/validation.go:97]: 🚀 ! Any chance to add a comment as example of an input?

Comment (hexfusion) [reply internal/util/validation/validation.go:97]: Saw another issue this function is actually cruft from my first pass and will be removed in favor of `ValidateCronSchedule` which was used in the bench.

Comment (hexfusion) [reply internal/util/validation/validation.go]: I ended up just taking a dep on this to github.com/robfig/cron in the end

File: api/v1alpha1/openapi.yaml [hunk 10]
```diff
@@ -3031,6 +3031,64 @@ components:
...
           required:
             - run
+    DeviceUpdatePolicySpec:
+      type: object
+      properties:
+        downloadSchedule:
+          $ref: '#/components/schemas/DownloadPolicy'
+        updateSchedule:
+          $ref: '#/components/schemas/UpdatePolicy' 
+          description: "Defines the policy for downloads and updates, including schedules and rate limits."
```
hexfusion, line 3041: nit

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,63 @@ components:
...
+      type: object
+      properties:
+        timeZone:
+          type: string
+          description: >
+            Time zone identifiers follow the IANA format AREA/LOCATION, where AREA
+            represents a continent or ocean, and LOCATION specifies a particular
+            site within that area.  e.g., America/New_York, Europe/Paris.
+            Only unambiguous 3-character time zones are supported ("GMT", "UTC").
+          default: "Local"
```
hexfusion, line 3059: For reference, k8s does not consider "Local"[1] valid.

[1] https://github.com/kubernetes/kubernetes/blob/776fb24e066d54cf84a52366aeea121fa5748e27/pkg/apis/batch/validation/validation.go#L834

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,63 @@ components:
               description: 'The directory in which the executable will be run from if it is left empty it will run from the users home directory.'
           required:
             - run
+    DeviceUpdatePolicySpec:
+      type: object
+      properties:
+        downloadSchedule:
+          $ref: '#/components/schemas/DownloadPolicy'
```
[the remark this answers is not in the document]
hexfusion: or maybe this is downloadPolicy and all the fields live here as allOf

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,63 @@ components:
...
+          description: |
+            The maximum duration after the scheduled time (`at`) within which the
+            task must start. If the task cannot start within this time, it will be
+            skipped.  The duration must be a positive integer followed by a time
+            unit. Supported time units are:
+              - 's' for seconds
+              - 'm' for minutes
+              - 'h' for hours
+      required:
+        - timeZone
```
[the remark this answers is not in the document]
hexfusion: good catch

api/v1alpha1/openapi.yaml
hexfusion: done

api/v1alpha1/openapi.yaml
```diff
@@ -3666,7 +3715,7 @@ components:
           $ref: '#/components/schemas/RolloutDeviceSelection'
         successThreshold:
           $ref: '#/components/schemas/Percentage'
-        defaultUpdateTimeout:
+        updateTimeout:
```
hexfusion, line 3718: I propose this change as its not technically the default meaning it can be changed by something else it is the `updateTimeout` this is aligned with UpdatePolicy.

fzdarsky (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -3031,6 +3031,55 @@ components:
...
+          description: 'Defines the schedule for applying updates.'
+    UpdateSchedule:
+      type: object
+      description: 'Defines the schedule for automatic updates, including timing and optional timeout.'
+      properties:
+        timeZone:
+          $ref: '#/components/schemas/TimeZone'
+        at:
+          $ref: '#/components/schemas/CronExpression'
+        startGraceDuration:
```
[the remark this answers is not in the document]
hexfusion: added validation

api/v1alpha1/validation.go
```diff
@@ -485,3 +538,82 @@ func getAppName(app ApplicationSpec, appType ApplicationProviderType) (string, e
...
+// * Do not use the file name components '.' and '..'.
+// * Within a file name component, use only ASCII letters, '.', '-' and '_'.
+// * Do not use digits, as that might create an ambiguity with POSIX TZ strings.
+// * A file name component must not exceed 14 characters or start with '-'
+//
+// 0-9 and + characters are tolerated to accommodate legacy compatibility names
+var validTimeZoneCharacters = regexp.MustCompile(`^[A-Za-z\.\-_0-9+]{1,14}$`)
+
+// validateTimeZone validates the time zone string. it must be a valid IANA time zone identifier.
+func validateTimeZone(timeZone string) []error {
```
fzdarsky, line 554: We're keeping most of the reusable, API version-independent validation code in https://github.com/flightctl/flightctl/tree/main/internal/util/validation, but we can move that later if needed.

fzdarsky (approved):
LGTM.

PR #656, opened 2024-11-17
EDM-681: Allow field selector when 'summaryOnly' is enabled

asafbss:
We initially decided that summaryOnly, when listing devices, would raise an error if the end-user attempted to use flags such as limit and continue for pagination. The logic behind this decision was that summaryOnly does not return a list of devices, so pagination is not applicable.

However, field selectors, such as metadata.owner and others, should be usable with summaryOnly. Even when applying a status filter inside the field-selector query, there is no reason to disallow filtering by it.

### DIFF
--- a/internal/service/device.go
+++ b/internal/service/device.go
@@ -75,19 +75,41 @@ func (h *ServiceHandler) ListDevices(ctx context.Context, request server.ListDev
 		return server.ListDevices400JSONResponse{Message: err.Error()}, nil
 	}
 
+	statusFilter := []string{}
+	if request.Params.StatusFilter != nil {
+		for _, filter := range *request.Params.StatusFilter {
+			statusFilter = append(statusFilter, fmt.Sprintf("status.%s", filter))
+		}
+	}
+
+	filterMap, err := ConvertFieldFilterParamsToMap(statusFilter)
+	if err != nil {
+		return server.ListDevices400JSONResponse{Message: fmt.Sprintf("failed to convert status filter: %v", err)}, nil
+	}
+
+	var fieldSelector k8sselector.Selector
+	if request.Params.FieldSelector != nil {
+		if fieldSelector, err = fields.ParseSelector(*request.Params.FieldSelector); err != nil {
+			return server.ListDevices400JSONResponse{Message: fmt.Sprintf("failed to parse field selector: %v", err)}, nil
+		}
+	}
+
 	// Check if SummaryOnly is true
 	if request.Params.SummaryOnly != nil && *request.Params.SummaryOnly {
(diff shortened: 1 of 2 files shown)


internal/service/device.go
```diff
@@ -75,19 +75,41 @@ func (h *ServiceHandler) ListDevices(ctx context.Context, request server.ListDev
 		return server.ListDevices400JSONResponse{Message: err.Error()}, nil
 	}
 
+	statusFilter := []string{}
+	if request.Params.StatusFilter != nil {
```
celdrake, line 79: We could already drop support for `statusFilter` here, given that's the old way of filtering for statuses and has been replaced by `fieldSelector`.

I'd expect we'll also remove the individual filter fields from the API soon, and keep only "fieldSelector" and "labelSelector".

asafbss: Hey @celdrake,
Removing this in this PR will break the functionality.
I agree with you about removing individual filter fields from the API, but this should be handled in a dedicated ticket that includes changes to the API, CLI, and backend functionality.

@amir-yogev-gh - FYI

celdrake (commented):
Works well for the UI use case (summaryOnly + filter by labels and owner).

internal/service/device.go
celdrake: Sure, though I will already include the change in the UI PR (which hasn't been merged yet), to use "fieldSelector" instead of statusFilter.

internal/service/device.go
```diff
@@ -101,30 +123,11 @@ func (h *ServiceHandler) ListDevices(ctx context.Context, request server.ListDev
...
-			statusFilter = append(statusFilter, fmt.Sprintf("status.%s", filter))
-		}
-	}
-
-	filterMap, err := ConvertFieldFilterParamsToMap(statusFilter)
-	if err != nil {
-		return server.ListDevices400JSONResponse{Message: fmt.Sprintf("failed to convert status filter: %v", err)}, nil
-	}
-
 	cont, err := store.ParseContinueString(request.Params.Continue)
```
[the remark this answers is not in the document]
asafbss: Hey @eldar101,
Logging APIs requires careful attention, especially when dealing with externally exposed endpoints. Anything that could potentially expose our service to blasting or DoS attacks must be handled with caution. In many cases, the log level for externally exposed APIs is critical and is often defined as the access log level, which can also be managed by a reverse proxy like NGINX.

To summarize, if you’re suggesting logging API requests, this should ideally cover all API endpoints and, in my opinion, should be addressed as a separate feature request. For this particular bug fix, it falls out of scope.

PR #660, opened 2024-11-18
EDM-310: Move the CA to an async task

kot-begemot-uk:
1. Move the CA to a singleton.
2. Provide support for multiple signers/CA identities.
3. Move signing to an async task.
4. Invoke the async task instead of signing.
5. Provide support for different CA types.


internal/config/config.go
```diff
@@ -58,6 +63,35 @@ type kvConfig struct {
...
+type HashiCAConfig struct {
+	AppRole		string	`json:"approle"`
+	SecretID	string	`json:"secretid"`
+	CAURL		string	`json:"caUrl"`
+	AppRoleURL	string	`json:"approleUrl"`
+	Signer		string	`json:"signer"`
+	Role		string	`json:"role"`
+}
+
+const (
```
asafbss, line 83: enum?

type CAType int
const (
	InternalCA CAType = iota + 1
	HashiVault
)

internal/config/config.go
```diff
@@ -58,6 +63,35 @@ type kvConfig struct {
...
+	InternalCA	= 1
+	HashiVault	= 2
+)
+
+type CryptographyConfigEntry struct {
+	CAType                  uint    `json:"catype,omitempty"`
```
asafbss, line 89: type CAType int

internal/tasks/casign.go
```diff
@@ -0,0 +1,168 @@
...
+func asyncSignEnrollment(ctx context.Context, resourceRef *ResourceReference, dbStore store.Store, callbackManager CallbackManager, log logrus.FieldLogger) (int, error) {
+
+	listParams := store.ListParams{Limit:DefaultBatchSize}
+	selector, err := selector.Parse("!status.certificate")
+
+	if err != nil {
+		log.Info(err)
+		return 0, err
+	}
+	listParams.FieldSelector = selector
```
asafbss, line 65: :heart:

internal/config/config.go
kot-begemot-uk: Sure. Will make it an enum in next revision

internal/crypto/cert.go
```diff
@@ -37,32 +48,104 @@ func CNFromDeviceFingerprint(fingerprint string) (string, error) {
...
+	MakeClientCertificate(certFile, keyFile string, subject string, expiryDays int) (*TLSCertificateConfig, error)
+	IssueRequestedClientCertificate(csr *x509.CertificateRequest, expirySeconds int) ([]byte, error)
+}
+
+
+
+var caMutex = &sync.Mutex{}
+var caList map[string]CA = make(map[string]CA)
+
+func EnsureCA(id string, cfg *config.Config) (CA, bool, error) {
```
hexfusion, line 78: I think we can reduce the scope of this signature to just the CryptographyConfig.

```suggestion
func EnsureCA(id string, cfg *config.CryptographyConfig) (CA, bool, error) {
```

internal/crypto/cert.go
```diff
@@ -37,32 +48,104 @@ func CNFromDeviceFingerprint(fingerprint string) (string, error) {
...
+	signCertificate(template *x509.Certificate, requestKey crypto.PublicKey) (*x509.Certificate, error)
+	EnsureClientCertificate(certFile, keyFile string, subjectName string, expireDays int) (*TLSCertificateConfig, bool, error)
+	MakeClientCertificate(certFile, keyFile string, subject string, expiryDays int) (*TLSCertificateConfig, error)
+	IssueRequestedClientCertificate(csr *x509.CertificateRequest, expirySeconds int) ([]byte, error)
+}
+
+
+
+var caMutex = &sync.Mutex{}
+var caList map[string]CA = make(map[string]CA)
```
hexfusion, line 76: How about we encapsulate this in a struct vs exposing via global? It is generally good practice to avoid globals if possible from a maintainability and testing perspective.

``` suggestion
type CARegistry struct {
   mu sync.Mutex
   certs map[string]CA
}

func (m *CARegistry) Get(key string) (CA, bool)
...
func (m *CARegistry) Put(key string, ca CA)
```

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	template.SerialNumber = big.NewInt(serial)
+	return signCertificate(template, requestKey, ca.Config.Certs[0], ca.Config.Key)
+}
+
+func (ca *internalCA) EnsureClientCertificate(certFile, keyFile string, subjectName string, expireDays int) (*TLSCertificateConfig, bool, error) {
+	certConfig, err := GetClientCertificate(certFile, keyFile, subjectName)
+	if err != nil {
+		certConfig, err = ca.MakeClientCertificate(certFile, keyFile, subjectName, expireDays)
+		return certConfig, true, err // true indicates we wrote the files.
```
hexfusion, line 108: should we document the interface with what the values of this bool mean?

internal/tasks/casign.go
```diff
@@ -0,0 +1,168 @@
...
+
+
+	ca := crypto.GetCachedCA(crypto.DefaultCA)
+
+	//listParams.Filter = filterMap
```
hexfusion, line 119: nit

internal/crypto/cert.go
kot-begemot-uk: Yeah fine. I know it is k8 style. It am not a fan of piling init args and carrying them across the whole codebase for something that is obviously single instance and global. But so be it.

hexfusion: The choice is yours I am just pointing out an observation that allowing any function/method access via global could be problematic down the road.

internal/crypto/cert.go
```diff
@@ -37,32 +48,104 @@ func CNFromDeviceFingerprint(fingerprint string) (string, error) {
...
+	loaded = &internalCA{Config: &config, SerialGenerator: ca.SerialGenerator, Id:id,}
+	caList[id] = loaded
+	return loaded, err
 }
 
-func MakeSelfSignedCA(certFile, keyFile, serialFile, subjectName string, expiryDays int) (*CA, error) {
+func GetCachedCA(id string) (CA) {
+	caMutex.Lock()
+	defer caMutex.Unlock()
+	ca, ok := caList[id]
```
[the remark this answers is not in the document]
kot-begemot-uk: No longer an issue. I switched to the k8 pattern of passing the CA/cfg top-down instead of having a singleton CA. I would not have coded it this way, but this seems to be the prevailing way of things in k8 land.

internal/crypto/internal_ca.go
kot-begemot-uk: This bool as undocumented. It was showing if the CA was generated or loaded from storage. Sorry, missed that one. I can add the relevant documentation.

internal/tasks/casign.go
kot-begemot-uk: Caching was removed.

internal/tasks/casign.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func asyncSignCSR(ctx context.Context, resourceRef *ResourceReference, dbStore store.Store, callbackManager CallbackManager, log logrus.FieldLogger, ca crypto.CA) (int, error) {
+
+
+	listParams := store.ListParams{Limit:DefaultBatchSize}
+	selector, err := selector.Parse("!status.certificate")
+
+	if err != nil {
+		log.Info(err)
```
hexfusion, line 109: should we provide more details for the log line and if its an error should we log.Error()?

internal/tasks/casign.go
```diff
@@ -0,0 +1,162 @@
...
+			}
+			csr.Subject.CommonName, err = crypto.BootstrapCNFromName(u)
+			if err != nil {
+				continue
+			}
+			expiry := DefaultCSRExpiry
+			if ereq.Spec.ExpirationSeconds != nil {
+				expiry = int(*ereq.Spec.ExpirationSeconds)
+			}
+			cert, err := ca.IssueRequestedClientCertificate(csr, expiry)
```
hexfusion, line 147: should we log/return this err?

internal/crypto/cert.go
```diff
@@ -37,32 +46,74 @@ func CNFromDeviceFingerprint(fingerprint string) (string, error) {
...
 
-type CA struct {
-	Config *TLSCertificateConfig
+type CA interface {
+	GetConfig() *TLSCertificateConfig
+	EnsureServerCertificate(certFile, keyFile string, hostnames []string, expireDays int) (*TLSCertificateConfig, bool, error)
+	MakeAndWriteServerCert(certFile, keyFile string, hostnames []string, expireDays int) (*TLSCertificateConfig, error)
+	MakeServerCert(hostnames []string, expiryDays int, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error)
+	signCertificate(template *x509.Certificate, requestKey crypto.PublicKey) (*x509.Certificate, error)
+	EnsureClientCertificate(certFile, keyFile string, subjectName string, expireDays int) (*TLSCertificateConfig, bool, error)
```
hexfusion, line 65: Can we add docs to this interface please :pray:

internal/tasks/consumer.go
```diff
@@ -6,14 +6,15 @@ import (
...
+	"github.com/flightctl/flightctl/internal/crypto"
 	"github.com/flightctl/flightctl/pkg/k8sclient"
 	"github.com/flightctl/flightctl/pkg/queues"
 	"github.com/sirupsen/logrus"
 )
 
 const TaskQueue = "task-queue"
 
-func dispatchTasks(store store.Store, callbackManager CallbackManager, k8sClient k8sclient.K8SClient, configStorage ConfigStorage) queues.ConsumeHandler {
+func dispatchTasks(store store.Store, callbackManager CallbackManager, k8sClient k8sclient.K8SClient, configStorage ConfigStorage, ca crypto.CA) queues.ConsumeHandler {
```
hexfusion, line 17: @avishayt, @ori-amizur is this how developers should extend tasks by extending the signature? The result is folks creating globals to work around it.

internal/tasks/casign.go
```diff
@@ -0,0 +1,162 @@
...
+			if err == nil {
+				ereq.Status.Certificate = &cert
+				_, err := dbStore.CertificateSigningRequest().UpdateStatus(ctx, orgId, &ereq)
+				if err != nil {
+					log.Infof("Failed to update csr %s", err)
```
hexfusion, line 152: Should this log.Errorf()?

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,209 @@
...
+	return certConfig, false, nil
+}
+
+func (ca *internalCA) MakeAndWriteServerCert(certFile, keyFile string, hostnames []string, expireDays int) (*TLSCertificateConfig, error) {
+	server, err := ca.MakeServerCert(hostnames, expireDays)
+	if err != nil {
+		return nil, err
+	}
+	if err := server.WriteCertConfigFile(certFile, keyFile); err != nil {
+		return server, err
```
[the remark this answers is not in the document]
kot-begemot-uk: In next version.

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,209 @@
...
+		SignatureAlgorithm: csr.SignatureAlgorithm,
+
+		PublicKey:          csr.PublicKey,
+		PublicKeyAlgorithm: csr.PublicKeyAlgorithm,
+
+		Issuer: ca.Config.Certs[0].Subject,
+
+		NotBefore:    now.Add(-1 * time.Second),
+		NotAfter:     now.Add(time.Duration(expirySeconds) * time.Second),
+		SerialNumber: big.NewInt(1),
```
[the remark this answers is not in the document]
kot-begemot-uk: This i a NOOP.
It is overwritten in:

func (ca *internalCA) signCertificate(template *x509.Certificate, requestKey crypto.PublicKey) (*x509.Certificate, error) {
	// Increment and persist serial
	serial, err := ca.SerialGenerator.Next(template)
	if err != nil {
		return nil, err
	}
	template.SerialNumber = big.NewInt(serial)
	return signCertificate(template, requestKey, ca.Config.Certs[0], ca.Config.Key)
}

I have cut-and-pasted this from the old CA code. we should be able to remove it.

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,209 @@
...
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		BasicConstraintsValid: true,
+
+		AuthorityKeyId: ca.Config.Certs[0].SubjectKeyId,
+	}
+	cert, err := ca.signCertificate(template, csr.PublicKey)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", flterrors.ErrSignCert, err.Error())
```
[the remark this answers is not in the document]
kot-begemot-uk: Cut-n-paste. This was in the existing code. Fixed.

internal/tasks/callback_manager.go
```diff
@@ -114,6 +119,17 @@ func (t *callbackManager) FleetUpdatedCallback(before *model.Fleet, after *model
 	}
 }
 
+func (t *callbackManager) CSRApproved(orgId uuid.UUID) {
+	ref := ResourceReference{OrgID: orgId}
+    t.log.Info("Submitting csr")
```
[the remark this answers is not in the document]
kot-begemot-uk: Sorry, did not kill this. This was for debugging only. As any async it does not like being debugged, sticking tracing stmts is usually easier :)

internal/tasks/casign.go
kot-begemot-uk: Fair point. I will try to gather whatever is applicable.

internal/tasks/consumer.go
```diff
@@ -33,6 +34,8 @@ func dispatchTasks(store store.Store, callbackManager CallbackManager, k8sClient
 			return deviceRender(ctx, &reference, store, callbackManager, k8sClient, configStorage, log)
 		case RepositoryUpdatesTask:
 			return repositoryUpdate(ctx, &reference, store, callbackManager, log)
+		case SignerTask:
+			return asyncSign(ctx, &reference, store, callbackManager, log, ca)
```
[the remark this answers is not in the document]
kot-begemot-uk: Retry revolves around internal logic in casign.go/asyncSign and friends. 

They will process all outstanding requests in the database. If a new request is added in the meantime it will be processed as well. This allows any number of worker threads as well scaling them horizontally.

The worker continues to process until there are zero candidates in the database. 

Unfortunately, the whole thing is prone to a race. There is a window if the record is added after the "get next records" query, but before the processing loop returns and releases the lock. It is not large and it is very unlikely because this means that insertion, submission of publish request, receiving it and handling has happened faster than for _, ereq := range ereqs.Items with len(Items) == 0. 

I will be honest - I do not know how to eliminate this one short of having a "cronjob" to do a regular re-run.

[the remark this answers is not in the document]
kot-begemot-uk: To be clear - the race can happen only if there are ZERO records to be processed in the database. If there is even one, adding other while it is being processed should be handled correctly.

internal/tasks/casign.go
kot-begemot-uk: We cannot return it. At all. We can log it, but that will be buried in the log. 

This is a common problem for all async jobs which I raised on the tech call last week. We have to record/log the state of async processing somewhere where it can be accessed by an API. Either tag an array of records on Status or start a proper database based event log.

internal/tasks/casign.go
kot-begemot-uk: Yes. Plus what we already discussed for API/API accessible record of the proceedings.

internal/crypto/cert.go
kot-begemot-uk: The interface is as ugly as it gets
It was copied 1:1 from the existing struct.
I am open to suggestions in terms of revising it and/or removing some of the functions from the public exports list.

kot-begemot-uk:
I have pushed the Vault support to my fork. It is incremental on top of this one.

If we decide to merge this, it can go on top. I still have some questions on e2e there, but they can wait until it is being actively worked on.

PR #662, opened 2024-11-18
EDM-708: Reduce git access

avishayt:
The device render task currently clones git repos for each device, which is inefficient for FC and can DoS the git server.  This commit caches the git data in valkey.  It is stored in ignition format, with a key per hash and path.


asafbss (commented):
LGTM, just a minor suggestion.

asafbss (approved):

asafbss (approved):

PR #663, opened 2024-11-19
EDM-696: Support AAP Gateway auth

rawagner:
This PR adds a new auth provider 'AAP Gateway' so that we can use AAP Gateway to proxy requests to FCTL service and use that Gateway as auth provider.

There are two possible auth tokens that we can recieve (AFAIK):
JWT that is typically recieved from UI client
Oauth2 tokens that users can generate for the CLI use.

The new auth provider understands both

## Release Notes

- **Authentication**
	- Updated authentication mechanism to use "Authorization" header instead of "Authentication".
	- Added support for AAP (Advanced Automation Platform) authentication type.
	- Simplified token validation process across authentication middleware.

- **Configuration**
	- Enhanced Helm chart to support AAP authentication configuration.
	- Added new authentication options in configuration files.
	- Introduced new `Target` configuration field and `AAP` authentication struct.

- **Deployment**
	- New Ansible playbook for registering FlightControl as an AAP Gateway service.
	- Updated README with instructions for AAP Gateway service setup.

- **API**
	- Modified OpenAPI specification to reflect authentication changes.
	- Updated client and server implementations to handle new authorization mechanism.
	- Introduced new configuration parameters for AAP authentication in relevant files.


internal/crypto/tls.go
```diff
@@ -28,7 +28,7 @@ func TLSConfigForServer(caConfig, serverConfig *TLSCertificateConfig) (*tls.Conf
 
 	tlsConfig := &tls.Config{
 		Certificates: []tls.Certificate{cert},
-		MinVersion:   tls.VersionTLS13,
+		MinVersion:   tls.VersionTLS10,
```
rawagner, line 31: AAP Gateway did not support TLS13, we will need to make this configurable

internal/config/config.go
```diff
@@ -62,6 +62,7 @@ type authConfig struct {
 	OpenShiftApiUrl       string `json:"openShiftApiUrl,omitempty"`
 	OIDCAuthority         string `json:"oidcAuthority,omitempty"`
 	InternalOIDCAuthority string `json:"internalOidcAuthority,omitempty"`
+	AAPGatewayApiUrl      string `json:"aapGatewayApiUrl,omitempty"`
```
fzdarsky, line 65: This authConfig struct is getting a bit messy, in particular that we add more and more fields for each AuthN/Z provider and do not check for incompatible or invalid config. Could we at a minimum add input validation for the auth in https://github.com/flightctl/flightctl/blob/main/internal/config/config.go#L173-L175, please? (I'll open a Jira for non-auth partS.)

Also wondering whether we can sub-structure a bit more. For example, ATM the OpenShiftApiUrl and AAPGatwayApiUrl assume that's the URL for everything: proxy, authN, and authZ. But down these may be different (e.g. using OpenShift for Auth, but SpiceDB instead of K8s RBAC), so wondering what are common vs provider-specific things that would belong into a sub-struct.

internal/auth/auth.go
```diff
@@ -91,6 +73,15 @@ func CreateAuthMiddleware(cfg *config.Config, log logrus.FieldLogger) (func(http
 			if err != nil {
 				return nil, fmt.Errorf("failed to create JWT AuthN: %w", err)
 			}
+		} else if cfg.Auth.AAPGatewayApiUrl != "" {
+			aapUrl := strings.TrimSuffix(cfg.Auth.AAPGatewayApiUrl, "/")
+			log.Println(fmt.Sprintf("AAP Gateway auth enabled: %s", aapUrl))
+			authZ = NilAuth{}
```
fzdarsky, line 79: Could we please log at start-up which authN and authZ has been configured and - in case of no authN / authZ log a warning?

internal/auth/authn/aapgateway.go
```diff
@@ -0,0 +1,104 @@
...
+			TLSClientConfig: a.ClientTlsConfig,
+		}}
+		res, err := client.Do(req)
+		if err != nil {
+			return false, err
+		}
+		return res.StatusCode == http.StatusOK, nil
+	} else {
+		_, err := jwt.Parse([]byte(token), jwt.WithKey(jwa.RS256, a.PublicKey))
+		return err == nil, err
```
fzdarsky, line 87: We probably want the middleware to store the scopes/claims in the request's ctx, so we can later do the authorization off of these.

internal/crypto/tls.go
fzdarsky: I hope AAP GW can be updated, but if not, we really should fall below `VersionTLS12`...

internal/auth/authn/aapgateway.go
```diff
@@ -0,0 +1,104 @@
...
+		client := &http.Client{Transport: &http.Transport{
+			TLSClientConfig: a.ClientTlsConfig,
+		}}
+		res, err := client.Do(req)
+		if err != nil {
+			return false, err
+		}
+		return res.StatusCode == http.StatusOK, nil
+	} else {
+		_, err := jwt.Parse([]byte(token), jwt.WithKey(jwa.RS256, a.PublicKey))
```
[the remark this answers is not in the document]
rawagner: I agree. We will want to follow their verification https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/common/auth.py#L170-L177

internal/auth/authn/aapgateway.go
```diff
@@ -0,0 +1,104 @@
...
+
+func NewAapGatewayAuth(gatewayUrl string, clientTlsConfig *tls.Config) (AapGatewayAuth, error) {
+	aapGatewayAuth := AapGatewayAuth{
+		GatewayUrl:      gatewayUrl,
+		ClientTlsConfig: clientTlsConfig,
+	}
+	client := &http.Client{Transport: &http.Transport{
+		TLSClientConfig: clientTlsConfig,
+	}}
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/v1/jwt_key/", gatewayUrl), nil)
```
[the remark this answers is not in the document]
rawagner: AFAIK single key

Im not sure about rotating keys, I will try to find out

internal/auth/authn/aapgateway.go
rawagner: service layer would just call authz.CheckPermission() and the proper AuthZ implementation will do the check. How the checking is done is abstracted from service layer.

Here is an example usage https://github.com/flightctl/flightctl/blob/main/internal/service/device.go#L50

so if there is 
- k8s AuthZ, it would check agaist the k8s API
- AAP AuthZ would match scopes/claims
- c.rh.c AuthZ would again match some scopes/claims

internal/config/config.go
rawagner: how about
```yaml
auth:
  // common props
  caCert: string
  insecureSkipTlsVerify: bool
  // provider specific props
  k8s:
    apiUrl: string
    externalOpenShiftApiUrl: string
    //additional config ie spiceDB
    spiceDB:
      foo: string
  odic:
    oidcAuthorityUrl: string
    externalOidcAuthorityUrl: string
  aap:
    gatewayUrl: string
    ...
```

internal/crypto/tls.go
```diff
@@ -28,7 +28,7 @@ func TLSConfigForServer(caConfig, serverConfig *TLSCertificateConfig) (*tls.Conf
 
 	tlsConfig := &tls.Config{
 		Certificates: []tls.Certificate{cert},
-		MinVersion:   tls.VersionTLS13,
+		MinVersion:   tls.VersionTLS12,
```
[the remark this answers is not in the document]
rawagner: FYI, the PR is approved but not yet merged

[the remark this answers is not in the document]
rawagner: Ok, removing this change as Gateway now supports TLS1.3 connections

rawagner:
Note: I've added Ansible playbook to `/deploy/aap` which registers FCTL as AAP Gateway service

internal/auth/authn/aapgateway.go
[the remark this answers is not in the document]
rawagner: `exp`, `iss` and `aud` are now being validated

internal/api_server/middleware/aap.go
```diff
@@ -0,0 +1,10 @@
+package middleware
+
+import "net/http"
+
+func AAPResponseMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Location", "__gateway_no_rewrite__=1")
+		next.ServeHTTP(w, r)
+	})
+}
```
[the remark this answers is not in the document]
rawagner: We need this middleware to inject `Location` header with specific `__gateway_no_rewrite__=1` which disables AAP's path rewrite lua script, which is rewriting urls in every HTTP response.
In our case, this is unwanted behavior.

rawagner:
Closing in favor of https://github.com/flightctl/flightctl/pull/917

PR #665, opened 2024-11-20
EDM-390: Add decommissioning route to device API

lkatalin:
A first step enabling a user to mark a device for decommissioning.


api/v1alpha1/openapi.yaml
```diff
@@ -2426,6 +2470,12 @@ paths:
                 $ref: '#/components/schemas/Error'
 components:
   schemas:
+    DeviceDesiredDecomState:
```
hexfusion, line 2473:
```suggestion
    DeviceDecommission:
      type: object
      properties:
      ...
```

What other minimal information would be useful here? What about a  `reason`? Do we need to know who it was `requestedBy` or is the auth enough? Should we consider this an object for future extensibility?

api/v1alpha1/openapi.yaml
```diff
@@ -2426,6 +2470,12 @@ paths:
...
+      type: string
+      enum:
+        - "DefaultDecom"
+        - "CustomDecom"
```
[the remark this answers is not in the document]
fzdarsky: We know we need custom actions anyway, so depending on the incremental effort (e.g. reusing hooks or not), why not.

To me the question is more the modelling: Isn't the parameter more about "how far" to decommission, i.e. to simply unenroll or to factory reset or to wipe/burn? Whether that's done by the agent or custom scripts or both is a different question and I'm not sure it belongs here.

api/v1alpha1/openapi.yaml
```diff
@@ -1038,6 +1038,50 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+  /api/v1/devices/{name}/decommission:
+    put:
+      tags:
+        - device
+      description: replace desired decommissioning state of the specified Device
+      operationId: replaceDeviceDesiredDecomState
```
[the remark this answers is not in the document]
lkatalin: >it is final / cannot be undone later

Agree with this. I'll update the operationId and description.

lkatalin:
@hexfusion @fzdarsky Updated!

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
fzdarsky: Ok, I think I'm confused about the use of this field now:
If this is a user-defined action, we cannot know whether/how that action performs reset, wipe, ...).
Also, we can define a well-known path in the device's filesystem (e.g. `/usr/flightctl/hooks.d/decommission.sh`) that - if present - gets executed at a certain point in the decom process. Means that'd replace "CustomDecom".
Therefore, I would have assumed this field to tell whether the agent shall just unenroll or go further?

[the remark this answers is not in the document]
lkatalin: I was thinking for this first iteration, we only have one setting that enables our only / default way of decommissioning: unenroll the device and send a signal to the device to do hardcoded / default actions, which means wipe the cert and key.

Later as we want to support more specific and custom options, we could add other settings: 'unenrollOnly', 'runCustomOnly', 'runDefaultAndCustom', etc., or whatever it is that we end up wanting to support. But not yet having to decide exactly what those distinctions are means we can support a set of initial default actions more quickly.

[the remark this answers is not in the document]
lkatalin: We could decide to name `default` something more descriptive that fits what we are going to support initially.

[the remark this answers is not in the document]
lkatalin: @fzdarsky I changed this to `DefaultUnenroll` to be more descriptive - thoughts?
In the future we could add `DefaultWipe`, `CustomWipe`, or whatever makes sense.

api/v1alpha1/openapi.yaml
```diff
@@ -2426,6 +2470,22 @@ paths:
...
 components:
   schemas:
+    DeviceDecommission:
+      type: object
+      properties:
+        decommissionType:
+          type: string
+          enum:
+            - 'DefaultUnenroll'
+          description: 'Specifies the desired decommissioning state of the device'
```
hexfusion, line 2480: If all we have is the default is this field even needed at this point? Have we decided what other options will exist and how they might look?

lkatalin: This is meant to sketch out where we'll add those options later, once we decide exactly what they should be.

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
fzdarsky: Naming is important in API design; names should be self-explanatory and naming something "default" isn't. And it seems we're not sure what that default is and which direction we want to take for the other values yet, so this seems merely a placeholder.

How about we first try to figure out what `decommissioningType` is? "Type" again is very unspecific. Are we thinking of
* a decommissioningTarget, e.g. Unenroll --> Unenroll+Wipe --> Unenroll+Wipe+FactoryReset --> Unenroll+Wipe+FactoryReset+Block
* a decommissioningPolicy, e.g. Unenroll plus whether or not to a) wipe data, b) run custom scripts, c) ...
* an Action? a Strategy? a Sequence?
Does it even make sense to give an operator the choice not to run a built-in or user-defined action? Arguably, if there is a user-defined action, it's been placed there deliberately by some security engineer and is mandatory.

My hunch is we're talking about a Target. And then your suggestion "UnenrollOnly" would be a valid target and maybe even a good default. It may also be the only target that's completely built-in, with others having at least one user-defined component.

api/v1alpha1/openapi.yaml
```diff
@@ -2426,6 +2470,22 @@ paths:
...
+        reason:
+          type: string
+          description: 'Reason for decommissioning device'
+        requestedBy:
```
avishayt, line 2484: What do we expect users to put here?  I would think that we'd implement an audit mechanism at some point to log each action along with the user who performed it according to their credentials.

api/v1alpha1/openapi.yaml
```diff
@@ -2426,6 +2470,22 @@ paths:
...
+          type: string
+          description: 'User or process requesting decommissioning of device'
+      required: 
+        - decommissionType
```
avishayt, line 2488: Should this be optional with a good default?

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
lkatalin: These are good questions to refine the idea, thanks.

>Does it even make sense to give an operator the choice not to run a built-in or user-defined action?

I'm not sure it does. And if each device is pre-loaded with the actions it should take upon decommissioning, then maybe there are no decisions to be made at this step...

api/v1alpha1/openapi.yaml
lkatalin: I believe this would be filled in by the system, but maybe this isn't the correct way to specify it then.

api/v1alpha1/openapi.yaml
lkatalin: I think we can leave in an open-ended `reason` field and then `requestedBy` would be filled in by the system later (which I think means we don't need it as part of this schema).

lkatalin: @fzdarsky suggests not adding anything we don't absolutely need as nothing can be removed later. So I've taken these extra properties back out, while leaving this as an extensible `object`.

api/v1alpha1/openapi.yaml
lkatalin: We decided we do need it. It is no longer required.

fzdarsky: It should be either required or defaulted. I'd vote for required, so we avoid testing for nil-pointers and so the user must explicitly choose one.

lkatalin:
@fzdarsky What do you think?

fzdarsky (approved):

PR #667, opened 2024-11-21
EDM-420: Start only one Podman Monitor per Agent

gshilin-sdb:
This is especially important for DeviceSimulator -- one monitor is enough and we don't need to kill others manually

### DIFF
--- a/internal/agent/device/applications/podman_monitor.go
+++ b/internal/agent/device/applications/podman_monitor.go
@@ -75,37 +75,23 @@ type PodmanMonitor struct {
 
 var podmanMonitor *PodmanMonitor
 var podmanMonitorMutex sync.Mutex
-var podmanMonitorSingle bool
-
-func newPodmanMonitor(log *log.PrefixLogger, exec executer.Executer, podman *client.Podman) *PodmanMonitor {
-	return &PodmanMonitor{
-		client:  podman,
-		boot:    client.NewBoot(exec),
-		compose: lifecycle.NewCompose(log, podman),
-		apps:    make(map[string]Application),
-		log:     log,
-	}
-}
-
-func UseSinglePodmanMonitor() {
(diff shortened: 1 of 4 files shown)


internal/agent/device/applications/podman_monitor.go
```diff
@@ -83,6 +87,27 @@ func NewPodmanMonitor(log *log.PrefixLogger, exec executer.Executer, podman *cli
...
+func UseSinglePodmanMonitor() {
+	podmanMonitorMutex.Lock()
+	defer podmanMonitorMutex.Unlock()
+	podmanMonitorSingle = true
+}
+
+// NewPodmanMonitor - initialize a podman monitor
+// Only one podman monitor runs per agent (this is not true for tests).
+// This is especially important in a case of device simulator and UseSinglePodmanMonitor should be called
+func NewPodmanMonitor(log *log.PrefixLogger, exec executer.Executer, podman *client.Podman) *PodmanMonitor {
```
amir-yogev-gh, line 99: For an efficient thread-safe singleton pattern, you can follow: "test, lock, test again to actually create".
The logic behind this is to protect concurrency with lock only during creation. If it's already created, you can test for != nil return the existing instance without the lock.
That might be considered as overkill depending on the usage.

Adjusting the body accordingly may look like:

	if podmanMonitorSingle {
		if podmanMonitor == nil {
			podmanMonitorMutex.Lock()
			defer podmanMonitorMutex.Unlock()
			if podmanMonitor == nil {
				podmanMonitor = newPodmanMonitor(log, exec, podman)
			}
		}
		return podmanMonitor
	}
	return newPodmanMonitor(log, exec, podman)

gshilin-sdb: This is "kind of" singleton. It should be, but the way it's used in tests we need a way to reset it. So changed it accordingly.

PR #668, opened 2024-11-22
EDM-417: Compute service-side statuses

fzdarsky:
This PR implement the service-side logic for updating the device summary status, device update status, and application summary status based on the information provided by the agent and information only available service-side.

Depends on PR #675 (which centralizes API consts to break import cycles).

User documentation is in PR #637.

### DIFF
--- a/internal/service/common/device.go
+++ b/internal/service/common/device.go
@@ -2,18 +2,36 @@ package common
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"strings"
 	"time"
 
+	"github.com/dustin/go-humanize"
+	api "github.com/flightctl/flightctl/api/v1alpha1"
 	"github.com/flightctl/flightctl/internal/api/server"
 	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/store"
+	"github.com/flightctl/flightctl/internal/util"
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
 )
 
-func ReplaceDeviceStatus(ctx context.Context, st store.Store, request server.ReplaceDeviceStatusRequestObject) (server.ReplaceDeviceStatusResponseObject, error) {
+const (
+	ApplicationStatusInfoHealthy = "All application workloads healthy."
+	DeviceStatusInfoHealthy      = "All system resources healthy."
+	DeviceStatusInfoRebooting    = "The device is rebooting."
+)
+
+func ReplaceDeviceStatus(ctx context.Context, st store.Store, log logrus.FieldLogger, request server.ReplaceDeviceStatusRequestObject) (server.ReplaceDeviceStatusResponseObject, error) {
 	orgId := store.NullOrgId
 
 	device := request.Body
+	if errs := validateDeviceStatus(device); len(errs) > 0 {
+		return server.ReplaceDeviceStatus400JSONResponse{Message: errors.Join(errs...).Error()}, nil
+	}
 	device.Status.LastSeen = time.Now()
+	UpdateServiceSideStatus(ctx, st, log, orgId, device)
 
 	result, err := st.Device().UpdateStatus(ctx, orgId, device)
 	switch err {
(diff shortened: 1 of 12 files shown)


internal/tasks/device_disconnected.go
```diff
@@ -37,30 +35,30 @@ func (t *DeviceDisconnected) Poll() {
...
 
-	statusInfoMessage := fmt.Sprintf("Did not check in for more than %d minutes", int(DeviceDisconnectedTimeout.Minutes()))
+	statusInfoMessage := fmt.Sprintf("Did not check in for more than %d minutes", int(api.DeviceDisconnectedTimeout.Minutes()))
 	// TODO: one thread per org?
 	orgID := uuid.UUID{}
 	// batch of 1000 devices
 	listParams := store.ListParams{Limit: ItemsPerPage}
 	for {
-		devices, err := t.deviceStore.List(ctx, orgID, listParams)
+		devices, err := t.store.Device().List(ctx, orgID, listParams)
```
hexfusion, line 44: :+1:

internal/service/common/device.go
```diff
@@ -2,18 +2,29 @@ package common
...
+	"github.com/sirupsen/logrus"
 )
 
-func ReplaceDeviceStatus(ctx context.Context, st store.Store, request server.ReplaceDeviceStatusRequestObject) (server.ReplaceDeviceStatusResponseObject, error) {
+func ReplaceDeviceStatus(ctx context.Context, st store.Store, log logrus.FieldLogger, request server.ReplaceDeviceStatusRequestObject) (server.ReplaceDeviceStatusResponseObject, error) {
 	orgId := store.NullOrgId
 
 	device := request.Body
+	if errs := validateDeviceStatus(device); len(errs) > 0 {
+		return server.ReplaceDeviceStatus400JSONResponse{Message: errors.Join(errs...).Error()}, nil
```
hexfusion, line 24: This bit me the other day but we need to be careful on this usage because if someone by mistake returns nil for []error it will panic here. Not sure what the action item is here but we should consider this issue with this pattern. I guess we could nill check err but not as clean.


```
	var errs []error
	errs = nil
	err := errors.Join(errs...).Error()
	fmt.Println(err)
```

hexfusion (approved):
Overall looks great left a few non blocking comments.

internal/service/common/device.go
```diff
@@ -2,18 +2,29 @@ package common
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"strings"
 	"time"
 
+	api "github.com/flightctl/flightctl/api/v1alpha1"
```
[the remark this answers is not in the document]
fzdarsky: Yeah, I also noticed that. I personally find it cleaner and easier on the eyes to use `api` and we can then update to a later version in a single line.

internal/service/common/device.go
fzdarsky: If `errs == nil`, then `len(errs) == 0`, so we should catch that case, or?

hexfusion: haha my bad tunnel vision I missed the upper check.

internal/service/common/device.go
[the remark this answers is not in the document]
hexfusion: That works until we have multiple versions of apis in the future.

celdrake (commented):
@fzdarsky Some small suggestions regarding the status messages.

**Note** I was not able to see most of the messages by actually setting the device configs. For example, I tried defining an application to later disconnect the device and check the message, but the agent was failing at pulling the image (despite it being public).

internal/service/common/device.go
```diff
@@ -30,7 +41,118 @@ func ReplaceDeviceStatus(ctx context.Context, st store.Store, request server.Rep
...
+		device.Status.Summary.Status = api.DeviceSummaryStatusRebooting
+		device.Status.Summary.Info = util.StrToPtr("Device is rebooting.")
+		return device.Status.Summary.Status != lastDeviceStatus
+	}
+
+	resourceErrors := []string{}
+	resourceDegradations := []string{}
+	switch device.Status.Resources.Cpu {
+	case api.DeviceResourceStatusCritical:
+		resourceErrors = append(resourceErrors, "CPU utilisation reached critical level") // TODO: add current threshold (>X% for more than Y minutes)
```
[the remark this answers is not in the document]
fzdarsky: Good catch, thank you!

avishayt (approved):

avishayt (approved):

PR #669, opened 2024-11-24
EDM-420: Increase DB parameters

gshilin-sdb:
Usage: make deploy DB_SIZE=medium-10k|small-1k|e2e(default)  STATUS_UPDATE_INTERVAL=2m0s SPEC_FETCH_INTERVAL=2m0s

e2e: minimal footprint for e2e testing.
small-1k: recommended setting for a demo environment 1000 device max
medium-10k: 10k max


deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
```diff
@@ -40,10 +42,13 @@ spec:
           volumeMounts:
             - mountPath: /var/lib/pgsql/data
               name: flightctl-db
-{{ if .Values.db.resources }}
```
hexfusion, line 43: why not define resources via helm values?

deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
               value: {{ .Values.db.userPassword }}
             - name: POSTGRESQL_USER
               value: {{ .Values.db.user }}
+            - name: POSTGRESQL_MAX_CONNECTIONS
+              value: "200"
```
hexfusion, line 35: should this be configurable?

deploy/helm/flightctl/templates/flightctl-db-persistentvolumeclaim.yaml
```diff
@@ -15,5 +15,5 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 2Gi
+      storage: 20Gi
```
hexfusion, line 18: should this be configurable?

hexfusion:
cc @mangelajo

deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
mangelajo: Yes, this needs to continue to be configurable from the values. Some environments are restrictive in terms of limits or requests and won’t accept just anything

mangelajo: Please leave it back and control the amount you need by setting it in values.db.resources

mangelajo (approved):

mangelajo (changes requested):
Look at the comments

gshilin-sdb:
Usage: make deploy DB_VERSION_PARAMS=demo-large|demo-small|e2e(default)

gshilin-sdb:
> Look at the comments

Done

docs/developer/README.md
```diff
@@ -61,9 +61,17 @@ Use an agent VM to test a device interaction, an image is automatically created
 hack/Containerfile.local and a qcow2 image is derived in output/qcow2/disk.qcow2, currently
 this only works on a Linux host.
 
+You can deploy a DB container of different sizes using a DB_VERSION variable for make command:
+* e2e (default) - minimal footprint for e2e testing
+* small - recommended setting for a demo environment 1000 devices max
+* demo-10k - recommended setting for a demo environment 10k devices max
```
mangelajo, line 67:
```suggestion
* demo-10k - recommended setting for a demo environment 10k devices max
```

Instead of demo-10k, I would call it medium-10k ?

Kind environment should not be used for demos, it's only ok for development, if we get customers understanding that they can demo on kind it has risks, I.e. the contents are volatile, when you bring down kind , all data is lost etc.

Can we change to:

small-1k
medium-10k

?
Then we always are explicit about the expected scale and not mention demos.

test/scripts/deploy_with_helm.sh
```diff
@@ -12,13 +13,30 @@ source "${SCRIPT_DIR}"/functions
...
     -a|--only-db) ONLY_DB="--set flightctl.api.enabled=false --set flightctl.worker.enabled=false --set flightctl.periodic.enabled=false --set flightctl.rabbitmq.enabled=false --set flightctl.kv.enabled=false" ; shift ;;
-    -h|--help) echo "Usage: $0 [--only-db]"; exit 0 ;;
+    -h|--help) echo "Usage: $0 $usage"; exit 0 ;;
+    --db-size)
+      db_size=$2
+      if [ "$db_size" == "e2e" ]; then
+        DB_SIZE_PARAMS=""
+      elif [ "$db_size" == "small" ]; then
+        DB_SIZE_PARAMS="--set db.resources.requests.cpu=1 --set db.resources.requests.memory=1Gi --set db.resources.limits.cpu=8 --set db.resources.limits.memory=64Gi"
+      elif [ "$db_size" == "prod" ]; then
```
mangelajo, line 31: there is no "10k" version here, and no mention of "prod" on the other side.
let's use.

small-1k
medium-10k

mangelajo (changes requested):
Thank you Gregory

deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
mangelajo: @gshilin-sdb can we make this a parameter of Values.db.max_connections and default that to 200 ?

deploy/helm/flightctl/templates/flightctl-db-persistentvolumeclaim.yaml
mangelajo: Can we make the same Values.db.storage.size: 20Gi ?

mangelajo (changes requested):
Just the couple of changes that sam requested too.

mangelajo (approved):

mangelajo:
> It seems like we have a few possible use cases for these helm charts .
> 
> e2e: minimal footprint for e2e testing. demo-small: recommended setting for a demo environment 1000 device max demo-large: 10k max
> 
> these are just examples but it should be easily configurable based on your env. does helm provide a way to set/consume diffent default values in this way?

we could provide values.yaml layers, but it's not easy to access on real installs...

hexfusion (approved):

mangelajo (approved):

PR #675, opened 2024-11-27
EDM-417: Centralize API consts

fzdarsky:
This PR moves several constants defined across internal packages into the API package to as a prep work for PR #668. This is mainly to break import cycles, but arguably these consts are anyway part of the API definition.


api/v1alpha1/consts.go
```diff
@@ -0,0 +1,77 @@
+package v1alpha1
+
+import "time"
+
+const (
+	CertificateSigningRequestAPI      = "v1alpha1"
```
mangelajo, line 6:
```suggestion
	CertificateSigningRequestAPIVersion      = "v1alpha1"
```
?

api/v1alpha1/consts.go
```diff
@@ -0,0 +1,77 @@
...
+	CertificateSigningRequestKind     = "CertificateSigningRequest"
+	CertificateSigningRequestListKind = "CertificateSigningRequestList"
+
+	DeviceAPI      = "v1alpha1"
```
mangelajo, line 10:
```suggestion
	DeviceAPIVersion      = "v1alpha1"
```
?

api/v1alpha1/consts.go
```diff
@@ -0,0 +1,77 @@
...
+
+	DeviceAnnotationConsole         = "device-controller/console"
+	DeviceAnnotationRenderedVersion = "device-controller/renderedVersion"
+	DeviceAnnotationTemplateVersion = "fleet-controller/templateVersion"
+
+	// TODO: make configurable
+	// DeviceDisconnectedTimeout is the duration after which a device is considered to be not reporting and set to unknown status.
+	DeviceDisconnectedTimeout = 5 * time.Minute
+
+	EnrollmentRequestAPI      = "v1alpha1"
```
mangelajo, line 22:
```suggestion
	EnrollmentRequestAPIVersion      = "v1alpha1"
```

api/v1alpha1/consts.go
```diff
@@ -0,0 +1,77 @@
...
+	EnrollmentRequestKind     = "EnrollmentRequest"
+	EnrollmentRequestListKind = "EnrollmentRequestList"
+
+	FleetAPI      = "v1alpha1"
+	FleetKind     = "Fleet"
+	FleetListKind = "FleetList"
+
+	FleetAnnotationTemplateVersion = "fleet-controller/templateVersion"
+
+	RepositoryAPI      = "v1alpha1"
```
mangelajo, line 32:
```suggestion
	RepositoryAPIVersion      = "v1alpha1"
```

api/v1alpha1/consts.go
```diff
@@ -0,0 +1,77 @@
...
+	RepositoryKind     = "Repository"
+	RepositoryListKind = "RepositoryList"
+
+	ResourceSyncAPI      = "v1alpha1"
```
mangelajo, line 36:
```suggestion
	ResourceSyncAPIVersion      = "v1alpha1"
```

api/v1alpha1/consts.go
```diff
@@ -0,0 +1,77 @@
...
+	ResourceSyncKind     = "ResourceSync"
+	ResourceSyncListKind = "ResourceSyncList"
+
+	TemplateVersionAPI      = "v1alpha1"
```
mangelajo, line 40:
```suggestion
	TemplateVersionApiVersion      = "v1alpha1"
```

api/v1alpha1/consts.go
mangelajo: I wonder, since all those API Versions ar under api/v1alpha/consts.go 

we could just have one const APIVersion = "v1alpha1"

mangelajo (changes requested):

api/v1alpha1/consts.go
fzdarsky: That depends on whether we'd like the individual resource kinds to evolve/mature somewhat independently. IOW, can we imagine that, say, the Device resource becomes stable (v1), while the EnrollmentRequests becomes v1alpha2?

Then, is there anything to version about our APIs _on top of_ the sum of the individual resource kinds? I believe there is, e.g. common patterns on the verbs, query params, etc. If we want to have a change later to evolve that, we also need this part of the API to be versioned.

@avishayt ^?

fzdarsky: Could we separate the API versioning from this PR, which is just a refactoring to unblock other work?

avishayt: This PR doesn't change anything.  Let's merge and discuss options.

avishayt:
Lint!!! :)

mangelajo (approved):

PR #678, opened 2024-11-28
GRPC API 

mangelajo:
Introduce the gRPC API for the agent.

A second step should move the agent implementation to this new API, a few key concepts of this API are:

* json is used as transport for the open-api bound objects. 
* the version of the requested/posted json object is explicitly set as part of the call, allowing the server to down-convert data to a format that the agent will understand, or to update received data to newer formats.


api/grpc/v1/agent.proto
```diff
@@ -0,0 +1,67 @@
+syntax = "proto3";
+
+package flightctl.v1;
```
fzdarsky, line 3: Can this be more mature than our user-facing API which is v1alpha1?

api/grpc/v1/agent.proto
```diff
@@ -0,0 +1,67 @@
...
+
+option go_package = "github.com/flightctl/flightctl/api/grpc/v1/grpc-v1";
```
fzdarsky, line 5: Should this be structured by purpose (`/api/agent`) rather than technology (`/api/grpc`)?
Along the same lines: Should the user API also move to a subdir?
I'm thinking we may need to collect more APIs here at some point, e.g. the Config APIs for the different services.

api/grpc/v1/agent.proto
mangelajo: Yes, in this case it's the format of this API.

When requesting objects related to the user-facing API, that would be ser/deserialized in json, and we would explicitly provide an api_version parameter from the agent.

PR #679, opened 2024-11-28
EDM-788: Add init script for PostgreSQL to set master user as superuser

asafbss:
- Added `enable-extensions.sh` script to initialize PostgreSQL with specified extensions.
- The script checks the `POSTGRESQL_EXTENSIONS` environment variable and enables each extension dynamically.
- Integrated the script into the `flightctl-db` ConfigMap and Deployment.

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-db-config.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-config.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: flightctl-db
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+data:
+  enable-superuser.sh: |
+    #!/bin/bash
+
+    _psql () { psql --set ON_ERROR_STOP=1 "$@" ; }
+
+    # Ensure POSTGRESQL_MASTER_USER is treated as a superuser
+    if [ -n "${POSTGRESQL_MASTER_USER}" ]; then
+      echo "Granting superuser privileges to ${POSTGRESQL_MASTER_USER}"
+      _psql -c "ALTER ROLE ${POSTGRESQL_MASTER_USER} WITH SUPERUSER;"
+    fi
--- a/deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
@@ -42,6 +42,9 @@ spec:
           volumeMounts:
             - mountPath: /var/lib/pgsql/data
               name: flightctl-db
+            - mountPath: /usr/share/container-scripts/postgresql/start/enable-superuser.sh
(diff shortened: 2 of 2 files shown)


asafbss:
@siserafin, Could you please test this thoroughly?

rawagner:
Can we now remove https://github.com/flightctl/flightctl/blob/main/test/scripts/deploy_with_helm.sh#L87 ?

siserafin:
> @siserafin, Could you please test this thoroughly?

I can test this in local, since I need a flightctl helm chart to install flightctl in ocp. Or do you know another way?

fzdarsky:
Could you please also add this change to the Quadlet deployment (https://github.com/flightctl/flightctl/blob/main/deploy/quadlets/flightctl-db/flightctl-db.container), so they don't get out-of-sync?

siserafin:
> > @siserafin, Could you please test this thoroughly?
> 
> I can test this in local, since I need a flightctl helm chart to install flightctl in ocp. Or do you know another way?

Locally I don't see the issue, I will test  in ocp when the PR is merged, and I have the updated helm chart

rawagner:
> > > @siserafin, Could you please test this thoroughly?
> > 
> > 
> > I can test this in local, since I need a flightctl helm chart to install flightctl in ocp. Or do you know another way?
> 
> Locally I don't see the issue, I will test in ocp when the PR is merged, and I have the updated helm chart

you can test before merging using
```
helm upgrade --install --namespace <ns> --create-namespace flightctl ./deploy/helm/flightctl --values <path_to_values_file>
```

siserafin:
Tested in ocp, and after the installation, the api pod is running: 
```
[kni@titan54 flightctl]$ oc get pods -n flightctl
NAME                                  READY   STATUS    RESTARTS        AGE

flightctl-api-65f5889b79-hh8cn        1/1     Running   2 (4m54s ago)   5m8s
```


LGTM

fzdarsky:
Please keep in mind that the Postgres database may not be the one we deploy by default, but an external one provided by the end-user. IOW, we cannot assume control over that external database.

fzdarsky:
> I agree with your point. Do you have any suggestions on how we can best accomplish this?
> 
> We are currently creating tables and indexes, as well as enabling database extensions. This requires the end-user to either provide a database user for the API server with the necessary permissions or enable the extensions themselves.

As a first step, I believe we should separate the PG admin and user secrets like in the Keycloak sub-chart (and not pass via env). Then we could use an init-container to do the required operations before starting the API server least privileged.

avishayt:
@fzdarsky I agree with you, but that seems to be outside the scope of this fix?

fzdarsky:
> @fzdarsky I agree with you, but that seems to be outside the scope of this fix?

Yeah, not saying this PR needs to do everything. Ok, let’s proceed and open a Jira for follow-up work.

avishayt (approved):

PR #684, opened 2024-12-03
EDM-659: agent/fileio: ensure ignition gid, uid and mode

hexfusion:
This PR fixes a bug where managed files did not consider mode or user and group changes. The result was an ignition file that maintains the same content but different perms would not update on disk

### DIFF
--- a/internal/agent/device/fileio/managed_file.go
+++ b/internal/agent/device/fileio/managed_file.go
@@ -2,20 +2,22 @@ package fileio
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"os"
+	"syscall"
 
 	"github.com/ccoveille/go-safecast"
 	ign3types "github.com/coreos/ignition/v2/config/v3_4/types"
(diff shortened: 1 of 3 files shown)


internal/agent/device/fileio/managed_file.go
```diff
@@ -58,18 +60,63 @@ func (m *managedFile) decodeFile() error {
...
+	}
+
+	gid, err := safecast.ToUint32(m.gid)
+	if err != nil {
+		return false, err
+	}
+
+	// compare file ownership
+	if stat.Uid != uid || stat.Gid != gid {
+		fmt.Printf("uid: %d, gid: %d\n", stat.Uid, stat.Gid)
```
avishayt, line 110: Remove?

avishayt (commented):
LGTM, just looks like you have a leftover debug print

internal/agent/device/fileio/managed_file.go
hexfusion: updated, thanks!

avishayt (approved):

PR #688, opened 2024-12-04
EDM-838: Update ACM registration doc

fzdarsky:
This PR updates the "Auto-Registering Devices with MicroShift into ACM" section of the documentation to align it with the updated Device Lifecycle Hooks API and remove a number of implementation details that go beyond what would be expected from a user documentation.

The PR also removes a stray section from the Hooks API description.

### DIFF
--- a/docs/user/registering-microshift-devices-acm.md
+++ b/docs/user/registering-microshift-devices-acm.md
@@ -1,62 +1,63 @@
 # Auto-Registering Devices with MicroShift into ACM
 
-## Pre-requisite
+If you have fleets of devices running an OS image that includes MicroShift, you can configure these fleets to auto-register MicroShift clusters with Red Hat Advanced Cluster Management (ACM).
 
-The Operating System image running on our fleet of devices must contain MicroShift. This is essential for running the Klusterlet, the agent responsible for communication with Red Hat Advanced Cluster Management (ACM)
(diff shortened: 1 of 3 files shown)


avishayt (approved):

docs/user/registering-microshift-devices-acm.md
```diff
@@ -70,94 +71,29 @@ spec:
...
+              run: kubectl apply -f /var/local/acm-import/import.yaml
+              envVars:
+                KUBECONFIG: /var/lib/microshift/resources/kubeadmin/kubeconfig
 ```
 
-Flight Control hooks are explained in the following [link](https://github.com/flightctl/flightctl/blob/main/docs/user/managing-devices.md#using-device-lifecycle-hooks).
-Hooks are triggered at specific moments on the lifecycle of a device such as before/after updating, before/after reboot, etc.
-The hooks definition shown above describe two hooks executed after updating the fleet, one to apply the CRD file and one to apply the rest of the Klusterlet manifests stored in the import.yaml file.
-
-## Device Reconciliation Process
```
hexfusion, line 153: why drop this? isnt this important information?

fzdarsky: Because it's way too detailed on implementation for a user documentation and describes either ACM processes documented in the referenced ACM doc or FC processes already described above.

PR #690, opened 2024-12-05
EDM-848: revert tzdata host mount

hexfusion:
This PR reverts the addition of a host mount on tzdata. tzdata is used by the API server to validate time zones. But this approach breaks in OCP as the pod does not have proper permissions.


hexfusion:
cc @rawagner

rawagner (approved):

PR #692, opened 2024-12-05
EDM-649: Update CSR error codes

lkatalin:
~This is built on #616 - the only new commit in this PR is [ad51a197548aaf1c36beea02532dd28f56c9bd42](https://github.com/flightctl/flightctl/commit/ad51a197548aaf1c36beea02532dd28f56c9bd42).~

~Will probably remove the error conversion (commented out) because it doesn't preserve the original error message. Instead of this we can choose an error code to represent the range of errors possible from `Approve`: 400, 401, 404, 409, 500. Thoughts?~

~The new 500 error on POST may or may not be necessary depending on the decision on above.~

The above is all outdated.


fzdarsky:
> Will probably remove the error conversion (commented out) because it doesn't preserve the original error message. Instead of this we can choose an error code to represent the range of errors possible from `Approve`: 400, 401, 404, 409, 500. Thoughts?
> 
> The new 500 error on POST may or may not be necessary depending on the decision on above.

A "500 - Internal Server error" means we've got a programming error, we should never deliberately return that and when we return it when catching an exception we must not leak implementation details by returning Golang `error` from deep down the service.

Let's think through the error cases:
When a user creates a CSR (POST to `/api/v1/certificatesigningrequests`), the only return codes can be
* 200 - OK
* 400 - Bad Request: only when the request did not pass Validation().
* 401 - Unauthorized
When a user updates a CSR (PUT/PATCH to `/api/v1/certificatesigningrequests/{name}`), the only return codes can be
* 200 - OK
* 201 - Created
* 400 - Bad Request: only when the request did not pass Validation(), including trying to modify immutable fields.
* 401 - Unauthorized.

Everything to do with approval and signing normally happens asynchronously. That we currently do it synchronously as a hack should not change the API. Result is that any errors from approval or signing cannot be returned via HTTP response codes. Instead, we need to surface them through a `.status.condition` of type "Failed".

lkatalin:
Thanks @fzdarsky .

>Result is that any errors from approval or signing cannot be returned via HTTP response codes. Instead, we need to surface them through a .status.condition of type "Failed"

Yes, exactly why I was struggling to represent this with the response code. I had not considered the `Failed` `.status.condition`, that's a great alternative so I will do that.

ACK re: only using `500` as a catch-all and not returning the error in those cases.

Btw, this isn't as relevant since we can use the `.status.condition`, but we do have more response codes available currently for CSRs than those you listed (201, 202, 208 for POST and 404 and 409 for PUT/PATCH). I believe this was the result of a union between the codes K8s uses and what we were using for enrollment requests.

lkatalin:
Actually it looks like we can just return `nil, err` if the CSR cannot be approved. But in the case of a newly created CSR that cannot be auto-approved, I've also added a `Failed` Condition.

lkatalin:
@fzdarsky This could use a review as well, it is one in a series of bug fix PRs with others to be built on top of it.

internal/service/certificatesigningrequest.go
```diff
@@ -310,8 +310,19 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, r
...
+				// do not update status on pre-existing enrollment CSR with same name
+				if created {
+					c := []api.Condition{{
+						Message: "CSR could not be auto-approved",
+						Reason:  "CSR could not be auto-approved",
+						Type:    "Failed"}}
+					err := h.store.CertificateSigningRequest().UpdateConditions(ctx, orgId, request.Name, c)
+					if err != nil {
+						return nil, fmt.Errorf("enrollment CSR for %s could not be auto-approved: %s\nstatus condition could not be set to 'failed': %s", request.Name, approveResp, err)
+					}
```
hexfusion, line 323: Did you mean to return here in a non-error case?` `If created` is always an error case, but maybe that is expected.

lkatalin: That is actually expected, but I'm now noticing another problem - hold on, I will update.

lkatalin: Okay, I simplified this a bit because we actually don't need to try auto-approving at all unless it was `created`.

api/v1alpha1/openapi.yaml
```diff
@@ -2022,7 +2022,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/EnrollmentRequest'
+                $ref: '#/components/schemas/Error'
```
fzdarsky, line 2025: POSTs (Create) of resources should only have 201, 400, 401, 403, 409, 503 responses (see the linting rules in [PR #714](https://github.com/flightctl/flightctl/pull/714)).

The correct code here would be 409, as the resource being created already exists. We could also remove the 200, 202, and 208 while we're at it.

fzdarsky (commented):
This is a good step forward. While running the linter, I noticed the approval needs to be PUT instead of POST and we need to merge the "Deny" into this (it's currently a DELETE...). But we can do that as a follow-up.

internal/service/certificatesigningrequest.go
```diff
@@ -351,18 +359,18 @@ func (h *ServiceHandler) ApproveCertificateSigningRequest(ctx context.Context, r
 		if errors.Is(err, flterrors.ErrResourceNotFound) {
 			return server.ApproveCertificateSigningRequest404JSONResponse{}, nil
 		}
-		return server.ApproveCertificateSigningRequest500JSONResponse{Message: err.Error()}, nil
+		return nil, err
 	}
 	if csr == nil {
-		return server.ApproveCertificateSigningRequest500JSONResponse{Message: "CertificateSigningRequest is nil"}, nil
+		return server.ApproveCertificateSigningRequest400JSONResponse{Message: "CertificateSigningRequest is nil"}, nil
```
fzdarsky, line 365: I don't think `csr` can ever be nil if `err == nil` (see the `Get()` function).

fzdarsky (approved):

PR #693, opened 2024-12-05
EDM-556: agent: fix sync distribution mismatch

hexfusion:
This PR resolves a long-standing bug where the distribution of spec and status syncs can become misaligned for a period of time. Although the current implementation does provide 50/50 distribution over time in the case where the durations are the same.  As with a coin toss you can get 5 or more heads in a row. Depending on the configuration this can result in long waits for status updates and or new spec reconciliation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,6 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.1.0
 	github.com/lestrrat-go/jwx/v2 v2.1.0
 	github.com/lib/pq v1.10.9
-	github.com/lthibault/jitterbug v2.0.0+incompatible
 	github.com/mackerelio/go-osstat v0.2.5
 	github.com/oapi-codegen/nethttp-middleware v1.0.1
 	github.com/oapi-codegen/runtime v1.1.1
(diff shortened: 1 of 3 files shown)


fzdarsky:
I’m probably triggering on the term “scheduler, (implying selecting from a pool of resources which resource to assign a task to and we only have one), but am immediately thinking “complexity”.

Can we take a step back and consider whether specFetch and statusUpdate intervals still make sense the way they are currently defined?

We have 2 kinds of status updates: mandatory ones that must be sent no matter what (assuming network connectivity, of course) and optional ones that may be rate-limited, i.e. can be skipped if the last update was less than X seconds ago. For state transitions during updates, for resource or app alarms, etc. we need to decide case by case which one to use.

We also know the calls for mandatory status updates need to be synchronous: It’d be incorrect for the caller to continue before the status update is received by the service.

This means the “scheduler” (can we simply call it “timer”?) just needs to ensure a _minimum_ rate of sending additional, mandatory updates. The rate is at least `1/(disconnectTimeout - maxJitter - safety margin)`. A simple timer that when firing sends a status update if `timeSinceLastUpdate' > 1/rate` ago and resets the timer to fire again for in `1/rate - timeSinceLastUpdate` would be all we need.

Heartbeats? Not needed, as the above performs exactly that function, especially if we send the delta since the last status update.
Spec fetch interval? I fail to find a scenario where we’d fetch less often than we send status and if we want agents to _more_ frequently poll for updates (because push from service isn’t available and the user wants faster reactions), why would we not send status at the same time?

hexfusion:
> I’m probably triggering on the term “scheduler" 

I agree the term may not be the best as a scheduler may draw an assumption that the tasks are async when in fact all tasks in the proposed change are synchronous.

> Heartbeats? Not needed

This is a possible design change but today when we demo the agent it can have unexpected patterns. short term I feel that should be resolved first then design/API changes can be prioritized/considered.

> Spec fetch interval? I fail to find a scenario where we’d fetch less often than we send status.

The bug itself and a lot of complexity stems from this existing API which allows two separate intervals. Changing that is a different conversation my assumption (poor) was that this was a requirement.

> If the user wants the agent to poll for spec updates more often and assuming push updates are not available, why would we not send status or use this as "heartbeat" at the same time?

today status updates (db write) are expensive so it makes sense that we would want to batch writes at a slower interval than we may want to poll for change (read). To your other point, I agree and I brought it up before that etcd for example, it uses any RPC as a heartbeat or proof of life. So requiring a status update may not be required.

avishayt (approved):

mangelajo (approved):

hexfusion:
I will clean this up in the next iteration but this helps the problem for now,

hexfusion:
This work has been continued via #761

PR #695, opened 2024-12-06
EDM-390: Soft delete device record on decommission

lkatalin:
~The first of several steps that need to happen server-side during device decommissioning.~
This will take place when a device has decommissioned, or timed out of decommissioning.

### DIFF
--- a/internal/service/device.go
+++ b/internal/service/device.go
@@ -339,5 +339,19 @@ func (h *ServiceHandler) PatchDevice(ctx context.Context, request server.PatchDe
 
 // (PUT /api/v1/devices/{name}/decommission)
 func (h *ServiceHandler) DecommissionDevice(ctx context.Context, request server.DecommissionDeviceRequestObject) (server.DecommissionDeviceResponseObject, error) {
-	return nil, fmt.Errorf("not yet implemented")
+	orgId := store.NullOrgId
+
+	// soft delete the device in the DB
+	err := h.store.Device().SoftDelete(ctx, orgId, request.Name)
+
+	switch err {
+	case nil:
+		return server.DecommissionDevice200JSONResponse{}, nil
+	case flterrors.ErrResourceIsNil, flterrors.ErrResourceNameIsNil, flterrors.ErrIllegalResourceVersionFormat:
+		return server.DecommissionDevice400JSONResponse{Message: err.Error()}, nil
+	case flterrors.ErrResourceNotFound:
+		return server.DecommissionDevice404JSONResponse{}, nil
+	default:
+		return nil, err
+	}
 }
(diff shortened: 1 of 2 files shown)


internal/store/device.go
```diff
@@ -440,6 +441,17 @@ func (s *DeviceStore) Delete(ctx context.Context, orgId uuid.UUID, name string,
 	return nil
 }
 
+func (s *DeviceStore) SoftDelete(ctx context.Context, orgId uuid.UUID, name string) error {
```
avishayt, line 444: You're missing some logic that exists in Delete above.  I suggest changing that method instead of creating a new one by removing "Unscoped()" from innerTx.Unscoped().Delete(&existingRecord).

fzdarsky: +1. The current logic _should_ (let's verify!) already be a [soft-delete](https://gorm.io/docs/delete.html#Soft-Delete), no need to add something new here.

internal/service/device.go
```diff
@@ -339,5 +339,19 @@ func (h *ServiceHandler) PatchDevice(ctx context.Context, request server.PatchDe
 
 // (PUT /api/v1/devices/{name}/decommission)
 func (h *ServiceHandler) DecommissionDevice(ctx context.Context, request server.DecommissionDeviceRequestObject) (server.DecommissionDeviceResponseObject, error) {
-	return nil, fmt.Errorf("not yet implemented")
+	orgId := store.NullOrgId
+
+	// soft delete the device in the DB
+	err := h.store.Device().SoftDelete(ctx, orgId, request.Name)
```
fzdarsky, line 345: Why delete the device here?

internal/store/device.go
avishayt: @fzdarsky None of our DB delete methods use soft delete today.  Do we want to change them all, or only devices (and ERs)?

internal/service/device.go
lkatalin: Based on today's conversation, we won't soft delete until the device signals that it is done or until the decommissioning process times out, so it won't happen here after all.

internal/store/device.go
lkatalin: I'm seeing that `store/device.go` the device uses [`Unscoped().Delete()`](https://github.com/flightctl/flightctl/blob/main/internal/store/device.go#L421) which is a hard delete. However I'm fine to change that method instead of creating this separate one.

lkatalin:
Thanks for taking a look @avishayt @fzdarsky .
Based on today's conversation we won't soft delete the device until its decommissioning process times out or until the device signals back that it has decommissioned. So, I can change this PR to just make a soft-delete available for devices and use that functionality in a later PR.

lkatalin:
Will we ever want a hard delete available for devices too? It seems like we might.

avishayt:
I don't think we'll ever want to soft-delete.  It causes bloat due to old records sticking around (slower indices).  Unless we ever want to support an "undo delete" action, it's not necessary.  The thought was to use soft-deleted records as an audit log, but I don't believe this meets customer requirements for an audit log.

lkatalin:
>I don't believe this meets customer requirements for an audit log
What would you recommend as an alternative for audit?

avishayt:
@lkatalin Audit would be a separate component, likely owned by the customer.  We'll need to gather requirements for that.  In the meantime, let's stick to regular deletion (not soft) because I'm worried about the side effects of soft deletion more than I'm worried about auditing at this point.

PR #700, opened 2024-12-08
EDM-854: Use go-templates for fleet template parameters

avishayt:
1. Use go-templates instead of {{ device.metadata.name }} and {{ device.metadata.labels[key] }}.
2. Add go-template helper functions: toUpper, toLower, replace, getOrDefault, getOrError.
3. Support parameters in OS images (EDM-856)


api/v1alpha1/validation.go
```diff
@@ -681,3 +754,39 @@ func validateGraceDuration(schedule cron.Schedule, duration string) error {
...
 	return nil
 }
+
+// To ensure valid go-template syntax, we create a fake device that contains only ObjectMeta
+// (what we allow to include in parameters), then evaluate.
+type whitelistDevice struct {
+	Metadata ObjectMeta
+}
+
+func ValidateParametersInString(s *string, path string) (bool, []error) {
```
fzdarsky, line 764: Should we - and is there a way to - exclude some of go-template's actions like "if" ,"range", and "with" to protect the service?

api/v1alpha1/validation.go
```diff
@@ -681,3 +754,39 @@ func validateGraceDuration(schedule cron.Schedule, duration string) error {
...
+	if s == nil {
+		return false, []error{}
+	}
+
+	allErrs := []error{}
+
+	t, err := template.New("t").Parse(*s)
```
fzdarsky, line 771: How expensive is validating with templates vs. using pre-compiled regexes here?

api/v1alpha1/validation.go
```diff
@@ -681,3 +754,39 @@ func validateGraceDuration(schedule cron.Schedule, duration string) error {
 
 	return nil
 }
+
+// To ensure valid go-template syntax, we create a fake device that contains only ObjectMeta
+// (what we allow to include in parameters), then evaluate.
+type whitelistDevice struct {
```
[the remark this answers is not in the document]
avishayt: Changed to allowedPropertiesDevice

api/v1alpha1/validation.go
avishayt: Do you mean to limit computational resources used?  I haven't seen a straight-forward way to do so.

api/v1alpha1/validation.go
avishayt: We can use a regex here if we want to limit the use of templates to a very small

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,21 @@ A fleet's device template contains a device specification that gets applied to a
 
 For example, you could specify in a fleet's device template that all devices in the fleet shall run the OS image `quay.io/flightctl/rhel:9.5`. The Flight Control service would then roll out this specification to all devices in the fleet and the Flight Control agents would update the devices accordingly. The same would apply to the other specification items described in [Managing Devices](managing-devices.md).
 
-However, it would be impractical if *all* of a fleet's devices had to have the *exact same specification*. Flight Control therefore allows templates to contain placeholders that get filled in based on a device's name or label values.
+However, it would be impractical if *all* of a fleet's devices had to have the *exact same specification*. Flight Control therefore allows templates to contain placeholders that get filled in based on a device's name or label values. The syntax for these placeholders matches that of [Go templates](https://pkg.go.dev/text/template). You may reference anything under a device's metadata such as `{{ index .Metadata.Labels \"key\" }}` or {{ .Metadata.Name }}`.
```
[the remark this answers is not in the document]
avishayt: Fixed

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,21 @@ A fleet's device template contains a device specification that gets applied to a
...
 
 Here are some examples what you can do with placeholders in device templates:
 
-* You can label devices by their deployment stage (say, `stage: testing` and `stage: production`) and then use the label with the key `stage` as placeholder when referencing the OS image to use (say, `quay.io/myorg/myimage:latest-{{ .Labels["stage"] }}`) or when referencing a folder with configuration in a Git repository.
+* You can label devices by their deployment stage (say, `stage: testing` and `stage: production`) and then use the label with the key `stage` as placeholder when referencing the OS image to use (say, `quay.io/myorg/myimage:latest-{{ index .Metadata.Labels \"stage\" }}`) or when referencing a folder with configuration in a Git repository.
```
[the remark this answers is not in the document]
avishayt: If a label doesn't exist then:
{{index .Metadata.Labels "key"}} defaults to ""
{{.Metadata.Labels.key}} defaults to "<no value>"

api/v1alpha1/validation.go
celdrake: Please let us know if there will be limited syntax, or if we'll allow the broader syntax from go-templates. This may affect the UX.

fzdarsky: Yeah. Like sending an infinite / long-running loop.

docs/user/managing-fleets.md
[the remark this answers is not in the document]
fzdarsky: So both have the same behaviour? In that case, I personally find the latter easier.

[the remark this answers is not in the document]
avishayt: {{index .Metadata.Labels "key"}} defaults to ""
{{.Metadata.Labels.key}} defaults to "<no value>"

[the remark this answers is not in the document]
avishayt: Sorry, github messed up my comment, please see above :)

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,29 @@ A fleet's device template contains a device specification that gets applied to a
...
 
-Here are some examples what you can do with placeholders in device templates:
+We also provide some helper functions:
 
-* You can label devices by their deployment stage (say, `stage: testing` and `stage: production`) and then use the label with the key `stage` as placeholder when referencing the OS image to use (say, `quay.io/myorg/myimage:latest-{{ .Labels["stage"] }}`) or when referencing a folder with configuration in a Git repository.
-* You can label devices by deployment site (say, `site: factory-berlin` and `site: factory-madrid`) and then use the label with the key `site` as parameter when referencing the secret with network access credentials in Kubernetes.
+* `toUpper`: Change to upper case. For example, `{{ toUpper .Metadata.Name }}`.
+* `toUpper`: Change to lower case. For example, `{{ toLower .Metadata.Labels.key }}`.
```
celdrake, line 99:
```suggestion
* `toLower`: Change to lower case. For example, `{{ toLower .Metadata.Labels.key }}`.
```

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,29 @@ A fleet's device template contains a device specification that gets applied to a
...
+* `replace`: Replace a substring with another string. For example, `{{ replace .Metadata.Labels.key \"old\" \"new\" }}`.
+* `getOrDefault`: Return a default value if accessing a missing label. For example, `{{ getOrDefault .Metadata.Labels \"key\" \"default\" }}`.
+* `getOrError`: Fail if accessing a missing label. For example, `{{ getOrError .Metadata.Labels \"key\" }}`.
 
-The following placeholders are supported:
+Here are some examples what you can do with placeholders in device templates:
```
celdrake, line 104:
```suggestion
Here are some examples of what you can do with placeholders in device templates:
```

docs/user/managing-fleets.md
avishayt: Fixed

docs/user/managing-fleets.md
avishayt: Fixed

api/v1alpha1/util.go
```diff
@@ -270,3 +272,39 @@ func FleetSpecsAreEqual(f1, f2 FleetSpec) bool {
...
+		return strings.Replace(input, old, new, -1)
+	}
+
+	return template.FuncMap{
+		"toUpper":      strings.ToUpper,
+		"toLower":      strings.ToLower,
+		"replace":      replace,
+		"getOrDefault": getOrDefault,
+		"getOrError":   getOrError,
+	}
```
fzdarsky, line 309: Would it make sense to use the same [function names as Helm](https://helm.sh/docs/chart_template_guide/function_list/#string-functions) which adds [Sprig functions](https://masterminds.github.io/sprig/) to the built-in go-template functions?

--> "upper", "lower", "", "replace", "default", ??

I couldn't find an equivalent of "getOrError". I think one would usually use the `index` built-in function (`index .Metadata.Labels "key"`), which returns the empty value if the key doesn't exist, which can then be defaulted. And so if we were to set `missingkey=error`, we'd get the behaviour of `getOrError` if the `.` was used to access `.Metadata.Labels.key`). It'd be familiar to Helm users at least, but may be surprising to others?

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,29 @@ A fleet's device template contains a device specification that gets applied to a
 
 For example, you could specify in a fleet's device template that all devices in the fleet shall run the OS image `quay.io/flightctl/rhel:9.5`. The Flight Control service would then roll out this specification to all devices in the fleet and the Flight Control agents would update the devices accordingly. The same would apply to the other specification items described in [Managing Devices](managing-devices.md).
 
-However, it would be impractical if *all* of a fleet's devices had to have the *exact same specification*. Flight Control therefore allows templates to contain placeholders that get filled in based on a device's name or label values.
+However, it would be impractical if *all* of a fleet's devices had to have the *exact same specification*. Flight Control therefore allows templates to contain placeholders that get filled in based on a device's name or label values. The syntax for these placeholders matches that of [Go templates](https://pkg.go.dev/text/template), but you may only use simple text or actions (no conditionals or loops, for example). You may reference anything under a device's metadata such as `{{ .Metadata.Labels.key }}` or `{{ .Metadata.Name }}`.
```
fzdarsky, line 94: I it surprising to users that they'd have to use the variable names as in our Go structs instead of those in the YAML files (IOW, they need to use `.Metadata.Name` instead of `.metadata.name`)? in these examples it's just the initial capital letter, but in theory the Go field and YAML field could be named very differently.

api/v1alpha1/util.go
avishayt: OK, so:
1. toUpper -> upper
2. toLower -> lower
3. Use missingkey=error and remove getOrError

> I think one would usually use the index built-in function (index .Metadata.Labels "key"), which returns the empty value if the key doesn't exist, which can then be defaulted.
What do you mean?  How can it "then be defaulted"?

docs/user/managing-fleets.md
avishayt: Good point.  Need to think about how to translate that.

docs/user/managing-fleets.md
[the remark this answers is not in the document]
fzdarsky: Got it. So if we change to missingkey=error as discussed below the {{.Metadata.Labels.key}} would return an error instead.

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,29 @@ A fleet's device template contains a device specification that gets applied to a
...
 
-Here are some examples what you can do with placeholders in device templates:
+We also provide some helper functions:
 
-* You can label devices by their deployment stage (say, `stage: testing` and `stage: production`) and then use the label with the key `stage` as placeholder when referencing the OS image to use (say, `quay.io/myorg/myimage:latest-{{ .Labels["stage"] }}`) or when referencing a folder with configuration in a Git repository.
-* You can label devices by deployment site (say, `site: factory-berlin` and `site: factory-madrid`) and then use the label with the key `site` as parameter when referencing the secret with network access credentials in Kubernetes.
+* `toUpper`: Change to upper case. For example, `{{ toUpper .Metadata.Name }}`.
```
fzdarsky, line 98: Should we support and prefer the pipeline syntax?

If there are multiple actions, the pipeline syntax feels cleaner and more intuitive, e.g.:
```
{{ .Metadata.Labels.stage | default "testing" | lower | replace " " "-" }}
```

docs/user/managing-fleets.md
```diff
@@ -91,27 +91,29 @@ A fleet's device template contains a device specification that gets applied to a
...
+* `toLower`: Change to lower case. For example, `{{ toLower .Metadata.Labels.key }}`.
+* `replace`: Replace a substring with another string. For example, `{{ replace .Metadata.Labels.key \"old\" \"new\" }}`.
```
fzdarsky, line 100:
```suggestion
* `replace`: Replace all occurrences of substring with another string. For example, `{{ replace .Metadata.Labels.key \"old\" \"new\" }}`.
```

fzdarsky (approved):

PR #701, opened 2024-12-08
EDM-642: Remove sorting mechanism for TP phase

asafbss:
Following our discussion about EDM-642, due to this bug, only sorting by the default field and direction will be supported in TP (i.e., sorting by metadata.name in ascending order). The sorting parameters have been removed from the API in this PR to revert to implicit sorting, which prevents the bug from occurring.


celdrake (commented):
LGTM. Data seems to be returned sorted by the default criteria, and I couldn't see an issue with paginated requests either.

The UI PR to remove the parameters is ready as well.

avishayt (approved):

PR #703, opened 2024-12-09
EDM-860: Add permission checks for all API endpoints

ori-amizur:
In order to implement RBAC authorization, permission check should be done for every API endpoint

EDN-861: Add document describing the authorization calls

Add document describing the authorization calls including routes, call names, resources, and verbs


docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
+# Authentication resources
+
+The table below contains the routes, names, resource names, and verbs for flightctl API endpoints:
+
```
fzdarsky, line 4: Should align with K8s's pattern:

- POST, GET, DELETE on `.../$resource` --> verbs create, list, deletecollection
- GET, PUT, PATCH, DELTE on `.../$resource/{name}` --> verbs get, **update**, patch, delete
- PUT, PATCH on `.../$resource/{name}/status` --> verbs **update, patch** on resource name `$resource**/status**` <-- need to have separate authorization on this endpoint

docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
...
+|Route| Name| Resource| Verb |
+|-----|-----|---------|------|
+|`DELETE /api/v1/certificatesigningrequests`|`DeleteCertificateSigningRequests`|`certificatessigningrequests`|`deletecollection`|
+|`GET /api/v1/certificatesigningrequests`|`ListCertificateSigningRequests`|`certificatessigningrequests`|`list`|
+|`POST /api/v1/certificatesigningrequests`|`CreateCertificateSigningRequest`|`certificatessigningrequests`|`create`|
+|`DELETE /api/v1/certificatesigningrequests/{name}`|`DeleteCertificateSigningRequest`|`certificatessigningrequests`|`delete`|
+|`GET /api/v1/certificatesigningrequests/{name}`|`ReadCertificateSigningRequest`|`certificatessigningrequests`|`read`|
+|`PATCH /api/v1/certificatesigningrequests/{name}`|`PatchCertificateSigningRequest`|`certificatessigningrequests`|`patch`|
+|`PUT /api/v1/certificatesigningrequests/{name}`|`ReplaceCertificateSigningRequest`|`certificatessigningrequests`|`replace`|
+|`DELETE /api/v1/certificatesigningrequests/{name}/approval`|`DenyCertificateSigningRequest`|`certificatessigningrequests`|`deny`|
```
fzdarsky, line 14: Looks like we have an issue here with the CSR API. This should be PUT/PATCH on the `certificatesigningrequests/approve` subresource to either set the "approved" or "denied" conditions...

docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
...
+|`POST /api/v1/devices`|`CreateDevice`|`devices`|`create`|
+|`GET /api/v1/devices`|`ListDevices`|`devices`|`list`|
+|`DELETE /api/v1/devices`|`DeleteDevices`|`devices`|`deletecollection`|
+|`GET /api/v1/devices/{name}`|`ReadDevice`|`devices`|`read`|
+|`PUT /api/v1/devices/{name}`|`ReplaceDevice`|`devices`|`replace`|
+|`DELETE /api/v1/devices/{name}`|`DeleteDevice`|`devices`|`delete`|
+|`GET /api/v1/devices/{name}/status`|`ReadDeviceStatus`|`devices`|`readstatus`|
+|`PUT /api/v1/devices/{name}/status`|`ReplaceDeviceStatus`|`devices`|`replacestatus`|
```
fzdarsky, line 22:
```suggestion
|`GET /api/v1/devices/{name}/status`|`ReadDeviceStatus`|`devices/status`|`get`|
|`PUT /api/v1/devices/{name}/status`|`ReplaceDeviceStatus`|`devices/status`|`update`|
```

docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
...
+|`GET /api/v1/devices/{name}/rendered`|`GetRenderedDeviceSpec`|`devices`|`getrenderedspec`|
```
fzdarsky, line 23:
```suggestion
|`GET /api/v1/devices/{name}/rendered`|`GetRenderedDeviceSpec`|`devices/rendered`|`get`|
```

docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
...
+|`PUT /api/v1/devices/{name}/decommission`|`DecommissionDevice`|`devices`|`decomission`|
```
fzdarsky, line 24:
```suggestion
|`PUT /api/v1/devices/{name}/decommission`|`DecommissionDevice`|`devices/decommission`|`update`|
```

docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
...
+|`POST /api/v1/enrollmentrequests`|`CreateEnrollmentRequest`|`enrollmentrequests`|`create`|
+|`GET /api/v1/enrollmentrequests`|`ListEnrollmentRequests`|`enrollmentrequests`|`list`|
+|`DELETE /api/v1/enrollmentrequests`|`DeleteEnrollmentRequests`|`enrollmentrequests`|`deletecollection`|
+|`GET /api/v1/enrollmentrequests/{name}`|`ReadEnrollmentRequest`|`enrollmentrequests`|`read`|
+|`PUT /api/v1/enrollmentrequests/{name}`|`ReplaceEnrollmentRequest`|`enrollmentrequests`|`replace`|
+|`DELETE /api/v1/enrollmentrequests/{name}`|`DeleteEnrollmentRequest`|`enrollmentrequests`|`delete`|
+|`GET /api/v1/enrollmentrequests/{name}/status`|`ReadEnrollmentRequestStatus`|`enrollmentrequests`|`readstatus`|
+|`POST /api/v1/enrollmentrequests/{name}/approval`|`ApproveEnrollmentRequest`|`enrollmentrequests`|`approve`|
```
fzdarsky, line 32:
```suggestion
|`POST /api/v1/enrollmentrequests/{name}/approval`|`ApproveEnrollmentRequest`|`enrollmentrequests/approval`|`update`|
```

fzdarsky: Again looks like an issue in our API that we use POST instead of PUT...

docs/user/auth-resources.md
ori-amizur: The resource being read/updated is the device.  Do we want to have different resource?
Can a resource contain a '/' ? I don't recall that a resource in CRD can contain a '/'.

docs/user/auth-resources.md
```diff
@@ -0,0 +1,56 @@
...
+|`POST /api/v1/resourcesyncs`|`CreateResourceSync`|`resourcesyncs`|`create`|
+|`GET /api/v1/resourcesyncs`|`ListResourceSync`|`resourcesyncs`|`list`|
+|`DELETE /api/v1/resourcesyncs`|`DeleteResourceSyncs`|`resourcesyncs`|`deletecollection`|
+|`GET /api/v1/resourcesyncs/{name}`|`ReadResourceSync`|`resourcesyncs`|`read`|
+|`PUT /api/v1/resourcesyncs/{name}`|`ReplaceResourceSync`|`resourcesyncs`|`replace`|
+|`DELETE /api/v1/resourcesyncs/{name}`|`DeleteResourceSync`|`resourcesyncs`|`delete`|
+|`GET /api/v1/api/v1/fleets/{fleet}/templateVersions`|`ListTemplateVersions`|`templateversions`|`list`|
+|`DELETE /api/v1/api/v1/fleets/{fleet}/templateVersions`|`DeleteTemplateVersions`|`templateversions`|`deletecollection`|
+|`GET /api/v1/fleets/{fleet}/templateVersions/{name}`|`ReadTemplateVersion`|`templateversions`|`read`|
+|`DELETE /api/v1/fleets/{fleet}/templateVersions/{name}`|`DeleteTemplateVersion`|`templateversions`|`delete`|
```
fzdarsky, line 56: This needs some thinking, as templateVersions is a subresource of a specific fleet.
I think the resource should be called `fleets/templateVersions` and the verbs list/deletecollection and get/delete.

internal/service/certificatesigningrequest.go
```diff
@@ -56,9 +57,16 @@ func signApprovedCertificateSigningRequest(ca *crypto.CA, request api.Certificat
 
 // (DELETE /api/v1/certificatesigningrequests)
 func (h *ServiceHandler) DeleteCertificateSigningRequests(ctx context.Context, request server.DeleteCertificateSigningRequestsRequestObject) (server.DeleteCertificateSigningRequestsResponseObject, error) {
+	allowed, err := auth.GetAuthZ().CheckPermission(ctx, "certificatessigningrequests", "deletecollection")
+	if err != nil {
+		return server.DeleteCertificateSigningRequests401JSONResponse{Message: fmt.Sprintf("auth failed: %v", err)}, nil
+	}
```
fzdarsky, line 63: 401 would be "Unauthorized", but when we get to this point in the code, middleware should already have rejected unauthorized requests, or?
Also, we shouldn't return detailed error messages to users, as they can accidentally leak info (I know we still do that in many places in our APIs).

internal/service/certificatesigningrequest.go
```diff
@@ -56,9 +57,16 @@ func signApprovedCertificateSigningRequest(ca *crypto.CA, request api.Certificat
...
+	if !allowed {
+		return server.DeleteCertificateSigningRequests403JSONResponse{Message: "cannot delete certificate signing requests"}, nil
```
fzdarsky, line 65: Could we define an error constant for this for consistency? I think a "Forbidden" would even be enough, as the rest is clear from the API call itself.

internal/service/certificatesigningrequest.go
ori-amizur: Do you think it should be removed? I don't think so

docs/user/auth-resources.md
ori-amizur: It is not in this PR scope

fzdarsky: What I meant is you correctly modelled it after our current API, but the API should be different.

docs/user/auth-resources.md
fzdarsky: The resource name is just a string, we can use whatever we want. The point is that the authorization for the status must be different, i.e. we need to allow updating status without being able to update spec, for example.

docs/user/auth-resources.md
ori-amizur: That's why I said it is not for this PR. Once the API changes, the auth model should change too.

internal/service/certificatesigningrequest.go
fzdarsky: No. What I'm wondering is whether a 401 is the appropriate response, because - despite its name being "Unauthorized" it is about attempt to access a resource without proper auth_entication_. Thus, the user would be asked to log in once more, but that wouldn't help.

Would it be a 403 - Forbidden, considering that the effect is the permissions check failed? I also don't think so, because CheckPermissions() failing isn't about lack of privileges, but about an error that's occurring on the side of the service. So... would a 5xx code be more appropriate?

ori-amizur: 500 sounds OK - it sounds like internal server error.

internal/service/certificatesigningrequest.go
```diff
@@ -56,9 +57,16 @@ func signApprovedCertificateSigningRequest(ca *crypto.CA, request api.Certificat
 
 // (DELETE /api/v1/certificatesigningrequests)
 func (h *ServiceHandler) DeleteCertificateSigningRequests(ctx context.Context, request server.DeleteCertificateSigningRequestsRequestObject) (server.DeleteCertificateSigningRequestsResponseObject, error) {
+	allowed, err := auth.GetAuthZ().CheckPermission(ctx, "certificatesigningrequests", "deletecollection")
+	if err != nil {
+		return server.DeleteCertificateSigningRequests503JSONResponse{Message: fmt.Sprintf("auth failed: %v", err)}, nil
```
fzdarsky, line 62: Can you please also update the error message ("Authorisation server unavailable") and not return the `err`?
Instead, we probably want to log the error.

fzdarsky (approved):

PR #706, opened 2024-12-10
Console on websocket

mangelajo:
1. This is keeping  the gRPC <-> gRPC transport for the console until UI migrates to websocket,
to avoid any release issues.
2. for the reason in (1) and to prove I am not breaking it, I am not doing the flightctl console ....  cmdline change in the same PR
3. This still continues to use the same annotation mechanisms, single console at once

but it's preparing for a few things:
a) Doing multiple consoles eventually, I create a ConsoleManager class
b) Moving from in-process channels for console comm (tx/rx), to distributed (through two queues per connection in redis/valkey)
that last thing would allow for horizontally scaling flightctl-api (or eventually splitting out the agent service) without breaking the console

can be tested with:
```
$ podman run --net=host --rm -ti ghcr.io/vi/websocat:nightly --binary -E -k wss://api.10.0.2.2.nip.io:3443/ws/v1/devices/o18jbk21am279cspp77lhb9f55eechk5v3on22booki0igbl1vb0/console
bash: cannot set terminal process group (2692): Inappropriate ioctl for device
bash: no job control in this shell
[root@localhost /]#

[root@localhost /]#
```

### DIFF
--- a/internal/api_server/agentserver/grpc.go
+++ b/internal/api_server/agentserver/grpc.go
@@ -11,6 +11,7 @@ import (
 	pb "github.com/flightctl/flightctl/api/grpc/v1"
 	"github.com/flightctl/flightctl/internal/api_server/middleware"
 	"github.com/flightctl/flightctl/internal/config"
+	"github.com/flightctl/flightctl/internal/console"
 	"github.com/flightctl/flightctl/internal/consts"
 	grpcAuth "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
 	"github.com/sirupsen/logrus"
@@ -45,7 +46,7 @@ func NewAgentGrpcServer(
 }
 
 func (s *AgentGrpcServer) Run(ctx context.Context) error {
-	s.log.Printf("Initializing Agent-side gRPC server: %s", s.cfg.Service.AgentGrpcAddress)
+	s.log.Infof("Initializing Agent-side gRPC server: %s", s.cfg.Service.AgentGrpcAddress)
 	tlsCredentials := credentials.NewTLS(s.tlsConfig)
 	server := grpc.NewServer(
 		grpc.Creds(tlsCredentials),
@@ -68,11 +69,34 @@ func (s *AgentGrpcServer) Run(ctx context.Context) error {
 }
 
 type streamCtx struct {
-	cancel context.CancelFunc
-	stream pb.RouterService_StreamServer
-	closed bool // one side closed the connection and we should not accept any more messages
+	cancel  context.CancelFunc
+	stream  pb.RouterService_StreamServer
+	session *console.ConsoleSession // we let the console sessions coexist for a while until UI gets fully migrated out of gRPC
+	closed  bool                    // one side closed the connection and we should not accept any more messages
 }
 
+// This function is called by the console session manager to signal that it is ready to accept connections
+// from the agent. This is a way to provide the gRPC service with the session UUID and the channels
+func (s *AgentGrpcServer) StartSession(session *console.ConsoleSession) error {
+	sctx := streamCtx{
+		cancel:  nil,
+		stream:  nil,
+		closed:  false,
+		session: session,
+	}
+	s.log.Infof("the console manager started the session %s for device %s, waiting for stream", session.UUID, session.DeviceName)
+	// we store the session in the map, so when the agent connects we can start forwarding
+	// bytes between the two client and the device agent
+	s.pendingStreams.Store(session.UUID, sctx)
+	return nil
+}
+
+// This function is called by the console session manager to signal that the session has been closed
+// from the other side
+func (s *AgentGrpcServer) CloseSession(session *console.ConsoleSession) error {
+	s.log.Infof("the console manager closed the session %s for device %s", session.UUID, session.DeviceName)
+	return nil
+}
 func (s *AgentGrpcServer) Stream(stream pb.RouterService_StreamServer) error {
 	ctx := stream.Context()
 	md, ok := metadata.FromIncomingContext(ctx)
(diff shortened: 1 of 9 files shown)


rawagner:
works great on UI side with https://github.com/flightctl/flightctl-ui/pull/173

rawagner (approved):

internal/api_server/agentserver/grpc.go
```diff
@@ -108,43 +132,116 @@ func (s *AgentGrpcServer) Stream(stream pb.RouterService_StreamServer) error {
...
 		// the map did not have a value, we are the first client, so we wait for the second
 		s.log.Infof("client %s waiting for peer %s", clientName, sessionId)
 		<-ctx.Done()
 		return nil
 	}
+	return nil
+}
+
+func (s *AgentGrpcServer) forwardConsoleSession(ctx context.Context, stream pb.RouterService_StreamServer, otherSideSession *console.ConsoleSession, sessionId string, actual any) error {
+	err := s.forwardChannels(ctx, stream, otherSideSession)
```
hexfusion, line 164: minor: 

```suggestion
        return s.forwardChannels(ctx, stream, otherSideSession)
````

File: internal/api_server/agentserver/grpc.go [hunk 2]
```diff
@@ -108,43 +132,116 @@ func (s *AgentGrpcServer) Stream(stream pb.RouterService_StreamServer) error {
...
+}
+
+func (s *AgentGrpcServer) forwardChannels(ctx context.Context, a pb.RouterService_StreamServer, b *console.ConsoleSession) error {
+	g, _ := errgroup.WithContext(ctx)
+	g.Go(func() error { return s.pipeStreamToChannel(a, b.RecvCh) })
+	g.Go(func() error { return s.pipeChannelToStream(b.SendCh, a) })
+	return g.Wait()
+}
+
+func (s *AgentGrpcServer) pipeStreamToChannel(a pb.RouterService_StreamServer, ch chan []byte) error {
```
hexfusion, line 177: can we add a ctx this this signature and add a select case for <-ctx.Done()?

internal/api_server/agentserver/grpc.go
```diff
@@ -108,43 +132,116 @@ func (s *AgentGrpcServer) Stream(stream pb.RouterService_StreamServer) error {
...
+		payload := msg.GetPayload()
+		closed := msg.GetClosed()
+		ch <- payload
+		if closed {
+			return io.EOF
+		}
+	}
+}
+
+func (s *AgentGrpcServer) pipeChannelToStream(ch chan []byte, a pb.RouterService_StreamServer) error {
```
hexfusion, line 196: same here it would be good for these to respect ctx

internal/console/console.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := m.store.Device().UpdateAnnotations(ctx, orgId, deviceName, annotations, []string{}); err != nil {
+		return nil, err
+	}
+
+	// tell the gRPC service, or the message queue (in the future) that there is a session waiting, and provide
+	// the channels to read and write to the websocket session
+	if err := m.sessionRegistration.StartSession(session); err != nil {
+		// if we fail to register the session we should remove the annotation (best effort)
+		deleteAnnotations := []string{api.DeviceAnnotationConsole}
+		_ = m.store.Device().UpdateAnnotations(ctx, orgId, deviceName, map[string]string{}, deleteAnnotations)
```
hexfusion, line 77: should we at least log this error?

internal/service/console.go
```diff
@@ -2,14 +2,132 @@ package service
...
+			h.log.Errorf("console requested for unknown device: %s", deviceName)
+			http.Error(w, "Device not found", http.StatusNotFound)
+		default:
+			h.log.Errorf("There was an error retrieving DB from database during console request: %v", err)
+			http.Error(w, "Internal server error", http.StatusInternalServerError)
+		}
+		return
+	}
+
+	// Upgrade the HTTP connection to a WebSocket
```
hexfusion, line 43: should we validate ws related headers exist before upgrade?

internal/service/console.go
```diff
@@ -2,14 +2,132 @@ package service
...
+	CheckOrigin: func(r *http.Request) bool {
+		return true // Allow connections from any origin
+	},
+}
+
+func (h *WebsocketHandler) HandleDeviceConsole(w http.ResponseWriter, r *http.Request) {
+	orgId := store.NullOrgId
+	deviceName := chi.URLParam(r, "name")
+
+	h.log.Infof("websocket console connection requested for device: %s", deviceName)
```
hexfusion, line 28: should this be debug?

hexfusion (commented):
Few minor things otherwise LGTM

mangelajo:
Thank you for all the comments, working on them.

internal/api_server/agentserver/grpc.go
mangelajo: Done, it's not perfect because a.Recv is blocking, but as soon as it loops once it will find the context Done and get out.

internal/console/console.go
mangelajo: this one if happens and the one at the top of the if.

internal/service/console.go
mangelajo: I haven't seen examples of that , at this point we are under the /ws/v1/  API path.

Ahh, investigating a bit more, this is handled by the upgrade function already:

image

mangelajo:
@rawagner @hexfusion suggestions applied, thanks for the reviews!, thank you
@rawagner for trying

internal/service/console.go
hexfusion: nice thx for digging

hexfusion (approved):

PR #708, opened 2024-12-11
EDM-846: Deploy default fctl roles in ACM env

rawagner:
@fzdarsky do you have some specific roles in mind ?

## Summary by CodeRabbit

- **New Features**
	- Added Kubernetes Role definitions for different user types in the `flightctl` application.
	- Introduced roles for admin, viewer, operator, and installer with specific access permissions.
	- Implemented role-based access control (RBAC) for fine-grained user management.


deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - devices
+      - fleets
+      - resourcesyncs
+  - verbs:
+      - get
+      - list
+    apiGroups:
+      - flightctl.io
+    resources:
+      - repositories
```
fzdarsky, line 58: Means they'll also be able to see the secrets for accessing the repos...

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - update
+      - patch
+      - post
+    apiGroups:
+      - flightctl.io
+    resources:
+      - enrollmentrequests
+      - enrollmentrequests/approval
+      - certificatesigningrequests
+      - certificatesigningrequests/approval
```
fzdarsky, line 82: I think they need to create/list/get CSRs for the enrollment-certs (though I have a headache that they'll be able to access also get certs they didn't request themselves, esp. agent certs this way), but the enrollment certs should be auto-approved, so they shouldn't need/have that permission, I think.

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - verbs:
+      - get
+      - list
+    apiGroups:
+      - flightctl.io
+    resources:
+      - devices
```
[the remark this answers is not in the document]
rawagner: Ok, removing

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
fzdarsky: Hm... yeah, the persona would need to get a list of available repos without being able to list the details of these repos and we don't support that AFAIK. Result of not having a first-class Secret resource IMO.

The UI needs this to create a drop-down list, or? Means I don't think we can remove...

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -0,0 +1,95 @@
...
+rules:
+  - verbs:
+      - get
+      - list
+    apiGroups:
+      - flightctl.io
+    resources:
+      - enrollmentrequests
+  - verbs:
+      - post
```
fzdarsky, line 82: Needs to be changed to `put` as result of PR #748.

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -0,0 +1,95 @@
+{{ if eq .Values.global.target "acm" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: flightctl-admin
```
fzdarsky, line 5: The `flightctl-` prefix on the roles looks a bit looks a bit verbose, but guess we'll need it in K8s to avoid collisions. :/

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
rawagner: Yep, UI is creating a drop-down.
Ok, so Im keeping it even though it isn't the best.

Hopefully we will add some kind of Secret resource later on.

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
rawagner: Even though HTTP method changed to `put`, the RBAC verb did not. It is still `post` https://github.com/flightctl/flightctl/blob/main/internal/service/enrollmentrequest.go#L359

So if we won't change it, the role definition is correct.

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
rawagner: +1

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -0,0 +1,96 @@
...
+    resources:
+      - devices/console
+  - verbs:
+      - get
+      - list
+    apiGroups:
+      - flightctl.io
+    resources:
+      - repositories
+      - fleets/templateVersions
```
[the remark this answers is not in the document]
rawagner: Im not sure - in any case, i've changed it to lowercase.

ori-amizur (approved):

PR #710, opened 2024-12-12
EDM-786 Agent updates status only on actual changes

gshilin-sdb:
A heartbit will be sent in case there are no changes in status or any error happened while collecting status data.

For now heartbit only updates a device.status.lastSeen field. Later it can be saved separately in a keyval storage.


api/v1alpha1/agent/openapi.yaml
```diff
@@ -52,6 +52,40 @@ paths:
             application/json:
               schema:
                 $ref: '../openapi.yaml#/components/schemas/Error'
+  /api/v1/devices/{name}/heartbeat:
```
hexfusion, line 55: All API changes should have a design doc do can you link that here so we can talk about the proposed solution?

hexfusion (changes requested):

api/v1alpha1/agent/openapi.yaml
gshilin-sdb: No... I made it based on ERM-786...
Can you point me to any design doc of this kind and I'll create a new one

hexfusion (commented):
High level what does a heartbeat mean? We should define this. Does it mean that the agent has connectivity and TLS credentials for the management service? In that case would any request to `GetRenderedDeviceSpec` [1] also be a heartbeat? 

We should talk this through a bit.

[1] https://github.com/flightctl/flightctl/blob/68dbb18ec506a6008bd7c6c1fba8a09d50ec365d/internal/service/common/device.go#L218

gshilin-sdb:
Agent sends a status every so seconds. The general idea is to send a minimum data in case state was not changed, just to remind that device is still alive. Today it will update lastSeen field. The next step may be to move that field outside of DB, for example to redis, and to take it from there for "get devices".

api/v1alpha1/agent/openapi.yaml
hexfusion: Gotcha EDM-786 was not marked as `Ready` so @avishayt was not aware it was being worked on.

hexfusion: image

hexfusion: > an you point me to any design doc of this kind and I'll create a new one

You could  simplify to meet your needs

[1] https://docs.google.com/document/d/1t0ymgOUwhkA99DTZNmHISesavagwl3G6OSTctKMndQ4/edit?tab=t.0#heading=h.bs36tijs8419

hexfusion:
> Agent sends a status every so seconds. The general idea is to send a minimum data in case state was not changed, just to remind that device is still alive. Today it will update lastSeen field. The next step may be to move that field outside of DB, for example to redis, and to take it from there for "get devices".

Makes sense so it sounds like it doesn't matter if status is sent or not for a heartbeat.

- Could we have a patch for ReplaceStatus which would send back 304 if no change vs updating the timestamp?
- Can we use GetRenderedDeviceSpec as a heartbeat?

Point being we should consider alternatives before we add new API. Thanks for creating the doc!

fzdarsky:
> Makes sense so it sounds like it doesn't matter if status is sent or not for a heartbeat.
> 
> * Could we have a patch for ReplaceStatus which would send back 304 if no change vs updating the timestamp?
> * Can we use GetRenderedDeviceSpec as a heartbeat?
> 
> Point being we should consider alternatives before we add new API. Thanks for creating the doc!

+1 to considering alternatives to adding a new API endpoint. The agent checking in for spec updates could be a heartbeat. If we added support for delta status updates (patches) an empty status update could be a heartbeat. When we have gRPC connections to the service, heartbeats may not even be needed.

mangelajo:
I also agree, we could take any device activity as "heartbeat" (get spec, update state... etc). Any contact made from the device would update the lastSeen status and we avoid an additional API method.

romfreiman:
@coderabbitai review

PR #714, opened 2024-12-13
EDM-768: Run linter on OpenAPI specs

fzdarsky:
Adds linting of OpenAPI specs with Spectral. A set of rules is included that
- enforce consistency for verbs, error codes, and request/response media types across resource kinds and collections
- enforce having descriptions for parameters and properties
- warn if best practices are not followed

This PR will be complemented by a number of other PRs cleaning up the specs.

## Summary by CodeRabbit

- **New Features**
	- Added a GitHub Actions workflow for API specification linting.
	- Introduced comprehensive OpenAPI specification validation rules.
	- Enhanced Makefile with new linting targets and improved help output.

- **Chores**
	- Set up automated checks for API specification quality and consistency.

### DIFF
--- a/.spectral.yaml
+++ b/.spectral.yaml
@@ -1,0 +1,500 @@
+extends:
+  - [spectral:oas, all]
+  # - "https://unpkg.com/@stoplight/spectral-owasp-ruleset/dist/ruleset.mjs"
+
+aliases:
+  PathToResourceCollection:
+    # To prevent enrollmentconfig from being classified as collection, we select everything
+    # except segments starting with 'e', then explicitly add enrollmentrequests.
+    - '$.paths[?(@property.match( "^/api/v1/[a-d,f-z][a-z]+$" ))]'
+    - '$.paths[?(@property.match( "^/api/v1/enrollmentrequests$" ))]'
+  PathToResource:
+    - '$.paths[?(@property.match( "^/api/v1/[a-z]+/{name}$" ))]'
+  PathToStatusSubresource:
+    - '$.paths[?(@property.match( "^/api/v1/[a-z]+/{name}/status$" ))]'
+  PathToApprovalSubresource:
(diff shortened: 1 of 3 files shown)


.spectral.yaml
```diff
@@ -0,0 +1,515 @@
...
+    severity: error
+    given: $path.[?(@property.match( /(get|delete)/ ))].parameters..in
+    then:
+      function: pattern
+      functionOptions:
+        notMatch: "/^body$/"
+
+  ## Allowed and required responses
+  responses-collection-post-required:
+    description: POST on resource collections must have responses 200, 400, 401, 403.
```
avishayt, line 186: I guess it should be 201 and not 200

.spectral.yaml
```diff
@@ -0,0 +1,515 @@
...
+        function: truthy
+      - field: responses.400
+        function: truthy
+      - field: responses.401
+        function: truthy
+      - field: responses.403
+        function: truthy
+
+  responses-collection-post-allowed:
+    description: POST on resource collections must not have responses other than 200, 400, 401, 403.
```
avishayt, line 201: 201 not 200

avishayt (commented):
Looks awesome!  One small comment, but we'll probably find more once we start fixing the errors.
Another thing to consider (not necessarily part of this PR) - we return 400 for any client-side error, but that seems to be more for cases where the server didn't understand the request (e.g., malformed syntax).  422 is more suitable for cases where the server understood the request but couldn't perform it.

avishayt:
We have a couple instances of returning 418, these should probably be 422.

mangelajo:
It looks awesome, it would be good if we could sanitize all the API before we release TP

avishayt:
I suggest
1. Make the check optional
2. Let's split the work to fix the issues
3. Make the check mandatory

PR #722, opened 2024-12-16
NO-ISSUE: Use a password for accessing Valkey

avishayt:
Two commits:
1. Move key-value code out of tasks package: Previously the code was in the tasks package but used outside of it.  Moved to internal/kvstore.
2. Use a password for accessing Valkey: Valkey now uses a password, but it's in values.yaml like the other components.  We'll move them to Secrets as a separate effort.


deploy/helm/flightctl/templates/flightctl-api-config.yaml
```diff
@@ -46,6 +46,7 @@ data:
     kv:
         hostname: flightctl-kv.{{ default .Release.Namespace .Values.global.internalNamespace }}.svc.cluster.local
         port: 6379
+        password: {{ .Values.kv.password }}   # we should funnel this via secrets instead
```
mangelajo, line 49: I agree, if we cannot do it now, can we track it somehow in JIRA?

mangelajo: we have the same issue with database, probably both should be on the same JIRA.

internal/console/console.go
```diff
@@ -32,18 +33,18 @@ type ConsoleSessionManager struct {
...
 }
 
-func NewConsoleSessionManager(store store.Store, callbackManager tasks.CallbackManager, configStorage tasks.ConfigStorage, log logrus.FieldLogger, sessionRegistration InternalSessionRegistration) *ConsoleSessionManager {
+func NewConsoleSessionManager(store store.Store, callbackManager tasks.CallbackManager, kvStore kvstore.KVStore, log logrus.FieldLogger, sessionRegistration InternalSessionRegistration) *ConsoleSessionManager {
 	return &ConsoleSessionManager{
 		store:               store,
 		log:                 log,
 		callbackManager:     callbackManager,
-		configStorage:       configStorage,
+		kvStore:             kvStore,
```
mangelajo, line 47: I plan to use the queues from valley here in next PRs

mangelajo (approved):

deploy/helm/flightctl/templates/flightctl-api-config.yaml
avishayt: I know, rabbitmq as well.  That's why I wanted to level set valkey first, then fix them all.

PR #725, opened 2024-12-17
EDM-922: Fine-tune HTTP server settings

avishayt:
Please pay special attention to the values and see if they seem suitable.

I tested by adjusting our CLI to send out bad requests:

**Too many headers:**
```
$ bin/flightctl apply -f examples/device.yaml 
device: applying examples/device.yaml/f62dfb4f5d2cdbb9339362b5f18f2ce268d75649bdc80003f0e04292a7ef8c3: 413 Request Entity Too Large
Request has too many headers, exceeds 10

Error: device: failed to apply examples/device.yaml/f62dfb4f5d2cdbb9339362b5f18f2ce268d75649bdc80003f0e04292a7ef8c3: 413 Request Entity Too Large
```

**Header too big:**
```
$ bin/flightctl apply -f examples/device.yaml 
device: applying examples/device.yaml/f62dfb4f5d2cdbb9339362b5f18f2ce268d75649bdc80003f0e04292a7ef8c3: 431 Request Header Fields Too Large
431 Request Header Fields Too Large
Error: device: failed to apply examples/device.yaml/f62dfb4f5d2cdbb9339362b5f18f2ce268d75649bdc80003f0e04292a7ef8c3: 431 Request Header Fields Too Large
```

**URL too long:**
```
$ bin/flightctl apply -f examples/device.yaml 
device: applying examples/device.yaml/f62dfb4f5d2cdbb9339362b5f18f2ce268d75649bdc80003f0e04292a7ef8c3: 414 Request URI Too Long
URL too long, exceeds 2000 characters

Error: device: failed to apply examples/device.yaml/f62dfb4f5d2cdbb9339362b5f18f2ce268d75649bdc80003f0e04292a7ef8c3: 414 Request URI Too Long
```


avishayt:
@fzdarsky Note that we're returning some 4XX HTTP codes that aren't defined in openapi and it works fine.

internal/api_server/server.go
```diff
@@ -68,6 +68,13 @@ func oapiErrorHandler(w http.ResponseWriter, message string, statusCode int) {
 	http.Error(w, fmt.Sprintf("API Error: %s", message), statusCode)
 }
 
+const (
+	readTimeout    = time.Second * 5 // Set request read timeout to 5 seconds
```
hexfusion, line 72: Question: what happens when the server is under resource starvation and this timeout is met? What is the advantage of having this set?

internal/api_server/server.go
```diff
@@ -99,8 +106,27 @@ func (s *Server) Run(ctx context.Context) error {
...
+				}
+
+				next.ServeHTTP(w, r)
+			})
+		}
+	}
+
 	// general middleware stack for all route groups
 	router.Use(
+		middleware.RequestSize(1024*1024), // Limit request body size to 1MB
```
fzdarsky, line 128: That seems a bit tight. In our validation code, we allow inline configs and certificates that are larger than this.
https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/validation.go#L16-L17

internal/api_server/server.go
```diff
@@ -99,8 +106,27 @@ func (s *Server) Run(ctx context.Context) error {
...
+		requestSizeLimiter(),
```
fzdarsky, line 129: I'm wondering whether these should be before or after RequestID and Logger? Feels to me these requests should be logged, too.

internal/api_server/server.go
```diff
@@ -126,6 +152,8 @@ func (s *Server) Run(ctx context.Context) error {
 	ws.RegisterRoutes(router)
 
 	srv := tlsmiddleware.NewHTTPServer(router, s.log, s.cfg.Service.Address)
+	srv.ReadTimeout = readTimeout
+	srv.MaxHeaderBytes = maxHeaderBytes
```
fzdarsky, line 156: Should these be moved into the server's constructor?

internal/api_server/server.go
avishayt: Good point.  Certificates can be up to 20MB?
What do you suggest for this value?  50MB?

internal/api_server/server.go
avishayt: Sure

internal/api_server/server.go
avishayt: Will do

internal/api_server/server.go
hexfusion: This is a good article[1] on the topic they recommend setting `ReadHeaderTimeout` vs `ReadTimeout`. I just want to make sure this is what we want.

[1] https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6

hexfusion: Brad notes `ReadTimeout` but this is pretty old[1].

[1] https://groups.google.com/g/golang-nuts/c/MFZd6b8zQTQ/m/ISD96eyXtD4J

internal/api_server/middleware/tls.go
```diff
@@ -20,10 +20,12 @@ const TLSCommonNameContextKey contextKey = "tls-cn"
...
-		Addr:         address,
-		Handler:      router,
-		ReadTimeout:  5 * time.Second,
-		WriteTimeout: 10 * time.Second,
+		Addr:              address,
+		Handler:           router,
+		ReadTimeout:       5 * time.Second,
+		ReadHeaderTimeout: 5 * time.Second,
+		WriteTimeout:      10 * time.Second,
+		MaxHeaderBytes:    4096,
```
fzdarsky, line 28: Comparing notes with [K8s's API server config(https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L165-L173) and [kubelet's](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/server.go#L176-L183) they define IdleTimeout and set much higher values for MaxHeaderBytes (though IDK of a situation we'd ever have such large headers?JWTs? ) and timeouts.

avishayt: 4 hours for ReadTimeout?  Odd.
I assume we'll change these values a few times along the way.  We just need something sane to start with.

mangelajo: Should we make all this configurable parameters? (with sensible defaults, of course)

I am just worried that under unexpected circumstances this could not work out on the field, and then customers have no escape hatch.

mangelajo: The timeouts are probably very low, and problematic for console over HTTP or gRPC over HTTP as we want to do, I'd at least bump those at least an order of magnitude

internal/api_server/server.go
```diff
@@ -68,6 +68,12 @@ func oapiErrorHandler(w http.ResponseWriter, message string, statusCode int) {
 	http.Error(w, fmt.Sprintf("API Error: %s", message), statusCode)
 }
 
+const (
+	maxUrlLength   = 2000             // Limit the length of URLs
+	maxNumHeaders  = 10               // Limit the number of headers in a request
```
mangelajo, line 73:
```suggestion
	maxNumHeaders  = 10               // Limit the number of headers in a request
```

Why so low?, If we have proxies of any sort in the middle, those can include extra and unexpected headers.

internal/api_server/middleware/tls.go
avishayt: Set all timeouts to 5 minutes by default, and headers to 32KB.  All configurable.

internal/api_server/server.go
avishayt: Increased the default to 32 and made all of these configurable too.

internal/config/config.go
```diff
@@ -153,7 +168,7 @@ func Load(cfgFile string) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("reading config file: %v", err)
 	}
-	c := &Config{}
+	c := NewDefault()
```
avishayt, line 171: Pay attention to this change in behavior - it uses the defaults if no value is specified

PR #727, opened 2024-12-17
EDM-943: Add document describing k8s RBAC authorization

ori-amizur:
## Summary by CodeRabbit

- **Documentation**
	- Added documentation for Kubernetes authorization and Role Based Access Control (RBAC)
	- Introduced configuration details for OpenShift authentication
	- Explained Kubernetes RBAC authorization mechanisms and API endpoint access

- **Chores**
	- Updated spelling configuration with new technical terms related to Kubernetes and OpenShift


docs/user/kubernetes-auth.md
```diff
@@ -0,0 +1,25 @@
+# Authorization with kubernetes
+
+Flight control kubernetes authorization uses Kubernetes RBAC to control authorization for flight control API endpoints.
```
fzdarsky, line 3: A few spelling errors: "Flight Control", "Kubernetes", "Helm"

docs/user/kubernetes-auth.md
```diff
@@ -0,0 +1,25 @@
...
+Some helm configuration values are needed to be modified (see [values.yaml](https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/values.yaml#L34-L41))
```
fzdarsky, line 4: Linking to specific lines is brittle. We'll probably want to link to a (to be created) "Installing Flight Control on Kubernetes" section later.

docs/user/kubernetes-auth.md
```diff
@@ -0,0 +1,25 @@
...
+
+The following variables need to be set:
+* Set **global.auth.type** to "openshift"
+* Set **global.auth.openShiftApiUrl** to the openshift API URL
+* Optionally set **caCert**, **insecureSkipTlsVerify**, **internalOpenShiftApiUrl**
+
+If deploying on ACM  (by global.target: acm), the k8s auth values are automatically calculated.
+
+## Kubernetes RBAC authorization
```
fzdarsky, line 13: Can you please add information on the default roles and their permissions as well as an example for how to create a new role? It's also not clear from the below description which scope (Role vs ClusterRole) a user should use and when.

rawagner: currently we are checking cluster-wide access (ClusterRole). Is that ok ? In case we deploy multiple FCTL, we wont be able to set different permissions for different deployments.

We could switch to checking namespace-scoped access (Role), where FCTL would check Role for the same namespace as FCTL deployment.

ori-amizur: Added explanation when to use Role/ClusterRole

docs/user/kubernetes-auth.md
ori-amizur: Fixed

docs/user/kubernetes-auth.md
rawagner: let me unresolve this, as we need to align on what should we be checking.
Currently our code checks for cluster-wide access **only**, which will cause issues with multiple FCTL deployments. Do we want to switch to namespace-scope check ? @fzdarsky

ori-amizur: Why do we need to switch? Can't we support both?

rawagner: we can support both, but now we are doing only cluster-wide

rawagner: on the other hand, it seems a bit dangerous doesn't it ?
If you'll have a deployment with ns-scoped check and then you deploy another one with cluster-scoped, creating a ClusterRole/ClusterRoleBiding will affect the ns-scoped deployment.

ori-amizur: Isn't it a limitation not to support multiple namespaces?
How will the namespace be set for authorization? Is it on flightctl level?

rawagner: Actually we may be misunderstanding each other.
After we switch to ns-scoped check, you can still use ClusterRole/ClusterRoleBinding and it would work fine.
just that the FCTL would always do ns-scoped check, never just cluster-scoped.

ori-amizur: OK, so no need to change the doc.

rawagner: Yes, if we merge https://github.com/flightctl/flightctl/pull/736

rawagner (approved):

rawagner (approved):

PR #733, opened 2024-12-18
NO-ISSUE: Update GH runner

fzdarsky:
We're currently using `ubuntu-latest` as GH runners, which currently still links to 22.04, but will soon automatically update to 24.04. See: https://github.com/actions/runner-images/issues/10636

This PR makes the switch now and pins the version to 24.04. As 24.04 comes with Podman 4.9, we can also remove the separate installation from SuSE repos.


celdrake (approved):
LGTM

PR #736, opened 2024-12-18
EDM-948: Allow specifying Namespace for k8s RBAC checks

rawagner:
This PR switches our k8s RBAC checks to be ns-scoped instead of cluster-scoped.
Admins of the k8s cluster then can decide whether to use Role/RoleBinding or ClusterRole/ClusterRoleBinding.
Without this change, only ClusterRole/ClusterRoleBinding was useable.

By default FCTL will check the same ns as is deployed in. This could be overridden in values.yaml.


deploy/helm/flightctl/templates/flightctl-api-config.yaml
```diff
@@ -53,6 +53,7 @@ data:
         openShiftApiUrl: {{ include "flightctl.getOpenShiftAPIUrl" . }}
         internalOpenShiftApiUrl: {{ .Values.global.auth.internalOpenShiftApiUrl }}
         insecureSkipTlsVerify: {{ .Values.global.auth.insecureSkipTlsVerify }}
+        k8sRbacNs: {{ default .Release.Namespace .Values.global.auth.k8sRbacNs }}
```
rawagner, line 56: I know this is `auth` section is already ugly. Lets address it in some followup.

PR #738, opened 2024-12-19
EDM-390: Add decommission CLI command

lkatalin:
CLI command to interact with the new `/api/v1/devices/{name}/decommission`.

Use:
```
flightctl decommission devices/<devicename>
```


internal/cli/decommission.go
```diff
@@ -0,0 +1,125 @@
...
+	var status string
+	if response.HTTPResponse != nil {
+		status = response.HTTPResponse.Status
+		statuscode := response.HTTPResponse.StatusCode
+		if statuscode != http.StatusOK {
+			return fmt.Errorf("unsuccessful decommissioning device request %s: %s", name, string(response.Body))
+		}
+	}
+
+	log.Infof("decommission request for %s: %s\n", status, name)
```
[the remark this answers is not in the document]
lkatalin: Should they get any status output at all? For some other commands we have a `log.Infof("%s: %s\n", status, name)`. But if we don't want it I can just make it a debug.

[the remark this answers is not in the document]
fzdarsky: Commands should print out whether they succeeded or failed, of course. But that should just be printed to stdout/stderr, not logged. If you can point me to other instances where there's logging I'll try to clean that up.

In this case, I think something as simple as "Success." (or because we're async, probably "Device scheduled for decommissioning.") and "Error: Not found." would probably do.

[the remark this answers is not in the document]
lkatalin: Changed it to stdout.

api/v1alpha1/openapi.yaml
```diff
@@ -3088,14 +3088,14 @@ components:
     DeviceDecommission:
       type: object
       properties:
-        decommissionTarget:
+        target:
```
fzdarsky, line 3091: That'd also remove the double "decommission" alright, but what meant wasn't changing the property name.
If you were to define a separate type like you did [here](https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/openapi.yaml#L3937-L3948) and referenced that here, you'd be able to get clean variable and type names.

internal/cli/decommission.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		SilenceUsage: true,
+	}
+	o.Bind(cmd.Flags())
+	return cmd
+}
+
+func (o *DecommissionOptions) Bind(fs *pflag.FlagSet) {
+	o.GlobalOptions.Bind(fs)
+	fs.StringVarP(&o.DecommissionTarget, "target", "t", o.DecommissionTarget, "Specify the type of decommissioning operation: unenroll, or factoryreset")
```
hexfusion, line 56: should we remove factoryreset from docs until impl?

hexfusion: otherwise could be bug food :)

lkatalin: You're right. Just changed it.

api/v1alpha1/openapi.yaml
fzdarsky: We could probably revert to `decommissionTarget` now, though I don't have a strong preference. The name should just be intuitive.

fzdarsky (approved):

api/v1alpha1/openapi.yaml
lkatalin: It looks like this when generated:
```
const (
	DeviceDecommissionTargetTypeFactoryReset DeviceDecommissionTargetType = "FactoryReset"
	DeviceDecommissionTargetTypeUnenroll     DeviceDecommissionTargetType = "Unenroll"
)
```
which I actually think is a bit nicer than the original one.

fzdarsky: Right. But that's for developers and to avoid naming conflicts later.
Through the change you've made, both `target` and `decommissionTarget` would now produce the above code, so now we can choose what's the most intuitive.

PR #739, opened 2024-12-19
NO-ISSUE: omit grpc from make generate

hexfusion:
add `generate-proto` which will be used much less frequently and omit api/grpc from `generate`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,10 @@ publish: build-containers
 .PHONY: publish
 
 generate:
-	go generate -v $(shell go list ./...)
+	go generate -v $(shell go list ./... | grep -v -e api/grpc)
+
+generate-proto:
+	go generate -v ./api/grpc/...
 
 tidy:
 	git ls-files go.mod '**/*go.mod' -z | xargs -0 -I{} bash -xc 'cd $$(dirname {}) && go mod tidy -v'


lkatalin (approved):

hexfusion:
hold for @mangelajo just so I don't break anything related.

fzdarsky:
I had just merged the `make generate-grpc` into `make generate`, so both happen together and - more importantly - that changes to gRPC are covered by our [linting check](https://github.com/flightctl/flightctl/blob/main/.github/workflows/lint.yaml#L49).

mangelajo:
Yes, but I guess we are generating issues with the versioning? although I guess that must have been handled by the go tooling ?

fzdarsky:
We need to update the developer docs, so devs who update the .proto and regenerate know how to install the right protoc binary. Or we have that done from a Bash script called from //go:generate, too.

We can split the “make generate” again so devs do not need to install protoc if they want to update only the openapi spec but not touching the .proto. But then let’s run both from the linter action.

mangelajo:
> We need to update the developer docs, so devs who update the .proto and regenerate know how to install the right protoc binary. Or we have that done from a Bash script called from //go:generate, too.
> 
> We can split the “make generate” again so devs do not need to install protoc if they want to update only the openapi spec but not touching the .proto. But then let’s run both from the linter action.

This will be fixed when we move to buf, which uses a containerized version of protoc.

I'd just go ahead for now and split it so we don't cause any more CI issues, or developer annoyance.

mangelajo (approved):

PR #740, opened 2024-12-19
EDM-390: View lifecycle status of devices

lkatalin:
The output looks like this:

Screenshot from 2024-12-19 16-49-29

### DIFF
--- a/internal/cli/get.go
+++ b/internal/cli/get.go
@@ -392,7 +392,7 @@ func (o *GetOptions) printDevicesSummaryTable(w *tabwriter.Writer, summary *api.
 
 func (o *GetOptions) printDevicesTable(w *tabwriter.Writer, devices ...api.Device) {
 	if o.Output == wideFormat {
-		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN\tLABELS")
+		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN\tLIFECYCLE\tLABELS")
 	} else {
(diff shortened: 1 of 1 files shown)


internal/cli/get.go
```diff
@@ -392,9 +392,9 @@ func (o *GetOptions) printDevicesSummaryTable(w *tabwriter.Writer, summary *api.
 
 func (o *GetOptions) printDevicesTable(w *tabwriter.Writer, devices ...api.Device) {
 	if o.Output == wideFormat {
-		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN\tLABELS")
+		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tLIFECYCLE\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN\tLABELS")
 	} else {
-		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN")
+		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tLIFECYCLE\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN")
```
hexfusion, line 397: I think it probably only makes sense in -o wide the data is pretty static right?

lkatalin: Sure, we can do that. Will make a note to add this to the documentation.

hexfusion: cc @fzdarsky

fzdarsky: Decommissioned devices are such a rare case and then they only live in that state for short time, I wonder whether it's necessary to have that info in the printed table at all. It can always be accessed through the `-o json`.

lkatalin: We can also do it that way.

PR #744, opened 2024-12-23
EDM-953: Use approved Postgres

avishayt:
Allow setting an alternate image via variables to "make deploy".  For example:
IMAGE_PULL_SECRET_PATH="/path/to/pull/secret.yaml" SQL_IMAGE="registry.redhat.io/rhel8/postgresql-12" SQL_VERSION="1-181"

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
@@ -17,6 +17,10 @@ spec:
       labels:
         flightctl.service: flightctl-db
     spec:
+      {{- if .Values.global.imagePullSecretName }}
+      imagePullSecrets:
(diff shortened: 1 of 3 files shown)


avishayt:
We need to decide how to manage this in terms of testing.  Ideally, we'd have one upstream repo test on upstream images and a downstream repo test on Red Hat-approved images.

amir-yogev-gh:
IMO, upstream images E2E testing is a must. This should be the immediate and simplest to provision and use.
Later we can add or consider downstream images E2E testing.

avishayt:
> IMO, upstream images E2E testing is a must. This should be the immediate and simplest to provision and use. Later we can add or consider downstream images E2E testing.

(Upstream images is what we had before)

fzdarsky (approved):

PR #754, opened 2025-01-01
Edm 439: Rollout policies device selection and disruption allowance implementation

ori-amizur:
This implements the logic for both device selection and disruption allowance.
Below is a spec for both (not full):

```
apiVersion: v1alpha1
kind: Fleet
metadata:
  name: default
spec:
  selector:
    matchLabels:
      fleet: default
  rolloutPolicy:
    deviceSelection:
      strategy: 'BatchSequence'
      sequence:
        - selector:
            matchLabels:
              site: madrid
        - limit: 50%
        - selector:
            matchLabels:
              site: paris
    disruptionBudget:
      minAvailable: 1
      groupBy: ['site']
```

The device selection and disruption allowance logic are implemented by reconcilers that are invoked by the periodic server.

This PR assumes the current flow is correct for the existing logic, and attempts to make as little as possible changes to the existing functionality.
Currently when a template in a fleet spec is modified, a flow starts that is executed mainly in the tasks by the worker server which results in an unconditional device rendering for all fleet devices.  This is of course undesirable when working with batches  that each batch needs to be approved and rolled out separately.  Also it is undesirable for disruption allowance that we want to control the rollout according to some grouping defined in the spec.
So the changes to the existing flow is done by manipulating the callbacks which are the means to propagate the flow between tasks.

For device selection, the rollout flow by the **fleet_validate** task is disabled if device selection is defined in the spec.  Instead the reconciler is responsible to invoke the rollout when conditions are met.

For disruption budget, the device render flow by the **fleet_rollout** task is disabled if disruption budget is defined in the spec.  Instead the reconciler is responsible to invoke the device render when conditions are met.

## Summary by CodeRabbit

Based on the comprehensive summary of changes, here are the release notes:

## Release Notes

- **New Features**
	- Introduced advanced rollout management with batch sequence device selection
	- Added disruption budget controls for fleet updates
	- Enhanced device rendering and selection mechanisms

- **Improvements**
	- Refined device and fleet annotation handling
	- Improved rollout strategy configuration
	- Added more granular control over device update processes

- **Bug Fixes**
	- Improved error handling in device and fleet reconciliation
	- Fixed potential issues with device template version tracking

- **Performance**
	- Optimized device selection and rendering logic
	- Implemented more efficient fleet and device count management


ori-amizur:
@coderabbitai review

internal/store/store.go
```diff
@@ -134,7 +134,7 @@ type ListParams struct {
 	Labels                      map[string]string
 	LabelMatchExpressions       v1alpha1.MatchExpressions
 	AnnotationsMatchExpressions v1alpha1.MatchExpressions
-	Filter                      map[string][]string
```
asafbss, line 137: see: https://github.com/flightctl/flightctl/pull/751

internal/store/common.go
```diff
@@ -140,7 +151,8 @@ func matchInQuery(query *gorm.DB, matchExpression api.MatchExpression, colName s
 
 func matchNotInQuery(query *gorm.DB, matchExpression api.MatchExpression, colName string) *gorm.DB {
 	whereStr, args := matchExpressionQueryAndArgs(matchExpression, colName)
-	return query.Not(whereStr, args...)
+	existsQueryStr := existsQuery(colName)
```
asafbss, line 154: see: https://github.com/flightctl/flightctl/pull/751

internal/store/common.go
```diff
@@ -109,6 +112,14 @@ func (lq *listQuery) Build(ctx context.Context, db *gorm.DB, orgId uuid.UUID, li
 	return query, nil
 }
 
+func (lq *listQuery) Build(ctx context.Context, db *gorm.DB, orgId uuid.UUID, listParams ListParams) (*gorm.DB, error) {
+	query, err := lq.BuildNoOrder(ctx, db, orgId, listParams)
```
asafbss, line 116: see: https://github.com/flightctl/flightctl/pull/751
Is this necessary?

asafbss (commented):
Hey, we have a few instances, mainly in MatchExpression, where there’s overlap. MatchExpression will be part of the label-selector as discussed. Please review this section in my PR: [PR #751](https://github.com/flightctl/flightctl/pull/751).

internal/store/common.go
asafbss: This will break the pagination mechanism

ori-amizur: It is needed to use the same query for aggregate queries that don't perform group by 'name'

internal/tasks/fleet_validate.go
```diff
@@ -73,8 +74,13 @@ func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Co
...
 	}
-
-	tv, err := t.store.TemplateVersion().Create(ctx, t.resourceRef.OrgID, &templateVersion, t.callbackManager.TemplateVersionCreatedCallback)
+	var callback store.TemplateVersionStoreCallback = func(m *model.TemplateVersion) {
+		t.log.Infof("fleet %v/%s: template version %s created without rollout", t.resourceRef.OrgID, t.resourceRef.Name, m.Name)
+	}
+	if fleet.Spec.RolloutPolicy == nil || fleet.Spec.RolloutPolicy.DeviceSelection == nil {
+		callback = t.callbackManager.TemplateVersionCreatedCallback
+	}
+	tv, err := t.store.TemplateVersion().Create(ctx, t.resourceRef.OrgID, &templateVersion, callback)
```
fzdarsky, line 83: Could we envision using a single code path / always using a rollout? In the sense of that if the user does not explicitly define a rollout policy, we default to a rollout policy that selects all devices in the fleet?

ori-amizur: If I understand the question correctly, does it mean that we should change the rollout flow to the new flow and not keep the old flow?
Assuming this is correct then of course we could envision that. But I would prefer the process to be gradual.  For couple of reasons:
* This would break many of the existing tests
* Users are not used to the new flow.  We might want to get some feedback before we move to the new flow completely
* This change might affect changes that are currently being worked on.  I would like to avoid as much as possible these disturbances.
* In a meeting we had with UX, We were asked specifically to keep supporting the old flow if there is no rollout definition

internal/store/model/fleet.go
```diff
@@ -10,6 +10,19 @@ import (
...
+	FleetListKind = "FleetList"
+
+	FleetAnnotationTemplateVersion            = "fleet-controller/templateVersion"
+	FleetAnnotationDeployingTemplateVersion   = "fleet-controller/deployingTemplateVersion"
+	FleetAnnotationBatchNumber                = "fleet-controller/batchNumber"
+	FleetAnnotationRolloutApproved            = "fleet-controller/rolloutApproved"
+	FleetAnnotationRolloutApprovalMethod      = "fleet-controller/rolloutApprovalMethod"
+	FleetAnnotationLastBatchSuccessPercentage = "fleet-controller/lastBatchSuccessPercentage"
+)
+
```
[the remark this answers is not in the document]
ori-amizur: OK

api/v1alpha1/properties.go
```diff
@@ -74,7 +74,7 @@ func (d *Device) IsUpdatedToFleetSpec(f *Fleet) bool {
 	if !ok {
 		return false
 	}
-	deviceTemplateVersion, ok := (*d.Metadata.Annotations)[FleetAnnotationTemplateVersion]
+	deviceTemplateVersion, ok := (*d.Metadata.Annotations)[DeviceAnnotationRenderedTemplateVersion]
```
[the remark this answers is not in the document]
ori-amizur: The changes in this PR are according the current state of the code.  If there are other plans to refactor the code, namely move rendering to the API server, we will have to adapt then.

To check if a device is updated to fleet spec then we check that the TV in the fleet is the same as TV in the device and RV and the annotation **DeviceAnnotationRenderedVersion** is the same as the TV in the status.
This is generally OK, but we have to keep in mind that the **FleetAnnotationTemplateVersion** is set in the fleet-rollout task, but the RV is set in the device render task in device render task (which runs right after fleet rollout task).
If the device rendering is delayed as result of disruption allowance, then there might be a long time that fleet TV is the same as the one in the annotation **FleetAnnotationTemplateVersion** of the device, and the rendered versions are the same but the device is still out of date.
One of the changes of this PR, an annotation was added **DeviceAnnotationRenderedTemplateVersion** which is set when a new rendered version is created (this annotations has other uses as well).  This annotation also contains the TV.  Since it is also set in the device render task, the test to find of a device is out of date is correct.

ori-amizur:
@coderabbitai review

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,456 @@
...
+	if err != nil {
+		return err
+	}
+	if err = selection.unmark(ctx); err != nil {
+		return err
+	}
+	nextBatch := currentBatch + 1
+	if nextBatch > len(lo.FromPtr(b.Sequence)) {
+		return fmt.Errorf("batch number overflow")
+	}
```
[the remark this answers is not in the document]
ori-amizur: This is the intended behavior since I wanted the batch number to increase passed the last batch.  The function currentSelection in batchSequenceSelector ensures that only valid batches are accessed in the sequence

internal/rollout/disruption_allowance/reconciler.go
```diff
@@ -0,0 +1,193 @@
...
+	}, groupBy, differentRenderedVersion, onlineOnly)
+	if err != nil {
+		return nil, err
+	}
+	ret, err := mergeDeviceAllowanceCounts(totalCounts, busyCounts, groupBy)
+	if err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
```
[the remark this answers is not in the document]
ori-amizur: The whole logic is retried in case of error since it is called from a reconciler that is called every 30 seconds.

api/v1alpha1/util.go
```diff
@@ -127,6 +128,18 @@ func getApplicationType(union json.RawMessage) (ApplicationProviderType, error)
...
+	index := strings.Index(p, "%")
+	if index <= 0 {
+		return 0, fmt.Errorf("%s is not in percentage format", p)
+	}
+	percentage, err := strconv.ParseInt(p[:index], 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return int(percentage), nil
+}
```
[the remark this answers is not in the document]
ori-amizur: Noted, although percentage validation is defined by openapi regular expression

internal/store/fleet.go
```diff
@@ -124,6 +126,51 @@ func fleetSelectStr(withDeviceCount bool) string {
...
+                 (select 1 from devices where org_id = ? and owner = CONCAT('%s/', main_query.name) and
+					main_query.tv = annotations ->> '%s' and
+                    main_query.tv <> COALESCE(annotations ->> '%s', ''))`,
+		api.FleetAnnotationTemplateVersion,
+		api.FleetKind, api.DeviceAnnotationTemplateVersion, api.DeviceAnnotationRenderedTemplateVersion), orgId, orgId).Scan(&fleets).Error
+	if err != nil {
+		return nil, ErrorFromGormError(err)
+	}
+	return lo.ToPtr(fleets.ToApiResource(nil, nil)), nil
+}
```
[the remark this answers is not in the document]
ori-amizur: Why is it a potential for SQL injection? How can it happen here ?

api/v1alpha1/openapi.yaml
```diff
@@ -4490,7 +4490,11 @@ components:
     BatchSequence:
       type: object
       description: BatchSequence defines the list of batches to be executed in sequence.
+      required:
+        - strategy
       properties:
+        strategy:
+          type: string
```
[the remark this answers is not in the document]
ori-amizur: Simultaneous use of minAvailable and maxUnavailable is legal.

internal/store/fleet.go
```diff
@@ -124,6 +126,51 @@ func fleetSelectStr(withDeviceCount bool) string {
...
+                                 main_query.tv <> COALESCE(annotations ->> '%s', '') or
+                           		 status -> 'config' ->> 'renderedVersion' <> COALESCE(annotations ->> '%s', '')) and
+                           		 org_id = ? and owner = CONCAT('%s/', main_query.name) limit 1)`,
+		api.FleetAnnotationTemplateVersion, api.DeviceAnnotationSelectedForRollout, api.DeviceAnnotationRenderedTemplateVersion, api.DeviceAnnotationRenderedVersion,
+		api.FleetKind), orgId, gorm.Expr("?"), orgId).Scan(&fleets).Error
+	if err != nil {
+		return nil, ErrorFromGormError(err)
+	}
+	return lo.ToPtr(fleets.ToApiResource(nil, nil)), nil
+}
```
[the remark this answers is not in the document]
ori-amizur: Why is it a potential for SQL injection? How can it happen here ?

internal/store/device.go
```diff
@@ -209,6 +214,110 @@ func (s *DeviceStore) List(ctx context.Context, orgId uuid.UUID, listParams List
...
+	if err != nil {
+		return nil, ErrorFromGormError(err)
+	}
+	ret := lo.Map(results, func(m map[string]any, _ int) map[string]any {
+		return lo.SliceToMap(append(groupBy, "count"), func(s string) (string, any) {
+			return s, m[labelKeyToSymbol(s)]
+		})
+	})
+	return ret, nil
+}
```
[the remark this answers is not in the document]
ori-amizur: All the labels are enclosed in quotes.  How can SQL injection happen in this case ?

internal/rollout/disruption_allowance/reconciler.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		_, exists := annotations[api.FleetAnnotationTemplateVersion]
+		if !exists {
+			continue
+		}
+		if err := r.reconcileFleet(ctx, orgId, fleet); err != nil {
+			r.log.WithError(err).Errorf("reconcile fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
+		}
+	}
+}
```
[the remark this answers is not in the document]
ori-amizur: The Reconcile function is invoked in the periodic_checker in a thread object.  The thread object already has support for graceful shutdowns.  Isn't it enough ?

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,456 @@
...
+	return c.RenderedTemplateVersion == b.templateVersionName && c.SameRenderedVersion
+}
+
+// IsComplete checks is the total number of devices in a batch is the same as the number of completed
+func (b *batchSelection) IsComplete(ctx context.Context) (bool, error) {
+	if b.batch == nil {
+		return true, nil
+	}
+	// TODO: Add timeout check
+	counts, err := b.store.Device().CompletionCounts(ctx, b.orgId, fmt.Sprintf("%s/%s", v1alpha1.FleetKind, b.fleetName))
```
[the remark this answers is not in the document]
ori-amizur: Yes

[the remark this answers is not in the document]
ori-amizur: The timeout has to rely on the time of devices being updated, not of the batch.  The reason for that is that a batch might not be approved for a long time, and devices might not start to perform update since they are stopped by disruption allowance mechanism.

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,461 @@
...
+	if err != nil {
+		return err
+	}
+	if err = selection.unmark(ctx); err != nil {
+		return err
+	}
+	nextBatch := currentBatch + 1
+	if nextBatch > len(lo.FromPtr(b.Sequence)) {
+		return fmt.Errorf("batch number overflow")
+	}
```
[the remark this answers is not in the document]
ori-amizur: It allows to exceed only by 1

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,461 @@
...
+	total := lo.Sum(lo.Map(counts, func(c store.CompletionCount, _ int) int64 {
+		return lo.Ternary(b.isUpdateCompletedSuccessfully(c) || c.SummaryStatus != Unknown || isInUpdate(c.UpdatingReason), c.Count, 0)
+	}))
+
+	// A device is counted as completed if it has completed successfully or, it is not disconnected, and it is in error state
+	complete := lo.Sum(lo.Map(counts, func(c store.CompletionCount, _ int) int64 {
+		return lo.Ternary(b.isUpdateCompletedSuccessfully(c) || c.SummaryStatus != Unknown && (c.UpdatingReason == Error || c.UpdateTimedOut && c.RenderedTemplateVersion == b.templateVersionName), c.Count, 0)
+	}))
+	return total == complete, nil
+}
```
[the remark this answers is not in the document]
ori-amizur: The UpdateTimedOut field in CompletionCount is calculated by taking into account the individual device timeouts

internal/store/device.go
```diff
@@ -24,6 +25,11 @@ type Device interface {
 	Create(ctx context.Context, orgId uuid.UUID, device *api.Device, callback DeviceStoreCallback) (*api.Device, error)
 	Update(ctx context.Context, orgId uuid.UUID, device *api.Device, fieldsToUnset []string, fromAPI bool, callback DeviceStoreCallback) (*api.Device, error)
 	List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*api.DeviceList, error)
+	Count(ctx context.Context, orgId uuid.UUID, listParams ListParams) (int64, error)
+	UnmarkRolloutSelection(ctx context.Context, orgId uuid.UUID, fleetName string) error
+	MarkRolloutSelection(ctx context.Context, orgId uuid.UUID, listParams ListParams, limit *int) error
+	CompletionCounts(ctx context.Context, orgId uuid.UUID, owner string, templateVersion string, updateTimeout *time.Duration) ([]CompletionCount, error)
+	CountByLabels(ctx context.Context, orgId uuid.UUID, listParams ListParams, groupBy []string, busyOnly bool) ([]map[string]any, error)
```
avishayt, line 32: As part of [EDM-1025](https://issues.redhat.com/browse/EDM-1025), we want tasks to use the service handler methods rather than access the store directly.  Can we use existing ReplaceDevice, ListDevices, etc. for these?  What's missing?

internal/store/fleet.go
```diff
@@ -23,6 +23,8 @@ type Fleet interface {
 	Create(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet, callback FleetStoreCallback) (*api.Fleet, error)
 	Update(ctx context.Context, orgId uuid.UUID, fleet *api.Fleet, callback FleetStoreCallback) (*api.Fleet, error)
 	List(ctx context.Context, orgId uuid.UUID, listParams ListParams, opts ...ListOption) (*api.FleetList, error)
+	ListRolloutDeviceSelection(ctx context.Context, orgId uuid.UUID) (*api.FleetList, error)
+	ListDisruptionAllowanceFleets(ctx context.Context, orgId uuid.UUID) (*api.FleetList, error)
```
avishayt, line 27: Same question as above about these - what's missing from ListFleets?

internal/tasks/callback_manager.go
```diff
@@ -186,8 +184,13 @@ func (t *callbackManager) DeviceUpdatedCallback(before *model.Device, after *mod
...
 		t.submitTask(FleetSelectorMatchTask, ref, op)
 	}
-	if specUpdated {
-		t.submitTask(DeviceRenderTask, ref, DeviceRenderOpUpdate)
+
+}
+
+func (t *callbackManager) DeviceUpdatedCallback(before *model.Device, after *model.Device) {
+	t.DeviceUpdatedNoRenderCallback(before, after)
+	if after != nil && (before == nil || !reflect.DeepEqual(before.Spec, after.Spec)) {
```
avishayt, line 192: Comparing device specs with reflect doesn't work - use api.DeviceSpecsAreEqual()

avishayt (commented):
First pass, still going through it

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,462 @@
...
+)
+
+const (
+	DefaultSuccessThreshold = 90
+	Unknown                 = "Unknown"
+	ApplyingUpdate          = "ApplyingUpdate"
+	Rebooting               = "Rebooting"
+	ActivatingConfig        = "ActivatingConfig"
+	RollingBack             = "RollingBack"
+	Error                   = "Error"
```
avishayt, line 28: Why redefine these?  They are in api/v1alpha1/consts.go

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,462 @@
...
+	store                store.Store
+	orgId                uuid.UUID
+	fleet                *api.Fleet
+	fleetName            string
+	templateVersionName  string
+	defaultUpdateTimeout *time.Duration
+	log                  logrus.FieldLogger
+}
+
+func (b *batchSequenceSelector) getAnnotation(annotation string) (string, bool) {
```
avishayt, line 55: Why not move to a util file, maybe api/v1alpha1/util.go?

api/v1alpha1/consts.go
```diff
@@ -11,9 +11,11 @@ const (
...
 	DeviceListKind   = "DeviceList"
 
-	DeviceAnnotationConsole         = "device-controller/console"
-	DeviceAnnotationRenderedVersion = "device-controller/renderedVersion"
-	DeviceAnnotationTemplateVersion = "fleet-controller/templateVersion"
+	DeviceAnnotationConsole                 = "device-controller/console"
+	DeviceAnnotationRenderedVersion         = "device-controller/renderedVersion"
+	DeviceAnnotationTemplateVersion         = "fleet-controller/templateVersion"
+	DeviceAnnotationRenderedTemplateVersion = "device-controller/renderedTemplateVersion"
+	DeviceAnnotationSelectedForRollout      = "device-controller/selectedForRollout"
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/store/device.go
ori-amizur: Assuming we need it for large scale, it is preferable to work with small data sets and offload computations (i.e count) to the database.  So these operations do exactly that, so we can avoid bringing large sets of devices just for counting in order to make decisions what needs to be done.

internal/store/fleet.go
ori-amizur: These use capabilities that are missing in ListFleets:
The first tries to bring fleets that have devices that have not been rolled out (where not been processed by the fleet-rollout task).
The second tries to bring fleets that have devices that have not been rendered yet.
In general,  ListFleets cannot use a filter which is based on the underlying device population.

internal/rollout/device_selection/batch_sequence_selector.go
ori-amizur: Will move

internal/tasks/callback_manager.go
ori-amizur: OK

internal/rollout/device_selection/batch_sequence_selector.go
ori-amizur: OK

internal/tasks/fleet_validate.go
avishayt: The UX will remain the same, but all the code will go through the same rollout logic, as if there was a single batch and infinite disruption allowance.

celdrake:
@ori-amizur There seems to be some issue with the current definition of `RolloutDeviceSelection`. The openapi-generator for Typescript generates it as
```
export type RolloutDeviceSelection = (BatchSequence | {
  strategy: string;
});

```

meaning either `sequence` OR `strategy` need to be populated, which is not what's in your example. 
Seems like it's the openapi definition what's not correct.

ori-amizur:
> Edit: As pointed by @ori-amizur , I was checking the types as generated from `main`. From this branch, the type has been fixed even though I find it a bit odd that the `strategy` field is in the `BatchSequence` object instead on the parent `RolloutDeviceSelection` object.

In order to use discriminator by golang, the discriminator field (i.e `strategy`), must be part of the `BatchSequence` object

internal/rollout/disruption_allowance/disruption_allowance.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/samber/lo"
+)
+
+const keySeparator = "@#$"
+
+type Counts struct {
+	TotalCount int
+	BusyCount  int
+	key        map[string]any
```
hexfusion, line 16: just curious why the `any` type here vs hard type.

internal/rollout/disruption_allowance/disruption_allowance.go
```diff
@@ -0,0 +1,109 @@
...
+	setter(counts, count)
+	return nil
+}
+
+func (g *GroupMap) InsertTotal(row map[string]any) error {
+	return g.insert(row, func(counts *Counts, count int) { counts.TotalCount = count })
+}
+
+func (g *GroupMap) InsertBusy(row map[string]any) error {
+	return g.insert(row, func(counts *Counts, count int) { counts.BusyCount = count })
```
hexfusion, line 93: question: can counts be nil?

internal/rollout/disruption_allowance/reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+		FieldSelector: selector.NewFieldSelectorFromMapOrDie(map[string]string{"metadata.owner": util.ResourceOwner(api.FleetKind, lo.FromPtr(fleet.Metadata.Name))}),
+	}, groupBy, true)
+	if err != nil {
+		return nil, err
+	}
+	ret, err := mergeDeviceAllowanceCounts(totalCounts, busyCounts, groupBy)
+	if err != nil {
+		return nil, err
+	}
+	return ret, nil
```
hexfusion, line 64: minor: since you dont add any additional ctx to error could 

```
return mergeDeviceAllowanceCounts(totalCounts, busyCounts, groupBy)
```

internal/rollout/disruption_allowance/reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	templateVersionName, exists := annotations[api.FleetAnnotationTemplateVersion]
+	if !exists {
+		return fmt.Errorf("template version doesn't exist")
+	}
+	listParams := store.ListParams{
+		Limit: numToRender,
+
+		// The list of labels is converted to MatchExpressions.  In case that the label does not exist
+		// (nil value), then the query requests explicitly that the label should not exist.
```
hexfusion, line 80: :+1:

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,520 @@
...
+		return 0, false, err
+	}
+	return int(ret), true, nil
+}
+
+// A batch may be approved atotmatically only if its approval method is "automatic" and the
+// success percentage of the previous batch is greater or equal to the success threshold
+func (b *batchSelection) MayApproveAutomatically() (bool, error) {
+	approvalMethod, exists := b.getAnnotation(api.FleetAnnotationRolloutApprovalMethod)
+	if !exists || approvalMethod != "automatic" {
```
hexfusion, line 368: if this doesn't exist I would expect approvalMethod to always be empty.  if its not might be something to look at.

internal/rollout/device_selection/batch_sequence_selector.go
```diff
@@ -0,0 +1,520 @@
...
+	rolledOut, err := b.store.Device().Count(ctx, b.orgId, parts.withRolledOut(b.templateVersionName).listParams())
+	if err != nil {
+		return 0, 0, err
+	}
+	return int(rolledOut), int(total), nil
+}
+
+func (b *batchSelection) calculateLimit(ctx context.Context) (*int, error) {
+	if b.batch == nil || b.batch.Limit == nil {
+		return nil, nil
```
hexfusion, line 457: nit: if I am a consumer of this method the general expectation is I should not need to nil check the returned value if the error is non nil. In other words it would be unexpected to panic using the value returned here.

Maybe return (int, error)?

internal/rollout/disruption_allowance/reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+	defer r.log.Infof("disruption allowance: finished reconciling fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
+
+	maxUnavailable := fleet.Spec.RolloutPolicy.DisruptionAllowance.MaxUnavailable
+	minAvailable := fleet.Spec.RolloutPolicy.DisruptionAllowance.MinAvailable
+	if maxUnavailable == nil && minAvailable == nil {
+		return fmt.Errorf("both maxUnavailable and minAvailable for fleet %s are nil", lo.FromPtr(fleet.Metadata.Name))
+	}
+	m, err := r.getFleetCounts(ctx, orgId, fleet)
+	if err != nil {
+		return fmt.Errorf("getFleetCounts: %v", err)
```
hexfusion, line 135:
```suggestion
		return fmt.Errorf("getFleetCounts: %w", err)
```

internal/store/fleet.go
```diff
@@ -124,6 +126,54 @@ func fleetSelectStr(withDeviceCount bool) string {
...
 }
 
+// ListRolloutDeviceSelection attempts to get all relevant fleets for rollout device selection.
+// A relevant fleet contains at least 1 device that at least one of the conditions below is true:
+// - marked as selected for rollout
+// - the template version of the fleet is not the same the template version in the annotation 'device-controller/renderedTemplateVersion'
+// - the field 'status.config.renderedVersion' is not the same as the annotation 'device-controller/renderedVersion'
+func (s *FleetStore) ListRolloutDeviceSelection(ctx context.Context, orgId uuid.UUID) (*api.FleetList, error) {
+	var fleets model.FleetList
+	err := s.db.Raw(fmt.Sprintf(`select * from (select *, annotations ->> '%s' as tv from fleets) as main_query
```
hexfusion, line 136: I know you have already considered this but I am just passing along the article from go team.

https://go.dev/doc/database/sql-injection

internal/rollout/device_selection/batch_sequence_selector.go
ori-amizur: In case a limit doesn't exist then nil is returned.  It means that limit should not be applied.

internal/rollout/device_selection/device_selection.go
```diff
@@ -0,0 +1,65 @@
...
+
+func getUpdateTimeout(defaultUpdateTimeoutStr *api.Duration) (time.Duration, error) {
+	timeout := DefaultUpdateTimeout
+
+	if defaultUpdateTimeoutStr != nil {
+		d, err := time.ParseDuration(*defaultUpdateTimeoutStr)
+		if err != nil {
+			return 0, fmt.Errorf("failed to parse duration %s: %w", *defaultUpdateTimeoutStr, err)
+		}
+		if d != 0 {
```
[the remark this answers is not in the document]
ori-amizur: 0 means it will fail on timeout immediately

api/v1alpha1/util.go
```diff
@@ -127,6 +128,21 @@ func getApplicationType(union json.RawMessage) (ApplicationProviderType, error)
...
 }
 
+func PercentageAsInt(p Percentage) (int, error) {
+	index := strings.Index(p, "%")
+	if index <= 0 || index != len(p)-1 {
+		return 0, fmt.Errorf("%s is not in percentage format", p)
+	}
+	percentage, err := strconv.ParseInt(p[:index], 10, 64)
+	if err != nil {
+		return 0, err
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/periodic_checker/server.go
```diff
@@ -69,6 +71,23 @@ func (s *Server) Run() error {
 	deviceDisconnectedThread.Start()
 	defer deviceDisconnectedThread.Stop()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Rollout device selection
+	rolloutDeviceSelection := device_selection.NewReconciler(s.store, callbackManager, s.log)
+	rolloutDeviceSelectionThread := thread.New(
+		s.log.WithField("pkg", "rollout-device-selection"), "Rollout device selection", device_selection.RolloutDeviceSelectionInterval, func() { rolloutDeviceSelection.Reconcile(ctx) })
```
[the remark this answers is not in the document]
ori-amizur: I don't understand the suggestion.  According to the suggested code the ctx is removed.

internal/rollout/device_selection/batch_sequence_selector.go
ori-amizur: OK

internal/rollout/device_selection/reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+		callbackManager: callbackManager,
+	}
+}
+
+func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet api.Fleet) {
+	r.log.Infof("device selection: starting reconciling fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
+	defer r.log.Infof("device selection: finished reconciling fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
+
+	if fleet.Spec.RolloutPolicy == nil || fleet.Spec.RolloutPolicy.DeviceSelection == nil {
+		return
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/rollout/device_selection/reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	annotations := lo.FromPtr(fleet.Metadata.Annotations)
```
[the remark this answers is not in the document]
ori-amizur: I think it is better to use the publicly available function than have and use it in util, since we are using this package for many other purposes.
My opinion is to remove from util

internal/rollout/device_selection/reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+	if annotations == nil {
+		r.log.Infof("no annotations for fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
+		return
+	}
+	templateVersionName, exists := annotations[api.FleetAnnotationTemplateVersion]
+	if !exists {
+		r.log.Infof("no template version for fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/rollout/device_selection/reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+		if !isRolledOut {
+			if !selection.IsApproved() {
+
+				// A batch may be approved either by a user or automatically
+				mayApprove, err := selection.MayApproveAutomatically()
+				if err != nil {
+					r.log.WithError(err).Errorf("%v/%s: MayApproveAutomatically", orgId, lo.FromPtr(fleet.Metadata.Name))
+					break
+				}
+				if mayApprove {
```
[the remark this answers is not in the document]
ori-amizur: It seems OK, since on line 91 it is actually approved.

internal/rollout/disruption_allowance/disruption_allowance.go
ori-amizur: Counts are created by `getOrCreate` so it verifies that it will not be nil.  Look at `insert` function.

internal/rollout/disruption_allowance/reconciler.go
ori-amizur: OK

internal/store/fleet.go
ori-amizur: Just note that all parameters passed to the `Sprintf` function are constants.

hexfusion: thanks

internal/rollout/disruption_allowance/reconciler.go
ori-amizur: OK

avishayt (approved):

PR #743, opened 2024-12-20
EDM-390: Add decommission request to device spec

lkatalin:
This includes the DecommissionRequest as a new field in the Device Spec and makes updates to service logic so that this field can't be updated in regular requests. There will also be logic in the `lifecycle` (later PR) to not allow a `Decommissioned` setting to be changed back to `Enrolled`.

## Release Notes

- **New Features**
	- Added support for device decommissioning requests.
	- Introduced validation for decommissioned device operations.
	- Enhanced API to manage device lifecycle events.

- **Bug Fixes**
	- Prevented creation, replacement, and updates of decommissioned devices.
	- Added error handling for decommissioning-related scenarios.

- **API Changes**
	- Added `decommissionRequested` property to device specifications.
	- Introduced new error handling for device decommissioning attempts.

PR #757, opened 2025-01-02
EDM-390: Set decommission request in device spec

lkatalin:
This sets the `DeviceSpec.DecommissionRequested` field during a decommission request.

~This builds on #743 ; the only relevant commit for this PR is `cba2d8b4809e68fc327b354f8b60a1a5d5804b5e`.~

## Release Notes

- **New Features**
  - Added device decommissioning functionality.
  - Introduced ability to track device lifecycle events.
  
- **Improvements**
  - Enhanced device management with new validation checks.
  - Added safeguards to prevent updates to decommissioned devices.

- **Bug Fixes**
  - Implemented controls to prevent creating or modifying decommissioned devices.

- **API Changes**
  - New `decommissioning` property added to device specifications.


api/v1alpha1/validation.go
```diff
@@ -68,6 +69,11 @@ func (r DeviceSpec) Validate(fleetTemplate bool) []error {
 			allErrs = append(allErrs, validation.ValidateString(&matchPattern, fmt.Sprintf("spec.systemd.matchPatterns[%d]", i), 1, 256, nil, "")...)
 		}
 	}
+
+	if r.DecommissionRequested != nil {
+		allErrs = append(allErrs, flterrors.ErrDecommission)
+	}
```
avishayt, line 75: These validations are stateless - they're just checking to ensure that the device struct is valid.  Validation should only be in the action.

internal/store/device.go
```diff
@@ -36,6 +36,7 @@ type Device interface {
 	UpdateRendered(ctx context.Context, orgId uuid.UUID, name, renderedConfig, renderedApplications string) error
 	GetRendered(ctx context.Context, orgId uuid.UUID, name string, knownRenderedVersion *string, consoleGrpcEndpoint string) (*api.RenderedDeviceSpec, error)
 	SetServiceConditions(ctx context.Context, orgId uuid.UUID, name string, conditions []api.Condition) error
+	SetDecommission(ctx context.Context, orgId uuid.UUID, name string, request *api.DeviceDecommission) error
```
avishayt, line 39: I'm currently standardizing the store API across resources and reducing the number of methods.  Can we use Update() instead, perhaps by passing fromAPI=false and thus avoid the checks that you added there?

internal/store/device.go
```diff
@@ -629,6 +638,35 @@ func (s *DeviceStore) SetServiceConditions(ctx context.Context, orgId uuid.UUID,
...
+
+	if existingRecord.Spec == nil {
+		existingRecord.Spec = model.MakeJSONField(api.DeviceSpec{})
+	}
+	if existingRecord.Spec.Data.DecommissionRequested != nil {
+		return fmt.Errorf("decommission request on already decommissioning device")
+	}
+	existingRecord.Spec.Data.DecommissionRequested = request
+
+	result = s.db.Model(existingRecord).Where("resource_version = ?", lo.FromPtr(existingRecord.ResourceVersion)).Updates(map[string]interface{}{
```
avishayt, line 656: This is missing retries on resourceVersion conflict.  In a comment above I suggested using Update(), which already has this.

internal/store/model/device.go
```diff
@@ -53,6 +53,9 @@ func NewDeviceFromApiResource(resource *api.Device) (*Device, error) {
 
 	spec := api.DeviceSpec{}
 	if resource.Spec != nil {
+		if resource.Spec.DecommissionRequested != nil {
+			return nil, flterrors.ErrDecommission
+		}
```
avishayt, line 58: Why is this necessary?  I think it would return an error whenever doing anything to any decommissioned device.

internal/service/device.go
```diff
@@ -418,5 +432,12 @@ func (h *ServiceHandler) DecommissionDevice(ctx context.Context, request server.
...
 	}
-	return nil, fmt.Errorf("not yet implemented")
+
+	orgId := store.NullOrgId
+	err = h.store.Device().SetDecommission(ctx, orgId, request.Name, request.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return server.DecommissionDevice200JSONResponse{}, nil
```
DakCrowder, line 442: If we aren't returning the latest representation of the device, could it make sense to return a `Status` object indicating a successful result rather than an empty `Device`?  From looking at the openapi spec I believe the `Status` is currently only returned for some of the delete operations so not sure if this fits the original intended usecase.

internal/store/device.go
lkatalin: Okay, let me take a look at doing it that way.

internal/store/model/device.go
lkatalin: Yes, that is the intended result as no changes should be made to decommissioned devices.

api/v1alpha1/validation.go
lkatalin: I'm trying to keep the user from passing in any DeviceSpec (through any method) that has a DecommissionRequested set, as that should never happen (decommissioned devices cannot be updated at all, and devices cannot be decommissioned via a general update of the device spec). Does this check not do this?

internal/store/model/device.go
lkatalin: I'm actually removing this check, as it no longer works if we use `Update()`.

internal/store/device.go
lkatalin: Updated - no pun intended :)

internal/service/device.go
lkatalin: I've revamped the code to use Update() so we're returning a non-empty result now.

internal/service/device.go
```diff
@@ -379,6 +390,9 @@ func (h *ServiceHandler) PatchDevice(ctx context.Context, request server.PatchDe
 	if !reflect.DeepEqual(currentObj.Status, newObj.Status) {
 		return server.PatchDevice400JSONResponse{Message: "status is immutable"}, nil
 	}
+	if newObj.Spec != nil && newObj.Spec.DecommissionRequested != nil {
```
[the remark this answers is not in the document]
lkatalin: done

api/v1alpha1/validation.go
DakCrowder: I wonder if what @avishayt is indicating (please correct if not!) is that structurally a spec having DecommissionRequested set can be a valid state.  For example, if I were to fetch an existing spec from the database with `DecommissionRequested` set (maybe in a context like a task where there may be no user)  is that spec then invalid?  Maybe not as structurally the spec conforms to our standards.

lkatalin: Ah okay, I was thinking these validations are only used for user input. If that's not the case then I agree, I'll remove it.

avishayt:
It looks good to me. The behavior, as implemented, allows decommissioning a device that's part of a fleet.  I'm OK with that, but I want to make sure it's intentional.

lkatalin:
> The behavior, as implemented, allows decommissioning a device that's part of a fleet

Another tiny PR coming in will remove the device from its fleet (by removing metadata.Owner) in the handling of the decommissioning request. Not sure if there should be other explicit checks in addition. I don't think we are requiring manually removing a device from a fleet before decommissioning, but decommissioning itself will remove the device from the fleet.

DakCrowder (approved):

avishayt (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -4620,6 +4620,8 @@ components:
           description: 'Array of resource monitor configurations.'
           items:
             $ref: '#/components/schemas/ResourceMonitor'
+        decommissionRequested:
```
fzdarsky, line 4623: It'd have to be "decommision**ing**Requested" (noun, not verb), or?
However, the user's intent would be to have the device decommissioned/wiped/..., not to request something. That intent is in the "target" of the decom schema, so maybe a grouping noun like "decommissioning" would be more appropriate?

internal/flterrors/flterrors.go
```diff
@@ -27,6 +27,7 @@ var (
 	ErrTemplateVersionIsNil   = errors.New("spec.templateVersion not set")
 	ErrInvalidTemplateVersion = errors.New("device's templateVersion is not valid")
 	ErrNoRenderedVersion      = errors.New("no rendered version for device")
+	ErrDecommission           = errors.New("decommissioned device cannot be created or updated; to request decommissioning, use /api/v1/devices/{name}/decommission")
```
fzdarsky, line 30: I'd remove the second part: We don't have instructions how to solve an error anywhere else, either. Actually also: When the error is printed, the device is already decommissioned, so the instructions aren't even relevant anymore.

internal/service/device.go
```diff
@@ -393,5 +407,33 @@ func (h *ServiceHandler) DecommissionDevice(ctx context.Context, request server.
...
+		Metadata: v1alpha1.ObjectMeta{
+			Name: &request.Name,
+		},
+		Spec: &v1alpha1.DeviceSpec{
+			DecommissionRequested: request.Body,
+		},
+	}
+
+	// set the fromAPI bool to 'false', otherwise updating the spec.decommissionRequested of a device is blocked
+	result, err := h.store.Device().Update(ctx, orgId, newObj, nil, false, updateCallback)
```
fzdarsky, line 427: Doesn't this erase all other `.spec` fields instead of just adding the `decommissionRequested` field?

lkatalin: True; fixed.

fzdarsky (approved):

PR #764, opened 2025-01-07
EDM-526: policy: ensure DST is properly covered

hexfusion:
This PR is a follow-up to #737 and supports daylight savings time corner cases.

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging capabilities for schedule parsing
	- Improved daylight saving time (DST) handling during time normalization

- **Documentation**
	- Added clarification comment about thread safety for device manager

- **Chores**
	- Updated internal tracking of template version in queue management
	- Modified method signatures to support enhanced logging

### DIFF
--- a/internal/agent/device/policy/manager.go
+++ b/internal/agent/device/policy/manager.go
@@ -18,6 +18,9 @@ type manager struct {
 	log *log.PrefixLogger
 }
 
+// NewManager returns a new device policy manager.
+// Note: This manager is designed for sequential operations only and is not
+// thread-safe.
 func NewManager(log *log.PrefixLogger) Manager {
 	return &manager{
 		log: log,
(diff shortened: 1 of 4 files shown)


hexfusion:
@coderabbitai review

internal/agent/device/policy/manager.go
```diff
@@ -155,3 +161,25 @@ func (s *schedule) IsReady(log *log.PrefixLogger) bool {
...
+	}
+
+	localTime := t.In(s.location)
+	if localTime.IsDST() {
+		// dst adjusted ok to return
+		return localTime
+	}
+
+	// ensure fall DST is handled
+	previousHour := localTime.Add(-time.Hour)
```
DakCrowder, line 178: This is a (very) extreme edge case - but DST transitions may not always be an hour.  There is an [island](https://en.wikipedia.org/wiki/Lord_Howe_Island) with a 30 minute transition.  Not sure if we care enough to change the implementation, but its certainly a good fun fact.

DakCrowder (approved):

internal/agent/device/policy/manager.go
hexfusion: is that the only case? curious if there are more nonhour cases.

hexfusion: I will research and followup if it seems viable, thanks for the info!

PR #765, opened 2025-01-08
EDM-390: Update device lifecycle status based on Condition

lkatalin:
When the device sets a `DeviceDecommission` `Condition`, the service side will update the `LifecycleStatus` to `Decommissioning`.

## Summary by CodeRabbit

- **New Features**
	- Added a method to check if a device is in the process of decommissioning.
	- Enhanced device lifecycle status management to handle decommissioning state.

- **Improvements**
	- Introduced explicit handling for devices undergoing decommissioning.
	- Added informative status updates for decommissioning devices.

### DIFF
--- a/api/v1alpha1/properties.go
+++ b/api/v1alpha1/properties.go
@@ -80,3 +80,16 @@ func (d *Device) IsUpdatedToFleetSpec(f *Fleet) bool {
 	}
 	return d.IsUpdatedToDeviceSpec() && deviceTemplateVersion == fleetTemplateVersion
 }
+
+// IsDecommissioning() is true if the device has added a DeviceDecommissioning ConditionType to its Conditions.
+func (d *Device) IsDecommissioning() bool {
+	if d.Status == nil || d.Status.Conditions == nil {
+		return false
+	}
+
+	decommissioningCondition := FindStatusCondition(d.Status.Conditions, DeviceDecommissioning)
+	if decommissioningCondition == nil {
+		return false
+	}
+	return decommissioningCondition.Status == ConditionStatusTrue
+}
--- a/internal/service/common/device.go
+++ b/internal/service/common/device.go
@@ -87,7 +87,8 @@ func UpdateServiceSideStatus(ctx context.Context, st store.Store, log logrus.Fie
 	deviceStatusChanged := updateServerSideDeviceStatus(device)
 	updatedStatusChanged := updateServerSideDeviceUpdatedStatus(ctx, st, log, orgId, device)
 	applicationStatusChanged := updateServerSideApplicationStatus(device)
-	return deviceStatusChanged || updatedStatusChanged || applicationStatusChanged
+	lifecycleStatusChanged := updateServerSideLifecycleStatus(device)
+	return deviceStatusChanged || updatedStatusChanged || applicationStatusChanged || lifecycleStatusChanged
 }
 
 func updateServerSideDeviceStatus(device *api.Device) bool {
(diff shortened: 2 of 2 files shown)


lkatalin:
The CodeRabbit has a reasonable point - if we want to track the lifecyclestatus changes the way we are tracking the other device status changes, I can add these checks.

internal/service/common/device.go
```diff
@@ -92,6 +92,13 @@ func UpdateServiceSideStatus(ctx context.Context, st store.Store, log logrus.Fie
 
 func updateServerSideDeviceStatus(device *api.Device) bool {
 	lastDeviceStatus := device.Status.Summary.Status
+	if device.IsDecommissioning() {
+		device.Status.Lifecycle = api.DeviceLifecycleStatus{
+			Info:   util.StrToPtr("Device has acknowledged decommissioning request"),
+			Status: api.DeviceLifecycleStatusDecommissioning,
+		}
+		// todo - not sure if we can just return here or we need to go through other updates
```
hexfusion, line 100: as per conv we should report/update the status until  we deem the device "dead,gone, etc"

internal/service/common/device.go
```diff
@@ -138,6 +139,19 @@ func updateServerSideDeviceStatus(device *api.Device) bool {
...
+	lastLifecycleStatus := device.Status.Lifecycle.Status
+
+	if device.IsDecommissioning() {
+		device.Status.Lifecycle = api.DeviceLifecycleStatus{
+			Info:   util.StrToPtr("Device has acknowledged decommissioning request"),
+			Status: api.DeviceLifecycleStatusDecommissioning,
+		}
+	}
+
+	return device.Status.Lifecycle.Status != lastLifecycleStatus
```
hexfusion, line 152: If we are not decommissioning, isn't this always false? This logic works, so this is just an observation.

api/v1alpha1/properties.go
```diff
@@ -80,3 +80,15 @@ func (d *Device) IsUpdatedToFleetSpec(f *Fleet) bool {
...
+// IsDecommissioning() is true if the device has added a DeviceDecommissioning ConditionType to its Conditions.
+func (d *Device) IsDecommissioning() bool {
+	if d.Status == nil || d.Status.Conditions == nil {
+		return false
+	}
+
+	if found := FindStatusCondition(d.Status.Conditions, DeviceDecommissioning); found == nil {
+		return false
+	}
+	return true
```
hexfusion, line 93: do you need to validate the status or reason or is the existence enough?

https://github.com/flightctl/flightctl/blob/915f4c0596b7216683e309ee6b7c3f4d63a3b166/api/v1alpha1/properties.go#L42

internal/service/common/device.go
lkatalin: Basically yes; the lifecycle under normal circumstances can transition only from Enrolled -> Decommissioning or from Decommissioning -> Decommissioned. There is also a `LifecycleStatusUnknown` value but there is no scenario setting it to that currently.

api/v1alpha1/properties.go
lkatalin: Currently if the Condition exists then that is the signal that the device has switched to Decommissioning, so we don't need other checks to determine it.

hexfusion: OK I just figured that the condition existed but the value was `ConditionStatusFalse` by default.

lkatalin: Yep it makes sense now! Updated

hexfusion (approved):

PR #771, opened 2025-01-08
EDM-390: Only allow deletes on decommissioning devices

lkatalin:
This ensures that devices go through the proper decommissioning process.

## Summary by CodeRabbit

- **New Features**
  - Added a new query parameter that allows retrieving a summarized view of device details.

- **Refactor**
  - Updated device removal so that only decommissioned devices can be deleted.
  - Removed the bulk deletion option for devices to enforce stricter management controls.


internal/service/device.go
```diff
@@ -148,23 +148,7 @@ func (h *ServiceHandler) ListDevices(ctx context.Context, request server.ListDev
...
-	orgId := store.NullOrgId
-
-	err = h.store.Device().DeleteAll(ctx, orgId, h.callbackManager.AllDevicesDeletedCallback)
-	switch err {
-	case nil:
-		return server.DeleteDevices200JSONResponse{}, nil
-	default:
-		return nil, err
-	}
+	return nil, fmt.Errorf("delete all operations on devices are not permitted")
```
celdrake, line 151: This seems to also throw an error 500 in the CLI and the error message is not displayed

delete-all-devices

avishayt: Is there a reason to keep this endpoint around if we're not implementing it?

test/e2e/field_selectors/field_selectors_test.go
```diff
@@ -182,12 +182,6 @@ var _ = Describe("Field Selectors in Flight Control", func() {
 			Expect(out).To(ContainSubstring(unknownSelector))
 		})
 
-		It("Deleting resources", func() {
-			_, _ = harness.CLI("delete", "repo")
-			_, _ = harness.CLI("delete", "fleet")
-			_, _ = harness.CLI("delete", "device")
-		})
-
```
avishayt, line 190: Don't we need this for cleanup?

celdrake:
@lkatalin Are we settling in that deleting devices with any possible status is allowed, as long as the users have the permission to do so? Should this PR be closed in that case?

lkatalin:
> @lkatalin Are we settling in that deleting devices with any possible status is allowed, as long as the users have the permission to do so? Should this PR be closed in that case?

@celdrake It looks like we reversed this decision at the 1/27 decom meeting: "The decoming/ed devices will not automatically disappear, but instead will need to be manually deleted by the user if desired" (from the [recap](https://redhat-internal.slack.com/archives/C01KER8QV33/p1738159862261259)) so we will need this PR merged.

internal/service/device.go
lkatalin: I would be fine removing it.

lkatalin: Actually it looks like removing `DELETE` for devices would violate one of our `openapi` rules: https://github.com/flightctl/flightctl/blob/main/.spectral.yaml#L37. Thoughts?

celdrake:
> > @lkatalin Are we settling in that deleting devices with any possible status is allowed, as long as the users have the permission to do so? Should this PR be closed in that case?
> 
> @celdrake It looks like we reversed this decision at the 1/27 decom meeting: "The decoming/ed devices will not automatically disappear, but instead will need to be manually deleted by the user if desired" (from the [recap](https://redhat-internal.slack.com/archives/C01KER8QV33/p1738159862261259)) so we will need this PR merged.

Ok, so now we are saying that decommissioned devices will eventually disappear from the system after some time, right?
And regarding enrolled devices, they cannot be deleted manually? Therefore only decommissioned devices can be deleted? I need to take this into account in the UI (I had done it, but deleted the code for that).

lkatalin:
We have [decided not to have this functionality](https://redhat-internal.slack.com/archives/C01KER8QV33/p1739196292337349?thread_ts=1739196034.351569&cid=C01KER8QV33), so I am closing the PR.

PR #779, opened 2025-01-09
Remove docker compose, use quadlets

DakCrowder:
This PR:
- Removes current compose code from the `/deploy/podman` directory and replaces it with quadlets config
- Updates make commands accordingly
  - Introduces some script files to aid in deploying / cleaning up quadlets
  - Introduces some 'standalone' deployments for services like the db, kv, mq so they can be spun up as one-offs.  An alternative to this might be a more advanced templating setup for the .container files but I figured this was the simplest way to get spun up and depending on how future templating work is done they could be refactored or removed.
- Fixed a few small issues to (hopefully) get quadlets running "out of the box"

Known needed improvements / future work:
- Passing secrets/config instead of hardcoding info like passwords
- Templating work for passing data to the quadlets config e.g. networking or specific images

Builds on the [originial quadlets pr](https://github.com/flightctl/flightctl/pull/646)

## Spinning up services
 ```make deploy-quadlets``` - Run all of the services
```make clean-quadlets``` - Clean up the deployment

```make depoy-db``` - Spin up a standalone db service
 ```make kill-db``` - Stop the standalone service

## Release Notes

- **Deployment Changes**
  - Transitioned from Podman Compose to systemd-based service management
  - Simplified deployment process using new deployment scripts
  - Removed RabbitMQ message queue configuration

- **Service Configuration**
  - Added standalone configurations for PostgreSQL and Redis services
  - Updated service dependencies and volume mappings
  - Removed observability stack (Grafana, Prometheus)

- **Configuration Management**
  - Centralized configuration file locations
  - Updated configuration templates
  - Added key-value store configuration for services

- **Cleanup and Maintenance**
  - Introduced new cleanup script for managing system resources
  - Updated Makefile with new targets for linting and cleaning

### CODE REVIEW

**Comment by DakCrowder** [comment]
@coderabbitai review

**Comment by DakCrowder** [comment]
@coderabbitai review

File: deploy/podman/flightctl-periodic/flightctl-periodic.container [hunk 1]
```diff
@@ -9,7 +9,7 @@ Image=quay.io/flightctl/flightctl-periodic:latest
 Network=flightctl.network
 Environment=HOME=/root
 
-Volume=./flightctl-periodic-config/config.yaml:/root/.flightctl/config.yaml
+Volume=/etc/containers/systemd/flightctl-periodic/flightctl-periodic-config/config.yaml:/root/.flightctl/config.yaml
```
hexfusion, line 12: Just so I am clear the goal is to use this for production or why not use `~/.config/containers/systemd/`?

deploy/scripts/deploy_quadlet_service.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
```
hexfusion, line 3: :+1:

deploy/podman/flightctl-periodic/flightctl-periodic.container
DakCrowder: The short answer:  Because I punted on thinking about users

Slightly longer answer: The current goal for this is not to get things ready for production, but continue to build towards prod in subsequent work (and really the ticket was about removing compose so we don't have to support it).  I'm specifying /etc/containers/system mainly because that's where we are currently configured to move the .container files.  However after looking up some docs It seems like `~/.config/containers/systemd/` is used for rootless users which we will need to handle.  I don't believe ~ or $HOME can be resolved in the .container file by default  (currently experimenting a bit with it atm) which then necessitates passing env vars or some templating work (which we also 100% need) but I'd prefer to follow up w/ that in a separate PR. 

Let me know if my interpretation wasn't correct or you have other concerns

hexfusion: mostly jus curious, appreciate the details

hexfusion (approved):

hexfusion (approved):
nice work!

PR #787, opened 2025-01-14
EDM-391: Update device side decommission handling

lkatalin:
EDM-391: Update device side decommission handling

## Summary by CodeRabbit

- **New Features**
	- Enhanced device lifecycle management with improved decommissioning process.
	- Added ability to manage system reboot and certificate wiping.
	- Introduced status tracking for device lifecycle events.
	- New method for securely overwriting and wiping files.

- **Improvements**
	- Expanded systemd interaction capabilities.
	- Added more detailed logging for lifecycle transitions.

- **Technical Updates**
	- Updated lifecycle manager with new status and systemd client integrations.
	- Introduced methods for file management and secure deletion.
	- Added mock implementations for new file operations.

### DIFF
--- a/internal/agent/client/systemd.go
+++ b/internal/agent/client/systemd.go
@@ -53,6 +53,15 @@ func (s *Systemd) Stop(ctx context.Context, name string) error {
 	return nil
 }
 
+func (s *Systemd) Reboot(ctx context.Context) error {
+	args := []string{"reboot"}
+	_, stderr, exitCode := s.exec.ExecuteWithContext(ctx, systemctlCommand, args...)
+	if exitCode != 0 {
+		return fmt.Errorf("reboot systemd: %w", errors.FromStderr(stderr, exitCode))
+	}
+	return nil
+}
+
 func (s *Systemd) Restart(ctx context.Context, name string) error {
 	args := []string{"restart", name}
 	_, stderr, exitCode := s.exec.ExecuteWithContext(ctx, systemctlCommand, args...)
--- a/internal/agent/device/lifecycle/manager.go
+++ b/internal/agent/device/lifecycle/manager.go
@@ -11,6 +11,7 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/client"
 	"github.com/flightctl/flightctl/internal/agent/device/errors"
 	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	"github.com/flightctl/flightctl/internal/agent/device/status"
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/skip2/go-qrcode"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -32,11 +33,14 @@ type LifecycleManager struct {
 	deviceName           string
 	enrollmentUIEndpoint string
 	managementCertPath   string
+	managementKeyPath    string
 	deviceReadWriter     fileio.ReadWriter
 
 	enrollmentClient client.Enrollment
 	defaultLabels    map[string]string
 	enrollmentCSR    []byte
+	statusManager    status.Manager
+	systemdClient    *client.Systemd
 
 	backoff wait.Backoff
 	log     *log.PrefixLogger
@@ -47,10 +51,13 @@ func NewManager(
 	deviceName string,
 	enrollmentUIEndpoint string,
 	managementCertPath string,
+	managementKeyPath string,
 	deviceReadWriter fileio.ReadWriter,
 	enrollmentClient client.Enrollment,
 	enrollmentCSR []byte,
 	defaultLabels map[string]string,
+	statusManager status.Manager,
+	systemdClient *client.Systemd,
 	backoff wait.Backoff,
 	log *log.PrefixLogger,
 ) *LifecycleManager {
@@ -59,11 +66,14 @@ func NewManager(
 		deviceName:           deviceName,
 		enrollmentUIEndpoint: enrollmentUIEndpoint,
 		managementCertPath:   managementCertPath,
+		managementKeyPath:    managementKeyPath,
 		deviceReadWriter:     deviceReadWriter,
 		enrollmentClient:     enrollmentClient,
 		enrollmentCSR:        enrollmentCSR,
 		defaultLabels:        defaultLabels,
 		backoff:              backoff,
+		statusManager:        statusManager,
+		systemdClient:        systemdClient,
 	}
 }
 
@@ -92,12 +102,119 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
 }
 
 func (m *LifecycleManager) Sync(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
-	// TODO: implement
+	// this controller currently does not implement a sync operation
 	return nil
 }
 
 func (m *LifecycleManager) AfterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
-	// TODO: implement
+	var errs []error
+	if current.Decommission == nil && desired.Decommission != nil {
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		m.log.Warn("Updating Condition to decommissioning started")
+		if err := m.updateWithStartedCondition(ctx); err != nil {
+			errs = append(errs, fmt.Errorf("failed to update status with decommission started Condition: %w", err))
+			m.log.Warn("Unable to update Condition to decommissioning started")
+		}
+
+		// TODO: add support for additional decommissioning target types.
+		// these are the steps that will take places between Started and Completed status
+
+		if len(errs) == 0 {
+			m.log.Warn("No errors during decommissioning prior to wiping key and cert; updating Condition to decommissioning completed")
+			if err := m.updateWithCompletedCondition(ctx); err != nil {
+				errs = append(errs, fmt.Errorf("failed to update status with decommission completed Condition: %w", err))
+				m.log.Warn("Unable to update Condition to decommissioning completed")
+			}
+		} else {
+			m.log.Warn("Errors encountered during decommissioning; updating Condition to decommission error")
+			if err := m.updateWithErrorCondition(ctx, errs); err != nil {
+				errs = append(errs, fmt.Errorf("failed to update status with decommission errored Condition: %w", err))
+				m.log.Warn("Unable to update Condition to decommissioning error")
+			}
+		}
+
+		// after this point the device will no longer be able to communicate with the management service
+		m.log.Warn("Preparing to wipe agent certificate and keys and reboot")
+		if err := m.wipeAndReboot(ctx); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+	return nil
+}
+
+func (m *LifecycleManager) updateWithStartedCondition(ctx context.Context) error {
+	updateErr := m.statusManager.UpdateCondition(ctx, v1alpha1.Condition{
+		Type:    v1alpha1.DeviceDecommissioning,
+		Status:  v1alpha1.ConditionStatusTrue,
+		Reason:  string(v1alpha1.DecommissionStateStarted),
+		Message: "The device has started decommissioning",
+	})
+	if updateErr != nil {
+		m.log.Warnf("Failed setting status: %v", updateErr)
+		return fmt.Errorf("failed to update decommission started status: %w", updateErr)
+	}
+	return nil
+}
+
+func (m *LifecycleManager) updateWithCompletedCondition(ctx context.Context) error {
+	updateErr := m.statusManager.UpdateCondition(ctx, v1alpha1.Condition{
+		Type:    v1alpha1.DeviceDecommissioning,
+		Status:  v1alpha1.ConditionStatusTrue,
+		Reason:  string(v1alpha1.DecommissionStateComplete),
+		Message: "The device has completed decommissioning and will wipe its management certificate",
+	})
+	if updateErr != nil {
+		m.log.Warnf("Failed setting status: %v", updateErr)
+		return fmt.Errorf("failed to update decommission completed status: %w", updateErr)
+	}
+	return nil
+}
+
+func (m *LifecycleManager) updateWithErrorCondition(ctx context.Context, errs []error) error {
+	updateErr := m.statusManager.UpdateCondition(ctx, v1alpha1.Condition{
+		Type:    v1alpha1.DeviceDecommissioning,
+		Status:  v1alpha1.ConditionStatusTrue,
+		Reason:  string(v1alpha1.DecommissionStateError),
+		Message: fmt.Sprintf("The device has encountered one or more errors during decommissioning: %v", errors.Join(errs...)),
+	})
+	if updateErr != nil {
+		m.log.Warnf("Failed setting status: %v", updateErr)
+		return fmt.Errorf("failed to update decommission errored status: %w", updateErr)
+	}
+	return nil
+}
+
+// point of no return - wipes management cert and keys
+func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
+	var errs []error
+	err := m.deviceReadWriter.OverwriteAndWipe(m.managementCertPath)
+	if err != nil {
+		m.log.Errorf("Failed to remove management certificate at %s: %v", m.managementCertPath, err)
+		errs = append(errs, fmt.Errorf("failed to remove management certificate: %w", err))
+	}
+
+	err = m.deviceReadWriter.OverwriteAndWipe(m.managementKeyPath)
+	if err != nil {
+		m.log.Errorf("Failed to remove management key at %s: %v", m.managementKeyPath, err)
+		errs = append(errs, fmt.Errorf("failed to remove management key: %w", err))
+	}
+
+	// Clear sensitive data ahead of time in case reboot fails
+	m.deviceName = ""
+	m.enrollmentUIEndpoint = ""
+	m.enrollmentClient = nil
+	m.enrollmentCSR = nil
+
+	// TODO: incorporate before-reboot hooks
+	if err = m.systemdClient.Reboot(ctx); err != nil {
+		errs = append(errs, fmt.Errorf("failed to initiate system reboot: %w", err))
+	}
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
 	return nil
 }
 
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -171,10 +171,13 @@ func (a *Agent) Run(ctx context.Context) error {
 		deviceName,
 		a.config.EnrollmentService.EnrollmentUIEndpoint,
 		a.config.ManagementService.GetClientCertificatePath(),
+		a.config.ManagementService.GetClientKeyPath(),
 		deviceReadWriter,
 		enrollmentClient,
 		csr,
 		a.config.DefaultLabels,
+		statusManager,
+		systemdClient,
 		backoff,
 		a.log,
 	)
--- a/internal/agent/device/fileio/fileio.go
+++ b/internal/agent/device/fileio/fileio.go
@@ -30,6 +30,7 @@ type Writer interface {
 	MkdirAll(path string, perm fs.FileMode) error
 	CopyFile(src, dst string) error
 	CreateManagedFile(file ign3types.File) (ManagedFile, error)
+	OverwriteAndWipe(file string) error
 }
 
 type Reader interface {
--- a/internal/agent/device/fileio/mock_fileio.go
+++ b/internal/agent/device/fileio/mock_fileio.go
@@ -164,6 +164,20 @@ func (mr *MockWriterMockRecorder) MkdirAll(path, perm any) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "MkdirAll", reflect.TypeOf((*MockWriter)(nil).MkdirAll), path, perm)
 }
 
+// OverwriteAndWipe mocks base method.
+func (m *MockWriter) OverwriteAndWipe(file string) error {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "OverwriteAndWipe", file)
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
+// OverwriteAndWipe indicates an expected call of OverwriteAndWipe.
+func (mr *MockWriterMockRecorder) OverwriteAndWipe(file any) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "OverwriteAndWipe", reflect.TypeOf((*MockWriter)(nil).OverwriteAndWipe), file)
+}
+
 // PathFor mocks base method.
 func (m *MockWriter) PathFor(filePath string) string {
 	m.ctrl.T.Helper()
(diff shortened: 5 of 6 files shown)


internal/agent/device/device.go
```diff
@@ -189,10 +189,48 @@ func (a *Agent) sync(ctx context.Context, current, desired *v1alpha1.RenderedDev
 		return fmt.Errorf("sync device: %w", err)
 	}
 
+	if a.lifecycleManager.StartedDecommissioning() {
```
hexfusion, line 192: Can we incorporate this into lifecycleManager.AfterUpdate()? I think in your case we can add statusManager as a dependency of  lifecycleManager.

then in that method you can handle all of these status calls.

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,39 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+	// once we support more types
+	if m.decommission != nil {
+		m.decommission.state = v1alpha1.DecommissionStateComplete
+	}
 	return nil
 }
 
+// point of no return - wipes management cert and keys
+func (m *LifecycleManager) WipeAndReboot() {
+	err := os.Remove(m.managementCertPath)
```
hexfusion, line 123: For anything touching the filesystem, we want to use the `fileio` package. So, in this case, you will want to add `fileio.Writer` as a dependency and call `RemoveFile`. This is to allow us to perform mock test against the file system inside a tmpDir.

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +96,79 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+		m.UpdateWithStartedCondition(ctx)
+
+		// todo - switch handling based on decommission target type once we support more types;
+		// keep track of whether we have an error
+
+		if !errored {
+			m.UpdateWithCompletedCondition(ctx)
+		}
+
+		// no return
```
hexfusion, line 115: lets log every step in here so we know exactly what happened

internal/agent/device/lifecycle/lifecycle.go
```diff
@@ -9,6 +9,10 @@ import (
 type Manager interface {
 	Sync(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error
 	AfterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error
+	WipeAndReboot()
+	UpdateWithErrorCondition(ctx context.Context, err error)
+	UpdateWithStartedCondition(ctx context.Context)
+	UpdateWithCompletedCondition(ctx context.Context)
```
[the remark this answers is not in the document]
lkatalin: That's a good point

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +96,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
 	return nil
 }
 
 func (m *LifecycleManager) AfterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
-	// TODO: implement
+	if current.Decommission == nil && desired.Decommission != nil {
+		m.log.Warnf("Detected decommissioning request from flightctl service")
+		errored := false
+		m.updateWithStartedCondition(ctx)
+		m.log.Warnf("Updated Condition to decommissioning started")
```
hexfusion, line 108: nit:
```suggestion
		m.log.Warn("Updated Condition to decommissioning started")
```

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +96,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+	err := m.deviceReadWriter.RemoveFile(m.managementCertPath)
+	if err != nil {
+		m.log.Errorf("Unable to delete mcert: %v", err)
+	}
+
+	m.enrollmentUIEndpoint = ""
+	m.enrollmentClient = nil
+	m.enrollmentCSR = nil
+
+	// todo - reboot
```
hexfusion, line 161: I think this is critical right? To do this you will need an executor to send a command via os exec. Or extend the systemd client.

https://github.com/flightctl/flightctl/blob/main/internal/agent/client/systemd.go#L18

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +96,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+		Reason:  string(v1alpha1.DecommissionStateComplete),
+		Message: "The device has completed decommissioning and will wipe its management certificate",
+	})
+	if updateErr != nil {
+		m.log.Warnf("Failed setting status: %v", updateErr)
+	}
+}
+
+// point of no return - wipes management cert and keys
+func (m *LifecycleManager) wipeAndReboot() {
```
hexfusion, line 151: this should return an error incase there is a failure by RemoveFile or Reboot

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +96,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+		// keep track of whether we have an error
+
+		if !errored {
+			m.log.Warnf("No errors during decommissioning, updating Condition to decommissioning completed")
+			m.updateWithCompletedCondition(ctx)
+		}
+
+		// no return
+		m.log.Warnf("Preparing to wipe agent certificate and keys and reboot")
+		m.wipeAndReboot()
```
hexfusion, line 120: since this will return an error its existence should send error status or what is the expected result

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +96,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
-	// TODO: implement
+	// this controller currently does not implement a sync operation
 	return nil
 }
 
 func (m *LifecycleManager) AfterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
-	// TODO: implement
+	if current.Decommission == nil && desired.Decommission != nil {
+		m.log.Warnf("Detected decommissioning request from flightctl service")
+		errored := false
```
hexfusion, line 106: nit: we should use actual errors as conditions of error

internal/agent/device/lifecycle/manager.go
lkatalin: I can bubble up the error

internal/agent/device/lifecycle/manager.go
lkatalin: Yeah, I guess once we have more handling here we can just return early if an error comes up so we won't need this bool

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,67 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		m.updateWithStartedCondition(ctx)
+		m.log.Warn("Updated Condition to decommissioning started")
+
+		// todo - switch handling based on decommission target type once we support more types
+
+		m.updateWithCompletedCondition(ctx)
+		m.log.Warn("No errors during decommissioning, updated Condition to decommissioning completed")
+
+		// no return
```
hexfusion, line 117: nit: 
```suggestion
		// after this point the device will no longer be able to communicate with the management service.
```

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,67 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+	if err != nil {
+		m.log.Errorf("Unable to delete mcert: %v", err)
+		return err
+	}
+
+	m.enrollmentUIEndpoint = ""
+	m.enrollmentClient = nil
+	m.enrollmentCSR = nil
+
+	return m.systemdClient.Reboot(ctx, name)
```
hexfusion, line 160: can we add a TODO:  incorporate before reboot hooks

internal/agent/client/systemd.go
```diff
@@ -53,6 +53,15 @@ func (s *Systemd) Stop(ctx context.Context, name string) error {
 	return nil
 }
 
+func (s *Systemd) Reboot(ctx context.Context, name string) error {
+	args := []string{"reboot", name}
+	_, stderr, exitCode := s.exec.ExecuteWithContext(ctx, systemctlCommand, args...)
+	if exitCode != 0 {
+		return fmt.Errorf("reboot systemd unit: %s: %w", name, errors.FromStderr(stderr, exitCode))
```
hexfusion, line 60:
```suggestion
		return fmt.Errorf("system reboot: %s: %w", name, errors.FromStderr(stderr, exitCode))
```

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,68 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
-	// TODO: implement
+	if current.Decommission == nil && desired.Decommission != nil {
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		m.updateWithStartedCondition(ctx)
+		m.log.Warn("Updated Condition to decommissioning started")
+
+		// todo - switch handling based on decommission target type once we support more types
+
+		m.updateWithCompletedCondition(ctx)
+		m.log.Warn("No errors during decommissioning, updated Condition to decommissioning completed")
```
hexfusion, line 115: is this true what if wipeAndReboot fails

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,68 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
 }
 
 func (m *LifecycleManager) AfterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
-	// TODO: implement
+	if current.Decommission == nil && desired.Decommission != nil {
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		m.updateWithStartedCondition(ctx)
+		m.log.Warn("Updated Condition to decommissioning started")
+
+		// todo - switch handling based on decommission target type once we support more types
```
hexfusion, line 112:
```suggestion
		// TODO:  add support for additional types
```

internal/agent/device/lifecycle/manager.go
lkatalin: `wipeAndReboot` is only called after this point. But, actually, `wipeAndReboot` needs to be called (but `updateWithCompletedCondition` should *not* be) even if the previous handling steps resulted in an error. so, I'm going to add back the `errored` bool.

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
-	// TODO: implement
+	if current.Decommission == nil && desired.Decommission != nil {
+		errored := false
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		m.updateWithStartedCondition(ctx)
+		m.log.Warn("Updated Condition to decommissioning started")
+
+		// TODO: add support for additional decommissioning target types
+
+		if !errored {
```
ori-amizur, line 115: If was set unconditionally to false - why check it here again ?

internal/agent/device/lifecycle/manager.go
hexfusion: I would say we don't need it now right it can't be true right?

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+
+// point of no return - wipes management cert and keys
+func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
+	err := m.deviceReadWriter.RemoveFile(m.managementCertPath)
+	if err != nil {
+		m.log.Errorf("Unable to delete mcert: %v", err)
+		return err
+	}
+
+	m.enrollmentUIEndpoint = ""
```
ori-amizur, line 159: What are we wiping here and what is the purpose? The wipe is only from memory which will not persist after reboot

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,15 +99,71 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+	if current.Decommission == nil && desired.Decommission != nil {
+		errored := false
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		m.updateWithStartedCondition(ctx)
+		m.log.Warn("Updated Condition to decommissioning started")
+
+		// TODO: add support for additional decommissioning target types
+
+		if !errored {
+			m.updateWithCompletedCondition(ctx)
```
ori-amizur, line 116: The status is set to starting and immediately set to completed without any action in-between.  So what is the purpose of setting it to "started" to begin with ?

hexfusion: It is a reasonable question status updates are not free so we should make sure they are required. At a minimum document intent if it is future functionality.

internal/agent/device/lifecycle/manager.go
lkatalin: The `todo` part that isn't implemented yet is where it may be set to true

internal/agent/device/lifecycle/manager.go
lkatalin: Yes, it is future functionality. Currently we only support "unenroll" which basically will just do `wipeAndReboot`. In the future other targets will add further handling steps between `Started` and `Completed`. But this is stubbing out the intent here. I can add a more detailed note

internal/agent/device/lifecycle/manager.go
lkatalin: Makes sense

internal/agent/device/lifecycle/manager.go
lkatalin: I'm implementing the coderabbit suggestions so this is going away anyhow

internal/agent/device/lifecycle/manager.go
hexfusion: can we add a note

lkatalin: Yep done

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,12 +99,78 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+	if updateErr != nil {
+		m.log.Warnf("Failed setting status: %v", updateErr)
+		return fmt.Errorf("failed to update decommission completed status: %w", updateErr)
+	}
+	return nil
+}
+
+// point of no return - wipes management cert and keys
+func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
+	err := m.deviceReadWriter.RemoveFile(m.managementCertPath)
```
fzdarsky, line 159: Why delete only the management cert and not the management key (which is what's required impersonate a device with the same fingerprint)?

lkatalin: Good catch, I missed this since it was not in the lifecycle manager but I've updated now to pass this in and remove the file.

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,12 +102,85 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+	// this controller currently does not implement a sync operation
 	return nil
 }
 
 func (m *LifecycleManager) AfterUpdate(ctx context.Context, current, desired *v1alpha1.RenderedDeviceSpec) error {
-	// TODO: implement
+	if current.Decommission == nil && desired.Decommission != nil {
+		m.log.Warn("Detected decommissioning request from flightctl service")
+		if err := m.updateWithStartedCondition(ctx); err != nil {
+			return fmt.Errorf("failed to update decommission started status: %w", err)
```
hexfusion, line 113: If the device can't connect to the management service to update status do you want to decommission anyway?

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,12 +102,85 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+		}
+		m.log.Warn("Updated Condition to decommissioning started")
+
+		// TODO: add support for additional decommissioning target types.
+		// these are the steps that will take places between Started and Completed status
+
+		if err := m.updateWithCompletedCondition(ctx); err != nil {
+			return fmt.Errorf("failed to update decommission completed status: %w", err)
```
hexfusion, line 121: The same is true here. The point of returning the error from the method is not to exit the decommission flow, right? If the device isn't connected to the service, do we still want to continue to decommission?

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,12 +102,85 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+		}
+		m.log.Warn("No errors during decommissioning, updated Condition to decommissioning completed")
+
+		// after this point the device will no longer be able to communicate with the management service
+		m.log.Warn("Preparing to wipe agent certificate and keys and reboot")
+		return m.wipeAndReboot(ctx)
```
hexfusion, line 127: If this fails what should the agent do?

internal/agent/device/lifecycle/manager.go
hexfusion: if one fails, logically do you want to continue and remove as much as you can or return?

internal/agent/device/lifecycle/manager.go
```diff
@@ -59,11 +66,14 @@ func NewManager(
 		deviceName:           deviceName,
 		enrollmentUIEndpoint: enrollmentUIEndpoint,
 		managementCertPath:   managementCertPath,
+		managementKeyPath:    managementKeyPath,
 		deviceReadWriter:     deviceReadWriter,
 		enrollmentClient:     enrollmentClient,
```
hexfusion, line 71: there is also the managementClient I believe the certs are in its memory?

internal/agent/device/lifecycle/manager.go
lkatalin: Yeah, I think we need to separate the actual flow (which should complete no matter what) from the status reports. This is why I was keeping that `errored := false` bool so that execution could complete but we would know which status to send back.

lkatalin: Maybe I can instead have an array of errors we can return at the end, I have seen that elsewhere in the code.

internal/agent/device/lifecycle/manager.go
lkatalin: I'm not sure there's much that it can do except retry, which may or may not be worth it. Thoughts?

internal/agent/device/lifecycle/manager.go
lkatalin: I've looked around and it seems like it just has the path to the cert, not the cert itself, unless I missed it. 
https://github.com/flightctl/flightctl/blob/ebe6e9a425c7ee1f164f729df1b534808709617b/internal/agent/device/bootstrap.go#L265

internal/agent/device/lifecycle/manager.go
lkatalin: We should remove as much as possible. There isn't a good way to do troubleshooting at this stage

internal/agent/device/lifecycle/manager.go
lkatalin: Updated

internal/agent/device/lifecycle/manager.go
lkatalin: Updated

internal/agent/device/lifecycle/manager.go
mangelajo: We should add an wipe method to deviceReaderWriter, that must store 0's over all the contents, then remove file.

This is not perfect and best efforts, in some filesystems/or when using lvm with snapshots, the block changes will be logged, and the older content is still persisted to disk.

What I am trying to prevent here is that somebody would scan the disk raw contents for keys/certs, and find ours this is actually something relatively easy to do technically.

mangelajo: Another question, will we incorporate support for the user running additional wipe scripts? 

i.e. customers may want to perform full disk block wipe under some conditions.

mangelajo (changes requested):
I believe we may want to do wiping of files instead of only deletion. If not now, at least in a follow up, WDYT?

mangelajo:
Beyond my comment, I think the code is great, very clean.

mangelajo:
A possible implementation to secure wipe:

```go
package main

import (
	"crypto/rand"
	"fmt"
	"os"
)

func wipeFileWithRandomData(filePath string) error {
	// Open the file for writing without truncating.
	file, err := os.OpenFile(filePath, os.O_WRONLY, 0)
	if err != nil {
		return fmt.Errorf("failed to open file: %w", err)
	}
	defer file.Close()

	// Get the file's size.
	fileInfo, err := file.Stat()
	if err != nil {
		return fmt.Errorf("failed to get file info: %w", err)
	}
	fileSize := fileInfo.Size()

	// Generate random data.
	randomData := make([]byte, fileSize)
	if _, err := rand.Read(randomData); err != nil {
		return fmt.Errorf("failed to generate random data: %w", err)
	}

	// Write the random data over the existing content.
	if _, err := file.WriteAt(randomData, 0); err != nil {
		return fmt.Errorf("failed to write random data: %w", err)
	}

	return nil
}

func main() {
	filePath := "example.txt" // Replace with your file path
	err := wipeFileWithRandomData(filePath)
	if err != nil {
		fmt.Printf("Error: %v\n", err)
	} else {
		fmt.Println("File successfully wiped with random data.")
	}
}
```

we may want to add a method to the file readerwriter interface we have

mangelajo:
@coderabitai fullreview

internal/agent/device/lifecycle/manager.go
lkatalin: That's a great idea @mangelajo , I'll make an update.
Re: your question, yes, supporting additional scripts and options is future functionality that we will add when we add additional decommissioning "targets".

internal/agent/device/lifecycle/manager.go
lkatalin: The service side expects the `Started` Condition followed by the `Completed` Condition, so we can't skip these.

internal/agent/device/lifecycle/manager.go
```diff
@@ -92,12 +102,105 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
...
+				errs = append(errs, fmt.Errorf("failed to update decommission completed status: %w", err))
+				m.log.Warn("Unable to update Condition to decommissioning completed")
+			}
+		} else {
+			m.log.Warn("Errors encountered during decommissioning; updating Condition to decommission error")
+			if err := m.updateWithCompletedCondition(ctx); err != nil {
+				errs = append(errs, fmt.Errorf("failed to update decommission completed status: %w", err))
+				m.log.Warn("Unable to update Condition to decommissioning completed")
+			}
+		}
```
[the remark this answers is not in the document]
lkatalin: Good catch

lkatalin:
Could use another review! Thanks

lkatalin:
@hexfusion @mangelajo Do you think we can merge?

hexfusion (approved):

mangelajo (approved):

PR #785, opened 2025-01-13
EDM-967: Add identity information for K8S

ori-amizur:
Add identity information to the context when auth type is k8s based on token review result.

## Release Notes

- **Dependencies**
  - Added new dependency: `github.com/jellydator/ttlcache/v3`
  - Updated `golang.org/x/sync` to latest version

- **New Features**
  - Enhanced authentication middleware with identity retrieval functionality
  - Added token review caching mechanism for improved performance

- **Improvements**
  - Introduced `GetIdentity` method across authentication providers
  - Added support for storing user identity in request context
  - Defined new `Identity` struct for user information management

PR #788, opened 2025-01-14
EDM-882: Approving ER adds user identity

rawagner:
Thanks to https://github.com/flightctl/flightctl/pull/785 , we can now retrieve user identity and populate `EnrollmentRequestApproval`

## Summary by CodeRabbit

- **New Features**
	- Added labels support for enrollment request approvals.
	- Enhanced error handling for enrollment request approval process.
	- Improved API response structure for enrollment request approvals.
	- Introduced a new function to retrieve user identity from the context.

- **Bug Fixes**
	- Updated validation and type handling for approval-related fields.
	- Refined identity retrieval during enrollment request approval.

- **Documentation**
	- Updated OpenAPI specification for enrollment request approval workflow.

- **Refactor**
	- Restructured enrollment request approval status representation.
	- Improved type safety for approval-related data structures.
	- Updated handling of approver information in output display.
	- Modified the logic for approval processing and identity retrieval.


internal/service/enrollmentrequest.go
```diff
@@ -383,24 +384,31 @@ func (h *ServiceHandler) ApproveEnrollmentRequest(ctx context.Context, request s
 		if v1alpha1.IsStatusConditionTrue(enrollmentReq.Status.Conditions, v1alpha1.EnrollmentRequestApproved) {
 			return server.ApproveEnrollmentRequest400JSONResponse{Message: "Enrollment request is already approved"}, nil
 		}
-		if request.Body.ApprovedAt != nil {
-			return server.ApproveEnrollmentRequest400JSONResponse{Message: "ApprovedAt is not allowed to be set when approving enrollment requests"}, nil
+
+		identity, err := authcommon.GetIdentity(ctx)
+		if err != nil {
+			return server.ApproveEnrollmentRequest500JSONResponse{Message: fmt.Sprintf("Failed to retrieve user identity while approving enrollment request: %v", err.Error())}, nil
```
fzdarsky, line 390: If you return the error as second parameter (instead of `nil`), this'll generate a 500 error.

api/v1alpha1/openapi.yaml
```diff
@@ -2064,6 +2064,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+        "500":
+          description: InternalServerError
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
fzdarsky, line 2072: Explicitly declaring 500 response codes shouldn't be necessary (they happen if we return a non-nil error, see below).

api/v1alpha1/openapi.yaml
```diff
@@ -4192,15 +4198,24 @@ components:
...
+          approvedBy:
+            type: string
+            description: The name of the approver.
+          approvedAt:
+            type: string
+            format: date-time
+            description: The time at which the request was approved.
+        required:
+        - approvedBy
+        - approvedAt
```
fzdarsky, line 4218: What's the motivation of splitting the `approvedBy` and `approvedAt` out of the `EnrollmentRequestApproval`? And not the approved labels, for example?

api/v1alpha1/openapi.yaml
```diff
@@ -4192,15 +4198,24 @@ components:
         approved:
           type: boolean
           description: Indicates whether the request has been approved.
```
fzdarsky, line 4200: Do we still need the `approved` field, considering that we have an `Approved` condition?

internal/service/enrollmentrequest.go
rawagner: great, thanks for pointing that out! Fixed.

api/v1alpha1/openapi.yaml
rawagner: Removed from schema.

api/v1alpha1/openapi.yaml
rawagner: The motivation is to have a cleaner API.
Before the change we used the same object (`EnrollmentRequestApproval`) for incoming approval requests and also for ER status. However the original `EnrollmentRequestApproval` contained `approvedBy` and `approvedAt` fields which do not make sense for the approval request, only for status.

So I split them to a separate objects - one for approval `EnrollmentRequestApproval` and one for ER status `EnrollmentRequestApprovalStatus`.

api/v1alpha1/openapi.yaml
rawagner: hmm..I guess we don't need it anymore. Do we want to remove the `labels` field from the too `EnrollmentRequestApprovalStatus` too? What value does it bring to keep the labels that the ER was approved with ?

celdrake: In the UI, we're currently filtering the list of Pending ER by querying those that have their `status.approval.approved` field either unset or false. So we still need this field, as it's not possible to filter by conditions IIUC.

celdrake: Sorry, I was confused by the comment regarding the Condition, I thought we were talking about the ER status field and not to the response of approving an ER.  
In the response for approving an ER, it's OK to remove the `approved` field.

rawagner: I believe we are talking about the ER status.approved field :)

lkatalin (approved):
I see there may be a couple of outstanding details but it looks like a good change IMO

PR #789, opened 2025-01-14
EDM-885: make build-cli-multiarch target

gshilin-sdb:
## Summary by CodeRabbit

- **Workflow Updates**
	- Enhanced GitHub Actions workflow for CLI binary publishing
	- Improved environment variable handling and job sequencing
	- Standardized output file naming conventions
	- Added a new setup job for caching and dependency installation
	- Modified jobs for building, verifying, and publishing binaries with additional integrity checks

- **Utility Changes**
	- Updated string to integer conversion method in validation function for more robust numeric parsing

### DIFF
--- a/.github/workflows/publish-cli-bins.yaml
+++ b/.github/workflows/publish-cli-bins.yaml
@@ -9,89 +9,159 @@ on:
   pull_request:
 
 jobs:
+  setup:
+    name: Set Up Shared Environment
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y make
+
   build:
+    name: Build Binaries
     strategy:
       matrix:
-        os: [linux, darwin]
-        arch: [amd64, arm64]
-    env:
-      NAME: flightctl-${{ matrix.os }}-${{ matrix.arch }}
-    runs-on: "ubuntu-24.04"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-cli-bins.yaml
```diff
@@ -9,89 +9,160 @@ on:
...
+            ${{ runner.os }}-go-
+
+      # Install dependencies
+      - name: Install Dependencies
+        run: go mod tidy
+
+      - name: Build
         run: |
-          make build-cli
+          GOOS="${GOOS}" GOARCH="${GOARCH}" make build-cli
```
mangelajo, line 74: What I was proposing was more around having a make build-cli-multiarch target itself that you could use from anywhere

inside the target you could loop. or have a script that runs

for GOARCH in amd64 arm64; do
     for GOOS in linux darwin windows; do
            GOOS="${GOOS}" GOARCH="${GOARCH}" make build-cli
            # SHA signature
            # renaming binaries
      done
done

But put this acessible from a make target.


The reason for asking this, is that then we can use this in a Container file to build the "container wrapped" cli binaries.

mangelajo (changes requested):
See the comments, sorry for keeping you looping on it, but I think it will be simpler for the CLI wrapping containerfile.

mangelajo (approved):

PR #801, opened 2025-01-19
EDM-1039: Make CLI code windows compatible

gshilin-sdb:
* Make CLI code windows compatible
* Add windows target to publish-cli-bins

## Summary by CodeRabbit

- **New Features**
  - Expanded our CLI binary release process to support a broader range of operating systems and architectures, including dedicated verification for Windows.
  - Improved the build tooling to deliver clearly named and properly archived CLI binaries (using appropriate formats like .exe, .zip, and .tar.gz) for a more robust distribution process.
  - Introduced a comprehensive configuration management structure for connecting to the Flight Control API server, enhancing validation and error handling.
  
- **Bug Fixes**
  - Updated various configuration handling methods to ensure compatibility with the new structure and types.

- **Documentation**
  - Enhanced comments and output messages for clarity throughout the configuration and build processes.

### DIFF
--- a/.github/workflows/publish-cli-bins.yaml
+++ b/.github/workflows/publish-cli-bins.yaml
@@ -20,80 +20,110 @@ jobs:
 
   build:
     name: Build Binaries
-    strategy:
-      matrix:
-        os_arch: [ "linux-amd64", "linux-arm64", "darwin-amd64", "darwin-arm64" ]
     runs-on: ubuntu-latest
     needs: setup
+    strategy:
+      matrix:
+        tag: [ linux-amd64, linux-arm64, mac-amd64, mac-arm64, windows-amd64, windows-arm64 ]
+        include:
+          # Linux builds
+          - &linux-base
+            os: linux
+            zip: .tar.gz
+            ext: ""
+          - <<: *linux-base
+            arch: amd64
+            tag: linux-amd64
+          - <<: *linux-base
+            arch: arm64
+            tag: linux-arm64
+
+          # macOS builds
+          - &mac-base
+            os: darwin
+            zip: .zip
+            ext: ""
+          - <<: *mac-base
+            arch: amd64
+            tag: mac-amd64
+          - <<: *mac-base
+            arch: arm64
+            tag: mac-arm64
+
+          # Windows builds
+          - &windows-base
+            os: windows
+            zip: .zip
+            ext: .exe
+          - <<: *windows-base
+            arch: amd64
+            tag: windows-amd64
+          - <<: *windows-base
+            arch: arm64
+            tag: windows-arm64
     steps:
-      - name: Set up Environment Variables
-        run: |
-          # Split os-arch into os and arch
-          IFS="-" read -r GOOS GOARCH <<< "${{ matrix.os_arch }}"
-          # Export variables to GITHUB_ENV for subsequent steps
-          {
-            echo "GOOS=$GOOS"
-            echo "GOARCH=$GOARCH"
-          } >> $GITHUB_ENV
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.21
 
       - name: Checkout Code
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: 1.21
-
-      # Cache Go modules
-      - name: Cache Go Modules
-        uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cache/go-build
-            ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-
-
-      # Install dependencies
       - name: Install Dependencies
         run: go mod tidy
 
       - name: Build
         run: |
-          GOOS="${GOOS}" GOARCH="${GOARCH}" make build-cli
-          SHA=$(shasum -a 256 bin/flightctl | awk '{ print $1 }')
-          echo "${SHA}" > "flightctl-${{ matrix.os_arch }}-sha256.txt"
-          mv bin/flightctl "flightctl-${{ matrix.os_arch }}"
-          tar cvf "flightctl-${{ matrix.os_arch }}.tar.gz" "flightctl-${{ matrix.os_arch }}"
+          GOOS="${{ matrix.os }}" GOARCH="${{ matrix.arch }}" make build-cli
+          mv bin/flightctl${{ matrix.ext }} flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
+          SHA=$(shasum -a 256 flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }} | cut -d' ' -f1)
+          echo ${SHA} > flightctl-${{ matrix.os }}-${{ matrix.arch}}-sha256.txt
+          if [ "${{ matrix.zip }}" = ".zip" ]; then
+            zip flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }} flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
+          else
+            tar cvf flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }} flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
+          fi
 
-      - name: Save tar binary
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Save zip
         uses: actions/upload-artifact@v4
         with:
-          name: flightctl-${{ matrix.os_arch }}.tar.gz
-          path: flightctl-${{ matrix.os_arch }}.tar.gz
+          name: flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }}
+          path: flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }}
+
+      - name: Save sha256
+        uses: actions/upload-artifact@v4
+        with:
+          path: flightctl-${{ matrix.os }}-${{ matrix.arch}}-sha256.txt
+          name: flightctl-${{ matrix.os }}-${{ matrix.arch}}-sha256.txt
 
       - name: Save binary
         uses: actions/upload-artifact@v4
         with:
-          name: flightctl-${{ matrix.os_arch }}
-          path: flightctl-${{ matrix.os_arch }}
+          name: flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
+          path: flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
 
-      - name: Save checksum
-        uses: actions/upload-artifact@v4
-        with:
-          name: flightctl-${{ matrix.os_arch }}-sha256.txt
-          path: flightctl-${{ matrix.os_arch }}-sha256.txt
-
-  verify:
-    name: Verify Binaries
+  verify-windows:
+    name: Verify Binaries on windows
+    runs-on: "windows-latest"
+    needs: build
     strategy:
       matrix:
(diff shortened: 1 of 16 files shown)


avishayt:
Do we have any tests for this? How do we know if it breaks?

Makefile
```diff
@@ -185,6 +185,7 @@ $(GOBIN)/golangci-lint:
 
 lint: tools
 	$(GOBIN)/golangci-lint run -v
+	CGO_ENABLED=0 GOOS=windows GOARCH=arm64 go build -buildvcs=false $(GO_BUILD_FLAGS) -o /dev/null ./cmd/flightctl
```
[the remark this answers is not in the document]
gshilin-sdb: This is build on Linux, so /dev/null is correct

mangelajo, line 188: why did you add this under lint? We will now have verification that it builds during the build phase, we can remove this here.


but, you need to add support here:

https://github.com/flightctl/flightctl/blob/main/hack/build_multiarch_clis.sh

So that we also get windows binaries into the CLI artifact container.

This is what I meant, that we should use that make target/script from here instead. So you don't have to maintain it at two different places. 

That one is used from: https://github.com/flightctl/flightctl/blob/main/Containerfile.cli-artifacts#L16

mangelajo (changes requested):
one comment :)

mangelajo:
> Do we have any tests for this? How do we know if it breaks?

image

For every PR we try to compile every combination of the CLI (we just don't publish it.)

.github/workflows/publish-cli-bins.yaml
```diff
@@ -20,88 +20,89 @@ jobs:
...
-          path: flightctl-${{ matrix.os_arch }}.tar.gz
+          name: "flightctl-${{ matrix.os_arch }}${{ matrix.zip }}"
+          path: "flightctl-${{ matrix.os_arch }}${{ matrix.zip }}"
 
       - name: Save binary
         uses: actions/upload-artifact@v4
         with:
-          name: flightctl-${{ matrix.os_arch }}
-          path: flightctl-${{ matrix.os_arch }}
+          name: "flightctl-${{ matrix.os_arch }}${{ matrix.ext }}"
```
mangelajo, line 93: you may need to fix the shell script to also generate the non zipped binaries, I didn't account for this, sorry.

mangelajo (changes requested):

gshilin-sdb:
> Do we have any tests for this? How do we know if it breaks?

We're creating a windows executable in GH workflow for each PR.

Makefile
gshilin-sdb: removed

.github/workflows/publish-cli-bins.yaml
```diff
@@ -20,88 +20,89 @@ jobs:
...
 
       - name: Build
         run: |
-          GOOS="${GOOS}" GOARCH="${GOARCH}" make build-cli
-          SHA=$(shasum -a 256 bin/flightctl | awk '{ print $1 }')
-          echo "${SHA}" > "flightctl-${{ matrix.os_arch }}-sha256.txt"
-          mv bin/flightctl "flightctl-${{ matrix.os_arch }}"
-          tar cvf "flightctl-${{ matrix.os_arch }}.tar.gz" "flightctl-${{ matrix.os_arch }}"
+          make multiarch-build-cli
+          pwd
```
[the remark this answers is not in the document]
gshilin-sdb: no, just testing it

Makefile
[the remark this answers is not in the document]
gshilin-sdb: removed

internal/agent/device/console/console_windows.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func init() {
+	// Ensure DLL and procedures are loaded
+	if err := kernel32.Load(); err != nil {
+		panic(fmt.Sprintf("failed to load kernel32.dll: %v", err))
+	}
+}
+
+func setConsoleSize(fd uintptr, size TerminalSize) error {
```
mangelajo, line 31: Why did you need to do anything with the console at all?

agent code should not be compiled for the CLI, where is the CLI pulling the internal/agent/device/console ?

gshilin-sdb: removed all `console` stuff

.github/workflows/publish-cli-bins.yaml
```diff
@@ -20,80 +20,98 @@ jobs:
...
+            ext: ""
+          - tag: mac-arm64
+            arch: arm64
+            os: darwin
+            zip: .zip
+            ext: ""
+          - tag: windows-amd64
+            arch: amd64
+            os: windows
+            zip: zip
```
mangelajo, line 52:
```suggestion
            zip: .zip
```

mangelajo: coderabbitt told you :D

.github/workflows/publish-cli-bins.yaml
```diff
@@ -20,80 +20,98 @@
...
-          GOOS="${GOOS}" GOARCH="${GOARCH}" make build-cli
-          SHA=$(shasum -a 256 bin/flightctl | awk '{ print $1 }')
-          echo "${SHA}" > "flightctl-${{ matrix.os_arch }}-sha256.txt"
-          mv bin/flightctl "flightctl-${{ matrix.os_arch }}"
-          tar cvf "flightctl-${{ matrix.os_arch }}.tar.gz" "flightctl-${{ matrix.os_arch }}"
+          GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }} make build-cli
+          mv bin/flightctl${{ matrix.ext }} flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
+          SHA=$(shasum -a 256 flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }} | cut -d' ' -f1)
+          echo ${SHA} > flightctl-${{ matrix.os }}-${{ matrix.arch}}-sha256.txt
+          tar cvf flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }} flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}
```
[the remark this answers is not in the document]
mangelajo: This is correct too... you cannot use tar to generate a zip you need to use zip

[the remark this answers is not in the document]
mangelajo: @gshilin-sdb

.github/workflows/publish-cli-bins.yaml
```diff
@@ -20,80 +20,98 @@
...
-      - name: Save tar binary
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Save zip
         uses: actions/upload-artifact@v4
         with:
-          name: flightctl-${{ matrix.os_arch }}.tar.gz
-          path: flightctl-${{ matrix.os_arch }}.tar.gz
+          name: flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }}
+          path: flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.zip }}
```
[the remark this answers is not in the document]
mangelajo: @gshilin-sdb

internal/cli/certificate.go
```diff
@@ -49,6 +49,17 @@ type CertificateOptions struct {
 	SignerName string
 }
 
+type enrollmentService struct {
+	baseclient.Config
+
+	// EnrollmentUIEndpoint is the address of the device enrollment UI
+	EnrollmentUIEndpoint string `json:"enrollment-ui-endpoint,omitempty"`
```
mangelajo, line 56: This will be a problem eventually if the config format is changed @lkatalin *

What gregory is trying to fix here is that otherwise the CLI imported the agent, pulling in a lot of agent code that then needed windows compatibility.

Could be beneficial to move those structures into a common place outside of the agent tree?

internal/agent/device/fileio/fileio_unix.go
```diff
@@ -0,0 +1,110 @@
+//go:build !windows
+
+package fileio
```
mangelajo, line 3: Is all the windows/file compatibility here necessary after the change in cli/certificate?

hexfusion: fileio is internal to the agent, was something in the cli depending on it?

internal/cli/certificate.go
lkatalin: I think I see the goal and potential issue, and agree with @mangelajo it would be best if the agent config is pulled in both here and to the `internal/agent` from a shared location that is separated from the other `agent` code into its own package.

.github/workflows/publish-cli-bins.yaml
gshilin-sdb: fixed

internal/agent/device/fileio/fileio_unix.go
gshilin-sdb: yes, removed it

internal/cli/certificate.go
gshilin-sdb: Done

mangelajo:
It looks better now, please make @avishayt and @hexfusion aware of the file movement, please note that that is done to avoid the cli pulling agent code for compilation where that is not necessary.

internal/agent/device/console/console_linux.go
```diff
@@ -0,0 +1,261 @@
...
+		// finally this should end the other forward function
+		_ = stream.CloseSend()
+		c.log.Info("startForwarding: closing stream for console")
+
+	}()
+	g, _ := errgroup.WithContext(ctx)
+
+	g.Go(func() error {
+		defer stdout.Close() // close the other side to make the other forward function leave
+		defer c.log.Debug("stream > bash: leaving forward loop")
```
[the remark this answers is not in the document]
gshilin-sdb: this is a standard way fro shells to redirect input/output

internal/agent/device/fileio/writer.go
```diff
@@ -187,37 +174,6 @@ func (w *writer) overwriteFileWithRandomData(file string) error {
 
 // writeFileAtomically uses the renameio package to provide atomic file writing, we can't use renameio.WriteFile
 // directly since we need to 1) Chown 2) go through a buffer since files provided can be big
-func writeFileAtomically(fpath string, b []byte, dirMode, fileMode os.FileMode, uid, gid int) error {
```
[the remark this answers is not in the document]
gshilin-sdb: already not

internal/types/config.go
```diff
@@ -0,0 +1,281 @@
+package types
```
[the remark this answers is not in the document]
gshilin-sdb: CLI uses these types to create certs (createEmbeddedConfig and createReferenceConfig). Including it directly from agent also includes the **whole** agent package. Which uses linux-special code to work with filesystem and with a console. So there are just two ways: to implement a windows-only mocks for all those functions or to make these types shared. After some discussion the second way was preferred.

internal/agent/device/fileio/writer.go
```diff
@@ -133,17 +130,7 @@ func (w *writer) copyFile(src, dst string) error {
...
-		return fmt.Errorf("failed to retrieve UID and GID")
-	}
-
-	// set file ownership
-	if err := os.Chown(dstTarget, int(stat.Uid), int(stat.Gid)); err != nil {
-		return fmt.Errorf("failed to set UID and GID: %w", err)
-	}
-
-	return nil
+	return setChown(srcFileInfo, dstTarget)
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

hexfusion:
from the agent perspwctive could we just move

```
type EnrollmentService struct {
	client.Config

	// EnrollmentUIEndpoint is the address of the device enrollment UI
	EnrollmentUIEndpoint string `json:"enrollment-ui-endpoint,omitempty"`
}

type ManagementService struct {
	client.Config
}

func (s *EnrollmentService) Equal(s2 *EnrollmentService) bool {
	if s == s2 {
		return true
	}
	return s.Config.Equal(&s2.Config) && s.EnrollmentUIEndpoint == s2.EnrollmentUIEndpoint
}

func (s *ManagementService) Equal(s2 *ManagementService) bool {
	if s == s2 {
		return true
	}
	return s.Config.Equal(&s2.Config)
}

avishayt:
@gshilin-sdb Is this still relevant?

PR #807, opened 2025-01-20
NO-ISSUE: device/application: add single container provider

hexfusion:
wip


fzdarsky:
Instead of introducing an additional provider just for the single container case, is there a way to model this as an additional application format supported by the ImageApplicationProvider? Basically, if the image referenced by an ImageField does not contain the app package label, then assume it's a single app container.

Obviously, there are more params that this type of "app package" would need besides env vars. Could these be considered "extra args" that are just passed to Podman (or podman-compose or helm or ...)?

hexfusion:
Initial conversation continued in gdoc https://docs.google.com/document/d/1R8b28cLoLsv7L-kGLllUgtatjRXReVqorYPGr6qaPFk/edit?tab=t.0

PR #808, opened 2025-01-20
EDM-707: device/applications/compose: enforce service container_name is not supported

hexfusion:
Podman dynamically generates container names to avoid conflicts. However, hardcoding a container_name in the compose spec can cause unexpected issues if shared between applications.

This PR introduces pre-update validation to return an error when a compose spec contains a hardcoded container_name.

## Summary by CodeRabbit

- **Improvements**
	- Updated JSON serialization for compose service definitions.
	- Added validation to prevent hard-coded container names.

- **Logging**
	- Enhanced debug logging for Podman event monitoring, providing more visibility into boot-time event replay.

### DIFF
--- a/internal/agent/client/compose.go
+++ b/internal/agent/client/compose.go
@@ -17,11 +17,12 @@ import (
 const defaultPodmanTimeout = 2 * time.Minute
 
 type ComposeSpec struct {
-	Services map[string]ComposeService `yaml:"services"`
+	Services map[string]ComposeService `json:"services"`
 }
 
 type ComposeService struct {
-	Image string `yaml:"image"`
+	Image         string `json:"image"`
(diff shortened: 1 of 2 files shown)


internal/agent/client/compose.go
```diff
@@ -17,11 +17,12 @@ import (
...
 
 type ComposeSpec struct {
-	Services map[string]ComposeService `yaml:"services"`
+	Services map[string]ComposeService `json:"services"`
 }
 
 type ComposeService struct {
-	Image string `yaml:"image"`
+	Image         string `json:"image"`
+	ContainerName string `json:"container_name"`
```
hexfusion, line 25: note: serialization would fail using yaml but works fine using json which the module supports both.

DakCrowder (approved):

PR #810, opened 2025-01-21
NO-ISSUE: make packit replace -rc to ~rc in tags

mangelajo:
## Summary by CodeRabbit

- **Chores**
	- Updated package configuration to improve version retrieval and tag formatting.
	- Modified version tag template to remove leading 'v'.
	- Added a new action to retrieve the current package version.
	- Enhanced GitHub Actions workflow to fetch complete history for all branches and tags.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -10,7 +10,9 @@ packages:
     upstream_package_name: flightctl
     downstream_package_name: flightctl
     upstream_tag_template: v{version} # remove v from the start of the version tags (vx.y.z)->x.y.z
-
+actions:
+  get-current-version:
+    - bash -c 'git describe --tags | sed "s/^v//; s/-/~/g"'
 jobs:
   - job: copr_build
     trigger: pull_request
--- a/.github/workflows/pr-e2e-testing.yaml
+++ b/.github/workflows/pr-e2e-testing.yaml
@@ -18,6 +18,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - uses: dorny/paths-filter@v3
         id: filter


.packit.yaml
```diff
@@ -10,6 +10,9 @@ packages:
     upstream_package_name: flightctl
     downstream_package_name: flightctl
     upstream_tag_template: v{version} # remove v from the start of the version tags (vx.y.z)->x.y.z
+actions:
+  get-current-version:
+    - bash -c 'git describe --tags | sed "s/-/~/"'
```
asafbss, line 15: v prefix (e.g., v0.4.0-rc1) ?

mangelajo: oh true!!!
good catch"

asafbss: Hey Miguel, should we add something like this to our prepare_sources step to dynamically update the .spec file version during the build?

mangelajo: what prepare sources step?

mangelajo: do you mean Downstream?

asafbss: I think the prepare_sources action in the .packit.yaml configuration would dynamically update the Version: field in the .spec file before the SRPM is built.
something like:
```
prepare_sources:
  - bash -c 'git describe --tags | sed "s/^v//; s/-/~/" | xargs -I {} rpmdev-bumpspec --new {} --comment "Packit automated version update" packaging/rpm/flightctl.spec'
```

But I don't know if this is necessary

mangelajo: This is isn't necessary as far as I tested.

mangelajo: Actually pack it will do it on the spec for you. We just don't keep it or maintain it on the .spec. We could chose to upgrade the way we work with packit anytime.

For the release I think we can merge it as it is. I tested it locally and works.

asafbss (approved):

PR #815, opened 2025-01-22
EDM-885: RPM build 

gshilin-sdb:
do not cache git version in workflow
in spec user proper values for SOURCE_GIT_TAG et al
added fedora 41 targets

## Summary by CodeRabbit

- **Build Configuration**
	- Added support for Fedora 41 (x86_64 and aarch64) in build targets
	- Updated upstream tag format to remove leading 'v'

- **Packaging**
	- Modified build process to improve version and commit hash extraction

- **Repository Management**
	- Added `.rpm` files to `.gitignore`
	- Updated GitHub Actions workflow by removing Git repository caching step
	- Enhanced verification steps for CLI binaries to include YAML output format


packaging/rpm/flightctl.spec
```diff
@@ -53,7 +53,8 @@
     if [[ ! -f "${GOENVFILE}" ]]; then
         export GOPROXY='https://proxy.golang.org,direct'
     fi
-    SOURCE_GIT_TAG=%{release} SOURCE_GIT_TREE_STATE=clean SOURCE_GIT_COMMIT=%{release} SOURCE_GIT_TAG_NO_V=%{version} make build-cli build-agent
+
+    SOURCE_GIT_TAG=%{version} SOURCE_GIT_TREE_STATE=clean SOURCE_GIT_COMMIT=$(echo %{version} | awk -F'~g' '{print $2}') SOURCE_GIT_TAG_NO_V=%{version} make build-cli build-agent
```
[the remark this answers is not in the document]
gshilin-sdb: unfortunately git doesn't always present

packaging/rpm/flightctl.spec
```diff
@@ -14,7 +14,7 @@ Summary:        Flightctl is a manager of the edge device fleets.
 License:        Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND ISC AND MIT
 URL:            %{gourl}
 
-Source0:        %{name}-%{version}.tar.gz
+Source0:        flightctl-0.4.0~rc1~11~ga05a4957.tar.gz
```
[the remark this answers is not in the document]
gshilin-sdb: done

packaging/rpm/flightctl.spec
```diff
@@ -14,7 +14,7 @@ Summary:        Flightctl is a manager of the edge device fleets.
 License:        Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND ISC AND MIT
 URL:            %{gourl}
 
-Source0:        %{name}-%{version}.tar.gz
+Source0:         %{name}-%{version}.tar.gz
```
fzdarsky, line 17: Accidental space

mangelajo (approved):
LGTM

packaging/rpm/flightctl.spec
```diff
@@ -53,7 +53,12 @@ Flightctl Agent is a component of the flightctl tool.
     if [[ ! -f "${GOENVFILE}" ]]; then
         export GOPROXY='https://proxy.golang.org,direct'
     fi
-    SOURCE_GIT_TAG=%{release} SOURCE_GIT_TREE_STATE=clean SOURCE_GIT_COMMIT=%{release} SOURCE_GIT_TAG_NO_V=%{version} make build-cli build-agent
+
+    SOURCE_GIT_TAG=%{version} \
+    SOURCE_GIT_TREE_STATE=clean \
+    SOURCE_GIT_COMMIT=$(echo %{version} | awk -F'~g' '{print $2}') \
+    SOURCE_GIT_TAG_NO_V=%{version} \
```
fzdarsky, line 60: Why not use the logic in the Makefile to set these vars? Having a single authoritative place seems to be the only way to ensure consistency. Concretely:
* The `%{version}` needs to be a function of the $SOURCE_GIT_TAG, not vice-versa (otherwise that commit isn't guaranteed to exist).
* We cannot use the same `%{version}` for SOURCE_GIT_TAG and SOURCE_GIT_TAG_NO_V, because the former includes the "v" prefix.
* SOURCE_GIT_COMMIT only works if the `%{version}` contains a git commit, but released versions would not!
* SOURCE_GIT_TREE_STATE is hardcoded to "clean", but if the tree contains changes it must be marked as "dirty".

fzdarsky (dismissed):

packaging/rpm/flightctl.spec
gshilin-sdb: Makefile uses git, while we don't have it. So we try to invent it from other sources, in this case from spec itself, i.e. that's what packit supplied us as it had access to git beforehand. This "invented" version is "v0.4.0-rc1-17-g469d9365", so it doesn't have the "v" prefix. There are no way to invent clear/dirty, but these RPMs are built for "release" (excluding two experimental ones built for PR), so they expected to be clean. Binaries built for PR got access to git, so there is a true information.

packaging/rpm/flightctl.spec
gshilin-sdb: fixed

packaging/rpm/flightctl.spec
mangelajo: It's not going to be a 1:1 with other compilations and targets, but it should work. This should keep us more or less consistent with the copr builds.

GH cli bins and downstream rpms shall be good.

mangelajo (approved):

PR #816, opened 2025-01-23
EDM-1052: Return full resources upon update

avishayt:
This adds RETURNING clauses to Update, CreateOrUpdate, UpdateStatus calls so that the full updated resource is returned from the DB.

Also changes Fleet UpdateStatus to use the generic code.

## Release Notes

- **Refactor**
	- Simplified fleet status update logic by centralizing update operations.
	- Enhanced generic store update methods to return updated resources more efficiently.

- **Bug Fixes**
	- Improved error handling during resource updates.
	- Ensured proper management of labels and annotations during updates.

- **Tests**
	- Updated test suite to validate fleet status update functionality.
	- Enhanced test coverage for fleet creation and update scenarios, including verification of API resource correctness.


internal/store/generic.go
```diff
@@ -328,12 +329,20 @@ func (s *GenericStore[P, M, A, AL]) UpdateStatus(ctx context.Context, orgId uuid
 		return nil, err
 	}
 
-	result := s.db.WithContext(ctx).Model(&model).Where("org_id = ? AND name = ?", orgId, model.GetName()).Updates(
+	result := s.db.WithContext(ctx).Model(model).Where("org_id = ? AND name = ?", orgId, model.GetName()).Clauses(clause.Returning{}).Updates(
```
ori-amizur, line 332: Shouldn't the model be passed as a pointer in order for it to be updated with the results?

ori-amizur (approved):

PR #817, opened 2025-01-23
EDM-1053: add helm chart tag 0-latest

mparram:
To keep the staging environment updated with the latest built version, the Helm repository needs a label that always points to the latest version, and the best place to define it is within the GitHub Actions workflow itself.
Since Helm interprets tags in a specific way, they must start with a number. Therefore, instead of using "latest", it needs to be named something like "0-latest"

## Summary by CodeRabbit

- **Chores**
	- Updated Helm chart publishing workflow to include explicit version tagging when pushing to OCI registry.


mangelajo (approved):
looks reasonable to me, we can try!

.github/workflows/publish-containers.yaml
```diff
@@ -79,6 +79,7 @@ jobs:
       - name: Push helm charts
         run: |
           helm push "flightctl-${{ needs.generate-tags.outputs.helm_tag }}.tgz" oci://${{ env.QUAY_CHARTS }}/
+          helm push "flightctl-${{ needs.generate-tags.outputs.helm_tag }}.tgz" oci://${{ env.QUAY_CHARTS }}/ --version 0-latest
```
mangelajo, line 82: https://github.com/jumpstarter-dev/jumpstarter-controller/pull/104

That push parameter doesn't seem to exist, where did you see it?

avishayt:
@mparram Is this PR still relevant?

mparram:
No need anymore, I fixed it by creating a custom ArgoCD plugin

PR #828, opened 2025-01-27
EDM-323: Ensure CSR Name mangling is not done at renewal

kot-begemot-uk:
Change certificate CN logic to ensure it does not mangle names in renewal requests.

This is the bare minimum to make the CSR API usable for renewals. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced certificate common name generation with support for multiple prefixes.
	- Improved flexibility in bootstrap common name processing.


avishayt (approved):

internal/crypto/cert.go
```diff
@@ -22,7 +23,15 @@ const ClientBootstrapCommonNamePrefix = "client-enrollment-"
 const AdminCommonName = "flightctl-admin"
 const DeviceCommonNamePrefix = "device:"
 
+var Prefixes = [...]string{ClientBootstrapCommonNamePrefix, DeviceCommonNamePrefix}
```
hexfusion, line 26: question: does this need to be exported?

kot-begemot-uk: Now - now. 

Gut feeling is that it will be needed in the CA shims as there are differences in the way a proper CA handles renew and first issue.

PR #832, opened 2025-01-27
EDM-240: applications: add inline provider

hexfusion:
```yaml
    spec:
      applications:
        - name: my-app
          type: compose
          inline:
            docker-compose.yaml:
              content: |
                version: "3.8"
                services:
                  app:
                    image: "docker.io/nginx:latest"
                    ports:
                      - "80:80"
```

design doc: https://docs.google.com/document/d/1oeHjScxmpddYMcHzuqUzoOjIo4uYmKbqeSxVOTkv-54/edit?tab=t.0#heading=h.xopq9msd32jw


celdrake:
@hexfusion Do we have a Jira for this task? I took a quick look, and it will require some changes in the UI as now we assume there will be an `app.image`, and we'd need to make sure we don't incorrectly overwrite an inline app due to its different spec. Ideally we could have also a Jira task for the UI to track this.

hexfusion:
The main issue is we have not agreed on which path to move forward with so the work will not make TP. But EDM-240 is the parent.

[1] https://issues.redhat.com/browse/EDM-240

PR #834, opened 2025-01-28
EDM-925: T17: Do not only rely on client-side authorization

gshilin-sdb:
Each and every user-faced endpoint must have an authorization and access permissions verification.
A middleware `Permissions` was created to perform this verification.
Verifications were removed from handlers.

Open problems (unnecessary for this PR):
* PUT (i.e. Update) actions - perform additional validations. These validations also should be performed in middleware

## Release Notes

- **Security Changes**
	- Removed authorization checks across multiple service methods, simplifying permission handling for various API resources.

- **API Specification**
	- Updated OpenAPI specification with new schema definitions for device decommissioning and patch request handling.
	- Minor correction in certificate signing request status description.

- **Middleware**
	- Introduced new permissions middleware to enhance request processing.

- **Documentation**
	- Corrected URL structure for API endpoints related to template versions for fleets.
	- Updated `.gitignore` to exclude `.tar` files from version control.

The changes primarily focus on modifying the authorization model and improving API specification clarity.


internal/auth/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+	case 3: // /api/v{version}/resources
+		switch r.Method {
+		case http.MethodGet:
+			action = "list"
+		case http.MethodDelete:
+			action = "deletecollection"
+		}
+	case 4: // /api/v{version}/resources/name
+		// No changes
+	case 5: // /api/v{version}/resources/name/sub-action
```
ori-amizur, line 93: What about 6?
Example: https://github.com/flightctl/flightctl/blob/175ae0484d70763bd78d621b3c6c683b0acca0cd/api/v1alpha1/openapi.yaml#L3108

internal/auth/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+		action = "update"
+	case http.MethodPatch:
+		action = "patch"
+	case http.MethodDelete:
+		action = "delete"
+	default:
+		return "", ""
+	}
+	switch len(parts) {
+	case 3: // /api/v{version}/resources
```
ori-amizur, line 84: It seems the counting is not correct.  Since the path starts with "/", then the first part will be an empty string.

internal/auth/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+
+			// If authorized, proceed to the next handler
+			next.ServeHTTP(w, r)
+		}
+		return http.HandlerFunc(fn)
+	}
+}
+
+func extractResourceAndAction(r *http.Request) (string, string) {
+	// Handle according to the URL structure
```
ori-amizur, line 58: In general template versions is not handled correctly

internal/auth/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+			if len(parts) <= 2 ||
+				parts[2] == "auth" {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			// Extract resource and action from the request
+			resource, action := extractResourceAndAction(r)
+			if resource == "" || action == "" {
+				http.Error(w, BadRequest, http.StatusServiceUnavailable)
```
ori-amizur, line 28: I would say to log it.  It might mean that the extraction logic failed to find the resource and action

internal/auth/permissions.go
```diff
@@ -0,0 +1,99 @@
...
+				return
+			}
+
+			// TODO [gshilin]: is there a way to handle exceptions automatically?
+			// Maybe we can add x-skip-permission-middleware to openapi.yaml?
+			// Or maybe we don't need this exception and can change the (GET /api/v1/enrollmentconfig) handler instead?
+			if resource == "certificatesigningrequests" && action == "get" {
```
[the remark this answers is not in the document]
gshilin-sdb: Verification there is done under condition, so I'm not sure whether it can be performed in middleware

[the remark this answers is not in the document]
ori-amizur: So is it valid to skip it? If not were will it be verified ?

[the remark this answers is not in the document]
gshilin-sdb: It was verified only if some change was required, but not always. So that's not an exception anymore.

internal/auth/permissions.go
gshilin-sdb: added handling

internal/auth/permissions.go
gshilin-sdb: refurbished parts, now it's correct

internal/auth/permissions.go
gshilin-sdb: fixed

internal/auth/permissions.go
gshilin-sdb: done

internal/auth/permissions.go
```diff
@@ -0,0 +1,106 @@
...
+		case http.MethodDelete:
+			action = "deletecollection"
+		}
+	case 2: // resources/{name}
+		// No changes
+	case 3:
+		// resources/{name}/sub-action
+		subAction := strings.ToLower(parts[2])
+		resource += "/" + subAction
+		if subAction == "templateversions" && action == "delete" {
```
ori-amizur, line 96: Isn't it generic? Why should we mention templateversions here ?

internal/auth/permissions.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		http.Error(w, AuthorizationServerUnavailable, http.StatusServiceUnavailable)
+		return false
+	}
+	if !allowed {
+		http.Error(w, Forbidden, http.StatusForbidden)
+		return false
+	}
+
+	return true
```
hexfusion, line 60: not: its good defensive practice to not assume true and the end of a function returning a bool.  especially since we are in the auth package.

we can do something like this 
```
if allowed {
    return true
} else {
   http.Error(w, Forbidden, http.StatusForbidden)
   return false
}

return false
 ```

File: internal/auth/permissions.go [hunk 8]
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func extractResourceAndAction(parts []string, method string) (string, string) {
+	// Handle according to the URL structure
+	// e.g., "device", "devices", "devices/{name}", "devices/{name}/sub-action"
+	resource := strings.ToLower(parts[0])
+	action := ""
+	switch method {
+	case http.MethodGet:
+		action = "get"
```
hexfusion, line 70: Could we make these actions consts to reduce scope of future typos?

internal/auth/permissions.go
```diff
@@ -0,0 +1,106 @@
...
+			action = "deletecollection"
+		}
+	case 4:
+		// resources/{name}/sub-action/{name}
+		subAction := strings.ToLower(parts[2])
+		resource += "/" + subAction
+	}
+
+	return resource, action
+}
```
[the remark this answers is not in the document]
gshilin-sdb: This covers all others as it's an error and we shouldn't accept them: return "", ""

internal/auth/permissions.go
gshilin-sdb: right

internal/auth/permissions.go
gshilin-sdb: fixed

internal/auth/permissions.go
gshilin-sdb: done

avishayt (approved):

avishayt (approved):

PR #843, opened 2025-01-29
EDM-1070: It's not possible to update a fleet setting only disruption budget without device selection

ori-amizur:
In the validation, rollout policy was expected to contain device selection.  Otherwise, an internal server error occured. This fixes the issue and adds some other missing validations.

## Summary by CodeRabbit

- **New Features**
  - Enhanced validation logic for various system components.
  - Added comprehensive validation checks for batches, batch sequences, device selections, disruption budgets, and rollout policies.
  - Improved error handling and validation consistency across different structures.
  - Updated OpenAPI specification to ensure at least one device must be available during rollouts.


api/v1alpha1/validation.go
```diff
@@ -384,25 +384,68 @@ func (r CertificateSigningRequest) Validate() []error {
...
+
+func (d *DisruptionBudget) Validate() []error {
+	if d == nil {
+		return nil
+	}
+	if d.MinAvailable == nil && d.MaxUnavailable == nil {
+		return []error{errors.New("at least one of [MinAvailable, MaxUnavailable] must be defined in disruption budget")}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
ori-amizur: Simultaneous use of both is valid

api/v1alpha1/validation.go
```diff
@@ -384,25 +384,68 @@ func (r CertificateSigningRequest) Validate() []error {
...
+		}
+	}
+	return errs
+}
+
+func (d *DisruptionBudget) Validate() []error {
+	if d == nil {
+		return nil
+	}
+	if d.MinAvailable == nil && d.MaxUnavailable == nil {
```
celdrake, line 423: I noticed that `MinAvailable` is allowed to be 0, whereas `MaxUnavailable` is allowed to be at least 1.

Maybe it would be best to keep the minimum also at least to be 1, otherwise it should be unset?

celdrake (commented):
Tested using the new UI for Rollout policies / Disruption budget and everything works well. Just left a comment regarding "disruptionBudget.minAvailable", about what's its minimum allowed value.

api/v1alpha1/validation.go
ori-amizur: There is a difference if they are defined, and if defined what are their allowed values.
For MinAvailable the minimum allowed is 0 - which means that if there is a single device in a group it can still be rolled out using disruption budget.
For MaxUnavailable the minimum is 1 - which means that we do not allow to define a rule that will make roll out impossible.
Anyway, here we just check that at least one of them is defined if disruption budget is defined.

celdrake: But in the end, setting `minAvailable` to 0 is not applying any restriction, IIUC the system would behave the same as if it was `nil`.

Then it would be possible to define a DB that has only `groupBy` + `minAvailable` which I assume would mean the system will calculate the groups, with no other effect.

Anyway, in the UI we will also accept  `minAvailable=0` so we are aligned to the API.

avishayt: I agree with @celdrake - what is the difference between these two disruption budgets?
{}
{minAvailable: 0}

celdrake: Changed the UI to restrict `minAvailable` >= 1 as well.

avishayt (approved):

PR #846, opened 2025-01-29
Update redis config

DakCrowder:


pkg/kvconfig/kvconfig.go
```diff
@@ -0,0 +1,67 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to configure TLS for redis: %w", err)
+		}
+		options.TLSConfig = tlsConfig
+	}
+
+	return options, nil
+}
+
+func loadTLSConfig(cfg *KvConfig) (*tls.Config, error) {
```
DakCrowder, line 42: @coderabbitai Generate unit testing code for this file

DakCrowder:
@coderabbitai review

internal/config/config.go
```diff
@@ -16,11 +17,11 @@ const (
...
 
 type Config struct {
-	Database   *dbConfig         `json:"database,omitempty"`
-	Service    *svcConfig        `json:"service,omitempty"`
-	KV         *kvConfig         `json:"kv,omitempty"`
-	Auth       *authConfig       `json:"auth,omitempty"`
-	Prometheus *prometheusConfig `json:"prometheus,omitempty"`
+	Database   *dbConfig          `json:"database,omitempty"`
+	Service    *svcConfig         `json:"service,omitempty"`
+	KV         *kvconfig.KvConfig `json:"kv,omitempty"`
```
hexfusion, line 22: maybe we could make the package more general so could do this.
```suggestion
	KV         *kv.Config `json:"kv,omitempty"`
```

avishayt:
While I think this is a fine addition, I have two questions:
1. Should we change the deploy scripts to take advantage of these options?
2. Did we decide that we will use AAP's Redis deployment instead of having our own?  Is their configuration suitable?  Are both teams willing to take the risk of one affecting the other?

DakCrowder:
@avishayt 
1.  Yes, I was poking around with this last night.  I wasn't planning on doing tls on the first pass but would like to get some of the other options in the template.
2.  Great question, short answer is no nothing concrete has been decided and there are some higher level discussions that have been raised in the past couple days about how flightctl fits in with AAP when installed/deployed.  I'm hoping some of those discussions will be completed soon and we can get more clarity there / have enough info for us to make decisions.

pkg/kv/kv_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
+			options, err := ConfigToRedisOptions(testCase.cfg)
+			if testCase.expectedErr != nil {
+				assert.ErrorIs(t, err, testCase.expectedErr)
+				return
+			}
+			require.NoError(t, err)
+
+			assert.Equal(t, fmt.Sprintf("%s:%d", testCase.cfg.Hostname, testCase.cfg.Port), options.Addr)
```
hexfusion, line 105: nit: we should standardize on require vs assert.

you can do this

```
require := require := require.New(t)
// then no need to pass t as a param
require.NotNil(options.TLSConfig.RootCAs)
```

avishayt:
Is this PR still relevant?

DakCrowder:
@avishayt No - closing it.  Probably a good idea to do at some point but definitely lower priority than other work atm and this is quite out of date.

PR #848, opened 2025-01-29
EDM-811: Enable status.lifecycle.status field selector for devices

lkatalin:
This is needed for the UI to show relevant views of devices ("live" only, decommissioning/ed only).

## Summary by CodeRabbit

- **Documentation**
	- Updated field selectors documentation for "Device" resource.
	- Added new field `status.lifecycle.status` for filtering devices by lifecycle status.

- **New Features**
	- Enhanced device filtering capabilities with lifecycle status selection.

### DIFF
--- a/internal/store/model/selectors.go
+++ b/internal/store/model/selectors.go
@@ -19,6 +19,7 @@ var (
 		selector.NewSelectorName("status.applicationsSummary.status"): selector.String,
 		selector.NewSelectorName("status.updated.status"):             selector.String,
 		selector.NewSelectorName("status.lastSeen"):                   selector.Timestamp,
+		selector.NewSelectorName("status.lifecycle.status"):           selector.String,
 	}
 	fleetSpecSelectors = selectorToTypeMap{
 		selector.NewSelectorName("spec.template.spec.os.image"): selector.String,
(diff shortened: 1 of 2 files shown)


internal/store/model/selectors.go
```diff
@@ -19,6 +19,7 @@ var (
 		selector.NewSelectorName("status.applicationsSummary.status"): selector.String,
 		selector.NewSelectorName("status.updated.status"):             selector.String,
 		selector.NewSelectorName("status.lastSeen"):                   selector.Timestamp,
+		selector.NewSelectorName("status.lifecycle.status"):           selector.String,
```
asafbss, line 22: Perfect @lkatalin!
You might want to consider hiding it from discovery using `selector.NewHiddenSelectorName`

asafbss (approved):

internal/store/model/selectors.go
lkatalin: Hmm. Do we not want it to be discoverable? What are the implications?

asafbss: It will be documented and users will be able to see it listed as a selector

celdrake: API / CLI users may also want to filter out the Decommissioning devices when they are interested in their "live" devices.

Specially if we keep the labels and owner field of those devices, as they would meet the definition of belonging to a fleet etc.

celdrake:
@lkatalin Tested the filter, and I can see it working correctly for GETting devices list, as well as when only the status breakdown is requested (using `summaryOnly=true`)

If we stop clearing the "labels" and "owner" field of Decom devices, I'd need to double-check the filtering in other requests, but I imagine those would also work correctly.

Thanks!

lkatalin:
Thanks for testing it out @celdrake !

PR #857, opened 2025-02-03
EDM-TBD: Bug fix for setting lifecycle to Enrolled when using agent-vm

lkatalin:
Adds enrollment lifecycle status on device-side.

## Summary by CodeRabbit

- **New Features**
  - Devices now automatically update their lifecycle status to "Enrolled" during initialization, ensuring a current enrollment state.
  - Enhanced support for updating lifecycle status details provides clearer device status information for improved tracking.

### DIFF
--- a/internal/agent/device/lifecycle/manager.go
+++ b/internal/agent/device/lifecycle/manager.go
@@ -11,8 +11,9 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/client"
 	"github.com/flightctl/flightctl/internal/agent/device/errors"
 	"github.com/flightctl/flightctl/internal/agent/device/fileio"
-	"github.com/flightctl/flightctl/internal/agent/device/status"
+	fcstatus "github.com/flightctl/flightctl/internal/agent/device/status"
 	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/samber/lo"
 	"github.com/skip2/go-qrcode"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/cert"
@@ -39,7 +40,7 @@ type LifecycleManager struct {
 	enrollmentClient client.Enrollment
 	defaultLabels    map[string]string
 	enrollmentCSR    []byte
-	statusManager    status.Manager
+	statusManager    fcstatus.Manager
 	systemdClient    *client.Systemd
 
 	backoff wait.Backoff
@@ -56,7 +57,7 @@ func NewManager(
 	enrollmentClient client.Enrollment,
 	enrollmentCSR []byte,
 	defaultLabels map[string]string,
-	statusManager status.Manager,
+	statusManager fcstatus.Manager,
 	systemdClient *client.Systemd,
 	backoff wait.Backoff,
 	log *log.PrefixLogger,
@@ -95,6 +96,14 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
 		if err != nil {
 			return err
 		}
+	}
+
+	_, updateErr := m.statusManager.Update(ctx, fcstatus.SetDeviceLifecycleStatus(v1alpha1.DeviceLifecycleStatus{
+		Status: "Enrolled",
+		Info:   lo.ToPtr("Device was enrolled"),
+	}))
+	if updateErr != nil {
+		m.log.Warnf("Failed setting initial 'enrolled' status: %v", updateErr)
 	}
 
 	// write the management banner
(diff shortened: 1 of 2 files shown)


internal/agent/device/lifecycle/manager.go
```diff
@@ -97,6 +98,14 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
 		}
 	}
 
+	_, updateErr := m.statusManager.Update(ctx, fcstatus.SetDeviceLifecycleStatus(v1alpha1.DeviceLifecycleStatus{
+		Status: "Enrolled",
+		Info:   lo.ToPtr("Device was enrolled"),
```
hexfusion, line 103: nit: please use util.FromPtr

internal/agent/device/lifecycle/manager.go
```diff
@@ -11,8 +11,9 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/client"
 	"github.com/flightctl/flightctl/internal/agent/device/errors"
 	"github.com/flightctl/flightctl/internal/agent/device/fileio"
-	"github.com/flightctl/flightctl/internal/agent/device/status"
+	fcstatus "github.com/flightctl/flightctl/internal/agent/device/status"
```
hexfusion, line 14: just curious why the prefix?

internal/agent/device/lifecycle/manager.go
```diff
@@ -97,6 +98,14 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1alpha1.Devi
 		}
 	}
 
+	_, updateErr := m.statusManager.Update(ctx, fcstatus.SetDeviceLifecycleStatus(v1alpha1.DeviceLifecycleStatus{
+		Status: "Enrolled",
```
hexfusion, line 102: This should be a const to eliminate future typos but I am trying to understand why this is necessary. The service knows if the agent is enrolled already and the agent could lie if it wanted to.

hexfusion: The main question is who owns this status? Should it be the agent or the service.

internal/agent/device/lifecycle/manager.go
lkatalin: Looks like that was removed in [this commit](https://github.com/flightctl/flightctl/commit/5198e4cad6e32fd7d0d118fe6013f9721df732a3)

internal/agent/device/lifecycle/manager.go
lkatalin: There was some confusion around `status` as a module vs. `status` as a variable - let me see if it was my confusion vs. the compiler's, though. If it works without the prefix I can remove it.

internal/agent/device/lifecycle/manager.go
lkatalin: This is a good point re: owning the status. We did decide the service would own it. However, it is getting wiped for some reason even after being set on the service side - I'm not completely sure but I think this may be because a blank status is created for the device during bootstrap if one does not exist already. Maybe a better solution would be to make sure the service-side status gets passed to the agent upon enrollment (so there is never a case where a device status doesn't exist)? I'm open to suggestions.

The related bug is [this one](https://redhat-internal.slack.com/archives/C01KER8QV33/p1738313965026339). This change did fix the issue for me when I tested it.

hexfusion: https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/constructors.go#L29-L31 this should be removed so the status is not sent from the agent at all.

internal/agent/device/lifecycle/manager.go
hexfusion: sigh

internal/agent/device/lifecycle/manager.go
lkatalin: That code is being used on the server-side, though.

lkatalin: This is where the device status gets set initially on the service side: https://github.com/flightctl/flightctl/blob/main/internal/service/enrollmentrequest.go#L81 . But I do not know whether that initial "device status" is ever sent to the agent - there isn't a reason to do it as the status is normally sent from agent to service. But in this case I think the "new" device status is spawned again on the device side and overwrites the previous status.

hexfusion: it's also being used by the agent in this case I would unset it after here[1] then.

[1] https://github.com/flightctl/flightctl/blob/dfc4b3ba042b21b5824082f29803dc77028f8a62/internal/agent/device/status/status.go#L22

lkatalin:
@hexfusion Actually, I think the better fix is #858 . If you agree, I'll close this one.

PR #858, opened 2025-02-04
EDM-TBD: Bug fix to ensure enrolled device lifecycle status is not reset

lkatalin:
Without this change, a newly initialized device status from the agent may overwrite the server's lifecycle status of `enrolled` for an enrolled device.


celdrake (commented):
LGTM. Tested, and I could see that since approving the device, it stays as "Enrolled", even after doing changes to its config, etc.

hexfusion (approved):

PR #860, opened 2025-02-04
EDM-390: Add diagram for device lifecycle statuses

lkatalin:
WIP


docs/user/device-api-statuses.md
```diff
@@ -209,6 +209,65 @@ stateDiagram
...
+
+The following state diagram shows the possible transitions between lifecycle statuses and states.
+
+```mermaid
+stateDiagram
+    classDef transientNotReported stroke-dasharray:5 5
+    class ActivatingConfig transientNotReported
+
+    classDef notYetSupported fill:white
+    class PoweringOff notYetSupported
```
celdrake, line 241: Do these PowerOff/On states belong here?

lkatalin: Likely not, good catch

docs/user/device-api-statuses.md
```diff
@@ -209,6 +209,57 @@ stateDiagram
...
+        direction LR
+
+        state Decommissioning {
+            [*] --> DecommissioningRequested
+            DecommissioningRequested --> DecommissioningStarted
+        }
+
+        state Decommissioned {
+            [*] --> DecommissioningCompleted
+            [*] --> DecommissioningError
```
fzdarsky: Match state names in table above:

```suggestion
            [*] --> Requested
            Requested --> Started
        }

        state Decommissioned {
            [*] --> Completed
            [*] --> Error
```

docs/user/references/device-api-statuses.md
```diff
@@ -209,6 +209,57 @@ stateDiagram
...
+| `Decommissioning` | The device has been requested to decommission by a user and is no longer available to carry out normal operations. | `device.spec.decommissioning != nil` |
+| `Decommissioned` | The device has either completed its decommissioning process or has encountered an unrecoverable error in doing so. | `∃ c ∈ device.status.conditions \| c.type == DeviceDecommissioning && c.status == true && c.reason ∈ {DecommissionStateComplete, DecommissionStateError}` |
+| `Unknown` | The device has not properly enrolled. No device available through management should be in this state. | `∄ er ∈ enrollmentrequest \| er[device] && er ∈{Complete}` |
+
+The `device.status.conditions` field may contain a Condition of type `DeviceDecommissioning` whose `Condition.Reason` field contains the current state of a decommissioning in progress and can contain the following values:
+
+| Decommissioning State | Description |
+| ------------ | ----------- |
+| `Started` | The agent has received the request to decommission from the service and will take a series of previously defined decommissioning actions. |
```
fzdarsky, line 242: Missing `Requested` state.

docs/user/device-api-statuses.md
```diff
@@ -209,6 +209,57 @@ stateDiagram
...
+
+The `device.status.lifecycle` field can have the following values:
+
+| Status | Description | Formal Definition<sup>1</sub> |
+| ------ | ----------- | ----------------------------- |
+| `Enrolled` | The device's Enrollment Request was approved and it will be able to connect to management to carry out normal operations using its management certificate. | `∃ er ∈ enrollmentrequest \| er[device] && er ∈{Complete}` |
+| `Decommissioning` | The device has been requested to decommission by a user and is no longer available to carry out normal operations. | `device.spec.decommissioning != nil` |
+| `Decommissioned` | The device has either completed its decommissioning process or has encountered an unrecoverable error in doing so. | `∃ c ∈ device.status.conditions \| c.type == DeviceDecommissioning && c.status == true && c.reason ∈ {DecommissionStateComplete, DecommissionStateError}` |
+| `Unknown` | The device has not properly enrolled. No device available through management should be in this state. | `∄ er ∈ enrollmentrequest \| er[device] && er ∈{Complete}` |
```
fzdarsky: Status and Description are ok. For the Formal Definition, I've been struggling to figure out whether from a boolean logic perspective the definition is such that the device must always be in exactly one of these statuses.

I've therefore tried to read the definition out of `updateServerSideLifecycleStatus`, but then noticed this function does not yet enforce any definition as contract; it just takes what the agent sends at face value.

What I think the definition _should_ look like (but this doesn't match implementation):
```suggestion
| `Enrolled` | The device's Enrollment Request was approved and it will be able to connect to management to carry out normal operations using its management certificate. | `enrollmentCompleted(device) && device.spec.decommissioning == nil` |
| `Decommissioning` | The device has been requested to decommission by a user and is no longer available to carry out normal operations. | `enrollmentCompleted(device) && device.spec.decommissioning != nil && (
  last.device.condition["DeviceDecommissioning"].status != "true"
  && (current.device.condition["DeviceDecommissioning"].status == "true" && current.device.condition["DeviceDecommissioning"].reason ∉ {Complete, Error})
)` |
| `Decommissioned` | The device has either completed its decommissioning process or has encountered an unrecoverable error in doing so. | `enrollmentCompleted(device) && device.spec.decommissioning != nil && (
  last.device.condition["DeviceDecommissioning"].status == "true"
  || (current.device.condition["DeviceDecommissioning"].status == "true" && current.device.condition["DeviceDecommissioning"].reason ∈ {Complete, Error})
)` |
| `Unknown` | The device has not properly enrolled. No device available through management should be in this state. | `!enrollmentCompleted(device)` |
```

Then the `enrollmentCompleted(device)` helper function could be defined as

```
// A device has completed enrollment if there exists an EnrollmentRequest record with the device's ID and that record contains a valid `er.status.certificate`.
enrollmentCompleted := enrollmentRequests[device.metadata.name].status.certificate != nil
```

fzdarsky: Does the `Unknown` state actually exist (i.e. it's reported on the API)? Otherwise, I'd suggest `NotEnrolled` as this better matches the diagram.

docs/user/references/device-api-statuses.md
lkatalin: These are meant to match our `const` values for the Decom State: https://github.com/flightctl/flightctl/blob/main/api/v1beta1/consts.go#L161 

Should we change the name of `Started` to `Requested`? Or add a new state to describe the time between when the agent receives the request and when it begins taking decommissioning actions?

docs/user/device-api-statuses.md
lkatalin: I agree, this should be `NotEnrolled`.

lkatalin: I like these changes and will update the PR.
How are we getting or defining `last.device.condition` and `current.device.condition`? Do these need to be defined as helper functions or are they predefined somewhere in the codebase that I'm forgetting?

docs/user/device-api-statuses.md
```diff
@@ -209,6 +209,57 @@ stateDiagram
...
+
+The `device.status.conditions` field may contain a Condition of type `DeviceDecommissioning` whose `Condition.Reason` field contains the current state of a decommissioning in progress and can contain the following values:
```
lkatalin: To clarify why there is no `Requested` state, I'm changing the wording here.
```suggestion
The `device.status.conditions` field may contain a Condition of type `DeviceDecommissioning` whose `Condition.Reason` field contains the current state of a device's progress on a request to decommission that  it has received. It can contain the following values:
```

docs/user/references/device-api-statuses.md
lkatalin: See above comment for some clarification.

docs/user/device-api-statuses.md
lkatalin: Oh I see, this is where `Requested` comes in.
The device has no real notion of `Requested`, only the service does. That is why it isn't included as an option in `Decommissioning State`. Should this change?

fzdarsky: No need, I think. Remember these diagrams do not describe the device agent's view but the service's view, which combines what the service knows about the device with what the agent reports / did not yet report. Concretely, the service knows the state is `Requested`, because it has updated the device spec to initiate decom, but the agent did not yet pick up that request and report it status change yet.

docs/user/device-api-statuses.md
fzdarsky: Yeah, it's more pseudo code than matching lines in the code. We've used something resembling a JSONPath as compromise between formality and legibility, so actually the path should probably be `device.status.condition` instead and maybe use `$last.` and `$currrent.` to denote the spec object being operated on.

docs/user/device-api-statuses.md
lkatalin: Makes sense - then I will not update the `Decommissioning State` table, below, as the states it shows are from the agent view (vs. this diagram, which is service view).

docs/user/device-api-statuses.md
lkatalin: That makes sense conceptually,  although it would be nice to know exactly how the service will interpret what "state" a device is in, so it seems to me we should add some implementation. I will look.

lkatalin: What is the difference between `last` and `current` in your definition? Edit: Oh I see, I think you are checking for a `DeviceDecommission` Condition status change from `false` to `true` (except that's now how it is implemented - we don't have any `DeviceDecommission` Condition present at all until the agent receives the request and starts the process).

~~Also - the agent never sets the `DeviceDecommission` Condition to anything other than `ConditionStatus = true` while it transitions through `Started` to `Complete` (or `Error`). See [ref](https://github.com/flightctl/flightctl/blob/main/internal/agent/device/lifecycle/manager.go#L165). Are you saying you want to change this implementation as well?~~

lkatalin: Based on the current implementation, it seems like we actually should have these definitions:

```
| `Decommissioning` | The device has been requested to decommission by a user and is no longer 
available to carry out normal operations. | `enrollmentCompleted(device) && 
device.spec.decommissioning != nil |


| `Decommissioned` | The device has either completed its decommissioning process or has 
encountered an unrecoverable error in doing so. | `enrollmentCompleted(device) && 
device.spec.decommissioning != nil && (
 ∃ `ConditionTypeDeviceDecommissioning` ∈ DeviceStatus.Conditions
  &&  `ConditionTypeDeviceDecommissioning`.reason ∈ {Complete, Error})`|
```

Edit: I removed the `DeviceStatus` checks on `Decommissioning` because from the server's POV the device is `Decommissioning` even if the agent never receives that request.

This doesn't have to do with the service enforcing any contract about the device state, so it should be separate from your earlier comment about changing implementation, but let me know if that's incorrect.

Edit 2: ... Looks like I went back to basically my original definitions. Maybe we have a quick sync about these.

lkatalin:
I've made updates since the last sync, we can review when you get a chance @fzdarsky .
Hopefully this comment will keep the bot from closing it.

lkatalin:
Please keep the PR open

lkatalin:
Reopening

lkatalin:
We want to get this merged

lkatalin:
Okay, looks like [a PR can't re reopened after a force push](https://github.com/isaacs/github/issues/361) so I'll have to open a new PR with this change.

PR #862, opened 2025-02-05
EDM-1014: Report versions of all software components phase 2 - agent

gshilin-sdb:
Added AgentVersion to service endpoint and to report

## Summary by CodeRabbit

- **New Features**
	- Added an agent version field to device information, which is now visible in CLI listings with a new "VERSION" column.
	- Updated API documentation to include agent version details, enriching the device system information provided to users.
	- Introduced a new mock implementation for the CallbackManager interface to enhance unit testing capabilities.

- **Bug Fixes**
	- Improved output formatting for device information in CLI, ensuring clarity and consistency in displayed data.


api/v1alpha1/openapi.yaml
```diff
@@ -3936,6 +3936,9 @@ components:
         operatingSystem:
           type: string
           description: The Operating System reported by the device.
+        agentVersion:
```
hexfusion, line 3939: One question, do we think the agent will be independent from the OS image? If no should this status be part of `os:`?

gshilin-sdb: AFAIK we can build agent-0.3.0 and -0.4.0 for RHEL9, so yes, it will be different

hexfusion: Right, I guess the question rephrased is can you upgrade the agent without a new OS image?

gshilin-sdb: If I understand you correctly: I can build different agents into the same OS image.
So Architecture (runtime.GOARCH) and OperatingSystem (runtime.GOOS) describes OS, but agentVersion describes agent. It can be different version of agent running on the same ARCH-OS.
So I think agent version is not connected to OS version, and I'd not mix them

fzdarsky: @hexfusion: You’re correct that with bootc/rpm-ostree we assume users bake the agent into the OS image, but we may relax that constraint. Say, when the agent is deployed on an rpm/deb package-based system or containerized.

internal/cli/get.go
```diff
@@ -386,9 +386,9 @@ func (o *GetOptions) printDevicesSummaryTable(w *tabwriter.Writer, summary *api.
 
 func (o *GetOptions) printDevicesTable(w *tabwriter.Writer, devices ...api.Device) {
 	if o.Output == wideFormat {
-		fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN\tLABELS")
+		_, _ = fmt.Fprintln(w, "NAME\tALIAS\tOWNER\tSYSTEM\tUPDATED\tAPPLICATIONS\tLAST SEEN\tVERSION\tLABELS")
```
fzdarsky, line 389: I don’t think the agent version is important enough to consume a column in the table output, not even in wide format.

api/v1alpha1/openapi.yaml
```diff
@@ -3936,6 +3936,9 @@ components:
...
+          type: string
+          description: The Agent version.
```
fzdarsky, line 3941: Any reason not to make the agentVersion required?

internal/util/util.go
```diff
@@ -108,13 +108,9 @@ func LabelMapToArray(labels *map[string]string) []string {
...
 		return []string{}
 	}
-	output := make([]string, len(*labels))
-	i := 0
-	for key, val := range *labels {
-		output[i] = fmt.Sprintf("%s=%s", key, val)
-		i++
-	}
-	return output
+	return lo.Map(lo.Entries(*labels), func(entry lo.Entry[string, string], _ int) string {
```
fzdarsky, line 111: This is great example for what we discussed on the team call today, where introducing `lo` may have undesirable side-effects besides adding a dependency, which is encouraging changing the programming style.

IMO, this change does not improve readability, conciseness, performance, etc, but on the contrary reduces readability as it assumes
knowledge of `lo` and isn’t aligned with the style used elsewhere in the codebase.

api/v1alpha1/openapi.yaml
hexfusion: fair

internal/cli/get.go
gshilin-sdb: There is no other place to display it... Any ideas?

api/v1alpha1/openapi.yaml
gshilin-sdb: To support older agents that don't send it.

internal/util/util.go
gshilin-sdb: reverted

internal/cli/get.go
avishayt: It will be displayed with -oyaml, for example

gshilin-sdb: restored to original

api/v1alpha1/openapi.yaml
gshilin-sdb: made

avishayt (approved):

PR #863, opened 2025-02-05
NO-ISSUE: Generate secrets via job

rawagner:
We cannot create secrets directly in our helm chart due to ACM operator's restrictions. So instead we create a job that generates these secrets in runtime.

## Summary by CodeRabbit

- **New Features**
	- Introduced an automated process that securely generates credentials during deployment with improved error handling.
	- Enhanced security by adding dedicated access controls for managing sensitive data across deployment environments.
- **Refactor**
	- Removed outdated secret management routines to streamline the deployment configuration.


mangelajo (approved):
Looks good to me,

PR #867, opened 2025-02-05
EDM-715: Quadlet configuration changes

DakCrowder:
Modifies the quadlets config in the following ways:
- Use rootless podman for running containers
  - Associated .container/.network/.volume files now moved to user owned `$HOME/.config/containers/systemd` rather than `/etc/containers/systemd`
  - This should be more inline with how we might deploy the quadlet services in prod contexts
- Moves config files before mounting into containers to a separate location
- Explicitly defines volumes our containers write to with .volume files
- More selective in copying files
- Uses the 0.4.0 tag for flightctl containers rather than :latest.  Eventually this should be configurable and/or work with locally built images

This pulls in some work done when building the containerized AAP installer PoC (which is no longer the path forward) but overall gets our quadlets config in a better place.

## Summary by CodeRabbit

- New Features
	- Added persistent storage options for key services to enhance data management.
- Chores
	- Updated container images to a fixed version (0.4.0) for improved stability.
	- Refined volume mappings with enhanced security settings.
	- Streamlined deployment scripts by removing elevated privilege requirements and switching to user-level execution.
	- Standardized service startup directives for more consistent launch behavior.

### DIFF
--- a/deploy/podman/flightctl-api/flightctl-api.container
+++ b/deploy/podman/flightctl-api/flightctl-api.container
@@ -7,7 +7,7 @@ Before=flightctl-periodic.service flightctl-worker.service
 
 [Container]
 ContainerName=flightctl-api
-Image=quay.io/flightctl/flightctl-api:latest
+Image=quay.io/flightctl/flightctl-api:0.4.0-rc3
 Network=flightctl.network
 Environment=HOME=/root
 Environment=FLIGHTCTL_DISABLE_AUTH=true
@@ -17,8 +17,8 @@ PublishPort=7443:7443
 PublishPort=7444:7444
 PublishPort=15690:15690
 PublishPort=15691:15691
-Volume=/etc/containers/systemd/flightctl-api/flightctl-api-config/config.yaml:/root/.flightctl/config.yaml
-Volume=flightctl-api-certs:/root/.flightctl/certs
+Volume=flightctl-api-certs:/root/.flightctl/certs:Z
+Volume=%h/.config/flightctl/flightctl-api-config/config.yaml:/root/.flightctl/config.yaml:Z
 
 [Service]
 Slice=flightctl.slice
--- a/deploy/podman/flightctl-api/flightctl-api.container
+++ b/deploy/podman/flightctl-api/flightctl-api.container
@@ -27,4 +27,3 @@ RestartSec=30
 
 [Install]
 WantedBy=flightctl.slice
-WantedBy=default.target
--- a/deploy/podman/flightctl-api/flightctl-api.container
+++ b/deploy/podman/flightctl-api/flightctl-api.container
@@ -27,3 +27,4 @@ RestartSec=30
 
 [Install]
 WantedBy=flightctl.slice
+WantedBy=default.target
--- a/deploy/podman/flightctl-api/flightctl-api.container
+++ b/deploy/podman/flightctl-api/flightctl-api.container
@@ -7,7 +7,7 @@ Before=flightctl-periodic.service flightctl-worker.service
 
 [Container]
 ContainerName=flightctl-api
-Image=quay.io/flightctl/flightctl-api:0.4.0-rc3
+Image=quay.io/flightctl/flightctl-api:0.4.0
 Network=flightctl.network
 Environment=HOME=/root
 Environment=FLIGHTCTL_DISABLE_AUTH=true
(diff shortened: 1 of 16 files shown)


deploy/podman/flightctl-api/flightctl-api.container
```diff
@@ -17,8 +17,8 @@ PublishPort=7443:7443
 PublishPort=7444:7444
 PublishPort=15690:15690
 PublishPort=15691:15691
-Volume=/etc/containers/systemd/flightctl-api/flightctl-api-config/config.yaml:/root/.flightctl/config.yaml
-Volume=flightctl-api-certs:/root/.flightctl/certs
+Volume=flightctl-api-certs:/root/.flightctl/certs:Z
+Volume=%h/.config/flightctl/flightctl-api-config/config.yaml:/root/.flightctl/config.yaml:Z
```
hexfusion, line 21: Could we try to make these paths dynamic in some way so we can accommodate local/prod paths. I think talked about go template or? Problem is in the soonish future we might need to accommodate both :)

deploy/podman/flightctl-kv/flightctl-kv-standalone.container
```diff
@@ -4,13 +4,18 @@ Description=FlightCtl Key Value service
 [Container]
 ContainerName=flightctl-kv
 Image=docker.io/redis:7.4.1
-Environment=HOME=/root
```
hexfusion, line 7: Just curious why we dropped this.

deploy/podman/flightctl-api/flightctl-api.container
```diff
@@ -27,3 +27,4 @@ RestartSec=30
 
 [Install]
 WantedBy=flightctl.slice
+WantedBy=default.target
```
hexfusion, line 30: What advantages does this provide?

deploy/podman/flightctl-kv/flightctl-kv-standalone.container
```diff
@@ -4,13 +4,18 @@ Description=FlightCtl Key Value service
...
+Exec=redis-server /usr/local/etc/redis/redis.conf
+Network=flightctl.network
+
+# The following is necessary so that the redis user which runs the redis process inside of the container
+# can have write access to the flightctl-kv volume
+PodmanArgs="--userns=keep-id:uid=999,gid=999"
```
hexfusion, line 12: neat :+1:

deploy/scripts/deploy_quadlets.sh
```diff
@@ -13,24 +13,40 @@ if ! PRIMARY_IP=$(bash -c 'source ./test/scripts/functions && get_ext_ip'); then
 fi
 export PRIMARY_IP
 
-envsubst "\$PRIMARY_IP" < deploy/podman/flightctl-api/flightctl-api-config/config.yaml.template > deploy/podman/flightctl-api/flightctl-api-config/config.yaml
+echo "Copying quadlet unit files"
+mkdir -p $SYSTEMD_DIR
```
hexfusion, line 17: nit: it's generally good practice to quote variables so we can reduce unexpected behavior when they are expanded.

``` suggestion
mkdir -p "${SYSTEMD_DIR}"
```

deploy/podman/flightctl-kv/flightctl-kv-standalone.container
DakCrowder: This isn't used by redis, it isn't causing any harm but is also doing nothing for us.  I actually need to remove it from the other kv .container file

deploy/podman/flightctl-api/flightctl-api.container
DakCrowder: I agree - templating should be coming in a follow up PR!

deploy/podman/flightctl-api/flightctl-api.container
DakCrowder: This ensures systemd spins the containers back up on boot/reboot - however until templating is done I think I'm actaully going to remove this line so if somebody runs it locally we aren't throwing all these containers up in the background

deploy/podman/flightctl-kv/flightctl-kv-standalone.container
DakCrowder: For a bit of added context redis is also running as a redis non-root user in the container so I'd say its actively confusing to have $HOME=/root

deploy/scripts/deploy_quadlets.sh
DakCrowder: Makes sense, coderabbit thought the same thing :) - will get these refs updated

deploy/podman/flightctl-kv/flightctl-kv-standalone.container
hexfusion: makes sense

deploy/podman/flightctl-api/flightctl-api.container
hexfusion: seems like a nice production condition for future templating perhaps :+1:

hexfusion (approved):
Nice progress :+1:

PR #872, opened 2025-02-06
EDM-973: packaging/rpm: make firewalld a hard dependency

hexfusion:
The default hooks that are shipped with the RPM depend of firewalld this PR makes it a required dependency. While we could try to evaluate if firewalld exists if we are going to be opinionated on the hook itself we should support it out of the box. 

This is meant to be a discussion as well as there are various solutions but `systemd` and `nmcli` are included already by default.

## Summary by CodeRabbit

- **Chores**
	- Updated the agent’s dependency management to require a firewall service, ensuring that the necessary firewall package is installed alongside the agent for seamless operation.

### DIFF
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -38,6 +38,7 @@ Flightctl is a command line interface for managing edge device fleets.
 Summary: Flightctl Agent
 
 Requires: bootc
+Requires: firewalld
 
 %description agent
 Flightctl Agent is a component of the flightctl tool.


mangelajo:
Firewalld uses memory and cpu, and imposes an additional footprint to customers.

While I cannot see why somebody would want to run devices without a firewall , they can always decide to go bare iptables/nft.

I would instead evaluate if firewalld is there before trying to apply, and produce an error/warning if it isn't there.

hexfusion:
The main problem with the inline run command is concluding the binary we need to verify. I can ensure firewall-cmd in this case but other rules make this nontrivial considering you can invoke inline bash from a subshell.

PR #880, opened 2025-02-07
EDM-1104: agent/rollback: set version failed before rollback

hexfusion:
## Summary by CodeRabbit

- Bug Fixes
	- Improved error handling and logging during device operations for a more stable experience.
	- Refined task management to better preserve important operations and prevent premature removals.
- Chores
	- Updated the agent package’s installation requirements to include an essential firewall management service for enhanced system support.

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -211,18 +211,25 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 		return
 	}
 
-	if err := a.sync(ctx, current, desired); err != nil {
+	if syncErr := a.sync(ctx, current, desired); syncErr != nil {
 		// if context is canceled return to exit the sync loop
-		if errors.Is(err, context.Canceled) {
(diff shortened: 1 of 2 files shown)


internal/agent/device/device.go
```diff
@@ -211,18 +211,25 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 		return
 	}
 
-	if err := a.sync(ctx, current, desired); err != nil {
+	if syncErr := a.sync(ctx, current, desired); syncErr != nil {
 		// if context is canceled return to exit the sync loop
 		if errors.Is(err, context.Canceled) {
```
DakCrowder, line 216: Should this check if the syncErr is Canceled?  from line 209 would err always be nil here?

hexfusion: good catch!

hexfusion: pushed

DakCrowder (approved):

PR #882, opened 2025-02-08
EDM-390: Update device lifecycle status immediately upon user request

lkatalin:
Updates device lifecycle status immediately upon user request to decommission.
 **This PR is required to unblock UI for Tech Preview**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved device status management to ensure that valid statuses are maintained when handling updates.
  - Enhanced the decommission process by validating a device’s current state before processing decommissioning requests, preventing duplicate actions.

### DIFF
--- a/internal/service/common/device.go
+++ b/internal/service/common/device.go
@@ -51,8 +51,9 @@ func ReplaceDeviceStatus(ctx context.Context, st store.Store, log logrus.FieldLo
 			return nil, err
 		}
 	}
-	if oldDevice.Status.Lifecycle.Status == api.DeviceLifecycleStatusEnrolled && device.Status.Lifecycle.Status == api.DeviceLifecycleStatusUnknown {
-		device.Status.Lifecycle.Status = api.DeviceLifecycleStatusEnrolled
+	// do not overwrite valid service-side lifecycle status with placeholder device-side status
+	if device.Status.Lifecycle.Status == api.DeviceLifecycleStatusUnknown {
+		device.Status.Lifecycle.Status = oldDevice.Status.Lifecycle.Status
 	}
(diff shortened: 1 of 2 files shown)


internal/service/common/device.go
```diff
@@ -51,8 +51,9 @@ func ReplaceDeviceStatus(ctx context.Context, st store.Store, log logrus.FieldLo
 			return nil, err
 		}
 	}
-	if oldDevice.Status.Lifecycle.Status == api.DeviceLifecycleStatusEnrolled && device.Status.Lifecycle.Status == api.DeviceLifecycleStatusUnknown {
-		device.Status.Lifecycle.Status = api.DeviceLifecycleStatusEnrolled
+	// do not overwrite valid service-side lifecycle status with placeholder device-side status
+	if device.Status.Lifecycle.Status == api.DeviceLifecycleStatusUnknown {
+		device.Status.Lifecycle.Status = oldDevice.Status.Lifecycle.Status
```
avishayt, line 56: It seems to me that both the old and new code here is not necessary.
A couple lines down we call UpdateServiceSideStatus(), which calls updateServerSideLifecycleStatus(), which should set device.Status.Lifecycle the status to decommissioning if necessary.

lkatalin: This logic is needed to update the data passed into the `updateServerSideLifecycleStatus()` function, otherwise there is a bootstrapping problem where the first "blank" `lifecycle status` created device-side overwrites the valid `enrolled` lifecycle status tracked service-side. The service-side owns the lifecycle status so it should be updated this way AFAIU.

avishayt (approved):

PR #886, opened 2025-02-10
EDM-1101: api/device/console: add support for a list of sessions

hexfusion:


api/v1alpha1/openapi.yaml
```diff
@@ -3461,12 +3461,23 @@ components:
...
           type: string
           description: The gRPC endpoint for the console connection.
-        sessionID:
-          type: string
-          description: The session ID for the console connection.
+        sessions:
+          type: array
+          description: A list of console sessions for the console connection.
+          items:
+            $ref: '#/components/schemas/ConsoleSession'
```
mangelajo, line 3468: I was looking into doing it the other way around. List of sessions where each session can have a different endpoint too. Let me try to post an alternative, I had it 60% ...

mangelajo:
an alternative proposal https://github.com/flightctl/flightctl/pull/887

Honestly I am not sure how useful it is that we make it easy to contact multiple endpoints at this time, so we could as well ditch the gRPC endpoint at all and not even keep one.


That was necessary in the past because part of the plan was to make the consumer of the console, and the device match on the same gRPC endpoint, but, since the end design is to just funnel it through a queue in redis, that detail won't matter at all.

Saying that... what I mean, may be *your* PR is closer to what we want, and we just ditch the gRPC endpoint at all.

mangelajo:
Superseded by https://github.com/flightctl/flightctl/pull/887

PR #893, opened 2025-02-11
EDM-1145: Add versioning to Status schema, replace Error with it

fzdarsky:
- Adds missing version info to the API's `Status` response.
- Instead of doing the same for the `Error` response, merges the `Error` into `Status`, like it's handled on K8s.

## Summary by CodeRabbit

- **Refactor**
  - Standardized error responses across the API and CLI to deliver a more structured and informative status message, improving clarity and consistency in error handling.
  - Updated various methods to utilize the new `Status` schema for error handling, ensuring consistent response formats across multiple endpoints.

- **Documentation**
  - Updated API specifications and documentation to reflect the new status format for error responses, ensuring clients receive enhanced error context and guidance.


api/v1alpha1/openapi.yaml
```diff
@@ -4796,6 +4796,18 @@ components:
...
       type: object
       properties:
+        apiVersion:
+          type: string
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources.'
+        kind:
+          type: string
+          description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds.'
+        metadata:
+          $ref: '#/components/schemas/ListMeta'
```
avishayt, line 4806: What's this for?

avishayt:
Where are the status fields populated?

api/v1alpha1/openapi.yaml
fzdarsky: Copied that from K8s API. It optionally contains the `continue` and `remainingItemCounts` that exist lists / collections of resources. I _suspect_ this is because the `Status` is mainly used in responses to `deletecollections` calls that return lists, but I didn't manage to reproduce a case where K8s API would actually return a non-empty metadata. Remove?

avishayt: If we don't use it, I say remove

celdrake: Just to be clear, you mean removing only `metadata` in the Status object, right?
In the UI, `metadata` is used for lists pagination so it's needed in `DeviceList`, ` FleetList` etc

api/v1alpha1/types.gen.go
```diff
@@ -1334,14 +1328,23 @@ type SshRepoSpec struct {
...
-	Message *string `json:"message,omitempty"`
+	Message string `json:"message"`
 
 	// Reason A machine-readable description of why this operation is in the "Failure" status. If this value is empty there is no information available. A Reason clarifies an HTTP status code but does not override it.
-	Reason *string `json:"reason,omitempty"`
+	Reason string `json:"reason"`
 
 	// Status Status of the operation. One of: "Success" or "Failure". More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status.
-	Status *string `json:"status,omitempty"`
+	Status string `json:"status"`
```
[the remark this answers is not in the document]
fzdarsky: The identified nil-checks are not related to the API's `Status` type, but operate on the `Status` fields of other types.

hexfusion (approved):

PR #894, opened 2025-02-11
NO-ISSUE: Add flightctl.io prefix to resource versions

avishayt:
## Summary by CodeRabbit

- **New Features**
  - Updated API version identifiers in configuration examples to use a consistent, namespaced format ("flightctl.io/v1alpha1") across multiple resource types. This change standardizes the API versioning scheme visible in public configurations.
  - Introduced a new constant `APIVersionPrefix` for improved clarity in API versioning.


api/v1alpha1/consts.go
```diff
@@ -3,9 +3,11 @@ package v1alpha1
 import "time"
 
 const (
-	CertificateSigningRequestAPI      = "v1alpha1"
-	CertificateSigningRequestKind     = "CertificateSigningRequest"
-	CertificateSigningRequestListKind = "CertificateSigningRequestList"
+	APIVersionPrefix = "flightctl.io/"
```
fzdarsky, line 6: I believe K8s would call that the API group. Shall we leave out the trailing `/` here and instead add it in the function that constructs the group/version string?

avishayt: Fixed

fzdarsky (approved):

avishayt:
FYI @celdrake . The UI should use this prefix.  However there is no validation so it shouldn't break. Can be done after TP if necessary.

PR #899, opened 2025-02-12
NO-ISSUE: Fix datarace detected by unittests in console

mangelajo:
The go-routines for copying stdio/stdout to the PTY were producing a couple of ouputs (stdinErr, stdoutErr) that we check/print later, but we didn't check that those have properly finished before we attempt to read the output.

This may fix the issue

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the handling of console data transfers, ensuring that operations complete fully before any errors are checked. This enhancement leads to more reliable and stable interactions for users operating within the console environment.


hexfusion:
tested with 

```
go test -v -timeout 10m -run ^TestConsoleControllerSuite$  github.com/flightctl/flightctl/internal/agent/device/console -race -count 1000
```

PR #901, opened 2025-02-12
NO-ISSUE: Remove redundant RBAC check for ws endpoint. The check is a…

rawagner:
…lready done in permissions middleware.


internal/auth/middleware.go
```diff
@@ -50,21 +84,28 @@ func CreatePermissionsMiddleware(log logrus.FieldLogger) func(http.Handler) http
 			} else {
```
fzdarsky, line 84: Setting `actionGet` irrespective of the `r.Method` means using other methods on `/api/version` would pass permission check, too. Handlers for those don't exist, of course, but still feels a bit off.

If you want to fix, a quick way could be to look up the defaultAction first, then pass the action (instead of `r.Method`) into extractResourceAndAction() that would modify the action if needed.

fzdarsky (approved):

internal/api_server/middleware/route_exists.go
```diff
@@ -0,0 +1,22 @@
...
+			if !router.Match(routeContext, r.Method, r.URL.Path) {
+				http.NotFound(w, r)
+				return
+			}
+			next.ServeHTTP(w, r)
+		}
+		return http.HandlerFunc(fn)
+	}
+
+}
```
fzdarsky, line 22: I'm ok with that for now, but wondering whether we can avoid the additional search through the routing tree by using route groups to separate what needs an AuthN or AuthZ check, then making the AuthN/AuthZ themselves look up whether a resource exists or not.

internal/auth/middleware.go
rawagner: Thanks for the suggestion. I've updated the logic to use the `defaultAction`

rawagner: Also note that I am not calling `extractResourceAndAction` afterwards which would return `NilAction` as it is trying to match the `method` passed to `defaultActions` [see](https://github.com/flightctl/flightctl/blob/main/internal/auth/permissions.go#L99-L102)

internal/auth/middleware_test.go
```diff
@@ -0,0 +1,280 @@
...
+	for _, m := range unsupportedMethods {
+		authZMock := NewMockAuthZMiddleware(ctrl)
+		authZ = authZMock
+		authZMock.EXPECT().CheckPermission(gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
+
+		req := httptest.NewRequest(m, "https://fctl.io/api/v1/devices", nil)
+		w := httptest.NewRecorder()
+		authZMiddleware.ServeHTTP(w, req)
+
+		require.Equal(http.StatusBadRequest, w.Code)
```
gshilin-sdb, line 196: **Bad Request** is used when the server cannot process the request due to a client error, typically malformed syntax, invalid parameters, missing data, or invalid JSON payload.

We should use one of:
* **405 Method Not Allowed** -- the resource is known but explicitly does not support the method used.
* **501 Not Implemented** -- the server doesn't recognize or doesn't support the requested method itself (e.g., a non-standard method like TRACE, if unsupported).

internal/auth/middleware_test.go
```diff
@@ -0,0 +1,280 @@
...
+}))
+
+type AuthNTestRequest struct {
+	url             string
+	shouldSkipCheck bool
+}
+
+var authNRequests []AuthNTestRequest = []AuthNTestRequest{
+	{
+		url:             "https://fctl.io/api/v1/auth/config",
```
gshilin-sdb, line 211: Nitpick: value should be much closer to "url:"

internal/auth/middleware_test.go
```diff
@@ -0,0 +1,280 @@
...
+		url: "https://fctl.io/api/v1/auth/foo/bar",
+	},
+	{
+		url: "https://fctl.io/api/v5",
+	},
+	{
+		url:        "https://fctl.io/api/foo",
+		badRequest: true,
+	},
+}
```
gshilin-sdb, line 156: Maybe to add all kinds of strange wss: endpoints, just to be sure...

internal/auth/middleware_test.go
rawagner: well, this is [your implementation ](https://github.com/flightctl/flightctl/pull/834) and you return 400 when mapping fails. It actually looks better to me than the 405/501. 400 is returned because mapping of http req to resource/action failed.
405/501 seems like internal implementation detail.

btw the url validation already happens in `OapiRequestValidatorWithOptions` middleware so we know we wont get a request that does not conform to OpenAPI schema.

gshilin-sdb: :D I was wrong. But your vision is quite good. Let it be 400

gshilin-sdb:
LGTM

gshilin-sdb (approved):

PR #907, opened 2025-02-12
EDM-1160: Add versioning to agent API

hexfusion:
This PR introduces versioning to the agent API by replacing RenderedDeviceSpec with a full Device resource, enabling proper versioning. This ensures backward compatibility, as the agent now reconciles the Device resource instead of relying on a non-versionable spec.

## Summary by CodeRabbit

- **Refactor**
  - Standardized how device configurations are handled by streamlining API endpoints and renaming internal types for better consistency. End users now receive a more unified, clearer device representation.
  - Updated application configuration types to a more descriptive format, which simplifies and clarifies how applications are defined and rendered.
  - Introduced a new function for converting ignition configurations into structured configurations, enhancing type specificity in device rendering.

- **Tests**
  - Revised integration and end-to-end tests to align with the new device and application specification models, ensuring ongoing reliability.
  - Updated tests to reflect changes in application specification types, ensuring accurate validation of device configurations.


fzdarsky:
LGTM. Just so we don't forget: Adding `device.spec.consoles` requires us to protect this field from being updated by the user, effectively bypassing the RBAC on the `/ws/v1/devices/{names}/console` endpoint.

internal/store/device.go
```diff
@@ -659,19 +659,25 @@ func (s *DeviceStore) GetRendered(ctx context.Context, orgId uuid.UUID, name str
...
-		Config:          device.RenderedConfig,
-		Os:              device.Spec.Data.Os,
-		Systemd:         device.Spec.Data.Systemd,
-		Resources:       device.Spec.Data.Resources,
+	device, err := deviceModel.ToApiResource()
+	if err != nil {
+		return nil, err
+	}
+
+	spec := api.DeviceSpec{
```
avishayt, line 667: Nit: In the future we may add more fields to device spec.  Why not just update config, applications, and consoles in device.Spec, so that all other fields go through as-is?

internal/tasks/device_render.go
```diff
@@ -605,3 +607,61 @@ func (t *DeviceRenderLogic) cloneCachedGitRepoToIgnition(ctx context.Context, re
 	ignToReturn := wrapper.AsIgnitionConfig()
 	return &ignToReturn, nil
 }
+
+// ignitionConfigToRenderedConfig converts an ignition config to rendered config bytes
+func ignitionConfigToRenderedConfig(ignition *config_latest_types.Config) ([]byte, error) {
```
avishayt, line 612: Why convert to ignition, and from ignition to inline, rather than converting directly to inline?

internal/cli/get.go
```diff
@@ -176,7 +176,7 @@ func (o *GetOptions) Run(ctx context.Context, args []string) error { //nolint:go
 	case kind == DeviceKind && len(name) > 0 && !o.Rendered:
 		response, err = c.ReadDeviceWithResponse(ctx, name)
 	case kind == DeviceKind && len(name) > 0 && o.Rendered:
-		response, err = c.GetRenderedDeviceSpecWithResponse(ctx, name, &api.GetRenderedDeviceSpecParams{})
+		response, err = c.GetRenderedDeviceWithResponse(ctx, name, &api.GetRenderedDeviceParams{})
```
avishayt, line 179: Currently the rendered output can only be printed as json or yaml.  Do we want to change this to simply print it as a single device?

internal/tasks/device_render.go
hexfusion: it was only to avoid implementing a merge. The goal is to remove.

internal/cli/get.go
fzdarsky: +1. The less we special-case the better.

api/v1alpha1/openapi.yaml
```diff
@@ -4550,9 +4495,9 @@ components:
             $ref: '#/components/schemas/ConfigProviderSpec'
         applications:
           type: array
-          description: List of applications.
+          description: List of application providers.
           items:
-            $ref: '#/components/schemas/ApplicationSpec'
+            $ref: '#/components/schemas/ApplicationProviderSpec'
```
celdrake, line 4500: The previous type `ApplicationSpec` is currently being used in the UI.
I'll create a PR for updating the UI to the new types, so it's ready when this PR is merged.

internal/cli/get.go
hexfusion: yeah going to need to be a followup

internal/store/device.go
hexfusion: moved this into the model

internal/cli/get.go
hexfusion: https://issues.redhat.com/browse/EDM-1168

internal/tasks/device_render.go
hexfusion: https://issues.redhat.com/browse/EDM-1167

fzdarsky (approved):

PR #908, opened 2025-02-12
EDM-1102: support user-provided certificates and generate self-signed…

asafbss:
… only if missing

- modified logic to prioritize user-provided certificates before generating or using a self-signed one
- replaced EnsureServerCertificate (which relied on oscrypto.GetServerCert) with crypto.GetTLSCertificateConfig
- logic will not call EnsureServerCertificate that verified hostnames and IPs, preventing unnecessary certificate regeneration
- if existing certificate cannot be loaded, the process will fail
- improved logging to warn about expired certificates instead of failing immediately

## Summary by CodeRabbit

- **New Features**
	- Enhanced certificate management: The system now validates user-supplied certificates, automatically falls back to a self-signed certificate when required, and warns for expired certificates.
	- Added support for RSA private keys with improved error reporting for key operations.
- **Refactor**
	- Streamlined file accessibility checks and error handling for certificate and key operations, boosting reliability and clarity.


asafbss:
Similar to the Kubernetes API Server approach but without storing the certificate in memory.

https://github.com/kubernetes/kubernetes/blob/c81431de59a3bf516489317433a165b050322339/staging/src/k8s.io/apiserver/pkg/server/options/serving.go#L340

fzdarsky (approved):

kot-begemot-uk (commented):
All OK. 
I may need to return to this code for the plugins, because plugins may do their own things with private keys (both certmanager and vault do so). At this point in time this is "what the doctor ordered". Good to go.

PR #931, opened 2025-02-18
EDM-701: Remove CSR 'usages' check

lkatalin:
This is [not used to generate the cert](https://github.com/flightctl/flightctl/blob/main/internal/crypto/csr.go#L78-L79) and just gives the user an unnecessary error if they do not get it right.


lkatalin:
Thank you @celdrake !

I agree, I will remove the "usages" from the API. It is best not to send this data even if it is ignored.

Re: the signerName, yes, the documentation could be better. There is an example command [here](https://github.com/flightctl/flightctl/blob/main/docs/user/getting-started.md#building-a-bootable-container-image-including-the-flight-control-agent) that demonstrates how to use it correctly but we should do more.

> the API doesn't specify anywhere which are its correct values
How would you suggest to do this?
The CLI does give a hint [here](https://github.com/flightctl/flightctl/blob/main/internal/cli/certificate.go#L93) but I can add more to make it more obvious.

celdrake:
> How would you suggest to do this? The CLI does give a hint [here](https://github.com/flightctl/flightctl/blob/main/internal/cli/certificate.go#L93) but I can add more to the API itself to make it more obvious, if we have ways to do that. Edit: Oh, maybe in this `description`? https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/openapi.yaml#L4913

@lkatalin This seems like a good option. Alternatively, the valid values could be listed in the error message itself, similar to how the CLI does it. Sorry as this is not strictly related to this PR's issue, so feel free to handle this later.

lkatalin:
> Why do you want to remove "usages" instead of fixing its validation and that it's being used properly by the CLI and by the signer(s)?

@fzdarsky This fix is related to this bug: [Usages Should Be Generated For User](https://issues.redhat.com/browse/EDM-701), which points out that we take and validate input for `usages` but these are [never used by the backend](https://github.com/flightctl/flightctl/blob/main/internal/crypto/csr.go#L78-L79) (where the usages are auto-generated / hardcoded). That means requiring correct `usages` as input makes things more complex for the user for no benefit.

We could go two ways with this, one is to use the user's input to create the `usages` on the backend and the other would be to simply do this for the user and not require them to correctly enter this input. IMO it makes sense to do the latter currently as the backend knows what the usages "should be". If we want to support a variety of user-provided usages later we could add that as necessary.

> We currently only have one signer ("enrollment"). The other one ("ca") doesn't make sense (all signers use a CA) and we should remove it IMO.

This is true as well. The "ca" right now is basically a placeholder for "signers that aren't enrollment". Removing it could make sense, but if we do plan to support more signers (do we know what these will be?) it might be less effort and code churn to simply add that logic where "ca" is currently instead of tearing it out and then adding something similar in later.

lkatalin:
> valid values could be listed in the error message itself

@celdrake Good idea, let me see about adding this too.

PR #938, opened 2025-02-20
EDM-1034: agent/hooks: add support for passing variables to hooks from inline cfg

hexfusion:
This PR adds support for passing hook variables via inline config by using a dollar bracket syntax `${ Var }`. This PR replaces the current regex with more performant string replacements and strings.Builder with benchmarked results. Unfortunately for Golang regex even when precompiled is a fair amount slower than alternatives in this case  ~10x

Note: confirmed change with @avishayt 

previous failed syntax

```
      - inline:
        - content: |-
            - if:
              - path: /var/home/user
                op: [created, updated, removed]
              run: "kubectl apply -k {{ Path }}"
          path: /etc/flightctl/hooks.d/afterupdating/custom-hook.yaml
```

new format

```
      - inline:
        - content: |-
            - if:
              - path: /var/home/user
                op: [created, updated, removed]
              run: "kubectl apply -k ${ Path }"
          path: /etc/flightctl/hooks.d/afterupdating/custom-hook.yaml
```

#### before
```
goos: linux
goarch: amd64
pkg: github.com/flightctl/flightctl/internal/agent/device/hook
BenchmarkReplaceTokensOld-12    	  448608	      2623 ns/op	     882 B/op	      19 allocs/op
```

#### after
```
goos: linux
goarch: amd64
pkg: github.com/flightctl/flightctl/internal/agent/device/hook
BenchmarkReplaceTokensOptimized-12    	 4047739	       279.1 ns/op	      80 B/op	       1 allocs/op
PASS
```

## Summary by CodeRabbit

- **Documentation**
  - Updated the device configuration guide to clarify the correct syntax for using dynamic inline variables, changing the format from double curly braces to dollar signs for better consistency.
  
- **Tests**
  - Added a new test function to validate the behavior of the token replacement functionality with various scenarios.
  - Introduced a benchmark function to measure the performance of the token replacement process.


fzdarsky:
The problem we're trying to address is that device templating (service-side) and the hooks templating (agent-side) both use the same `{{...}}` delimiters and the (only?) place this clashes is with InlineConfig, right?

Escaping is a solution, of course, though one that's notoriously hard to use. I'm wondering whether a better solution would be to use different delimiters?

For example, we use `{{...}}` service-side because go-template delimiters felt natural there, due to the K8s-/Helm-like syntax and functions.
In contrast, the hooks are probably more for an audience using shell scripting, where `${...}` or `${{...}}` would seem natural.

hexfusion:
The general problem is that the go template will try to manage ${{ Token }} directly while it will treat ${ Token } as a literal. If we mutate ${{ -> ${ the output might seem awkward in the case where ${{ NotVald }} would print ${ NotValid }.

fzdarsky (approved):

PR #943, opened 2025-02-24
EDM-799: added tests for decommission via CLI

Samudelacruz77:
## Summary by CodeRabbit

- **Tests**
  - Introduced an end-to-end testing suite to validate device decommissioning via the CLI. The new tests simulate typical user interactions and ensure that devices are accurately processed for decommissioning with clear confirmations, enhancing system reliability.
  - Added a new method to check specific conditions on a device's status within the testing harness.


test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,59 @@
...
+		harness = e2e.NewTestHarness()
+		deviceId = harness.StartVMAndEnroll()
+	})
+
+	AfterEach(func() {
+		harness.Cleanup(false)
+	})
+
+	Context("decommission", func() {
+		It("should decommission a device via CLI", Label("decommission", "rh-799"), func() {
```
amalykhi, line 37: Do you have Polarion test case for this?

Samudelacruz77: Need to do it, will do and add to the test

test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,59 @@
...
+			out, err := harness.CLI("decommission", "devices/"+deviceId)
+			Expect(err).NotTo(HaveOccurred())
+			logrus.Info(out)
+			Expect(out).To(ContainSubstring("Device scheduled for decommissioning: 200 OK:"))
```
lkatalin, line 41: Will this always have the `:` at the end?

test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,59 @@
...
+			harness.WaitForDeviceContents(deviceId, "The device has completed decommissioning and will wipe its management certificate",
+				func(device *v1alpha1.Device) bool {
+					return conditionExists(device, "DeviceDecommissioning", "True", string(v1alpha1.DecommissionStateComplete))
+				}, "10m")
+		})
+	})
+})
+
+func conditionExists(device *v1alpha1.Device, conditionType, conditionStatus, conditionReason string) bool {
```
lkatalin, line 50: We have a couple of similar functions available too (but not sure if it should be defined this way separately for testing):

https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/conditions.go#L95 

https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/properties.go#L105

lkatalin (commented):
I think it looks pretty good, thank you so much for writing this.

test/e2e/decommission/decommission.go
siserafin, line 50: Maybe this function could be moved inside harness, I think it could be useful for other test cases

Samudelacruz77: Ok, will put in harness library

Samudelacruz77: We have a more general function that allows us to use every condition

test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,50 @@
...
+		harness = e2e.NewTestHarness()
+		deviceId = harness.StartVMAndEnroll()
+	})
+
+	AfterEach(func() {
+		harness.Cleanup(false)
+	})
+
+	Context("decommission", func() {
+		It("should decommission a device via CLI", Label("decommission", "rh-799"), func() {
```
siserafin: I see the tc code is 81782

test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,50 @@
...
+			logrus.Infof("decommission device with id: %s", deviceId)
+
+			out, err := harness.CLI("decommission", "devices/"+deviceId)
+			Expect(err).NotTo(HaveOccurred())
+			logrus.Info(out)
+			Expect(out).To(ContainSubstring("Device scheduled for decommissioning: 200 OK:"))
+			harness.WaitForDeviceContents(deviceId, "The device has completed decommissioning and will wipe its management certificate",
+				func(device *v1alpha1.Device) bool {
+					return harness.ConditionExists(device, "DeviceDecommissioning", "True", string(v1alpha1.DecommissionStateComplete))
+				}, "2m")
```
siserafin: Maybe we could add the check that the device is no more in the 'flightctl get dev' list

siserafin (commented):
I think that we should merge this PR after fixing the small comments and close the related issue.

test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,51 @@
...
```
siserafin: This '2m' can be substitute by TIMEOUT constant that can be changed to 2m.

test/e2e/decommission/decommission.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+)
+
+const TIMEOUT = "1m"
+const POLLING = "250ms"
```
siserafin: Are you using POLLING constant?

Samudelacruz77: no, will remove

PR #955, opened 2025-02-26
Update CLI version command to use context and print version field names

fzdarsky:
- **New Feature**
	- Enabled the `--context/-c` flag to work with the version command, too.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the version display to clearly separate client and server details, ensuring the server version is shown only when available.

- **Bug Fixes**
	- Enhanced error messaging to provide clearer feedback when issues occur while retrieving version information.
	- Improved error handling for server connectivity issues, ensuring messages are printed to standard error without treating them as critical errors.

### DIFF
--- a/internal/cli/version.go
+++ b/internal/cli/version.go
@@ -5,9 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"slices"
 	"strings"
 
+	api "github.com/flightctl/flightctl/api/v1alpha1"
+	apiclient "github.com/flightctl/flightctl/internal/api/client"
 	"github.com/flightctl/flightctl/internal/client"
 	"github.com/flightctl/flightctl/pkg/version"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


mangelajo:
@fzdarsky It should be relatively easy to add e2e testing for this here:

https://github.com/flightctl/flightctl/blob/main/test/e2e/cli/cli_test.go#L80

internal/cli/version.go
```diff
@@ -26,16 +29,16 @@ type VersionOptions struct {
...
-	errReadingVersion = "reading service version"
-	errUnmarshalling  = "unmarshalling error"
+	errReadingVersion = "Could not read server version"
+	errUnmarshalling  = "Could not unmarshal server response"
 )
 
-type serviceVersion struct {
-	Version string `json:"version,omitempty"`
-	Error   string `json:"error,omitempty"`
+type Version struct {
```
gshilin-sdb, line 39: It shouldn't be global

gshilin-sdb:
BTW, which context is it about? "Update CLI version command to use context..."

oglok:
Looks good to me. It's true that the PR message is confusing, since this patch is not modifying any behaviour on the context, but that's just part of the CLI arguments. Or am I missing something?

oglok (approved):

PR #960, opened 2025-02-26
EDM-1202: Fix missing Flightctl API Server version in upstream images

asafbss:
Ensure version metadata (`SOURCE_GIT_TAG`, `SOURCE_GIT_TREE_STATE`, `SOURCE_GIT_COMMIT`) is correctly passed during the build process.

## Summary by CodeRabbit

- **Chores**
  - Enhanced the container build pipeline to capture comprehensive version metadata during builds. As a result, container images now reflect more accurate version and state details, improving traceability and consistency for deployments.

### DIFF
--- a/.github/workflows/publish-containers.yaml
+++ b/.github/workflows/publish-containers.yaml
@@ -100,6 +100,19 @@ jobs:
     runs-on: "ubuntu-24.04"
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Capture Git Metadata
+        run: |
+          echo "SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo v0.0.0)" >> $GITHUB_ENV
+          if git diff --quiet; then 
+            echo "SOURCE_GIT_TREE_STATE=clean" >> $GITHUB_ENV
+          else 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-containers.yaml
```diff
@@ -100,6 +100,19 @@ jobs:
...
+
+      - name: Capture Git Metadata
+        run: |
+          echo "SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo v0.0.0)" >> $GITHUB_ENV
+          if git diff --quiet; then 
+            echo "SOURCE_GIT_TREE_STATE=clean" >> $GITHUB_ENV
+          else 
+            echo "SOURCE_GIT_TREE_STATE=dirty" >> $GITHUB_ENV
+          fi
+          echo "SOURCE_GIT_COMMIT=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
```
fzdarsky, line 115: There's a tiny diff compared to the Makefile (`-- short "HEAD^{commit}"`). The logic in the Makefile ensures that HEAD is always a "commit-ish" (commit or annotated tag pointing to a commit). IDK if there's a relevant scenario where the result would differ, though.

fzdarsky (approved):

.github/workflows/publish-containers.yaml
asafbss: Fixing,,,

fzdarsky (approved):

PR #971, opened 2025-03-05
Ca separation v3

kot-begemot-uk:
This is the CA refactoring needed for pluggable CAs split into smaller logical chunks. 

The last commit is rather big, but it is unavoidable as it introduces an API change which results in changes to all test files which deal with crypto.

I have decided not to include the config changes needed for pluggability in this PR. It will come as a separate on top once we are done with this one.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined certificate management and TLS configuration by transitioning to a dedicated certificate bundle retrieval method, ensuring improved error handling and stability.

- **Tests**
  - Updated test routines to reflect the new certificate processing approach, resulting in more robust and reliable secure client-server communications.

- **Chores**
  - Reorganized internal certificate issuance and management logic for better maintainability and performance in secure connections.


internal/crypto/internal_ca.go
```diff
@@ -108,112 +105,6 @@ func makeSelfSignedCAConfig(subject pkix.Name, caLifetime time.Duration) (*oscry
 	return caConfig, nil
 }
 
-type CertificateExtensionFunc func(*x509.Certificate) error
-
-func (ca *CA) MakeServerCert(hostnames []string, expiryDays int, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
```
lkatalin: This function moved from `internal/crypto/cert.go` to `internal/crypto/internal_ca.go` in the first commit but moved back to the original file in the second commit.

internal/crypto/internal_ca.go
```diff
@@ -108,112 +105,6 @@ func makeSelfSignedCAConfig(subject pkix.Name, caLifetime time.Duration) (*oscry
...
-		return nil, err
-	}
-	server := &TLSCertificateConfig{
-		Certs: append([]*x509.Certificate{serverCrt}, ca.Config.Certs...),
-		Key:   serverPrivateKey,
-	}
-	return server, nil
-}
-
-func (ca *CA) MakeClientCertificate(certFile, keyFile string, subject string, expiryDays int) (*TLSCertificateConfig, error) {
```
lkatalin: Same with this one

internal/crypto/cert.go
```diff
@@ -35,120 +28,6 @@ func CNFromDeviceFingerprint(fingerprint string) (string, error) {
...
-
-// func (ca *CA) EnsureSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, bool, error) {
-// 	if subCA, err := GetCA(certFile, keyFile, serialFile); err == nil {
-// 		return subCA, false, err
-// 	}
-// 	subCA, err := ca.MakeAndWriteSubCA(certFile, keyFile, serialFile, name, expireDays)
-// 	return subCA, true, err
-// }
-
 func (ca *CA) EnsureServerCertificate(certFile, keyFile string, hostnames []string, expireDays int) (*TLSCertificateConfig, bool, error) {
```
lkatalin: Is there a reason this method wasn't transferred to the `internal_ca` file? Since the `CA` struct moved it seems like its methods should all be moved too.

internal/crypto/cert.go
```diff
@@ -179,60 +58,6 @@ func (ca *CA) MakeAndWriteServerCert(certFile, keyFile string, hostnames []strin
...
-	// Increment and persist serial
-	serial, err := ca.SerialGenerator.Next(template)
-	if err != nil {
-		return nil, err
-	}
-	template.SerialNumber = big.NewInt(serial)
-	return signCertificate(template, requestKey, ca.Config.Certs[0], ca.Config.Key)
-}
-
 func (ca *CA) EnsureClientCertificate(certFile, keyFile string, subjectName string, expireDays int) (*TLSCertificateConfig, bool, error) {
```
lkatalin: Same question as above

internal/crypto/cert.go
```diff
@@ -109,6 +177,40 @@ func (c *TLSCertificateConfig) GetPEMBytes() ([]byte, []byte, error) {
 	return certBytes, keyBytes, nil
 }
 
+func (ca *CA) IssueRequestedClientCertificateAsX509(csr *x509.CertificateRequest, expirySeconds int) (*x509.Certificate, error) {
+	return ca.IssueRequestedCertificateAsX509(csr, expirySeconds, []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth})
```
lkatalin: Why does this need to be wrapped?

lkatalin (commented):
Thanks, this is already more readable and easier to follow.

I reviewed the first two commits - two comments so far overall:

1. I'm seeing some functions like `MakeServerCert` move to a new file `internal_ca.go` in the first commit and then back again to `cert.go` in the next. They should either move to the new file and stay there or not move at all.

2. Some of the methods on the CA struct didn't move to the new `internal_ca.go` file and I'm not sure why - this means we have some methods on that struct in `internal_ca.go` but others are still in `cert.go`, which isn't ideal. (and I think may be the reason for some of the function wrapping?)

If you could update to fix those two issues I think that would be the next step.

test/harness/test_harness.go
```diff
@@ -165,7 +165,11 @@ func NewTestHarness(testDirPath string, goRoutineErrorHandler func(error)) (*Tes
 	}
 
 	// create client to talk to the server
-	client, err := testutil.NewClient("https://"+listener.Addr().String(), ca.Config)
+
+	pemCA, _ := ca.GetCABundle()
+
+	serverLog.Infof("New Client ca %s", pemCA)
+	client, err := testutil.NewClient("https://"+listener.Addr().String(), ca.GetCABundleX509())
```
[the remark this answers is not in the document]
kot-begemot-uk: There is no possible errors in this operation. If the CA plugin has initialized correctly the bundle should be available.

kot-begemot-uk:
> Thanks, this is already more readable and easier to follow.
> 
> I reviewed the first two commits - two comments so far overall:
> 
>     1. I'm seeing some functions like `MakeServerCert` move to a new file `internal_ca.go` in the first commit and then back again to `cert.go` in the next. They should either move to the new file and stay there or not move at all.

There is logic in the move. 

Initially, they are written as "private key abuse and inappropriate use of signCertificate. The first phase pulls all operations on the private key into the "plugin" file. Hence, they move. 

During the second phase inappropriate access to keys is replaced by a Make CSR->Sign CSR. 

Make CSR moves to the outer side, Sign on the "plugin" side. 

> 
>     2. Some of the methods on the CA struct didn't move to the new `internal_ca.go` file and I'm not sure why - this means we have some methods on that struct in `internal_ca.go` but others are still in `cert.go`, which isn't ideal. (and I think may be the reason for some of the function wrapping?)

See the last commit. The last commit rearranges the API so that internal_ca contains ONLY methods which are needed to do a plugin.
These are :

1. Init/Load
2. Sign
3. Provide cert bundle.

Everything else can be derived from these 3. 

> 
> 
> If you could update to fix those two issues I think that would be the next step.

See above. internal_ca will should not have any other methods as it does not need them. You need to init, sign and provide bundle as needed to init a TLS context. 

So if you have a vault plugin it will have an init and 2 methods - sign and provide bundle. Ditto for others. 

As far as the moves I can remove the moves from the first commit. It will actually make the sequence more confusing  as the private key ops will remain sprinkled all over the place.

internal/crypto/internal_ca.go
kot-begemot-uk: It think it is removed later on. It is not portable to plugins. At all.

internal/crypto/internal_ca.go
kot-begemot-uk: Yes. 

Phase 1 - move all private key operations to internal_ca.go

We get an ugly, but functional "plugin" - all ops in one place

Phase 2 - remove all private key ops in signing. Some wrappers pop up from there and end up in the outer. 

Phase 3 - API is cleaned up for pluggability

internal/crypto/cert.go
kot-begemot-uk: Same answer - it does not touch private keys and cannot be expressed as a CSR->Sign operation in the later phases.

internal/crypto/cert.go
kot-begemot-uk: Otherwise we have to supply key use constants in the upper layers. F.E. - we need to modify the certificatesigningrequest.go in the service package to pass key usage.

I am happy either way. I did not want to touch those without necessity and I wanted to keep the changes mostly to crypto/

lkatalin: What if we just used the `internal_ca`'s `IssueRequestedCertificateAsX509` directly, and some other function to translate it to bytes as necessary? I think it would be easier to follow and there could be fewer functions overall. Whatever is calling this `IssueRequestedClientCertificateAsX509` could just call `ca.IssueRequestedCertificateAsX509` with these parameters.

I see in certificatesigningrequest we do call it [here](https://github.com/flightctl/flightctl/blob/main/internal/service/certificatesigningrequest.go#L92). I think in this case it will be fine to change that line to the new function name and parameters.

lkatalin (commented):
I would suggest removing the wrapper functions in `cert.go` and just using the `internal_ca`'s cert issuing function from `certificatesigningrequest.go`. We could see what it looks like after that change.

lkatalin:
> Make CSR moves to the outer side, Sign on the "plugin" side.
> So if you have a vault plugin it will have an init and 2 methods - sign and provide bundle. Ditto for others.

Okay, I see the direction and I think that makes sense as a separation of concerns.

kot-begemot-uk:
Updated version on top of the recent CSR fixes.

test/harness/test_harness.go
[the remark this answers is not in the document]
hexfusion: errors should always be handeled even if the assumption is an error is not possible. we can not know the future so we should protect against it.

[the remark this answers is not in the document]
kot-begemot-uk: Ack. It is indeed the case where an error is practically impossible - it will be a fail of X509 which most likely came from PEM in the first place to encode again to PEM. 
Will fix.

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,160 @@
...
+	expire := time.Duration(expirySeconds) * time.Second
+	template := &x509.Certificate{
+		Subject:               csr.Subject,
+		Signature:             csr.Signature,
+		SignatureAlgorithm:    csr.SignatureAlgorithm,
+		PublicKey:             csr.PublicKey,
+		PublicKeyAlgorithm:    csr.PublicKeyAlgorithm,
+		IPAddresses:           csr.IPAddresses,
+		DNSNames:              csr.DNSNames,
+		Issuer:                caBackend.Config.Certs[0].Subject,
```
hexfusion, line 146: can we add a note that the length is ensured by signCertificate or would it make sense to ensure here

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (caBackend *internalCA) signCertificate(template *x509.Certificate, requestKey crypto.PublicKey) (*x509.Certificate, error) {
+	// Increment and persist serial
+	serial, err := caBackend.SerialGenerator.Next(template)
+	if err != nil {
+		return nil, err
+	}
+	template.SerialNumber = big.NewInt(serial)
+	return signCertificate(template, requestKey, caBackend.Config.Certs[0], caBackend.Config.Key)
```
hexfusion, line 113: again the length freaks me out could you expose a method that returns the cert and ensures its length as an option

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,160 @@
...
+
+	oscrypto "github.com/openshift/library-go/pkg/crypto"
+)
+
+type internalCA struct {
+	Config          *TLSCertificateConfig
+	SerialGenerator oscrypto.SerialGenerator
+}
+
+func ensureInternalCA(certFile, keyFile, serialFile, subjectName string, expireDays int) (CABackend, bool, error) {
```
hexfusion, line 21: can you add some go docs on this specifically what the bool represents

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,160 @@
...
+	if ca, err := GetCA(certFile, keyFile, serialFile); err == nil {
+		return ca, false, err
+	}
+	ca, err := MakeSelfSignedCA(certFile, keyFile, serialFile, subjectName, expireDays)
+	return ca, true, err
```
hexfusion, line 26: if this is an error the general expectation that the rest of the values are discarded. here it seems you want to pass true if its an error or not. is that expected

internal/crypto/cert.go
```diff
@@ -194,65 +97,46 @@ func (ca *CA) MakeAndWriteServerCert(certFile, keyFile string, hostnames []strin
...
 	certConfig, err := GetClientCertificate(certFile, keyFile, subjectName)
 	if err != nil {
-		certConfig, err = ca.MakeClientCertificate(certFile, keyFile, subjectName, expireDays)
-		return certConfig, true, err // true indicates we wrote the files.
+		certConfig, err := caClient.MakeClientCert(certFile, keyFile, subjectName, expireDays)
+		if err != nil {
+			return nil, false, err
+		}
+		err = certConfig.WriteCertConfigFile(certFile, keyFile)
+		return certConfig, err == nil, err // true indicates we wrote the files.
```
hexfusion, line 139: isnt the presence of the error itself indicative of writing files why do we need also a bool

hexfusion (commented):
first passa

test/util/util.go
```diff
@@ -125,7 +131,12 @@ func NewTestApiServer(log logrus.FieldLogger, cfg *config.Config, store store.St
 // NewTestServer creates a new test server and returns the server and the listener listening on localhost's next available port.
 func NewTestAgentServer(log logrus.FieldLogger, cfg *config.Config, store store.Store, ca *crypto.CA, serverCerts *crypto.TLSCertificateConfig, queuesProvider queues.Provider) (*agentserver.AgentServer, net.Listener, error) {
 	// create a listener using the next available port
-	_, tlsConfig, err := crypto.TLSConfigForServer(ca.Config, serverCerts)
+	_, tlsConfig, err := crypto.TLSConfigForServer(ca.GetCABundleX509(), serverCerts)
+
+	pemCert, _ := ca.GetCABundle()
+
+	log.Infof("Agent Server caCert %s", pemCert)
```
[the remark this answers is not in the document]
kot-begemot-uk: I forgot to remove it. Debug output :)

internal/crypto/internal_ca.go
kot-begemot-uk: Ack

internal/crypto/internal_ca.go
kot-begemot-uk: History. 
I did not change this - it is inherited from the original CA code and signifies was the CA loaded or generated on the fly.

internal/crypto/cert.go
kot-begemot-uk: See elsewhere - generated versus loaded.

internal/crypto/internal_ca.go
kot-begemot-uk: That should happen one level up. At this point (in the backend) we return crypto.x509
I will add it to the frontend

internal/crypto/cert.go
```diff
@@ -28,6 +25,30 @@ var allowedPrefixes = [...]string{
...
+// In case of errors a non-nil error is returned.
+func EnsureCA(certFile, keyFile, serialFile, subjectName string, expireDays int) (*CA, bool, error) {
+	caBackend, fresh, err := ensureInternalCA(certFile, keyFile, serialFile, subjectName, expireDays)
+	if err != nil {
+		return nil, fresh, err
+	}
+	ret := &CA{
+		caBackend: caBackend,
+	}
+	return ret, fresh, err
```
hexfusion, line 49:
```suggestion
	return ret, fresh, nil
```

internal/crypto/cert.go
```diff
@@ -50,124 +71,10 @@ func CNFromDeviceFingerprint(fingerprint string) (string, error) {
...
-
-// func (ca *CA) EnsureSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, bool, error) {
-// 	if subCA, err := GetCA(certFile, keyFile, serialFile); err == nil {
-// 		return subCA, false, err
-// 	}
-// 	subCA, err := ca.MakeAndWriteSubCA(certFile, keyFile, serialFile, name, expireDays)
-// 	return subCA, true, err
-// }
-
-func (ca *CA) EnsureServerCertificate(certFile, keyFile string, hostnames []string, expireDays int) (*TLSCertificateConfig, bool, error) {
```
hexfusion, line 167: just my 2c this naming is more idiomatic, the short name makes it more straightforward to reason about that this is a method off ca. where caClient feels like an internal object.

internal/crypto/cert.go
```diff
@@ -361,6 +214,48 @@ func (c *TLSCertificateConfig) GetPEMBytes() ([]byte, []byte, error) {
...
+	return certData, nil
+}
+
+func (caClient *CA) IssueRequestedServerCertificateAsX509(csr *x509.CertificateRequest, expirySeconds int) (*x509.Certificate, error) {
+	return caClient.caBackend.IssueRequestedCertificateAsX509(csr, expirySeconds, []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth})
+}
+func (caClient *CA) IssueRequestedServerCertificate(csr *x509.CertificateRequest, expirySeconds int) ([]byte, error) {
+	cert, err := caClient.IssueRequestedServerCertificateAsX509(csr, expirySeconds)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", flterrors.ErrSignCert, err.Error())
```
hexfusion, line 240:
```suggestion
		return nil, fmt.Errorf("%w: %w", flterrors.ErrSignCert, err)
```

internal/crypto/cert.go
```diff
@@ -361,6 +214,48 @@ func (c *TLSCertificateConfig) GetPEMBytes() ([]byte, []byte, error) {
...
+	}
+	certData, err := oscrypto.EncodeCertificates(cert)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", flterrors.ErrEncodeCert, err.Error())
```
hexfusion, line 244: same

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func GetCA(certFile, keyFile, serialFile string) (*internalCA, error) {
+	ca, err := oscrypto.GetCA(certFile, keyFile, serialFile)
+	if err != nil {
+		return nil, err
+	}
+	config := TLSCertificateConfig(*ca.Config)
+	return &internalCA{Config: &config, SerialGenerator: ca.SerialGenerator}, err
```
hexfusion, line 38:
```suggestion
	return &internalCA{Config: &config, SerialGenerator: ca.SerialGenerator}, nil
```

test/util/util.go
```diff
@@ -106,7 +107,12 @@ func (t *testProvider) Consume(ctx context.Context, handler queues.ConsumeHandle
 // NewTestServer creates a new test server and returns the server and the listener listening on localhost's next available port.
 func NewTestApiServer(log logrus.FieldLogger, cfg *config.Config, store store.Store, ca *crypto.CA, serverCerts *crypto.TLSCertificateConfig, queuesProvider queues.Provider) (*apiserver.Server, net.Listener, error) {
 	// create a listener using the next available port
-	tlsConfig, _, err := crypto.TLSConfigForServer(ca.Config, serverCerts)
+	tlsConfig, _, err := crypto.TLSConfigForServer(ca.GetCABundleX509(), serverCerts)
+
+	pemCert, _ := ca.GetCABundle()
+
+	log.Infof("Server caCert %s", pemCert)
```
hexfusion, line 114: another one

test/util/util.go
```diff
@@ -208,21 +214,20 @@ func NewClient(serverUrl string, caCert *crypto.TLSCertificateConfig) (*client.C
...
 }
 
-func NewBareHTTPsClient(caCert, clientCert *crypto.TLSCertificateConfig) (*http.Client, error) {
+func NewBareHTTPsClient(caBundle []*x509.Certificate, clientCert *crypto.TLSCertificateConfig) (*http.Client, error) {
 
 	httpClient := &http.Client{}
-	if caCert != nil || clientCert != nil {
-		var err error
-		tlsConfig, err := crypto.TLSConfigForClient(caCert, clientCert)
+	if len(caBundle) > 0 || clientCert != nil {
```
hexfusion, line 229: should this be and vs or?

kot-begemot-uk: IMHO it is correct. 
The way test are set up the connection is moved to TLS mode if you have either a CA bundle or a client cert mTLS.
It is an equivalent of the old condition (see the diff). caCert != nil is one or more certs in the bundle concatenated as PEM.

test/util/util.go
kot-begemot-uk: Terminated with extreme prejudice :)

internal/crypto/internal_ca.go
kot-begemot-uk: Ack.

internal/crypto/cert.go
kot-begemot-uk: I would prefer that it is not named CA so that someone who comes and tries to modify this code later one has one more warning that any CA functions (f.e. if you need to add CRLs, explicit renew(), etc) do not belong here. It is a wrapper. Client. The CA is in the backend which is pluggable and has multiple possible implementations.

lkatalin (changes requested):
Some comments inline; possibly more incoming

internal/crypto/cert.go
```diff
@@ -145,66 +135,32 @@ func GetClientCertificate(certFile, keyFile string, subjectName string) (*TLSCer
...
-	clientPublicKeyHash, err := HashPublicKey(clientPublicKey)
-	if err != nil {
-		return nil, err
-	}
-
-	now := time.Now()
-	clientTemplate := &x509.Certificate{
 
-		SignatureAlgorithm: x509.ECDSAWithSHA256,
+func (ca *CA) MakeClientCert(certFile, keyFile string, subjectName string, expiryDays int) (*TLSCertificateConfig, error) {
```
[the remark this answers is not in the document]
kot-begemot-uk: It is inconsistent at source. Most of the original cert is cargoculted out of https://github.com/openshift/library-go/

That, in turn uses both and is not particularly consistent. I am happy to pick one and stick with it.

cmd/flightctl-api/main.go
```diff
@@ -106,7 +106,13 @@ func main() {
...
 
 	// also write out a client config file
-	err = client.WriteConfig(config.ClientConfigFile(), cfg.Service.BaseUrl, "", ca.Config, nil)
+
+	caPemBytes, err := ca.GetCABundle()
+	if err != nil {
+		log.Fatalf("loading CA certificate bundle: %v", err)
+	}
+
+	err = client.WriteConfig(config.ClientConfigFile(), cfg.Service.BaseUrl, "", caPemBytes, nil)
```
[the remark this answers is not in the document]
kot-begemot-uk: https://issues.redhat.com/browse/EDM-1318

internal/crypto/cert.go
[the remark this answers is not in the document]
lkatalin: Thanks. Could you please avoid making the change in the second commit

internal/crypto/internal_ca.go
```diff
@@ -133,42 +128,29 @@ func signCertificate(template *x509.Certificate, requestKey crypto.PublicKey, is
...
+		Signature:             csr.Signature,
+		SignatureAlgorithm:    csr.SignatureAlgorithm,
+		PublicKey:             csr.PublicKey,
+		PublicKeyAlgorithm:    csr.PublicKeyAlgorithm,
+		IPAddresses:           csr.IPAddresses,
+		DNSNames:              csr.DNSNames,
+		Issuer:                ca.Config.Certs[0].Subject,
+		NotBefore:             now.Add(-time.Second),
+		NotAfter:              now.Add(expire),
+		SerialNumber:          big.NewInt(1),
```
lkatalin, line 155: Should the SerialNumber not be generated with the CA's `SerialGenerator`? I'm not very familiar with this but when I looked it up it looked like it would be relevant

lkatalin: Looks like this is from pre-existing code anyway

internal/crypto/internal_ca.go
```diff
@@ -0,0 +1,166 @@
...
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           usage,
+		BasicConstraintsValid: true,
+		AuthorityKeyId:        caBackend.Config.Certs[0].SubjectKeyId,
+	}
+	return caBackend.signCertificate(template, csr.PublicKey)
+}
```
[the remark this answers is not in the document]
kot-begemot-uk: This is not needed - if something is wrong with the CA certs, it will fail the initialization checks in EnsureCA. A CA which is instantiated successfully will always have certs/keys to sign with.

internal/crypto/internal_ca.go
kot-begemot-uk: That is a non-issue. This is the template, not the sign. The number is correctly generated when the CA signs the cert. 
However, in the process of reviewing it I found that we are generating the CA serial itself "by hand". That is incorrect - it also needs to be generated using the serial number generator. It is the first number for a sequential or random if all certs are random.

lkatalin (approved):
I think this looks good in terms of:

- Using `GetCABundle()` to get just the CA's cert instead of its cert + key with the previous `ca.Config` in functions that are not part of the internal CA

- Separating the previous CA functionality into `internalCA` (which may in the future be an external plugged-in CA) and `CAClient` that will interact with it

- Cleaning up templates for Server and Client certs on the CA Client side and separating CSR creation into its own function using `x509` before passing the CSR to the internal CA

PR #983, opened 2025-03-11
EDM-1025: Use ServiceHandler rather than Store in tasks

avishayt:


internal/tasks/repotester.go
```diff
@@ -41,38 +40,51 @@ func (r *RepoTester) TestRepositories() {
...
+			return
 		}
 
-		accessErr := r.TypeSpecificRepoTester.TestAccess(&repository)
+		for i := range repositories.Items {
+			repository := repositories.Items[i]
+
+			repoSpec, _ := repository.Spec.GetGenericRepoSpec()
+			switch repoSpec.Type {
+			case "http":
```
hexfusion, line 61: nit: could we make these consts :pray:

internal/tasks/resourcesync.go
```diff
@@ -53,42 +50,56 @@ func (r *ResourceSync) Poll() {
...
+			Continue: continueToken,
+		})
+		if status.Code != 200 {
+			log.Errorf("error fetching repositories: %s", status.Message)
+			return
+		}
+
+		for i := range resourcesyncs.Items {
+			rs := &resourcesyncs.Items[i]
+			_ = r.run(ctx, log, rs)
```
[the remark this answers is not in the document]
hexfusion: always for more observabilty

internal/tasks/resourcesync.go
```diff
@@ -97,68 +108,64 @@ func (r *ResourceSync) run(ctx context.Context, log logrus.FieldLogger, rs *mode
...
-			err = fmt.Errorf("one or more fleets are managed by a different resource. %w", err)
+		_, status := r.serviceHandler.ReplaceFleet(ctx, *resource.Metadata.Name, *resource)
+		if status.Code != http.StatusOK && status.Code != http.StatusCreated {
+			var err error
+			if status.Message == flterrors.ErrUpdatingResourceWithOwnerNotAllowed.Error() {
+				err = fmt.Errorf("one or more fleets are managed by a different resource")
+			} else {
+				err = service.ApiStatusToErr(status)
+			}
+			errs = append(errs, err)
```
hexfusion, line 167: optional: no need to allocate

```suggestion
			if status.Message == flterrors.ErrUpdatingResourceWithOwnerNotAllowed.Error() {
			    errs = append(errs, fmt.Errorf("one or more fleets are managed by a different resource"))
			} else {
    			    errs = append(errs, service.ApiStatusToErr(status))
			}
```

hexfusion (commented):
couple small notes otherwise looks clean

internal/tasks/repotester.go
avishayt: Yes, for sure - especially when they're already defined :)

internal/tasks/resourcesync.go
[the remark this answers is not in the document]
avishayt: Added

internal/tasks/resourcesync.go
avishayt: Done

kot-begemot-uk:
Using service handler instead of store means that tasks cannot be used as a fully featured interface for a CA plugin. 

Server certificates will have to be generated offline and renewed offline.

kot-begemot-uk:
In order to initialize the API handler you have to have a valid certificate. At present the certificate is autogenerated and this needs to be done before the API handler starts.

The CA plugin needs to be asynchronous by design and the agreement so far was to make it run as a task. If we keep that design it needs to be able to convey certificates to the front-end. As the API mTLS endpoint is not yet initialized (it has no cert), the only means of doing this is the database. It cannot be done via the API handler.

Additionally, there may be other pieces of data which the CA may need to convey to the API from time to time - the CA certificate bundle, CRLs, etc.


Options going forward:
1. Run it async, but not in tasks. Thread in the API handler. 
2. Run it in tasks, have access to the database via existing data structures (keep database access, not just API handler)
3. Run it in tasks, have access to the database via a table just for this purpose so that we do not deal with access control via the API handler.

avishayt:
> In order to initialize the API handler you have to have a valid certificate. At present the certificate is autogenerated and this needs to be done before the API handler starts.

We recently split the service layer into transport+service.  So the API service layers are transport (HTTP) -> service (validation, tenancy, etc) -> store (DB).  This PR changes the worker service to tasks -> service -> store.  Certificates should be dealt with at the transport layer.

kot-begemot-uk:
If you cannot initialize the transport layer, you cannot have access to the service layer, right?
If you do not have certificates, you cannot initialize the transport layer, right?

I am happy if we kill the autogeneration completely by the way. It will make life easier.

avishayt:
> If you cannot initialize the transport layer, you cannot have access to the service layer, right?

Wrong, this PR has flightctl-worker and flightctl-periodic initializing the service layer but not the transport layer.

> If you do not have certificates, you cannot initialize the transport layer, right?

Probably right.

> I am happy if we kill the autogeneration completely by the way. It will make life easier.

Which autogeneration?

kot-begemot-uk:
EnsureServerCertificate(), EnsureCACertificate(), EnsureClientCertificate() generating self-signed certs instead of failing if a cert is not present.

PR #986, opened 2025-03-11
EDM-1214: Perform better validation of CSR and enrollment requests

kot-begemot-uk:
1. Add comparisons between ASN.1 and JSON CSR metadata
2. Add checks between mTLS identity and requested CSR identity to allow the use of the CSR interface for renewals
3. Make adding prefixes one-off and allow for prefixes in CN comparisons
4. This should allow the use of CSR interface for renewals

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced validation checks during certificate signing requests to ensure proper handling of Common Names (CN), preventing identity mismatches and unauthorized certificate renewals.
  - Improved error handling to provide clearer feedback on CN discrepancies.
  - Strengthened validation in enrollment requests to ensure only valid and expected CNs are processed, reducing the risk of identity theft.
  - Added checks for specific prefixes in names and fingerprints to streamline processing.
  - Introduced a new method to validate Certificate Signing Request (CSR) parameters, enhancing error handling.

### DIFF
--- a/internal/cli/certificate.go
+++ b/internal/cli/certificate.go
@@ -6,6 +6,7 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
@@ -281,9 +282,14 @@ func createCsr(o *CertificateOptions, name string, priv crypto.PrivateKey) ([]by
 		return nil, err
 	}
 
-	// the CN is going to be a FC-generated UUID, populated at signing time
+	if name == "" {
+		name = uuid.NewString()
+	}
 	template := &x509.CertificateRequest{
 		SignatureAlgorithm: x509.ECDSAWithSHA256,
+		Subject: pkix.Name{
+			CommonName: name,
+		},
 	}
 
 	csrInner, err := x509.CreateCertificateRequest(rand.Reader, template, priv)
--- a/internal/crypto/cert.go
+++ b/internal/crypto/cert.go
@@ -10,6 +10,7 @@ import (
 	"math/big"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	oscrypto "github.com/openshift/library-go/pkg/crypto"
@@ -22,13 +23,27 @@ const ClientBootstrapCommonNamePrefix = "client-enrollment-"
 const AdminCommonName = "flightctl-admin"
 const DeviceCommonNamePrefix = "device:"
 
+var allowedPrefixes = [...]string{
+	DeviceCommonNamePrefix,
+	ClientBootstrapCommonNamePrefix,
+}
+
 func BootstrapCNFromName(name string) string {
+
+	for _, prefix := range allowedPrefixes {
+		if strings.HasPrefix(name, prefix) {
+			return name
+		}
+	}
 	return ClientBootstrapCommonNamePrefix + name
 }
 
 func CNFromDeviceFingerprint(fingerprint string) (string, error) {
 	if len(fingerprint) < 16 {
 		return "", errors.New("device fingerprint must have 16 characters at least")
+	}
+	if strings.HasPrefix(fingerprint, DeviceCommonNamePrefix) {
+		return fingerprint, nil
 	}
 	return DeviceCommonNamePrefix + fingerprint, nil
 }
--- a/internal/service/certificatesigningrequest.go
+++ b/internal/service/certificatesigningrequest.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	api "github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/api_server/middleware"
 	"github.com/flightctl/flightctl/internal/crypto"
 	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/store"
@@ -77,11 +78,18 @@ func signApprovedCertificateSigningRequest(ca *crypto.CA, request api.Certificat
 	}
 
 	// the CN will need the enrollment prefix applied;
-	// if a CN is not specified in the CSR, generate a UUID to represent the future device
+	// if the certificate is being renewed, the name will have an existing prefix.
+	// we do not touch in this case.
+
 	u := csr.Subject.CommonName
-	if u == "" {
-		u = uuid.NewString()
+
+	// Once we move all prefixes/name formation to the client this can become a simple
+	// comparison of u and *request.Metadata.Name
+
+	if crypto.BootstrapCNFromName(u) != crypto.BootstrapCNFromName(*request.Metadata.Name) {
+		return nil, fmt.Errorf("%w - CN %s Metadata %s mismatch", flterrors.ErrSignCert, u, *request.Metadata.Name)
 	}
+
 	csr.Subject.CommonName = crypto.BootstrapCNFromName(u)
 
 	expiry := DefaultEnrollmentCertExpirySeconds
@@ -155,6 +163,27 @@ func (h *ServiceHandler) ListCertificateSigningRequests(ctx context.Context, par
 	}
 }
 
+func (h *ServiceHandler) verifyCSRParameters(ctx context.Context, csr api.CertificateSigningRequest) error {
+
+	// Crypto validation
+	cn, ok := ctx.Value(middleware.TLSCommonNameContextKey).(string)
+
+	// Note - if auth is disabled and there is no mTLS handshake we get ok == False.
+	// We cannot check anything in that case.
+
+	if ok {
+		if csr.Spec.SignerName != "enrollment" {
+			if csr.Metadata.Name == nil {
+				return errors.New("invalid csr record - no name in metadata")
+			}
+			if cn != crypto.BootstrapCNFromName(*csr.Metadata.Name) {
+				return errors.New("denied attempt to renew other entity certificate")
+			}
+		}
+	}
+	return nil
+}
+
 func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, csr api.CertificateSigningRequest) (*api.CertificateSigningRequest, api.Status) {
 	orgId := store.NullOrgId
 
@@ -166,6 +195,11 @@ func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, cs
 		return nil, api.StatusBadRequest(errors.Join(errs...).Error())
 	}
 
+	err := h.verifyCSRParameters(ctx, csr)
+	if err != nil {
+		return nil, api.StatusUnauthorized(err.Error())
+	}
+
 	result, err := h.store.CertificateSigningRequest().Create(ctx, orgId, &csr)
 	if err != nil {
 		return nil, StoreErrorToApiStatus(err, true, api.CertificateSigningRequestKind, csr.Metadata.Name)
@@ -174,6 +208,7 @@ func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, cs
 	if result.Spec.SignerName == "enrollment" {
 		h.autoApprove(ctx, orgId, result)
 	}
+
 	if api.IsStatusConditionTrue(result.Status.Conditions, api.CertificateSigningRequestApproved) {
 		h.signApprovedCertificateSigningRequest(ctx, orgId, result)
 	}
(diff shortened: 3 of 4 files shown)


internal/service/certificatesigningrequest.go
```diff
@@ -165,6 +168,15 @@ func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, cs
 	if errs := csr.Validate(); len(errs) > 0 {
 		return nil, api.StatusBadRequest(errors.Join(errs...).Error())
 	}
+	// Crypto validation
+	cn, ok := ctx.Value(middleware.TLSCommonNameContextKey).(string)
+	if ok {
```
lkatalin, line 173: What do we do if `!ok`?

lkatalin (commented):
A few small comments / questions

internal/service/certificatesigningrequest.go
```diff
@@ -79,9 +80,11 @@ func signApprovedCertificateSigningRequest(ca *crypto.CA, request api.Certificat
 	// the CN will need the enrollment prefix applied;
 	// if a CN is not specified in the CSR, generate a UUID to represent the future device
 	u := csr.Subject.CommonName
-	if u == "" {
-		u = uuid.NewString()
+
+	if crypto.BootstrapCNFromName(u) != crypto.BootstrapCNFromName(*request.Metadata.Name) {
```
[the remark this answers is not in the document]
kot-begemot-uk: If we allow both client and server side CN formation, one can be transformed the other one not.
Example - client supplies a "name". At present it is the server which adds enrollment- to it. So a comparison needs to check for both cases because it may or may not contain it.
If we move all CN modifications to the client the CNF transform as a part of the check is no longer necessary.

internal/crypto/cert.go
```diff
@@ -23,13 +24,22 @@ const AdminCommonName = "flightctl-admin"
 const DeviceCommonNamePrefix = "device:"
 
 func BootstrapCNFromName(name string) string {
+	if strings.HasPrefix(name, ClientBootstrapCommonNamePrefix) {
```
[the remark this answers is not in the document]
kot-begemot-uk: I had it in an earlier version. Sure. We can do that.

internal/service/certificatesigningrequest.go
kot-begemot-uk: Not OK means we are running with auth disabled. We simply skip all of this as it does not apply.

internal/service/certificatesigningrequest.go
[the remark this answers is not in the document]
lkatalin: Okay, that makes sense. It's fine this way

internal/service/certificatesigningrequest.go
```diff
@@ -165,6 +173,19 @@ func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, cs
...
+	// Crypto validation
+	cn, ok := ctx.Value(middleware.TLSCommonNameContextKey).(string)
+
+	// Note - if auth is disabled and there is no mTLS handshake we get ok == False.
+	// We cannot check anything in that case.
+
+	if ok {
+		if csr.Spec.SignerName != "enrollment" {
+			if cn != crypto.BootstrapCNFromName(*csr.Metadata.Name) {
+				return nil, api.StatusUnauthorized(errors.New("Denied attempt to renew other entity certificate").Error())
```
hexfusion, line 185: no need to create an error if you need a string?
```suggestion
				return nil, api.StatusUnauthorized("Denied attempt to renew other entity certificate")
```

kot-begemot-uk: Fair point. Will do.

internal/service/enrollmentrequest.go
```diff
@@ -33,11 +33,21 @@ func approveAndSignEnrollmentRequest(ca *crypto.CA, enrollmentRequest *api.Enrol
 		return fmt.Errorf("approveAndSignEnrollmentRequest: error parsing CSR: %w", err)
 	}
 
-	csr.Subject.CommonName, err = crypto.CNFromDeviceFingerprint(*enrollmentRequest.Metadata.Name)
+	supplied, err := crypto.CNFromDeviceFingerprint(csr.Subject.CommonName)
+	if err != nil {
+		return fmt.Errorf("approveAndSignEnrollmentRequest: invalid CN supplied in CSR: %w", err)
+	}
+
+	desired, err := crypto.CNFromDeviceFingerprint(*enrollmentRequest.Metadata.Name)
```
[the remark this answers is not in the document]
kot-begemot-uk: No with a legit client - it is always initialized. This is not an excuse not to check.

kot-begemot-uk:
Good points, I will have a look Replace, though it somewhat returns us to our perennial discussion - do we keep CSRs or not :)

kot-begemot-uk:
I had a look. That is YET ANOTHER name. Name from the query string parameters. 
All other checks should still be performed in this case so we should abstract them into a func.

lkatalin:
> I had a look. That is YET ANOTHER name - the name from the query string parameters. All other checks should still be performed in this case so we should abstract them into a func.

Yeah, it may be good to put all the validations in one place (such as `api/v1alpha/validation.go` and the corresponding `internal/util/validation/validation.go`) and use them in both `Create...` and `Replace`.

internal/cli/certificate.go
```diff
@@ -278,9 +279,14 @@ func createCsr(o *CertificateOptions, name string, priv crypto.PrivateKey) ([]by
 		return nil, err
 	}
 
-	// the CN is going to be a FC-generated UUID, populated at signing time
+	if name == "" {
+		name = uuid.NewString()
+	}
```
lkatalin, line 284: Also, we should do this earlier - maybe as the first step in `Run()` [here](https://github.com/flightctl/flightctl/blob/main/internal/cli/certificate.go#L142) - and populate this into `o.Name` to make sure that the files written out for the user correspond to this CN as well.

internal/service/certificatesigningrequest.go
```diff
@@ -166,6 +195,12 @@ func (h *ServiceHandler) CreateCertificateSigningRequest(ctx context.Context, cs
 		return nil, api.StatusBadRequest(errors.Join(errs...).Error())
 	}
 
+	ok, err := h.verifyCSRParameters(ctx, csr)
+
```
hexfusion, line 199: style; drop newline

internal/service/certificatesigningrequest.go
```diff
@@ -254,6 +290,12 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, n
 		return nil, api.StatusBadRequest("resource name specified in metadata does not match name in path")
 	}
 
+	ok, err := h.verifyCSRParameters(ctx, csr)
```
hexfusion, line 293: just something to consider what value does the bool provide here that a non nil error could not solve

kot-begemot-uk: Fair point.

internal/cli/certificate.go
lkatalin: That's okay, I can also do it as a follow-on PR if you want.

kot-begemot-uk: Let's do it this way.

internal/service/certificatesigningrequest.go
```diff
@@ -254,6 +289,12 @@ func (h *ServiceHandler) ReplaceCertificateSigningRequest(ctx context.Context, n
 		return nil, api.StatusBadRequest("resource name specified in metadata does not match name in path")
 	}
 
+	err := h.verifyCSRParameters(ctx, csr)
+
+	if err != nil {
```
hexfusion, line 294: style: drop newline

internal/service/certificatesigningrequest.go
```diff
@@ -155,6 +163,27 @@ func (h *ServiceHandler) ListCertificateSigningRequests(ctx context.Context, par
...
+	// Crypto validation
+	cn, ok := ctx.Value(middleware.TLSCommonNameContextKey).(string)
+
+	// Note - if auth is disabled and there is no mTLS handshake we get ok == False.
+	// We cannot check anything in that case.
+
+	if ok {
+		if csr.Spec.SignerName != "enrollment" {
+			if csr.Metadata.Name == nil {
+				return errors.New("Invalid csr record - no name in metadata")
```
hexfusion, line 177: style: errors should not start with a capital letter

internal/crypto/cert.go
```diff
@@ -22,14 +23,28 @@ const ClientBootstrapCommonNamePrefix = "client-enrollment-"
 const AdminCommonName = "flightctl-admin"
 const DeviceCommonNamePrefix = "device:"
 
+var allowedPrefixes = [...]string{
+	DeviceCommonNamePrefix,
+	ClientBootstrapCommonNamePrefix,
+}
+
```
fzdarsky, line 30: I'm aware this isn't what the PR tries to address, but could we make the use of prefixes a bit more consistent?
Maybe agree on a format of `[prefix:]name` (using `:` as separator, unlike in `client-enrollment-`)? Shorten `client-enrollment` to `enrollment`? Are we even using the `AdminCommonName` and do we need prefixes for users/service acconts?

internal/service/enrollmentrequest.go
```diff
@@ -33,11 +33,21 @@ func approveAndSignEnrollmentRequest(ca *crypto.CA, enrollmentRequest *api.Enrol
...
 	if err != nil {
 		return fmt.Errorf("approveAndSignEnrollmentRequest: error setting CN in CSR: %w", err)
 	}
 
+	if desired != supplied {
+		return fmt.Errorf("approveAndSignEnrollmentRequest: attempt to supply a fake CN, possible identity thef, csr: %s, metadata %s", supplied, desired)
```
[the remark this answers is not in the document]
hexfusion: @kot-begemot-uk why not clean this up?

internal/crypto/cert.go
kot-begemot-uk: If we are going to address this and make it consistent, can I suggest we obsolete them. Fully. 

Not in this PR of course.

This is what OU field is usually used for. 

Then the question of separator does not stand as it is a separate ASN.1 field to start off with which is added to the subject by default.

internal/service/enrollmentrequest.go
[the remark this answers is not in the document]
kot-begemot-uk: Missed it :)
Will fix it right away.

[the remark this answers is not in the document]
kot-begemot-uk: It has been fixed. Just the message is different because the code for all cases was merged into a separate validate function. 

```
		if csr.Spec.SignerName != "enrollment" {
			if csr.Metadata.Name == nil {
				return errors.New("invalid csr record - no name in metadata")
			}
			if cn != crypto.BootstrapCNFromName(*csr.Metadata.Name) {
				return errors.New("denied attempt to renew other entity certificate")
			}
		}
```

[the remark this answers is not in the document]
hexfusion: https://github.com/flightctl/flightctl/blob/f06add45a6bc07c1fd330816250c8e8b8562d9a0/internal/service/enrollmentrequest.go#L47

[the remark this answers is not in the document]
kot-begemot-uk: Sorry, looking at wrong file. Fixed and pushed.

internal/service/enrollmentrequest.go
```diff
@@ -33,11 +33,21 @@ func approveAndSignEnrollmentRequest(ca *crypto.CA, enrollmentRequest *api.Enrol
...
+		return fmt.Errorf("approveAndSignEnrollmentRequest: invalid CN supplied in CSR: %w", err)
+	}
+
+	desired, err := crypto.CNFromDeviceFingerprint(*enrollmentRequest.Metadata.Name)
 	if err != nil {
 		return fmt.Errorf("approveAndSignEnrollmentRequest: error setting CN in CSR: %w", err)
 	}
 
+	if desired != supplied {
+		return fmt.Errorf("approveAndSignEnrollmentRequest: attempt to supply a fake CN, possible identity theft, csr: %s, metadata %s", supplied, desired)
```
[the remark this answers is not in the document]
kot-begemot-uk: The CN itself is not particularly private. Neither is the metadata name. 

My assessment is that it is OK and it needs to be logged.

hexfusion (approved):

lkatalin (approved):

PR #987, opened 2025-03-11
NO-ISSUE: Update CLI login docs

rawagner:
## Summary by CodeRabbit

- **Documentation**
	- Refined login instructions for both CLI and web interfaces with a more consistent command format.
	- Updated guidance for entering user credentials and clarified the distinction between headless and web methods.
	- Enhanced documentation on managing TLS verification issues for self-signed certificates by providing additional options for certificate handling.
	- Emphasized the requirement for the OIDC provider to have Direct Access grant enabled for headless login.
	- Added a new section for gathering deployment information, including access URLs for the Flight Control UI and API.
	- Introduced commands for storing API URL and demo user password in environment variables for improved usability.
- **Chores**
	- Added a new entry, `API_URL`, to the spelling configuration for enhanced spelling checks.


fzdarsky (approved):

docs/user/getting-started.md
```diff
@@ -199,52 +199,51 @@ Finally, move it into a location within your shell's search path.
...
-Use the CLI to log into the Flight Control Service:
+You will be logging in with ACM login credentials.
 
-```console
-$ flightctl login https://api.flightctl.127.0.0.1.nip.io/ --web --insecure-skip-tls-verify
+Headless login:
 
-[...]
+```console
+flightctl login <flightctl_api_url> -u <acm_user> -p <password>
```
fzdarsky, line 226: How can the user figure out what that URL is?

docs/user/getting-started.md
```diff
@@ -199,52 +199,51 @@ Finally, move it into a location within your shell's search path.
...
 ```
 
-In the web browser that opens, use your ACM login credentials.
-
-Verify you can now access the service via the CLI:
+Login using a user token:
 
 ```console
-$ flightctl get devices
+flightctl login <flightctl_api_url> -t $(oc whoami -t)
```
fzdarsky, line 232: Instead of listing all options, should we only list this one, which for an ACM user seems to be the most reliable?
Is there any benefit to the other ones? Ok, maybe that you do not need to login via `oc` first.

docs/user/getting-started.md
rawagner: For ACM deployment, it will always be through looking at routes, but the Standalone deployment is tricky due to multiple options (Route, Gateway or Node port). After `helm install`, we actually print out API URL so I've added this info to the doc.

docs/user/getting-started.md
rawagner: I agree with the `-t` being the best. Since the guide goes through the helm install process too, i'd say we are already logged it via `oc`.

docs/user/getting-started.md
```diff
@@ -40,9 +39,8 @@ local-path-storage   local-path-provisioner-7577fdbbfb-wxbck      1/1     Runnin
 Verify Helm is installed and can access the cluster:
 
 ```console
-$ helm list
```
rawagner, line 43: the commands the do not have any output in the `console`, i've removed the `$`sign. It makes it much easier to copy.

avishayt (approved):

PR #988, opened 2025-03-11
EDM-1211: api: add application inline provider

hexfusion:
```
spec:
  applications:
   - name: my-app
     appType: helm <-- possible future support
      inline:
        - path: "Chart.yaml"
          content: "YXBwTmFtZTogbXktYXBwCmFwcFR5cGU6IHdlYgo"
          contentEncoding: base64
        - path: "templates/deployment.yaml"
          content: "YXBwTmFtZTogbXktYXBwCmFwcFR5cGU6IHdlYgo"
          contentEncoding: base64 
        - path: "templates/something/.gitkeep"
  - name: my-app-2
    appType: compose
    inline:
      - path: "docker-compose.yaml"
        content: |
          version: "3.8"
          services:
          app:
            image: "docker.io/nginx:latest"
            ports:
              - "80:80"
```

## Summary by CodeRabbit

- **New Features**
  - Updated API specifications to offer enhanced flexibility in file and application configurations, including support for new file path and encoding options.
  - Added support for inline application provider configurations.
  - Introduced a new validation function for relative paths.

- **Refactor**
  - Streamlined application provider types for improved consistency and type safety.
  - Enhanced validation logic for configuration content and file paths.
  - Updated method signatures to align with new type definitions.

- **Tests**
  - Expanded test coverage to verify new validations and configuration changes.
  - Added tests for the new relative path validation functionality.
  - Introduced tests for the inline application provider specifications.
  - Updated existing tests to reflect changes in encoding constants and application provider types.


api/v1alpha1/validation.go
```diff
@@ -684,6 +711,13 @@ func validateAppProvider(app ApplicationProviderSpec, appType ApplicationProvide
 			errs = append(errs, fmt.Errorf("image reference cannot be empty when application name is not provided"))
 		}
 		errs = append(errs, validation.ValidateOciImageReference(&provider.Image, "spec.applications[].image")...)
+	case InlineApplicationProviderType:
```
celdrake, line 714: Looks like there's some missing validation for Inline apps.
If I attempt to PATCH a device to add an inline application, I get 400 request
with the message `Error 400: Bad Request: unsupported application provider type: inline`, and it seems to be coming from https://github.com/flightctl/flightctl/blob/5fb9d71868534401e54f18b8245978331a19b349/api/v1alpha1/validation.go#L762

Patch body is
```
[{"path":"/spec/applications","op":"replace","value":[{"type":"compose","inline":{"encoding":"base64","data":{"Chart.yaml":"YXBwTmFtZTogbXktYXBwCmFwcFR5cGU6IHdlYgo="}}}]}]
```

In this regard, I imagine `name` should be required for inline applications, while we may want to keep it optional for `ImageApplicationProviderType`?

api/v1alpha1/openapi.yaml
```diff
@@ -3650,6 +3669,29 @@ components:
           description: Reference to the container image for the application package.
       required:
         - image
+    InlineApplicationProvider:
```
celdrake, line 3672: A small issue regarding consistency naming:


For configs, all the provider are named `TypeConfigProviderSpec`:
- GitConfigProviderSpec
- HttpConfigProviderSpec
- InlineConfigProviderSpec
- KubernetesSecretProviderSpec

Then it looks like we didn't name `ImageApplicationProvider` following that convention. That makes the new provider consistent with the existing Application provider, but the application providers are not consistent with the config providers.

api/v1alpha1/validation.go
celdrake: It may be because you're still working on this PR, but FYI there's another missing validation regarding inline applications
https://github.com/flightctl/flightctl/blob/5fb9d71868534401e54f18b8245978331a19b349/internal/tasks/device_render.go#L260

hexfusion: thanks for the heads up, yeah I am working on the updates and will make sure these are all covered

api/v1alpha1/openapi.yaml
```diff
@@ -3285,7 +3285,7 @@ components:
       type: object
       properties:
         target:
-          $ref: '#/components/schemas/DeviceDecommissionTargetType'
+          $ref: "#/components/schemas/DeviceDecommissionTargetType"
```
hexfusion, line 3288: not sure what automation added this diff

api/v1alpha1/openapi.yaml
hexfusion: resolved

api/v1alpha1/validation.go
hexfusion: i stubbed out the support and it will be implemented on the next pr

api/v1alpha1/types.gen.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/oapi-codegen/runtime"
 )
 
+// Defines values for AppType.
+const (
+	AppTypeCompose AppType = "compose"
+)
```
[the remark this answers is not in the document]
hexfusion: helm is not yet supported

hexfusion:
rebase issues

api/v1alpha1/openapi.yaml
```diff
@@ -3639,17 +3668,36 @@ components:
...
+          description: A list of application content.
+          items:
+            $ref: "#/components/schemas/ApplicationContent"
+      required:
+        - inline
+    ApplicationContent:
+        allOf:
+          - $ref: '#/components/schemas/FileContent'
+          - $ref: '#/components/schemas/RelativePath'
+          - type: object
```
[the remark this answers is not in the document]
hexfusion: yeah it seems this is a bug in typescript generators. the usage is verified as valid per the spec. details are outlined here[1]. it makes sense for reuse of objects that in base schema can define it as required. on the other hand I don't want to make a problem for the UI. 

[1] https://github.com/oapi-codegen/oapi-codegen/issues/531

api/v1alpha1/validation.go
```diff
@@ -655,15 +655,44 @@ func (a ApplicationProviderSpec) Validate() []error {
 	return allErrs
 }
 
-func validateApplications(apps []ApplicationProviderSpec) []error {
+func (a InlineApplicationProviderSpec) Validate(fleetTemplate bool) []error {
+	allErrs := []error{}
+	for i := range a.Inline {
```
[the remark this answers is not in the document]
hexfusion: updated the path should be unique per application

api/v1alpha1/validation.go
```diff
@@ -655,15 +655,44 @@ func (a ApplicationProviderSpec) Validate() []error {
...
+			b, _ := base64.StdEncoding.DecodeString(content)
+			_, paramErrs := validateParametersInString(lo.ToPtr(string(b)), fmt.Sprintf("spec.applications[].inline.[%d].content", i), fleetTemplate)
+			allErrs = append(allErrs, paramErrs...)
+		} else if a.Inline[i].ContentEncoding == nil || (a.Inline[i].ContentEncoding != nil && *(a.Inline[i].ContentEncoding) == EncodingPlain) && content != "" {
+			// Contents should be limited to 1MB (1024*1024=1048576 bytes)
+			allErrs = append(allErrs, validation.ValidateString(&content, fmt.Sprintf("spec.applications[].inline.[%d].content", i), 0, maxInlineLength, nil, "")...)
+			_, paramErrs := validateParametersInString(&content, fmt.Sprintf("spec.applications[].inline.[%d].content", i), fleetTemplate)
+			allErrs = append(allErrs, paramErrs...)
+		} else {
+			allErrs = append(allErrs, fmt.Errorf("unknown encoding type: %s", *(a.Inline[i].ContentEncoding)))
```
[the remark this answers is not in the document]
hexfusion: ok this was a change from the first proposed version. encoding should be contentEncoding vs encoding

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
celdrake: In the UI there shouldn't be a problem, as it would be a bug in our code if we would actually send arbitrary fields when updating fleets / devices.

However, I suspect it's more than just the Typescript generator, since I was able to save those extra arbitrary fields. Meaning those using the API directly could run into this issue too. 

So feel free to dismiss the comment as it's not a big deal for the UI.

avishayt (approved):

PR #994, opened 2025-03-13
NO-ISSUE: make SQL validated

gshilin-sdb:
* Avoids Raw SQL Formatting Issues – Uses gorm.Expr to pass expressions safely.
* More Secure – Prevents SQL injection risks by using parameterized queries.
* More Readable & Maintainable – No fmt.Sprintf, making it easier to modify.

## Summary by CodeRabbit

- **Refactor**
	- Improved backend update processes to enhance security, stability, and maintainability. These changes streamline internal operations for better overall system reliability.

### DIFF
--- a/internal/store/device.go
+++ b/internal/store/device.go
@@ -483,16 +483,12 @@ func (s *DeviceStore) UpdateSummaryStatusBatch(ctx context.Context, orgId uuid.U
 		return nil
 	}
 
-	tokens := strings.Repeat("?,", len(deviceNames))
-	// trim tailing comma
-	tokens = tokens[:len(tokens)-1]
-
 	// https://www.postgresql.org/docs/current/functions-json.html
 	// jsonb_set(target jsonb, path text[], new_value jsonb, create_missing boolean)
 	createMissing := "false"
 	return s.db.WithContext(ctx).
 		Model(&model.Device{}).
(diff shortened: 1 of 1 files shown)


internal/store/device.go
```diff
@@ -490,22 +490,17 @@ func (s *DeviceStore) UpdateSummaryStatusBatch(ctx context.Context, orgId uuid.U
...
-	args := make([]interface{}, len(deviceNames))
-	for i, name := range deviceNames {
-		args[i] = name
-	}
-
-	return s.db.WithContext(ctx).Exec(query, args...).Error
+	return s.db.WithContext(ctx).
+		Model(&model.Device{}).
+		Where("name IN ?", deviceNames).
+		UpdateColumns(map[string]interface{}{
```
hexfusion, line 496: I argue that this makes it harder to read. what is the trade-off to making the store more dependent on gorm syntax? I agree that all statements should be evaluated for injection.

gshilin-sdb: * It will create a parameterized query, so each parameter will be validated against SQL injection, for example.
* It's more readable as you can see an intention (update columns), and table and where parts, which is not that easy to see in a printf-based raw SQL.

gshilin-sdb: Let's hope future code reviews will cover this problem :)

PR #995, opened 2025-03-13
EDM-1251: fix JSONB fields in CompositeSelectorResolver

asafbss:
## Summary by CodeRabbit

- **Refactor**
	- Streamlined the field selection process for more consistent query behavior.
	- Enhanced validation rules for selector fields, resulting in clearer error messaging.
	- Updated logic for handling JSONB selectors, improving query construction.
	- Simplified the logic for resolving fields, removing complex handling of array indexing.
	- Removed unnecessary methods related to array handling, reducing complexity.

- **Bug Fixes**
	- Removed outdated array handling to refine JSONB field query operations.
	- Simplified error handling for unsupported operations in field resolution.

- **Documentation**
	- Updated descriptions for the "Exists" operator and "Array" field type in user documentation for clarity.
	- Modified the `FieldName` in the `ResolveSelector` method for improved JSONB querying in developer documentation.
	- Clarified the functionality of the "Exists" operator with practical examples.

These improvements lead to more reliable and accurate custom field queries across the system.


internal/store/model/selectors.go
```diff
@@ -177,3 +153,25 @@ func (m *CertificateSigningRequest) ListSelectors() selector.SelectorNameSet {
 	}
 	return selector.NewSelectorFieldNameSet().Add(keys...)
 }
+
+func makeJSONBSelectorField(selectorName selector.SelectorName, selectorType selector.SelectorType) (*selector.SelectorField, error) {
+	var params strings.Builder
+	parts := strings.Split(selectorName.String(), ".")
+	params.WriteString(parts[0])
```
hexfusion, line 160: question: do we need to check the length first? if not can we document  this assumption?

asafbss: Hey @hexfusion ,

Each defined selector is covered by unit tests. If a given path doesn’t exist in the API structure (e.g., status or spec), the unit test will fail.

In the worst-case scenario, if a developer calls this function with an empty string or a single string, the result will simply be a string.

I can add a length check, but do you think it’s necessary?

hexfusion: It just looks awkward that we assume length, adding a note about the assumption would be helpful for our future selves

asafbss: Thanks, Sam, I understand.
I've added a length check for clarity.

internal/store/selector/resolvers_test.go
```diff
@@ -0,0 +1,93 @@
...
+		t.Fatalf("failed to create CompositeSelectorResolver: %v", err)
+	}
+
+	for input, expectedQuery := range testCases {
+		fieldSelector, err := NewFieldSelector(input)
+		if err != nil {
+			t.Fatalf("failed to parse field selector for %q: %v", input, err)
+		}
+
+		actualTokens, err := fieldSelector.Tokenize(ctx, selectorParserSession{selector: fieldSelector.selector, resolver: resolver})
```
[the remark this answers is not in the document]
asafbss: In that case, a proper error would indicate this

internal/store/selector/selector.go
```diff
@@ -147,44 +145,17 @@ func (sr *selectorFieldResolver) resolveSelector(name SelectorName) (*SelectorFi
...
-				}
-
-				arrayIndex, err := strconv.Atoi(selectorName[strings.Index(selectorName, "[")+1 : len(selectorName)-1])
-				if err != nil {
-					return nil, err
-				}
-
-				if arrayIndex == math.MaxInt {
-					return nil, fmt.Errorf("array index overflow for selector %s", selectorName)
+			if selectorType == Jsonb && selectorName[len(sn.String())] == '.' {
```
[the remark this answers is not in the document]
asafbss: covered :)

avishayt (approved):

internal/store/selector/resolvers_test.go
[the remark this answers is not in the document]
asafbss: I've added the following check to handle this:
https://github.com/flightctl/flightctl/blob/80fbb50ad1b95c86c7eafde6b69a6d2e8dc1f32b/internal/store/selector/field.go#L232

PR #996, opened 2025-03-13
More descriptive errors for CSRs

lkatalin:
Based on the changes suggested in PR #931 - however, we have decided that `usages` will be passed in from the client so those checks will not be removed. The `usages` will be used to populate the certificate in a future PR.

The PR also removes the unused 'ca' signer and adds a 'renewal' signer in preparation for certificate renewals and does more validation of input.


celdrake:
@lkatalin IIRC, the `validateCSRUsages` function will remain.
In that case, someone using the API directly for creating the CSR still needs to "guess" what `usages` they need to use in their API request.
Using this branch's PR, if not sending the `usages`, the API returns error 500. If passing a `usages` array whether empty or with some values, the API returns error 400
```
Error 400: Bad Request: spec.usages: Invalid value: []string{}: required usages must be present in request: [ clientAuth, CA:false, ] 
```

Therefore, my original point that `usages` is not properly documented would still be present with the solution from this PR.  If we stick to making the field required, IMO users should know what values they need to use before having to guess by trial an error.

lkatalin:
Okay good point @celdrake , I'll fix this in another commit and add it to this PR.

internal/util/validation/validation.go
```diff
@@ -259,20 +259,42 @@ func ValidateCSRUsages(u *[]string) []error {
...
+
+	if _, exists := validEnrollmentConfigOutputFormats[s]; exists {
+		return nil
+	}
+
+	msg := "must specify a valid output format. options include: "
+	for k := range validEnrollmentConfigOutputFormats {
+		msg += k + ", "
+	}
+	errs = append(errs, field.Invalid(fieldPathFor("enrollment config output format"), s, msg))
```
fzdarsky, line 297: `fieldPathFor` is for validation errors in complex (JSON) data structures where you need to specify the JSONPath to the specific field having the validation error.

internal/util/validation/validation.go
```diff
@@ -259,20 +259,42 @@ func ValidateCSRUsages(u *[]string) []error {
...
-	errs = append(errs, field.Invalid(fieldPathFor("spec.signerName"), s, "must specify a valid signer"))
+	msg := "must specify a valid signer. options include: "
+	for k := range validSigners {
+		msg += k + ", "
+	}
+	errs = append(errs, field.Invalid(fieldPathFor("spec.signerName"), s, msg))
+	return asErrors(errs)
+}
+
+func ValidateOutputFormat(s string) []error {
```
fzdarsky, line 282: This package is for input validation on the API data types, so mostly used service-side. For CLI flag validation, you probably want to limit the validation code to the respective command.

internal/util/validation/validation.go
```diff
@@ -259,20 +259,42 @@ func ValidateCSRUsages(u *[]string) []error {
...
+	errs := field.ErrorList{}
+
+	if _, exists := validEnrollmentConfigOutputFormats[s]; exists {
+		return nil
+	}
+
+	msg := "must specify a valid output format. options include: "
+	for k := range validEnrollmentConfigOutputFormats {
+		msg += k + ", "
+	}
```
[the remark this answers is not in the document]
fzdarsky: (more concise and more robust, as it handles the case of single option correctly)

api/v1alpha1/openapi.yaml
```diff
@@ -4999,7 +4999,7 @@ components:
           format: byte
           type: string
         signerName:
-          description: Indicates the requested signer, and is a qualified name.
+          description: Indicates the requested signer, and is a qualified name. Supported values are "enrollment" and "renewal".
```
fzdarsky, line 5002: We don't have a backend implementation for `renewal` signer yet, so I would remove this option until the implementation is added again.

PR #1000, opened 2025-03-17
EDM-1211: agent/applications: refactor 

hexfusion:
This PR refactors the application implementation with the following goals:

- Separate concerns between the application provider and the application itself. This will provide a clean GC flow for potentially large applications defined in spec.
- Move controller logic into the manager and monitor to reduce repetition.

## Summary by CodeRabbit

- **New Features**
  - Enhanced application lifecycle management with streamlined installation, updates, removals, and verification.
  - Improved data copying capabilities for more reliable container data handling.

- **Refactor**
  - Transitioned from container-based to workload-based terminology for clearer status indicators and error feedback.

- **Tests**
  - Updated testing procedures to validate the new lifecycle and workload management improvements, ensuring more robust logging and reliability.

### DIFF
--- a/internal/agent/client/podman.go
+++ b/internal/agent/client/podman.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -22,24 +23,24 @@ const (
 )
 
 type Podman struct {
-	exec    executer.Executer
-	log     *log.PrefixLogger
-	timeout time.Duration
-	reader  fileio.Reader
-	backoff wait.Backoff
+	exec       executer.Executer
+	log        *log.PrefixLogger
+	timeout    time.Duration
+	readWriter fileio.ReadWriter
+	backoff    wait.Backoff
 }
 
 type ImageConfig struct {
 	Labels map[string]string `json:"Labels"`
 }
 
-func NewPodman(log *log.PrefixLogger, exec executer.Executer, reader fileio.Reader, backoff wait.Backoff) *Podman {
+func NewPodman(log *log.PrefixLogger, exec executer.Executer, readWriter fileio.ReadWriter, backoff wait.Backoff) *Podman {
 	return &Podman{
-		log:     log,
-		exec:    exec,
-		timeout: defaultPodmanTimeout,
-		reader:  reader,
-		backoff: backoff,
+		log:        log,
+		exec:       exec,
+		timeout:    defaultPodmanTimeout,
+		readWriter: readWriter,
+		backoff:    backoff,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


api/v1alpha1/openapi.yaml
```diff
@@ -3639,17 +3668,36 @@ components:
...
+          type: array
+          description: A list of application content.
+          items:
+            $ref: "#/components/schemas/ApplicationContent"
+      required:
+        - inline
+    ApplicationContent:
+        allOf:
+          - $ref: '#/components/schemas/FileContent'
+          - $ref: '#/components/schemas/AbsolutePath'
```
celdrake, line 3697: Shouldn't the path be actually `RelativePath`? There's even a validation indicating so.

```
Error 400: Bad Request: spec.applications[].inline[0].path: Invalid value: "/some-path": must be a relative path
```

Also, when generating the types using this branch, `ApplicationContent` is generated with an extra Map<string, "any"> which means I can set any property I want on the object.

This is the generated Typescript type:
```
type ApplicationContent = (FileContent & AbsolutePath & Record<string, any>);         
```

and this is some example data I've been able to save into the DB

```
 "applications": [
      {
        "name": "my-app",
        "appType": "compose",
        "envVars": {},
        "inline": [
          {
            "path": "some-path",
            "content": "just a test",
            "encoding": "plain",
            "whatever": "test",
            "someNumber": 34333
          }
        ],
      }
    ]
```

hexfusion: you are correct i introduced this regression via rebase issue fixed in #988

internal/agent/device/applications/provider.go
```diff
@@ -0,0 +1,254 @@
...
+
+	// set the app name to the image name if not provided
+	appName := lo.FromPtr(spec.Name)
+	if appName == "" {
+		appName = provider.Image
+	}
+	embedded := false
+	path, err := pathFromAppType(v1alpha1.AppTypeCompose, appName, embedded)
+	if err != nil {
+		return nil, fmt.Errorf("getting app path: %w", err)
```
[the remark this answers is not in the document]
hexfusion: the signature for pathFromAppType takes an owned type and spec in this context provides an owned type.

internal/agent/client/podman.go
```diff
@@ -344,6 +349,69 @@ func IsPodmanRootless() bool {
 	return os.Geteuid() != 0
 }
 
+func copyContainerData(ctx context.Context, log *log.PrefixLogger, writer fileio.Writer, podman *Podman, image, destPath string) (err error) {
+	var mountPoint string
+
+	rootless := IsPodmanRootless()
+	if rootless {
+		log.Warnf("Running in rootless mode this is for testing only")
+		mountPoint, err = podman.Unshare(ctx, "podman", "image", "mount", image)
```
DakCrowder, line 358: I see this has been moved here so not net new but have a question - any thoughts on the podman Mount method abstracting out rootless so callers wouldn't have to care?

DakCrowder (approved):

PR #1003, opened 2025-03-19
EDM-786: Agent updates status only on actual changes -- step 1

gshilin-sdb:
Separate between 'status update' and heartbit in agent.


hexfusion:
can we define what a heartbeat is and then also what its purpose is?

gshilin-sdb:
> can we define what a heartbeat is and then also what its purpose is?

For some period of time heartbeat is the same as status update, it's supposed to give us an option to slowly move to a different implementation of it, like storing "created_at" separately in redis, for example

fzdarsky:
Why close #710 and open a new PR, especially without linking that previous PR and without
addressing the concerns against having a heartbeat API raised there?

PR #1006, opened 2025-03-23
EDM-1274: Define event in the API, DB, and CLI

avishayt:
1. Defines Event in OpenAPI, with a few types to get started.
2. This changed caused the code generation to add a prefix to conditions, so I needed to change condition types throughout the code.
3. Add event create and list to the service and store layers, expose list in the transport layer.
4. Add list events to the CLI.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new event retrieval capability via a dedicated API endpoint that supports filtering by type, severity, time range, and more.
  - Extended CLI functionality to recognize and display events as a distinct resource, enhancing operational visibility.
  - Added a new configuration option for event retention period to manage event lifecycle.

- **Refactor**
  - Improved internal consistency by standardizing condition type names, ensuring the reliability of status updates and event processing.
  - Enhanced clarity and maintainability by renaming constants related to condition types across various components.
  - Added new constants and structured definitions for event handling, improving the overall organization of event-related data.
  - Introduced new methods for event creation and listing, streamlining event management within the application.
  - Updated existing methods and tests to utilize the new condition type naming conventions, ensuring consistency throughout the codebase.
  - Implemented a new event cleanup mechanism to manage event retention based on specified periods.


api/v1alpha1/openapi.yaml
```diff
@@ -5044,4 +5147,168 @@ components:
...
+        - kind
+        - apiVersion
+        - id
+        - timestamp
+        - type
+        - source
+        - status
+        - message
+        - severity
+        - correlationId
```
ori-amizur, line 5162: Why is correlationId required? Can't there be events that are not correlated to other events ?

avishayt: In that case there will be a single event with that ID.  When the system emits an event it doesn't know if other correlated events will come after it.

api/v1alpha1/openapi.yaml
```diff
@@ -5098,4 +5188,169 @@ components:
...
+          nullable: true
+          description: User who triggered the event (if applicable).
+        actorService:
+          type: string
+          nullable: true
+          description: Automated service or system that triggered the event (if applicable).
+        status:
+          type: string
+          enum: [success, failure, in-progress]
+          description: Current state of the event, indicating whether the action was completed successfully, failed, or is still ongoing.
```
hexfusion, line 5260: This feels potentially overly complex. An event should be emitted after some observation like a snapshot. in-progress makes me think that an event also tracks the lifecycle of the action. is that what we want?

api/v1alpha1/openapi.yaml
```diff
@@ -3276,8 +3278,96 @@ paths:
...
+        - event
+      description: |
+        Retrieves a list of events.
+      operationId: listEvents
+      parameters:
+        - name: kind
+          in: query
+          description: Filter by resource type.
+          schema:
+            $ref: "#/components/schemas/ResourceKind"
```
asafbss, line 3293: Should we also update /api/v1/labels to support this? It uses the same parameter.

avishayt:
Replaced by https://github.com/flightctl/flightctl/pull/1112

PR #1014, opened 2025-03-25
Add basic AAP AuthZ

rawagner:
A minimalistic AuthZ implementation for AAP. Only two user roles are recognized - admin or viewer.


internal/auth/authz/aap.go
```diff
@@ -0,0 +1,131 @@
...
+	"slices"
+
+	"github.com/flightctl/flightctl/internal/auth/common"
+	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	AAP_PERMISSIONS_CONFIG_FILE = "/etc/permissions/config.yaml"
```
fzdarsky, line 16: Should we keep config files together?
```suggestion
	AAP_PERMISSIONS_CONFIG_FILE = "/etc/flightctl/permissions.yaml"
```

fzdarsky (commented):
According to the PR description, "auditor" is mapped to "operator" and "everyone" to "viewer". Shouldn't that be the other way around?

internal/auth/authz/aap.go
rawagner: +100, updated

rawagner:
> According to the PR description, "auditor" is mapped to "operator" and "everyone" to "viewer". Shouldn't that be the other way around?

My reasoning was, that unless we specifically add (IOW check the admin/auditor checkbox), the users should have the least amount of permissions. 

But you're right that it does not actually align with the target platform here. I've updated the roles to admin, auditor and operator. Let me know if it's better.

rawagner:
closing - the authz has been re-worked since

PR #1030, opened 2025-03-27
EDM-1329: agent/resource: calculate CPU usage with delta-based sampling

hexfusion:
My initial implementation was a bit naïve — it only took a single snapshot of CPU stats, which ended up reflecting the average CPU usage since boot. This gave inflated results, especially when the system was idle.

After digging in, I realized tracking the delta between successive readings gives a much more accurate view of real-time usage. I picked up this approach by looking at how tools like cargo calculate CPU usage in Rust and aligned more with how top calculates usage.

ref. https://github.com/rust-lang/cargo/blob/master/src/cargo/util/cpu.rs

## Summary by CodeRabbit

- **New Features**
  - Enhanced CPU monitoring with refined usage calculations and improved alert synchronization.
  - Introduced new computations to capture both total and active CPU time for better performance insights.
  - Upgraded disk monitoring with additional logging for clearer resource usage trends.
  - Expanded memory monitoring with improved logging for more transparent usage reporting.
  - Introduced a new interface for modular usage collection, separating concerns from the monitoring interface.

- **Bug Fixes**
  - Improved logging for disk and memory usage to enhance observability without altering existing functionality.

- **Tests**
  - Transitioned CPU monitoring tests to a table-driven approach for better modularity and clarity.
  - Enhanced test organization and clarity with new testing structures and logging improvements.

### DIFF
--- a/internal/agent/device/resource/cpu.go
+++ b/internal/agent/device/resource/cpu.go
@@ -23,12 +23,13 @@ const (
 var _ Monitor[CPUUsage] = (*CPUMonitor)(nil)
 
 type CPUMonitor struct {
-	mu       sync.Mutex
-	alerts   map[v1alpha1.ResourceAlertSeverityType]*Alert
-	statPath string
+	mu     sync.Mutex
+	alerts map[v1alpha1.ResourceAlertSeverityType]*Alert
 
 	updateIntervalCh chan time.Duration
 	samplingInterval time.Duration
+	collector        Collector[CPUUsage]
+	prevUsage        *CPUUsage
 
 	log *log.PrefixLogger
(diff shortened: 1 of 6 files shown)


celdrake:
@hexfusion Tested, working as expected. I also tested that the resource status is updated not only based on the CPU load, but also by a modification in the fleet/device spec alert thresholds.

IMO it would be great if the TODOs regarding the message for the breached alert threshold ("CPU utilization reached critical/warning level") were implemented. 

In that way, if the samplingInterval is high and users modify the alert thresholds,  it's easier to understand that the degraded status hasn't had time to adjust yet.

PR #1034, opened 2025-03-27
Edm 1179

noga-magen:
## Summary by CodeRabbit

- **New Features**
  - Improved device enrollment: The registration flow now supports both adding new devices and updating existing ones, offering a smoother, more flexible experience.
  - Enhanced error notifications: Users will receive clear alerts when attempting to register a device that’s already in the system.
  - Streamlined enrollment handling: Adjustments have been made to expedite the approval process for enrollment requests.

### DIFF
--- a/internal/flterrors/flterrors.go
+++ b/internal/flterrors/flterrors.go
@@ -36,4 +36,7 @@ var (
 	ErrSignature       = errors.New("signature error")
 	ErrSignCert        = errors.New("error signing certificate")
 	ErrEncodeCert      = errors.New("error encoding certificate")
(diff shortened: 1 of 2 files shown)


cmd/flightctl-api/__debug_bin694798359
fzdarsky, line 1: Please remove this accidental commit.

fzdarsky (changes requested):
What would be a scenario where we'd want to allow updating a previously created device through an enrollment request? Even more so by overwriting the device record with only a new status?

PR #1048, opened 2025-03-30
Edm 1179

noga-magen:


internal/service/enrollmentrequest.go
```diff
@@ -207,6 +207,14 @@ func (h *ServiceHandler) ReplaceEnrollmentRequest(ctx context.Context, name stri
 	}
 
 	AddStatusIfNeeded(&er)
+	enrollment_requests, _ := h.GetEnrollmentRequest(ctx, name)
```
asafbss, line 210: I’d play it safe and check whether the device exists—not only during updates, but also when creating.
I’d recommend placing this logic inside the store, rather than in the service.
The goal is to enforce this check at the lowest level, preventing other components from calling the store directly and bypassing the validation.

lkatalin: I agree with this, so the check would go in both `[Create()](https://github.com/flightctl/flightctl/blob/main/internal/store/enrollmentrequest.go#L80)` and `[CreateOrUpdate()](https://github.com/flightctl/flightctl/blob/main/internal/store/enrollmentrequest.go#L88)` in the store.

A lot of the logic in the store was moved to `generic.go`, but since this is specific to ERs I would assume it should be added in `enrollmentrequest.go`.

noga-magen: I agree - applying the changes
Regarding the the implementation in both create and create and CreateOrUpdate  - should I apply the change to the Update function as well?

lkatalin (changes requested):
According to the [bug](https://issues.redhat.com/browse/EDM-1179), the error occurs when a previously-approved ER is already deleted but a corresponding device exists. The bug fix should check for a corresponding [Device](https://github.com/flightctl/flightctl/blob/main/internal/store/device.go) with the same `resource.Metadata.Name`, rather than for an already-approved ER which may not exist anymore.

internal/service/enrollmentrequest.go
lkatalin: That's a good question. I think it should also go in `Update` as ERs should not need to be updated if there is already a corresponding device.

noga-magen: Great!
applying the changes and updating the PR

noga-magen:
> According to the [bug](https://issues.redhat.com/browse/EDM-1179), the error occurs when a previously-approved ER is already deleted but a corresponding device exists. The bug fix should check for a corresponding [Device](https://github.com/flightctl/flightctl/blob/main/internal/store/device.go) with the same `resource.Metadata.Name`, rather than for an already-approved ER which may not exist anymore.

I believe it should do both of those thing - since an enrollment request is also an object

lkatalin:
> I believe it should do both of those things

Yeah, that's true. I believe we do have a check for the existing resource with the same name, though: https://github.com/flightctl/flightctl/blob/main/internal/store/generic.go#L99

noga-magen:
> > I believe it should do both of those things
> 
> Yeah, that's true. I believe we do have a check for the existing resource with the same name, though: https://github.com/flightctl/flightctl/blob/main/internal/store/generic.go#L99

Yes but the function update an approved ER - which is redundant 
Anyway I'm working on the error handling applying an enrollment request for an existing device

lkatalin:
I chatted with @fzdarsky earlier and it will actually be better to have this check in the service layer like you orginally did @noga-magen .

noga-magen:
> I chatted with @fzdarsky earlier and it will actually be better to have this check in the service layer like you orginally did @noga-magen .

Amazing - will do thank you Lily! (checking both the existence  of an approved ER and the existence of  a device)

internal/service/enrollmentrequest.go
```diff
@@ -322,3 +327,26 @@ func (h *ServiceHandler) ReplaceEnrollmentRequestStatus(ctx context.Context, nam
...
+	enrollmentRequest, err := s.store.EnrollmentRequest().Get(ctx, orgId, name)
+	if err == nil && enrollmentRequest != nil {
+		if enrollmentRequest.Status != nil &&
+			enrollmentRequest.Status.Approval != nil &&
+			enrollmentRequest.Status.Approval.Approved {
+			return api.StatusBadRequest("enrollment request already approved")
+		}
+	} else if err != nil && !errors.Is(err, flterrors.ErrResourceNotFound) {
+		return api.StatusInternalServerError(fmt.Sprintf("failed to check enrollment request: %v", err))
+	}
```
lkatalin, line 342: It will be easier to remove this from this PR. The bug fix will not depend on it, and it should be part of a larger set of changes if we add this in and requires a bit more workshopping.

lkatalin (changes requested):
@noga-magen Mostly looks good with some small changes. 

As well, if you could change the commit message and PR title to be of the form: `EDM-1179: Bug fix to check for device during ER` or something like this, we like to have descriptive names for these before merge.

internal/service/enrollmentrequest.go
noga-magen: This part will be removed but in raises a logical problem - when trying to apply the same enrollment request you end up updating an approved enrollment request. 
As discussed is the slack I'm opening a new PR (:

PR #1050, opened 2025-03-30
EDM-1305: Console enhancements

ori-amizur:
- Separate stderr and stdout.  Only when working in non terminal mode.
- Support exit code transfer.  The exit code of the remote is
  transferred to local.
- Support command line and arguments. This is done by adding "--" to the
  console command line and then the command and arguments to run.
- Support non-terminal mode. It is detected automatically if both stdin
  and stdout belong to a terminal.
- Support multiple console sessions.

The change is based on remotecommand in https://github.com/kubernetes/client-go. The CLI
is changed to use that library.  The gRPC based code in the agent was
changed to match the library protocol.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new constant for console session metadata.
  - Added support for multiple console sessions and improved session management.
  - Enhanced command execution in the CLI with flexible terminal options and improved argument handling.
  - Implemented keepalive parameters for better gRPC connection management.
  - Added a new method for updating device details with enhanced validation.
  - Added a new function to calculate absolute values for numeric types.
  - Enhanced logging for session handling and connection requests.
  - Introduced a method for modifying device annotations with improved error handling.
- **Bug Fixes**
  - Refined error handling and timeout management in various components to improve reliability.
- **Chores**
  - Upgraded underlying dependencies to boost overall system stability and performance.


go.mod
```diff
@@ -1,11 +1,12 @@
 module github.com/flightctl/flightctl
 
-go 1.22
+go 1.22.0
+
+toolchain go1.23.7
```
fzdarsky, line 5: Please don't bump up the Golang version, we're relying on specific versions for the up- and downstream builds.

fzdarsky:
> The change is based on remotecommand in https://github.com/kubernetes/client-go. The CLI is changed to use that library. The gRPC based code in the agent was changed to match the library protocol.

Any chance we can copy parts of remotecommand over rather than pulling in heavy new dependencies? These dependencies have increased the agent's size by +25% and the CLI's size by +28%. (Thanks for bumping the K8s dep version, BTW!)

Another issue is that this change breaks out users again, which is something we really shouldn't be doing anymore now that we're TP.

mangelajo:
Same comments as frank.

1) Is it feasible to copy out any remotecommand functionality that we use to our modules, as long as it's the same license it should be ok. We could put a note at the top of the file/copied content to indicate provenance.

2) Is there a way to provide backwards compatibility let an older agent behave without channels/channel decoding/etc. Just the CLI or UI must be aware.

About backwards compatibility, you should probably decide on a strategy, the most common is
a) Server is updated first, and server supports 0.v.x and 0.v-1.x agents.
b) agents are updated.
c) A server should not be allowed to update (eventually) if older (v-2) agents are alive.

ori-amizur:
> Any chance we can copy parts of remotecommand over rather than pulling in heavy new dependencies ?

Of course we can.  Do we want to?  Isn't this approach true about all packages we use?  Assuming we copy all the code we want to use, we will probably be able to decrease the size substantially.

go.mod
```diff
@@ -135,53 +142,52 @@ require (
...
-	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
+	sigs.k8s.io/randfill v1.0.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 )
 
 replace (
 	dario.cat/mergo => github.com/imdario/mergo v1.0.0
 	github.com/google/gnostic => github.com/google/gnostic v0.5.7-v3refs
-	golang.org/x/crypto => github.com/openshift/golang-crypto v0.33.1-0.20250310193910-9003f682e581
-	golang.org/x/oauth2 => github.com/openshift/golang-oauth2 v0.26.1-0.20250310184649-06a918c6239d
```
fzdarsky, line 160: IIRC, @rawagner recently added these to address a CVE, so please check with him.
In general, I feel we should add comments to these lines motivating why they are in place.

fzdarsky:
> > Any chance we can copy parts of remotecommand over rather than pulling in heavy new dependencies ?
> 
> Of course we can. Do we want to? Isn't this approach true about all packages we use? Assuming we copy all the code we want to use, we will probably be able to decrease the size substantially.

I agree we're too careless when adding deps and the same argument also applies to previously added deps. We should review those older deps and find a good balance.

go.mod
ori-amizur: Changed

ori-amizur:
> I agree we're too careless when adding deps and the same argument also applies to previously added deps. We should review those older deps and in general find a good balance. This one just stood out because it's such a large increase relatively speaking.

As mentioned before (https://github.com/flightctl/flightctl/pull/1050#issuecomment-2768706951), the difference is about 2% when we remove the console code completely from the CLI.

ori-amizur:
@coderabbitai review

ori-amizur:
> As mentioned before ([#1050 (comment)](https://github.com/flightctl/flightctl/pull/1050#issuecomment-2768706951)), the difference is about 2% when we remove the console code completely from the CLI.

Tried again. I probably made a mistake. It removes ~23%.
There is also an option to strip the executable.  This removes ~45%.

internal/agent/device/console/outgoing_streams.go
```diff
@@ -0,0 +1,208 @@
...
+		packets++
+		sent += len(msg)
+		err := o.streamClient.Send(&grpc_v1.StreamRequest{
+			Payload: msg,
+		})
+		if err != nil {
+			o.log.Errorf("failed sending outgoing message: %v", err)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
ori-amizur: gRPC already has builtin retries ( https://grpc.io/docs/guides/retry/#:~:text=Retries%20are%20enabled%20by%20default,been%20processed%20by%20a%20server. ).  Otherwise, I am not sure it is correct to retry streaming operation since sending the message twice might have undesirable effect.

internal/transport/console.go
```diff
@@ -7,24 +7,22 @@ import (
...
 func (h *WebsocketHandler) HandleDeviceConsole(w http.ResponseWriter, r *http.Request) {
 	orgId := store.NullOrgId
 	deviceName := chi.URLParam(r, "name")
 
-	h.log.Debugf("websocket console connection requested for device: %s", deviceName)
-	consoleSession, err := h.consoleSessionManager.StartSession(r.Context(), orgId, deviceName)
+	h.log.Infof("websocket console connection requested for device: %s", deviceName)
+
+	// Extract metadata
+	metadata := r.URL.Query().Get(api.DeviceQueryConsoleSessionMetadata)
```
rawagner, line 24: for the `ws` connection, i'd rather use the ws native subprotocols header, then encoding subprotocol into url query. This header is also supported when opening ws connection from browsers.

ori-amizur: The negotiation is between the agent and the cli.  The cli uses ws but the agent doesn't.  The the protocols are used by the agent to select a specific protocol.  This then is returned by gRPC, and used by the server in the standard way to indicate the selected protocol.

ori-amizur: OK

rawagner (approved):

rawagner (approved):

PR #1069, opened 2025-04-06
EDM-996: flightctl-agent rpm should create /etc/flightctl/ directory

gshilin-sdb:
## Summary by CodeRabbit

- **New Features**
	- The agent package now includes a dedicated configuration directory to support enhanced system configuration management.
	- A new directory `/etc/flightctl` is created during the installation process for improved organization.

### DIFF
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -85,6 +85,7 @@ The flightctl-selinux package provides the SELinux policy modules required by th
 
 %install
     mkdir -p %{buildroot}/usr/bin
+    mkdir -p %{buildroot}/etc/flightctl
     cp bin/flightctl %{buildroot}/usr/bin
     mkdir -p %{buildroot}/usr/lib/systemd/system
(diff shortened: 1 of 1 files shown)


packaging/rpm/flightctl.spec
```diff
@@ -152,6 +153,7 @@ fi
 
 %files agent -f licenses.list
     %license LICENSE
+    %dir /etc/flightctl
```
hexfusion, line 156: Today the config-dir is variable and created by the agent; do we need both?

gshilin-sdb: It's default value is
```go
DefaultConfigDir = "/etc/flightctl"
```
and it is mentioned in getting-started.md, managing-devices.md, provisioning-devices.md
So it should be created by default.
WDYT?

hexfusion (approved):
since the rpm creates the dir for the expected data-dir location this seems fine. the only corner cases is if the user declares the data-dir in a non standard location the rpm will not create the file but the agent will.

PR #1070, opened 2025-04-06
EDM-1179: Bug fix to check for device when applying ER

noga-magen:
Checking in the service layer, if there is a device with the same name as the ER, using the store Device.Get() function.


internal/service/enrollmentrequest.go
```diff
@@ -322,3 +327,14 @@ func (h *ServiceHandler) ReplaceEnrollmentRequestStatus(ctx context.Context, nam
...
+
+func (s *ServiceHandler) allowCreationOrUpdate(ctx context.Context, orgId uuid.UUID, name string) bool {
+	device, err := s.store.Device().Get(ctx, orgId, name)
+	if err == nil && device != nil {
+		return false
+	}
+	if err != nil && !errors.Is(err, flterrors.ErrResourceNotFound) {
+		return false
+	}
+	return true
```
hexfusion, line 339: nit: it's a good defensive practice to return the default as false. could we rearange this to achieve that?

internal/service/enrollmentrequest.go
```diff
@@ -322,3 +327,14 @@ func (h *ServiceHandler) ReplaceEnrollmentRequestStatus(ctx context.Context, nam
 	result, err := h.store.EnrollmentRequest().UpdateStatus(ctx, orgId, &er)
 	return result, StoreErrorToApiStatus(err, false, api.EnrollmentRequestKind, &name)
 }
+
+func (s *ServiceHandler) allowCreationOrUpdate(ctx context.Context, orgId uuid.UUID, name string) bool {
+	device, err := s.store.Device().Get(ctx, orgId, name)
+	if err == nil && device != nil {
+		return false
+	}
+	if err != nil && !errors.Is(err, flterrors.ErrResourceNotFound) {
```
hexfusion, line 336: question: if the error isn't flterrors.ErrResourceNotFound would it be useful to know what that error is for debugging?

internal/service/enrollmentrequest.go
```diff
@@ -322,3 +327,14 @@ func (h *ServiceHandler) ReplaceEnrollmentRequestStatus(ctx context.Context, nam
 	result, err := h.store.EnrollmentRequest().UpdateStatus(ctx, orgId, &er)
 	return result, StoreErrorToApiStatus(err, false, api.EnrollmentRequestKind, &name)
 }
+
+func (s *ServiceHandler) allowCreationOrUpdate(ctx context.Context, orgId uuid.UUID, name string) bool {
+	device, err := s.store.Device().Get(ctx, orgId, name)
+	if err == nil && device != nil {
```
hexfusion, line 333: The error handling pattern in golang is to handle the error first then perform additional checks. This might help your flow here as well.

internal/service/enrollmentrequest.go
noga-magen: sure

internal/service/enrollmentrequest.go
noga-magen: yes of course - I have changed the type the function returns to an error type in order to debug the the err it appears.

lkatalin (commented):
This is looking much better to me. Could you please add the check also to `CreateEnrollmentRequest`?

internal/service/enrollmentrequest.go
```diff
@@ -121,6 +122,12 @@ func (h *ServiceHandler) CreateEnrollmentRequest(ctx context.Context, er api.Enr
 	if errs := er.Validate(); len(errs) > 0 {
 		return nil, api.StatusBadRequest(errors.Join(errs...).Error())
 	}
+
+	err := h.allowCreationOrUpdate(ctx, orgId, *er.Metadata.Name)
+	if err != nil {
+		return nil, api.StatusBadRequest(err.Error())
```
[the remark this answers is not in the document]
noga-magen: regarding the comments?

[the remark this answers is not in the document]
noga-magen: By the way when trying to add the new function to the create enrollment request function encountered a problem  with a test that seems problematic - I've asked Lily to look at it, could you look at it as well? You could have some extra input on the test ((:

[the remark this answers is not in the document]
noga-magen: > Regarding
> 
> > why give out this error if, when allowCreationOrUpdate fails, it already gives out an error too?


since this error is raised by the function - if a condition happen then the error is raised by it does not come up from the Get() function

internal/service/enrollmentrequest.go
```diff
@@ -322,3 +333,14 @@ func (h *ServiceHandler) ReplaceEnrollmentRequestStatus(ctx context.Context, nam
 	result, err := h.store.EnrollmentRequest().UpdateStatus(ctx, orgId, &er)
 	return result, StoreErrorToApiStatus(err, false, api.EnrollmentRequestKind, &name)
 }
+
+func (h *ServiceHandler) allowCreationOrUpdate(ctx context.Context, orgId uuid.UUID, name string) error {
+	device, err := h.store.Device().Get(ctx, orgId, name)
+	if errors.Is(err, flterrors.ErrResourceNotFound) {
+		return nil // Device not found: allow create or update
+	}
```
avishayt, line 341:
```suggestion
	if err != nil {
		// If it's an expected error (device not found), allow creation or update
		if errors.Is(err, flterrors.ErrResourceNotFound) {
			return nil
		}
		// For any other errors, return them so they can be handled upstream
		return fmt.Errorf("failed to get device: %w", err)
	}
```

lkatalin:
I'm not completely sure, but I think the tests may be failing because the tests themselves are not cleaning up existing devices between updating enrollment requests. Maybe take a look around [here](https://github.com/flightctl/flightctl/blob/main/test/integration/agent/agent_test.go#L158) @noga-magen ?

lkatalin:
Awesome job getting the tests to pass @noga-magen !
Why is it important to check if the device has been seen?

Also could you incorporate @avishayt 's suggested changes from April 21?

This looks pretty close to done IMO and hopefully we can merge it shortly.

internal/service/enrollmentrequest.go
noga-magen: I applied to the new function ((:

noga-magen:
> Awesome job getting the tests to pass @noga-magen ! Why is it important to check if the device has been seen?
> 
> Also could you incorporate @avishayt 's suggested changes from April 21?
> 
> This looks pretty close to done IMO and hopefully we can merge it shortly.



Thank you! ((:
And about the question- 
The old logic failed because the device existed in the database from the initial enrollment attempt but had never been "seen" (i.e., the device had not completed at least one successful interaction with the API) since the agent was stopped. When the agent was restarted, the device could not complete enrollment because its record already existed in the database, which caused the re-enrollment to fail and resulted in the test failure.

avishayt (approved):

PR #1071, opened 2025-04-06
EDM-724: Agent failed create enrollmentrequest failed: 500 Internal S…

gshilin-sdb:
…erver Error while running Device Simulator

### DIFF
--- a/cmd/devicesimulator/main.go
+++ b/cmd/devicesimulator/main.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"math/rand"
 	"os"
 	"os/signal"
@@ -94,7 +93,7 @@ func main() {
 
 	formattedLables := formatLabels(labels)
 
-	agentConfigTemplate := createAgentConfigTemplate(*dataDir, *configFile)
+	agentConfigTemplate := createAgentConfigTemplate(log, *dataDir, *configFile)
 
 	log.Infoln("starting device simulator")
 	defer log.Infoln("device simulator stopped")
@@ -125,8 +124,7 @@ func main() {
 	for i := 0; i < *numDevices; i++ {
 		// stagger the start of each agent
 		time.Sleep(time.Duration(rand.Float64() * float64(agentConfigTemplate.StatusUpdateInterval))) //nolint:gosec
-		agent := agents[i]
-		go startAgent(ctx, agent, log, i)
+		go startAgent(ctx, agents[i], log, i)
 		go approveAgent(ctx, log, serviceClient, agentsFolders[i], formattedLables)
 	}
 	if stopAfter != nil && *stopAfter > 0 {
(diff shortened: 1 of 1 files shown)


cmd/devicesimulator/main.go
```diff
@@ -167,23 +165,35 @@ func reportVersion(versionFormat *string) error {
...
+	err := wait.PollImmediateWithContext(ctx, 2*time.Second, 5*time.Minute, func(ctx context.Context) (bool, error) {
+		// timeout after 30s and retry
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+		defer cancel()
+
+		prefix := agent.GetLogPrefix()
+		if err := agent.Run(ctx); err != nil {
+			// agent timeout waiting for enrollment approval
+			if errors.Is(err, wait.ErrWaitTimeout) {
+				log.Errorf("%s: agent %d timed out: %v", prefix, agentInstance, err)
```
hexfusion, line 177: Internally how long does it take for the agent to timeout and return this error? Will it ever be hit here now?

cmd/devicesimulator/main.go
```diff
@@ -167,23 +165,35 @@ func reportVersion(versionFormat *string) error {
...
+				return false, nil
+			} else if ctx.Err() != nil {
+				// normal teardown
+				log.Infof("%s: agent %d stopped due to context cancellation.", prefix, agentInstance)
+				return false, nil
+			} else {
+				log.Fatalf("%s: %v", prefix, err)
```
hexfusion, line 184: If the agent hits a 500 error will we retry? Do we know why we received a 500 error?

cmd/devicesimulator/main.go
```diff
@@ -167,23 +165,35 @@ func reportVersion(versionFormat *string) error {
...
-		// agent timeout waiting for enrollment approval
-		if errors.Is(err, wait.ErrWaitTimeout) {
-			log.Errorf("%s: agent timed out: %v", prefix, err)
-		} else if ctx.Err() != nil {
-			// normal teardown
-			log.Infof("%s: agent stopped due to context cancellation.", prefix)
-		} else {
-			log.Fatalf("%s: %v", prefix, err)
+	err := wait.PollImmediateWithContext(ctx, 2*time.Second, 5*time.Minute, func(ctx context.Context) (bool, error) {
+		// timeout after 30s and retry
```
hexfusion, line 169: why?

hexfusion (changes requested):

cmd/devicesimulator/main.go
gshilin-sdb: The above code tries every 2 seconds for 5 minutes. This 30 sec timeout is for each attempt, so an attempt will be cancelled if it's got stuck for any reason.
BTW this and code is copy/paste from `approveAgent()`  func, i.e. all numbers are from there.

gshilin-sdb: Moved all those numbers to constants.

cmd/devicesimulator/main.go
gshilin-sdb: You mentioned in Jira card that this error is `just a warning`, so we need just to retry it. Changed this func to retry in any case. In the worst case it will fail after `wait.PollImmediateWithContext` with a `timed out waiting for the condition` error.

cmd/devicesimulator/main.go
```diff
@@ -214,7 +224,7 @@ func createAgents(log *logrus.Logger, numDevices int, initialDeviceIndex int, ag
 		certDir := filepath.Join(agentConfigTemplate.ConfigDir, "certs")
 		agentDir := filepath.Join(agentConfigTemplate.DataDir, agentName)
 		// Cleanup if exists and initialize the agent's expected
-		os.RemoveAll(agentDir)
+		_ = os.RemoveAll(agentDir)
```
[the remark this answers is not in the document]
gshilin-sdb: added

PR #1082, opened 2025-04-10
EDM-1295: Configuration for cli artifacts in UI standalone deployment

celdrake:
Adds the necessary configuration for the UI to be able to talk to the container for the CLI artfacts.

## Summary by CodeRabbit

- **New Features**
  - **CLI Artifacts Server**: Introduced a dynamic configuration option that allows administrators to enable or disable CLI artifacts integration through a new deployment setting.
  - **Enhanced URL Handling**: Improved how deployment endpoints are determined by adapting to environment and protocol settings for more flexible and reliable service configuration.
  - **New Configuration Parameter**: Added a new parameter `cliArtifacts.enabled` to control the activation of CLI artifacts in the UI.

### DIFF
--- a/deploy/helm/flightctl/charts/ui/values.yaml
+++ b/deploy/helm/flightctl/charts/ui/values.yaml
@@ -41,4 +41,6 @@ api:
   url: https://flightctl-api:3443/
   insecureSkipTlsVerify: false
   caCert: ""
-
+## @param cliArtifacts.enabled Set to true if Cli artifacts are enabled in the UI
+cliArtifacts:
+  enabled: true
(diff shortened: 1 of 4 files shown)


deploy/helm/flightctl/charts/ui/values.yaml
```diff
@@ -41,4 +41,7 @@ api:
   url: https://flightctl-api:3443/
   insecureSkipTlsVerify: false
   caCert: ""
-
+## @param cliArtifacts.enabled Set to true if Cli artifacts are enabled in the UI
+# TODO should it be part of global??
```
celdrake, line 45: I believe it would be better to move whether the cliArtifacts are enabled to the global scope.
See https://github.com/flightctl/flightctl/commit/03561147a21f02f3c65d1ab06ea94f65b7553f43

celdrake: Discussed with @rawagner , if we do this, it should apply to all other components and in a separate PR.
Keeping this PR with the current approach.

rawagner (approved):

PR #1083, opened 2025-04-10
Services rpm improvements

DakCrowder:
Currently, we use a [.slice](https://www.freedesktop.org/software/systemd/man/latest/systemd.slice.html) file for grouping our services and providing a parent to start/stop the services.  However, slices are more focused on resource management while [.target](https://www.freedesktop.org/software/systemd/man/latest/systemd.target.html) files are described as a way to " group units and to set synchronization points for ordering dependencies with other unit files."

This PR:
- Modifies the quadlet setup to use a .target file rather than a .slice.  A small side effect is we now mount this .target file to a different location than the existing quadlet files.
- Refactors how some of the service dependencies are defined.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new consolidated system target for flight control services that streamlines dependency management and ordering.
  - Updated service configurations to improve startup sequencing and integration among related services.

- **Chores**
  - Enhanced deployment scripts by updating unit handling, eliminating obsolete configurations, and triggering systemd reloads.
  - Refined installation specifications for seamless packaging and unit file management.
  - Added a new environment variable for systemd unit file output directory.


deploy/podman/flightctl.target
```diff
@@ -1,18 +1,26 @@
...
 
+Requires=flightctl-network.service
+Requires=flightctl-db.service
+Requires=flightctl-kv.service
+Requires=flightctl-api-init.service
+Requires=flightctl-api.service
+Requires=flightctl-worker.service
+Requires=flightctl-periodic.service
+Requires=flightctl-ui-init.service
+Requires=flightctl-ui.service
```
hexfusion, line 12: looks like this ordering is important would a note be helpful to our future selves

hexfusion (approved):

deploy/podman/flightctl.target
DakCrowder: The ordering actually isn't important for the inter-service startup sequence - but to your point this is confusing so added a comment to indicate that.

hexfusion (approved):

PR #1084, opened 2025-04-10
EDM-1412: Add HTTP (openapi) endpoint to support file download from device

ori-amizur:
This is based on the console API.


fzdarsky (changes requested):
The improvements to the console make sense to me, but adding a new API just for downloading doesn't, when users could use the console to download using compression:

```
flighctl console device/xxx -- 'tar -c --zstd <src_files>' | tar -x --zstd
```

or even upload:

```
tar -c --zstd <src_files> | flightctl console device/xxx -- 'cd target_dir && tar -x --zstd'
```

If we wanted a new API endpoint, we'd need to add input validation and RBAC at least.

ori-amizur:
> The improvements to the console make sense to me, but adding a new API just for downloading doesn't, when users could use the console to download using compression:

Usually many operations are supported both by UI and by CLI.
If we don't allow it then it means that downloading SOS report will not be supported in the UI, since the above example is only relevant for CLI users.

> If we wanted a new API endpoint, we'd need to add input validation and RBAC at least.

These can be added if we decide we want to support download by UI.

fzdarsky:
> > If we wanted a new API endpoint, we'd need to add input validation and RBAC at least.
> 
> These can be added if we decide we want to support download by UI. BTW, operations with similar outcomes are allowed by CLI and don't require RBAC and validation. Why are there different approaches for CLI and UI users?

If you know of *any* other API operation that doesn't have RBAC, please LMK. We need to fix that asap.

ori-amizur:
> If you know of _any_ other API operation that doesn't have RBAC, please LMK. We need to fix that asap.

In general, console operations are not restricted by RBAC.

internal/transport/remotecommand/session.go
```diff
@@ -0,0 +1,200 @@
...
+	switch payload[0] {
+	case api.StdoutStreamID:
+		return s.emitStdoutStream(payload[1:])
+	case api.StderrStreamID:
+		return s.emitStderrStream(payload[1:])
+	case api.ErrStreamID:
+		return s.emitErrorStream(payload[1:])
+	default:
+		s.log.Errorf("unexpected stream id %d", int(payload[0]))
+		return fmt.Errorf("unexpected stream id %d", int(payload[0]))
```
hexfusion, line 154:
```suggestion
		return fmt.Errorf("unexpected stream id: %d", int(payload[0]))
```

internal/transport/remotecommand/session.go
```diff
@@ -0,0 +1,200 @@
...
+func (s *session) emitClose() {
+	if s.callbacks != nil {
+		if err := s.callbacks.OnClose(); err != nil {
+			s.log.Errorf("emitClose: %v", err)
+		}
+	}
+}
+
+func (s *session) handlePayload(payload []byte) error {
+	switch payload[0] {
```
hexfusion, line 145: can payload be empty or just have a single element?

internal/transport/remotecommand/session.go
ori-amizur: OK

internal/transport/remotecommand/session.go
ori-amizur: I will add verification

rawagner (approved):
looks good to me. I've also tried the endpoint from UI and it worked well.

PR #1085, opened 2025-04-11
EDM-1413: Set image tag for installed quadlets based off of rpm version

DakCrowder:
Updates the quadlets image tags such that:

- Tags used in the quadlets bundled in the RPM are set from the RPM version
- Tags used for local dev are :latest

One note:  Because we don't build and push containers as part of the pull request build the constructed RPM tags for the PR triggered COPR builds won't reference current tags.  The build will run, but the installable RPM won't work and code will need to be merged into main before an RPM referencing it is usable.  This could be solved by pushing images per PR or possibly falling back to :latest or some other value but my priority right now is removing the hardcoded tags for the 0.6.0 release.

## Summary by CodeRabbit

- **Chores**
  - Updated container images for the API, periodic operations, UI, and worker components to always deploy the latest releases.
  - Enhanced the installation process with dynamic image tag support.
  - Improved the packaging system with refined image versioning and updated changelog entries.

### DIFF
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -128,7 +128,12 @@ The flightctl-services package provides installation and setup of files for runn
     done
 
     # flightctl-services sub-package steps
-    # Run the install script to move the quadlet files
+    # Run the install script to move the quadlet files.
+    #
+    # The IMAGE_TAG is derived from the RPM version, which may include tildes (~)
+    # for proper version sorting (e.g., 0.5.1~rc1-1). However, the tagged images
+    # always use hyphens (-) instead of tildes (~). To ensure valid image tags we need
+    # to transform the version string by replacing tildes with hyphens.
     CONFIG_READONLY_DIR="%{buildroot}%{_datadir}/flightctl" \
     CONFIG_WRITEABLE_DIR="%{buildroot}%{_sysconfdir}/flightctl" \
     QUADLET_FILES_OUTPUT_DIR="%{buildroot}%{_datadir}/containers/systemd" \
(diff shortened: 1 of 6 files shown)


packaging/rpm/flightctl.spec
```diff
@@ -133,6 +133,7 @@ The flightctl-services package provides installation and setup of files for runn
     CONFIG_WRITEABLE_DIR="%{buildroot}%{_sysconfdir}/flightctl" \
     QUADLET_FILES_OUTPUT_DIR="%{buildroot}%{_datadir}/containers/systemd" \
     SYSTEMD_UNIT_OUTPUT_DIR="%{buildroot}/usr/lib/systemd/system" \
+    IMAGE_TAG=$(echo %{version} | tr '~' '-') \
```
hexfusion, line 136: remind me why again

DakCrowder: The RPM versions that get injected into the specs are supposed to have tildes for pre-release versions so they get sorted properly.

[Fedora docs](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_handling_non_sorting_versions_with_tilde_dot_and_caret) about the tilde and in our pipelines for the [more recent rc candidates](https://copr.fedorainfracloud.org/coprs/g/redhat-et/flightctl/package/flightctl/) you can see a tilde.  The [packit](https://github.com/flightctl/flightctl/blob/main/.packit.yaml#L15) config sets the RPM version w/ the tilde from the git tags

hexfusion: thanks add those notes as a concise comment then lgtm

DakCrowder: Also scrounged this up for more context: https://redhat-internal.slack.com/archives/C01KER8QV33/p1737460977290809?thread_ts=1737455289.009299&cid=C01KER8QV33

hexfusion: thanks i forget everytime haha

hexfusion (approved):

PR #1091, opened 2025-04-15
Create aap oauth app

DakCrowder:
Allows users to pass an oauth token to the quadlet scripting and we will create a Fight control oauth application for them.  An OAuth app is created when aap auth is enabled and a token is passed in the config and a client ID does not exist in the config.

This is quite a happy path implementation and does not gracefully handle many errors.  The created app requires a primary key identifier for an "Organization", rather than build out any additional logic for now the value has been hardcoded to the default aap managed organization.

## Summary by CodeRabbit

- **New Features**
  - Added automation for creating OAuth applications in Ansible Automation Platform (AAP), including a new script and service configuration support.
- **Improvements**
  - Enhanced container setup with writable service configuration and dedicated network assignment.
  - Improved handling of extracted configuration values by trimming whitespace.
- **Configuration**
  - Updated service configuration to support OAuth tokens for AAP authentication.
- **Packaging**
  - Included the new OAuth application creation script in the RPM package.

### DIFF
--- a/deploy/podman/flightctl-api/flightctl-api-config/create_aap_application.sh
+++ b/deploy/podman/flightctl-api/flightctl-api-config/create_aap_application.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+create_oauth_application() {
+    local oauth_token="$1"
+    local base_domain="$2"
+    local aap_url="$3"
+    local insecure_skip_tls="$4"
+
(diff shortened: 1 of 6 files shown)


hexfusion (approved):
can we ensure that any followups todo are captured in jira

DakCrowder:
@hexfusion Yes, I need to see where they should live but 100% there should be some follow up

PR #1093, opened 2025-04-16
EDM-1184: changed SELinux policy to allow systemd to start the service

mparram:
## Summary by CodeRabbit

- **Chores**
	- Updated SELinux policy rules for the flightctl agent to refine role associations and permissions. This enhances process domain management and executable access controls. No changes to user-facing functionality.


packaging/selinux/flightctl_agent.te
```diff
@@ -1,6 +1,7 @@
 policy_module(flightctl_agent, 1.0.0)
```
hexfusion, line 1: should we be bumping the version

hexfusion (approved):

PR #1095, opened 2025-04-16
EDM-1436: api: extend systemInfo to include facts

hexfusion:
this is just for testing at this point


api/v1alpha1/openapi.yaml
```diff
@@ -4025,7 +4025,15 @@ components:
         agentVersion:
           type: string
           description: The Agent version.
+        facts:
```
avishayt, line 4028: I know naming is hard, but "facts" seems to generic to me, and may also imply that other information returned is not facts?

hexfusion: I took the naming from Ansible[1], open to other ideas `details` maybe? But I believe we want the name fairly generic to also accommodate custom user defined fields.

[1] https://www.redhat.com/en/blog/playing-ansible-facts

avishayt: I see, got it.  Yea, "details" or "attributes" seem less loaded.

hexfusion: cc @fzdarsky

avishayt: Between these 3 choices, I like "attributes".

hexfusion: i thought that would not be legal but I will explore

avishayt:
LGTM.  @fzdarsky WDYT?

fzdarsky (approved):

PR #1096, opened 2025-04-17
EDM-1016: Add TPM functionality and unit tests to support attestation

lkatalin:
This PR adds and tests some TPM functionality that can be used in a future PR to do the following agent-side:
- Detect whether a TPM is present
- Get a Quote or Attestation from the TPM
- Generate an Attestation Key to be used for signing the Attestation, as well as potentially for the device ID

Note: According to the [TCG's guide to keys for identity and attestation](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf), it is best practice to separate the Attestation Key from the DeviceID key. However, the `go-tpm-tools` library (which seems to be the best option for TPM interaction in Go) [does not yet seem to support](https://github.com/google/go-tpm-tools/blob/v0.4.5/client/keys.go#L194) generating this DeviceID key as a non-primary key. So we may temporarily need to use the Attestation Key for both purposes.

## Summary by CodeRabbit

- **New Features**
  - Added TPM detection and version validation.
  - Introduced capabilities to create local attestation keys and generate TPM quotes with nonce validation.
  - Enabled attestation data retrieval using TPM keys.

- **Tests**
  - Added unit tests covering TPM key creation, quote generation, attestation retrieval, and PCR value reading.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -99,7 +99,10 @@ require (
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/gnostic-models v0.6.9 // indirect
 	github.com/google/go-configfs-tsm v0.2.2 // indirect
+	github.com/google/go-sev-guest v0.9.3 // indirect
+	github.com/google/go-tdx-guest v0.3.1 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
+	github.com/google/logger v1.1.1 // indirect
 	github.com/google/pprof v0.0.0-20240727154555-813a5fbdbec8 // indirect
 	github.com/gorilla/mux v1.8.1 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
@@ -131,6 +134,7 @@ require (
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
+	github.com/pborman/uuid v1.2.1 // indirect
 	github.com/perimeterx/marshmallow v1.1.5 // indirect
 	github.com/pjbgf/sha1cd v0.3.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
@@ -142,6 +146,7 @@ require (
 	github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3 // indirect
 	github.com/skeema/knownhosts v1.3.0 // indirect
 	github.com/xanzy/ssh-agent v0.3.3 // indirect
+	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/mod v0.20.0 // indirect
 	golang.org/x/net v0.33.0 // indirect
 	golang.org/x/oauth2 v0.21.0 // indirect
--- a/internal/tpm/tpm.go
+++ b/internal/tpm/tpm.go
@@ -1,18 +1,52 @@
 package tpm
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 	"io"
+	"os"
 
+	"github.com/google/go-tpm-tools/client"
+	pbattest "github.com/google/go-tpm-tools/proto/attest"
+	pbtpm "github.com/google/go-tpm-tools/proto/tpm"
 	"github.com/google/go-tpm-tools/simulator"
 	"github.com/google/go-tpm/legacy/tpm2"
 	"github.com/google/go-tpm/tpmutil"
 )
 
+const (
+	MinNonceLength     = 8
+	TpmSystemPath      = "/dev/tpm/tpm0"
+	TpmVersionInfoPath = "/sys/class/tpm/tpm0/tpm_version_major"
+)
+
 type TPM struct {
 	devicePath string
 	channel    io.ReadWriteCloser
+}
+
+// Note: this may be a hardware TPM or a software or emulated TPM available to the system
+func TpmExists() bool {
+	if _, err := os.Stat(TpmSystemPath); err == nil {
+		return true
+	}
+	return false
+}
+
+func ValidateTpmVersion2() error {
+	if !TpmExists() {
+		return fmt.Errorf("no TPM detected at %s", TpmSystemPath)
+	}
+	versionBytes, err := os.ReadFile(TpmVersionInfoPath)
+	if err != nil {
+		return err
+	}
+	versionStr := string(bytes.TrimSpace(versionBytes))
+	if versionStr != "2" {
+		return fmt.Errorf("TPM is not version 2.0")
+	}
+	return nil
 }
 
 func OpenTPM(devicePath string) (*TPM, error) {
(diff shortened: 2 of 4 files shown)


lkatalin:
The unit test for `GetAttestation` is failing because it requires running as root in order to get the TCG event log (this was true locally). What is the right way to test this?

```
=== Failed
=== FAIL: internal/tpm TestGetAttestation (0.01s)
    tpm_test.go:103: failed to get attestation: failed to retrieve TCG Event Log: open /sys/kernel/security/tpm0/binary_bios_measurements: no such file or directory
```

internal/tpm/tpm.go
```diff
@@ -51,3 +92,42 @@ func (t *TPM) GetPCRValues(measurements map[string]string) error {
...
+func (t *TPM) GetAttestation(nonce []byte, ak *client.Key) (*pbattest.Attestation, error) {
+	// TODO - may want to use CertChainFetcher in the AttestOpts in the future
+	// see https://pkg.go.dev/github.com/google/go-tpm-tools/client#AttestOpts
+
+	att, err := ak.Attest(client.AttestOpts{Nonce: nonce})
+	if err != nil {
+		return nil, err
+	}
+	return att, nil
+}
```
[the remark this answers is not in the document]
lkatalin: This would be a good idea if we used Github issues to keep track of todos.

internal/tpm/tpm.go
```diff
@@ -51,3 +92,42 @@ func (t *TPM) GetPCRValues(measurements map[string]string) error {
...
+
+func (t *TPM) GetQuote(nonce []byte, ak *client.Key, pcr_selection *tpm2.PCRSelection) (*pbtpm.Quote, error) {
+	if len(nonce) < MinNonceLength {
+		return nil, fmt.Errorf("nonce does not meet minimum length of %d bytes", MinNonceLength)
+	}
+
+	quote, err := ak.Quote(*pcr_selection, nonce)
```
hexfusion, line 128: if your not going to add additional context here just return ak.Quote...

internal/tpm/tpm_test.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func TestGetQuote(t *testing.T) {
+	tpm, key, nonce, selection := setupTestData(t)
+	defer tpm.Close()
+	defer key.Close()
+
+	q, err := tpm.GetQuote(nonce, key, selection)
+	if err != nil {
+		t.Errorf("failed to get quote: %v", err)
```
hexfusion, line 90: for unit tests can we use testify

```go
require := require.New(t)
[...]
q, err := tpm.GetQuote(nonce, key, selection)
require.NoError(err)
```

internal/tpm/tpm.go
```diff
@@ -1,20 +1,61 @@
...
+	}
+	if out.String() != "2" {
+		return fmt.Errorf("TPM is not version 2.0")
+	}
+	return nil
+}
+
+func GetTpmVendorInfo() (string, error) {
+	if !TpmExists() {
+		return "", fmt.Errorf("No TPM detected at /dev/tpm/tpm0")
```
hexfusion, line 53: nit: /dev/tpm/tpm0 should be a const

internal/tpm/tpm.go
```diff
@@ -51,3 +92,42 @@ func (t *TPM) GetPCRValues(measurements map[string]string) error {
...
+// Restricted: yes
+// Sign: yes
+// Decrypt: no
+// FixedTPM: yes (cannot migrate or be duplicated)
+// SensitiveDataOrigin: yes (was created in the TPM)
+func (t *TPM) CreateLAK() (*client.Key, error) {
+	// AttestationKeyECC generates and loads a key from AKTemplateECC in the Owner (aka 'Storage') hierarchy.
+	lak, err := client.AttestationKeyECC(t.channel)
+	if err != nil {
+		return nil, err
```
hexfusion, line 107: might be worth adding more context on the error here, otherwise just return

internal/tpm/tpm.go
```diff
@@ -1,20 +1,60 @@
...
+	pbattest "github.com/google/go-tpm-tools/proto/attest"
+	pbtpm "github.com/google/go-tpm-tools/proto/tpm"
 	"github.com/google/go-tpm-tools/simulator"
 	"github.com/google/go-tpm/legacy/tpm2"
 	"github.com/google/go-tpm/tpmutil"
 )
 
+const MinNonceLength = 8
+const TpmSystemPath = "/dev/tpm/tpm0"
+const TpmVersionInfoPath = "/sys/class/tpm/tpm0/tpm_version_major"
```
hexfusion, line 20: nit: single const () plz

internal/tpm/tpm.go
```diff
@@ -1,20 +1,60 @@
...
+func TpmExists() bool {
+	if _, err := os.Stat(TpmSystemPath); err == nil {
+		return true
+	}
+	return false
+}
+
+func TpmIsVersion2() error {
+	if !TpmExists() {
+		return fmt.Errorf("No TPM detected at %s", TpmSystemPath)
```
hexfusion, line 37: nit errors should not start with caps vs log

internal/tpm/tpm.go
```diff
@@ -1,20 +1,60 @@
...
+	}
+	versionBytes, err := os.ReadFile(TpmVersionInfoPath)
+	if err != nil {
+		return err
+	}
+	versionStr := string(bytes.TrimSpace(versionBytes))
+	if versionStr != "2" {
+		return fmt.Errorf("TPM is not version 2.0")
```
hexfusion, line 45: same

internal/tpm/tpm.go
```diff
@@ -51,3 +91,34 @@ func (t *TPM) GetPCRValues(measurements map[string]string) error {
...
+
+	return ak.Attest(client.AttestOpts{Nonce: nonce})
+}
+
+func (t *TPM) GetQuote(nonce []byte, ak *client.Key, pcr_selection *tpm2.PCRSelection) (*pbtpm.Quote, error) {
+	if len(nonce) < MinNonceLength {
+		return nil, fmt.Errorf("nonce does not meet minimum length of %d bytes", MinNonceLength)
+	}
+
+	return ak.Quote(*pcr_selection, nonce)
```
hexfusion, line 123: question: Usual go patterns would guarantee these params would not be nil, are we certain of that here.

internal/tpm/tpm_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+type TestFixture struct {
+	tpm *TPM
+}
+
+func setupTestFixture(t *testing.T) *TestFixture {
+	t.Helper()
+
+	tpm, err := OpenTPMSimulator()
+	if err != nil {
```
hexfusion, line 26: if this is an error wont tpm be nil I think you want this setup to return an error and you can handle error with testify after.

internal/tpm/tpm_test.go
```diff
@@ -0,0 +1,123 @@
...
+	f := setupTestFixture(t)
+
+	key, err := f.tpm.CreateLAK()
+	if err != nil {
+		t.Errorf("unable to create local attestation key")
+	}
+
+	nonce := make([]byte, 8)
+	if _, err := io.ReadFull(rand.Reader, nonce); err != nil {
+		t.Errorf("failed to create nonce: %v", err)
```
hexfusion, line 45: same here you need to stop progress when there is an error minimally. either return a proper error or return after the log

internal/tpm/tpm_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	selection := client.FullPcrSel(tpm2.AlgSHA256)
+
+	return f.tpm, key, nonce, &selection
+}
+
+func TestLAK(t *testing.T) {
+	tpm, key, _, _ := setupTestData(t)
+	defer tpm.Close()
+	defer key.Close()
```
hexfusion, line 56: optional:

```suggestion
defer func() {
     tpm.Close()
     key.Close()
}()
```

internal/tpm/tpm.go
lkatalin: This is a good catch, I always forget you have to do this check manually in Go.

internal/tpm/tpm_test.go
lkatalin: Okay, trying that way

internal/tpm/tpm.go
hexfusion: well the pattern should ensure they were nil checked before passed to the method just making sure that is true.

internal/tpm/tpm_test.go
```diff
@@ -0,0 +1,123 @@
...
+	tpm, key, nonce, _ := setupTestData(t)
+	defer tpm.Close()
+	defer key.Close()
+
+	_, err = tpm.GetAttestation(nonce, key)
+	require.NoError(err)
+}
+
+func TestGetPCRValues(t *testing.T) {
+	require := require.New(t)
```
hexfusion, line 114: :+1:

internal/tpm/tpm.go
lkatalin: I wouldn't assume it gets checked, so I added the check.

internal/tpm/tpm.go
lkatalin: this one is just in caps because `TPM` is always in caps

go.mod
```diff
@@ -99,7 +99,10 @@ require (
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/gnostic-models v0.6.9 // indirect
 	github.com/google/go-configfs-tsm v0.2.2 // indirect
+	github.com/google/go-sev-guest v0.9.3 // indirect
+	github.com/google/go-tdx-guest v0.3.1 // indirect
```
hexfusion, line 103: question: its generally felt these dependencies are a requirement?

internal/tpm/tpm.go
```diff
@@ -1,20 +1,62 @@
...
+		return err
+	}
+	versionStr := string(bytes.TrimSpace(versionBytes))
+	if versionStr != "2" {
+		return fmt.Errorf("TPM is not version 2.0")
+	}
+	return nil
+}
+
+func GetTpmVendorInfo() (string, error) {
```
hexfusion, line 52: question: why are we exporting this if it's not implemented?

go.mod
lkatalin: Can we get rid of them? I think `go-tpm-tools` brings them in.

internal/tpm/tpm.go
lkatalin: I've un-exported it. I can re-export it when I add the implementation.

lkatalin: The linter didn't like it, so I just removed it.

go.mod
hexfusion: is go-tpm-tools required?

hexfusion: I guess we can try to trim this back down the road to just go-tpm?

internal/tpm/tpm.go
```diff
@@ -1,20 +1,54 @@
...
 
+// Note: this may be a hardware TPM or a software or emulated TPM available to the system
+func TpmExists() bool {
+	if _, err := os.Stat(TpmSystemPath); err == nil {
+		return true
+	}
+	return false
+}
+
+func TpmIsVersion2() error {
```
hexfusion, line 37: `Is` naming is generally used for a bool return maybe
```suggestion
func ValidateTpmVersion2() error {
```

lkatalin: That makes sense

hexfusion (approved):
These code changes are not plumbed into any functional aspects of the project and are  providing the foundation to future work.

hexfusion:
cc @amir-yogev-gh

PR #1101, opened 2025-04-21
EDM-1205 Load Testing Agent TImeout

SiddarthR56:


internal/agent/device/lifecycle/manager.go
```diff
@@ -348,7 +349,9 @@ func (b *LifecycleManager) enrollmentRequest(ctx context.Context, deviceStatus *
 	}
 
 	err := wait.ExponentialBackoffWithContext(ctx, b.backoff, func(ctx context.Context) (bool, error) {
-		_, err := b.enrollmentClient.CreateEnrollmentRequest(ctx, req)
+		timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
hexfusion, line 352: nit: can we define the timeout as a const. 

If I understand this, you are putting the timeout in the agent vs simulator to use the retry mechanism already defined via ExponentialBackoffWithContext. This solves the greater problem with the agent itself vs just simulator?

hexfusion: If my assessment is true or if not, please in your own words describe why you are making the change in the PR description so its clear to reviewers. The automated rabbit comments should not be a replacement. I know its just a draft.

avishayt: Would it be more robust to add a default timeout to the client instead of using context.WithTimeout in some calls?

PR #1102, opened 2025-04-21
EDM-1435: agent/systemInfo: harden override logic

hexfusion:


internal/agent/device/systeminfo/system_info.go
```diff
@@ -29,8 +29,8 @@ const (
 	ipv4RoutePath    = "/proc/net/route"
 	ipv6RoutePath    = "/proc/net/ipv6_route"
 
-	// ScriptOverrideDir is the directory where custom/overide scripts are stored
-	ScriptOverrideDir = "collect.d"
+	// DetailsDir is the directory where custom/overide scripts are stored in the agent data dir
+	DetailsDir = "details.d"
```
hexfusion, line 33: note: I don't love this but I also don't think its appropriate to include these in $PATH

hexfusion: I think a `plugins` dir would be more intuitive so `plugins/system-info`

lkatalin (approved):

internal/agent/agent.go
```diff
@@ -203,6 +204,9 @@ func (a *Agent) Run(ctx context.Context) error {
 	statusManager.RegisterStatusExporter(specManager)
 	statusManager.RegisterStatusExporter(systemInfoManager)
 
+	// register internal system info collectors
+	// systemInfoManager.RegisterCollector(ctx, "attestation", lifecycleManager.GetAttestation)
```
DakCrowder, line 208: Is this intended / a future TODO?

test/scripts/agent-images/Containerfile-e2e-base.local
```diff
@@ -41,3 +41,4 @@ RUN useradd -ms /bin/bash user && \
 ## Add your flightctl configuration and certificates
 ADD bin/agent/etc/flightctl/config.yaml /etc/flightctl/
 ADD bin/agent/etc/flightctl/certs/* /etc/flightctl/certs/
+RUN  mkdir -p /etc/containers/registries.conf.d/ && echo -e '[[registry]]\nlocation = "192.168.1.42:5000"\ninsecure = true' > /etc/containers/registries.conf.d/flightctl-e2e.conf
```
DakCrowder, line 44: Probably a miscommit here?

internal/agent/agent.go
hexfusion: example but meant to remove good catch

DakCrowder (approved):

PR #1104, opened 2025-04-22
EDM -1450: Unable to Approve/Deny CSR Requests

SiddarthR56:
Approval updates to certificate signing requests are incorrectly being submitted through the full replacement endpoint instead of the dedicated approval endpoint, bypassing validation and approval policy checks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the handling of certificate signing request approvals to ensure correct processing.


internal/transport/certificatesigningrequest.go
```diff
@@ -75,6 +75,6 @@ func (h *TransportHandler) UpdateCertificateSigningRequestApproval(w http.Respon
 		return
 	}
 
-	body, status := h.serviceHandler.ReplaceCertificateSigningRequest(r.Context(), name, csr)
+	body, status := h.serviceHandler.UpdateCertificateSigningRequestApproval(r.Context(), name, csr)
```
hexfusion, line 78: why is this wrong :) I know its just a draft.

SiddarthR56: Approval updates to CSRs are being sent via the full replacement endpoint instead of the proper approval endpoint.

hexfusion:
cc @lkatalin

lkatalin:
Yeah, this relates to a question I had a few weeks ago - basically, why do we have both of these methods? Will we ever want to "update" a CSR beyond changing its approval status? (IMO, we won't - a CSR is a one-time event and shouldn't be changed beyond approval/denial.) And if we don't need to update it, can we restrict the available functionality to only make available updating the approval/denial of the CSR?

avishayt (approved):

PR #1110, opened 2025-04-23
NO-ISSUE: Wrong code is generated for python

gshilin-sdb:
Original code generated a wrong python client types for `ResourceMonitor`.

[OpenAPI Specification v3.1.0](https://spec.openapis.org/oas/v3.1.0.html)
Discriminator lives on a composite schema
The discriminator object is legal only when using one of the composite keywords oneOf, anyOf, allOf.
 …
 When using the discriminator, inline schemas will not be considered.
 — Discriminator Object (OAS 3.1.0 § 4.8.25)
Inline schemas (i.e. schemas without an explicit component name) cannot participate
While composition offers model extensibility, it does not imply a hierarchy between the models.
 To support polymorphism, the OpenAPI Specification adds the discriminator field. … If a new value exists, this takes precedence over the schema name. As such, inline schema definitions, which do not have a given id, cannot be used in polymorphism.
 — Composition and Inheritance (Polymorphism) (OAS 3.1.0 § 4.8.24.2.1)
Wrapping schema (the one that uses oneOf) is the only place you can put the discriminator.
All of the variants you want to discriminate must be named component schemas (so they have an identifier).
Inline‐defined schemas (e.g. anonymous allOf fragments) will simply be ignored by the discriminator mechanism.
ResourceMonitorSpec isn’t itself a polymorphic container (it has no oneOf/anyOf to list its variants), so there’s nothing for the discriminator to drive.
Inline schemas are ignored by the discriminator (because they have no component key to map against)—see “When using the discriminator, inline schemas will not be considered.”
Only named components can participate in polymorphism. The discriminator must live on a named component that directly references other named components via $ref .

The following test may be added to flightctl-python-client project:
```python
from flightctl.models.cpu_resource_monitor_spec import CpuResourceMonitorSpec
from flightctl.models.device import Device
from flightctl.models.device_spec import DeviceSpec
from flightctl.models.disk_resource_monitor_spec import DiskResourceMonitorSpec
from flightctl.models.memory_resource_monitor_spec import MemoryResourceMonitorSpec
from flightctl.models.resource_monitor import ResourceMonitor


def test_resource_monitor_from_dict_picks_cpu_subtype():
    # use from_dict to dispatch via the discriminator
    wrapper = ResourceMonitor.from_dict({"monitorType": "CPU"})
    # the wrapper itself is ResourceMonitor…
    assert isinstance(wrapper, ResourceMonitor)
    # …but the actual subtype lives in `actual_instance`:
    actual = wrapper.actual_instance
    assert isinstance(actual, CpuResourceMonitorSpec)
    assert actual.monitor_type == "CPU"


def test_resource_monitor_from_dict_picks_memory_subtype():
    wrapper = ResourceMonitor.from_dict({"monitorType": "Memory"})
    actual = wrapper.actual_instance
    assert isinstance(actual, MemoryResourceMonitorSpec)
    assert actual.monitor_type == "Memory"


def test_resource_monitor_from_dict_picks_disk_subtype():
    wrapper = ResourceMonitor.from_dict({"monitorType": "Disk", "path": "path/to/disk"})
    actual = wrapper.actual_instance
    assert isinstance(actual, DiskResourceMonitorSpec)
    assert actual.monitor_type == "Disk"


def test_device_spec_accepts_any_resource_monitor_subtype():
    cpu_wrapper = ResourceMonitor.from_dict({"monitorType": "CPU"})
    mem_wrapper = ResourceMonitor.from_dict({"monitorType": "Memory"})
    disk_wrapper = ResourceMonitor.from_dict({"monitorType": "Disk", "path": "path/to/disk"})

    spec = DeviceSpec(resources=[cpu_wrapper, mem_wrapper, disk_wrapper])
    # your DeviceSpec.resources stays as a list of ResourceMonitor
    assert spec.resources == [cpu_wrapper, mem_wrapper, disk_wrapper]


def test_device_to_dict_and_from_dict_roundtrip():
    cpu_wrapper = ResourceMonitor.from_dict({"monitorType": "CPU"})
    disk_wrapper = ResourceMonitor.from_dict({"monitorType": "Disk", "path": "path/to/disk"})
    spec = DeviceSpec(resources=[cpu_wrapper, disk_wrapper])
    device = Device(spec=spec)

    data = device.to_dict()
    # make sure the raw JSON has the right shape
    assert data["spec"]["resources"][0]["monitorType"] == "CPU"
    assert data["spec"]["resources"][1]["monitorType"] == "Disk"

    recon = Device.from_dict(data)
    # full object equality
    assert recon == device
```

## Summary by CodeRabbit

- **Refactor**
  - Updated resource monitor specifications to define the monitor type property within each specific monitor, rather than in the shared base schema.
  - Enhanced type safety and clarity by introducing explicit monitor type fields and struct definitions for CPU, Memory, and Disk resource monitors.


gshilin-sdb:
CC @DakCrowder

api/v1alpha1/types.gen.go
```diff
@@ -118,6 +123,11 @@
 	DeviceUpdatedStatusUpdating  DeviceUpdatedStatusType = "Updating"
 )
 
+// Defines values for DiskResourceMonitorSpecMonitorType.
+const (
+	DiskResourceMonitorSpecMonitorTypeCPU DiskResourceMonitorSpecMonitorType = "CPU"
+)
+
```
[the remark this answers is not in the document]
gshilin-sdb: this is a generated code, I cannot change it

DakCrowder:
@gshilin-sdb Before committing to changing the openapi spec I found an option with the python client generator tool that might solve this issue for us, and is at the very least worth exploring.

Per the [client generator docs](https://openapi-generator.tech/docs/generators/python/) there is an useOneOfDiscriminatorLookup option that it looks like we can set to `true` and results in "Validation (e.g. one and only one match in oneOf's schemas) will be skipped." which seems to be aligned with what we want.

hexfusion:
> @gshilin-sdb Before committing to changing the openapi spec I found an option with the python client generator tool that might solve this issue for us, and is at the very least worth exploring.
> 
> Per the [client generator docs](https://openapi-generator.tech/docs/generators/python/) there is an useOneOfDiscriminatorLookup option that it looks like we can set to `true` and results in "Validation (e.g. one and only one match in oneOf's schemas) will be skipped." which seems to be aligned with what we want.

this makes sense to explore

api/v1alpha1/openapi.yaml
```diff
@@ -3747,18 +3744,38 @@ components:
...
         - $ref: '#/components/schemas/ResourceMonitorSpec'
     MemoryResourceMonitorSpec:
       allOf:
+        - type: object
+          required: [ monitorType ]
+          properties:
+            monitorType:
+              type: string
+              description: The type of resource to monitor.
+              enum: [ 'CPU' ]
```
hexfusion, line 3767: question: should this be Memory?

api/v1alpha1/openapi.yaml
```diff
@@ -3747,18 +3744,38 @@ components:
...
     DiskResourceMonitorSpec:
       allOf:
         - $ref: '#/components/schemas/ResourceMonitorSpec'
+        - type: object
+          required: [ monitorType ]
+          properties:
+            monitorType:
+              type: string
+              description: The type of resource to monitor.
+              enum: [ 'CPU' ]
```
hexfusion, line 3778: question: should this be Disk?

api/v1alpha1/openapi.yaml
gshilin-sdb: fixed

api/v1alpha1/openapi.yaml
gshilin-sdb: fixed

gshilin-sdb:
> results in "Validation (e.g. one and only one match in oneOf's schemas) will be skipped." which seems to be aligned with what we want.

IMHO we want to ensure a contract between code and data. We want only three types of ResourceMonitor to be enabled, and we want to distinguish between those three types.

hexfusion:
> With `useOneOfDiscriminatorLookup=true` got the following code, which generates better code for CPU,Disk,Memory, but in addition it also tries to handle (incorrectly IMHO) the other "**possible???**" values of the monitorTypes discriminator. And it also tries to count them in case it's not one of the above proper/improper monitorTypes.
> 
> So it looks like we need this fix.

Just so I understand, the concern is that the generated code would allow strings `Memory` and `MemoryResourceMonitorSpec`. Then, because of that possibility we should change the API?  Or is there are a greater problem I am missing. Ideally, API changes are the last resort.

gshilin-sdb:
Both original version of openapi.yaml and this PR are generating the same code for `go`. But different for `python`. Currently the python one is `just wrong`. After this PR the generated python will handle ResourceMonitor exactly as expected, enabling just three sub-ResourceMonitor values, parsing them properly etc. etc. And from go point of view nothing will change.

hexfusion:
is it actually wrong or not perfect I think the goal is generate working code automatically from our spec. We do not generally audit the automated code. So the question is will the automated code work as expected? The API that exists that spec is supported and valid.

gshilin-sdb:
The fix is to get a discriminator field monitorType from a ResourceMonitorSpec, which includes other common fields, and to put it directly to each <whatever>ResourceMonitor.

gshilin-sdb:
It is wrong. With a flag it's still wrong.

DakCrowder:
Following up - after some conversations to unblock shipping the collection code the `useOneOfDiscriminatorLookup` option has been [used in the python client](https://github.com/flightctl/flightctl-python-client/blob/main/Makefile#L7) to alleviate the issue

hexfusion:
after reviewing openapi spec it seems that the discriminator should be defined in the type itself.

ref. https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/ 

so I believe that removing the enum from your pr would make this work? can we try updating that alone see if the generated code is as expected? also my assumption is the change is not breaking on the wire.

api/v1alpha1/openapi.yaml
```diff
@@ -3747,18 +3744,38 @@ components:
...
         - samplingInterval
     CpuResourceMonitorSpec:
       allOf:
+        - type: object
+          required: [ monitorType ]
+          properties:
+            monitorType:
+              type: string
+              description: The type of resource to monitor.
+              enum: [ 'CPU' ]
```
hexfusion, line 3757: with the mapping below i believe the enum can be removed can we verify

api/v1alpha1/openapi.yaml
```diff
@@ -3734,9 +3734,6 @@ components:
       type: object
       description: Specification for monitoring a resource.
       properties:
-        monitorType:
-          type: string
-          description: The type of resource to monitor.
```
hexfusion, line 3739: this removal makes sense per the spec

hexfusion:
in general this seems fine now but I want to ensure the UI is ready and this makes sense for them as well. cc @celdrake

celdrake:
> in general this seems fine now but I want to ensure the UI is ready and this makes sense for them as well. cc @celdrake

Hi, I just updated the PR for the UI so it's aligned with the new changes.

celdrake (approved):

PR #1113, opened 2025-04-28
EDM-1447: The console cannot connect to device if the the device is busy in sync loop

ori-amizur:
## Summary by CodeRabbit

- **New Features**
  - Introduced a subscription-based mechanism for device specification updates, enabling asynchronous and efficient distribution of updates to multiple components.
  - Added a generic, thread-safe ring buffer for managing update queues.
  - Added a new console controller run loop that manages console sessions based on device spec subscriptions.

- **Bug Fixes**
  - Improved handling of device spec updates to prevent missed or duplicate updates during synchronization.

- **Refactor**
  - Replaced direct management API polling and retry logic with a publisher-subscriber model for device spec management.
  - Streamlined console and agent synchronization to use event-driven updates.
  - Removed direct client setting and retry polling from spec manager, shifting to subscription consumption.

- **Tests**
  - Added comprehensive unit tests for the new publisher and ring buffer components.
  - Updated existing tests to align with the new subscription-based architecture.


internal/agent/device/device.go
```diff
@@ -106,15 +109,26 @@ func NewAgent(
...
+	go notifierEngine.Run(ctx, &wg)
 
-	return engine.Run(ctx)
+	// Engine to handle device spec and status updates
+	specEngine := NewEngine(
+		NewUnit(time.Duration(a.fetchSpecInterval), a.syncDeviceSpec),
+		NewUnit(time.Duration(a.statusUpdateInterval), a.statusUpdate),
+	)
+	specEngine.Run(ctx, &wg)
+	wg.Wait()
```
hexfusion, line 130: nit: we should not need a wait group if all goroutines honor ctx. engine.Run can just block?

internal/agent/device/device.go
```diff
@@ -106,15 +109,26 @@ func NewAgent(
...
+	// Goroutine to handle spec notifier which reads rendered devices from the server
+	notifierEngine := NewEngine(
+		NewUnit(time.Duration(a.fetchSpecInterval), a.specNotifier.PollAndNotify),
 	)
+	go notifierEngine.Run(ctx, &wg)
 
-	return engine.Run(ctx)
+	// Engine to handle device spec and status updates
+	specEngine := NewEngine(
+		NewUnit(time.Duration(a.fetchSpecInterval), a.syncDeviceSpec),
```
hexfusion, line 126: question: Why are we making this generic if it is unnecessary for this change? I believe we can leave the existing sync logic in place, including the engine. My concern is this gives the impression that other things besides spec and status should be part of this loop. Which if anything we would be reducing down to one before we extend,

internal/agent/device/device.go
```diff
@@ -195,6 +209,26 @@ func (a *Agent) sync(ctx context.Context, current, desired *v1alpha1.Device) err
 	return nil
 }
 
+func (a *Agent) syncConsoles(ctx context.Context, wg *sync.WaitGroup) {
```
hexfusion, line 212: this should live in the consile controller shouldn't it

internal/agent/device/device.go
```diff
@@ -195,6 +209,26 @@ func (a *Agent) sync(ctx context.Context, current, desired *v1alpha1.Device) err
...
+	defer wg.Done()
```
hexfusion, line 213: wait group is sort of redundant if we return on ctx cancel.

internal/agent/device/engine.go
```diff
@@ -2,93 +2,101 @@ package device
...
+	u.Fn(ctx)
+	u.lastRun = now
+}
+
 type Engine struct {
-	fetchSpecInterval  util.Duration
-	fetchSpecFn        func(context.Context)
-	pushStatusInterval util.Duration
-	pushStatusFn       func(context.Context)
+	units []*Unit
```
hexfusion, line 35: engine logic change does not seem necessary and could lead to future confusion.

internal/agent/device/spec/notifier.go
```diff
@@ -0,0 +1,124 @@
...
+		return
+	}
+
+	n.lastKnownVersion = newDesired.Version()
+
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	// notify all subscribers of the new device spec
+	for _, ch := range n.subscribers {
```
hexfusion, line 105: does this channel need to be closed?

internal/agent/device/spec/notifier.go
```diff
@@ -0,0 +1,124 @@
...
+		default:
+			n.log.Warnf("Subscriber channel is full, skipping notification")
+		}
+	}
+}
+
+func (n *notifier) Subscribe() chan *v1alpha1.Device {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	ch := make(chan *v1alpha1.Device, 100)
```
hexfusion, line 117: how large is 100 devices object could any device handle that. I get its a trade off of blocking vs memory but this is complexity introduced with channels

internal/util/util.go
```diff
@@ -311,3 +311,11 @@ func Clone[T any](t *T) *T {
 	ret := *t
 	return &ret
 }
+
+// Greatest Common Divisor (Euclidean Algorithm)
+func Gcd[N constraints.Integer](a, b N) N {
```
hexfusion, line 316: i dont see where this is used

hexfusion (changes requested):
please hold merge until qe has validated.

internal/agent/device/spec/notifier.go
ori-amizur: It is an slice of channels.  Not a single channel

internal/agent/device/spec/notifier.go
ori-amizur: It never blocks but it should be a limit that is never reached.  If the channel is full then the data is lost.

internal/agent/device/device.go
ori-amizur: The notifier runs with an engine.  It is possible to create a dedicated loop instead.

internal/agent/device/device.go
ori-amizur: OK

internal/util/util.go
ori-amizur: By the engine to calculate the needed wait duration.

internal/agent/device/device.go
ori-amizur: To verify that other goroutines terminated.

internal/agent/device/engine.go
ori-amizur: If it seems confusing it can be reverted.  For the notifier a similar logic needs to be created.

internal/agent/device/device.go
hexfusion: yeah or move the engine out of agent. if we can keep this logic the same for this pr I think it will help with risk.

internal/agent/device/device.go
hexfusion: why would that matter here is the coordination part of some shutdown routine? if yes then maybe we should comment on what we are waiting for.

ori-amizur: It is cleaner this way.  When the function terminates we know that all the created goroutines were terminated

internal/agent/device/device.go
ori-amizur: Reverted the engine

internal/agent/device/spec/notifier.go
ori-amizur: Changed to ring buffer implementation where the newest overrides the oldest.  This means that the newest will never be discarded even with small capacity.  Therefore, the limit was changed to 3.

internal/util/util.go
ori-amizur: Removed

hexfusion (approved):
thanks @ori-amizur

PR #1114, opened 2025-04-28
NO-ISSUE: Add documentation for sos reports and console

ori-amizur:
## Summary by CodeRabbit

- **Documentation**
  - Expanded instructions for remote device access with updated disconnect commands and new examples for running commands and downloading files via the console.
  - Replaced log bundle generation with a detailed guide on creating and downloading SOS reports for troubleshooting, including examples and checksum verification.


docs/user/troubleshooting.md
```diff
@@ -27,3 +27,36 @@ sudo flightctl-must-gather
 ```
 
 The output is a tarball named `must-gather-$timestamp.tgz`, whereby `$timestamp` is the current date and time. Include this tarball in the bug report.
+
```
avishayt, line 30: We can remove the must-gather section above.

docs/user/managing-devices.md
```diff
@@ -661,7 +661,19 @@ To connect, use the `flightctl console` command specifying the device's name, an
 flightctl console device/<some_device_name>
 ```
 
-To disconnect, enter "exit" on the console. To force-disconnect, press `<ctrl>+b` three times.
+To disconnect, enter "exit" on the console.
```
fzdarsky, line 664: Does this mean we no longer have a way to recover from a hanging console?

docs/user/troubleshooting.md
```diff
@@ -27,3 +27,36 @@ sudo flightctl-must-gather
...
+## Generating and downloading SOS report
```
fzdarsky, line 31:
```suggestion
## Generating and downloading an SOS report
```

docs/user/managing-devices.md
ori-amizur: Since we can send binary information through the console, we cannot assume a certain sequence causes disconnection.
SSH doesn't have such sequence.
To disconnect a session there are the following possibilities:
- Kill the console session locally
- Connect with another console session to the device and kill the session.

docs/user/troubleshooting.md
ori-amizur: OK

docs/user/troubleshooting.md
ori-amizur: It still exists.

docs/user/managing-devices.md
fzdarsky: > SSH doesn't have such sequence.
It does: The escape sequence is `Enter`, `~`, `.`.  See, e.g., [here](https://foosel.net/til/about-ssh-escape-sequences/).

docs/user/troubleshooting.md
```diff
@@ -27,3 +27,36 @@ sudo flightctl-must-gather
 ```
 
 The output is a tarball named `must-gather-$timestamp.tgz`, whereby `$timestamp` is the current date and time. Include this tarball in the bug report.
+
+## Generating and downloading an SOS report
+
+SOS reports are a collection of system information that can be used to debug issues with the device agent. The SOS report is generated by the `sosreport` command, which is included in the device image.
```
fzdarsky, line 33:
```suggestion
SOS reports are a collection of system information that can be used to debug issues with the device agent. The SOS report is generated by the `sos report` command, which is included in the device image.
```

docs/user/troubleshooting.md
```diff
@@ -27,3 +27,36 @@ sudo flightctl-must-gather
...
 The output is a tarball named `must-gather-$timestamp.tgz`, whereby `$timestamp` is the current date and time. Include this tarball in the bug report.
+
+## Generating and downloading SOS report
+
+SOS reports are a collection of system information that can be used to debug issues with the device agent. The SOS report is generated by the `sosreport` command, which is included in the device image.
+To generate an SOS report, run the following console command on the device:
+
+```console
+flightctl console device/${device_name} -- sos report --batch --quiet
+```
```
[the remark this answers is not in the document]
fzdarsky: `sosreport` was the command for RHEL7 and earlier, the command for RHEL8 and higher is `sos report` (see [this KB article](https://access.redhat.com/solutions/3592#sos_report)).

docs/user/troubleshooting.md
fzdarsky: It should be obsolete with `sos report` now, so we should remove from docs and RPM package.

ori-amizur: OK

docs/user/managing-devices.md
ori-amizur: OK, so we need to distinguish between valid binary input and such escape sequence.
The previous approach [scans](https://github.com/flightctl/flightctl/blob/76486ae190a6d9eebfd54b93498ea43f8d298e39/internal/cli/console.go#L195-L202)  the input.  It is not practical for tool supporting uploads.

docs/user/troubleshooting.md
ori-amizur: OK

docs/user/managing-devices.md
avishayt: There are a couple ways to fix this:
1. SSH only escapes if there's a newline before "~.", reducing the chance of accidentally killing the connection.  It also contains two characters, whereas ours contains one.
2. When running in non-interactive mode, disable this functionality.  SSH does this when executing remote commands as well (e.g., ssh host "cmd").

ori-amizur: Only the second is a valid option, since when uploading a file, we cannot assume any sequence is unexpected.
Will implement only for interactive mode.

ori-amizur: Will change to "~." since <ctrl>-B is used by "less" and "vim" as page-up.

avishayt (approved):

PR #1116, opened 2025-04-28
EDM-1016: Add TPM client agent-side

lkatalin:
This functionality will be used by the system info collector to gather and send TPM information during the device's enrollment request.

## Summary by CodeRabbit

- **New Features**
	- Introduced support for TPM (Trusted Platform Module) device interactions, including attestation and nonce management.
- **Bug Fixes**
	- Improved error handling when closing TPM connections by propagating errors properly.

### DIFF
--- a/internal/agent/device/lifecycle/tpm.go
+++ b/internal/agent/device/lifecycle/tpm.go
@@ -1,0 +1,87 @@
+package lifecycle
+
+import (
+	"context"
+	"fmt"
+
+	fcrypto "github.com/flightctl/flightctl/internal/crypto"
+	"github.com/flightctl/flightctl/internal/tpm"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/google/go-tpm-tools/client"
+)
+
+type TpmClient struct {
+	// path where the tpm will be found, if it exists
+	tpmSysPath string
+	// handle to the open TPM
+	tpm *tpm.TPM
+	// local attestation public key
+	lak *client.Key
+	// local attestation public key hash
+	lakPubKeyHash []byte
+	// meant to be updated for each attestation
+	currNonce []byte
+	log       *log.PrefixLogger
+}
+
+func NewTpmClient(log *log.PrefixLogger) (*TpmClient, error) {
+	tc := TpmClient{
+		tpmSysPath: tpm.TpmSystemPath,
+		log:        log,
+	}
+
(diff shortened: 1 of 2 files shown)


internal/agent/device/lifecycle/tpm.go
```diff
@@ -0,0 +1,87 @@
...
+	if tc.tpm != nil {
+		err := tc.tpm.Close()
+		tc.tpm = nil
+		return err
+	}
+	return nil
+}
+
+func (tc *TpmClient) TpmAttestationCollector(ctx context.Context) string {
+	att, err := tc.tpm.GetAttestation(tc.currNonce, tc.lak)
```
hexfusion, line 73: if tpm is nil until we OpenTPM shouldn't we do that first or at least ensure its not nil

internal/agent/device/lifecycle/tpm.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		tc.log.Infof("Unable to get TPM attestation: %v", err)
```
hexfusion, line 75: info or warn or is this an error

internal/agent/device/lifecycle/tpm.go
```diff
@@ -0,0 +1,87 @@
...
+		return ""
+	}
+	return att.String()
+}
+
+func (tc *TpmClient) UpdateNonce(nonce []byte) error {
+	if len(nonce) < tpm.MinNonceLength {
+		return fmt.Errorf("nonce does not meet minimum length of %d bytes", tpm.MinNonceLength)
+	}
+	tc.currNonce = nonce
```
hexfusion, line 85: is it an error if the nonce is the same as before?

internal/agent/device/lifecycle/tpm.go
lkatalin: I suppose in the log it could be an error, yeah

internal/agent/device/lifecycle/tpm.go
lkatalin: It should be open via the constructor, but I'll add a check

internal/agent/device/lifecycle/tpm.go
lkatalin: Hmmm this is a good question

lkatalin:
@hexfusion I'm closing this in favor of #1121 but I'll update based on your comments

PR #1117, opened 2025-04-29
NO-ISSUE: api/proto: add GetSessions rpc

hexfusion:
Since the agent already has a client for RouterService, we should expose active sessions directly through a unary gRPC call rather than embedding them into the DeviceSpec.

This approach has several advantages:

Decouples consoles from spec updates: consoles can query sessions independently without relying on generating a new rendered device spec.

Simplifies DeviceSpec: keeps session state out of the desired/observed spec, which should focus on long-lived configuration, not ephemeral connection state.

Improves responsiveness: consoles can immediately reflect session changes without waiting for agent heartbeats or spec resyncs.


hexfusion:
cc @ori-amizur

PR #1119, opened 2025-04-29
EDM-1061: Add function to get TPM vendor info

lkatalin:
See title. We may refine this function to add more information in a future PR.

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to retrieve TPM vendor information.

### DIFF
--- a/internal/tpm/tpm.go
+++ b/internal/tpm/tpm.go
@@ -71,6 +71,16 @@ func (t *TPM) Close() {
 	}
 }
 
+func (t *TPM) GetTpmVendorInfo() ([]byte, error) {
+	if t == nil {
+		return nil, fmt.Errorf("cannot get TPM vendor info: nil receiver in TPM struct")
+	}
+	if t.channel == nil {
+		return nil, fmt.Errorf("cannot get TPM vendor info: no channel available in TPM struct")
+	}
+	return tpm2.GetManufacturer(t.channel)
+}
+
 func (t *TPM) GetPCRValues(measurements map[string]string) error {
 	if t == nil {
 		return nil


internal/tpm/tpm.go
```diff
@@ -71,6 +71,11 @@ func (t *TPM) Close() {
 	}
 }
 
+func (t *TPM) GetTpmVendorInfo() ([]byte, error) {
+	return tpm2.GetManufacturer(t.channel)
+
```
hexfusion, line 76: nit: drop newline afer return

internal/tpm/tpm.go
```diff
@@ -71,6 +71,11 @@ func (t *TPM) Close() {
 	}
 }
 
+func (t *TPM) GetTpmVendorInfo() ([]byte, error) {
+	return tpm2.GetManufacturer(t.channel)
```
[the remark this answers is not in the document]
lkatalin: Added a check

internal/tpm/tpm.go
```diff
@@ -71,6 +71,13 @@ func (t *TPM) Close() {
 	}
 }
 
+func (t *TPM) GetTpmVendorInfo() ([]byte, error) {
+	if t == nil || t.channel == nil {
+		return []byte{}, nil
```
hexfusion, line 76: this isn't an error? should we log this at least it feels pretty unexpected

lkatalin: Hmm. Initially I thought it might be fine, but I think you are right.

hexfusion (approved):

PR #1121, opened 2025-04-29
EDM-1016: Add experimental agent-side TPM functionality

lkatalin:
This PR:
- Adds a TPMClient to interact with the agent-side TPM
- Adds functionality to use a TPM-based device identity key (which also signs the CSR for enrollment) behind an experimental feature gate
- Adds functionality to collect TPM vendor info and attestation behind an experimental feature gate


internal/agent/agent.go
```diff
@@ -76,9 +77,44 @@ func (a *Agent) Run(ctx context.Context) error {
...
 	}
-	publicKey, privateKey, _, err := fcrypto.EnsureKey(deviceReadWriter.PathFor(a.config.ManagementService.AuthInfo.ClientKey))
-	if err != nil {
-		return err
+
+	experimentalFeatures := experimental.NewFeatures()
+	var publicKey crypto.PublicKey
+	var privateKey crypto.PrivateKey
+	var signer crypto.Signer
+	var tpmClient *lifecycle.TpmClient
```
hexfusion, line 85: style: single var please

internal/agent/agent.go
```diff
@@ -76,9 +77,44 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			a.log.Infof("tpm is not available: %v", err)
+			tpmClient = nil
+		} else {
+
+			publicKey = tpmClient.GetLocalAttestationPubKey()
+			signer, err = tpmClient.GetSigner()
+			if err != nil {
+				a.log.Errorf("unable to get TPM-based signer for CSR: switching to software keys instead")
+				tpmClient = nil
+			}
```
hexfusion, line 101: can encapsulate this logic into something where the output is the key and signer etc and the implementation including experimental is abstracted away.

internal/agent/agent.go
```diff
@@ -204,6 +240,15 @@ func (a *Agent) Run(ctx context.Context) error {
 	statusManager.RegisterStatusExporter(specManager)
 	statusManager.RegisterStatusExporter(systemInfoManager)
 
+	// register internal system info collectors
+	if experimentalFeatures.IsEnabled() {
+		a.log.Infof("experimental features enabled: registering TPM info collection functions")
```
hexfusion, line 245: lets warn here

internal/agent/device/lifecycle/tpm.go
```diff
@@ -0,0 +1,101 @@
...
+	if err := tc.OpenTPM(); err != nil {
+		return nil, err
+	}
+
+	lak, err := tc.tpm.CreateLAK()
+	if err != nil {
+		return nil, err
+	}
+	tc.lak = lak
+
```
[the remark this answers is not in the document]
hexfusion: what are the expectations for open close is this like a file where we leak if we don't close

PR #1122, opened 2025-04-29
EDM-1016: Include collector function data in system info

lkatalin:
Include collector function data in system info


internal/agent/device/systeminfo/system_info.go
```diff
@@ -564,6 +564,16 @@ func getSystemInfoMap(ctx context.Context, log *log.PrefixLogger, info *Info, in
 		infoMap[key] = val
 	}
 
+	for key, collectorfn := range collectors {
+		_, alreadyExists := infoMap[key]
+		if alreadyExists {
+			log.Debugf("SystemInfo collector already populated: %s is %s", key, infoMap[key])
```
avishayt, line 570: IIUC, this is a user-defined collector trying to override a key whose value the agent already determined?  If so, we should allow the user to override it in case the agent isn't collecting the information correctly for the given hardware.

lkatalin: That's a good insight, @hexfusion what do you think?

hexfusion: The contract is that root keys in systemInfo are system-generated, best effort. In the case use wants to provide another mechanism it can be declared on the filesystem as part of customInfo. In this case this would be considered a system-level collector. Docs will reflect.

hexfusion, line 570: nit: we should really warn here as this would be unexpected to duplicate an internal implementation.

internal/agent/device/systeminfo/manager.go
```diff
@@ -155,6 +156,7 @@ func collectDeviceSystemInfo(
 	infoKeys []string,
 	customKeys []string,
 	bootID string,
+	collectors map[string]CollectorFn,
```
hexfusion, line 159: :+1:

hexfusion (approved):
non functional change because no collectors are registered

PR #1124, opened 2025-04-30
EDM-1497: api: extend ApplicationProviderSpec to support volume providers

hexfusion:
This PR adds image based volume providers to the application API.

### examples
```
     applications:
        - name: my-inline
          appType: compose
          inline:
            - path: docker-compose.yaml
              content: |
                version: "3.8"
                services:
                  service1:
                    image:  quay.io/flightctl-tests/alpine:v1
                    command: ["sleep", "infinity"]
                    volumes:
                      - my-inline-data:/data
                volumes:
                  my-inline-data:
                    external: true      
          volumes:
            - name: my-data
              image:
                reference: quay.io/rawagner/models/gpt2
                pullPolicy: IfNotPresent
        - name: my-image-app
          image: quay.io/flightctl-tests/alpine:v1
          volumes:
            - name: my-image-data
              image:
                reference: quay.io/rawagner/models/gemma
                pullPolicy: Always
```

## Summary by CodeRabbit

- **New Features**
  - Added support for defining and tracking application volumes, including image-based volumes with configurable image pull policies.
  - Extended application and device status schemas to include volume information and statuses.
- **Bug Fixes**
  - Improved validation to enforce unique application and volume names, and to validate image references and volume names.
- **Tests**
  - Introduced comprehensive tests for application and volume validation scenarios.


api/v1alpha1/openapi.yaml
```diff
@@ -3756,8 +3756,26 @@ components:
...
             $ref: "#/components/schemas/ApplicationContent"
+        volumes:
+          type: array
+          description: List of volume providers.
+          items:
+            oneOf:
+              - $ref: "#/components/schemas/ImageVolumeProviderSpec"
       required:
         - inline
+    ImageVolumeProviderSpec:
```
rawagner, line 3767: Should this be called OCI Artifact ?
```suggestion
    OciArtifactVolumeProviderSpec:
```

api/v1alpha1/openapi.yaml
```diff
@@ -3756,8 +3756,26 @@ components:
...
+      type: object
+      properties:
+        image:
+          type: string
+          description: Reference to the container image for the volume provider.
```
rawagner, line 3772:
```suggestion
        ociArtifact:
          type: string
          description: Reference to the OCI Artifact for the volume provider. 
```

api/v1alpha1/openapi.yaml
```diff
@@ -3756,8 +3756,26 @@ components:
           description: A list of application content.
           items:
             $ref: "#/components/schemas/ApplicationContent"
+        volumes:
```
celdrake, line 3759: Since this PR adds a new field to an existing object, we'd need an adaptation in the UI to make sure that for inline applications that have defined volumes, they won't get removed as the field wouldn't be included in the PATCH request the UI sends to update the application.

Also, we may need to discuss whether this is only a API/CLI feature, or it should also be included in the UI.

I imagine this feature is not for 0.7?

hexfusion: correct this is a sketch for some work @rawagner has been working on nd not planned for 0.7

api/v1alpha1/openapi.yaml
fzdarsky: Would it _always_ be an OCI artifact or could it also be populated from OCI image? Compare the [OCI Volume Source](https://github.com/kubernetes/enhancements/issues/4639) in K8s.

rawagner (approved):

api/v1alpha1/openapi.yaml
```diff
@@ -4131,6 +4191,20 @@ components:
           description: Number of restarts observed for the application.
         status:
           $ref: "#/components/schemas/ApplicationStatusType"
+        volumes:
+          type: array
+          description: Status of volumes used by this application.
```
hexfusion, line 3984: We could assume this as long as the application container is running. I am not sure of the value we gain here.

celdrake (commented):
~~LGTM.~~ ~~Adding comments for ApplicationVolume as I noticed a problem there.~~
Back to LGTM

api/v1alpha1/openapi.yaml
```diff
@@ -3741,23 +3741,71 @@ components:
...
+      properties:
+        volumes:
+          type: array
+          description: List of application volumes.
+          items:
+            $ref: "#/components/schemas/ApplicationVolume"
+    ApplicationVolume:
+      allOf:
+        - type: object
+          description: Defines a named volume used by an application.
```
celdrake, line 3570: `ApplicationVolume` defines the `name` property twice.
Once here, as an optional prop. Second as part of `ImageVolumeProviderSpec`

hexfusion: yikes good catch

api/v1alpha1/openapi.yaml
```diff
@@ -4131,6 +4187,20 @@ components:
...
+            $ref: "#/components/schemas/ApplicationVolumeStatus"
+    ApplicationVolumeStatus:
+      type: object
+      description: Status of a volume used by an application.
+      required:
+        - name
+      properties:
+        name:
+          type: string
+          description: Name of the volume.
```
fzdarsky, line 3996: So we'd only show which volumes have been defined for the app, but not their status. If a download of a volume takes significant amount of time, would we show the status as part of the application status instead? Do we foresee the ability to update volumes without updating the running application?

hexfusion: I have considered this and I think it sort of ties into a larger question. If the volume is tied to an application similar to a pod, in kube volumes are considered immutable. In this case the lifecycle of the volume is tied to that of the application.

If we want to have volumes that have a lifecycle outside of an application then I agree this would need a more complete status. So I think we are talking about persistent volumes vs pod/application volumes?

rawagner:
We should consider putting the volumes directly under `spec` (on the same level as apps).
This way we would align with what we are actually doing on the device - creating podman volumes which are not scoped to specific application. These volumes can be reused across multiple apps.

hexfusion:
> We should consider putting the volumes directly under `spec` (on the same level as apps). This way we would align with what we are actually doing on the device - creating podman volumes which are not scoped to specific application. These volumes can be reused across multiple apps.

Good topic. So speaking with @fzdarsky we agreed that in terms of isolation each application could be seen as its own namespace. In that context, volumes fit more of the configmap/secret model in terms of namespaces.

However, in terms of implementation details, the underlying volume is as you said global similar to a PV in RWX.

api/v1alpha1/openapi.yaml
hexfusion: Application volume data will be treated the same as all beforeUpdate image pulls (os, application) and pulled before the application would try to start. So to answer your question volume image fetch status would be reported as part of the update status. This would mean for a new application the image/artifact would be pulled before the application existed. Also then on volume reference update the data would be pulled beforeUpdate, then remounted. If for example we lack auth to pull the image this would bubble up as an update status error.

So I think reporting the name of the running volume and the reference is reasonable. You could know if the underlying content for the mount has be mounted if the reference matches the spec. If in the future we want to extend this we can WDYT?

fzdarsky: Yeah, that's a very good point. I agree that because we're pulling all content pre update, it'd be reasonable to report that in the update status.

I'm wondering how it'd work with updates that need to pull multiple images, though (say, a bootc image, a few app containers, a data volume). What would the status show in the good case and in the error case? And do we need that info to be both human-readable and machine-readable?

In the good case, I'd assume a user would want to at least know the download progress ("downloading (80%, 2 out of 10 complete)".
In the error case, a user would want to know which download failed (could be multiple) and why (missing pull secret, server unavailable, ...).

PR #1136, opened 2025-05-05
EDM 1362:  CLI commands use different styles of version argument

SiddarthR56:


DakCrowder (approved):

gshilin-sdb (approved):

fzdarsky:
What's the motivation for replacing the argument parsing with pflags that we use in all of our commands and implement implement this functionality manually and in a format that no longer does the POSIX/GNU-style --flags?

hexfusion (changes requested):
I talked to @SiddarthR56 about this and this is not what we want for an end solution. Do not merge

hexfusion (approved):

PR #1140, opened 2025-05-06
EDM-1106: The CLI should report the versions of the CLI and Service

eldar101:
## Summary by CodeRabbit

- **Tests**
  - Added end-to-end test to verify that the CLI client and server versions are displayed and match when running the version command.

### DIFF
--- a/test/e2e/cli/cli_test.go
+++ b/test/e2e/cli/cli_test.go
@@ -284,6 +284,26 @@ var _ = Describe("cli operation", func() {
 		})
 	})
 
+	Context("Flightctl Version Checks", func() {
+		It("should show matching client and server versions", Label("79621"), func() {
+			By("Getting the version output")
+			out, err := harness.CLI("version")
+			clientVersionPrefix := "Client Version:"
+			serverVersionPrefix := "Server Version:"
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring(clientVersionPrefix))
+			Expect(out).To(ContainSubstring(serverVersionPrefix))
+
+			By("Parsing client and server versions")
+			clientVersion := GetVersionByPrefix(out, clientVersionPrefix)
+			serverVersion := GetVersionByPrefix(out, serverVersionPrefix)
+
+			Expect(clientVersion).ToNot(BeEmpty(), "client version should be found")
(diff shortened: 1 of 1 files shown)


test/e2e/cli/cli_test.go
```diff
@@ -284,6 +284,32 @@ var _ = Describe("cli operation", func() {
...
+			out, err := harness.CLI("version")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring("Client Version:"))
+			Expect(out).To(ContainSubstring("Server Version:"))
+
+			By("Parsing client and server versions")
+			lines := strings.Split(out, "\n")
+			var clientVersion, serverVersion string
+			for _, line := range lines {
+				if strings.HasPrefix(line, "Client Version:") {
```
siserafin, line 299: Maybe here I would create two helper functions GetOutputClientVersion and GetOutputServerVersion, just to make more readable the test

test/e2e/cli/cli_test.go
```diff
@@ -341,6 +359,24 @@ var _ = Describe("cli login", func() {
 	})
 })
 
+func GetOutputClientVersion(output string) string {
+	for _, line := range strings.Split(output, "\n") {
```
siserafin, line 363: These two functions repeat the same code. To even better optimize,  you could create a third generic function GetOutputVersion(output, type) output, that is called from the other two.

test/e2e/cli/cli_test.go
```diff
@@ -284,6 +284,24 @@ var _ = Describe("cli operation", func() {
 		})
 	})
 
+	Context("Flightctl Version Checks", func() {
+		It("should show matching client and server versions", Label("697347"), func() {
```
amalykhi, line 288: The TC doesn't exist in P-n

test/e2e/cli/cli_test.go
```diff
@@ -284,6 +284,24 @@ var _ = Describe("cli operation", func() {
...
+			By("Getting the version output")
+			out, err := harness.CLI("version")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring("Client Version:"))
```
siserafin, line 292: "Client Version:" and "Service Version:" can be variables, since they are now repeated twice

siserafin (approved):

PR #1143, opened 2025-05-07
NO-ISSUE: TPM simulator is not supported on ppc64le and s390x, i.e. !(amd64 || argm64)

gshilin-sdb:
## Summary by CodeRabbit

- **New Features**
  - Improved support for TPM simulator functionality, now available only on supported platforms.
- **Bug Fixes**
  - Provides clear error messaging when TPM simulator is not supported on the current platform.

### DIFF
--- a/internal/tpm/nosimulator.go
+++ b/internal/tpm/nosimulator.go
@@ -1,0 +1,9 @@
+//go:build !(amd64 || arm64)
+
+package tpm
+
+import "errors"
+
+func OpenTPMSimulator() (*TPM, error) {
+	return &TPM{}, errors.New("TPM simulator not supported")
+}
--- a/internal/tpm/simulator.go
+++ b/internal/tpm/simulator.go
@@ -1,0 +1,15 @@
+//go:build amd64 || arm64
+
+package tpm
+
+import (
+	tpmSimulator "github.com/google/go-tpm-tools/simulator"
+)
+
+func OpenTPMSimulator() (*TPM, error) {
+	simulator, err := tpmSimulator.Get()
+	if err != nil {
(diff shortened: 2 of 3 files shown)


lkatalin (approved):
This fix makes sense IMO, thank you.
@hexfusion @fzdarsky or others any thoughts?

internal/tpm/nosimulator.go
```diff
@@ -0,0 +1,9 @@
+//go:build !(amd64 || arm64)
+
+package tpm
+
+import "errors"
+
+func OpenTPMSimulator() (*TPM, error) {
+	return &TPM{}, errors.New("TPM simulator not supported")
```
hexfusion, line 8: Doesnt this mean tests will always fail for other distros

lkatalin: We currently skip the event log test if the event log isn't available, so we could do similar for TPM in general.

hexfusion: Makes sense vs perm fail.

lkatalin:
@fzdarsky Has also pointed out that since we use the simulator only for testing, it should live in a `_test.go` file and that would sidestep this problem. I can also submit a PR for that.

lkatalin:
I have an alternative fix in #1146 based on the suggestion from @fzdarsky . PTAL @hexfusion .
@gshilin-sdb Has told me he will need a fix in by tomorrow (Thursday) morning (European time).

gshilin-sdb:
This solution is better: to require TPM simulator for testing only

NO-ISSUE: Move TPM Simulator to testing only #1146

PR #1148, opened 2025-05-07
NO-ISSUE: device/applications: cleanup race

hexfusion:
Stop is called by the shutdown manager in another goroutine so its important that cmd and cancelFn are guarded to prevent races. Although this should not happen in practice it should be cleaned up.

## Summary by CodeRabbit

- **Refactor**
  - Improved concurrency safety and thread management for application monitoring, reducing the risk of race conditions and enhancing overall stability.

### DIFF
--- a/internal/agent/device/applications/podman_monitor.go
+++ b/internal/agent/device/applications/podman_monitor.go
@@ -56,21 +56,28 @@ func NewPodmanMonitor(
 
 func (m *PodmanMonitor) Run(ctx context.Context) error {
 	m.log.Debugf("Starting podman monitor")
-	ctx, m.cancelFn = context.WithCancel(ctx)
+	ctx, cancelFn := context.WithCancel(ctx)
 
 	// list of podman events to listen for
 	events := []string{"create", "init", "start", "stop", "die", "sync", "remove", "exited"}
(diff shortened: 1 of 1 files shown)


internal/agent/device/applications/podman_monitor.go
```diff
@@ -409,11 +431,6 @@ func (m *PodmanMonitor) updateAppStatus(ctx context.Context, app Application, ev
 		return
 	}
 
-	restarts, err := m.getContainerRestarts(inspectData)
-	if err != nil {
-		m.log.Errorf("Failed to get container restarts: %v", err)
-	}
-
```
hexfusion, line 416: note: moved so lock is held after

internal/agent/device/applications/podman_monitor.go
```diff
@@ -392,14 +409,19 @@ func (m *PodmanMonitor) handleEvent(ctx context.Context, data []byte) {
 }
 
 func (m *PodmanMonitor) updateAppStatus(ctx context.Context, app Application, event *client.PodmanEvent) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
```
hexfusion, line 396: tightened up locks

PR #1159, opened 2025-05-09
EDM-1550: agent/applications/providers: add image volume support

hexfusion:


api/v1alpha1/openapi.yaml
```diff
@@ -3756,8 +3756,36 @@ components:
           description: A list of application content.
           items:
             $ref: "#/components/schemas/ApplicationContent"
+        volumes:
```
rawagner, line 3759: Can you add this for the ` ImageApplicationProviderSpec` - which is actually more useful in context of catalog right now.

hexfusion: pushed

avishayt:
Tests are failing, but otherwise nice work! Very cool!

internal/agent/device/applications/provider/volume_test.go
```diff
@@ -0,0 +1,179 @@
...
+			},
+			renamed: map[string]string{"vol1": "vol1-renamed"},
+			expected: `services:
+  svc1:
+    image: nginx
+    volumes:
+    - vol1-renamed:/mnt
+volumes:
+  vol1-renamed:
+    external: true`,
```
rawagner, line 102: The override content should be just
```suggestion
			expected: `volumes:
  vol1:
    external:
      name: vol1-renamed`,
```

You do not need to modify the service definition or volume key. You just need to add the `name: <unique_name>` See [compose-spec docs](https://github.com/compose-spec/compose-spec/blob/main/07-volumes.md#name)

internal/agent/client/podman.go
```diff
@@ -287,14 +309,63 @@ func (p *Podman) RemoveContainer(ctx context.Context, labels []string) error {
...
+	args := []string{"volume", "inspect", name, "--format", "{{.Mountpoint}}"}
+	stdout, stderr, exitCode := p.exec.ExecuteWithContext(ctx, podmanCmd, args...)
+	if exitCode != 0 {
+		return "", fmt.Errorf("inspect volume mountpoint: %w", errors.FromStderr(stderr, exitCode))
+	}
+
+	return strings.TrimSpace(stdout), nil
+}
+
+func (p *Podman) RemoveVolumes(ctx context.Context, volumes ...string) error {
```
kkyrazis, line 359: Is there a preference for removing one by one vs all in one go?
I.E.
```go
if len(volumes) == 0 {
  return nil
}

nctx, cancel := context.WithTimeout(ctx, p.timeout)
defer cancel()
args := append([]string{"volume", "rm"}, volumes...)
_, stderr, exitCode := p.exec.ExecuteWithContext(nctx, podmanCmd, args...)
if exitCode != 0 {			
  return fmt.Errorf("remove volumes: %w", errors.FromStderr(stderr, exitCode))
}
return nil
```

internal/agent/device/applications/provider/volume.go
```diff
@@ -0,0 +1,298 @@
...
+		switch vType {
+		case v1alpha1.ImageApplicationVolumeProviderType:
+			v, err := volume.AsImageVolumeProviderSpec()
+			if err != nil {
+				return fmt.Errorf("getting image volume provider spec: %w", err)
+			}
+			var pullPolicy v1alpha1.ImagePullPolicy
+			if v.Image.PullPolicy == nil {
+				pullPolicy = v1alpha1.PullIfNotPresent
+			}
```
kkyrazis: Pull policy is either the zero value or PullIfNotPresent. I assume the intention is to use the Image's pull policy if it's available.

```suggestion
			pullPolicy := v1alpha1.PullIfNotPresent
			if v.Image.PullPolicy != nil {
				pullPolicy = *v.Image.PullPolicy
			}
```

internal/agent/client/podman.go
hexfusion: yeah this was stale we need to be perscriptive I don't know why this doesn't show out of date.

internal/agent/device/applications/provider/volume.go
hexfusion: yep that makes more sense

internal/agent/device/applications/provider/volume_test.go
hexfusion: I am going to follow up on this change as it will require a bit of rework but will make the release thanks again.

internal/agent/client/podman.go
```diff
@@ -147,9 +155,23 @@ func (p *Podman) pullImage(ctx context.Context, image string, pullSecretPath str
...
+	}
+	return strings.TrimSpace(stdout), nil
+}
+
+func (p *Podman) ExtractArtifact(ctx context.Context, artifact, destination string) (string, error) {
+	ctx, cancel := context.WithTimeout(ctx, p.timeout)
+	defer cancel()
+
+	// TODO: only available in podman >= 4.5.0
+	args := []string{"artifact", "extract", artifact, destination}
```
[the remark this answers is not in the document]
hexfusion: ensureDependenciesFromVolumes ensures  5.5 or greater

internal/agent/device/applications/provider/volume.go
```diff
@@ -0,0 +1,297 @@
...
+				return fmt.Errorf("pulling image volume: %w", err)
+			}
+		default:
+			return fmt.Errorf("%w: %s", errors.ErrUnsupportedVolumeType, vType)
+		}
+	}
+	return nil
+}
+
+// ensurePodmanVolumes creates and populates each image-backed volume in Podman.
```
DakCrowder: nit: comment for the method is slightly incorrect, mentions ensurePodmanVolumes and implies it handles more than one volume

internal/agent/device/applications/provider/volume.go
```diff
@@ -0,0 +1,297 @@
...
+	return nil
+}
+
+// ensureArtifactExists checks if an artifact exists locally and pulls it if needed.
+func ensureArtifactExists(ctx context.Context, log *log.PrefixLogger, podman *client.Podman, artifact string, pullPolicy v1alpha1.ImagePullPolicy) error {
+	if pullPolicy == v1alpha1.PullNever {
+		log.Tracef("Pull policy is set to never, skipping artifact pull: %s", artifact)
+		return nil
+	}
+	if podman.ImageExists(ctx, artifact) {
```
DakCrowder: Does this need to take into account the `Always` pull policy like in the application provider?  e.g.

```
if podman.ImageExists(ctx, image) && pullPolicy != v1alpha1.PullAlways {
...
```

hexfusion: good catch pull policy will need some cleanup but this helps  a lot thank

DakCrowder (approved):

PR #1175, opened 2025-05-15
NO-ISSUE: Console sessions are not terminated in case a close request arrives from server

ori-amizur:
If a close request arrives from server, the bash process may be left alive in the agent.
This change causes the bash process to terminate and clean up the session.
Also, when a websocket is closed and the remote stream running on the agent has not been closed, make sure that the reader from the stream terminates by returning from the gRPC call when the websocket is closed.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for outgoing streams to prevent unnecessary error logs and handle end-of-stream scenarios more gracefully.
	- Fixed potential deadlocks in channel forwarding for device console sessions, ensuring smoother session termination.

- **New Features**
	- Added explicit cancellation support for incoming streams and command execution, allowing sessions to be cancelled more reliably.

### DIFF
--- a/internal/agent/device/console/incoming_streams.go
+++ b/internal/agent/device/console/incoming_streams.go
@@ -1,6 +1,7 @@
 package console
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -19,13 +20,19 @@ type incomingStream interface {
 type incomingStreams struct {
 	streamClient grpc_v1.RouterService_StreamClient
 	streams      map[byte]incomingStream
+	cancel       context.CancelFunc
 	log          *log.PrefixLogger
 }
 
-func newIncomingStreams(streamClient grpc_v1.RouterService_StreamClient, stdin io.WriteCloser, resizeFd uintptr, closers map[byte]io.Closer, log *log.PrefixLogger) *incomingStreams {
+func newIncomingStreams(streamClient grpc_v1.RouterService_StreamClient,
+	stdin io.WriteCloser, resizeFd uintptr,
+	closers map[byte]io.Closer,
+	cancel context.CancelFunc,
+	log *log.PrefixLogger) *incomingStreams {
 	ret := &incomingStreams{
 		streamClient: streamClient,
 		streams:      make(map[byte]incomingStream),
+		cancel:       cancel,
 		log:          log,
 	}
 	ret.streams[StdinID] = newStdinStream(stdin, log)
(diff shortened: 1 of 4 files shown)


internal/agent/device/console/incoming_streams.go
```diff
@@ -38,6 +45,7 @@ func newIncomingStreams(streamClient grpc_v1.RouterService_StreamClient, stdin i
 
 func (i *incomingStreams) run(wg *sync.WaitGroup) {
```
hexfusion, line 46: if run was context-aware you could also ensure shutdown when the parent context is cancelled 

```go
      for {
		select {
		case <-ctx.Done():
			i.log.Debug("incoming streams cancelled")
			return
```

internal/agent/device/console/session.go
```diff
@@ -86,7 +87,7 @@ func (s *session) startProcess(metadata *api.DeviceConsoleSessionMetadata, cmd *
 }
 
 func (s *session) initialize(ctx context.Context, metadata *api.DeviceConsoleSessionMetadata) (*incomingStreams, *outgoingStreams, error) {
-	cmd := s.buildBashCommand(ctx, metadata)
```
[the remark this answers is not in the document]
ori-amizur: Removed

internal/api_server/agentserver/grpc.go
```diff
@@ -158,33 +158,38 @@ func (s *AgentGrpcServer) forwardConsoleSession(ctx context.Context, stream pb.R
...
 	for {
 		select {
 		case <-ctx.Done():
 			s.log.Debug("context is closed")
-			return io.EOF
+			return
 		default:
 			msg, err := a.Recv()
 			if err != nil {
 				s.log.Debugf("failed to receive a message: %v", err)
```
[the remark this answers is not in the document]
ori-amizur: Not really

internal/agent/device/console/incoming_streams.go
ori-amizur: The purpose of it is to call cancel when a request to close the stream is received by gRPC at line 59

hexfusion: right I just mean lets say the agent process exits shutdown will cancel the parent context. maybe this is already covered just an observation.

ori-amizur: Right, but the purpose of this change was to propagate connection closure which is signaled from the api-server.  Until now, it was not handled properly.

hexfusion (approved):

PR #1177, opened 2025-05-15
Local bootc services

DakCrowder:


deploy/services-vm.mk
```diff
@@ -0,0 +1,67 @@
...
+#
+# Example cmd:
+# PACKIT_CURRENT_VERSION=0.7.0 make services-container
+services-container: rpm
+	sudo podman build -t flightctl-services:latest -f test/scripts/services-images/Containerfile.services .
+
+run-services-container: services-container
+	sudo podman run -d --privileged --replace \
+	--name flightctl-services \
+	-p 8080:443 \
```
hexfusion, line 18: this feels awkward, is that correct? 8080 is usually not secure

packaging/rpm/flightctl.spec
```diff
@@ -10,6 +10,15 @@
...
     restorecon -v /usr/bin/flightctl-agent
+%define services_relabel_files() \
+    semanage fcontext -a -t container_file_t "%{_datadir}/flightctl(/.*)?"; \
+    semanage fcontext -a -t bin_t "%{_datadir}/flightctl/init_host.sh"; \
+    semanage fcontext -a -t bin_t "%{_datadir}/flightctl/secrets.sh"; \
+    semanage fcontext -a -t bin_t "%{_datadir}/flightctl/init_utils_host.sh"; \
+    restorecon -Rv %{_datadir}/flightctl
+%define services_remove_labels() \
+    semanage fcontext -d "%{_datadir}/flightctl(/.*)?" >/dev/null 2>&1 || true; \
+    restorecon -Rvf "%{_datadir}/flightctl(/.*)?" >/dev/null 2>&1 || true
```
hexfusion, line 21: super nitpicky, but if you are using recursive, is the regex needed like above?

deploy/scripts/shared.sh
```diff
@@ -69,7 +69,12 @@ move_shared_files() {
     cp "${source_dir}/podman/service-config.yaml" "${CONFIG_WRITEABLE_DIR}/service-config.yaml"
 
     # Copy read only files
+    #
+    # Due to selinux tag constraints and the fact init_utils is shared between scripts run on
+    # the host and in containers, we need to copy it twice so each can be tagged with
+    # the correct labels on rpm install
     cp "${source_dir}/scripts/init_utils.sh" "${CONFIG_READONLY_DIR}/init_utils.sh"
+    cp "${source_dir}/scripts/init_utils.sh" "${CONFIG_READONLY_DIR}/init_utils_host.sh"
```
hexfusion, line 77: question: Is it intentional to copy the same file to two locations or should the source here be init_utils_host.sh

deploy/services-vm.mk
DakCrowder: Ahh good catch, updated

deploy/scripts/shared.sh
DakCrowder: This is intentional - the issue is the file  is run in different contexts once installed and since I can't share a container_file_t tag and a bin_t tag on the file its a way to get the utils tagged/available in both contexts.  Could declare some new selinux policy but since it is only this file thought this was simpler for now.

PR #1182, opened 2025-05-19
Edm 1538 enable deletion of multiple resources via cli

SiddarthR56:
This PR enhances the flightctl delete command to support bulk deletion by allowing multiple resource names to be passed in a single invocation. For example:

flightctl delete device device1 device2

Additionally, the CLI supports shell expansions such as:

flightctl delete device $(flightctl get devices -o name)

### DIFF
--- a/internal/cli/delete.go
+++ b/internal/cli/delete.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 
 	"github.com/flightctl/flightctl/internal/client"
+	apiclient "github.com/flightctl/flightctl/internal/api/client"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
@@ -27,9 +28,9 @@ func DefaultDeleteOptions() *DeleteOptions {
 func NewCmdDelete() *cobra.Command {
 	o := DefaultDeleteOptions()
 	cmd := &cobra.Command{
-		Use:       "delete (TYPE | TYPE/NAME)",
-		Short:     "Delete resources by resources or owner.",
-		Args:      cobra.ExactArgs(1),
+		Use:       "delete (TYPE NAME [NAME...] | TYPE/NAME)",
+		Short:     "Delete one or more resources by name.",
+		Args:      cobra.MinimumNArgs(1),
 		ValidArgs: getValidResourceKinds(),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(cmd, args); err != nil {
@@ -67,9 +68,15 @@ func (o *DeleteOptions) Validate(args []string) error {
 		return err
 	}
 
-	kind, _, err := parseAndValidateKindName(args[0])
+	kind, name, err := parseAndValidateKindName(args[0])
 	if err != nil {
 		return err
+	}
+	if len(name) == 0 && len(args) < 2 {
+		return fmt.Errorf("name must be specified when deleting %s", kind)
+	}
+	if len(name) > 0 && len(args) > 1 {
+		return fmt.Errorf("invalid format: cannot mix TYPE/NAME syntax with additional resource names. Use either 'delete TYPE/NAME' or 'delete TYPE NAME [NAME...]'")
 	}
 	if kind == TemplateVersionKind && len(o.FleetName) == 0 {
 		return fmt.Errorf("fleetname must be specified when deleting templateversions")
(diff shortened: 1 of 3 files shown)


internal/cli/get.go
```diff
@@ -583,15 +574,45 @@ func (o *GetOptions) printCSRTable(w *tabwriter.Writer, csrs ...api.CertificateS
...
-			e.InvolvedObject.Name,
-			e.Type,
-			e.Message,
-		)
+func (o *GetOptions) printNames(responseJSON interface{}, name string) error {
+	jsonBytes, err := json.Marshal(responseJSON)
+	if err != nil {
+		return fmt.Errorf("marshalling JSON200: %w", err)
+	}
+
```
hexfusion, line 582: maybe we could reduce some of the boiler plate here with an inline type.

```go
type metadata struct {
	Name string `json:"name"`
}
var resource struct {
	Metadata metadata `json:"metadata"`
}
```

test/harness/e2e/harness.go
```diff
@@ -715,8 +715,22 @@ func (h *Harness) WaitForDeviceNewGeneration(deviceId string, newGeneration int6
...
-	return h.CLI("delete", resourceType)
 
+	resources, err := h.CLI("get", resourceType, "-o", "name")
+	if err != nil {
+		return "", fmt.Errorf("failed to get %s resources: %w", resourceType, err)
+	}
+	resources = strings.TrimSpace(resources)
+	if resources == "" {
+		logrus.Infof("No %s resources found to delete", resourceType)
+		return "No resources to delete", nil
```
hexfusion, line 726: this feels strange what is the context of this string it reads like an error

internal/cli/delete.go
```diff
@@ -88,42 +108,66 @@ func (o *DeleteOptions) Run(ctx context.Context, args []string) error { //nolint
...
+			return err
+		}
+		if err := processDeletionReponse(response, err, kind, name); err != nil {
+			return err
+		}
+		fmt.Printf("%s/%s deleted\n", kind, name)
+		return nil
+	}
+
+	names := args[1:]
```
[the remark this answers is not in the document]
SiddarthR56: Yes, One of the initial checks is that at least 1 arg is provided

internal/cli/delete.go
```diff
@@ -88,42 +108,66 @@ func (o *DeleteOptions) Run(ctx context.Context, args []string) error { //nolint
...
+func (o *DeleteOptions) deleteMultiple(ctx context.Context, c *apiclient.ClientWithResponses, kind string, names []string) error {
+	var err error
+
+	for _, name := range names {
+		response, deleteErr := o.deleteOne(ctx, c, kind, name)
+
+		processErr := processDeletionReponse(response, deleteErr, kind, name)
+		if processErr != nil {
+			fmt.Printf("Error: %v\n", processErr)
+			err = processErr
```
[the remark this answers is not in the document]
SiddarthR56: I don't return a slice of errors because I'm already printing each error as it occurs for immediate user feedback. The return error  just  signals that something failed to set the proper exit code for scripts. Returning all errors would be redundant since callers don't need to process the array. 
Changed approach to use bool var to track any failures and returns a simple summary error

test/harness/e2e/harness.go
SiddarthR56: The function returns a status message to let the caller know what happened. When there are no resources to delete, it's actually a successful operation , so we return that message with a nil error.
I see how "No resources to delete" sounds like something went wrong. Should i change it to something more neutral like "No resources found"?

internal/cli/delete.go
```diff
@@ -67,17 +70,34 @@ func (o *DeleteOptions) Validate(args []string) error {
...
+	if len(name) > 0 {
+		if len(args) > 1 {
+			return fmt.Errorf("invalid format: cannot mix TYPE/NAME syntax with additional resource names. Use either 'delete TYPE/NAME' or 'delete TYPE NAME [NAME...]'")
+		}
+		if errs := validation.ValidateResourceName(&name); len(errs) > 0 {
+			return fmt.Errorf("invalid resource name: %s", errors.Join(errs...).Error())
+		}
+	}
+	for _, resName := range args[1:] {
+		resNameCopy := resName
```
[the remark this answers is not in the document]
hexfusion: gotcha please see https://go.dev/blog/loopvar-preview this was resolved in core via go 1.22. I believe it covers this case.

internal/cli/delete.go
```diff
@@ -88,42 +107,69 @@ func (o *DeleteOptions) Run(ctx context.Context, args []string) error { //nolint
...
+		if processErr != nil {
+			fmt.Printf("Error: %v\n", processErr)
+			hasErrors = true
+		} else {
+			fmt.Printf("%s \"%s\" deleted\n", kind, name)
+		}
+	}
+
+	if hasErrors {
+		return fmt.Errorf("one or more deletions failed")
```
hexfusion, line 143: Since this is more or less syntactic sugar this would work. But under the hood should this be a single transaction that has either success or failure? My general concern is that for the user what delete failed exactly is missing. If we are not going to use a transaction why not show all errors?

SiddarthR56: Each failure is printed immediately, so users get specific error details in real-time. Looking at kubectl's source code, it doesn't return an error for individual deletion failures.
However, I didn't follow kubectl's approach of returning nil because this command might be used in scripts and automation where detecting partial failure is important. 

Should I follow kubectl's way (return nil, real-time errors only), or collect and print errors at the end instead of real-time, or leave it as is with the current approach?

hexfusion: Good response, as long as we are ensuring our ux matches kubectl then that's great. Thanks for digging in.

hexfusion (approved):
Looks good, nice work. Please ask QE for a tag as well.

PR #1187, opened 2025-05-20
EDM-1188:  deleting git mount path

noga-magen:
## Summary by CodeRabbit

- **New Features**
  - None.

- **Bug Fixes**
  - None.

- **Documentation**
  - Removed references to the `mountPath` parameter from user and developer documentation.
  - Updated configuration examples to reflect the removal of `mountPath`.

- **Refactor**
  - Removed the `mountPath` field from Git configuration options and related logic.
  - Simplified configuration by merging or updating path fields where necessary.
  - Adjusted internal logic and tests to eliminate all handling of `mountPath`.

- **Tests**
  - Updated test cases and test data to remove `mountPath` usage and reflect new configuration structure.


internal/tasks/fleet_rollout.go
```diff
@@ -378,14 +378,6 @@ func (f FleetRolloutsLogic) replaceGitConfigParameters(device *api.Device, confi
 		errs = append(errs, fmt.Errorf("failed replacing parameters in path in git config %s: %w", gitSpec.Name, err))
 	}
 
-	if gitSpec.GitRef.MountPath != nil {
```
celdrake, line 381: 1. The code for replacing the parameters was removed, but the [documentation](https://raw.githubusercontent.com/flightctl/flightctl/refs/heads/EDM-1188/docs/user/managing-devices.md) still refers to `mountPath`.
2. Some [example](https://github.com/flightctl/flightctl/blob/EDM-1188/examples/fleet-b.yaml#L22) still refer to it as well. 

The files I linked are just examples, please check if there are more places that refer to `mountPath`.

hexfusion: This would be a breaking change, but it is unclear to me why we need to change the API to fix the bug. @noga-magen can you elaborate? Or even better, can we bring this up in our next group call?

hexfusion (changes requested):
API changes need concensus if you have consensus please provide context.

amir-yogev-gh:
We are in a point where we should try to avoid breaking changes. If we must, we should also provide backward compatibility.

examples/device-standalone.yaml
```diff
@@ -12,8 +12,7 @@ spec:
       gitRef:
         repository: flightctl-demos
         targetRevision: main
-        path: /demos/basic-nginx-demo/configuration/etc/microshift/manifests
-        mountPath: /etc/microshift/manifests
+        path: /etc/microshift/manifests/demos/basic-nginx-demo/configuration/etc/microshift/manifests
```
siserafin: Is this path correct?  /etc/microshift/manifests is repeated twice

noga-magen: You can see the [way](https://github.com/flightctl/flightctl/commit/2672d90cb377eb0de3a376093c36f0cd34495cd0#diff-5b681129d485bc3c384fec32b6499817bb86278a9982d9cffcb713f19d966db1) the paths were configured before the MountPath addition

test/data/examples/device-standalone.yaml
```diff
@@ -12,8 +12,7 @@ spec:
       gitRef:
         repository: flightctl-demos
         targetRevision: main
-        path: /demos/basic-nginx-demo/configuration/etc/microshift/manifests
-        mountPath: /etc/microshift/manifests
+        path: /etc/microshift/manifests/demos/basic-nginx-demo/configuration/etc/microshift/manifests
```
celdrake: This example wouldn't work. Given the current demo project,it should be:

```suggestion
        path: /demos/basic-nginx-demo/configuration
```

https://github.com/flightctl/flightctl-demos/tree/main/demos/basic-nginx-demo/configuration

test/data/examples/fleet.yaml
```diff
@@ -19,8 +19,7 @@ spec:
           gitRef:
             repository: flightctl-demos
             targetRevision: main
-            path: /demos/basic-nginx-demo/configuration/etc/microshift/manifests
-            mountPath: /etc/microshift/manifests
+            path: /etc/microshift/manifests/demos/basic-nginx-demo/configuration/etc/microshift/manifests
```
celdrake: Same

```suggestion
            path: /demos/basic-nginx-demo/configuration
```

test/e2e/parametrisable_templates/parametrisable_templates_test.go
```diff
@@ -325,12 +322,11 @@ var (
 	contentWithFunction   = "{{ replace \"a\" \"c\" .metadata.labels.team }}"
 	pathWithFunction      = "/var/home/user/{{ upper .metadata.labels.team | lower }}/test.txt"
 	repoTestName          = "git-repo"
-	repoTestUrl           = "https://github.com/flightctl/flightctl-demos"
+	repoTestUrl           = "https://github.com/noga-magen/noga_tests"
```
celdrake: Should be back to the original
```suggestion
	repoTestUrl           = "https://github.com/flightctl/flightctl-demos"
```

test/e2e/parametrisable_templates/parametrisable_templates_test.go
```diff
@@ -325,12 +322,11 @@ var (
...
 	deviceAlias           = "base"
-	mountPath             = "/var/home/user/{{ .metadata.labels.team }}/file.txt"
 	branchTargetRevision  = "demo"
 	httpRepoName          = "http-repo"
-	gitRepoConfigPath     = "/{{ .metadata.labels.config }}/bootc/Containerfile.arm64"
+	gitRepoConfigPath     = "/testing"
```
celdrake: I'm not sure about this one, should it be back to the original?

`gitRepoConfigPath = "/{{ .metadata.labels.config }}/bootc/Containerfile.arm64"`

docs/user/managing-devices.md
```diff
@@ -301,7 +301,6 @@ The Git Config Provider takes the following parameters:
 | Repository | The name of a Repository resource defined in Flight Control. |
```
celdrake, line 301: I'm wondering if we should consider to explain in the docs that the mount path for the git config files replicates that of the directory that's pointed to. It may not be obvious for everyone.

docs/developer/enhancements/fep-000-api-device-fleet.md
```diff
@@ -126,8 +126,6 @@ The proposed changes include:
 
 * Some config items reference external configurations (`KubernetesSecretProviderSpec`, `GitConfigProviderSpec`)
 
-* `GitConfigProviderSpec` config types add a `mountPath` field for consistency, in the same way k8s secrets add such field.
-
```
hexfusion, line 130: if the fep was merged I dont know if we need to maintain that moving forward. @fzdarsky thoughts?

examples/device-standalone.yaml
```diff
@@ -12,8 +12,7 @@ spec:
       gitRef:
         repository: flightctl-demos
         targetRevision: main
-        path: /demos/basic-nginx-demo/configuration/etc/microshift/manifests
-        mountPath: /etc/microshift/manifests
+        path: /demos/basic-nginx-demo/configuration
```
hexfusion, line 15: I think we want a better example using different paths

test/e2e/parametrisable_templates/parametrisable_templates_test.go
```diff
@@ -327,10 +324,9 @@ var (
 	repoTestName          = "git-repo"
 	repoTestUrl           = "https://github.com/flightctl/flightctl-demos"
 	deviceAlias           = "base"
-	mountPath             = "/var/home/user/{{ .metadata.labels.team }}/file.txt"
 	branchTargetRevision  = "demo"
 	httpRepoName          = "http-repo"
-	gitRepoConfigPath     = "/{{ .metadata.labels.config }}/bootc/Containerfile.arm64"
+	gitRepoConfigPath     = "/demos/basic-nginx-demo/configuration"
```
hexfusion, line 329: is this correct?

docs/user/managing-devices.md
hexfusion: this should be resolved before merge.

noga-magen: Got i will do .

test/e2e/parametrisable_templates/parametrisable_templates_test.go
noga-magen: Yes

examples/device-standalone.yaml
noga-magen: What can a new path demonstrate that the current one doesn't?

hexfusion: From the user perspective `/etc/microshift/manifests` felt more intuitive for config gitops considering those paths will change in the demos repo.

noga-magen: So this needs to be merged today - how about we open another ticket or reopen this when the those paths will change in the demos repo?

test/e2e/parametrisable_templates/parametrisable_templates_test.go
hexfusion: I would expect the test to fail then. If not then we have another problem.

siserafin: yes, it looks correct, since at that path there is an /etc/ folder with a configuration inside, which is one of the folders already present in our vm device.

examples/device-standalone.yaml
siserafin: We have a qe task, to create a local git server for the E2E to avoid issues in case of changes in flightctl-demos.

PR #1204, opened 2025-05-28
NO-ISSUE: Update e2e running command

siserafin:
Updating the E2E running command, including the flag "--label-filter" only if GINKGO_LABEL_FILTER is not empty.


Samudelacruz77 (approved):
LGTM

PR #1206, opened 2025-05-28
EDM-1632: Upgrade go to RH-supported (and FIPS-compatible) version: 1.23.6

gshilin-sdb:
## Summary by CodeRabbit

- **Chores**
	- Updated Go version requirement from 1.22 to 1.23 across documentation, build workflows, container images, and configuration files.
	- Made a minor grammatical correction in developer documentation.
	- Adjusted the order of build targets in the build configuration.
	- Updated a comment reference to the Go standard library in Linux process handling.

### DIFF
--- a/test/scripts/run_e2e_tests.sh
+++ b/test/scripts/run_e2e_tests.sh
@@ -4,7 +4,7 @@ SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 source "${SCRIPT_DIR}"/functions
 
 REPORTS=${1}
-GO_E2E_DIRS=${@:2}
+GO_E2E_DIRS=("${@:2}")
 GINKGO_FOCUS=${GINKGO_FOCUS:-""}
 #Filtering e2e tests by labels
 GINKGO_LABEL_FILTER=${GINKGO_LABEL_FILTER:-""}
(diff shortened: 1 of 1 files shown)


.github/actions/setup-dependencies/action.yaml
```diff
@@ -12,7 +12,7 @@ runs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.22'
+          go-version: '1.23.6'
```
hexfusion, line 15: question: should we be using the redhat fips golang directly? Is it enough to use the same version upstream? I guess why should upstream  not continue to use latest version of 1.23?

gshilin-sdb: if this last attempt will not work will move to 1.23

hexfusion (approved):

.github/actions/setup-dependencies/action.yaml
gshilin-sdb: Current release is 1.24.0, so 1.22 has no official support (excluding critical security problems).

[Release Policy](https://go.dev/doc/devel/release#policy)
Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

PR #1209, opened 2025-05-30
NO-ISSUE: agent/console: isolate shell with bwrap and explicit bind mounts

hexfusion:
This PR is an experiment using bubblewrap to create an isolated shell for the console. This will allow us to gate some of the functionality based on user permissions and also be declarative in our permissions required for selinux. generally feels much more powerful than chroot and we get it for free via bootc/rpm-ostree.

ref. https://github.com/containers/bubblewrap


hexfusion:
cc @ori-amizur I think this might be a nice cheap way to provide isolation.

PR #1210, opened 2025-05-30
EDM-1613: Automation of refresh expired token

kkyrazis:
* Added certs and keys to the .gitignore that are generated via `make e2e-test` 
* Updated the keycloak deployment to allow sessions to persist across restarts
* Added a few comments to the docs with some issues I ran into.
* Added some testing utils to allow for manipulation of the running k8s cluster
* Refactored the login flow to enable logging in when `oc` doesn't exist.
* Added a CLI test to ensure token refreshing works. (Only runs when auth is enabled in the cluster)

## Summary by CodeRabbit

- **Documentation**
  - Added notes clarifying CLI login requirements and troubleshooting firewall connectivity in deployment documentation.

- **New Features**
  - Enhanced test harness with Kubernetes API access and improved network failure simulation tools.
  - Introduced utility functions for IP and URI handling in test utilities.

- **Bug Fixes**
  - Improved login test suite to cover token expiration and network failure scenarios.

- **Chores**
  - Updated .gitignore to exclude certificate and key files in test directories.
  - Added a new indirect dependency for improved code functionality.

- **Refactor**
  - Modularized and clarified login logic to support multiple authentication methods with better error handling.


docs/developer/README.md
```diff
@@ -57,10 +57,18 @@ bin/flightctl apply -f examples/fleet.yaml
 bin/flightctl get fleets
 ```
 
+Note: If deployed without auth enabled then there is no need to login.
```
hexfusion, line 60: :)

test/harness/e2e/harness.go
```diff
@@ -16,7 +16,7 @@ import (
 
 	"github.com/flightctl/flightctl/api/v1alpha1"
 	apiclient "github.com/flightctl/flightctl/internal/api/client"
-	client "github.com/flightctl/flightctl/internal/client"
```
hexfusion, line 19: :+1:

hexfusion (commented):
looking good so far!

test/e2e/cli/cli_test.go
```diff
@@ -552,6 +497,132 @@ var _ = Describe("cli login", func() {
...
+		It("Validations work when logging into flightctl CLI", Label("78748", "sanity"), func() {
+			By("Prepare invalid API endpoint")
+			invalidEndpoint := "https://not-existing.lab.redhat.com"
+			loginArgs := []string{"login", invalidEndpoint}
+
+			By("Try login using a wrong API endpoint without --insecure-skip-tls-verify flag")
+			logrus.Infof("Executing CLI with args: %v", loginArgs)
+			out, err := harness.CLI(loginArgs...)
+			Expect(err).To(HaveOccurred())
+			Expect(out).To(ContainSubstring("failed to get auth info"))
```
eldar101, line 577: If a string repeats more than once in our code, put in a var please.

test/login/login.go
```diff
@@ -1,53 +1,130 @@
...
+	// AuthUsernamePassword indicates authentication using a username and password.
+	AuthUsernamePassword
+)
+
+const (
+	openshift = "oc"
+)
+
+func baseLoginArgs() []string {
+	return []string{"login", "${API_ENDPOINT}", "--insecure-skip-tls-verify"}
```
eldar101, line 34: ditto about repeated strings.

test/login/login.go
```diff
@@ -1,53 +1,130 @@
...
+}
+
+// LoginToAPIWithToken attempts to log in to the flightctl API via different methods depending on what is available.
+// If auth is disabled then this is largely a noop. This method panics if login fails
+func LoginToAPIWithToken(harness *e2e.Harness) AuthMethod {
```
eldar101, line 39: don't we have something similar? @siserafin

test/e2e/cli/cli_test.go
```diff
@@ -552,6 +497,132 @@ var _ = Describe("cli login", func() {
...
+	Context("login validation", func() {
+		BeforeEach(func() {
+			harness = e2e.NewTestHarness()
+			clusterManager = util.NewClusterDeploymentManager(harness.Context, harness.Cluster, "")
+		})
+		AfterEach(func() {
+			harness.Cleanup(false) // do not print console on error
+		})
+
+		It("Validations work when logging into flightctl CLI", Label("78748", "sanity"), func() {
```
amir-yogev-gh, line 568: What's the added value in grouping all these test cases into a single flow?
How the third test case is going to be run if the second test case fails?
IMO, it will be better to break these. You can consider using DescribeTable to feed identical flow with different args.

test/e2e/cli/cli_test.go
```diff
@@ -552,6 +497,132 @@ var _ = Describe("cli login", func() {
...
+				logrus.Infof("Executing CLI with args: %v", loginArgsPassword)
+				out, _ = harness.CLI(loginArgsPassword...)
+				// We don't check for error here as we're only interested in the output message
+				Expect(out).To(Or(
+					ContainSubstring("Invalid user credentials"),
+					ContainSubstring("unexpected http code: 401")))
+			}
+		})
+
+		It("Should refresh token when expiration is reached", Label("GET A LABEL"), func() {
```
amir-yogev-gh: See my previous comment. Applies also here.

amir-yogev-gh (changes requested):

deploy/helm/flightctl/charts/keycloak/templates/keycloak-deployment.yaml
```diff
@@ -26,6 +26,7 @@ spec:
         - --http-port=8081
         - --proxy=edge
         - --import-realm
+        - --features=persistent-user-sessions
```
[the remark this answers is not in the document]
siserafin: I agree with this comment, we don’t need to modify or bring down the Keycloak deployment, since our goal is simply to make it temporarily unavailable. We already have harness functions like harness.SimulateNetworkFailure() and harness.FixNetworkFailure() that can simulate network disruptions. Please check if you can achieve the same result using these instead.

test/harness/e2e/harness.go
```diff
@@ -60,6 +63,30 @@ func findTopLevelDir() string {
 	return ""
 }
 
+func kubernetesClient() (kubernetes.Interface, error) {
+	var kubeconfig string
+	if kc, ok := os.LookupEnv("KUBECONFIG"); ok && kc != "" {
```
siserafin, line 69: Please test this function in an ocp env (we can provide you one). I believe that with our current implementation this is going to fail, since in our ci hosts we are storing the kubeconfig at path /home/kni/clusterconfigs/kubeconfig, and we are not exporting the KUBECONFIG variable, so the flow might need to be updated.

test/login/login.go
```diff
@@ -1,53 +1,130 @@
...
+		namespaces = append(namespaces, strings.ToLower(strings.TrimSpace(item.Name)))
+	}
+	return namespaces
+}
+
+func loginWithPassword(harness *e2e.Harness) (AuthMethod, error) {
+	namespaces := getActiveNamespaces(harness)
+	namespace := ""
+	// is this a legacy thing?
+	if slices.Contains(namespaces, "flightctl") {
```
siserafin: We can't assume only flightctl or flightctl-external, In some cases we use different namespaces, (When we are testing enabling flightctl tech preview we can have ocm or rhacm). But there is an env variable FLIGHTCTL_NS that can be checked.

test/login/login.go
siserafin: Yes, he's making modifications to the existing implementation.

siserafin, line 39: Please test this updated function in ocp+ACM+flightctl and ocp+standalone  flightctl deployments.

test/e2e/cli/cli_test.go
siserafin: In this case, since it’s an existing test already linked to a specific Polarion id, I’d suggest keeping its current structure as it is. Using DescribeTable would change how individual entries are reported (each would show up as a separate test) which would modify the current behavior.

test/e2e/cli/cli_test.go
siserafin: Polarion id for this test case is 81481

test/e2e/cli/cli_test.go
```diff
@@ -552,6 +497,132 @@ var _ = Describe("cli login", func() {
...
+			const (
+				authDeployment = "keycloak"
+				configPath     = "" // default to nothing to let the default path be used
+			)
+			By("Login to the service")
+			// We want to log in directly with user/pass on this flow so that we can force the auth mechanism
+			// to occur with our keycloak deployment
+			if login.WithPassword(harness) == login.AuthDisabled {
```
siserafin: From what I understand, this feature isn’t supported in ocp. I believe that in the ocp + standalone setup, the test wouldn’t be skipped since login using a username and password would succeed.

test/e2e/cli/cli_test.go
```diff
@@ -552,6 +497,132 @@ var _ = Describe("cli login", func() {
...
+			Expect(err).ToNot(HaveOccurred(), "Failed to get device info after expiring the token")
+
+			By("Ensure a new token was generated")
+			// Note: The old token is still valid at this point as it hasn't actually expired
+			cfg, err = harness.ReadClientConfig(configPath)
+			Expect(err).ToNot(HaveOccurred(), "Failed to read client config after expiring token")
+			secondToken := cfg.AuthInfo.Token
+			Expect(secondToken).ToNot(Equal(initialToken), "Token should have been refreshed")
+
+			By("Bring down the auth service")
```
siserafin: Please, test by simulating a network interruption using harness.SimulateNetworkFailure() and harness.FixNetworkFailure(). (see my previous comment)

siserafin (changes requested):
Thanks for helping with automating this test case! Please also test the changes, especially those in login.go, by running a regression in both the ocp+ACM and ocp + standalone environments.

test/e2e/cli/cli_test.go
siserafin: Please, see my comment. This should be reported as a single e2e test.

test/e2e/cli/cli_test.go
Samudelacruz77: This is an E2E case, so the flow is unique, in Polarion we define steps for testcase, so this is like this, one testcase with multiple steps that should be run in a certain order

kkyrazis: There was already a `Describe("cli login")` block. The `It("Validations work when logging into flight CLI"` tests also already existed. Previously the `Events API Tests` context was inside the `cli login` `Describe` which seemed off to me so I just separated this out and added only the tests inside the 'GET A LABEL'. I can move it back if you'd like 


Before:
```
Describe("cli operation") ...
Describe("cli login") 
                                Context("login validation")
                                Context("Events API Tests")
```

After:
```
Describe("cli operation")
                                Context("Events API Tests")
Describe("cli login")
                                Context("login validation")
```

Github is unfortunately making it look like I did more to this than I did

test/e2e/cli/cli_test.go
kkyrazis: Thank you

test/login/login.go
kkyrazis: Of course.

test/harness/e2e/harness.go
kkyrazis: If the `NetworkFailure` functions work I'll just remove all of this.

test/e2e/cli/cli_test.go
kkyrazis: The `SimulateNetworkFailure` and `FixNetworkFailure` methods only appear to set tables on Agent VMs (of which this test has none). Regardless, I'll make some changes so that I can utilize this method rather than bringing the deployment down.

test/util/ip.go
```diff
@@ -44,3 +47,69 @@ func GetExtIP() string {
 	Expect(err).ToNot(HaveOccurred())
 	return strings.TrimSpace(string(output))
 }
+
+func BinaryExistsOnPath(binaryName string) bool {
+	_, err := exec.LookPath(binaryName)
+	return err == nil
+}
+
+func resolveHostNameIP4Address(hostname string) (string, error) {
```
eldar101: don't you think it should be a generic rseolve host name? for both ipv4 and ipv6?

test/e2e/cli/cli_test.go
kkyrazis: Adjusted by calling the original login method and ensuring that the result is `UsernamePassword`

test/util/ip.go
kkyrazis: Sure, I'll update this so that it prefers ip4, but will pick ip6 if there are no ip4 addresses.

test/harness/e2e/harness.go
kkyrazis: I ended up keeping this and just adding a few extra well known kubeconfig locations.

test/e2e/cli/cli_test.go
kkyrazis: Github is showing this as me having made changes, but really I just moved the events tests out of Login tests. I can make changes to this test if you would like, but I was trying to avoid making changes to existing tests

test/login/login.go
kkyrazis: Ok I have verified that both of these login mechanisms work when deployed with ocp. And in the case of k8s auth being used, the test correctly skips running:
```
cli login
/home/kni/flightctl/test/e2e/cli/cli_test.go:501
  login validation
  /home/kni/flightctl/test/e2e/cli/cli_test.go:504
    Should refresh token when expiration is reached [81481]
    <elided>
time="2025-06-04T11:15:34-04:00" level=info msg="running: oc whoami -t"
time="2025-06-04T11:15:34-04:00" level=info msg="running: /home/kni/flightctl/bin/flightctl login https://api.flightctl.apps.ocp-edge-cluster-0.qe.lab.redhat.com --insecure-skip-tls-verify --token <who ami token>"
  [SKIPPED] This test requires authentication to be enabled
```

I've updated the skipped message since this run, but this test behaves correctly in OIDC, k8s auth, or no auth mode.

test/harness/e2e/harness.go
kkyrazis: All cases should be covered now. 
1. prefer the KUBCONFIG env variable
2. check ~/.kube/config
3. check the QA locations

It should be noted that within the `make deploy-e2e-ocp-test-vm` VM, that once you `oc login` a `~/.kube/config` file is created for you so that case should be covered to.

test/e2e/cli/cli_test.go
kkyrazis: I don't think this test actually benefits from a Table. It's one test case that builds off the previous steps. If any step fails the next action should not be attempted.

test/e2e/cli/cli_test.go
kkyrazis: While I agree that this `login validation` test could be a table, I think that the QA team would prefer that I don't modify the structure of this test.

test/login/login.go
```diff
@@ -137,7 +137,9 @@ func loginWithOpenshift(harness *e2e.Harness) (AuthMethod, error) {
 	Expect(token).ToNot(BeEmpty(), "Token from 'oc whoami' should not be empty")
 	loginArgsOcp := append(baseLoginArgs(), "--token", token)
 	out, err := harness.CLI(loginArgsOcp...)
-	Expect(err).ToNot(HaveOccurred())
+	if err != nil {
+		return AuthDisabled, fmt.Errorf("failed to sign in with OpenShift token: %w", err)
```
hexfusion: if this returns an error why do we also set the string, my expectation is that on error the string is never read. if we break that pattern it adds complexity long term.

kkyrazis: I'm not sure I fully understand what you mean. Why is the `out` variable set if an `err` occurs? Or are you talking about the `Auth[Disabled|Token|UsernamePassword]` enum?

kkyrazis: I ended up simplifying this block by only returning an error to indicate that token login failed. The caller will return the authmethod if there is no error

test/login/login.go
```diff
@@ -1,53 +1,147 @@
...
-	Expect(err).ToNot(HaveOccurred())
-	Expect(strings.TrimSpace(out)).To(BeElementOf("Auth is disabled", "Login successful", "Login successful."))
+	if err != nil {
+		return AuthDisabled, fmt.Errorf("error executing login: %w", err)
+	}
+	if isLoginSuccessful(out) {
+		return AuthUsernamePassword, nil
+	}
+	return AuthDisabled, errors.New("failed to sign in with user name and password")
+}
```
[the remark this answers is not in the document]
kkyrazis: These are tests with secrets that are regenerated whenever a cluster is re-deployed specifically for test users. The information is not actually sensitive.

amalykhi (approved):

PR #1211, opened 2025-06-03
NO-ISSUE: Specify Go 1.23 toolchain for hermetic builds

asafbss:
Add `toolchain go1.23.6` to go.mod to ensure hermetic build environments (e.g., Konflux, CPaaS) use the correct Go version explicitly.

## Summary by CodeRabbit

- **Chores**
  - Updated the Go toolchain version to 1.23.6.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/flightctl/flightctl
 
 go 1.23
 
+toolchain go1.23.6
+
 require (
 	github.com/ccoveille/go-safecast v1.1.0
 	github.com/containers/image/v5 v5.30.1


celdrake:
@asafbss Is this same change necessary in the UI project?

gshilin-sdb:
> @asafbss Is this same change necessary in the UI project?

[Release Policy](https://go.dev/doc/devel/release#policy)
Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

We have to be FIPS-compatible, so yes...

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/flightctl/flightctl
 
 go 1.23
 
+toolchain go1.23.0
```
hexfusion: My understanding of the docs [1] is that this explicitly requires/pulls 1.23.0 in the case that the local toolchain is less than 1.23.0. But I wonder is that what we want as that version potentially not secure etc.

```
gvm use go1.21.8 --default
Now using version go1.21.8
$ make build
go: downloading go1.23.0 (linux/amd64)
[ build noise here ]

$ go version -m bin/flightctl-agent 
bin/flightctl-agent: go1.23.0
```

alternatively without this declaration the build could do 

```
GOTOOLCHAIN=go1.23.9 make build
go: downloading go1.23.9 (linux/amd64)
....
```
[1] https://go.dev/doc/toolchain

hexfusion: I will also admit I read the docs 3 times and just had to try it as I was more confused each time. :)

asafbss: You're right, Sam — it's better to use a more recent patch version. I'll update from toolchain go1.23.0 to go1.23.9 to avoid relying on an older, potentially less secure version.

Using a fixed, up-to-date version ensures more predictable and stable builds, and helps reduce the risk of unexpected behavior on systems that require strict toolchain consistency, WDYT?

asafbss: @gshilin-sdb @celdrake FYI — we might update to toolchain go1.23.9.

@gshilin-sdb you mentioned something earlier about CentOS only supporting up to Go 1.23.4 — is that still relevant in our case?

hexfusion: Can we declare the ENV where we need it instead of declaring this as all in go.mod? `GOTOOLCHAIN=` overrides the value in go.mod.

asafbss: It’s explicitly required by the downstream build systems — they scan the go.mod file to detect the toolchain version in order to cache dependencies and ensure a hermetic build.

asafbss: Even if it can technically be set elsewhere, relying on external configuration for this isn't a good practice — it's better to declare it in go.mod so the version is always consistent and transparent across environments.

hexfusion: feels like a downsteam only problem is the awkward part. but I don't want to block the release

asafbss: Totally understandable.
I've seen toolchain recommended as a best practice for reproducible builds across different environments — not just downstream. It helps ensure consistency and avoid surprises when others build the project.

Still waiting to hear back from @gshilin-sdb regarding the CentOS part.

asafbss: @hexfusion just made GitHub search and found a known repo for an example.
https://github.com/rancher/rancher/blob/main/go.mod

asafbss: https://github.com/etcd-io/raft/blob/main/go.mod

fzdarsky: At this point in time, the latest patch version of the upstream Go toolchain is 1.23.9, same as the latest downstream `registry.access.redhat.com/ubi9/go-toolset` toolchain. However, latest downstream may trail upstream and we're not even using the latest but pin `go-toolset:1.23.6-1747333074`.

How do we ensure that we use exactly the pinned downstream version and do not blend in code from upstream?

asafbss: Hey @fzdarsky ,
Thanks for pointing this out — I was rushing and missed it (sorry @hexfusion ❤️)

I tested this directly, and it turns out the Red Hat go-toolset image defaults to GOTOOLCHAIN=local, which disables Go’s automatic toolchain switching. So even if we declare toolchain go1.23.9 in go.mod, it doesn’t fetch or use that version unless GOTOOLCHAIN=auto is explicitly set.

```shell
[1/3] STEP 18/19: RUN go env GOTOOLCHAIN
local
[1/3] STEP 19/19: RUN make build-api
mkdir -p bin
GOOS=linux GOARCH=amd64 go build -buildvcs=false -ldflags " -X github.com/flightctl/flightctl/pkg/version.majorFromGit=0 -X github.com/flightctl/flightctl/pkg/version.minorFromGit=8 -X github.com/flightctl/flightctl/pkg/version.patchFromGit=0 -X github.com/flightctl/flightctl/pkg/version.versionFromGit=v0.8.0-main-82-ge4d4e6d3 -X github.com/flightctl/flightctl/pkg/version.commitFromGit=e4d4e6d3 -X github.com/flightctl/flightctl/pkg/version.gitTreeState=dirty -X github.com/flightctl/flightctl/pkg/version.buildDate=20250605 " -o /app/bin/ ./cmd/flightctl-api
go: downloading github.com/sirupsen/logrus v1.9.3
``` 

```shell
[1/3] STEP 18/20: ENV GOTOOLCHAIN=auto
[1/3] STEP 19/20: RUN go env GOTOOLCHAIN
auto
[1/3] STEP 20/20: RUN make build-api
go: downloading go1.23.9 (linux/amd64)
``` 

Given that, I suggest we align the go.mod toolchain to go1.23.6, to match the version bundled in go-toolset, especially for consistency in non-container builds.

gshilin-sdb: @asafbss everything is ok as far as centos rpm is built (and it is)

celdrake (approved):

fzdarsky (approved):

go.mod
hexfusion: Just going to point out that upstream and downstream golang patch versions **should** diverge here. While downstream may pin X.Y.Z they will potentially backport CVE security fixes into the patch version. This is not the same for upstream.

asafbss: Completely understandable. Just to clarify — this only applies to local builds (i.e., non-container builds). In that case, the end user can see the Go version we recommend via the toolchain line and decide whether to use it or stick with their own local version.

We’re not assuming they’ll use Red Hat’s version — only giving a precise and reproducible default. The alternative is leaving it vague with go 1.23, which gives them no clear guidance.

PR #1213, opened 2025-06-03
EDM-1565: Repository not removed

SiddarthR56:
Issue: Repository deletion appears to succeed (HTTP 200) but leaves the repository in an inconsistent state when dependencies exist. The current deletion logic  leaves behind metadata records that maintain referential integrity for dependent resources like Fleets. When Fleets reference a repository via gitRef.repository, the system keeps these metadata records alive, causing subsequent recreation attempts to fail with HTTP 409 Conflict errors. 

Root Cause: The createOrUpdate method has inconsistent logic:
- Business Logic: Correctly identifies nil-spec resources as "creating" operations: 
      creating := !exists || P(existing).HasNilSpec()  
- Mode Validation: Ignores the creating flag and only checks raw existence:   
      if exists && mode == ModeCreateOnly {
         return ErrDuplicateName 
      }

Solution: Align mode validation with the creating flag 

Impact:
Breaking Change: This fix affects ALL resource types since it's implemented in GenericStore. Any POST request to a resource with nil spec will now succeed and recreate the resource, rather than returning 409 conflict.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for create-only and update-only operations to provide more accurate feedback when creating or updating resources.

### DIFF
--- a/internal/store/generic.go
+++ b/internal/store/generic.go
@@ -257,7 +257,7 @@ func (s *GenericStore[P, M, A, AL]) Get(ctx context.Context, orgId uuid.UUID, na
 	return apiResource, err
 }
 
-func (s *GenericStore[P, M, A, AL]) Delete(ctx context.Context, resource M, callback func(ctx context.Context, orgId uuid.UUID, before, after *A), associatedResources ...Resource) error {
+func (s *GenericStore[P, M, A, AL]) Delete(ctx context.Context, resource M, callback func(ctx context.Context, orgId uuid.UUID, before, after *A), associatedResources ...Resource) (bool, error) {
 	var deleted bool
 	var err error
 	if len(associatedResources) == 0 {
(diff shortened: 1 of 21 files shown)


avishayt:
Thanks for the work here. I think we should take a step back and consider the declarative model we're following in Flight Control.

Right now, it's possible to create a fleet that references a non-existent repository — and that's totally fine; the system surfaces the issue via the resource's status. In that spirit, I don't think it's consistent for a repository delete to be blocked just because something references it. Declarative systems should let users express their desired state (e.g., "this repository should be gone") even if it temporarily causes inconsistencies elsewhere.

So instead of blocking deletion, we should fix the create behavior so it doesn't error if a stale mapping exists (i.e., treat the mapping as soft or reconstruct it as needed)

That would align better with the model we're using elsewhere. Happy to discuss further if needed.

celdrake:
> So instead of blocking deletion, we should fix the create behavior so it doesn't error if a stale mapping exists (i.e., treat the mapping as soft or reconstruct it as needed)
> 
I was looking at the PR and I agree with the suggested approach.

internal/store/generic.go
```diff
@@ -128,10 +128,11 @@ func (s *GenericStore[P, M, A, AL]) createOrUpdate(ctx context.Context, orgId uu
 		}
 	}
 
-	if exists && mode == ModeCreateOnly {
+	if !creating && mode == ModeCreateOnly {
```
SiddarthR56, line 135: When a repository has nil spec, it's effectively like it doesn't exist - you can create it but you can't update it. The fix makes sure the validation logic treats these recreatable repositories the same way it treats completely missing ones.

avishayt (changes requested):
The code looks great.  Please just clean up the commits (1 commit in this case with a clear message).

PR #1218, opened 2025-06-04
EDM-1581: A flightctl console session attempt during a reboot cannot be stopped with ctrl+c

ori-amizur:
Move terminal to raw only after connection has been established which means there is direct connection with the device.


avishayt (approved):

ori-amizur:
> maybe put the time in a variable so it will be easily modified late. Are you sure this is the least amount of time needeD?

What is the benefit?

PR #1222, opened 2025-06-06
EDM-1543: agent/spec clarify rollback logic

hexfusion:
While reviewing another bug I came
Across this and wanted to clarify some of the rollback logic with improved testing and comments. This PR does not result in logical change. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rollback behavior to ensure the desired state is accurately updated from the current state, both on disk and in memory.

- **Tests**
  - Enhanced rollback test coverage with multiple scenarios, realistic file operations, and additional validation of in-memory and on-disk states.

### DIFF
--- a/internal/agent/device/spec/manager.go
+++ b/internal/agent/device/spec/manager.go
@@ -260,26 +260,24 @@ func (s *manager) Rollback(ctx context.Context, opts ...RollbackOption) error {
 			return err
 		}
 		// set the desired spec as failed
-
 		s.queue.SetFailed(version)
 	}
 
-	// copy the current rendered spec to the desired rendered spec
-	// this will reconcile the device with the desired "rollback" state
+	// rollback on disk state current == desired
 	if err := s.deviceReadWriter.CopyFile(s.currentPath, s.desiredPath); err != nil {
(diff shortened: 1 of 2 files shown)


internal/agent/device/spec/manager.go
```diff
@@ -260,26 +260,24 @@ func (s *manager) Rollback(ctx context.Context, opts ...RollbackOption) error {
...
-	// copy the current rendered spec to the desired rendered spec
-	// this will reconcile the device with the desired "rollback" state
+	// rollback on disk state current -> desired
 	if err := s.deviceReadWriter.CopyFile(s.currentPath, s.desiredPath); err != nil {
 		return fmt.Errorf("%w: copy %q to %q: %w", errors.ErrCopySpec, s.currentPath, s.desiredPath, err)
 	}
 
-	// update cache
-	desired, err := s.Read(Desired)
+	current, err := s.Read(Current)
```
hexfusion, line 271: note: this isn’t technically a logical change, since after the rollback the desired and current states are the same. but it could be confusing to read. lower in the method, we explicitly update the cached desired version, which was the original intent behind the read. this is because the cache is updated on each read or write to disk.

SiddarthR56 (approved):

PR #1232, opened 2025-06-09
EDM-1365: agent/spec: skip beforeUpdate validation if IsRollback

hexfusion:


internal/agent/device/device_test.go
```diff
@@ -74,47 +76,72 @@ func TestSync(t *testing.T) {
...
+					mockSpecManager.EXPECT().IsUpgrading().Return(true),
 					mockManagementClient.EXPECT().UpdateDeviceStatus(ctx, deviceName, gomock.Any()).Return(nil),
+					mockSpecManager.EXPECT().IsUpgrading().Return(true),
 					mockManagementClient.EXPECT().UpdateDeviceStatus(ctx, deviceName, gomock.Any()).Return(nil),
 					mockHookManager.EXPECT().Sync(current.Spec, desired.Spec).Return(nil),
 					mockResourceManager.EXPECT().ResetAlertDefaults().Return(nil),
 					mockSystemdManager.EXPECT().EnsurePatterns(gomock.Any()).Return(nil),
 					mockLifecycleManager.EXPECT().Sync(ctx, current.Spec, desired.Spec).Return(nil),
 					mockLifecycleManager.EXPECT().AfterUpdate(ctx, current.Spec, desired.Spec).Return(nil),
-					mockOSClient.EXPECT().Status(ctx).Return(&os.Status{}, nil),
```
hexfusion, line 90: Note: The status was called by the actual spec Manager, and since we are not using the mock, it is not visible directly.

internal/agent/device/device_test.go
```diff
@@ -74,47 +76,72 @@ func TestSync(t *testing.T) {
...
+				mockPolicyManager *policy.MockManager,
+				mockSpecManager *spec.MockManager,
 			) {
 				nonRetryableHookError := errors.New("hook error")
 				gomock.InOrder(
+					mockSpecManager.EXPECT().GetDesired(ctx).Return(desired, false, nil),
+					mockSpecManager.EXPECT().Read(spec.Current).Return(current, nil),
+					mockPolicyManager.EXPECT().Sync(ctx, desired.Spec).Return(nil),
+					mockSpecManager.EXPECT().CheckPolicy(ctx, policy.Download, desired.Version()).Return(nil),
+					mockSpecManager.EXPECT().IsUpgrading().Return(true),
```
hexfusion, line 88: note: we were using an actual spec manager before which didn't provide the same level of observability to the flow.

internal/agent/device/spec/queue.go
```diff
@@ -129,6 +129,12 @@ func (m *queueManager) Next(ctx context.Context) (*v1alpha1.Device, bool) {
 		m.log.Debugf("Template version policy updated: %d", item.Version)
 	}
 
+	if !requeue.downloadPolicySatisfied && !requeue.updatePolicySatisfied {
+		m.queue.Add(item)
+		m.log.Debugf("Template version %d policies are not satisfied skipping...", item.Version)
+		return nil, false
+	}
```
hexfusion, line 134: note: this is an optimization; before, we would pop this then fail on policy check and rollback which is unnecessary.

internal/agent/device/device.go
```diff
@@ -133,51 +133,9 @@ func (a *Agent) Run(ctx context.Context) error {
...
-		updateErr := a.statusManager.UpdateCondition(ctx, v1alpha1.Condition{
-			Type:    v1alpha1.DeviceUpdating,
-			Status:  v1alpha1.ConditionStatusTrue,
-			Reason:  string(v1alpha1.UpdateStatePreparing),
-			Message: fmt.Sprintf("The device is preparing an update to renderedVersion: %s", desired.Version()),
-		})
-		if updateErr != nil {
-			a.log.Warnf("Failed setting status: %v", updateErr)
-		}
-	}
```
hexfusion, line 159: note: this was all moved into `beforeUpdate`

internal/agent/device/device_test.go
```diff
@@ -169,23 +198,12 @@ func TestSync(t *testing.T) {
...
-				dataDir,
-				policyManager,
-				readWriter,
-				mockOSClient,
-				devicePublisher,
-				log,
-			)
-
-			err := specManager.Initialize(ctx)
-			require.NoError(err)
```
hexfusion, line 182: note: this was replaced by the mock

internal/agent/device/spec/queue.go
```diff
@@ -130,6 +130,12 @@ func (m *queueManager) Next(ctx context.Context) (*v1alpha1.Device, bool) {
 	}
 
 	if exists {
+		if !requeue.downloadPolicySatisfied && !requeue.updatePolicySatisfied {
```
kkyrazis: Semi-related, but why is it safe to call `m.updatePolicy` above without considering the result of `exists`?

hexfusion: Yep good call its not, I already have `getOrCreateRequeueState` so why not use it as well.

internal/agent/device/spec/queue.go
```diff
@@ -116,29 +116,31 @@ func (m *queueManager) Next(ctx context.Context) (*v1alpha1.Device, bool) {
...
-	requeue, exists := m.requeueLookup[item.Version]
-	if exists && now.Before(requeue.nextAvailable) {
+	requeue := m.getOrCreateRequeueState(ctx, version)
+	if now.Before(requeue.nextAvailable) {
 		m.queue.Add(item)
-		m.log.Debugf("Template version %d requeue is currently in backoff. Available after: %s", item.Version, requeue.nextAvailable.Format(time.RFC3339))
+		m.log.Debugf("Template version %d requeue is currently in backoff. Available after: %s", version, requeue.nextAvailable.Format(time.RFC3339))
 		return nil, false
 	}
-	if m.updatePolicy(ctx, requeue) {
```
hexfusion, line 128: note: this is handled in getOrCreateRequeueState

kkyrazis (approved):

internal/agent/device/spec/queue.go
hexfusion: this was actually a bad idea we need the regular flow for proper status reporting and ensuring new template versions properly update the policy manager.

PR #1235, opened 2025-06-10
EDM-1058: Implement LDevID key for device's TPM hardware identity

lkatalin:
This PR creates LDevID keys that can be used as the device's TPM-based identity. These keys are created in the TPM's Storage Hierarchy as recommended in Section 6 of [TPM 2.0 Keys for Device Identity and Attestation](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM-2p0-DevID_v1p00_r10_12july2021.pdf). The key attributes are based on Section 7 of the same document.

## Summary by CodeRabbit

- **New Features**
	- Added support for generating a Storage Root Key (SRK) and creating an LDevID key pair in alignment with TPM 2.0 Device Identity standards.
- **Bug Fixes**
	- Updated TPM device path to improve compatibility with modern systems.
- **Refactor**
	- Streamlined TPM test setup by grouping related test data into a single struct and standardizing usage of legacy TPM2 types.

### DIFF
--- a/internal/tpm/templates.go
+++ b/internal/tpm/templates.go
@@ -1,0 +1,49 @@
+package tpm
+
+import (
+	"github.com/google/go-tpm/tpm2"
+)
+
+// This key template uses the Storage Root Key as the parent key.
+// Other key attributes are aligned with definitions from https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM-2p0-DevID_v1p00_r10_12july2021.pdf. Specifically, for key attribute and parameter recommendations, see Sections 7.3.4.1 and 7.3.4.3.
+var (
+	LDevIDTemplate = tpm2.TPMTPublic{
+		Type:    tpm2.TPMAlgECC,
+		NameAlg: tpm2.TPMAlgSHA256,
+		ObjectAttributes: tpm2.TPMAObject{
+			FixedTPM:             true,  //true = must stay in TPM
+			STClear:              false, //true = cannot be loaded after tpm2_clear
+			FixedParent:          true,  //true = can't be re-parented
+			SensitiveDataOrigin:  true,  //true = TPM generates all sensitive data during creation
+			UserWithAuth:         true,  //true = pw or hmac can be used in addition to authpolicy
+			AdminWithPolicy:      true,  //true = authValue cannot be used for auth
+			NoDA:                 false, //true = there are dictionary attack protections
+			EncryptedDuplication: false, //true = there are more robust protections for duplication
+			Restricted:           false, //true = cannot be used to sign data from outside tpm
+			Decrypt:              false, //true = can be used to decrypt
+			SignEncrypt:          true,  //true = for asymm, may be used to sign
+		},
+		Parameters: tpm2.NewTPMUPublicParms(
+			tpm2.TPMAlgECC,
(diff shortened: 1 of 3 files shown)


internal/tpm/tpm.go
```diff
@@ -117,7 +135,31 @@ func (t *TPM) GetAttestation(nonce []byte, ak *client.Key) (*pbattest.Attestatio
 	return ak.Attest(client.AttestOpts{Nonce: nonce})
 }
 
-func (t *TPM) GetQuote(nonce []byte, ak *client.Key, pcr_selection *tpm2.PCRSelection) (*pbtpm.Quote, error) {
+// This function creates an ECC LDevID key pair under the Storage/Owner hierarchy with the Storage Root Key as parent.
+func (t *TPM) CreateLDevID(srk tpm2.CreatePrimaryResponse) (*tpm2.TPMHandle, error) {
```
fzdarsky: Instead of passing in a `tpm2.CreatePrimaryResponse`, I'd pass in a `tpm2.NamedHandle`, as this makes the function more generic (and readable).

internal/tpm/tpm.go
```diff
@@ -91,6 +93,22 @@ func (t *TPM) GetPCRValues(measurements map[string]string) error {
 	return nil
 }
 
+// This function (re-)creates an ECC Primary Storage Root Key in the Owner/Storage Hierarchy.
+// This key is deterministically generated from the Storage Primary Seed + input parameters.
+func (t *TPM) GenerateSRKPrimary() (*tpm2.CreatePrimaryResponse, error) {
```
fzdarsky: Would the Primary SRK already exist on some systems? If so, would it be acceptable (or even allowed) to overwrite it like this? Wondering whether we should use an existing Primary SRK if that already exists instead?

internal/tpm/templates.go
```diff
@@ -0,0 +1,73 @@
...
+
+// This key template uses the Storage Root Key as the parent key.
+// Other key attributes are aligned with definitions from https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM-2p0-DevID_v1p00_r10_12july2021.pdf. Specifically, for key attribute and parameter recommendations, see Sections 7.3.4.1 and 7.3.4.3.
+func CreateOwnerLDevIDTemplate(srk tpm2.CreatePrimaryResponse) tpm2.Create {
+	return tpm2.Create{
+		ParentHandle: tpm2.NamedHandle{
+			Handle: srk.ObjectHandle,
+			Name:   srk.Name,
+		},
+		InPublic: tpm2.New2B(tpm2.TPMTPublic{
```
fzdarsky: Instead of defining a new function, how about defining the LDevIDTemplates as a global (like it's done in the tpm2 lib for the SRK)? This would also make the calling function more similar to the GenerateSRKPrimary() funciton (improves readability).

internal/tpm/tpm.go
lkatalin: Lily todo -  investigate whether SRK with RSA template / SRK with ECC template would overwrite each other. "Ensure" SRK for example

lkatalin: I'll have to do this as follow-up

fzdarsky (approved):

PR #1242, opened 2025-06-12
EDM-1669: Agent: device publisher does not facilitate retry

ori-amizur:
In EDM-1447, a publishing mechanism for new device spec has been added to support multiple parralel consumers.  In this change, a bug was introduced causing retry to be unavailable for failed rendered versions. This change will cause the failed device spec to be consumed again to enable retry.

## Summary by CodeRabbit

- **New Features**
	- Improved handling of device specifications to automatically reprocess the last known device in rollback or version mismatch scenarios, ensuring device state consistency.

### DIFF
--- a/internal/agent/device/spec/manager.go
+++ b/internal/agent/device/spec/manager.go
@@ -30,7 +30,8 @@ type manager struct {
 	cache            *cache
 	queue            PriorityQueue
 
-	log *log.PrefixLogger
+	lastConsumedDevice *v1alpha1.Device
+	log                *log.PrefixLogger
 }
 
 // NewManager creates a new device spec manager.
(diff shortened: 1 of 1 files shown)


internal/agent/device/spec/manager.go
```diff
@@ -320,8 +321,21 @@ func (s *manager) consumeLatest(ctx context.Context) (bool, error) {
 		}
 		s.log.Debugf("New template version received from management service: %s", newDesired.Version())
 		s.queue.Add(ctx, newDesired)
+		s.lastConsumedDevice = newDesired
 		consumed = true
 	}
+	if !consumed && s.lastConsumedDevice != nil {
```
hexfusion, line 327: this is all technically in the same logical flow as if !popped could we move this into the for loop to save the dupe calls to queue consumed bool etc.

```
if !popped {
   if s.lastConsumedDevice != nil {
   // rest of logic
```

ori-amizur: It is not the same.
The code tries to empty the queue.  So if there is 1 element in the queue, in the first iteration popped is true and the device is added to the queue and lastConsumedDevice is set.  In the second iteration popped is false.
I would like to add the lastConsumedDevice only if  no device was consumed.

hexfusion (approved):

PR #1244, opened 2025-06-12
NO-ISSUE: agent/hooks: add tests for splitWithQuotes

hexfusion:


siserafin:
The cluster registration using the acm-registration fleet worked from this PR branch.

hexfusion:
> I could just be missing it, but as '\t' was added as a valid delimiter, it could be useful to include that character at least once in one of the cases.

good point on discoverability

hexfusion:
test failures unrelated

kkyrazis (approved):

PR #1253, opened 2025-06-16
EDM-1697: Fix multiple owners removal

kkyrazis:
* Fixes a minor issue where devices with no owner couldn't have their MultipleOwners condition removed

Reproduction Steps:
1. Create fleet 1 with selector labels: `label1 = value`
2. Create fleet 2 with selector labels: `label2 = value`
3. Create device with labels: `label1 = value, label2 = value`
Ensure that no owner is applied to the device and that it has the multiple owners condition
5. Remove all labels from the device
**Note bug is here**. Previously the multiple owners condition would stay attached, but now it is set to false.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of devices without labels or owners to ensure correct owner updates and condition resets.

- **Tests**
  - Enhanced test coverage by adding a scenario for devices with no labels and no owner, verifying correct behavior in these cases.

### DIFF
--- a/internal/tasks/fleet_selector.go
+++ b/internal/tasks/fleet_selector.go
@@ -444,7 +444,8 @@ func (f FleetSelectorMatchingLogic) handleDeviceWithPotentialOverlap(ctx context
 			if err != nil {
 				return nil, err
 			}
-
+		}
+		if api.IsStatusConditionTrue(device.Status.Conditions, api.DeviceMultipleOwners) {
 			condition := api.Condition{
 				Type:   api.DeviceMultipleOwners,
 				Status: api.ConditionStatusFalse,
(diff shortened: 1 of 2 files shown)


internal/tasks/fleet_selector.go
```diff
@@ -444,7 +444,8 @@ func (f FleetSelectorMatchingLogic) handleDeviceWithPotentialOverlap(ctx context
 			if err != nil {
 				return nil, err
 			}
-
+		}
+		if api.IsStatusConditionTrue(device.Status.Conditions, api.DeviceMultipleOwners) {
```
avishayt, line 448: Can you explain how this fixes the issue?

kkyrazis: The actual fix is to just close the if block sooner. The `if conditionTrue` check was a suggestion from the rabbit to avoid an unnecessary call to the api if it's not necessary.
```
before
if hasOwner {
   clearOwner()
   clearMultipleOwners()
}

now 
if hasOwner {
  clearOwner()
}
if hasMultipleOwnersCond {
   clearMultipleOwners()
}
```

avishayt (approved):

internal/tasks/fleet_selector.go
avishayt: Thanks!

PR #1255, opened 2025-06-17
EDM-1709: Deterministic ordering for events

avishayt:
Added support for secondary sorting (e.g., by name) to ensure deterministic ordering when multiple resources share the same primary sort value (e.g., created_at).

## Summary by CodeRabbit

- **New Features**
  - Improved event listing and pagination with multi-column sorting by creation time and event name for consistent ordering.
- **Bug Fixes**
  - Enhanced pagination logic to support multi-column sorting and updated continuation tokens, improving listing accuracy and remaining item counts.
- **Tests**
  - Expanded test coverage with dynamically generated events and refined assertions to validate sorting, filtering, and pagination enhancements.

### DIFF
--- a/internal/store/common.go
+++ b/internal/store/common.go
@@ -11,6 +11,7 @@ import (
 	"github.com/flightctl/flightctl/internal/store/model"
 	"github.com/flightctl/flightctl/internal/store/selector"
 	"github.com/google/uuid"
+	"github.com/samber/lo"
 	"gorm.io/gorm"
 )
 
@@ -124,17 +125,19 @@ func (lq *listQuery) BuildNoOrder(ctx context.Context, db *gorm.DB, orgId uuid.U
 	return query, nil
 }
 
-func getColumnAndOrder(listParams ListParams) (SortColumn, SortOrder) {
-	// default is name in ascending order
-	column := SortByName
-	if listParams.SortColumn != nil {
-		column = *listParams.SortColumn
-	}
+func getSortColumns(listParams ListParams) ([]SortColumn, SortOrder, string) {
 	order := SortAsc
 	if listParams.SortOrder != nil {
 		order = *listParams.SortOrder
 	}
-	return column, order
+	op := map[SortOrder]string{SortAsc: ">=", SortDesc: "<="}[order]
+
+	columns := listParams.SortColumns
+	if len(columns) == 0 {
+		columns = []SortColumn{SortByName}
+	}
+
+	return columns, order, op
 }
 
 func (lq *listQuery) Build(ctx context.Context, db *gorm.DB, orgId uuid.UUID, listParams ListParams) (*gorm.DB, error) {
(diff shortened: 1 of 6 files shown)


internal/store/common.go
```diff
@@ -170,16 +173,33 @@ func AddPaginationToQuery(query *gorm.DB, limit int, cont *Continue, listParams
...
+			fmt.Sprintf("(%s, %s) %s (?, ?)", column, *listParams.SecondarySortColumn, op),
+			lastItemName, *lastItemSecondaryName,
+		)
+	} else {
+		query = query.Where(fmt.Sprintf("%s %s ?", column, op), lastItemName)
+	}
 
-	query.Where(fmt.Sprintf("%s %s ?", column, map[SortOrder]string{SortAsc: ">=", SortDesc: "<="}[order]), lastItemName).Count(&count)
+	query.Count(&count)
 	return count
```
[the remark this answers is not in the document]
avishayt: No.  The List method in generic.go (and fleet.go) read an extra item and discard it.  lastItemName and lastItemSecondaryName refer to this extra item, which should be the first in the next iteration.  That's why it should be >= or <=.

internal/store/store.go
```diff
@@ -157,27 +157,32 @@ const (
...
-	SortOrder          *SortOrder
-	SortColumn         *SortColumn
+	Limit               int
+	Continue            *Continue
+	FieldSelector       *selector.FieldSelector
+	LabelSelector       *selector.LabelSelector
+	AnnotationSelector  *selector.AnnotationSelector
+	SortOrder           *SortOrder
+	SortColumn          *SortColumn
+	SecondarySortColumn *SortColumn
```
ori-amizur: Maybe consider making sort columns a slice.  If it is then the function `getColumnAndOrder` will return the slice and not just a single column.  This will remove the need to check in many places the existence of the `SecondarySortColumn`

internal/store/common.go
```diff
@@ -170,16 +173,33 @@ func AddPaginationToQuery(query *gorm.DB, limit int, cont *Continue, listParams
...
 }
 
-func CountRemainingItems(query *gorm.DB, lastItemName string, listParams ListParams) int64 {
+func CountRemainingItems(query *gorm.DB, lastItemName string, lastItemSecondaryName *string, listParams ListParams) int64 {
 	var count int64
 
 	column, order := getColumnAndOrder(listParams)
+	op := map[SortOrder]string{SortAsc: ">=", SortDesc: "<="}[order]
+
+	if listParams.SecondarySortColumn != nil && lastItemSecondaryName != nil {
```
ori-amizur: The condition code seems to repeat itself between `resolveOrDefault` and `CountRemainingItems`. Maybe a single function can do it once?

internal/store/store.go
avishayt: Done

internal/store/common.go
avishayt: Done

internal/store/common.go
```diff
@@ -168,18 +175,49 @@ func AddPaginationToQuery(query *gorm.DB, limit int, cont *Continue, listParams
...
+func CountRemainingItems(query *gorm.DB, nextValues []string, listParams ListParams) int64 {
 	var count int64
 
-	column, order := getColumnAndOrder(listParams)
+	columns, _, op := getSortColumns(listParams)
+	if len(columns) != len(nextValues) {
+		return 0
+	}
+
+	if len(columns) == 1 {
```
ori-amizur: Isn't 1 a private case of multiple ? Maybe we can omit the condition for 1?

avishayt: Done

ori-amizur (approved):

PR #1257, opened 2025-06-17
EDM-1627: A clear error message is missing when we try to start a console session with a not existing device name

ori-amizur:
With the current kubernetes API, a clear error message is not extracted from the websocket upgrade error.
This is fixed from kubernetes api 0.32.0 or later. An error is now created by trying to perform "get" operation on the device.  If an error occurs it is displayed.
Since the generated error is not necessarily the real error the occured, the solution for this bug should be removed when moving to kubernetes API 0.32.0 or later.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error messages for specific device connection failures, providing clearer feedback for issues such as unauthorized access, forbidden access, device not found, or service unavailable.
	- Enhanced handling of canceled operations, ensuring the app exits with an appropriate code when interrupted.

- **New Features**
	- Added more detailed error reporting for device connection upgrade failures.

### DIFF
--- a/internal/cli/console.go
+++ b/internal/cli/console.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


siserafin:
Tested, LGTM

PR #1258, opened 2025-06-17
EDM-1058: Implement crypto.Signer for TPM LDevID

lkatalin:
~~Draft WIP~~

- [x] This will likely make more sense and will pass tests once it is rebased on #1235 

This PR implements a `crypto.Signer` for TPM-based keys (LDevID) and uses it during agent startup to sign the CSR if experimental features are enabled and the TPM is available.

## Summary by CodeRabbit

- **New Features**
  - Added support for using TPM-based cryptographic keys when experimental features are enabled, with automatic fallback to software keys if TPM setup is unavailable.
  - Enabled the TPM to act as a standard cryptographic signer, allowing use of the TPM’s LDevID key for signing operations and public key retrieval.

- **Bug Fixes**
  - Improved error handling and logging during TPM setup and key generation processes.

- **Refactor**
  - Updated internal logic to conditionally register TPM-related information collectors and streamline key management based on feature flags.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -73,23 +73,56 @@ func (a *Agent) Run(ctx context.Context) error {
 	// create file io writer and reader
 	deviceReadWriter := fileio.NewReadWriter(fileio.WithTestRootDir(a.config.GetTestRootDir()))
 
-	// ensure the agent key exists if not create it.
-	if !a.config.ManagementService.Config.HasCredentials() {
-		a.config.ManagementService.Config.AuthInfo.ClientCertificate = filepath.Join(a.config.DataDir, agent_config.DefaultCertsDirName, agent_config.GeneratedCertFile)
-		a.config.ManagementService.Config.AuthInfo.ClientKey = filepath.Join(a.config.DataDir, agent_config.DefaultCertsDirName, agent_config.KeyFile)
-	}
-	publicKey, privateKey, _, err := fcrypto.EnsureKey(deviceReadWriter.PathFor(a.config.ManagementService.AuthInfo.ClientKey))
-	if err != nil {
-		return err
-	}
+	// generate tpm client only if experimental features are enabled
+	experimentalFeatures := experimental.NewFeatures()
+	var tpmClient *lifecycle.TpmClient
+	var publicKeyHash []byte
+	var signer crypto.Signer
+	tpmSuccess := false
 
-	publicKeyHash, err := fcrypto.HashPublicKey(publicKey)
-	if err != nil {
-		return err
+	if experimentalFeatures.IsEnabled() {
+		a.log.Warn("Experimental features enabled: creating TPM client")
+		tc, err := lifecycle.NewTpmClient(a.log)
+		if err != nil {
+			a.log.Warnf("Experimental feature: tpm is not available: %v", err)
+		} else {
+			a.log.Warnf("Experimental feature: creating TPM keys for device identity")
+			tpmClient = tc
+			// note: the signer does not reveal the tpm's private key
+			tpmSigner := tpmClient.GetLDevIDSigner()
+			publicKey := tpmClient.GetLDevIDPublic()
+			hash, err := fcrypto.HashPublicKey(publicKey)
+			if err != nil {
+				a.log.Errorf("Experimental feature: Unable to create public key hash for TPM key: %v", err)
+			} else {
+				a.log.Warnf("Experimental feature: TPM-based keys successfully created")
+				signer = tpmSigner
+				publicKeyHash = hash
+				tpmSuccess = true
+			}
+		}
+	}
+	if !experimentalFeatures.IsEnabled() || !tpmSuccess {
+		a.log.Warnf("Experimental features are not enabled, or TPM setup did not succeed: proceeding with software keys")
+		// ensure the agent key exists if not create it.
+		if !a.config.ManagementService.Config.HasCredentials() {
+			a.config.ManagementService.Config.AuthInfo.ClientCertificate = filepath.Join(a.config.DataDir, agent_config.DefaultCertsDirName, agent_config.GeneratedCertFile)
+			a.config.ManagementService.Config.AuthInfo.ClientKey = filepath.Join(a.config.DataDir, agent_config.DefaultCertsDirName, agent_config.KeyFile)
+		}
+		publicKey, privateKey, _, err := fcrypto.EnsureKey(deviceReadWriter.PathFor(a.config.ManagementService.AuthInfo.ClientKey))
+		if err != nil {
+			return err
+		}
+		signer = privateKey.(crypto.Signer)
+		hash, err := fcrypto.HashPublicKey(publicKey)
+		if err != nil {
+			return err
+		}
+		publicKeyHash = hash
 	}
 
 	deviceName := strings.ToLower(base32.HexEncoding.WithPadding(base32.NoPadding).EncodeToString(publicKeyHash))
-	csr, err := fcrypto.MakeCSR(privateKey.(crypto.Signer), deviceName)
+	csr, err := fcrypto.MakeCSR(signer, deviceName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/agent/agent.go
```diff
@@ -148,6 +148,22 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			publicKeyHash, err := fcrypto.HashPublicKey(publicKey)
+			if err != nil {
+				a.log.Errorf("Unable to create public key hash for TPM key: %v", err)
+			}
+			deviceName := strings.ToLower(base32.HexEncoding.WithPadding(base32.NoPadding).EncodeToString(publicKeyHash))
+			tpmCsr, err := fcrypto.MakeCSR(tpmSigner, deviceName)
+			if err != nil {
+				a.log.Errorf("Unable to create CSR with TPM key: %v", err)
+			}
+			csr = tpmCsr
```
fzdarsky: This updates the key pair after the agent already created one and stored it on disk. I suggest moving this up to lines 76-91 and use if/else based on experimental flag, so it's only created once.

fzdarsky: Also, the CSR generation should then be outside the if/else and be a single call for both.

internal/agent/agent.go
```diff
@@ -148,6 +148,22 @@ func (a *Agent) Run(ctx context.Context) error {
...
 			systemInfoManager.RegisterCollector(ctx, "attestation", tpmClient.TpmAttestationCollector)
+
+			a.log.Warn("Experimental features enabled: creating TPM hardware identity")
+
+			// overwrite previous CSR and device name
+			tpmSigner := tpmClient.GetLDevIDSigner()
+			publicKey := tpmClient.GetLDevIDPublic()
+			publicKeyHash, err := fcrypto.HashPublicKey(publicKey)
+			if err != nil {
+				a.log.Errorf("Unable to create public key hash for TPM key: %v", err)
```
hexfusion: do we want to return here? if we can't generate the hash I assume `publicKeyHash` is not usable?

internal/agent/agent.go
lkatalin: I've moved it

internal/agent/agent.go
lkatalin: Yeah, I added some logic to go ahead and create SW keys in this case

internal/agent/agent.go
lkatalin: I think there is more concern about the TPM code impacting users in this configuration though. The `Sign` implementation is not called until we try to `MakeCSR` and things can go wrong at that stage. So if we want to be maximally cautious I would want to put it back where it was because that way there is no chance of it impacting users.

lkatalin: As @fzdarsky pointed out, nothing experimental will be involved in `MakeCSR` unless the experimental feature is enabled for TPM keys, so I reverse my previous comment -  we should be okay.

internal/agent/agent.go
```diff
@@ -73,23 +73,56 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	var tpmClient *lifecycle.TpmClient
+	var publicKeyHash []byte
+	var signer crypto.Signer
+	tpmSuccess := false
 
-	publicKeyHash, err := fcrypto.HashPublicKey(publicKey)
-	if err != nil {
-		return err
+	if experimentalFeatures.IsEnabled() {
+		a.log.Warn("Experimental features enabled: creating TPM client")
```
fzdarsky, line 84: Maybe we can emit the warning "Experimental features enabled: [TPM-bound identity, TPM info collection]" once at the beginning and then no longer repeat the "Experimental feature:" in the log messages later? Also, should we move this one up to after line 77.

internal/agent/agent.go
```diff
@@ -73,23 +73,56 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		tc, err := lifecycle.NewTpmClient(a.log)
+		if err != nil {
+			a.log.Warnf("Experimental feature: tpm is not available: %v", err)
```
fzdarsky, line 87: When this is no longer experimental, what would be the expected behaviour? For example, "If a TPM v2 exists and the user did not explicitly opt-out of using it, use the TPM and fail on errors."?

Asking because here we fall back to using the non-secure identity if anything goes wrong and that seems risky.

fzdarsky: Spelling: Always capitalise acronyms like TPM.

internal/agent/agent.go
```diff
@@ -73,23 +73,56 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		} else {
+			a.log.Warnf("Experimental feature: creating TPM keys for device identity")
```
fzdarsky, line 89:
```suggestion
			a.log.Infof("Creating TPM-bound device identity")
```

internal/agent/agent.go
```diff
@@ -73,23 +73,56 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			tpmClient = tc
+			// note: the signer does not reveal the tpm's private key
+			tpmSigner := tpmClient.GetLDevIDSigner()
+			publicKey := tpmClient.GetLDevIDPublic()
+			hash, err := fcrypto.HashPublicKey(publicKey)
+			if err != nil {
+				a.log.Errorf("Experimental feature: Unable to create public key hash for TPM key: %v", err)
+			} else {
+				a.log.Warnf("Experimental feature: TPM-based keys successfully created")
```
fzdarsky, line 98:
```suggestion
				a.log.Infof("Using TPM-bound device identity")
```

fzdarsky: It's more informative for users to log the consequence (using/not using TPM for X) than that a step towards that succeeded.

internal/agent/agent.go
```diff
@@ -73,23 +73,56 @@ func (a *Agent) Run(ctx context.Context) error {
...
+				signer = tpmSigner
+				publicKeyHash = hash
+				tpmSuccess = true
+			}
+		}
+	}
+	if !experimentalFeatures.IsEnabled() || !tpmSuccess {
+		a.log.Warnf("Experimental features are not enabled, or TPM setup did not succeed: proceeding with software keys")
```
fzdarsky, line 106:
```suggestion
		a.log.Warnf("Using software-based device identity")
```

internal/agent/agent.go
```diff
@@ -133,24 +166,17 @@ func (a *Agent) Run(ctx context.Context) error {
...
 	}
 
-	// generate tpm client only if experimental features are enabled
-	experimentalFeatures := experimental.NewFeatures()
-	if experimentalFeatures.IsEnabled() {
-		a.log.Warn("Experimental features enabled: creating TPM client")
-		tpmClient, err := lifecycle.NewTpmClient(a.log)
+	// register tpm info collectors if experimental features are enabled
+	if experimentalFeatures.IsEnabled() && tpmClient != nil {
+		a.log.Warn("Experimental features enabled: registering TPM info collection functions")
```
fzdarsky, line 171:
```suggestion
		a.log.Info("Registering TPM info collection functions")
```

fzdarsky: Is this important enough to print at info level or should this be debug level instead?

internal/agent/agent.go
```diff
@@ -133,24 +166,17 @@ func (a *Agent) Run(ctx context.Context) error {
...
-			a.log.Warnf("Experimental feature: tpm is not available: %v", err)
-		} else {
-			a.log.Warn("Experimental features enabled: registering TPM info collection functions")
-			err := tpmClient.UpdateNonce(make([]byte, 8))
-			if err != nil {
-				a.log.Errorf("Unable to update nonce in tpm client: %v", err)
-			}
-			systemInfoManager.RegisterCollector(ctx, "tpmVendorInfo", tpmClient.TpmVendorInfoCollector)
-			systemInfoManager.RegisterCollector(ctx, "attestation", tpmClient.TpmAttestationCollector)
+			a.log.Errorf("Unable to update nonce in tpm client: %v", err)
```
fzdarsky, line 174: What is the consequence of not updating the nonce? Is that a security risk or is it acceptable to let the system continue?

internal/agent/agent.go
```diff
@@ -133,24 +166,17 @@ func (a *Agent) Run(ctx context.Context) error {
...
 		}
+		systemInfoManager.RegisterCollector(ctx, "tpmVendorInfo", tpmClient.TpmVendorInfoCollector)
+		systemInfoManager.RegisterCollector(ctx, "attestation", tpmClient.TpmAttestationCollector)
 	} else {
-		a.log.Debug("Experimental features are not enabled: skipping creation of TPM client and registration of TPM collection functions")
+		a.log.Debug("Experimental features are not enabled: skipping registration of TPM collection functions")
```
fzdarsky, line 179: Necessary?

lkatalin:
Since we're working out the bugs in the agent code, I created #1401 to merge just the Signer implementation functionality.

lkatalin:
Closing this to avoid confusion. May reopen if needed.

PR #1261, opened 2025-06-17
EDM-1681 and EDM-1682: Add Validation and Allow for templated for systemd services

SiddarthR56:
## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced validation for systemd match patterns in device specifications with stricter format and character rules to prevent invalid entries.
- **Tests**
  - Added tests to verify systemd unit name pattern validation, ensuring proper acceptance and rejection of valid and invalid formats.


eldar101:
Pass the code checks, consider the coderabbit suggestions and after that i'll approve.

fzdarsky:
The regex needs to be more flexible than this. Please see related [EDM-1682](https://issues.redhat.com/browse/EDM-1682).

internal/util/validation/formats_test.go
```diff
@@ -88,3 +88,28 @@ func TestValidateGitRevision(t *testing.T) {
...
+		"service",
+		"foo[0-9].service",
+		"foo?.service",
+		"foo\\.service",
+	}
+	for _, val := range goodValues {
+		assert.Empty(ValidateSystemdName(&val, "good.unit"))
+	}
+
+	badValues := []string{
```
kkyrazis, line 106: simply because the valid length changed from 256 to 63 it might be useful to include a 64 character length string in this test

SiddarthR56: Thanks for catching that! I didn’t intend to change the max length to 63—I'm not sure how I missed it. I’ve now updated the validation check back to the original max length.

internal/util/validation/formats_test.go
```diff
@@ -88,3 +88,33 @@ func TestValidateGitRevision(t *testing.T) {
...
+		"foo@@bar.service",
+		"",
+		"@bar.service",
+		"foo;bar.service",
+		"junk!@#$%^&*()_+-=[]{}|;:',.<>?/~`1234567890" +
```
kkyrazis: This is an excellent string and I'm not sure if this would actually work, but I feel like `string(make([]rune,257))` should work and is easier to ensure that its actually appropriately sized.

Update: I've confirmed that this approach should work

celdrake (approved):

PR #1263, opened 2025-06-18
NO-ISSUE: Fix issue with update policy not triggering version updates

kkyrazis:
Independent updates to an `UpdatePolicy` does not trigger a new rendered version. This fixes that and adds a very limited set of tests.


hexfusion (approved):
nice :+1:

PR #1268, opened 2025-06-20
EDM-991: enabling to change the client.yaml directory

noga-magen:


internal/cli/global.go
```diff
@@ -4,13 +4,13 @@ import (
...
 	"errors"
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/client-go/util/homedir"
```
hexfusion: do we need client-go here? could we achieve this in another way with the standard library ([os](https://pkg.go.dev/os#UserHomeDir))? the important take away is client-go is pretty heavy and the more we rely on it the harder it will be to remove.

fzdarsky: I agree it's heavy, just wanted to point out we're already using it today, especially the console support.

IMHO, the problem is more client-go's non-standard [heuristic](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/homedir/homedir.go#L33-L89) for selecting the dir and that we do not need the user's home but the config dir (see my other comment below).

internal/cli/global.go
```diff
@@ -70,17 +79,17 @@ func (o *GlobalOptions) WithTimeout(ctx context.Context) (context.Context, conte
...
 
-func ConfigFilePath(context string) string {
-	if len(context) > 0 && context != "default" {
-		return filepath.Join(ConfigDir(), defaultConfigFileName+"_"+context+"."+defaultConfigFileExt)
+func (o *GlobalOptions) determineConfigDir() string {
+	if o.ConfigDirOverride != "" {
+		return o.ConfigDirOverride
 	}
-	return filepath.Join(ConfigDir(), defaultConfigFileName+"."+defaultConfigFileExt)
+	return filepath.Join(homedir.HomeDir(), ".config", "flightctl")
```
fzdarsky: Note that `$HOME/.config/flightctl` is only correct on Linux and even only on _some_ distros. To ensure it works correctly everywhere (incl. Darwin and Windows), we need to keep using `os.UserConfigDir` instead.

Nit: Changing the order and naming of these functions to get the config dir and file path (instead of just turning them into a method) makes the diff unnecessarily harder to review.

internal/cli/login.go
```diff
@@ -253,6 +253,12 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return fmt.Errorf("persisting client config: %w", err)
 	}
+
+	defaultConfigPath := client.DefaultFlightctlClientConfigPath()
```
fzdarsky: Ouch. I just noticed we're defining the client's config file in two locations (`internal/cli/global` and `internal/client`) and inconsistently.

@noga-magen could I please ask you to introduce a `client.DefaultFlightctlClientConfigDir()` based on `os.UserConfigDir()` and update both the `client.DefaultFlightctlClientConfigPath()` and the `ConfigDir()` to use it? Thanks.

internal/cli/global.go
noga-magen: sure! I will also change the  DefaultFlightctlClientConfigPath() (mentioned in the other comment below)

internal/cli/global.go
```diff
@@ -49,6 +51,14 @@ func (o *GlobalOptions) Validate(args []string) error {
 		return fmt.Errorf("request-timeout must be greater than 0")
 	}
 
+	if o.ConfigDir != "" {
+		path := filepath.Clean(o.ConfigDir)
+		ext := filepath.Ext(path)
+		if ext != "" {
+			return fmt.Errorf("config-dir should specify a directory path")
+		}
```
fzdarsky, line 59: Even without extension, `o.ConfigDir` could point to a file, so you may want to test the path exists and is a dir.

Also, good call that you clean the `o.ConfigDir` path. You're not writing it back, though, so later code uses the uncleaned version. Should we move the clean into the `Complete()` method?

internal/cli/login.go
```diff
@@ -253,6 +253,15 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
...
 		return fmt.Errorf("persisting client config: %w", err)
 	}
+
+	defaultConfigPath, err := client.DefaultFlightctlClientConfigPath()
+	if err != nil {
+		return fmt.Errorf("could not get user config directory: %w", err)
+	}
+	if o.ConfigFilePath != defaultConfigPath {
+		fmt.Printf("Using a non-default configuration file path: %s (Default: %s)\n", o.ConfigFilePath, defaultConfigPath)
+	}
```
fzdarsky, line 263: The config file dir/path are being used before this, e.g. in lines 200 and 252, so this code needs to be before that. I'd suggest moving it into the `.Complete()` method.

lkatalin: Looks like this change is being implemented in the new PR https://github.com/flightctl/flightctl/pull/1285/files

PR #1271, opened 2025-06-23
EDM-1271: Make tasks idempotent (first step)

avishayt:
Three commits:
1. Make fleet_validate task idempotent
2. Add comments to device_render, fleet_rollout, repo_update which are already idempotent
3. Remove bulk delete logic from tasks

The fleet_selector task is idempotent but racy - still have work to do there.

## Summary by CodeRabbit

- **New Features**
  - Template versions are now listed with improved sorting, showing the most recently created versions first.

- **Bug Fixes**
  - Template version creation during fleet validation is now idempotent, preventing duplicate versions if the process is retried.

- **Documentation**
  - Added detailed explanations to several background tasks, clarifying their behavior and reliability.

- **Chores**
  - Removed support for "delete all" operations for repositories, fleets, and devices, along with related callbacks and tests.


internal/tasks/fleet_validate.go
```diff
@@ -80,15 +103,19 @@ func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Co
 	immediateRollout := fleet.Spec.RolloutPolicy == nil || fleet.Spec.RolloutPolicy.DeviceSelection == nil
 	tv, status := t.serviceHandler.CreateTemplateVersion(ctx, templateVersion, immediateRollout)
 	if status.Code != http.StatusCreated {
-		return t.setStatus(ctx, fmt.Errorf("creating templateVersion for valid fleet: %s", status.Message))
+		if status.Code == http.StatusConflict {
```
ori-amizur, line 99: If a fleet spec changes and rendered, and then the previous spec is set in the fleet.  Wouldn't we miss the second rollout ?

avishayt: Good call!  I changed it to use the fleet's current generation number (incremented each time spec is changed).

internal/tasks/fleet_validate.go
```diff
@@ -231,3 +255,7 @@ func (t *FleetValidateLogic) validateHttpProviderConfig(ctx context.Context, con
 
 	return &httpConfigProviderSpec.Name, &httpConfigProviderSpec.HttpRef.Repository, nil
 }
+
+func generateTemplateVersionName(fleet *api.Fleet) (string, error) {
+	return fmt.Sprintf("%s-%d", *fleet.Metadata.Name, *fleet.Metadata.Generation), nil
```
ori-amizur: No error is needed here.  This function cannot fail

ori-amizur: If the fleet is deleted and re-created.  Will the generation start from 0?  If yes, then the template version name will be the same

avishayt: I checked that - the templateversions are deleted together with the fleet, so no issue there

avishayt: Fixed

ori-amizur: OK

PR #1272, opened 2025-06-23
EDM-1739: Fix device comparisons

avishayt:
We have a special comparison method for Device specs because reflect.DeepEqual doesn't work with unions. The problem is that we keep forgetting to update this method as we extend the API. This commit replaces this delicate code - it converts both specs to JSON and compares the output. This works for unions and is actually faster than DeepEqual.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability of device and fleet specification comparisons with enhanced equality checks that handle complex nested and union-typed fields more accurately.

- **Tests**
	- Expanded test coverage for device specification equality, including scenarios with applications, volumes, and comprehensive nested fields.
	- Added tests to ensure consistent and deterministic equality check behavior.
	- Enhanced integration tests simulating real-world usage and database serialization round-trips.
	- Updated device store tests to validate updates involving comprehensive specs and metadata label changes.

### DIFF
--- a/api/v1alpha1/util.go
+++ b/api/v1alpha1/util.go
@@ -5,12 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"slices"
 	"strconv"
 	"strings"
 	"text/template"
-
-	"github.com/samber/lo"
 )
 
 type DeviceCompletionCount struct {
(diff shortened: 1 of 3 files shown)


avishayt:
>  We have a special comparison method for Device specs because reflect.DeepEqual doesn't work with unions. The problem is that we keep forgetting to update this method as we extend the API. This commit replaces this delicate code - it replaces manual field enumeration with automatic reflection-based deep comparison that handles union types safely and doesn't break when API fields are added.
> 
> ## Summary by CodeRabbit
> * **Bug Fixes**
>   
>   * Improved reliability of device and fleet specification comparisons, ensuring more consistent and accurate equality checks.
> * **Tests**
>   
>   * Expanded test coverage for device specification equality, including new scenarios with applications, volumes, and complex nested fields.
>   * Added tests to verify consistent and deterministic behavior of equality checks.
>   * Enhanced integration tests to simulate real-world scenarios and database serialization effects.
>   * Updated device store tests to validate updates with comprehensive specs and label changes.

siserafin:
Re-tested after the updated. The conflict issue is not reproduced.

hexfusion (approved):
Spoke offline on reviewing the performance on the reflect usage and looking into alternatives for the future. For today, it solves the problem.

PR #1283, opened 2025-06-24
EDM-1747: Fix policy syncing

kkyrazis:
Assisted by: Cursor / claude 4 sonnet

* Enhances `policy.Manager` to maintain the state that was kept in `spec/queue.go`. 
* Allows the policy manager to maintain multiple versions as opposed to only a single one
* Whenever a new version is observed by the spec manager, a call to Sync is made on the policy manager to register the new policies.
* If a version's policies aren't ready, it's put back into the queue with a delay equal to whatever policy's schedule triggers next.
* Adds conditions to the spec manager so that failures are externally visible.

## Summary by CodeRabbit

* **New Features**
  * Improved device update scheduling with precise readiness checks and status reporting, clearly indicating when updates are delayed due to policy or schedule.
  * Added support for delayed processing of device updates, ensuring updates occur only when all conditions are met.
  * Introduced detailed status conditions reflecting update retry states, errors, and policy readiness.
  * Integrated versioned policy management with caching and enhanced readiness evaluation.
  * Added explicit policy readiness checks and requeueing with delay when policies are not ready.
* **Bug Fixes**
  * Enhanced handling of device update errors and retries, providing more robust fallback and recovery during update failures.
* **Documentation**
  * Clarified documentation for scheduling updates and device status reporting, including updated examples and best practices.
* **Tests**
  * Added comprehensive tests for scheduling logic, policy readiness, delayed update processing, and status conditions to ensure reliability and correctness.
  * Included new tests for priority queue delayed availability and ordering.
  * Expanded coverage for update state tracking and policy manager interactions.


internal/agent/device/spec/queue.go
```diff
@@ -196,7 +222,11 @@ func (m *queueManager) getOrCreateRequeueState(ctx context.Context, version int6
 		}
 		m.requeueLookup[version] = state
 	}
-
+	// reconcile any version specific spec policies before applying an update.
+	// i.e. the update/download policies are spec dependent.
+	if err := m.reconcileSpec(ctx, item.Spec); err != nil {
```
eldar101:
```suggestion
if err := m.reconcileSpec(ctx, item.Spec); err != nil {
    m.log.Warnf("Failed to execute reconcilers for version %d: %v", version, err)
    if isCriticalError(err) {
        m.setFailed(version)
        return nil
    }
}
```

internal/agent/device/spec/queue.go
```diff
@@ -176,6 +178,29 @@ func (m *queueManager) SetFailed(version int64) {
 	m.setFailed(version)
 }
 
+// RegisterReconciler registers a reconciler that will be invoked when a device is retrieved from the queue.
+func (m *queueManager) RegisterReconciler(reconciler Reconciler) {
+	m.reconcilers = append(m.reconcilers, reconciler)
```
eldar101: add validation please.

kkyrazis: This PR isn't ready for anything, I was discussing some items with Sam and this was the simplest way to bounce ideas off of each other

internal/agent/device/spec/queue.go
hexfusion: why?

eldar101: Reconciler errors are logged but don't prevent queue processing

kkyrazis: This is all fake and just an attempt to get Eldar to waste his premium prompts. Mission accomplished.

internal/agent/device/policy/manager.go
```diff
@@ -11,9 +11,27 @@ import (
...
+	isReady  bool // defaults to false
+}
+
+// versionSchedules holds the schedules and cached results for a specific version
+type versionSchedules struct {
+	download *policySchedule
+	update   *policySchedule
+
+	// Timestamp of when this version was last accessed (for LRU eviction)
+	lastAccessed time.Time
```
hexfusion: regarding evictions, if we know the current version and can sort by version, can we safely evict items from earlier versions? do we already have something that could be generalized and reused for this purpose?

internal/agent/device/policy/manager.go
```diff
@@ -23,66 +41,183 @@ type manager struct {
...
 	return &manager{
-		log: log,
+		versions:    make(map[string]*versionSchedules),
+		versionsCap: maxVersionsCapacity,
+		log:         log,
 	}
 }
 
-func (m *manager) Sync(ctx context.Context, desired *v1alpha1.DeviceSpec) error {
+func (m *manager) Sync(ctx context.Context, desired *v1alpha1.DeviceSpec, version string) error {
```
hexfusion: nit: we talked about changing the signature to include Device so we get Version() for free and it aligns with other Sync() methods.

internal/agent/device/spec/spec.go
```diff
@@ -61,14 +67,15 @@ type Manager interface {
...
-	CheckPolicy(ctx context.Context, policyType policy.Type, version string) error
+
 	status.Exporter
 }
 
 type PriorityQueue interface {
 	// Add adds a new spec to the scheduler
 	Add(ctx context.Context, spec *v1alpha1.Device)
+	// AddWithDelay adds a new spec to the scheduler with a specific next available time
+	AddWithDelay(ctx context.Context, spec *v1alpha1.Device, nextAvailable time.Time)
```
hexfusion, line 79: question: in the case that there is an error in the policy lets say its set to update in 1 year. now I want to fix that and run in 1 hour. how do we handle this for a version added to queue with the 1 year delay.

hexfusion (commented):
I generally like the direction, few questions to think about.

internal/agent/device/spec/spec.go
kkyrazis: In the current case where the Queue's size is 1, this isn't an issue as the delayed entry gets evicted. In a case in which we would want to allow applying multiple specs in order... I'm not entirely sure. Either you respect the spec or you allow a future spec to control the pace.

internal/agent/device/policy/manager.go
kkyrazis: Your point is definitely valid and we certainly can evict anything before our 'current' spec. I was trying to avoid having to indicate to the policy manager what 'current' means and just letting general access patterns handle things. 

E.g. Current is 2, specs 3-100 have all failed, we're trying spec 101 now

With the LRU approach the current spec and more recent attempts would remain in the cache, but perhaps thats too "lucky".

I'll revise this approach to add a better compaction strategy.

internal/agent/device/spec/spec_test.go
```diff
@@ -398,17 +501,23 @@ func TestUpgrade(t *testing.T) {
...
 
 			if tc.expectedError != nil {
-				require.ErrorIs(err, tc.expectedError)
+				if tc.expectedError.Error() != "" {
+					require.Error(err)
+					require.Contains(err.Error(), tc.expectedError.Error())
+				} else {
+					require.ErrorIs(err, tc.expectedError)
+				}
 				return
```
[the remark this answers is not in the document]
kkyrazis: Tell your brother to do better next time.

internal/agent/device/policy/ordered_cache.go
```diff
@@ -0,0 +1,143 @@
...
+	key   K
+	value V
+}
+
+// OrderedCache is a generic thread-safe cache that maintains insertion order
+// New entries are added to the back, and entries can be compacted from the front
+type OrderedCache[K comparable, V any] struct {
+	items map[K]*list.Element
+	order *list.List
+	mutex sync.RWMutex
```
hexfusion: in general, we want to keep the flow of sync as serial as possible. do you foresee a thread-safe requirement?

kkyrazis: The original intention was for this to be a reusable component, but I think this is an optimization for this usecase that we ultimately won't really care about. I'll spend some time making the queue that we have in the spec package more generic and just reuse that. It won't be as efficient, but it ultimately won't matter. I'll leave off the mutexes.

pkg/queues/priority_queue.go
```diff
@@ -0,0 +1,221 @@
...
+	return item
+}
+
+// IndexedPriorityQueue is a generic, thread-unsafe priority queue.
+type IndexedPriorityQueue[T any, P comparable] struct {
+	heap      *itemHeap[T, P]
+	items     map[P]*Item[T, P] // The key is now the generic, comparable priority type P
+	maxSize   int
+	extractor Extractor[T, P] // The new extractor function
+}
```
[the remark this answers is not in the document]
kkyrazis: The current comment already mentions that it's unsafe... I will improve the documentation though

docs/user/managing-devices.md
```diff
@@ -844,7 +844,12 @@ Each schedule supports:
 | `timeZone`             | (Optional) The time zone used to evaluate the schedule. Defaults to the device’s local system time zone. Must be a valid [IANA time zone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones). |
 | `startGraceDuration`   | (Optional) A duration string that extends the allowed start time window after a schedule trigger. Follows the [Go duration format](https://pkg.go.dev/time#ParseDuration), such as `"1h"` or `"45m"`. |
 
-The Flight Control agent evaluates these schedules during its control loop to determine whether each policy is currently allowed to proceed. While the device waits for the update window the device status will read `OutOfDate`. For more details please see [Device API Statuses](device-api-statuses.md#device-api-statuses).
+The Flight Control agent evaluates these schedules during its control loop to determine whether each policy is currently allowed to proceed.
+While the device waits for the update window the device status will read `Pending` and indicate the next time at which it will re-evaluate its update conditions.
```
celdrake: 1. There is no such state "Pending".  I thought you probably meant "Preparing", but I tested it with  a VM and I see it's actually "Updating" which could be misleading. The "updated" status has:
```
  updated:
    info: 'renderedVersion: 2 is not allowed to proceed with updates until 2025-07-04T00:00:00Z'
    status: Updating
```

Should we modify the state in these cases to "Preparing"?


2. When I defined the policy as immediate (* * * * *, which doesn't make sense TBH), the device also had for a brief moment the same information, just that the time in this occassion was "now". If it's a matter of a couple of seconds, could we prevent this transition? As it would generate an event, etc.
```
  updated:
    info: 'renderedVersion: 2 is not allowed to proceed with updates until 2025-06-30T12:57:00Z'
    status: Updating
```

kkyrazis: 1. You are correct, the code is Preparing but I wrote 'Pending'. I'll update the docs, thanks for catching this. 
2. This is something I've thought about, perhaps it's a documentation issue, perhaps it's behavioral, but I'm definitely open to suggestions.

The expression `* * * * *` means updates are allowed every minute. If no `startGraceDuration` is specified then the 'can update' condition only evaluates to true at `00:01:00.000, 00:02:00.000,` .etc. A `startGraceDuration` of `20s` changes the evaluation to return true at `00:01:00.000 - 00:01:20.000, 00:02:00.000 - 00:02:20.000`. (crucially there are 2/3rds of a minute in which you cannot update and that message above will be triggered) A `startGraceDuration` of `1m` for the above cron expression would mean effectively mean "update whenever".

Because there is already a user configurable grace duration, I am hesitant to include another grace duration as defined by us on top of what they specify. But if you think it's worth adding, I'm open to it.

celdrake: 1.  Ok to changing to `Preparing`, but I also think  the wording could be more explicit, and mention the exact field  to check for (`device.status.conditions.Updating.Reason`).

2. Given that I don't think users would define almost immediate update policies (as they are almost identical to defining no policy at all), I think you're right and there's no need to introduce a new grade duration.

kkyrazis:
Will be re-addressed in a future feature. Will open a simpler bug fix PR

PR #1285, opened 2025-06-25
EDM-991: enabling to change the client.yaml directory

noga-magen:
## Summary by CodeRabbit

- **Improvements**
  - Configuration file paths are now automatically cleaned and normalized.
  - Validation ensures configuration directories exist and are accessible.
  - Users are notified earlier when using a non-default configuration file path.
  - Enhanced error handling when determining the default configuration file path.

### DIFF
--- a/internal/cli/global.go
+++ b/internal/cli/global.go
@@ -28,7 +28,7 @@ type GlobalOptions struct {
 
 func DefaultGlobalOptions() GlobalOptions {
 	return GlobalOptions{
-		ConfigFilePath: ConfigFilePath("", ""),
+		ConfigFilePath: filepath.Clean(ConfigFilePath("", "")),
 		Context:        "",
 		RequestTimeout: 0,
 	}
@@ -51,11 +51,17 @@ func (o *GlobalOptions) Validate(args []string) error {
 		return fmt.Errorf("request-timeout must be greater than 0")
 	}
 
+	// If user provided --config-dir, validate it's actually a directory
 	if o.ConfigDir != "" {
-		path := filepath.Clean(o.ConfigDir)
-		ext := filepath.Ext(path)
-		if ext != "" {
-			return fmt.Errorf("config-dir should specify a directory path")
+		stat, err := os.Stat(o.ConfigDir)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("config directory %q does not exist", o.ConfigDir)
+			}
+			return fmt.Errorf("failed to check config directory %q: %w", o.ConfigDir, err)
+		}
+		if !stat.IsDir() {
+			return fmt.Errorf("config directory path %q exists but is not a directory", o.ConfigDir)
 		}
 	}
 
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -94,6 +94,13 @@ func (o *LoginOptions) Complete(cmd *cobra.Command, args []string) error {
 	if err := o.GlobalOptions.Complete(cmd, args); err != nil {
 		return err
 	}
+	defaultConfigPath, err := client.DefaultFlightctlClientConfigPath()
+	if err != nil {
+		return fmt.Errorf("could not get user config directory: %w", err)
+	}
+	if o.ConfigFilePath != defaultConfigPath {
+		fmt.Printf("Using a non-default configuration file path: %s (Default: %s)\n", o.ConfigFilePath, defaultConfigPath)
+	}
 	return nil
 }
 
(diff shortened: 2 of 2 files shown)


internal/cli/login.go
```diff
@@ -91,6 +91,13 @@ func (o *LoginOptions) Bind(fs *pflag.FlagSet) {
 }
 
 func (o *LoginOptions) Complete(cmd *cobra.Command, args []string) error {
+	defaultConfigPath, err := client.DefaultFlightctlClientConfigPath()
```
eldar101, line 97: any validation for defaultconfigpath being empty or incorrect?

noga-magen:
Applying Frank's comments from the [previous PR](https://github.com/flightctl/flightctl/pull/1268)

lkatalin:
Based on the comments from @fzdarsky  in #1268 , this PR needs to accomplish the following:

- [x] Getting the defaultConfigPath and checking it against o.ConfigFilePath should move up into the `.Complete()` method
- [x] Test that o.ConfigDir exists and is a dir (rather than just checking that it's empty)
- [x] Move the cleaned version of o.ConfigDir into `.Complete()` and make sure later code does not use the uncleaned version

lkatalin:
@noga-magen Please add the checks on the ConfigFilePath that it exists and is a directory. Otherwise I think it looks good

internal/cli/global.go
```diff
@@ -51,12 +51,9 @@ func (o *GlobalOptions) Validate(args []string) error {
...
 	}
 
-	if o.ConfigDir != "" {
-		path := filepath.Clean(o.ConfigDir)
-		ext := filepath.Ext(path)
-		if ext != "" {
-			return fmt.Errorf("config-dir should specify a directory path")
-		}
+	// Validate ConfigDir exists and is a directory
+	if err := validateConfigDir(filepath.Dir(o.ConfigFilePath)); err != nil {
```
lkatalin: `filepath.Dir` is going to take only the directory from `o.ConfigFilePath` so it will pass the validation check that it is a directory, even if the full `o.ConfigFilePath` is not a directory. (see [description](https://pkg.go.dev/path/filepath#Dir)) If the intent is to use the full `o.ConfigFilePath` as a directory then you'll need to call `validateConfigDir` on the full `o.ConfigFilePath`.

internal/cli/global.go
```diff
@@ -51,19 +51,22 @@ func (o *GlobalOptions) Validate(args []string) error {
...
+		return err
 	}
 
-	if _, err := os.Stat(o.ConfigFilePath); errors.Is(err, os.ErrNotExist) {
-		if o.Context != "" {
-			return fmt.Errorf("context '%s' does not exist", o.Context)
+	// Validate ConfigFilePath is not a directory and exists as a file
+	if stat, err := os.Stat(o.ConfigFilePath); err != nil {
```
lkatalin: Why not just call `validateConfigDir` on `o.ConfigFilePath` above instead of adding this check? This is duplicating functionality

lkatalin:
I don't understand why we have 3 different validation functions. As far as I understand, we just need to take `o.ConfigFilePath` and make sure it is a directory that exists. Is this not the case?

internal/cli/global.go
```diff
@@ -41,7 +41,7 @@ func (o *GlobalOptions) Bind(fs *pflag.FlagSet) {
 }
 
 func (o *GlobalOptions) Complete(cmd *cobra.Command, args []string) error {
-	o.ConfigFilePath = ConfigFilePath(o.Context, o.ConfigDir)
+	o.ConfigFilePath = filepath.Clean(ConfigFilePath(o.Context, o.ConfigDir))
```
hexfusion: if we feel filepath.Clean is useful in this context should we move it into `ConfigFilePath` ?

internal/cli/global.go
```diff
@@ -51,11 +51,15 @@ func (o *GlobalOptions) Validate(args []string) error {
...
-		path := filepath.Clean(o.ConfigDir)
-		ext := filepath.Ext(path)
-		if ext != "" {
-			return fmt.Errorf("config-dir should specify a directory path")
+		if stat, err := os.Stat(o.ConfigDir); err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("config directory %q does not exist", o.ConfigDir)
+			}
+			return fmt.Errorf("failed to check config directory %q: %w", o.ConfigDir, err)
+		} else if !stat.IsDir() {
```
hexfusion: nit: go error handling should be very boring :) but its boring pattern is easy to read.

```suggestion
stat, err := os.Stat(o.ConfigDir)
if err != nil {
  if os.IsNotExist(err) {
    return fmt.Errorf("config directory %q does not exist", o.ConfigDir)
  }
  return fmt.Errorf("failed to check config directory %q: %w", o.ConfigDir, err)
}
if !stat.IsDir() {
  // stuff
...
```

gshilin-sdb (approved):

PR #1290, opened 2025-06-26
NO-ISSUE: stop rollout in case a device has multiple owners

gshilin-sdb:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of devices with multiple owners by skipping rollout processing and logging an error message for clearer issue visibility.

### DIFF
--- a/internal/tasks/fleet_rollout.go
+++ b/internal/tasks/fleet_rollout.go
@@ -147,7 +147,8 @@ func (f FleetRolloutsLogic) RolloutDevice(ctx context.Context) error {
 	}
 
 	if api.IsStatusConditionTrue(device.Status.Conditions, api.ConditionTypeDeviceMultipleOwners) {
-		f.log.Warnf("Device has multiple owners, skipping rollout")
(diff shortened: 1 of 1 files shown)


internal/tasks/fleet_rollout.go
```diff
@@ -148,6 +148,7 @@ func (f FleetRolloutsLogic) RolloutDevice(ctx context.Context) error {
 
 	if api.IsStatusConditionTrue(device.Status.Conditions, api.DeviceMultipleOwners) {
 		f.log.Warnf("Device has multiple owners, skipping rollout")
```
hexfusion: Is this still a warn in this case? seems very unexpected

gshilin-sdb: Don't know. If this is error then we also need to return something non-nil...

avishayt: It should be an error log. Return nil for now. That will change when we implement retries.

gshilin-sdb: Talked with Avishay. It will be log.Error + return nil

gshilin-sdb: done

PR #1299, opened 2025-06-30
EDM-1205: Podman monitor runs only if necessary

kkyrazis:
Currently the podman monitor starts when a device starts (more or less), and stops when a device is stopped. This is regardless of whether there is anything that actually needs monitoring. This change introduces the ability for the monitor to start or stop depending on the state of a device's applications. 

This change will greatly help the device simulator as the monitor is a very heavy operation and is a major limiting factoring in the number of devices that can be spun up on a single machine.


internal/agent/device/applications/manager_test.go
```diff
@@ -61,31 +59,6 @@ func TestManager(t *testing.T) {
 			},
 			wantAppNames: []string{"app-new"},
 		},
-		{
```
kkyrazis, line 61: Note: moved to a separate test as this now requires an extra step to start the Monitor before stopping it.

internal/agent/device/applications/podman_monitor.go
```diff
@@ -20,21 +23,26 @@ import (
 	"github.com/samber/lo"
 )
 
+const (
+	expectedPodmanExitCode = 1
```
hexfusion:
```suggestion
	expectedPodmanSigKillExitCode = 1
```

internal/agent/device/applications/podman_monitor.go
```diff
@@ -46,23 +54,46 @@ func NewPodmanMonitor(
...
+	defer m.mu.Unlock()
+	return m.running
+}
+
+// startMonitor starts the podman monitor if it's not already running
+func (m *PodmanMonitor) startMonitor(ctx context.Context) error {
+	m.mu.Lock()
+	if m.running {
+		m.mu.Unlock()
+		return nil // already running
```
hexfusion: nit: please add debug log

internal/agent/device/applications/podman_monitor.go
```diff
@@ -46,23 +54,46 @@ func NewPodmanMonitor(
...
+	m.mu.Unlock()
+
+	m.log.Infof("Starting podman monitor")
+	monitorCtx, cancelFn := context.WithCancel(ctx)
 
 	// list of podman events to listen for
 	events := []string{"create", "init", "start", "stop", "die", "sync", "remove", "exited"}
-	m.log.Debugf("Replaying podman events since boot time: %s", m.bootTime)
-	cmd := m.client.EventsSinceCmd(ctx, events, m.bootTime)
+	since := m.lastStopTime
```
hexfusion: question: is it possible for apps to start before the monitor. if yes should we add a short time buffer so nothing is missed?

internal/agent/device/applications/podman_monitor.go
```diff
@@ -78,45 +109,75 @@ func (m *PodmanMonitor) Run(ctx context.Context) error {
 	m.mu.Lock()
 	m.cancelFn = cancelFn
 	m.cmd = cmd
+	m.running = true
+	m.initialized = true
```
hexfusion: question: do we need initialized anymore?

internal/agent/device/applications/podman_monitor.go
```diff
@@ -46,23 +54,46 @@ func NewPodmanMonitor(
...
+func (m *PodmanMonitor) startMonitor(ctx context.Context) error {
+	m.mu.Lock()
+	if m.running {
+		m.mu.Unlock()
+		return nil // already running
+	}
+	m.mu.Unlock()
+
+	m.log.Infof("Starting podman monitor")
+	monitorCtx, cancelFn := context.WithCancel(ctx)
```
hexfusion: its a common pattern to reuse the ctx var I know this looks more sane to rename but if the parent cancels so does the child.

ref. https://github.com/kubernetes/apiserver/blob/7710d9ff3e4f6db0e8831d3a2cd053fae50d35ac/pkg/registry/generic/registry/decorated_watcher.go#L33

```suggestion
	ctx, cancelFn := context.WithCancel(ctx)
```

internal/agent/device/applications/podman_monitor.go
```diff
@@ -131,6 +192,38 @@ func (m *PodmanMonitor) getApps() []Application {
 	return apps
 }
 
+// isExpectedShutdownError determines if an error is expected during shutdown
+func (m *PodmanMonitor) isExpectedShutdownError(err error) bool {
```
hexfusion: :+1: this was annoying having it log error thanks

internal/agent/device/applications/podman_monitor.go
kkyrazis: No, will remove.

internal/agent/device/applications/podman_monitor.go
kkyrazis: If even apps couldn't start before the monitor, we don't necessarily process all events before stopping the monitor. I've changed it so that we listen from the the timestamp of the last processed event.

hexfusion (approved):

internal/agent/client/podman.go
```diff
@@ -51,13 +51,17 @@ type PodmanContainerConfig struct {
 	Labels map[string]string `json:"Labels"`
 }
 
+// PodmanEvent represents the structure of a podman event as produced via a CLI events command.
+// It should be noted that the CLI represents events differently from libpod. (notably the time properties)
+// https://github.com/containers/podman/blob/main/cmd/podman/system/events.go#L81-L96
```
hexfusion, line 56: :+1:

PR #1300, opened 2025-06-30
EDM-1483: Emit all business critical events from tasks

gshilin-sdb:
## Summary by CodeRabbit

* **New Features**
  * Expanded event tracking with new event types for fleet rollouts and resource synchronization, including detailed event payloads for rollout starts and sync completions.
  * Enhanced event logging and status reporting for repository accessibility and fleet rollout lifecycle events.
  * Improved annotation handling and rollout state detection for fleets.

* **Bug Fixes**
  * More robust and consistent event callback handling across resource operations, reducing risk of missed or incorrect event reporting.

* **Chores**
  * Refactored internal callback and task infrastructure for improved maintainability and reliability.
  * Updated tests and internal APIs to support new callback signatures and event details.
  * Centralized context and logging setup for periodic tasks.
  * Simplified repository status updates and resource sync task event creation.


internal/rollout/device_selection/reconciler.go
```diff
@@ -23,6 +27,9 @@ type reconciler struct {
 	callbackManager tasks_client.CallbackManager
 }
 
+const Name = "Rollout device selection"
```
celdrake: The events for the reconcisling of fleets doesn't seem correct to me.

1. The "involvedObject" details don't represent a valid Fleet. Is is possible to create one event per fleet?
```
    {
      "actor": "rollout-device-selection",
      "apiVersion": "flightctl.io/v1alpha1",
      "details": {
        "newOwner": null,
        "previousOwner": null,
        "updatedFields": null
      },
      "involvedObject": {
        "kind": "Fleet",
        "name": "Rollout device selection"
      },
      "kind": "Event",
      "message": "fleets reconciled successfully",
      "metadata": {
        "creationTimestamp": "2025-07-01T07:17:47.721366Z",
        "name": "task:rollout-device-selection-Fleet-Rollout device selection-Pkg/shcBnv9ZUN-000087"
      },
      "reason": "TaskSuccess",
      "source": {
        "component": "flightctl-periodic"
      },
      "type": "Normal"
    },
```

2. The task generates an event also when there's no changes to be made. Would it be possible to only generate an event when the task has performed some changes in the system?

3. Maybe I'm doing something wrong, but I can't seem to filter events with the field containing spaces.
```
fc get events --field-selector=involvedObject.name='Rollout device selection'
Error: listing events: 400, message: failed to parse field selector: invalid field selector syntax: found 'device', expected: ',' or 'end of string'
```
(Edit: "rollout device selection" string is being also used as part of `metadata.name`, and spaces are invalid characters for this. So we shouldn't include spaces in values that are also used as a name/ID)

api/v1alpha1/openapi.yaml
```diff
@@ -5091,6 +5091,9 @@ components:
             - ResourceUpdateFailed
             - ResourceDeleted
             - ResourceDeletionFailed
+            - ResourceUpdateFound
+            - ResourceUpdateNotFound
+            - ResourceNoop
```
celdrake: What's the purpose of "ResourceNoop"?
IMO, an event should represent some kind of action that happened in the system. If nothing happened, there shouldn't be an event in the first place.

api/v1alpha1/openapi.yaml
```diff
@@ -5186,7 +5193,7 @@ components:
           type: array
           items:
             type: string
-            enum: [owner, labels, spec]
+            enum: [ owner, labels, spec ]
```
celdrake: This definition of `updatedFields` is not according to the current implementation. The enum restriction should be dropped.

internal/tasks/repotester.go
```diff
@@ -66,9 +72,19 @@ func (r *RepoTester) TestRepositories(ctx context.Context) {
...
 				log.Errorf("unsupported repository type: %s", repoSpec.Type)
+				continue
 			}
 
 			accessErr := r.TypeSpecificRepoTester.TestAccess(&repository)
+			status = api.StatusOK()
+			if accessErr != nil {
+				status = api.NewSuccessStatus(http.StatusNotAcceptable, http.StatusText(http.StatusNotAcceptable), "")
+			}
+			r.serviceHandler.CreateEvent(ctx,
```
celdrake: There are several issues with the events generated by the "repotester":
 1. involvedObject.name is set to "repotester", not to the actual repository name. So it's an invalid reference to any existing resource in the system.
 2. metadata.name is invalid (contains semicolon, uppercase letters and slash). This is also true in other event types. It's probably worth adding a test for that @eldar101 
Example:
```
      "metadata": {
        "creationTimestamp": "2025-07-01T14:37:51.41664Z",
        "name": "task:repotester-Repository-repotester-repotester/hGiXXNuu64-000048"
      },
```
3. Too many EventReasonRepositoryAccessible  events: these are generated on every run of the task, even when no observed accessibility changes ocurred. Is it possible to modify this behaviour?
4. No EventReasonRepositoryInaccessible event: I changed the repository URL to an unexisting one, and no such event was generated.

internal/tasks/resourcesync.go
```diff
@@ -240,6 +245,25 @@ func (r *ResourceSync) parseAndValidateResources(rs *api.ResourceSync, repo *api
...
+	updateDetails := &api.ResourceUpdatedDetails{
+		UpdatedFields: []api.ResourceUpdatedDetailsUpdatedFields{
+			api.ResourceUpdatedDetailsUpdatedFields(fmt.Sprintf("resource %s", *rs.Metadata.Name)),
+			api.ResourceUpdatedDetailsUpdatedFields(fmt.Sprintf("op %s", "???")), // TODO: where to get this?
+			api.ResourceUpdatedDetailsUpdatedFields(fmt.Sprintf("commit %s", hash)),
+		},
+	}
+	if r.serviceHandler != nil {
+		r.serviceHandler.CreateEvent(ctx,
+			r.serviceHandler.CreateGenericEvent(ctx, api.RepositoryKind, resourceSyncName, resourceSyncTask, api.StatusOK(),
```
celdrake: There are similar issues for ResourceSync that for Repository events:

1. involvedObject.name is "resourcesync" and not the actual RS name.
2. metadata.name is also invalid according to the k8s spec.
3. Needs solving the "op ???" TODO as it appears in updatedFields and message.
4. "kind" is Repository, should be ResourceSync.

api/v1alpha1/openapi.yaml
```diff
@@ -1344,7 +1344,7 @@ paths:
                 $ref: '#/components/schemas/Device'
         "204":
           description: No Content
-          content: {}
+          content: { }
```
avishayt: Please don't make these cosmetic changes unless there's a good reason to.

api/v1alpha1/openapi.yaml
avishayt: Agree for sure.  For ResourceSync we discussed ResourceUpdateFound.

internal/rollout/device_selection/reconciler.go
```diff
@@ -134,39 +147,50 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
...
+			updates = append(updates, r.logAndSetUpdate(orgId, fleetName, "Advance", err))
 			break
 		}
 	}
+
+	return updates
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) {
+	reqid.OverridePrefix("Pkg")
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

internal/rollout/device_selection/reconciler.go
gshilin-sdb: 1. involvedObject.name is now the name of the fleet.
2. Event emission in tasks was changed to a minimum. Sometimes in addition to an event from a task, an event from Fleet itself will be created, but its actor will be task. I don't see a good way to disable it.
3. removed spaces

internal/tasks/resourcesync.go
```diff
@@ -240,6 +245,25 @@ func (r *ResourceSync) parseAndValidateResources(rs *api.ResourceSync, repo *api
...
+				fmt.Sprintf("resource %s op??? commit #%s", *rs.Metadata.Name, hash),
+				"",
+				api.EventReasonResourceUpdateFound, api.EventReasonResourceNoop,
+				updateDetails,
+			),
+		)
+	}
```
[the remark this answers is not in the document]
gshilin-sdb: removed

api/v1alpha1/openapi.yaml
gshilin-sdb: removed

api/v1alpha1/openapi.yaml
gshilin-sdb: @avishayt WDYT?

internal/tasks/repotester.go
gshilin-sdb: 1. Fixed
2. fixed
3, 4. refactored

internal/tasks/resourcesync.go
gshilin-sdb: fixed

api/v1alpha1/openapi.yaml
gshilin-sdb: that's auto-formatted...
will undo

internal/rollout/device_selection/reconciler.go
celdrake: All good!

internal/tasks/repotester.go
celdrake: 1 and 3 now fixed.

2 This problem still persists across many of the events in the System, but we can ignore it in this PR since I created a Jira bug to track it. Eg. A ResourceCreated event with metadata.name="resource-create-ResourceSync-rs-a-flightctl-api-65f6555786-99scc/Zny7pJahwS-000428"

4. I'm still unable to see this event. I created a repository with a non-existing URL, and its status was "Innaccessible", but no such event was created. Later I corrected the repository URL so it would become Accessible, then broke it again. The statuses are always correct, but we never see a `EventReasonRepositoryInaccessible` event.

api/v1alpha1/openapi.yaml
```diff
@@ -5091,6 +5091,8 @@ components:
             - ResourceUpdateFailed
             - ResourceDeleted
             - ResourceDeletionFailed
+            - ResourceUpdateFound
```
celdrake: It would help having a definition of when each event should be generated. It would also help testing them and ensuring the system behaves correctly.

IIUC, `ResourceUpdateFound` events should trigger when a ResourceSync or a Repository are synced and changes were done in the git remote. IIRC, yesterday I could see those events which included the commit of the change, but now they are not being generated anymore.

api/v1alpha1/openapi.yaml
celdrake: I was testing this again and never saw a value that was outside the range. It may have been different in previous PRs.

So we can keep the current definition.

api/v1alpha1/openapi.yaml
```diff
@@ -3602,12 +3602,12 @@ components:
...
-          - $ref: '#/components/schemas/RelativePath'
-          - type: object
-            required:
-              - path
+      allOf:
+      - $ref: '#/components/schemas/FileContent'
+      - $ref: '#/components/schemas/RelativePath'
+      - type: object
+        required:
+          - path
```
avishayt: ?

api/v1alpha1/openapi.yaml
avishayt: I agree.  We should also have more descriptive names.  Where was the resource update found?  How about GitResourceChangeDetected?

api/v1alpha1/openapi.yaml
```diff
@@ -5112,6 +5114,10 @@ components:
             - DeviceContentUpdating
             - EnrollmentRequestApproved
             - EnrollmentRequestApprovalFailed
+            - FleetReconciled
+            - FleetReconcileFailed
```
avishayt: @ori-amizur Please review.  These event names aren't descriptive.  Check where they are emitted, if it makes sense, and suggest better names.

internal/tasks/repotester.go
```diff
@@ -144,6 +150,17 @@ func (r *RepoTester) SetAccessCondition(ctx context.Context, repository *api.Rep
...
+	if err != nil {
+		status = api.NewFailureStatus(http.StatusNotAcceptable, http.StatusText(http.StatusNotAcceptable), "")
+	}
+	r.serviceHandler.CreateEvent(ctx,
+		r.serviceHandler.CreateGenericEvent(ctx, api.RepositoryKind, *repository.Metadata.Name, repotesterName, status,
+			fmt.Sprintf("repository %s is accessible", *repository.Metadata.Name),
+			fmt.Sprintf("repository %s is inaccessible: %v", *repository.Metadata.Name, err),
+			api.EventReasonRepositoryAccessible, api.EventReasonRepositoryInaccessible, nil))
+
+	_, status = r.serviceHandler.ReplaceRepositoryStatus(ctx, *repository.Metadata.Name, *repository)
```
avishayt: 1. Emit the event after setting the status
2. Why do you need that fake status?

api/v1alpha1/openapi.yaml
avishayt: I would summarize what happened in the task, in the message and details.
1. What is the git hash
2. How many changes were detected
3. How many errors were detected

If errors > 0, set type to warning.

No need for ResourceUpdateNotFound

internal/tasks/repotester.go
gshilin-sdb: @avishayt Problem number 2 arises from
```go
Name: lo.ToPtr(fmt.Sprintf("%s-%s-%s-%s", resourceEvent.Prefix, resourceEvent.ResourceKind, resourceEvent.ResourceName, requestIDstr)),
```
Should we change it? Or to change definition of Event.Metadata.Name to be just string?

gshilin-sdb: @celdrake solved 4

api/v1alpha1/openapi.yaml
gshilin-sdb: fixed

api/v1alpha1/openapi.yaml
gshilin-sdb: done

api/v1alpha1/openapi.yaml
gshilin-sdb: un-done

internal/tasks/repotester.go
gshilin-sdb: 1. This event relates to repository being accessible, i.e. to err that is passed as a parameter to SetAccessCondition and on current condition. Not to result of ReplaceRepositoryStatus. Changed and now it works as expected.
2. removed

internal/tasks/repotester_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/flightctl/flightctl/pkg/reqid"
+	"github.com/go-chi/chi/v5/middleware"
+	"github.com/samber/lo"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRepoTester_SetAccessCondition(t *testing.T) {
+	require := require.New(t)
+
+	reqid.OverridePrefix(repotesterName)
```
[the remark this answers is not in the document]
gshilin-sdb: this is test file and that const is defined in its code file

internal/service/event.go
```diff
@@ -447,3 +467,21 @@ func GetFleetSelectorProcessingCompletedEvent(ctx context.Context, fleetName str
...
+	return getBaseEvent(ctx, resourceEvent{
+		ResourceKind:   api.RepositoryKind,
+		ResourceName:   repoName,
+		Prefix:         "repotester",
+		ReasonSuccess:  api.EventReasonRepositoryAccessible,
+		ReasonFailure:  api.EventReasonRepositoryInaccessible,
+		OutcomeSuccess: "Repository is accessible",
+		OutcomeFailure: func() string { return fmt.Sprintf("Repository is inaccessible: %v", err) },
+		Status:         status,
+	}, nil)
```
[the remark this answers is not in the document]
gshilin-sdb: done

internal/tasks/teststore_framework_test.go
```diff
@@ -0,0 +1,614 @@
...
+func (s *DummyRepository) Update(ctx context.Context, orgId uuid.UUID, repository *api.Repository, callback store.RepositoryStoreCallback) (*api.Repository, api.ResourceUpdatedDetails, error) {
+	for i, repo := range *s.repositories {
+		if *repository.Metadata.Name == *repo.Metadata.Name {
+			*s.repositories = append((*s.repositories)[:i], (*s.repositories)[i+1:]...)
+			*s.repositories = append(*s.repositories, *repository)
+			return repository, api.ResourceUpdatedDetails{}, nil
+		}
+	}
+	return nil, api.ResourceUpdatedDetails{}, flterrors.ErrResourceNotFound
+}
```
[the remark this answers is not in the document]
gshilin-sdb: done

internal/tasks/teststore_framework_test.go
```diff
@@ -0,0 +1,620 @@
...
+			OutcomeSuccess: successMessage,
+			OutcomeFailure: failureFunc,
+			UpdateDetails:  updateDetails,
+		}, s.log)
+}
+
+type ServiceHandler struct {
+	store store.Store
+	log   logrus.FieldLogger
+}
```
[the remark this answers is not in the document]
gshilin-sdb: added

internal/tasks/teststore_framework_test.go
```diff
@@ -0,0 +1,620 @@
...
+	return list, nil
+}
+
+func (s *DummyRepository) Create(ctx context.Context, orgId uuid.UUID, repository *api.Repository, callback store.RepositoryStoreCallback) (*api.Repository, error) {
+	if s.repositories == nil {
+		s.repositories = &[]api.Repository{}
+	}
+	*s.repositories = append(*s.repositories, *repository)
+	return repository, nil
+}
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

internal/tasks/repotester.go
celdrake: 2 is being addressed in a separate [PR](https://github.com/flightctl/flightctl/pull/1335), and the others have been fixed here. Thanks!

internal/tasks/resourcesync.go
```diff
@@ -65,44 +77,73 @@ func (r *ResourceSync) Poll(ctx context.Context) {
...
+
+	if len(updatedFields) == 0 || r.serviceHandler == nil {
+		return
+	}
+	status := api.StatusOK()
+	if totalErrors > 0 {
+		status = api.StatusInternalServerError(fmt.Sprintf("#%d errors", totalErrors))
+	}
+	r.serviceHandler.CreateEvent(ctx,
+		r.serviceHandler.CreateGenericEvent(ctx, api.RepositoryKind, resourceSyncName, resourceSyncName, status,
```
celdrake: Testing the following flow for a RS:
a) I initially created a RS with an invalid path
b) Then I fixed the path
c) Later I pushed to the git repository with a change in the device spec.

These were the issues I noticed:

1. The event is correctly generated as a warning when it was initially wrongly created. However, a new `GitResourceChangeDetected` event was not generated when I fixed the path of the RS.
2. `involvedObject` is incorrect. should be `ResourceSync/rs-a`
```
"involvedObject": {
        "kind": "Repository",
        "name": "resourcesync"
      },
```
3. Given the enum in `updatedFields`, the value seems incorrect `"updatedFields": [
          "resourcesync/rs-a  repo/basic-nginx-demo commit/#35e58ed0147576a1708a559221c4f28b384b855f: parsing failed: file does not exist"
        ]`
4. The message is not clear IMO. The format "#n" is difficult to understand, and probably the info we have in the `updatedFields` is more relevant that the number of changes. 
```
 "message": "ResourceSync Polling: #0 changes and #1 failures",
```
5. When I pushed an update to the git repository linked to the RS, the `GitResourceChangeDetected` event was not generated.

internal/rollout/disruption_budget/reconciler.go
```diff
@@ -206,10 +221,20 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
 			}
 		}
 	}
+	r.serviceHandler.CreateEvent(ctx,
```
celdrake: I tested the event defining a disruption budget.

1. The event is being generated every 30 seconds even when there's no action taken. Could it be generated only when there was some reconciliation done?
2. `involvedObject` is wrong, fleet name is "test-fleet"
```
      "involvedObject": {
        "kind": "Fleet",
        "name": "disruption-budget"
      },
```
3. The message includes the OrgId. I imagine when we have multiple orgs, the orgId will need to be part of `involvedObject`, so I'd remove it from the message for clarity. `"message": "00000000-0000-0000-0000-000000000000/test-fleet: Reconciled for Disruption Budget"`

Suggestion: something like "Fleet 'test-fleet' reconciled according to disruption budget'

internal/rollout/device_selection/reconciler.go
```diff
@@ -31,59 +38,73 @@ func NewReconciler(serviceHandler service.Service, callbackManager tasks_client.
...
+			r.sendLog(orgId, fleetName, "OnNewRollout", err)
 			return
 		}
-		if err := selector.Reset(ctx); err != nil {
-			r.log.WithError(err).Errorf("%v/%s: Reset", orgId, lo.FromPtr(fleet.Metadata.Name))
+		if err = selector.Reset(ctx); err != nil {
+			r.sendLog(orgId, fleetName, "Reset", err)
 			return
 		}
+		r.sendEvent(ctx, fleetName, fmt.Sprintf("%v/%s: On New Rollout", orgId, fleetName))
```
celdrake: Tested the rollout events.
Here everything looks great, I'd just make the comment to remove the orgId in the message for the same reason as in the disruption budget

` "message": "00000000-0000-0000-0000-000000000000/test-fleet: On New Rollout",`

Maybe something like "Fleet 'test-fleet' reconciled on new rollout" or similar.

gshilin-sdb: We're adding different orgs, so we shouldn't use only fleet names AFAIK.
Re wording: sure, no problem to change

celdrake: Yes, but the event is associated to a particular fleet as identified by its unique key in the system. 
When organizations exist in the system, the identification (`involvedObject`) should include the orgID or it could be ambiguous.  

That's why I think the orgID in the message is redundant.

gshilin-sdb: removed :D

internal/rollout/disruption_budget/reconciler.go
gshilin-sdb: fixed

internal/tasks/resourcesync.go
gshilin-sdb: I believe it's better now

celdrake: Regarding 2, I meant that `involvedObject` should be 
```
      "involvedObject": {
        "kind": "ResourceSync",
        "name": "rs-a"
      },
```
but now it is
```
      "involvedObject": {
        "kind": "Repository",
        "name": "ResourceSync/rs-a"
      },
```
which doesn't correctly identify a resource in the system.

Everything else is perfect now :)

internal/rollout/disruption_budget/reconciler.go
celdrake: Only 1 seems fixed.
2  `involvedObject.name` is still "disruption-budget" (must be fixed IMO)
3 the message still includes the OrgId. (nice to have it removed IMO, plus it's inconsistent with almost? all other messages)

internal/rollout/disruption_budget/reconciler.go
```diff
@@ -206,10 +221,20 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
...
+		r.serviceHandler.CreateGenericEvent(ctx, api.FleetKind, DisruptionBudgetName, api.StatusOK(),
```
celdrake: I imagine this would fix 2
```suggestion
		r.serviceHandler.CreateGenericEvent(ctx, api.FleetKind, *fleet.Metadata.Name, api.StatusOK(),
```

internal/rollout/disruption_budget/reconciler.go
```diff
@@ -206,10 +221,20 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
...
+			fmt.Sprintf("%v/%s: Reconciled for Disruption Budget", orgId, *fleet.Metadata.Name), "",
```
celdrake: Suggestion for 3
```suggestion
			fmt.Sprintf("Fleet %s: reconciled for Disruption Budget", *fleet.Metadata.Name), "",
```

internal/service/event.go
```diff
@@ -410,3 +431,58 @@ func GetFleetSelectorProcessingCompletedEvent(ctx context.Context, fleetName str
...
+		ReasonSuccess:  api.EventReasonGitResourceChangeDetected,
+		ReasonFailure:  api.EventReasonGitResourceChangeDetected,
+		OutcomeSuccess: fmt.Sprintf("ResourceSync Polling Success: %d changes", totalChanges),
+		OutcomeFailure: func() string {
+			return fmt.Sprintf("ResourceSync Polling: %d changes and %d failures", totalChanges, totalErrors)
+		},
+		Status:        status,
+		UpdateDetails: resourceDetails,
+	}, log)
+}
```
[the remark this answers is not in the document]
gshilin-sdb: done

internal/rollout/disruption_budget/reconciler.go
gshilin-sdb: done

internal/tasks/resourcesync.go
```diff
@@ -135,30 +175,40 @@ func (r *ResourceSync) run(ctx context.Context, log logrus.FieldLogger, rs *api.
...
 			}
 		}
 	}
 	api.SetStatusConditionByError(&rs.Status.Conditions, api.ConditionTypeResourceSyncSynced, "success", "fail", createUpdateErr)
 	if createUpdateErr != nil {
-		log.Errorf("resourcesync/%s: failed to apply resource. error: %s", *rs.Metadata.Name, err.Error())
-		return createUpdateErr
+		log.Errorf("Resource %s: failed to apply resource. error: %s", resourceName, err.Error())
+		updatedFields = append(updatedFields, api.ResourceUpdatedDetailsUpdatedFields(fmt.Sprintf("Resource %s repo %s commit #%s: failed to apply resource: %s", resourceName, repoName, commitHash, err.Error())))
+		return updatedFields, totalChanges, totalErrors + 1, createUpdateErr
```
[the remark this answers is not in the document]
gshilin-sdb: done

api/v1alpha1/openapi.yaml
avishayt: ?

celdrake:
LGTM from the UI's perspective

internal/rollout/disruption_budget/reconciler.go
```diff
@@ -153,7 +158,16 @@ func (r *reconciler) reconcileSelectionDevices(ctx context.Context, orgId uuid.U
 	}
 	remaining := lo.Ternary(numToRender > 0, numToRender, math.MaxInt)
 	for {
-		listParams.Limit = lo.ToPtr(int32(math.Min(float64(remaining), float64(maxItemsToRender))))
+		minVal := lo.Min([]int{remaining, maxItemsToRender})
+		var safeLimit int32
+		if minVal > math.MaxInt32 {
+			safeLimit = math.MaxInt32
+		} else if minVal < math.MinInt32 {
+			safeLimit = math.MinInt32
```
ori-amizur: Limit must not be negative

internal/tasks/repotester.go
```diff
@@ -144,6 +152,15 @@ func (r *RepoTester) SetAccessCondition(ctx context.Context, repository *api.Rep
...
 	}
-	_, status := r.serviceHandler.ReplaceRepositoryStatus(ctx, *repository.Metadata.Name, *repository)
+
+	repoName := lo.FromPtr(repository.Metadata.Name)
+	_, status := r.serviceHandler.ReplaceRepositoryStatus(ctx, repoName, *repository)
+
+	// This event is about the error that was transferred as a parameter to SetAccessCondition,
+	// not about success or failure to access database.
+	// It should be emitted only in a case that Condition was changed.
+	event := service.GetReporterTaskEvent(ctx, repoName, err)
```
avishayt:
```suggestion
	event := service.GetRepositoryAccessibleEvent(ctx, repoName, err)
```

internal/rollout/disruption_budget/reconciler.go
```diff
@@ -153,7 +158,16 @@ func (r *reconciler) reconcileSelectionDevices(ctx context.Context, orgId uuid.U
 	}
 	remaining := lo.Ternary(numToRender > 0, numToRender, math.MaxInt)
 	for {
-		listParams.Limit = lo.ToPtr(int32(math.Min(float64(remaining), float64(maxItemsToRender))))
+		minVal := lo.Min([]int{remaining, maxItemsToRender})
```
ori-amizur: I don't understand why is this change needed.  Is there any known issue?
I would prefer simpler code:
```suggestion
                remaining = o.Ternary(numToRender > 0 && numToRender <= int(MaxInt32), int32(numToRender), math.MaxInt32)
                .....
                
		listParams.Limit = lo.ToPtr(util.Min(remaining, maxItemsToRender))
```

internal/service/event.go
```diff
@@ -410,3 +412,85 @@ func GetFleetSelectorProcessingCompletedEvent(ctx context.Context, fleetName str
...
+		resourceDetails = &api.ResourceUpdatedDetails{
+			UpdatedFields: updatedFields,
+		}
+	}
+	return getBaseEvent(ctx, resourceEvent{
+		ResourceKind:   api.ResourceSyncKind,
+		ResourceName:   repoName,
+		ReasonSuccess:  api.EventReasonGitResourceChangeDetected,
+		ReasonFailure:  api.EventReasonGitResourceChangeDetected,
+		OutcomeSuccess: fmt.Sprintf("ResourceSync Polling Success: %d changes", totalChanges),
```
avishayt: 1. Add the git commit to the messages and details
2. The message can be "Processed %d changes for commit %s (%d failures)"

internal/service/event.go
```diff
@@ -410,3 +412,85 @@ func GetFleetSelectorProcessingCompletedEvent(ctx context.Context, fleetName str
 		}, log)
 	}
 }
+
+// GetReporterTaskEvent creates an event for repotester task
+func GetReporterTaskEvent(ctx context.Context, repoName string, err error) *api.Event {
```
avishayt:
```suggestion
func GetRepositoryAccessibleEvent(ctx context.Context, repoName string, err error) *api.Event {
```

api/v1alpha1/openapi.yaml
ori-amizur: Maybe ?
RolloutDeviceSelectionSucceded / RolloutDeviceSelectionFailed
RolloutDisruptionBudgetSucceeded / RolloutDisruptionFailed ?

We might want more granular events (batch completion, rollout completion)

internal/rollout/device_selection/reconciler.go
```diff
@@ -134,33 +153,37 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
 
 		// Once the batch is complete, set the success percentage of the current batch
 		if err = selection.SetCompletionReport(ctx); err != nil {
-			r.log.WithError(err).Errorf("%v/%s: SetCompletionReport", orgId, lo.FromPtr(fleet.Metadata.Name))
+			r.sendLog(orgId, fleetName, "SetCompletionReport", err)
```
ori-amizur: Maybe send event when a batch is completed ?

internal/rollout/device_selection/reconciler.go
```diff
@@ -134,33 +153,37 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
...
 			break
 		}
 		hasMoreSelections, err := selector.HasMoreSelections(ctx)
 		if err != nil {
-			r.log.WithError(err).Errorf("%v/%s: HasMoreSelections", orgId, lo.FromPtr(fleet.Metadata.Name))
+			r.sendLog(orgId, fleetName, "HasMoreSelections", err)
 			break
 		}
 		if !hasMoreSelections {
 			if err = selection.OnFinish(ctx); err != nil {
```
ori-amizur, line 151: Maybe emit event when there are no more selections? It means fleet rollout was complete

internal/rollout/disruption_budget/reconciler.go
gshilin-sdb: changed to 
```go
for remaining > 0 {
    minVal := lo.Min([]int{remaining, maxItemsToRender})
    listParams.Limit = lo.ToPtr(int32(minVal)) // #nosec G115
```

internal/rollout/disruption_budget/reconciler.go
gshilin-sdb: It causes linter error:
```go
G115: integer overflow conversion int -> int32 (gosec)
		listParams.Limit = lo.ToPtr(int32(util.Min(remaining, maxItemsToRender)))
```
so  we have to add "// #nosec G115"

Also util.Min is used in this file only, and it uses lo.Ternary internally, and there exists lo.Min already, so:
```go
listParams.Limit = lo.ToPtr(int32(lo.Min([]int{remaining, maxItemsToRender}))) // #nosec G115
```

internal/service/event.go
```diff
@@ -424,7 +425,79 @@ func GetInternalTaskFailedEvent(ctx context.Context, resourceKind api.ResourceKi
...
+			UpdatedFields: updatedFields,
+		}
+	}
+	return getBaseEvent(ctx, resourceEvent{
+		resourceKind:   api.ResourceSyncKind,
+		resourceName:   repoName,
+		reasonSuccess:  api.EventReasonGitResourceChangeDetected,
+		reasonFailure:  api.EventReasonGitResourceChangeDetected,
+		outcomeSuccess: fmt.Sprintf("Processed %d changes for commit %s", totalChanges, commitHash),
+		outcomeFailure: func() string {
```
avishayt, line 480: Add totalChanges, commitHash, and totalErrors to the event details

internal/tasks/repotester.go
```diff
@@ -144,6 +145,13 @@ func (r *RepoTester) SetAccessCondition(ctx context.Context, repository *api.Rep
...
 	}
-	_, status := r.serviceHandler.ReplaceRepositoryStatus(ctx, *repository.Metadata.Name, *repository)
+	repoName := lo.FromPtr(repository.Metadata.Name)
+	_, status := r.serviceHandler.ReplaceRepositoryStatus(ctx, repoName, *repository)
+
+	// This event is about the error that was transferred as a parameter to SetAccessCondition,
+	// not about success or failure to access database.
+	// It should be emitted only in a case that Condition was changed.
+	event := service.GetRepositoryAccessibleEvent(ctx, repoName, err)
+	r.serviceHandler.CreateEvent(ctx, event)
```
avishayt: Move the logic from here to a service/store callback pattern

avishayt (changes requested):

internal/rollout/device_selection/reconciler.go
```diff
@@ -17,73 +20,83 @@ type Reconciler interface {
...
+type eventCallbackCaller func(ctx context.Context, callbackEvent eventCallback, orgId uuid.UUID, name string, reason api.EventReason, message string, err error)
+
 type reconciler struct {
-	serviceHandler  service.Service
-	log             logrus.FieldLogger
-	callbackManager tasks_client.CallbackManager
+	serviceHandler          service.Service
+	log                     logrus.FieldLogger
+	callbackManager         tasks_client.CallbackManager
+	callEventCallbackCaller eventCallbackCaller
```
hexfusion: I have not deeply dug into the implementation and I know naming is hard but maybe `callEventCallbackCaller` deserves another pass?

internal/rollout/device_selection/reconciler.go
```diff
@@ -17,73 +20,83 @@ type Reconciler interface {
...
 	}
 }
 
-func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet api.Fleet) {
-	r.log.Infof("device selection: starting reconciling fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
-	defer r.log.Infof("device selection: finished reconciling fleet %v/%s", orgId, lo.FromPtr(fleet.Metadata.Name))
+func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet api.Fleet, eventCallback eventCallback) {
+	fleetName := lo.FromPtr(fleet.Metadata.Name)
+
+	r.log.Infof("device selection: starting reconciling fleet %v/%s", orgId, fleetName)
```
hexfusion: Usually, logs start with a caps while error messages do not. I can tell this was introduced before but I wonder if we should continue it.

internal/rollout/device_selection/reconciler.go
```diff
@@ -17,73 +20,83 @@ type Reconciler interface {
...
-		r.log.WithError(err).Errorf("%v/%s: IsDefinitionUpdated", orgId, lo.FromPtr(fleet.Metadata.Name))
+		r.log.WithError(err).Errorf("%v/%s: IsDefinitionUpdated", orgId, fleetName)
 		return
 	}
 	if selector.IsRolloutNew() || definitionUpdated {
+		r.callEventCallbackCaller(ctx, eventCallback, store.NullOrgId, fleetName, api.EventReasonFleetRolloutCreated, "Fleet rollout created", nil)
 		// There is either a new template version, or the rollout definition was updated
-		if err := selector.OnNewRollout(ctx); err != nil {
-			r.log.WithError(err).Errorf("%v/%s: OnNewRollout", orgId, lo.FromPtr(fleet.Metadata.Name))
+		if err = selector.OnNewRollout(ctx); err != nil {
```
hexfusion, line 77: not sure I am following why the scope of the err is not isolated? generally that is a nice reason for this type of usage `if err := something(ctx); err != nil { oops do something with err }

internal/rollout/device_selection/reconciler.go
```diff
@@ -134,30 +147,37 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
...
 			break
 		}
 	}
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) {
+	reqid.OverridePrefix("task:rollout-device-selection")
+	requestID := reqid.NextRequestID()
+	ctx = context.WithValue(ctx, middleware.RequestIDKey, requestID)
+	ctx = context.WithValue(ctx, consts.EventActorCtxKey, "task:rollout-device-selection")
```
hexfusion: `task:rollout-device-selection` looks like something we could typo by accident, should it be a const?

internal/rollout/device_selection/reconciler.go
```diff
@@ -167,6 +187,31 @@ func (r *reconciler) Reconcile(ctx context.Context) {
...
+		// Call callback in a defer with error recovery to prevent callback failures from affecting the main operation
+		defer func() {
+			if r := recover(); r != nil {
+				log.Errorf("Event Callback panicked during service status update: %v", r)
+			}
+		}()
+
+		// Call the callbackEvent - if it fails, log the error but don't propagate it
+		func() {
+			defer func() {
```
hexfusion: I might be misunderstanding intent but this doesn't seem necessary. since you have a defer on return from this function covering any panic which makes me nervous a little in itself. cant we just call callbackEvent(ctx, orgId, name, reason, message, err) at line 206

internal/service/event.go
```diff
@@ -424,7 +425,41 @@ func GetInternalTaskFailedEvent(ctx context.Context, resourceKind api.ResourceKi
 	}, log)
 }
 
-// Callbacks
+// GetResourceSyncTaskEvent creates an event for resourcesync task
+func GetResourceSyncTaskEvent(ctx context.Context, repoName string, commitHash string, totalChanges int, totalErrors int, updatedFields []api.ResourceUpdatedDetailsUpdatedFields, log logrus.FieldLogger) *api.Event {
+	statusFunc := lo.Ternary(totalErrors == 0,
```
hexfusion: I know you didn't introduce the use of lo.Ternary but go doesn't support it internally so I wonder why we introduce it to the project. Anyways this is very simple logic if else logic. I would appreciate if you did that.
```go
var status api.Status
if totalErrors == 0 {
	status = api.StatusOK()
} else {
	status = api.StatusInternalServerError(fmt.Sprintf("%d errors", totalErrors))
}
```

internal/service/event.go
```diff
@@ -477,3 +512,71 @@ func (h *ServiceHandler) eventDeleteCallback(ctx context.Context, resourceKind a
...
+		status := StoreErrorToApiStatus(err, false, api.FleetKind, &name)
+		h.CreateEvent(ctx, getBaseEvent(ctx, resourceEvent{
+			resourceKind:   api.FleetKind,
+			resourceName:   name,
+			reasonFailure:  reason,
+			outcomeFailure: func() string { return message },
+			status:         status,
+		}, nil))
+	}
+	h.CreateEvent(ctx, getBaseEvent(ctx, resourceEvent{
```
hexfusion, line 638: on error you would create 2 events is that intended?

internal/rollout/device_selection/reconciler.go
```diff
@@ -17,73 +20,83 @@ type Reconciler interface {
...
 		return
 	}
 	definitionUpdated, err := selector.IsDefinitionUpdated()
 	if err != nil {
-		r.log.WithError(err).Errorf("%v/%s: IsDefinitionUpdated", orgId, lo.FromPtr(fleet.Metadata.Name))
+		r.log.WithError(err).Errorf("%v/%s: IsDefinitionUpdated", orgId, fleetName)
 		return
 	}
 	if selector.IsRolloutNew() || definitionUpdated {
+		r.callEventCallbackCaller(ctx, eventCallback, store.NullOrgId, fleetName, api.EventReasonFleetRolloutCreated, "Fleet rollout created", nil)
```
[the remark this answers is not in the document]
gshilin-sdb: done

internal/rollout/device_selection/reconciler.go
```diff
@@ -167,6 +187,31 @@
...
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					log.Errorf("Event Callback panicked: %v", r)
+				}
+			}()
+			callbackEvent(ctx, orgId, name, reason, message, err)
+		}()
 	}
 }
```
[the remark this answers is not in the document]
gshilin-sdb: parameters to callbackEvent are different, so I cannot

internal/tasks/repotester.go
```diff
@@ -83,6 +87,38 @@ func (r *RepoTester) TestRepositories(ctx context.Context) {
...
+	if err != nil {
+		r.log.Errorf("Failed to update repository status for %s: %v", repoName, err)
+		return
+	}
+
+	// This event is about the error that was transferred as a parameter to SetAccessCondition,
+	// not about success or failure to access database.
+	// It should be emitted only in a case that Condition was changed.
+	r.callEventCallbackCaller(ctx, eventCallback, store.NullOrgId, repoName, false, nil, err)
+}
```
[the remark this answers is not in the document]
gshilin-sdb: done

internal/rollout/device_selection/reconciler.go
gshilin-sdb: it is now eventCallbackCaller

internal/rollout/device_selection/reconciler.go
gshilin-sdb: I'll change it, though yes, it was introduced before...

internal/rollout/device_selection/reconciler.go
gshilin-sdb: It is isolated, it's just declaration issue: `err` was declared on a (new) line 72, so there is no need to re-declare it and to shadow the original one.

internal/rollout/device_selection/reconciler.go
gshilin-sdb: added

internal/rollout/device_selection/reconciler.go
gshilin-sdb: Yep, that's too much.
Also consolidated three functions into one SafeEventCallback

internal/service/event.go
gshilin-sdb: just for you :)

internal/service/event.go
gshilin-sdb: oops

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,48 @@ components:
...
+        - commitHash
+        - totalChanges
+        - totalErrors
+        - errorMessages
+      properties:
+        detailType:
+          type: string
+          enum: [GitResourceChange]
+          description: The type of detail for discriminator purposes.
+        repoName:
```
avishayt: Not needed

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,48 @@ components:
...
+          type: string
+          description: Name of the repository.
+        commitHash:
+          type: string
+          description: Hash of the last commit.
+        totalChanges:
+          type: integer
+          description: Total number of changes this Git Resource introduces.
+        errorMessages:
```
avishayt: Details is for machines to understand, so better to have the number of errors rather than messages.

```suggestion
        totalErrors:
```

internal/store/common.go
```diff
@@ -324,27 +327,28 @@ func retryUpdate(fn func() (bool, error)) error {
 	return err
 }
 
-func callEventCallback(resourceKind api.ResourceKind, log logrus.FieldLogger) func(ctx context.Context, callbackEvent EventCallback, orgId uuid.UUID, name string, created bool, updateDesc *api.ResourceUpdatedDetails, err error) {
-	return func(ctx context.Context, callbackEvent EventCallback, orgId uuid.UUID, name string, created bool, updateDesc *api.ResourceUpdatedDetails, err error) {
-		// Call callback if provided (but don't fail the operation if callback fails)
-		if callbackEvent == nil {
-			return
+func GetAnnotation(fleet *api.Fleet, annotation string) (string, bool) {
```
avishayt: This method operates on an api-level object, so it should go into api/v1alpha1/util.go

internal/rollout/device_selection/reconciler.go
```diff
@@ -134,30 +140,37 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
...
 
 func (r *reconciler) Reconcile(ctx context.Context) {
+	const taskName = "rollout-device-selection"
+
+	reqid.OverridePrefix(taskName)
+	requestID := reqid.NextRequestID()
+	ctx = context.WithValue(ctx, middleware.RequestIDKey, requestID)
+	ctx = context.WithValue(ctx, consts.EventActorCtxKey, taskName)
```
avishayt: We need to make sure to set this for all tasks

internal/store/fleet.go
```diff
@@ -60,13 +59,30 @@ func NewFleet(db *gorm.DB, log logrus.FieldLogger) Fleet {
...
 	)
-	return &FleetStore{dbHandler: db, log: log, genericStore: genericStore, callEventCallbackCaller: callEventCallback(api.FleetKind, log)}
+	return &FleetStore{dbHandler: db, log: log, genericStore: genericStore}
 }
 
+func (s *FleetStore) callEventCallback(ctx context.Context, eventCallback EventFleetCallback, orgId uuid.UUID, name string, created bool, updateDesc *api.ResourceUpdatedDetails, isRolloutNew, isRolloutCompleted bool, err error) {
+	SafeEventCallback(s.log, func() {
+		eventCallback(ctx, orgId, name, created, updateDesc, isRolloutNew, isRolloutCompleted, err)
+	})
+}
```
avishayt: All these checks should be in the service level.  Just like devices - the callback in the service level gets the old and new fleet, and we check what changed there.

internal/tasks/repotester.go
```diff
@@ -53,6 +60,10 @@ func (r *RepoTester) TestRepositories(ctx context.Context) {
 			return
 		}
 
+		eventCallback := func(ctx context.Context, resourceKind api.ResourceKind, orgId uuid.UUID, name string, created bool, updates *api.ResourceUpdatedDetails, err error) {
+			service.EventRepositoryAccessible(ctx, resourceKind, orgId, name, created, updates, err, r.serviceHandler.CreateEvent)
+		}
```
avishayt: Do the same store/service pattern - ReplaceRepositoryStatus should have a callback that gets the old and new status, do a diff, and emit events.

api/v1alpha1/openapi.yaml
gshilin-sdb: removed

api/v1alpha1/openapi.yaml
gshilin-sdb: replaced

internal/store/common.go
gshilin-sdb: moved

internal/rollout/device_selection/reconciler.go
gshilin-sdb: ensured

internal/store/fleet.go
gshilin-sdb: moved

api/v1alpha1/openapi.yaml
```diff
@@ -5089,6 +5089,7 @@ components:
             - ResourceUpdateFailed
             - ResourceDeleted
             - ResourceDeletionFailed
+            - GitResourceChangeDetected
```
avishayt:
```suggestion
            - ResourceSyncCompleted
```

api/v1alpha1/openapi.yaml
```diff
@@ -5189,12 +5196,16 @@ components:
...
           InternalTaskFailed: "#/components/schemas/InternalTaskFailedDetails"
+          GitResourceChange: "#/components/schemas/GitResourceChangeDetails"
+          FleetRolloutStarted: "#/components/schemas/FleetRolloutStartedDetails"
       oneOf:
         - $ref: "#/components/schemas/ResourceUpdatedDetails"
         - $ref: "#/components/schemas/DeviceOwnershipChangedDetails"
         - $ref: "#/components/schemas/DeviceMultipleOwnersDetectedDetails"
         - $ref: "#/components/schemas/DeviceMultipleOwnersResolvedDetails"
         - $ref: "#/components/schemas/InternalTaskFailedDetails"
+        - $ref: "#/components/schemas/GitResourceChangeDetails"
```
avishayt:
```suggestion
        - $ref: "#/components/schemas/ResourceSyncCompletedDetails"
```

api/v1alpha1/openapi.yaml
```diff
@@ -5189,12 +5196,16 @@ components:
           DeviceMultipleOwnersDetected: "#/components/schemas/DeviceMultipleOwnersDetectedDetails"
           DeviceMultipleOwnersResolved: "#/components/schemas/DeviceMultipleOwnersResolvedDetails"
           InternalTaskFailed: "#/components/schemas/InternalTaskFailedDetails"
+          GitResourceChange: "#/components/schemas/GitResourceChangeDetails"
```
avishayt:
```suggestion
          ResourceSyncCompleted: "#/components/schemas/ResourceSyncCompletedDetails"
```

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,41 @@ components:
           additionalProperties:
             type: string
           description: Parameters needed to retry the task.
+    GitResourceChangeDetails:
```
avishayt:
```suggestion
    ResourceSyncCompletedDetails:
```

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,41 @@ components:
...
+      type: object
+      required:
+        - detailType
+        - commitHash
+        - totalChanges
+        - errorCount
+      properties:
+        detailType:
+          type: string
+          enum: [GitResourceChange]
```
avishayt:
```suggestion
          enum: [ResourceSyncCompleted]
```

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,41 @@ components:
...
+          description: The type of detail for discriminator purposes.
+        commitHash:
+          type: string
+          description: Hash of the last commit.
+        totalChanges:
+          type: integer
+          description: Total number of changes this Git Resource introduces.
+        errorCount:
+          type: integer
+          description: Number of errors discovered while applying this Git Resource changes.
```
avishayt:
```suggestion
        changeCount:
          type: integer
          description: Number of changes introduced by this ResourceSync update.
        errorCount:
          type: integer
          description: Number of errors encountered by this ResourceSync update.
```

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,41 @@ components:
...
+        - templateVersion
+        - isImmediate
+      properties:
+        detailType:
+          type: string
+          enum: [FleetRolloutStarted]
+          description: The type of detail for discriminator purposes.
+        templateVersion:
+          type: string
+          description: Version of the template to rollout.
```
avishayt:
```suggestion
          description: The name of the TemplateVersion that is rolling out.
```

api/v1alpha1/openapi.yaml
```diff
@@ -5299,3 +5310,41 @@ components:
...
+        isImmediate:
+          type: boolean
+          description: True for immediate rollout, false for a batch one.
```
avishayt: Change to enum named "rolloutType" with values "batched" and "immediate"

api/v1alpha1/util.go
```diff
@@ -376,3 +377,10 @@ func (r *RepositoryList) HideSensitiveData() error {
 	}
 	return nil
 }
+
+func GetAnnotation(fleet *Fleet, annotation string) (string, bool) {
```
avishayt: Even better - you can move this to api/v1alpha1/properties.go, and then call it via fleet.GetAnnotation(name) - you can even have more logic there if you need, like fleet.HasRolloutInProgress or fleet.HasCompletedRollout, or whatever you need there.

internal/service/event.go
```diff
@@ -324,6 +325,14 @@ func getEventType(reason api.EventReason) api.EventType {
 	return api.Normal
 }
 
+func getStatus(resourceKind api.ResourceKind, name *string, created bool, err error) api.Status {
+	if err != nil {
+		return StoreErrorToApiStatus(err, created, string(resourceKind), name)
+	} else {
+		return api.StatusOK()
```
avishayt: If created is true, shouldn't we return api.StatusCreated()?

internal/service/event.go
```diff
@@ -424,15 +433,70 @@ func GetInternalTaskFailedEvent(ctx context.Context, resourceKind api.ResourceKi
...
+	return getBaseEvent(ctx, resourceEvent{
+		resourceKind:   api.ResourceSyncKind,
+		resourceName:   lo.FromPtr(resourceSyncName),
+		reasonSuccess:  api.EventReasonGitResourceChangeDetected,
+		outcomeSuccess: successMessage,
+		reasonFailure:  api.EventReasonGitResourceChangeDetected,
+		outcomeFailure: func() string {
+			return failureMessage
+		},
+		status:        status,
```
avishayt, line 483: It's weird that getBaseEvent gets a status parameter since many events don't depend on status

internal/service/event.go
```diff
@@ -442,25 +506,74 @@ func (h *ServiceHandler) eventCallbackDevice(ctx context.Context, _ api.Resource
...
+		reasonSuccess:  api.EventReasonFleetRolloutBatchCompleted,
+		outcomeSuccess: "Fleet rollout batch completed",
+		reasonFailure:  api.EventReasonFleetRolloutBatchCompleted,
+		outcomeFailure: func() string { return "Fleet rollout batch completion failure" },
+		status:         status,
+	}, nil)
 }
 
-func (h *ServiceHandler) eventCallbackDeviceDecommission(ctx context.Context, _ api.ResourceKind, _ uuid.UUID, name string, _, _ *api.Device, created bool, updateDesc *api.ResourceUpdatedDetails, err error) {
+func (h *ServiceHandler) eventFleetCallback(ctx context.Context, _ api.ResourceKind, _ uuid.UUID, name string, oldResource, newResource interface{}, created bool, updatedDetails *api.ResourceUpdatedDetails, err error) {
```
avishayt, line 553: Why do we need updateDetails if we have the old and new?

internal/service/event.go
```diff
@@ -442,25 +506,74 @@ func (h *ServiceHandler) eventCallbackDevice(ctx context.Context, _ api.Resource
...
+		status := StoreErrorToApiStatus(err, created, api.FleetKind, &name)
+		h.CreateEvent(ctx, eventRolloutNew(ctx, name, oldFleet, newFleet, status))
+		h.CreateEvent(ctx, GetResourceCreatedOrUpdatedFailureEvent(ctx, created, api.FleetKind, name, status, updatedDetails))
 	} else {
-		h.CreateEvent(ctx, GetDeviceDecommissionedSuccessEvent(ctx, created, api.DeviceKind, name, updateDesc, nil))
+		status := api.StatusOK()
+		// Fleet cannot be in both states at the same time, so only one of the 2 following lines will emit event
+		h.CreateEvent(ctx, eventRolloutNew(ctx, name, oldFleet, newFleet, status))
+		h.CreateEvent(ctx, eventRolloutCompleted(ctx, name, oldFleet, newFleet, status))
+		h.CreateEvent(ctx, GetResourceCreatedOrUpdatedSuccessEvent(ctx, created, api.FleetKind, name, updatedDetails, nil))
```
avishayt: I don't understand these lines.
1. Why do we always send resource updated events?  It should only be if the owner/spec/labels changed, not annotations for example.
2. If there's an error, why only eventRolloutNew?

internal/service/event_benchmark_test.go
```diff
@@ -0,0 +1,281 @@
+// Benchmark tests for event creation and handling performance.
```
avishayt: No need for this file - we're not benchmarking events at this point

internal/service/event.go
gshilin-sdb: it needs it for Failure cases

internal/service/event.go
gshilin-sdb: updateDesc is calculated in a storage.generic, so why to re-calculated it?

api/v1alpha1/openapi.yaml
gshilin-sdb: renamed

api/v1alpha1/openapi.yaml
gshilin-sdb: renamed

api/v1alpha1/openapi.yaml
gshilin-sdb: renamed

api/v1alpha1/openapi.yaml
gshilin-sdb: renamed

api/v1alpha1/openapi.yaml
gshilin-sdb: renamed

api/v1alpha1/openapi.yaml
gshilin-sdb: renamed

api/v1alpha1/openapi.yaml
gshilin-sdb: done

api/v1alpha1/openapi.yaml
gshilin-sdb: done

api/v1alpha1/util.go
gshilin-sdb: moved

internal/service/event.go
gshilin-sdb: This func is called just in two places, so removed it.

internal/service/event.go
gshilin-sdb: 1. eventRolloutNew and eventRolloutCompleted return nil events in case nothing was changed and CreateEvent knows to ignore it.
2. added eventRolloutCompleted too

internal/service/event_benchmark_test.go
gshilin-sdb: deleted

internal/service/event.go
avishayt: I think we'll need to refactor this file, but we'll leave that for after

internal/service/event.go
avishayt: let's remove updateDetails from store in a followup PR

PR #1301, opened 2025-06-30
EDM-1712: Migrating HTTP Metrics 

SiddarthR56:
This is a draft commit to introduce service-level metrics for the HTTP API. Final metrics definitions are still under discussion and subject to change.

### DIFF
--- a/cmd/flightctl-api/main.go
+++ b/cmd/flightctl-api/main.go
@@ -144,13 +144,19 @@ func main() {
 		log.Fatalf("failed connecting to Redis queue: %v", err)
 	}
 
+	// Initialize HTTP collector for metrics
+	var httpCollector *metrics.HTTPCollector
+	if cfg.Prometheus != nil {
+		httpCollector = metrics.NewHTTPCollector(cfg)
+	}
+
 	// create the agent service listener as tcp (combined HTTP+gRPC)
 	agentListener, err := net.Listen("tcp", cfg.Service.AgentEndpointAddress)
 	if err != nil {
 		log.Fatalf("creating listener: %s", err)
 	}
 
-	agentserver := agentserver.New(log, cfg, store, ca, agentListener, provider, agentTlsConfig)
+	agentserver := agentserver.New(log, cfg, store, ca, agentListener, provider, agentTlsConfig, httpCollector)
 
 	go func() {
 		listener, err := middleware.NewTLSListener(cfg.Service.Address, tlsConfig)
@@ -158,7 +164,7 @@ func main() {
 			log.Fatalf("creating listener: %s", err)
 		}
 		// we pass the grpc server for now, to let the console sessions to establish a connection in grpc
-		server := apiserver.New(log, cfg, store, ca, listener, provider, agentserver.GetGRPCServer())
+		server := apiserver.New(log, cfg, store, ca, listener, provider, agentserver.GetGRPCServer(), httpCollector)
 		if err := server.Run(ctx); err != nil {
 			log.Fatalf("Error running server: %s", err)
 		}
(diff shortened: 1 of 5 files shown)


cmd/flightctl-api/main.go
```diff
@@ -173,8 +179,14 @@ func main() {
 	}()
 
 	if cfg.Prometheus != nil {
+		var collectors []metrics.NamedCollector
+		collectors = append(collectors, metrics.NewSystemCollector(ctx))
+		if httpCollector != nil {
```
asafbss, line 184: It would be better to add a nil check for the collector inside NewMetricsServer, instead of checking each collector before appending it to the slice.

https://github.com/flightctl/flightctl/blob/91318412a8aec2196d18d86fbb13a514c40f2cc8/internal/instrumentation/instrumentation.go#L36

The handler itself can remain as is and will simply return an empty response if no collectors are registered, which is fine.

https://github.com/flightctl/flightctl/blob/91318412a8aec2196d18d86fbb13a514c40f2cc8/internal/instrumentation/metrics/metrics.go#L106

internal/api_server/agentserver/server.go
```diff
@@ -137,9 +144,15 @@ func (s *AgentServer) prepareHTTPHandler(serviceHandler service.Service) (http.H
 		middleware.RequestID,
 		middleware.Logger,
 		middleware.Recoverer,
-		oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapiOpts),
 	}
 
+	// Add HTTP metrics middleware if available
+	if s.httpCollector != nil {
+		middlewares = append(middlewares, s.httpCollector.AgentServerMiddleware)
```
asafbss, line 151: It’s better to insert the middleware at position 0 in the slice, as in:
https://github.com/flightctl/flightctl/blob/863c7debfcf5eb0e72039177f235ccfde6c08305/internal/api_server/agentserver/server.go#L149

internal/api_server/server.go
```diff
@@ -163,14 +170,21 @@ func (s *Server) Run(ctx context.Context) error {
...
 		fcmiddleware.RequestSizeLimiter(s.cfg.Service.HttpMaxUrlLength, s.cfg.Service.HttpMaxNumHeaders),
 		fcmiddleware.RequestID,
 		fcmiddleware.AddEventMetadataToCtx,
 		middleware.Logger,
 		middleware.Recoverer,
-	)
+	}
+
+	// Add HTTP metrics middleware if available
+	if s.httpCollector != nil {
```
asafbss, line 183: Please verify its correctness:
https://github.com/flightctl/flightctl/blob/863c7debfcf5eb0e72039177f235ccfde6c08305/internal/api_server/server.go#L185

internal/instrumentation/metrics/http.go
```diff
@@ -0,0 +1,136 @@
...
+	sloMax float64
+
+	requestsTotal *prometheus.CounterVec
+	latencies     *prometheus.HistogramVec
+	errorsTotal   *prometheus.CounterVec
+}
+
+// NewHTTPCollector creates a new HTTP metrics collector with the provided configuration
+func NewHTTPCollector(cfg *config.Config) *HTTPCollector {
+	if cfg.Prometheus == nil {
```
asafbss, line 23: Unless NewHTTPCollector uses the config for internal configuration and behavior, it should be replaced with a ctx context.Context parameter as a best practice.

There’s no need to check cfg.Prometheus; the caller has already verified that and decided whether to call it.

test/util/util.go
```diff
@@ -119,7 +119,7 @@ func NewTestApiServer(log logrus.FieldLogger, cfg *config.Config, store store.St
 		return nil, nil, fmt.Errorf("NewTLSListener: error creating TLS certs: %w", err)
 	}
 
-	return apiserver.New(log, cfg, store, ca, listener, queuesProvider, nil), listener, nil
+	return apiserver.New(log, cfg, store, ca, listener, queuesProvider, nil, nil), listener, nil
```
asafbss, line 122: https://github.com/flightctl/flightctl/blob/863c7debfcf5eb0e72039177f235ccfde6c08305/test/util/util.go#L123

test/util/util.go
```diff
@@ -136,7 +136,7 @@ func NewTestAgentServer(log logrus.FieldLogger, cfg *config.Config, store store.
 		return nil, nil, fmt.Errorf("NewTestAgentServer: error creating TLS certs: %w", err)
 	}
 
-	return agentserver.New(log, cfg, store, ca, listener, queuesProvider, tlsConfig), listener, nil
+	return agentserver.New(log, cfg, store, ca, listener, queuesProvider, tlsConfig, nil), listener, nil
```
asafbss, line 139: https://github.com/flightctl/flightctl/blob/863c7debfcf5eb0e72039177f235ccfde6c08305/test/util/util.go#L142

asafbss (commented):
See comments

PR #1302, opened 2025-06-30
EDM-1659: improve e2e duration part 1

siserafin:
Improving e2e image building adding a parameter to set multiple threads.


test/scripts/agent-images/README.md
```diff
@@ -8,6 +8,20 @@ directory.
 
 And can be triggered from the top-level makefile with: `make e2e-agent-images`
 
+## Build Process
+
+The script builds images in two phases:
+1. **Base image**: Built first since other images depend on it
+2. **Remaining images**: Built in parallel to speed up the process
+
+The number of parallel build jobs can be controlled with the `PARALLEL_JOBS`
```
eldar101, line 17: any error handling for PARALLEL_JOBS? as in not entering a negative number. IS there a limit for the number of jobs?

siserafin: added some validations, thanks. I think we are going to be the only users :)

test/scripts/agent-images/create_agent_images.sh
```diff
@@ -27,35 +28,71 @@ if [ -n "${FLIGHTCTL_RPM:-}" ]; then
     BUILD_ARGS="${BUILD_ARGS} --build-arg=RPM_PACKAGE=${RPM_PACKAGE}"
 fi
 
+build_single_image() {
```
hexfusion, line 31: note: just thinking out loud, the level of complexity we're managing in bash is becoming unmaintainable. would it make sense to move the test logic and orchestration into a separate go module and use the podman bindings/os.exec directly? that would let us use go templates for config generation, structure things more cleanly, and write proper unit tests. 

to be clear the intent is not to block this great work :)

hexfusion: cc @kkyrazis, @SiddarthR56 @sk-ilya

siserafin: Thanks for sharing. I see your point of moving the logic to go. For now I’ve tried to keep things minimal and just made changes within the bounds of the existing bash setup. Switching to a go module feels like a bigger shift, and I’m not sure it’s something I’m in a position to take on at the moment.
Totally open to discussing this more if it’s something we want to prioritize as a team, but just wanted to clarify that my changes were limited to existing functions without introducing new structure or tools.

hexfusion: just food for future thought

amir-yogev-gh:
@siserafin another idea worth considering:
Reduce the tests to the bare minimal required (unit + integration + smoke + selected E2E only?). In other words - skip many tests.
This will buy us the time to optimize running the tests in parallel or any other constructive idea.
I think the goal should be run time below ~25 minutes. Based on current run: 2m + 7m + 10m + ?.

siserafin:
Yes, @amir-yogev-gh, I agree, the test duration is long. We're already running a subset of the E2E tests labeled "sanity", but I think we should reduce it even further. Since the full suite runs daily in the QE CI anyway, we can afford to keep this run lighter for faster feedback. Developers can always run specific tests locally for a quick check, if those tests aren’t included in the "sanity"

amalykhi (approved):

PR #1312, opened 2025-07-02
EDM-1553: The owner is not set in a fleet added by a resource sync

ori-amizur:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the behavior of fleet replacement to ensure certain fields are only cleared for non-internal requests.

* **Tests**
  * Added integration tests to verify correct handling of fleet owner fields based on request type (internal vs. non-internal).

### DIFF
--- a/api/v1alpha1/util.go
+++ b/api/v1alpha1/util.go
@@ -322,3 +322,57 @@ func GetNextDeviceRenderedVersion(annotations map[string]string) (string, error)
 	currentRenderedVersion++
 	return strconv.FormatInt(currentRenderedVersion, 10), nil
 }
+
+type SensitiveDataHider interface {
+	HideSensitiveData() error
+}
+
+func hideValue(value *string) {
+	if value != nil {
+		*value = "*****"
+	}
+}
+
+func (r *Repository) HideSensitiveData() error {
+	if r == nil {
+		return nil
+	}
+	spec := r.Spec
+
+	_, err := spec.GetGenericRepoSpec()
+	if err != nil {
+		gitHttpSpec, err := spec.GetHttpRepoSpec()
+		if err == nil {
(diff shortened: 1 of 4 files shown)


internal/service/fleet.go
```diff
@@ -63,14 +63,16 @@ func (h *ServiceHandler) GetFleet(ctx context.Context, name string, params api.G
 	return result, StoreErrorToApiStatus(err, false, api.FleetKind, &name)
 }
 
-func (h *ServiceHandler) ReplaceFleet(ctx context.Context, name string, fleet api.Fleet) (*api.Fleet, api.Status) {
+func (h *ServiceHandler) ReplaceFleet(ctx context.Context, name string, fleet api.Fleet, isInternal bool) (*api.Fleet, api.Status) {
```
avishayt: Why are you passing isInternal explicitly rather than using IsInternalRequest(ctx)?

avishayt: Also make sure this bug is covered in our tests so it doesn't return.

avishayt (changes requested):

internal/service/fleet.go
ori-amizur: OK

avishayt (changes requested):
You need to pass the value to CreateOrUpdate as well

ori-amizur:
> You need to pass the value to CreateOrUpdate as well

Which value ?

avishayt:
> > You need to pass the value to CreateOrUpdate as well
> 
> Which value ?

See how "isNotInternal" is used in device.go (it's passed to CreateOrUpdate).

ori-amizur:
> See how "isNotInternal" is used in device.go (it's passed to CreateOrUpdate).

Why it is not using the context ?

PR #1313, opened 2025-07-02
EDM-1792: agent: implement prefetch manager

hexfusion:
## Summary by CodeRabbit

* **New Features**
  * Introduced an advanced OCI image and artifact prefetching system for improved update readiness and reliability.
  * Added configuration options for image pull timeouts and retry steps.
  * Integrated prefetching into the device update workflow with explicit handling for ongoing prefetch states.
  * Added OCI target collection interfaces to application and OS managers for extensible prefetching.

* **Bug Fixes**
  * Improved error handling and retry logic for image prefetching scenarios, including rollback support on prefetch not ready errors.

* **Refactor**
  * Modularized application and OS image handling by separating OCI target collection and verification from image pulling.
  * Simplified update and prefetch logic for clearer flow and maintainability.
  * Updated backoff logic to rely solely on context cancellation for timeouts.

* **Tests**
  * Added comprehensive unit tests for the OCI prefetching system, including scheduling, readiness, status reporting, and cleanup.
  * Updated existing device and application tests to cover interactions with the prefetch manager.

* **Chores**
  * Added and updated mock implementations for prefetch manager and OCI collector interfaces to support new tests.
  * Automated mock generation for dependency package interfaces.


internal/agent/client/client.go
```diff
@@ -172,15 +173,8 @@ func authFromSpec(log *log.PrefixLogger, device *v1alpha1.DeviceSpec, authPath s
 type ClientOption func(*clientOptions)
 
 type clientOptions struct {
-	retry          bool
 	pullSecretPath string
-}
-
-// WithRetry enables enables retry based on the backoff config provided.
-func WithRetry() ClientOption {
```
hexfusion, line 180: note: retry is now assumed

internal/agent/device/applications/manager_test.go
```diff
@@ -219,122 +244,6 @@ func TestManagerRemoveApplication(t *testing.T) {
 	require.False(manager.podmanMonitor.isRunning())
 }
 
-func TestBeforeUpdate(t *testing.T) {
```
hexfusion, line 222: note: most of this test logic was moved to the dependency package

internal/agent/device/applications/manager.go
```diff
@@ -43,7 +43,6 @@ func NewManager(
 	}
 }
 
-// Add an application to be managed
```
hexfusion, line 46: note: this is an implementation of an interface and interface is already doced

internal/agent/device/spec/queue.go
```diff
@@ -202,13 +202,14 @@ func (m *queueManager) getOrCreateRequeueState(ctx context.Context, version int6
 }
 
 func (m *queueManager) shouldEnforceDelay(state *requeueState) bool {
-	if !m.queue.IsEmpty() || state.tries == 0 {
```
hexfusion, line 205: note: this change is required to ensure that during rollback flow the retryable version can be backedoff

hexfusion:
>  Error: deleting fleet/default: 409 Conflict (409)v

e2e flake

internal/agent/client/podman.go
```diff
@@ -93,12 +93,9 @@ func (p *Podman) Pull(ctx context.Context, image string, opts ...ClientOption) (
 	}
 
 	return logProgress(ctx, p.log, "Pulling image, please wait...", func(ctx context.Context) (string, error) {
-		if options.retry {
-			return retryWithBackoff(ctx, p.log, p.backoff, func(ctx context.Context) (string, error) {
-				return p.pullImage(ctx, image, options.pullSecretPath)
-			})
-		}
-		return p.pullImage(ctx, image, options.pullSecretPath)
```
hexfusion, line 101: note: still working on pull timeout logic

internal/agent/client/compose.go
```diff
@@ -25,7 +25,6 @@ const (
 	ComposeOverrideFilename      = "99-compose-flightctl-agent.override.yaml"
 	ComposeDockerProjectLabelKey = "com.docker.compose.project"
 	defaultPodmanTimeout         = 10 * time.Minute
-	defaultPodmanMaxRetryTimeout = 60 * time.Minute
```
hexfusion, line 28: note: simplified backoff to not require timeout and use only max steps

internal/agent/device/dependency/dependency.go
```diff
@@ -1,20 +1,406 @@
...
+
+// OCICollectorFn defines a function that collects OCI targets from device specs
+type OCICollectorFn func(ctx context.Context, current, desired *v1alpha1.DeviceSpec) ([]OCIPullTarget, error)
+
+var _ PrefetchManager = (*prefetchManager)(nil)
+
+// PrefetchManager orchestrates OCI target collection and prefetching
+type PrefetchManager interface {
+	// RegisterOCICollector registers a function that can collect OCI targets from a device spec
+	RegisterOCICollector(collectorFn OCICollectorFn)
```
kkyrazis: Why the preference for a CollectorFn as the registration item instead of the `OCICollector` interface defined below (assuming that the  collector interface was changed slightly to return the targets and an error)

internal/agent/device/device.go
```diff
@@ -317,10 +330,13 @@ func (a *Agent) beforeUpdate(ctx context.Context, current, desired *v1alpha1.Dev
 		}
 	}
 
+	a.prefetchManager.RegisterOCICollector(a.appManager.CollectOCITargets(ctx))
 	if a.specManager.IsOSUpdate() {
-		if err := a.osManager.BeforeUpdate(ctx, current.Spec, desired.Spec); err != nil {
-			return fmt.Errorf("os: %w", err)
-		}
+		a.prefetchManager.RegisterOCICollector(a.osManager.CollectOCITargets(ctx))
```
kkyrazis: Won't this register the same collectors multiple times? I could have missed it, but I don't see where collectors are being removed in the prefetchManager.

internal/agent/device/dependency/dependency.go
```diff
@@ -1,20 +1,406 @@
...
+	for _, task := range m.tasks {
+		if task.cancelFn != nil {
+			task.cancelFn()
+		}
+		// fire cleanups
+		for _, cleanup := range task.cleanupFns {
+			cleanup()
+		}
+	}
+	m.mu.Unlock()
```
kkyrazis: should the tasks also be removed during cleanup?

internal/agent/device/dependency/dependency.go
```diff
@@ -1,20 +1,406 @@
...
+			return fmt.Errorf("prefetch schedule: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (m *prefetchManager) schedule(ctx context.Context, image string, ociType OCIType, cleanupFns []func(), opts ...client.ClientOption) error {
+	m.mu.Lock()
+	if _, exists := m.tasks[image]; exists {
```
kkyrazis: In the event that a spec has `maxQueueSize + 1` images/artifacts to pull this check could cause an event to not get send to the download worker
 
(assume slow download times)
1. have 11 images in a spec
2. the first 10 get scheduled just fine
3. number 11 gets a task registered (line 293)
4. timer elapses and the image isn't enqueued (line 306)
5. retry eventually
6. try to schedule 11, task already exists (line 269), never sent to the worker thread.

internal/agent/device/dependency/dependency.go
```diff
@@ -1,20 +1,406 @@
...
+
+	for {
+		if ctx.Err() != nil {
+			m.setResult(target, fmt.Errorf("pulling oci target %s: %w", target, ctx.Err()))
+			m.log.Warnf("Context error pulling oci target: %s", target)
+			return
+		}
+		if err := m.pull(ctx, target, task); err != nil {
+			if errors.IsRetryable(err) {
+				time.Sleep(queueNextItemAfter)
```
kkyrazis: minor: is a:
```go
select {
<-time.After(queueNextItemAfter)
<-ctx.Done()
}
```
preferable here to save those potential sweet sweet seconds of sleeping

internal/agent/device/dependency/dependency.go
```diff
@@ -1,20 +1,406 @@
...
+		// fire cleanups
+		for _, cleanup := range task.cleanupFns {
+			cleanup()
+		}
+	}
+	m.mu.Unlock()
+
+	for {
+		select {
+		case <-m.queue:
```
kkyrazis, line 373: I _think_ there is a race condition here between the worker routine and Cleanup where the worker routine could set a cancelFunc on a task after we have already cleaned up the existing tasks. In the real flow I don't think it is likely to be hit.

internal/agent/device/dependency/dependency.go
hexfusion: yeah good point the callbackFn pattern I have used in other areas but its also pretty clean to add the manager directly

hexfusion: updated

internal/agent/device/dependency/dependency.go
hexfusion: sleep was lazy

internal/agent/device/dependency/dependency.go
hexfusion: good catch looking

hexfusion: cleaned up task

internal/agent/device/dependency/dependency.go
hexfusion: hmm will look at some closed boolish signal

hexfusion: fixed thanks

internal/agent/device/dependency/dependency.go
```diff
@@ -1,20 +1,425 @@
...
+}
+
+func (m *prefetchManager) worker(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			m.log.Debugf("Prefetch worker exiting: %v", ctx.Err())
+			return
+
+		case image := <-m.queue:
```
kkyrazis, line 118: what are your thoughts on parallelizing this? Is it just safer to download a single item at a time?

internal/agent/device/device.go
```diff
@@ -514,6 +530,7 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1alpha1.Device, s
 		conditionUpdate.Reason = string(v1alpha1.UpdateStateError)
 		conditionUpdate.Message = log.Truncate(msg, status.MaxMessageLength)
 		conditionUpdate.Status = v1alpha1.ConditionStatusFalse
+		a.prefetchManager.Cleanup()
```
kkyrazis, line 533: I'm just putting this comment here even though it's not particularly tied to it. The current behavior and the new behavior is the exact same with respect to the following flow:

1. spec 2
2. start spec 2 update
3. download images
4. spec 3
5. continue downloading images
6. proceed with upgrade of spec 2
7. download spec 3 images
8. proceed with upgrade of spec 3

With this new flow we could abort downloading images that may no longer be required. Is that a future item that will be added?

internal/agent/device/dependency/dependency.go
hexfusion: I am a little nervous about network and storage I/O so I wanted to keep it serial for starters but you're right to ask.

internal/agent/device/device.go
hexfusion: we need to do a few followups highlevel
- what you describe and general polishing
- come up with a flow to prune unused images
- provide `dependency` and maybe others a way to validate no resource alerts are firing (disk) before we pull.

internal/agent/config/config.go
```diff
@@ -98,6 +103,12 @@ type Config struct {
 	// SystemInfoTimeout is the timeout for collecting system info.
 	SystemInfoTimeout util.Duration `json:"system-info-timeout,omitempty"`
 
+	// PullTimeout is the max duration a single OCI target will try to pull.
+	PullTimeout util.Duration `json:"pull-timeout,omitempty"`
+
+	// PullRetrySteps defines how many retry attempts are allowed for pulling an OCI target.
+	PullRetrySteps int `json:"pull-retry-steps,omitempty"`
```
hexfusion, line 110: TODO: update docs

PR #1314, opened 2025-07-03
EDM-342: Implement Helm post-delete cleanup job

avishayt:
This commit adds a Kubernetes Job that runs after Helm uninstall to delete leftover namespace-scoped resources such as Secrets and PVCs that Helm does not remove automatically (e.g., StatefulSet PVCs, manually created Secrets).

The cleanup job is defined as a `post-delete` Helm hook. It is granted the necessary permissions via a Role and RoleBinding, also defined as post-delete hooks with `hook-delete-policy: before-hook-creation` to ensure they persist long enough for the job to execute.

The cleanup script avoids deleting any resource explicitly labeled to be kept, e.g., `helm.sh/resource-policy: keep`.

This ensures a clean uninstall, with only explicitly retained resources left behind.


celdrake (approved):
LGTM. There were not leftover PVCs or other resources after uninstalling the helm chart.

PR #1317, opened 2025-07-03
EDM-1179: Prevent ER application when device exists

noga-magen:


internal/service/enrollmentrequest.go
```diff
@@ -327,16 +322,11 @@ func (h *ServiceHandler) ReplaceEnrollmentRequestStatus(ctx context.Context, nam
...
 func (h *ServiceHandler) allowCreationOrUpdate(ctx context.Context, orgId uuid.UUID, name string) error {
 	device, err := h.store.Device().Get(ctx, orgId, name)
-	if err != nil {
-		// Device not found, allow creation or update
-		if errors.Is(err, flterrors.ErrResourceNotFound) {
-			return nil
-		}
-		return fmt.Errorf("failed to get device: %w", err)
+	if errors.Is(err, flterrors.ErrResourceNotFound) {
+		return nil // Device not found: allow create or update
```
eldar101, line 326: shouldn't it return an error for not finding the device?

noga-magen: No - this is exactly what we want.
It's an error that the resource does not exist - not that the function failed (:

noga-magen:
Why allowCreationOrUpdate Was Removed from CreateEnrollmentRequest

CreateEnrollmentRequest — Used by the agent when it first tries to enroll.
ReplaceEnrollmentRequest — Used by the CLI for manually managing enrollments.

What Happened in the Failing Test:
The agent sends a request to enroll (CreateEnrollmentRequest).
The agent goes offline (e.g. network issue, crash).
While it’s offline, an admin approves the request — this creates the device in the database.
The agent comes back online and tries to enroll again using the same request.

The Problem: 
Database Locking and Deadlocks
Database operations: 2 simple INSERTs
INSERT INTO [enrollment_requests](https://github.com/flightctl/flightctl/blob/1cb8fb24e84adef161d08d8b55f79eed3d9a0992/internal/service/enrollmentrequest.go#L140)
INSERT INTO [events](https://github.com/flightctl/flightctl/blob/1cb8fb24e84adef161d08d8b55f79eed3d9a0992/internal/service/enrollmentrequest.go#L142)

Broken Flow (with allowCreationOrUpdate added):

the allowCreationOrUpdate uses this query: SELECT FROM devices WHERE name = ? ← NEW: Creates lock contention
INSERT INTO enrollment_requests (Original)
INSERT INTO events (Original)

Creates:
- Lock contention
- Deadlocks
- Context timeouts
- Server crashes

The agent client already has proper [conflict resolution](https://github.com/flightctl/flightctl/blob/1cb8fb24e84adef161d08d8b55f79eed3d9a0992/internal/agent/client/enrollment.go#L51)
No need to pre-check the device table. The system is already safe and resilient without it.

Lock Contention Chain:
Extra Device Query: allowCreationOrUpdate() queries devices table
Cross-Table Locks: Creates dependencies between devices and enrollment_requests tables
Transaction Deadlock: Concurrent operations on both tables create circular waits
Context Cancellation: Database timeouts cancel contexts
Server Shutdown: Cascading failures shut down test infrastructure

Error Sequence:
Database Deadlock → Query Timeouts → Context Canceled → Server Shutdown → Test Failure

lkatalin:
Thanks @noga-magen , your detailed write-up above is very helpful.

> The agent comes back online and tries to enroll again using the same request.

This is not related to your bug fix, but I have questions about this being correct functionality (ex. could the device poll for its agent/mgmt certificate before deciding to submit an ER?). I will bring that up elsewhere. If we decide to change that behavior, the integration test would likely pass with `allowCreateorUpdate` called in the `Create` function, so it's possible we may want to put it back in the future.

However, this PR looks fine to me. Also after testing the branch locally I have verified that it does fix the bug. We now get the following error when submitting an ER that has a pre-existing device:

```
[lily@fedora flightctl]$ bin/flightctl apply -f examples/enrollmentrequest-waiting.yaml 
enrollmentrequest: applying examples/enrollmentrequest-waiting.yaml/mul6f5j1qiak0nsm9bl4n3f32bo4eub6oemntql56daoupdli2n0: 400 Bad Request
{"apiVersion":"v1alpha1","code":400,"kind":"Status","message":"a resource with this name already exists","reason":"Bad Request","status":"Failure"}

Error: enrollmentrequest: failed to apply examples/enrollmentrequest-waiting.yaml/mul6f5j1qiak0nsm9bl4n3f32bo4eub6oemntql56daoupdli2n0: 400 Bad Request
```

This is what I would expect to happen.

lkatalin (approved):

lkatalin:
This will need to be rebased on main but otherwise it seems fine to merge.

PR #1319, opened 2025-07-03
EDM-1204: Support for printing unlimited number of resources in table mode by fetching in batches

sk-ilya:
## Summary by CodeRabbit

* **New Features**
  * Added support for client-side batching and pagination when listing large numbers of resources in table output format.
  * Enforced a maximum limit for list requests, with special handling for table output to allow batching beyond the server-side maximum.

* **Bug Fixes**
  * Improved error handling during table formatting to ensure errors are reported immediately.

* **Tests**
  * Introduced comprehensive tests for list batching, pagination, and output formatting behaviors.

### DIFF
--- a/internal/cli/display/table.go
+++ b/internal/cli/display/table.go
@@ -32,11 +32,15 @@ func (f *TableFormatter) Format(data interface{}, options FormatOptions) error {
 		err = f.formatSingle(w, data, options)
 	}
 
+	if err != nil {
+		return err
+	}
+
 	// ensure that after this call headers are suppressed on subsequent calls
 	// noHeaders is set to true after the first successful formatting run
 	f.noHeaders = true
 
-	return err
+	return nil
 }
 
 // formatList handles formatting for list endpoints (TYPE)
--- a/internal/cli/get.go
+++ b/internal/cli/get.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"reflect"
 	"slices"
 	"strings"
 
@@ -19,9 +20,9 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var (
-	legalOutputTypes = []string{string(display.JSONFormat), string(display.YAMLFormat), string(display.NameFormat), string(display.WideFormat)}
-)
+var legalOutputTypes = []string{string(display.JSONFormat), string(display.YAMLFormat), string(display.NameFormat), string(display.WideFormat)}
+
+const maxRequestLimit = 1000 // At most the server side constraint
 
 type GetOptions struct {
 	GlobalOptions
@@ -79,7 +80,7 @@ func (o *GetOptions) Bind(fs *pflag.FlagSet) {
 	fs.StringVarP(&o.LabelSelector, "selector", "l", o.LabelSelector, "Selector (label query) to filter on, supporting operators like '=', '!=', and 'in' (e.g., -l='key1=value1,key2!=value2,key3 in (value3, value4)').")
 	fs.StringVar(&o.FieldSelector, "field-selector", o.FieldSelector, "Selector (field query) to filter on, supporting operators like '=', '==', and '!=' (e.g., --field-selector='key1=value1,key2!=value2').")
 	fs.StringVarP(&o.Output, "output", "o", o.Output, fmt.Sprintf("Output format. One of: (%s).", strings.Join(legalOutputTypes, ", ")))
-	fs.Int32Var(&o.Limit, "limit", o.Limit, "The maximum number of results returned in the list response.")
+	fs.Int32Var(&o.Limit, "limit", o.Limit, "The maximum number of results returned in the list response. If the value is 0, then the result is not limited.")
 	fs.StringVar(&o.Continue, "continue", o.Continue, "Query more results starting from the value of the 'continue' field in the previous response.")
 	fs.StringVar(&o.FleetName, "fleetname", o.FleetName, "Fleet name for accessing templateversions (use only when getting templateversions).")
 	fs.BoolVar(&o.Rendered, "rendered", false, "Return the rendered device configuration that is presented to the device (use only when getting a single device).")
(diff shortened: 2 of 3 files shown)


internal/cli/display/table.go
```diff
@@ -32,11 +32,15 @@ func (f *TableFormatter) Format(data interface{}, options FormatOptions) error {
...
+	if err != nil {
+		return err
+	}
+
 	// ensure that after this call headers are suppressed on subsequent calls
 	// noHeaders is set to true after the first successful formatting run
 	f.noHeaders = true
 
-	return err
+	return nil
```
hexfusion, line 43: :+1:

internal/cli/get.go
```diff
@@ -157,16 +161,90 @@ func (o *GetOptions) Run(ctx context.Context, args []string) error {
...
+
+	return o.handleSingle(ctx, formatter, clientWithResponses, kind, name)
+}
+
+// handleSingle fetches and displays a single resource.
+func (o *GetOptions) handleSingle(ctx context.Context, formatter display.OutputFormatter, c *apiclient.ClientWithResponses, kind, name string) error {
+	response, err := o.getSingleResource(ctx, c, kind, name)
+	errMsg := fmt.Sprintf("reading %s/%s", kind, name)
+	if err != nil {
+		return fmt.Errorf(errMsg+": %w", err)
```
hexfusion: for all instances
```suggestion
		return fmt.Errorf("%s: %w", errMsg, err)
```

internal/cli/get.go
```diff
@@ -157,16 +161,90 @@ func (o *GetOptions) Run(ctx context.Context, args []string) error {
...
+}
+
+// handleList fetches and displays a list of resources, optionally processing the list in batches
+func (o *GetOptions) handleList(ctx context.Context, formatter display.OutputFormatter, c *apiclient.ClientWithResponses, kind string) error {
+	// Batching is only supported for table output (i.e. when no --output flag is provided)
+	isTableOutput := o.Output == ""
+	requestedLimit := int(o.Limit) // keep a copy of user-requested limit
+	shouldLoadInBatches := isTableOutput && (requestedLimit == 0 || requestedLimit > maxRequestLimit)
+
+	errMsg := fmt.Sprintf("listing %s", plural(kind))
```
hexfusion: nit: why allocate something we may not need?

internal/cli/get.go
```diff
@@ -157,16 +161,90 @@ func (o *GetOptions) Run(ctx context.Context, args []string) error {
...
+		if err != nil {
+			return fmt.Errorf(errMsg+": %w", err)
+		}
+		if err = validateResponse(response, errMsg); err != nil {
+			return fmt.Errorf(errMsg+": %w", err)
+		}
+
+		if err = o.displayResponse(formatter, response, kind, ""); err != nil {
+			return fmt.Errorf(errMsg+": %w", err)
+		}
```
hexfusion: if you encapsolated all of this logic you could do
```go
 if err  := functionOrMethod(ctx. kind, respond...);err != nil {
      return  fmt.Errorf("listing %s: %w", plural(kind), err)
 }
```

internal/cli/get.go
```diff
@@ -294,12 +363,14 @@ func (o *GetOptions) processResponse(response interface{}, err error, kind strin
 			if err := json.Unmarshal(responseBody, &dest); err != nil {
 				return fmt.Errorf("unmarshalling error: %w", err)
 			}
-			return fmt.Errorf(errorPrefix+": %d, message: %s", httpResponse.StatusCode, dest.Message)
+			return fmt.Errorf(msg+": %d, message: %s", httpResponse.StatusCode, dest.Message)
```
hexfusion:
```suggestion
			return fmt.Errorf("%s: %d, message: %s", msg, httpResponse.StatusCode, dest.Message)
```

internal/cli/get.go
```diff
@@ -321,3 +392,52 @@ func (o *GetOptions) processResponse(response interface{}, err error, kind strin
 	// For table format, pass the full response
 	return formatter.Format(response, options)
 }
+
+// getListMetadata extracts the ListMeta and the number of items from a list response.
+func getListMetadata(response interface{}) (api.ListMeta, int, error) {
```
hexfusion: question: why we are returning the owned object instead of a pointer

internal/cli/get.go
```diff
@@ -321,3 +392,52 @@ func (o *GetOptions) processResponse(response interface{}, err error, kind strin
...
+	var zeroMeta api.ListMeta
+
+	json200, err := responseField[interface{}](response, "JSON200")
+	if err != nil {
+		return zeroMeta, 0, err
```
hexfusion: if this is a pointer on error you could do like below. in golang the response object should never be consumed if err != nil so nil is safe and idiomatic.
```suggestion
		return nil, 0, err
```

internal/cli/get.go
sk-ilya: To avoid repeating `return fmt.Errorf("listing %s: %w", plural(kind), err)` four times in the following for loop. Is there a better way?

internal/cli/get.go
sk-ilya: Fixed

internal/cli/get.go
sk-ilya: Added a `listOnce` method that handles ther get -> validate -> display chain for lists.

internal/cli/get.go
sk-ilya: UPD: reorganized the batch listing method a bit

internal/cli/get.go
```diff
@@ -157,16 +161,97 @@ func (o *GetOptions) Run(ctx context.Context, args []string) error {
...
+			return fmt.Errorf("reading list metadata: %w", err)
+		}
+
+		printedCount += int32(listItemsCount) // #nosec G115
+
+		if requestedLimit > 0 && printedCount >= requestedLimit {
+			break // Reached user-requested limit
+		}
+
+		if listMetadata.RemainingItemCount == nil || *listMetadata.RemainingItemCount == 0 {
```
avishayt: It's safer to loop as long as listMatadata.Continue isn't nil

internal/cli/get.go
```diff
@@ -321,3 +399,33 @@ func (o *GetOptions) processResponse(response interface{}, err error, kind strin
...
+	}
+
+	// Retrieve metadata
+	listMeta, err := responseField[api.ListMeta](json200, "Metadata")
+	if err != nil {
+		return nil, 0, err
+	}
+
+	// Determine number of items via reflection
+	v := reflect.ValueOf(json200)
```
hexfusion, line 417: nit: whenever I see the use of `reflect` I wonder is this really needed? Is there no other way to get the data we need from the response? If we have explored that and using reflection is the only way then that is fine. But please explore. high-level reflection is very expensive in this context it might not be a big deal but its good practice to try and avoid.

PR #1321, opened 2025-07-06
Edm 1783 keep rpms on github2

gshilin-sdb:
- Add copr-to-gh_pages-download.sh script for downloading RPMs from COPR
- Add create-github-pages.sh script for generating GitHub Pages structure
- Add GitHub Actions workflow for automated RPM repository updates
- Support multiple versions and platforms
- Generate beautiful responsive web interface
- Add temporary directories to .gitignore"


.gitignore
```diff
@@ -10,3 +10,10 @@ test/e2e/cli/*.key
 /packaging/rpm/flightctl-*-vendor.tar.bz2
 flightctl-*.rpm
 *.tar
+
+# COPR to GitHub Pages temporary directories
+copr-rpms-temp/
+pages-content/
```
fzdarsky, line 16: Could we start moving all build artifacts below `/output` and ignore that dir only? The repo's root is really starting to get messy. All other artifacts (bins, rpms, test artifacts, ...) should move there, too, eventually.

.github/workflows/update-rpm-repo.yml
```diff
@@ -1,436 +1,218 @@
...
-          find copr-rpms-temp -name "*debuginfo*.rpm" -delete
-          find copr-rpms-temp -name "*debugsource*.rpm" -delete
-          find copr-rpms-temp -name "*.src.rpm" -delete
+          echo "=== Creating GitHub Pages structure ==="
           
-          echo "Files after cleanup:"
-          find copr-rpms-temp -name "*.rpm" | head -10
-          echo "RPMs remaining: $(find copr-rpms-temp -name "*.rpm" | wc -l)"
+          # Make the script executable
+          chmod +x ./hack/create-github-pages.sh
```
fzdarsky, line 160: Shouldn't be necessary. Just commit the script with exec permissions already. Or fix with:
`git update-index --chmod=+x hack/create-github-pages.sh`

hack/copr-to-gh_pages-download.sh
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env bash
+
+# COPR to GitHub Pages RPM Upload Script
+# This script finds COPR builds, downloads RPMs, and prepares them for GitHub Pages
+
+set -euo pipefail
+
+# Configuration
+COPR_PROJECT="@redhat-et/flightctl"
+DEST_DIR="copr-rpms-temp"
```
fzdarsky, line 10:
```suggestion
DEST_DIR="_output/copr-rpms-temp"
```

hack/copr-to-gh_pages-download.sh
```diff
@@ -0,0 +1,323 @@
...
+            if curl -f -L --connect-timeout 20 --max-time 180 -o \"\$local_file\" \"\$file_url\" 2>/dev/null; then
+                echo \"[$chroot]   ✅ \$rpm_file\"
+                exit 0
+            else
+                if [ \$retry -eq $MAX_RETRIES ]; then
+                    echo \"[$chroot]   ❌ Failed: \$rpm_file\"
+                fi
+                sleep 2
+            fi
+        done
```
fzdarsky, line 117: Use curl's built-in retries instead?
`curl --retry $MAX_RETRIES --retry-all-errors --retry-max-time 120 ...`
If you want to be less aggressive on errors, could also be `--retry-connrefused`.

Let's also consider whether we need the variants with `copr-cli` or hybrid to download. Just seems to add unnecessary complexity.

hack/create-github-pages.sh
```diff
@@ -0,0 +1,593 @@
...
+
+cat > "$PAGES_OUTPUT_DIR/flightctl-fedora.repo" << EOF
+[flightctl]
+name=FlightCtl RPM Repository (Fedora)
+baseurl=https://$REPO_OWNER.github.io/$REPO_NAME/fedora-\$releasever-\$basearch/
+enabled=1
+gpgcheck=0
+repo_gpgcheck=0
+type=rpm
+metadata_expire=1d
```
fzdarsky, line 112: The `.repo` file is missing the GPG check.  The `baseurl` should contain the `$releasever`, e.g. `fedora-$releasever-$basearch/` that for EPEL should be hardcoded major `epel-9-$basearch/`. The repo type should be `rpm-md` to make `dnf` more efficient.

```suggestion
[flightctl]
name=Flight Control RPM repository
baseurl=https://rpm.flightctl.io/$REPO_NAME/epel-9-\$basearch/
type=rpm-md
enabled=1
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/@redhat-et/flightctl/pubkey.gpg
repo_gpgcheck=0
metadata_expire=1d
enabled_metadata=1
```

hack/create-github-pages.sh
```diff
@@ -0,0 +1,593 @@
...
+cat > "$PAGES_OUTPUT_DIR/flightctl-centos9.repo" << EOF
+[flightctl]
+name=FlightCtl RPM Repository (CentOS Stream 9)
+baseurl=https://$REPO_OWNER.github.io/$REPO_NAME/centos-stream-9-\$basearch/
+enabled=1
+gpgcheck=0
+repo_gpgcheck=0
+type=rpm
+metadata_expire=1d
+EOF
```
fzdarsky, line 102: We should no longer use/publish the `centos` and `rhel` chroots, only `epel`.

hack/create-github-pages.sh
```diff
@@ -0,0 +1,593 @@
...
+        <div class="code-block"># CentOS Stream 9
+sudo dnf install https://$REPO_OWNER.github.io/$REPO_NAME/centos-stream-9-x86_64/flightctl-*.rpm
+
+# Fedora
+sudo dnf install https://$REPO_OWNER.github.io/$REPO_NAME/fedora-*-x86_64/flightctl-*.rpm</div>
+
+        <h3>Specific Version</h3>
+        <div class="code-block"># Install specific version
+sudo dnf install https://$REPO_OWNER.github.io/$REPO_NAME/centos-stream-9-x86_64/flightctl-$LATEST_VERSION-*.rpm</div>
+    </div>
```
fzdarsky, line 572: Not sure we should show how to directly download&install RPMs, especially since most have dependencies. Why not stay with the repo only?

Then, to install a specific version, we could simply tell users to `sudo dnf install flightctl-0.7.2`.

hack/create-github-pages.sh
```diff
@@ -0,0 +1,593 @@
...
+
+    <div class="instructions">
+        <h2>📋 Quick Setup</h2>
+        <p>Add this repository to your system:</p>
+        <div class="code-block">sudo dnf config-manager --add-repo https://$REPO_OWNER.github.io/$REPO_NAME/flightctl.repo<br />sudo dnf install flightctl</div>
+    </div>
+
+    <div class="stats">
+        <h3>📊 Repository Overview</h3>
+        <p><strong>Total packages:</strong> $total_rpms RPMs</p>
```
fzdarsky, line 483: This seems to count the individual package versions. There's only `flightctl-agent` package, no matter how many versions/archs/... exist of it.

hack/create-github-pages.sh
```diff
@@ -0,0 +1,593 @@
...
+EOF
+
+# Add version badges to header
+if [ -n "$versions" ]; then
+    for version in $versions; do
+        echo "            <span class=\"version-badge badge\">$version</span>" >> "$PAGES_OUTPUT_DIR/index.html"
+    done
+else
+    echo "            <span class=\"version-badge badge\">$LATEST_VERSION</span>" >> "$PAGES_OUTPUT_DIR/index.html"
+fi
```
fzdarsky, line 469: Personally, I feel the presentation with cards is cluttered and presents a lot of redundant info that makes things hard to find. How about we have one page per chroot with a simple sorted lists of all packages+versions?

gshilin-sdb:
> Before we point users to the new locations, we should create a DNS entry for rpm.flightctl.io, so we can move repo location later without breaking users another time.

I don't have enough permissions, but here's what needs to be done if we go for it (not that I understand each line):

Domain ownership: The FlightCtl project must own flightctl.io
DNS configuration: Set up DNS records pointing to GitHub Pages
GitHub Pages custom domain: Configure the repository to use the custom domain

Setup Steps:
1. DNS Configuration (in your domain provider):
```
# Add CNAME record
rpm.flightctl.io → flightctl.github.io
```
2. GitHub Repository Configuration:

Go to `https://github.com/flightctl/flightctl/settings/pages`
In `**Custom domain**` field, enter: `rpm.flightctl.io`
Check `**Enforce HTTPS**`
GitHub will verify domain ownership

3. Domain verification (one-time):

GitHub will require domain verification
Add a TXT record to `flightctl.io` with verification code

Result:

Your RPM repository would be accessible at: https://rpm.flightctl.io/
DNF commands would use: https://rpm.flightctl.io/flightctl.repo

> Also, we probably want to move the docs into a separate repo at some point, so the web site can be evolved independently of the code. Would that impact our decision where to put the repo metadata?

Independently from the above:
Now `https://flightctl.github.io/flightctl/` page is served from branch `gh-pages`, from directory `docs`.
We can make there two subdirectories: `docs` and `rpms`

Another way: to create a `flightctl-docs` project and keep docs there

hack/copr-to-gh_pages-download.sh
gshilin-sdb: Verified and every time code was fallback-ed to `copr-cli`. So removing `curl` instead

hack/create-github-pages.sh
gshilin-sdb: yes, and -cli

gshilin-sdb:
all moved to flightctl/flightctl-rpm

PR #1323, opened 2025-07-06
EDM-1675: Create observability backend stanalone quadlets

ori-amizur:
EDM-1678: Support observability SSO against AAP

## Summary by CodeRabbit

* **New Features**
  * Introduced a standalone observability stack with Grafana, Prometheus, and OpenTelemetry Collector, including new systemd targets, container templates, and network configuration.
  * Added a UserInfo Proxy service for OpenID Connect integration with Ansible Automation Platform, including a new proxy server and container build.
  * Provided comprehensive user documentation for deploying and configuring the observability stack.
  * Added configuration options for observability components in the main service config.

* **Improvements**
  * Enhanced deployment scripts with explicit readiness checks for database and key-value store, improved secret synchronization, and service restarts.
  * Updated container configurations across services to pull newer images automatically.
  * Improved error messages for certificate and key file checks.
  * Modified systemd unit dependencies to use soft dependencies for Flight Control services.

* **Chores**
  * Added new RPM sub-packages for observability and OpenTelemetry Collector with lifecycle scripts and packaging support.
  * Added a template rendering script and a dedicated rendering command for observability configuration.
  * Extended spell-check dictionary with new domain-specific terms.

### DIFF
--- a/Containerfile.userinfo
+++ b/Containerfile.userinfo
@@ -1,0 +1,39 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.23.6-1747333074 as build
+WORKDIR /app
+COPY ./api api
+COPY ./cmd cmd
+COPY ./deploy deploy
+COPY ./hack hack
+COPY ./internal internal
+COPY ./go.* ./
+COPY ./pkg pkg
+COPY ./test test
+COPY ./Makefile .
+# make sure that version extraction works
+COPY .git .git
+
+USER 0
+RUN make build-userinfo-proxy
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal as certs
+RUN microdnf update --nodocs -y  && microdnf install ca-certificates --nodocs -y
+
+FROM registry.access.redhat.com/ubi9/ubi-micro
+WORKDIR /app
+LABEL \
+  com.redhat.component="flightctl-userinfo-proxy-container" \
+  description="Flight Control UserInfo proxy for Grafana authentication" \
+  io.k8s.description="Flight Control UserInfo proxy for Grafana authentication" \
+  io.k8s.display-name="Flight Control UserInfo Proxy" \
+  name="flightctl-userinfo-proxy" \
+  summary="Flight Control UserInfo proxy for Grafana authentication"
+COPY --from=build /app/bin/flightctl-userinfo-proxy .
+COPY --from=certs /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/
+
+EXPOSE 8080
+
+# Health check
+HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
+  CMD wget --no-verbose --tries=1 --spider http://localhost:8080/health || exit 1
+
+CMD ./flightctl-userinfo-proxy 
--- a/deploy/scripts/deploy_quadlets.sh
+++ b/deploy/scripts/deploy_quadlets.sh
@@ -14,35 +14,123 @@ if ! deploy/scripts/install.sh; then
     exit 1
 fi
 
+echo "Ensuring secrets are available..."
+# Always ensure secrets exist before starting services
+if ! "${CONFIG_READONLY_DIR}/init_host.sh"; then
+    echo "Error: Failed to initialize secrets"
+    exit 1
+fi
+
+echo "Starting all FlightCtl services via target..."
 start_service "flightctl.target"
 
-echo "Checking if all services are running..."
-
-timeout --foreground 300s bash -c '
-    expected_services=("flightctl-api" "flightctl-worker" "flightctl-periodic" "flightctl-alert-exporter" "flightctl-db" "flightctl-kv" "flightctl-alertmanager" "flightctl-alertmanager-proxy" "flightctl-cli-artifacts" "flightctl-ui")
-
+echo "Waiting for services to initialize..."
+# Wait for database to be ready first
+timeout --foreground 120s bash -c '
     while true; do
-        missing_services=()
-
-        for service in "${expected_services[@]}"; do
-            if ! podman ps --quiet --filter "name=$service" | grep -q .; then
-                missing_services+=("$service")
-            fi
-        done
-
-        if [ ${#missing_services[@]} -eq 0 ]; then
-            echo "All services are running"
-            exit 0
+        if podman ps --quiet --filter "name=flightctl-db" | grep -q . && \
+           podman exec flightctl-db pg_isready -U postgres >/dev/null 2>&1; then
+            echo "Database is ready"
+            break
         fi
-
-        echo "Waiting for (${#missing_services[@]}/${#expected_services[@]}) services to start: ${missing_services[*]}"
-        sleep 10
+        echo "Waiting for database to become ready..."
+        sleep 3
     done
-' || {
-    echo "Timeout reached while waiting for services"
-    exit 1
-}
+'
 
-echo "Deployment completed. Please log in to Flight Control with the following command:"
-echo "Login: flightctl login --insecure-skip-tls-verify $(grep baseUrl $CONFIG_WRITEABLE_DIR/flightctl-api/config.yaml | awk '{print $2}')"
-echo "Console URL: $(grep baseUIUrl $CONFIG_WRITEABLE_DIR/flightctl-api/config.yaml | awk '{print $2}')"
+# Sync database password with secrets
+echo "Synchronizing database password with secrets..."
+DB_ACTUAL_PASSWORD=$(sudo podman exec flightctl-db printenv POSTGRESQL_MASTER_PASSWORD)
+if ! sudo podman run --rm --network flightctl \
+    --secret flightctl-postgresql-master-password,type=env,target=DB_PASSWORD \
+    quay.io/sclorg/postgresql-16-c9s:latest \
+    bash -c 'PGPASSWORD="$DB_PASSWORD" psql -h flightctl-db -U admin -d flightctl -c "SELECT 1" >/dev/null 2>&1'; then
+    
+    echo "Password mismatch detected! Fixing secret..."
+    sudo podman secret rm flightctl-postgresql-master-password
+    echo "$DB_ACTUAL_PASSWORD" | sudo podman secret create flightctl-postgresql-master-password -
+    echo "Secret updated to match database password"
+fi
+
+# Ensure admin has superuser privileges
+echo "Ensuring database admin has superuser privileges..."
+if sudo podman exec flightctl-db psql -U postgres -tAc "SELECT rolsuper FROM pg_roles WHERE rolname = 'admin'" | grep -q "f"; then
+    echo "Granting superuser privileges to admin user..."
+    sudo podman exec flightctl-db psql -U postgres -c "ALTER USER admin WITH SUPERUSER;"
+fi
+
+# Wait for key-value service
+timeout --foreground 60s bash -c '
+    while true; do
+        if podman ps --quiet --filter "name=flightctl-kv" | grep -q . && \
+           podman exec flightctl-kv redis-cli ping >/dev/null 2>&1; then
+            echo "Key-value service is ready"
+            break
+        fi
+        echo "Waiting for key-value service..."
+        sleep 2
+    done
+'
+
+# Restart any failed services due to initial password/permission issues
+echo "Restarting API services to apply fixes..."
+sudo systemctl restart flightctl-api.service flightctl-worker.service flightctl-periodic.service
+
+echo "Waiting for all services to be fully ready..."
+# Get all services from flightctl.target
+ALL_SERVICES=$(systemctl show flightctl.target -p Wants --value | tr ' ' '\n' | grep -E '^flightctl-.*\.service$' | sort)
+
+# Wait for core services to be ready
+start_time=$(date +%s)
+timeout_seconds=120
+
+while true; do
+    current_time=$(date +%s)
+    elapsed=$((current_time - start_time))
+    
+    if [ $elapsed -ge $timeout_seconds ]; then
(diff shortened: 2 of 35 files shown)


Containerfile.userinfo
```diff
@@ -0,0 +1,39 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.23.6-1747333074 as build
```
ori-amizur, line 1: Converter between AAP userinfo to the expected format (https://connect2id.com/products/server/docs/api/userinfo)

amir-yogev-gh:
@gshilin-sdb this will eventually generate a new RPM file

packaging/observability/flightctl-otel-collector-config-reloader.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -eo pipefail # Exit immediately if a command exits with a non-zero status or a command in a pipeline fails
+
+CONFIG_FILE="/etc/flightctl/service-config.yaml"
+TEMPLATES_DIR="/opt/flightctl-observability/templates"
+DEFINITIONS_FILE="/etc/flightctl/definitions/otel-collector.defs"
+
+# Source shared logic
+source /etc/flightctl/scripts/render-templates.sh
+
```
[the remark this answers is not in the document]
ori-amizur: It is part of the same RPM

packaging/observability/flightctl-userinfo-proxy.container.template
```diff
@@ -0,0 +1,26 @@
...
+ContainerName=flightctl-userinfo-proxy
+
+# Network configuration
+Network=flightctl
+
+# Environment variables for configuration
+Environment=USERINFO_LISTEN_PORT=8080
+Environment=USERINFO_UPSTREAM_URL=${USERINFO_UPSTREAM_URL}
+Environment=USERINFO_SKIP_TLS_VERIFY=${USERINFO_SKIP_TLS_VERIFY}
+
```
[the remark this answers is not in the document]
ori-amizur: If these values are missing then the proxy is not started at all

packaging/rpm/flightctl.spec
```diff
@@ -69,6 +69,370 @@ Requires: podman
...
+%dir /etc/grafana/provisioning/datasources
+%dir /etc/grafana/provisioning/dashboards
+%dir /etc/grafana/provisioning/dashboards/flightctl
+%dir /etc/grafana/certs
+%dir /var/lib/prometheus
+%dir /var/lib/grafana
+%dir /etc/flightctl
+%dir /etc/flightctl/scripts
+%dir /etc/flightctl/definitions
+%dir /usr/local/bin
```
DakCrowder: I'm not sure if `/usr/local/bin` itself should be owned by the observability rpm - seems like its a broader system directory that the rpm shouldn't own

packaging/rpm/flightctl.spec
```diff
@@ -69,6 +69,370 @@ Requires: podman
...
+    fi
+done
+
+if [ ${#MISSING_SERVICES[@]} -gt 0 ]; then
+    echo "ERROR: FlightCtl OpenTelemetry collector requires the following services to be running:"
+    for service in "${MISSING_SERVICES[@]}"; do
+        echo "  - $service"
+    done
+    echo ""
+    echo "Please deploy the FlightCtl services first by running:"
```
DakCrowder: Having an prm installable depending on runtime state seems like an antipattern, I wonder if this should be part of spinning up/down the associated services after install?

packaging/rpm/flightctl.spec
```diff
@@ -69,6 +69,370 @@ Requires: podman
...
+
+/usr/bin/systemctl daemon-reload
+echo "FlightCtl OpenTelemetry Collector uninstalled."
+
+
+%pre observability
+# This script runs BEFORE the files are installed onto the system.
+echo "Checking FlightCtl service prerequisites for observability stack..."
+
+# Check if FlightCtl services from 'make deploy-quadlets' are running
```
DakCrowder: This comment should be modified as `make deploy-quadlets` is not ever run as a part of the rpm install (its only for local dev)

packaging/rpm/flightctl.spec
```diff
@@ -69,6 +69,370 @@ Requires: podman
...
+# This script runs AFTER the files have been installed onto the system.
+echo "Running post-install actions for FlightCtl OpenTelemetry Collector..."
+
+# Create necessary directories on the host if they don't already exist.
+/usr/bin/mkdir -p /etc/otelcol /var/lib/otelcol
+/usr/bin/mkdir -p /opt/flightctl-observability/templates
+/usr/bin/mkdir -p /etc/flightctl /etc/flightctl/scripts /etc/flightctl/definitions
+
+# Apply persistent SELinux contexts for volumes and configuration files.
+/usr/sbin/semanage fcontext -a -t container_file_t "/etc/otelcol/otelcol-config.yaml" >/dev/null 2>&1 || :
```
DakCrowder, line 199: Are semanage and restorecon guaranteed to be present on installed machines?  The current call in this .spec file (`agent_relabel_files`) which references them is actually _broken_ because iirc the necessary `policycoreutils` and `policycoreutils-python-utils` requirements aren't properly specified in this .spec

deploy/scripts/deploy_quadlets.sh
```diff
@@ -14,33 +14,90 @@ if ! deploy/scripts/install.sh; then
...
+        sleep 3
+    done
+'
 
-echo "Deployment completed. Please log in to Flight Control with the following command:"
-echo "Login: flightctl login --insecure-skip-tls-verify $(grep baseUrl $CONFIG_WRITEABLE_DIR/flightctl-api/config.yaml | awk '{print $2}')"
-echo "Console URL: $(grep baseUIUrl $CONFIG_WRITEABLE_DIR/flightctl-api/config.yaml | awk '{print $2}')"
+echo "Deployment completed successfully!"
+echo ""
+echo "All FlightCtl services should now be running:"
```
DakCrowder: This doesn't account for services like the ui or the recently added alertmanager / proxy / alert-exporter services

deploy/scripts/deploy_quadlets.sh
```diff
@@ -14,33 +14,90 @@ if ! deploy/scripts/install.sh; then
...
+
+# Sync database password with secrets
+echo "Synchronizing database password with secrets..."
+DB_ACTUAL_PASSWORD=$(sudo podman exec flightctl-db printenv POSTGRESQL_MASTER_PASSWORD)
+if ! sudo podman run --rm --network flightctl \
+    --secret flightctl-postgresql-master-password,type=env,target=DB_PASSWORD \
+    quay.io/sclorg/postgresql-16-c9s:latest \
+    bash -c 'PGPASSWORD="$DB_PASSWORD" psql -h flightctl-db -U admin -d flightctl -c "SELECT 1" >/dev/null 2>&1'; then
+    
+    echo "Password mismatch detected! Fixing secret..."
```
DakCrowder, line 49: :rocket:

packaging/observability/observability.defs
```diff
@@ -0,0 +1,25 @@
...
+GRAFANA_CERT_KEY               | .observability.grafana.https.cert_key                      | /etc/grafana/certs/grafana.key               | grafana.ini.template                          | /etc/grafana/grafana.ini
+
+GRAFANA_IMAGE                  | .observability.grafana.image                               | docker.io/grafana/grafana:latest             | flightctl-grafana.container.template          | /etc/containers/systemd/flightctl-grafana.container
+GRAFANA_PUBLISHED_PORT         | .observability.grafana.published_port                      | 3000                                         | flightctl-grafana.container.template          | /etc/containers/systemd/flightctl-grafana.container
+
+PROMETHEUS_IMAGE               | .observability.prometheus.image                            | docker.io/prom/prometheus:latest             | flightctl-prometheus.container.template       | /etc/containers/systemd/flightctl-prometheus.container
+
+USERINFO_PROXY_IMAGE           | .observability.userinfo_proxy.image                       | flightctl/userinfo-proxy:latest              | flightctl-userinfo-proxy.container.template   | /etc/containers/systemd/flightctl-userinfo-proxy.container
```
DakCrowder, line 22: Does the userinfo-proxy need to be added to the publish-containers github action similar to https://github.com/flightctl/flightctl/pull/1340/files?

packaging/rpm/flightctl.spec
```diff
@@ -69,6 +69,370 @@ Requires: podman
 %description services
 The flightctl-services package provides installation and setup of files for running containerized Flight Control services
 
+%package otel-collector
+Summary: OpenTelemetry Collector for FlightCtl
+Requires:       podman
+Requires:       systemd
+Requires:       yq
```
DakCrowder, line 76: Are there downstream effects or security considerations for including the yq dependency?  It would be great to use for some of the other installation scripts but I was unsure about including a non-RH managed dep.

Here is a somewhat related discussion around yq usage in konflux
https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1738055819157789

deploy/scripts/deploy_quadlets.sh
ori-amizur: OK.  Will add

packaging/observability/observability.defs
ori-amizur: Yes. Thanks.

packaging/rpm/flightctl.spec
ori-amizur: There is no sense (and it might fail), if flight control quadlets do not run before installation.
Another option is to pack all the quadlets in a separate RPM (which contradicts the way it is deployed now).  
Then the dependency will be between RPMs.

packaging/rpm/flightctl.spec
ori-amizur: OK

packaging/rpm/flightctl.spec
ori-amizur: Good point.  Need to check

packaging/rpm/flightctl.spec
ori-amizur: I am not sure that declaring the directory here means ownership.  It is just a requirement that the directory should exist.

DakCrowder: These docs mentions %dir implies ownership?  https://rpm-guide.readthedocs.io/en/latest/rpm-guide.html  but it isn't really clear exactly what that entails....  After looking it up I don't think anything bad will happen, its just the rpm would try to remove /usr/local/bin if it was empty on uninstall.

packaging/rpm/flightctl.spec
DakCrowder: Ahh I just realized the line below this also has
`make deploy-quadlets`
which isn't how they are installed via the rpms - users would do something like
`sudo dnf install -y flightctl-services`
then
`sudo systemctl enable flightctl.target
sudo systemctl start flightctl.target`
to get the services up and running.

ori-amizur: We don't have it here.  Here we deploy via "make deploy-quadlets"
We can create such rpm, but this wasn't the requirement

packaging/rpm/flightctl.spec
ori-amizur: OK.  Will remove

packaging/rpm/flightctl.spec
ori-amizur: It seems there are couple of possible options from the thread.  One of them is just to use the container image
quay.io/konflux-ci/yq .  Another is to pack for our purpose our own or publicly available yq executable in the RPM.
Since there will be other observability PRs following this one, I would like to postpone the decision to later stage.

packaging/rpm/flightctl.spec
ori-amizur: Added requirement

packaging/rpm/flightctl.spec
DakCrowder: I'm still a bit confused - `make deploy-quadlets` has no meaning for a user who has installed the rpm and references to that command shouldn't exist in this context.  As I understand it users should install and run the `flightctl-services` RPM, then install `flightctl-observability` RPM on top of it.

`make deploy-quadlets` is for local development only and we don't ship the scripts to run it - I think we should just replace an reference to `make deploy-quadlets` with a more generic message to please start the services.

DakCrowder: Additionally, I remembered an important detail about the rpm installation experience.  The services RPM used to perform some similar commands (dynamic configuration, calling systemd).  However this was modified to better support bootc installs in the future, where users can rpm install in a .Containerfile and ship everything bundled in that fashion.  In the context of a bootc .Containerfile creating dynamic data is potentially problematic and making calls to start services via systemd is not meaningful.

Thus the services rpm takes an approach like:

1. Lays down and installs necessary files - this is all the RPM really does
2. Provides instructions to the user to run services
3. Services generate provided configuration (including dynamic data per installation like podman secrets for password) they need at runtime.

Deviating from this pattern introduces a second model that is not only different from the way we install and package the flightctl services themselves, it would be problematic with our bootc appliance vision of the future.  I don't think that is a requirement for the observability work in the short term but wanted to call that out as it will require some refactoring if we wanted users to be able to package the observability stack in an image/appliance in the future.

internal/config/config.go
```diff
@@ -43,7 +43,7 @@ type svcConfig struct {
 	BaseUrl                string        `json:"baseUrl,omitempty"`
 	BaseAgentEndpointUrl   string        `json:"baseAgentEndpointUrl,omitempty"`
 	BaseUIUrl              string        `json:"baseUIUrl,omitempty"`
-	SrvCertFile            string        `json:"srvCertificateFile,omitempty"`
+	SrvCertFile            string        `json:"srvCertFile,omitempty"`
```
DakCrowder: I merged a change to main yesterday to align/fix the underlying mismatch between this key and the api `conifg.yaml.template` - can be removed

PR #1324, opened 2025-07-06
EDM-1567: Prevent multiple resourceSyncs to fleets with identical names

noga-magen:
## Summary by CodeRabbit

* **New Features**
  * Added validation to detect and prevent conflicts when creating or updating fleets with names already owned by another ResourceSync. Users will now receive an error listing any conflicting fleet names, ensuring fleet name ownership remains unique.

### DIFF
--- a/internal/cli/helpers.go
+++ b/internal/cli/helpers.go
@@ -44,6 +44,7 @@ var (
 		ResourceSyncKind:              "rs",
 		TemplateVersionKind:           "tv",
 		CertificateSigningRequestKind: "csr",
+		EventKind:                     "ev",
 	}
 )
 


internal/tasks/resourcesync.go
```diff
@@ -367,3 +376,27 @@ func isValidFile(filename string) bool {
 	}
 	return false
 }
+
+func (r *ResourceSync) validateFleetNameConflicts(ctx context.Context, fleets []*api.Fleet, owner string) error {
+	var conflictingFleets []string
+
+	for _, fleet := range fleets {
+		// Check if a fleet with this name already exists
+		existingFleet, status := r.serviceHandler.GetFleet(ctx, *fleet.Metadata.Name, api.GetFleetParams{})
```
avishayt, line 385: Ideally this check would be in the store layer to make sure nobody is modifying a resource owned by someone else, but this is OK for now.

internal/tasks/resourcesync.go
```diff
@@ -105,6 +105,15 @@ func (r *ResourceSync) run(ctx context.Context, log logrus.FieldLogger, rs *api.
 		return err
 	}
 
+	// Validate that no fleet names conflict with fleets owned by other ResourceSyncs
+	err = r.validateFleetNameConflicts(ctx, fleets, *owner)
+	if err != nil {
+		api.SetStatusConditionByError(&rs.Status.Conditions, api.ConditionTypeResourceSyncResourceParsed, "success", "fail", err)
+		err = fmt.Errorf("resourcesync/%s: error: %w", *rs.Metadata.Name, err)
+		log.Errorf("%e", err)
```
hexfusion, line 113: if err is type error then this should be
```suggestion
		log.Error(err)
```
%e is for scientific notation

- https://pkg.go.dev/fmt

internal/tasks/resourcesync.go
hexfusion, line 385: nit: this is a small issue but if we must dereference, we should try to just do it once. for example you could do

fleetName := *fleet.Metadata.Name at the top of the loop

PR #1326, opened 2025-07-06
NO-ISSUE: Optimize DeviceDisconnected task

avishayt:
This change addresses the scalability issue where the DeviceDisconnected task was scanning ALL devices every 2 minutes just to check for disconnected ones.

Changes:
- Use field selector 'status.lastSeen<cutoff_time' to query only devices that haven't checked in for more than 5 minutes (DeviceDisconnectedTimeout)
- Add comprehensive integration tests covering all scenarios including the field selector optimization, pagination, edge cases, and mixed device states


internal/tasks/device_disconnected.go
```diff
@@ -37,7 +37,17 @@ func (t *DeviceDisconnected) Poll(ctx context.Context) {
...
 	statusInfoMessage := fmt.Sprintf("Did not check in for more than %d minutes", int(api.DeviceDisconnectedTimeout.Minutes()))
 
-	listParams := api.ListDevicesParams{Limit: lo.ToPtr(int32(ItemsPerPage))}
+	// Calculate the cutoff time for disconnected devices
+	cutoffTime := time.Now().Add(-api.DeviceDisconnectedTimeout)
+
+	// Create a field selector to only get devices that haven't been seen for more than DeviceDisconnectedTimeout
+	fieldSelector := fmt.Sprintf("status.lastSeen<%s", cutoffTime.Format(time.RFC3339))
+
+	listParams := api.ListDevicesParams{
```
ori-amizur, line 47: If a device is already marked as unknown, are we marking it again? It means that if most of the devices are disconnected, we are marking them on every iteration

internal/tasks/device_disconnected.go
```diff
@@ -47,17 +57,22 @@ func (t *DeviceDisconnected) Poll(ctx context.Context) {
 
 		var batch []string
 		for _, device := range devices.Items {
+			// Update the device status on the service side
```
ori-amizur, line 61: It seems that the function UpdateServiceSideDeviceStatus updates the device in memory, but the change is not saved to the database.  If not then the same operation will be done again during the next iteration

avishayt: It's saved to the DB below - UpdateDeviceSummaryStatusBatch()

internal/tasks/device_disconnected.go
avishayt: You're correct - updated the field selector to exclude "unknown" devices.

internal/tasks/device_disconnected.go
ori-amizur: Only the summary status.  The function UpdateServiceSideDeviceStatus updates the field `Status.Updated.Status` but this is not updated in the DB.

avishayt: OK, this is a pre-existing bug.  Good catch!

PR #1327, opened 2025-07-07
EDM-518: Parse resource time in days and weeks

noga-magen:


internal/agent/device/resource/common.go
```diff
@@ -21,7 +22,7 @@ func updateMonitor(
 		return false, err
 	}
 
-	newSamplingInterval, err := time.ParseDuration(spec.ResourceMonitorSpec.SamplingInterval)
+	newSamplingInterval, err := util.ExtendedParseDuration(spec.ResourceMonitorSpec.SamplingInterval)
```
hexfusion, line 25: few questions:
- what does the `SamplingInterval` represent in the context of a resource monitor? then what would be the usecase for it to be defined in days or weeks.
- where does this value come from?

internal/util/duration.go
```diff
@@ -0,0 +1,105 @@
+package util
```
hexfusion, line 1: I believe we have a util.Duration already if we were to make this change we should consider using that.

avishayt:
@hexfusion I saw you wrote in Jira that the bug is not relevant?

internal/util/duration.go
```diff
@@ -0,0 +1,105 @@
...
+func parseExtendedDuration(s string) (time.Duration, error) {
+	var total time.Duration
+	remaining := s
+
+	// Parse extended units first (days and weeks)
+	for _, unit := range []struct {
+		suffix     string
+		multiplier time.Duration
+	}{
+		{"w", Week},
```
hexfusion, line 38: a week is 7 days, is that a requirement? do our other API's support weeks?

celdrake:
Hi, if this change goes forward, it would require changes in the UI. Please let us know before merging.

Also, the openapi [spec](https://github.com/flightctl/flightctl/blob/main/api/v1alpha1/openapi.yaml#L4513) has not been updated, and I suspect validations at the API level are also missing.

internal/util/duration_test.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+		{
+			name:     "standard hours",
+			input:    "2h",
+			expected: 2 * time.Hour,
+			wantErr:  false,
+		},
+		{
+			name:     "complex standard duration",
+			input:    "1h30m45s",
```
celdrake, line 40: These complex durations are not allowed by the current Duration openAPI spec `^[1-9]\d*[smh]$`, and therefore they are not supported in the UI.

I think we should stick to the single unit rule that we have now.

internal/util/duration.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Common duration constants for convenience
+const (
+	Day  = 24 * time.Hour
```
ori-amizur, line 13: Are we OK that a day might not always equal to 24 hours?
Example: Daylight savings (23 or 25), leap seconds

PR #1335, opened 2025-07-07
EDM-1793: Ensure event names conform to k8s standards

avishayt:
Resources in Flight Control API have a "metadata.name" which adhere to the restrictions of k8s API. We were using the requestID as part of the event name so that events can be easily correlated with requests, but requestIDs sometimes had illegal characters like slashes and upper-case letters.

The solution is to use a UUID for the event's name, and store the requestID in an annotation.

```
avishay@fedora:~/work/flightctl$ bin/flightctl get events -oyaml
apiVersion: flightctl.io/v1alpha1
items:
- actor: user:none
  apiVersion: flightctl.io/v1alpha1
  involvedObject:
    name: client-enrollment-70f71f74
  kind: Event
  message: CertificateSigningRequest was created successfully.
  metadata:
    annotations:
      event-controller/requestID: fedora/myQJmQMYkB-000001
    creationTimestamp: "2025-07-07T17:11:28.837744Z"
    name: 538dd08d-c463-4a92-996b-447f3118c6f1
  reason: ResourceCreated
  source:
    component: flightctl-api
  type: Normal
kind: EventList
metadata: {}

```

## Summary by CodeRabbit

* **New Features**
  * Events now include a request ID annotation for improved traceability.

* **Improvements**
  * Event names are generated as UUIDs to ensure compliance with Kubernetes naming standards.
  * Annotations are consistently preserved when converting events to and from API resources.

### DIFF
--- a/internal/tasks/device_disconnected.go
+++ b/internal/tasks/device_disconnected.go
@@ -37,7 +37,18 @@ func (t *DeviceDisconnected) Poll(ctx context.Context) {
 
 	statusInfoMessage := fmt.Sprintf("Did not check in for more than %d minutes", int(api.DeviceDisconnectedTimeout.Minutes()))
 
-	listParams := api.ListDevicesParams{Limit: lo.ToPtr(int32(ItemsPerPage))}
+	// Calculate the cutoff time for disconnected devices
+	cutoffTime := time.Now().Add(-api.DeviceDisconnectedTimeout)
+
+	// Create a field selector to only get devices that haven't been seen for more than DeviceDisconnectedTimeout
+	// and don't already have "Unknown" status to avoid reprocessing the same devices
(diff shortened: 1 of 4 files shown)


celdrake (approved):

internal/service/event.go
```diff
@@ -172,9 +169,17 @@ func getBaseEvent(ctx context.Context, resourceEvent resourceEvent, log logrus.F
 		componentStr = component.(string)
 	}
 
+	// Generate a UUID for the event name to ensure k8s compliance
+	eventName := uuid.New().String()
+
+	// Create annotations map with the request ID for correlation
```
gshilin-sdb: Maybe to create annotations instead of lines 154-160 as we don't need requestIDstr anymore?

gshilin-sdb (approved):

PR #1336, opened 2025-07-07
EDM-1747: Fix policy syncing

kkyrazis:
After a discussion with @hexfusion it was decided to fix the existing issue in a non-invasive manner while a more comprehensive approach for a long term fix is designed for a future release. This change solves the existing problem by syncing policy updates immediately upon receiving a new version so that the most recent policy rules will be applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy synchronization to prevent update stalls by ensuring policy state is updated before processing new device specs.

* **Tests**
  * Enhanced test coverage for policy synchronization scenarios, including new and updated test cases to verify correct behavior when policy readiness states vary.
  * Adjusted existing tests to reflect changes in policy synchronization flow.

### DIFF
--- a/internal/agent/device/spec/queue_test.go
+++ b/internal/agent/device/spec/queue_test.go
@@ -1,7 +1,7 @@
 package spec
 
 import (
-	context "context"
+	"context"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


internal/agent/device/spec/queue_test.go
```diff
@@ -1,7 +1,7 @@
 package spec
 
 import (
-	context "context"
+	"context"
```
hexfusion, line 4: ha :+1:

PR #1354, opened 2025-07-10
NO-ISSUE: stop building centos9, rhel, fedora-eln. For PR: build epel-9-aarch64 instead

gshilin-sdb:


fzdarsky (approved):
LGTM. For the record, the choice of building F42 for x86_64 and EPEL9 aarch64 only when the trigger is PR was to keep the # of builds small while sampling opposite ends of the space.

PR #1359, opened 2025-07-10
Add default organization to api layer

DakCrowder:
Goal of this PR: Add the organization model without modifying current api behavior.  Fall back in all cases to the current org id while paving the path forward to supporting multi org.

Changes are focused around:
- Creating a default org
- Establishing the api for fetching organizations
  - Currently list orgs will only be the single org -  creating new orgs / pagination not implemented
- Establishing an api for segmenting calls by organization
  - ?org_name parameter
- Some refactoring to centralize pulling the currently scoped from context in the service layer
  - Currently falls back to the default org in all cases

## Summary by CodeRabbit

* **New Features**
  * Introduced organization management as a first-class API resource, including new endpoints and data models for organizations.
  * Added an endpoint to list organizations.
  * Implemented middleware and context utilities for organization ID handling.
  * Added support for organizations in the data store and service layers.

* **Bug Fixes**
  * Improved test setups to ensure organizations are created before associating resources.

* **Tests**
  * Added comprehensive unit and integration tests for organization-related functionality.

* **Documentation**
  * Updated OpenAPI specification to document new organization endpoints and security schemes.

* **Chores**
  * Updated Makefile for more flexible integration test directory assignment.


internal/store/model/resource.go
```diff
@@ -11,11 +11,12 @@ import (
 type Resource struct {
 	// Composite Primary Key: Unique within a tenant (OrgID, Name)
 
-	// Uniquely identifies the tenant the resource belongs to.
+	// Uniquely identifies the organization the resource belongs to.
 	// Assigned by IAM. Immutable.
-	OrgID uuid.UUID `gorm:"type:uuid;primaryKey;index:,composite:org_name,priority:1"`
+	OrgID        uuid.UUID    `gorm:"type:uuid;primaryKey;index:,composite:org_name,priority:1"`
+	Organization Organization `gorm:"foreignKey:OrgID"`
```
ori-amizur, line 17: Why is there a need to add the Organization as part of a resource if OrgID already exists ?

internal/store/model/organization.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/google/uuid"
+)
+
+type Organization struct {
+	ID uuid.UUID `gorm:"type:uuid;primary_key"`
+
+	CreatedAt time.Time `json:"created_at"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
```
avishayt: What's the benefit of storing the org IDs in a separate table?
1. When do we create an org?
2. Can an org be deleted?

internal/store/model/resource.go
DakCrowder: This is how the gorm [docs](https://gorm.io/docs/belongs_to.html#Belongs-To) indicate defining a "Belongs To" relationship

internal/store/model/organization.go
DakCrowder: For storing org IDs, I think there are a few different benefits:
- Ultimately since they segment so much of our data I'd like to control them, ensuring uniqueness and stability.  A composite primary key on basically all of the Flight Control data to me indicates an identifier we want to control
- Having them be uniform for clients to consume / use
- We can't guarantee organization ids in external systems will always be UUIDs, and having some arbitrary values could provide less efficient queries if it was something like a composite key of (varchar, name) for a `Resource`

I'm really not trying to think _too_ much about the future but I think it also keeps the door open a bit more into cases like org-specific config like certs/IdPs or additional management features in the UI.

1.  Currently, only the default org is created as part of an initial migration.  I think once IdPs like AAP are configured we can create orgs when we reach out to AAP to see what orgs a user has access to - mapping from the external org identifiers AAP gives us to internal ones and storing them if needed.
2. Currently no without manual intervention at the db level, although in the future we could have some kind of org management if required (e.g. maybe an org gets deleted in an IdP) and allow for some type of cascade on delete if users were super sure.

api/v1alpha1/openapi.yaml
```diff
@@ -5298,3 +5319,46 @@ components:
...
+          type: string
+          description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds.'
+        id:
+          type: string
+          format: uuid
+          description: The unique identifier for the organization.
+        displayName:
+          type: string
+          description: Human readable name shown to users.
+    OrganizationList:
```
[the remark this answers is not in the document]
avishayt: I agree with this

internal/api_server/middleware/middleware.go
```diff
@@ -57,3 +60,26 @@ func AddEventMetadataToCtx(next http.Handler) http.Handler {
...
+		if orgIDParam != "" {
+			parsedID, err := uuid.Parse(orgIDParam)
+			if err != nil {
+				http.Error(w, fmt.Sprintf("Invalid org_id parameter: %s", err.Error()), http.StatusBadRequest)
+				return
+			}
+			orgID = parsedID
+		} else {
+			// Fall back to the default organization.
+			orgID = store.NullOrgId
```
avishayt: What's the behavior of the Null org?  Is everyone a member?  If so, I'm concerned that it would lead to information leaks when users forget to pass an orgID, and then their resources can be viewed and modified by everyone.  Do you intend to fix this with RBAC?

internal/service/organization.go
```diff
@@ -0,0 +1,44 @@
...
+var organizationApiVersion = fmt.Sprintf("%s/%s", api.APIGroup, api.OrganizationAPIVersion)
+
+func (h *ServiceHandler) ListOrganizations(ctx context.Context) (*api.OrganizationList, api.Status) {
+	orgs, err := h.store.Organization().List(ctx)
+	status := StoreErrorToApiStatus(err, false, OrganizationKind, nil)
+	if err != nil {
+		return nil, status
+	}
+
+	apiOrgs := make([]api.Organization, len(orgs))
```
avishayt, line 19: It seems all this code is because the org might not have a display name?  Wouldn't it be easier and more efficient to set the default upon storing it?

internal/service/organization.go
```diff
@@ -0,0 +1,44 @@
+package service
```
avishayt, line 1: Consider adding integration tests in test/integration/store/

api/v1alpha1/openapi.yaml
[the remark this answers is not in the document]
DakCrowder: Agreed, implementing now :+1:

internal/api_server/middleware/middleware.go
DakCrowder: The goal here is to have no functional changes (yet) so for now everyone would implicitly be a member which mirrors the current behavior w/ how the null or default org owns all data.

I think as the org work progresses there are a few different scenarios to consider:
- No Auth - everyone will be an implicit member of the default org
- Auth is configured but they don't care about orgs (e.g. keycloak provider is setup but orgs not configured) - everyone will be an implicit member of the default org
- Auth is configured and they care about orgs - this is where we should "deny by default" and there will be no implicit membership.  We always use the IdP as the source of truth, and if we want to provide an option to "link" the default org if it exists to some representation in the IdP to give access we can.

internal/service/organization.go
DakCrowder: It would in this case, but the question then becomes what do we do when we are pulling this information from IdPs?  I was considering always pulling the display name from some IdP value if we can get it so it stays relatively up to date in case of a change.  Although we could decide to just store it at initial create time and not care if it goes stale for now, or store it and do some comparisons / updates dynamically if we see differences.

Alternatively the display name could be exposed as a managed field (e.g. users could set a distinct display name in Flight Control that is different than anything in the IdP) but not sure if we want to expose any real organizational management features as I don't think its been noted as a requirement.

avishayt: I assumed the display name was the ID coming from the IdP.  Does the IdP have both an ID and display name?

avishayt: If the IdP has both ID and name, where do we store each of them?

internal/api_server/middleware/middleware.go
avishayt: ACK

internal/service/organization.go
DakCrowder: Refactored to store both and not have the conditional logic

api/v1alpha1/openapi.yaml
```diff
@@ -3131,12 +3133,55 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
 components:
   securitySchemes:
     bearerAuth:
       type: http
       scheme: bearer
       bearerFormat: JWT
+    orgName:
```
celdrake: Fetching resources with a non-existing org_name returns an empty list of items.
IMO it should either return 400 or 404.

```
curl -k -X 'GET' 'https://api.192.168.1.142.nip.io:3443/api/v1/devices?org_name=10000000-0000-0000-0000-000000000000&limit=2&sortBy=metadata.name&sortOrder=Asc' -H 'accept: application/json'                                  
{"apiVersion":"flightctl.io/v1alpha1","items":[],"kind":"DeviceList","metadata":{},"summary":{"applicationStatus":{},"summaryStatus":{},"total":0,"updateStatus":{}}}

```

By comparison, error 400 is returned if the org_name does not have the correct format.

```
curl -k -X 'GET' 'https://api.192.168.1.142.nip.io:3443/api/v1/devices?org_name=00000-0000-0000-0000-000000000000&limit=2&sortBy=metadata.name&sortOrder=Asc' -H 'accept: application/json' 
Invalid org_name parameter: invalid UUID length: 33

```

api/v1alpha1/openapi.yaml
```diff
@@ -3131,12 +3133,55 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/organizations:
+    get:
+      tags:
+        - organization
+      summary: List organizations
+      description: Retrieves a list of organizations.
```
celdrake: I'm assuming that the organizations that someone can retrieve would also be only those they have access to? Is it necessary to clarify in the description?

api/v1alpha1/openapi.yaml
```diff
@@ -3131,12 +3133,55 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/organizations:
```
celdrake, line 3136: The CLI command will be implemented separately?

DakCrowder: Correct, that is my plan.  Other parts of the codebase (tasks, agent api) also need to be updated as well before we can properly handle many orgs

api/v1alpha1/openapi.yaml
DakCrowder: This is an excellent point and clarification, my expectations are that in the future (once we integrate with IdPs) yes we would return only resources a user has access to (effectively a form of ReBAC) and that is worth calling out.

api/v1alpha1/openapi.yaml
DakCrowder: Updated to check and return a not found

api/v1alpha1/openapi.yaml
```diff
@@ -5298,3 +5343,44 @@ components:
           additionalProperties:
             type: string
           description: Parameters needed to retry the task.
+    Organization:
```
avishayt, line 5397: Right now, Organization has metadata and a top-level displayName field.  I recommend adopting a more consistent structure aligned with our existing K8s-style resource model by moving org-specific data into a spec object.

Suggested structure:
```
Organization:
  type: object
  required:
    - metadata
    - spec
  properties:
    metadata:
      $ref: "#/components/schemas/ObjectMeta"
    spec:
      type: object
      required:
        - externalID
      properties:
        externalID:
          type: string
        displayName:
          type: string
        isDefault:
          type: boolean
```

Why this is better:
1. Consistent with how we define Device, Fleet, etc.
2. Clean separation between system metadata (metadata) and domain-specific fields (spec)
3. Easier to extend in the future (e.g., adding org settings, sync state, etc.)
4. Avoids cluttering the top-level resource with functional fields

This also gives us a clear place to put fields like isDefault and makes future evolution more natural.

api/v1alpha1/openapi.yaml
```diff
@@ -3131,12 +3133,55 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
 components:
   securitySchemes:
     bearerAuth:
       type: http
       scheme: bearer
       bearerFormat: JWT
+    orgName:
```
avishayt: The current use of orgName as a request header is potentially confusing. In our schema, organization.metadata.name is the stable identifier, but calling the header orgName could lead users to think it should be the human-readable displayName.

Recommendation: Rename the header to Org-ID. This:
1. Clearly signals that this is an internal, stable identifier (not a UI label)
2. Aligns with conventions in other REST APIs
3. Reduces ambiguity for users unfamiliar with the K8s-style metadata.name convention
4. To avoid user error, we should also document that this header should be set to organization.metadata.name.

Let me know what you think.

internal/store/organization.go
```diff
@@ -0,0 +1,100 @@
...
+	return db.Transaction(func(tx *gorm.DB) error {
+		var count int64
+		if err := tx.Model(&model.Organization{}).Count(&count).Error; err != nil {
+			return err
+		}
+
+		// If there are no organizations, create a default one
+		if count == 0 {
+			if err := tx.Create(&model.Organization{
+				ID:          NullOrgId,
```
avishayt, line 50: We probably want to rename NullOrgId to DefaultOrgId, but not necessarily in this PR

api/v1alpha1/openapi.yaml
DakCrowder: All makes sense to me - added the `spec` and should now be more aligned w/ other resource types.

FYI _for now_ I also removed the `isDefault` field as there is no application logic that makes use of it, after thinking more I'd like to avoid special cases and would like to treat the default as "just another org" unless a clear reason not to arises.

api/v1alpha1/openapi.yaml
DakCrowder: Moved back to org_id - and updated the description to indicate the metadata.name should be used

avishayt: Change orgName to OrgId?

internal/api_server/middleware/middleware.go
```diff
@@ -57,3 +64,53 @@ func AddEventMetadataToCtx(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
+
+func AddOrgIDToCtx(orgStore store.Organization) func(http.Handler) http.Handler {
+	// Create a TTL cache with 5-minute expiration for organization validation
+	cache := ttlcache.New[uuid.UUID, bool](
```
avishayt, line 70: Fine for now, we can consider moving to Redis when we support HA API services

internal/consts/consts.go
```diff
@@ -17,4 +17,5 @@ const (
 	EventSourceComponentCtxKey ctxKey = "event_source"
 	EventActorCtxKey           ctxKey = "event_actor"
 	TLSPeerCertificateCtxKey   ctxKey = "tls-peer-certificate"
+	OrganizationIDCtxKey       ctxKey = "organization-id"
```
avishayt, line 20: For 6 consts we have 3 different formats - underscores, hyphens, and camelCase.  Can you please set them all to hyphens as long as you're editing here?

avishayt (approved):

PR #1364, opened 2025-07-13
EDM -712: Otel http metrics

SiddarthR56:
- Add edcustom FlightCtl HTTP metrics (requests_total, latencies_seconds, errors_total)
- Implemented OpenTelemetry HTTP metrics with configurable OTLP endpoints
- Differentiate agent-api and user-api traffic with service type constants
- Simplify metric labels: binary success/failed status, remove bucket dimension
- Add flexible configuration via httpOtelMetrics and prometheus.customHttpEnabled
- Use interface-based architecture with null object pattern for clean error handling
- Support graceful shutdown and resource management for production use
- Integrate metrics middleware with both main API server and agent server


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/flightctl/flightctl
 
-go 1.23
+go 1.23.0
 
-toolchain go1.23.6
+toolchain go1.23.10
```
asafbss, line 5: We cannot change the Go version — please avoid this.
Updating it would impact both the downstream build and the upstream, since we build with a specific Go toolset.
Reference: https://github.com/flightctl/flightctl/blob/016392f59c1b02b07473c7fb1e40f75f59b2a319/Containerfile.api#L1

internal/api_server/server.go
```diff
@@ -195,8 +203,8 @@ func (s *Server) Run(ctx context.Context) error {
 		ws.RegisterRoutes(r)
 	})
 
-	handler := otelhttp.NewHandler(router, "http-server")
```
asafbss, line 198: You’ve removed the tracing

internal/api_server/agentserver/server.go
```diff
@@ -143,10 +146,15 @@ func (s *AgentServer) prepareHTTPHandler(serviceHandler service.Service) (http.H
...
 
+	// Add HTTP metrics middleware if available
+	if s.httpMetrics != nil {
+		router.Use(s.httpMetrics.GetMiddleware(metrics.ServiceTypeAgentAPI))
+	}
+
 	h := transport.NewAgentTransportHandler(serviceHandler, s.ca, s.log)
 	server.HandlerFromMux(h, router)
 
-	return otelhttp.NewHandler(router, "agent-http-Server"), nil
```
asafbss, line 149: You’ve removed the tracing

internal/config/config.go
```diff
@@ -101,6 +103,13 @@ type tracingConfig struct {
 	Insecure bool   `json:"insecure,omitempty"`
 }
 
+type httpOtelConfig struct {
```
asafbss, line 106: Do we need a separate config just for HTTP? We already have prometheusConfig, which enables all metrics in the system — HTTP, system-level, and business logic.

internal/config/config.go
```diff
@@ -162,9 +171,14 @@ func NewDefault(opts ...ConfigOption) *Config {
...
 		Prometheus: &prometheusConfig{
-			Address:        ":15690",
-			SloMax:         4.0,
-			ApiLatencyBins: []float64{1e-7, 1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 1e-1, 1e0},
+			Address:           ":15690",
+			SloMax:            4.0,
+			ApiLatencyBins:    []float64{1e-7, 1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 1e-1, 1e0},
+			CustomHTTPEnabled: true,
+		},
+		HTTPOtelMetrics: &httpOtelConfig{
```
asafbss, line 179: Do we need a separate config just for HTTP? We already have prometheusConfig, which enables all metrics in the system — HTTP, system-level, and business logic.

internal/instrumentation/instrumentation.go
```diff
@@ -129,3 +131,64 @@ func InitTracer(log logrus.FieldLogger, cfg *config.Config, serviceName string)
 	log.Info("Tracing initialized")
 	return tp.Shutdown
 }
+
+// InitHTTPMetrics initializes OpenTelemetry HTTP metrics with OTLP exporter if configured.
+// The returned shutdown function should be called on application exit to ensure all metrics are flushed.
+func InitHTTPMetrics(log logrus.FieldLogger, cfg *config.Config, serviceName string) func(context.Context) error {
```
asafbss, line 137: We should avoid adding this to instrumentation.go. Can this logic be moved to http.go, similar to how it’s done in system.go? The logic in instrumentation.go should remain focused on higher-level concerns, such as general (parent) metrics and tracing. I’d like to avoid putting specific child-level logic there.

internal/instrumentation/metrics/http.go
```diff
@@ -0,0 +1,229 @@
...
+
+type HTTPMetrics interface {
+	RecordRequest(statusCode int, duration time.Duration, source string)
+	GetMiddleware(serviceType ServiceType) func(http.Handler) http.Handler
+	MetricsName() string
+	Describe(ch chan<- *prometheus.Desc)
+	Collect(ch chan<- prometheus.Metric)
+}
+
+type HTTPMetricsCollector struct {
```
asafbss, line 30: I need some clarification about this. Are we not fully relying on otelhttp, or is there additional custom logic involved? Just want to make sure I understand the design choices here.

PR #1367, opened 2025-07-14
EDM-896: Configure secure software settings

avishayt:
1. Don't use default passwords for Postgres and Redis
2. Don't use "latest" tags for software

Tested with:
- AUTH=true make deploy
- make deploy
- make deploy-quadlets

## Summary by CodeRabbit

* **New Features**
  * Added templates for Kubernetes secrets managing database and key-value store credentials with secure password generation and multi-namespace support.
  * Removed the previous secret generation job to streamline secret management.

* **Bug Fixes**
  * Cleanup job now exclusively deletes orphaned Persistent Volume Claims (PVCs), improving resource cleanup accuracy.
  * Updated Redis key-value store configuration to support automatic secure password generation when no password is specified.

* **Chores**
  * Updated PostgreSQL database container images to use a specific version tag instead of "latest" for improved stability and consistency.
  * Removed global secret generation configuration parameter to simplify deployment settings.


kkyrazis (approved):
Just a couple of questions from my side:
1. When is `generateSecrets` false? Is it just a general configuration value that we want to support? I couldn't find any instances of our automation setting it to false.
2. Why is specifying a kv store password allowed but not db passwords?

avishayt:
> Just a couple of questions from my side:
> 
> 1. When is `generateSecrets` false? Is it just a general configuration value that we want to support? I couldn't find any instances of our automation setting it to false.
> 2. Why is specifying a kv store password allowed but not db passwords?

Great questions!
The idea was to allow the user to provide existing passwords instead of generating new ones, but I saw that it wouldn't work.  I changed the approach and also made KV and DB the same behavior.

PR #1371, opened 2025-07-14
EDM-1820: Lifecycle hook templates e2e test

kkyrazis:
Adds a test that adds an inline, after updating, hook to a device that will write the contents of the files that were created or updated and the base path of the triggering hook. The test verifies that the contents are logged to the device's journal.

## Summary by CodeRabbit

* **Tests**
  * Added a new end-to-end test to verify lifecycle hooks defined with template variables, confirming proper logging of file paths and contents when files are created or updated in a monitored directory.


test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
 			Expect(logs).To(ContainSubstring("this is a test message from beforerebooting hook"))
 			Expect(logs).To(ContainSubstring("this is a test message from beforeupdating hook"))
 		})
+		It("Verifies that lifecycle hooks can be defined with template variables", Label("80022"), func() {
+			const (
+				firstMessage   = "this is a test message from afteradding hook"
```
eldar101: clearer const names please...maybe firstHookMessage?

test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
...
+				firstFile      = "temp-dir/file1.txt"
+				secondMessage  = "this is a second test message from afteradding hook"
+				secondFile     = "secondary-dir/file2.txt"
+				updatedMessage = "this has been updated"
+			)
+
+			By("Adding a template hook that prints the contents of all created/updated files")
+			nextRenderedVersion, err := harness.PrepareNextDeviceVersion(deviceId)
+			Expect(err).ToNot(HaveOccurred())
```
eldar101, line 190: Besides expecting an error, please check what the output in  nextRenderedVersion contains

test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
...
+
+			inlineConfigProviderSpec := v1alpha1.ConfigProviderSpec{}
+			err = inlineConfigProviderSpec.FromInlineConfigProviderSpec(inlineConfigValidTemplateLifeCycle)
+			Expect(err).ToNot(HaveOccurred())
+			err = harness.AddConfigToDeviceWithRetries(deviceId, inlineConfigProviderSpec)
+			Expect(err).ToNot(HaveOccurred())
+			err = harness.WaitForDeviceNewRenderedVersion(deviceId, nextRenderedVersion)
+			Expect(err).ToNot(HaveOccurred())
```
eldar101, line 199: ditto

test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
...
+
+			By("adding a new file to the hooks watch directory")
+			nextRenderedVersion, err = harness.PrepareNextDeviceVersion(deviceId)
+			Expect(err).ToNot(HaveOccurred())
```
eldar101, line 203: ditto

test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
...
+
+			provider, err := newTemplateSpecProvider("first-create", firstFile, firstMessage)
+			Expect(err).ToNot(HaveOccurred())
+			err = harness.AddConfigToDeviceWithRetries(deviceId, provider)
+			Expect(err).ToNot(HaveOccurred())
```
eldar101, line 213: ditto

test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
...
+
+			err = harness.WaitForDeviceNewRenderedVersion(deviceId, nextRenderedVersion)
+			Expect(err).ToNot(HaveOccurred())
+			// ensure we see our expected messages
+			logs, err := harness.ReadPrimaryVMAgentLogs("")
+			Expect(err).NotTo(HaveOccurred())
```
eldar101, line 218: ditto

test/e2e/hooks/hooks_test.go
```diff
@@ -174,6 +174,71 @@ var _ = Describe("Device lifecycles and embedded hooks tests", func() {
...
+			nextRenderedVersion, err = harness.PrepareNextDeviceVersion(deviceId)
+			Expect(err).ToNot(HaveOccurred())
+
+			updatedProvider, err := newTemplateSpecProvider("first-create", firstFile, updatedMessage)
+			Expect(err).ToNot(HaveOccurred())
+			// create a second file
+			newProvider, err := newTemplateSpecProvider("second-create", secondFile, secondMessage)
+			Expect(err).ToNot(HaveOccurred())
+
+			modifiedSpec := []v1alpha1.ConfigProviderSpec{
```
eldar101: please move the spec out of the block, for readability.

test/e2e/hooks/hooks_test.go
kkyrazis: The very next line is checking the output of the logs.

test/e2e/hooks/hooks_test.go
kkyrazis: What else is there to check here? `WaitForDeviceNewRenderedVersion` only returns an error which occurs if it is unable to update to the expected version. If there is no error we can be sure that we've updated properly. If we aren't happy with the `WaitForDeviceNewRenderedVersion` function and would prefer for all of our code to explicitly check for renderedVersion changes we can make those changes too. Do you just want me to enhance the error message so that it's clear that we failed to update to our expected version?

test/e2e/hooks/hooks_test.go
kkyrazis: Why is that relevant? All we really care about is that we have some steady state version to start with. If today all devices start at 1 and in a month they all start at 10 it doesn't really matter for this test. The future checks just use this to ensure that the device has successfully applied the spec by checking that the currentVersion == expectedVersion

keitwb: I agree; the whole point of `PrepareNextDeviceVersion` is to query the API for the current device version and increment it, to relieve the test of the burden of having to keep up with the current version. If there is some concern about versions getting out of sync somehow that should be handled by using a different harness method and not done explicitly in each test node as that would be very distracting from the test content if followed consistently.

PR #1376, opened 2025-07-14
NO-ISSUE: Fix slow integration tests

kkyrazis:
Agent startup involves attempting to resolve network information. It polls for about 45 seconds waiting for a route to use for diagnostic reasons. If no route is found the agent will continue its startup after logging a warning. This means that every agent integration test is waiting 45 seconds for startup.  (6.5 minutes down to 30 seconds when running locally). 

This fix just copies the root `/proc/net/route` table into the temp dir `/tmp/some-dir/proc/net/route` so that tests can startup immediately. 

Unfortunately it can't just be a static route table with fake data filled in as there is logic that ensures the routes returned match the currently available interfaces. 

## Summary by CodeRabbit

* **New Features**
  * The test environment now includes a mock network route table file, improving support for agent onboarding scenarios.

* **Tests**
  * Enhanced the test harness to better simulate system networking conditions by replicating the route table structure.


eldar101:
LGTM, if others don't have comments, ping me and I'll approve.

PR #1377, opened 2025-07-14
NO-ISSUE: Reassign the qcow image back to the current user

keitwb:
Sudo is invoked purely for the build process and not so that the output image is owned by root so it seems appropriate to assign it back to the build user in the script itself.

## Summary by CodeRabbit

* **Chores**
  * Updated image creation script to ensure output files are owned by the user running the script.


hexfusion (approved):
seems reaonable, thanks @keitwb!

PR #1379, opened 2025-07-14
EDM-1839: Build binaries using FIPS crypto

fzdarsky:
## Summary by CodeRabbit

* **Chores**
  * Improved build configuration by centralizing and simplifying environment variable and flag management for various build scenarios.
  * Updated multi-architecture CLI build process to explicitly set the FIPS mode flag during builds.
  * Added new build targets for device simulator and base container image creation.
  * Updated container images to use a custom base image simplifying certificate management.
  * Enhanced CI workflows to disable FIPS mode during testing, deployment, and publishing steps.
  * Expanded developer prerequisites with additional container tooling requirements.
  * Introduced new container image and script for RPM building environment setup.

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -7,7 +7,7 @@ Flight Control is a service for declarative, GitOps-driven management of edge de
 ## Building
 
 Prerequisites:
-* `git`, `make`, and `go` (>= 1.23), `openssl`, `openssl-devel`, `podman-compose` and `go-rpm-macros` (in case one needs to build RPM's)
+* `git`, `make`, and `go` (>= 1.23), `openssl`, `openssl-devel`, `buildah`, `podman`, `podman-compose`, and `go-rpm-macros` (in case one needs to build RPM's)
 
 Flightctl agent reports the status of running rootless containers. Ensure the podman socket is enabled:
 
--- a/hack/build_flightctl-base.sh
+++ b/hack/build_flightctl-base.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+IMAGE_TAG=9.6-1752500771
+
+container=$(buildah from registry.redhat.io/ubi9-micro:$IMAGE_TAG)
+
+mountdir=$(buildah mount $container)
+dnf install \
+    --installroot $mountdir \
+    --releasever 9 \
+    --setopt install_weak_deps=false \
+    --nodocs -y \
+    openssl-libs tzdata
+dnf clean all \
+    --installroot $mountdir
+buildah umount $container
+
+buildah commit $container flightctl-base:$IMAGE_TAG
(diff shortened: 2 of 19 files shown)


avishayt:
We also need to build in FIPS-enabled environments, right?

fzdarsky:
> We also need to build in FIPS-enabled environments, right?

For downstream probably yes. For upstream, too?
Currently focussing on building with FIPS-compliant toolchain and need to switch away from building on Ubuntu host for this.

hack/build_flightctl-base.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
rawagner, line 1: why this script instead of Containerfile ?

hack/build_flightctl-base.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+
+IMAGE_TAG=9.6-1752500771
+
+container=$(buildah from registry.redhat.io/ubi9-micro:$IMAGE_TAG)
+
+mountdir=$(buildah mount $container)
+yum install \
```
rawagner: dnf instead of yum?

hack/build_flightctl-base.sh
fzdarsky: We're using a ubi-micro image that does not contain `dnf` or `microdnf`. To install RPMs, one therefore needs to use `buildah` instead of Podman. Alternative would be to install to a builder image and the copy from there, but that gets finnicky with more than a few files. Here, we need to install `openssl-libs` shared libraries correctly.

packaging/rpm/flightctl.spec
```diff
@@ -414,7 +414,11 @@ echo "Flightctl Observability Stack uninstalled."
     SOURCE_GIT_TREE_STATE=clean \
     SOURCE_GIT_COMMIT=$(echo %{version} | awk -F'[-~]g' '{print $2}') \
     SOURCE_GIT_TAG_NO_V=%{version} \
-    make build-cli build-agent
+    %if 0%{?rhel} == 9
```
hexfusion, line 417: what about rhel 10 should we do `>=`

docs/developer/README.md
```diff
@@ -7,7 +7,7 @@ Flight Control is a service for declarative, GitOps-driven management of edge de
 ## Building
 
 Prerequisites:
-* `git`, `make`, and `go` (>= 1.23), `openssl`, `openssl-devel`, `podman-compose` and `go-rpm-macros` (in case one needs to build RPM's)
+* `git`, `make`, and `go` (>= 1.23), `openssl`, `openssl-devel`, `buildah`, `podman`, `podman-compose`, and `go-rpm-macros` (in case one needs to build RPM's)
```
hexfusion, line 10: :+1:

packaging/rpm/flightctl.spec
fzdarsky: We probably should, but building for RHEL 10 will fail elsewhere first I’m afraid.

hexfusion (approved):

PR #1382, opened 2025-07-15
EDM-1545: Certificate manager infrastructure to flightctl agent

asafbss:


internal/agent/device/certmanager/provider/provider.go
```diff
@@ -0,0 +1,176 @@
...
+		return false
+	}
+	if !c.Storage.Equal(other.Storage) {
+		return false
+	}
+	if c.RenewalThreshold != other.RenewalThreshold {
+		return false
+	}
+	return true
+}
```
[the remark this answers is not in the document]
asafbss: fixed

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+func (f *FileSystemStorageFactory) Validate(log provider.Logger, cc provider.CertificateConfig) error {
+	storage := cc.Storage
+
+	if storage.Type != "filesystem" {
+		return fmt.Errorf("not a filesystem Storage")
+	}
+
+	var fsConfig FileSystemStorageConfig
+	if err := json.Unmarshal(storage.Config, &fsConfig); err != nil {
+		return fmt.Errorf("failed to decode CSR config for certificate %q: %w", cc.Name, err)
```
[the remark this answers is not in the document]
asafbss: fixed

internal/agent/client/management.go
```diff
@@ -111,3 +111,55 @@ func (m *management) GetRenderedDevice(ctx context.Context, name string, params
...
+	if resp.JSON400 != nil {
+		return nil, resp.StatusCode(), fmt.Errorf("create certificate signing request failed: %s", resp.JSON400.Message)
+	}
+
+	if resp.JSON201 != nil {
+		return resp.JSON201, resp.StatusCode(), nil
+	}
+
+	return nil, resp.StatusCode(), nil
+}
```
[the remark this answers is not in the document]
asafbss: fixed

internal/agent/device/certmanager/provider/config/file.go
```diff
@@ -0,0 +1,103 @@
...
+				continue
+			}
+
+			if !bytes.Equal(sum, p.lastHash) {
+				p.changeNotifier.TriggerConfigChange()
+				p.lastHash = sum
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
asafbss: RegisterConfigChangeChannel will be called once by the cert manager

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+	KeyPath string `json:"key-path"`
+	// Permissions for the certificate and key files
+	Permissions *int `json:"permissions,omitempty"`
+}
+
+// FileSystemStorage handles certificate storage on the local filesystem.
+// It stores certificates and private keys as managed files with appropriate permissions
+// and supports loading existing certificates from the filesystem.
+type FileSystemStorage struct {
+	CertPath         string            // Path where the certificate file will be stored
```
hexfusion: nit: can comments go above struct fields please

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+	KeyPath          string            // Path where the private key file will be stored
+	Permissions      int               // File permissions for certificate and key files
+	deviceReadWriter fileio.ReadWriter // File I/O interface for reading and writing files
+	log              provider.Logger   // Logger for storage operations
+}
+
+// NewFileSystemStorage creates a new filesystem storage provider with the specified configuration.
+// It uses the provided file I/O interface and logger for operations.
+func NewFileSystemStorage(certPath, keyPath string, permissions int, rw fileio.ReadWriter, log provider.Logger) *FileSystemStorage {
```
hexfusion: what is the usecase for declaring permissions instead of enforcing best practices by default?

hexfusion: do we need a separate logger interface? Can we achieve the same functionality with the log.PrefixLogger?

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+		deviceReadWriter: rw,
+		log:              log,
+	}
+}
+
+// LoadCertificate loads a certificate from the filesystem.
+// It reads the certificate file and parses it as a PEM-encoded X.509 certificate.
+func (fs *FileSystemStorage) LoadCertificate(ctx context.Context) (*x509.Certificate, error) {
+	certPath := fs.deviceReadWriter.PathFor(fs.CertPath)
+	certPEM, err := fs.deviceReadWriter.ReadFile(certPath)
```
hexfusion: you can use readWriter with the regular path PathFor is used only outside the context of fileio. Here I believe the path will end up being duplicated
```suggestion
	certPEM, err := fs.deviceReadWriter.ReadFile(fs.CertPath)
```

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+	cert, err := fccrypto.ParsePEMCertificate(certPEM)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse PEM certificate: %w", err)
+	}
+	return cert, nil
+}
+
+// Write stores a certificate and private key to the filesystem.
+// It creates the necessary directories and writes both files with appropriate permissions.
+// The files are stored as base64-encoded managed files for consistency with the agent's file management.
```
hexfusion: question: can you clarify is base64 needed for storing on disk when you say consistency with agent file management do you mean in the context of cert-manager or in general.

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// writeBase64 writes data to a file as a base64-encoded managed file.
+// This ensures consistency with the agent's file management system and proper permissions.
+func (fs *FileSystemStorage) writeBase64(path string, data []byte) error {
+	content := base64.StdEncoding.EncodeToString(data)
+	mode := fs.Permissions
+	encoding := v1alpha1.EncodingBase64
+
+	mf, err := fs.deviceReadWriter.CreateManagedFile(v1alpha1.FileSpec{
```
hexfusion: we only need managed files in the case where we are trying to show change unless you need that its probably better to use WriteFile directly and avoid base64 encoding on disk unless there is another reason to do so.

ref. https://github.com/flightctl/flightctl/blob/880c3985c932e4f97893c1d70adefd0f756e5bc3/internal/agent/device/config/controller.go#L103-L117

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return NewFileSystemStorage(fsConfig.CertPath, fsConfig.KeyPath, permissions, f.rw, log), nil
+}
+
+// Validate checks whether the provided configuration is valid for filesystem storage.
+// It ensures required fields are present and the configuration is properly formatted.
+func (f *FileSystemStorageFactory) Validate(log provider.Logger, cc provider.CertificateConfig) error {
+	storage := cc.Storage
+
+	if storage.Type != "filesystem" {
```
hexfusion: nit: can we make this an enum

internal/agent/agent.go
```diff
@@ -14,6 +14,11 @@ import (
 	agent_config "github.com/flightctl/flightctl/internal/agent/config"
 	"github.com/flightctl/flightctl/internal/agent/device"
 	"github.com/flightctl/flightctl/internal/agent/device/applications"
+	"github.com/flightctl/flightctl/internal/agent/device/certmanager"
+	cm_config "github.com/flightctl/flightctl/internal/agent/device/certmanager/provider/config"
+	cm_provisioner "github.com/flightctl/flightctl/internal/agent/device/certmanager/provider/provisioner"
+	cm_state "github.com/flightctl/flightctl/internal/agent/device/certmanager/provider/state"
+	cm_storage "github.com/flightctl/flightctl/internal/agent/device/certmanager/provider/storage"
```
hexfusion: question: could we have a single constructor import here or are these nessisary?

internal/agent/client/client.go
```diff
@@ -46,6 +46,8 @@ type Management interface {
 	GetRenderedDevice(ctx context.Context, name string, params *v1alpha1.GetRenderedDeviceParams, rcb ...client.RequestEditorFn) (*v1alpha1.Device, int, error)
 	PatchDeviceStatus(ctx context.Context, name string, patch v1alpha1.PatchRequest, rcb ...client.RequestEditorFn) error
 	SetRPCMetricsCallback(cb RPCMetricsCallback)
+	CreateCertificateSigningRequest(ctx context.Context, csr v1alpha1.CertificateSigningRequest, rcb ...client.RequestEditorFn) (*v1alpha1.CertificateSigningRequest, int, error)
+	GetCertificateSigningRequest(ctx context.Context, name string, rcb ...client.RequestEditorFn) (*v1alpha1.CertificateSigningRequest, int, error)
```
hexfusion, line 50: nit: can we create a seperate interface that only exposes these? the cert-manager does not need full access to the management client.

internal/agent/config/config.go
```diff
@@ -113,6 +114,11 @@ type Config struct {
 	// PullRetrySteps defines how many retry attempts are allowed for pulling an OCI target.
 	PullRetrySteps int `json:"pull-retry-steps,omitempty"`
 
+	// Certificates defines the certificates to be managed by the certificate manager.
+	// These certificates are automatically provisioned, renewed, and stored based on their configuration.
+	// Each certificate can use different provisioners (CSR, self-signed) and storage backends (filesystem).
+	Certificates []provider.CertificateConfig `json:"certificates,omitempty"`
```
hexfusion: nit: i'd prefer the cert-manager config be separate from the agent. this separation has several benefits:

- parsing failures in cert config won’t block the agent (temporary otel downtime is acceptable).

- the same config pattern can be reused by internal agent-driven apps: /etc/compose/manifests/*/certs{.yaml,.d/*.yaml}

```sh
/etc/flightctl/
  config.yaml          # agent-only (tiny, stable)
  certs.yaml           # base cert-manager config
  certs.d/             # drop-ins for per-cert requests or overrides
```

internal/agent/device/bootstrap.go
```diff
@@ -302,3 +304,44 @@ func (b *Bootstrap) setManagementClient() error {
...
+	}
+
+	configBytes, err := json.Marshal(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal storage config: %w", err)
+	}
+
+	return []provider.CertificateConfig{{
+		Name:       "device",        // Primary device certificate for management server communication
+		AllowRenew: lo.ToPtr(false), // Disable renewal: this certificate is managed externally
```
hexfusion: nit: for this PR can we remove/make dead all cert rotation code? we can bring it back once we have consensus on that process/design. :pray:

internal/agent/device/certmanager/cert_processing_queue.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+// processHandlerFunc defines the function signature used to process a certificate.
+// It receives the certificate processing context and returns a retry delay if the operation
+// should be retried, or nil if the operation completed successfully or should not be retried.
+type processHandlerFunc func(ctx context.Context, providerName string, cert *certificate, cfg *provider.CertificateConfig) *time.Duration
+
+// certificateProcess represents an active certificate processing task.
+type certificateProcess struct {
+	ctx      context.Context            // Context for this processing task, can be canceled
```
hexfusion: nit: there should be no need to store the context in the struct. context should be passed and if you need to cancel you can call the cancelFn.

internal/agent/device/certmanager/cert_processing_queue.go
```diff
@@ -0,0 +1,172 @@
...
+	cancel   context.CancelFunc         // Cancel function to stop processing
+	provider string                     // Name of the configuration provider
+	cert     *certificate               // Certificate being processed
+	cfg      provider.CertificateConfig // Configuration for certificate processing
```
hexfusion: is there a reason we are copying the configuration instead of a reference to the config?

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,649 @@
...
+const DefaultRequeueDelay = 10 * time.Second
+
+// CertManager manages the complete certificate lifecycle for flight control agents.
+// It coordinates certificate provisioning, storage, renewal, and cleanup across multiple
+// configuration providers and implements retry logic for failed operations.
+// The manager supports pluggable provisioners (CSR, self-signed, etc.) and storage
+// backends (filesystem, etc.) through factory patterns.
+type CertManager struct {
+	log             provider.Logger
+	certificates    *certStorage                           // In-memory certificate state with optional persistent backing
```
hexfusion: nit: if comments are needed, please add above struct fields

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,649 @@
...
+	if !ready {
+		return &cm.requeueDelay, nil
+	}
+
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	// check storage drift
+	if !cert.Config.Storage.Equal(cfg.Storage) {
```
hexfusion, line 399: the only way for the cert manager to get a new config would be after a SIGHUP signal could you use the reload manager to observe config changes?

https://github.com/flightctl/flightctl/blob/main/internal/agent/reload/reload.go#L81

internal/agent/device/certmanager/provider/provider.go
```diff
@@ -0,0 +1,196 @@
...
+type ConfigProvider interface {
+	// Name returns a unique identifier for this configuration provider
+	Name() string
+	// GetCertificateConfigs returns all certificate configurations from this provider
+	GetCertificateConfigs() ([]CertificateConfig, error)
+}
+
+// SupportsNotify is an optional interface for configuration providers that can notify
+// about configuration changes. This enables reactive certificate management.
+type SupportsNotify interface {
```
hexfusion: question: is this used?

internal/agent/device/certmanager/provider/provider.go
```diff
@@ -0,0 +1,196 @@
...
+// This is typically called during certificate manager initialization.
+func (n *ConfigProviderChangeNotifier) RegisterConfigChangeChannel(ch chan<- ConfigProvider, cp ConfigProvider) error {
+	n.self = cp
+	n.changeCh = ch
+	return nil
+}
+
+// TriggerConfigChange sends a configuration change notification to the registered channel.
+// This should be called when the configuration provider detects a change in its configuration.
+func (n *ConfigProviderChangeNotifier) TriggerConfigChange() {
```
hexfusion: the config can only change if the agent recieves a SIGHUP signal for example during `systemctl reload flightclt-agent`. can we use the current agent/reload manager?

internal/agent/device/certmanager/provider/provisioner/self_signed.go
```diff
@@ -0,0 +1,150 @@
...
+
+// Provision generates a self-signed certificate using a temporary certificate authority.
+// It creates a new CA, generates a private key and CSR, then uses the CA to sign the certificate.
+// This method always returns ready=true since self-signed certificates are generated synchronously.
+func (p *SelfSignedProvisioner) Provision(ctx context.Context) (bool, *x509.Certificate, []byte, error) {
+	if p.cfg.CommonName == "" {
+		return false, nil, nil, fmt.Errorf("common name must be set for self-signed certificate")
+	}
+
+	tmpDir, err := os.MkdirTemp("", "self-signed-ca-*")
```
hexfusion: nit: if we are in the context of the agent all file interactions should be with fileio.

internal/agent/device/certmanager/provider/state/filestorage.go
```diff
@@ -0,0 +1,57 @@
+package state
```
hexfusion: question: do we need this package or could this be provider.FileStorage

internal/agent/device/certmanager/retryqueue.go
```diff
@@ -0,0 +1,91 @@
...
+// queuedItem represents an item in the retry queue with its attempt count.
+// The attempt count is used to implement exponential backoff or attempt-based logic.
+type queuedItem[T any] struct {
+	value   T   // The item to be processed
+	attempt int // Number of processing attempts (0 for first attempt)
+}
+
+// NewRetryQueue creates a new RetryQueue with the given handler function.
+// The handler will be called for each item in the queue and can control retry behavior.
+func NewRetryQueue[T any](handler HandlerFunc[T]) *RetryQueue[T] {
```
hexfusion, line 42: The agent is starting to grow a few generic queues please review the usage in the publisher and see if we can align I don't care which one as much as growing more.

https://github.com/flightctl/flightctl/blob/main/internal/agent/device/publisher/publisher.go#L22

internal/agent/device/certmanager/provider/state/filestorage.go
```diff
@@ -0,0 +1,57 @@
...
+func NewFileStorage(path string) *FileStorage {
+	return &FileStorage{
+		filePath: path,
+	}
+}
+
+// StoreState reads data from the provided reader and writes it to the file.
+// This implements the StateStorageProvider interface for persisting certificate state.
+func (f *FileStorage) StoreState(r io.Reader) error {
+	file, err := os.Create(f.filePath)
```
hexfusion: please use fileio in this package.

internal/agent/device/certmanager/provider/storage/empty.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"crypto/x509"
+	"fmt"
+
+	"github.com/flightctl/flightctl/internal/agent/device/certmanager/provider"
+)
+
+// EmptyStorage is a placeholder storage provider that always fails storage operations.
+// It's used for testing, development, or scenarios where certificates should not be persisted.
+// All operations return errors indicating this is an empty storage provider.
```
hexfusion: if this is just for testing can it live in one of the _test files?

hexfusion: Speaking of there are no tests :)

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: I’d keep the small provider.Logger interface for decoupling and easier testing/mocking. Since log.PrefixLogger already implements it; we’re not adding overhead, but we gain flexibility to swap/mock in tests.

internal/agent/client/client.go
asafbss: I’ve added a minimal CSR-specific interface and updated the CSR provisioner to depend on it. This interface is only implemented and used within the CSR provisioner — the cert manager itself doesn’t take or use it directly.

internal/agent/config/config.go
asafbss: Hey Sam, just to clarify the ask around separating certs.yaml and certs.d/:

What I’m thinking is to load a base config file (certs.yaml) and then merge in any drop-ins from certs.d/. Each drop-in would define one or more certificates, and if a drop-in has the same name as a base cert, it overrides it. That way we get a clean layering (base + overrides) and can safely handle partial updates without breaking the agent.

Does this match what you had in mind? Also, do you expect us to support passing explicit paths for the config file and drop-in directory via CLI args, or should we assume fixed well-known paths under /etc/flightctl/?

internal/agent/device/certmanager/cert_processing_queue.go
asafbss: We intentionally copy CertificateConfig into the work item rather than keep a reference. This gives the handler a stable snapshot for the entire processing/retry cycle and avoids races if the provider reloads and mutates the live config while work is in flight. The struct is small, so the copy cost is negligible compared to the safety of preventing mid-run config drift

internal/agent/device/certmanager/manager.go
asafbss: Let’s keep config updates provider-driven with the periodic sync as a backstop. Providers already expose change signals (SupportsNotify/TriggerConfigChange), and the cert-manager registers for those; providers that can’t notify are still covered by the periodic reconcile (hourly by default).

Using the agent’s SIGHUP via the reload manager would couple cert-manager behavior to agent-level signals instead of the provider abstraction. This also limits future providers (e.g., HTTP/remote) that can notify directly without a process signal.

internal/agent/device/certmanager/provider/provider.go
asafbss: Yes—it's used. In certmanager.WithConfigProvider we type-assert the provider to SupportsNotify and, if it implements it, we call RegisterConfigChangeChannel(...). That lets providers invoke TriggerConfigChange() to drive an immediate resync (on top of the periodic sync).

internal/agent/device/certmanager/manager.go
asafbss: https://github.com/flightctl/flightctl/pull/1382#discussion_r2282103495

internal/agent/device/certmanager/provider/provider.go
asafbss: https://github.com/flightctl/flightctl/pull/1382#discussion_r2282099281

asafbss: WDYT?

internal/agent/device/certmanager/provider/provider.go
asafbss: WDYT?

internal/agent/device/certmanager/manager.go
asafbss: WDYT?

internal/agent/config/config.go
hexfusion: Yeah, everything you need to inject into the agent would be covered in the cert.yaml. Then any updates during runtime would be drop-in via cert.d. The cert-manager would gain access to the new config after receiving `SIGHUP` signal. In terms of paths you would only require one.  This is the same as for example systemd unit/drop-in model.

internal/agent/device/certmanager/manager.go
hexfusion: Signal based reload is an established pattern. As the manager is part of the agent process using the established reload signal makes sense. Watching the filesystem for config change seems like an entirely new pattern. Adding that for possible future features does not feel like what we want. What am I missing? What value does it bring today?

internal/agent/device/certmanager/provider/provider.go
hexfusion: If the goal is to allow configuration change based on signal. IE systemctl reload flightctl-agent (SIGHUP) then that makes sense. I just didnt see any reference to its use on first review.

internal/agent/device/certmanager/provider/provider.go
hexfusion: since any config change would be the result of a config API written to disk and observable by the hooks system, I would like to see us using existing patterns instead of creating new.

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,699 @@
...
+			// Remove old queued item
+			cm.processingQueue.Remove(providerName, cert.Name)
+
+			// Re-queue with new config
+			if err := cm.renewCertificate(ctx, providerName, cert, cfg); err != nil {
+				return fmt.Errorf("failed to renew certificate %q from provider %q: %w", cert.Name, providerName, err)
+			}
+			cm.log.Infof("Config changed during processing — re-queued renewal for certificate %q of provider %q", certName, providerName)
+		}
+		return nil
```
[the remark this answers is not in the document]
asafbss: Not accurate, please review: func (c CertificateConfig) Equal(other CertificateConfig) bool

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,705 @@
...
+	// Attempt to ensure certificate (provision or renew)
+	retryDelay, err := cm.ensureCertificate_do(ctx, providerName, cert, cfg)
+	if err != nil {
+		// On failure, reset provisioner and storage to force re-init next time
+		cert.Provisioner = nil
+		cert.Storage = nil
+		cert.RetryFailures++
+		cert.Err = err.Error()
+		return nil
+	}
```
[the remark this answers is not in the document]
asafbss: This is by design...The 'retry' is just for the provisioner until it gets ready.
Ready also means 'error', and we stop the processing.

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,167 @@
...
+// since deletion is a cleanup operation.
+func (fs *FileSystemStorage) Delete(ctx context.Context) error {
+	if err := fs.deviceReadWriter.RemoveFile(fs.CertPath); err != nil {
+		fs.log.Warnf("failed to delete cert file %s: %v", fs.CertPath, err)
+	}
+	if err := fs.deviceReadWriter.RemoveFile(fs.KeyPath); err != nil {
+		fs.log.Warnf("failed to delete key file %s: %v", fs.KeyPath, err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
asafbss: Storage providers will always receive a logger via certManager.
certManager is checking its logger itself.

internal/agent/agent.go
```diff
@@ -259,6 +260,22 @@ func (a *Agent) Run(ctx context.Context) error {
...
 
+	// Initialize certificate
+	certManager, err := certmanager.NewManager(a.log,
+		certmanager.WithConfigProvider(bootstrap),
+		certmanager.WithBuiltins(
+			ctx,
+			deviceName,
+			bootstrap.ManagementClient(),
+			deviceReadWriter,
+		),
```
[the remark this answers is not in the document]
asafbss: We want the config dir

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,699 @@
...
+	return nil, nil
+}
+
+// addCertificateInfo extracts and stores certificate information from a parsed X.509 certificate.
+func (cm *CertManager) addCertificateInfo(cert *certificate, parsedCert *x509.Certificate) {
+	cert.Info.NotBefore = &parsedCert.NotBefore
+	cert.Info.NotAfter = &parsedCert.NotAfter
+	cert.Info.CommonName = &parsedCert.Subject.CommonName
+	cert.Info.SerialNumber = lo.ToPtr(parsedCert.SerialNumber.String())
+}
```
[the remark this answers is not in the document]
asafbss: fixed

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,699 @@
...
+		// Mark existing certificates as handled so they won't be deleted
+		providerObj, err := cm.certificates.EnsureProvider(providerName)
+		if err != nil {
+			return err
+		}
+		for _, cert := range providerObj.Certificates {
+			handledCertificates = append(handledCertificates, cert.Name)
+		}
+		return fmt.Errorf("failed to load certificate configs from provider %q: %w", providerName, err)
+	}
```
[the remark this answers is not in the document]
asafbss: changed, please check again

internal/agent/device/certmanager/builtins.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		// Provisioner providers
+		if err := WithProvisionerProvider(provisioner.NewCSRProvisionerFactory(deviceName, managementClient))(cm); err != nil {
+			return err
+		}
+
+		// Storage providers
+		if err := WithStorageProvider(storage.NewFileSystemStorageFactory(readWriter))(cm); err != nil {
+			return err
```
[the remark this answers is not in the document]
asafbss: No, self-signed is not part of the built-ins. 
It's mainly gonna be used for tests

internal/agent/agent.go
```diff
@@ -259,6 +260,22 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		// Temporary: disable cert rotation at this stage; see WithRotationDisabled.
+		certmanager.WithRotationDisabled(),
+	)
```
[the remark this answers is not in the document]
asafbss: No, we do want the config dir

internal/agent/device/certmanager/manager.go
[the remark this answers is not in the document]
asafbss: Correct, if the provisioner wants to delay and retry, it can implement it.

internal/agent/agent.go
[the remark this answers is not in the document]
asafbss: You were right.
'cert-state.json' should be inside the data dir.
And it should come from a config instance, check the latest, please.

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	var snapshot map[string]*certificateProvider
+	if err := json.NewDecoder(&buf).Decode(&snapshot); err != nil {
+		return fmt.Errorf("failed to decode state: %w", err)
+	}
+
+	s.providers = snapshot
+	return nil
+}
```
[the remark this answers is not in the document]
asafbss: Thank you, check the latest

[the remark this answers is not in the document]
asafbss: Thank you, check the latest

internal/agent/device/certmanager/provider/provisioner/csr.go
```diff
@@ -0,0 +1,263 @@
...
+
+	if p.cfg.CommonName == "" {
+		return false, nil, nil, fmt.Errorf("commonName must be set")
+	}
+
+	// Generate unique CSR object name for Kubernetes resource
+	p.csrName = fmt.Sprintf("%s-%s", p.cfg.CommonName, uuid.NewString()[:8])
+
+	// Generate private key and CSR using the configured CommonName (without suffix)
+	key, csr, err := generateKeyAndCSR(p.cfg.CommonName)
```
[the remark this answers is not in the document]
asafbss: No...worst case it will be rejected by the server and a corresponding error will be written.

internal/agent/device/certmanager/provider/storage/fs.go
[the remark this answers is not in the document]
asafbss: fixed, check again

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading cert file: %w", err)
+	}
+
+	cert, err := fccrypto.ParsePEMCertificate(certPEM)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse PEM certificate: %w", err)
+	}
+	return cert, nil
+}
```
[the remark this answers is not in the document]
asafbss: fixed, check again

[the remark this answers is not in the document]
asafbss: fixed, check again

internal/agent/device/certmanager/provider/provisioner/csr.go
```diff
@@ -0,0 +1,263 @@
...
+// This provisioner generates a private key and CSR, submits it to the management server,
+// and waits for approval and signing by the certificate authority.
+type CSRProvisionerConfig struct {
+	// Signer is the name of the signer for CSR provisioning
+	Signer string `json:"signer"`
+	// CommonName is the common name for the certificate
+	CommonName string `json:"common-name,omitempty"`
+	// KeyUsage specifies the key usage for the certificate (e.g., "clientAuth", "serverAuth")
+	KeyUsage []string `json:"key-usage,omitempty"`
+	// ExpirationSeconds requests a specific certificate validity duration (in seconds); signer may ignore
```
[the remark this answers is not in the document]
asafbss: fixed, check again

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,714 @@
...
+// asynchronous sync request that the manager's Run loop will handle.
+func (cm *CertManager) SyncNow(ctx context.Context, _ *config.Config) error {
+	select {
+	case cm.syncRequestCh <- struct{}{}:
+	default:
+	}
+
+	cm.log.Info("Starting certificate sync")
+	return nil
+}
```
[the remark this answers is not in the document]
asafbss: fixed, check again

internal/agent/device/certmanager/cert_processing_queue.go
```diff
@@ -0,0 +1,185 @@
...
+		if cp.ctx.Err() != nil {
+			q.Remove(cp.provider, cp.cert.Name)
+			return nil
+		}
+		return retry
+	}
+
+	q.Remove(cp.provider, cp.cert.Name)
+	return nil
+}
```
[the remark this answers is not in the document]
asafbss: fixed, check again

internal/agent/device/certmanager/builtins.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// WithBuiltins registers the standard certificate manager providers and factories.
+func WithBuiltins(
+	ctx context.Context,
+	deviceName string,
+	managementClient client.Management,
+	readWriter fileio.ReadWriter,
+	cfg *agent_config.Config,
+) ManagerOption {
```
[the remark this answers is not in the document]
asafbss: fixed, check again

internal/agent/device/certmanager/provider/storage/empty.go
asafbss: I’ve removed the empty provider/storage entirely.
With your permission, I’ll add tests in a follow-up PR to keep this change focused

internal/agent/device/certmanager/provider/state/filestorage.go
asafbss: Done

internal/agent/device/certmanager/provider/provisioner/self_signed.go
asafbss: Done

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: Done

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: Done - removed

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: fixed

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: Done

internal/agent/device/certmanager/retryqueue.go
asafbss: These “queues” serve different contracts and can’t be merged without losing behavior:

Publisher’s RingBuffer (internal/agent/device/ring_buffer): bounded FIFO with overwrite-on-full semantics. Push may drop the oldest when capacity is reached; Pop blocks until data arrives; also has TryPop. This is a lossy, throughput-oriented buffer by design.

Cert-manager RetryQueue (retryqueue.go): not a ring and currently unbounded ([]queuedItem[T]). It drives a single worker loop, supports timed requeue/backoff (via time. After and attempts), and aims for at-least-once processing. No overwrite semantics.

Because one is a bounded, lossy ring and the other is an unbounded, retry-aware worker, unifying them would either add unnecessary timers/backoff to the publisher path or weaken retry guarantees in cert-manager.

internal/agent/device/certmanager/provider/state/filestorage.go
asafbss: Hey, we do need a separate state package. This is specifically about the CertManager’s internal state, which can be persisted via a pluggable storage backend (e.g., state.FileStorage). The idea is that this state can later be swapped with a different implementation without touching **state** storage logic.

provider.FileStorage, on the other hand, implements the **certificate** storage interface that handles persisting issued certs/keys as part of the certificate lifecycle. That’s orthogonal to the CertManager’s state management.

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: Done

internal/agent/agent.go
asafbss: Done, see 'WithBuiltins'

internal/agent/client/client.go
asafbss: Done

internal/agent/config/config.go
asafbss: Done, I didn't include a CLI arg at this stage

internal/agent/device/bootstrap.go
asafbss: Done, see 'certmanager.WithRotationDisabled()'

internal/agent/device/certmanager/cert_processing_queue.go
asafbss: We actually need to keep the context in the struct because it’s passed down and used later when we call process(). The cancelFn alone isn’t enough — the process() logic requires access to the full context for cancellation and propagation.

internal/agent/device/certmanager/manager.go
asafbss: Done

internal/agent/device/certmanager/manager.go
asafbss: Done -  reloadmanager integration was added

internal/agent/device/certmanager/provider/provider.go
asafbss: Done - 'SupportsNotify' was removed, reloadmanager integration was added

internal/agent/device/certmanager/provider/provider.go
asafbss: Done - 'SupportsNotify' was removed, reloadmanager integration was added

internal/agent/agent.go
```diff
@@ -320,11 +337,13 @@ func (a *Agent) Run(ctx context.Context) error {
 	reloadManager.Register(agent.ReloadConfig)
 	reloadManager.Register(systemInfoManager.ReloadConfig)
 	reloadManager.Register(statusManager.ReloadCollect)
+	reloadManager.Register(certManager.SyncNow)
```
hexfusion: :+1:

internal/agent/device/bootstrap.go
```diff
@@ -294,3 +296,39 @@ func (b *Bootstrap) setManagementClient() error {
...
+// Name returns the configuration provider name for the certificate manager.
+// This identifies the bootstrap provider in certificate manager logs and operations.
+func (p *Bootstrap) Name() string {
+	return "agent-bootstrap"
+}
+
+// GetCertificateConfigs provides the bootstrap certificate configuration to the certificate manager.
+// This tracks the existing device certificate used for management server communication.
+// The certificate is marked as non-renewable since it's managed outside the certificate manager.
+func (p *Bootstrap) GetCertificateConfigs() ([]provider.CertificateConfig, error) {
```
hexfusion: Bootstrap is not expected to be portable, since there is nothing really exposed by these methods from bootstrap I am wondering if this can be done in another way. Note how other managers get access to the ManagementClient. I get your trying to implement an interface but the agent is trying to not share this logic.

internal/agent/device/certmanager/cert_processing_queue.go
```diff
@@ -0,0 +1,185 @@
...
+
+// processHandlerFunc defines the function signature used to process a certificate.
+// It receives the certificate processing context and returns a retry delay if the operation
+// should be retried, or nil if the operation completed successfully or should not be retried.
+type processHandlerFunc func(ctx context.Context, providerName string, cert *certificate, cfg *provider.CertificateConfig) *time.Duration
+
+// certificateProcess represents an active certificate processing task.
+type certificateProcess struct {
+	// Context for this processing task, can be canceled
+	ctx context.Context
```
hexfusion, line 20: I think I noted in previous review ctx stored in the struct is a general antipattern. If you want to provide shutdown you can still attach the cancelFn but can we please pass context not store it.

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,297 @@
...
+// It includes provisioner and storage providers for certificate lifecycle management,
+// retry tracking, and detailed certificate information for renewal decisions.
+type certificate struct {
+	mu            sync.RWMutex                 `json:"-"`              // Mutex for thread-safe access to certificate fields
+	Provisioner   provider.ProvisionerProvider `json:"-"`              // Provisioner instance for certificate generation (not persisted)
+	Storage       provider.StorageProvider     `json:"-"`              // Storage instance for certificate persistence (not persisted)
+	Name          string                       `json:"name"`           // Certificate name/identifier
+	Config        provider.CertificateConfig   `json:"config"`         // Certificate configuration including provisioner and storage settings
+	Info          CertificateInfo              `json:"info,omitempty"` // Certificate metadata and validity information
+	RetryFailures int                          `json:"retry_failures"` // Number of consecutive provisioning failures
```
hexfusion: question: what happens after these retry failures?

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// certificateProvider groups certificates from a single configuration provider
+// with tracking of last synchronization time for cleanup purposes.
+type certificateProvider struct {
+	Certificates map[string]*certificate `json:"certificates"`   // Map of certificate name to certificate object
+	LastSyncedAt time.Time               `json:"last_synced_at"` // Last time this provider was synchronized
+}
+
+// certStorage manages in-memory certificate state with optional persistent backing.
```
hexfusion: if this can run in memory why do we need to persist it. My expectation would be that in memory state should be rebuildable from cert metadata and the config. If this true why do we need this JSON serialization?

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,297 @@
...
+
+// CertificateInfo contains parsed certificate metadata used for renewal decisions
+// and monitoring certificate lifecycle status.
+type CertificateInfo struct {
+	LastProvisioned *time.Time `json:"last_provisioned,omitempty"` // Timestamp of last successful provisioning
+	NotBefore       *time.Time `json:"not_before,omitempty"`       // Certificate validity start time
+	NotAfter        *time.Time `json:"not_after,omitempty"`        // Certificate validity end time (expiration)
+	CommonName      *string    `json:"common_name,omitempty"`      // Subject common name from certificate
+	SerialNumber    *string    `json:"serial_number,omitempty"`    // Certificate serial number
+	RenewalCount    int        `json:"renewal_count"`              // Number of times certificate has been renewed
```
hexfusion: why is this important information? the other data should be parseable from the certs directly. the path to the certs should be in the config.

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// certificateProvider groups certificates from a single configuration provider
+// with tracking of last synchronization time for cleanup purposes.
+type certificateProvider struct {
+	Certificates map[string]*certificate `json:"certificates"`   // Map of certificate name to certificate object
+	LastSyncedAt time.Time               `json:"last_synced_at"` // Last time this provider was synchronized
```
hexfusion: during runtime this is in memory when we restart the process we should sync by default. do we really need to store this to disk?

internal/agent/device/certmanager/certificate.go
hexfusion: My general point is there is a lot of complexity that we can and probably should shed before we merge. If we don't need it now we should remove it.

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,714 @@
...
+				cm.log.Errorf("certificate management sync failed: %v", err)
+			}
+		}
+	}
+}
+
+// SyncNow is intended to be registered with the reload manager.
+// It does not perform a synchronous sync; instead it enqueues an
+// asynchronous sync request that the manager's Run loop will handle.
+func (cm *CertManager) SyncNow(_ context.Context, _ *config.Config) error {
```
hexfusion: Should the Reload manager learn about the cert-manager config and treat it in the same way? Goal being to handle config reload the same across the application. This can be a followup, but we want to try to keep a single way of doing things from a maintenance perspective. The goal of the reload manager is to reload config. Since we are growing a config we might need to extend support.

ref. https://github.com/flightctl/flightctl/blob/main/internal/agent/reload/reload.go#L63

internal/agent/device/certmanager/manager.go
hexfusion: see above I would like tighter coupling so the reload manger actually reloads configs. The config pattern is established in agent/config. I would expect cert-manager to adopt vs create a new pattern.

internal/agent/device/certmanager/provider/config/dropin.go
```diff
@@ -0,0 +1,135 @@
...
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/flightctl/flightctl/internal/agent/device/certmanager/provider"
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	"sigs.k8s.io/yaml"
+)
+
+// DropInConfigProvider reads a base certificate configuration file and merges
```
hexfusion, line 16: why do we need a provider to handle dropin config? the dropin config pattern is established please adopt. if there is more than one way to handle configs I would ask why.

ref. https://github.com/flightctl/flightctl/blob/024f0c1ec9a9be30af6b1b2faa87fe614a3d56fb/internal/agent/reload/reload.go#L63

This should allow you to remove config provider altogether?

internal/agent/device/certmanager/provider/provisioner/csr.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// CSRProvisioner handles certificate provisioning through Certificate Signing Requests.
+// It generates a private key and CSR, submits it to the management server, and polls
+// for approval and certificate issuance. This supports the standard Kubernetes CSR workflow.
+type CSRProvisioner struct {
+	// Name of the device requesting the certificate
+	deviceName string
+	// Client for communicating with management server
+	managementClient csrClient
```
hexfusion:
```suggestion
	csrClient csrClient
```

internal/agent/device/certmanager/provider/state/filestorage.go
hexfusion: the internal state of the cert manager isnt this just the config plus the metadata from the certs on disk? what other state is required? why would we need a pluggable backend for this data which can  be recreated on process start?

internal/agent/device/certmanager/provider/storage/fs.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/flightctl/flightctl/internal/agent/device/certmanager/provider"
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	fccrypto "github.com/flightctl/flightctl/pkg/crypto"
+	oscrypto "github.com/openshift/library-go/pkg/crypto"
+)
+
+// FileSystemStorageConfig defines configuration for filesystem-based certificate storage.
+// It specifies where certificates and private keys should be stored on the filesystem
+// and what permissions should be applied to the files.
+type FileSystemStorageConfig struct {
```
hexfusion, line 19: can you provide a usecase for non-file storage?

internal/agent/device/certmanager/retryqueue.go
hexfusion: I wonder if we need a ring buffer for storing a handful of items. Anyways thanks for looking.

internal/agent/device/certmanager/provider/storage/fs.go
asafbss: Tpm

internal/agent/device/certmanager/provider/config/dropin.go
asafbss: Static for tests? Agnostic

internal/agent/device/certmanager/certificate.go
asafbss: Renewal count. Last sync. Error count.

internal/agent/device/certmanager/certificate.go
asafbss: We log the error in the state file

internal/agent/device/bootstrap.go
asafbss: It's just for observability at the moment. We can get information about the certificate like when it's going to be expired. Having said that, i can remove it

internal/agent/device/certmanager/provider/state/filestorage.go
asafbss: Last sync, renewal count, etc...

internal/agent/device/certmanager/cert_processing_queue.go
asafbss: But the process function where we provision the certificate and store it checks for the context cancellation. 
It needs  the context to do so

internal/agent/device/certmanager/certificate.go
asafbss: When was the last time we were able to sync this provider

internal/agent/device/certmanager/cert_processing_queue.go
avishayt: It’s idiomatic because context is built for hierarchical cancellation: you create a per-task child with context.WithCancel(parent), so the parent can stop everything while each child can be cancelled independently. You pass that child context down the worker’s call chain (without persisting external contexts in structs), keeping only the cancel handle, which avoids stale or unrelated deadlines leaking across tasks.

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,551 @@
...
+
+		certs, err := cm.certificates.ReadCertificates(providerName)
+		if err != nil {
+			return err
+		}
+		for _, c := range certs {
+			handledCertificates = append(handledCertificates, c.Name)
+		}
+		return fmt.Errorf("failed to load certificate configs from provider %q: %w", providerName, err)
+	}
```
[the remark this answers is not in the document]
asafbss: we run:  _, err := cm.certificates.EnsureProvider(providerName); afterward...what's wrong with that.
please recheck

internal/agent/device/certmanager/builtins.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+)
+
+// WithBuiltins registers the standard certificate manager providers and factories.
+func WithBuiltins(
+	deviceName string,
+	managementClient client.Management,
+	readWriter fileio.ReadWriter,
+	cfg *agent_config.Config,
+) ManagerOption {
```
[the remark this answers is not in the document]
asafbss: > This means agent restarts will lose all certificate state.

Yes, this was the latest requirement

internal/agent/device/certmanager/manager.go
```diff
@@ -0,0 +1,554 @@
...
+		return true
+	}
+
+	if !cert.Config.Provisioner.Equal(cfg.Provisioner) || !cert.Config.Storage.Equal(cfg.Storage) {
+		cm.log.Debugf("Certificate %q for provider %q: provisioner or storage changed - needs provisioning", cert.Name, providerName)
+		return true
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
asafbss: We decided not to include renewal at this stage

internal/agent/device/certmanager/provider/provisioner/csr.go
asafbss: done

internal/agent/device/certmanager/certificate.go
asafbss: I’ve removed the persistent storage for certificates, and I agree the in-memory state can be rebuilt from cert metadata + the current config.

Keep in mind that a gap this leaves, for example, is cleanup when configs change storage/provisioner settings. Example: a user switches the provisioner or points storage to a new location, then fully restarts the agent. After restart, we can rebuild the state for the new location, but we no longer have a reference to the old location, so we can’t safely purge the previous certificate.

internal/agent/device/certmanager/certificate.go
asafbss: I have removed the RetryFailures field

internal/agent/device/bootstrap.go
asafbss: Removed

internal/agent/device/certmanager/certificate.go
asafbss: I have removed the unnecessary fields

internal/agent/device/certmanager/certificate.go
asafbss: I’ve removed the persistent storage for certificates, and I agree the in-memory state can be rebuilt from cert metadata + the current config.

The gap this leaves is cleanup when configs change storage/provisioner settings. Example: a user switches the provisioner or points storage to a new location, then fully restarts the agent. After restart we can rebuild state for the new location, but we no longer have a reference to the old location, so we can’t safely purge/migrate the previous certificate artifacts.

internal/agent/device/certmanager/manager.go
asafbss: I agree. Let’s follow up on this. The cert-manager’s config differs from the agent’s, so for now let’s keep it simple and proceed as-is. Once we decide to extend the reload manager’s support to cover cert-manager, we can align on a single, consistent reload path across the application.

internal/agent/device/certmanager/provider/config/dropin.go
asafbss: https://github.com/flightctl/flightctl/pull/1382#discussion_r2303696681

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,188 @@
...
+
+	provider, exists := s.providers[providerName]
+	if !exists {
+		provider = &certificateProvider{
+			Certificates: make(map[string]*certificate),
+		}
+		s.providers[providerName] = provider
+	}
+	return provider, nil
+}
```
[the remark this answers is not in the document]
asafbss: fixed

internal/agent/device/certmanager/certificate.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	provider, ok := s.providers[providerName]
+	if !ok {
+		return fmt.Errorf("provider %q not found", providerName)
+	}
+
+	provider.Certificates[cert.Name] = cert
+	return nil
+}
```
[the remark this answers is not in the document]
asafbss: fixed

internal/agent/device/certmanager/cert_processing_queue.go
```diff
@@ -0,0 +1,182 @@
...
+		cancel:   cancel,
+		provider: providerName,
+		cert:     cert,
+		cfg:      cfg,
+	}
+
+	q.inProcess[generateName(providerName, cert.Name)] = c
+	q.queue.Add(c)
+	return nil
+}
```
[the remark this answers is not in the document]
asafbss: We don't want the forwarded context to be part of the certificate provisioning, which might take some time (async). Imagine we have a context with a timeout; the context will be cancelled, and the certificate will still be in provisioning status.

avishayt (approved):
Looks like everything is addressed, thanks Assaf!  We can do more improvements in the future if necessary but I feel we're in a good place with the current version.

internal/agent/device/certmanager/provider/storage/fs.go
hexfusion: We can take a deeper look at this when i get back, I think the existing identity provider already abstrats this so the cert manager should only need to reason about files.

PR #1383, opened 2025-07-15
EDM-1205: agent/network fix minor route selection issue

kkyrazis:
Assisted by Cursor:

Fixes minor route default route selection issue where if the loopback interface is defined before a desired interface, a real default interface could potentially not be selected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced default network route detection by filtering out loopback interfaces earlier in the process with improved logging.

* **Tests**
  * Added extensive unit tests covering IPv4 and IPv6 default route parsing, loopback interface exclusion, and various error conditions.

### DIFF
--- a/internal/agent/device/systeminfo/network.go
+++ b/internal/agent/device/systeminfo/network.go
@@ -180,7 +180,7 @@ func waitForDefaultRoute(ctx context.Context, log *log.PrefixLogger, reader file
 	var route *DefaultRoute
 
 	err := wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(ctx context.Context) (bool, error) {
-		r, err := getDefaultRoute(reader)
+		r, err := getDefaultRoute(log, reader)
 		if err != nil {
 			log.Infof("Waiting for default route...")
 			return false, nil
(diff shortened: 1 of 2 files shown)


internal/agent/device/systeminfo/network.go
```diff
@@ -268,14 +268,22 @@ func getDefaultRouteIPv4(reader fileio.Reader) (*DefaultRoute, error) {
 
 		// is the default route (destination 0.0.0.0)
 		if fields[1] == "00000000" {
+			interfaceName := fields[0]
+			gatewayHex := fields[2]
+
+			// skip loopback interfaces
+			if isLoopback(interfaceName) {
+				continue
```
hexfusion, line 277: question: should we add some trace logging here, not sure exactly how far away a logger is.

internal/agent/device/systeminfo/network.go
```diff
@@ -307,14 +315,22 @@ func getDefaultRouteIPv6(reader fileio.Reader) (*DefaultRoute, error) {
 		// check if this is the default route (destination ::/0, which is 00000000000000000000000000000000 with prefix 00)
 		// https://mirrors.deepspace6.net/Linux+IPv6-HOWTO/proc-net.html
 		if fields[0] == "00000000000000000000000000000000" && fields[1] == "00" {
+			interfaceName := fields[9]
```
hexfusion, line 319: just making sure we are len checking before somewhere.

internal/agent/device/systeminfo/network.go
kkyrazis: It's only one level up, I'll pass it down and add logging.

internal/agent/device/systeminfo/network.go
kkyrazis: Len checking the size of the name? We aren't doing that. Len checking to ensure we have at least 10 fields? That happens a couple of lines up.

hexfusion (approved):

hexfusion (approved):

PR #1384, opened 2025-07-15
EDM-433: Test K8s Secret Config

keitwb:
This adds an e2e test to confirm that the worker will render device config from a k8s secret source.

## Summary by CodeRabbit

* **New Features**
  * Added support for using Kubernetes Secrets as a configuration source for devices.
  * Introduced a new test secret and permissions to facilitate secret-based configuration in end-to-end tests.
  * Implemented new cluster roles and role bindings to enable secure secret access for the worker component.

* **Tests**
  * Added an end-to-end test to verify device configuration using a Kubernetes Secret and confirm correct file mounting.

### DIFF
--- a/docs/user/managing-devices.md
+++ b/docs/user/managing-devices.md
@@ -607,11 +607,15 @@ spec:
 Volume images must follow the OCI artifact specification:
 
 * Published as OCI images (media type: `application/vnd.oci.image.manifest.v1+json`).
-* Contain one or more tar layers representing the volume contents.
+* Contain one or more layers representing the volume file contents.
 * Hosted on any OCI-compatible registry accessible by the device.
 
 > [!TIP]
-> Typically, volume artifacts contain a single tar layer with the full extracted content.
+> If the artifact contains more than one layer, the mount path should be an already existing
+> directory, into which the layers will each be copied as a separate file using the name specified
+> in the layer's `org.opencontainers.image.title` field. For single layer archives if the mount path
+> does not exist as a directory the single layer will be extracted as a file at that path, otherwise
+> it will be placed into the existing directory using the file name in the name field for the layer.
 
 #### Device Requirements
 
--- a/internal/agent/client/podman.go
+++ b/internal/agent/client/podman.go
@@ -174,6 +174,12 @@ func (p *Podman) pullArtifact(ctx context.Context, artifact string, options *cli
 	return strings.TrimSpace(stdout), nil
 }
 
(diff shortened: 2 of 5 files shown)


deploy/helm/e2e-extras/templates/flightctl-worker-clusterrole.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    flightctl.service: flightctl-worker
+  name: flightctl-worker
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
```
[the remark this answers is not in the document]
asafbennatan: this is probably correct ?

test/e2e/agent/agent_test.go
```diff
@@ -328,6 +328,45 @@ var _ = Describe("VM Agent behavior", func() {
...
+				device.Spec.Config = &[]v1alpha1.ConfigProviderSpec{configProviderSpec}
+				logrus.Infof("Updating %s with config %s", deviceId, device.Spec.Config)
+			})
+			Expect(err).ToNot(HaveOccurred())
+
+			logrus.Infof("Waiting for the device to pick the config")
+			harness.WaitForDeviceContents(deviceId, fmt.Sprintf("the device is updated to renderedVersion: %s", strconv.Itoa(newRenderedVersion)),
+				func(device *v1alpha1.Device) bool {
+					for _, condition := range device.Status.Conditions {
+						if condition.Type == "Updating" && condition.Reason == "Updated" && condition.Status == "False" &&
```
asafbennatan: while i notice this pattern all over the e2e - i think we should be using the proper enums here :

`
if condition.Type == v1alpha1.DeviceUpdating && condition.Reason == string(v1alpha1.UpdateStateUpdated) && condition.Status == v1alpha1.ConditionStatusFalse &&.....`

additionally i find it weird that we are checking condition.Reason if its a free text field , if its not it should be backed by its own enum IMO
@eldar101 thoughts?

eldar101: As long as it's just internal logic and not some input from a json that might change, then sure, go ahead.

kkyrazis: There's a helper function that was created at some point. `e2e.ConditionExists` that uses the enums (minus the reason).

kkyrazis: There's also just `harness.WaitForDeviceNewRenderedVersion` which probably does some similar check to what you have here.

keitwb: Ok, yeah those are definitely better, some of the existing tests can definitely be refactored to use it.

kkyrazis (approved):

PR #1392, opened 2025-07-17
EDM-1847: Add TPM LDevID persistence

kkyrazis:
Adds the ability to persist an LDevID created via a TPM so that it'll survive power cycles.
1. **Default** `key-blob`: persists the key blobs that are generated by the TPM to disk. The public portion of the blob is well public and safe to store on disk. The private portion is encrypted by the TPM such that it is only decryptable and usable by the original TPM itself. This is also safe to store.

This PR also adds TPM device discovery so that we don't have to rely on hardcoded values.

## Summary by CodeRabbit

* **New Features**
  * Added experimental TPM device simulation support with automatic TPM device detection and configuration per simulated agent.
  * Enabled TPM device and sysfs symlink creation in simulated environments, controlled via an experimental feature flag.

* **Improvements**
  * Enhanced TPM configuration with detailed settings for device paths and key persistence.
  * Refined TPM client initialization, lifecycle management, and resource cleanup for improved reliability.
  * Improved error handling, logging, and TPM device discovery logic.

* **Bug Fixes**
  * Fixed path existence checks for files and directories, including handling empty files and directories with optional content verification.

* **Tests**
  * Added extensive tests for TPM device discovery, validation, key blob persistence, and TPM client operations.
  * Expanded tests for file existence checks with and without content validation.

* **Chores**
  * Updated dependencies for better stability and compatibility.


lkatalin:
I didn't get into all the details, esp. with the simulator, but these changes generally look good to me.

One thing we should make sure to do is enforce use of the TPM resource manager. So, `/dev/tpmrm0` should be a valid device path, but `/dev/tpm0` should not (or it should translate to use of the resource manager if a path without is specified). Otherwise we will get errors.

Saving the LDevID and other child keys as a blob seems fine IMO.

kkyrazis:
> I didn't get into all the details, esp. with the simulator, but these changes generally look good to me.
> 
> One thing we should make sure to do is enforce use of the TPM resource manager. So, `/dev/tpmrm0` should be a valid device path, but `/dev/tpm0` should not (or it should translate to use of the resource manager if a path without is specified). Otherwise we will get errors.
> 
> Saving the LDevID and other child keys as a blob seems fine IMO.

The way I have it implemented is that it'll allow a user specified path to match either the resource manager or the device path, but the act of `Open`ing a device only uses the resource manager path.

```go
for _, device := range devices {
		if device.DevicePath == devicePath || device.ResourceMgrPath == devicePath {
			if err := device.ValidateVersion2(); err != nil {
				return nil, fmt.Errorf("invalid TPM device %q: %w", devicePath, err)
			}
			return &device, nil
		}
	}
```

```go
func (d *Device) Open() (*TPM, error) {
	if d.tpm != nil {
		return d.tpm, nil
	}

	tpm, err := OpenTPM(d.rw, d.ResourceMgrPath)
	if err != nil {
		return nil, err
	}
	d.tpm = tpm
	return tpm, nil
}
```

internal/tpm/discovery.go
```diff
@@ -0,0 +1,135 @@
...
+
+const (
+	devicePathTemplate   = "/dev/tpm%s"
+	deviceRMPathTemplate = "/dev/tpmrm%s"
+	versionPathTemplate  = "/sys/class/tpm/%s/tpm_version_major"
+	sysClassPath         = "/sys/class/tpm"
+	sysFsPathTemplate    = "/sys/class/tpm/%s"
+)
+
+var tpmDeviceRegex = regexp.MustCompile(`^tpm(\d+)$`)
```
hexfusion: note: is regex required? almost like reflect if you see regexp make sure its needed because performance is terrible.

internal/agent/device/fileio/reader.go
```diff
@@ -50,48 +50,75 @@ func (r *reader) ReadDir(dirPath string) ([]fs.DirEntry, error) {
...
-func checkPathExists(path string) (bool, error) {
+func checkPathExists(path string, options *pathExistsOptions) (bool, error) {
 	info, err := os.Stat(path)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return false, nil
 		}
 		return false, fmt.Errorf("error checking path: %w", err)
 	}
+	pathType := "file"
```
hexfusion: lets make this an enum, my guess is we have other uses.

internal/agent/device/fileio/reader.go
```diff
@@ -50,48 +50,75 @@ func (r *reader) ReadDir(dirPath string) ([]fs.DirEntry, error) {
...
-		if err != nil {
-			return false, fmt.Errorf("directory exists but %w: %w", errors.ErrReadingPath, err)
-		}
-		defer dir.Close()
 		// read a single entry from the directory to confirm readability
-		_, err = dir.Readdirnames(1)
+		_, err = file.Readdirnames(1)
 		if err != nil {
 			if errors.Is(err, io.EOF) {
 				// readable but empty
```
hexfusion: so we were already doing skipContentCheck or just on dir

internal/agent/device/fileio/reader.go
```diff
@@ -50,48 +50,75 @@ func (r *reader) ReadDir(dirPath string) ([]fs.DirEntry, error) {
...
-		defer file.Close()
 		// read a single byte from the file to ensure permissions are correct
 		buffer := make([]byte, 1)
-		if _, err := file.Read(buffer); err != nil {
-			return false, fmt.Errorf("file exists but %w: %w", errors.ErrReadingPath, err)
-		}
+		_, err = file.Read(buffer)
+	}
+
+	if err != nil {
```
hexfusion, line 128: this feels awkward why is err handled outside of else?

internal/agent/device/lifecycle/tpm.go
```diff
@@ -5,15 +5,30 @@ import (
...
+	PersistenceType     string
+	PersistenceMetadata string
+	DevicePath          string
+	DataDir             string // Used for default metadata path construction
+}
+
 type TpmClient struct {
 	// path where the tpm will be found, if it exists
 	tpmSysPath string
+	rw         fileio.ReadWriter
```
hexfusion, line 28: :+1:

internal/tpm/discovery.go
```diff
@@ -0,0 +1,135 @@
...
+			rw:              rw,
+		}
+
+		devices = append(devices, device)
+	}
+
+	return devices, nil
+}
+
+func ResolveDevice(rw fileio.ReadWriter, devicePath string) (*Device, error) {
```
hexfusion: add docs to public function

internal/tpm/discovery.go
```diff
@@ -0,0 +1,135 @@
...
+
+	var devices []Device
+	for _, entry := range entries {
+		matches := tpmDeviceRegex.FindStringSubmatch(entry.Name())
+		if len(matches) != 2 {
+			continue
+		}
+		deviceNum := matches[1]
+
+		device := Device{
```
hexfusion: sort of feels like we are overloading the Device name might not be part of this PR but something to think about.  I also see a fair amount of ecoing.  for example I would expect device.Number vs device.DeviceNumber.

internal/tpm/tpm.go
```diff
@@ -3,74 +3,201 @@ package tpm
...
+// Capabilities represents TPM capabilities and resource information.
+type Capabilities struct {
+	// PersistentHandleCount is the total number of persistent handles supported by the TPM.
+	PersistentHandleCount uint32
+	// PersistentHandleAvailCount is the number of available persistent handles.
+	PersistentHandleAvailCount uint32
+}
+
+// Device represents a TPM device and its associated file paths.
+type Device struct {
```
hexfusion: Is this just a `TPM` and the other `TPM` feels like a tpm.Client can we make sure these object names are appropriate.

internal/tpm/tpm.go
```diff
@@ -3,74 +3,201 @@ package tpm
...
+func WithTransientKey() EnsureLDevIDOption {
+	return func(opts *ensureLDevIDOptions) {
+		opts.ensureStrategy = &transientStrategy{}
 	}
-	return nil
 }
 
-func OpenTPM(devicePath string) (*TPM, error) {
-	ch, err := tpmutil.OpenTPM(devicePath)
+// TPM represents a connection to a TPM device and manages TPM operations.
```
hexfusion: so like a conn?

hexfusion (commented):
looking good few small notes

internal/tpm/discovery.go
kkyrazis: Fair point. I'll spend some time coming up with a better name

internal/tpm/discovery.go
kkyrazis: I just liked it better than "if starts with tpm and doesn't start with tpmrm then trim tpm to get the number"

internal/agent/device/fileio/reader.go
kkyrazis: in the dir case is was checking to see if an entry exists in the directory. In the file case it was checking to see if there are contents. SkipContentCheck validates neither of those.

internal/agent/device/fileio/reader.go
kkyrazis: The error logging was very repetitive, but I'll come up with a better approach for this.

internal/agent/agent.go
```diff
@@ -75,6 +75,16 @@ func (a *Agent) Run(ctx context.Context) error {
 	// create file io writer and reader
 	deviceReadWriter := fileio.NewReadWriter(fileio.WithTestRootDir(a.config.GetTestRootDir()))
 
+	tpmClient := a.tryLoadTPM(deviceReadWriter)
+
+	// update nonce if TPM client was successfully created
+	if tpmClient != nil {
```
hexfusion, line 81: we still need a way for this to be opt-in I am thinking like `DisableTPMAuth` or similar, default to true for now.

internal/agent/agent.go
```diff
@@ -354,3 +355,24 @@ func newGrpcClient(cfg *baseconfig.ManagementService) (grpc_v1.RouterServiceClie
 	}
 	return client, nil
 }
+
+func (a *Agent) tryLoadTPM(writer fileio.ReadWriter) *lifecycle.TpmClient {
```
hexfusion, line 359: lets make it easier to reason about
```
func (a *Agent) tryLoadTPM(writer fileio.ReadWriter) (*lifecycle.TpmClient, bool) {
...

if isTPMEnabled {
```

internal/agent/config/config.go
```diff
@@ -56,6 +56,10 @@ const (
 	EnrollmentCertFile = "client-enrollment.crt"
 	// name of the enrollment key file
 	EnrollmentKeyFile = "client-enrollment.key"
+	// DefaultTPMDevicePath is the default TPM device path
+	DefaultTPMDevicePath = "/dev/tpm0"
+	// DefaultTPMKeyBlobFile is the default filename for TPM key blob persistence
+	DefaultTPMKeyBlobFile = "agent-ldevid-blob.yaml"
```
hexfusion: we should know this is agent material
```suggestion
	DefaultTPMKeyBlobFile = "tpm-ldevid.yaml"
```

internal/agent/config/config.go
```diff
@@ -73,8 +77,8 @@ type Config struct {
 	// StatusUpdateInterval is the interval between two status updates
 	StatusUpdateInterval util.Duration `json:"status-update-interval,omitempty"`
 
-	// TPMPath is the path to the TPM device
-	TPMPath string `json:"tpm-path,omitempty"`
+	// TPMConfig holds all TPM-related configuration
+	TPMConfig
```
hexfusion: we are in the config already.
```suggestion
	TPM
```

agentConfig.TPM.Something

internal/agent/config/config.go
```diff
@@ -336,6 +340,9 @@ func mergeConfigs(base, override *Config) {
 	overrideSliceIfNotNil(&base.SystemInfoCustom, override.SystemInfoCustom)
 	overrideIfNotEmpty(&base.SystemInfoTimeout, override.SystemInfoTimeout)
 
+	// TPM config
+	base.TPMConfig.MergeWith(&override.TPMConfig)
```
hexfusion: nit: update the user-docs to reflect we have reload support for TPM config.

internal/agent/config/tpm_config.go
```diff
@@ -0,0 +1,15 @@
+package config
```
hexfusion: `internal/agent/config/tpm.go`

internal/agent/device/fileio/reader.go
```diff
@@ -50,49 +55,78 @@ func (r *reader) ReadDir(dirPath string) ([]fs.DirEntry, error) {
...
+// PathExistsOption represents options for PathExists function
+type PathExistsOption func(*pathExistsOptions)
+
+type pathExistsOptions struct {
+	skipContentCheck bool
+}
+
+// WithSkipContentCheck configures PathExists to skip content verification
+// and only check if the path can be opened
+func WithSkipContentCheck() PathExistsOption {
```
hexfusion, line 67: :+1:

internal/agent/device/lifecycle/tpm.go
```diff
@@ -5,42 +5,96 @@ import (
...
+	"path/filepath"
 
+	agent_config "github.com/flightctl/flightctl/internal/agent/config"
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
 	"github.com/flightctl/flightctl/internal/tpm"
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/google/go-tpm-tools/client"
 )
 
+type TPMClientConfig struct {
```
hexfusion, line 17: we are in the tpm package no need to repeat always try to think like the caller.

`tpm.ClientConfig`

internal/agent/device/lifecycle/tpm.go
```diff
@@ -5,42 +5,96 @@ import (
...
+	Log             *log.PrefixLogger
```
hexfusion, line 18: its good to be defensive on exporting fields. I default to never it will reduce the scope of misusing the object in the future.

internal/agent/device/lifecycle/tpm.go
```diff
@@ -5,42 +5,96 @@ import (
...
+	DeviceWriter    fileio.ReadWriter
+	PersistencePath string
+	DevicePath      string
+	DataDir         string // Used for default persistence path construction
```
hexfusion, line 22: if you have this object should the caller be able to mutate these? if not lets not export

internal/agent/device/lifecycle/tpm.go
```diff
@@ -5,42 +5,96 @@ import (
...
 	// meant to be updated for each attestation
 	currNonce []byte
 	log       *log.PrefixLogger
 }
 
-func NewTpmClient(log *log.PrefixLogger) (*TpmClient, error) {
-	tc := TpmClient{
-		tpmSysPath: tpm.TpmSystemPath,
-		log:        log,
+func NewTPMClient(config TPMClientConfig) (*TpmClient, error) {
```
hexfusion, line 38: we already know and love tpm :)
```suggestion
func NewClient(config TPMClientConfig) (*Client, error) {
```

internal/tpm/discovery.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		devices = append(devices, device)
+	}
+
+	return devices, nil
+}
+
+// ResolveDevice returns the device specified by the device path if it exists and if the specified device is version 2
+func ResolveDevice(rw fileio.ReadWriter, devicePath string) (*TPM, error) {
```
hexfusion: No more device right
`s/device/tpm/g`

internal/tpm/discovery.go
```diff
@@ -0,0 +1,136 @@
...
+	if !d.Exists() {
+		return fmt.Errorf("no TPM detected at %s", d.ResourceMgrPath)
+	}
+	versionBytes, err := d.rw.ReadFile(d.VersionPath)
+	if err != nil {
+		return fmt.Errorf("reading tpm version file: %w", err)
+	}
+	versionStr := string(bytes.TrimSpace(versionBytes))
+	if versionStr != "2" {
+		return fmt.Errorf("TPM is not version 2.0")
```
hexfusion: lets include what version we found.

internal/tpm/tpm.go
```diff
@@ -1,76 +1,118 @@
...
 	"github.com/google/go-tpm/tpmutil"
+	"sigs.k8s.io/yaml"
 )
 
 const (
-	MinNonceLength     = 8
-	TpmSystemPath      = "/dev/tpmrm0"
-	TpmVersionInfoPath = "/sys/class/tpm/tpm0/tpm_version_major"
+	// MinNonceLength is the minimum required length for nonces used in TPM operations.
+	MinNonceLength = 8
```
hexfusion, line 29: is this a spec can we get a ref?

internal/tpm/tpm.go
```diff
@@ -1,76 +1,118 @@
...
 )
 
+// TPM represents a TPM device and its associated file paths.
 type TPM struct {
-	devicePath string
-	channel    io.ReadWriteCloser
-	srk        *tpm2.NamedHandle
-	ldevid     *tpm2.NamedHandle
+	// DeviceNumber is the numeric identifier of the TPM device (e.g., "0" for /dev/tpm0).
+	DeviceNumber string
```
hexfusion: wat is device?? tpm.Number

internal/tpm/tpm.go
```diff
@@ -159,7 +203,30 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+func (t *Client) loadLDevIDFromBlob(public tpm2.TPM2BPublic, private tpm2.TPM2BPrivate) (*tpm2.NamedHandle, error) {
+	loadCmd := tpm2.Load{
+		ParentHandle: t.srk,
+		InPrivate:    private,
+		InPublic:     public,
+	}
+
+	loadRsp, err := loadCmd.Execute(transport.FromReadWriter(t.channel))
+	if err != nil {
+		return nil, fmt.Errorf("error loading ldevid key: %w", err)
```
hexfusion: we know its an error ;)
```suggestion
		return nil, fmt.Errorf("loading ldevid key: %w", err)
```

internal/tpm/tpm.go
```diff
@@ -169,7 +236,40 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
 	return t.ldevid, nil
 }
 
-func (t *TPM) GetQuote(nonce []byte, ak *client.Key, pcr_selection *legacy.PCRSelection) (*pbtpm.Quote, error) {
+func (t *Client) flushContext() error {
+	if t.ldevid == nil {
+		return nil
```
hexfusion: this would be unexpected right or no can we at least log this

internal/tpm/tpm.go
```diff
@@ -183,3 +283,80 @@ func (t *TPM) GetQuote(nonce []byte, ak *client.Key, pcr_selection *legacy.PCRSe
...
+	if err != nil {
+		return fmt.Errorf("writing blob to file %s: %v", path, err)
+	}
+
+	return nil
+}
+
+func (t *Client) loadLDevIDBlob(path string) (tpm2.TPM2BPublic, tpm2.TPM2BPrivate, error) {
+	var public tpm2.TPM2BPublic
+	var private tpm2.TPM2BPrivate
```
hexfusion, line 470: are these interaces regardless if there is an error we should just return nil as the opjects should never be consumed

internal/tpm/tpm.go
```diff
@@ -183,3 +283,80 @@ func (t *TPM) GetQuote(nonce []byte, ak *client.Key, pcr_selection *legacy.PCRSe
...
+
+	data, err := t.rw.ReadFile(path)
+	if err != nil {
+		return public, private, err
```
hexfusion:
```suggestion
		return nil, nil, err
```

internal/agent/device/lifecycle/tpm.go
kkyrazis: This is the lifecycle package which leads me to a question I've had. Why is this in the lifecycle package?

internal/agent/device/lifecycle/tpm.go
kkyrazis: This is lifecycle though

internal/agent/device/lifecycle/tpm.go
kkyrazis: The intention was to use this in place of N number of arguments to `NewTPMClient`, but I can just remove this and pass the args directly.

internal/tpm/tpm.go
```diff
@@ -159,7 +203,30 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+		Handle: loadRsp.ObjectHandle,
+		Name:   loadRsp.Name,
+	}
+
+	return t.ldevid, nil
+}
+
+// loadLDevIDFromBlob will load a LDevID for the existing SRK from key blob parts
+// According to https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf
+// it is safe to persist these blobs are they can only be decrypted by the originating TPM
```
hexfusion: revisit these docs

internal/tpm/tpm.go
kkyrazis: They aren't interfaces. I've changed this signature to return pointers rather than the zero value for these structs.

internal/tpm/tpm.go
kkyrazis: I couldn't find anything definitive. I'll keep doing research to see if there's something available.

hexfusion (approved):

PR #1396, opened 2025-07-20
NO-ISSUE: Upgrade Redis image tag to 9.6 in values.acm.yaml

amir-yogev-gh:
## Summary by CodeRabbit

* **Chores**
  * Updated PostgreSQL and Redis image versions in deployment configuration to newer specific build tags.

### DIFF
--- a/deploy/helm/flightctl/values.acm.yaml
+++ b/deploy/helm/flightctl/values.acm.yaml
@@ -9,7 +9,7 @@ db:
 kv:
   image:
     image: registry.redhat.io/rhel9/redis-7
-    tag: 9.5-1736454843
+    tag: 9.6
 
 alertmanager:
   image:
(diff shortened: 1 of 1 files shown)


deploy/helm/flightctl/values.acm.yaml
```diff
@@ -9,7 +9,7 @@ db:
 kv:
   image:
     image: registry.redhat.io/rhel9/redis-7
-    tag: 9.5-1736454843
+    tag: 9.6
```
avishayt: I suggest pinning it to an exact version before each release, so 9.6-1752567986

amir-yogev-gh: Any newer 9.6 version should be backward compatible. Why should we insist on an exact version?

amir-yogev-gh: The same comment could be applied on Postgres version. It's also not using an exact version.

avishayt: Using an exact version allows us to test the exact configuration.  That way if there buggy image released after we release then it won't break us.

amir-yogev-gh: Done

PR #1401, opened 2025-07-21
EDM-1058: Implement Signer for TPM LDevID

lkatalin:
This adds the `Signer` functionality we need to enable the TPM-based signing of CSRs for the agent enrollment.


internal/tpm/tpm.go
```diff
@@ -169,6 +174,88 @@
...
+	}
+	pubkey := &ecdsa.PublicKey{
+		Curve: curve,
+		X:     big.NewInt(0).SetBytes(unique.X.Buffer),
+		Y:     big.NewInt(0).SetBytes(unique.Y.Buffer),
+	}
+	var cryptopubkey crypto.PublicKey = pubkey
+	t.ldevidPub = cryptopubkey
+	return &cryptopubkey, nil
+}
```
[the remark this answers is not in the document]
lkatalin: Actually we have to leave `var cryptopubkey crypto.PublicKey = pubkey` for type conversion.

internal/tpm/tpm.go
```diff
@@ -27,6 +31,7 @@ type TPM struct {
 	channel    io.ReadWriteCloser
 	srk        *tpm2.NamedHandle
 	ldevid     *tpm2.NamedHandle
+	ldevidPub  crypto.PublicKey
```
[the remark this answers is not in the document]
lkatalin: This is probably true in several places and will need to be solved in a separate PR.

internal/tpm/tpm.go
```diff
@@ -27,6 +31,7 @@ type TPM struct {
 	channel    io.ReadWriteCloser
```
hexfusion, line 31: nit: its not part of this PR but using the name `channel`  is something we should change in this context is this a conn? Closing a conn for a client seems natural.

internal/tpm/tpm.go
```diff
@@ -169,6 +174,96 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get curve id for ldevid key: %v", err)
+	}
+	unique, err := outpub.Unique.ECC()
+	if err != nil {
+		return nil, fmt.Errorf("could not get unique parameters for ldevid key: %v", err)
+	}
+	pubkey := &ecdsa.PublicKey{
+		Curve: curve,
+		X:     big.NewInt(0).SetBytes(unique.X.Buffer),
```
hexfusion, line 210: nit: as per godocs and other implementations it is more idiomatic to initialize using `new`
```suggestion
 		X:    new(big.Int).SetBytes(unique.X.Buffer),
```

internal/tpm/tpm.go
```diff
@@ -169,6 +174,96 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+		Y:     big.NewInt(0).SetBytes(unique.Y.Buffer),
+	}
+	// converts ecdsa.PublicKey to crypto.PublicKey
+	var cryptopubkey crypto.PublicKey = pubkey
+	t.ldevidPub = cryptopubkey
+	return &cryptopubkey, nil
```
hexfusion, line 216: nit: this conversion should not be necessary because `crypto.PublicKey` interface is implemented by `ecdsa.PublicKey`

```suggestion
t.ldevidPub = pubkey
return pubkey, nil
```

internal/tpm/tpm.go
```diff
@@ -169,6 +174,96 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+}
+
+func (t TPM) Public() crypto.PublicKey {
+	if t.ldevidPub == nil {
+		return nil
```
hexfusion, line 221: question: is this what we want, or should this be an error? I personally do not like returning nil interfaces.

internal/tpm/tpm.go
```diff
@@ -169,6 +174,96 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+		},
+		Digest: tpm2.TPM2BDigest{
+			Buffer: data[:],
+		},
+		Validation: tpm2.TPMTTKHashCheck{
+			Tag: tpm2.TPMSTHashCheck,
+		},
+	}
+
+	transportTPM := transport.FromReadWriter(t.channel)
```
hexfusion, line 244: question: am I right to think this os creating a client connection to the tpm? generally clients are reused but maybe we have a reason for a new one on every sign?

internal/tpm/tpm.go
```diff
@@ -169,6 +174,96 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+	signRsp, err := sign.Execute(transportTPM)
+	if err != nil {
+		return nil, fmt.Errorf("failed to sign digest with ldevid: %v", err)
+	}
+	ecdsaSig, err := signRsp.Signature.Signature.ECDSA()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ECDSA signature from sign response: %v", err)
+	}
+	bigR := big.NewInt(0).SetBytes(ecdsaSig.SignatureR.Buffer)
```
hexfusion, line 253: see above on using `new` as big.Int contstructor.

internal/tpm/tpm.go
kkyrazis: It's just wrapping the already existing device that is opened. That said, it probably could be created a single time.

internal/tpm/tpm.go
kkyrazis: The docs of OpenTPM describe it as a channel, but I can change the name.

internal/tpm/tpm.go
kkyrazis: Kind of annoying, but `Public` is part of the crypto.Signer interface. I can panic, but it's probably sufficient to just always return `t.ldevidPub`

internal/tpm/tpm.go
hexfusion: just look like conn even if its not the terms channel in go assumes chan type

internal/tpm/tpm.go
kkyrazis: I tried to create this a single time, but there is some oddity with some of the functions that use the legacy package vs the newer package. Those would only work with the raw unwrapped io.ReadWriter while everything else worked fine. Eventually that should be addressed, but I decided to leave it alone for now

hexfusion: more of an open question which is fine

hexfusion:
this work is now being tracked here https://github.com/flightctl/flightctl/pull/1409

PR #1402, opened 2025-07-21
EDM-1766: RPM/SELinux build tweaks/docs

keitwb:
The most important thing here is adding the conditional definition of the cgroup_type to the SELinux policy, as it is not defined in either Centos 9 or 10 but is present in the Fedora distros used to build RPMs.

The rest is just docs/script changes to help avoid some of the pitfalls I ran into.


packaging/selinux/flightctl_agent.te
```diff
@@ -29,6 +32,10 @@ gen_require(`
     attribute domain;
 ')
 
+init_daemon_domain(flightctl_agent_t, flightctl_agent_exec_t)
```
keitwb, line 39: These were just moved after the gen_require to make them consistent with the other rules.

hack/build_rpms.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+if [[ -n "$(git status -s)" ]]; then
+  # See https://packit.dev/source-git/work-with-source-git/build-locally
+  echo "WARNING: There are uncommitted git changes. This RPM build will NOT include them."
```
hexfusion, line 5: nice

hexfusion (approved):
great stuff thanks!

PR #1409, opened 2025-07-22
EDM-1749: Implement Signer for TPM LDevID

kkyrazis:
This is Lily's PR from #1401 but since she's out for a few weeks I'll be addressing the PR feedback. 

Lily: This adds the Signer functionality we need to enable the TPM-based signing of CSRs for the agent enrollment.

Kyle: Address PR feedback and added tests.

## Summary by CodeRabbit

* **New Features**
  * Added TPM-based cryptographic signing capabilities, allowing the TPM to sign data and provide the corresponding public key.
  * TPM now supports integration with cryptographic interfaces for seamless signing operations.

* **Bug Fixes**
  * Improved resource management and cleanup when working with TPM connections.
  * Enhanced error messages with detailed context for TPM operations.

* **Tests**
  * Introduced comprehensive tests for TPM key management, signing, public key retrieval, and cryptographic interface compliance to ensure reliability and correctness.


internal/tpm/tpm.go
```diff
@@ -53,31 +62,34 @@ func ValidateTpmVersion2() error {
...
+	return t.cleanup()
 }
 
 func (t *TPM) GetTpmVendorInfo() ([]byte, error) {
 	if t == nil {
 		return nil, fmt.Errorf("cannot get TPM vendor info: nil receiver in TPM struct")
 	}
-	if t.channel == nil {
+	if t.conn == nil {
 		return nil, fmt.Errorf("cannot get TPM vendor info: no channel available in TPM struct")
```
hexfusion:
```suggestion
		return nil, fmt.Errorf("cannot get TPM vendor info: no conn available")
```

internal/tpm/tpm.go
```diff
@@ -169,6 +179,93 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+
+	pub, err := tpm2.ReadPublic{
+		ObjectHandle: t.ldevid.Handle,
+	}.Execute(transport.FromReadWriter(t.conn))
+	if err != nil {
+		return nil, fmt.Errorf("could not read public key: %v", err)
+	}
+	outpub, err := pub.OutPublic.Contents()
+	if err != nil {
+		return nil, fmt.Errorf("could not get contents of TPM2Bpublic: %v", err)
```
hexfusion: lets wrap all errors `%w` more below

internal/tpm/tpm.go
```diff
@@ -53,31 +62,34 @@ func ValidateTpmVersion2() error {
...
 
 func (t *TPM) Close() error {
 	if t == nil {
 		return nil
 	}
-	if t.channel == nil {
-		return nil
-	}
-	return t.channel.Close()
+	return t.cleanup()
```
hexfusion, line 82: question: io.ReadWriteCloser Close should be idempotent what kind of errors can we get here. are they retryable?

internal/tpm/tpm.go
```diff
@@ -169,6 +179,93 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get unique parameters for ldevid key: %v", err)
+	}
+	pubkey := &ecdsa.PublicKey{
+		Curve: curve,
+		X:     new(big.Int).SetBytes(unique.X.Buffer),
+		Y:     new(big.Int).SetBytes(unique.Y.Buffer),
+	}
+	// converts ecdsa.PublicKey to crypto.PublicKey
+	t.ldevidPub = pubkey
```
hexfusion, line 226: question: if this already exists do we do this all over or just return it

internal/tpm/tpm.go
```diff
@@ -169,6 +179,93 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+	return pubkey, nil
+}
+
+func (t TPM) Public() crypto.PublicKey {
+	return t.ldevidPub
```
hexfusion, line 231: this seems strange why not jsut have a single way to GetLDevIDPubKey

internal/tpm/tpm_test.go
```diff
@@ -68,6 +75,10 @@ func setupTestData(t *testing.T) TestData {
 	ldevid, err := f.tpm.CreateLDevID(*srk)
 	require.NoError(err)
 
+	// Populate the public key cache
```
hexfusion: get is get :)

internal/tpm/tpm.go
kkyrazis: This is for the `crypto.Signer` interface. I like the `GetLDevIDPubKey` method as it returns an error and we'll be able to diagnose the issue before attempting to use it as a Signer.

internal/tpm/tpm.go
kkyrazis: technically a new ldevid _could_ be recreated in between calls. Perhaps it would be simpler if the opening the TPM resulted in all of the necessary items generated.

internal/tpm/tpm.go
```diff
@@ -169,6 +179,93 @@ func (t *TPM) CreateLDevID(srk tpm2.NamedHandle) (*tpm2.NamedHandle, error) {
...
+}
```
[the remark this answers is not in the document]
kkyrazis: I don't love this approach as it hides errors.

internal/tpm/tpm.go
hexfusion: it ensures the key seems useful why not always use it and return the key if it exists?

internal/tpm/tpm.go
hexfusion: yes

internal/tpm/tpm.go
kkyrazis: It's the result of an `os.OpenFile` call. In the majority of cases here I would expect we wouldn't want to retry closing it.

internal/tpm/tpm_test.go
kkyrazis: The AI overlords love commenting. No matter how many times I tell them to pipe down.

internal/tpm/tpm.go
kkyrazis: I think doing this: https://github.com/flightctl/flightctl/pull/1409/files/5ac82085a1ff805777f4cfcd13f8fc755e50f0d9#r2222993702 would simplify a lot (including this discussion) would you like me to do that now? Or as a separate PR? 

If we have `Public` first call `GetLDevIDKeyPublic` then we should probably implement some sort of method for not redoing all of the validation if there's no need for it. It also wouldn't be able to do anything with errors other than panic.

hexfusion: followup is fine but we need it soon so we don't forget

hexfusion (approved):

PR #1415, opened 2025-07-23
EDM-1838: don't allow to delete ER of a live device

sk-ilya:
## Summary by CodeRabbit

* **New Features**
  * The system now blocks deletion of an enrollment request if a device with the same name exists, returning a clear conflict error message.
  * The API and client now support and handle HTTP 409 Conflict responses for enrollment request deletion attempts.

* **Bug Fixes**
  * Improved error reporting when attempting to delete enrollment requests tied to existing devices.

* **Tests**
  * Added integration and end-to-end tests to verify correct behavior when deleting enrollment requests with or without a device of the same name.

* **Documentation**
  * Updated API documentation to include the new 409 Conflict response for enrollment request deletion.


avishayt:
1. I assume this will break test cleanups
2. Do we want to have a "force" option for test cleanup and to allow users to get out of strange situations (e.g., decommission stuck)?

sk-ilya:
@avishayt 
> 1. I assume this will break test cleanups

Will see soon ;) 
> 2. Do we want to have a "force" option for test cleanup and to allow users to get out of strange situations (e.g., decommission stuck)?

Good idea. I can add that as well @hexfusion  @fzdarsky WDYT?

fzdarsky:
> @avishayt
> 
> > 1. I assume this will break test cleanups
> 
> Will see soon ;)
> 
> > 2. Do we want to have a "force" option for test cleanup and to allow users to get out of strange situations (e.g., decommission stuck)?
> 
> Good idea. I can add that as well @hexfusion @fzdarsky WDYT?

As @avishayt hinted at, we had a similar situation with `DELETE device`, where we were considering a business logic to not allow deletes when a device has not been decommissioned. In the end the decision was that from an API perspective that should be allowed, but the UI/CLI would issue a warning and ask for confirmation. IOW, we didn't use a `--force` there.
One advantage: From RBAC perspective we could grant an Operator persona the permission to decommission, but not to delete, while an Admin persona could have the permission to delete, too. This would be something we couldn't do with our current RBAC model with a `--force` flag.

sk-ilya:
@fzdarsky currently, the only operation that we have is "delete". Is it worth adding an additional (sub)resource for ER's similar to `device/decomission`  for RBAC? 
cc @avishayt

avishayt:
Hold until after code freeze

fzdarsky:
@sk-ilya after discussing through the use cases and options, let's do the following:

- Yes, we want to enforce that an ER cannot be deleted until the corresponding Device is deleted --> 409. But please remove the check whether a device is decom'ed.
- There is no need for a `--force` flag, because users with the right permissions can delete the Device, then delete the ER. Device deletion and decommission are separate permissions.

test/e2e/cli/cli_test.go
```diff
@@ -132,11 +132,6 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(device).ToNot(BeNil())
 
-			By("Deleting the enrollment request")
-			out, err = harness.ManageResource("delete", fmt.Sprintf("er/%s", erName))
-			Expect(err).ToNot(HaveOccurred())
-			Expect(out).To(ContainSubstring("completed"))
```
amalykhi, line 138: `By("Attempting to delete the enrollment request for live device")
			out, err = harness.ManageResource("delete", fmt.Sprintf("er/%s", erName))
			Expect(out).To(ContainSubstring("device exists"))`

avishayt (approved):

PR #1419, opened 2025-07-23
EDM-1810: svc signer added

asafbennatan:
svc signer added


internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (s *SignerServerSvc) Name() string {
+	return s.name
+}
+
+func (s *SignerServerSvc) Verify(ctx context.Context, request api.CertificateSigningRequest) error {
+	// For server service certificates, we require CLI authentication
+	// Users must authenticate through the CLI (same as other operations)
+	identity, err := authcommon.GetIdentity(ctx)
```
asafbennatan: anyone reviewing this - i wonder if this is the correct way to verify we only allow access from CLI and not for device

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,129 @@
...
+	if serviceName == "" {
+		return fmt.Errorf("CSR CommonName %q must include a service name after prefix %q", parsedCSR.Subject.CommonName, servicePrefix)
+	}
+
+	return nil
+}
+
+func (s *SignerServerSvc) Sign(ctx context.Context, request api.CertificateSigningRequest) ([]byte, error) {
+
+	signer := s.ca.PeerCertificateSignerFromCtx(ctx)
```
asafbss: The peer certificate (where we set the signer in context) happens synchronously during the end-user's request, but signing typically occurs asynchronously—after the request has already returned. By that time, the context is no longer valid.

internal/crypto/cert.go
```diff
@@ -282,6 +282,7 @@ func (caClient *CAClient) IssueRequestedClientCertificate(ctx context.Context, c
 func (caClient *CAClient) IssueRequestedServerCertificateAsX509(ctx context.Context, csr *x509.CertificateRequest, expirySeconds int, opts ...CertOption) (*x509.Certificate, error) {
 	return caClient.caBackend.IssueRequestedCertificateAsX509(ctx, csr, expirySeconds, []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth}, opts...)
 }
+
```
asafbss: Remove. The file should not be part of the commit.

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (s *SignerServerSvc) Name() string {
+	return s.name
+}
+
+func (s *SignerServerSvc) Verify(ctx context.Context, request api.CertificateSigningRequest) error {
+	// For server service certificates, we require CLI authentication
+	// Users must authenticate through the CLI (same as other operations)
+	signer := s.ca.PeerCertificateSignerFromCtx(ctx)
```
asafbss: consider using the same as:
https://github.com/flightctl/flightctl/blob/381456599e345065bf3dbace16c3f6926cb653e8/internal/crypto/signer/signer_client_bootstrap.go#L34

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, fmt.Errorf("error getting identity for service certificate approval: %w", err)
+	}
+	// Validate that the authenticated user has appropriate permissions
+	if identity != nil && identity.Username != "" {
+		// Restrict server certificate requests to specific roles
+		// Only allow users with admin or installer roles to request server certificates
+		allowedRoles := []string{"flightctl-admin", "flightctl-installer"}
+		hasAllowedRole := false
+
+		for _, role := range allowedRoles {
```
asafbss: Consider moving to a small (not-exported) function in the same file. You can easily use 'return'

asafbss: Or even better, move to common.go file in the same package

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,129 @@
...
+
+	expirySeconds := signerServerSvcExpiryDays * 24 * 60 * 60
+	if request.Spec.ExpirationSeconds != nil && *request.Spec.ExpirationSeconds < expirySeconds {
+		expirySeconds = *request.Spec.ExpirationSeconds
+	}
+
+	opts := []certOption{
+		WithExtension(OIDOrgID, NullOrgId.String()),
+	}
+	return s.ca.IssueRequestedServerCertificate(
```
asafbss, line 91: Make sure that Subject Alternative Names (SANs) are fully supported

internal/crypto/signer/signer_server_svc.go
asafbennatan: done

internal/crypto/signer/signer_server_svc.go
asafbennatan: it is handled the actual CSR bytes are passed in the CertificateRequestObject->Spec->Request

and these are simply signed

the only place i needed to adjust is cli - which is done

internal/crypto/signer/signer_server_svc.go
asafbss: Might be better in 'common.go' if other signers use the same logic in the future.

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,137 @@
...
+			return fmt.Errorf("user %s is not authorized to request server certificates. Required roles: %v",
+				identity.Username, allowedRoles)
+		}
+	}
+	return nil
+}
+
+func (s *SignerServerSvc) Sign(ctx context.Context, request api.CertificateSigningRequest) ([]byte, error) {
+
+	signer := s.ca.PeerCertificateSignerFromCtx(ctx)
```
asafbss: The peer certificate (where we set the signer in context) happens synchronously during the end-user's request, but signing typically occurs asynchronously—after the request has already returned. By that time, the context is no longer valid (i.e., should use in the verify)

internal/cli/certificate.go
```diff
@@ -279,6 +284,16 @@ func (o *CertificateOptions) submitCsrWithRetries(ctx context.Context, c *apicli
 	return "", fmt.Errorf("submitting CSR failed after %d attempts, giving up", maxAttempts)
 }
 
+// getUsagesForSigner returns the appropriate usage strings based on the signer name
+func getUsagesForSigner(signerName string) *[]string {
```
asafbss: I don't think this is necessary, since the usage is determined by the signer.
The new signer calls IssueRequestedServerCertificate, which sets the appropriate key usage:
https://github.com/flightctl/flightctl/blob/832eb63e92ff6661616e883cde87d1f35c2d6dad/internal/crypto/cert.go#L283

internal/cli/certificate.go
```diff
@@ -323,7 +356,7 @@ func createCsr(o *CertificateOptions, name string, priv crypto.PrivateKey) ([]by
 			ExpirationSeconds: &expirationSeconds,
 			Request:           csrPEM,
 			SignerName:        o.SignerName,
-			Usages:            &[]string{"clientAuth", "CA:false"},
+			Usages:            getUsagesForSigner(o.SignerName),
```
asafbss: Let's remove this entirely or keep the default AS IS.
See my comment I added for getUsagesForSigner

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (s *SignerServerSvc) Name() string {
+	return s.name
+}
+
+func (s *SignerServerSvc) Verify(ctx context.Context, request api.CertificateSigningRequest) error {
+	// For server service certificates, we require CLI authentication
+	// Users must authenticate through the CLI (same as other operations)
+
+	if _, err := PeerCertificateFromCtx(ctx); err == nil {
```
asafbss, line 32: consider using the same as:
https://github.com/flightctl/flightctl/blob/381456599e345065bf3dbace16c3f6926cb653e8/internal/crypto/signer/signer_client_bootstrap.go#L34

internal/crypto/signer/signer_server_svc.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	// Validate that the authenticated user has appropriate permissions
+	if identity != nil && identity.Username != "" {
+		// Restrict server certificate requests to specific roles
+		// Only allow users with admin or installer roles to request server certificates
+		allowedRoles := []string{"flightctl-admin", "flightctl-installer"}
+		hasAllowedRole := false
+
+		for _, role := range allowedRoles {
+			for _, userRole := range identity.Groups {
```
asafbss: I don't see any usage/reference in the repo for the user group.

internal/cli/certificate.go
```diff
@@ -96,6 +99,8 @@ func (o *CertificateOptions) Bind(fs *pflag.FlagSet) {
 	fs.StringVarP(&o.OutputDir, "output-dir", "d", o.OutputDir, "Specify desired output directory for key, cert, and ca files.")
 	fs.StringVarP(&o.SignerName, "signer", "s", o.SignerName, "Specify the signer of the certificate request.")
 	fs.BoolVarP(&o.EncryptKey, "encrypt", "e", o.EncryptKey, "Option to encrypt key file with a password from env var $FCPASS, or if $FCPASS is not set password must be provided during runtime.")
+	fs.StringSliceVarP(&o.DNSNames, "dns-names", "", o.DNSNames, "Comma-separated list of DNS names to include in the certificate's Subject Alternative Names.")
+	fs.StringSliceVarP(&o.IPAddresses, "ip-addresses", "", o.IPAddresses, "Comma-separated list of IP addresses to include in the certificate's Subject Alternative Names.")
```
asafbss: We'll need architecture approval for this:
@avishayt

internal/crypto/signer/signer_server_svc.go
asafbss: @rawagner WDYT?

internal/cli/certificate.go
asafbss: I'm tempted to leave this out for now to keep the CLI surface minimal and focused. We can follow the Kubernetes approach and have users generate the CSR themselves (including SANs) and submit it via a CertificateSigningRequest resource.

jira-ai-issue-solver: is there an actual way to create a CertificateSigningRequest from the CLI by providing a raw CSR ? i dont see such option (yes the api does receive that as raw bytes , but the cli does not allow this as far as i can see) 
regardless i find the specific parameter approach a better api

internal/crypto/signer/signer_server_svc.go
jira-ai-issue-solver: i dont see where the signing is done in an async manner - can you explain ?
IMO even if it was done async - we should still know the details of who initiated this - in any case i cant find async stuff here related to approve

PR #1425, opened 2025-07-24
EDM-1766: Small SELinux Related Tweaks

keitwb:
- make: Depend on the files in the selinux dir and not just directory entries

 - Use `restorecon` when updating the agent executable via `make update-vm-agent`

 - Always build the RPMs with the ci builder image to avoid issues when building with newer Fedora versions that will cause backwards compatibility issues with SELinux on Centos 9.
 
 - Add docs for policy diffing

## Summary by CodeRabbit

* **Chores**
  * Improved RPM build reliability by always building in a container environment, ensuring consistent SELinux policy handling.
  * Enhanced Makefile dependencies so RPMs are rebuilt when any file in the SELinux packaging directory changes.
  * Updated VM deployment process to automatically restore SELinux security context on the agent binary after updates.
  * Added documentation clarifying SELinux policy restrictions by comparing agent permissions to an unconfined root domain.

### DIFF
--- a/cmd/flightctl/main.go
+++ b/cmd/flightctl/main.go
@@ -28,6 +28,7 @@ func NewFlightCtlCommand() *cobra.Command {
 	cmd.AddCommand(cli.NewCmdDelete())
 	cmd.AddCommand(cli.NewCmdApprove())
 	cmd.AddCommand(cli.NewCmdCSRConfig())
+	cmd.AddCommand(cli.NewCmdConfig())
 	cmd.AddCommand(cli.NewCmdDecommission())
 	cmd.AddCommand(cli.NewCmdDeny())
 	cmd.AddCommand(cli.NewCmdLogin())
--- a/internal/cli/apply.go
+++ b/internal/cli/apply.go
@@ -14,7 +14,6 @@ import (
 	"strings"
 
 	apiclient "github.com/flightctl/flightctl/internal/api/client"
-	"github.com/flightctl/flightctl/internal/client"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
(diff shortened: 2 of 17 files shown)


packaging/selinux/README.md
```diff
@@ -98,6 +98,16 @@ ls -Z /usr/bin/flightctl-agent
...
 
+## Check what we are really blocking
+
+`unconfined_t` as root is equivalent to root access with SELinux disabled on a normal Linux system. To
+see how close the agent is to the limitless policy, run this in the agent:
+
+`$ grep -Fxvf <(sudo sesearch --allow -s flightctl_agent_t) <(sudo sesearch --allow -s unconfined_t)`
+
+This will print everything that is allowed in unconfied that is not specified in our own policy.
+This is a way to kind of answer the question, "What are we actually blocking with this policy?".
```
hexfusion, line 109: nice  :+1:

hack/build_rpms.sh
```diff
@@ -18,18 +18,17 @@ if [ -n "${FLIGHTCTL_RPM:-}" ]; then
     exit 0
 fi
 
-# our RPM build process works in rpm bases systems so we wrap it if necessary
-if [[ -n "$FORCE_PACKIT_CONTAINER" ]] || ! command -v packit >/dev/null 2>&1; then
-    echo "Building RPMs on a system without packit, using container"
-    cat >bin/build_rpms.sh <<EOF
+# Given that the SELinux policies are so sensitive to versioning issues, make sure to always build
+# the RPM in a known environment.
+echo "Building RPMs in container"
```
hexfusion, line 23: @gshilin-sdb can you confirm these changes will not affect our downstream builds.

hexfusion (commented):
looks good just would like @gshilin-sdb to take a quick peak with the rpm build actions I don't think it will affect us but just to make sure.

hack/build_rpms.sh
avishayt: @gshilin-sdb

gshilin-sdb: We don't use this script to build downstream rpms, but spec only.

keitwb:
@avishayt Are you good with merging this before code freeze?

PR #1428, opened 2025-07-24
EDM-1702: Adopt task queue consumer and publisher pattern for periodi…

DakCrowder:
…c tasks split by organization

### DIFF
--- a/internal/periodic_checker/consumer.go
+++ b/internal/periodic_checker/consumer.go
@@ -1,0 +1,63 @@
+package periodic
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/flightctl/flightctl/internal/consts"
+	"github.com/flightctl/flightctl/internal/util"
+	"github.com/flightctl/flightctl/pkg/queues"
+	"github.com/sirupsen/logrus"
+)
+
+func consumeTasks(executors map[PeriodicTaskType]PeriodicTaskExecutor) queues.ConsumeHandler {
+	return func(ctx context.Context, payload []byte, log logrus.FieldLogger) error {
+		var reference PeriodicTaskReference
+		if err := json.Unmarshal(payload, &reference); err != nil {
+			log.WithError(err).Error("failed to unmarshal consume payload")
+			return err
+		}
+
+		log.Infof("consuming %s task for organization %s", reference.Type, reference.OrgID)
+
+		executor, exists := executors[reference.Type]
+		if !exists {
+			log.Errorf("no executor found for task type %s", reference.Type)
+			return nil
+		}
+
+		// Add the orgID to the context
+		ctx = util.WithOrganizationID(ctx, reference.OrgID)
+
+		executor.Execute(ctx, log)
+
+		return nil
+	}
+}
+
+type PeriodicTaskConsumer struct {
+	queuesProvider queues.Provider
+	log            logrus.FieldLogger
+	executors      map[PeriodicTaskType]PeriodicTaskExecutor
+}
+
+func NewPeriodicTaskConsumer(queuesProvider queues.Provider, log logrus.FieldLogger, executors map[PeriodicTaskType]PeriodicTaskExecutor) *PeriodicTaskConsumer {
+	return &PeriodicTaskConsumer{
+		queuesProvider: queuesProvider,
+		log:            log,
+		executors:      executors,
+	}
+}
+
+func (c *PeriodicTaskConsumer) Start(ctx context.Context) error {
+	consumer, err := c.queuesProvider.NewConsumer(consts.PeriodicTaskQueue)
+	if err != nil {
+		return err
+	}
+
+	if err = consumer.Consume(ctx, consumeTasks(c.executors)); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/internal/periodic_checker/publisher.go
+++ b/internal/periodic_checker/publisher.go
@@ -1,0 +1,186 @@
+package periodic
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	api "github.com/flightctl/flightctl/api/v1alpha1"
+	"github.com/flightctl/flightctl/internal/consts"
+	"github.com/flightctl/flightctl/internal/kvstore"
+	"github.com/flightctl/flightctl/pkg/queues"
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	DefaultTaskTickerInterval = 5 * time.Second
+	DefaultOrgTickerInterval  = 5 * time.Minute
+)
+
+const (
+	// RedisKeyPeriodicTaskLastRun is the Redis key pattern for storing the last run time of periodic tasks
+	RedisKeyPeriodicTaskLastRun = "periodic_task:last_run:"
+)
+
+type OrganizationService interface {
+	ListOrganizations(ctx context.Context) (*api.OrganizationList, api.Status)
+}
+
+type PeriodicTaskPublisher struct {
+	publisher     queues.Publisher
+	log           logrus.FieldLogger
+	tasksMetadata []PeriodicTaskMetadata
+	orgService    OrganizationService
+	kvStore       kvstore.KVStore
+	organizations map[uuid.UUID]bool // Track which organizations are registered
+
+	// Configurable intervals, currently only overridden for testing
+	taskTickerInterval time.Duration
+	orgTickerInterval  time.Duration
+}
+
+func NewPeriodicTaskPublisher(log logrus.FieldLogger, kvStore kvstore.KVStore, orgService OrganizationService, queuesProvider queues.Provider, tasksMetadata []PeriodicTaskMetadata) (*PeriodicTaskPublisher, error) {
+	publisher, err := queuesProvider.NewPublisher(consts.PeriodicTaskQueue)
+	if err != nil {
+		log.WithError(err).Error("failed to create periodic task publisher")
+		return nil, err
+	}
+	return &PeriodicTaskPublisher{
+		publisher:          publisher,
+		log:                log,
+		orgService:         orgService,
+		kvStore:            kvStore,
+		tasksMetadata:      tasksMetadata,
+		taskTickerInterval: DefaultTaskTickerInterval,
+		orgTickerInterval:  DefaultOrgTickerInterval,
+	}, nil
+}
+
+func (p *PeriodicTaskPublisher) publishTasks(ctx context.Context) {
+	for orgID, _ := range p.organizations {
+		for _, taskMetadata := range p.tasksMetadata {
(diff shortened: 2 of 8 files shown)


DakCrowder:
@coderabbitai review

DakCrowder:
@coderabbitai review

avishayt:
Hold until after code freeze

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,268 @@
...
+		organizations[orgId] = true
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	for orgID := range p.organizations {
+		if !organizations[orgID] {
+			p.log.Infof("Organization %s is no longer registered, removing from tracking", orgID)
+			delete(p.organizations, orgID)
```
kkyrazis: minor: If we're just setting the organizations to the new list on line 234 that we've just fetched there's no real need to delete the organizations. The Info log still seems useful for debugging purposes though.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+
+	p.organizations = organizations
+}
+
+func (p *PeriodicTaskPublisher) stopAll() {
```
kkyrazis: Perhaps clearOrgs is a better name as there doesn't really seem to be much stopping required.

internal/periodic_checker/publisher.go
keitwb: This seems redundant with what you are doing below when you just reassign the whole org map, which makes any changes to the old `p.organizations` irrelevant.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,268 @@
...
+
+	for _, org := range orgList.Items {
+		orgId, err := uuid.Parse(*org.Metadata.Name)
+		if err != nil {
+			p.log.Errorf("Failed to parse organization ID %s: %v", *org.Metadata.Name, err)
+			continue
+		}
+
+		p.mu.RLock()
+		isRegistered := p.organizations[orgId]
```
keitwb: I guess this is helpful for generating the log message but it doesn't seem necessary as a separate var otherwise.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,268 @@
...
+	p.organizations[orgID] = true
+}
+
+func (p *PeriodicTaskPublisher) hasOrganization(orgID uuid.UUID) bool {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	return p.organizations[orgID]
+}
+
+func (p *PeriodicTaskPublisher) SetTaskTicker(interval time.Duration) {
```
keitwb: Since this is only used in the Start method and isn't going to have any effect after starting, it should really be part of the New... method only. Having it like this implies you can dynamically change it after startup.

pkg/queues/redis_provider.go
```diff
@@ -162,10 +184,12 @@ func (r *redisQueue) consumeOnce(ctx context.Context, handler ConsumeHandler) er
...
-	msgs, err := r.client.XRead(ctx, &redis.XReadArgs{
-		Streams: []string{r.name, "0"},
-		Count:   1,
-		Block:   0,
+	msgs, err := r.client.XReadGroup(ctx, &redis.XReadGroupArgs{
+		Group:    r.name,
+		Consumer: r.consumerID,
+		Streams:  []string{r.name, ">"}, // ">" means new messages only
+		Count:    1,
+		Block:    0,
```
kkyrazis, line 193: if this is just run in a tight loop it may be beneficial to add some blocking.

pkg/queues/redis_provider.go
```diff
@@ -69,16 +71,35 @@ func (r *redisProvider) newQueue(queueName string) (*redisQueue, error) {
...
 	queue := &redisQueue{
-		name:   queueName,
-		client: r.client,
-		log:    r.log,
-		wg:     r.wg,
+		name:       queueName,
+		client:     r.client,
+		log:        r.log,
+		wg:         r.wg,
+		consumerID: uuid.New().String(),
```
keitwb, line 81: This should ideally be stable across restarts (which would be easier to implement if the periodic service ran as a StatefulSet and not a Deployment). Otherwise if the service crashes while processing a message you would have to do the logic described at https://redis.io/docs/latest/develop/data-types/streams/#recovering-from-permanent-failures to process the message. Maybe it doesn't matter since it would just get rescheduled so many minutes later anyway but something to consider.

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,148 @@
...
+	"sync"
+
+	"github.com/flightctl/flightctl/internal/consts"
+	"github.com/flightctl/flightctl/internal/util"
+	"github.com/flightctl/flightctl/pkg/queues"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	DefaultNumConsumers = 5
```
kkyrazis: Is there any concern with multiple consumers processing the same task in parallel? Hypothetically, if a periodic task took long enough to complete that a second message was published requesting another worker perform the task would we be in trouble?

DakCrowder: Looking at the prior implementation calls would block and multiple runs of the same task could not occur.  Tasks are in theory idempotent but not 100% sure how concurrent execution affects all of them.  The non-periodic tasks could currently get into situations where they get enqueued multiple times before being processed for an entity or have mid execution contention and are designed around that.  We could introduce a redis lock on each task but that also raises additional questions (not in a bad way, but things we would need to further clarify) around what does our periodic interval mean.  If a locked task is consumed do we error then wait for the next publish, do we try and execute as soon as we can or re-enqueue etc.

Need to think about this a bit more, but either way I'd also argue for splitting that out into a second block of work / PR.

internal/periodic_checker/publisher.go
DakCrowder: Good catch, resolved

internal/periodic_checker/consumer.go
kkyrazis: I guess the problem exists today if you just spin up a second periodic worker replica. How easy would it be to wire up a configuration to allow us to fall back to a single consumer if everything falls over?

> We could introduce a redis lock on each task but that also raises additional questions

We could also take a more Kafka like approach and partition streams such that each consumer always processes messages for the same org. But that opens up another can of worms.

DakCrowder: Yeah the existing implementation (and this one too especially due to the publisher) really only work for one deployed periodic service but I see this as a _step_ towards being able to scale out workers.  

> How easy would it be to wire up a configuration to allow us to fall back to a single consumer if everything falls over?

Depends on what "everything falls over?" entails ha, to be clear were you thinking something automatic or just a config option for numConsumers?  The latter we probably want anyway and should be trivial to add.  The former would also be good to be coupled with some form of alerting / indicator to users things are in a degraded state which also doesn't exist.

kkyrazis: Just a simple config value. Auto scaling would be way too much.

pkg/queues/redis_provider.go
keitwb: I think `0` means to block without any timeout, which is the most efficient thing to do. See https://redis.io/docs/latest/develop/data-types/streams/#listening-for-new-items-with-xread. Apparently in the go-redis client you have to pass a negative number to disable blocking.

kkyrazis: Good catch, I forgot 0 == :infinity:

internal/periodic_checker/consumer.go
DakCrowder: Added an option for the consumers to be specified and updated the helm deployment

pkg/queues/redis_provider.go
DakCrowder: Sounds good - reverted back to 0

pkg/queues/redis_provider.go
```diff
@@ -213,12 +237,12 @@ func (r *redisQueue) consumeOnce(ctx context.Context, handler ConsumeHandler) er
 				handlerSpan.RecordError(err)
 				handlerSpan.SetStatus(codes.Error, "unexpected body type")
 
-				// Attempt to delete the bad message
-				if _, delErr := r.client.XDel(ctx, r.name, entry.ID).Result(); delErr != nil {
-					parentSpan.RecordError(delErr)
-					parentSpan.SetStatus(codes.Error, delErr.Error())
+				// Acknowledge the bad message to remove it from pending
```
kkyrazis: XAck removes the message from the consumer groups pending list but I would expect that it doesn't actually remove the item from the stream itself. (I've clearly been wrong about expected behavior before). Do we have some other mechanism to prevent unbounded growth of the stream?

DakCrowder: I think you are correct, I read this line in the docs "the PEL entry about this message is also purged, releasing memory from the Redis server." and got confused, it looks like a second call is needed.  For now will add an delete for the msg.

pkg/queues/redis_provider.go
DakCrowder: We also have non k8s deployments that could make this interesting as we couldn't use a StatefulSet all the time :cry: 

For periodic tasks for now I think its ok to just drop + reschedule - but in general additional hardening / retry logic / failure recovery would all be good improvements for the service.  Honestly could be a whole feature around periodic improvements.  Also I don't think consistent consumer ids would be all that is needed, we would also need to have the consumers check which messages they have marked as pending but not ack'd.

It does raise an interesting point about the non-periodic tasks though.  The tasks are currently always run by a single consumer (using the previous xRead and no consumer groups) and I think would pick back up where they left off on a tasks service crash (w/ potential reprocessing).  However adding the consumer groups and no recovery mechanism means that functionality is now changed.  But we also also deleted every message on failure so a finnicky network call or something would also result in a failure without retry so its not like we were providing strict guarantees about execution (more of a we'll try it at least once).

So with all that being said, I think it would be worth looking at using [autoclaim](https://redis.io/docs/latest/commands/xautoclaim/) (would have to define heuristics around when to claim) to clean up potentially crashed tasks.

pkg/queues/redis_provider.go
DakCrowder: Added delete coupled w/ the ack + some test coverage

avishayt:
I'm not sure about this architecture.
If you just want goroutines, why not use channels?  Redis seems overkill in this case.
If the goal is scaling out, maybe flightctl-periodic should just be a simple loop that sends tasks to flightctl-worker?

DakCrowder:
Closing for now in favor of https://github.com/flightctl/flightctl/pull/1493 which simplifies the architecture to use channels and punts on and redis-based refactoring until further need is demonstrated / some work shakes out with other tasks and we can conform to that structure.

PR #1434, opened 2025-07-25
EDM-1766: Remove bwrap for console shell

keitwb:
This directly invokes bash. SELinux policy will be shared with the agent for now as bash will run in the same domain as the agent.


internal/agent/device/console/session.go
```diff
@@ -37,31 +37,22 @@ func (s *session) close() error {
...
 	if metadata.TTY {
 		args = append(args, "-i", "-l")
 	}
+
 	if metadata.Command != nil && metadata.Command.Command != "" {
 		args = append(args, "-c", strings.Join(append([]string{metadata.Command.Command}, metadata.Command.Args...), " "))
 	}
-	ret := s.executor.CommandContext(ctx, "bwrap", args...)
+
+	ret := s.executor.CommandContext(ctx, "/bin/bash", args...)
```
hexfusion, line 50: bubblewrap would be useful in providing a more restricted env. while we might need something like this in the future its overly complicated for now trying to provide full root access.

PR #1444, opened 2025-07-28
EDM-917: Perform message throttling in Web APIs

avishayt:
Implements in-memory HTTP rate limiting using github.com/go-chi/httprate for both user-facing and agent-facing APIs. Features include:

- IP-based rate limiting with configurable limits (default: 60 req/min)
- Environment variable overrides (RATE_LIMIT_REQUESTS, RATE_LIMIT_WINDOW)
- Stricter login endpoint protection (default: 10 req/hour)
- Proper HTTP headers (X-Ratelimit-*, Retry-After)
- Comprehensive test coverage
- Complete user documentation

Applied to both main API server and agent server middleware stacks.

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable IP-based rate limiting for API requests, with distinct limits for general and authentication endpoints.
  * API returns standardized HTTP 429 (Too Many Requests) responses including appropriate headers and JSON error details when limits are exceeded.
  * Rate limiting settings can be managed via environment variables, configuration files, and deployment templates (Helm and Podman).
* **Bug Fixes**
  * Enhanced client and server support for handling HTTP 429 responses consistently across all API endpoints.
* **Documentation**
  * Added detailed user documentation on rate limiting configuration, behavior, and troubleshooting.
  * Updated security guidelines to include rate limiting best practices and considerations.
  * OpenAPI specification updated to document HTTP 429 responses for all relevant endpoints.
* **Tests**
  * Added extensive tests verifying rate limiting enforcement, header correctness, error responses, and client IP detection logic.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -35,6 +35,8 @@ jobs:
       - name: Running Integration tests
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
         run: make integration-test VERBOSE=true
+        env:
+          DISABLE_FIPS: true
 
 
 
--- a/.github/workflows/pr-e2e-testing.yaml
+++ b/.github/workflows/pr-e2e-testing.yaml
@@ -55,7 +55,7 @@ jobs:
 
       - name: Deploy the flightctl server
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
-        run: make deploy
+        run: DISABLE_FIPS="true" make deploy
 
       - name: Deploy the E2E side services, registry and git
         run: make deploy-e2e-extras
--- a/.github/workflows/pr-smoke-testing.yaml
+++ b/.github/workflows/pr-smoke-testing.yaml
@@ -39,16 +39,12 @@ jobs:
 
       - name: Deploy
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
-        run: make deploy
+        run: DISABLE_FIPS="true" make deploy
 
       - name: Check
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
         run: |
           kubectl get pods --all-namespaces
-
-      - name: Make the cmdline tool
-        if: ${{ steps.filter.outputs.notdocs == 'true' }}
-        run: make build
 
       - name: Apply device
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
@@ -69,6 +65,10 @@ jobs:
       - name: Apply resourcesync
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
         run: bin/flightctl apply -f examples/resourcesync.yaml
+
+      - name: Build the simulator
+        if: ${{ steps.filter.outputs.notdocs == 'true' }}
+        run: DISABLE_FIPS="true" make build-devicesimulator
 
       - name: Simulator run
         if: ${{ steps.filter.outputs.notdocs == 'true' }}
--- a/.github/workflows/publish-cli-bins.yaml
+++ b/.github/workflows/publish-cli-bins.yaml
@@ -79,7 +79,7 @@ jobs:
 
       - name: Build
         run: |
-          GOOS="${{ matrix.os }}" GOARCH="${{ matrix.arch }}" make build-cli
+          DISABLE_FIPS="true" GOOS="${{ matrix.os }}" GOARCH="${{ matrix.arch }}" make build-cli
           mv "bin/flightctl${{ matrix.ext }}" "flightctl-${{ matrix.os }}-${{ matrix.arch}}${{ matrix.ext }}"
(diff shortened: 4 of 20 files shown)


celdrake:
I'd like to test it out to see if there could be any issues with the current UI.
I'll check it tomorrow after I finish with Alerts UI.

celdrake:
> I'd like to test it out to see if there could be any issues with the current UI. I'll check it tomorrow after I finish with Alerts UI.

The UI is working normally, I couldn't see any request rejection or abnormal behaviour.
LGTM.

internal/api_server/middleware/ratelimit.go
```diff
@@ -0,0 +1,83 @@
...
+
+		// Return API status response
+		status := api.StatusTooManyRequests(errorMessage)
+		transport.SetResponse(w, nil, status)
+	}
+
+	// Create the rate limiter with custom limit handler
+	limiter := httprate.NewRateLimiter(requests, window,
+		httprate.WithLimitHandler(limitHandler),
+	)
```
rawagner: maybe a silly question as Im not familiar with this rate limiter, but where do we specify that we are limiting by IP?

Also note that UI uses reverse-proxy so we need to ensure that the IP is determined by `X-Forwarded-For` header and also that our reverse-proxy is properly sending this header.

avishayt: Great question - I adjusted it to use `httprate.WithKeyByRealIP()`, so it will use those headers if they exist.  Also updated the documentation.

docs/user/rate-limiting.md
```diff
@@ -0,0 +1,265 @@
...
+## How It Works
+
+### IP Detection
+
+Flight Control automatically detects the real client IP address using the following priority order:
+
+1. **X-Forwarded-For header** (most common with reverse proxies)
+2. **X-Real-IP header**
+3. **True-Client-IP header**
+4. **Direct connection IP** (fallback)
```
rawagner: I think the order is actually 
3, 2, 1, 4 

[ref](https://github.com/go-chi/httprate/blob/master/httprate.go#L55-L59)

rawagner: I think we will want some additional hardening here. The header values should be used only for some specific sources - like the UI proxy, otherwise anybody can set these and change them for every request which would overcome our rate limiter.

docs/user/rate-limiting.md
```diff
@@ -0,0 +1,265 @@
...
+    authWindow: "1h"    # Auth time window
+```
+
+### Default Values
+
+If not configured, rate limiting is **disabled by default**. To enable it, you must explicitly set the configuration values.
+
+### Environment Variables
+
+You can also configure rate limiting using environment variables:
```
DakCrowder, line 58: Can a mention be added that the env vars override the yaml config?

docs/user/rate-limiting.md
```diff
@@ -0,0 +1,265 @@
...
+    window: "1m"        # Time window (e.g., "1m", "1h", "1d")
+    
+    # Authentication-specific rate limiting (stricter)
+    authRequests: 10    # Maximum auth requests per window
+    authWindow: "1h"    # Auth time window
+```
+
+### Default Values
+
+If not configured, rate limiting is **disabled by default**. To enable it, you must explicitly set the configuration values.
```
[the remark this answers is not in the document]
avishayt: This is the code.  If not set in the config, there's no rate limiting.

if cfg.Service == nil || cfg.Service.RateLimit == nil {
		// If no rate limit config, return a no-op middleware
		return func(next http.Handler) http.Handler {
			return next
		}
	}

docs/user/rate-limiting.md
avishayt: You're absolutely right about both points - thanks for that.  I updated the documentation, and also changed the implementation to allow configuring "trusted proxies".  The middleware will now use information from headers only from trusted proxies.

docs/user/rate-limiting.md
avishayt: Done!

internal/api_server/middleware/ratelimit_test.go
```diff
@@ -0,0 +1,784 @@
...
+
+	req := httptest.NewRequest("GET", "/test", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	assert.Equal(t, http.StatusOK, w.Code)
+	assert.Equal(t, "100", w.Header().Get("X-RateLimit-Limit"))
+	assert.NotEmpty(t, w.Header().Get("X-RateLimit-Remaining"))
+	assert.NotEmpty(t, w.Header().Get("X-RateLimit-Reset"))
+}
```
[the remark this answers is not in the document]
avishayt: There are several tests that check for http.StatusTooManyRequests

DakCrowder (approved):

PR #1447, opened 2025-07-28
EDM-1754: TPM Take Owner Ownership

kkyrazis:
This PR adds functionality for setting a password on the Owner Hierarchy on startup. 

**Note**: This is probably not something that we want to run on personal TPMs. There's a non-zero chance that someone running this will lose their password and you'll be stuck.

* By default, we disable taking ownership of the hierarchy.
* If ownership is not established, we'll take ownership by generating a random 32 byte password 
* Modifies the persistence file to allow for more values to be added. Now we'll persist the LDevID, the LAK, and OwnerHierarchy Password in the same file. (breaking change to the current persistence schema)

Note: In doing this I had to change the AK creation as the APIs that were being used don't have support for auth values.

Note: This PR does not handle the following portions of the ticket:
```
*FC agent should set up distinct auth values (different from Storage Hierarchy auth) for each child key OTel needs to access and make sure OTel can access those values.
*Make sure OTel is not (attempting to or) taking ownership of the storage hierarchy during its operations.
```
as there is a separate ticket directly of OTEL. I think it's best to defer it to that as this PR is already very large.

## Summary by CodeRabbit

* **New Features**
  * Introduced persistent storage for TPM keys and sealed passwords, enabling secure storage and retrieval across restarts.
  * Added support for managing TPM ownership and storage hierarchy passwords, including automatic password generation, sealing, and reset functionality.
  * Enhanced key management with robust attestation and quote generation, supporting explicit PCR policy enforcement.
  * Added new TPM key templates for storage password sealing and attestation keys.
  * Provided flexible TPM client creation supporting simulator connections for testing.
  * Added functions to handle PCR selections for TPM operations.

* **Bug Fixes**
  * Improved error handling and robustness in TPM key and password management workflows.

* **Tests**
  * Expanded and refactored test coverage for TPM key management, ownership handling, and password lifecycle scenarios.


internal/tpm/client.go
```diff
@@ -283,14 +308,19 @@ func (t *Client) generateSRKPrimary() (*tpm2.NamedHandle, error) {
 // The LAK is an asymmetric key that persists for the device's lifecycle and can be used
 // to sign TPM-internal data such as attestations. This is a Restricted signing key.
 // Key attributes: Restricted=yes, Sign=yes, Decrypt=no, FixedTPM=yes, SensitiveDataOrigin=yes
-func (t *Client) CreateLAK() (*client.Key, error) {
-	// AttestationKeyECC generates and loads a key from AKTemplateECC in the Owner (aka 'Storage') hierarchy.
-	return client.AttestationKeyECC(t.conn)
+// The LAK is created as a child of the SRK to properly handle storage hierarchy authentication.
+func (t *Client) ensureLAK() (*tpm2.NamedHandle, error) {
+	var err error
+	t.lak, err = t.ensureKey(t.persistence.loadLAKBlob, t.persistence.saveLAKBlob, AttestationKeyTemplate)
```
kkyrazis, line 314: note: the previous method doesn't support supplying auth values. Also, the implementation persisted the key created via evict_control to a hardcoded key. I've reworked things to use the api that supports auth and reuse our key persistence mechanism. I did my best to keep the previous Quote and Attestion methods similar.

internal/agent/config/config.go
```diff
@@ -127,6 +127,8 @@ type TPM struct {
 	Path string `json:"tpm-path,omitempty"`
 	// PersistencePath specifies the file path for TPM key blob persistence
 	PersistencePath string `json:"tpm-persistence-path,omitempty"`
+	// SkipOwnership indicates whether to skip TPM ownership establishment
+	SkipOwnership bool `json:"tpm-skip-ownership,omitempty"`
```
hexfusion: Skip vs Enable I think we use Enable for another

internal/tpm/client.go
```diff
@@ -118,11 +138,11 @@ func (t *Client) GetPath() string {
...
 
 // GetLocalAttestationPubKey returns the public key of the Local Attestation Key.
-func (t *Client) GetLocalAttestationPubKey() crypto.PublicKey {
+func (t *Client) GetLocalAttestationPubKey() (crypto.PublicKey, error) {
 	if t.lak == nil {
-		return nil
+		return nil, fmt.Errorf("lak is not yet initialized")
 	}
-	return t.lak.PublicKey()
+	return t.eccPublicKey(t.lak)
```
hexfusion, line 145: @fzdarsky was worried about ecc support with fips can we return either or?n

kkyrazis: Do you have any objections to me doing this in a follow up PR?

hexfusion: no

PR #1452, opened 2025-07-29
NO-ISSUE: Fix alertmanager-proxy deployment

avishayt:
1. Make ca.crt optional
2. Move it to the external namespace (it was failing on reaching keycloak, but the proxy should be accessible to users so it should be in the external namespace)

## Summary by CodeRabbit

* **Refactor**
  * Updated resource definitions to consistently use the Helm release namespace, removing fallback to internal namespace values.
  * Improved handling of certificate fields by conditionally including or setting empty values as appropriate.
  * Marked the certificate key in secret volumes as optional for more flexible deployments.
  * Updated environment variable for service discovery to use a fully qualified domain name format.

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-alertmanager-proxy-deployment.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-alertmanager-proxy-deployment.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     flightctl.service: flightctl-alertmanager-proxy
   name: flightctl-alertmanager-proxy
-  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+  namespace: {{ .Release.Namespace }}
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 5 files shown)


celdrake:
@avishayt Should we combine our two PRs into 1?
https://github.com/flightctl/flightctl/pull/1454 requires the changes here, and fixes the alert-manager proxy for ACM.

deploy/helm/flightctl/templates/flightctl-alertmanager-proxy-deployment.yaml
```diff
@@ -95,6 +95,7 @@ spec:
             items:
               - key: ca.crt
                 path: ca.crt
+                optional: true
```
celdrake, line 98: There's a warning when installing the helm chart,
```I0730 11:29:27.365397   17955 warnings.go:110] "Warning: unknown field \"spec.template.spec.volumes[0].secret.items[0].optional\""```

Seems like `optional` should be outside `items`

PR #1463, opened 2025-07-30
EDM-1875: Fix K8s Secrets Test in OpenShift Test Environments

keitwb:
In those environments, flightctl can be deployed in a single namespace and we must give the flightctl-worker access to the secrets to be able to deploys apps that use them.

Also adding the test label.

## Summary by CodeRabbit

* **Chores**
  * Updated configuration to support environments using a single namespace for all components.
* **Tests**
  * Added a label to an existing end-to-end test case for improved identification.


siserafin (approved):
LGTM

PR #1467, opened 2025-07-30
EDM-1756: Add ability to clear hierarchy

kkyrazis:
## Summary by CodeRabbit

* **New Features**
  * Added the ability to fully clear and reset TPM credentials and state, including persistent data removal.
* **Bug Fixes**
  * Improved reliability of credential wiping by ensuring TPM state is cleared, not just in-memory data.
* **Tests**
  * Introduced comprehensive tests to validate TPM clearing, fallback scenarios, and re-initialization after a clear operation.
* **Chores**
  * Updated default TPM key blob filename for persistence from "ldevid-blob.yaml" to "tpm-blob.yaml".

### DIFF
--- a/internal/agent/config/config.go
+++ b/internal/agent/config/config.go
@@ -59,7 +59,7 @@ const (
 	// DefaultTPMDevicePath is the default TPM device path
 	DefaultTPMDevicePath = "/dev/tpm0"
 	// DefaultTPMKeyBlobFile is the default filename for TPM key blob persistence
-	DefaultTPMKeyBlobFile = "ldevid-blob.yaml"
+	DefaultTPMKeyBlobFile = "tpm-blob.yaml"
 	// TestRootDirEnvKey is the environment variable key used to set the file system root when testing.
 	TestRootDirEnvKey = "FLIGHTCTL_TEST_ROOT_DIR"
 )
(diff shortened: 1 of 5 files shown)


internal/agent/config/config.go
```diff
@@ -59,7 +59,7 @@ const (
 	// DefaultTPMDevicePath is the default TPM device path
 	DefaultTPMDevicePath = "/dev/tpm0"
 	// DefaultTPMKeyBlobFile is the default filename for TPM key blob persistence
-	DefaultTPMKeyBlobFile = "ldevid-blob.yaml"
+	DefaultTPMKeyBlobFile = "tpm-blob.yaml"
```
hexfusion, line 62: 👍

internal/agent/identity/tpm.go
```diff
@@ -106,22 +106,26 @@ func (t *tpmProvider) CreateManagementClient(config *base_client.Config, metrics
 	if err != nil {
 		return nil, err
 	}
+	configCopy := config.DeepCopy()
+	if err := configCopy.Flatten(); err != nil {
+		return nil, err
```
hexfusion, line 111: 👀

hexfusion (approved):

internal/agent/identity/tpm.go
kkyrazis: communication with the management client was failing without this enabled in my test setup. Every other flow (file and tpm.gprc) eventually calls this, but I can spend some more time tomorrow cleaning it up. I suspect an issue with populating the CertAuthorityData.

PR #1475, opened 2025-08-03
EDM-1845: signer refactoring - avoid parsing CSR multiple times

sk-ilya:
Refactor signers to avoid parsing CSR multiple times. No functional changes. 
This is a prep for enabling embedding OrgID in certificates (EDM-1845)

## Summary by CodeRabbit

* **New Features**
  * Introduced a flexible abstraction for certificate signing requests, supporting multiple input sources and robust validation.
  * Added helper functions for certificate signing, verification, and PEM encoding/decoding.
  * New error reporting for invalid certificate signing requests.

* **Refactor**
  * Simplified and unified the certificate issuance and signing flow, removing redundant conversions and encoding steps.
  * Updated interfaces and methods to use higher-level abstractions and return typed certificates rather than raw bytes.
  * Centralized error handling and status updates for certificate and enrollment requests.
  * Removed direct dependencies on legacy CSR types and parsing utilities in favor of a unified request abstraction.

* **Bug Fixes**
  * Improved validation of certificate requests and error reporting for invalid or malformed CSRs.

* **Tests**
  * Added tests to ensure round-trip correctness of certificate PEM encoding and parsing.


internal/crypto/signer/pemutil.go
```diff
@@ -0,0 +1,33 @@
+package signer
```
asafbss: Check pkg/crypto

internal/crypto/signer/helpers.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+)
+
+// SignAndEncodeCertificate verifies, then signs the request and returns the certificate in PEM format.
+func SignAndEncodeCertificate(ctx context.Context, ca CA, request SignRequest) ([]byte, error) {
+	cert, err := VerifyAndSign(ctx, ca, request)
```
asafbss: This approach may cause issues.
Some signers rely on the peer certificate for verification. When a request is approved and signed, this typically happens asynchronously, and the context at that stage no longer includes the original peer certificate from the client that submitted the CSR.

See: https://github.com/flightctl/flightctl/blob/9108a7a96073ebe0e73e4557a93de7852da04a28/internal/crypto/signer/signer_device_svc_client.go#L44

internal/util/validation/validation.go
```diff
@@ -315,6 +316,16 @@ func ValidateExpirationSeconds(e *int32) []error {
 	return nil
 }
 
+func ValidateX509CSR(c *x509.CertificateRequest) []error {
```
asafbss: Consider moving this to pkg/crypto since it’s not specific to Flightctl.
Now that the main validation function has been split and ValidateX509CSR has been added, it feels more appropriate to place it there.

internal/crypto/signer/helpers.go
sk-ilya: These helpers just combine the operations that are commonly executed together. There are no functional changes (at least that was the intention), and it is always possible to rollback to the original pattern of
```go
signer := caClient.GetSigner(signerName)

if signer == nil {...}
if err := s.Verify(ctx, req); err != nil {...}

cert, err := s.Sign(ctx, req)
if err != nil {...}
```

Will adding a helper for signing a certificate without verification resolve the issue  (it will currently be unused)? 

I’m also wondering whether we should enforce a rule that we only sign requests that have been validated at some point

internal/util/validation/validation.go
sk-ilya: The actual validation is just a `csr.CheckSignature()` call :) `ValidateX509CSR` is a wrapper from the API domain. 

We could further decouple signers from the API by parsing the CSR and calling `CheckSignature` directly inside `WithCSRValidation`, and revert the change in validation.go as there will be no need to split it anymore.

The signer can return a typed error (ex. `ErrInvalidCSR`) so the API layer will map it to the appropriate error response generated with `field.Invalid`. I think it would make the overall structure more clean.

```go
var ErrInvalidCSR = errors.New("invalid CSR")

func WithCSRValidation(s Signer) Signer {
<...>
		verify: func(ctx context.Context, request SignRequest) error {
			x509CSR := request.X509()
			if err := x509CSR.CheckSignature(); err != nil {
				return errors.Join(ErrInvalidCSR, err) 
			}
			return s.Verify(ctx, request)
		},
		<...>
	}
}
```

internal/crypto/signer/pemutil.go
sk-ilya: It's in use here with `fccrypto.GetPEMBlock(certPEM)`. Maybe it would be better to move both of the `EncodeCertificatePEM` and `ParseCertificatePEM` to `pkg/crypto` instead of keeping the `pemutil.go`?

internal/util/validation/validation.go
sk-ilya: Moved it to `pkg/crypto`

internal/crypto/signer/pemutil.go
sk-ilya: Moved these utils to a new `cert.go` in `pkg/crypto`

internal/crypto/signer/helpers.go
sk-ilya: Resolved, please see internal/service/certificatesigningrequest.go & internal/service/enrollmentrequest.go

sk-ilya:
@asafbss thanks 🙏 but now I have some violent merge conflicts to resolve... @hexfusion your input would be greatly welcomed

sk-ilya:
@coderabbitai summary

sk-ilya:
@kkyrazis  thanks! That's great to know as this was my main concern after #1492

PR #1480, opened 2025-08-04
EDM-1986: Enable Source IP binding for the simulator

kkyrazis:
Updates to the device simulator to make it more responsive:

1. Allow passing in multiple source IPs. The devices will be round robin assigned to the specified IPs. This allows for an increased number of devices to be onboarded (avoids running out of ephemeral ports)
2. Propagates the log level passed in to the device simulator down to the agents themselves to prevent over logging (or enable overlogging). Any log-level specified explicitly in the agent config will override the simulators level.
3. Adds a `max-concurrency` option. This specifies the maximum number of agents that can be creating at a single time. This massively increases the onboarding speed so that instead of a new Agent being added every (sync_interval / 2) devices are onboarded much more quickly.
4. The default number of http idle connections per host is 2. We now allow overriding that as we expect there to be up to (max-concurrency) number of connections active at a time. In reality, it isn't that high, but allowing more than 2 idle connections decreases the number of connections that are torn down and recreated.
5. copies some config files (ip route tables) to the agents so that they aren't slowed down by trying to find a default route.
6. Fixes a few logs in the agent's that don't use the default logger when they could.
7. Enables caching http.Transports across creations of http.Clients
8. Added more documentation.

## Summary by CodeRabbit

* **New Features**
  * Device Simulator: CLI options for source IP distribution, max concurrency, startup jitter, per-agent log level/prefix, TPM linking, certificate handling, and a metrics endpoint.

* **Performance**
  * Improved HTTP connection reuse, pooling, and configurable transport options for higher scalability.

* **Bug Fixes**
  * More resilient, race-tolerant enrollment approval with timeouts, polling, and retries.

* **Refactor**
  * Startup orchestration streamlined with concurrency gating, staggered launches, enrollment orchestration, and improved readiness/console behavior.

* **Documentation**
  * Major overhaul: setup, CLI reference, examples, tuning, monitoring, remote operation, troubleshooting.


internal/client/config.go
```diff
@@ -586,3 +607,48 @@ func resolvePath(path string, baseDir string) string {
 	}
 	return path
 }
+
+func resolveTransport(client *http.Client) *http.Transport {
+	transport, ok := client.Transport.(*http.Transport)
+	if !ok {
```
ori-amizur, line 630: Is it OK to discard the previous transport in case the casting failed ?

internal/agent/identity/tpm.go
```diff
@@ -137,6 +137,10 @@ func (t *tpmProvider) CreateManagementClient(config *base_client.Config, metrics
 		},
 	}
 
+	for _, opt := range configCopy.HTTPOptions {
+		opt(httpClient)
```
ori-amizur: What about file identity (not tpm) ?

cmd/devicesimulator/main.go
```diff
@@ -153,12 +186,22 @@ func main() {
...
-		agent := agents[i]
-		go startAgent(ctx, agent, log, i)
-		go approveAgent(ctx, log, serviceClient, agentsFolders[i], formattedLables)
+	// limit the maximum number of devices that are being approved.
+	sem := semaphore.NewWeighted(int64(*maxConcurrency))
+	for i := range *numDevices {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			break
+		}
+		go func() {
```
ori-amizur: What is the advantage of starting a goroutine to start other goroutines ?

cmd/devicesimulator/main.go
```diff
@@ -233,101 +288,167 @@ func createAgentConfigTemplate(dataDir string, configFile string) *agent_config.
...
-	agentsFolders := make([]string, numDevices)
+func copyAgentFiles(log *logrus.Logger, certDir, agentDir string) {
+	copyDir := func(path string) string {
+		return filepath.Join(agentDir, path)
+	}
+	fileCopyMap := map[string]struct {
+		dir   string
+		files []string
+	}{
+		certDir: {dir: copyDir(agent_config.DefaultConfigDir), files: []string{"ca.crt", "client-enrollment.crt", "client-enrollment.key"}},
```
ori-amizur: Since this is a device simulator, maybe it is better for it not to be aware of "real" network configurations?

internal/agent/identity/tpm.go
kkyrazis: The file identity provider just calls `NewFromConfig` which applies the options already.

internal/client/config.go
kkyrazis: It's probably not great to blindly override a transport we don't understand. I'll adjust these options to allow propagation of errors.

cmd/devicesimulator/main.go
kkyrazis: I wanted to keep the stagger before launching the agent goroutines so that they don't all start at exactly the same time, but time.Sleep or <-time.After block the current routine. To allow multiple agents to start concurrently I need to wait in a separate goroutine. To avoid refactoring the other functions I just went for the simple approach.

cmd/devicesimulator/main.go
kkyrazis: I tried supplying it with a well formatted routing table that contained faked data, but the default route finding logic matches the routing table against network interfaces that it discovers via golangs `net.Interfaces`. So unless the fake table matches interfaces locally, it'll still take 45 seconds to resolve nothing.

As the network info is just used for reporting purposes, I do think it would be preferable to be able to disable collecting the network info, but that is a more involved agent change.

cmd/devicesimulator/main.go
```diff
@@ -233,101 +288,167 @@ func createAgentConfigTemplate(dataDir string, configFile string) *agent_config.
 		log.Fatalf("Error validating config: %v", err)
 	}
 
-	return agentConfigTemplate
+	// if the config does not have an explicit log level we can override it
+	return agentConfigTemplate, tmpConfig.LogLevel == ""
```
ori-amizur: There is no good reason to read the configuration value here and to return a boolean flag which is used to override a structure (agentConfigTemplate) that is created here.  If you decided that you need to override it - then override it here.
BTW, aren't there other values that might be overridden by the configuration file ?

cmd/devicesimulator/main.go
ori-amizur: BTW, agent uses its own reader and writer.  Each of them may have some pre-set prefix as part of the file.
With this you can use it to use some your own fake data.
Is the 45 seconds accumulative or the whole process is delayed only by  this duration?

cmd/devicesimulator/main.go
```diff
@@ -233,101 +288,167 @@ func createAgentConfigTemplate(dataDir string, configFile string) *agent_config.
...
+		// Device management currently requires setting up individual mTLS connections. This makes HTTP connection reuse
+		// effectively impossible across agents. When a device is onboarded, a new http connection must be generated.
+		// To avoid ephemeral port exhaustion, we allow the management client to bind to specific IPs.
+		if len(agentCfg.parsedSourceIPs) > 0 {
+			// Assign source IP if provided (round-robin distribution)
+			sourceIP := agentCfg.parsedSourceIPs[i%len(agentCfg.parsedSourceIPs)]
+			// Create dialer with source IP, using same defaults as current implementation (30s timeout/keepalive)
+			cfg.ManagementService.Config.AddHTTPOptions(baseclient.WithDialer(&net.Dialer{
+				LocalAddr: &net.TCPAddr{IP: sourceIP},
+				Timeout:   30 * time.Second,
```
ori-amizur, line 423: I think these are the defaults? Right ?

kkyrazis: These are the same values as the `http.DefaultTransport`, but they are not the Dialer default values. The default timeout is no timeout, and the default keepalive is 15 seconds.

cmd/devicesimulator/main.go
kkyrazis: > If you decided that you need to override it - then override it here.

:+1: 

> BTW, aren't there other values that might be overridden by the configuration file ?

This is the only one that I'm allowing to be overridden directly via cli arguments arguments (if not explicitly set)

cmd/devicesimulator/main.go
kkyrazis: It's 45 seconds per agent as each agent uses it's reader writer to read from its configured directory. This setup is copying the routing tables from `/proc/net` into each devices directory: (`/home/user/.flightctl/data-****/proc/net`) so that the reader/writers that are setup can find the routing table. The agent then takes the contents of the routing table and compares it against the result of a syscall (which I can't fake) to find the default route

cmd/devicesimulator/main.go
```diff
@@ -236,98 +295,169 @@ func createAgentConfigTemplate(dataDir string, configFile string) *agent_config.
...
+	fileCopyMap := map[string]struct {
+		dir   string
+		files []string
+	}{
+		certDir: {
+			dir:   copyDir(agent_config.DefaultConfigDir),
+			files: []string{"ca.crt", "client-enrollment.crt", "client-enrollment.key"},
+		},
+		// copies the routing tables from /proc/net into the agent's directories so that it will be able to parse
+		// them and select a default route. Currently, it spends 45 seconds attempting to resolve a default route
```
hexfusion: the default route and really all the systemInfo is just cosmetic status why not just disable `system-info: []` during simulation. are we using it?

cmd/devicesimulator/main.go
ori-amizur: I think the change could lead to bursty behavior.  Assuming the time to enroll is fixed and relatively long and we use the default timeout (2 seconds), so in the first 2 seconds all the goroutines will be spawned.  And then we need to wait for them until they are over.
The previous implementation without the goroutine was better.  We can enhance it with the maxConcurrency.  Also the time to wait between launching the goroutines should not depend on `StatusUpdateInterval` but it should be a separate command line parameter.

cmd/devicesimulator/main.go
kkyrazis: You had mentioned that previously and I had tried disabling it that way, but it didn't work. I'll take a look further though

cmd/devicesimulator/main.go
kkyrazis: >  so in the first 2 seconds all the goroutines will be spawned.

This is only true if:
1. maxConcurrency >= count
2. We're actually able to onboard all devices within 2 seconds

Otherwise, the semaphore limits the number of spawned goroutines to 'maxConcurrency'

> The previous implementation without the goroutine was better.

The previous implementation launches a single agent every `StatusUpdateInterval` / 2 (on average). This, in my opinion, is very very slow. 

Agree with StatusUpdateInterval not being used as our jitter interval. I'm assuming someone chose it for simplicity and to align the jitter with the interval that the agent itself will actually be doing work, but I have no issues changing this to a parameter.

cmd/devicesimulator/main.go
kkyrazis: I can spend some time improving this if you think it worthwhile, but what I've found is that while cosmetic, the system info is always collected, and config value simply filters what is displayed. 

`system_info/manager.Status` => `collectDeviceSystemInfo` => `Collect` (gets everything) => `getSystemInfoMap` (filters out what should be displayed)

https://github.com/flightctl/flightctl/blob/main/internal/agent/device/systeminfo/manager.go#L223
https://github.com/flightctl/flightctl/blob/main/internal/agent/device/systeminfo/system_info.go#L255

kkyrazis: Done here: #1526

cmd/devicesimulator/main.go
kkyrazis: I implemented the fix in the agent to not make this required anymore: #1526

cmd/devicesimulator/main.go
```diff
@@ -214,12 +278,28 @@ func startAgent(ctx context.Context, agent *agent.Agent, log *logrus.Logger, age
...
+	}
+
+	err = yaml.Unmarshal(fileBytes, &tmpConfig)
+	if err != nil {
+		log.Fatalf("Error parsing config: %v", err)
+	}
+
+	if tmpConfig.LogLevel == "" {
+		agentConfigTemplate.LogLevel = logLevelOverride
+	}
```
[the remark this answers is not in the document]
kkyrazis: The issue is that NewDefault will always set a value. ParseConfigFile will overwrite the value if one exists, but a method for determining if the value actual exists in the file is needed. I don't want to pollute any real code with this logic as this is a weird usecase and only for the simulator.

ori-amizur:
Maybe we can add options to disable banner printing and disable launching of podman events ?

kkyrazis:
> Maybe we can add options to disable banner printing and disable launching of podman events ?

`podman events` should only launch if there are applications defined in the device's or fleet's spec. I can add another flag to disable it entirely, but I would prefer to avoid adding code to disable that flow if we can just add documentation advising against it.

I'll disable the banner printing.

ori-amizur:
> > Maybe we can add options to disable banner printing and disable launching of podman events ?
> 
> `podman events` should only launch if there are applications defined in the device's or fleet's spec. I can add another flag to disable it entirely, but I would prefer to avoid adding code to disable that flow if we can just add documentation advising against it.

We just have to make sure that there is a way for devicesimulator to not use such spec.  Either by updating the default spec it is using or by some option.

kkyrazis:
> > > Maybe we can add options to disable banner printing and disable launching of podman events ?
> > 
> > 
> > `podman events` should only launch if there are applications defined in the device's or fleet's spec. I can add another flag to disable it entirely, but I would prefer to avoid adding code to disable that flow if we can just add documentation advising against it.
> 
> We just have to make sure that there is a way for devicesimulator to not use such spec. Either by updating the default spec it is using or by some option.

Ok, yeah, since there is no way to prevent someone from updating the spec of devices or fleets from whatever we start with, I'll look into disabling the monitor.

docs/developer/devicesimulator.md
```diff
@@ -27,14 +260,16 @@ To simulate a large number of devices (10,000 for example) a
...
+```sh
     # on remote host
     mkdir -p ~/.flightctl/certs/ ~/.config/flightctl ~/bin
     scp bin/agent/etc/flightctl/certs/* your_username@remotehost:~/.flightctl/certs/
     scp bin/agent/etc/flightctl/config.yaml your_username@remotehost:~/.flightctl/agent.yaml
     scp ~/.config/flightctl/client.yaml your_username@remotehost:~/.config/flightctl/client.yaml
     scp bin/devicesimulator your_username@remotehost:~/bin/
     # on remote host 
-    chmod -R 755 ~/.flightctl/ ~/.config/flightctl bin/devicesimulator
+    chmod -R 755 ~/.flightctl/ ~/.config/flightctl ~/bin/devicesimulator
```
ori-amizur, line 312: Are you sure about this change ?

docs/developer/devicesimulator.md
```diff
@@ -46,14 +281,56 @@ On each remote host run `devicesimulator` with at least two parameters:
     > bin/devicesimulator --count=200 --initial-device-index=200 \# device-00200 to device-00399
     > 
     > bin/devicesimulator --count=200 --initial-device-index=400 \# device-00400 to device-00599
- 
-Per each device a `podman event` will be created to consume its events. These tasks will never finish and will consume a considerable amount of memory. As a workaround run the following script on each host:
 
+When running remotely it is important to consider [connection limits](devicesimulator.md#system-level-connection-limits) when running large scale performance tests.
+
+**Source IP Considerations for Remote Hosts**: When distributing device simulation across multiple remote hosts, each host can benefit from using multiple source IP addresses to avoid ephemeral port exhaustion. This is especially important when running high device counts per host (>10,000 devices). Configure network aliases on each remote host before running the simulator:
```
ori-amizur: It is important to note that we need to verify we do not cause any IP address conflicts with other hosts in the network.
We might want to consider creating a virtual nic (for example bridge) and assigning IP addresses to it.  These IPs should be in a network that doesn't currently exist.

docs/developer/devicesimulator.md
kkyrazis: It probably isn't necessary, but based on the setup in the previous steps everything should be available in the home directory.

internal/agent/device/applications/sim_manager.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/flightctl/flightctl/internal/agent/device/applications/provider"
+	"github.com/flightctl/flightctl/internal/agent/device/dependency"
+	"github.com/flightctl/flightctl/internal/agent/device/status"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/samber/lo"
+)
+
+var _ Manager = (*simManager)(nil)
+
+type simManager struct {
```
hexfusion: the agent should not care or reason about simulation. please reconsider this approach. we have mocks if we wanted to have the managers do no-ops. but then I wonder why use the agent at all.

hexfusion (dismissed):

internal/agent/device/applications/sim_manager.go
kkyrazis: @ori-amizur are you ok with pushing this off in favor of the ticket  https://issues.redhat.com/browse/EDM-1829 where I can probably come up with something better.

kkyrazis: I can leave it documented as "DO NOT DO THIS" for now while we work through it.

ori-amizur: I am OK of delaying it for now.

kkyrazis: Ok, I've rolled back the changes.

cmd/devicesimulator/main.go
```diff
@@ -153,13 +189,33 @@ func main() {
...
+		formattedLabels: formattedLables,
+		sem:             sem,
+		jitterDuration:  jitterDuration,
+	}
+	for i := range *numDevices {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			break
+		}
+		go launchAgent(ctx, i, launchParams)
 	}
```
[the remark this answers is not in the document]
kkyrazis: That is incorrect and has been since go 1.22

cmd/devicesimulator/main.go
```diff
@@ -236,99 +321,156 @@
...
 		err := os.Setenv(client.TestRootDirEnvKey, agentDir)
 		if err != nil {
-			log.Fatalf("Error setting environment variable: %v", err)
-		}
-		for _, filename := range []string{"ca.crt", "client-enrollment.crt", "client-enrollment.key"} {
-			if err := copyFile(filepath.Join(certDir, filename), filepath.Join(agentDir, agent_config.DefaultConfigDir, filename)); err != nil {
-				log.Fatalf("copying %s: %v", filename, err)
-			}
+			logger.Fatalf("Error setting environment variable: %v", err)
 		}
```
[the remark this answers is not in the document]
kkyrazis: That is not correct. It exists there.

ori-amizur (approved):

PR #1493, opened 2025-08-05
EDM-1702: Organization aware periodic checker

DakCrowder:
Modifies the periodic checker from a thread for each task to a publisher/consumer pattern.  Tasks are scoped by organization, and executed by dedicated consumers.  The tasks scheduling and consumers are completely isolated within the periodic service at this time, and go channels are used to publish / consume tasks.

Architecture Overview:
1. Task Definitions (task_definitions.go):
  - Defines periodic tasks types and intervals
  - Tasks have a common wrapper and injects shared values into context
2. Publisher (publisher.go)
  - Runs two goroutines
    - Org Sync (on a 5 minute ticker): Polls the service to get an updated list of organizations.
    - Task Publishing (execution based on next task runtime): Publishes tasks for execution.  Publish fails are retried with an increasing backoff. 
  - Tasks are scheduled by previous _publish_ time, not execution.  That is, if a task is defined with a 5 minute interval but takes 2 minutes to run, we will try to enqueue the task every 5 minutes
3. Consumer (consumer.go)
  - Runs (5 by default) consumers
  - Adds organization ID to context before task execution (scoping each task)
  - Minimal error handling or monitoring, basically we call .Execute() on a given task and will recover from panics but does not have any retry logic or additional config at this time

High Level Execution Flow of the publish / consume loop:
  1. Publisher scheduling loop calculates next run time from the next task to run
  2. Publisher wakes up at that time and attempts to publish all ready tasks at that time
  3. Consumers pick off tasks from the channel and execute

## Summary by CodeRabbit

* **New Features**
  * Centralized publisher–consumer for periodic background tasks with buffered channel, configurable consumer count (Helm default 5), per‑organization scheduling (staggered), retries/backoff, and bounded graceful shutdown.
  * New periodic task types with pluggable executors and per‑organization dispatch.

* **Bug Fixes**
  * Improved propagation and validation of organization IDs across periodic workflows and reconcilers.

* **Tests**
  * Extensive unit and integration tests covering scheduling, publishing, consumers, channel behavior, retries/backoff, panic recovery, and multi‑org scenarios.

### DIFF
--- a/internal/periodic_checker/consumer.go
+++ b/internal/periodic_checker/consumer.go
@@ -2,6 +2,7 @@ package periodic
 
 import (
 	"context"
+	"fmt"
 	"runtime/debug"
 	"sync"
 
@@ -13,7 +14,7 @@ const (
 	DefaultConsumerCount = 5
 )
 
-func executeWithRecover(executor PeriodicTaskExecutor, ctx context.Context, log logrus.FieldLogger, taskType PeriodicTaskType, orgID uuid.UUID) {
+func executeWithRecover(ctx context.Context, executor PeriodicTaskExecutor, log logrus.FieldLogger, taskType PeriodicTaskType, orgID uuid.UUID) {
 	defer func() {
 		if r := recover(); r != nil {
 			log.WithFields(logrus.Fields{
@@ -39,7 +40,7 @@ func (c *PeriodicTaskConsumer) processTask(ctx context.Context, reference Period
 		return
 	}
 
-	executeWithRecover(executor, ctx, c.log, reference.Type, reference.OrgID)
+	executeWithRecover(ctx, executor, c.log, reference.Type, reference.OrgID)
 }
 
 type PeriodicTaskConsumer struct {
@@ -57,7 +58,17 @@ type PeriodicTaskConsumerConfig struct {
 	ConsumerCount  int
 }
 
-func NewPeriodicTaskConsumer(config PeriodicTaskConsumerConfig) *PeriodicTaskConsumer {
+func NewPeriodicTaskConsumer(config PeriodicTaskConsumerConfig) (*PeriodicTaskConsumer, error) {
+	if config.ChannelManager == nil {
+		return nil, fmt.Errorf("channel manager is required")
+	}
+	if config.Log == nil {
+		return nil, fmt.Errorf("log is required")
+	}
+	if config.Executors == nil {
+		return nil, fmt.Errorf("executors are required")
+	}
+
 	if config.ConsumerCount <= 0 {
 		config.ConsumerCount = DefaultConsumerCount
 	}
@@ -67,7 +78,7 @@ func NewPeriodicTaskConsumer(config PeriodicTaskConsumerConfig) *PeriodicTaskCon
 		log:            config.Log,
 		executors:      config.Executors,
 		consumerCount:  config.ConsumerCount,
-	}
+	}, nil
 }
 
 // runConsumer runs a single consumer goroutine
@@ -91,7 +102,9 @@ func (c *PeriodicTaskConsumer) runConsumer(ctx context.Context, consumerID int) 
 	}
 }
 
-func (c *PeriodicTaskConsumer) Start(ctx context.Context) {
+// Run spins up the consumer goroutines.
+// It blocks until the context is done.
+func (c *PeriodicTaskConsumer) Run(ctx context.Context) {
 	// Start all consumer goroutines
 	for i := 0; i < c.consumerCount; i++ {
 		c.wg.Add(1)
--- a/internal/periodic_checker/consumer_test.go
+++ b/internal/periodic_checker/consumer_test.go
@@ -18,7 +18,6 @@ type mockPeriodicTaskExecutor struct {
 }
 
 type executeCallArgs struct {
-	ctx   context.Context
 	log   logrus.FieldLogger
 	orgID uuid.UUID
 }
@@ -27,7 +26,7 @@ func (m *mockPeriodicTaskExecutor) Execute(ctx context.Context, log logrus.Field
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.executeCallCount++
-	m.executeCallArgs = append(m.executeCallArgs, executeCallArgs{ctx: ctx, log: log, orgID: orgID})
+	m.executeCallArgs = append(m.executeCallArgs, executeCallArgs{log: log, orgID: orgID})
 }
 
 func (m *mockPeriodicTaskExecutor) GetExecuteCallCount() int {
@@ -106,18 +105,20 @@ func TestConsumer_processTask_Success(t *testing.T) {
 			taskRef := createTaskReference(tt.taskType)
 
 			channelManagerConfig := ChannelManagerConfig{Log: log}
-			channelManager := NewChannelManager(channelManagerConfig)
+			channelManager, err := NewChannelManager(channelManagerConfig)
+			require.NoError(t, err)
 			defer channelManager.Close()
 
-			consumer := NewPeriodicTaskConsumer(PeriodicTaskConsumerConfig{
+			consumer, err := NewPeriodicTaskConsumer(PeriodicTaskConsumerConfig{
 				ChannelManager: channelManager,
 				Log:            log,
 				Executors:      executors,
 			})
+			require.NoError(t, err)
 
-			go consumer.Start(ctx)
+			go consumer.Run(ctx)
 
-			err := channelManager.PublishTask(ctx, taskRef)
+			err = channelManager.PublishTask(ctx, taskRef)
 			require.NoError(t, err)
 
 			require.Eventually(t, func() bool {
@@ -146,16 +147,18 @@ func TestConsumer_processTask_MultipleTaskTypes(t *testing.T) {
 	}
 
 	channelManagerConfig := ChannelManagerConfig{Log: log}
-	channelManager := NewChannelManager(channelManagerConfig)
+	channelManager, err := NewChannelManager(channelManagerConfig)
+	require.NoError(t, err)
 	defer channelManager.Close()
 
-	consumer := NewPeriodicTaskConsumer(PeriodicTaskConsumerConfig{
+	consumer, err := NewPeriodicTaskConsumer(PeriodicTaskConsumerConfig{
 		ChannelManager: channelManager,
 		Log:            log,
 		Executors:      executors,
 	})
+	require.NoError(t, err)
 
-	go consumer.Start(ctx)
+	go consumer.Run(ctx)
 
 	// Execute each task type
 	for _, taskType := range testCases {
@@ -206,19 +209,21 @@ func TestConsumer_processTask_MultipleTasks(t *testing.T) {
 	log := logrus.New()
 
 	channelManagerConfig := ChannelManagerConfig{Log: log}
-	channelManager := NewChannelManager(channelManagerConfig)
+	channelManager, err := NewChannelManager(channelManagerConfig)
+	require.NoError(t, err)
 	defer channelManager.Close()
 
-	consumer := NewPeriodicTaskConsumer(PeriodicTaskConsumerConfig{
+	consumer, err := NewPeriodicTaskConsumer(PeriodicTaskConsumerConfig{
 		ChannelManager: channelManager,
 		Log:            log,
 		Executors:      executors,
 	})
+	require.NoError(t, err)
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	go consumer.Start(ctx)
+	go consumer.Run(ctx)
 
 	// Publish multiple tasks
 	numTasks := 10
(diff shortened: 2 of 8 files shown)


DakCrowder:
@coderabbitai review

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,126 @@
...
+	}).Info("Consuming task")
+
+	executor, exists := c.executors[reference.Type]
+	if !exists {
+		c.log.Errorf("no executor found for task type %s", reference.Type)
+		return
+	}
+
+	// Add the orgID to the context
+	ctx = util.WithOrganizationID(ctx, reference.OrgID)
```
keitwb: Using context to pass the org id to the executors seems a bit sketchy, since org id is a prominent, domain-specific piece of information for the executor, and not just a cross-cutting concern here.

Is it not feasible to just extend the `executor.Execute` method to support passing the org id explicitly?

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	c.log.Infof("Starting periodic task consumer %d", consumerID)
+
+	for {
+		select {
+		case <-c.ctx.Done():
+			c.log.Infof("Consumer %d stopped", consumerID)
+			return
+		default:
+			taskRef, ok := c.channelManager.ConsumeTask(c.ctx)
```
keitwb: A common pattern in Go is to have a type return a consume-only channel directly from a method (or it can just be an exposed field) so that you can use it in the normal for-select loop without having to resort to `default` like this.

```
for {
  select {
    case <-c.ctx.Done():
      ...
    case taskRef, ok := <-c.channelManager.Tasks():
      ...
  }
}
```
This makes it easiser to later add more select cases as well.

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,126 @@
...
+				// Channel closed or context cancelled
+				c.log.Infof("Consumer %d stopped - channel closed", consumerID)
+				return
+			}
+			c.processTask(c.ctx, taskRef)
+		}
+	}
+}
+
+func (c *PeriodicTaskConsumer) Start(ctx context.Context) error {
```
keitwb: If you are going to pass in `ctx` here then you can just let that context control the shutdown of this instance instead of having a separate `Stop` method. See https://go.dev/blog/context-and-structs

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,207 @@
...
+				taskLastPublish[taskType] = time.Unix(0, 0)
+			}
+			p.orgTasksMetadata[orgID] = &OrgTaskMetadata{
+				TaskLastPublish: taskLastPublish,
+			}
+		}
+	}
+}
+
+func (p *PeriodicTaskPublisher) Stop() {
```
keitwb: This should be called `WaitForStopped` or something like that as it doesn't actually stop anything.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,207 @@
...
+		p.syncOrganizations(ctx)
+
+		taskTicker := time.NewTicker(p.taskTickerInterval)
+		defer taskTicker.Stop()
+		orgTicker := time.NewTicker(p.orgTickerInterval)
+		defer orgTicker.Stop()
+
+		for {
+			select {
+			case <-taskTicker.C:
```
keitwb: I guess the # of tasks is pretty small and there probably won't be crazy amounts of orgs in a deployment, but the natural way to do scheduling is with a minheap which could avoid the nested looping and also provide more immediate scheduling of the tasks if timeliness is important. As long as CPU consumption with a large # of orgs isn't very high what you have is fine if we are ok with a potential delay of 5 seconds for task scheduling.

internal/periodic_checker/server.go
```diff
@@ -78,48 +54,54 @@ func (s *Server) Run(ctx context.Context) error {
...
+		OrgService:     serviceHandler,
+		TasksMetadata:  periodicTasks,
+		ChannelManager: channelManager,
+	}
+	periodicTaskPublisher, err := NewPeriodicTaskPublisher(publisherConfig)
+	if err != nil {
+		return err
+	}
+	periodicTaskPublisher.Start(ctx)
+	defer periodicTaskPublisher.Stop()
```
keitwb: This I think will just hang since you are dependent on the context to be cancelled for the wait group in this method to be done, but cancel won't be called until this defer returns (defers are processed in reverse order).

internal/rollout/device_selection/reconciler.go
```diff
@@ -163,9 +163,11 @@ func (r *reconciler) reconcileFleet(ctx context.Context, orgId uuid.UUID, fleet
 }
 
 func (r *reconciler) Reconcile(ctx context.Context) {
-	// Get all relevant fleets
-	orgId := store.NullOrgId
-
+	orgId, ok := util.GetOrgIdFromContext(ctx)
+	if !ok {
+		r.log.Error("No organization ID found in context")
```
keitwb, line 192: If this is non-optional, why isn't it just passed in as a function arg?

internal/periodic_checker/consumer.go
DakCrowder: Yeah i think that sounds reasonable.

internal/rollout/device_selection/reconciler.go
DakCrowder: The way things are currently built the job would then have to add the org id to context as the underlying service handler implementation [(example here](https://github.com/flightctl/flightctl/blob/main/internal/service/fleet.go#L16C11-L16C30) is reliant on it being set.  The other periodic tasks are (right now) implicitly reliant on the same behavior, this job is just written differently and appends org id to many of the logs and makes use of it in that way.

It does raise an interesting point though, maybe the transport layer(s) over the apis could handle pulling the org id out of context and the service handler methods updated to expect an explicit org id.  Then these tasks could be updated to be directly passed the org ids.

keitwb: Yeah I think the http handler should pull it out immediately and stick it either in the API resource metadata or passed as an explicit arg to the handlers.  That way seems much clearer.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,207 @@
...
+	return nil
+}
+
+func (p *PeriodicTaskPublisher) publishTask(ctx context.Context, taskType PeriodicTaskType, orgID uuid.UUID) error {
+	taskReference := PeriodicTaskReference{
+		Type:  taskType,
+		OrgID: orgID,
+	}
+
+	if err := p.channelManager.PublishTask(ctx, taskReference); err != nil {
```
kkyrazis: Is there any concern with publishing blocking until consumers are available to pick up a task?

internal/periodic_checker/publisher.go
DakCrowder: This is a good question and definitely  worth considering.  I would ask if it is worth additional work to implement (not that its a massive lift) but would it be better to again take a look at something like [redis sorted sets](https://redis.io/docs/latest/develop/data-types/sorted-sets/) if task persistence is more important?

internal/rollout/device_selection/reconciler.go
DakCrowder: Created a new [ticket](https://issues.redhat.com/browse/EDM-1977) to take a look at this

internal/periodic_checker/consumer.go
DakCrowder: Refactored to do so and have more clear shutdown logic - will push tomorrow

internal/periodic_checker/publisher.go
DakCrowder: Yes and no - realistically without retries or any additional consumer management logic we are kinda screwed either way if consumers _actually_ do something really bad like enter loops then we block forever.  Although hitting the channel limit could reasonably happen w/ the current implementation if a few orgs were "discovered" in a sync and a bunch of tasks would then be set to be enqueued.  Would be lessened if we used a minheap as @keitwb mentioned or slightly lest bursty approach or at the very least might be good to have a timeout on publish attempt so there is an indication _something_ is up.

internal/periodic_checker/publisher.go
DakCrowder: Approach has been modified to use a minheap w/ a bit more intelligent scheduling logic based upon the expected publish time of the "next task up"

internal/periodic_checker/publisher.go
DakCrowder: Implementation has been modified so we'll fail to publish, but re-try with a configured backoff.  Right now there are no backoff limits or anything like a DLQ but should be better than what it was.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,363 @@
...
+
+	orgList, status := p.orgService.ListOrganizations(ctx)
+	if status.Code < 200 || status.Code >= 300 {
+		p.log.Errorf("Failed to list organizations: %v", status)
+		return
+	}
+
+	currentOrgs := make(map[uuid.UUID]struct{})
+	for _, org := range orgList.Items {
+		orgID, err := uuid.Parse(*org.Metadata.Name)
```
kkyrazis: Can we use lo.FromPtr here to avoid potential panics?

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,384 @@
...
+	taskTimer := time.NewTimer(DefaultTaskTickerInterval)
+	defer taskTimer.Stop()
+
+	for {
+		p.heapMu.Lock()
+		nextTask := p.taskHeap.Peek()
+		p.heapMu.Unlock()
+
+		if nextTask == nil {
+			taskTimer = time.NewTimer(DefaultTaskTickerInterval)
```
kkyrazis: Do we need the default ticker interval now that we have a wake up mechanism? Theoretically you either need to sleep until a known event time, or if there are no events, until someone (the org syncing) adds events.

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	DefaultConsumerCount = 5
+)
+
+func executeWithRecover(executor PeriodicTaskExecutor, ctx context.Context, log logrus.FieldLogger, taskType PeriodicTaskType, orgID uuid.UUID) {
```
hexfusion: minor: can we please keep ctx as first param

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,107 @@
...
+	log            logrus.FieldLogger
+	executors      map[PeriodicTaskType]PeriodicTaskExecutor
+	consumerCount  int
+	wg             sync.WaitGroup
+}
+
+type PeriodicTaskConsumerConfig struct {
+	ChannelManager *ChannelManager
+	Log            logrus.FieldLogger
+	Executors      map[PeriodicTaskType]PeriodicTaskExecutor
```
hexfusion, line 57: might be missing it but how do we ensure this map is initialized or is nil vaild.

internal/periodic_checker/consumer.go
```diff
@@ -0,0 +1,107 @@
...
+			if !ok {
+				c.log.Infof("Task channel closed – consumer %d stopping", consumerID)
+				return
+			}
+			c.processTask(ctx, taskRef)
+		}
+	}
+}
+
+func (c *PeriodicTaskConsumer) Start(ctx context.Context) {
```
hexfusion: minor: `s/Start/Run`?

internal/periodic_checker/consumer_test.go
```diff
@@ -0,0 +1,339 @@
...
+)
+
+type mockPeriodicTaskExecutor struct {
+	mu               sync.Mutex
+	executeCallCount int
+	executeCallArgs  []executeCallArgs
+}
+
+type executeCallArgs struct {
+	ctx   context.Context
```
hexfusion: why is the context in the struct here it should be passed not stored, almost exclusively. I know that sometimes in a server context, it can be necessary. but I would like not to continue this pattern unless absolutely necessary even in a test.

internal/periodic_checker/publisher.go
```diff
@@ -0,0 +1,397 @@
...
+	channelManager TaskChannelManager
+	taskBackoff    poll.Config
+
+	// Heap for scheduled tasks
+	taskHeap *TaskHeap
+	heapMu   sync.Mutex
+
+	// Organizations tracking
+	organizations map[uuid.UUID]struct{}
+	orgsMu        sync.Mutex
```
hexfusion: question: do we absolutly need seperate locks?

internal/periodic_checker/consumer.go
DakCrowder: Added some additional validations to avoid nil refs

internal/periodic_checker/publisher.go
DakCrowder: Hmmmm don't think its a need, consolidated the two mutexes into a single one

PR #1495, opened 2025-08-06
labels endpoint with field-selector=metadata.labels returns error 400

noga-magen:
The [current fix](https://github.com/flightctl/flightctl/commit/ff0907707f28a7cf374851dd2da1bcd576485854) is causing a fail at this endpoint: 
/api/v1/labels?kind=Device&fieldSelector=metadata.labels.keyOrValue=x'

In order to solve the problem the problem the current fix is trying to solve I'll create another ticket. 

## Summary by CodeRabbit

* **Refactor**
  * Improved the logic for resolving selector fields to prioritize visible fields before hidden fields. This change may affect which fields are selected in certain scenarios.


internal/store/selector/field.go
```diff
@@ -594,10 +594,10 @@ func (fs *FieldSelector) queryField(args ...string) (*sql.FunctionResult, error)
 // resolveSelectorField attempts to resolve a field using both visible and hidden selectors.
 func (fs *FieldSelector) resolveSelectorField(resolver Resolver, key string) ([]*SelectorField, error) {
 	// First, try resolving as a hidden selector
```
asafbss: "First, try resolving as a hidden selector" - typo

Please revert the exact commit:
https://github.com/flightctl/flightctl/commit/ff0907707f28a7cf374851dd2da1bcd576485854

asafbss (approved):

PR #1507, opened 2025-08-08
NO-ISSUE: reverting EDM-1871

noga-magen:
Since the commit https://github.com/flightctl/flightctl/commit/1a21ea8e7b73dc2c0945d856758e5f195c6add68 had caused this [bug](https://issues.redhat.com/browse/EDM-1935)
I'm reverting it . I'll contact Shaun (who opened the og problem) and together we will think of a solution.


hexfusion (commented):
Can you provide some content on why this should be backported please.

amir-yogev-gh (approved):

PR #1523, opened 2025-08-11
EDM-1987: Disallow label updates for owned fleets

avishayt:
Fleets are owned by ResourceSyncs, which are responsible for syncing the entire fleet to Flight Control, including its labels. Therefore, users should not be allowed to modify the labels of an owned fleet.

Also, fix a broken link in the docs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevent updating a Fleet’s spec or labels when the resource is owned, returning a clear error instead of allowing unintended changes.
  - Improved ownership checks to consistently block prohibited updates from all sources.

- **Documentation**
  - Updated the Alerts and Monitoring link to point to the correct destination.

- **Tests**
  - Added integration tests to verify that spec and label updates are rejected for owned Fleet resources and that the resource remains unchanged.

### DIFF
--- a/internal/store/generic.go
+++ b/internal/store/generic.go
@@ -3,6 +3,7 @@ package store
 import (
 	"context"
 	"errors"
+	"reflect"
 	"time"
 
 	api "github.com/flightctl/flightctl/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


internal/store/generic.go
```diff
@@ -186,20 +187,27 @@ func (s *GenericStore[P, M, A, AL]) createResource(ctx context.Context, resource
...
+			// Don't let the user update the spec if it has an owner
+			return false, flterrors.ErrUpdatingResourceWithOwnerNotAllowed
 		}
 
 		// Update the generation if the spec was updated
 		resource.SetGeneration(lo.ToPtr(lo.FromPtr(existing.GetGeneration()) + 1))
 	}
 
+	// Don't let the user update a fleet's labels if it has an owner
+	if fromAPI && hasOwner && resource.GetKind() == api.FleetKind {
```
sk-ilya, line 204: If this is a fleet-specific check, maybe it's worth keeping it in the concrete `store/fleet.go` or at the service level?

PR #1526, opened 2025-08-11
EDM-1992: Agent SystemInfo smart collection

kkyrazis:
Changes System Info collection to only collect information that it will actually use. Currently, all system info is collected on startup and then, depending on the `system-info` config property, only partial values are displayed. This is particularly noticeable in instances where the network info is being collected as that can take upwards of 45 seconds to resolve. Now, that will only be collected in instances where it is specifically requested.

## Summary by CodeRabbit

- **New Features**
  - Modular, configurable system information collection with selectable categories (CPU, GPU, Memory, Network, BIOS, System, Kernel, Distribution).
  - Default collectors enabled by default; system-info command now uses the full default collection set.

- **Bug Fixes**
  - Prevented nil system-info lists by initializing them to an empty set.
  - Refresh of cached system info is now conditional — only invalidated when new collectors are added.
  - Boot JSON fields renamed: bootTime → time, bootID → id.

- **Tests**
  - Added coverage for collection-option mapping and collector behaviors; updated collection-related tests.

### DIFF
--- a/cmd/devicesimulator/main.go
+++ b/cmd/devicesimulator/main.go
@@ -296,6 +296,7 @@ func createAgents(log *logrus.Logger, numDevices int, initialDeviceIndex int, ag
 			Server:               agentConfigTemplate.ManagementService.Config.Service.Server,
 			CertificateAuthority: filepath.Join(cfg.ConfigDir, agent_config.CacertFile),
 		}
+		cfg.SystemInfo = []string{}
 
 		cfg.SetEnrollmentMetricsCallback(rpcMetricsCallback)
 		if err := cfg.Complete(); err != nil {
--- a/cmd/flightctl-agent/main.go
+++ b/cmd/flightctl-agent/main.go
@@ -122,7 +122,7 @@ func (s *systemInfoCmd) Execute() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 	exec := &executer.CommonExecuter{}
-	info, err := systeminfo.Collect(ctx, s.log, exec, reader, nil, s.hardwareMapPath, systeminfo.WithAllDefault(), systeminfo.WithAllCustom())
+	info, err := systeminfo.Collect(ctx, s.log, exec, reader, nil, s.hardwareMapPath, systeminfo.WithAll())
 	if err != nil {
 		s.log.Fatalf("Error collecting system info: %v", err)
 	}
--- a/cmd/flightctl-agent/main.go
+++ b/cmd/flightctl-agent/main.go
@@ -122,7 +122,7 @@ func (s *systemInfoCmd) Execute() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 	exec := &executer.CommonExecuter{}
-	info, err := systeminfo.Collect(ctx, s.log, exec, reader, nil, s.hardwareMapPath, systeminfo.WithAllCustom())
+	info, err := systeminfo.Collect(ctx, s.log, exec, reader, nil, s.hardwareMapPath, systeminfo.WithAllDefault(), systeminfo.WithAllCustom())
 	if err != nil {
 		s.log.Fatalf("Error collecting system info: %v", err)
 	}
--- a/internal/agent/device/systeminfo/system_info.go
+++ b/internal/agent/device/systeminfo/system_info.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"runtime"
 	"sort"
@@ -39,6 +40,40 @@ const (
 	SystemFileName = "system.json"
 	// HardwareMapFileName is the name of the file where the hardware map is stored.
 	HardwareMapFileName = "hardware-map.yaml"
+
+	// Info key constants for system information collection
+	infoKeyHostname     = "hostname"
+	infoKeyArchitecture = "architecture"
+	infoKeyKernel       = "kernel"
+
+	// CPU info keys
+	infoKeyCPUCores      = "cpuCores"
+	infoKeyCPUProcessors = "cpuProcessors"
+	infoKeyCPUModel      = "cpuModel"
+
+	// GPU info keys
+	infoKeyGPU = "gpu"
+
+	// Memory info keys
+	infoKeyMemoryTotalKb = "memoryTotalKb"
+
+	// Network info keys
+	infoKeyNetInterfaceDefault = "netInterfaceDefault"
+	infoKeyNetIPDefault        = "netIpDefault"
+	infoKeyNetMACDefault       = "netMacDefault"
+
+	// BIOS info keys
+	infoKeyBIOSVendor  = "biosVendor"
+	infoKeyBIOSVersion = "biosVersion"
+
+	// System info keys
+	infoKeyProductName   = "productName"
+	infoKeyProductSerial = "productSerial"
+	infoKeyProductUUID   = "productUuid"
+
+	// Distribution info keys
+	infoKeyDistroName    = "distroName"
+	infoKeyDistroVersion = "distroVersion"
 )
 
 type Manager interface {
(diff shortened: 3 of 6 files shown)


cmd/devicesimulator/main.go
```diff
@@ -296,6 +296,7 @@ func createAgents(log *logrus.Logger, numDevices int, initialDeviceIndex int, ag
 			Server:               agentConfigTemplate.ManagementService.Config.Service.Server,
 			CertificateAuthority: filepath.Join(cfg.ConfigDir, agent_config.CacertFile),
 		}
+		cfg.SystemInfo = []string{}
```
hexfusion, line 299: :+1:

cmd/flightctl-agent/main.go
```diff
@@ -122,7 +122,7 @@ func (s *systemInfoCmd) Execute() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 	exec := &executer.CommonExecuter{}
-	info, err := systeminfo.Collect(ctx, s.log, exec, reader, nil, s.hardwareMapPath, systeminfo.WithAllCustom())
+	info, err := systeminfo.Collect(ctx, s.log, exec, reader, nil, s.hardwareMapPath, systeminfo.WithAllDefault(), systeminfo.WithAllCustom())
```
hexfusion: we should make sure the documentation reflects this. my main questions is by design this comand should return all internal system-info collectors vs default list or is default all.

internal/agent/device/systeminfo/system_info.go
```diff
@@ -225,7 +260,31 @@ type SystemInfo struct {
 	SKU          string `json:"sku,omitempty"`
 }
 
-type InfoMap map[string]string
+type infoMap map[string]string
+
+type collectContext struct {
+	ctx                 context.Context
```
hexfusion: i get what you're trying to do here, but it's more idiomatic in go to standardize the function signature. context is almost always meant to be passed, except in fairly rare situations.

internal/agent/device/systeminfo/system_info.go
```diff
@@ -238,145 +297,281 @@ func (b *Boot) IsEmpty() bool {
...
+
+// WithAllCustom sets the allCustom option to true enabling all custom scripts to be collected.
+func WithAllCustom() CollectOpt {
+	return func(cfg *collectCfg) {
+		cfg.collectAllCustom = true
+	}
+}
+
+// WithAllDefault enables collection of all default hardware information.
+func WithAllDefault() CollectOpt {
```
hexfusion: is there another default set in the agent config? lets just make sure default set is a single definition.

internal/agent/device/systeminfo/system_info.go
```diff
@@ -238,145 +297,281 @@ func (b *Boot) IsEmpty() bool {
...
+		cfg.addCollector(collectorBIOS, collectBIOSFunc)
+		cfg.addCollector(collectorSystem, collectSystemFunc)
+		cfg.addCollector(collectorKernel, collectKernelFunc)
+		cfg.addCollector(collectorDistribution, collectDistributionFunc)
+	}
+}
+
+// WithCPUCollection enables CPU information collection.
+func WithCPUCollection() CollectOpt {
+	return func(cfg *collectCfg) { cfg.addCollector(collectorCPU, collectCPUFunc) }
```
hexfusion: this pattern repeats a fair amount since this is more internal facing I wonder if this would good enough to just do

```
func WithCollector(key string, collectFn whateverTheObjectIs) CollectOpt {
	return func(cfg *collectCfg) { cfg.addCollector(key, collectFn) }
}
```

internal/agent/device/systeminfo/system_info.go
```diff
@@ -400,10 +595,10 @@ func Collect(ctx context.Context, log *log.PrefixLogger, exec executer.Executer,
 
 // SupportedInfoKeys is a map of supported info keys to their corresponding functions.
 var SupportedInfoKeys = map[string]func(info *Info) string{
-	"hostname":     func(i *Info) string { return i.Hostname },
-	"architecture": func(i *Info) string { return i.Architecture },
-	"kernel":       func(i *Info) string { return i.Kernel },
-	"distroName": func(i *Info) string {
+	infoKeyHostname:     func(i *Info) string { return i.Hostname },
```
hexfusion: infoKey sort of stutters here just curious if its needed

internal/agent/device/systeminfo/system_info.go
```diff
@@ -238,145 +297,281 @@ func (b *Boot) IsEmpty() bool {
...
+		case infoKeyNetInterfaceDefault, infoKeyNetIPDefault, infoKeyNetMACDefault:
+			opts = append(opts, WithNetworkCollection())
+		case infoKeyBIOSVendor, infoKeyBIOSVersion:
+			opts = append(opts, WithBIOSCollection())
+		case infoKeyProductName, infoKeyProductSerial, infoKeyProductUUID:
+			opts = append(opts, WithSystemCollection())
+		case infoKeyKernel:
+			opts = append(opts, WithKernelCollection())
+		case infoKeyDistroName, infoKeyDistroVersion:
+			opts = append(opts, WithDistributionCollection())
```
hexfusion: might be good to catch an error if we have a key with no match

internal/agent/device/systeminfo/system_info.go
```diff
@@ -238,145 +297,281 @@ func (b *Boot) IsEmpty() bool {
...
+	memInfo, err := collectMemoryInfo(collectCtx.log, collectCtx.reader)
 	if err != nil {
-		if errors.IsContext(err) {
-			log.Warningf("Context canceled during distribution info collection: %v", err)
-			return info, err
-		}
-		log.Warningf("Failed to get distribution info: %v", err)
-	} else {
-		info.Distribution = distroInfo
+		return fmt.Errorf("memory collector failed: %w", err)
```
hexfusion, line 341: nit: collection is all best effort if one collector fails we should not exit and instead send what we can

internal/agent/device/systeminfo/system_info.go
kkyrazis: The systemInfo.Collect function is exported which is why I leaned this way, but if you'd like I can just expose the WithAll() options and leave the rest internal.

internal/agent/device/systeminfo/system_info.go
kkyrazis: The individual "collectXYZ" return errors, but the caller is behaving the same way as previously. Log the error returned, move on to the next collector.

internal/agent/device/systeminfo/system_info.go
kkyrazis: I'll work out something better.

internal/agent/device/systeminfo/system_info.go
kkyrazis: Yeah, I just didn't like the signature being 50% unused for most of the calls, but I'll make the change.

cmd/flightctl-agent/main.go
kkyrazis: I was trying to differentiate between [All]Default vs [All]Custom, but it'll just combine this into a single `WithAll()` that combines the two

hexfusion: makes sense

internal/agent/device/systeminfo/system_info.go
hexfusion: I see now

internal/agent/device/systeminfo/system_info.go
hexfusion: well if we need an object just dont include the ctx

internal/agent/device/systeminfo/system_info.go
```diff
@@ -238,145 +272,264 @@ func (b *Boot) IsEmpty() bool {
...
+
+// WithAllCustom enables all custom collection.
+func WithAllCustom() CollectOpt {
+	return func(cfg *collectCfg) {
+		cfg.collectAllCustom = true
+	}
+}
+
+// withCollector creates a CollectOpt that adds a specific collector
+func withCollector(cType collectorType, fn collectorFunc) CollectOpt {
```
hexfusion, line 432: :+1:

hexfusion (approved):

PR #1471, opened 2025-07-31
EDM-1845: middleware agent cert OrgID support

sk-ilya:
Update middleware to scope agents' requests based on OrgID certs. CSR verification by signers will be implemented in a separate PR.

## Summary by CodeRabbit

* **New Features**
  * Improved organization ID validation and extraction from API requests, supporting both query parameters and client certificates.
  * Enhanced certificate signing requests to include organization ID information as an ASN.1 extension.
  * Introduced caching for organization ID lookups to improve performance and resource management.
  * Added centralized method to determine effective organization for CLI client configuration.

* **Bug Fixes**
  * Standardized error handling for missing certificate extensions with a dedicated error variable.

* **Refactor**
  * Streamlined middleware for organization ID extraction and validation using modular extractors.
  * Unified organization ID handling across the application with internal parsing utilities replacing external UUID packages.
  * Updated middleware chains to improve organization ID context injection.

* **Tests**
  * Added comprehensive tests for organization ID extraction from requests and certificates, middleware integration, and configuration validation.

PR #1424, opened 2025-07-24
EDM-1846: CLI organizations support

sk-ilya:
# Summary

1. **`flightctl config` sub-commands**

   • `flightctl config current-organization`  
     Prints the organization UUID currently stored in the CLI config file.  

   • `flightctl config set-organization <organization-uuid>`  
     Persists `<organization-uuid>` as the default organization in the user’s config file.  
     – Pass an empty string to unset.  
     – Basic validation rejects malformed UUIDs.

2. **Listing accessible organizations**

   • `flightctl get organizations` (alias: `orgs`)  
     Returns every organization the client is allowed to see.  

3. **Global flag `--org`**

   • Available on every CLI command (wired through `GlobalOptions`).  
   • Precedence order:  
     1. `--org` flag (highest)  
     2. Value stored in the `client.yaml`
     3. None 
   • Enables ad-hoc overrides without touching the persisted config, e.g.:

     ```bash
     flightctl device list --org 00000000-0000-0000-0000-000000000001
     ```

---

## Usage Examples

```bash
# Show current org
$ flightctl config current-organization
00000000-0000-0000-0000-000000000000

# Change default org (persist in client.yaml)
$ flightctl config set-organization 00000000-0000-0000-0000-000000000001
Current organization set to: 00000000-0000-0000-0000-000000000001

# Unset
$ flightctl config set-organization ""
Current organization unset

# List organizations
$ flightctl get orgs
NAME                                 DISPLAY NAME   EXTERNAL ID   OWNER
00000000-0000-0000-0000-000000000000 My Org         <none>        <none>
00000000-0000-0000-0000-000000000001 Other Org      <none>        <none>

# List devices in a different org (does not touch config file)
$ flightctl device list --org 00000000-0000-0000-0000-000000000001
```

---

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to view and set the current organization in the configuration.
  * Introduced support for listing organizations via the CLI.
  * Added a new CLI flag to override the organization for requests.
  * Enabled automatic inclusion of organization ID as a query parameter in API requests.

* **Enhancements**
  * Improved validation and error messages for organization-related operations.
  * Refined output formatting for organization lists in the CLI.
  * Modularized validation logic for resource retrieval commands.
  * Updated client creation to incorporate organization overrides seamlessly.

* **Bug Fixes**
  * Enhanced client creation to support organization overrides and configuration consistency.

* **Tests**
  * Added comprehensive tests for organization-related CLI configuration commands.

PR #1527, opened 2025-08-11
EDM-1845: support embedding OrgID in certificates

sk-ilya:
# Device OrgID Scoping

Each request is scoped to a single organization identified by an `OrgID` carried as an X.509 extension in the device’s peer certificate.  
The service middleware extracts the `OrgID` from the peer certificate and stores it in the request context (#1471); all authorization and data access for that request use that value.

If the certificate does not include an `OrgID`, the request is scoped to the default organization, making the change backward compatible and aligning with the current user scoping logic.

---

## Provisioning & Enrollment

**Bootstrap phase:**  
An operator creates a CSR and may include the OrgID as an X.509 extension. After approval, the server issues a short-lived peer certificate for the device.  
The new behavior is that the signed certificate includes the OrgID extension derived from the operator’s request and request context.

**Enrollment phase:**  
On the device’s first boot, it loads its short-lived bootstrap certificate and uses it as the peer certificate when making the Enrollment Request to the server.  
After the Enrollment Request is approved, the device retrieves the newly issued long-lived device certificate from the server. This certificate includes the OrgID, and all future requests will be scoped based on the OrgID in this certificate.

---

## Signing & Validation Rules

The API signer chain is updated to inject the OrgID extension based on the X.509 certificate signing request in the CSR/ER request body and the request context:

1. If both CSR and context contain OrgID **and they differ** → **fail verification/issuance**.
2. If CSR is **missing** OrgID and context has one → **use context OrgID**.
3. If CSR **has** OrgID → **use it**.
4. If neither has OrgID → **do not include OrgID** in the certificate.

---
# Demo using flightctl CLI

Below is an example of how the process works using the CLI. 
The support for org-scoped requests using CLI is implemented in #1424.
Requesting an enrollment certificate:
```bash
> flightctl certificate request --org=00000000-0000-0000-0000-000000000001 \
    --signer=enrollment \
    --expiration=365d
Creating new ECDSA key pair and writing to "/etc/flightctl/certs/client-enrollment.key".
Submitting certificate signing request "client-enrollment-f9d0355e"... success.
Waiting for certificate to be approved and issued.... success.
```
Inspecting the certificate content:

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 5453664005379253296 (0x4baf4e84a4f6e430)
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: CN=ca
        Validity
            Not Before: Aug 12 10:22:18 2025 GMT
            Not After : Aug 12 10:22:19 2026 GMT
        Subject: CN=client-enrollment-f9d0355e
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment
            X509v3 Extended Key Usage:
                TLS Web Client Authentication
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Authority Key Identifier:
                67:C6:52:51:92:6C:9A:33:5C:80:27:77:D2:98:2C:B1:9E:C2:9F:8C:2C:D1:03:B8:16:E5:A8:9C:2F:CE:DB:0B
            1.3.6.1.4.1.99999.1.2:
                .$00000000-0000-0000-0000-000000000001 <-- OrgID
            1.3.6.1.4.1.99999.1.1:
                ..flightctl.io/enrollment
```
The `1.3.6.1.4.1.99999.1.2` extension contains the OrgID, confirming that the certificate is correctly scoped to the specified organization.

After the device is booted, the Enrollment Request can be listed as a resource belonging to the same organization:
```bash
> flightctl get er --org=00000000-0000-0000-0000-000000000001
NAME                                                    APPROVAL        APPROVER        APPROVED LABELS
01l9bb80uf5himu7mjrec32rnig0sj1bek2ljhp0mvaf062moca0    Approved        unknown

> flightctl approve er/01l9bb80uf5himu7mjrec32rnig0sj1bek2ljhp0mvaf062moca0 --org=00000000-0000-0000-0000-000000000001
```
Once approved, the device belongs to that same organization:

```bash
> flightctl get dev --org=00000000-0000-0000-0000-000000000001
NAME                                                    ALIAS   OWNER   SYSTEM  UPDATED         APPLICATIONS    LAST SEEN
01l9bb80uf5himu7mjrec32rnig0sj1bek2ljhp0mvaf062moca0            <none>  Online  UpToDate        Healthy         27 seconds ago
```

**Note**: A default organization can be set for all flightctl requests to avoid specifying `--org` each time: `flightctl config set-organization 00000000-0000-0000-0000-000000000001`

## Summary by CodeRabbit

* **New Features**
  * Org ID handling added to certificate flows: Org IDs can be read from CSRs or context, embedded when present, and validated for consistency during issuance and verification.

* **Refactor**
  * Signer pipeline reorganized into decorator-style wrappers to centralize extension handling and wiring.

* **Changes**
  * Some issued certificates (device enrollment and certain clients) no longer include the Org ID extension.
  * Extension extraction now returns explicit presence vs absence.

* **Tests**
  * Adds comprehensive signer-chain tests covering issuance, verification, reuse, restrictions, and Org ID scenarios.

### DIFF
--- a/internal/crypto/signer/signer.go
+++ b/internal/crypto/signer/signer.go
@@ -3,10 +3,13 @@ package signer
 import (
 	"context"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/flightctl/flightctl/internal/config/ca"
+	"github.com/flightctl/flightctl/internal/flterrors"
+	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/pkg/crypto"
 )
 
@@ -40,33 +43,43 @@ func NewCASigners(ca CA) *CASigners {
 	ret := &CASigners{
 		ca: ca,
 		signers: map[string]Signer{
-			cfg.ClientBootstrapSignerName: WithSignerNameValidation(
-				WithCertificateReuse(
-					WithCSRValidation(
-						WithSignerNameExtension(NewClientBootstrap, ca),
-					),
-				),
+			cfg.ClientBootstrapSignerName: compose(
+				compose(
+					NewClientBootstrap,
+					WithOrgIDExtension,
+					WithSignerNameExtension,
+				)(ca),
+				WithCSRValidation,
+				WithCertificateReuse,
+				WithSignerNameValidation,
 			),
-			cfg.DeviceEnrollmentSignerName: WithSignerNameValidation(
-				WithCertificateReuse(
-					WithCSRValidation(
-						WithSignerNameExtension(NewSignerDeviceEnrollment, ca),
-					),
-				),
+			cfg.DeviceEnrollmentSignerName: compose(
+				compose(
+					NewSignerDeviceEnrollment,
+					WithOrgIDExtension,
+					WithSignerNameExtension,
+				)(ca),
+				WithCSRValidation,
+				WithCertificateReuse,
+				WithSignerNameValidation,
 			),
-			cfg.DeviceSvcClientSignerName: WithSignerNameValidation(
-				WithCertificateReuse(
-					WithCSRValidation(
-						WithSignerNameExtension(NewSignerDeviceSvcClient, ca),
-					),
-				),
+			cfg.DeviceSvcClientSignerName: compose(
+				compose(
+					NewSignerDeviceSvcClient,
+					WithSignerNameExtension,
+				)(ca),
+				WithCSRValidation,
+				WithCertificateReuse,
+				WithSignerNameValidation,
 			),
-			cfg.ServerSvcSignerName: WithSignerNameValidation(
-				WithCertificateReuse(
-					WithCSRValidation(
-						WithSignerNameExtension(NewSignerServerSvc, ca),
-					),
-				),
+			cfg.ServerSvcSignerName: compose(
+				compose(
+					NewSignerServerSvc,
+					WithSignerNameExtension,
+				)(ca),
+				WithCSRValidation,
+				WithCertificateReuse,
+				WithSignerNameValidation,
 			),
 		},
 	}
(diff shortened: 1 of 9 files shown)


internal/crypto/signer/signer.go
```diff
@@ -40,33 +43,43 @@ func NewCASigners(ca CA) *CASigners {
 	ret := &CASigners{
 		ca: ca,
 		signers: map[string]Signer{
-			cfg.ClientBootstrapSignerName: WithSignerNameValidation(
-				WithCertificateReuse(
-					WithCSRValidation(
-						WithSignerNameExtension(NewClientBootstrap, ca),
-					),
-				),
+			cfg.ClientBootstrapSignerName: compose(
```
sk-ilya, line 46: No functional changes except the addition of `WithOrgIDExtension`. 
The use of `compose` makes the decorator chain read in the actual execution order top-to-bottom instead of right-to-left.

hexfusion (commented):
couple questions
- will a device in the future ever cross multiple orgs?
- can a device change org-id if so how is/will that be handeled?

sk-ilya:
@hexfusion It's hard to predict the future :) but for now, all resources (including CSRs, ERs, and devices themselves) belong to exactly one org and cannot be moved across orgs. This is consistent with k8s namespaces.

@DakCrowder  might have more input.

DakCrowder:
@sk-ilya That tracks with my thoughts

internal/crypto/signer/signer.go
```diff
@@ -209,25 +222,93 @@ type ctxKey string
...
+				return inst.Sign(ctx, request)
+			},
+		}
+	}
+}
+
+// WithOrgIDExtension Injects OrgID extension (from CSR or context) when issuing client certificates via CA.
+// Rules:
+// - If both CSR and context contain OrgID and they differ: fail verification/issuance.
+// - If CSR is missing OrgID and context has one: use context OrgID.
```
kkyrazis, line 241: Does this condition make it possible to have ERs exist with "no org", but have devices assigned to a specific org?

sk-ilya: It shouldn't happen - every request originating from a device is scoped by OrgID, which is derived from its peer certificate (bootstrap or long living), falling back to `org.DefaultID` if there is no OrgID extension:
https://github.com/flightctl/flightctl/blob/10256e570c85eab4fe8ce60dd57de1ec7d62f2c9/internal/api_server/middleware/middleware.go#L142-L160
Normally, the context should always contain the requester’s OrgID, ensuring that it will be stamped (or rejected if it mismatches the CSR).

DakCrowder: I was going to ask something similar, right now I _think_ the following flow exists:

The flightctl api writes a client-enrollment keypair on startup (https://github.com/flightctl/flightctl/blob/main/cmd/flightctl-api/main.go#L100).  This is used by some tests and the device simulator, and is currently provisioned outside the context of an org.  However later on I think the api would fall back to the default org id when proceeding with enrollment so it is effectively implicitly owned by the default org but ideally I think we tighten things up there.

Which speaking of @sk-ilya just curious have you tested the device sim with these changes?  Just want to verify we don't cause a regression.

sk-ilya: @DakCrowder 
>  I think the api would fall back to the default org id when proceeding with enrollment so it is effectively implicitly owned by the default org

That _is_ the idea (see my reply above). This way, we keep some backward compatibility.

> Just want to verify we don't cause a regression.

I ran `bin/flightctl-agent` before with both no/default org cert as well as with non-default org cert (by populating a new org in the DB) - it looked OK with all resources scoped properly. 
Just ran the simulator and saw no regression (yay). But we 100% need some e2e coverage

DakCrowder (approved):

internal/crypto/signer/signer.go
```diff
@@ -40,33 +43,43 @@ func NewCASigners(ca CA) *CASigners {
...
-					),
-				),
+			cfg.ClientBootstrapSignerName: compose(
+				compose(
+					NewClientBootstrap,
+					WithOrgIDExtension,
+					WithSignerNameExtension,
+				)(ca),
+				WithCSRValidation,
+				WithCertificateReuse,
```
asafbss: The compose function is great — I like how it makes the decorator chaining clean and easy 👍.
One thing to watch out for: it runs functions in reverse order from how they’re written.
For example, WithSignerNameValidation will run before WithCertificateReuse even though it’s written after.
This can be a bit confusing and is prone to misuse or errors

asafbss (dismissed):

asafbss (approved):

PR #1528, opened 2025-08-11
Updating Metrics Docs

SiddarthR56:


docs/user/metrics.md
```diff
@@ -21,26 +21,44 @@ Monitors system resource usage on the Flight Control server.
...
 
 - `enabled`: Enable/disable system collector (default: `true`)
 - `tickerInterval`: Collection frequency (default: `"5s"`)
 
 ### HTTP Collector
 
-Tracks HTTP API performance and service level compliance.
+Tracks HTTP API performance using OpenTelemetry instrumentation. This collector integrates OpenTelemetry HTTP metrics with the existing Prometheus metrics infrastructure, providing standardized HTTP observability.
+
+**Features:**
```
asafbss, line 35: I need more clarification about this header...
Should we even mention the tracing?

SiddarthR56: Will remove the part about tracing. 
Do you think the header feels too technical and implementation-focused? Would it work better if it were simpler?

asafbss: @avishayt wdyt?

PR #1531, opened 2025-08-12
Emit single device disconnected event; add summary to alerts

avishayt:
Fix two bugs:
1. EDM-1996: Emit single event for disconnected device
    We previously emitted 3 events because each property that became
    "unknown" triggered a separate event. We now make sure to emit only one.
2. EDM-1997: Annotate alerts with the event message
    Add a "summary" to each alert with a human-readable message.

## Summary by CodeRabbit

- New Features
  - Alerts now include a summary derived from event messages, improving context and readability in downstream alerting systems.
- Bug Fixes
  - Prevents duplicate “Device Disconnected” notifications when multiple status fields change in the same update, ensuring a single notification per update.
- Tests
  - Added tests to verify deduplication, ensuring only one “Device Disconnected” event is emitted per update cycle.

### DIFF
--- a/internal/service/event_handler.go
+++ b/internal/service/event_handler.go
@@ -77,8 +77,19 @@ func (h *EventHandler) HandleDeviceUpdatedEvents(ctx context.Context, resourceKi
 	// Only generate status change events when the device is not being created
 	if !created {
 		statusUpdates := common.ComputeDeviceStatusChanges(ctx, oldDevice, newDevice, orgId, h.store)
+
+		// Deduplicate DeviceDisconnected events - if multiple status fields changed to Unknown,
+		// only emit one DeviceDisconnected event
+		deviceDisconnectedEmitted := false
 		for _, update := range statusUpdates {
-			h.CreateEvent(ctx, common.GetDeviceEventFromUpdateDetails(ctx, name, update))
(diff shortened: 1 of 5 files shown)


internal/service/event_handler.go
```diff
@@ -77,8 +77,19 @@ func (h *EventHandler) HandleDeviceUpdatedEvents(ctx context.Context, resourceKi
 	// Only generate status change events when the device is not being created
 	if !created {
 		statusUpdates := common.ComputeDeviceStatusChanges(ctx, oldDevice, newDevice, orgId, h.store)
+
+		// Deduplicate DeviceDisconnected events - if multiple status fields changed to Unknown,
+		// only emit one DeviceDisconnected event
+		deviceDisconnectedEmitted := false
```
kkyrazis, line 83: I took a quick look and it seemed like this is the only type of event that can be duplicated across the different device statuses, but was curious about your thoughts on just generically deduplicating all events here. Not until we need it?

kkyrazis (approved):

internal/service/event_handler.go
avishayt: In general you're right.  However, where possible, I'm emitting events as callbacks from the store layer with the resources before and after DB updates.  This should guarantee emitting events once.

PR #1534, opened 2025-08-13
Keycloak multi org auth

DakCrowder:


kkyrazis (commented):
I understand adding the configuration to disable organizations when creating OIDC auth objects in the code (maybe someones still deploying keycloak 25), but why do we need to have organizations optionally configured in the deployments?

DakCrowder:
@kkyrazis To be clear, are you talking about the additions to the keycloak helm charts?  The additions to the keycloak realm config is  simply to aid in making testing easier - it is currently configured to setup an admin and demouser for use signing into the admin portal and Flight Control client when testing.  The org specific config can all be setup using in the keycloak UI but having a one shot deployment toggle that creates orgs, the scopes etc. has been nice for development

kkyrazis (approved):

PR #1537, opened 2025-08-14
EDM-2001: Performance improvements for steady state scenario

ori-amizur:
This change addresses 2 main problems during testing of large population during steady state scenario:
- The periodic update status operation done by the device agent takes a lot of resources - mainly database.
- The periodic get rendered device takes a lot of resources - also database.

This PR tries to reduce the load by performming the following:
- Reduce the frequency of the messages.
- Remove periodic access to the databse on behalf of a single device.

The following changes were made:
- Add new healthcheck endpoint.  Let the agent send this request every 2 minutes.
- Make the agent send update status only if the status changed from the last time it was sent.
- Make the server check if rendered version changed in redis. If it has not, wait for such notification for 2 minutes before returning to the client.
- Let the server accumulate healthchecdk requests, and update last_seen field for multiple devices instead of single device.

## Summary by CodeRabbit

- **New Features**
  - Added listing endpoint for EnrollmentRequests and a per-device rendered-version wait/notify system.
  - New healthcheck batching service to record device heartbeats.

- **Improvements**
  - Agent-integrated health checks, standardized OutOfDate/OutOfSync messages, HTTP/2 client support, and higher Kubernetes client QPS/Burst defaults.
  - More robust long-polling and version-aware rendering sync.

- **Operations**
  - Increased API container file-descriptor limits and DB max connections; deploy script image matching improved.

- **Tests**
  - Redis Pub/Sub tests added; integration tests updated for enqueue-based queue APIs.


deploy/podman/flightctl-api/flightctl-api.container
```diff
@@ -24,6 +24,7 @@ Volume=/etc/flightctl/flightctl-api/config.yaml:/root/.flightctl/config.yaml:ro,
 [Service]
 Restart=always
 RestartSec=30
+LimitNOFILE=524288
```
hexfusion: question: why 524288?

deploy/podman/flightctl-db/flightctl-db.container
```diff
@@ -5,6 +5,7 @@ PartOf=flightctl.target
 [Container]
 ContainerName=flightctl-db
 Environment=POSTGRESQL_DATABASE=flightctl POSTGRESQL_USER=flightctl_app POSTGRESQL_MASTER_USER=admin
+Environment=POSTGRESQL_MAX_CONNECTIONS=300
```
hexfusion: question: why 300

api/v1alpha1/agent/openapi.yaml
```diff
@@ -168,6 +168,49 @@ paths:
             application/json:
               schema:
                 $ref: '../openapi.yaml#/components/schemas/Status'
+
+  /api/v1/devices/{name}/healthcheck:
```
hexfusion: question: can you help me understand why this is nessisary

internal/agent/agent.go
```diff
@@ -154,6 +155,10 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	policyManager := policy.NewManager(a.log)
 
+	healthcheckManager := healthcheck.New(deviceName,
```
hexfusion: i thought “healthy” simply meant the service responded to a request from the device, so in normal operations even an empty status update would signal it’s alive. I understand avoiding unnecessary updates, but I’m not clear why we need a separate endpoint and manager for something the system already handles automatically. more concisely why isnt an empty status or a call to get spec enough?

internal/agent/device/status/status.go
```diff
@@ -141,6 +144,21 @@ func (m *StatusManager) ReloadCollect(ctx context.Context, _ *config.Config) err
...
 }
 
+func (m *StatusManager) update(ctx context.Context) {
+	if !reflect.DeepEqual(m.lastStatus, m.device.Status) {
+		if err := m.managementClient.UpdateDeviceStatus(ctx, m.deviceName, *m.device); err != nil {
+			m.log.Warnf("Failed to update device status: %v", err)
+		} else {
+			st, ok := deepcopy.Copy(m.device.Status).(*v1alpha1.DeviceStatus)
+			if !ok {
+				m.log.Warnf("Failed to deep copy device status: %v", err)
```
hexfusion: can error be not nil here?

internal/console/console.go
```diff
@@ -80,6 +91,9 @@ func (m *ConsoleSessionManager) modifyAnnotations(ctx context.Context, deviceNam
 			break
 		}
 	}
+	if err == nil {
+		err = rendered_version.Bus.Instance().StoreAndNotify(ctx, getOrgIdFromContext(ctx), deviceName, nextRenderedVersion)
```
hexfusion: question: why should error never be nil

internal/rendered_version/pubsub.go
```diff
@@ -0,0 +1,80 @@
...
+
+const queueName = "rendered_version_notifier"
+
+type Broadcaster interface {
+	Broadcast(ctx context.Context, orgId uuid.UUID, name string, renderedVersion string) error
+}
+
+type Subscriber interface {
+	Subscribe(ctx context.Context, handler func(ctx context.Context, orgId uuid.UUID, name string, renderedVersion string) error) error
+}
```
hexfusion: I guess the expectation is if we have multiple API servers each has its own ?

internal/store/model/device.go
```diff
@@ -101,9 +104,10 @@ func NewDeviceFromApiResource(resource *api.Device) (*Device, error) {
 			Owner:           resource.Metadata.Owner,
 			ResourceVersion: resourceVersion,
 		},
-		Alias:  alias,
-		Spec:   MakeJSONField(spec),
-		Status: MakeJSONField(status),
+		Alias:    alias,
+		Spec:     MakeJSONField(spec),
+		Status:   MakeJSONField(status),
+		LastSeen: status.LastSeen,
```
hexfusion: why duplicate  `status.LastSeen`, which is the canonnical source of truth

pkg/queues/provider.go
```diff
@@ -9,6 +9,8 @@ import (
 type Provider interface {
 	NewConsumer(queueName string) (Consumer, error)
 	NewPublisher(queueName string) (Publisher, error)
+	NewBroadcaster(channelName string) (Broadcaster, error)
+	NewSubscriber(channelName string) (Subscriber, error)
```
hexfusion: what is the difference between a publisher and broadcaster and then subscriber and consumer maybe adding docs to the interface would help here

internal/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,103 @@
...
+			if len(pending[req.orgId]) < maxPendingNames {
+				continue
+			}
+		case <-ticker.C:
+		case <-ctx.Done():
+			h.log.Info("Healthcheck worker context cancelled, stopping")
+			return
+		}
+		for orgId := range pending {
+			h.save(context.Background(), orgId, pending[orgId])
```
kkyrazis:
```suggestion
			h.save(ctx, orgId, pending[orgId])
```

internal/agent/agent.go
ori-amizur: Since healthcheck is a required operation, what is the harm to have such endpoint?
As far as the update status is concerned, in this PR the update status is not sent if there is no status change.  So if it is not sent then it cannot be used as healthcheck.
About the get rendered device, this call can be called not from the device (CLI supports this API call).
So only when this call is issued, it is clear that it comes from a device.  It is explicit and clear what is the intention behind it.

deploy/podman/flightctl-db/flightctl-db.container
ori-amizur: I would say it is worked during my tests, although after these changes it might not be needed.  I will test.

deploy/podman/flightctl-api/flightctl-api.container
ori-amizur: It is the number of file descriptors.
Since we now have at least 2 file descriptors per device so 100000 devices means at least 200000 constantly open file descriptors - without counting gRPC etc.
But maybe 300000 will be enough.

internal/agent/device/status/status.go
ori-amizur: Right.  Will fix

internal/console/console.go
ori-amizur: 1. The number of retries (10) exhausted.
2. Non nil error which is not  flterrors.ErrResourceVersionConflict in line 91

internal/rendered_version/pubsub.go
ori-amizur: Correct

internal/service/device.go
```diff
@@ -292,12 +294,46 @@ func (h *ServiceHandler) PatchDeviceStatus(ctx context.Context, name string, pat
...
+		err               error
+	)
+
 	orgId := getOrgIdFromContext(ctx)
 
-	result, err := h.store.Device().GetRendered(ctx, orgId, name, params.KnownRenderedVersion, h.agentEndpoint)
-	if err == nil && result == nil {
-		return nil, api.StatusNoContent()
+	if params.KnownRenderedVersion != nil {
+		isNew, kvRenderedVersion, err = rendered_version.Bus.Instance().WaitForNewVersion(ctx, orgId, name, *params.KnownRenderedVersion)
```
kkyrazis: Why do we have the handler block for 2 minutes waiting for a new version instead of just immediately returning NoContent if there isn't a new version? This feels like an attempt at pushing updates to devices, but hoping that they're attempting a poll when the update occurs. I'm unsure how often we expect real devices to be updating, but I would guess that it's unexpected that every `GetRenderedDevice` results in a new version.

We probably also need to update the agent's poller to remove that log for slow api calls.

internal/store/model/device.go
ori-amizur: The interface of the API hasn't changed.  So since we return "API"  record and not "db" record, the field in the API record must remain.
So when transferring between db and API record this field must be assigned as well.
The reason to have additional field is that update of this type of a field (not JSON), is much more lightweight and can be executed for multiple records easily.

pkg/queues/provider.go
ori-amizur: Maybe I need to think of better names...
These 2 represent a model that there is a single broadcaster and multiple consumers (for our case, the consumers are API servers). In the existing model there is a queue, so if one consumer consumes a message, the other will not see it.

api/v1alpha1/agent/openapi.yaml
ori-amizur: I answered in https://github.com/flightctl/flightctl/pull/1537/files#r2277480605 .  I think it is relevant also for this comment.

internal/store/device.go
```diff
@@ -659,18 +666,25 @@ func (s *DeviceStore) updateRendered(ctx context.Context, orgId uuid.UUID, name,
...
-	return retryUpdate(func() (bool, error) {
-		return s.updateRendered(ctx, orgId, name, renderedConfig, renderedApplications)
-	})
+func (s *DeviceStore) UpdateRendered(ctx context.Context, orgId uuid.UUID, name, renderedConfig, renderedApplications string) (string, error) {
+	var (
+		retry bool
+		rv    string
+		err   error
+	)
+	i := 0
```
kkyrazis: what about something like:
```go
func (s *DeviceStore) UpdateRendered(ctx context.Context, orgId uuid.UUID, name, renderedConfig, renderedApplications string) (string, error)
    var (
		rv    string
	)

	wrapper := func() (bool, error) {
		var retry bool
        var err error
		retry, rv, err = s.updateRendered(ctx, orgId, name, renderedConfig, renderedApplications)
		return retry, err
	}

	err  := retryUpdate(wrapper)
	return rv, err
}
```

Obviously up to you, but this certainly is a for loop

deploy/podman/flightctl-db/flightctl-db.container
kkyrazis: The helm deployments set it at 200, so perhaps set it to that to keep it aligned? https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/values.yaml#L102

internal/service/device.go
ori-amizur: If we are waiting then we are able to respond immediately for changes, but still reduce the number of API calls.
If we return immediately then we might not be able to reduce the number of API calls and the number of queries we are making (even to redis) to find if a version has changed.
For steady state scenario we do not expect the devices to be updated often.
I don't think there is a log in the agent for long api calls.

internal/store/device.go
ori-amizur: Possible.

internal/service/device.go
kkyrazis: https://github.com/flightctl/flightctl/blob/main/internal/agent/device/publisher/publisher.go#L112-L121

pkg/queues/provider.go
kkyrazis: Could the provider interface just be NewPublisher/Consumer and have separate constructors? `queues.NewRedisStreamProvider` and `queues.NewRedisPubSubProvider`

internal/service/device.go
ori-amizur: Debug, but OK

pkg/queues/provider.go
ori-amizur: Changed naming.  Hope it is clearer

internal/store/model/device.go
ori-amizur: Reverted the adding of additional column

internal/agent/agent.go
hexfusion: Since this is an internal contract between the agent and the service, could we add a HEAD verb to the status endpoint? It would be documented as part of the agent–service contract and used exclusively by the agent.

Could we consider this as an alternative?

agent/openapi.yaml
```
/api/v1/devices/{name}/status:
  head:
    tags:
      - device
    description: Report no status change, this also acts as a device heartbeat.
    operationId: heartbeatDeviceStatus
```

hexfusion: > Since healthcheck is a required operation, what is the harm to have such endpoint?

Isn’t a health check just an internal contract between agent and service, vs a user facing API? Adding a manager, timer, and goroutine feels like unnecessary complexity. We already have separate duration configs for device pull and status push do we really need a third?

ori-amizur: Even if it is internal, is implicit hidden functionality better ?
The additional duration is needed even for the empty status.  It indicates at what interval status must to be sent even if there is no status change.
It is possible to merge these calls, but my opinion that it is patchy.  But it is possible functionality wise - assuming we still have the additional duration.  Otherwise we will need it hard coded.

ori-amizur: One more thing: If the system is now busy with an update, healthcheck will not be sent (if we keep it as part of the update status loop).  Is this the behavior we want?

deploy/scripts/deploy_quadlets.sh
```diff
@@ -33,7 +33,7 @@ switch_to_local_images() {
     # Update db-setup image in container files to use locally built image
     db_migrate_container="${QUADLET_FILES_OUTPUT_DIR}/flightctl-db-migrate.container"
     if [[ -f "$db_migrate_container" ]] && grep -q "flightctl-db-setup:" "$db_migrate_container"; then
-        sed -i "s|flightctl-db-setup:[^ ]*|flightctl-db-setup:latest|g" "$db_migrate_container"
+        sed -i "s|[^ ]*flightctl-db-setup:[^ ]*|flightctl-db-setup:latest|g" "$db_migrate_container"
```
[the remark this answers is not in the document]
ori-amizur: That's the meaning - to replace the remote image with local one.  So the prefix before the image name is removed

internal/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,103 @@
...
+			h.log.Info("Healthcheck worker context cancelled, stopping")
+			return
+		}
+		for orgId := range pending {
+			h.save(ctx, orgId, pending[orgId])
+		}
+		pending = make(map[uuid.UUID][]string)
+		ticker.Reset(saveInterval)
+	}
+}
```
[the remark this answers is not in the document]
ori-amizur: If the context is cancelled we don't care about the remaining heartbeats

internal/rendered_version/manager.go
```diff
@@ -0,0 +1,134 @@
...
+	renderedVersionTimeout = 120 * time.Second // Timeout for waiting for a new rendered version
+)
+
+type Manager struct {
+	kvStore     kvstore.KVStore
+	subscriber  Subscriber
+	broadcaster Broadcaster
+	subscribers sync.Map
+	log         logrus.FieldLogger
+}
```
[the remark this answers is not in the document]
ori-amizur: There are no multiple subscribers per device

internal/agent/agent.go
ori-amizur: Another thing, the healthcheck is only part of the agent openapi - so it is not visible by the user facing api

hexfusion: > One more thing: If the system is now busy with an update, healthcheck will not be sent (if we keep it as part of the update status loop). Is this the behavior we want?

Does this require a new endpoint? In the background is the already added publisher that lives in its own go routine. Its calling to the service on a given interval regardless of work being done.

hexfusion: > Even if it is internal, is implicit hidden functionality better ?

Correct it should be a documented contract.

ori-amizur: The publisher isn't calling the service, it just adds to the queue.  The service still has its own loop.  If we keep it the way it is, update status (and healthcheck) are blocked when update is in progress.  If update takes more than 5 minutes, the device will be marked disconnected.

What is the harm to have a new endpoint?

ori-amizur: > Correct it should be a documented contract.

There are 2 issues here:
1. Should the healthcheck be independent from the main loop in case the main loop is busy for a long time?
2. Should we use the same openapi endpoint?

internal/client/config.go
```diff
@@ -280,10 +281,21 @@ func NewHTTPClientFromConfig(config *Config) (*http.Client, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("NewHTTPClientFromConfig: configuring HTTP/2 transport: %w", err)
+	}
+
 	httpClient := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: tlsConfig,
-		},
+		Transport: transport,
 	}
```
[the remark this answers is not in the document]
ori-amizur: @coderabbitai can you show in the code where ForceAttemptHTTP2 is set by http2.ConfigureTransport ?

internal/store/device.go
```diff
@@ -630,11 +616,49 @@ func (s *DeviceStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, na
...
+		orgId, owner, api.DeviceUpdatedStatusUpToDate).Updates(map[string]any{
+		"status": gorm.Expr(`jsonb_set(jsonb_set(status, '{updated,status}', to_jsonb(?::text)),'{updated,info}', to_jsonb(?::text))`,
+			api.DeviceUpdatedStatusOutOfDate, api.DeviceOutOfSyncWithFleetText),
+		"resource_version": gorm.Expr("resource_version + 1"),
+	}).Error
+	if err != nil {
+		return strings.Contains(err.Error(), "deadlock"), ErrorFromGormError(err)
+	}
+	return false, nil
+}
```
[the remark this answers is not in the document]
ori-amizur: What do you mean by null propagation ?

pkg/queues/publish_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+func TestPublishSubscribe(t *testing.T) {
+	// This test requires a Redis instance running on localhost:6379
+	// Skip if Redis is not available
+	provider, err := NewRedisProvider(context.Background(), logrus.New(), "localhost", 6379, "")
+	if err != nil {
+		t.Skipf("Redis not available: %v", err)
+		return
+	}
+	defer provider.Stop()
```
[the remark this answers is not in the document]
ori-amizur: SecureString type derives from a string so setting a string constant compiles.

internal/agent/device/publisher/publisher.go
```diff
@@ -109,14 +111,17 @@ func (n *publisher) pollAndPublish(ctx context.Context) {
 
 	newDesired := &v1alpha1.Device{}
 
+	var cancel context.CancelFunc
 	startTime := time.Now()
+	ctx, cancel = context.WithTimeout(ctx, longPollTimeout)
```
hexfusion, line 116: nit: no need to declare cancelFn below provides the proper scope
```suggestion
	ctx, cancel := context.WithTimeout(ctx, longPollTimeout)
```

internal/agent/device/publisher/publisher.go
```diff
@@ -16,6 +16,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+const longPollTimeout = 4 * time.Minute
```
hexfusion, line 19: question: one minute is a really long time to wait for a distributed system why 4 minutes?

internal/agent/device/status/status.go
```diff
@@ -141,6 +144,21 @@ func (m *StatusManager) ReloadCollect(ctx context.Context, _ *config.Config) err
...
 }
 
+func (m *StatusManager) update(ctx context.Context) {
+	if !reflect.DeepEqual(m.lastStatus, m.device.Status) {
+		if err := m.managementClient.UpdateDeviceStatus(ctx, m.deviceName, *m.device); err != nil {
+			m.log.Warnf("Failed to update device status: %v", err)
+		} else {
+			st, ok := deepcopy.Copy(m.device.Status).(*v1alpha1.DeviceStatus)
+			if !ok {
+				m.log.Warn("Failed to deep copy device status")
```
hexfusion, line 154: This seems like its something very unexpected should it be an error?

internal/api_server/agentserver/server.go
```diff
@@ -9,10 +9,11 @@ import (
 	"net/http"
 	"time"
 
-	api "github.com/flightctl/flightctl/api/v1alpha1"
+	agent "github.com/flightctl/flightctl/api/v1alpha1/agent"
```
hexfusion, line 12:
```suggestion
	"github.com/flightctl/flightctl/api/v1alpha1/agent"
```

internal/client/config.go
```diff
@@ -295,10 +296,21 @@ func NewHTTPClientFromConfig(config *Config) (*http.Client, error) {
...
 
+	// Configure transport for HTTP/2 support
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+		// Enable HTTP/2
+		ForceAttemptHTTP2: true,
+	}
+
+	// Configure HTTP/2
+	err = http2.ConfigureTransport(transport)
```
hexfusion, line 307: neat!

test/integration/store/device_test.go
```diff
@@ -660,36 +660,6 @@ var _ = Describe("DeviceStore create", func() {
 			Expect(*dev.Metadata.Annotations).To(HaveLen(0))
 		})
 
-		It("UpdateDeviceAnnotations console", func() {
```
hexfusion, line 663: question: was this test just moved?

pkg/queues/redis_provider.go
```diff
@@ -24,12 +25,13 @@ import (
...
-	wg      *sync.WaitGroup
-	queues  []*redisQueue
-	stopped atomic.Bool
-	mu      sync.Mutex
+	client   *redis.Client
+	log      logrus.FieldLogger
+	wg       *sync.WaitGroup
+	queues   []*redisQueue
+	channels []*redisChannel
+	stopped  atomic.Bool
```
hexfusion: nit: if you already have a mutex cheaper to just use a real bool

pkg/k8sclient/k8s_client.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 const (
+	defaultBurst         = 1000
+	defaultQPS           = 500
```
hexfusion, line 15: question: just for my understanding how were these numbers chosen? I do see for example the KAS using a similar ratio 2:1 QPS to Burst.

ref. https://github.com/openshift/kubernetes-apiserver/blob/7742d76eabcb5f3d49e995b91b499b478d4f2f2e/pkg/server/options/authentication.go#L455

internal/service/device.go
hexfusion: > For steady state scenario we do not expect the devices to be updated often.

Maybe to help us make the decision we can see the actual vs perceived performance gains? Do we have that data? Will having a larger number of concurrent requests open cause any performance issues at scale? Will future load balancers or proxies cause interference at all?

For example I am curious if we could start at say 30-45s. It would be nice to see the metrics at 100k. Starting with 30s then move up to 2m.

hexfusion: In general I think we should start lower and increase as required. WDYT?

internal/store/device.go
```diff
@@ -55,15 +55,18 @@ type Device interface {
...
 	// Used internally
 	UpdateAnnotations(ctx context.Context, orgId uuid.UUID, name string, annotations map[string]string, deleteKeys []string) error
-	UpdateRendered(ctx context.Context, orgId uuid.UUID, name, renderedConfig, renderedApplications string) error
+	UpdateRendered(ctx context.Context, orgId uuid.UUID, name, renderedConfig, renderedApplications string) (string, error)
 	SetServiceConditions(ctx context.Context, orgId uuid.UUID, name string, conditions []api.Condition, callback ServiceConditionsCallback) error
 	OverwriteRepositoryRefs(ctx context.Context, orgId uuid.UUID, name string, repositoryNames ...string) error
 	GetRepositoryRefs(ctx context.Context, orgId uuid.UUID, name string) (*api.RepositoryList, error)
 	PrepareDevicesAfterRestore(ctx context.Context) (int64, error)
 	RemoveConflictPausedAnnotation(ctx context.Context, orgId uuid.UUID, listParams ListParams) (int64, []string, error)
+	UpdateToOutOfDateByOwner(ctx context.Context, orgId uuid.UUID, owner string) error
```
hexfusion:
```suggestion
	SetOutOfDate(ctx context.Context, orgId uuid.UUID, owner string) error
```

internal/store/device.go
```diff
@@ -632,11 +618,49 @@ func (s *DeviceStore) UpdateAnnotations(ctx context.Context, orgId uuid.UUID, na
 	})
 }
 
-func (s *DeviceStore) updateRendered(ctx context.Context, orgId uuid.UUID, name, renderedConfig, renderedApplications string) (retry bool, err error) {
+func (s *DeviceStore) healthcheck(ctx context.Context, orgId uuid.UUID, names []string) (bool, error) {
+	result := s.getDB(ctx).Model(&model.Device{}).Where("org_id = ? and name in (?)", orgId, names).Update(
+		"status", gorm.Expr(fmt.Sprintf(`jsonb_set(status, '{lastSeen}', '"%s"', true)`, time.Now().Format(time.RFC3339))))
+	err := ErrorFromGormError(result.Error)
+	if err != nil {
+		return strings.Contains(err.Error(), "deadlock"), err
```
hexfusion, line 626: this feels a bit brittle should we extend ErrorFromGormError to perform this and return an error we can unwrap like errors.Is(...?

internal/tasks/device_disconnected.go
```diff
@@ -36,11 +36,11 @@ func (t *DeviceDisconnected) Poll(ctx context.Context) {
 	defer cancel()
 
 	// Calculate the cutoff time for disconnected devices
-	cutoffTime := time.Now().Add(-api.DeviceDisconnectedTimeout)
+	cutoffTime := time.Now().Add(-api.DeviceDisconnectedTimeout).Format(time.RFC3339)
```
hexfusion: just for my own understanding why these changes.

internal/agent/device/publisher/publisher.go
ori-amizur: OK

internal/agent/device/publisher/publisher.go
ori-amizur: For 2 minutes interval, 4 minutes timeout sounds reasonable.

internal/agent/device/status/status.go
ori-amizur: OK

internal/api_server/agentserver/server.go
ori-amizur: OK

test/integration/store/device_test.go
ori-amizur: The method of updating console annotations changed with the console changes.  So this test is not relevant any more

pkg/k8sclient/k8s_client.go
ori-amizur: I took the default numbers and multiplied them by 100.  It solved latency problems especially for API calls.
I can't say I used any smart method to calculate these numbers.

internal/service/device.go
ori-amizur: The frequency of the API calls has direct affect on performance.  When measured, 30s interval lead to 90% api-server CPU, while 2 minutes lead to 30% api-server CPU.
Usually we would decrease the interval to gain responsiveness and increase to gain scalability and performance.  With the current implementation we have 2 minutes timeout but with immediate responsiveness.  So why decrease it?

internal/store/device.go
ori-amizur: That's the pattern we are using everywhere in this package. 
Anyway, I will take a look

internal/tasks/device_disconnected.go
ori-amizur: I can revert.

internal/service/device.go
ori-amizur: In general, there is no benefit that we gain if we start low.

deploy/scripts/deploy_quadlets.sh
```diff
@@ -24,7 +24,7 @@ switch_to_local_images() {
     # Update db-setup image in service files to use locally built image
     db_migrate_service="${SYSTEMD_UNIT_OUTPUT_DIR}/flightctl-db-migrate.service"
     if [[ -f "$db_migrate_service" ]] && grep -q "flightctl-db-setup:" "$db_migrate_service"; then
-        sed -i "s|flightctl-db-setup:[^ ]*|flightctl-db-setup:latest|g" "$db_migrate_service"
+        sed -i "s|[^ ]*flightctl-db-setup:[^ ]*|flightctl-db-setup:latest|g" "$db_migrate_service"
```
hexfusion, line 27: question: was this a bug or why this change?

internal/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (h *Healthchecker) Add(ctx context.Context, orgId uuid.UUID, name string) error {
+	select {
+	case h.input <- heartbeatRequest{orgId: orgId, name: name}:
+	default:
+		h.log.Warnf("Healthcheck channel is full, dropping request for %s in org %s", name, orgId)
+		return fmt.Errorf("healthcheck channel is full, dropping request for %s in org %s", name, orgId)
```
hexfusion, line 100: since we are returning an error here should this also log error?

internal/healthchecker/healthchecker.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return nil
+}
+
+func startSpan(ctx context.Context, method string) (context.Context, trace.Span) {
+	ctx, span := tracing.StartSpan(ctx, "flightctl/healthcheck", method)
+	return ctx, span
```
hexfusion: minor
```suggestion
	return tracing.StartSpan(ctx, "flightctl/healthcheck", method)
```

internal/rendered_version/manager.go
```diff
@@ -0,0 +1,137 @@
...
+
+var Bus util.Singleton[Manager]
+
+func New(ctx context.Context, kvStore kvstore.KVStore, provider queues.Provider, log logrus.FieldLogger) *Manager {
+	broadcaster, err := NewBroadcaster(ctx, provider)
+	if err != nil {
+		log.Fatalf("Failed to create publisher for rendered version: %v", err)
```
hexfusion: nit: fatal is pretty heavy for a constructor this tells the process to exit non zero as a consiquence all defer statements will not fire.  Please return err.

internal/rendered_version/manager.go
```diff
@@ -0,0 +1,137 @@
+package rendered_version
```
hexfusion: nit: in general high level package names should be more general to allow further reuse.

```suggestion
package render
```

usage

render.NewVersionManager()

internal/rendered_version/pubsub.go
```diff
@@ -0,0 +1,92 @@
...
+			return err
+		}
+		return handler(ctx, s.OrgId, s.Name, s.RenderVersion)
+	}
+
+	_, err := c.subscriber.Subscribe(ctx, queuesHandler)
+	if err == nil {
+		c.subscriptions = append(c.subscriptions, c.subscriber)
+	}
+	return err
```
hexfusion: optional style
```
	_, err := c.subscriber.Subscribe(ctx, queuesHandler)
	if err != nil {
		return err
   }
   c.subscriptions = append(c.subscriptions, c.subscriber)
   return nil
```

internal/service/device.go
```diff
@@ -314,11 +317,44 @@ func (h *ServiceHandler) PatchDeviceStatus(ctx context.Context, name string, pat
...
 
 func (h *ServiceHandler) GetRenderedDevice(ctx context.Context, name string, params api.GetRenderedDeviceParams) (*api.Device, api.Status) {
+	var (
+		isNew             bool
+		kvRenderedVersion string
+		err               error
+	)
+
+	if _, ok := ctx.Value(consts.AgentCtxKey).(string); ok {
+		if err := healthchecker.HealthChecks.Instance().Add(ctx, getOrgIdFromContext(ctx), name); err != nil {
```
hexfusion, line 336: should we also use status to report health?

we need to document this contract.

```
Heartbeat Semantics

Definition: Any successful call from the agent to the service constitutes a heartbeat signal.
- Scope: The specific API endpoint invoked is not relevant; the ability to successfully reach and authenticate with the service is sufficient.
- Cadence: The service infers agent health based on the most recent call timestamp. No separate, dedicated “heartbeat” endpoint is required.
```

internal/service/device.go
hexfusion: This is a change in the way the device configuration works. Basically it is not longer true that the user can define `SpecFetchInterval` as the interval is set by the service. We need to make those adjustments to the docs and I guess deprecate/remove the config?

internal/service/device.go
```diff
@@ -389,7 +445,19 @@ func (h *ServiceHandler) UpdateDeviceAnnotations(ctx context.Context, name strin
 
 func (h *ServiceHandler) UpdateRenderedDevice(ctx context.Context, name, renderedConfig, renderedApplications string) api.Status {
 	orgId := getOrgIdFromContext(ctx)
-	err := h.store.Device().UpdateRendered(ctx, orgId, name, renderedConfig, renderedApplications)
+	renderedVersion, err := h.store.Device().UpdateRendered(ctx, orgId, name, renderedConfig, renderedApplications)
+	if err == nil {
+		err = h.UpdateServerSideDeviceStatus(ctx, name)
+	}
+	if err == nil {
```
hexfusion: can we keep the standard  golang err flow.
```go
if err != nil {
    return StoreErrorToApiStatus(err, false, api.DeviceKind, &name)
}
if err := h.UpdateServerSideDeviceStatus(ctx, name); err != nil {
	return StoreErrorToApiStatus(err, false, api.DeviceKind, &name)
}
```

test/integration/store/device_test.go
```diff
@@ -660,36 +660,6 @@ var _ = Describe("DeviceStore create", func() {
...
-
-			err = devStore.UpdateAnnotations(ctx, orgId, "mydevice-1", nil, []string{api.DeviceAnnotationConsole})
-			Expect(err).ToNot(HaveOccurred())
-			dev, err = devStore.Get(ctx, orgId, "mydevice-1")
-			Expect(err).ToNot(HaveOccurred())
-			Expect(dev.Metadata.Annotations).ToNot(BeNil())
-			Expect(*dev.Metadata.Annotations).To(HaveLen(2))
-			Expect((*dev.Metadata.Annotations)["key1"]).To(Equal("val1"))
-			Expect((*dev.Metadata.Annotations)[api.DeviceAnnotationRenderedVersion]).To(Equal("2"))
-		})
```
hexfusion, line 691: did this test move?

hexfusion (commented):
getting closer! there are some changes to agent behaviour and contracts that should be documented at part of this change. thanks!

deploy/scripts/deploy_quadlets.sh
ori-amizur: Bug: We need to substitute the image quay.io/flightctl/flightctl-db-setup:latest to flightctl-db-setup:latest

internal/healthchecker/healthchecker.go
ori-amizur: OK

internal/rendered_version/manager.go
ori-amizur: OK

internal/rendered_version/manager.go
ori-amizur: BTW, why do we care for defer statements if the process exits?
I can return error but the usage will be less convenient in several places:
During initialization currently we use the following pattern:
This:
`
rendered.Bus.GetOrInit(rendered.NewVersionManager(ctx, kvStore, provider, log)).Start(ctx)
`
Or this:
`
rendered.Bus.GetOrInit(rendered.NewVersionManager(ctx, kvStore, provider, log))
`
Returning an error will force us to handle them in any location we call in this way:
```
vm, err := rendered.NewVersionManager(ctx, kvStore, provider, log)
if err != nil {
      log.Fatalf(".....")
}
rendered.GetOrInit(vm)
```

internal/rendered_version/pubsub.go
ori-amizur: OK

internal/service/device.go
ori-amizur: Healthcheck updates the `lastSeen` field in the status.  Same for `UpdateStatus`.  So both have same affect on the `lastSeen` field.
But currently, `UpdateStatus` is sent only on status modification.

internal/service/device.go
ori-amizur: OK

internal/rendered_version/manager.go
hexfusion: yes fatal from err makes sense and is more observable. my nit is more fatal on a constructor is not a common pattern. there is assumptions about defer I just want it to be clear if you fatal inline defer does not fire. this could be unexpected.

internal/service/device.go
hexfusion: ok so the semantics above are actually still true, lgtm

internal/service/device.go
ori-amizur: OK

internal/store/device.go
ori-amizur: OK

test/integration/store/device_test.go
ori-amizur: no

internal/api_server/agentserver/server.go
ori-amizur: It seems that if it is removed, golang doesn't like it

internal/rendered_version/manager.go
ori-amizur: Changed.  Now returning error

internal/rendered/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	subscriber, err := NewSubscriber(ctx, provider)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create subscriber for rendered version: %v", err)
+	}
+	return &VersionManager{
+		kvStore:             kvStore,
+		broadcaster:         broadcaster,
+		subscriber:          subscriber,
+		renderedWaitTimeout: lo.Ternary(renderedWaitTimeout > 0, renderedWaitTimeout, renderedVersionDefaultTimeout),
```
hexfusion, line 67: nit `renderedWaitTimeout timeout`  is declared at the config level. should the default value handling also be handled by `config`?

PR #1538, opened 2025-08-14
EDM-2004: Fix alertmanager access when deploying with nodeports

avishayt:
When using NodePort expose method, the UI should connect to alertmanager-proxy
using internal service name instead of external URL to avoid 502 errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Alertmanager proxy URL generation for NodePort deployments. The chart now targets the in-cluster proxy service over HTTPS on port 8443, instead of constructing an external NodePort URL. This improves connectivity consistency in NodePort environments, reduces misrouting risks, and ensures more reliable access to Alertmanager through the proxy.


celdrake:
Can we merge? This seems to be an issue also in the Staging environment.

PR #1554, opened 2025-08-19
RBAC Testing an admin role

itzikb-redhat:
First test for RBAC.
Creating a role with creating,listing and getting flightctl resources and a a role binding to a user with no prior permissions.

Making sure the user can:
Create, update, list, get a specific and delete a:
 - Device
 - Fleet
 - Repository

Create and list:
- Enrollment requests
- Events

## Summary by CodeRabbit

- Tests
  - Added comprehensive end-to-end RBAC suites exercising namespace- and cluster-scoped permissions with per-test VM/agent isolation, setup/teardown, and gating to skip when ACM isn’t available.
- Refactor
  - Centralized ACM presence detection into a shared test utility for consistent test gating across suites.
- Chores
  - Expanded test harness and tooling: RBAC and resource lifecycle helpers, YAML label tooling, Kubernetes context utilities, non-admin login helper, repository fetch helper, and an example Repository test manifest.

### DIFF
--- a/test/e2e/rbac/rbac_suite_test.go
+++ b/test/e2e/rbac/rbac_suite_test.go
@@ -1,0 +1,76 @@
+package rbac_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestRbac(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "RBAC E2E Suite")
+}
+
+var (
+	flightCtlNs string
+)
+
+var _ = BeforeSuite(func() {
+	_, _, err := e2e.SetupWorkerHarness()
+	Expect(err).ToNot(HaveOccurred())
+
+	// Check if ACM is installed before running any tests
+	isAcmInstalled, err := util.IsAcmInstalled()
+	if err != nil || !isAcmInstalled {
+		Skip("Skipping test suite because ACM is not installed.")
+	}
+})
+
+var _ = BeforeEach(func() {
+	flightCtlNs = os.Getenv("FLIGHTCTL_NS")
+	if flightCtlNs == "" {
+		Skip("FLIGHTCTL_NS environment variable should be set")
+	}
+
+	// Get the harness and context directly - no package-level variables
+	workerID := GinkgoParallelProcess()
+	harness := e2e.GetWorkerHarness()
+	suiteCtx := e2e.GetWorkerContext()
+
+	GinkgoWriter.Printf("🔄 [BeforeEach] Worker %d: Setting up test with VM from pool\n", workerID)
+
+	// Create test-specific context for proper tracing
+	ctx := util.StartSpecTracerForGinkgo(suiteCtx)
+
+	// Set the test context in the harness
+	harness.SetTestContext(ctx)
+
+	// Setup VM from pool, revert to pristine snapshot, and start agent
+	err := harness.SetupVMFromPoolAndStartAgent(workerID)
+	Expect(err).ToNot(HaveOccurred())
+
+	GinkgoWriter.Printf("✅ [BeforeEach] Worker %d: Test setup completed\n", workerID)
+})
+
+var _ = AfterEach(func() {
+	workerID := GinkgoParallelProcess()
+	GinkgoWriter.Printf("🔄 [AfterEach] Worker %d: Cleaning up test resources\n", workerID)
+
+	// Get the harness and context directly - no shared variables needed
+	harness := e2e.GetWorkerHarness()
+	suiteCtx := e2e.GetWorkerContext()
+
+	// Clean up test resources BEFORE switching back to suite context
+	// This ensures we use the correct test ID for resource cleanup
+	err := harness.CleanUpAllTestResources()
+	Expect(err).ToNot(HaveOccurred())
+
+	// Now restore suite context for any remaining cleanup operations
+	harness.SetTestContext(suiteCtx)
+
+	GinkgoWriter.Printf("✅ [AfterEach] Worker %d: Test cleanup completed\n", workerID)
+})
--- a/test/e2e/rbac/rbac_test.go
+++ b/test/e2e/rbac/rbac_test.go
@@ -1,0 +1,202 @@
+package rbac_test
+
+import (
+	"context"
+
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/login"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	nonAdminUser         = "demouser1"
+	adminRoleName        = "rbac-test-admin-role"
+	adminRoleBindingName = "rbac-test-admin-role-binding"
+	defaultNs            = "default"
+)
+
+var _ = Describe("RBAC Authorization Tests", Label("rbac", "authorization"), func() {
+	var (
+		harness           *e2e.Harness
+		suiteCtx          context.Context
+		defaultK8sContext string
+		k8sApiEndpoint    string
+	)
+
+	roles := []string{
+		adminRoleName,
+	}
+	roleBindings := []string{
+		adminRoleBindingName,
+	}
+	clusterRoles := []string{
+		adminRoleName,
+	}
+	clusterRoleBindings := []string{
+		adminRoleBindingName,
+	}
+	adminTestLabels := &map[string]string{"test": "rbac-admin"}
+	userTestLabels := &map[string]string{"test": "rbac-user"}
+
+	BeforeEach(func() {
+		var err error
+		// Get the harness and context set up by the suite
+		harness = e2e.GetWorkerHarness()
+		suiteCtx = e2e.GetWorkerContext()
+
+		// Get the default K8s context
+		defaultK8sContext, err = harness.GetDefaultK8sContext()
+		Expect(err).ToNot(HaveOccurred(), "Failed to get default K8s context")
+		k8sApiEndpoint, err = harness.GetK8sApiEndpoint(suiteCtx, defaultK8sContext)
+		Expect(err).ToNot(HaveOccurred(), "Failed to get Kubernetes API endpoint")
+	})
+
+	AfterEach(func() {
+		err := harness.ChangeK8sContext(suiteCtx, defaultK8sContext)
+		Expect(err).ToNot(HaveOccurred(), "Failed to change K8s context")
+		login.LoginToAPIWithToken(harness)
+
+		harness.CleanupRoles(suiteCtx, harness.Cluster, roles, roleBindings, flightCtlNs)
+		harness.CleanupClusterRoles(suiteCtx, harness.Cluster, clusterRoles, clusterRoleBindings)
+	})
+
+	Context("FlightCtl user", func() {
+		adminRole := &rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      adminRoleName,
+				Namespace: flightCtlNs,
+			},
+		}
+		adminRole.Rules = []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"flightctl.io"},
+				Resources: []string{"*"},
+				Verbs:     []string{"*"},
+			},
+		}
+		adminRoleBinding := &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      adminRoleBindingName,
+				Namespace: flightCtlNs,
+			},
+		}
+		adminRoleBinding.RoleRef = rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     adminRoleName,
+		}
+		adminRoleBinding.Subjects = []rbacv1.Subject{
+			{
+				Kind: "User",
+				Name: nonAdminUser,
+			},
+		}
+		adminClusterRole := &rbacv1.ClusterRole{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: adminRoleName,
+			},
+		}
+		adminClusterRole.Rules = []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"flightctl.io"},
+				Resources: []string{"*"},
+				Verbs:     []string{"*"},
+			},
+		}
+		adminClusterRoleBinding := &rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: adminRoleBindingName,
+			},
+		}
+		adminClusterRoleBinding.RoleRef = rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     adminRoleName,
+		}
+		adminClusterRoleBinding.Subjects = []rbacv1.Subject{
+			{
+				Kind: "User",
+				Name: nonAdminUser,
+			},
+		}
+
+		It("should have access full access with an admin role", Label("83842"), func() {
+			By("Login to the cluster as a user without a role")
+			err := login.LoginAsNonAdmin(harness, nonAdminUser, nonAdminUser, defaultK8sContext, k8sApiEndpoint)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Testing that operations should fail without a role")
+			operations := []string{e2e.OperationCreate, e2e.OperationList}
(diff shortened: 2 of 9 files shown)


test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+type flightCtlResource struct {
+	ResourceType string
+	ResourceName string
+}
+
+var _ = Describe("RBAC Authorization Tests", Label("rbac", "authorization"), func() {
+	var harness *e2e.Harness
+
+	flightCtlNs := os.Getenv("FLIGHTCTL_NS")
```
siserafin: Should we add a default value or fail the test if the env var is empty?

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,277 @@
...
+	BeforeEach(func() {
+		harness = e2e.NewTestHarness(suiteCtx)
+		flightCtlResources = []flightCtlResource{}
+	})
+
+	AfterEach(func() {
+		ChangeContext("default")
+		login.LoginToAPIWithToken(harness)
+
+		for _, resource := range flightCtlResources {
```
siserafin: Maybe you could use:

```
err := harness.CleanUpAllResources()

File: test/e2e/rbac/rbac_test.go (hunk 1 above)
Comment (itzikb-redhat) [reply test/e2e/rbac/rbac_test.go]: I think it should fail, wdyt?

File: test/e2e/rbac/rbac_test.go [hunk 3]
```diff
@@ -0,0 +1,277 @@
...
+			output, err := harness.CleanUpResource(resource.ResourceType, resource.ResourceName)
+			if err != nil {
+				logrus.Errorf("Failed to clean up resource %s of resource type %s: %v", resource.ResourceName, resource.ResourceType, err)
+			} else {
+				logrus.Infof("Cleaned up resource %s of resource type %s: %s", resource.ResourceName, resource.ResourceType, output)
+			}
+		}
+
+		for _, role := range roles {
```
siserafin: We could create a local function to cleanup the roles

test/e2e/rbac/rbac_test.go
itzikb-redhat: Won't that be a problem when running tests in parallel?

test/e2e/rbac/rbac_test.go
itzikb-redhat: Will do

test/e2e/rbac/rbac_test.go
itzikb-redhat: I've made it skip

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,288 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/yaml"
+)
+
+const (
```
siserafin, line 14: I would set these as constants:
```
const (
	nonAdminUser         = "demouser1"
	adminRoleName        = "rbac-test-admin-role"
	adminRoleBindingName = "rbac-test-admin-role-binding"
)

File: test/e2e/rbac/rbac_test.go [hunk 5]
```diff
@@ -0,0 +1,288 @@
...
+
+			adminRoleBinding, err := createRoleBinding(harness.Cluster, flightCtlNs, adminRoleBinding)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Login to the cluster by a user without a role")
+			cmd := exec.Command("bash", "-c", "oc whoami --show-server")
+			kubernetesApiEndpoint, err := cmd.CombinedOutput()
+			Expect(err).ToNot(HaveOccurred(), "Failed to get kubernetes api endpoint")
+
+			loginCommand := fmt.Sprintf("oc login -u %s -p %s %s", nonAdminUser, nonAdminUser, string(kubernetesApiEndpoint))
```
siserafin: login can be a helper function loginAsNonAdminUser(harness *e2e.Harness)

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,288 @@
...
+			cmd = exec.Command("bash", "-c", loginCommand)
+			err = cmd.Run()
+			Expect(err).ToNot(HaveOccurred())
+			defer ChangeContext("default")
+
+			method := login.LoginToAPIWithToken(harness)
+			Expect(method).ToNot(Equal(login.AuthDisabled))
+
+			By("Testing device operations - admin should have full access")
+			By("Testing creating a device")
```
siserafin: Resource creation and update steps are repeated . I think we could have a function like:
```
out, err := testResourceCreation(harness *e2e.Harness, resourceType, yamlFile string, flightCtlResources  []flightCtlResource)
Same for update.

File: test/e2e/rbac/rbac_test.go [hunk 7]
```diff
@@ -0,0 +1,288 @@
...
+			device.Metadata.Labels = &map[string]string{"test": "rbac-admin"}
+			deviceData, err := yaml.Marshal(&device)
+			Expect(err).ToNot(HaveOccurred())
+			logrus.Println("deviceData: ", string(deviceData))
+			out, err = harness.CLIWithStdin(string(deviceData), "apply", "-f", "-")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(MatchRegexp(`200 OK`))
+
+			By("Testing getting a specific device")
+			_, err = harness.CLI("get", "device"+"/"+*deviceName)
```
siserafin: We have GetResourcesByName

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,288 @@
...
+				out, err = harness.CLI("apply", "-f", resourceYamls["device"])
+				return err
+			}).Should(BeNil(), "failed to create a Device")
+			Expect(out).To(MatchRegexp(`201 Created`))
+			device := harness.GetDeviceByYaml(resourceYamls["device"])
+			deviceName := device.Metadata.Name
+			flightCtlResources = append(flightCtlResources, flightCtlResource{ResourceType: "device", ResourceName: *deviceName})
+
+			By("Testing updating a device")
+			device.Metadata.Labels = &map[string]string{"test": "rbac-admin"}
```
siserafin: We could have a helper function:
func addTestLabel(resource interface{})

siserafin (commented):
Looks good, I think we can have some optimizations, but the flow is ok.

test/e2e/rbac/rbac_test.go
itzikb-redhat: Thanks! Using it for listing resources as well.

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,338 @@
...
+			Expect(err).ToNot(HaveOccurred(), "Admin should be able to delete role binding")
+
+			By("Testing listing repositories without a role - should not be able to list")
+			_, err = harness.GetResourcesByName("repositories")
+			Expect(err).To(HaveOccurred(), "A user without a role should not be able to list repositories")
+		})
+	})
+})
+
+func changeK8sContext(harness *e2e.Harness, k8sContext string) {
```
siserafin: Please move this function to harness, we might re-use it

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,338 @@
...
+	cmd := exec.Command("oc", "config", "use-context", k8sContext)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		GinkgoWriter.Printf("❌ Failed to change context to %s: %v\n", k8sContext, err)
+	} else {
+		GinkgoWriter.Printf("✅ Changed context to %s: %s\n", k8sContext, output)
+	}
+}
+
+func deleteRoleBinding(ctx context.Context, client kubernetes.Interface, namespace string, roleBindingName string) error {
```
siserafin: All these role-related functions might be re-used. I think we could move these to a file test/harness/e2e/harness_rbac.go

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,338 @@
...
+		err := deleteRoleBinding(suiteCtx, harness.Cluster, flightCtlNs, roleBinding)
+		if err != nil {
+			logrus.Errorf("Failed to delete role binding %s: %v", roleBinding, err)
+		} else {
+			GinkgoWriter.Printf("Deleted role binding %s", roleBinding)
+		}
+	}
+}
+
+func createResource(harness *e2e.Harness, resourceType string, flightCtlResources []flightCtlResource) (string, []byte, error) {
```
siserafin: please move this function to harness

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,338 @@
...
+	changeK8sContext(harness, k8sContext)
+	cmd := exec.Command("bash", "-c", "oc whoami --show-server")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("Failed to get Kubernetes API endpoint: %v", err)
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+func loginAsNonAdmin(harness *e2e.Harness, user string, password string, k8sContext string, k8sApiEndpoint string) error {
```
siserafin: this function can be moved to login.go

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,338 @@
...
+			return "", nil, err
+		}
+		return *resourceName, resourceData, nil
+	} else {
+		GinkgoWriter.Printf("Apply output: %s\n", applyOutput)
+		return "", nil, fmt.Errorf("Failed to create a %s", resourceType)
+	}
+}
+
+func getDefaultK8sContext() (string, error) {
```
siserafin: same comment here

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,338 @@
...
+	for _, context := range contexts {
+		if strings.Contains(context, "default") {
+			GinkgoWriter.Printf("🔍 [DEBUG] Found default context: %s\n", context)
+			return context, nil
+		}
+	}
+	return "", fmt.Errorf("no context with 'default' in name found")
+}
+
+func getK8sApiEndpoint(harness *e2e.Harness, k8sContext string) (string, error) {
```
siserafin: same comment here

test/e2e/rbac/rbac_suite_test.go
```diff
@@ -0,0 +1,81 @@
...
+var (
+	flightCtlNs string
+)
+
+var _ = BeforeSuite(func() {
+	_, _, err := e2e.SetupWorkerHarness()
+	Expect(err).ToNot(HaveOccurred())
+
+	// Check if ACM is installed before running any tests
+	isAcmInstalled, err := util.IsAcmInstalled()
```
siserafin, line 27: I think this test would work also in OCP + standalone. Maybe we can just check the context.

itzikb-redhat: As it is right now we don't have user demouser1 in standalone

test/e2e/rbac/rbac_test.go
itzikb-redhat: Done

test/e2e/rbac/rbac_test.go
itzikb-redhat: Done

test/e2e/rbac/rbac_test.go
itzikb-redhat: Done

test/e2e/rbac/rbac_test.go
itzikb-redhat: Done

test/e2e/rbac/rbac_test.go
itzikb-redhat: Done

itzikb-redhat:
> So far it looks good! Is this tested in OCP?

It was tested locally with ACM.
One point:
A VM is created as part of the setup - NewTestHarnessWithoutVM is going to be removed and I haven't found another way

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,143 @@
...
+			Name:     adminRoleName,
+		}
+		adminRoleBinding.Subjects = []rbacv1.Subject{
+			{
+				Kind: "User",
+				Name: nonAdminUser,
+			},
+		}
+
+		It("should have full access to all resources and operations", Label("sanity", "83842"), func() {
```
rawagner: I'd make the test a bit more comprehensive by:
1. running the test scenario without role/rolebinding - and expect it to fail with 403
2. running the test again with role/rolebinding created (as you do now)
3. create the role/rolebinding in different NS and expect it to fail with 403
4. create the clusterrole/clusterolebinding and expect it to pass

rawagner (approved):

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,293 @@
...
+			Expect(err).ToNot(HaveOccurred(), "Admin should be able to delete cluster role")
+			err = harness.DeleteClusterRoleBinding(suiteCtx, harness.Cluster, createdAdminClusterRoleBinding.Name)
+			Expect(err).ToNot(HaveOccurred(), "Admin should be able to delete cluster role binding")
+		})
+	})
+})
+
+// testResourceOperations tests all CRUD operations for the given resource types
+// shouldSucceed determines whether the operations are expected to succeed or fail
+func testResourceOperations(ctx context.Context, harness *e2e.Harness, resourceTypes []string, shouldSucceed bool, testLabels *map[string]string, namespace string, operations []string) {
```
siserafin: Nice! I think this function should return output and error, and validate with Expect from the test.

siserafin: Maybe consider to move it to harness. I guess we can re-use it.

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,293 @@
...
+				}
+			}
+		}
+	}
+}
+
+// testReadOnlyResourceOperations tests read-only operations for the given resource types
+// for the given resource types
+// shouldSucceed determines whether the operations are expected to succeed or fail
+func testReadOnlyResourceOperations(harness *e2e.Harness, resourceTypes []string, shouldSucceed bool) {
```
siserafin: Same comment here....I would move the Expect to the test case.

test/harness/e2e/harness.go
```diff
@@ -1928,3 +1964,107 @@ func (h *Harness) CreateGitRepositoryWithContent(repoName, filePath, content str
...
+	if err != nil {
+		GinkgoWriter.Printf("❌ Failed to change context to %s: %v\n", k8sContext, err)
+		return fmt.Errorf("failed to change context to %s: %w", k8sContext, err)
+	} else {
+		GinkgoWriter.Printf("✅ Changed context to %s: %s\n", k8sContext, output)
+		return nil
+	}
+}
+
+func (h *Harness) CreateResource(resourceType string) (string, string, []byte, error) {
```
siserafin, line 1998: Maybe I would call this function apply resource, since there are different way to create resources, here you are using the client

test/harness/e2e/harness.go
```diff
@@ -1928,3 +1964,107 @@ func (h *Harness) CreateGitRepositoryWithContent(repoName, filePath, content str
...
+	}
+	return "", fmt.Errorf("no context with 'default' in name found")
+}
+
+// GetK8sApiEndpoint returns the API endpoint for a given K8s context
+func (h *Harness) GetK8sApiEndpoint(k8sContext string) (string, error) {
+	if err := h.ChangeK8sContext(k8sContext); err != nil {
+		return "", fmt.Errorf("failed to change K8s context to %s: %w", k8sContext, err)
+	}
+	cmd := exec.Command("bash", "-c", "oc whoami --show-server")
```
siserafin, line 2084: I would add a check that oc client is installed in this function.

test/e2e/rbac/rbac_test.go
```diff
@@ -0,0 +1,293 @@
...
+			Name:     adminRoleName,
+		}
+		adminClusterRoleBinding.Subjects = []rbacv1.Subject{
+			{
+				Kind: "User",
+				Name: nonAdminUser,
+			},
+		}
+
+		It("should have access full access with an admin role", Label("sanity", "83842"), func() {
```
siserafin: Not sure If we should add the sanity label. This test is going to be skipped in the pre-PR checks, since ACM is not part of the env.

test/e2e/rbac/rbac_test.go
itzikb-redhat: Moved

test/e2e/rbac/rbac_test.go
itzikb-redhat: Moved

test/harness/e2e/harness.go
itzikb-redhat: Done

PR #1568, opened 2025-08-26
EDM-2075: added resume api and cli

asafbennatan:
## Summary by CodeRabbit

- **New Features**
  - Added POST /api/v1/deviceactions/resume (single & bulk) with resumedDevices response; flightctl resume CLI; client/server HTTP support; GET /api/v1/devices/{name}/rendered accepts knownRenderedVersion; new DeviceConflictResolved event reason and details.

- **Tests**
  - Added unit, integration, CLI and e2e tests for resume flows, selectors, idempotence, event filtering, and error cases.

- **Chores**
  - Build now includes an extra flightctl binary; added System/SystemComponentDB constants.


api/v1alpha1/openapi.yaml
```diff
@@ -3552,6 +3657,53 @@ components:
       required:
         - target
       description: Metadata about a device decommissioning request.
+    DeviceResumeRequest:
+      type: object
+      properties:
+        labelSelector:
```
avishayt, line 3613: Also allow fieldSelector

api/v1alpha1/openapi.yaml
```diff
@@ -1456,6 +1456,62 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/devices/{name}/resume:
```
avishayt: If we have the resumeDevices API then we don't need this one too.  Better to avoid multiple ways of doing something.  The CLI for a single device can call the resumeDevices API with an appropriate fieldSelector to choose a single device.

api/v1alpha1/openapi.yaml
```diff
@@ -1521,6 +1577,55 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/devices/resume:
```
avishayt: This conflicts with /api/v1/devices/{name}.

I recommend POST /api/v1/deviceactions/resume

api/v1alpha1/openapi.yaml
```diff
@@ -3552,6 +3657,53 @@ components:
...
+    DeviceResumeRequest:
+      type: object
+      properties:
+        labelSelector:
+          type: string
+          description: A selector to restrict the list of devices to resume by their labels. Uses the same format as Kubernetes label selectors (e.g., "key1=value1,key2!=value2").
+      required:
+        - labelSelector
+      description: Request to resume devices based on label selector.
+    DeviceResumeResponse:
```
avishayt, line 3628: This return value is complicated and can be huge if applied to many devices.  It also seems like it's not all implemented (e.g., failedDevices).  It's a single DB query, so either the whole thing should pass or fail - no partial success.

internal/service/device.go
```diff
@@ -481,6 +482,32 @@ func (h *ServiceHandler) UpdateServiceSideDeviceStatus(ctx context.Context, devi
...
+	}
+
+	// Remove conflictPaused annotation from all matching devices in a single SQL query
+	resumedCount, err := h.store.Device().RemoveConflictPausedAnnotation(ctx, orgId, *listParams)
+	if err != nil {
+		return 0, api.StatusInternalServerError(fmt.Sprintf("failed to resume devices: %v", err))
+	}
+
+	return int(resumedCount), api.StatusOK()
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/cli/resume_test.go
```diff
@@ -0,0 +1,297 @@
...
+	if len(name) == 0 {
+		return fmt.Errorf("specify a specific device to resume")
+	}
+
+	if opts.LabelSelector != "" {
+		return fmt.Errorf("label selector cannot be used when resuming a specific device")
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

api/v1alpha1/openapi.yaml
asafbennatan: added

api/v1alpha1/openapi.yaml
asafbennatan: done

api/v1alpha1/openapi.yaml
asafbennatan: done

api/v1alpha1/openapi.yaml
asafbennatan: done - returning the count of updated devices only

internal/api/client/client.gen.go
```diff
@@ -7558,6 +7676,67 @@ func ParseUpdateCertificateSigningRequestApprovalResponse(rsp *http.Response) (*
...
+		if err := json.Unmarshal(bodyBytes, &dest); err != nil {
+			return nil, err
+		}
+		response.JSON503 = &dest
+
+	}
+
+	return response, nil
+}
+
```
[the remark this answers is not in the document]
asafbennatan: this is not a pattern we currently have in the codebase

api/v1alpha1/types.gen.go
```diff
@@ -801,6 +801,22 @@ type DeviceResourceStatus struct {
...
 
+// DeviceResumeRequest Request to resume devices based on label selector and/or field selector. At least one selector must be provided.
+type DeviceResumeRequest struct {
+	// FieldSelector A selector to restrict the list of devices to resume by their fields. Uses the same format as Kubernetes field selectors (e.g., "metadata.name=device1,status.phase!=Pending").
+	FieldSelector *string `json:"fieldSelector,omitempty"`
+
+	// LabelSelector A selector to restrict the list of devices to resume by their labels. Uses the same format as Kubernetes label selectors (e.g., "key1=value1,key2!=value2").
+	LabelSelector *string `json:"labelSelector,omitempty"`
+	union         json.RawMessage
+}
```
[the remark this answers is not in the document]
asafbennatan: we do that already

internal/cli/resume.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (o *ResumeOptions) Validate(args []string) error {
+	if err := o.GlobalOptions.Validate(args); err != nil {
+		return err
+	}
+
+	arg := args[0]
+
+	// Handle bulk resume case
+	if arg == "devices" {
```
avishayt: Please make consistent with other CLI commands.  We treat plural and singular the same.
Call parseAndValidateKindName here, and then if name != "" it's singular, otherwise bulk.

internal/cli/resume.go
```diff
@@ -0,0 +1,195 @@
...
+		if o.LabelSelector == "" {
```
avishayt: FieldSelector is also valid.
But is there any reason to enforce this?  Maybe they just want to unpause all devices?

internal/cli/resume.go
```diff
@@ -0,0 +1,195 @@
...
+func (o *ResumeOptions) Run(ctx context.Context, args []string) error {
+	c, err := o.BuildClient()
+	if err != nil {
+		return fmt.Errorf("creating client: %w", err)
+	}
+
+	arg := args[0]
+
+	// Handle bulk resume case
+	if arg == "devices" {
```
avishayt: Also here, call parseAndValidateKindName and check if name != ""

internal/cli/resume.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if response.HTTPResponse != nil {
+		switch response.HTTPResponse.StatusCode {
+		case http.StatusOK:
+			if response.JSON200 != nil {
+				count := *response.JSON200
+				if count > 0 {
+					fmt.Printf("Device resumed successfully: %s\n", name)
+				} else {
+					return fmt.Errorf("device not found or not in conflictPaused state: %s", name)
```
avishayt: The call should be idempotent, like delete.  If it's not in conflictPaused state then we should return 200.

You should also match the success message to delete:
"Resume request for Device \"%s\" completed"

internal/cli/resume.go
```diff
@@ -0,0 +1,195 @@
...
+		switch response.HTTPResponse.StatusCode {
+		case http.StatusOK:
+			if response.JSON200 != nil {
+				count := *response.JSON200
+				fmt.Printf("Resume operation completed:\n")
+				fmt.Printf("  Devices resumed: %d\n", count)
+
+				if count == 0 {
+					fmt.Printf("No devices matched the selector or were in conflictPaused state\n")
+				}
```
avishayt, line 221: Similar to above

internal/service/common/events.go
```diff
@@ -572,11 +572,11 @@ func GetReferencedRepositoryUpdatedEvent(ctx context.Context, kind api.ResourceK
...
 
 // GetSystemRestoredEvent creates an event for system restoration completion
-// Following the pattern of InternalTaskFailed events, this associates the event with a system-level resource
+// Associates the event with a system-level resource using the System kind
 func GetSystemRestoredEvent(ctx context.Context, devicesUpdated int64) *api.Event {
 	return getBaseEvent(ctx, resourceEvent{
-		resourceKind: api.FleetKind,
-		resourceName: api.FlightCtlSystemResourceName,
+		resourceKind: api.SystemKind,
+		resourceName: api.SystemKind,
```
avishayt: Database

internal/service/device.go
```diff
@@ -519,6 +521,50 @@ func (h *ServiceHandler) UpdateServiceSideDeviceStatus(ctx context.Context, devi
...
+		trimmed := strings.TrimSpace(*request.FieldSelector)
+		if trimmed != "" {
+			normalizedFieldSelector = &trimmed
+		}
+	}
+
+	// Validate that at least one selector is provided after normalization
+	if normalizedLabelSelector == nil && normalizedFieldSelector == nil {
+		return 0, api.StatusBadRequest("at least one non-empty selector (labelSelector or fieldSelector) must be provided to avoid accidental resume all")
+	}
```
avishayt: Like I said above, I understand the intention here, but I'm not sure we want to enforce it.
Maybe we should add a "--all" flag in the CLI and enforce it only there?
Please ask Paola (UX).

internal/service/device.go
```diff
@@ -519,6 +521,50 @@ func (h *ServiceHandler) UpdateServiceSideDeviceStatus(ctx context.Context, devi
...
+			normalizedLabelSelector = &trimmed
+		}
+	}
+
+	if request.FieldSelector != nil {
+		trimmed := strings.TrimSpace(*request.FieldSelector)
+		if trimmed != "" {
+			normalizedFieldSelector = &trimmed
+		}
+	}
```
avishayt: Is this necessary?  I don't see that we do it anywhere else.

internal/service/device.go
```diff
@@ -519,6 +521,50 @@ func (h *ServiceHandler) UpdateServiceSideDeviceStatus(ctx context.Context, devi
...
+	// Create list params with both label and field selectors
+	listParams, status := prepareListParams(nil, normalizedLabelSelector, normalizedFieldSelector, nil)
+	if status.Code != http.StatusOK {
+		return 0, status
+	}
+
+	// Remove conflictPaused annotation from all matching devices in a single SQL query
+	resumedCount, err := h.store.Device().RemoveConflictPausedAnnotation(ctx, orgId, lo.FromPtr(listParams))
+	if err != nil {
+		return 0, api.StatusInternalServerError(fmt.Sprintf("failed to resume devices: %v", err))
```
avishayt: I think you need to check selector.AsSelectorError?

internal/service/device.go
```diff
@@ -519,6 +521,50 @@ func (h *ServiceHandler) UpdateServiceSideDeviceStatus(ctx context.Context, devi
 	return common.UpdateServiceSideStatus(ctx, orgId, &device, h.store, h.log)
 }
 
+func (h *ServiceHandler) ResumeDevices(ctx context.Context, request api.DeviceResumeRequest) (int, api.Status) {
```
celdrake: Few questions about Resuming devices.

1. What happens when the request would match both devices in COnflictPaused or other statuses? Are the non-paused devices just skipped?
2. It appears like you are returning a plain text with the result (based on the examples from the tests). 
IMO it would be better to return a typed response where the number of resumed devices is on a separate field, for easier consumption on the clients.

api/v1alpha1/openapi.yaml
```diff
@@ -1523,6 +1525,56 @@ paths:
               schema:
                 $ref: '#/components/schemas/Status'
 
+  /api/v1/deviceactions/resume:
+    post:
+      tags:
+        - deviceactions
+      description: Resume devices based on label selector and/or field selector.
```
celdrake, line 1532: IIUC, via the API it's not possible to explicitly resume a single device, and it needs to be done via the "field-selector=metadata.name=x".

internal/cli/resume.go
asafbennatan: done

internal/cli/resume.go
asafbennatan: done

internal/cli/resume.go
asafbennatan: done ,added --all flag for cli

internal/cli/resume.go
asafbennatan: done

internal/service/device.go
asafbennatan: removed due to all usecase

internal/service/device.go
asafbennatan: fixed

internal/service/device.go
asafbennatan: 1. yes we are only taking paused devices
2. we are returning the count of devices changed , will adjust the examples

api/v1alpha1/openapi.yaml
asafbennatan: yes this is the intended behaviour

internal/service/device.go
asafbennatan: added all flag based on our talk

internal/cli/resume.go
asafbennatan: what about bulk operation resposewhat about bulk operation respose
we dont have any existing bulk operation to my understanding
so we dont have an example to go with
currently resume bulk cli succeeds regardless but just prints an extra line (not error) saying that no devices matched

api/v1alpha1/consts.go
```diff
@@ -87,6 +87,9 @@ const (
 	OrganizationAPIVersion = "v1alpha1"
 	OrganizationKind       = "Organization"
 	OrganizationListKind   = "OrganizationList"
+
+	SystemKind        = "System"
+	SystemComponentDB = "db"
```
avishayt:
```suggestion
	SystemComponentDB = "database"
```

api/v1alpha1/openapi.yaml
```diff
@@ -1523,6 +1525,57 @@ paths:
...
+        required: true
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: integer
+                description: Number of devices that were successfully resumed
+              example: 3
```
avishayt: Make it an object so that it's extensible

internal/cli/resume.go
avishayt: OK

api/v1alpha1/openapi.yaml

api/v1alpha1/consts.go
asafbennatan: fixed

internal/cli/resume.go
```diff
@@ -0,0 +1,231 @@
...
+func NewCmdResume() *cobra.Command {
+	o := DefaultResumeOptions()
+	cmd := &cobra.Command{
+		Use:   "resume (device/NAME | device | devices)",
+		Short: "Resume a device or devices based on selectors or all devices.",
+		Long: `Resume devices that are in a conflictPaused state.
+
+Examples:
+  # Resume a specific device
+  flightctl resume device/my-device
```
celdrake, line 39: Resume for an individual device which doesn't exist does not return an error and it should return 404.

```
➜ flightctl get dev/f
Error: reading device/f: response status: 404, message: Device of name "f" not found
```

```
❯ flightctl resume dev/f
Resume request for device "f" completed
```

internal/cli/resume.go
```diff
@@ -0,0 +1,231 @@
...
+	if response.HTTPResponse != nil {
+		switch response.HTTPResponse.StatusCode {
+		case http.StatusOK:
+			if response.JSON200 != nil {
+				count := response.JSON200.ResumedDevices
+				fmt.Printf("Resume operation completed:\n")
+				fmt.Printf("  Devices resumed: %d\n", count)
+
+				if count == 0 {
+					fmt.Printf("No devices matched the selector or were in conflictPaused state\n")
```
celdrake, line 220: Should be `ConflictPaused`.
And this new statuses (`ConflictPaused` and `AwaitingReconnect`) are missing in the user documentation. 
https://github.com/flightctl/flightctl/blob/main/docs/user/device-api-statuses.md?plain=1#L9

asafbennatan: added in the new PR we have for docs

internal/cli/resume.go
asafbennatan: see this:
https://github.com/flightctl/flightctl/pull/1620

celdrake: As I mentioned in that PR, I think the behaviour is inconsistent with the rest of the CLI commands.
I'd expect an error 404 when the device does not exist vs statusCode 200 when it does exist and it could not be resumed (eg. it already was)

PR #1575, opened 2025-08-26
EDM-1961: Guided Onboarding for FlightCtl CLI

SiddarthR56:
- Extend /auth/config to return the external FlightCtl service URL, enabling the UI to generate accurate CLI login commands.

- Enhanced user experience when encountering TLS certificate issues during flightctl login by translating cryptic errors into user-friendly messages with specific resolution guidance. The CLI now provides clear instructions for common certificate scenarios (unknown authority, self-signed, expired, hostname mismatch) and includes dedicated OAuth certificate handling with the --auth-certificate-authority flag.


docs/user/getting-started.md
```diff
@@ -249,12 +249,53 @@ flightctl login ${FC_API_URL} -t $(oc whoami -t)
...
-* add the path for the CA certificate via the `--certificate-authority=<path_to_ca_crt>` flag to your command
-* or bypass the server verification (insecure!) by adding the `--insecure-skip-tls-verify` flag to your command
+#### Common Certificate Issues
+
+#### Certificate Not Trusted
+
+If you see "Cause: certificate not trusted", the server uses a custom certificate authority:
+
+```console
+flightctl login ${FC_API_URL} --certificate-authority=/path/to/ca.crt
```
celdrake, line 265: These examples are missing the required paramters token/username+password/web.
Maybe add something like 
```suggestion
flightctl login ${FC_API_URL} <loginParams/options> --certificate-authority=/path/to/ca.crt
```

docs/user/getting-started.md
```diff
@@ -249,12 +249,53 @@ flightctl login ${FC_API_URL} -t $(oc whoami -t)
...
+```
+
+#### Self-Signed Certificates
+
+For self-signed certificates, you can either:
+
+```console
+flightctl login ${FC_API_URL} --insecure-skip-tls-verify
```
celdrake, line 273: Same here and the rest of the cases.

internal/cli/login.go
```diff
@@ -18,6 +19,118 @@ import (
...
+}
+
+// formatTLSErrorForGeneral formats TLS errors for general API endpoints
+func formatTLSErrorForGeneral(errorInfo TLSErrorInfo) string {
+	cause := fmt.Sprintf("Cause: %s", errorInfo.Cause)
+
+	switch errorInfo.Type {
+	case TLSErrorUnknownAuthority:
+		return cause + "\n" +
+			"To resolve this issue, you can:\n" +
```
celdrake: Maybe best to check with UX experts, but to me this reads more like an "AND" rather than an "OR".

Users may think that they need to follow all the steps, and the first one may have them confused as they may not know where to obtain the certificates from.

internal/cli/login.go
```diff
@@ -222,6 +336,14 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
 	if token == "" {
 		authInfo, err := o.authProvider.Auth(o.Web, o.Username, o.Password)
 		if err != nil {
+			// Check if this is an OAuth certificate issue
+			if o.authConfig != nil && o.authConfig.AuthURL != "" {
```
celdrake: I attempt to login using the following command `flightctl login https://192.168.1.142.nip.io:3443 --token=$token`
The API url is correct, and the problem is with the certificate check. Adding "--insecure-skip-tls-verify" allows me to login correctly.

The CLI response is:
```
Error: failed to get auth info: failed to get auth info:
Cause: certificate hostname mismatch
To resolve this issue, you can:
  1. Check if the URL is correct
  2. --insecure-skip-tls-verify       (skip certificate verification)
```


Compare to how the OC cli handles the same situation. It clearly identifies that the issue is with the cerficate check, and it suggests if the user wants to skip it while warning them of the security issues. And it doesn't mention that the problem could be with the URL, since that's actually correct.

```
oc login --token=<token> --server=<some-server>:6443
The server uses a certificate signed by an unknown authority.
You can bypass the certificate check, but any data you send to the server could be intercepted by others.
Use insecure connections? (y/n): 

```

internal/cli/login.go
```diff
@@ -18,6 +19,118 @@ import (
...
+		return cause + "\n" +
+			"To resolve this issue, you can:\n" +
+			"  1. --certificate-authority=<path>   (provide the CA certificate file)\n" +
+			"  2. --insecure-skip-tls-verify       (skip certificate verification)"
+
+	case TLSErrorExpired:
+		return cause + "\n" +
+			"To resolve this issue, you can:\n" +
+			"  1. Contact your administrator to renew the certificate\n" +
+			"  2. --insecure-skip-tls-verify       (skip certificate verification)"
```
celdrake: There is no warning to the user about the consequences of skipping certificate verification.
Again, the OC cli states it clearly "You can bypass the certificate check, but any data you send to the server could be intercepted by others."

celdrake (changes requested):

internal/auth/authn/aapgateway.go
```diff
@@ -93,8 +95,9 @@ func (a AapGatewayAuth) ValidateToken(ctx context.Context, token string) error {
 
 func (a AapGatewayAuth) GetAuthConfig() common.AuthConfig {
 	return common.AuthConfig{
-		Type: common.AuthTypeAAP,
-		Url:  a.externalGatewayUrl,
+		Type:       common.AuthTypeAAP,
+		Url:        a.externalGatewayUrl,
+		ServiceUrl: a.serviceUrl,
 	}
```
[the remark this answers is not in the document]
celdrake: FYI the UI is using AAPGateway.
https://github.com/flightctl/flightctl-ui/blob/main/proxy/auth/auth.go#L44

api/v1alpha1/openapi.yaml
```diff
@@ -5187,11 +5187,15 @@ components:
...
           description: Auth URL.
+        serviceUrl:
+          type: string
+          description: Service URL.
         authOrganizationsConfig:
           $ref: '#/components/schemas/AuthOrganizationsConfig'
       required:
         - authType
         - authURL
+        - serviceUrl
```
fzdarsky: Why this change to the API? To be able to fetch this value, the client already needs to know the service URL??

celdrake: The client might have an internal URL which cannot be used externally.
This is the case for the UI, where we have the internal API URL, and for the login command we need to obtain the external URL.

internal/cli/login.go
```diff
@@ -18,6 +19,118 @@ import (
...
+			Cause:    "certificate hostname mismatch",
+			RawError: errStr,
+		}
+	default:
+		return TLSErrorInfo{
+			Type:     TLSErrorGeneric,
+			Cause:    "certificate verification failed",
+			RawError: errStr,
+		}
+	}
```
rawagner: instead of looking at the err msg, lets check if the err is of `x509` error
```
var uaErr x509.UnknownAuthorityError
	if errors.As(err, &uaErr) {
		return TLSErrorInfo{
			Type:     TLSErrorUnknownAuthority,
			Cause:    "certificate not trusted",
			RawError: errStr,
		}
	}
```

same for other `x509` errors

also the cases `case strings.Contains(errStr, tlsUnknownAuthority):` and `case strings.Contains(errStr, tlsSelfSigned):` are the same

rawagner (changes requested):

api/v1alpha1/openapi.yaml
fzdarsky: I see, thanks. So we're talking about service endpoint discovery. Feels a bit strange to repurpose the `/auth` endpoint for this instead of, say, using a dedicated discovery endpoint (e.g. a `/config`) or passing the external service URL to the UI proxy as configuration. I don't have a concrete other use case for a `/config` in mind, though, so hard to argue for it. Just noting that the proposed change will break users again...

internal/cli/login.go
celdrake: The new wording is much clearer, thanks.

internal/cli/login.go
celdrake: The new way of handling the "insecure-skip-tls-verify" as the "oc" CLI does is much clearer and user friendly.

docs/user/getting-started.md
```diff
@@ -249,12 +249,53 @@ flightctl login ${FC_API_URL} -t $(oc whoami -t)
 
 > 📌 You can also login with your ACM login credentials using the `--web` or `--username` and `--password` flags
 
-### Self-signed certificates
+### Certificate Configuration and Troubleshooting
 
-The CLI uses the host's certificate authority (CA) pool to verify the Flight Control service's identity. When using self-signed certificates, this can lead to a TLS verification error if you have not added your CA's certificate to that pool. You can:
+The CLI uses the host's certificate authority (CA) pool to verify the Flight Control service's identity. When certificate verification fails, the CLI provides user-friendly error messages with specific guidance for resolution.
```
celdrake: With the latest changes, I only noticed a small problem if I attempt to login using the agent-api URL instead of the API url.  IMO it's a minor issue, but maybe we can detect it?

Notice I'm using the "insecure-skip-tls-verify" flag, and the CLI suggests adding it.
The actual issue is that I'm using por 7443 rather than 3443.

```
 flightctl login  https://api.192.168.1.131.nip.io:7443 --username demouser --password <password> --insecure-skip-tls-verify 
Error: failed to get auth info: failed to get auth info:
Cause: certificate verification failed (Get "https://api.192.168.1.131.nip.io:7443/api/v1/auth/config": remote error: tls: certificate required)
Options (choose one):
  1. Provide a trusted CA certificate (recommended)
     flightctl login <API_URL> <login options> --certificate-authority=/path/to/ca.crt
  2. Skip certificate verification (not recommended)
     flightctl login <API_URL> <login options> --insecure-skip-tls-verify
     WARNING: Skipping certificate verification makes your HTTPS connection insecure and could expose your credentials and data to interception.
```

deploy/helm/flightctl/charts/ui/templates/flightctl-ui-cm.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 data:
   FLIGHTCTL_SERVER: {{ .Values.api.url | quote }}
+  FLIGHTCTL_EXTERNAL_URL: '{{ if eq (include "flightctl.getServiceExposeMethod" .) "nodePort" }}https://api.{{ include "flightctl.getBaseDomain" . }}:{{ .Values.global.nodePorts.api }}/{{ else if and (eq (include "flightctl.getServiceExposeMethod" .) "gateway") (not (eq (int .Values.global.gatewayPorts.tls) 443)) }}https://api.{{ include "flightctl.getBaseDomain" . }}:{{ .Values.global.gatewayPorts.tls }}/{{ else }}https://api.{{ include "flightctl.getBaseDomain" . }}/{{ end }}'
```
rawagner: We already have a helper to get the API URL https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/templates/_helpers.tpl#L83 - lets use it.

internal/cli/login.go
```diff
@@ -2,6 +2,8 @@ package cli
```
celdrake, line 2: (Comment related to general login flow)-

I deployed flightctl with the following command:
`make clean && AUTH=true ORGS=true make deploy`

Then I logged in, it worked well.

But when I run the `make agent-vm` command, it fails to complete.
I believe the new interactive prompts are preventing the login command to succeed in this script.

deploy/helm/flightctl/charts/ui/templates/flightctl-ui-cm.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     {{- include "flightctl.standardLabels" . | nindent 4 }}
 data:
   FLIGHTCTL_SERVER: {{ .Values.api.url | quote }}
+  FLIGHTCTL_EXTERNAL_URL: {{ include "flightctl.getApiUrl" . }}
```
celdrake: I'd rename so it's clear that it's the same as the FLIGHTCTL_SERVER, just external.

```suggestion
  FLIGHTCTL_SERVER_EXTERNAL: {{ include "flightctl.getApiUrl" . }}
```

internal/cli/login.go
celdrake: I checked again with the latest branch and I couldn't see any issue.

internal/cli/login.go
```diff
@@ -342,6 +585,86 @@ func (o *LoginOptions) getClientConfig(apiUrl string) (*client.Config, error) {
...
+	_, _ = fmt.Scanln(&resp)
+	resp = strings.TrimSpace(strings.ToLower(resp))
+	return resp == "y" || resp == "yes"
+}
+
+// promptUseInsecureForOAuth shows a warning specific to OAuth endpoints
+func (o *LoginOptions) promptUseInsecureForOAuth(errorInfo TLSErrorInfo) bool {
+	fmt.Println("The authentication server's certificate could not be verified (" + errorInfo.Cause + ").")
+	fmt.Println("You can bypass the certificate check, but any data you send to the server could be intercepted by others.")
+	fmt.Print("Use insecure connections for authentication? (y/N): ")
```
rawagner: in the end, we configure insecure connection for the API server as well. So the msg should be just 
```suggestion
	fmt.Print("Use insecure connections? (y/N): ")
```

internal/cli/login.go
```diff
@@ -342,6 +585,86 @@ func (o *LoginOptions) getClientConfig(apiUrl string) (*client.Config, error) {
...
+	var resp string
+	_, _ = fmt.Scanln(&resp)
+	resp = strings.TrimSpace(strings.ToLower(resp))
+	return resp == "y" || resp == "yes"
```
rawagner: Should we make a separate, reusable function out it these lines, which are shared between `promptUseInsecureForOAuth` and `promptUseInsecureForGeneral` ?

internal/cli/login.go
```diff
@@ -342,6 +585,86 @@ func (o *LoginOptions) getClientConfig(apiUrl string) (*client.Config, error) {
...
+	fmt.Println("You can bypass the certificate check, but any data you send to the server could be intercepted by others.")
+	fmt.Print("Use insecure connections? (y/N): ")
+	var resp string
+	_, _ = fmt.Scanln(&resp)
+	resp = strings.TrimSpace(strings.ToLower(resp))
+	return resp == "y" || resp == "yes"
+}
+
+// promptUseInsecureForOAuth shows a warning specific to OAuth endpoints
+func (o *LoginOptions) promptUseInsecureForOAuth(errorInfo TLSErrorInfo) bool {
```
rawagner: This is generically for Auth, not just OAuth, right ?

```suggestion
// promptUseInsecureForAuth shows a warning specific to Auth endpoints
func (o *LoginOptions) promptUseInsecureForAuth(errorInfo TLSErrorInfo) bool {
```

internal/cli/login.go
```diff
@@ -144,6 +272,141 @@ func (o *LoginOptions) Validate(args []string) error {
...
+	o.clientConfig.AuthInfo.AuthProvider = &client.AuthProviderConfig{
+		Name: o.authConfig.AuthType,
+		Config: map[string]string{
+			client.AuthUrlKey:      o.authConfig.AuthURL,
+			client.AuthClientIdKey: o.ClientId,
+		},
+	}
+}
+
+// fetchToken retrieves an access token, handling OAuth TLS prompts and retries as needed
```
rawagner: not just OAuth
```suggestion
// fetchToken retrieves an access token, handling Auth TLS prompts and retries as needed
```

internal/cli/login.go
```diff
@@ -144,6 +272,141 @@ func (o *LoginOptions) Validate(args []string) error {
...
+func (o *LoginOptions) fetchToken() (string, error) {
+	if o.AccessToken != "" {
+		return o.AccessToken, nil
+	}
+
+	authInfo, err := o.authProvider.Auth(o.Web, o.Username, o.Password)
+	if err != nil {
+		// Check if this is an OAuth certificate issue
```
rawagner:
```suggestion
		// Check if this is an Auth certificate issue
```

internal/cli/login.go
```diff
@@ -144,6 +272,141 @@ func (o *LoginOptions) Validate(args []string) error {
...
+					// enable insecure and recreate provider, then retry once
+					o.enableInsecure()
+					if err := o.recreateAuthProvider(); err != nil {
+						return "", fmt.Errorf("OAuth authentication failed\n%s", formatTLSErrorForOAuth(errorInfo))
+					}
+					authInfo, err = o.authProvider.Auth(o.Web, o.Username, o.Password)
+				}
+				if err != nil {
+					oauthErrMsg := formatTLSErrorForOAuth(errorInfo)
+					return "", fmt.Errorf("OAuth authentication failed\n%s", oauthErrMsg)
```
rawagner:
```suggestion
					return "", fmt.Errorf("authentication failed\n%s", oauthErrMsg)
```

rawagner: ..well, you get the idea..any OAuth mention should be just auth :)

internal/auth/auth.go
```diff
@@ -44,7 +44,7 @@ const (
 	AuthTypeNil  AuthType = "nil"
 	AuthTypeK8s  AuthType = "k8s"
 	AuthTypeOIDC AuthType = "oidc"
-	AuthTypeAAP  AuthType = "aap"
+	AuthTypeAAP  AuthType = "AAPGateway"
```
rawagner: why the change ?

celdrake: I believe the reason is to align with the value being checked in the UI https://github.com/flightctl/flightctl-ui/blob/main/proxy/auth/auth.go#L44

Maybe we should change it instead? WDYT?

rawagner: UI is checking against these https://github.com/flightctl/flightctl/blob/main/internal/auth/common/common.go#L19

the `internal/auth/auth.go` consts are not reported anywhere

celdrake: Ah, I mixed those. Then the "aap" value could be kept here, thanks.

internal/cli/login.go
```diff
@@ -342,6 +585,70 @@ func (o *LoginOptions) getClientConfig(apiUrl string) (*client.Config, error) {
...
+}
+
+// enableInsecure flips the config to insecure and updates related fields
+func (o *LoginOptions) enableInsecure() {
+	o.InsecureSkipVerify = true
+	o.clientConfig.Service.InsecureSkipVerify = true
+}
+
+// recreateAuthProvider rebuilds the auth provider with the current insecurity setting
+func (o *LoginOptions) recreateAuthProvider() error {
```
rawagner: One more thing - this function is exactly the same as `initAuthProvider()`

rawagner (approved):

PR #1576, opened 2025-08-27
EDM-1911: Implement JWT specific identity and organization provider

DakCrowder:
This change implements a new JWT identity and multi org support by pulling org info from passed token claims. Changes are architected so future additions for AAP or other sources of external organizational membership info for users can be added as pluggable org providers.

Organizations are setup to be supplied as JWT claims like so:
``` 
"organization": {
    "pinkcorp": {
      "id": "a6e97659-16a5-4b18-9d90-7fe88a744e2a"
    },
    "orangecorp": {
      "id": "7ca05aab-652c-46a4-aef5-1093e573865c"
    }
  }
```

  Added Functionality / Overview:

- Identity updates
  - Single common identity interface added and base identity added that all auth providers implement
  - New JWT based identity added that extends a base identity with JWT claim info
- Organization providers added
  - ExternalOrganizationProvider interface added for pluggable organization sources
  - Right now the only used provider knows how to pull out organization info from above token claims / identity
  - AAP / K8s / other providers can be added here in the future
- Organization resolver expanded
  - Moved from a simple resolver to an interface with (for now) two concrete implementations
  - External resolver combines cached organization lookup with external organization knowledge / mappings
  - Default resolver acts as a shim that provides user access to the default org
  - An important note is organization models are created on the fly as we see new org identifiers

## Summary by CodeRabbit

* **New Features**
  * Multi-organization support: org-aware resolution with a TTL-backed org cache; APIs and CLI now respect organization scope.

* **Security / Auth**
  * Per-instance authentication & authorization: token/identity handling updated, JWT-based membership checks, org-aware permission enforcement, health-check bypass option.

* **Documentation**
  * Expanded Organizations guide with scoping rules, token-claim format, and examples.

* **Refactor**
  * Organization resolution and identity wiring converted to dependency-injected components.

* **Tests**
  * Tests and test utilities updated to exercise multi-org flows and resolver injection.


internal/auth/auth.go
```diff
@@ -47,6 +51,10 @@ func GetAuthN() AuthNMiddleware {
 	return authN
 }
 
+func GetOrgProvider() providers.ExternalOrganizationProvider {
```
kkyrazis: Do we need this? If so, should can we set this up in a way that doesn't require us to synchronize the type of provider when building the Resolver and when initializing auth?

internal/org/resolvers/external.go
```diff
@@ -0,0 +1,154 @@
...
+func (r *ExternalResolver) IsMemberOf(ctx context.Context, identity common.Identity, id uuid.UUID) (bool, error) {
+	org, err := r.getOrg(ctx, id)
+	if err != nil {
+		return false, err
+	}
+
+	return r.externalOrgProvider.IsMemberOf(ctx, identity, org.ExternalID)
+}
+
+func (r *ExternalResolver) GetUserOrganizations(ctx context.Context, identity common.Identity) ([]*model.Organization, error) {
```
kkyrazis, line 65: I'm probably missing something, but I'm a little confused by the overall flow. It appears that external -> internal org mapping/creation is done in this call, but this is only referenced from `service.ListOrganizations`. If this isn't called, regular API requests will fail as the org doesn't exist in the store. How are we ensuring that List Organizations is called?

internal/auth/auth.go
DakCrowder: Unused - removed

internal/org/resolvers/external.go
DakCrowder: Yep good question.  Because the api only ingests **internal** org identifiers (the primary id UUIDs from the org model we control) via the org_id api param, how do users or our clients know which identifiers they can send?  They need to be requested via the list organizations call, and that call becomes almost like an entrypoint that the cli/ui will need to fetch that info.

I think some actions can be taken to reduce friction specifically in the cli /ui (maybe we do this request after login and set an org in context automatically, especially if a user only has access to one org) but when multi org is configured its a natural starting point since they scope effectively all the resources the api surfaces.

internal/auth/auth.go
```diff
@@ -107,11 +109,11 @@ func getOrgConfig(cfg *config.Config) *common.AuthOrganizationsConfig {
 	}
 }
 
-func initOIDCAuth(cfg *config.Config, log logrus.FieldLogger) error {
+func initOIDCAuth(cfg *config.Config, log logrus.FieldLogger, orgResolver resolvers.Resolver) error {
 	oidcUrl := strings.TrimSuffix(cfg.Auth.OIDC.OIDCAuthority, "/")
 	externalOidcUrl := strings.TrimSuffix(cfg.Auth.OIDC.ExternalOIDCAuthority, "/")
 	log.Infof("OIDC auth enabled: %s", oidcUrl)
-	authZ = NilAuth{}
+	authZ = authz.NewJWTAuthZ(orgResolver)
```
hexfusion: I haven’t looked deeply yet, but would it be simpler to construct the authz and pass it as a parameter instead of plumbing orgResolver? Have I mentioned how much I dislike globals? Is there really no other way? I understand you didn’t introduce it.

internal/auth/authn/aapgateway.go
```diff
@@ -102,13 +102,11 @@ func (AapGatewayAuth) GetAuthToken(r *http.Request) (string, error) {
...
+func (a AapGatewayAuth) GetIdentity(ctx context.Context, token string) (common.Identity, error) {
 	userInfo, err := a.loadUserInfo(token)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get identity: %w", err)
 	}
 
-	return &common.Identity{
-		Username: userInfo.Username,
-	}, nil
+	return common.NewBaseIdentity(userInfo.Username, "", []string{}), nil
```
hexfusion, line 111: :+1:

internal/auth/authn/jwtverifier.go
```diff
@@ -58,17 +77,53 @@ func NewJWTAuth(oidcAuthority string, externalOIDCAuthority string, clientTlsCon
...
-	return err
+
+	parsedToken, err := jwt.Parse([]byte(token), jwt.WithKeySet(jwkSet), jwt.WithValidate(true))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse JWT token: %w", err)
+	}
+
+	identity := &JWTIdentity{
+		BaseIdentity: common.BaseIdentity{},
+	}
```
hexfusion: optional: since BaseIdentity is owned, you can do 

```suggestion
	identity := &JWTIdentity{}
```

internal/auth/authn/jwtverifier.go
```diff
@@ -58,17 +77,53 @@ func NewJWTAuth(oidcAuthority string, externalOIDCAuthority string, clientTlsCon
...
+	identity.parsedToken = parsedToken
+
+	if sub, exists := parsedToken.Get("sub"); exists {
+		if uid, ok := sub.(string); ok {
+			identity.SetUID(uid)
+		}
+	}
+
+	if preferredUsername, exists := parsedToken.Get("preferred_username"); exists {
```
hexfusion: nit: it's a good idea to move `preferred_username` to a const so I don't fat finger it next week.

internal/auth/common/common.go
```diff
@@ -31,18 +29,59 @@ type AuthOrganizationsConfig struct {
...
-	identityVal := ctx.Value(IdentityCtxKey)
+func GetIdentity(ctx context.Context) (Identity, error) {
+	identityVal := ctx.Value(consts.IdentityCtxKey)
 	if identityVal == nil {
 		return nil, fmt.Errorf("failed to get identity from context")
 	}
-	identity, ok := identityVal.(*Identity)
+	identity, ok := identityVal.(Identity)
 	if !ok {
 		return nil, fmt.Errorf("incorrect type of identity in context")
```
hexfusion: nice to let caller know what the type observed was
```suggestion
        return nil, fmt.Errorf("incorrect type of identity in context (got %T)", identityVal)
```

internal/flterrors/flterrors.go
```diff
@@ -41,4 +41,19 @@ var (
 
 	// certificate extensions
 	ErrExtensionNotFound = errors.New("certificate extension not found")
+
+	// authentication/authorization
+	ErrInvalidTokenClaims = errors.New("invalid token claims")
+	ErrMissingTokenClaims = errors.New("missing required token claims")
 )
+
+func IsClientAuthError(err error) bool {
```
hexfusion, line 50: nice

internal/org/providers/claims_provider.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return false, nil
+}
+
+// claimsFromIdentity returns a map of orgID -> orgName
+// pulled from the claims of the token identity
+func claimsFromIdentity(identity common.Identity) (map[orgID]orgName, error) {
+	tokenIdentity, ok := identity.(authn.TokenIdentity)
+	if !ok {
+		return nil, fmt.Errorf("cannot get organizations claims from a non-token identity")
```
hexfusion: same here should we return what type we found `%T`

internal/org/providers/claims_provider.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, fmt.Errorf("%w: %s claim not found", flterrors.ErrMissingTokenClaims, organizationClaimName)
+	}
+
+	// Organization claims are a map in the format of
+	// {
+	//   "organization-name": {
+	// 	   "id": "organization-unique-identifier"
+	// 	 }
+	//   ...
+	// }
```
hexfusion: oops :)

internal/org/providers/claims_provider.go
```diff
@@ -0,0 +1,100 @@
...
+	rawOrgs, ok := organizationClaims.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("%w: invalid organizations claim format", flterrors.ErrInvalidTokenClaims)
```
hexfusion: another to consider returning the type you observed `%T` these bugs are a pain so any more details helps a lot

internal/org/resolvers/default.go
```diff
@@ -0,0 +1,76 @@
...
+	store      OrgStore
+	defaultOrg *model.Organization
+	mu         sync.Mutex
+}
+
+func NewDefaultResolver(store OrgStore) *DefaultResolver {
+	return &DefaultResolver{store: store}
+}
+
+func (r *DefaultResolver) getDefaultOrg(ctx context.Context, id uuid.UUID) (*model.Organization, error) {
```
hexfusion, line 26: nit: feels like we’re conflating “get” and “set” here. might not need to change just an observation

internal/org/resolvers/external.go
```diff
@@ -0,0 +1,154 @@
...
+	if ttl > 0 {
+		opts = append(opts, ttlcache.WithTTL[uuid.UUID, *model.Organization](ttl))
+	}
+	c := ttlcache.New(opts...)
+
+	go func() {
+		c.Start()
+		<-ctx.Done()
+		c.Stop()
+	}()
```
hexfusion: question: do we have to fire a go routine in the constructor. Would it be simpler and more explicit if you intialized your cache before and passed it as a param. that would nix the context param on the constructor which feels a little awkward.

internal/org/resolvers/resolver.go
```diff
@@ -0,0 +1,43 @@
...
+
+type Resolver interface {
+	EnsureExists(ctx context.Context, id uuid.UUID) error
+	IsMemberOf(ctx context.Context, identity common.Identity, id uuid.UUID) (bool, error)
+	GetUserOrganizations(ctx context.Context, identity common.Identity) ([]*model.Organization, error)
+}
+
+const cacheExpirationTime = 10 * time.Minute
+
+func BuildResolver(ctx context.Context, cfg *config.Config, store OrgStore, log logrus.FieldLogger) Resolver {
```
hexfusion: by removing the ctx from NewExternalResolver you can drop it here as wel

hexfusion (commented):
Will take a deeper look next week

internal/auth/auth.go
DakCrowder: > Have I mentioned how much I dislike globals? Is there really no other way?

What are you some kind of anti globalist???  In all seriousness I was wary of refactoring this too much but it seems like a good idea, the number of callers is also pretty low so shouldn't bee to difficult.  I'm going to run through some of the other low hanging fruit comments you added then will take a pass at this whole initialization procedure.

internal/org/resolvers/external.go
DakCrowder: Makes sense to me, then the caller can setup the TTL, manage the lifecycle, and the resolver just cares it has a cache that satisfies some Get/Set interface

cmd/flightctl-alert-exporter/main.go
```diff
@@ -78,7 +80,18 @@ func main() {
 	}
 	defer publisher.Close()
 	workerClient := worker_client.NewWorkerClient(publisher, log)
-	serviceHandler := service.WrapWithTracing(service.NewServiceHandler(store, workerClient, kvStore, nil, log, "", "", []string{}))
+
+	orgCache := cache.NewOrganizationTTLCache(cache.DefaultCacheTTL)
+	go orgCache.Start()
+	defer orgCache.Stop()
```
hexfusion, line 86: note: defer is not honored in fatal case not sure how big of a deal that is here

cmd/flightctl-api/main.go
```diff
@@ -165,23 +167,35 @@ func main() {
 		log.Fatalf("creating listener: %s", err)
 	}
 
-	agentserver := agentserver.New(log, cfg, store, ca, agentListener, provider, agentTlsConfig)
+	orgCache := cache.NewOrganizationTTLCache(cache.DefaultCacheTTL)
```
hexfusion: no change needed, but we already know its a cache right 

```
cache.NewOrganizationTTL
```

internal/auth/auth.go
```diff
@@ -28,25 +30,14 @@ const (
...
-var authN AuthNMiddleware
-
-func GetAuthZ() AuthZMiddleware {
-	return authZ
-}
-
-func GetAuthN() AuthNMiddleware {
-	return authN
-}
-
```
hexfusion, line 49: omg thank you  :heart:

internal/auth/authn/jwtverifier.go
```diff
@@ -58,17 +82,51 @@ func NewJWTAuth(oidcAuthority string, externalOIDCAuthority string, clientTlsCon
...
 func (j JWTAuth) ValidateToken(ctx context.Context, token string) error {
+	_, err := j.parseAndCreateIdentity(ctx, token)
+	return err
+}
+
+func (j JWTAuth) parseAndCreateIdentity(ctx context.Context, token string) (*JWTIdentity, error) {
+	// Check if we already have a parsed identity in the context
+	if existingIdentity, ok := ctx.Value(consts.IdentityCtxKey).(*JWTIdentity); ok {
+		return existingIdentity, nil
+	}
```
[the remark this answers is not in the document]
hexfusion: this might be worth looking at

internal/store/model/organization.go
```diff
@@ -13,7 +13,7 @@ type Organization struct {
 	DisplayName string `json:"display_name"`
 
 	// External identifier of the organization in the configured IdP.
-	ExternalID string `gorm:"index" json:"external_id"`
+	ExternalID string `json:"external_id"`
```
hexfusion, line 16: double checking this change is needed

DakCrowder: Yes, the index logic has been moved to what is being run in the store file and not derived from gorm tags

PR #1578, opened 2025-08-28
EDM-2064: Fix CLI version showing -dirty by cleaning git working tree

galel12:
## EDM-2064: Fix CLI version showing -dirty

**Problem**: CLI artifacts downloaded from UI showed `v0.9.2-dirty` instead of clean `v0.9.2`

**Root Cause**: Selective COPY in containers caused git to perceive working tree as dirty

**Solution**: 
- Remove duplicated version calculation from GitHub Actions workflow
- Add git stash technique to clean working tree state in containers  
- Make Makefile single source of truth for versioning

**Tested**: ✅ CLI artifacts now show clean versions without `-dirty` suffix

## Summary by CodeRabbit

- Chores
  - Reduced files copied during container builds to improve caching and build stability.
  - Stopped exporting git metadata as runtime environment variables for publish job; git metadata no longer passed to that CI container build.

- Refactor
  - Hardened version-extraction during image builds and adjusted Makefile to derive normalized tag and major version.

Impact
- More reproducible, cache-friendly builds with no runtime behavior changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,9 +39,10 @@ GOARCH := $(shell go env GOARCH)
 
 VERBOSE ?= false
 
-SOURCE_GIT_TAG ?=$(shell git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
-SOURCE_GIT_TREE_STATE ?=$(shell ( ( [ ! -d ".git/" ] || git diff --quiet ) && echo 'clean' ) || echo 'dirty')
-SOURCE_GIT_COMMIT ?=$(shell git rev-parse --short "HEAD^{commit}" 2>/dev/null || echo "unknown")
+# Check ENV first; if unset or empty, fallback to git (robust for CI/env quirks)
+SOURCE_GIT_TAG := $(if $(strip $(SOURCE_GIT_TAG)),$(SOURCE_GIT_TAG),$(shell git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown"))
+SOURCE_GIT_TREE_STATE := $(if $(strip $(SOURCE_GIT_TREE_STATE)),$(SOURCE_GIT_TREE_STATE),$(shell ( ( [ ! -d ".git/" ] || git diff --quiet ) && echo clean ) || echo dirty))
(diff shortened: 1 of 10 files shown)


eldar101:
Please close and resolve all suggestions.
ping me for approval.

galel12:
@eldar101 resolved all suggestions.

Makefile
```diff
@@ -39,9 +39,10 @@ GOARCH := $(shell go env GOARCH)
 
 VERBOSE ?= false
 
-SOURCE_GIT_TAG ?=$(shell git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
-SOURCE_GIT_TREE_STATE ?=$(shell ( ( [ ! -d ".git/" ] || git diff --quiet ) && echo 'clean' ) || echo 'dirty')
-SOURCE_GIT_COMMIT ?=$(shell git rev-parse --short "HEAD^{commit}" 2>/dev/null || echo "unknown")
+# Check for ENV vars first (from container), fallback to git commands (for host builds)
+SOURCE_GIT_TAG ?= $(or $(shell echo $$SOURCE_GIT_TAG 2>/dev/null), $(shell git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown"))
+SOURCE_GIT_TREE_STATE ?= $(or $(shell echo $$SOURCE_GIT_TREE_STATE 2>/dev/null), $(shell ( ( [ ! -d ".git/" ] || git diff --quiet ) && echo 'clean' ) || echo 'dirty'))
```
fzdarsky: Is there a scenario where we'd want to allow the `SOURCE_GIT_TREE_STATE` to be passed in as parameter rather than determining the true state here?

asafbss: I think the main case where it makes sense to allow passing SOURCE_GIT_TREE_STATE as a parameter is in CI. In that environment we don’t want to run git diff at all, and we can just pass "clean" explicitly. So yes, it’s useful to keep the ability to pass it in, at least for CI runs. Otherwise, if it isn’t provided, the Makefile will fall back to calculating it locally. We do it for the downstream build.

PR #1580, opened 2025-08-28
NO-ISSUE: address nil pointer in login flow

DakCrowder:


internal/cli/login.go
```diff
@@ -212,6 +212,14 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("must provide --token")
 	}
 
+	o.clientConfig.AuthInfo.AuthProvider = &client.AuthProviderConfig{
```
amir-yogev-gh, line 215: The same block you added seems to be already part of the block at line 182.
What is the difference?
Don't you need the third line with the CA file?

DakCrowder: Two different auth providers are set in the file - o.authProvider and o.clientConfig.AuthInfo.AuthProvider.  The ca file for the o.clientConfig.AuthInfo.AuthProvider is set further down in the method.  Seems like maybe this should be consolidated a bit but my focus right now is just getting the login working.

SiddarthR56 (approved):

PR #1586, opened 2025-08-31
EDM-2058: agent retry and 404 handling 

asafbennatan:
1. added infinite retry for for enrollmentClient (as creating / getting ers is not a periodic task)
2. added 404 device not found handler when getting specs - this had to be implemented with a callback since publisher does not have access to identity providers (we need access to wipe the certificate and then restart the flightctl-agent)
3. server will not return an older spec than the one reported by agent unless there are consoles
4. added jitter for retry policy
5. even after the change that was recently merged to queue.go to ignore old versions we still need the spec manager to specifically ignore older versions for desired , without this the desired version can be downgraded (consider current=5 , desired =7 and incoming =6 , without the change desired will be downgraded to 6)

## Summary by CodeRabbit

* **New Features**
  * Automatic recovery when the management service reports a specific "device not found": the agent will wipe only the device certificate and restart to re-enroll.
  * Enrollment and polling now use jittered backoff with per-run RNG seeding; enrollment also supports infinite HTTP retry.

* **Bug Fixes**
  * Rejects older desired versions to prevent accidental rollbacks.
  * Avoids redundant writes when desired and rendered versions match.

* **Tests**
  * Expanded coverage for jitter/backoff, nuanced device-not-found handling, and certificate-wipe behaviors.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"time"
 
 	"github.com/flightctl/flightctl/internal/agent/client"
@@ -114,10 +115,12 @@ func (a *Agent) Run(ctx context.Context) error {
 	}
 
 	pollBackoff := poll.Config{
-		MaxDelay:  1 * time.Minute,
-		BaseDelay: 10 * time.Second,
-		Factor:    1.5,
-		MaxSteps:  a.config.PullRetrySteps,
+		MaxDelay:     1 * time.Minute,
+		BaseDelay:    10 * time.Second,
+		Factor:       1.5,
+		MaxSteps:     a.config.PullRetrySteps,
+		JitterFactor: 0.1,                                             // jitterFactor (10% jitter to prevent thundering herd)
+		Rand:         rand.New(rand.NewSource(time.Now().UnixNano())), //nolint:gosec
 	}
 
 	// create os client
@@ -158,7 +161,10 @@ func (a *Agent) Run(ctx context.Context) error {
 	devicePublisher := publisher.New(deviceName,
 		time.Duration(a.config.SpecFetchInterval),
 		backoff,
-		a.log)
+		a.log,
+		func() error {
+			return wipeCertificateAndRestart(ctx, identityProvider, executer, a.log)
+		})
 
 	// create spec manager
 	specManager := spec.NewManager(
@@ -350,7 +356,18 @@ func (a *Agent) Run(ctx context.Context) error {
 }
 
 func newEnrollmentClient(cfg *agent_config.Config, log *log.PrefixLogger) (client.Enrollment, error) {
-	httpClient, err := client.NewFromConfig(&cfg.EnrollmentService.Config, log)
+	// Create infinite retry policy for enrollment requests using same config as management client
+	// but with infinite retries (MaxSteps: 0)
+	infiniteRetryConfig := poll.Config{
+		BaseDelay:    10 * time.Second,                                // baseDelay (same as management client)
+		Factor:       1.5,                                             // factor (same as management client)
+		MaxDelay:     1 * time.Minute,                                 // maxDelay (same as management client)
+		MaxSteps:     0,                                               // maxSteps (0 means infinite retries until context timeout)
+		JitterFactor: 0.1,                                             // jitterFactor (10% jitter to prevent thundering herd)
+		Rand:         rand.New(rand.NewSource(time.Now().UnixNano())), //nolint:gosec
+	}
+
+	httpClient, err := client.NewFromConfig(&cfg.EnrollmentService.Config, log, client.WithHTTPRetry(infiniteRetryConfig))
 	if err != nil {
 		return nil, err
 	}
@@ -368,4 +385,23 @@ func (a *Agent) tryLoadTPM(writer fileio.ReadWriter) (*tpm.Client, error) {
 		return nil, fmt.Errorf("creating TPM client: %w", err)
 	}
 	return tpmClient, nil
+}
+
+// wipeCertificateAndRestart wipes only the certificate (not keys or CSR) and restarts the flightctl-agent service
+func wipeCertificateAndRestart(ctx context.Context, identityProvider identity.Provider, executer executer.Executer, log *log.PrefixLogger) error {
+	log.Warn("Device not found on management server - wiping certificate and restarting agent")
+
+	// Wipe only the certificate, preserving keys and CSR
+	if err := identityProvider.WipeCertificateOnly(); err != nil {
+		return fmt.Errorf("failed to wipe certificate: %w", err)
+	}
+
+	// Restart the flightctl-agent service
+	systemdClient := client.NewSystemd(executer)
+	if err := systemdClient.Restart(ctx, "flightctl-agent"); err != nil {
+		return fmt.Errorf("failed to restart flightctl-agent service: %w", err)
+	}
+
+	log.Info("Successfully wiped certificate and restarted flightctl-agent service")
+	return nil
 }
--- a/internal/agent/device/publisher/publisher.go
+++ b/internal/agent/device/publisher/publisher.go
@@ -31,26 +31,29 @@ type Publisher interface {
 }
 
 type publisher struct {
-	managementClient client.Management
-	deviceName       string
-	subscribers      []Subscription
-	lastKnownVersion string
-	interval         time.Duration
-	stopped          atomic.Bool
-	log              *log.PrefixLogger
-	backoff          wait.Backoff
-	mu               sync.Mutex
+	managementClient      client.Management
+	deviceName            string
+	subscribers           []Subscription
+	lastKnownVersion      string
+	interval              time.Duration
+	stopped               atomic.Bool
+	log                   *log.PrefixLogger
+	backoff               wait.Backoff
+	deviceNotFoundHandler func() error
+	mu                    sync.Mutex
 }
 
 func New(deviceName string,
 	interval time.Duration,
 	backoff wait.Backoff,
-	log *log.PrefixLogger) Publisher {
+	log *log.PrefixLogger,
+	deviceNotFoundHandler func() error) Publisher {
 	return &publisher{
-		deviceName: deviceName,
-		interval:   interval,
-		backoff:    backoff,
-		log:        log,
+		deviceName:            deviceName,
+		interval:              interval,
+		backoff:               backoff,
+		log:                   log,
+		deviceNotFoundHandler: deviceNotFoundHandler,
 	}
 }
 
@@ -125,6 +128,19 @@ func (n *publisher) pollAndPublish(ctx context.Context) {
 		n.log.Debugf("Dialing management API took: %v", duration)
 	}
 	if err != nil {
+		// Check for device not found error - handle certificate wiping and restart
+		if errors.Is(err, client.ErrDeviceNotFound) {
+			n.log.Warn("Device not found on management server")
+			if n.deviceNotFoundHandler != nil {
+				if handlerErr := n.deviceNotFoundHandler(); handlerErr != nil {
+					n.log.Warnf("Failed to handle device not found: %v", handlerErr)
+					return
+				}
+				n.log.Info("Successfully handled device not found - certificate wiped and agent restarted")
+			}
+			return
+		}
+
 		if errors.Is(err, errors.ErrNoContent) || errors.IsTimeoutError(err) {
 			n.log.Debug("No new template version from management service")
 			return
--- a/internal/agent/device/spec/manager.go
+++ b/internal/agent/device/spec/manager.go
@@ -381,6 +381,20 @@ func (s *manager) getDeviceFromQueue(ctx context.Context) (*v1alpha1.Device, boo
 
 	// if this is a new version ensure we persist it to disk
 	if desired.Version() != s.cache.getRenderedVersion(Desired) {
+		// Guard check: ensure we don't allow older versions
+		desiredVersion, err := stringToInt64(desired.Version())
+		if err != nil {
+			return nil, false, fmt.Errorf("invalid desired version: %w", err)
+		}
+		currentDesiredVersion, err := stringToInt64(s.cache.getRenderedVersion(Desired))
+		if err != nil {
+			return nil, false, fmt.Errorf("invalid current desired version: %w", err)
+		}
+		if desiredVersion < currentDesiredVersion {
+			s.log.Errorf("Rejecting older version: %d < %d", desiredVersion, currentDesiredVersion)
+			return nil, false, fmt.Errorf("version %d is older than current desired version %d", desiredVersion, currentDesiredVersion)
+		}
+
 		if s.isNewDesiredVersion(desired) {
 			s.log.Infof("Writing new desired rendered spec to disk version: %s", desired.Version())
 		}
--- a/internal/agent/device/spec/spec.go
+++ b/internal/agent/device/spec/spec.go
@@ -2,6 +2,7 @@ package spec
 
 import (
 	"context"
+	"math/rand"
 	"time"
 
 	"github.com/flightctl/flightctl/api/v1alpha1"
@@ -27,9 +28,11 @@ const (
 
 // defaultSpecPollConfig is the default poll configuration for the spec priority queue.
 var defaultSpecPollConfig = poll.Config{
-	BaseDelay: 30 * time.Second,
-	Factor:    1.5,
-	MaxDelay:  5 * time.Minute,
+	BaseDelay:    30 * time.Second,
+	Factor:       1.5,
+	MaxDelay:     5 * time.Minute,
+	JitterFactor: 0.1,
+	Rand:         rand.New(rand.NewSource(time.Now().UnixNano())), //nolint:gosec
 }
 
 type Manager interface {
--- a/internal/agent/identity/tpm.go
+++ b/internal/agent/identity/tpm.go
@@ -221,6 +221,14 @@ func (t *tpmProvider) WipeCredentials() error {
 	return nil
 }
 
+func (t *tpmProvider) WipeCertificateOnly() error {
+	// For TPM provider, only clear certificate data from memory
+	// The key and CSR remain in TPM storage
+	t.certificateData = nil
+	t.log.Info("Wiped TPM-stored certificate data from memory (key and CSR preserved)")
+	return nil
+}
+
 func (t *tpmProvider) Close(ctx context.Context) error {
 	if t.client != nil {
 		return t.client.Close(ctx)
--- a/internal/store/model/device.go
+++ b/internal/store/model/device.go
@@ -142,8 +142,22 @@ func (d *Device) ToApiResource(opts ...APIResourceOption) (*api.Device, error) {
 
 		// if we have a console request we ignore the rendered version
 		// TODO: bump the rendered version instead?
-		if len(consoles) == 0 && apiOpts.knownRenderedVersion != nil && renderedVersion == *apiOpts.knownRenderedVersion {
-			return nil, nil
+		if len(consoles) == 0 && apiOpts.knownRenderedVersion != nil {
+			// Try numeric comparison first, fall back to lexicographic if parsing fails
+			renderedVersionInt, err1 := strconv.ParseInt(renderedVersion, 10, 64)
+			knownRenderedVersionInt, err2 := strconv.ParseInt(*apiOpts.knownRenderedVersion, 10, 64)
+
+			if err1 == nil && err2 == nil {
+				// Both versions are numeric, use numeric comparison
+				if renderedVersionInt <= knownRenderedVersionInt {
+					return nil, nil
+				}
+			} else {
+				// Fall back to lexicographic comparison for non-numeric versions
+				if renderedVersion <= *apiOpts.knownRenderedVersion {
+					return nil, nil
+				}
+			}
 		}
 		// TODO: handle multiple consoles, for now we just encapsulate our one console in a list
 		spec.Config = d.RenderedConfig.Data
--- a/pkg/poll/poll.go
+++ b/pkg/poll/poll.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"time"
 )
 
@@ -23,6 +24,20 @@ type Config struct {
 	MaxDelay time.Duration
 	// MaxSteps limits the number of retries. If 0, retries will continue until timeout.
 	MaxSteps int
+	// JitterFactor adds randomization to prevent thundering herd (0.0 to 1.0)
+	JitterFactor float64
+	// Rand is the random number generator for jitter calculation
+	Rand *rand.Rand
+}
+
+// NewConfig creates a new Config with a properly seeded random number generator.
+func NewConfig(baseDelay time.Duration, factor float64) *Config {
+	return &Config{
+		BaseDelay:    baseDelay,
+		Factor:       factor,
+		JitterFactor: 0.1,                                             // Default jitter factor
+		Rand:         rand.New(rand.NewSource(time.Now().UnixNano())), //nolint:gosec
+	}
 }
 
 // Validate checks if the configuration parameters are valid
@@ -36,7 +51,22 @@ func (c *Config) Validate() error {
 	if c.MaxDelay > 0 && c.MaxDelay < c.BaseDelay {
 		return errors.New("poll MaxDelay must be greater than or equal to BaseDelay")
 	}
+	if c.JitterFactor < 0 || c.JitterFactor > 1 {
+		return errors.New("poll JitterFactor must be between 0.0 and 1.0")
+	}
+	if c.Rand == nil {
+		return errors.New("poll Rand must not be nil")
+	}
 	return nil
+}
+
+// calculateJitter calculates jitter for a given delay based on the jitter factor.
+func calculateJitter(rng *rand.Rand, delay time.Duration, jitterFactor float64) time.Duration {
+	if jitterFactor <= 0 {
+		return 0
+	}
+	jitterRange := float64(delay) * jitterFactor
+	return time.Duration((rng.Float64()*2 - 1) * jitterRange)
 }
 
 // BackoffWithContext repeatedly calls the operation until timeout is reached,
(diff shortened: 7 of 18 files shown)


internal/agent/client/client.go
```diff
@@ -23,17 +27,157 @@ import (
...
+		// Success or non-retryable error
+		return nil
+	}, contextBackoff)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return finalResp, finalErr
+}
```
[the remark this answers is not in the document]
asafbennatan: this was the intention

internal/agent/client/client.go
```diff
@@ -23,17 +27,157 @@
...
+		// For retries to work with request bodies, the original request should use
+		// GetBody function or the body should be seekable. For now, we'll assume
+		// the OpenAPI client handles this properly.
+		reqClone.Body = req.Body
+		reqClone.GetBody = req.GetBody
+	}
+
+	return reqClone
+}
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/agent/device/spec/manager.go
```diff
@@ -377,8 +377,16 @@ func (s *manager) getDeviceFromQueue(ctx context.Context) (*v1alpha1.Device, boo
 		return nil, true, nil
 	}
 
+	currentDesiredVersion := s.cache.getRenderedVersion(Desired)
+
+	// Skip older versions - they are outdated
+	if isOlderVersion(desired.Version(), currentDesiredVersion) {
+		s.log.Warnf("Skipping older version %s (current desired: %s) from queue", desired.Version(), currentDesiredVersion)
+		return nil, true, nil // Signal requeue to get next item
+	}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/agent/client/client.go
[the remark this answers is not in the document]
asafbennatan: this is the intention

internal/agent/client/client.go
```diff
@@ -23,17 +27,177 @@ import (
...
+		}
+
+		// Last resort: if we can't rewind or recreate the body,
+		// we'll have to use the original body (this may fail on retry)
+		reqClone.Body = req.Body
+		reqClone.GetBody = req.GetBody
+	}
+
+	return reqClone
+}
```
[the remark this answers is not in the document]
asafbennatan: according to my understanding these requests are seekable - which we handle

asafbennatan:
@hexfusion  the question is how can we distinguish between an accidental table delete and the device actually not being there in the restore , note that we have a specific message we are expecting on that 404 so a generic 404 wouldnt cause us to do that

hexfusion:
this is a unique situation where this problem could exist, would you agree? the result decommision could be very disruptive and unexpected during normal operations. (some process or fat finger (DELETE table ...). I believe in this case API would respond with 404 and the same message?  I am asking that at a minimim we gate the scope of this to disaster recovery and harden it with a more explicit signal. does that seem reasonable?

asafbennatan:
I agree , question is how , since in both cases we don't have the device so
we cannot annotate anything on it .
We can mark something systemwide but how long would it take effect ?
devices that connect only once every week would not behave correctly in
that case .

What should be the behavior when a device thinks its registered but server
doesn't know about irrespective of restore.

Perhaps we could save unknown devices as a new store ?

hexfusion:
cc @lkatalin would like you to chime in on this as it pertains to decom flow.

asafbennatan:
Note that currently I do not wipe key and csr only the certificate

lkatalin:
Looking at this and the related issue, there are critical security implications that need to be addressed. A device that has a valid mTLS cert but no related server-side data could be a compromised device, as this is not an expected state in the device lifecycle. I think it would be important to align the improvements proposed here with the secure device lifecycle, expected decommissioning / enrollment states, etc. Is there a design doc for disaster recovery already or can one be created?

> What should be the behavior when a device thinks its registered but server doesn't know about irrespective of restore

Our source of truth has to be server-side when there is any divergence between server and agent, because only the server-side is trusted.

avishayt:
> Looking at this and the related issue, there are critical security implications that need to be addressed. A device that has a valid mTLS cert but no related server-side data could be a compromised device, as this is not an expected state in the device lifecycle. I think it would be important to align the improvements proposed here with the secure device lifecycle, expected decommissioning / enrollment states, etc. Is there a design doc for disaster recovery already or can one be created?
> 
> > What should be the behavior when a device thinks its registered but server doesn't know about irrespective of restore
> 
> Our source of truth has to be server-side when there is any divergence between server and agent, because only the server-side is trusted.

Here's the design doc: https://docs.google.com/document/d/1uM-eIx6o1e-t5VA3g6Fm_dciRYCYW7mMwlfbCT00SnU/edit?tab=t.0#heading=h.ph7rewjls114

IMO regardless of how the device got to this state, it's not enrolled in the system and therefore it should re-enroll.  Then the operator can either approve or deny the request depending on if the record was lost or the device was compromised.

internal/agent/agent.go
```diff
@@ -238,6 +238,20 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		// Get current device status for the wipe operation
+		currentStatus := statusManager.Get(ctx)
+		if currentStatus == nil {
+			a.log.Warn("Device status is nil during certificate wipe, continuing anyway")
+		}
+
+		// Wipe only the certificate (preserving private key/CSR) and reboot
+		return lifecycleManager.WipeCertificateAndReboot(ctx)
+	}
+
```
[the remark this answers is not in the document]
asafbennatan: this is the case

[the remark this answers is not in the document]
asafbennatan: there are already safe guards , a plain 404 will not cause this , only a 404 with the specific body with a device not found error

internal/agent/identity/tpm.go
```diff
@@ -221,6 +220,13 @@ func (t *tpmProvider) WipeCredentials() error {
 	return nil
 }
 
+func (t *tpmProvider) WipeCertificate() error {
+	// clear only certificate data from memory, keeping TPM key intact
+	t.certificateData = nil
```
lkatalin, line 227: What about in the non-TPM case?
It looks like it is stored as a file: https://github.com/flightctl/flightctl/blob/main/internal/agent/identity/file.go#L74

lkatalin: Actually does this flow also keep the same device identity (keys) if it's a non-TPM device?

lkatalin:
Thanks for the doc @avishayt . 

> IMO regardless of how the device got to this state, it's not enrolled in the system and therefore it should re-enroll. Then the operator can either approve or deny the request depending on if the record was lost or the device was compromised.

That is reasonable, I just want to be clear on the details. This is the lifecycle design we had:

flightctl decommissioning design

As this 404 would only wipe the current agent certificate but keep the same device ID (at least in the TPM case, see inline for question about non-TPM) I agree this would not count as an EOL state, so it could be acceptable with the proper ER checks.

The DB needs to retain decommissioned device IDs (this would act as the EOL "signal" after which devices cannot re-enroll). Any ER will need to be checked against those IDs and flagged or denied if previously decommissioned. Did we ever make a decision on whether we are keeping decommissioned device IDs? I know there was some desire to eventually delete that data for performance reasons.

Additionally it would be good to log this server-side (maybe there is an alert) if such a 404 is encountered. I think this may be implemented already.

go.mod
```diff
@@ -105,6 +105,7 @@ require (
 )
 
 require (
+	github.com/cenkalti/backoff/v4 v4.3.0
```
kkyrazis: There's already a few ways to exponentially backoff in the agent. k8s.io/apimachinery and our pkg/poll. Can we just use those?

internal/agent/agent.go
```diff
@@ -238,6 +238,20 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log,
 	)
 
+	// Create device not found callback for HTTP client level handling
+	deviceNotFoundCallback := func(ctx context.Context) error {
+		a.log.Warn("Device not found on server - wiping certificate and rebooting for fresh enrollment")
+
+		// Get current device status for the wipe operation
+		currentStatus := statusManager.Get(ctx)
```
kkyrazis: What is the purpose of this check?

internal/agent/device/spec/manager.go
```diff
@@ -377,8 +377,26 @@ func (s *manager) getDeviceFromQueue(ctx context.Context) (*v1alpha1.Device, boo
 		return nil, true, nil
 	}
 
+	currentDesiredVersion := s.cache.getRenderedVersion(Desired)
+
+	// Skip older versions - they are outdated
+	if isOlderVersion(desired.Version(), currentDesiredVersion) {
```
kkyrazis: I'm curious as to why exactly this happens. Restarting the agent process and triggering re-enrollment should send the current device status with the current rendered version in it. Wouldn't the approval of the new enrollment request create a device where the rendered version starts at non-zero?

internal/agent/agent.go
```diff
@@ -238,6 +238,20 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log,
 	)
 
+	// Create device not found callback for HTTP client level handling
+	deviceNotFoundCallback := func(ctx context.Context) error {
```
kkyrazis: As a whole, does this need to be a callback created in the top level of the agent? I think we could get away with pushing this into the internals

internal/agent/client/client.go
```diff
@@ -23,17 +27,177 @@ import (
...
+	return NewFromConfigWithHTTPClient(config, httpClient, deviceNotFoundCallback)
+}
+
+// NewFromConfigWithHTTPClient returns a new Flight Control API client using a pre-configured HTTP client
+// with global 5xx retry and device 404 handling.
+func NewFromConfigWithHTTPClient(config *baseclient.Config, httpClient *http.Client, deviceNotFoundCallback DeviceNotFoundCallback) (*client.ClientWithResponses, error) {
+	// Wrap the HTTP client with device 404 detection and 5xx retry logic
+	wrappedClient := &DeviceNotFoundHTTPClient{
+		Client:   httpClient,
+		Callback: deviceNotFoundCallback,
```
kkyrazis: If no callback is defined is it still worth wrapping?

internal/agent/client/client.go
```diff
@@ -23,17 +27,177 @@ import (
...
+			// Network errors should be retried
+			return err
+		}
+
+		// Store response for final return
+		finalResp = resp
+		finalErr = nil
+
+		// Check for 5xx errors to retry
+		if resp.StatusCode >= 500 && resp.StatusCode < 600 {
```
kkyrazis: Is it possible to add more visibility here. I guess it's nice that we have retries built in on server side issues, but theres no real visibility from the caller that something is amiss.

internal/agent/device/spec/manager.go
hexfusion: The agent knows its current state. When a request is made requesting the latest rendered spec it includes its current version. The expectation is that the service only responds with newer versions. Is this assumption broken by DR?

kkyrazis: IIRC the publisher's query on start is versionless and uses the result of the initial query to guide future queries. Perhaps it would be simpler to wire up the publisher so that it always has an initial version.

hexfusion: Yes this is a bug , current version is always know even if it is zero and the publisher should never call with a version less than current.

internal/agent/agent.go
asafbennatan: we have more than one api that may result with 404 not found - this is why i did this

internal/agent/client/client.go
asafbennatan: yes since it adds 5xx retry as well

internal/agent/client/client.go
asafbennatan: what do you mean here ? do you want to add logs ?
i dont think that the caller should care if the request was retired or not - it is an internal implementation detail , we could add a retry metric in the RPCMetricsCallback

for now i have added logs

asafbennatan:
@hexfusion this just a single location , what about all other api calls ?
i think keeping like it is makes sense :
 1. all requests should retry 5xx
 2. in the specific case of getRenderedDevice if we get 409/204 - retry as well

internal/agent/device/spec/manager.go
asafbennatan: @hexfusion 
according to what i understand the assumption was already broken by consoles? 
 #https://github.com/flightctl/flightctl/blob/4bd040d1a6c80c36bee72830d3556b69680b5a72/internal/store/model/device.go#L145

asafbennatan: i have changed == to <= , but the current protection - agent side is required nonetheless , the agent should not return to an older version regardless of the service behavior (as i said currently it will return older or eq version if there are consoles)

internal/agent/agent.go
kkyrazis: Sure, but it doesn't have to be created here does it? This could be scoped to bootstrap rather than the agent.

internal/agent/client/client.go
kkyrazis: I'm probably wrong, but I'm not convinced that that is that big of a net positive. What about for metrics? Are we measuring the time it takes it takes to get a 200 or 400? Or should it be expected that a call to a Client is measuring time for a single request?

internal/agent/client/client.go
```diff
@@ -18,22 +22,233 @@ import (
...
+	// Try to parse as Status object
+	var status v1alpha1.Status
+	if err := json.Unmarshal(body, &status); err != nil {
+		// If not JSON, check raw body for device not found message
+		bodyStr := string(body)
+		return strings.Contains(bodyStr, "Device of name") && strings.Contains(bodyStr, "not found")
+	}
+
+	// Check if it's a device-specific not found error
+	return status.Code == http.StatusNotFound &&
```
kkyrazis: I know this is explicitly the `DeviceNotFoundHttpClient`, but I'm assuming that this specific check is necessary to prevent other resources (CSRs and ERs) managed by the agent from triggering a restart of the service. This feels too targeted for what should be a generic agent management client. How would I add a `CSRNotFoundClient` if I needed to take specific action on a 404 for those in the future?

internal/agent/client/client.go
asafbennatan: its not a question of a net positive - we need retry for all requests this is why we are wrapping it , do you suggest that we handle retry for each request made individually ?

internal/agent/client/client.go
asafbennatan: |Sure, but it doesn't have to be created here does it? This could be scoped to bootstrap rather than the agent. 
    this will require changing a bunch of interfaces to support this since LifecycleInitalizer does not support the new WipeCertificateAndRestartService function , i have preferred creating this here rather then changing the interfaces.

|Or should it be expected that a call to a Client is measuring time for a single request? - i am not sure 

|How would I add a CSRNotFoundClient - once we have this specific use case we can adjust if needed .

we need this PR to merge to move forward with the backup and restore feature , if not critical - please allow this to be merged .
Thanks.

kkyrazis: > we need this PR to merge to move forward with the backup and restore feature , if not critical - please allow this to be merged .

I haven't blocked it by requesting changes. I'm asking questions and commenting on things that look a little smelly to me, but I am not a gatekeeper.

internal/agent/client/client.go
kkyrazis: Yes, that is what I would expect callers to do if they cared. Perhaps a caller is already on an interval and doesn't require infinite retries of 500s. Maybe it's fine waiting for its next interval. Maybe a caller has a very specific retry policy that they're required to use.

internal/agent/device/publisher/publisher.go
```diff
@@ -31,26 +31,29 @@ type Publisher interface {
...
+	managementClient      client.Management
+	deviceName            string
+	subscribers           []Subscription
+	lastKnownVersion      string
+	interval              time.Duration
+	stopped               atomic.Bool
+	log                   *log.PrefixLogger
+	backoff               wait.Backoff
+	deviceNotFoundHandler func() error
+	mu                    sync.Mutex
```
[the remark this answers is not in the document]
asafbennatan: there are no multiple polls

internal/agent/device/publisher/publisher.go
```diff
@@ -31,26 +31,29 @@
...
-		deviceName: deviceName,
-		interval:   interval,
-		backoff:    backoff,
-		log:        log,
+		deviceName:            deviceName,
+		interval:              interval,
+		backoff:               backoff,
+		log:                   log,
+		deviceNotFoundHandler: deviceNotFoundHandler,
 	}
```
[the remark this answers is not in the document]
asafbennatan: i dont think this is suitable for a non-optional parameter

internal/agent/device/spec/spec.go
```diff
@@ -27,9 +27,10 @@ const (
 
 // defaultSpecPollConfig is the default poll configuration for the spec priority queue.
 var defaultSpecPollConfig = poll.Config{
-	BaseDelay: 30 * time.Second,
-	Factor:    1.5,
-	MaxDelay:  5 * time.Minute,
+	BaseDelay:    30 * time.Second,
+	Factor:       1.5,
+	MaxDelay:     5 * time.Minute,
+	JitterFactor: 0.1,
```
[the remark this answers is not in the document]
asafbennatan: fixed

pkg/poll/poll.go
```diff
@@ -106,5 +144,13 @@
...
 
+	// Add pre-calculated jitter to prevent thundering herd
+	delayDuration += cfg.jitterOffset
+
+	// Ensure delay doesn't go negative
+	if delayDuration < 0 {
+		delayDuration = 0
+	}
+
 	return delayDuration
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/store/model/device.go
```diff
@@ -142,7 +142,7 @@ func (d *Device) ToApiResource(opts ...APIResourceOption) (*api.Device, error) {
 
 		// if we have a console request we ignore the rendered version
 		// TODO: bump the rendered version instead?
-		if len(consoles) == 0 && apiOpts.knownRenderedVersion != nil && renderedVersion == *apiOpts.knownRenderedVersion {
+		if len(consoles) == 0 && apiOpts.knownRenderedVersion != nil && renderedVersion <= *apiOpts.knownRenderedVersion {
 			return nil, nil
 		}
```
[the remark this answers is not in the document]
asafbennatan: fixed

pkg/poll/poll.go
```diff
@@ -23,6 +24,10 @@ type Config struct {
 	MaxDelay time.Duration
 	// MaxSteps limits the number of retries. If 0, retries will continue until timeout.
 	MaxSteps int
+	// JitterFactor adds randomization to prevent thundering herd (0.0 to 1.0)
+	JitterFactor float64
+	// jitterOffset is the pre-calculated jitter offset for efficiency
+	jitterOffset time.Duration
 }
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/agent/device/publisher/publisher.go
```diff
@@ -125,6 +128,19 @@
...
+			n.log.Warn("Device not found on management server")
+			if n.deviceNotFoundHandler != nil {
+				if handlerErr := n.deviceNotFoundHandler(); handlerErr != nil {
+					n.log.Errorf("Failed to handle device not found: %v", handlerErr)
+					return
+				}
+				n.log.Info("Successfully handled device not found - certificate wiped and agent restarted")
+			}
+			return
+		}
```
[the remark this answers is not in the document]
asafbennatan: no need , this is called once

pkg/poll/poll.go
```diff
@@ -36,9 +39,32 @@ func (c *Config) Validate() error {
...
 		return errors.New("poll MaxDelay must be greater than or equal to BaseDelay")
 	}
+	if c.JitterFactor < 0 || c.JitterFactor > 1 {
+		return errors.New("poll JitterFactor must be between 0.0 and 1.0")
+	}
 	return nil
 }
 
+// NewConfig creates a new poll.Config.
+func NewConfig(baseDelay time.Duration, factor float64, maxDelay time.Duration, maxSteps int, jitterFactor float64) *Config {
```
hexfusion: the downside of a constructor for config is you lose the self-documenting structs.

pkg/poll/poll.go
```diff
@@ -36,9 +39,32 @@ func (c *Config) Validate() error {
...
+		BaseDelay:    baseDelay,
+		Factor:       factor,
+		MaxDelay:     maxDelay,
+		MaxSteps:     maxSteps,
+		JitterFactor: jitterFactor,
+	}
+}
+
+// calculateJitter calculates jitter for a given delay based on the jitter factor.
+func calculateJitter(delay time.Duration, jitterFactor float64) time.Duration {
```
hexfusion: :+1:

internal/agent/agent.go
```diff
@@ -350,7 +354,17 @@ func (a *Agent) Run(ctx context.Context) error {
 }
 
 func newEnrollmentClient(cfg *agent_config.Config, log *log.PrefixLogger) (client.Enrollment, error) {
-	httpClient, err := client.NewFromConfig(&cfg.EnrollmentService.Config, log)
+	// Create infinite retry policy for enrollment requests using same config as management client
+	// but with infinite retries (MaxSteps: 0)
+	infiniteRetryConfig := *poll.NewConfig(
+		10*time.Second, // baseDelay (same as management client)
```
hexfusion: nit: if we use poll.Config these fields will self document

internal/agent/device/spec/manager.go
```diff
@@ -381,6 +381,20 @@ func (s *manager) getDeviceFromQueue(ctx context.Context) (*v1alpha1.Device, boo
 
 	// if this is a new version ensure we persist it to disk
 	if desired.Version() != s.cache.getRenderedVersion(Desired) {
+		// Guard check: ensure we don't allow older versions
+		desiredVersion, err := stringToInt64(desired.Version())
```
hexfusion, line 385: I understand the concern but desired.Version() comes from the queue. The queue will not allow a version to be added that is less than the currentVersion. So if anything needs to be validated it should happen before the item is added to the queue.

https://github.com/flightctl/flightctl/blob/e34982f80082149711920f47018ee534c917ec02/internal/agent/device/spec/manager.go#L376

https://github.com/flightctl/flightctl/blob/e34982f80082149711920f47018ee534c917ec02/internal/agent/device/spec/queue.go#L94

hexfusion (commented):
Few small nits then lgtm.

internal/agent/identity/tpm.go
hexfusion: this will not result in the TPM-based agent re-enrolling after the process restarts. the actual problem is being tracked and resolution is not blocking this PR.

pkg/poll/poll.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"time"
```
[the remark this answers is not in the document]
hexfusion: @asafbennatan this probably makes sense to consider.

internal/agent/device/publisher/publisher.go
```diff
@@ -125,6 +128,19 @@ func (n *publisher) pollAndPublish(ctx context.Context) {
 		n.log.Debugf("Dialing management API took: %v", duration)
 	}
 	if err != nil {
+		// Check for device not found error - handle certificate wiping and restart
+		if errors.Is(err, client.ErrDeviceNotFound) {
+			n.log.Warn("Device not found on management server")
```
hexfusion, line 133: question: is this an error I think this is very unexpected during normal operation

internal/agent/identity/tpm.go
lkatalin: In this question I was mainly trying to understand if TPM and non-TPM devices are being treated similarly so that we don't miss something

internal/agent/device/spec/manager.go
asafbennatan: The problem is that it compares against current rendered and not against desired
Consider (rendered =5 , desired=7 , incoming 6 ) the queue will change desired to 6

internal/agent/device/publisher/publisher.go
asafbennatan: If it's not an error what should it be ? I don't think that the frequency of an error should dictate if it's an error or not

pkg/poll/poll.go
asafbennatan: What do you mean ?  Wouldn't the struct still be documented, how does the constructor effects this ?

internal/agent/device/publisher/publisher.go
hexfusion: The log level is at warn `n.log.Warn` should that be `n.log.Error`?

internal/agent/identity/tpm.go
hexfusion: Yeah these are 2 seperate conversations that was my fault in not separating.

pkg/poll/poll.go
hexfusion: Its just cosmetic but my point is the field name document parameters for free.
```
poll.NewConfig{
        MaxDelay:     1 * time.Minute,
		BaseDelay:    10 * time.Second,
		Factor:       1.5,
		MaxSteps:     a.config.PullRetrySteps,
		JitterFactor: 0.1,
}
```	
vs
```
infiniteRetryConfig := *poll.NewConfig(
		10*time.Second, // baseDelay (same as management client)
		1.5,            // factor (same as management client)
		1*time.Minute,  // maxDelay (same as management client)
		0,              // maxSteps (0 means infinite retries until context timeout)
		0.1,            // jitterFactor (10% jitter to prevent thundering herd)
	)
```

internal/agent/device/spec/manager.go
hexfusion: I am fine with the logic my misunderstanding came from the naming 
```
currentVersion, err := stringToInt64(s.cache.getRenderedVersion(Desired))
```
I read as `currentVersion` but in this case it is the `currentDesiredVersion`.  The justification makes sense we should add some tests around this but can be a followup.

internal/agent/device/publisher/publisher.go
asafbennatan: fixed

internal/agent/device/spec/manager.go
asafbennatan: renamed to currentDesiredVersion

pkg/poll/poll.go
asafbennatan: we dont need the constructor anyway so removed

pkg/poll/poll.go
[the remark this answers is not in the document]
asafbennatan: fixed

test/util/create_utils.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math/rand"
```
hexfusion, line 7: go v1.22 introduced `math/rand/v2`, we can consider moving in a followup 

ref. https://go.dev/blog/randv2

PR #1596, opened 2025-09-02
EDM-941: updating application doesn't result in new application deployed on device

jira-ai-issue-solver:
This PR addresses the issue described in [EDM-941](https://issues.redhat.com/browse/EDM-941).

**Summary:** updating application doesn't result in new application deployed on device

**Description:** *Description of the problem:*

This is most likely caused by the logic based on updating only on tag update (new tag?).

While it might make sense to tag all version of the Standard Operating Environment system, it doesn't for applications, where most of the times you push always to latest tag (alternatively to beta for new functionalities).

There is also most likely some image caching involved somewhere, because on removing and readding the application to the device configuration, I would expect flightctl to pull the latest version of the tag, while it is not what's happening.

*How reproducible:*

 

*Steps to reproduce:*
 # Add application to device configuration

 #  

Update application image on container registry

 #  

No update triggered on application on device

 # removing the application from device configuration and readding it doesn't make the trick either

*Actual results:*

 

*Expected results:*

Application updates on devices triggered by new sha values of tag (in the end that's already supported on podman autoupdate)

**Assignee:** Asaf Ben Natan (abennata@redhat.com)

### DIFF
--- a/internal/agent/device/applications/provider/embedded.go
+++ b/internal/agent/device/applications/provider/embedded.go
@@ -60,7 +60,7 @@ func (p *embeddedProvider) OCITargets(pullSecret *client.PullSecret) ([]dependen
 				targets = append(targets, dependency.OCIPullTarget{
 					Type:       dependency.OCITypeImage,
 					Reference:  svc.Image,
-					PullPolicy: v1alpha1.PullIfNotPresent,
+					PullPolicy: v1alpha1.PullAlways,
 					PullSecret: pullSecret,
 				})
 			}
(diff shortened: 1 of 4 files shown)


jira-ai-issue-solver:
🤖 AI Processing Timestamp: 2025-09-02T08:26:57Z

AI has processed feedback for ticket EDM-941 at this time. Future processing will only consider feedback submitted after this timestamp.

jira-ai-issue-solver:
🤖 AI Processing Timestamp: 2025-09-02T08:31:25Z

AI has processed feedback for ticket EDM-941 at this time. Future processing will only consider feedback submitted after this timestamp.

PR #1610, opened 2025-09-04
EDM-2110: explicit rbac support from openapi.yaml

asafbennatan:
this adds supports for x-rbac extension in openapi.yaml which generates api_metadata_registry.gen.go
this contains metadata for the api that can be obtained based on the url , in this usecase we are extracting the x-rbac data and enforcing rbac based on it .

i have also granted the required permissions for the resume operation for the existing k8 roles.

we can use this for all operations or not , out choice , meanwhile if rbac details are not found based on the above approch we continue parsing them from the url like we did before


avishayt:
LGTM.  @fzdarsky WDYT?

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -63,6 +63,13 @@ rules:
     resources:
       - repositories
       - fleets/templateversions
+  # Device lifecycle actions (granular subresource approach)
+  - verbs:
+      - '*'  # All verbs on resume subresource
+    apiGroups:
+      - flightctl.io
+    resources:
+      - device/resume
```
celdrake: This is inconsistent with the other definitions, see example https://github.com/flightctl/flightctl/blob/a85f6257862ea97fb71d6f20a725d4c806b85af5/deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml#L57.

```suggestion
      - devices/resume
```

hexfusion (commented):
We talked about this in slack but there are some open questions on how native RBAC API's will look for standalone. Should this change be part of that greater conversation?

cc @DakCrowder

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
asafbennatan: fixed - it makes sense to have this as devices/resume , but actually it doesnt make sense for console to be devices/resume (since it operates on a single device)

asafbennatan:
@hexfusion i think how we can do rbac in standalone is a separate conversation , this was done here just so we can get reasonable action and resource for this api , i do think that adding x-rbac is useful on a bigger scope especially if we plan to remain compatible with k8 rbac and not create separate conventions for standalone

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
celdrake: I thought it's more a pattern for all the "resources" to always use the plural.

asafbennatan: youre right , IDK why i thought our url was .../device/{name} for some reason , in any case the above was fixed

api/v1alpha1/openapi.yaml
```diff
@@ -1531,6 +1531,9 @@ paths:
         - deviceactions
       description: Resume devices based on label selector and/or field selector.
       operationId: resumeDevices
+      x-rbac:
+        resource: devices/resume
+        action: resume
```
hexfusion: question: i am trying to understand what action means in this context, could it ever be anything other than resume?

asafbennatan: no , but according to RBAC k8 specs action can only be one of a very short list of list,update,create etc
in this case it doesnt make sense to use one of these so we use a "generic" resume , this requires the k8 rules to have action = * (wildcard)
and have the resource include the actual thing we are doing (devices/resume)

but - looking into this further - i guess it makes more sense to change the action to update so it will work with existing tooling

asafbennatan: no , but according to RBAC k8 specs action can only be one of a very short list of list,update,create etc
in this case it doesnt make sense to use one of these so we use a "generic" resume , this requires the k8 rules to have action = * (wildcard)
and have the resource include the actual thing we are doing (devices/resume)

but - looking into this further - i guess it makes more sense to change the action to update so it will work with existing tooling

hexfusion: > but - looking into this further - i guess it makes more sense to change the action to update so it will work with existing tooling

is `action` (update) then a `verb`?

hexfusion: In our case these are inlined into the spec I am just trying to understand this change. if you add docs to the types that would make this more clear ;)

asafbennatan: it is an action , k8 implementation of rbac includes only the http verbs as possible actions - to my understanding

asafbennatan: it is an action , k8 implementation of rbac includes only the http verbs ( and some additional ones)  as possible actions - to my understanding , so for this specific implementation this would be a verb

tools/api-metadata-extractor/go.mod
```diff
@@ -0,0 +1,7 @@
+module api-metadata-extractor
+
+go 1.22.5
```
hexfusion: nit: should we align with the repo on 1.23.0

ref. https://github.com/flightctl/flightctl/blob/7a7136932bd5785438386538e4fa9f3f329c766a/go.mod#L3

tools/api-metadata-extractor/go.mod
```diff
@@ -0,0 +1,7 @@
...
+
+toolchain go1.23.9
+
+require gopkg.in/yaml.v3 v3.0.1
```
hexfusion: nit: in a lot of areas of the code base we use `sigs.k8s.io/yaml` would it makes sense for us to align on a single yaml package?

tools/api-metadata-extractor/go.mod
asafbennatan: fixed

tools/api-metadata-extractor/main.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// EndpointMetadata contains metadata for an API endpoint
+type EndpointMetadata struct {
+	URLPattern  string
+	Method      string
+	OperationID string
+	Resource    string
+	Action      string
+}
```
hexfusion, line 60: nit: can you please add brief docs to these fields

tools/api-metadata-extractor/go.mod
asafbennatan: fixed

tools/api-metadata-extractor/main.go
asafbennatan: added

tools/api-metadata-extractor/go.mod
asafbennatan: note that sigs.k8s.io/yaml does not support inline , i have adjusted regardless

internal/api/server/docs.go
```diff
@@ -1,3 +1,4 @@
 package server
 
 //go:generate go run -modfile=../../../tools/go.mod github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen --config=server.gen.cfg ../../../api/v1alpha1/openapi.yaml
+//go:generate go run ../../../tools/api-metadata-extractor/main.go ../../../api/v1alpha1/openapi.yaml api_metadata_registry.gen.go server
```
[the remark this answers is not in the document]
hexfusion: @asafbennatan was we review this change request it seems logical. I just want to make sure we are not introducing a new pattern if its not required. :pray:

tools/api-metadata-extractor/go.mod
```diff
@@ -0,0 +1,7 @@
+module api-metadata-extractor
+
+go 1.23.0
+
+toolchain go1.23.9
+
+require sigs.k8s.io/yaml v1.5.0
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/api/server/docs.go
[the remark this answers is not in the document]
asafbennatan: fixed according to the ai comment

internal/auth/middleware.go
```diff
@@ -82,26 +88,38 @@ func CreateAuthZMiddleware(log logrus.FieldLogger) func(http.Handler) http.Handl
...
+					}
+					if len(parts) < 3 || matchesAPIVPath || (len(parts) >= 4 && parts[3] == "auth") {
+						next.ServeHTTP(w, r)
+						return
+					}
+
+					parts = parts[3:]
+					resource, action = extractResourceAndAction(parts, r.Method)
+				}
 			}
```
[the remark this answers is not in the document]
asafbennatan: this was also like that before this PR , if we do want to allow HEAD or Options and support other api versions - this will be done in its own PR

tools/api-metadata-extractor/main.go
```diff
@@ -0,0 +1,195 @@
...
+// APIMetadataMap provides O(1) lookup for endpoint metadata using pattern+method as key
+var APIMetadataMap = map[string]EndpointMetadata{
+{{- range .Endpoints}}
+	"{{.Method}}:{{.URLPattern}}": {
+		OperationID: "{{.OperationID}}",
+		Resource:    "{{.Resource}}",
+		Action:      "{{.Action}}",
+	},
+{{- end}}
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/middleware.go
```diff
@@ -82,26 +88,38 @@ func CreateAuthZMiddleware(log logrus.FieldLogger) func(http.Handler) http.Handl
...
+			if resource == "" || action == actionNil {
+				if r.URL.Path == "/api/version" {
+					resource = "version"
+					var ok bool
+					if action, ok = defaultActions[r.Method]; !ok {
+						action = actionNil
+					}
+				} else {
+					parts := strings.Split(r.URL.Path, "/")
+					// /, /api, /api/v{api-version} and /api/v{api-version}/auth don't require permissions
```
hexfusion, line 110: nit

asafbennatan: the previous implementation excluded the mentioned path from permission checking , the comment just mentions this fact

PR #1615, opened 2025-09-04
EDM-2072: db migration dry run hook and other Helm enhancements

sk-ilya:
The main changes include:
- DB setup is now executed post-install and pre-upgrade (previously post-upgrade).
- DB migration **dry run** is executed pre-upgrade. 
- The following parameters are configurable:
```yaml  
db.migration:
    backoffLimit: 3
    activeDeadlineSeconds: 600
    dryRun:
      preUpgrade:
        enabled: true
        hookWeight: "-10"
```
- The DB migration hook deletion policy now includes before-hook-creation to avoid blocking upgrades/rollbacks if the hook failed and was not deleted.

Also: `flightctl-db-app-secret.yaml` is aligned with `flightctl-db-admin-secret.yaml`. The original version was failing on `helm upgrade`

## Summary by CodeRabbit

- New Features
  - Configurable database migration Job with optional pre-upgrade dry-run, hook weighting, backoff and timeout controls.

- Refactor
  - Migration Job content centralized into a reusable template; orchestrator emits dry-run and real runs.
  - Canonicalized DB password lookup to reuse a single password across namespaces; prevent duplicate namespace targets.

- Behavior
  - DB type/port/name are now required (missing values cause rendering to fail).
  - PersistentVolumeClaims omit storageClass when unset.

- Documentation
  - Added values documentation for DB setup and migration parameters.

- Style
  - Minor whitespace/formatting cleanup in secret handling.

- Chore
  - Example storageClass entry added (commented) in values.

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-alert-exporter-config.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-alert-exporter-config.yaml
@@ -10,9 +10,9 @@ data:
   config.yaml: |-
     database:
         hostname: {{ include "flightctl.dbHostname" . }}
-        type: {{ $.Values.db.type | quote }}
-        port: {{ $.Values.db.port }}
-        name: {{ $.Values.db.name }}
+        type: {{ required "Values.db.type is required" $.Values.db.type | quote }}
+        port: {{ required "Values.db.port is required" $.Values.db.port }}
+        name: {{ required "Values.db.name is required" $.Values.db.name }}
     service:
       alertPollingInterval: "1m"
     kv:
--- a/deploy/helm/flightctl/templates/flightctl-api-config.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-api-config.yaml
@@ -8,9 +8,9 @@ data:
   config.yaml: |-
     database:
         hostname: {{ include "flightctl.dbHostname" . }}
-        type: {{ $.Values.db.type | quote }}
-        port: {{ $.Values.db.port }}
-        name: {{ $.Values.db.name }}
+        type: {{ required "Values.db.type is required" $.Values.db.type | quote }}
+        port: {{ required "Values.db.port is required" $.Values.db.port }}
+        name: {{ required "Values.db.name is required" $.Values.db.name }}
     service:
         address: :3443
         agentEndpointAddress: :7443
--- a/deploy/helm/flightctl/templates/flightctl-db-admin-secret.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-admin-secret.yaml
@@ -1,5 +1,5 @@
 {{- $namespaces := list .Release.Namespace }}
-{{- if .Values.global.internalNamespace }}
+{{- if and .Values.global.internalNamespace (ne .Values.global.internalNamespace .Release.Namespace) }}
 {{- $namespaces = append $namespaces .Values.global.internalNamespace }}
 {{- end }}
 {{- /* Look for the secret in any target namespace to keep one canonical password */}}
--- a/deploy/helm/flightctl/templates/flightctl-db-app-secret.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-app-secret.yaml
@@ -1,5 +1,5 @@
 {{- $namespaces := list .Release.Namespace }}
-{{- if .Values.global.internalNamespace }}
+{{- if and .Values.global.internalNamespace (ne .Values.global.internalNamespace .Release.Namespace) }}
 {{- $namespaces = append $namespaces .Values.global.internalNamespace }}
 {{- end }}
 {{- /* Look for the secret in any target namespace to keep one canonical password */}}
--- a/deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-deployment.yaml
@@ -28,7 +28,7 @@ spec:
       containers:
         - env:
             - name: POSTGRESQL_DATABASE
-              value: {{ $.Values.db.name }}
+              value: {{ required "Values.db.name is required" $.Values.db.name }}
             - name: POSTGRESQL_MASTER_PASSWORD
               valueFrom:
                 secretKeyRef:
--- a/deploy/helm/flightctl/templates/flightctl-db-migration-config.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-migration-config.yaml
@@ -7,9 +7,9 @@ data:
   config.yaml: |-
     database:
         hostname: {{ include "flightctl.dbHostname" . }}
-        type: {{ $.Values.db.type | quote }}
-        port: {{ $.Values.db.port }}
-        name: {{ $.Values.db.name }}
+        type: {{ required "Values.db.type is required" $.Values.db.type | quote }}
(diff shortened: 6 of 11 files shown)


hexfusion:
@rawagner PTAL

deploy/helm/flightctl/values.yaml
```diff
@@ -37,7 +37,7 @@ global:
   baseDomainTls:
     cert: ""
     key: ""
-  storageClassName: ""
+  storageClassName: "standard"
```
rawagner: `standard` storage class does not exist on every cluster.

deploy/helm/flightctl/templates/_db-migration-job.tpl
```diff
@@ -0,0 +1,192 @@
...
+        app: flightctl-db-migration
+        release: {{ $ctx.Release.Name }}
+    spec:
+      restartPolicy: Never
+      serviceAccountName: flightctl-db-migration
+      initContainers:
+      {{- include "flightctl.databaseWaitInitContainer" (dict "context" $ctx "userType" "admin" "timeout" 120 "sleep" 2) | nindent 6 }}
+      {{- if not $isDryRun }}
+      - name: setup-database-users
+        image: "{{ $ctx.Values.dbSetup.image.image }}:{{ $ctx.Values.dbSetup.image.tag | default $ctx.Chart.AppVersion }}"
```
rawagner: lets be consistent with usage of `default`
```suggestion
        image: "{{ $ctx.Values.dbSetup.image.image }}:{{ default $ctx.Chart.AppVersion $ctx.Values.dbSetup.image.tag }}"
```

deploy/helm/flightctl/templates/_db-migration-job.tpl
```diff
@@ -0,0 +1,192 @@
...
+          PGPASSWORD="$DB_ADMIN_PASSWORD" psql -v ON_ERROR_STOP=1 -h "$DB_HOST" -p "$DB_PORT" -U "$DB_ADMIN_USER" -d "$DB_NAME" -f "$SQL_FILE"
+
+          # Clean up temporary file
+          rm -f "$SQL_FILE"
+
+          echo "Database users setup completed successfully!"
+      {{- end }}
+      containers:
+      - name: run-migrations
+        image: "{{ $ctx.Values.dbSetup.image.image }}:{{ $ctx.Values.dbSetup.image.tag | default $ctx.Chart.AppVersion }}"
```
rawagner:
```suggestion
        image: "{{ $ctx.Values.dbSetup.image.image }}:{{ default $ctx.Chart.AppVersion $ctx.Values.dbSetup.image.tag }}"
```

deploy/helm/flightctl/templates/_db-migration-job.tpl
```diff
@@ -0,0 +1,192 @@
...
+      - name: setup-database-users
+        image: "{{ $ctx.Values.dbSetup.image.image }}:{{ $ctx.Values.dbSetup.image.tag | default $ctx.Chart.AppVersion }}"
+        imagePullPolicy: {{ default $ctx.Values.global.imagePullPolicy $ctx.Values.dbSetup.image.pullPolicy }}
+        env:
+        - name: DB_HOST
+          value: "{{ include "flightctl.dbHostname" $ctx }}"
+        - name: DB_PORT
+          value: "{{ (default 5432 $ctx.Values.db.port) }}"
+        - name: DB_NAME
+          value: "{{ (default "flightctl" $ctx.Values.db.name) }}"
```
rawagner: these `default` values are the defaults from values.yaml. Why specify them again ?
```suggestion
        - name: DB_PORT
          value: "{{ $ctx.Values.db.port }}"
        - name: DB_NAME
          value: "{{ $ctx.Values.db.name }}"
```

deploy/helm/flightctl/templates/_db-migration-job.tpl
```diff
@@ -0,0 +1,192 @@
...
+          echo "Migrations completed successfully!"
+
+          {{- if not $isDryRun }}
+          # Grant permissions on all existing tables to the application user
+          echo "Granting permissions on existing tables to application user..."
+          # Need to get admin credentials from init container environment
+          DB_HOST="{{ include "flightctl.dbHostname" $ctx }}"
+          # Get admin credentials from the same secrets used by init container
+          export PGPASSWORD="$DB_ADMIN_PASSWORD"
+          psql -h "$DB_HOST" -p {{ (default 5432 $ctx.Values.db.port) }} -U "$DB_ADMIN_USER" -d "{{ (default "flightctl" $ctx.Values.db.name) }}" -c "SELECT grant_app_permissions_on_existing_tables();"
```
rawagner:
```suggestion
          psql -h "$DB_HOST" -p {{ $ctx.Values.db.port }} -U "$DB_ADMIN_USER" -d "{{ $ctx.Values.db.name }}" -c "SELECT grant_app_permissions_on_existing_tables();"
```

deploy/helm/flightctl/templates/flightctl-db-migration-job.yaml
```diff
@@ -1,164 +1,34 @@
...
+{{- /* Defaults applied if any key along the path is missing */ -}}
+{{- $dbSetup := default (dict) .Values.dbSetup -}}
+{{- $migration := default (dict) $dbSetup.migration -}}
+{{- $dryRun := default (dict) $migration.dryRun -}}
+{{- $preUpgrade := default (dict) $dryRun.preUpgrade -}}
+
+{{- $dryRunEnabled := default true  $preUpgrade.enabled -}}
+{{- $dryRunHookWeight := default "-10" $preUpgrade.hookWeight -}}
+{{- $backoffLimit := default 3     $migration.backoffLimit -}}
+{{- $activeDeadlineSeconds := default 600   $migration.activeDeadlineSeconds -}}
```
rawagner: the defaults are already in `values.yaml`. Why default here again ? It will be a problem to keep these values in-sync as well.

deploy/helm/flightctl/values.yaml
sk-ilya: It is a configurable parameter. If not defined, k8s sets it to `standard` anyway (same behaviour as before), which will lead to an upgrade failure because Helm will think the spec attempts to change the value from `standard` to an empty string.

rawagner: no, it will set it to the storage class which is marked as `default`. Which can be anything.
If we want to make the upgrade possible a better way would be to get the default storage class via `lookup` and then in upgrade flows, do not try to override it.

deploy/helm/flightctl/templates/flightctl-db-migration-job.yaml
```diff
@@ -1,164 +1,43 @@
...
+  "hookDeletePolicy" "before-hook-creation,hook-succeeded"
+  "isDryRun" false
+) }}
+{{- else if .Release.IsInstall }}
+{{- include "flightctl.dbMigrationJob" (dict
+  "context" $root
+  "name" $baseName
+  "hooks" (list)
+  "isDryRun" false
+) }}
```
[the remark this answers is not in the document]
sk-ilya: For install - should start together with other services, no point in running it as a hook like when upgrading. It will fail the installation with `--wait`

rawagner (approved):

PR #1619, opened 2025-09-05
EDM-2016: Add Activate Credential infra

kkyrazis:
Adds the functionality to generate a credential blob and solve it. Adds a test to ensure the credentials are able to be generated and solved. This does not hook it up to anywhere yet. 

It is important that we know what type (ECC or RSA) of EK cert we've sent so that we can construct the TPM Public portion properly. For that reason I had to change the EKCert retrieval code a bit.

In the future the server will call `CreateCredentialChallenge` with the contents of the ER's CSR and the agent will solve the challenge. 

## Summary by CodeRabbit

* **New Features**
  * ECC-based Endorsement Key flows added alongside RSA (supports P-256/P-384/P-521).
  * New API to create TPM credential challenges that returns credential blob, encrypted secret, and expected secret.
  * Session-level abilities to generate and solve credential challenges.

* **Bug Fixes**
  * Improved session cleanup to flush keys and surface aggregated errors.

* **Tests**
  * End-to-end simulator tests and mocks updated to exercise RSA and ECC challenge flows.

### DIFF
--- a/internal/tpm/server.go
+++ b/internal/tpm/server.go
@@ -812,9 +812,8 @@ func convertEKPublicKeyToTPMTPublic(publicKey crypto.PublicKey) (*tpm2.TPMTPubli
 
 // convertEKECDSAPublicKey converts an ECDSA public key to TPM format
 func convertEKECDSAPublicKey(key *ecdsa.PublicKey) (*tpm2.TPMTPublic, error) {
-	// we currently only check the well known indexes of RSA2048 and ECC P256.
-	// We could support more, but we'd need to ensure that the Template's match up and
-	// only the P256 template is currently defined
+	// Currently only RSA2048 and ECC P256 EK Certs are supported. Should support for more ECC curves be added, this will
+	// need to change.
 	switch key.Curve.Params().Name {
 	case "P-256":
 	default:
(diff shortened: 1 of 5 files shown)


internal/tpm/server.go
```diff
@@ -728,3 +729,126 @@ func NormalizeEnrollmentCSR(csrString string) ([]byte, bool, error) {
...
+	case *rsa.PublicKey:
+		return convertEKRSAPublicKey(key)
+	default:
+		return nil, fmt.Errorf("unsupported public key type: %T", publicKey)
+	}
+}
+
+// convertEKECDSAPublicKey converts an ECDSA public key to TPM format
+func convertEKECDSAPublicKey(key *ecdsa.PublicKey) (*tpm2.TPMTPublic, error) {
+	// we currently only check the well known indexes of RSA2048 and ECC P256.
```
hexfusion: style: who is we :) its hard but try to use an imperative voice

internal/tpm/session.go
```diff
@@ -440,35 +440,76 @@ func (s *tpmSession) Close() error {
 	return nil
 }
 
+// checkEKCertExists checks if an EK certificate exists at the given NVRAM index
+// without reading the actual certificate data
+func (s *tpmSession) checkEKCertExists(nvIndex uint32) bool {
```
hexfusion: style: functions returning bool in Go are typically written as predicates so maybe `isEKCertPresent`

internal/tpm/session.go
```diff
@@ -440,35 +440,76 @@ func (s *tpmSession) Close() error {
...
+
+// detectEKAlgorithm determines the EK algorithm based on which certificate
+// exists in NVRAM, prioritizing RSA first
+func (s *tpmSession) detectEKAlgorithm() (KeyAlgorithm, error) {
+	// Try RSA first (maintaining current priority)
+	if s.checkEKCertExists(client.EKCertNVIndexRSA) {
+		return RSA, nil
+	}
+
+	// Try ECC second
```
hexfusion: good to know :)

kkyrazis: dang it siri.

hexfusion (approved):

internal/tpm/server.go
```diff
@@ -728,3 +729,127 @@ func NormalizeEnrollmentCSR(csrString string) ([]byte, bool, error) {
...
+	tpmPublic := tpm2.RSAEKTemplate
+
+	// Only replace the Unique field with the actual public key data
+	tpmPublic.Unique = tpm2.NewTPMUPublicID(
+		tpm2.TPMAlgRSA,
+		&tpm2.TPM2BPublicKeyRSA{Buffer: key.N.Bytes()},
+	)
+
+	return &tpmPublic, nil
+}
```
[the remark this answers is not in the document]
kkyrazis: The comment you suggested for deletion explains the decision. The templates specify the Key size directly and any adjustment to them may have policy changes that are currently unknown. I'd rather be restrictive than have it fail for something unknown.

PR #1620, opened 2025-09-06
NO-ISSUE: suitable output for CLI when resuming a single device that …

asafbennatan:
suitable output for CLI when resuming a single device that doesnt exist


avishayt (approved):

celdrake:
@asafbennatan @avishayt 
This behaviour is not well aligned with the rest of the commands.

As I described in the original PR, if I attempt to get an unexisting device, approve an unexisting ER, etc, I will get a 404.
In this PR we're returning statusCode 200 and no proper information if the devices does not exist.

So typos might not be caught etc.

PR #1623, opened 2025-09-07
EDM-1265: Enable task execution resilience after Redis restarts

avishayt:
Tracks the timestamp before which all tasks were completed and periodically saves it to the database. When Redis restarts and loses this timestamp, the system replays all events since the last saved timestamp to ensure no tasks are lost, then continues normal tracking.

This ensures all queued tasks are executed even after Redis failures, with idempotent task handling to safely re-execute tasks when needed.

## Summary by CodeRabbit

- New Features
  - Global checkpointing for task processing with forward-only advancement and cleanup.
  - Automatic recovery by replaying recent events across organizations.
  - Improved handling and visibility of failed and timed-out tasks, with safer replay and cross-service visibility.

- Documentation
  - Adds Key-Value Store & event-queue architecture doc.
  - Removes outdated tasks architecture doc.

- Tests
  - New integration tests covering checkpoint advancement, recovery, multi-organization replay, and Redis provider behaviors.


internal/periodic_checker/task_definitions.go
```diff
@@ -147,6 +147,14 @@ type QueueMaintenanceExecutor struct {
 func (e *QueueMaintenanceExecutor) Execute(ctx context.Context, log logrus.FieldLogger, orgID uuid.UUID) {
 	taskCtx := createTaskContext(ctx, PeriodicTaskTypeQueueMaintenance, orgID)
 
+	// Create publisher for event republishing during recovery
+	publisher, err := e.queuesProvider.NewPublisher(taskCtx, consts.TaskQueue)
```
asafbennatan: @avishayt i dont understand why are we creating a publisher here , it is also not used as far as i can see

internal/store/device.go
```diff
@@ -259,7 +259,7 @@ func (s *DeviceStore) createDeviceLabelsTrigger(db *gorm.DB) error {
 		triggerSQL := `
 		DROP TRIGGER IF EXISTS device_labels_insert ON devices;
 		DROP TRIGGER IF EXISTS device_labels_update ON devices;
-	
+
```
asafbennatan: consider removing the space changes to device

internal/tasks/queue_maintenance.go
```diff
@@ -50,9 +54,73 @@ func (t *QueueMaintenanceTask) Execute(ctx context.Context) error {
...
+			// Failed to parse the original event, just log an error
+			log.WithField("entryID", entryID).WithError(err).Error("Failed to parse original event from message body, skipping event emission")
+		}
+
+		return nil
+	})
+}
+
+// retryFailedMessages retries failed messages and handles permanently failed messages
+func (t *QueueMaintenanceTask) retryFailedMessages(ctx context.Context, log logrus.FieldLogger) (int, error) {
```
asafbennatan, line 122: anything preventing from the handler to be called more than retry count ? 
retryfailedmessages will still call the handler

avishayt: RetryFailedMessages will call the handler only if we exceeded retryCount.  The handler emits an event that this reconcile permanently failed.  Otherwise, if we didn't reach retryCount, the queue provider just re-publishes the event.

internal/periodic_checker/task_definitions.go
avishayt: Fixed

internal/store/device.go
avishayt: Fixed

asafbss (approved):

PR #1628, opened 2025-09-09
EDM-762: automatic helm chart docs generation

asafbennatan:
automatic helm chart docs generation

## Summary by CodeRabbit

* **New Features**
  * Expanded Helm chart configuration surface: DB & DB setup, key‑value store, API (rate limits, probes, baseUIUrl), Keycloak, Alertmanager, telemetryGateway, UI, worker/periodic/cli components and richer global knobs (ports, exposure, RBAC, metrics, tracing, image policies).

* **Documentation**
  * Added a templated Flightctl chart README plus README files for Keycloak and UI subcharts; includes installation, upgrade, uninstall, usage, configuration examples, environment-specific values, and values reference.

* **Chores**
  * Added helm-docs tooling, generation directive, and tool dependencies.


deploy/helm/flightctl/README.md
```diff
@@ -0,0 +1,255 @@
...
+|------------|------|---------|
+| keycloak | keycloak | 0.0.1 |
+| ui | ui | 0.0.1 |
+
+## Installation
+
+### Add Helm Repository
+
+```bash
+# If using a Helm repository (update with your actual repository)
```
sk-ilya: > update with your actual repository

deploy/helm/flightctl/README.md
```diff
@@ -0,0 +1,255 @@
...
+helm repo add flightctl https://flightctl.github.io/flightctl
+helm repo update
+```
+
+### Install Chart
+
+```bash
+# Install with default values
+helm install my-flightctl flightctl/flightctl
```
sk-ilya: There are a few mandatory params without which the installation will fail. 
More info in https://github.com/flightctl/flightctl/blob/main/docs/user/getting-started.md#deploying-the-flight-control-service in general I think this section kind of duplicates what we have in "Getting started"

deploy/helm/flightctl/README.md
```diff
@@ -0,0 +1,255 @@
...
+  target: "acm"
+  auth:
+    type: "k8s"
+    k8s:
+      apiUrl: "https://api.cluster.example.com:6443"
+```
+
+For more detailed configuration options, see the [Values](#values) section below.
+
+## Values
```
sk-ilya, line 80: The descriptions are missing. Shouldn't they be extracted from the `values.yaml`? 

Also, we have sections defined there as well (e.g., `## @section Global parameters`) which will make the doc more readable

sk-ilya:
Do we need to add codeowners for `values*.yaml`? @avishayt @hexfusion @fzdarsky  wdyt?

fzdarsky:
This is great, thank you!

Now, before going into text polishing, I'm wondering whether it'd make sense to generate the content into a `/docs/user/deploying-on-kubernetes.md` document instead / in addition?

The diff would be mainly in style, i.e. writing for a user that needs to understand the concepts and procedures plus tables of Helm config parameters vs. only a reference doc for the Helm config parameters. Might also have both: user doc with most common params, then linking for reference doc for details.

fzdarsky:
> Do we need to add codeowners for `values*.yaml`? @avishayt @hexfusion @fzdarsky wdyt?

It's an API, so that'd make sense to me.

deploy/helm/flightctl/README.md
fzdarsky: I agree we should refactor our docs a bit to remove duplication (related to my comment below). The "Getting started" was meant to be like a "Quickstart", i.e. a fool-proof, opinionated way of kicking the tyres, but now it includes many options that we should probably move into separate docs.

For simplicity, I'd probably keep refactoring to a separate PR and just do the minimum here to produce the reference doc.

asafbennatan:
> This is great, thank you!
> 
> Now, before going into text polishing, I'm wondering whether it'd make sense to generate the content into a `/docs/user/deploying-on-kubernetes.md` document instead / in addition?
> 
> The diff would be mainly in style, i.e. writing for a user that needs to understand the concepts and procedures plus tables of Helm config parameters vs. only a reference doc for the Helm config parameters. Might also have both: user doc with most common params, then linking for reference doc for details.

note that i think that the main use case here - is to have a readme displayed where we deploy the helm chart , we can have a separate docs/user/deploying-on-kubernetes.md - would that be generated ? can we auto generate info such that users will understand it ?
@fzdarsky

deploy/helm/flightctl/README.md
asafbennatan: we do not need a repo at all - we are using OCI charts , fixing in the readme

deploy/helm/flightctl/README.md
asafbennatan: note that all of these annotations we have in the values.yaml files are incorrectly written - they use java doc style and not the proper helm one , fixed , and now we got proper description

sk-ilya: That looks great now 👍

PR #1630, opened 2025-09-09
flightctl-worker metrics + configure redis as in-memory only

avishayt:
Two commits:

First:
    EDM-1269: worker metrics collection
    
    - Add WorkerCollector with metrics for message processing, task
      execution, and queue health
    - Enable worker metrics in Helm configmap and deployment templates
    
    Worker now exposes Prometheus metrics on port 8080 including message
    processing duration, task execution metrics, queue depth, and Redis
    connection health.
    
Second:
    EDM-2129: configure Redis as in-memory only
    
    Change Redis from persistent StatefulSet to in-memory Deployment in both
    Helm and podman deployments. This ensures our event replaying mechanism
    works correctly by guaranteeing that any Redis restart results in
    complete data loss, allowing the system to rebuild state from scratch
    rather than having inconsistent partial data from RDB/AOF persistence.
    
    Changes:
    - Convert Redis StatefulSet to Deployment in Helm charts
    - Remove persistent volumes and volumeClaimTemplates
    - Disable Redis persistence (save "", appendonly no)
    - Remove save configuration from Helm values.yaml
    - Update podman Redis container to remove persistent volume mount
    - Rename flightctl-kv-statefulset.yaml to flightctl-kv-deployment.yaml
    
    Upgrade Risk: Helm upgrade will fail due to StatefulSet→Deployment change.
    Manual StatefulSet deletion required before upgrade. All Redis data will
    be lost on restart (by design).
    
    Performance benefits: Faster startup, no storage I/O, simplified config.

## Summary by CodeRabbit

- **New Features**
  - Optional worker metrics with Prometheus endpoint (port 8080 when enabled) reporting processing counts, per-task/type durations, retries/failures, queue depth, active consumers, Redis health, and last-success timestamps; runs as a background metrics server.

- **Chores**
  - KV switched to ephemeral in-memory Redis (Deployment, no persistence), with liveness/readiness probes, memory constraints, and startup-generated config; worker container exposes metrics port conditionally; Helm deploy script updated for Deployment rollout; packaging bumped and KV removed from RPM.

- **Tests**
  - Added unit/integration tests for worker metrics exposition and queue maintenance metrics.


deploy/podman/flightctl-kv/flightctl-kv-config/redis.conf
```diff
@@ -1,18 +1,14 @@
...
-# AOF configuration
-appendonly yes
-appendfilename "appendonly.aof"
-appendfsync everysec
\ No newline at end of file
+# Disable persistence for in-memory Redis
+save ""
+appendonly no
+#
+# Prevent unbounded memory growth (tune to your environment)
```
ori-amizur: If it is tuned for environment, maybe it should be controlled by some configuration variable?

test/scripts/deploy_with_helm.sh
```diff
@@ -105,8 +105,6 @@ helm upgrade --install --namespace flightctl-external \
                   ${ONLY_DB} ${DB_SIZE_PARAMS} ${AUTH_ARGS} ${SQL_ARG} ${GATEWAY_ARGS} ${KV_ARG} ${ORGS_ARGS} flightctl \
               ./deploy/helm/flightctl/ --kube-context kind-kind
 
-kubectl rollout status statefulset flightctl-kv -n flightctl-internal -w --timeout=300s
```
ori-amizur, line 108: Don't we want to wait for redis even if it is memory-only ?

internal/tasks/consumer.go
```diff
@@ -21,15 +23,27 @@ import (
...
 		// Add timeout for the entire event processing
 		ctx, cancel := context.WithTimeout(ctx, EventProcessingTimeout)
 		defer cancel()
 
 		var eventWithOrgId worker_client.EventWithOrgId
 		if err := json.Unmarshal(payload, &eventWithOrgId); err != nil {
 			log.WithError(err).Error("failed to unmarshal consume payload")
+			// Record processing duration for unmarshal errors
+			if workerMetrics != nil {
+				workerMetrics.ObserveProcessingDuration(time.Since(startTime))
```
asafbennatan: did you intend to use a different meter here ?

deploy/podman/flightctl-kv/flightctl-kv-config/redis.conf
avishayt: Changed to environment variables for quadlets, and added docs explaining the configuration.

internal/tasks/consumer.go
avishayt: Great catch.  I changed this to be a permanent failure instead of looking like success.

test/scripts/deploy_with_helm.sh
avishayt: Sure, re-added it

ori-amizur (approved):

asafbss (approved):

PR #1631, opened 2025-09-09
EDM-551: edit command added

asafbennatan:
support for edit command

## Summary by CodeRabbit

- **New Features**
  - Added an in-place "edit" CLI command (TYPE/NAME or TYPE NAME) with editor override, YAML/JSON output, fleet scoping, patch-based updates, status messages, and recovery file saving; command registered in the CLI.

- **Documentation**
  - Updated device management guide with two workflows: recommended in-editor edit and export–edit–apply, including examples, editor customization, and error-recovery guidance.

- **Tests**
  - Added comprehensive tests for validation, editor resolution, formatting, patch calculation/apply, argument parsing, and recovery.

- **Chores**
  - Added JSON Patch dependency and updated embedded OpenAPI/Swagger payloads to support the "test" patch operation.


avishayt:
Do we still update resourceVersion on every agent heartbeat?

asafbennatan:
not sure , but i have changed the approch to use patch - so this should be less of an issue here

internal/cli/edit.go
```diff
@@ -0,0 +1,515 @@
...
+	case EventKind:
+		return fmt.Errorf("you cannot edit events")
+	case OrganizationKind:
+		return fmt.Errorf("you cannot edit organizations")
+	case TemplateVersionKind:
+		return fmt.Errorf("you cannot edit templateversions")
+	case CertificateSigningRequestKind:
+		return fmt.Errorf("you cannot edit certificatesigningrequests")
+	case EnrollmentRequestKind:
+		return fmt.Errorf("you cannot edit enrollmentrequests")
```
avishayt: You can do 'flightctl apply' for CSRs and ERs, so why not edit?

internal/cli/edit.go
```diff
@@ -0,0 +1,515 @@
...
+
+func NewCmdEdit() *cobra.Command {
+	o := DefaultEditOptions()
+	cmd := &cobra.Command{
+		Use:   "edit (TYPE NAME | TYPE/NAME)",
+		Short: "Edit a resource in your default editor.",
+		Long: `Edit a resource from your default command-line editor.
+
+The edit command allows you to directly edit an API resource. It opens the editor defined by --editor,
+then KUBE_EDITOR, then VISUAL, then EDITOR environment variables, or falls back to 'vi'.
```
avishayt:
```suggestion
then FLIGHTCTL_EDITOR, then VISUAL, then EDITOR environment variables, or falls back to 'vi'.
```

internal/cli/edit.go
```diff
@@ -0,0 +1,515 @@
...
+	fetchCancel()
+	if err != nil {
+		return fmt.Errorf("fetching %s/%s: %w", kind, name, err)
+	}
+
+	// Convert to the specified format for editing
+	originalContent, err := o.resourceToFormat(originalResource)
+	if err != nil {
+		return fmt.Errorf("converting resource to %s: %w", o.Output, err)
+	}
```
avishayt: Is it worth sharing code with internal/cli/get.go?

internal/cli/edit.go
asafbennatan: fixed

internal/cli/edit.go
asafbennatan: added some sharing between the two

internal/cli/edit.go
```diff
@@ -0,0 +1,577 @@
...
+
+	// 3. Filter the patch to protect immutable fields
+	var filteredOps []jsonpatchcreate.JsonPatchOperation
+
+	// Define paths that cannot be changed
+	protectedPaths := []string{
+		"/metadata/name",
+		"/kind",
+		"/apiVersion",
+	}
```
avishayt, line 293: The service returns 400 in case the user tries to change these, so better to pass them along and keep the behavior consistent.

avishayt (approved):

PR #1636, opened 2025-09-09
EDM-2087: add Helm upgrade + db migration dry run docs

sk-ilya:


sk-ilya:
@coderabbitai review

docs/user/database-migration.md
```diff
@@ -286,8 +305,9 @@ FROM (
 2. **Rotate passwords regularly** using your secret management system
 3. **Monitor database permissions** to ensure proper access restrictions
 4. **Test migrations** in non-production environments first
-5. **Review migration logs** for any permission or connectivity issues
-6. **Audit user access** - ensure services only have the minimum required credentials:
+5. **Use dry-run validation** to verify migration scripts before applying them in production
+6. **Review migration logs** for any permission or connectivity issues
+7. **Audit user access** - ensure services only have the minimum required credentials:
```
hexfusion, line 310: just a note I believe we can just use 1. for all of these and markdown handles this on rendering.

docs/user/getting-started.md
```diff
@@ -174,6 +174,127 @@ FC_API_URL=<API_URL>
 
 ```
 
+## Managing Flight Control Deployments
+
+### Upgrading Flight Control
```
hexfusion: This doesn't belong in the 'Getting Started' section. This is more to get the system up.  Upgrade is a day 2 opperation it feels like. Looking at what exists now there should certainly a section dedicated to upgrade.

here is something for ideas

ref. https://docs.redhat.com/en/documentation/openshift_container_platform/4.19#Upgrade%20and%20migrate

docs/user/database-migration.md
sk-ilya: Yep, that'd work. But this needs to be changed consistently across all files

docs/user/getting-started.md
sk-ilya: Moved to the main Helm readme

deploy/helm/flightctl/README.md
```diff
@@ -36,10 +36,103 @@ helm install my-flightctl oci://quay.io/flightctl/charts/flightctl --namespace f
...
+* **Preview the diff before upgrading** with the [Helm diff plugin](https://github.com/databus23/helm-diff).
+
+### Rollbacks
+
+Use rollbacks to revert to a previously successful revision if an upgrade causes issues. Use `helm history` to identify the target revision, then roll back if needed.
+
+Show release history and see failure reasons in the DESCRIPTION column:
+
+```bash
+$ helm history my-flightctl
```
siserafin, line 86: I think that you should add the namespace to this command, since in the previous installation you used it:
```
 helm history flightctl --namespace flightctl

sk-ilya: Makes sense... But I think it would be better to remove it from here to be consistent with the Install Chart section

siserafin (approved):
lgtm

PR #1637, opened 2025-09-09
EDM-2016: TPM Activate Credential Service Side

kkyrazis:
Adds the service side implementation of the TPM Activation Credential flow. 

The current Agent Server allows for both gRPC and REST endpoints to exist on the same port. This piggybacks on that existing behavior and adds a new gRPC service to the existing server. The gRPC validation ensure that there is a valid cert, of which the agent will be supplying the enrollment credentials for. 

As the new grpc handlers require the services, I had to update the initialization logic of the agent server.

[Design Doc](https://docs.google.com/document/d/1NJQMWRVA0YJsU9vbsA3JgnoxKfl0TUxc2So9QGoPT_s/edit?tab=t.0#heading=h.aalzxayhzgtt)


api/v1alpha1/consts.go
```diff
@@ -160,6 +160,17 @@ const (
...
 	FlightCtlSystemResourceName = "flightctl-system"
+
+	// TPM Validation Reasons
+
+	// TPMVerificationFailedReason indicates a TPM Request failed initial validation
+	TPMVerificationFailedReason = "TPMVerificationFailed"
+	// TPMChallengeRequiredReason indicates a TPM Challenge is required
+	TPMChallengeRequiredReason = "TPMChallengeRequired"
+	// TPMChallengeFailedReason indicates that a TPM Challenge attempt failed
+	TPMChallengeFailedReason = "TPMChallengeFailed"
```
lkatalin, line 171: When do we use `ChallengeFailed` vs `VerificationFailed`?

internal/api_server/agentserver/enrollmentrequest_grpc.go
```diff
@@ -0,0 +1,241 @@
...
+	if status.Code != http.StatusOK {
+		if err := s.sendErrorResponse(stream, fmt.Sprintf("Enrollment request not found: %s", enrollmentRequestName)); err != nil {
+			s.log.Errorf("Failed to send error response: %v", err)
+		}
+		return nil, nil, fmt.Errorf("getting enrollment request %s: %s", enrollmentRequestName, status.Message)
+	}
+
+	tpmCond := api.FindStatusCondition(lo.FromPtr(enrollmentRequest.Status).Conditions, api.ConditionTypeEnrollmentRequestTPMVerified)
+	// 1. There should be a tpm verified condition at this point
+	// 2. The enrollment request should not already be verified
```
lkatalin, line 116: If it's showing up as already verified for some reason, that seems pretty odd

kkyrazis: In the happy path I agree. My main thought in including this is to prevent others from interfering with a verified request.

api/v1alpha1/consts.go
kkyrazis: `VerificationFailed` is currently only being used if the initial ER post fails validation while `ChallengeFailed` I have being used for the ActivateCredential handshake flow. I kept them separate so that I have the agent differentiate between, 'failure and don't try activating credentials' and 'failure but you could try again'

internal/api_server/agentserver/enrollmentrequest_grpc.go
```diff
@@ -0,0 +1,241 @@
...
+		return "", status.Error(codes.InvalidArgument, "expected challenge request")
+	}
+
+	enrollmentRequestName := challengeRequest.EnrollmentRequestName
+	return enrollmentRequestName, nil
+}
+
+// validateEnrollmentRequest validates the enrollment request and parses the TPM CSR
+func (s *AgentGrpcServer) validateEnrollmentRequest(ctx context.Context, stream pb.EnrollmentRequestService_PerformTPMChallengeServer, enrollmentRequestName string) (*api.EnrollmentRequest, *tpm.ParsedTCGCSR, error) {
+	enrollmentRequest, status := s.service.GetEnrollmentRequest(ctx, enrollmentRequestName)
```
hexfusion: question: in this context are we ok to use the service directly?

internal/api_server/agentserver/enrollmentrequest_grpc.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return challenge, nil
+}
+
+// receiveAndVerifyResponse receives the challenge response and verifies the secret
+func (s *AgentGrpcServer) receiveAndVerifyResponse(stream pb.EnrollmentRequestService_PerformTPMChallengeServer, challenge *tpm.CredentialChallenge) error {
+	responseMsg, err := stream.Recv()
+	if err != nil {
+		return status.Error(codes.Internal, "failed to receive challenge response")
```
hexfusion: do we want to handle special cases at all like io.EOF just wondering if 500 is always accurate here.

internal/api_server/agentserver/enrollmentrequest_grpc.go
```diff
@@ -0,0 +1,241 @@
...
+func (s *AgentGrpcServer) updateStatusToFailed(ctx context.Context, stream pb.EnrollmentRequestService_PerformTPMChallengeServer, enrollmentRequest *api.EnrollmentRequest, enrollmentRequestName string, errorMessage string) error {
+	condition := api.Condition{
+		Type:    api.ConditionTypeEnrollmentRequestTPMVerified,
+		Status:  api.ConditionStatusFalse,
+		Reason:  api.TPMChallengeFailedReason,
+		Message: fmt.Sprintf("TPM challenge failed: %s", errorMessage),
+	}
+	api.SetStatusCondition(&enrollmentRequest.Status.Conditions, condition)
+
+	_, status := s.service.ReplaceEnrollmentRequestStatus(ctx, enrollmentRequestName, *enrollmentRequest)
```
hexfusion: question: lets ensure this is the only status writer for er in this challenge flow I believe that is yes but we don't want to flap.

internal/api_server/agentserver/enrollmentrequest_grpc.go
```diff
@@ -0,0 +1,241 @@
...
+		Status:  api.ConditionStatusTrue,
+		Reason:  api.TPMChallengeSucceededReason,
+		Message: "TPM activate credential challenge completed successfully",
+	}
+	api.SetStatusCondition(&enrollmentRequest.Status.Conditions, condition)
+
+	_, status := s.service.ReplaceEnrollmentRequestStatus(ctx, enrollmentRequestName, *enrollmentRequest)
+	if status.Code != http.StatusOK {
+		if err := s.sendErrorResponse(stream, "Failed to update enrollment status"); err != nil {
+			s.log.Errorf("Failed to send error response: %v", err)
```
hexfusion: question: if response fails to send how is the agent handling this?

api/grpc/v1/enrollment.proto
```diff
@@ -0,0 +1,66 @@
+syntax = "proto3";
+
+package flightctl.v1;
+
+option go_package = "github.com/flightctl/flightctl/api/grpc/v1/grpc-v1";
+
+// EnrollmentRequestService provides methods for TPM-based enrollment challenges.
+service EnrollmentRequestService {
```
hexfusion: lets make this more general so we can consider extension
```suggestion
service Enrollment {
```

api/grpc/v1/enrollment.proto
```diff
@@ -0,0 +1,66 @@
...
+  // PerformTPMChallenge establishes a bidirectional stream to conduct the TPM challenge-response.
+  rpc PerformTPMChallenge(stream AgentTPMChallengeMessage) returns (stream ServerTPMChallengeMessage);
```
hexfusion: since we are in the TPMChallenge rpc already
```suggestion
  rpc TPMChallenge(stream AgentChallenge) returns (stream ServerChallenge);
```

internal/api_server/agentserver/grpc.go
```diff
@@ -52,6 +57,7 @@ func (s *AgentGrpcServer) PrepareGRPCService() *grpc.Server {
 			Timeout:           20 * time.Second, // Wait 20s for client response before closing
 		}))
 	pb.RegisterRouterServiceServer(server, s)
+	pb.RegisterEnrollmentRequestServiceServer(server, s)
```
hexfusion: question: while this is generally a good practice. is it possible that enrollment service lives on another port? we might of removed that option.

internal/api_server/agentserver/server.go
```diff
@@ -37,49 +37,53 @@ const (
...
-	orgResolver    *org.Resolver
+	log             logrus.FieldLogger
+	cfg             *config.Config
+	store           store.Store
+	ca              *crypto.CAClient
+	listener        net.Listener
+	queuesProvider  queues.Provider
+	tlsConfig       *tls.Config
+	orgResolver     *org.Resolver
+	agentGrpcServer *AgentGrpcServer
```
hexfusion, line 46: why do we need grpcServer and agentGrpcServer

internal/api_server/agentserver/enrollmentrequest_grpc.go
kkyrazis: There's updateToFailed and updateToSuccess. I'll collapse them into a single version.

internal/api_server/agentserver/server.go
kkyrazis: AgentGrpcServer implements an interface depended on by the api service for console connection management, and grpcServer is the raw server. Will update the AgentGrpcServer to contain this closing logic so that it doesn't have to leak out here.

internal/api_server/agentserver/enrollmentrequest_grpc.go
kkyrazis: I'm not sure I follow. As opposed to using the Store? My intention with the service is to take advantage of any business/eventing logic that may exist.

internal/api_server/agentserver/enrollmentrequest_grpc.go
kkyrazis: I'll spend some more time around this and see what I can come up with.

internal/api_server/agentserver/enrollmentrequest_grpc.go
kkyrazis: I'll confirm that this stream closes once the handshake is over, but in general the agent will handle failures and attempt to retry. In this particular case, on retry it'll observe that the status is ChallengeSucceeded and not redo the challenge.

internal/api_server/agentserver/grpc.go
kkyrazis: The server that runs on the agent port is already doing some logic to route grpc vs non-grpc traffic to appropriate handlers so it seemed appropriate to piggy back on the AgentGrpcServer. 

I don't see any reason why we couldn't create a completely separate port and bind the new server directly to it.  I know you had mentioned it previously, but while implementing it seemed appropriate to reuse the existing port. Would you prefer me to setup a new port?

internal/api_server/agentserver/enrollmentrequest_grpc.go
kkyrazis: added more error codes.

internal/api_server/agentserver/grpc.go
hexfusion: > Would you prefer me to setup a new port?

no I just want to ensure if we setup a seperate port, we do not also bring along the console service. I think it's fine for now just wanted to keep that in mind

PR #1641, opened 2025-09-10
EDM-2031: Added checking number of devices in each fleet test

hferber1:
## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end CLI scenario that validates device counts per fleet across creation, querying, device add/remove, and deletion workflows.
  * Covers multiple fleets and verifies count transitions (e.g., 0 → 1 → 0) throughout the lifecycle; uses CLI output parsing to assert device counts.
  * Note: the scenario was accidentally added twice (duplicate test blocks).

### DIFF
--- a/test/e2e/cli/cli_test.go
+++ b/test/e2e/cli/cli_test.go
@@ -20,7 +20,22 @@ var (
 	unspecifiedResource  = "Error: name must be specified when deleting"
 	resourceCreated      = `(200 OK|201 Created)`
 	strictfipsruntimeTag = "X:strictfipsruntime"
+	applyOperation       = "apply"
 )
+
+type fleetTestManager struct {
+	harness          *e2e.Harness
+	testID           string
+	fleetA           v1alpha1.Fleet
+	fleetB           v1alpha1.Fleet
+	device           v1alpha1.Device
+	uniqueFleetAYAML string
+	uniqueFleetBYAML string
+	uniqueDeviceYAML string
+	fleetAName       string
+	fleetBName       string
+	deviceName       string
+}
 
 // _ is a blank identifier used to ignore values or expressions, often applied to satisfy interface or assignment requirements.
 var _ = Describe("cli operation", func() {
@@ -36,7 +51,7 @@ var _ = Describe("cli operation", func() {
 			harness := e2e.GetWorkerHarness()
 
 			By("should error when creating incomplete fleet")
-			out, err := harness.CLIWithStdin(incompleteFleetYaml, "apply", "-f", "-")
+			out, err := harness.CLIWithStdin(incompleteFleetYaml, applyOperation, "-f", "-")
 			Expect(err).To(HaveOccurred())
 			Expect(out).To(ContainSubstring("fleet: failed to apply"))
 
@@ -46,7 +61,7 @@ var _ = Describe("cli operation", func() {
 
 			By("Should error when creating a device with decimal in percentages")
 
-			out, err = harness.CLI("apply", "-f", util.GetTestExamplesYamlPath("badfleetrequest.yaml"))
+			out, err = harness.CLI(applyOperation, "-f", util.GetTestExamplesYamlPath("badfleetrequest.yaml"))
 			Expect(err).To(HaveOccurred())
 			Expect(out).To(MatchRegexp(`doesn't match percentage pattern`))
 
@@ -55,12 +70,12 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			defer util.CleanupTempYAMLFile(uniqueFleetYAML)
 
-			out, err = harness.ManageResource("apply", uniqueFleetYAML)
+			out, err = harness.ManageResource(applyOperation, uniqueFleetYAML)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(out).To(ContainSubstring("201 Created"))
 
 			// Applying a 2nd time it should also work, the fleet is just updated
-			out, err = harness.ManageResource("apply", uniqueFleetYAML)
+			out, err = harness.ManageResource(applyOperation, uniqueFleetYAML)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(out).To(ContainSubstring("200 OK"))
 		})
@@ -96,7 +111,7 @@ var _ = Describe("cli operation", func() {
 			Expect(out).To(ContainSubstring(deviceID))
 
 			By("Should let you list fleets")
-			_, err = harness.CLIWithStdin(completeFleetYaml, "apply", "-f", "-")
+			_, err = harness.CLIWithStdin(completeFleetYaml, applyOperation, "-f", "-")
 			Expect(err).ToNot(HaveOccurred())
 			out, err = harness.CLI("get", "fleets")
 			Expect(err).ToNot(HaveOccurred())
@@ -154,7 +169,7 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			defer util.CleanupTempYAMLFile(uniqueDeviceYAML)
 
-			out, err := harness.ManageResource("apply", uniqueDeviceYAML)
+			out, err := harness.ManageResource(applyOperation, uniqueDeviceYAML)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(MatchRegexp(resourceCreated))
 
@@ -195,7 +210,7 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			defer util.CleanupTempYAMLFile(uniqueFleetYAML)
 
-			out, err = harness.ManageResource("apply", uniqueFleetYAML)
+			out, err = harness.ManageResource(applyOperation, uniqueFleetYAML)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(MatchRegexp(resourceCreated))
 			fleet := harness.GetFleetByYaml(uniqueFleetYAML)
@@ -229,7 +244,7 @@ var _ = Describe("cli operation", func() {
 			defer util.CleanupTempYAMLFile(uniqueRepoYAML)
 
 			Eventually(func() error {
-				out, err = harness.ManageResource("apply", uniqueRepoYAML)
+				out, err = harness.ManageResource(applyOperation, uniqueRepoYAML)
 				return err
 			}).Should(BeNil(), "failed to apply Repository")
 			Expect(out).To(MatchRegexp(resourceCreated))
@@ -241,7 +256,7 @@ var _ = Describe("cli operation", func() {
 			*repo.Metadata.Name = updatedName
 			repoData, err := yaml.Marshal(&repo)
 			Expect(err).ToNot(HaveOccurred())
-			out, err = harness.CLIWithStdin(string(repoData), "apply", "-f", "-")
+			out, err = harness.CLIWithStdin(string(repoData), applyOperation, "-f", "-")
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(ContainSubstring(updatedName))
 
@@ -253,7 +268,7 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			defer util.CleanupTempYAMLFile(uniqueResourceSyncYAML)
 
-			out, err = harness.ManageResource("apply", uniqueResourceSyncYAML)
+			out, err = harness.ManageResource(applyOperation, uniqueResourceSyncYAML)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(MatchRegexp(resourceCreated))
 			rSync := harness.GetResourceSyncByYaml(uniqueResourceSyncYAML)
@@ -263,7 +278,7 @@ var _ = Describe("cli operation", func() {
 			*rSync.Metadata.Name = rSyncNewName
 			rSyncData, err := yaml.Marshal(&rSync)
 			Expect(err).ToNot(HaveOccurred())
-			out, err = harness.CLIWithStdin(string(rSyncData), "apply", "-f", "-")
+			out, err = harness.CLIWithStdin(string(rSyncData), applyOperation, "-f", "-")
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(ContainSubstring(rSyncNewName))
 
@@ -275,7 +290,7 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			defer os.Remove(erYAMLPath)
 
-			out, err = harness.ManageResource("apply", erYAMLPath)
+			out, err = harness.ManageResource(applyOperation, erYAMLPath)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(MatchRegexp(resourceCreated))
 			er := harness.GetEnrollmentRequestByYaml(erYAMLPath)
@@ -286,7 +301,7 @@ var _ = Describe("cli operation", func() {
 			*er.Metadata.Name = erNewName
 			erData, err := yaml.Marshal(&er)
 			Expect(err).ToNot(HaveOccurred())
-			out, err = harness.CLIWithStdin(string(erData), "apply", "-f", "-")
+			out, err = harness.CLIWithStdin(string(erData), applyOperation, "-f", "-")
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(ContainSubstring(erNewName))
 
@@ -299,7 +314,7 @@ var _ = Describe("cli operation", func() {
 			defer util.CleanupTempYAMLFile(uniqueCsrYAML)
 
 			Eventually(func() error {
-				out, err = harness.CLI("apply", "-f", uniqueCsrYAML)
+				out, err = harness.CLI(applyOperation, "-f", uniqueCsrYAML)
 				return err
 			}).Should(BeNil(), "failed to apply CSR")
 			Expect(out).To(MatchRegexp(resourceCreated))
@@ -311,7 +326,7 @@ var _ = Describe("cli operation", func() {
 			*csr.Metadata.Name = csrNewName
 			csrData, err := yaml.Marshal(&csr)
 			Expect(err).ToNot(HaveOccurred())
-			out, err = harness.CLIWithStdin(string(csrData), "apply", "-f", "-")
+			out, err = harness.CLIWithStdin(string(csrData), applyOperation, "-f", "-")
 
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(ContainSubstring(csrNewName))
@@ -339,13 +354,13 @@ var _ = Describe("cli operation", func() {
 			Expect(err).ToNot(HaveOccurred())
 			defer util.CleanupTempYAMLFile(uniqueDeviceBYAML)
 
-			out, err := harness.ManageResource("apply", uniqueDeviceYAML)
+			out, err := harness.ManageResource(applyOperation, uniqueDeviceYAML)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(MatchRegexp(resourceCreated))
 			device1 := harness.GetDeviceByYaml(uniqueDeviceYAML)
 			device1Name := *device1.Metadata.Name
 
-			out, err = harness.ManageResource("apply", uniqueDeviceBYAML)
+			out, err = harness.ManageResource(applyOperation, uniqueDeviceBYAML)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(out).To(MatchRegexp(resourceCreated))
 			device2 := harness.GetDeviceByYaml(uniqueDeviceBYAML)
@@ -410,7 +425,7 @@ var _ = Describe("cli operation", func() {
 			}
 
 			for _, file := range applyResources {
-				out, err := harness.ManageResource("apply", file)
+				out, err := harness.ManageResource(applyOperation, file)
 				Expect(err).ToNot(HaveOccurred())
 				Expect(out).To(MatchRegexp(resourceCreated))
 			}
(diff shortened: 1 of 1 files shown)


test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
 		})
 	})
 
+	Context("Number of devices in fleet check", func() {
```
eldar101: Context("Verify fleet check shows devices", func() {

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+		It("Show number of devices associated with each fleet", Label("84266", "sanity"), func() {
+			// Get harness directly - no shared package-level variable
```
eldar101: Remove this comment please.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			harness := e2e.GetWorkerHarness()
+			testID := harness.GetTestIDFromContext()
+
+			// Creating a fleet
+			By("Creating an empty fleet")
```
eldar101: Remove the "empty"

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			uniqueFleetYAML, err := util.CreateUniqueYAMLFile("fleet.yaml", testID)
```
eldar101: Create a fleet using
`_, err := harness.CLIWithStdin(completeFleetYaml, "apply", "-f", "-")
			Expect(err).ToNot(HaveOccurred())`

and remove the redundant code such as testIDf

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			defer util.CleanupTempYAMLFile(uniqueFleetYAML)
```
eldar101: remove

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+
+			// Checking the fleet was created
+			out, err := harness.ManageResource("apply", uniqueFleetYAML)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(out).To(ContainSubstring("201 Created"))
```
eldar101: no need for this, we create fleets multiple times in this test and verifying it each time will just make the file huge and bloated.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+
+			// Check if number of devices is shown
+			By("Checking if number of devices is shown")
+			out, err = harness.CLI("get", "fleet -s | awk '{print $1, $5}' | column -t")
```
eldar101: we can possibly replace this with the results from the object
for example:

`// Example: build a map of fleetName → deviceCount
counts := map[string]int{}
for _, item := range result.Items {
    counts[item.Metadata.Name] = item.Status.Devices
}`

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring("DEVICES"))
```
eldar101: Put "DEVICES" in a constant at the top of the file since you're checking for it multiple times in this test.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			Expect(out).To(ContainSubstring("0"))
+
+			// Creating a device in the same fleet
+			By("Creating a device in the same fleet")
+			uniqueDeviceYAML, err := util.CreateUniqueYAMLFile("device.yaml", testID)
```
eldar101: I see that the flow for  device creation has changed, you can keep this as shown in line 333

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			defer util.CleanupTempYAMLFile(uniqueDeviceYAML)
+
+			// Checking the device was created and getting its name
+			out, err = harness.ManageResource("apply", uniqueDeviceYAML)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(MatchRegexp(resourceCreated))
```
eldar101: No need for this validation, we already do it in device validation tests.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			deviceOfFirstFleet := harness.GetDeviceByYaml(uniqueDeviceYAML)
+			deviceOfFirstFleetName := *deviceOfFirstFleet.Metadata.Name
+
+			// Checking if the number of devices is shown and changed
+			By("Checking if number of devices is shown and changed")
+			out, err = harness.CLI("get", "fleet -s | awk '{print $1, $5}' | column -t")
```
eldar101: like before, we can find a way to not make it hardcoded..please research this.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			Expect(out).To(ContainSubstring("1"))
+			Expect(out).ToNot(ContainSubstring("0"))
+
+			// Create another fleet
+			By("Creating another empty fleet")
+			uniqueFleetBYAML, err := util.CreateUniqueYAMLFile("fleet-b.yaml", testID)
+			Expect(err).ToNot(HaveOccurred())
+			defer util.CleanupTempYAMLFile(uniqueFleetBYAML)
+
+			// Checking the second fleet was created
```
eldar101: no need for repetitive fleet validation each time besides the original test that validates resource creation.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +461,86 @@ var _ = Describe("cli operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring("completed"))
+
+			// Check if the number is changed after deletion
+			By("Checking if number of devices is shown and changed")
+			out, err = harness.CLI("get", "fleet -s | awk '{print $1, $5}' | column -t")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring("DEVICES"))
+			Expect(out).To(ContainSubstring("0"))
+			Expect(out).ToNot(ContainSubstring("1"))
```
eldar101: you check the number of devices in fleet multiple times and it's the same check.
I suggest you put this in a helper func to avoid repetition of multiple lines.

test/e2e/cli/cli_test.go
```diff
@@ -624,6 +716,35 @@ func GetVersionByPrefix(output, prefix string) string {
 	return ""
 }
 
+// Comparing devices count output with fleet data in map
+func checkDeviceCountCLIOutput(output string, expected map[string]int64) {
```
eldar101:
```suggestion
func CheckDeviceCountCLIOutput(output string, expected map[string]int64) {
```

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,96 @@ var _ = Describe("cli operation", func() {
...
+			fleetName := *fleet.Metadata.Name
+			fleetSummary := *fleet.Status.DevicesSummary
+			fleetCount := fleetSummary.Total
+			fleetDevicesCount[fleetName] = fleetCount
+
+			// Checking if number of devices is shown
+			By("Checking if number of devices is shown")
+			out, err = harness.CLIWithStdin("get", "fleet", "-s")
+			Expect(err).ToNot(HaveOccurred())
+			checkDeviceCountCLIOutput(out, fleetDevicesCount)
```
eldar101: Fix the func's name capitalization please.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,96 @@ var _ = Describe("cli operation", func() {
...
+			// Checking the device was created and updating in map
+			out, err = harness.ManageResource("apply", uniqueDeviceYAML)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(out).To(MatchRegexp(resourceCreated))
+			fleetSummary = *fleet.Status.DevicesSummary
+			fleetCount = fleetSummary.Total
+			fleetDevicesCount[fleetName] = fleetCount
+
+			// Getting the device name
+			deviceOfFirstFleet := harness.GetDeviceByYaml(uniqueDeviceYAML)
```
eldar101: You later use FleetB,
So please just use one naming scheme...fleetA and FleetB for example

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,96 @@ var _ = Describe("cli operation", func() {
...
+			Expect(out).To(MatchRegexp(resourceCreated))
+			fleetB := harness.GetFleetByYaml(uniqueFleetBYAML)
+			fleetBName := *fleetB.Metadata.Name
+			fleetBSummary := *fleetB.Status.DevicesSummary
+			fleetBCount := fleetBSummary.Total
+			fleetDevicesCount[fleetBName] = fleetBCount
+
+			// Checking if the number of devices is shown for both fleets
+			By("Checking if number of devices is shown for both fleets")
+			out, err = harness.CLIWithStdin("get", "fleet", "-s")
```
eldar101: you use this multiple times, So I suggest you add it as a func to harness.go...
something like:

// RunGetFleets executes "get fleets" CLI command with optional arguments.
func (h *Harness) RunGetFleets(args ...string) (string, error) {
	allArgs := append([]string{"get", "fleets"}, args...)
	return h.CLIWithStdin(allArgs...)
}

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,96 @@ var _ = Describe("cli operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			checkDeviceCountCLIOutput(out, fleetDevicesCount)
+
+			// Deleting a device from the first fleet
+			By("Deleting a device from the first fleet")
+			out, err = harness.CLI("delete", util.Device, deviceOfFirstFleetName)
```
eldar101: You can delete resources with the same func you used to apply ..
harness.ManageResource

test/e2e/cli/cli_test.go
```diff
@@ -624,6 +716,35 @@ func GetVersionByPrefix(output, prefix string) string {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		// Comparing the count in output to the count in map
+		expectedCount := expected[name]
+		Expect(count).To(Equal(expectedCount))
+		matched++
+	}
+
+	// Making sure number of fleets (lines in output) match the number of fleets in map
+	Expect(matched).To(Equal(len(expected)))
```
eldar101: we don't put "expect" in funcs, only in tests.
Please replace it with a returned value andf check it's value in the test.

itzikb-redhat:
@hferber1 Please add a description to the PR

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,88 @@ var _ = Describe("cli operation", func() {
...
+			fleetDevicesCount := map[string]int64{}
+
+			// Creating a fleet
+			By("Creating a fleet")
+			uniqueFleetAYAML, err := util.CreateUniqueYAMLFile("fleet.yaml", testID)
+			Expect(err).ToNot(HaveOccurred())
+			defer util.CleanupTempYAMLFile(uniqueFleetAYAML)
+
+			// Checking the fleet was created and updating in map
+			out, err := harness.ManageResource("apply", uniqueFleetAYAML)
```
eldar101: please put the "apply" string in a var and replace it along the file...I've noticed it's been used a lot.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,88 @@ var _ = Describe("cli operation", func() {
...
+			By("Checking if number of devices is shown and changed")
+			notMatched, err = CheckDevicesInFleetStatus(harness, fleetDevicesCount)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(notMatched).To(BeEmpty())
+
+			// Creating another fleet
+			By("Creating another fleet")
+			uniqueFleetBYAML, err := util.CreateUniqueYAMLFile("fleet-b.yaml", testID)
+			Expect(err).ToNot(HaveOccurred())
+			defer util.CleanupTempYAMLFile(uniqueFleetBYAML)
```
eldar101: besides capitalization at the begininnig you have to follow the rules regarding CLI Too,
so it should be:
CheckDeviceCountCliOutput

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,88 @@ var _ = Describe("cli operation", func() {
...
+			fleetBName := *fleetB.Metadata.Name
+			fleetBCount := (*fleetB.Status.DevicesSummary).Total
+			fleetDevicesCount[fleetBName] = fleetBCount
+
+			// Checking if the number of devices is shown for both fleets
+			By("Checking if number of devices is shown for both fleets")
+			notMatched, err = CheckDevicesInFleetStatus(harness, fleetDevicesCount)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(notMatched).To(BeEmpty())
+
```
eldar101, line 553: you check total devices multiple times.
This duplicates code,
Please put this block in a helper func

eldar101:
Please try to maintain a structure where a by block ends with the expect lines.

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,88 @@ var _ = Describe("cli operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(ContainSubstring("completed"))
+			fleetACount = (*fleetA.Status.DevicesSummary).Total
+			fleetDevicesCount[fleetAName] = fleetACount
+
+			// Check if the number is changed after deletion
+			By("Checking if number of devices is shown and changed")
+			notMatched, err = CheckDevicesInFleetStatus(harness, fleetDevicesCount)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(notMatched).To(BeEmpty())
```
itzikb-redhat, line 565: You may want to output the difference in case there is one, otherwise you can just use a counter

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,88 @@ var _ = Describe("cli operation", func() {
...
+			fleetDevicesCount[fleetAName] = fleetACount
+
+			// Checking if the number of devices is shown and changed
+			By("Checking if number of devices is shown and changed")
+			notMatched, err = CheckDevicesInFleetStatus(harness, fleetDevicesCount)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(notMatched).To(BeEmpty())
+
+			// Creating another fleet
+			By("Creating another fleet")
```
itzikb-redhat, line 542: Can you create the two fleets together instead of one by one?

test/e2e/cli/cli_test.go
```diff
@@ -624,6 +708,47 @@ func GetVersionByPrefix(output, prefix string) string {
 	return ""
 }
 
+// Comparing devices count output with fleet data in map
+func CompareDeviceCountCLIOutput(output string, expected map[string]int64) []string {
```
itzikb-redhat: Are the two functions going to be used outside of this package?

test/e2e/cli/cli_test.go
```diff
@@ -461,6 +463,88 @@ var _ = Describe("cli operation", func() {
...
 
+	Context("Verify fleet check shows devices", func() {
+		It("Show number of devices associated with each fleet", Label("84266", "sanity"), func() {
+			harness := e2e.GetWorkerHarness()
+			testID := harness.GetTestIDFromContext()
+			fleetDevicesCount := map[string]int64{}
+
+			// Creating a fleet
+			By("Creating a fleet")
+			uniqueFleetAYAML, err := util.CreateUniqueYAMLFile("fleet.yaml", testID)
```
siserafin: To avoid the repetitions I would suggest to use a helper struct like:
```
type FleetTestManager struct {
    harness *e2e.Harness
    testID  string
    fleetA  *v1alpha1.Fleet
    fleetB  *v1alpha1.Fleet
    device  *v1alpha1.Device
}
```
And you can add methods to create the fleets and the device, remove a device from a fleet, get the number of devices associated to a fleet, verification methods...

test/e2e/cli/cli_test.go
```diff
@@ -624,6 +708,47 @@ func GetVersionByPrefix(output, prefix string) string {
...
+func CompareDeviceCountCLIOutput(output string, expected map[string]int64) []string {
+
+	var notMatched []string
+	lines := strings.Split(strings.TrimSpace(output), "\n")
+
+	for _, line := range lines {
+		columns := strings.Fields(line)
+
+		// Skipping the first line of output containing the titles
+		if columns[4] == deviceCountColumn {
```
siserafin: I would get the column number 4 from the expected title, and return an error if not found.  If in the future a column is added this function would fail, and it would be difficult to debug.

test/e2e/cli/cli_test.go
```diff
@@ -624,6 +708,47 @@ func GetVersionByPrefix(output, prefix string) string {
...
+		if count != expectedCount {
+			notMatched = append(notMatched, name)
+		}
+	}
+
+	// Making sure number of fleets (lines in output) match the number of fleets in map
+	return notMatched
+}
+
+// Checking the status of devices in fleet and devices count
```
eldar101, line 815: I suggest you add some log lines

test/e2e/cli/cli_test.go
```diff
@@ -624,6 +708,47 @@ func GetVersionByPrefix(output, prefix string) string {
...
+		// Skipping the first line of output containing the titles
+		if columns[4] == deviceCountColumn {
+			continue
+		}
+
+		name := columns[0]
+		strCount := columns[4]
+		count, err := strconv.ParseInt(strCount, 10, 64)
+		if err != nil {
+			notMatched = append(notMatched, name)
```
eldar101: log lines

test/e2e/cli/cli_test.go
siserafin: If you don't want to go with this solution, maybe you can create a local function to create a test fleet, and another to create the test device related to a fleet. Otherwise the same code is repeated twice.

eldar101:
Please fix according to Silvia's suggestions, rebase, pass checks, and I'll approve.

test/e2e/cli/cli_test.go
```diff
@@ -655,6 +728,102 @@ func GetVersionByPrefix(output, prefix string) string {
...
+	// Checking the device was created and updating in map
+	out, err := fleetTestManager.harness.ManageResource(applyOperation, uniqueDeviceYAML)
+	fleetTestManager.device = fleetTestManager.harness.GetDeviceByYaml(uniqueDeviceYAML)
+	fleetTestManager.deviceName = *fleetTestManager.device.Metadata.Name
+	fleetTestManager.uniqueDeviceYAML = uniqueDeviceYAML
+
+	if strings.Contains(fleetIdentifier, "fleetA") {
+		fleetDevicesCount[fleetTestManager.fleetAName]++
+	} else if strings.Contains(fleetIdentifier, "fleetB") {
+		fleetDevicesCount[fleetTestManager.fleetAName]++
```
siserafin: Should this be fleetBName?

test/e2e/cli/cli_test.go
```diff
@@ -492,6 +507,64 @@ var _ = Describe("cli operation", func() {
...
+	Context("Verify fleet check shows devices", func() {
+		It("Show number of devices associated with each fleet", Label("84266", "sanity"), func() {
+			harness := e2e.GetWorkerHarness()
+			testID := harness.GetTestIDFromContext()
+			fleetDevicesCount := map[string]int64{}
+			fleetManager := fleetTestManager{harness: harness, testID: testID}
+
+			By("Creating a fleet")
+			out, err := createTestFleet(&fleetManager, fleetDevicesCount, "fleet.yaml", "fleetA")
+			defer util.CleanupTempYAMLFile(fleetManager.uniqueFleetAYAML)
```
siserafin: I think that the defer cleanup should happen after confirming the YAML file is created, or it could be called on unset values

test/e2e/cli/cli_test.go
```diff
@@ -492,6 +507,64 @@ var _ = Describe("cli operation", func() {
...
+			// Checking if number of devices is shown
+			By("Checking if number of devices is shown")
+			notMatched, err := checkDevicesInFleetStatus(harness, fleetDevicesCount)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(notMatched).To(BeEmpty())
+
+			// Creating a device in the same fleet
+			By("Creating a device in the same fleet")
+			out, err = createDeviceInFleet(&fleetManager, fleetDevicesCount, "device.yaml", "fleetA")
+			defer util.CleanupTempYAMLFile(fleetManager.uniqueDeviceYAML)
```
siserafin: same here

test/e2e/cli/cli_test.go
```diff
@@ -492,6 +507,64 @@ var _ = Describe("cli operation", func() {
...
+
+			// Checking if the number of devices is shown and changed
+			By("Checking if number of devices is shown and changed")
+			notMatched, err = checkDevicesInFleetStatus(harness, fleetDevicesCount)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(notMatched).To(BeEmpty())
+
+			By("Creating another fleet")
+			out, err = createTestFleet(&fleetManager, fleetDevicesCount, "fleet-b.yaml", "fleetB")
+			defer util.CleanupTempYAMLFile(fleetManager.uniqueFleetBYAML)
```
siserafin: same here

PR #1648, opened 2025-09-11
NO-ISSUE: agent/device/spec: ensure agent never goes backwards

hexfusion:
Thanks to @asafbennatan for pointing this issue out.

## Summary by CodeRabbit

- New Features
  - Queue and manager now share a single rendered-state cache for consistent state.

- Bug Fixes
  - Enqueue logic rejects older spec versions based on current rendered state to avoid stale work and divergence.
  - Retry and delay handling now apply to the proposed version to prevent incorrect retry accounting.

- Tests
  - Added tests verifying rejection of older versions and acceptance/processing of newer ones.

- Refactor
  - Initialization updated to wire the shared cache and gating behavior between queue and manager.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -344,3 +344,11 @@ TN1330
 - docs/user/metrics.md
 sloMax
 SLO
+- docs/user/postgresql-backup-strategy.md
+RPOs
+RPO
+Bitbucket
+GitLab
+GitHub
+Configs
+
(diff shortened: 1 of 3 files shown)


internal/agent/device/spec/manager.go
```diff
@@ -47,11 +47,13 @@ func NewManager(
 	devicePublisher publisher.Subscription,
 	log *log.PrefixLogger,
 ) Manager {
-	queue := newPriorityQueue(
+	cache := newCache(log)
```
kkyrazis, line 50: I can't comment on the correct line, but I think this should be shared with the manager right? (line 65)

PR #1651, opened 2025-09-12
EDM-2058: agent/client: add retry transport and RequestHook

hexfusion:
```
hook404 := client.WithResponseHook(func(resp *http.Response, attempt int) bool {
    if resp.StatusCode == 404 {
        log.Errorf("Device not found (404) on attempt %d", attempt)
        // Run special logic for 404
        handleMissingDevice("device-123")
        // Don't retry on 404s
        return false  
    }
    return true // Continue with normal retry logic
})
```

## Summary by CodeRabbit

- New Features
  - Adds automatic HTTP retry with exponential backoff for network calls, improving reliability under transient errors (e.g., 429 and 5xx).
  - Honors Retry-After headers and caps delays to prevent long waits.
  - Retries preserve request bodies for safe re-attempts on POST/other methods.
  - Enhanced retry logging for better observability during network issues.

- Chores
  - Internal updates to adopt the new retry-capable HTTP client across components.

### DIFF
--- a/internal/agent/client/transport.go
+++ b/internal/agent/client/transport.go
@@ -1,0 +1,154 @@
+package client
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/poll"
+)
+
+type contextKey string
+
+const responseHookKey contextKey = "responseHook"
+
+// ResponseHook is called before retry decision is made
+// Return true to continue with retry logic, false to stop retrying
+type ResponseHook func(resp *http.Response, attempt int) bool
+
+type RetryTransport struct {
+	transport  http.RoundTripper
+	pollConfig poll.Config
+	log        *log.PrefixLogger
+}
+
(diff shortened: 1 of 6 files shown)


internal/agent/client/transport.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+			return resp, nil
+		}
+
+		wait := poll.CalculateBackoffDelay(&r.pollConfig, attempt+1)
+
+		// check Retry-After header
+		if retryAfter := r.parseRetryAfter(resp); retryAfter > 0 {
+			wait = retryAfter
+			if r.log != nil {
```
kkyrazis: should other logs be nil checked?

internal/agent/client/transport.go
```diff
@@ -0,0 +1,157 @@
...
+				r.log.Debugf("Using Retry-After header value: %v", wait)
+			}
+		}
+
+		r.log.Debugf("Request failed with status %d, retrying in %v", resp.StatusCode, wait)
+
+		resp.Body.Close()
+
+		time.Sleep(wait)
```
kkyrazis: should the requests context be used to prevent sleeping through canceled contexts?

internal/agent/client/transport.go
hexfusion: none

PR #1652, opened 2025-09-12
EDM-2058: agent: do not drain workloads when process stops

hexfusion:
the agent currently drains Podman workloads whenever its process stops.
this PR separates that behavior from node restarts or shutdowns.

## Summary by CodeRabbit

- New Features
  - Detects pending system shutdown/reboot via system signals, enabling graceful, one-time draining of applications.
  - Integrates systemd job awareness to inform shutdown handling.

- Bug Fixes
  - Prevents crashes when updating missing dependency tasks.
  - Improves error handling: “no space left on device” is non-retryable; better mapping for “unit not found.”
  - More reliable shutdown sequencing and application drain behavior.

- Tests
  - Expanded coverage for shutdown detection, application drain lifecycle, podman monitor behavior, and TPM client closure.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -150,7 +150,7 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	if tpmClient != nil {
 		systemInfoManager.RegisterCollector(ctx, "tpmVendorInfo", tpmClient.VendorInfoCollector)
-		shutdownManager.Register("tpmClient", tpmClient.Close)
+		defer tpmClient.Close()
 	}
 
 	reloadManager := reload.NewManager(a.configFile, a.log)
@@ -187,14 +187,12 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log,
 		deviceReadWriter,
 		podmanClient,
+		systemdClient,
 		systemInfoManager,
 	)
 
 	// register the application manager with the shutdown manager
-	shutdownManager.Register("applications", applicationsManager.Stop)
-
-	// register identity provider with shutdown manager
-	shutdownManager.Register("identity", identityProvider.Close)
+	shutdownManager.Register("applications", applicationsManager.Drain)
 
 	// create systemd manager
 	systemdManager := systemd.NewManager(a.log, systemdClient)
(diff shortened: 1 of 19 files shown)


internal/agent/agent.go
```diff
@@ -338,9 +339,6 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log,
 	)
 
-	// register agent with shutdown manager
-	shutdownManager.Register("agent", agent.Stop)
```
hexfusion, line 338: note: ctx is cancelled via shutdown already so this is not needed

internal/tpm/client_test.go
```diff
@@ -691,160 +690,6 @@ func TestResolveTPMPath(t *testing.T) {
 	}
 }
 
-func TestConvertTPMLPCRSelectionToPCRSelection(t *testing.T) {
```
hexfusion, line 694: note: these tested logic that was not actually used

internal/agent/device/applications/podman_monitor.go
```diff
@@ -105,108 +90,22 @@ func (m *PodmanMonitor) startMonitor(ctx context.Context) error {
...
-
-func (m *PodmanMonitor) Stop(ctx context.Context) error {
-	if err := m.drain(ctx); err != nil {
-		return err
-	}
-	return nil
+func (m *PodmanMonitor) Drain(ctx context.Context) error {
+	m.log.Info("Draining application workloads for system shutdown")
+	return m.drain(ctx)
 }
```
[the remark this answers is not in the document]
hexfusion: this timeout is handeled at a higher level

internal/agent/device/applications/podman_monitor.go
```diff
@@ -389,9 +268,10 @@ func (m *PodmanMonitor) ExecuteActions(ctx context.Context) error {
 			return fmt.Errorf("failed to start podman monitor: %w", err)
 		}
 	} else {
-		if err := m.stopMonitor(); err != nil {
-			return fmt.Errorf("failed to stop podman monitor: %w", err)
-		}
+		// When there are no apps, the monitor will stop naturally when context is cancelled
```
kkyrazis, line 271: In the event that a user adds and removes applications across many specs or the agent restarts multiple times without restarting the VM,  `podman events` processes could be left running. We know that that is relatively high CPU intensive process.

In the add/remove apps to the running process case, I believe this would result in multiple event listeners updating an application. I know that the application map is thread safe, but I'm not sure if the applications themselves are.

hexfusion: I will double check that I thought because podman events  is a sub process that there is teardown logic but good callout.

PR #1658, opened 2025-09-14
EDM-1735: Device disconnected alertManager test

eldar101:
https://issues.redhat.com/browse/EDM-1735

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test verifying alert lifecycle during agent restarts: confirms DeviceDisconnected alerts appear when the agent stops and clear after reconnection.
  * Uses extended timeouts and polling to ensure stability across restart scenarios.
  * Validates baseline counts, alert surge during outage, and return to baseline after recovery, strengthening regression coverage for CLI-driven workflows.

### DIFF
--- a/test/e2e/cli/cli_events_test.go
+++ b/test/e2e/cli/cli_events_test.go
@@ -424,6 +424,47 @@ var _ = Describe("cli events operation", func() {
 				return out
 			}, "60s", "2s").Should(ContainSubstring("no application workloads defined"))
 		})
+
+		It("emits DeviceDisconnected then clears after agent restarts", Label("84692"), func() {
+			harness := e2e.GetWorkerHarness()
+
+			token, err := harness.GetToken()
+			Expect(err).ToNot(HaveOccurred())
+			token = strings.TrimSpace(token)
+
+			host, err := harness.CLI("get", "route", "flightctl-alertmanager-proxy-route",
+				"-n", "flightctl", "-o", "jsonpath={.spec.host}")
+			Expect(err).ToNot(HaveOccurred())
+			host = strings.TrimSpace(host)
+
+			countDisc := func() int {
+				cmd := fmt.Sprintf(`curl -ks -H "Authorization: Bearer %s" "https://%s/api/v2/alerts?active=true&silenced=false&inhibited=false&filter=alertname%%3DDeviceDisconnected" | jq 'length'`, token, host)
+				out, err := harness.CLI("bash", "-lc", cmd)
+				Expect(err).ToNot(HaveOccurred(), "Alertmanager query failed: %s", out)
+				var n int
+				_, _ = fmt.Sscanf(strings.TrimSpace(out), "%d", &n)
+				return n
+			}
+
+			initial := countDisc()
+
+			By("stopping flightctl-agent")
+			Expect(harness.StopAgent()).To(Succeed())
+			defer func() {
+				By("(defer) starting flightctl-agent")
+				_ = harness.StartAgent()
+			}()
+
+			By("waiting for DeviceDisconnected to appear while agent is stopped")
+			Eventually(countDisc, LONG_TIMEOUT, LONG_POLLING).Should(BeNumerically(">", initial))
+
+			By("starting flightctl-agent to reconnect")
+			Expect(harness.StartAgent()).To(Succeed())
+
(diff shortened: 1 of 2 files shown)


eldar101:
might need to edit this to this format:
create an app with bad yaml info/path/duplicate names/whaetever and get an alert

test/e2e/cli/cli_events_test.go
```diff
@@ -424,6 +424,62 @@ var _ = Describe("cli events operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			token = strings.TrimSpace(token)
+
+			host, err := harness.CLI("get", "route", "flightctl-alertmanager-proxy-route",
+				"-n", "flightctl", "-o", "jsonpath={.spec.host}")
+			Expect(err).ToNot(HaveOccurred())
+			host = strings.TrimSpace(host)
+
+			countDisc := func() int {
+				cmd := fmt.Sprintf(`curl -ks -H "Authorization: Bearer %s" "https://%s/api/v2/alerts?active=true&silenced=false&inhibited=false" | jq '[.[] | select(.labels.alertname=="DeviceDisconnected")] | length'`, token, host)
```
siserafin: I would create a helper function for curl

test/e2e/cli/cli_events_test.go
```diff
@@ -424,6 +424,62 @@ var _ = Describe("cli events operation", func() {
...
+				var n int
+				_, _ = fmt.Sscanf(strings.TrimSpace(out), "%d", &n)
+				return n
+			}
+
+			initial := countDisc()
+			GinkgoWriter.Printf("[DEBUG] Initial DeviceDisconnected count: %d\n", initial)
+
+			1
+			GinkgoWriter.Println("[DEBUG] Stopping flightctl-agent on VM")
```
siserafin: I think that stop and start agent can be harness functions

test/e2e/cli/cli_events_test.go
```diff
@@ -424,6 +424,62 @@ var _ = Describe("cli events operation", func() {
 				return out
 			}, "60s", "2s").Should(ContainSubstring("no application workloads defined"))
 		})
+
+		It("emits DeviceDisconnected then clears after agent restarts", Label("sanity", "84692"), func() {
+			harness := e2e.GetWorkerHarness()
+
+			token, err := harness.CLI("whoami", "-t")
```
siserafin: This could be GetToken function

siserafin: Is this compatible with standalone, ACM and kind?

test/e2e/cli/cli_events_test.go
```diff
@@ -424,6 +424,62 @@ var _ = Describe("cli events operation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			token = strings.TrimSpace(token)
+
+			host, err := harness.CLI("get", "route", "flightctl-alertmanager-proxy-route",
```
siserafin, line 435: This command works only in OCP. If is meant we should skip it in case of different context. And remove the sanity label.

test/e2e/cli/cli_events_test.go
```diff
@@ -424,6 +424,62 @@ var _ = Describe("cli events operation", func() {
...
+				"-n", "flightctl", "-o", "jsonpath={.spec.host}")
```
siserafin, line 436: We have FLIGHTCTL_NS env. The tests are also executed enabling flightctl from ACM, in this case the ns name is different

siserafin (commented):
Please check my comments

test/e2e/cli/cli_events_test.go
eldar101: https://curlconverter.com/go/

test/e2e/cli/cli_events_test.go
eldar101: yes

eldar101:
Closing as it is redundant
I'm not sure I can disconnect a device in our e2e tests by turning off the interface.
the only safe trigger for device alerts is, Other than disconnection, the most deterministic is application failure,
As in Creating an app with bad yaml info/path/duplicate names/whaetever and get an alert

Already covered by:
https://github.com/eldar101/flightctl/blob/main/test/e2e/cli/cli_events_test.go#L370

PR #1662, opened 2025-09-15
Edm 285: Prevent User API from auto-updating device.status.lastSeen

galel12:
### Problem
The `device.status.lastSeen` field was being automatically updated to `time.Now()` for **all** device status requests, including User API calls. This made it impossible to track actual agent connectivity since user requests would "hijack" the timestamp.

### Root Cause
The existing `IsInternalRequest()` function didn't differentiate between Agent and User API requests - both were treated as "not internal" and thus both received the auto-update `lastSeen` behavior.

### Solution
Implemented request-context-aware `lastSeen` handling:
- **Agent requests**: Auto-update `lastSeen` to current time (preserves existing behavior)
- **User/Internal requests**: Preserve existing `lastSeen` value, ignore any incoming `lastSeen` in payload

**Key changes:**
- Added `IsAgentRequest()` helper function to specifically identify agent contexts
- Updated service layer to conditionally set `lastSeen` based on request source
- Fixed boolean consistency for `AgentCtxKey` across all components

### Testing
- Added comprehensive TDD unit test (`TestDeviceReplaceStatusLastSeen`) covering agent, user, and internal request contexts
- Fixed existing tests to work with new context-aware logic
- Validated with real agent VM and User API attack scenarios - fix confirmed working

### Impact
- ✅ Agent behavior unchanged
- ✅ User API can no longer manipulate device communication timestamps  
- ✅ `lastSeen` now accurately reflects actual agent connectivity


celdrake:
I think this PR is not needed anymore due to the recent changes in https://github.com/flightctl/flightctl/pull/1681 that removed the field.

galel12:
> I think this PR is not needed anymore due to the recent changes in #1681 that removed the field. @celdrake 

Yeah i figured EDM-2149 might affect this one after the conversation of removing this subfield but looking at the implementation, the core issue is still there. The new code in EDM-2149 still auto-updates lastSeen for ALL non-internal requests (including user API calls), when it should only update for agent requests.
My EDM-285 fix addresses exactly this, distinguishing between agent vs user requests.
I'll just need to adapt it for the new pointer-based structure (*time.Time instead of time.Time) after rebasing onto the latest main.

From what i understand the field still exists in DeviceStatus (just hidden from JSON) and they also promoted it to a top-level database column with its own /lastseen endpoint. So the underlying update logic is still there and still has the same bug - user API requests shouldn't auto-update this timestamp.
So the logic is still needed, just needs a small tweak for the new architecture 👍

avishayt:
> > I think this PR is not needed anymore due to the recent changes in #1681 that removed the field. @celdrake
> 
> Yeah i figured EDM-2149 might affect this one after the conversation of removing this subfield but looking at the implementation, the core issue is still there. The new code in EDM-2149 still auto-updates lastSeen for ALL non-internal requests (including user API calls), when it should only update for agent requests. My EDM-285 fix addresses exactly this, distinguishing between agent vs user requests. I'll just need to adapt it for the new pointer-based structure (*time.Time instead of time.Time) after rebasing onto the latest main.
> 
> From what i understand the field still exists in DeviceStatus (just hidden from JSON) and they also promoted it to a top-level database column with its own /lastseen endpoint. So the underlying update logic is still there and still has the same bug - user API requests shouldn't auto-update this timestamp. So the logic is still needed, just needs a small tweak for the new architecture 👍

Sorry for not syncing with you on this.  I moved lastSeen to its own DB column as part of the change, so now UpdateStatus shouldn't update lastSeen anymore.  Only the "healthcheck" should be updating.  Please make sure.

hexfusion:
closing this and the bug

hexfusion:
actually I will let @galel12 do that once verified @avishayt comment  https://github.com/flightctl/flightctl/pull/1662#issuecomment-3335141038

galel12:
@avishayt Ahh I see what you did there! Thanks for the clarification. I tested it myself and confirmed it works correctly now - user API calls can no longer manipulate lastSeen timestamps. 

Closing this PR as fixed by #1681. Thanks for the great work! 🚀

galel12:
https://github.com/flightctl/flightctl/pull/1662#issuecomment-3411167403

PR #1665, opened 2025-09-15
EDM-830: Add RBAC user role and cluster role tests

itzikb-redhat:
- Checking that a user without a role can't create and list devices and fleets
- Adding a role and a role binding to be able to create and list devices but not fleets
- Update the role to be able to also create and list fleets
- Do the same for ClusterRole and ClusterRoleBinding


test/e2e/rbac/rbac_test.go
```diff
@@ -198,5 +275,97 @@ var _ = Describe("RBAC Authorization Tests", Label("rbac", "authorization"), fun
 			err = harness.DeleteClusterRoleBinding(suiteCtx, harness.Cluster, createdAdminClusterRoleBinding.Name)
 			Expect(err).ToNot(HaveOccurred(), "Admin should be able to delete cluster role binding")
 		})
+		It("should have a limited access with a non-admin role", Label("84169"), func() {
+			By("Login to the cluster as a user without a role")
+			err := login.LoginAsNonAdmin(harness, nonAdminUser, nonAdminUser, defaultK8sContext, k8sApiEndpoint)
+			Expect(err).ToNot(HaveOccurred())
```
eldar101, line 284: does logging in contain an expected subtstring?
Please in general try to add a expected output check in the blocks of this PR.

itzikb-redhat: Thanks. I think in this case an error check is enough. I've added a check to ChangeK8sContext

siserafin (approved):
lgtm

PR #1676, opened 2025-09-17
EDM-2132:Update Agent rollback logs

SiddarthR56:


internal/agent/device/device.go
```diff
@@ -210,6 +210,12 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 			return
 		}
 
+		// handle policy not ready (download/update)
+		if errors.Is(syncErr, errors.ErrUpdatePolicyNotReady) || errors.Is(syncErr, errors.ErrDownloadPolicyNotReady) {
+			a.handleSyncError(ctx, desired, syncErr)
+			return
```
hexfusion, line 216: nit: although we are not upgrading yet so to speak the desired state has still been bumped. the expectation is that we rollback before we try again. after rollback the current and desired state will be equal. 

so we still need to rollbackDevice but if we can identify expected cases, perhaps we can handle the log message differently?

internal/agent/device/device.go
```diff
@@ -210,6 +210,13 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 			return
 		}
 
+		// Policy may defer update; in all cases, warn and roll back to the previous renderedVersion.
+		if errors.Is(syncErr, errors.ErrUpdatePolicyNotReady) || errors.Is(syncErr, errors.ErrDownloadPolicyNotReady) {
+			a.log.Warnf("Policy not ready for update; Attempting to rollback to previous renderedVersion: %s", current.Version())
```
hexfusion: Looking better the bug is around the use of the word `rollback` as it seems like something that would happen in error but in this case its expected. so maybe we can do.

```suggestion
			a.log.Warnf("Requeuing version %s: %s", current.Version(), syncErr.Error())
```
would look like 
```
Requeuing version 23: update policy not ready

PR #1677, opened 2025-09-17
EDM-1959: Safegaurd flightctl config from being modified

SiddarthR56:


internal/tasks/device_render.go
```diff
@@ -279,6 +279,35 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+// The check denies the root itself and any subpath under these roots.
+func denyUnsafeDevicePath(p string) error {
+	// Only absolute Linux paths are allowed in rendered configs
+	if p == "" || !strings.HasPrefix(p, "/") {
+		return fmt.Errorf("invalid device path (must be absolute): %q", p)
+	}
+	clean := filepath.Clean(p)
+
+	deniedRoots := []string{
+		"/etc/flightctl",
```
hexfusion: nit: we want to be more perscriptive here. we have certain paths that we should be able to write to in /etc/flightctl/ like `/etc/flightctl/hooks.d/${lifecyclehook}/` I think you will want to block explicitly `/etc/flightctl/config.yaml` and instead prefer direct usage of dropin dir `/etc/flightctl/conf.d/`. 

see https://github.com/flightctl/flightctl/blob/main/docs/user/configuring-agent.md for more paths.

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+	if strings.ContainsRune(p, ':') {
+		return fmt.Errorf("invalid device path (must not contain ':'): %q", p)
+	}
+	clean := filepath.Clean(p)
+
+	// Always-denied roots
+	deniedRoots := []string{
+		"/var/lib/flightctl",
+		"/usr/lib/flightctl",
+		"/etc/flightctl/certs",
```
kkyrazis: I'm just curious what the use case is for allowing writing to `/etc/flightctl/` is if we're blocking other flightctl directories.

kkyrazis: This was a dumb question, please ignore it. `/etc/flightctl` is largely open for the user.

hexfusion (commented):
this all makes sense to me can we add some coverage to a new device_renger_test.go please we really lack coverage. thank you!

internal/tasks/device_render.go
```diff
@@ -333,6 +374,13 @@ func (t *DeviceRenderLogic) renderGitConfig(ctx context.Context, configItem *api
 		}
 	}
 
+	// Validate all file paths produced by the git config before merging
+	for _, f := range ignition.Storage.Files {
+		if err := denyUnsafeDevicePath(f.Path); err != nil {
+			return &gitSpec.Name, &gitSpec.GitRef.Repository, fmt.Errorf("invalid path from git config: %w", err)
```
hexfusion: note: I know you didnt setup this file but I don't understand why in error condition why we are not just returning nil for name and repo

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+	// Always-denied roots
+	deniedRoots := []string{
+		"/var/lib/flightctl",
+		"/usr/lib/flightctl",
+		"/etc/flightctl/certs",
+	}
+
+	for _, root := range deniedRoots {
+		if clean == root || strings.HasPrefix(clean, root+"/") {
+			return fmt.Errorf("unsafe device path %q: writing under %q is not allowed", p, root)
```
hexfusion: unsafe is a little scary sounding can we say forbidden

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+		}
+	}
+
+	etcFlightctlConfig := filepath.Clean("/etc/flightctl/config.yaml")
+	if clean == etcFlightctlConfig {
+		return fmt.Errorf("unsafe device path %q: writing %q is not allowed", p, etcFlightctlConfig)
```
hexfusion: same

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+// denyUnsafeDevicePath returns an error if the given absolute device path targets
+// agent-managed or read-only directories/files that must not be written by config providers.
+// Denied paths:
+//   - /var/lib/flightctl (and all subpaths)
+//   - /usr/lib/flightctl (and all subpaths)
+//   - /etc/flightctl/certs (and all subpaths)
+//   - /etc/flightctl/config.yaml (file itself)
+//
+// The check denies the exact file and the roots listed above and any subpath under those roots.
+func denyUnsafeDevicePath(p string) error {
```
hexfusion: nit: unsafe is scary
```suggestion
func denyForbiddenDevicePath(p string) error {
```

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+	// Only single absolute Linux paths are allowed in rendered configs
+	if p == "" || !filepath.IsAbs(p) {
+		return fmt.Errorf("invalid device path (must be absolute): %q", p)
+	}
```
hexfusion: Is this validation specific to the deny list or a general validation that should always be checked

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+	// Reject PATH-like lists which could alter semantics downstream
+	if strings.ContainsRune(p, ':') {
+		return fmt.Errorf("invalid device path (must not contain ':'): %q", p)
+	}
```
hexfusion: same here

internal/tasks/device_render.go
SiddarthR56: it's both general validation AND required for the deny list to work correctly. For Git configs this validation is not done at the API Level.

internal/tasks/device_render.go
```diff
@@ -292,6 +292,47 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *ap
...
+	for _, root := range deniedRoots {
+		if clean == root || strings.HasPrefix(clean, root+"/") {
+			return fmt.Errorf("forbibben device path %q: writing under %q is not allowed", p, root)
+		}
+	}
+
+	etcFlightctlConfig := filepath.Clean("/etc/flightctl/config.yaml")
+	if clean == etcFlightctlConfig {
+		return fmt.Errorf("forbidden device path %q: writing %q is not allowed", p, etcFlightctlConfig)
+	}
```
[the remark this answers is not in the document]
SiddarthR56: The agent's main config file is hardcoded to /etc/flightctl/config.yaml. The agent never checks for /etc/flightctl/config.yml, so blocking it would be unnecessary.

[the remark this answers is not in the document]
hexfusion: we should fix both but lets cover it here first plz

test/integration/tasks/device_render_test.go
```diff
@@ -66,14 +93,201 @@ var _ = Describe("DeviceRender", func() {
...
+					event := api.Event{
+						Reason:         api.EventReasonResourceUpdated,
+						InvolvedObject: api.ObjectReference{Kind: api.DeviceKind, Name: testDeviceName},
+					}
+					logic := tasks.NewDeviceRenderLogic(log, serviceHandler, nil, kvStoreInst, orgId, event)
+					err = logic.RenderDevice(ctx)
+
+					if shouldFailWithPathError {
+						Expect(err).To(HaveOccurred())
+						Expect(err.Error()).To(ContainSubstring("unsafe device path"))
```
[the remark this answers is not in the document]
hexfusion: ideally these are error types that we can eval with errors.Is(err, ErrTypeHere)

PR #1697, opened 2025-09-22
EDM-1915: Implement AAP based organization authorization

DakCrowder:
This PR creates an organization provider and updates our middleware to make use of AAP organizations.  A new AAP client has been added to abstract out some of the specific logic around interacting with their api, and caching added around membership checks so we're not hitting that api on every request.

The AAP api does not have a way to directly check for organization membership or fetch all orgs for a given user at once based upon membership so we have to do some of our own logic / wrapping, namely around team based membership.

Users can:
- Be assigned membership in AAP directly to orgs (orgs have many users), these mappings can be inferred by fetching data from /organizations endpoints
- Be assigned membership via teams.  Orgs have many teams which have many users.  We can fetch teams a user is assigned to, and these teams are returned with their parent orgs in the request body

There are also a couple of superuser roles (superuser and auditor) that are handled differently, these users have access to all orgs in AAP and we simply fetch all the orgs when those special case users are in context.

## Summary by CodeRabbit

* **New Features**
  * AAP integration: gateway client, user/org/team endpoints, AAP-based org provider, and AAP identity with role flags.
  * Membership TTL cache to speed membership checks.

* **Improvements**
  * Authorization switched to organization-membership-based checks.
  * Resolver construction now returns errors and accepts an optional membership cache; callers handle initialization failures.
  * TLS/pagination support for AAP gateway client.

* **Tests**
  * Expanded unit/integration tests for AAP client, provider, cache, and membership flows.

### DIFF
--- a/cmd/flightctl-alertmanager-proxy/main.go
+++ b/cmd/flightctl-alertmanager-proxy/main.go
@@ -288,7 +288,10 @@ func main() {
 		Log:    logger,
 		Cache:  orgCache,
 	}
-	orgResolver := resolvers.BuildResolver(buildResolverOpts)
+	orgResolver, err := resolvers.BuildResolver(buildResolverOpts)
+	if err != nil {
+		logger.Fatalf("Failed to build organization resolver: %v", err)
+	}
 
 	// Initialize auth system
 	authN, authZ, err := auth.InitAuth(cfg, logger, orgResolver)
--- a/internal/auth/authn/aapgateway.go
+++ b/internal/auth/authn/aapgateway.go
@@ -19,7 +19,7 @@ type AAPGatewayUserIdentity interface {
 	IsPlatformAuditor() bool
 }
 
-// AAPGatewayIdentity extends common.Identity with AAP-specific fields
+// AAPIdentity extends common.Identity with AAP-specific fields
 type AAPIdentity struct {
 	common.BaseIdentity
 	superUser       bool


internal/org/cache/membership.go
```diff
@@ -0,0 +1,47 @@
+package cache
+
+import (
+	"time"
+
+	"github.com/jellydator/ttlcache/v3"
+)
+
+type MembershipCache interface {
+	Get(key string) (bool, bool)
```
hexfusion, line 10: note: double bool feels a little awkward, is it nessisary?

pkg/aap_client/client.go
```diff
@@ -0,0 +1,121 @@
+package aap_client
```
hexfusion: in the future aap/client but for now consider making the namespace reusable.
```suggestion
package aap
```

internal/org/cache/membership.go
```diff
@@ -0,0 +1,47 @@
...
+
+type MembershipTTL struct {
+	cache *ttlcache.Cache[string, bool]
+}
+
+func NewMembershipTTL(ttl time.Duration) *MembershipTTL {
+	opts := []ttlcache.Option[string, bool]{}
+	if ttl > 0 {
+		opts = append(opts, ttlcache.WithTTL[string, bool](ttl))
+	} else {
```
hexfusion: if the ttl is negative, I would expect an error what is the usecase for this?

internal/org/providers/aap_provider.go
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/flightctl/flightctl/internal/consts"
+	"github.com/flightctl/flightctl/internal/org"
+	"github.com/flightctl/flightctl/internal/org/cache"
+	"github.com/flightctl/flightctl/pkg/aap_client"
+)
+
+type AAPClientInterface interface {
+	GetOrganization(token string, organizationID string) (*aap_client.AAPOrganization, error)
+	GetUserTeams(token string, userID string) ([]*aap_client.AAPTeam, error)
+	GetOrganizations(token string) ([]*aap_client.AAPOrganization, error)
```
hexfusion: is this `ListOrganizations`?

internal/org/providers/aap_provider.go
```diff
@@ -0,0 +1,220 @@
...
+	GetUserOrganizations(token string, userID string) ([]*aap_client.AAPOrganization, error)
```
hexfusion: could this be `ListOrganizations(token, WithUserID(userId))`

internal/org/cache/membership.go
```diff
@@ -0,0 +1,47 @@
+package cache
+
+import (
+	"time"
+
+	"github.com/jellydator/ttlcache/v3"
+)
+
+type MembershipCache interface {
```
hexfusion: we already know we are a `cache` so could this be `cache.Membership`?

internal/org/providers/aap_provider.go
```diff
@@ -0,0 +1,220 @@
...
+		GatewayUrl:      apiUrl,
+		TLSClientConfig: tlsConfig,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if cache == nil {
+		return nil, fmt.Errorf("AAP organization provider requires a membership cache")
+	}
```
hexfusion, line 45: do we really need to check this? its good to be defensive, but in practice is it nessisary? checkout your code and you can decide.

internal/org/providers/aap_provider.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	aapTeams, err := p.client.GetUserTeams(token, userID)
+	if err != nil {
+		return nil, err
+	}
+
+	aapOrganizationsMap := make(map[int]*aap_client.AAPOrganization)
```
hexfusion: preallocate?
```suggestion
	aapOrganizationsMap := make(map[int]*aap_client.AAPOrganization, len(aapOrganizations))
```

internal/org/providers/aap_provider.go
```diff
@@ -0,0 +1,220 @@
...
+	if !ok {
+		return false, fmt.Errorf("cannot get organizations claims from a non-token identity (got %T)", identity)
+	}
+
+	userID := aapIdentity.GetUID()
+	if userID == "" {
+		return false, fmt.Errorf("user ID is required")
+	}
+
+	if present, isMember := p.cache.Get(cacheKey(userID, externalOrgID)); present {
```
hexfusion, line 147: question: should cache reason about membership maybe decoupling would simplify?

pkg/aap_client/client.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, errors.New("aap_client: GatewayUrl is required")
+	}
+	if options.TLSClientConfig == nil {
+		return nil, errors.New("aap_client: TLSClientConfig is required")
+	}
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: options.TLSClientConfig,
+		},
+		Timeout: 10 * time.Second,
```
hexfusion: const?

internal/org/cache/membership.go
DakCrowder: After thinking about it more just modified to accept the passed ttl

internal/org/cache/membership.go
DakCrowder: It is an attempt to resolve cache results for negative lookup results - we actually care about 3 states from the cache, true / false/ doesn't exist in the cache and this is an attempt to expose that so we aren't making extra api calls when a user does NOT have membership to some org.  Should be rare to get requests like this (most likely because a users membership was revoked) but tried to handle.

internal/org/providers/aap_provider.go
DakCrowder: Ahh right we don't know the total number of orgs yet but it is at least len(aapOrganizations), updated

PR #1700, opened 2025-09-23
EDM-1804: Add ability to generate Application Identities

kkyrazis:
Adds to ability to create Application Keys that are structured as defined in the following diagram. Due to the expected use of these requests, I've created the ability for multiple sessions to be created so that they can be used independently. The underlying shared storage was made thread safe. This will allow the device's primary identity to continue establishing tls sessions while the cert manager manages application keys. The entire operation is self contained so that all information that is relevant is generated, the TCG CSR and the TSS2 file, and then the session is closed.

image

## Summary by CodeRabbit

- **New Features**
  - Create, load, remove per-application TPM keys; generate CSRs signed by per-key signers; exportable device identities; new client methods to manage application keys.

- **Refactor**
  - Lazy, option-driven TPM connection/session lifecycle with connection-factory support; session APIs extended for app-key flows; storage key template selection standardized; CSR generation extracted for reuse.

- **Chores**
  - File storage is concurrency-safe and persists per-application key data.

- **Tests**
  - Simulator-backed tests updated to use connection factories and exercise application-key lifecycle; mocks expanded for new app-key and device-ID APIs.

### DIFF
--- a/internal/tpm/client.go
+++ b/internal/tpm/client.go
@@ -7,9 +7,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/rsa"
-	"crypto/x509"
-	"crypto/x509/pkix"
-	"encoding/pem"
 	"fmt"
 	"io"
 	"math/big"
@@ -18,6 +15,7 @@ import (
 
 	agent_config "github.com/flightctl/flightctl/internal/agent/config"
 	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	fccrypto "github.com/flightctl/flightctl/pkg/crypto"
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/google/go-tpm/tpm2"
 	"github.com/google/go-tpm/tpmutil"
@@ -29,10 +27,14 @@ var _ crypto.Signer = (*client)(nil)
 // Ensure client implements Client interface
 var _ Client = (*client)(nil)
 
+type connFactory func() (io.ReadWriteCloser, error)
+
 // Client represents a simplified TPM client that exposes signing capabilities
 // and attestation data for CSR generation.
 type client struct {
 	session       Session
+	connFactory   connFactory
+	sessionOpts   []SessionOption
 	log           *log.PrefixLogger
 	productModel  string
 	productSerial string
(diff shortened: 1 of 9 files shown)


internal/tpm/client.go
```diff
@@ -179,6 +200,100 @@ func (c *client) MakeCSR(deviceName string, qualifyingData []byte) ([]byte, erro
 	return tcgCSR, nil
 }
 
+// RemoveApplicationKey removes the associated key from storage
+func (c *client) RemoveApplicationKey(appName string) error {
+	conn, err := c.connFactory()
```
hexfusion, line 203: question: what are the tradeoffs of having short lives conns.

internal/tpm/device_id.go
```diff
@@ -0,0 +1,51 @@
...
+	pub          tpm2.TPM2BPublic
+	priv         tpm2.TPM2BPrivate
+	loadedHandle tpm2.AuthHandle
+	conn         io.ReadWriteCloser
+}
+
+func (e *exportableDeviceID) Public() crypto.PublicKey {
+	pub, err := convertTPM2BPublicToPublicKey(&e.pub)
+	if err != nil {
+		return nil
```
hexfusion, line 26: nit: can we log the error ?

internal/tpm/session.go
```diff
@@ -21,37 +20,72 @@ const (
 	persistentHandleMax = tpm2.TPMHandle(0x81FFFFFF)
 	permanentHandleMin  = tpm2.TPMHandle(0x40000000)
 	permanentHandleMax  = tpm2.TPMHandle(0x4004FFFF)
-	nvReadChunkSize     = uint16(1024) // Maximum chunk size for NVRead operations
-	ekRSANVIndex        = gotpmclient.EKCertNVIndexRSA
-	ekECCNVIndex        = gotpmclient.EKCertNVIndexECC
+	// Table 11 of https://trustedcomputinggroup.org/wp-content/uploads/RegistryOfReservedTPM2HandlesAndLocalities_v1p1_pub.pdf
```
hexfusion, line 23: :+1:

internal/tpm/session.go
```diff
@@ -121,6 +155,45 @@ func (s *tpmSession) CreateKey(keyType KeyType) (*tpm2.CreateResponse, error) {
...
+	}
+
+	return &exportableDeviceID{
+		parentHandle: key.ParentHandle,
+		pub:          key.Public,
+		priv:         key.Private,
+		loadedHandle: tpm2.AuthHandle{
+			Handle: loadRsp.ObjectHandle,
+			Name:   loadRsp.Name,
+			Auth:   tpm2.PasswordAuth(nil), // todo passwords
```
hexfusion, line 192: :sob:

internal/tpm/client.go
kkyrazis: The major benefits are:
* we can an isolate these application keys from the main session we have. 
* There's less contention for transient space as the RM will deal with it
* don't have to worry about concurrent access of the conn (the cert manager runs on a separate goroutine)

Cons:
* more conns

hexfusion: ok I assume the performance impact is reasonable.

internal/tpm/session.go
```diff
@@ -1030,3 +1038,206 @@ func (s *tpmSession) SolveChallenge(credentialBlob, encryptedSecret []byte) ([]b
...
+	// Load the storage key so that it can be immediately evicted
+	loadCmd := tpm2.Load{
+		ParentHandle: s.handles[SRK],
+		InPrivate:    createResp.OutPrivate,
+		InPublic:     createResp.OutPublic,
+	}
+	loadRsp, err := loadCmd.Execute(transport.FromReadWriter(s.conn))
+	if err != nil {
+		return fmt.Errorf("loading app storage key: %w", err)
+	}
```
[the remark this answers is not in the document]
kkyrazis: It expects an interface of which the pointer satisfies.

PR #1702, opened 2025-09-23
EDM-1804: Cert Manager with TPM identity support

kkyrazis:
Adds the ability for CSRs created by Agents to use the TPM as an identity provider. 

The type of identity: tpm or software, is configurable in the certs.yaml config file. TPM identities can only be used if the tpm was used as the primary identity.

## Summary by CodeRabbit

* **New Features**
  * Added TPM-based device identity and CSR flows, TPM verification, and TPMVerified condition.
  * Agent can provision identities via TPM or software using a new factory; deviceVerificationIdentifier exposed.
  * Auto-approval tightened to require matching signer and no failed TPM verification.

* **Tests**
  * Added unit tests for TPM and identity providers (software/TPM) and exportable behaviors.

* **Documentation**
  * OpenAPI/Swagger updated to include TPMVerified enum/varnames and deviceVerificationIdentifier.


internal/agent/device/certmanager/provider/provisioner/csr.go
```diff
@@ -49,20 +49,21 @@ type CSRProvisioner struct {
...
-	cfg *CSRProvisionerConfig
+	cfg              *CSRProvisionerConfig
+	identityProvider identity.ExternalProvider
 
 	// Name of the CSR resource on the server
-	csrName string
-	// Generated private key for this certificate
-	privateKey crypto.Signer
+	csrName  string
+	identity *identity.ExportableIdentity
```
hexfusion: optional: total nit pick but you can see the opp `identity.Exportable`

internal/service/certificatesigningrequest.go
```diff
@@ -269,7 +347,12 @@ func (h *ServiceHandler) UpdateCertificateSigningRequestApproval(ctx context.Con
 	if newCSR.Status == nil {
 		return nil, api.StatusBadRequest("status is required")
 	}
-	allowedConditionTypes := []api.ConditionType{api.ConditionTypeCertificateSigningRequestApproved, api.ConditionTypeCertificateSigningRequestDenied, api.ConditionTypeCertificateSigningRequestFailed}
+	allowedConditionTypes := []api.ConditionType{
+		api.ConditionTypeCertificateSigningRequestApproved,
+		api.ConditionTypeCertificateSigningRequestDenied,
+		api.ConditionTypeCertificateSigningRequestFailed,
+		api.ConditionTypeCertificateSigningRequestTPMVerified,
```
hexfusion, line 362: question: isn't verification a seperate operation from approval? an ER could be TPMVerified=True AND Approved=False.

kkyrazis: Yeah, this is part of the `UpdateCertificateSigningRequestApproval` function. Without this, it would fail to approve because of the unknown 'TPMVerified' condition.

api/v1alpha1/openapi.yaml
```diff
@@ -4671,6 +4671,9 @@ components:
         info:
           type: string
           description: Human readable information about the last integrity transition.
+        deviceVerificationIdentifier:
```
hexfusion: question: is the value used outside of the csr request, can its value be trusted?
cc @lkatalin

internal/agent/device/certmanager/provider/provisioner/csr.go
```diff
@@ -50,19 +50,32 @@ type CSRProvisioner struct {
...
+	if csrClient == nil {
+		return nil, fmt.Errorf("csrClient is nil")
+	}
+	if identityProvider == nil {
+		return nil, fmt.Errorf("identityProvider is nil")
+	}
+	if cfg == nil {
+		return nil, fmt.Errorf("cfg is nil")
+	}
+
```
hexfusion: question: can these be nil in practice? I get being defensive. idiomatic go means we can make some sane assumptions. i dont want to say don't be defensive but also be practical.

internal/agent/device/certmanager/provider/provisioner/csr.go
```diff
@@ -157,9 +173,9 @@ func (p *CSRProvisioner) check(ctx context.Context) (bool, *x509.Certificate, []
 			return false, nil, nil, fmt.Errorf("failed to parse CSR PEM certificate: %w", err)
 		}
 
-		keyPEM, err := fccrypto.PEMEncodeKey(p.privateKey)
-		if err != nil {
-			return false, nil, nil, fmt.Errorf("failed to encode private key: %w", err)
+		keyPEM := p.identity.KeyPEM
```
hexfusion: wonder if its worth making this a method and returning an error if empty

api/v1alpha1/openapi.yaml
lkatalin: I'm not sure why we need a new field for this - can someone point me to context?

kkyrazis: To wrap up this conversation: We synced offline and decided that this **does** need to live somewhere, but this isn't the place. We'll need to spend some more time evaluating this before coming up with the solution. For now, we'll use the ER as the source of truth.

hexfusion (approved):

PR #1704, opened 2025-09-24
EDM-1977: Org update transport layer

SiddarthR56:
## Summary by CodeRabbit

- **New Features**
  - APIs now accept an explicit organization identifier to scope operations by organization.

- **Improvements**
  - Organization scoping applied broadly: events, devices, fleets, rollouts, resource syncs, repositories, auth providers, console sessions, and background tasks.
  - Event creation/listing and internal task emissions consistently include org context.

- **Tests**
  - Test suites and mocks updated to the org-scoped interfaces.

- **Breaking Changes**
  - Many public method signatures changed; callers must now provide an organization ID.


internal/service/service.go
```diff
@@ -7,113 +7,114 @@ import (
...
-	DeleteTemplateVersion(ctx context.Context, fleet string, name string) api.Status
-	GetLatestTemplateVersion(ctx context.Context, fleet string) (*api.TemplateVersion, api.Status)
+	CreateTemplateVersion(ctx context.Context, orgId uuid.UUID, tv api.TemplateVersion, immediateRollout bool) (*api.TemplateVersion, api.Status)
+	ListTemplateVersions(ctx context.Context, orgId uuid.UUID, fleet string, params api.ListTemplateVersionsParams) (*api.TemplateVersionList, api.Status)
+	GetTemplateVersion(ctx context.Context, orgId uuid.UUID, fleet string, name string) (*api.TemplateVersion, api.Status)
+	DeleteTemplateVersion(ctx context.Context, orgId uuid.UUID, fleet string, name string) api.Status
+	GetLatestTemplateVersion(ctx context.Context, orgId uuid.UUID, fleet string) (*api.TemplateVersion, api.Status)
 
 	// Event
 	CreateEvent(ctx context.Context, event *api.Event)
```
DakCrowder: Should CreateEvent also take in an OrgID?

fzdarsky:
The orgId is already passed through in the context, right? What is the advantage of extracting it early and then change all method signatures to pass it as duplicate parameter?

SiddarthR56:
> The orgId is already passed through in the context, right? What is the advantage of extracting it early and then change all method signatures to pass it as duplicate parameter?

From my understanding, the reasoning behind the task is that while orgId is already in context, passing it explicitly makes the dependency clearer, simplifies testing and reuse for non-HTTP callers by making org flow visible in signatures

internal/console/console.go
```diff
@@ -157,7 +148,10 @@ func (m *ConsoleSessionManager) StartSession(ctx context.Context, deviceName, se
 	}
 	m.log.Infof("Start session. Metadata %s", sessionMetadata)
 
-	orgId := getOrgIdFromContext(ctx)
+	orgId, ok := util.GetOrgIdFromContext(ctx)
```
DakCrowder: Rather than pulling from ctx here, could we also move this up to the internal/transport/console.go layer?

internal/service/device.go
```diff
@@ -96,21 +96,20 @@ func (h *ServiceHandler) PrepareDevicesAfterRestore(ctx context.Context) error {
 	if h.eventHandler != nil {
 		event := common.GetSystemRestoredEvent(ctx, devicesUpdated)
 		if event != nil {
-			h.eventHandler.CreateEvent(ctx, event)
+			// No org available in this context; emit under NullOrgId explicitly
```
DakCrowder, line 99: :+1:

internal/tasks/consumer_test.go
```diff
@@ -486,7 +486,7 @@ func TestDispatchTasks_WithNilMetrics_SuccessfulProcessing(t *testing.T) {
 	mockConsumer := &MockConsumer{}
 
 	// Create a simple valid payload that won't trigger any task processing
-	orgId := uuid.New()
+	orgId := store.NullOrgId
```
DakCrowder: Is this replacement necessary?  I would think the code shouldn't care if the uuid is some random value

internal/tasks/fleet_rollout_test.go
```diff
@@ -212,7 +213,7 @@ func TestFleetRolloutsLogic_FullDelayDeviceRenderPropagation(t *testing.T) {
 			ctrl := gomock.NewController(t)
 			defer ctrl.Finish()
 
-			orgId := uuid.New()
+			orgId := store.NullOrgId
```
DakCrowder: Similar question here as with the other test - in general I think tests are probably a bit better not being so tightly tied to store.NullOrgId.  I know MANY are right now, but in general the further we an move away from that over time I think the better.

internal/tasks/queue_maintenance.go
```diff
@@ -164,8 +157,8 @@ func (t *QueueMaintenanceTask) retryFailedMessages(ctx context.Context, log logr
 				event.Details = &details
 			}
 
-			// Emit the event
-			t.serviceHandler.CreateEvent(orgCtx, event)
+			// Emit the event under the correct organization without mutating caller ctx
```
DakCrowder: Don't know if we need this comment expansion "without mutating caller ctx", might raise more questions in the future.

internal/periodic_checker/task_definitions.go
```diff
@@ -64,22 +64,20 @@ type RepositoryTesterExecutor struct {
...
+func createTaskContext(ctx context.Context, taskType PeriodicTaskType) context.Context {
 	taskName := string(taskType)
 	reqid.OverridePrefix(taskName)
 	requestID := reqid.NextRequestID()
 	ctx = context.WithValue(ctx, middleware.RequestIDKey, requestID)
 
-	ctx = util.WithOrganizationID(ctx, orgID)
-
 	return context.WithValue(ctx, consts.EventActorCtxKey, taskName)
 }
```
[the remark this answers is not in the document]
SiddarthR56: eventCleanup is a systemwide operation

internal/tasks/repotester.go
```diff
@@ -76,22 +77,22 @@ func (r *RepoTester) TestRepositories(ctx context.Context) {
...
 
-func (r *RepoTester) testRepository(ctx context.Context, repository api.Repository) {
+func (r *RepoTester) testRepository(ctx context.Context, orgId uuid.UUID, repository api.Repository) {
 	repoName := *repository.Metadata.Name
 	accessErr := r.TypeSpecificRepoTester.TestAccess(&repository)
-	if err := r.SetAccessCondition(ctx, &repository, accessErr); err != nil {
+	if err := r.SetAccessCondition(ctx, orgId, &repository, accessErr); err != nil {
 		r.log.Errorf("Failed to update repository status for %s: %v", repoName, err)
 	}
 }
```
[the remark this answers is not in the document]
SiddarthR56: The OpenAPI schema defines RepoSpecType as an enum with only two valid values: git and http

DakCrowder (approved):

PR #1714, opened 2025-09-28
NO-ISSUE: LastSeen CLI use "never" not "none"

avishayt:
Show <never> instead of <none> for device last seen when no timestamp exists.


itzikb-redhat (approved):
LGTM

PR #1724, opened 2025-09-29
EDM-1960: Error Message for Client-Server Version Incompatibility

noga-magen:


internal/cli/certificate.go
```diff
@@ -147,6 +148,12 @@ func (o *CertificateOptions) Validate(args []string) error {
 }
 
 func (o *CertificateOptions) Run(ctx context.Context, args []string) error {
+	// Build client first to surface incompatibility early
+	c, err := o.BuildVersionAwareClient()
```
eldar101, line 151: is c unused?

internal/client/config.go
```diff
@@ -274,6 +274,25 @@ func NewFromConfigFile(filename string, opts ...client.ClientOption) (*client.Cl
...
+func NewVersionAwareClientFromConfigFile(filename string, opts ...client.ClientOption) (*VersionAwareClient, error) {
+	config, err := ParseConfigFile(filename)
+	if err != nil {
+		return nil, err
+	}
+	return NewVersionAwareClientFromConfig(config, filename, opts...)
+}
+
+// NewVersionAwareClientFromConfig returns a new version-aware Flight Control API client from the given config.
+func NewVersionAwareClientFromConfig(config *Config, configFilePath string, opts ...client.ClientOption) (*VersionAwareClient, error) {
```
eldar101, line 288: these funs seem similar.
Any way to combine them into a more generic func?

pkg/version/compatibility_test.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+		{
+			name:          "invalid client version - should skip check",
+			clientVersion: "invalid",
+			serverVersion: "0.9.1",
+			expectError:   false,
+		},
+		{
+			name:          "invalid server version - should skip check",
+			clientVersion: "0.9.1",
```
eldar101, line 84: Strings that are repeated multiple times in one file should be put in a var.

internal/cli/certificate.go
noga-magen: It is used (:

fzdarsky:
We need to think a bit about how to test compatibility:
We currently still have minor breakages between minor versions, so minor version +/-1 does not guarantee compatibility.
Once we're at major version 1, we should actually have mechanisms in place to gracefully handle API changes, i.e. minor versions should be compatible.

PR #1725, opened 2025-09-29
EDM-1246: Add FIPS validation to CI

fzdarsky:


.github/workflows/publish-containers.yaml
```diff
@@ -179,6 +179,13 @@ jobs:
           containerfiles: Containerfile.${{ matrix.image }}
           context: .
 
+      - name: Validate FIPS
+        id: fips
+        run: |
+          go install github.com/flightctl/fips-validator@latest
```
hexfusion, line 185: question: should we version this at all? might not be an issue.

PR #1726, opened 2025-09-29
EDM-2221: agent/identity: implement parent/child auth

hexfusion:


internal/tpm/storage.go
```diff
@@ -355,79 +343,6 @@ func (s *fileStorage) ClearApplicationKey(appName string) error {
 	return s.writeData(data)
 }
 
-func (s *fileStorage) GetPassword() ([]byte, error) {
```
hexfusion, line 358: note: password management is not handled in identity via systemd

PR #1743, opened 2025-10-02
NO-ISSUE: deploy/podman: improve dependency management for api and ui units

hexfusion:


siserafin:
Tested in RHEL9.6. All the containers and services are running. LGTM.

PR #1744, opened 2025-10-02
EDM-2240: Flags shown in help for 'get' commands are contextual

celdrake:
I gave this a try with the help of Cursor. 

For "get" commands, in the help output, we only show the flags specific to the command the user is checking.


internal/cli/get.go
```diff
@@ -72,22 +94,109 @@ func NewCmdGet() *cobra.Command {
...
+			// Parse the arguments after "get" and get just the resource arguments
+			remainingArgs := args[i+1:]
+			var resourceArgs []string
+			for _, a := range remainingArgs {
+				if !strings.HasPrefix(a, "-") {
+					resourceArgs = append(resourceArgs, a)
+				}
+			}
+			return resourceArgs
+		}
```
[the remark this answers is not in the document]
celdrake: Can you re-check? I made some additional changes too.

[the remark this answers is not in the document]
celdrake: - The parsing logic was updated and now it stops at the first flag it detects, therefore the "environment=prod" token is not identified as resource arguments. With your example, the help flags shown are those of a list of devices.

- The command "flightctl get devices $name1 $name2 -o yaml", shows the output is a DeviceList with only those two items, but it's still a "list" operation.
As another example,
"flightctl get dev $dev1 $dev2 --rendered"
Error: '--rendered' can only be used when getting a single device

internal/cli/get.go
```diff
@@ -72,28 +94,121 @@ func NewCmdGet() *cobra.Command {
...
+	fs.StringVar(&o.FieldSelector, FlagFieldSelector, o.FieldSelector, "Selector (field query) to filter on, supporting operators like '=', '!=', 'in', 'contains', '>', '<', etc. (e.g., --field-selector='metadata.name in (device1,device2)', --field-selector='metadata.owner=Fleet/test').")
+	fs.StringVarP(&o.Output, FlagOutput, "o", o.Output, fmt.Sprintf("Output format. One of: (%s).", strings.Join(legalOutputTypes, ", ")))
+	fs.Int32Var(&o.Limit, FlagLimit, o.Limit, "The maximum number of results returned in the list response. If the value is 0, then the result is not limited.")
+	fs.StringVar(&o.Continue, FlagContinue, o.Continue, "Query more results starting from the value of the 'continue' field in the previous response.")
+	fs.StringVar(&o.FleetName, FlagFleetName, o.FleetName, "Fleet name for accessing templateversions (use only when getting templateversions).")
+	fs.BoolVar(&o.Rendered, FlagRendered, false, "Return the rendered device configuration that is presented to the device. Default output format is YAML.")
```
celdrake, line 116: Note I modified the description for `--rendered` to indicate that the default output format is YAML.

internal/cli/get.go
```diff
@@ -72,28 +94,121 @@ func NewCmdGet() *cobra.Command {
...
+			}
+			return resourceArgs
+		}
+	}
+	return nil
+}
+
+func (o *GetOptions) shouldShowHelpFlag(rule FlagContextualRule, resourceKind, operation string) bool {
+	resourceMatches := false
+	for _, rt := range rule.ResourceKind {
```
kkyrazis: very minor, but this could just be a `slices.Contains(rule.ResourceKind, resourceKind)`

kkyrazis (approved):

internal/cli/get.go
celdrake: Done, thanks for the review!

PR #1746, opened 2025-10-02
EDM-2211: docs: clarify bootc image building dep

hexfusion:
Error after failed BIB build

```
Error: OCI runtime error: crun: systemd failed to install eBPF device filter on cgroup.
```

This was tracked by upstream issue https://github.com/containers/container-selinux/issues/389

## Summary by CodeRabbit

* **Documentation**
  * Added container-selinux (>= 2.241) to the prerequisites for building images.
  * Updated developer and user building guides to list the new dependency required by the image builder.
  * Clarified prerequisite versions to reduce setup issues and improve build reliability across environments.

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -7,7 +7,7 @@ Flight Control is a service for declarative, GitOps-driven management of edge de
 ## Building
 
 Prerequisites:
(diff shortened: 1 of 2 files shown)


siserafin (commented):
LGTM. I tested it in RHEL, and I can build the bootc image using that container-selinux version. 
My worry is that from the original error message is not clear that the solution is to update this dependency. The user is supposed to check all the dependencies in the doc.

hexfusion:
Makes sense if it was something related to rpms we could pin a min version but BIB is really an external tool. I think if someone hits the error msg a search of the repo should find this.

PR #1753, opened 2025-10-05
EDM-2255: added flightctl-restore to the rpm

asafbennatan:


packaging/rpm/flightctl.spec
```diff
@@ -578,6 +579,7 @@ fi
 
 %files cli -f licenses.list
     %{_bindir}/flightctl
+    %{_bindir}/flightctl-restore
```
sk-ilya, line 582: The `flightctl-restore` connects directly to the DB and KV, which is a very different use case from the `cli` package, which contains the `flightctl-cli` that communicates with external APIs.
Maybe we need a separate package purely for admin tools instead of shipping these together? :thinking:

asafbennatan: i assume this will require some work from the release team and others ? 

we need this for the 0.10 release , if this separation is easily accomplished then fine , i am not sure exactly what would it take from them

hexfusion: Lets track improving this in Jira. We have a lot of binaries we are shipping, and it needs to be reviewed.

hexfusion (approved):

hexfusion:
@asafbennatan please rebase

hexfusion:
replaced by #1767

packaging/rpm/flightctl.spec
kkyrazis: Does it make sense to include this as part of the the https://github.com/flightctl/flightctl/blob/main/hack/build_multiarch_clis.sh? For our releases, this is ultimately what we use for the CLI. If we don't want it as a part of that we'd need to create a separate package which should probably be done as Ilya suggested.

hexfusion: Yeah I think we will need to address this properly before we backport

PR #1754, opened 2025-10-05
EDM-2254: fixed quadlets docs , added flightctl-restore version option

asafbennatan:
## Summary by CodeRabbit

- **New Features**
  - Added a dedicated restore CLI workflow with a restore-specific version command.

- **Documentation**
  - Revised restore guide: added explicit version compatibility checks, switched to per-service stop/status/readiness checks, reorganized and renumbered steps, added container-based port exposure and cleanup guidance, and expanded post-restore verifications.

- **Chores**
  - Updated global spelling dictionary: added "localhost" and removed an obsolete docs entry.

### DIFF
--- a/docs/user/restore.md
+++ b/docs/user/restore.md
@@ -37,7 +37,25 @@ This document covers restore procedures for both Kubernetes and Quadlets deploym
 
 ## Kubernetes Deployment Restore Process
 
-### Step 1: Stop Flight Control Services
+### Step 1: Verify Version Compatibility
+
+Verify that the `flightctl-restore` version matches the Flight Control server version to ensure compatibility:
+
+```bash
+# Get the Flight Control server version
+flightctl version
+
+# Get the flightctl-restore version
+flightctl-restore version
+
+# Compare the versions - they should match
+echo "Server version: $(flightctl version)"
+echo "Restore version: $(flightctl-restore version)"
+```
+
+⚠️ **Important**: The versions must match to ensure compatibility. If they don't match, update your `flightctl-restore` binary to match the server version before proceeding.
+
+### Step 2: Stop Flight Control Services
 
 Scale down all Flight Control service deployments to prevent data conflicts during the restore process:
 
@@ -69,7 +87,7 @@ kubectl get pods -n flightctl-external
 kubectl get pods -n flightctl-internal
 ```
 
-### Step 2: Restore Database from Backup
+### Step 3: Restore Database from Backup
 
 Now that all Flight Control services are stopped, restore your database from backup using your preferred restoration method.
 
@@ -101,7 +119,7 @@ KV_PASSWORD=$(kubectl get secret flightctl-kv-secret -n flightctl-internal -o js
 echo "KV Password retrieved successfully"
 ```
 
-### Step 4: Set Up Port Forwarding
+### Step 5: Set Up Port Forwarding
 
 Open separate terminal sessions for each port forward, or run them in the background:
 
@@ -127,7 +145,7 @@ KV_PORT_FORWARD_PID=$!
 REDISCLI_AUTH="$KV_PASSWORD" redis-cli -h localhost -p 6379 ping
 ```
 
-### Step 5: Run the Restore Command
+### Step 6: Run the Restore Command
 
 Execute the flightctl-restore command using environment variables for database and KV store passwords:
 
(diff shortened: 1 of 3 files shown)


docs/user/restore.md
```diff
@@ -262,7 +263,23 @@ sudo podman exec flightctl-kv redis-cli ping
 sudo podman exec -it flightctl-kv redis-cli
 ```
 
-### Step 5: Run the Restore Command
+### Step 5: Verify Port Access
+
+The database and KV store containers are already configured with port forwarding to the host system:
```
sk-ilya: Are they? IIRC the ports are not exposed if installed via RPM/DNF. 

There is no `PublishPort` in [flightctl-db.container](https://github.com/flightctl/flightctl/blob/main/deploy/podman/flightctl-db/flightctl-db.container)

docs/user/restore.md
```diff
@@ -262,7 +263,23 @@ sudo podman exec flightctl-kv redis-cli ping
...
+# Verify ports are accessible on localhost
+netstat -tlnp | grep :5432
+netstat -tlnp | grep :6379
+
+# Test database connectivity
+pg_isready -h localhost -p 5432
+
+# Test KV store connectivity
+redis-cli -h localhost -p 6379 ping
+```
```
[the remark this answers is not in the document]
sk-ilya: ^

hexfusion (commented):
Should we verify `flightctl-restore version` before restoring? I assume over time the logic shipped in this binary could change?

docs/user/restore.md
asafbennatan: fixed

asafbennatan:
> Should we verify `flightctl-restore version` before restoring? I assume over time the logic shipped in this binary could change?

done - added flightctl-restore version and mentioned that it should be compared with flightctl version (server)

docs/user/restore.md
```diff
@@ -262,7 +299,41 @@ sudo podman exec flightctl-kv redis-cli ping
...
+
+# Get the container file locations dynamically
+DB_CONTAINER_FILE=$(systemctl show flightctl-db.service -p SourcePath --value)
+KV_CONTAINER_FILE=$(systemctl show flightctl-kv.service -p SourcePath --value)
+
+echo "Database container file: $DB_CONTAINER_FILE"
+echo "KV store container file: $KV_CONTAINER_FILE"
+
+# Add port publishing to the container configurations
+sudo sed -i '/^\[Container\]/a PublishPort=5432:5432' "$DB_CONTAINER_FILE"
```
sk-ilya: Nit:
Perhaps a cleaner and safer way would be to create a drop-in like /`usr/share/containers/systemd/flightctl-db.container.d/10-publish-port.conf` with:
```
[Container]
PublishPort=5432:5432
```

…and a similar one for the kv container. Then simply delete these files in “Step 8: Clean Up Port Forwarding.”

Also, you can just run `daemon-reload` && `restart `(no need to `stop`).

docs/user/restore.md
```diff
@@ -262,7 +299,41 @@ sudo podman exec flightctl-kv redis-cli ping
...
+
+# Verify ports are accessible
+netstat -tlnp | grep :5432
+netstat -tlnp | grep :6379
+
+# Test database connectivity
+pg_isready -h localhost -p 5432
+
+# Test KV store connectivity
+redis-cli -h localhost -p 6379 ping
```
sk-ilya, line 348: It won't work without the instance password. Also `localhost:6379` is already the default. 

Consider:
```suggestion
REDISCLI_AUTH="$KV_PASSWORD" redis-cli ping
```

sk-ilya (commented):
A couple of comments and lgtm

docs/user/restore.md
asafbennatan: we dont need that we just need to verify the port is accepting traffic , yes it will not authenticate which is fine

docs/user/restore.md
asafbennatan: nice , fixed

sk-ilya (approved):

docs/user/restore.md
```diff
@@ -180,18 +198,37 @@ kubectl get pods -n flightctl-internal
...
+
+Stop the Flight Control application services to prevent data conflicts during the restore process:
+
+```bash
+# Stop only the application services (keep database and KV store running)
+sudo systemctl stop flightctl-api.service
+sudo systemctl stop flightctl-worker.service
+sudo systemctl stop flightctl-periodic.service
+sudo systemctl stop flightctl-alert-exporter.service
+sudo systemctl stop flightctl-alertmanager-proxy.service
```
siserafin, line 229: Should I also stop flightctl-alertmanager or doesn't matter?

PR #1761, opened 2025-10-06
EDM-2266: Append org id param to console requests from cli and addres…

DakCrowder:
…s outdated const refs

## Summary by CodeRabbit

- New Features
  - Console URLs now include the organization identifier for organization-aware console sessions.
  - Requests can specify organization via a standardized query parameter for consistent routing.

- Bug Fixes
  - Device existence checks and enrollment request deletion now respect the selected organization, preventing cross-organization mismatches.

- Refactor
  - Console session startup now derives organization from context for more consistent behavior.
  - Middleware and transport updated to use the standardized organization query parameter.

### DIFF
--- a/api/v1alpha1/consts.go
+++ b/api/v1alpha1/consts.go
@@ -89,6 +89,7 @@ const (
 	OrganizationAPIVersion = "v1alpha1"
 	OrganizationKind       = "Organization"
 	OrganizationListKind   = "OrganizationList"
+	OrganizationIDQueryKey = "org_id"
 
 	SystemKind           = "System"
 	SystemComponentDB    = "database"
--- a/internal/cli/console.go
+++ b/internal/cli/console.go
@@ -181,6 +181,7 @@ func (o *ConsoleOptions) buildURL(baseURL, metadata string) (string, error) {
 	// Create query parameters
 	query := url.Values{}
 	query.Set(api.DeviceQueryConsoleSessionMetadata, metadata)
+	query.Set(api.OrganizationIDQueryKey, o.GetEffectiveOrganization())
 
 	// Encode the query parameters and attach them to the URL
 	u.RawQuery = query.Encode()
--- a/internal/service/enrollmentrequest.go
+++ b/internal/service/enrollmentrequest.go
@@ -17,7 +17,6 @@ import (
 	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/kvstore"
 	"github.com/flightctl/flightctl/internal/service/common"
-	"github.com/flightctl/flightctl/internal/store"
 	"github.com/flightctl/flightctl/internal/store/selector"
 	"github.com/flightctl/flightctl/internal/tpm"
 	"github.com/flightctl/flightctl/internal/util"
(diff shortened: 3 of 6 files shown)


api/v1alpha1/consts.go
```diff
@@ -86,9 +86,10 @@ const (
 
 	EventAnnotationDelayDeviceRender = "fleet-controller/delayDeviceRender"
 
-	OrganizationAPIVersion = "v1alpha1"
-	OrganizationKind       = "Organization"
-	OrganizationListKind   = "OrganizationList"
+	OrganizationAPIVersion   = "v1alpha1"
+	OrganizationKind         = "Organization"
+	OrganizationListKind     = "OrganizationList"
+	OrganizationIDQueryParam = "org_id"
```
hexfusion: minor this is really the query param key so maybe 
```suggestion
	OrganizationIDQueryKey= "org_id"
```

internal/service/enrollmentrequest.go
```diff
@@ -526,8 +525,8 @@ func (h *ServiceHandler) allowCreationOrUpdate(ctx context.Context, orgId uuid.U
 
 // deviceExists checks if a device with the given name exists in the store.
 // Error is returned if there is an error other than ErrResourceNotFound.
-func (h *ServiceHandler) deviceExists(ctx context.Context, name string) (bool, error) {
-	dev, err := h.store.Device().Get(ctx, store.NullOrgId, name)
+func (h *ServiceHandler) deviceExists(ctx context.Context, orgId uuid.UUID, name string) (bool, error) {
+	dev, err := h.store.Device().Get(ctx, orgId, name)
```
DakCrowder, line 529: Note: bonus bugfix find for DeleteEnrollmentRequest, will mention in the ticket and ensure it is tested.

internal/transport/console.go
```diff
@@ -29,7 +28,6 @@ func (h *WebsocketHandler) injectProtocolsToMetadata(metadataStr string, protoco
 }
 
 func (h *WebsocketHandler) HandleDeviceConsole(w http.ResponseWriter, r *http.Request) {
-	orgId := store.NullOrgId
```
DakCrowder, line 32: This wasn't actually the cause of the bug as the org id is correctly pulled from context in the service layer when updating the device annotations, but it is misleading so cleaning up usage for clarity.

hexfusion (approved):

internal/cli/console.go
```diff
@@ -181,6 +181,7 @@ func (o *ConsoleOptions) buildURL(baseURL, metadata string) (string, error) {
 	// Create query parameters
 	query := url.Values{}
 	query.Set(api.DeviceQueryConsoleSessionMetadata, metadata)
+	query.Set(api.OrganizationIDQueryKey, o.GetEffectiveOrganization())
```
celdrake, line 184: IMO this doesn't solve the original problem, since IIUC, you're defaulting to the last selected organization in the CLI from the config file, which assumes the user had previously successfully logged in the system.

For a new a UI user which has never before used the CLI, now we'll give them the login command to the CLI, and all they will see is that they fail to retrieve any entities, as all requests will fail with error 403. There's no informative message telling them what is wrong.

```
flightctl get fleets                                                                                                               
Error: listing fleets: response status: 403
```

hexfusion: cc @DakCrowder

DakCrowder: @celdrake This pr addresses a specific issue with the console endpoint where the 403 could occur even after the user is logged in as we weren't sending the org properly.  What you raised is still an issue and needs to be tackled separately.

celdrake: In that case, it looks like this PR is referencing the wrong Jira ticket.
EDM-2266 is about the situation I described.

DakCrowder: Ugh my wires got totally crossed yesterday between this issue and another I was looking at, will get that sorted out

PR #1763, opened 2025-10-06
EDM-2271: Fix TPM Activate Credential with tracing disabled

kkyrazis:
The TPM Activate Credential flow fails if tracing is disabled. This is because it's the only gRPC handler (as of now) that calls the services to read/write data to/from the database. When initializing the gorm db instance, callbacks are registered the ensure that tracing spans are setup appropriately. This check fails when the call originates from a gRPC originated call compared to a typical REST call. 

The difference is in how otel is configured:
[gRPC config](https://github.com/flightctl/flightctl/blob/main/internal/api_server/agentserver/grpc.go#L55)
[http config](https://github.com/flightctl/flightctl/blob/main/internal/api_server/server.go#L310)

Debugging shows that the span available to the traced services flowing from gRPC is a `trace.nonRecordingSpan` while the span from the http flow is a `noop.Span`. In both instances, the same `noop.TracerProvider` is used. Since only `noop.Span` was checked for, the gRPC flow would fail.

I've changed the behavior so that registration of the callbacks only occurs if necessary (tracing is enabled) 

## Summary by CodeRabbit

* **Refactor**
  * Initialize database tracing only when tracing is enabled to reduce overhead.
  * Enforce tracing only when a valid trace context is present for more accurate telemetry.

* **Chores**
  * Added a warning log for retryable enrollment errors to aid diagnostics.
  * Removed unused imports and helper code; minor formatting cleanups.

### DIFF
--- a/internal/agent/client/podman.go
+++ b/internal/agent/client/podman.go
@@ -450,6 +450,59 @@ func (p *Podman) RemoveNetworks(ctx context.Context, networks ...string) error {
 	return nil
 }
 
+func (p *Podman) ListPods(ctx context.Context, labels []string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, p.timeout)
+	defer cancel()
+
+	// pods created by podman-compose don't have the compose project label,
+	// so we need to get pod IDs from the containers that do have the label
+	args := []string{
+		"ps",
+		"-a",
+		"--format",
+		"{{.Pod}}",
+	}
+	for _, label := range labels {
+		args = append(args, "--filter", fmt.Sprintf("label=%s", label))
+	}
+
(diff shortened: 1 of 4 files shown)


internal/agent/device/lifecycle/manager.go
```diff
@@ -283,6 +283,7 @@ func (m *LifecycleManager) verifyEnrollment(ctx context.Context) (bool, error) {
 			if errors.Is(err, identity.ErrIdentityProofFailed) {
 				return false, fmt.Errorf("proving identity: %w", err)
 			}
+			m.log.Infof("A retryable error occurred while proving the agent's identity: %v", err)
```
hexfusion: warn?

hexfusion (approved):

PR #1766, opened 2025-10-07
EDM-2273: device/dependency: ensure stale images are removed on version change

hexfusion:
The prefetch manager continues pulling OCI images for an old version even after a new rendered version is pushed. This can create a situation where invalid images are never removed and block updates.

```
go test -benchmem -run=^$ -bench ^BenchmarkPrefetchTargetChange$ github.com/flightctl/flightctl/internal/agent/device/dependency

goos: linux
goarch: amd64
pkg: github.com/flightctl/flightctl/internal/agent/device/dependency
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkPrefetchTargetChange/small_partial_change-12         	 1423314	       848.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkPrefetchTargetChange/medium_partial_change-12        	  435942	      2794 ns/op	     663 B/op	       1 allocs/op
BenchmarkPrefetchTargetChange/large_partial_change-12         	   94155	     12661 ns/op	    2839 B/op	       2 allocs/op
PASS
ok  	github.com/flightctl/flightctl/internal/agent/device/dependency	72.854s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale tasks when targets change to prevent leftover work and reduce resource use.
  * Added safeguards around optional cleanup handlers to avoid rare nil-call crashes.
  * More accurate detection of target changes to avoid unnecessary reprocessing.

* **Refactor**
  * Streamlined target-change handling for greater stability and maintainability.

* **Tests**
  * Added tests for selective cleanup and target-change scenarios.
  * Added benchmarks to measure performance with varying target sets.

### DIFF
--- a/internal/agent/device/dependency/dependency.go
+++ b/internal/agent/device/dependency/dependency.go
@@ -140,6 +140,26 @@ func (m *prefetchManager) RegisterOCICollector(collector OCICollector) {
 	m.collectors = append(m.collectors, collector)
 }
 
+// isTargetsChanged checks if targets changed using pre-built reference set
+// caller must hold m.mu lock
+func (m *prefetchManager) isTargetsChanged(seenTargets map[string]struct{}) bool {
+	if len(m.tasks) == 0 {
+		return len(seenTargets) > 0
+	}
+
+	if len(seenTargets) != len(m.tasks) {
+		return true
+	}
+
+	for existingRef := range m.tasks {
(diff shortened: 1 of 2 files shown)


siserafin:
Tested with this PR branch. I tried to add an image application with typo to a device, and update to a correct one. The device updated to the latest, and was not stuck. LGTM.

PR #1767, opened 2025-10-07
EDM-2255: added flightctl-restore to the rpm

hexfusion:
replacement for https://github.com/flightctl/flightctl/pull/1753

## Summary by CodeRabbit

- New Features
  - Introduces a new flightctl-restore command, installed as /usr/bin/flightctl-restore and included in the CLI package to support restore operations.

- Chores
  - Updated packaging to include the new restore utility in builds and installation.
  - Ensured compatibility across target platforms, including RHEL 9 and FIPS/non-FIPS variants.

PR #1768, opened 2025-10-07
EDM-2255: added flightctl-restore to the rpm

hexfusion:
This is a manual cherry-pick of #1767


kkyrazis (approved):

hexfusion:
I dont think we really want this holding to consider options

hexfusion:
Moving forward, the fixes can be applied as a follow-up for GA.

PR #1775, opened 2025-10-07
EDM-2043: Create SSH known_hosts file during installation

SiddarthR56:


DakCrowder (changes requested):
We also need to update the .spec so the rpm knows it owns the ssh directory, and if we are creating the known_hosts file on install think that needs to be added too.

Similar to this: https://github.com/flightctl/flightctl/blob/main/packaging/rpm/flightctl.spec#L616 
and
https://github.com/flightctl/flightctl/blob/main/packaging/rpm/flightctl.spec#L621

kkyrazis:
> We also need to update the .spec so the rpm knows it owns the ssh directory, and if we are creating the known_hosts file on install think that needs to be added too.
> 
> Similar to this: https://github.com/flightctl/flightctl/blob/main/packaging/rpm/flightctl.spec#L616 and https://github.com/flightctl/flightctl/blob/main/packaging/rpm/flightctl.spec#L621

What Dakota said. The rpm-build logs fortunately failed and output the following logs.

```
error: Installed (but unpackaged) file(s) found:
   /etc/flightctl/ssh/known_hosts
    %changelog not in descending chronological order
    Installed (but unpackaged) file(s) found:
   /etc/flightctl/ssh/known_hosts
```

DakCrowder (approved):

PR #1785, opened 2025-10-09
EDM-2214: helm add migration wait init containers to services

SiddarthR56:
Problem: When deploying FlightCtl with the Helm chart, if the database takes time to become available, the migration job can fail permanently after exhausting its retry attempts (`backoffLimit: 3`). Even after the database recovers, services start before migrations complete, causing API server to  reports "Ready" but logs errors about missing database tables (`devices`, `fleets`, etc.)

Add migration wait init containers to all services that directly access the database. Services now wait for the migration job to complete successfully before starting their main containers.

**Tested migration dependency with multiple failure scenarios:**
- Slow database connection - dependent services wait correctly
- Wrong database password - migration fails, dependent services blocked
-  Wrong database port - migration fails fast, dependent services properly blocked
- Successful migration - dependent services start only after migration completes

All dependent services (API, Worker, Periodic, Alert-Exporter, Alertmanager-Proxy) now properly wait for database migrations to complete before starting, preventing connection failures and ensuring schema is ready.

Approach: Init container uses `kubectl` to check migration job status (checking for job completion is more reliable than attempting to infer schema state).

## Summary by CodeRabbit

* **New Features**
  * Adds a migration-wait init step that pauses startup until DB migrations complete for API, worker, periodic, alert-exporter and alertmanager-proxy; configurable timeout (default 10 minutes).

* **Enhancements**
  * Deployments now set explicit service accounts for periodic and alert-exporter and include the migration-wait init step where applicable.

* **Chores**
  * Added service accounts and a job-status-reader role with expanded cross-namespace bindings for job status access.

* **Documentation**
  * Added troubleshooting guidance for migration failures (logs and common causes).

### DIFF
--- a/deploy/helm/flightctl/templates/_helpers.tpl
+++ b/deploy/helm/flightctl/templates/_helpers.tpl
@@ -272,6 +272,79 @@ Parameters:
   {{- include "flightctl.dbSslVolumeMounts" $context | nindent 2 }}
 {{- end }}
 
+{{/*
+Migration wait init container template.
+Waits for database migration job to complete before starting the main container.
+Usage: {{- include "flightctl.migrationWaitInitContainer" (dict "context" .) | nindent 6 }}
+Parameters:
+- context: The root template context (.)
+- timeout: Optional timeout in seconds (default: 600)
+*/}}
+{{- define "flightctl.migrationWaitInitContainer" }}
+{{- $ctx := .context }}
+{{- $timeout := .timeout | default 600 | int }}
+- name: wait-for-migration
+  image: "{{ $ctx.Values.clusterCli.image.image }}:{{ $ctx.Values.clusterCli.image.tag }}"
+  imagePullPolicy: {{ default $ctx.Values.global.imagePullPolicy $ctx.Values.clusterCli.image.pullPolicy }}
+  command:
+  - /bin/bash
+  - -c
+  - |
+    set -euo pipefail
+
+    LABEL_SELECTOR="app=flightctl-db-migration,flightctl.io/migration-revision={{ $ctx.Release.Revision }}"
+    TIMEOUT={{ $timeout }}
+    NS="{{ default $ctx.Release.Namespace $ctx.Values.global.internalNamespace }}"
+
+    echo "Waiting for migration job with labels: $LABEL_SELECTOR (timeout ${TIMEOUT}s)"
+    start=$(date +%s)
+
+    while true; do
+      elapsed=$(( $(date +%s) - start ))
+
+      if [ $elapsed -ge $TIMEOUT ]; then
+        echo "Timeout waiting for migration job after ${TIMEOUT}s"
+        exit 1
+      fi
+
+      # Find job by label selector
+      JOB=$(kubectl get jobs -n "$NS" -l "$LABEL_SELECTOR" -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
+      
+      # Check if job exists
+      if [ -n "$JOB" ]; then
+        # Check for parallel execution (dangerous even with idempotent migrations)
+        parallelism=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.spec.parallelism}' 2>/dev/null || echo 1)
+        if (( parallelism > 1 )); then
+            echo "ERROR: Migration job has spec.parallelism=$parallelism (must be 1)"
+            echo "Database migrations must not run in parallel due to race condition risks"
+            exit 1
+        fi
+
+        succeeded=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.status.succeeded}' 2>/dev/null || echo 0)
+        failed=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.status.failed}' 2>/dev/null || echo 0)
+
+        if (( succeeded > 0 )); then
+            # The 'greater than 1' scenario would only occur if the job spec.completions is set to more than 1
+            if (( succeeded > 1 )); then
+                echo "Warning: Migration job completed $succeeded times (expected 1). Ensure spec.completions is set to 1."
+            fi
+            echo "Migration job $JOB completed successfully"
+            exit 0
+        elif (( failed > 0 )); then
+            echo "Migration job $JOB failed"
+            exit 1
+        else
+            echo "Migration job $JOB is still running..."
+        fi
+      else
+        # Job not found - could be not created yet, RBAC issue, or other problem
+        echo "Migration job not found yet, waiting..."
+      fi
+
+      sleep 5
+    done
+{{- end }}
+
 {{- /*
 SSL certificate volume mounts for database connections.
 Usage: {{- include "flightctl.dbSslVolumeMounts" . | nindent X }}
--- a/deploy/helm/flightctl/templates/flightctl-db-admin-rbac.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-db-admin-rbac.yaml
@@ -11,6 +11,25 @@ metadata:
     security.level: high-privilege
 
 ---
+# ServiceAccounts for internal namespace services
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: flightctl-periodic
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+  labels:
+    flightctl.service: flightctl-periodic
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: flightctl-alert-exporter
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+  labels:
+    flightctl.service: flightctl-alert-exporter
+
+---
 # Role for accessing admin database secrets (superuser only)
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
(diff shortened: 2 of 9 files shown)


deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -272,6 +272,62 @@ Parameters:
...
+Waits for database migration job to complete before starting the main container.
+Usage: {{- include "flightctl.migrationWaitInitContainer" (dict "context" .) | nindent 6 }}
+Parameters:
+- context: The root template context (.)
+- timeout: Optional timeout in seconds (default: 600)
+*/}}
+{{- define "flightctl.migrationWaitInitContainer" }}
+{{- $ctx := .context }}
+{{- $timeout := .timeout | default 600 | int }}
+{{- $jobName := printf "flightctl-db-migration-%d" $ctx.Release.Revision }}
```
sk-ilya: Nit: consider adding `standardLabels` to the migration job

https://github.com/flightctl/flightctl/blob/887936cfc75093170846705612fd92ca60250469/deploy/helm/flightctl/templates/_helpers.tpl#L21-L24

... and querying the jobs in a more reliable way by `labels.app`=`flightctl-db-migration` && `labels.helm.sh/chart`=`{the value of this deployment}` instead of by the job name.

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -272,6 +272,62 @@ Parameters:
...
+        failed=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.status.conditions[?(@.type=="Failed")].status}' || true)
+
+        if [ "$complete" = "True" ]; then
+          echo "Migration job completed successfully"
+          exit 0
+        fi
+
+        if [ "$failed" = "True" ]; then
+          echo "Migration job failed"
+          kubectl get job "$JOB" -n "$NS" -o yaml | tail -20
```
hexfusion: I have not looked at this output but are we looking for the `Reason` and `Message` for the `Failed` condition or what other interesting data would we find in this dump? just from a user perspective can we make the output more perscriptive?

deploy/helm/flightctl/templates/_helpers.tpl
siserafin: I deployed flightctl from this PR branch in OCP standalone and ACM. It seems it took a little longer than usual, but all the pods became ready. I noticed that the wait-for-migration container was Created and Started.  I could login with no issues with both the deployments. lgtm

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -272,6 +272,62 @@ Parameters:
...
+
+    while true; do
+      elapsed=$(( $(date +%s) - start ))
+
+      if [ $elapsed -ge $TIMEOUT ]; then
+        echo "Timeout waiting for migration job after ${TIMEOUT}s"
+        exit 1
+      fi
+
+      if kubectl get job "$JOB" -n "$NS" &>/dev/null; then
```
hexfusion: question: what is the net result if the job does not exist. is it safe to assume

SiddarthR56: If the job doesn’t exist, the command will fail, the check will keep retrying every 5 seconds, and eventually time out after 10 minutes.

hexfusion: can we log this condition for observability?

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -308,20 +308,26 @@ Parameters:
...
-        if [ "$complete" = "True" ]; then
-          echo "Migration job completed successfully"
-          exit 0
-        fi
-
-        if [ "$failed" = "True" ]; then
-          echo "Migration job failed"
-          kubectl get job "$JOB" -n "$NS" -o yaml | tail -20
-          exit 1
+        if (( succeeded > 0 )); then
```
hexfusion, line 326: would we ever want more than one to succeed when would that happen, would it be idempotent?

SiddarthR56:
**NOTE**: When the migration job fails, the init container continuously restarts because it repeatedly detects the same failed job and exits with a non-zero status. This causes the parent Pod (e.g., API or worker) to enter a CrashLoopBackOff, preventing the main container from ever starting. To recover, either fix the migration issue and rerun the job, or delete the failed migration job so the init container can detect a new one and proceed once it completes successfully. If the failed job remains unchanged, the application will stay unavailable indefinitely.

deploy/helm/flightctl/templates/_helpers.tpl
SiddarthR56: We would almost never want more than one pod to succeed for a migration job—normally completions: 1. Multiple successes could only happen if the job spec is misconfigured with completions > 1. Even then, the migration should be idempotent, so it’s safe to treat any succeeded > 0 as migration done

deploy/helm/flightctl/templates/flightctl-db-admin-rbac.yaml
```diff
@@ -80,6 +99,20 @@ roleRef:
...
+kind: Role
+metadata:
+  name: flightctl-job-status-reader
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+  labels:
+    app: flightctl-job-status
+rules:
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["get", "list"]
```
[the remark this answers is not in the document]
SiddarthR56: Not required for current implementation, using kubectl get with polling (every 5 seconds), not kubectl wait.

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -272,6 +272,68 @@ Parameters:
...
+        elif (( failed > 0 )); then
+            echo "Migration job $JOB failed"
+            exit 1
+        else
+            echo "Migration job $JOB is still running..."
+        fi
+      else
+        # Job not found - could be not created yet, RBAC issue, or other problem
+        echo "Migration job not found yet, waiting..."
+      fi
```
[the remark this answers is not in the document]
SiddarthR56: IMO Not necessary, RBAC issues should surface clearly in pod events and users can also check the logs

hexfusion:
> **NOTE**: When the migration job fails, the init container continuously restarts because it repeatedly detects the same failed job and exits with a non-zero status. This causes the parent Pod (e.g., API or worker) to enter a CrashLoopBackOff, preventing the main container from ever starting. To recover, either fix the migration issue and rerun the job, or delete the failed migration job so the init container can detect a new one and proceed once it completes successfully. If the failed job remains unchanged, the application will stay unavailable indefinitely.

should we add this type of context to the docs?

deploy/helm/flightctl/templates/_helpers.tpl
hexfusion: lets make sure it IS idempotent then lgtm  ;)

SiddarthR56:
> > **NOTE**: When the migration job fails, the init container continuously restarts because it repeatedly detects the same failed job and exits with a non-zero status. This causes the parent Pod (e.g., API or worker) to enter a CrashLoopBackOff, preventing the main container from ever starting. To recover, either fix the migration issue and rerun the job, or delete the failed migration job so the init container can detect a new one and proceed once it completes successfully. If the failed job remains unchanged, the application will stay unavailable indefinitely.
> 
> should we add this type of context to the docs?

Added [this](https://github.com/flightctl/flightctl/pull/1785/files#diff-cf1894d213434e5c520cc84d541f557212d4647e7576bad68958337d1cef9822) to the docs

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -272,6 +272,67 @@ Parameters:
...
+      # Find job by label selector
+      JOB=$(kubectl get jobs -n "$NS" -l "$LABEL_SELECTOR" -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
+      
+      # Check if job exists
+      if [ -n "$JOB" ]; then
+        succeeded=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.status.succeeded}' 2>/dev/null || echo 0)
+        failed=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.status.failed}' 2>/dev/null || echo 0)
+
+        if (( succeeded > 0 )); then
+            echo "Migration job $JOB completed successfully"
```
hexfusion, line 331: I think we should log the case where we see succeeded > 1 as this would be IMO unexpected. wdyt?

SiddarthR56: The migration uses GORM’s AutoMigrate(), which is idempotent but not safe to run concurrently.

docs/user/database-migration.md
```diff
@@ -112,6 +112,8 @@ The Helm deployment automatically handles user creation and migration:
 # 3. Run database migrations using migration user
 ```
 
+> **Troubleshooting Migration Failures**: If the migration job fails, check logs with `kubectl logs -n flightctl job/flightctl-db-migration-<revision> --all-containers` to identify the issue. Common causes include DB connection problems, incorrect credentials, missing database users, or database not ready.
```
hexfusion, line 115: 👍

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -272,6 +272,79 @@ Parameters:
...
+      
+      # Check if job exists
+      if [ -n "$JOB" ]; then
+        # Check for parallel execution (dangerous even with idempotent migrations)
+        parallelism=$(kubectl get job "$JOB" -n "$NS" -o jsonpath='{.spec.parallelism}' 2>/dev/null || echo 1)
+        if (( parallelism > 1 )); then
+            echo "ERROR: Migration job has spec.parallelism=$parallelism (must be 1)"
+            echo "Database migrations must not run in parallel due to race condition risks"
+            exit 1
+        fi
```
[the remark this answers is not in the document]
SiddarthR56: The migration uses GORM’s AutoMigrate(), which is idempotent. Logging a warning in case of Multiple Succeeds

PR #1794, opened 2025-10-15
EDM-2214: create migration pod just once and retry without a limit

sk-ilya:
## Summary by CodeRabbit

- New Features
  - Migration job supports effectively unlimited retries (very large default) and no deadline when set to 0.

- Chores
  - Migration job now restarts on failure, runs a single completion and single parallel instance by default.

- Documentation
  - Clarified configuration semantics: "0 = no deadline" and updated retry default description to reflect unbounded retries.

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-preupgrade-scale-jobs.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-preupgrade-scale-jobs.yaml
@@ -22,7 +22,7 @@
   {{- end -}}
 {{- end -}}
 
-{{- if gt (len $deploymentsToScale) 0 }}
+{{- if and (gt (len $deploymentsToScale) 0) (ne .Values.global.target "acm") }}
 ---
 # ServiceAccount and RBAC for pre-upgrade scaling
 apiVersion: v1
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -3,7 +3,7 @@
 ## Building
 
 Prerequisites:
-* `git`, `make`, and `go` (>= 1.23), `openssl`, `openssl-devel`, `buildah`, `podman`, `podman-compose`, `container-selinux` (>= 2.241) and `go-rpm-macros` (in case one needs to build RPM's)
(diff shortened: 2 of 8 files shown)


deploy/helm/flightctl/values.yaml
```diff
@@ -345,10 +345,10 @@ dbSetup:
     # Default sleep interval between connection attempts
     sleep: 2
   migration:
-    # -- Number of retries for the migration Job on failure
-    backoffLimit: 3
-    # -- Maximum runtime in seconds for the migration Job
-    activeDeadlineSeconds: 600
+    # -- Number of retries for the migration Job on failure 
+    backoffLimit: 2147483647
```
SiddarthR56, line 349: Would setting infinite retries with no timeout risk causing the pre-upgrade hook to retry indefinitely if the database is misconfigured, potentially blocking deployments without providing a clear failure signal?

sk-ilya: The upgrade will fail after a timeout with:
```
Error: UPGRADE FAILED: pre-upgrade hooks failed: 1 error occurred:
        * timed out waiting for the condition
```
The default timeout is 5 mins

deploy/helm/flightctl/templates/_db-migration-job.tpl
```diff
@@ -42,7 +46,7 @@ spec:
         flightctl.io/migration-revision: "{{ $ctx.Release.Revision }}"
         {{- include "flightctl.standardLabels" $ctx | nindent 8 }}
     spec:
-      restartPolicy: Never
+      restartPolicy: OnFailure
```
SiddarthR56, line 49: Do you think keeping it as Never might be safer to ensure the init logic runs on each retry, or would that be unnecessary?

sk-ilya: The main reason is to keep a single pod in `CrashLoopBackOff` state. "Never" will result in a new pod created each time.

SiddarthR56: Is the debugging convenience of having a single Pod in CrashLoopBackOff worth skipping the database availability recheck when the main migration container fails?

PR #1795, opened 2025-10-15
EDM-2195: Use k8s job to create secrets. Do not render cleanup/db mig…

rawagner:
…ration jobs for ACM.

Due to ACM's approach of running helm template + applying manifests manually, we can not use Secret resources directly. We need to generate the Secret resources in runtime. Due to the same reason, hooks also do not work so we skip cleanup and db migration jobs rendering when ACM is target platform.

## Summary by CodeRabbit

* **New Features**
  * Added a job that generates and ensures DB and KV secrets across namespaces when missing, plus a service account and RBAC updates to support it.
  * Keycloak secrets now create when target is "standalone" and auth is "builtin" regardless of the old toggle.
  * Cleanup job now runs only when target is not "acm".

* **Refactor**
  * Removed default secret/password config fields and the global secret-generation toggle; secrets are now managed via the new job.

* **Documentation**
  * Updated Helm chart docs to reflect the new secret management flow and removed config options.


siserafin:
I deployed and login to edge manager in both ACM and OCP standalone. All the secrets are created. LGTM

rawagner:
closing, we are decoupling from ACM operator, so we can keep generating secrets.

PR #1803, opened 2025-10-16
EDM-1834: Agent attempts to reconcile before reboot on OS upgrade

noga-magen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of canceled device configuration upgrades to suppress spurious errors, reduce noise in logs, and improve system stability and reliability.


internal/agent/device/device.go
```diff
@@ -217,6 +217,10 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 
 	// reconciliation is a success, upgrade the current spec
 	if err := a.specManager.Upgrade(ctx); err != nil {
+		if errors.Is(err, context.Canceled) {
+			a.log.Warnf("Failed to upgrade spec: %v", err)
```
hexfusion: since there is a known condition here in upgrade flow can we ask the ctx before we try to upgrade if there is a ctx error and log.

```go
    if err := ctx.Err(); err != nil {
    	 a.log.Debugf("Sync is shutting down: %v", err)
         return
    }

	if err := a.specManager.Upgrade(ctx); err != nil {
```

ref https://github.com/flightctl/flightctl/blob/ef5bb2375a3171c2120e3ea51e64528c494e406d/internal/agent/device/errors/errors.go#L205C6-L205C15

PR #1745, opened 2025-10-02
EDM-2248: Replace yq dep with jq/pyyaml

DakCrowder:
## Summary by CodeRabbit

* **New Features**
  * Built-in YAML→JSON converter added and provided as an executable for runtime use.

* **Refactor**
  * Switched YAML parsing everywhere to a python3 + jq pipeline, preserving previous behavior and defaults.

* **Chores**
  * Images/packages and runtime requirements updated to include python3, PyYAML and jq; dependency checks and warnings adjusted.

PR #1805, opened 2025-10-16
EDM-2351: Add python utility for transforming yaml

DakCrowder:
There are two issues this PR seeks to solve:
1.  The [.observability.telemetry_gateway.config here](https://github.com/flightctl/flightctl/blob/main/packaging/observability/telemetry-gateway.defs#L7) that is referenced in telemetry setup scripts is nested yaml - wheras the previous script [added here](https://github.com/flightctl/flightctl/pull/1745/files#diff-176cb95c435dea467d22cd5f183a9d063c1d81d8699bffdbf810f06acf94e766) would put out json.

Solution: Updated the python script to handle and output values from yaml including nested yaml.  Removes jq dependency from those commands.

2. File paths in setup_telemetry_gateway_certs.sh always referenced the rpm installed location, but that script is also run as `make deploy`.

Solution: added an arg to the setup_telemetry_gateway_certs.sh script so the script path is not hardcoded.

## Summary by CodeRabbit

* **Chores**
  * Updated package dependencies: removed jq requirement, retained python3-pyyaml across services and observability components.
  * Refactored internal YAML configuration extraction utilities for improved maintainability.
  * Updated deployment and initialization scripts to use simplified configuration handling.

* **Documentation**
  * Updated developer documentation to reflect configuration extraction changes.


deploy/scripts/yaml_helpers.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
```
hexfusion, line 1: could we unify this into something like a yaml_helpers.py and expose subcommands?

```
yaml_helpers.py extract <path> <data>
etc
```

DakCrowder: makes sense to me - will refactor

deploy/podman/flightctl-db-migrate/flightctl-db-users-init.container
```diff
@@ -32,7 +32,7 @@ Type=oneshot
 RemainAfterExit=yes
 Restart=on-failure
 RestartSec=5
-ExecCondition=python3 /usr/share/flightctl/yaml-to-json.py < /etc/flightctl/service-config.yaml | jq -e '.db.external != "enabled"'
+ExecCondition=/bin/bash -c 'test "$(python3 /usr/share/flightctl/yaml_helpers.py extract ".db.external" /etc/flightctl/service-config.yaml --default "disabled")" != "enabled"'
```
hexfusion, line 35: why test prefix?

DakCrowder: Allows for the != and translates into an error code so the ExecCondition skips the run properly, the -e arg to jq previously handled it

PR #1808, opened 2025-10-19
EDM-2236: Add support for space-separated approve command syntax

galel12:
## Problem

The \`flightctl approve\` command currently only supports the slash-separated syntax (\`flightctl approve er/name\`). Users expect to be able to use space-separated syntax (\`flightctl approve er name\`) similar to other CLI tools like kubectl.

## Root Cause

The \`approve\` command was configured with \`cobra.ExactArgs(1)\` which only accepts a single argument, and used \`parseAndValidateKindName()\` which expects the \`TYPE/NAME\` format.

## Solution

Following guidance from the existing codebase patterns, I implemented the following changes:

* **Updated command configuration** to accept 1-2 arguments using \`cobra.RangeArgs(1, 2)\`
* **Replaced argument parsing** to use \`parseAndValidateKindNameFromArgsSingle()\` which handles both formats
* **Updated usage documentation** to show both supported syntaxes: \`TYPE/NAME or TYPE NAME\`
* **Maintained backward compatibility** - existing \`TYPE/NAME\` format continues to work

## Comprehensive Code Quality

* **Unit Tests**: Added 11 comprehensive test cases in \`approve_test.go\`
* **Real-world Testing**: Verified with actual enrollment request from agent VM  
* **Backward Compatibility**: Confirmed existing slash format still works
* **Error Handling**: Proper validation for invalid inputs and edge cases

## Usage Examples

``` bash
# New space-separated syntax ✅
flightctl approve er myrequest
flightctl approve enrollmentrequest myrequest

# Existing slash syntax (still supported) ✅  
flightctl approve er/myrequest
flightctl approve enrollmentrequest/myrequest
```

## Summary by CodeRabbit

* **New Features**
  * Approve command now accepts resource specifications as either "TYPE/NAME" or "TYPE NAME" and updated usage text.

* **Behavior Changes**
  * Argument validation now allows either one or two arguments for resource specification.

* **Tests**
  * Added tests covering parsing and validation for multiple resource kinds and syntax scenarios.

### DIFF
--- a/internal/agent/device/systeminfo/system_info.go
+++ b/internal/agent/device/systeminfo/system_info.go
@@ -457,6 +457,8 @@ func collectionOptsFromInfoKeys(infoKeys []string) ([]CollectOpt, error) {
 			opts = append(opts, withCollector(collectorKernel, collectKernelFunc))
 		case distroNameKey, distroVersionKey:
 			opts = append(opts, withCollector(collectorDistribution, collectDistributionFunc))
+		case hostnameKey, architectureKey:
+			// No specific collector needed - hostname and architecture are always collected
 		default:
 			errs = append(errs, fmt.Errorf("unknown key: %q", key))
 		}
--- a/internal/agent/device/systeminfo/system_info_test.go
+++ b/internal/agent/device/systeminfo/system_info_test.go
@@ -259,6 +259,20 @@ func TestGetCollectionOptsFromInfoKeys(t *testing.T) {
 			expectGPU: true,
 			expectErr: true,
 		},
+		{
+			name:      "hostname key should be supported",
+			infoKeys:  []string{hostnameKey},
+			expectErr: false, // Fixed: hostname is now properly supported
+		},
+		{
+			name:          "default system info with hostname key",
+			infoKeys:      config.DefaultSystemInfo, // This includes "hostname" which is now supported
+			expectErr:     false,                    // Fixed: default config now works properly
+			expectKernel:  true,                     // kernel is in default config
+			expectDistro:  true,                     // distro keys are in default config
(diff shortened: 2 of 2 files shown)


internal/cli/approve_test.go
```diff
@@ -0,0 +1,163 @@
...
+				}
+			} else {
+				if err != nil {
+					t.Errorf("unexpected error: %v", err)
+				}
+				if kind != tc.expectedKind {
+					t.Errorf("expected kind %q, got %q", tc.expectedKind, kind)
+				}
+				if name != tc.expectedName {
+					t.Errorf("expected name %q, got %q", tc.expectedName, name)
```
hexfusion: minor: with unit tests we have tried to standardize on using require pkg so something like below would be the translation

```go
for _, tc := range tests {
	t.Run(tc.name, func(t *testing.T) {
		require := require.New(t)

		kind, name, err := parseAndValidateKindNameFromArgsSingle(tc.args)
		if tc.expectError {
			require.Error(err, "expected an error but got none")
			if tc.errorContains != "" {
				require.Contains(err.Error(), tc.errorContains,
					"expected error to contain %q", tc.errorContains)
			}
            return
		} 

	    require.NoError(err, "unexpected error")
	    require.Equal(tc.expectedKind, kind, "unexpected kind")
	    require.Equal(tc.expectedName, name, "unexpected name")
	})
}
```

hexfusion (approved):

internal/cli/approve_test.go
galel12: Thanks for the feedback!
Fixed in https://github.com/flightctl/flightctl/pull/1808/commits/f210af1348f391f6592cb695c107e6dd9df931d7

PR #1815, opened 2025-10-20
EDM-2251: remove start condition for db-migrate

sk-ilya:
Sometimes `config.yaml` isn’t ready _yet_, preventing the migration from starting 

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified database migration startup behavior to improve initialization without requiring a specific configuration file path check.

### DIFF
--- a/deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
+++ b/deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
@@ -2,8 +2,7 @@
 Description=Flight Control Database Migration service
 PartOf=flightctl.target
 After=flightctl-db-wait.service flightctl-db-users-init.service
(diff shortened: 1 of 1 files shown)


deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
```diff
@@ -2,8 +2,7 @@
 Description=Flight Control Database Migration service
 PartOf=flightctl.target
 After=flightctl-db-wait.service flightctl-db-users-init.service
-Wants=flightctl-db-wait.service flightctl-db-users-init.service
-ConditionPathExists=/etc/flightctl/flightctl-api/config.yaml
```
hexfusion, line 6: makes sense that db migrate does not depend on api config

hexfusion (approved):

PR #1601, opened 2025-09-02
EDM-2036: CLI Update Flightctl completion

SiddarthR56:
## Summary by CodeRabbit

* **New Features**
  * Added PowerShell autocompletion support, accepting "powershell" and the "pwsh" alias.
  * Enforced a single optional shell argument for the completion command.
  * Supported "completion help" as a shortcut to view guidance.
  * Printed a shell-specific header when generating completion scripts to a terminal.
  * Accepts shell names case-insensitively (e.g., "PWSh", "PowerShell").

* **Documentation**
  * Updated CLI help with multi-shell instructions, examples, and clearer usage listing (bash, zsh, fish, powershell).

PR #1821, opened 2025-10-21
EDM-2377: automate flightctl completion

eldar101:
## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests that generate and validate shell completion for bash, zsh, fish, and PowerShell (PowerShell gated on non-Windows). Tests write completion scripts with secure permissions, optionally run syntax checks, and source them in non-interactive shells to verify suggestions.
  * Wired completion checks into the test suite and added a lightweight validation utility plus reusable test helpers to assert autocompletion across resource commands.

### DIFF
--- a/test/e2e/cli/cli_test.go
+++ b/test/e2e/cli/cli_test.go
@@ -1,9 +1,13 @@
 package cli_test
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


test/e2e/cli/cli_test.go
```diff
@@ -622,6 +626,149 @@ var _ = Describe("cli login", func() {
...
+	It("generates completion and can be sourced for each supported shell (harness.CLI only for flightctl calls)", Label("85470"), func() {
+		harness := e2e.GetWorkerHarness()
+		type sh struct {
+			name       string
+			markers    []string
+			subprocess func(scriptPath, tmpHome string) *exec.Cmd // returns a non-interactive one-shot subshell
+			syntax     func(scriptPath string) *exec.Cmd          // optional syntax check command
+		}
+
+		look := func(bin string) bool { _, err := exec.LookPath(bin); return err == nil }
```
itzikb-redhat: look - There is a similar function BinaryExistsOnPath in test/util/ip.go maybe you can use it
must - Maybe it's worth to add this to util , it can be handy for other tests

itzikb-redhat (commented):
2 nits.

test/e2e/cli/cli_test.go
eldar101: done. thanks

eldar101:
@itzikb-redhat please approve if lgtm.

test/e2e/cli/cli_test.go
```diff
@@ -622,6 +627,168 @@ var _ = Describe("cli login", func() {
...
+					return cmd
+				},
+			},
+		}
+
+		for _, sc := range shells {
+			home, tmp := mkHome(sc.name)
+
+			// Check if generation is supported
+			if sc.name == "powershell" && runtime.GOOS != "windows" {
```
siserafin, line 1099: Is this tested in windows?

eldar101: See the PR
https://github.com/flightctl/flightctl/pull/1601

siserafin: ok, but did you try this test in windows? I think all these cases should be tested, or how can we tell they are working?

eldar101: @SiddarthR56 should i keep this pws scenario?

SiddarthR56: testd completion on windows; worth keeping pwsh scenario

siserafin: It's ok to keep it, I just think we should at least try this test once in window before merging, or we don't know if it works.

test/e2e/cli/cli_test.go
```diff
@@ -1195,6 +1468,22 @@ func AssertTableValue(out, resourceName, column, expected string) {
...
+	// Cobra prints candidates one-per-line, sometimes with "\t" description and a directive line at the end.
+	lines := strings.Split(strings.TrimSpace(out), "\n")
+	Expect(strings.Join(lines, "\n")).To(ContainSubstring(expected),
+		"expected completion %q for args: flightctl %s\nFull output:\n%s",
+		expected, strings.Join(args, " "), out)
+}
+
 // completeFleetYaml defines a YAML template for creating a Fleet resource with specified metadata and spec configuration.
 const completeFleetYaml = `
 apiVersion: v1alpha1
```
siserafin, line 1492: Do the binary files need to be committed?

test/e2e/cli/cli_test.go
```diff
@@ -1195,6 +1468,22 @@ func AssertTableValue(out, resourceName, column, expected string) {
...
 
+// CLIRunner is the minimal surface we need from the harness.
+type CLIRunner interface {
+	CLI(args ...string) (string, error)
+}
+
+// ExpectCompletion runs `flightctl __complete <args...>` and asserts that a suggestion contains `expected`.
+func ExpectCompletion(h CLIRunner, args []string, expected string) {
+	out, err := h.CLI(append([]string{"__complete"}, args...)...)
+	Expect(err).ToNot(HaveOccurred(), "completion failed for: flightctl %s", strings.Join(args, " "))
```
siserafin, line 1482: Normally we would move the expect validations to the test, but in this case I like better your solution, since the test is more readable... @amalykhi what do you think?

test/e2e/cli/cli_test.go
```diff
@@ -976,6 +980,275 @@ var _ = Describe("cli login", func() {
...
+		defer util.CleanupTempYAMLFile(csrYAML)
+		Eventually(func() error {
+			_, err := harness.CLI(applyOperation, "-f", csrYAML)
+			return err
+		}).Should(BeNil(), "failed to apply CSR")
+		csr := harness.GetCertificateSigningRequestByYaml(csrYAML)
+		csrName = *csr.Metadata.Name
+		Expect(csrName).NotTo(BeEmpty())
+
+		By("Autocompletion: Scenario A (get <resource>/<prefix>)")
```
siserafin: Could be nice to change some of these get \<resource\>/\<prefix\> to get \<resource\> \<prefix\>, since this notation should also support validation,

test/e2e/cli/cli_test.go
```diff
@@ -976,6 +980,275 @@ var _ = Describe("cli login", func() {
...
+		csrYAML, err := util.CreateUniqueYAMLFile("csr.yaml", harness.GetTestIDFromContext())
+		Expect(err).ToNot(HaveOccurred())
+		defer util.CleanupTempYAMLFile(csrYAML)
+		Eventually(func() error {
+			_, err := harness.CLI(applyOperation, "-f", csrYAML)
+			return err
+		}).Should(BeNil(), "failed to apply CSR")
+		csr := harness.GetCertificateSigningRequestByYaml(csrYAML)
+		csrName = *csr.Metadata.Name
+		Expect(csrName).NotTo(BeEmpty())
```
siserafin, line 1202: I think this comment by coderabbit makes sense, I don't see where you are extracting the resources names, except for csr

siserafin (commented):
Looks good, added a few comments on the new part.

test/e2e/cli/cli_test.go
eldar101: https://issues.redhat.com/browse/EDM-2608

test/e2e/cli/cli_test.go
eldar101: removed.

test/e2e/cli/cli_test.go
eldar101: Thank you. I will keep whatever you prefer.

test/e2e/cli/cli_test.go
eldar101: Done, added for each scenario

siserafin (approved):
lgtm

PR #1822, opened 2025-10-21
EDM-2385:Redis integration test race  condition fix

amalykhi:
## Summary by CodeRabbit

* **Tests**
  * Replaced fixed sleeps with Redis-polling helpers to reduce flakiness and enable deterministic synchronization.
  * Introduced cancellable contexts per consumer for explicit lifecycle control (start/stop) and coordinated multi-consumer teardown.
  * Hardened cleanup ordering and added polling around backoff, retry, checkpoint progression, in-flight tracking, failed/timed-out processing.
  * No public API changes.


DakCrowder (approved):
Think its worth merging + monitoring.  We could add more of a polling based approach w/ timeout to query redis state rather than sleeps, but would be more refactoring so if this gets the job done :heavy_check_mark:

amalykhi:
@DakCrowder I added  Fresh Consumer for Retry Phase to avoid Redis error "failed to read from stream: redis: client is closed". Please, reapprove

amalykhi:
@DakCrowder on the fina TR another test failed so I applied the wait for actual Redis state changes, I'll rerun the integration tests multiple times .
Please, review

PR #1838, opened 2025-10-22
EDM-2396: Fix build source tag detection

keitwb:
If there is one or more tags at the current commit, use the highest tag (as determined by git's tag version sorting with suffix specified).

If there is no tag at the HEAD commit, use 'git describe' to generate an artificial tag.

## Summary by CodeRabbit

* **Chores**
  * Improved build/version detection: tag resolution now prefers an exact tag at HEAD and falls back to the previous describe-based method when none found.
  * Unified version sourcing across build, CI and packaging steps to produce more consistent version strings for non-tag builds and published images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ GOARCH := $(shell go env GOARCH)
 
 VERBOSE ?= false
 
-SOURCE_GIT_TAG ?=$(shell git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
+SOURCE_GIT_TAG ?=$(shell ./hack/current-version)
 SOURCE_GIT_TREE_STATE ?=$(shell ( ( [ ! -d ".git/" ] || git diff --quiet ) && echo 'clean' ) || echo 'dirty')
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -40,7 +40,13 @@ GOARCH := $(shell go env GOARCH)
 
 VERBOSE ?= false
 
-SOURCE_GIT_TAG ?=$(shell git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
+# Prefer proper release tags pointing at this exact commit first, prioritizing the highest version
```
kkyrazis: Should  .packits get-current-version and the publish containers workflow be updated too?

https://github.com/flightctl/flightctl/blob/main/.packit.yaml#L15
https://github.com/flightctl/flightctl/blob/main/.github/workflows/publish-containers.yaml#L114

keitwb: > Should .packits get-current-version and the publish containers workflow be updated too?
> 
> https://github.com/flightctl/flightctl/blob/main/.packit.yaml#L15 https://github.com/flightctl/flightctl/blob/main/.github/workflows/publish-containers.yaml#L114

Yes, definitely. I made it a common script that is used in all three places now.

PR #1841, opened 2025-10-23
Fix for test VM Agent behavior status Device status tests 

itzikb-redhat:
In case of use a non existing image WaitForDeviceContents panics. For now the check is removed. A follow up patch will add a better error for WaitForDeviceContents and UpdateDeviceWithRetries

## Summary by CodeRabbit

* **Tests**
  * Updated an end-to-end VM agent test to stop waiting for a new image generation and proceed directly with the device update, verifying displayed device version handling without synchronizing on generation availability.

### DIFF
--- a/test/e2e/agent/agent_test.go
+++ b/test/e2e/agent/agent_test.go
@@ -222,8 +222,6 @@ var _ = Describe("VM Agent behavior", func() {
 
 			By("should report the correct device status when trying to upgrade to a not existing image")
 			previousRenderedVersion := newRenderedVersion
-			nextGeneration, err := harness.PrepareNextDeviceGeneration(deviceId)
-			Expect(err).ToNot(HaveOccurred())
 			newRenderedVersion, err = harness.PrepareNextDeviceVersion(deviceId)
 			Expect(err).ToNot(HaveOccurred())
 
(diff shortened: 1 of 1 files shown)


test/e2e/agent/agent_test.go
```diff
@@ -222,7 +222,7 @@ var _ = Describe("VM Agent behavior", func() {
 
 			By("should report the correct device status when trying to upgrade to a not existing image")
 			previousRenderedVersion := newRenderedVersion
-			nextGeneration, err := harness.PrepareNextDeviceGeneration(deviceId)
+			_, err = harness.PrepareNextDeviceGeneration(deviceId)
```
siserafin: Can we just remove this and the following line? Since we are not checking the Generation.

itzikb-redhat: Isn't it used in WaitForDeviceContents?

itzikb-redhat:
I'll rebase and test on both

test/e2e/agent/agent_test.go
siserafin: But here you are not saving the result of the output, I see "_"

itzikb-redhat: right 🤦

itzikb-redhat:
Tested

PR #1844, opened 2025-10-23
Increase timeout waiting for test vm

itzikb-redhat:
It may happen that 30 seconds are not enough

## Summary by CodeRabbit

* **Tests**
  * Increased the VM startup timeout from 30 seconds to 60 seconds; automated tests will wait longer during VM provisioning to reduce intermittent failures caused by slow startups.


itzikb-redhat:
Deciding to abandon. no need.

PR #1496, opened 2025-08-06
NO-ISSUE: reverting EDM-1871

noga-magen:
Since the commit 1a21ea8e7b73dc2c0945d856758e5f195c6add68 
had caused this [bug](https://issues.redhat.com/browse/EDM-1935)
I'm reverting it . I'll contact Shaun (who opened the og problem) and together we will think of a solution. 

## Summary by CodeRabbit

* **Chores**
  * Updated the cleanup job to use a different container image and command set for managing persistent volume claims.
  * Renamed configuration keys related to the cleanup job for consistency.
  * Adjusted command-line options for improved compatibility and reliability.

PR #1851, opened 2025-10-24
EDM-2195: Use k8s job to create secrets. Do not render cleanup/db

hexfusion:
This PR is a manual cherrypick of  #1496 and #1701. Both of these PRs were merged into main and backported to 0.9 but NOT 0.10. Because they are considering shipping 0.10 in ACM 2.15 we are cleaning up this gap.


hexfusion:
some rebase issues need to rework this

PR #1852, opened 2025-10-24
EDM-2195: Use k8s job to create secrets. Do not render cleanup/db

rawagner:
This PR is a manual cherrypick of https://github.com/flightctl/flightctl/pull/1496 and https://github.com/flightctl/flightctl/pull/1701. Both of these PRs were merged into main and backported to 0.9 but NOT 0.10. Because they are considering shipping 0.10 in ACM 2.15 we are cleaning up this gap.


siserafin:
Tested in OCP+ACM. I could deploy and login to edge manager. All the secrets were created. LGTM.

PR #1855, opened 2025-10-24
EDM-2393: Added quadlet spec definition

kkyrazis:
* Adds initial definition for Quadlet application type
* Adds ArtifactProvider as a separate type from ImageProvider
* Adds validation to reject Quadlet types from all flows (API, and Device) that will be removed as support is added.
* **Note** adds enforcement for `appType` in the `ArtifactProvider` api validation flow

## Summary by CodeRabbit

* **New Features**
  * Added "quadlet" as a new application type in the API.

* **Improvements**
  * Validation updated to prevent "quadlet" apps from being used with inline application providers.
  * Image-based application provider now rejects unsupported app types (including "quadlet") to ensure compatibility.

### DIFF
--- a/cmd/devicesimulator/main.go
+++ b/cmd/devicesimulator/main.go
@@ -647,10 +647,10 @@ func createSimulatorFleet(ctx context.Context, serviceClient *apiClient.ClientWi
 		return fmt.Errorf("creating fleet: %w", err)
 	}
 
-	if createResponse.HTTPResponse != nil && createResponse.HTTPResponse.StatusCode >= 200 && createResponse.HTTPResponse.StatusCode < 300 {
+	if createResponse.StatusCode() >= 200 && createResponse.StatusCode() < 300 {
 		log.Infof("Successfully created fleet: %s", fleetName)
 		return nil
 	}
 
-	return fmt.Errorf("failed to create fleet: status %d, body: %s", createResponse.HTTPResponse.StatusCode, string(createResponse.Body))
+	return fmt.Errorf("failed to create fleet: status %d, body: %s", createResponse.StatusCode(), string(createResponse.Body))
 }
--- a/internal/agent/device/hook/actions.go
+++ b/internal/agent/device/hook/actions.go
@@ -62,7 +62,13 @@ func newActionContext(hook api.DeviceLifecycleHookType, current *api.DeviceSpec,
 	}
 	resetCommandLineVars(actionContext)
 	if current != nil || desired != nil {
-		computeFileDiff(actionContext, current, desired)
+		defaultIfNil := func(spec *api.DeviceSpec) *api.DeviceSpec {
+			if spec == nil {
+				return &api.DeviceSpec{}
+			}
+			return spec
+		}
+		computeFileDiff(actionContext, defaultIfNil(current), defaultIfNil(desired))
 	}
 	return actionContext
 }
--- a/internal/api_server/server.go
+++ b/internal/api_server/server.go
@@ -129,7 +129,7 @@ func oapiMultiErrorHandler(errs openapi3.MultiError) (int, error) {
 		}
 	}
 
-	if allMatchFormat && longestPathErrorIndex >= 0 {
+	if allMatchFormat && longestPathError != nil {
 		shortErrorMessages = append(shortErrorMessages[:longestPathErrorIndex], shortErrorMessages[longestPathErrorIndex+1:]...)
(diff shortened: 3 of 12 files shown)


api/v1alpha1/validation.go
```diff
@@ -908,6 +915,20 @@ func validateApplications(apps []ApplicationProviderSpec, fleetTemplate bool) []
...
+		case ArtifactApplicationProviderType:
+			_, err = app.AsArtifactApplicationProviderSpec()
+			if err != nil {
+				allErrs = append(allErrs, fmt.Errorf("invalid artifact application provider: %w", err))
+				continue
+			}
+			if app.AppType == nil {
+				allErrs = append(allErrs, fmt.Errorf("artifact application type cannot be empty"))
+			}
+			if lo.FromPtr(app.AppType) == AppTypeCompose {
```
celdrake: IIUC, artifacts will initially only be supported for quadlets.
So I'd modify the validation (and the validation message)
```suggestion
			if lo.FromPtr(app.AppType) != AppTypeQuadlet {
				allErrs = append(allErrs, fmt.Errorf("artifact application provider only supports %q application type", AppTypeQuadlet))
			}
```

internal/agent/device/applications/provider/image.go
```diff
@@ -98,6 +98,10 @@ func (p *imageProvider) Verify(ctx context.Context) error {
 		p.spec.AppType = appType
 	}
 
+	if p.spec.AppType == v1alpha1.AppTypeQuadlet {
```
celdrake: Same comment as before, for future-proof I think it's better to test for what it should be rather that what it should not be (in case we add new app types)

celdrake (commented):
Some nits regarding the comparison done to raise an error on type mismatch.

api/v1alpha1/openapi.yaml
```diff
@@ -4086,6 +4087,16 @@ components:
               description: Reference to the container image for the application package.
           required:
             - image
+    ArtifactApplicationProviderSpec:
+      type: object
+      allOf:
+        - type: object
+          properties:
+            artifact:
```
hexfusion: @fzdarsky I was thinking it would be better to be explicit here instead of trying to discover image vs artifact. Its going to make the API a lot saner to implement. Auto discovery adds some complexity. But curious your thoughts.

api/v1alpha1/validation.go
```diff
@@ -908,6 +915,21 @@ func validateApplications(apps []ApplicationProviderSpec, fleetTemplate bool) []
...
+				allErrs = append(allErrs, fmt.Errorf("invalid artifact application provider: %w", err))
+				continue
+			}
+			if app.AppType == nil {
+				allErrs = append(allErrs, fmt.Errorf("artifact application type cannot be empty"))
+			}
+			if lo.FromPtr(app.AppType) != AppTypeQuadlet {
+				allErrs = append(allErrs, fmt.Errorf("artifact application provider does not support %q application type", lo.FromPtr(app.AppType)))
+			}
+			// TODO add support for appropriate AppTypes
```
hexfusion: nit: does this todo need closed now?

kkyrazis: I've got a separate PR for turning the "artifact application provider is currently unsupported" error on line 931 into real validations.

api/v1alpha1/openapi.yaml
fzdarsky: I was assuming auto discover boils down to fetching the referenced manifest, checking whether that contains the `artifactType` field. If so, it's an artifact of the specified type (e.g. `application/vnd.cncf.helm.chart.v2`). If not, fetch the config object and check whether its `mediaType` is `application/vnd.oci.image.config.v1+json` (= container image) or not (= artifact).

Would fields other than the `artifact` or `image` field change or would the _service_ (who may not be able to fetch the manifest) do things differently depending on artifact or image? If not, I'd lean towards keeping things simple for the API user.

api/v1alpha1/validation.go
```diff
@@ -764,6 +764,10 @@ func (a InlineApplicationProviderSpec) Validate(appTypeRef *AppType, fleetTempla
 	allErrs := []error{}
 	appType := lo.FromPtr(appTypeRef)
 
+	if appType == AppTypeQuadlet {
+		return []error{fmt.Errorf("inline provider is unsupported for %q applications", appType)}
```
hexfusion, line 768: is this temporary

kkyrazis: very

PR #1861, opened 2025-10-26
EDM-2424: add missing labels

sk-ilya:
## Summary by CodeRabbit

* **Chores**
  * Enhanced Helm deployment configuration metadata for improved release tracking and resource organization.

PR #1865, opened 2025-10-27
EDM-2424: add missing labels

sk-ilya:
The outcome is similar to #1861 but the resolution is specific to 0.10


hexfusion:
holding until we have a plan for 0.10 changes

PR #1867, opened 2025-10-27
EDM-2424: randomize names of immutable resources

sk-ilya:
A workaround to resolve mutation errors when a rendered helm chart is re-applied 

## Summary by CodeRabbit

* **Chores**
  * Updated Kubernetes job resource naming in deployment templates to include timestamps and release revisions, ensuring unique identifiers during upgrades and deployments.

### DIFF
--- a/deploy/helm/flightctl/templates/db/flightctl-db-migration-job.yaml
+++ b/deploy/helm/flightctl/templates/db/flightctl-db-migration-job.yaml
@@ -1,5 +1,5 @@
 {{- $root := . -}}
-{{- $baseName := printf "flightctl-db-migration-%d" .Release.Revision -}}
+{{- $baseName := printf "flightctl-db-migration-%s-%d" (now | date "20060102-150405") .Release.Revision -}}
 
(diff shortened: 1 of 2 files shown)


deploy/helm/flightctl/templates/db/flightctl-db-migration-job.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $root := . -}}
-{{- $baseName := printf "flightctl-db-migration-%d" .Release.Revision -}}
+{{- $baseName := printf "flightctl-db-migration-%s-%d" (now | date "20060102-150405") .Release.Revision -}}
```
hexfusion, line 2: only thing we need to consider is how this will work with ACM. Today they use a static definition by name

https://github.com/cameronmwall/multicloudhub-operator/blob/0673dc0f718aa9ca10741c5027cee84dd72155e3/hack/bundle-automation/charts-config.yaml#L39

hexfusion (changes requested):
generally think this is fine but lets verify ACM pattern with @cameronmwall first

PR #1868, opened 2025-10-27
EDM-2375: Fix csr example

SiddarthR56:
metadata.name not matching CSR Subject CN causing csr being set to approved but not issued as signing failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve Denied and Approved status when replacing CSRs so prior decisions are not cleared.
  * Auto-approval now respects already-decided CSRs and will not overwrite existing status conditions.

* **Documentation**
  * Updated example CSR configuration with a revised encoded request payload.


eldar101:
LGTM. I Will approe but please first read my Request here in the comments:
https://issues.redhat.com/browse/EDM-2375?filter=-2

hexfusion (approved):

eldar101 (approved):

PR #1869, opened 2025-10-27
EDM-2300: Services must-gather script

DakCrowder:
## Summary by CodeRabbit

* **New Features**
  * Added a diagnostic must-gather utility that (with confirmation and elevated privileges) gathers system info, service and systemd state, recent service logs, and Podman data into a timestamped compressed archive.

* **Chores**
  * The diagnostic utility is packaged and installed as an executable in the services package and included in release artifacts.

### DIFF
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -477,6 +477,9 @@ echo "Flightctl Observability Stack uninstalled."
     IMAGE_TAG=$(echo %{version} | tr '~' '-') \
     deploy/scripts/install.sh
 
+    # Copy services must gather script
+    cp packaging/must-gather/flightctl-services-must-gather %{buildroot}%{_bindir}
+
     # Copy sos report flightctl plugin
     mkdir -p %{buildroot}/usr/share/sosreport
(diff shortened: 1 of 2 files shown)


packaging/rpm/flightctl.spec
```diff
@@ -477,6 +477,9 @@ echo "Flightctl Observability Stack uninstalled."
     IMAGE_TAG=$(echo %{version} | tr '~' '-') \
     deploy/scripts/install.sh
 
+    # Copy services must gather script
+    cp packaging/must-gather/flightctl-services-must-gather %{buildroot}%{_bindir}
```
hexfusion, line 481: makes sense to keep the name different in the repo but should we standardize on `flightctl-must-gather`?

DakCrowder: I think the separate names might be more clear with our current setup, the %install section (which i think we can only have one of for all subpackages) of the .spec places both must-gather scripts in /usr/bin so the same file name would conflict

hexfusion: gotcha makes sense

PR #1871, opened 2025-10-27
EDM-2234: CLI Kind/Name Autocomplete

keitwb:
- Add autocomplete for the 'approve', 'decommission', 'deny', 'edit', 'get', and 'resume' commands
- It is not added for the enrollmentconfig command since that one does
  not follow the same TYPE/NAME or TYPE [NAME]... pattern as the others.
- You can complete both short and long kind names, as well as the single
  arg <kind>/<name> form, as well as the space separate two arg form. It
  also works for multiple names as with the 'get' command.
- Use a newtype for the ResourceKind so enforce some consistency on
  which version of the kind is used.
- Fix bug in editor envvar handling when VISUAL envvar is set locally

## Summary by CodeRabbit

- **New Features**
  - Shell autocomplete for resource kinds and names across commands; accepts singular/plural and both "kind/NAME" and "kind NAME" forms.
  - Commands (approve, deny, console, resume, get, edit, delete) accept relaxed argument forms (1–2 args) and improved examples/error text.

- **Refactor**
  - Centralized resource-kind handling for consistent validation, routing, pluralization, and clearer user-facing messages.

- **Tests**
  - Added autocomplete unit tests and updated tests/expected error messages to match new argument handling.

### DIFF
--- a/packaging/selinux/flightctl_agent.te
+++ b/packaging/selinux/flightctl_agent.te
@@ -178,7 +178,7 @@ systemd_start_all_unit_files(flightctl_agent_t)
 systemd_status_all_unit_files(flightctl_agent_t)
 systemd_reload_all_services(flightctl_agent_t)
 systemd_start_all_services(flightctl_agent_t)
-allow flightctl_agent_t systemd_unit_file_t:service stop;
+systemd_stop_systemd_services(flightctl_agent_t)
 allow flightctl_agent_t init_t:system status;
 
 # sshd config/service permissions


internal/cli/completion.go
```diff
@@ -150,3 +154,161 @@ func (o *CompletionOptions) Validate(args []string) error {
...
+func (kna *KindNameAutocomplete) getAutocompleteNames(cmd *cobra.Command, o ClientBuilderOptions, kind ResourceKind) []string {
+	var names []string
+	if err := o.Complete(cmd, nil); err != nil {
+		return nil
+	}
+	c, err := o.BuildClient()
+	if err == nil {
+		switch kind {
+		case DeviceKind:
+			resp, err := c.ListDevicesWithResponse(context.Background(), &api.ListDevicesParams{})
```
DakCrowder, line 227: Is it possible for labelSelectors or other filter modifiers to be taken into account for these queries?  Or what if we have more devices than would be returned from ListDevicesWithResponse?

internal/cli/apply.go
```diff
@@ -199,19 +199,20 @@ func applyFromReader(ctx context.Context, client *apiclient.ClientWithResponses,
...
 		if err != nil {
-			errs = append(errs, fmt.Errorf("%s: skipping resource of kind %q: %w", filename, kind, err))
+			errs = append(errs, fmt.Errorf("%s: skipping resource of kind %q: %w", filename, kindLike, err))
 		}
 
 		var httpResponse *http.Response
 		var message string
 
-		switch strings.ToLower(kind) {
+		kind, _ := ResourceKindFromString(kindLike)
```
hexfusion, line 214: question: why not handle error?

internal/cli/approve_test.go
```diff
@@ -10,7 +10,7 @@ func TestParseAndValidateKindNameFromArgsApprove(t *testing.T) {
 	tests := []struct {
 		name          string
 		args          []string
-		expectedKind  string
+		expectedKind  ResourceKind
```
hexfusion, line 13: :+1:

internal/cli/completion.go
```diff
@@ -150,3 +154,161 @@ func (o *CompletionOptions) Validate(args []string) error {
...
+			} else {
+				kindStrs = append(kindStrs, k.String())
+			}
+		}
+		return kindStrs, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	existingNames := args[1:]
+
+	var err error
```
hexfusion: why allocate here?

internal/cli/completion.go
```diff
@@ -150,3 +154,161 @@ func (o *CompletionOptions) Validate(args []string) error {
...
+	kind, err := ResourceKindFromString(args[0])
+	if err != nil {
```
hexfusion, line 202: should we at least log the exact error

internal/cli/completion.go
```diff
@@ -150,3 +154,161 @@ func (o *CompletionOptions) Validate(args []string) error {
...
+	})
+
+	uniqueNames = slices.DeleteFunc(uniqueNames, func(n string) bool {
+		return !strings.HasPrefix(n, toComplete)
+	})
+
+	return uniqueNames, cobra.ShellCompDirectiveNoFileComp
+}
+
+//nolint:gocyclo
```
hexfusion, line 217: can we avoid this?

internal/cli/completion.go
```diff
@@ -150,3 +154,161 @@ func (o *CompletionOptions) Validate(args []string) error {
...
+	var names []string
+	if err := o.Complete(cmd, nil); err != nil {
+		return nil
+	}
+	c, err := o.BuildClient()
+	if err == nil {
+		switch kind {
+		case DeviceKind:
+			resp, err := c.ListDevicesWithResponse(context.Background(), &api.ListDevicesParams{})
+			if err == nil && resp.JSON200 != nil {
```
hexfusion, line 228: what if err != nil?

internal/cli/completion.go
keitwb: As far as labelSelectors/filters, those would only be used when using commands that only specify the kind and not specific names of devices, right? I'm trying to think when you would need to autocomplete a specific device name but also use labels/filter flags at the same time.

If there are a very high number of devices, then autocomplete is kind of limited in what it can do. We could paginate with the continue token to get everything but I'm not sure if we want to bombard the shell with that many options. Unfortunately there is no name prefix query option in the API for us to do completion based on what the user has already typed (e.g. if they type `ab<TAB>`, we can't do an API call for only devices starting with `ab` as the name).

internal/cli/apply.go
keitwb: If this errors, the default case of the switch below it will be followed, leading to an error eventually that is handled in a common manner to other apply errors in such a way that only that particular resource fails and not everything.

internal/cli/completion.go
keitwb: Logging seems to be pretty useless in the autocomplete invocations, at least for cobra and bash. Cobra has a debug mode where you can set a `BASH_COMP_DEBUG_FILE` envvar to log interactions with the autocomplete queries but it doesn't seem to show any log output. We could log to a separate file but that seems a bit overkill for this.

internal/cli/completion.go
keitwb: I'm not sure, but I removed it.

internal/cli/completion.go
keitwb: Then we just don't return any names to the shell. The ability to give feedback through bash during autocompletion at least seems pretty limited. There is no "error" output as far as I can tell.

internal/cli/completion.go
keitwb: This comes from having each kind's request laid out in the big switch body. There doesn't appear to be any way to use Go generics to do a generic `list` operation, or any way to process the responses since they are all direct structs with no common method interface. If you know a way to do this let me know.

We could of course just put the kind-specific operations in functions and call those out from the switch but that would result in 9 additional functions that are all very similar and I'm not really confident that makes it more readable or better.

internal/cli/completion.go
DakCrowder: All makes sense!  wrt the selectors I was thinking of a scenario where a user might be searching for some device they know has a label (e.g. at some site) but can't remember the exact name but might be a misaligned UX and probably needs the name prefix completion api as you mentioned anyway.

DakCrowder (approved):

internal/cli/completion.go
hexfusion: fair thanks for the response

hexfusion (approved):

siserafin:
Is there a reason why you didn't add the autocompletion for decommission kind/name and delete kind/name?
I also noticed that enrollmentrequest is one of the resource shown by flightctl edit. I am not sure if it applies.

keitwb:
> Is there a reason why you didn't add the autocompletion for decommission kind/name and delete kind/name? I also noticed that enrollmentrequest is one of the resource shown by flightctl edit. I am not sure if it applies.

Ok, I added decommission autocomplete.

Editing ERs has always been permitted so this does not change that.

hexfusion (approved):

PR #1873, opened 2025-10-28
EDM-2404: Add debug logs to snapshot restore

siserafin:
## Summary by CodeRabbit

* **Tests**
  * Improved test harness diagnostics: adds contextual diagnostic output of agent-related files and certificates at two key steps—after snapshot creation and after snapshot revert—to simplify troubleshooting, with per-step headers and clear delimiters for easier log inspection.


itzikb-redhat:
Should agent logs be in or commented?

siserafin:
> Should agent logs be in or commented?

@itzikb-redhat I removed them, since these logs would create confusion if printed before/after each snapshot revert. I think it's enough to print the agent's files, since the device version 0 or 1 is dependent on their content.

PR #1874, opened 2025-10-28
EDM-2395: Add Inline Quadlet Spec Validation

kkyrazis:
Adds Validation for Inline Quadlet Application's

## Summary by CodeRabbit

* **New Features**
  * Added end-user support for Quadlet applications: detection, parsing, and validation of quadlet files and references.

* **Improvements**
  * Application validation now supports both Compose and Quadlet types with improved path-aware checks and clearer errors.
  * Stronger mount and image reference handling and explicit constraints for Quadlet volume definitions.

* **Tests**
  * Expanded validation test coverage for Quadlet parsing, paths, and error cases.

### DIFF
--- a/api/v1alpha1/validation.go
+++ b/api/v1alpha1/validation.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/flightctl/flightctl/internal/api/common"
+	"github.com/flightctl/flightctl/internal/quadlet"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/internal/util/validation"
 	"github.com/robfig/cron/v3"
@@ -764,10 +765,6 @@ func (a InlineApplicationProviderSpec) Validate(appTypeRef *AppType, fleetTempla
 	allErrs := []error{}
 	appType := lo.FromPtr(appTypeRef)
 
-	if appType == AppTypeQuadlet {
-		return []error{fmt.Errorf("inline provider is unsupported for %q applications", appType)}
-	}
-
 	seenPath := make(map[string]struct{}, len(a.Inline))
 	for i := range a.Inline {
 		path := a.Inline[i].Path
@@ -777,18 +774,22 @@ func (a InlineApplicationProviderSpec) Validate(appTypeRef *AppType, fleetTempla
 		} else {
 			seenPath[path] = struct{}{}
 		}
-
 		allErrs = append(allErrs, a.Inline[i].Validate(i, appType, fleetTemplate)...)
 	}
 
-	if appType == AppTypeCompose {
-		paths := make([]string, 0, len(seenPath))
-		for path := range seenPath {
-			paths = append(paths, path)
-		}
-		if err := validation.ValidateComposePaths(paths); err != nil {
-			allErrs = append(allErrs, fmt.Errorf("spec.applications[].inline[].path: %w", err))
-		}
+	paths := make([]string, 0, len(seenPath))
+	for path := range seenPath {
+		paths = append(paths, path)
+	}
+	var pathErr error
+	switch appType {
+	case AppTypeCompose:
+		pathErr = validation.ValidateComposePaths(paths)
+	case AppTypeQuadlet:
+		pathErr = validation.ValidateQuadletPaths(paths)
+	}
+	if pathErr != nil {
+		allErrs = append(allErrs, fmt.Errorf("spec.applications[].inline[].path: %w", pathErr))
 	}
 
 	return allErrs
@@ -833,7 +834,7 @@ func (c ApplicationContent) Validate(index int, appType AppType, fleetTemplate b
 	allErrs = append(allErrs, validation.ValidateString(&decodedStr, contentPath, 0, maxInlineLength, nil, "")...)
 	_, paramErrs = validateParametersInString(&decodedStr, contentPath, fleetTemplate)
 	allErrs = append(allErrs, paramErrs...)
-	allErrs = append(allErrs, ValidateApplicationContent(decodedBytes, appType)...)
+	allErrs = append(allErrs, ValidateApplicationContent(decodedBytes, appType, c.Path)...)
 
 	return allErrs
 }
@@ -846,7 +847,7 @@ func (c ApplicationContent) IsPlain() bool {
 	return c.ContentEncoding == nil || *c.ContentEncoding == EncodingPlain
 }
 
-func ValidateApplicationContent(content []byte, appType AppType) []error {
+func ValidateApplicationContent(content []byte, appType AppType, path string) []error {
 	var allErrs []error
 	switch appType {
 	case AppTypeCompose:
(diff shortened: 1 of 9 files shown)


api/v1alpha1/validation.go
```diff
@@ -906,7 +924,25 @@ func validateApplications(apps []ApplicationProviderSpec, fleetTemplate bool) []
...
+				allErrs = append(allErrs, fmt.Errorf("invalid artifact application provider: %w", err))
+				continue
+			}
+			if app.AppType == nil {
+				allErrs = append(allErrs, fmt.Errorf("artifact application type cannot be empty"))
+			}
+			if lo.FromPtr(app.AppType) == AppTypeCompose {
+				allErrs = append(allErrs, fmt.Errorf("artifact application provider does not support %q application type", AppTypeCompose))
+			}
+			allErrs = append(allErrs, fmt.Errorf("artifact application provider is currently unsupported"))
```
celdrake: Shouldn't this check be removed in this PR?

internal/util/validation/quadlet.go
```diff
@@ -0,0 +1,119 @@
...
+		if _, ok := supportedQuadletExtensions[ext]; ok {
+			if !isAtRoot(path) {
+				errs = append(errs, fmt.Errorf("quadlet file must be at root level: %q", path))
+			}
+			foundSupported = true
+			continue
+		}
+
+		if _, ok := unsupportedQuadletExtensions[ext]; ok {
+			errs = append(errs, fmt.Errorf("unsupported quadlet type %q in path: %s", ext, path))
```
celdrake, line 110:
```suggestion
			errs = append(errs, fmt.Errorf("unsupported quadlet file type %q in path: %s", ext, path))
```

internal/util/validation/quadlet.go
```diff
@@ -0,0 +1,119 @@
...
+	"errors"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/flightctl/flightctl/internal/api/common"
+)
+
+// ValidateQuadletSpec verifies the QuadletSpec for common issues
+func ValidateQuadletSpec(spec *common.QuadletSpec) []error {
```
celdrake: We should be also validating the "Header" in the files.

eg. .container files should have a `[Container]` header, etc.

internal/api/common/common.go
```diff
@@ -30,3 +32,84 @@ func ParseComposeSpec(data []byte) (*ComposeSpec, error) {
...
+
+// QuadletSpec represents a Quadlet unit file specification.
+type QuadletSpec struct {
+	Type  QuadletType `json:"type"`
+	Image *string     `json:"image,omitempty"`
+}
+
+// ParseQuadletSpec parses INI data into a QuadletSpec
+func ParseQuadletSpec(data []byte) (*QuadletSpec, error) {
+	cfg, err := ini.Load(data)
```
celdrake: This seems to be validating any "pass-through" file as if they're INI files, therefore validations will fail depending on the content.

Works
```
      {
        path: 'test/config.json',
        content: '{"key": "value"}', 
      },
```

Fails
```
      {
        path: 'README.txt',
        content: 'This is a test README file',
        contentEncoding: 'plain',
      },
```
Error 400: Bad Request: parse quadlet spec: invalid Quadlet INI: key-value delimiter not found: This is a test README file

api/v1alpha1/validation.go
kkyrazis: This is under the Artifact Provider which I didn't add validation for. So this should still exist in this PR.

internal/api/common/common.go
kkyrazis: The ini parsing of quadlet files is correct. Attempting to parse pass through files is incorrect.

api/v1alpha1/validation.go
```diff
@@ -989,6 +1026,21 @@ func ensureAppName(app ApplicationProviderSpec, appType ApplicationProviderType)
...
 
+		return *app.Name, nil
+	case ArtifactApplicationProviderType:
+		provider, err := app.AsArtifactApplicationProviderSpec()
+		if err != nil {
+			return "", fmt.Errorf("invalid artifact application provider: %w", err)
+		}
+
+		// default name to artifact if not provided
+		if app.Name == nil {
```
hexfusion: wondering if we should continue this pattern or just make this required for artifacts

internal/agent/device/applications/provider/image.go
```diff
@@ -98,6 +98,10 @@ func (p *imageProvider) Verify(ctx context.Context) error {
 		p.spec.AppType = appType
 	}
 
+	if p.spec.AppType == v1alpha1.AppTypeQuadlet {
```
hexfusion: this is still up in the air but I am leaning on not supporting image here

api/v1alpha1/validation.go
```diff
@@ -855,6 +856,16 @@ func ValidateApplicationContent(content []byte, appType AppType) []error {
 			return []error{fmt.Errorf("parse compose spec: %w", err)}
 		}
 		allErrs = append(allErrs, validation.ValidateComposeSpec(composeSpec)...)
+	case AppTypeQuadlet:
+		// Quadlet apps can come with misc files, so only validate that the quadlet files are defined correctly
```
hexfusion: question: what is the expectation for these misc files. can you give an example? lets make sure docs are eventually solid around this.

api/v1alpha1/validation.go
```diff
@@ -913,7 +924,11 @@ func validateApplications(apps []ApplicationProviderSpec, fleetTemplate bool) []
 				allErrs = append(allErrs, fmt.Errorf("inline application type cannot be empty"))
 			}
 			allErrs = append(allErrs, provider.Validate(app.AppType, fleetTemplate)...)
-			volumes = provider.Volumes
+			if len(lo.FromPtr(provider.Volumes)) > 0 && lo.FromPtr(app.AppType) == AppTypeQuadlet {
+				allErrs = append(allErrs, fmt.Errorf("quadlet application volumes should be defined as quadlets"))
```
hexfusion, line 928: question: could we  translate an application volume to something.volume? i could see an images based quadlet that the user wants to overload the volume image as an example.

internal/quadlet/quadlet.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, err
+	}
+	if len(records) != 1 {
+		return nil, fmt.Errorf("invalid mount format")
+	}
+	return records[0], nil
+}
+
+// MountType returns the type (volume, image, .etc) associated with the mount
+// see https://github.com/containers/podman/blob/main/pkg/specgenutilexternal/mount.go
```
hexfusion, line 110: :+1:

internal/quadlet/unit.go
```diff
@@ -0,0 +1,105 @@
...
+// ErrSectionNotFound is returned when a requested section is not found in a unit file.
+var ErrSectionNotFound = errors.New("section not found")
+
+// Unit represents a systemd unit file with sections and key-value entries.
+type Unit struct {
+	sections []*unit.UnitSection
+}
+
+// NewUnitFromReader creates a new Unit by deserializing systemd unit file data from a reader.
+func NewUnitFromReader(reader io.Reader) (*Unit, error) {
```
hexfusion, line 24: cool

api/v1alpha1/validation.go
hexfusion: could be a followup

hexfusion: I guess that is sort of in general right for quadlets should we allow inline and image?

kkyrazis: As in, there is an image based app with the following structure:
```
/service.container
/service.volume
```

And the contents: 
```
[Container]
...
Volume=service.volume

[Volume]
...
Image=quay.io/some-image
```

and then as an override allow the definition of:
```
      volumes:
        - name:  service
          image:
            reference: quay.io/something-else
```

and then we do some quick rewriting to change the original Volume's image from `some-image` to `something-else`

kkyrazis: I think that would make sense for `image` providers, but for `inline` I'd argue for the caller to make that change themselves.

hexfusion: agree

hexfusion (approved):

PR #1881, opened 2025-10-29
EDM-2455: Add quadlet installation logic

kkyrazis:
Adds `InstallQuadlet` which  applies the following transformations to the quadlet application:

1. Namespacing of quadlets. This is not a native feature of quadlets, and is pretty invasive but the rules are defined as following:
   1. All quadlet files (`.container`, `.volume`, `.image`, `.network`, `.pod`) are renamed to `<app-id>-<quadletfile>`
   2. All references to quadlet files inside quadlet files are rewritten to match the new values (e.g. `Image=xyz.image` => `Image=appid-xyz.image`)
   3. Drop-in directories are renamed to adhere to the namespace
   4. .conf files within drop-in directories that reference quadlets have their quadlet references updated.
5. Ensures the specified `.env` file is referenced by all containers in an application
6. Adds a Label to all Quadlet files to to allow processing any podman events by application

Note: Adds a dependency on podman/containers v5.4.2 for parsing unit files. There are later versions of podman available, but anything above that version increases our required go version. v5.4.2 has enough functionality for what we're trying to accomplish. 

## Summary by CodeRabbit

* **New Features**
  * Podman quadlet support: automated quadlet installation with per-app namespacing, reference rewriting, and per-app override drop-ins.

* **Improvements**
  * More reliable application identification from Podman events, improving event handling and logging.

* **Tests**
  * Extensive unit tests covering quadlet namespacing, reference updates, mount/volume transformations, drop-ins, and multi-file scenarios.

* **Chores**
  * Dependency declaration updated.


go.mod
```diff
@@ -104,6 +104,7 @@ require (
 )
 
 require (
+	github.com/containers/podman/v5 v5.4.2
 	github.com/goccy/go-yaml v1.18.0
```
hexfusion, line 108: we need to be careful here how large is the podman dependency and is it required?

internal/agent/client/quadlet.go
```diff
@@ -0,0 +1,508 @@
...
+	QuadletImageExtension:     quadlet.ImageGroup,
+	QuadletPodExtension:       quadlet.PodGroup,
+}
+
+// InstallQuadlet prepares Podman quadlet files for use with flightctl by applying namespacing,
+// updating cross-references, and adding flightctl-specific overrides.
+//
+// This function performs three main operations on quadlet files located at the specified path:
+//
+//  1. NAMESPACING: Renames quadlet files and drop-in directories to include the appID as a prefix.
```
hexfusion: question: should we seperate application vs non application logic from the client?

internal/agent/client/quadlet.go
```diff
@@ -0,0 +1,508 @@
+package client
+
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/podman/v5/pkg/systemd/parser"
+	"github.com/containers/podman/v5/pkg/systemd/quadlet"
```
hexfusion: are these deps needed? is seems like this is mostly using basic types?

go.mod
kkyrazis: With the packages being used, the binary size doesn't increase much. However, it does make it too easy to reference the bindings package which definitely does increase the binary size.

There's a couple of functions on the units that are quadlet specific that I am using in other areas that I would like to keep, but there's no reason we couldn't replicate the logic ourselves.

internal/agent/client/quadlet.go
kkyrazis: In this use case it's simple and the logic was previously implemented with the coreos package. I swapped to this package because its `parser` has a few Quadlet specific functions that simplify some of the logic later. This is logic that we could maintain ourselves, but  I would prefer not to if possible.

hexfusion: OK lets just try to validate the size of the deps I think it pulls in podman

kkyrazis: Current Main:
```
ls -s ./bin/ | grep flightctl-agent
23280 flightctl-agent
```

Main + this PR with a reference to client.QuadletInstall  added here https://github.com/flightctl/flightctl/blob/main/internal/agent/device/applications/manager.go#L52 (close enough to where it would be)

```
ls -s ./bin/ | grep flightctl-agent
23332 flightctl-agent
```

internal/agent/client/quadlet.go
kkyrazis: Moved this logic into `provider` so that `install` can occur at the application level.

go.mod
hexfusion: Can we explore before merge please

kkyrazis: The code that would need to be replicated would be: https://github.com/containers/podman/blob/main/pkg/systemd/parser/unitfile.go#L952 and then the merging logic to apply the overrides. It is when I got to this logic that I decided to swap from https://github.com/coreos/go-systemd/tree/main/unit to podman. As I said, it's possible to add (I had it in a previous commit), but it felt like a lot when the podman package was right there. If we are strongly against this dependency, that's fine and I'll bring back the other version.

kkyrazis: Updated to only use the coreos/go-systemd

internal/agent/device/applications/provider/quadlet.go
```diff
@@ -0,0 +1,508 @@
...
+//	    myapp-web.container      (namespaced, references updated)
+//	    myapp-data.volume        (namespaced)
+//	    myapp-web.container.d/
+//	      10-custom.conf         (references updated)
+//	    myapp-.container.d/
+//	      99-flightctl.conf      (flightctl overrides)
+//	    myapp-.volume.d/
+//	      99-flightctl.conf      (flightctl overrides)
+//	    .env                     (preserved as-is)
+func installQuadlet(readWriter fileio.ReadWriter, path string, appID string) error {
```
keitwb, line 87: Is this supposed to be `InstallQuadlet` or `installQuadlet`?

kkyrazis: `installQuadlet`. Originally it needed to be exported, but I've since migrated it to a package that won't require it to be exported. I'll update the documentation.

internal/agent/device/applications/provider/quadlet.go
```diff
@@ -0,0 +1,508 @@
...
+			mountType = val
+		}
+
+		if key == "source" && (mountType == "volume" || mountType == "image") {
+			parts[i] = fmt.Sprintf("%s=%s", key, prefixQuadletReference(val, appID))
+		}
+	}
+
+	return strings.Join(parts, ",")
+}
```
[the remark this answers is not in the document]
DakCrowder: @kkyrazis is the rabbit right about this?  Idk if the 2 step approach is necessary (could put keys/values in a map or something) but think the ordering constraint it raises is valid if i understand things correctly.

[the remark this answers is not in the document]
kkyrazis: The rabbit is correct. There's a more robust solution in podman's systemd pkg that I'll bring over to handle this.

hexfusion (approved):

PR #1882, opened 2025-10-29
EDM-2329: Auto select org on login

SiddarthR56:
## Summary by CodeRabbit

* **New Features**
  * Login flow now automatically selects your organization when exactly one is accessible, streamlining authentication.
  * Improved feedback messages display the auto-selected organization and its display name when available.
* **Bug Fixes**
  * More reliable token validation during login and persistence of updated configuration after auto-selection.

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -376,10 +376,10 @@ func (o *LoginOptions) getAbsAuthCAFile() (string, error) {
 }
 
 // validateTokenWithServer validates the token with the server, handling TLS prompts and a single retry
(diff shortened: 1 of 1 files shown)


celdrake:
@SiddarthR56 This only solves the case when only 1 organization exists, but I think in this case there isn't really a problem because the backend will default to that organization.

The general case remains unsolved, when it's actually necessary if more than 1 org is available. I updated the Jiras to reflect this.

DakCrowder (approved):
I think we can get this in and then follow up w/ one more pr for the case @celdrake has raised

PR #1884, opened 2025-10-30
EDM-2328: Adding a selector for fetching orgs by id

noga-magen:
## Summary by CodeRabbit

* **New Features**
  * Organizations list supports a fieldSelector query parameter to filter results by resource fields (e.g., metadata.name) using =, ==, !=.

* **Chores**
  * List endpoints and clients now accept a parameters object for list options; CLI, services, transport, stores, and tests updated accordingly.
  * Storage listing applies field-selector filtering; integration tests added to verify selector-based behavior.


api/v1alpha1/openapi.yaml
```diff
@@ -3590,6 +3591,25 @@
...
+          description: A selector to restrict the list of returned objects by their fields, supporting operators like '=', '==', and '!=' (e.g., "key1=value1,key2!=value2").
+          schema:
+            type: string
+        - name: limit
+          in: query
+          description: The maximum number of results returned in the list response. The server will set the 'continue' field in the list response if more results exist. The continue value may then be specified as parameter in a subsequent query.
+          required: false
+          schema:
+            type: integer
+            format: int32
```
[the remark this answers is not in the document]
noga-magen: There are no examples for the other field selectors.
This is common practice.

internal/service/organization.go
```diff
@@ -3,27 +3,70 @@ package service
...
 	if err != nil {
 		return nil, status
 	}
+	listMeta := api.ListMeta{}
+	if params.Limit != nil && *params.Limit > 0 && len(orgs) > int(*params.Limit) {
+		last := orgs[len(orgs)-1]
+		next := store.BuildContinueString([]string{last.ID.String()}, 1)
+		listMeta.Continue = next
+		orgs = orgs[:int(*params.Limit)]
+	}
```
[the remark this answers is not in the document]
noga-magen: implemented.

internal/service/organization.go
```diff
@@ -3,27 +3,99 @@ package service
...
 	var err error
+	listParams, status := prepareListParams(params.Continue, nil, params.FieldSelector, params.Limit)
+	if status.Code != http.StatusOK {
+		return nil, status
+	}
+
 	if IsInternalRequest(ctx) {
-		orgs, err = h.listAllOrganizations(ctx)
+		orgs, err = h.store.Organization().List(ctx, *listParams)
+		status := StoreErrorToApiStatus(err, false, api.OrganizationKind, nil)
```
gshilin-sdb: nitpick: maybe to move `status` inside `if err != nil` ?

internal/service/organization.go
```diff
@@ -3,27 +3,99 @@ package service
...
+				return nil, status
+			}
+			listParams.Continue = nextParsed
+		}
+
+		orgs = filtered
 	}
 
-	status := StoreErrorToApiStatus(err, false, api.OrganizationKind, nil)
+	status = StoreErrorToApiStatus(err, false, api.OrganizationKind, nil)
```
gshilin-sdb: nitpick: move it inside the `if`

test/integration/store/organization_test.go
```diff
@@ -130,5 +131,57 @@ var _ = Describe("OrganizationStore Integration Tests", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(createdOrg).To(BeNil())
 		})
+
+		It("Should support field selector with limit and continue", func() {
+
+			u1 := uuid.MustParse("00000000-0000-0000-0000-000000000011")
```
gshilin-sdb, line 139: I'm not sure whether to do it here...
This is the code from internal/store/organization.go:
```
	if listParams.Continue != nil && len(listParams.Continue.Names) == 1 {
		if id, err := uuid.Parse(listParams.Continue.Names[0]); err == nil {
			db = db.Where("id >= ?", id)
		} else {
			db = db.Where("id >= ?", listParams.Continue.Names[0])
		}
	}
```
so ID can be not UUID (else part). Should we try it somewhere too?

noga-magen: Not really because the [model](https://github.com/flightctl/flightctl/blob/776008a480dc492bde40c170a0b096c2b37de325/internal/store/model/organization.go#L10) explicitly says that the ID must be a UUID (so we cant create an ID that is not a UUID)

internal/store/organization.go
```diff
@@ -162,8 +163,35 @@ func (s *OrganizationStore) GetByID(ctx context.Context, id uuid.UUID) (*model.O
...
 }
 
-func (s *OrganizationStore) List(ctx context.Context) ([]*model.Organization, error) {
-	db := s.getDB(ctx)
+func (s *OrganizationStore) List(ctx context.Context, listParams ListParams) ([]*model.Organization, error) {
+
+	db := s.getDB(ctx).Model(&model.Organization{})
+
+	if listParams.FieldSelector != nil {
+		resolver, err := selector.NewCompositeSelectorResolver(&model.Organization{})
```
asafbss: Why CompositeSelectorResolver in this case? The composite is meant for cross-object queries to unify field selectors. Example for context: https://github.com/flightctl/flightctl/blob/437adff968da7ad06215485aa7ea7cc194298676/internal/store/device.go#L477

internal/store/organization.go
```diff
@@ -162,8 +163,35 @@ func (s *OrganizationStore) GetByID(ctx context.Context, id uuid.UUID) (*model.O
 	return &org, nil
 }
 
-func (s *OrganizationStore) List(ctx context.Context) ([]*model.Organization, error) {
-	db := s.getDB(ctx)
+func (s *OrganizationStore) List(ctx context.Context, listParams ListParams) ([]*model.Organization, error) {
+
+	db := s.getDB(ctx).Model(&model.Organization{})
```
asafbss: Any particular reason for not using ListQuery from the genericStore here? It also provides built-in pagination support.
https://github.com/flightctl/flightctl/blob/437adff968da7ad06215485aa7ea7cc194298676/internal/store/device.go#L408

internal/cli/config.go
```diff
@@ -169,7 +170,9 @@ func (o *ConfigOptions) getOrganizationDisplayName(ctx context.Context, organiza
 		return "", fmt.Errorf("failed to create API client: %w", err)
 	}
 
-	response, err := c.ListOrganizationsWithResponse(ctx)
+	field := fmt.Sprintf("metadata.name=%s", organizationId)
```
asafbss, line 172: `for _, item := range response.JSON200.Items {`
If we’re selecting a single organizationId via the field selector, is there any reason to keep the for loop? There should only be one item returned.

internal/service/organization.go
```diff
@@ -3,27 +3,99 @@ package service
...
+		userOrgSet := make(map[string]struct{}, len(userOrgs))
+		for _, m := range userOrgs {
+			userOrgSet[m.ID.String()] = struct{}{}
+		}
+
+		filtered := make([]*model.Organization, 0)
+		seen := make(map[string]struct{})
+
+		for {
+			selectedOrgs, err := h.store.Organization().List(ctx, *listParams)
```
asafbss: I’m not in favor of this pagination intervention. @hexfusion , @fzdarsky , @avishayt  — what do you think?
Could we just return the filtered organizations, even if all of them are filtered, and let the UI handle it on its side (similar to the decommissioned device case)?
Do we actually need pagination for organizationId? How many organizations are we realistically expecting in practice?

hexfusion: Seems like we could add if we had an ask as a followup.

internal/cli/get.go
```diff
@@ -542,7 +542,12 @@ func (o *GetOptions) getResourceList(ctx context.Context, c *apiclient.ClientWit
...
 		return c.ListFleetsWithResponse(ctx, &params)
 	case OrganizationKind:
-		return c.ListOrganizationsWithResponse(ctx)
+		params := api.ListOrganizationsParams{
+			FieldSelector: util.ToPtrWithNilDefault(o.FieldSelector),
+			Limit:         util.ToPtrWithNilDefault(o.Limit),
+			Continue:      util.ToPtrWithNilDefault(o.Continue),
+		}
+		return c.ListOrganizationsWithResponse(ctx, &params)
 	case TemplateVersionKind:
```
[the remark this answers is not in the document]
noga-magen: the users should only see the orgs in their context.
the orgs that are being returned are not filtered by the org_id in the context.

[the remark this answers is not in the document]
noga-magen: **Expected Behavior:**
When a user runs `flightctl get organizations`, they should see ALL organizations 
they are members of, not just the organization in their current context.

**How the backend works:**
Looking at `internal/service/organization.go` lines 16-87, the `ListOrganizations` 
function does NOT use the `org_id` query parameter for filtering. Instead, it:

1. Extracts `mappedIdentity` from context via `GetMappedIdentityFromContext(ctx)`
2. Gets the user's organizations from `mappedIdentity.Organizations`
3. Filters the store results to only include organizations the user is a member of

The `org_id` query parameter is used by the `ValidateOrgMembership` middleware for 
access control, not by the ListOrganizations handler for result filtering.

**Why current code should work:**
The CLI client passes `org_id` from the current context, but the backend ignores 
it for filtering purposes and returns all organizations based on the user's identity.

internal/service/organization_test.go
```diff
@@ -52,12 +53,30 @@ func setupMockStoreWithError(mockStore *TestStore, err error) {
...
+		if o.ID == id {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+func (m *mockResolver) GetUserOrganizations(ctx context.Context, identity common.Identity) ([]*model.Organization, error) {
+	return m.orgs, nil
+}
+
```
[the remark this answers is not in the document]
noga-magen: this was removed.

internal/service/organization_test.go
```diff
@@ -134,9 +153,48 @@ func TestListOrganizations_ResourceNotFoundError(t *testing.T) {
...
+	// Second page: use continue from first response, expect U3 and no continue token
+	cont := list1.Metadata.Continue
+	list2, status2 := handler.ListOrganizations(ctx, api.ListOrganizationsParams{Limit: &limit, Continue: cont})
+	require.Equal(t, api.StatusOK(), status2)
+	require.NotNil(t, list2)
+	require.Len(t, list2.Items, 1)
+	require.Nil(t, list2.Metadata.Continue)
+	require.Equal(t, u3.String(), *list2.Items[0].Metadata.Name)
+}
```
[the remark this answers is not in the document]
noga-magen: done

internal/store/organization.go
```diff
@@ -165,11 +165,36 @@ func (s *OrganizationStore) GetByID(ctx context.Context, id uuid.UUID) (*model.O
 	return &org, nil
 }
 
-func (s *OrganizationStore) List(ctx context.Context) ([]*model.Organization, error) {
-	db := s.getDB(ctx)
+func (s *OrganizationStore) List(ctx context.Context, listParams ListParams) ([]*model.Organization, error) {
```
asafbss, line 169: Okay, let's try to keep it simple at the moment.
Please verify if pagination is part of the requirements.
If not, please consider (and evaluate) the following draft:
```
func (s *OrganizationStore) List(ctx context.Context, listParams ListParams) ([]*model.Organization, error) {
	db := s.getDB(ctx)

	resolver, err := selector.SelectorFieldResolver(&model.Organization{})
	if err != nil {
		return nil, err
	}

	if listParams.FieldSelector != nil {
		q, p, err := listParams.FieldSelector.Parse(ctx, resolver)
		if err != nil {
			return nil, err
		}
		db = db.Where(q, p...)
	}

	var orgs []*model.Organization
	if err := db.Find(&orgs).Error; err != nil {
		return nil, err
	}

	return orgs, nil
}
```

internal/service/organization.go
```diff
@@ -3,10 +3,13 @@ package service
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"regexp"
```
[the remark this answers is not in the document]
noga-magen: what about line 34? 

internal/service/organization.go
```diff
@@ -70,3 +87,17 @@ func (h *ServiceHandler) listUserOrganizations(ctx context.Context) ([]*model.Or
...
+	allowedIDs := make(map[uuid.UUID]struct{})
+	matches := selectorRegex.FindAllStringSubmatch(selectorStr, -1)
+
+	for _, match := range matches {
+		if id, err := uuid.Parse(match[1]); err == nil {
+			allowedIDs[id] = struct{}{}
+		}
+	}
+	return allowedIDs
+}
```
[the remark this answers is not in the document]
noga-magen: there is no filterOrganizationsByFieldSelector function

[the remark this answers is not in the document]
noga-magen: we don't want another db query since it's the slower than searching for the org in memory

PR #1886, opened 2025-11-01
NO-ISSUE: Use enabled for db control

JasonN3:
For all other components, it uses `enabled` to determine whether to use the internal option or allow for an external/custom options. This changes the variable name to `enabled` and replaces the values of `enabled`/`disabled` with `true`/`false` so it can be used as a boolean instead of a string.

## Summary by CodeRabbit

* **Configuration**
  * Database config simplified to a single boolean (db.enabled); internal DB enabled by default. Set db.enabled=false to use an external DB.
  * Resource creation (internal DB deployment, service, PVC, ConfigMap, migration/users flow, TLS config) now follows db.enabled rather than the old external flag.

* **Documentation**
  * Deployment guides, examples and verification steps updated to use db.enabled semantics.

### DIFF
--- a/deploy/helm/flightctl/README.md
+++ b/deploy/helm/flightctl/README.md
@@ -186,7 +186,7 @@ kubectl create secret generic postgres-client-certs \
 ```yaml
 # Configure in values.yaml
 db:
-  external: "enabled"
+  enabled: false
   hostname: "postgres.example.com"
   sslmode: "verify-ca"
   sslConfigMap: "postgres-ca-cert"     # ConfigMap containing CA certificate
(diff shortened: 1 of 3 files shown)


gshilin-sdb (commented):
1. What are "all other components" that you mention in description? There is only one component in the context of FlightCtl that may be external or internal, and it's DB.
2. "enabled: false" bears a connotation of 'disabled' and not 'external'. This will be confusing for users.
3. A non-boolean "external" will permit future developments. Using an enum-like approach (e.g., internal, external, prefer-external) would be more extensible and clearer about intent.

JasonN3:
> 1. What are "all other components" that you mention in description? There is only one component in the context of FlightCtl that may be external or internal, and it's DB.

For this Helm chart: kv and ui. It's also fairly standard for Helm charts to use `enabled` to define whether the Helm chart will be deploying a component or it will leave it for you to configure externally. Look at Gitea or Netbox for examples or pretty much any other Helm chart that has components that are commonly externally configured.
>    2. "enabled: false" bears a connotation of 'disabled' and not 'external'. This will be confusing for users.

It would be disabled. It would be disabled in this Helm chart leaving you to setup whatever other one you would like outside of this Helm chart. Values is configuring what the Helm chart will do.

>    3. A non-boolean "external" will permit future developments. Using an enum-like approach (e.g., internal, external, prefer-external) would be more extensible and clearer about intent.

In what situation would you want an app to randomly change from an external database server with data in it to suddenly an empty, unconfigured database? Either you want your app configured to an external database server or you want the Helm chart to deploy and configure a database server for you. Enabled or disabled should be the only options you need. I've never come across a Helm chart that doesn't use a boolean for this.

PR #1890, opened 2025-11-02
NO-ISSUE: run backward compatibility tests against a migrated database

sk-ilya:
* Execute a N->N+1 migration for backward compatibility checks instead of just initializing a fresh N+1 DB. 
* Don't drop the `last_seen` column. 
 

## Summary by CodeRabbit

* **Chores**
  * Enhanced database migration testing with a two‑phase (initial + expand) validation path and clearer phase-aware logs.
  * Improved integration/backward‑compatibility checks to validate both baseline and expanded migrations.

* **Bug Fixes / Data**
  * Initial migration no longer removes the last_seen column, preserving existing data during upgrades.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -147,7 +147,7 @@ jobs:
       # Temporarily disabled backward compatibility tests
       # TODO: Re-enable after fixing schema migration issues
       - name: Running backward compatibility tests in template mode
-        if: ${{ steps.changed-files.outputs.any_changed == 'true' }}
+        if: ${{ github.event_name == 'pull_request' && steps.changed-files.outputs.any_changed == 'true' }}
         run: |
           make integration-test VERBOSE=true
         env:
--- a/test/test.mk
+++ b/test/test.mk
@@ -85,7 +85,7 @@ _wait_for_db:
 _run_template_migration:
 	@INITIAL_MIGRATION_IMAGE="$(INITIAL_MIGRATION_IMAGE)" EXPAND_MIGRATION_IMAGE="$(EXPAND_MIGRATION_IMAGE)" bash -euo pipefail -c '\
 	  echo "Template strategy: resolving migration strategy..."; \
-	  if [ -n "$$EXPAND_MIGRATION_IMAGE" ]; then \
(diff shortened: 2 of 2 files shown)


internal/store/device.go
```diff
@@ -349,13 +345,6 @@ func (s *DeviceStore) backfillDeviceTimestamps(db *gorm.DB) error {
 		SELECT org_id, name FROM devices WHERE (org_id, name) NOT IN (SELECT org_id, name FROM device_timestamps)`).Error
 }
 
-func (s *DeviceStore) dropLastSeenColumnIfExists(db *gorm.DB) error {
```
sk-ilya, line 352: This caused failures from old code after the migration is executed

.github/workflows/integration-tests.yaml
```diff
@@ -30,29 +30,28 @@ jobs:
...
 
           # Build the db-setup container with commit tag
           sudo make flightctl-db-setup-container
 
-          # Tag the image with the commit SHA
+          # Tag the image with commit SHA
           sudo podman tag flightctl-db-setup:latest "flightctl-db-setup:${COMMIT_SHA}"
 
-          echo "MIGRATION_IMAGE=flightctl-db-setup:${COMMIT_SHA}" >> "$GITHUB_ENV"
+          echo "EXPAND_MIGRATION_IMAGE=flightctl-db-setup:${COMMIT_SHA}" >> "$GITHUB_ENV"
```
hexfusion, line 54: why expand its not clear to me what we are expanding?

sk-ilya: DB schema expansion. It's the same migration, ran against a non empty DB

hexfusion: OK but the image is the same right?

sk-ilya: Yes

internal/store/device.go
ori-amizur: Since we are not GA yet, it should not be removed.
If we decide to keep it, then we need to decide when and how the column is dropped.

ori-amizur:
Before such change is done, we should know how contract will work.

internal/store/device.go
hexfusion: Lets define a deprecation policy for the schema? I started a draft here please comment.

ref. https://github.com/flightctl/flightctl/pull/1898

PR #1891, opened 2025-11-02
EDM-1124: (SDLC) null_method_call: Calling a method on null object

ori-amizur:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened nil-safety across multiple components to prevent crashes (fleet creation, device actions, enrollment requests, event handling, and table display).
  * Made spec comparisons treat two nil specs as equal in device, fleet, repository, resource sync, and enrollment models.
  * Improved multi-error API response logic to pick the most relevant error reliably.
  * Switched status reads to accessor methods for more robust status reporting.


internal/agent/device/hook/actions.go
```diff
@@ -62,7 +62,13 @@ func newActionContext(hook api.DeviceLifecycleHookType, current *api.DeviceSpec,
 	}
 	resetCommandLineVars(actionContext)
 	if current != nil || desired != nil {
-		computeFileDiff(actionContext, current, desired)
+		defaultIfNil := func(spec *api.DeviceSpec) *api.DeviceSpec {
```
hexfusion, line 65: note: spec nil checks happen in the priority queue[1] and all Device is initialized as non nil spec[2]. Was there something specific you saw in logs to add this?

[1] https://github.com/flightctl/flightctl/blob/437adff968da7ad06215485aa7ea7cc194298676/internal/agent/device/device_test.go#L394
[2] https://github.com/flightctl/flightctl/blob/437adff968da7ad06215485aa7ea7cc194298676/internal/agent/device/spec/queue.go#L74

ori-amizur: It is part of SDLC warnings/errors.
https://cov01.lab.eng.brq2.redhat.com/osh/task/979762/log/flightctl-0.9.2-1.el9fc/scan-results-all.html

hexfusion: interesting considering all the areas that use DeviceSpec{} but what do I know :)

hexfusion (approved):

PR #1894, opened 2025-11-03
EDM-2346: replaced keycloak with PAM issuer

asafbennatan:
a split from the full PR , 
1. removes keyclock
2.adds pam issuer that exposes an oidc interface it is deployed in its own container: flightctl-pam-issuer
3.uses k8s authentication for vanialla k8s deployment / oc and not when running make deploy

## Summary by CodeRabbit

* **New Features**
  * PAM-based OpenID Connect issuer added as the default authentication provider, including OIDC discovery, authorize, login, token and userinfo endpoints.

* **Changes**
  * Keycloak removed; default auth type now OIDC with the PAM issuer.
  * Authentication service port changed from 8081 to 8444.
  * Deployment, packaging, templates and documentation updated for PAM OIDC (Helm/Podman/quadlet/packaging).
  * New integration and unit tests for the PAM issuer.

### DIFF
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -180,6 +180,7 @@ services to be running. This package automatically includes the flightctl-teleme
 %dir /etc/grafana
 %dir /etc/grafana/provisioning
 %dir /etc/grafana/provisioning/datasources
+%dir /etc/grafana/provisioning/alerting
 %dir /etc/grafana/provisioning/dashboards
 %dir /etc/grafana/provisioning/dashboards/flightctl
 %dir /etc/grafana/certs
@@ -292,7 +293,7 @@ echo "Running post-install actions for Flightctl Observability Stack..."
 
 # Create necessary directories on the host if they don't already exist.
 /usr/bin/mkdir -p /etc/prometheus /var/lib/prometheus
-/usr/bin/mkdir -p /etc/grafana /etc/grafana/provisioning /etc/grafana/provisioning/datasources /var/lib/grafana
+/usr/bin/mkdir -p /etc/grafana /etc/grafana/provisioning /etc/grafana/provisioning/datasources /etc/grafana/provisioning/alerting /var/lib/grafana
 /usr/bin/mkdir -p /etc/grafana/provisioning/dashboards /etc/grafana/provisioning/dashboards/flightctl
 /usr/bin/mkdir -p /etc/grafana/certs
 /usr/bin/mkdir -p /etc/flightctl /opt/flightctl-observability/templates
@@ -492,6 +493,7 @@ echo "Flightctl Observability Stack uninstalled."
      mkdir -p %{buildroot}/etc/containers/systemd
      mkdir -p %{buildroot}/etc/prometheus
      mkdir -p %{buildroot}/etc/grafana/provisioning/datasources
+     mkdir -p %{buildroot}/etc/grafana/provisioning/alerting
      mkdir -p %{buildroot}/etc/grafana/provisioning/dashboards/flightctl
      mkdir -p %{buildroot}/etc/grafana/certs
      mkdir -p %{buildroot}/var/lib/prometheus
--- a/.spelling
+++ b/.spelling
@@ -208,6 +208,7 @@ timestamped
 SELinux
 dnf
 rpm.flightctl.io
+wildcards
  - docs/user/api-resources.md
 approvedBy
 creationTimestamp
--- a/docs/user/standalone-observability.md
+++ b/docs/user/standalone-observability.md
@@ -348,6 +348,7 @@ observability:
     https:
       cert_file: /etc/grafana/certs/grafana.crt
       cert_key: /etc/grafana/certs/grafana.key
+    allowed_unsigned_plugins: ""  # Optional: comma-separated list of plugin IDs to allow unsigned plugins
 ```
 
 ### Prometheus Metrics
@@ -501,6 +502,9 @@ observability:
     https:
       cert_file: /etc/grafana/certs/grafana.crt
       cert_key: /etc/grafana/certs/grafana.key
+    
+    # Plugin Configuration (Optional)
+    allowed_unsigned_plugins: ""  # Comma-separated list of plugin IDs to allow unsigned plugins
 
   # Prometheus Configuration (internal only)
   prometheus:
@@ -1085,6 +1089,17 @@ This section provides detailed documentation for every configuration variable av
 - **Default**: `http://localhost:3000`
 - **Description**: Root URL for Grafana. Used for OAuth redirects and asset loading. Should match the external URL where Grafana is accessible.
 - **Example**: `https://grafana.yourdomain.com`, `http://server-ip:3000`
+
+#### Plugin Configuration
+
+**`observability.grafana.allowed_unsigned_plugins`**
+
+- **Type**: String (comma-separated list)
+- **Default**: Empty (no unsigned plugins allowed)
+- **Description**: Comma-separated list of plugin IDs that are allowed to load even if they are not signed by Grafana Labs. This is useful for custom plugins or community plugins that haven't been signed. Use with caution as unsigned plugins can pose security risks.
+- **Security**: Only specify plugins you trust. Use specific plugin IDs rather than wildcards in production environments.
+- **Example**: `my-custom-plugin,another-plugin`
+- **Warning**: Setting this to `*` allows all unsigned plugins, which is not recommended for security reasons.
 
 ### Prometheus Configuration Variables
 
--- a/packaging/observability/flightctl-grafana.container.template
+++ b/packaging/observability/flightctl-grafana.container.template
@@ -22,6 +22,9 @@ Volume=/etc/grafana/provisioning/datasources:/etc/grafana/provisioning/datasourc
 # Mount directory for TLS certificates (if using HTTPS)
 Volume=/etc/grafana/certs:/etc/grafana/certs:ro,z
 
+# Downstream grafana image expects this
+Volume=/etc/grafana/provisioning/alerting:/etc/grafana/provisioning/alerting:ro,z
+
 Network=flightctl
 
 [Service]
--- a/packaging/observability/grafana.ini.template
+++ b/packaging/observability/grafana.ini.template
@@ -55,4 +55,5 @@ scopes = ${GRAFANA_OAUTH_SCOPES}
 role_attribute_path = (contains(roles[*], 'grafana-admin') || contains(roles[*], 'Admin')) && 'Admin' ||  (contains(roles[*], 'grafana-editor') || contains(roles[*], 'Editor')) && 'Editor' ||  'Viewer'
 tls_skip_verify_insecure = ${GRAFANA_TLS_SKIP_VERIFY}
 
-
+[plugins]
+allow_loading_unsigned_plugins = ${GRAFANA_ALLOWED_UNSIGNED_PLUGINS}
--- a/packaging/observability/observability.defs
+++ b/packaging/observability/observability.defs
@@ -16,6 +16,7 @@ GRAFANA_TLS_SKIP_VERIFY        | .observability.grafana.oauth.tls_skip_verify   
 GRAFANA_PROTOCOL               | .observability.grafana.protocol                            | http                                         | grafana.ini.template                          | /etc/grafana/grafana.ini
 GRAFANA_CERT_FILE              | .observability.grafana.https.cert_file                     | /etc/grafana/certs/grafana.crt               | grafana.ini.template                          | /etc/grafana/grafana.ini
 GRAFANA_CERT_KEY               | .observability.grafana.https.cert_key                      | /etc/grafana/certs/grafana.key               | grafana.ini.template                          | /etc/grafana/grafana.ini
+GRAFANA_ALLOWED_UNSIGNED_PLUGINS  | .observability.grafana.allowed_unsigned_plugins         |                                              | grafana.ini.template                          | /etc/grafana/grafana.ini
 
 GRAFANA_IMAGE                  | .observability.grafana.image                               | docker.io/grafana/grafana:latest             | flightctl-grafana.container.template          | /etc/containers/systemd/flightctl-grafana.container
 GRAFANA_PUBLISHED_PORT         | .observability.grafana.published_port                      | 3000                                         | flightctl-grafana.container.template          | /etc/containers/systemd/flightctl-grafana.container


packaging/rpm/flightctl.spec
```diff
@@ -11,16 +11,16 @@
 Name:           flightctl
 # Version and Release are automatically updated by Packit during build
 # Do not manually change these values - they will be overwritten
-Version:        0.6.0
-Release:        1%{?dist}
+Version:        1.0.0~main~189~gccc1a335
```
sk-ilya: The templates got expanded

asafbennatan:
@coderabitai pause

deploy/helm/flightctl/values.yaml
```diff
@@ -19,8 +19,8 @@ global:
   # -- Alternative to global.auth.k8s.externalOpenShiftApiUrl with the same meaning, used by the multiclusterhub operator
   apiUrl: "" # alternative to global.auth.k8s.externalOpenShiftApiUrl used by multiclusterhub operator
   auth:
-    # -- Type of the auth to use. Can be one of 'builtin', 'k8s', 'oidc', or 'none'
-    type: "builtin" # builtin, k8s, oidc, aap, none
+    # -- Type of the auth to use. Can be one of 'k8s', 'oidc', or 'none'
+    type: "oidc" # k8s, oidc, aap, none
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/helm/flightctl/values.yaml
```diff
@@ -35,10 +35,27 @@ global:
...
+      # -- OAuth2 client ID for the OIDC issuer
+      clientId: "flightctl-client"
+      # -- OAuth2 client secret for the OIDC issuer
+      clientSecret: ""
+      # -- Supported OAuth2 scopes
+      scopes: ["openid", "profile", "email", "roles"]
+      # -- Allowed redirect URIs for OAuth2 flows
+      redirectUris: []
+      # -- PAM service name for authentication (default: "flightctl")
+      pamService: "flightctl"
```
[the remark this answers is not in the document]
asafbennatan: this is incorrect , oidc existed before the current commit

deploy/podman/flightctl-api/flightctl-api-config/init.sh
```diff
@@ -149,20 +151,108 @@ if [ "$AUTH_TYPE" == "aap" ]; then
...
+
+  # Set defaults for PAM
+  PAM_OIDC_ISSUER_ENABLED=${PAM_OIDC_ISSUER_ENABLED:-true}
+  PAM_OIDC_ISSUER=${PAM_OIDC_ISSUER:-${BASE_URL}}
+  PAM_OIDC_CLIENT_ID=${PAM_OIDC_CLIENT_ID:-flightctl-client}
+  PAM_OIDC_SERVICE=${PAM_OIDC_SERVICE:-flightctl}
+  
+  # Set PAM OIDC issuer URL for API server to connect to
+  # This is the URL where the PAM issuer service is accessible
+  PAM_OIDC_ISSUER_URL=${PAM_OIDC_ISSUER:-"https://${BASE_DOMAIN}:8444/api/v1/auth"}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,1217 @@
...
+		}
+	}
+
+	// Check for session_id query parameter
+	if req.Scope != nil && *req.Scope != "" {
+		// Check if the scope contains a session ID (prefixed with "session:")
+		if strings.HasPrefix(*req.Scope, "session:") {
+			return strings.TrimPrefix(*req.Scope, "session:")
+		}
+	}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,1217 @@
...
+		req.RedirectURI,
+		req.State,
+		req.Username, // Using username as scope for demo
+	)
+
+	return &LoginResponse{
+		Success:   true,
+		ReturnURL: authURL,
+	}, nil
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,1217 @@
...
+            </div>
+            
+            <button type="submit">Sign In</button>
+            
+            <div class="error"></div>
+        </form>
+    </div>
+</body>
+</html>`, clientID, redirectURI, state)
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

packaging/rpm/flightctl.spec
asafbennatan: fixed

deploy/helm/flightctl/values.yaml
```diff
@@ -243,6 +255,29 @@ api:
     readinessPath: /readyz
     # -- HTTP path for liveness probe
     livenessPath: /healthz
+
+# -- PAM OIDC Issuer Configuration
+pamIssuer:
+  # -- Enable PAM OIDC Issuer service deployment
+  enabled: false
```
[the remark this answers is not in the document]
asafbennatan: no

internal/config/config.go
```diff
@@ -431,6 +450,27 @@ func Load(cfgFile string) (*Config, error) {
...
+			}
+			if c.Auth.PAMOIDCIssuer.ClientID == "" {
+				c.Auth.PAMOIDCIssuer.ClientID = "flightctl-client"
+			}
+			if len(c.Auth.PAMOIDCIssuer.Scopes) == 0 {
+				c.Auth.PAMOIDCIssuer.Scopes = []string{"openid", "profile", "email", "roles"}
+			}
+			if len(c.Auth.PAMOIDCIssuer.RedirectURIs) == 0 {
+				c.Auth.PAMOIDCIssuer.RedirectURIs = []string{c.Service.BaseUIUrl + "/auth/callback"}
+			}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,364 @@
...
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			h.log.Errorf("Failed to decode JSON request: %v", err)
+			writeError(w, http.StatusBadRequest, "Failed to decode request")
+			return
+		}
+	}
+
+	// Call PAM provider's Token method directly with pamapi types
+	tokenResponse, err := h.pamProvider.Token(r.Context(), &req)
+	if err != nil {
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/scripts/init_host.sh
```diff
@@ -27,6 +33,13 @@ main() {
     else
         echo "Base domain already set to: $base_domain"
     fi
+    
+    # Configure PAM authentication when auth type is oidc and pamOidcIssuer is enabled
+    auth_type=$(extract_value "global.auth.type" "$SERVICE_CONFIG_FILE")
+    pam_enabled=$(extract_value "global.auth.pamOidcIssuer.enabled" "$SERVICE_CONFIG_FILE")
+    
+    # PAM configuration is baked into the flightctl-pam-issuer container image
+    # No host-level PAM configuration needed
```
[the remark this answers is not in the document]
asafbennatan: fixed

docs/developer/pam-issuer-deployment.md
```diff
@@ -0,0 +1,475 @@
...
+auth:
+  oidc:
+    - clientId: flightctl-client
+      enabled: true
+      issuer: https://pam-issuer.<baseDomain>
+      # ... other OIDC configuration
+```
+
+This allows the main API server to validate tokens issued by the PAM issuer without directly depending on PAM libraries.
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/authn/jwt_generator.go
```diff
@@ -0,0 +1,468 @@
...
+		if err := jwkKey.Set(jwk.AlgorithmKey, jwa.RS256); err != nil {
+			return "", fmt.Errorf("failed to set algorithm: %w", err)
+		}
+	default:
+		// For ECDSA keys, use ES256
+		signingAlg = jwa.ES256
+		if err := jwkKey.Set(jwk.AlgorithmKey, jwa.ES256); err != nil {
+			return "", fmt.Errorf("failed to set algorithm: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,949 @@
...
+	// If clientSecret is configured (non-empty), this is a confidential client and we require authentication
+	// If clientSecret is empty, this is a public client (CLI, SPA) and we don't require secret (should use PKCE in production)
+	if s.config.ClientSecret != "" {
+		// Confidential client - require client_secret_post authentication
+		if req.ClientSecret == nil || *req.ClientSecret != s.config.ClientSecret {
+			return &pamapi.TokenResponse{Error: lo.ToPtr("invalid_client")}, nil
+		}
+	}
+	// For public clients (empty secret), we accept the request without secret validation
+	// In production, this should be combined with PKCE (code_challenge/code_verifier) for security
```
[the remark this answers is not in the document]
asafbennatan: currently we need this for ui and we dont have a BFF

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,949 @@
...
+	s.CreateUserSession(sessionID, username, clientID, redirectURI, state)
+	s.log.Infof("Login: created user session for %s", username)
+
+	// Redirect back to authorization endpoint with session ID
+	authURL := fmt.Sprintf("/api/v1/auth/authorize?response_type=code&client_id=%s&redirect_uri=%s", clientID, redirectURI)
+	if state != "" {
+		authURL += fmt.Sprintf("&state=%s", state)
+	}
+	authURL += fmt.Sprintf("&session_id=%s", sessionID)
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,369 @@
...
+
+	// Start cleanup goroutines
+	go func() {
+		ticker := time.NewTicker(5 * time.Minute)
+		defer ticker.Stop()
+		for range ticker.C {
+			pamProvider.CleanupExpiredCodes()
+			pamProvider.CleanupExpiredSessions()
+		}
+	}()
```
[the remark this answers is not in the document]
asafbennatan: fixed

asafbennatan:
@coderabbitai pause

Makefile
```diff
@@ -154,7 +154,7 @@ build: bin build-cli
 		./cmd/flightctl-userinfo-proxy \
 		./cmd/flightctl-db-migrate \
 		./cmd/flightctl-restore \
-		./cmd/flightctl-telemetry-gateway
+		./cmd/flightctl-telemetry-gateway \
```
hexfusion: is that change correct?

api/v1alpha1/pam-issuer/docs.go
```diff
@@ -0,0 +1,5 @@
+// Package pam_issuer provides primitives to interact with the PAM Issuer openapi HTTP API.
+package pam_issuer
```
hexfusion, line 2: It isnt common to name packages snake case I know we have done this is this project before but I would advise breaking the pattern. Also since this is an OIDC provider is this a pam provider?

ref. https://github.com/kubernetes/kubernetes/tree/master/pkg

```
go list std
archive/tar
archive/zip
bufio
bytes
cmp
[...]
```

api/v1alpha1/pam-issuer/openapi.yaml
```diff
@@ -0,0 +1,473 @@
+openapi: 3.0.1
+info:
+  title: Flight Control PAM Issuer API
```
hexfusion:
```suggestion
  title: Flight Control PAM OIDC Provider API
```

api/v1alpha1/pam-issuer/types.gen.cfg
```diff
@@ -0,0 +1,8 @@
+package: pam_issuer
+generate:
+  models: true
+output: types.gen.go
+output-options:
+  skip-prune: true
+
+
```
hexfusion: minor: extra newline

cmd/flightctl-pam-issuer/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/flightctl/flightctl/internal/api_server/middleware"
+	"github.com/flightctl/flightctl/internal/config"
+	"github.com/flightctl/flightctl/internal/crypto"
+	"github.com/flightctl/flightctl/internal/instrumentation"
+	"github.com/flightctl/flightctl/internal/pam_issuer_server"
```
hexfusion, line 16: food for thought
```suggestion
	"github.com/flightctl/flightctl/internal/pam"
```


```go
   server, err := pam.NewServer()
   
   whatever, err := pam.NewWhatever()
```

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -1,4 +1,3 @@
-{{ if eq .Values.global.target "acm" }}
```
hexfusion, line 1: walk me through this change a little. why is this not gated for acm env anymore?

internal/auth/issuer/pam/oidc_issuer.go
```diff
@@ -0,0 +1,45 @@
+package pam
```
hexfusion, line 1: a little echoing in the namespace would something like below be more clear?

```
internal/auth/oidc/pam/provider.go
```

internal/auth/issuer/pam/pam_authenticator.go
```diff
@@ -0,0 +1,107 @@
+//go:build linux
+
+package pam
```
hexfusion, line 3: I know I harp on cosmetics but it does help in keeping things clean over time. since we are in the pam package it is assumed.

```
internal/auth/issuer/pam/authenticator.go

File: internal/auth/issuer/pam/pam_authenticator.go [hunk 25]
```diff
@@ -0,0 +1,107 @@
...
+	"os/user"
+
+	"github.com/msteinert/pam/v2"
+)
+
+// RealPAMAuthenticator implements Linux authentication using PAM and NSS
+// PAM (Pluggable Authentication Modules) handles authentication
+// NSS (Name Service Switch) handles user/group lookups via user.Lookup()
+// Works with any system-configured authentication backend
+type RealPAMAuthenticator struct {
```
hexfusion: Real vs Mock? Is that designation required or is there another reason for the naming.

internal/auth/issuer/pam/pam_authenticator.go
```diff
@@ -0,0 +1,107 @@
...
+	pamService string
+}
+
+// NewRealPAMAuthenticator creates a new Linux authenticator
+// Uses PAM for authentication and NSS for user/group information
+// Automatically works with any configured Linux authentication backend
+func NewRealPAMAuthenticator() (*RealPAMAuthenticator, error) {
```
hexfusion:
```suggestion
func NewAuthenticator() (*Authenticator, error) {

// usage
auth, err := pam.NewAuthenticator()
```

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	Scope       string
+	State       string
+	Username    string
+	ExpiresAt   time.Time
+	CreatedAt   time.Time
+}
+
+// AuthorizationCodeStore manages temporary authorization codes
+type AuthorizationCodeStore struct {
+	codes sync.Map
```
hexfusion, line 51: sync.Map is pretty specialized. In this context if the struct will never grow it could make sense. But in general I would advise what the docs say. 

```
The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content. 
```
ref. https://pkg.go.dev/sync#Map

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	s.codes.Delete(code)
+	return codeData, true
+}
+
+// CleanupExpiredCodes removes expired codes
+func (s *AuthorizationCodeStore) CleanupExpiredCodes() {
+	now := time.Now()
+	s.codes.Range(func(key, value interface{}) bool {
+		code := key.(string)
+		codeData := value.(*AuthorizationCodeData)
```
hexfusion: nit: interface param should have type check

```suggestion
		codeData, ok := value.(*AuthorizationCodeData)
        if !ok {
           // handle else panic
```

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+
+	// Validate the refresh token and ensure it's actually a refresh token
+	identity, err := s.jwtGenerator.ValidateTokenWithType(*req.RefreshToken, "refresh_token")
+	if err != nil {
+		return &pamapi.TokenResponse{Error: lo.ToPtr("invalid_grant")}, nil
+	}
+
+	// Get current user information from NSS to ensure user still exists
+	systemUser, err := s.pamAuthenticator.LookupUser(identity.GetUsername())
+	if err != nil {
```
hexfusion, line 192: question: would printing the actual error be useful for debug?

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	tokenResponse := &pamapi.TokenResponse{
+		AccessToken: lo.ToPtr(accessToken),
+		TokenType:   lo.ToPtr(pamapi.Bearer),
+		ExpiresIn:   lo.ToPtr(int(time.Hour.Seconds())),
+	}
+
+	// Always issue a new refresh token when using refresh_token grant
+	// (if we have a refresh token, it means offline_access was originally granted)
+	refreshToken, err := s.jwtGenerator.GenerateTokenWithType(tokenGenerationRequest, 7*24*time.Hour, "refresh_token")
+	if err != nil {
```
hexfusion, line 234: question: would printing the actual error be useful for debug?

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+		return &pamapi.TokenResponse{Error: lo.ToPtr("server_error")}, nil
+	}
+	tokenResponse.RefreshToken = lo.ToPtr(refreshToken)
+
+	return tokenResponse, nil
+}
+
+// handleAuthorizationCodeGrant handles the authorization_code grant type
+func (s *PAMOIDCProvider) handleAuthorizationCodeGrant(ctx context.Context, req *pamapi.TokenRequest) (*pamapi.TokenResponse, error) {
```
hexfusion, line 244: is it ever possible req is nil? If we follow golangs error patterns my expectation is no.

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+
+	// Create token response
+	tokenResponse := &pamapi.TokenResponse{
+		AccessToken: lo.ToPtr(accessToken),
+		TokenType:   lo.ToPtr(pamapi.Bearer),
+		ExpiresIn:   lo.ToPtr(int(time.Hour.Seconds())),
+	}
+
+	// Only generate refresh token if offline_access was requested
+	if strings.Contains(codeData.Scope, "offline_access") {
```
hexfusion: `offline_access`  and friends should be consts?

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+		}
+		tokenResponse.RefreshToken = lo.ToPtr(refreshToken)
+	}
+
+	return tokenResponse, nil
+}
+
+// Authorize handles the authorization endpoint for authorization code flow
+func (s *PAMOIDCProvider) Authorize(ctx context.Context, req *pamapi.AuthAuthorizeParams) (*AuthorizeResponse, error) {
+	s.log.Infof("Authorize: starting authorization request - ClientId=%s, RedirectUri=%s, ResponseType=%s, Scope=%s, State=%s",
```
hexfusion: question: how chatty is this should it be debug?

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	// Validate client ID
+	if s.config == nil {
+		s.log.Warnf("Authorize: config is nil")
+		return nil, fmt.Errorf("invalid_client")
+	}
+	if s.config.ClientID != req.ClientId {
+		s.log.Warnf("Authorize: invalid client ID - expected=%s, got=%s", s.config.ClientID, req.ClientId)
+		return nil, fmt.Errorf("invalid_client")
+	}
+	s.log.Infof("Authorize: client ID validation passed - %s", req.ClientId)
```
hexfusion: same question here probably for all info logs

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	}
+
+	// User is authenticated, get username from session
+	username := sessionData.Username
+	s.log.Infof("Authorize: user authenticated via session - username=%s, sessionID=%s", username, sessionID)
+
+	// Generate authorization code
+	authCode, err := generateAuthorizationCode()
+	if err != nil {
+		s.log.Errorf("Authorize: failed to generate authorization code - %v", err)
```
hexfusion, line 450: :+1:

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	s.log.Infof("Authorize: returning redirect URL - %s", redirectURL)
+	return &AuthorizeResponse{
+		Type:    AuthorizeResponseTypeRedirect,
+		Content: redirectURL,
+	}, nil
+}
+
+// Login handles the login form submission
+func (s *PAMOIDCProvider) Login(ctx context.Context, username, password, clientID, redirectURI, state string) (*LoginResult, error) {
+	s.log.Infof("Login: attempting authentication for user %s, clientID=%s, redirectURI=%s", username, clientID, redirectURI)
```
hexfusion: can we review all info level logs in this file if the information is appropriate

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	scopes := []string{"openid", "profile", "email", "roles"}
+	if s.config != nil && len(s.config.Scopes) > 0 {
+		scopes = s.config.Scopes
+	}
+
+	// Get the signing algorithms from all keys in the JWKS
+	// This is dynamic based on the actual key types (RSA=RS256, EC=ES256)
+	// Supports key rotation scenarios where multiple keys may be active
+	idTokenSigningAlgs := []string{"RS256"} // Default to RS256
+	if jwks, err := s.jwtGenerator.GetJWKS(); err == nil && jwks.Keys != nil && len(*jwks.Keys) > 0 {
```
hexfusion: question: what if err is ! nil should we continue execution of return?

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -0,0 +1,951 @@
...
+	}
+
+	return roles
+}
+
+// extractOrganizations extracts organization names from groups
+// Groups starting with "org:" are treated as organizations
+func (s *PAMOIDCProvider) extractOrganizations(groups []string) []string {
+	var organizations []string
+	orgSet := make(map[string]bool) // Use set to avoid duplicates
```
hexfusion: unless you have a use for the bool a cheaper pattern is 

```suggestion
	orgSet := make(map[string]struct{}) // Use set to avoid duplicates
```

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,389 @@
...
+	if cfg.Auth == nil || cfg.Auth.PAMOIDCIssuer == nil {
+		return nil, fmt.Errorf("PAM OIDC issuer not configured")
+	}
+
+	pamProvider, err := pam.NewPAMOIDCProvider(ca, cfg.Auth.PAMOIDCIssuer)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create PAM OIDC provider: %w", err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
hexfusion: the ctx does not escape the constructor does it? generally speaking it is unexpected to have a ctx created or consumed in a constructor.

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,389 @@
...
+	h := &Handler{
+		log:         log,
+		cfg:         cfg,
+		pamProvider: pamProvider,
+		cancel:      cancel,
+	}
+
+	// Start cleanup goroutine
+	h.wg.Add(1)
+	go func() {
```
hexfusion, line 76: can we create  a handler.Run(ctx) func that could create a child ctx and fire the go routine

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// Close cleans up resources
+func (h *Handler) Close() {
+	if h.cancel != nil {
+		h.cancel()
+	}
+	h.wg.Wait()
+	if h.pamProvider != nil {
+		h.pamProvider.Close()
```
hexfusion: not certain how this works but do.Once is nice to ensure Close is idempotent. https://pkg.go.dev/sync#example-Once

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,389 @@
...
+		Reason:     "Error",
+		Status:     "Failure",
+	}
+	writeJSON(w, statusCode, status)
+}
+
+// extractSessionContext extracts session information from HTTP request and adds it to context
+func (h *Handler) extractSessionContext(ctx context.Context, r *http.Request) context.Context {
+	// Extract session cookie (primary method for session management)
+	if cookie, err := r.Cookie("session_id"); err == nil && cookie.Value != "" {
```
hexfusion, line 141: what if err != do we care?

internal/pam_issuer_server/handler.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// AuthToken handles OAuth2 token endpoint
+// (POST /api/v1/auth/token)
+func (h *Handler) AuthToken(w http.ResponseWriter, r *http.Request) {
+	var req pamapi.TokenRequest
+
+	// Support both JSON and form-encoded requests
+	contentType := r.Header.Get("Content-Type")
+	if strings.Contains(contentType, "application/x-www-form-urlencoded") {
```
hexfusion, line 290: Why contains vs equal?

internal/pam_issuer_server/server.go
```diff
@@ -0,0 +1,137 @@
...
+		Handler:           httpHandler,
+		ReadHeaderTimeout: 32 * time.Second,
+	}
+
+	idleConnsClosed := make(chan struct{})
+	go func() {
+		<-ctx.Done()
+		s.log.Println("Shutting down PAM issuer server")
+
+		ctx, cancel := context.WithTimeout(context.Background(), gracefulShutdownTimeout)
```
hexfusion, line 119: should the context created external to the go routine so we can shut it down cleanly?

hexfusion (commented):
first pass looking good will dig deeper next. the input on naming is for consideration.

api/v1alpha1/pam-issuer/docs.go
asafbennatan: its a pam implementation of an oidc provider

internal/pam_issuer_server/handler.go
asafbennatan: i think we have just one entity calling close and thats the server (after i have changed to use handler.Run(ctx) like you suggested

internal/pam_issuer_server/handler.go
asafbennatan: ok

internal/pam_issuer_server/handler.go
asafbennatan: application/x-www-form-urlencoded; charset=utf-8
is also valid

internal/pam_issuer_server/server.go
asafbennatan: this already happens now with the change requested above (Run(ctx))

internal/pam_issuer_server/handler.go
asafbennatan: here is a philosophical question , do we want to wrap run and close with sync.once - or return an error if they are called more than once ?
IMO i like the second approach better as it makes sure we are not hiding a bug from the caller ( the contract is that the caller shouldn't call run and close more than once , if they break it we should let them know , not hide it)

Makefile
asafbennatan: it is working but removed it since its unnecessary

Containerfile.pam-issuer
```diff
@@ -0,0 +1,78 @@
...
+COPY --from=build /runtime-root/ /
+
+# Copy custom PAM config for flightctl
+COPY --from=build /app/deploy/podman/flightctl-pam-issuer/pam-flightctl /etc/pam.d/flightctl
+
+COPY --from=build /app/bin/flightctl-pam-issuer .
+
+# Note: User database files (/etc/passwd, /etc/shadow, /etc/group) are managed
+# inside the container. For persistence across container restarts, mount them
+# as volumes in the quadlet file if needed.
```
DakCrowder: Can this comment be edited to remove the specific quadlet reference as that won't always be the runtime?

Makefile
```diff
@@ -171,6 +171,9 @@ build-agent: bin
 build-api: bin
 	$(GOENV) GOOS=$(GOOS) GOARCH=$(GOARCH) go build -buildvcs=false $(GO_BUILD_FLAGS) -o $(GOBIN) ./cmd/flightctl-api
 
+build-pam-issuer: bin
+	$(GOENV) GOOS=linux GOARCH=$(GOARCH) CGO_ENABLED=1 CGO_CFLAGS="-D_GNU_SOURCE" CGO_LDFLAGS="-ldl" go build -tags linux -buildvcs=false $(GO_BUILD_FLAGS) -o $(GOBIN) ./cmd/flightctl-pam-issuer
```
DakCrowder: This build command differs a bit from the others (e.g. build-api), should those be aligned?  Things like GOOS being hardcoded to linux and the linux tag etc.

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/init.sh
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+echo "Initializing flightctl-pam-issuer configuration"
+
+source "/utils/init_utils.sh"
+
+# Define paths
+export CERTS_SOURCE_PATH="/certs"
```
DakCrowder: It looks like CERTS_SOURCE_PATH and CERTS_DEST_PATH aren't referenced are they needed here?

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-init.container
```diff
@@ -0,0 +1,23 @@
...
+
+[Container]
+ContainerName=flightctl-pam-issuer-init
+Image=registry.access.redhat.com/ubi9/ubi-minimal:latest
+Exec=/bin/sh /config-source/init.sh
+PodmanArgs=--replace
+Network=flightctl.network
+Volume=/usr/share/flightctl/flightctl-pam-issuer:/config-source:ro,z
+Volume=/usr/share/flightctl/init_utils.sh:/utils/init_utils.sh:ro,z
+Volume=/etc/flightctl/pki:/certs:rw,z
```
DakCrowder: Can we remove the certs volume if it isn't needed by the init container?

cmd/flightctl-pam-issuer/main.go
```diff
@@ -0,0 +1,132 @@
...
+				log.Fatalf("failed to load existing self-signed certificate: %v", err)
+			}
+		} else {
+			// default to localhost if no alternative names are set
+			altNames := cfg.Service.AltNames
+			if len(altNames) == 0 {
+				altNames = []string{"localhost"}
+			}
+
+			serverCerts, err = ca.MakeAndWriteServerCertificate(ctx, srvCertFile, srvKeyFile, altNames, cfg.CA.ServerCertValidityDays)
```
DakCrowder, line 85: I think this introduces a race between the flightctl-api and this service.  What if both are started and attempt to write certs at the same time?  I don't think the way we're doing it right now is ideal but seems like something that should at least be centralized.

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
```diff
@@ -0,0 +1,34 @@
...
+Image=quay.io/flightctl/flightctl-pam-issuer:latest
+Pull=newer
+Network=flightctl.network
+EnvironmentFile=/etc/flightctl/flightctl-pam-issuer/env
+Secret=flightctl-postgresql-user-password,type=env,target=DB_PASSWORD
+Secret=flightctl-kv-password,type=env,target=KV_PASSWORD
+Environment=DB_USER=flightctl_app
+
+PublishPort=8444:8444
+Volume=/etc/flightctl/pki:/root/.flightctl/certs:rw,z
```
DakCrowder: Depending on the certs race resolution wrt the api, we may be able to remove write perms from this volume

deploy/scripts/init_host.sh
```diff
@@ -27,6 +33,13 @@ main() {
     else
         echo "Base domain already set to: $base_domain"
     fi
+    
+    # Configure PAM authentication when auth type is oidc and pamOidcIssuer is enabled
+    auth_type=$(extract_value "global.auth.type" "$SERVICE_CONFIG_FILE")
```
DakCrowder: Are auth_type and pam_enabled referenced or used?  Can we remove them  / changes to this file?

DakCrowder (commented):
One additional note - in order for the image to be pushed to quay i believe you will also need to update the publish-containers.yaml github workflow for your new image.

deploy/helm/flightctl/templates/flightctl-default-org.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: org-default
```
DakCrowder: Is the implication here that using k8s auth will necessitate creating a role to represent an organization and use rolebindings in the future to denote organizational membership for individual users?

api/v1alpha1/pam-issuer/openapi.yaml
```diff
@@ -0,0 +1,473 @@
...
+  version: v1alpha1
+  description: |
+    PAM-based OIDC authentication service for Flight Control.
+    This service provides OAuth2/OIDC authentication using system PAM (Pluggable Authentication Modules).
```
hexfusion, line 7: Just looking at the spec are we missing `subject_types_supported` which here[1] it claims as required.

```
subject_types_supported
    REQUIRED. JSON array containing a list of the Subject Identifier types that this OP supports. Valid types include pairwise and public. 
```

[1] https://openid.net/specs/openid-connect-discovery-1_0.html

internal/pam_issuer_server/handler.go
hexfusion: I don't disagree with regards to bug in caller. The caveat is that if it does happen its controlled.

internal/auth/issuer/pam/pam_issuer.go
hexfusion: @asafbennatan why not handle this? If you are certain value is always that type why use interface?

internal/auth/issuer/pam/pam_issuer.go
hexfusion: I still have a fair number of open asks with regards to error handling. discarding errors can lead to complicated debug efforts. is it expected to error here?

Makefile
fzdarsky: Related: Please ensure you don't overwrite the flags needed for FIPS enablement
https://github.com/flightctl/flightctl/blob/main/Makefile#L57

deploy/helm/flightctl/README.md
```diff
@@ -156,7 +155,7 @@ global:
   target: "standalone"
   baseDomain: "flightctl.example.com"
   auth:
-    type: "builtin"
+    type: "oidc"
```
fzdarsky: Default should be the respective platforms's auth, i.e. OpenShift Auth (OAuth2) on OpenShift.

deploy/helm/flightctl/README.md
```diff
@@ -288,9 +287,17 @@ For more detailed configuration options, see the [Values](#values) section below
...
+| global.auth.oidc.issuer | string | `""` | The base URL for the OIDC provider that is reachable by flightctl services. Example: https://auth.foo.internal/realms/flightctl |
+| global.auth.pamOidcIssuer.clientId | string | `"flightctl-client"` | OAuth2 client ID for the OIDC issuer |
+| global.auth.pamOidcIssuer.clientSecret | string | `""` | OAuth2 client secret for the OIDC issuer |
+| global.auth.pamOidcIssuer.issuer | string | `""` | The base URL for the OIDC issuer (defaults to API server URL) |
+| global.auth.pamOidcIssuer.pamService | string | `"flightctl"` | PAM service name for authentication (default: "flightctl") |
+| global.auth.pamOidcIssuer.redirectUris | list | `[]` | Allowed redirect URIs for OAuth2 flows |
+| global.auth.pamOidcIssuer.scopes | list | `["openid","profile","email","roles"]` | Supported OAuth2 scopes |
+| global.auth.type | string | `"oidc"` | Type of the auth to use. Can be one of 'k8s', 'oidc', 'builtin', 'aap', or 'none' Note: 'builtin' is a legacy mode that translates to 'oidc' with PAM issuer automatically enabled For new deployments, explicitly set type to 'oidc' and configure pamOidcIssuer settings |
```
fzdarsky, line 305: On OpenShift, I don't think a PAM issuer makes sense, because there are no user accounts on the host and OpenShift itself has the possibility to broker different backends.

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -1,13 +1,21 @@
 {{- define "flightctl.getBaseDomain" }}
   {{- if .Values.global.baseDomain }}
     {{- printf .Values.global.baseDomain }}
-  {{- else }}
-    {{- $openShiftBaseDomain := (lookup "config.openshift.io/v1" "DNS" "" "cluster").spec.baseDomain }}
-    {{- if .noNs }}
-      {{- printf "apps.%s" $openShiftBaseDomain }}
+  {{- else if eq .Values.global.target "acm" }}
+    {{- /* For ACM deployments on OpenShift, try to lookup the base domain */}}
```
fzdarsky: This shouldn't be specific to ACM (which relies on OpenShift for Auth), but simply what RHEM uses by default whenever it's deployed on OpenShift (incl. Standalone).

api/v1alpha1/pam-issuer/openapi.yaml
asafbennatan: done

cmd/flightctl-pam-issuer/main.go
asafbennatan: will it be possible to tackle these re-namings for the next PR ( i dont want it to screw up the other PR which is based on this one)

internal/pam_issuer_server/handler.go
asafbennatan: i have this as Run returning error and Close is idempotent with (sync.once)

internal/pam_issuer_server/handler.go
asafbennatan: no , we handle this downstream , if not there it might just means that the user hasnt logged in yet

Containerfile.pam-issuer
asafbennatan: done

Makefile
asafbennatan: @fzdarsky should be fine now , i have changed this so it appends to the cgo flags making sure that the required flags to build pam take precedence (appended last ) in any case there is no overlap in these flags, 
there is an overlap with CGO_ENABLE , but this is fine since i have set the CGO_ENABLE=1 for that container after the regular flags so it will take precedence

@DakCrowder pam is only available on linux we dont want to build it on other envs (you need the pam-devel headers to build the pam library to run)

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/init.sh
asafbennatan: fixed

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-init.container
asafbennatan: we are using these certs to sign jwt tokens

cmd/flightctl-pam-issuer/main.go
asafbennatan: fixed - we will just mount separate volumes for each one - certs dont have to match

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
asafbennatan: see solution above , it will write its own certificate to a different place

deploy/scripts/init_host.sh
asafbennatan: fixed

deploy/helm/flightctl/templates/flightctl-default-org.yaml
asafbennatan: possibly yes (depending on what we want to do) , but this is a matter for the next PR , i shouldnt have this in this PR - removed

api/v1alpha1/pam-issuer/openapi.yaml
asafbennatan: great catch - fixed

deploy/helm/flightctl/README.md
asafbennatan: 1. done (see comment above next to @DakCrowder  comment)
2. i changed it LMK if this is what you meant - FYI target field values are confusing as i understand that standalone there refers to standalone k8 not to standalone deployment

deploy/helm/flightctl/templates/_helpers.tpl
asafbennatan: fixed

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
asafbennatan: previously our K8 only env used oidc authentication - this is now switched for using normal k8 token reviews (unless the user defines differently)
so we need the default roles to be available in that env as well

internal/auth/issuer/pam/oidc_issuer.go
asafbennatan: will move this on the next PR if possible

internal/auth/issuer/pam/pam_authenticator.go
asafbennatan: will rename this on the next PR if possible

internal/auth/issuer/pam/pam_authenticator.go
asafbennatan: fixed , a leftover from older version where pam was inside the container so i had to provide mocks for envs that did not support pam lib (linux)

internal/auth/issuer/pam/pam_authenticator.go
asafbennatan: see comment above

internal/auth/issuer/pam/pam_issuer.go
asafbennatan:
```
The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.
```

i think we fall under the second usecase - if we just use plain locks we will have very high contention here ( as every user reads/writes to the map)

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: fixed

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -2,11 +2,17 @@
   {{- if .Values.global.baseDomain }}
     {{- printf .Values.global.baseDomain }}
   {{- else }}
-    {{- $openShiftBaseDomain := (lookup "config.openshift.io/v1" "DNS" "" "cluster").spec.baseDomain }}
-    {{- if .noNs }}
-      {{- printf "apps.%s" $openShiftBaseDomain }}
+    {{- /* For OpenShift deployments, try to lookup the base domain */}}
+    {{- $dnsConfig := (lookup "config.openshift.io/v1" "DNS" "" "cluster") }}
```
fzdarsky, line 6: Just as reminder to ourselves: This won't work with the current ACM integration as there's no cluster at the time of rendering.

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: add logs throughout

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: add logs throughout

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: it is not , do you mean that we need to add a guard check against that ?

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: done

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: fixed

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: done

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
fzdarsky: Wondering whether we still need the `global.target`, it's become a bit confusing.

Original intention was to simplify configuration by selecting a "profile" that'd determine various config parameters:
1. auth backend configuration (OpenShift vs. AAP vs. the built-in Keycloak)
2. whether to deploy Keycloak or not
3. the UI experience to deploy
4. whether to enable some OpenShift-specifics like Routes

Without Keycloak, 2) is no longer relevant, 1) defaults to OpenShift (if on OpenShift) and can be reconfigured at runtime per org, 3)+4) are still relevant.

Should we replace this by a more explicit `enableOpenShiftExtensions=auto|true|false` and a `integrations=none|acm|aap` (if we need more than just configuring images)?

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: should be fine now , LMK if you find anything else (fixed a bunch of these )

internal/auth/issuer/pam/pam_issuer.go
asafbennatan: done

internal/pam_issuer_server/handler.go
asafbennatan: done

cmd/flightctl-pam-issuer/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/flightctl/flightctl/internal/api_server/middleware"
+	"github.com/flightctl/flightctl/internal/config"
+	"github.com/flightctl/flightctl/internal/crypto"
+	"github.com/flightctl/flightctl/internal/instrumentation"
```
celdrake:
```suggestion
	"github.com/flightctl/flightctl/internal/instrumentation/tracing"
```
OTherwise "make deploy" fails.

cmd/flightctl-pam-issuer/main.go
```diff
@@ -0,0 +1,132 @@
...
+				x509Cert.Subject.CommonName, x509Cert.Issuer.CommonName, x509Cert.NotAfter)
+		}
+	}
+
+	tlsConfig, _, err := crypto.TLSConfigForServer(ca.GetCABundleX509(), serverCerts)
+	if err != nil {
+		log.Fatalf("failed creating TLS config: %v", err)
+	}
+
+	tracerShutdown := instrumentation.InitTracer(log, cfg, "flightctl-pam-issuer")
```
celdrake:
```suggestion
tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-pam-issuer")
```
Seems to be the pattern in the rest of the modules.

fzdarsky (approved):

PR #1895, opened 2025-11-03
EDM-2465: Add upstream RHEL/CS-10 targets

gshilin-sdb:
## Summary by CodeRabbit

* **Chores**
  * Standardized CI build targets to Fedora 43 and added EPEL 10 variants (x86_64, aarch64); removed legacy Fedora 41/Rawhide targets.
  * Switched the current-version action to use an RPM-oriented generator.
  * Added a package-level network enable flag for module fetching and enabled a failure-comment notification to trigger a rebuild command.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -10,28 +10,27 @@ packages:
     upstream_package_name: flightctl
     downstream_package_name: flightctl
     upstream_tag_template: v{version} # remove v from the start of the version tags (vx.y.z)->x.y.z
+    enable_net: True # this is necessary for go modules to download the sources
+    notifications:
+      failure_comment:
+        message: "/packit build"
 actions:
   get-current-version:
     - bash -c './hack/current-version --rpm | sed "s/^v//; s/-/~/g"'
 jobs:
   - job: copr_build
     trigger: pull_request
-    enable_net: True # this is necessary for go modules to download the sources
     targets:
       - fedora-43-x86_64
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
     targets:
       - fedora-42-x86_64
       - epel-9-aarch64
+      - epel-10-x86_64
+      - centos-stream-10-x86_64
```
hexfusion: is this missing `centos-stream-10-aarch64`

gshilin-sdb: This is `trigger: pull_request` section, so this is just for verification...
But there is no clear reason not to add all targets here, so added them all.

gshilin-sdb: Comparing to e2e time, this is not a problem at all

fzdarsky: We don't publish the feddora-rawhide or centos-stream chroots, so why not drop them (incl. from COPR)?

gshilin-sdb: @avishayt asked for centos-stream. It doesn't take time to build them, so why not to have?

avishayt: Let’s drop the CentOS Stream 9 chroots since EPEL 9 is sufficient, and add EPEL 10 chroots now to get ahead of RHEL 10 support, and also drop rawhide

avishayt (approved):

PR #1909, opened 2025-11-05
EDM-2304: Add setup binary in containers for rendering go templates

DakCrowder:
Adds a shared "setup" binary that does the following:
- Takes in an arbitrary yaml file as a data source
- Reads go templates in a specified /templates directory
- Overlays the supplied data over the templates and places corresponding files in a /rendered directory

The binary has been added to some shipped containers.  A new ui-setup container was also added to mirror the db-setup container for performing setup/tasks for images that this repo doesn't control.  Some initialization tasks that are still needed from the init.sh bash scripts have been moved to "post-init" containers.

Effective usage looks something like:

```
[Container]
ContainerName=flightctl-api-init
# Binary is mounted in the flightctl-api image
Image=quay.io/flightctl/flightctl-api:latest

# Service config, templates, and rendered output directory are all mounted
Volume=/etc/flightctl/service-config.yaml:/app/service-config.yaml:ro,z
Volume=/usr/share/flightctl/flightctl-api/env.template:/app/templates/env.template:ro,z
Volume=/usr/share/flightctl/flightctl-api/config.yaml.template:/app/templates/config.yaml.template:ro,z
Volume=/etc/flightctl/flightctl-api:/app/rendered:rw,z

# Setup is called as the Exec command to perform the templating
Exec=/app/flightctl-service-setup
```

In the future the per-service setup might get more specific / granular, and could do things like wait for db, perform some other setup logic (e.g. creating records or other derived config).

A couple other notes:
- "builtin" was never supported as an option in service-config.yaml as an authentication option.  I have removed it and tried to mirror the logic added recently for the PAM service without considering "builtin"
- Static alert-exporter config was removed.  For now the alert-exporter service has been modified to use the generated api config (as do other services like periodic and worker) which I think should be changed over the long term but gets things working for now

## Summary by CodeRabbit

* **New Features**
  * Added a standalone service-setup CLI and a UI-setup container; CI now builds and publishes the UI-setup image.

* **Improvements**
  * Templates gain richer defaults and conditional auth, OIDC/PAM and rate‑limit settings.
  * Deployment units add post‑init services, improved restart/oneshot behavior, and a SKIP_BUILD guard for optional builds.

* **Chores**
  * Removed legacy per‑service init scripts, simplified init flows, updated container images/volumes/exec to use the new setup flow, and adjusted packaging entries.

* **Tests**
  * Added comprehensive template‑rendering tests.


deploy/podman/flightctl-alert-exporter/flightctl-alert-exporter.container
```diff
@@ -13,7 +13,7 @@ Environment=HOME=/root
 Secret=flightctl-postgresql-user-password,type=env,target=DB_PASSWORD
 Secret=flightctl-kv-password,type=env,target=KV_PASSWORD
 Environment=DB_USER=flightctl_app
-Volume=/usr/share/flightctl/flightctl-alert-exporter/config.yaml:/root/.flightctl/config.yaml:ro,z
+Volume=/etc/flightctl/flightctl-api/config.yaml:/root/.flightctl/config.yaml:ro,z
```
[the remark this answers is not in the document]
DakCrowder: This follows a pattern other services (such as periodic and worker) currently follow for sharing the /etc/flightctl/flightctl-api/config.yaml.  This might change in the future, but for now since the config generation did not exist and was using a static file that wouldn't work going forward this is ok

deploy/podman/flightctl-api/flightctl-api-config/config.yaml.template
```diff
@@ -1,40 +1,82 @@
...
+      redirectUris:
+{{- if .global.auth.pamOidcIssuer.redirectUris}}
+{{- range .global.auth.pamOidcIssuer.redirectUris}}
+        - {{.}}
+{{- end}}
+{{- else}}
+        - https://{{.global.baseDomain}}/callback
+{{- end}}
+{{- end}}{{/* end pamOidcIssuer conditional */}}
+{{- end}}{{/* end auth type conditional */}}
```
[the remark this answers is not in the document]
DakCrowder: The deploy/podman/service-config.yaml only supports "aap" not "AAPGateway" (there are no refs to AAPGateway in the deploy/podman directory)

deploy/podman/flightctl-alert-exporter/flightctl-alert-exporter.container
```diff
@@ -13,7 +13,7 @@ Environment=HOME=/root
 Secret=flightctl-postgresql-user-password,type=env,target=DB_PASSWORD
 Secret=flightctl-kv-password,type=env,target=KV_PASSWORD
 Environment=DB_USER=flightctl_app
-Volume=/usr/share/flightctl/flightctl-alert-exporter/config.yaml:/root/.flightctl/config.yaml:ro,z
```
DakCrowder, line 16: Note: this config was static and has been moved to sharing the api config which is what some other services (worker, periodic etc.) do.

I would like to split those out in the future to having more control over their own config (and not being reliant on flightctl-api rendering the confg) but going with this for now.

deploy/podman/service-config.yaml
```diff
@@ -1,7 +1,7 @@
 global:
   baseDomain:
   auth:
-    type: none # aap, oidc, builtin (legacy, translates to oidc with PAM), or none
```
DakCrowder, line 4: builtin was never exposed to users previously in this context - removed as we don't need to support

deploy/podman/flightctl-ui/flightctl-ui-config/env.template
```diff
@@ -1,12 +1,22 @@
...
 TLS_KEY=/app/certs/server.key
-AUTH_CLIENT_ID={{AUTH_CLIENT_ID}}
-INTERNAL_AUTH_URL={{INTERNAL_AUTH_URL}}
-AUTH_INSECURE_SKIP_VERIFY={{AUTH_INSECURE_SKIP_VERIFY}}
+AUTH_INSECURE_SKIP_VERIFY={{.global.auth.insecureSkipTlsVerify}}
+{{if eq .global.auth.type "aap"}}
+AUTH_CLIENT_ID={{.global.auth.aap.oAuthApplicationClientId}}
+INTERNAL_AUTH_URL={{.global.auth.aap.apiUrl}}
+{{else if eq .global.auth.type "oidc"}}
+{{if .global.auth.pamOidcIssuer.enabled}}
```
DakCrowder, line 15: @asafbennatan I tried to be sensible with fallbacks / default values for the pamOidcIssuer enabled case but not sure if this matches your expectations for how it should be configured from the service-config.yaml

asafbennatan: i believe that these two will be removed and ignored anyway ? @celdrake ?

deploy/podman/service-config.yaml
asafbennatan: thought we had it as default for stuff  - its good that we remove it

deploy/podman/flightctl-alert-exporter/flightctl-alert-exporter.container
asafbennatan: unrelated to this PR: i would add that i see a pattern in the code where the flightctl-api config structure is being used around by other services , seems like flightctl-api config is trying to be somekind of service-config.yaml structure - but we shouldnt do that , example: 

dbConfig

i would argue that the api config requires only the connection details there 
for example:
	MigrationUser     string       `json:"migrationUser,omitempty"`
	MigrationPassword SecureString `json:"migrationPassword,omitempty"`


should not be there , IMO it should be in its own config.go structure file that is referenced only by flightctl-db-migrate

service-config.yaml should have a section 

db-migrate:
    <db migrate specific config>

deploy/podman/flightctl-ui/flightctl-ui-config/env.template
celdrake: Yes, these are the currently hardcoded settings that the UI uses to access the static auth provider.
I have a branch that removes these, but I didn't publish it yet.

https://github.com/flightctl/flightctl/tree/EDM-2373-remove-static-auth-ui-configuration

deploy/podman/flightctl-alert-exporter/flightctl-alert-exporter.container
DakCrowder: +1, agreed.  I've added a ticket to track separating out some of these config files better so the api config yaml isn't just used as a catch all and shared among all the services

deploy/podman/flightctl-ui/flightctl-ui-config/env.template
DakCrowder: Added a ticket to audit/clean up once that branch is merged

deploy/podman/flightctl-ui/flightctl-ui-init.container
```diff
@@ -1,23 +1,22 @@
...
-Image=registry.access.redhat.com/ubi9/ubi-minimal
 ContainerName=flightctl-ui-init
-Volume=/usr/share/flightctl/init_utils.sh:/utils/init_utils.sh:ro,z
-Volume=/usr/share/flightctl/flightctl-ui:/config-source:ro,Z
-Volume=/etc/flightctl/pki:/certs-source:ro,z
-Volume=flightctl-ui-certs:/certs-destination:rw,z
-Volume=/etc/flightctl/flightctl-ui:/config-destination:rw,Z
-Volume=/etc/flightctl/service-config.yaml:/service-config.yaml:ro,Z
-Exec=/bin/sh /config-source/init.sh
+Image=quay.io/flightctl/flightctl-ui-setup:latest
```
keitwb, line 8: Is it not feasible to follow the same pattern as the API init and put this in the UI container? You can have a separate build container within the UI Containerfile for Go stuff and only pull in the necessary binaries into the final image to not mix with the nodejs stuff.

I'm just pretty wary of adding even more containers for such a small thing.

deploy/podman/flightctl-api/flightctl-api-init.container
```diff
@@ -1,18 +1,20 @@
...
-Image=registry.access.redhat.com/ubi9/ubi-minimal
 ContainerName=flightctl-api-init
-Network=flightctl.network
-Volume=/usr/share/flightctl/flightctl-api:/config-source:ro,z
-Volume=/usr/share/flightctl/init_utils.sh:/utils/init_utils.sh:ro,z
-Volume=/etc/flightctl/pki:/certs:rw,z
-Volume=/etc/flightctl/flightctl-api:/config-destination:rw,z
-Volume=/etc/flightctl/service-config.yaml:/service-config.yaml:rw,z
-Exec=/bin/sh /config-source/init.sh
+Image=quay.io/flightctl/flightctl-api:latest
```
kkyrazis, line 9: @gshilin-sdb @kkyrazis The base image changed for this service. We'll need to ensure downstream reflects this.

deploy/podman/flightctl-alertmanager-proxy/flightctl-alertmanager-proxy-init.container
```diff
@@ -5,14 +5,13 @@ After=flightctl-alertmanager.service
 Wants=flightctl-alertmanager.service
 
 [Container]
-Image=registry.access.redhat.com/ubi9/ubi-minimal
+Image=quay.io/flightctl/flightctl-alertmanager-proxy:latest
```
kkyrazis, line 8: @gshilin-sdb @kkyrazis The base image changed for this service. We'll need to ensure downstream reflects this.

deploy/podman/flightctl-cli-artifacts/flightctl-cli-artifacts-init.container
```diff
@@ -1,18 +1,19 @@
...
-ContainerName=flightctl-cli-artifacts-config
-Volume=/usr/share/flightctl/flightctl-cli-artifacts:/config-source:ro,z
-Volume=/usr/share/flightctl/init_utils.sh:/utils/init_utils.sh:ro,z
-Volume=/etc/flightctl/flightctl-cli-artifacts:/config-destination:rw,z
-Volume=/etc/flightctl/service-config.yaml:/service-config.yaml:ro,z
-Volume=/etc/flightctl/pki:/certs-source:ro,z
-Volume=flightctl-cli-artifacts-certs:/certs-destination:rw,z
-Exec=/bin/sh /config-source/init.sh
+ContainerName=flightctl-cli-artifacts-init
+Image=quay.io/flightctl/flightctl-cli-artifacts:latest
```
kkyrazis, line 9: @gshilin-sdb @kkyrazis The base image changed for this service. We'll need to ensure downstream reflects this.

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-init.container
```diff
@@ -4,19 +4,18 @@ PartOf=flightctl.target
 
 [Container]
 ContainerName=flightctl-pam-issuer-init
-Image=registry.access.redhat.com/ubi9/ubi-minimal:latest
-Exec=/bin/sh /config-source/init.sh
+Image=quay.io/flightctl/flightctl-pam-issuer:latest
```
kkyrazis, line 7: @gshilin-sdb @kkyrazis The base image changed for this service. We'll need to ensure downstream reflects this.

deploy/podman/flightctl-ui/flightctl-ui-post-init.container
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=Flight Control UI Post-Initialization
+PartOf=flightctl.target
+After=flightctl-api.service
+Wants=flightctl-api.service
+
+[Container]
+Image=registry.access.redhat.com/ubi9/ubi-minimal
```
keitwb: We should at least put the `latest` tag to make it clear that we just want the latest version of this.

deploy/podman/flightctl-ui/flightctl-ui-init.container
kkyrazis, line 8: @gshilin-sdb @kkyrazis The base image changed for this service. We'll need to ensure downstream reflects this.

deploy/podman/flightctl-alert-exporter/flightctl-alert-exporter.container
ori-amizur, line 16: If this configuration is shared between services, it shouldn't reside under /etc/flightctl-api

deploy/podman/flightctl-ui/flightctl-ui-post-init.container
DakCrowder: Might pin to another tag to avoid latest but will take a look

deploy/podman/flightctl-ui/flightctl-ui-init.container
DakCrowder: I'm not sure if its worth the overhead and since the ui container is packaged in another repo I'm viewing it as effectively external to this repo.  We also have concerns like the db (which already has a setup container), kv, telemetry etc. that might require similar treatment in the future.

DakCrowder: To add a bit more context, I think we could have a more generalized container that could be shared by some of these services as well - but based on the initial conversations before implementation with @asafbennatan and @fzdarsky it sounded like the setup processes might become more service-specific in some cases (e.g. api might want to wait for db or some external service as well) which led me to continuing the per-service approach

deploy/podman/flightctl-alert-exporter/flightctl-alert-exporter.container
DakCrowder: Agreed - have some work to follow up on decoupling this [here](https://issues.redhat.com/browse/EDM-2526)

deploy/podman/flightctl-ui/flightctl-ui-init.container
keitwb: Yeah I guess I don't understand why we aren't leveraging the systemd functionality for this though, especially since this is all quadlet-specific changes. `ExecStartPre` can do all of the config rendering, and `ExecStartPost` on the db service can wait for the startup/migrations before the service will be marked as ready which will automatically block other services that have an `After` dependency on it without the need for separate init services. Has anybody looked into this?

DakCrowder: I think part of the vision with the containerized approach is some functionality could be shared between the quadlets AND helm deployments.  As for `ExecStartPre` etc. we still have the question of exactly what is doing the go template rendering and setup, how we ship it etc.  We're moving away from bash so are we then shipping a go installer helper binary in the rpm, could that also do _some_ things and live alongside our containerized setup, I'm not 100% sure exactly where that line is.

PR #1912, opened 2025-11-05
EDM-2265: fix E2E rollout tests beforeEach in OCP

siserafin:
Rollout tests in OCP are failing at BeforeEach step, when the snapshot is restored since the agent fails to send a new ER with error: "private key does not match public key".
In this PR I make sure: 

- the agent is shut down before removing the agent's files in /var/lib/flightctl and the snapshot is created
- the agent is restarted after the snapshot is restored

## Summary by CodeRabbit

* **Tests**
  * Improved VM snapshot creation and management of the agent service during test harness setup for more reliable test runs and clearer logs.
  * Added a medium-length timeout and updated rollout tests to use it, adjusting wait time for device-update verification.

### DIFF
--- a/test/e2e/rollout/rollout_suite_test.go
+++ b/test/e2e/rollout/rollout_suite_test.go
@@ -11,6 +11,7 @@ import (
 
 const TIMEOUT = "5m"
 const POLLING = "125ms"
+const MEDIUMTIMEOUT = "10m"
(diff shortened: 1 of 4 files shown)


itzikb-redhat:
/LGTM , Not sure how increasing the timeout should helps in this case.

amalykhi (approved):

PR #1913, opened 2025-11-05
EDM-2443: Add Inline Provider for Quadlet Apps

kkyrazis:
This PR enables the full flow of Quadlet applications via inline providers.

E.g.
```
spec:
  applications:
      - name: my-app
        appType: quadlet
        inline:
          - content: |
              [Container]
              Image=10.192.27.200:5000/flightctl-tests/alpine:v1
              Exec=sleep infinity
              
              [Install]
              WantedBy=default.target
            path: test.container
```

results in:

```
status:
  applications:
    - name: my-app
      ready: 1/1
      restarts: 0
      status: Running
  applicationsSummary:
    info: Device's application workloads are healthy.
    status: Healthy
```

## Summary by CodeRabbit

* **New Features**
  * Full Quadlet app support: parsing from inline specs and directories, drop-in (.conf) merging, image/mount target extraction, and install/verify flows.
  * Improved .env handling when environment variables are present.
  * SELinux rules added to allow agent interactions with system/init for unit management.

* **Bug Fixes / Reliability**
  * Consistent, descriptive error reporting for quadlet parsing and missing quadlet files.

* **Tests**
  * Extensive unit tests covering quadlet parsing, drop-in behavior, and target extraction.

### DIFF
--- a/hack/current-version
+++ b/hack/current-version
@@ -13,11 +13,12 @@ if [[ -z "$tag" ]]; then
   tag=$(git describe --tags --exclude latest 2>/dev/null || echo -n "v0.0.0-unknown")
 fi
 
-# For RPM packaging (when called from packit), remove the git commit hash suffix
-# since COPR will add its own commit hash, preventing duplication
+# For RPM packaging (when called from packit), remove branch info and git commit hash
+# since COPR will add its own branch and commit information, preventing duplication
 if [[ "${PACKIT_BUILD:-}" == "1" ]] || [[ "$1" == "--rpm" ]]; then
-  # Remove the git commit hash part (e.g., "-g1234567") from the tag
-  tag=$(echo "$tag" | sed 's/-g[0-9a-f]*$//')
+  # Remove everything after the base version (e.g., "-main-211-g1234567") from the tag
+  # This leaves only the base version like "v1.0.0" and lets COPR handle the rest
+  tag=$(echo "$tag" | sed 's/-.*$//')
 fi
 
 echo -n "$tag"


packaging/selinux/flightctl_agent.te
```diff
@@ -178,6 +178,7 @@ systemd_start_all_unit_files(flightctl_agent_t)
 systemd_status_all_unit_files(flightctl_agent_t)
 systemd_reload_all_services(flightctl_agent_t)
 systemd_start_all_services(flightctl_agent_t)
+allow flightctl_agent_t systemd_unit_file_t:service stop;
```
kkyrazis, line 181: note: the `systemd_stop_all_services` macro doesn't appear to be available everywhere

internal/agent/client/quadlet.go
```diff
@@ -1,37 +1,235 @@
...
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	"github.com/flightctl/flightctl/internal/api/common"
+	"github.com/flightctl/flightctl/internal/quadlet"
+	"github.com/samber/lo"
+)
+
 const (
 	QuadletProjectLabelKey = "io.flightctl.quadlet.project"
 
-	QuadletContainerExtension = ".container"
```
kkyrazis, line 6: moved to the quadlet package

internal/agent/client/quadlet.go
```diff
@@ -1,37 +1,235 @@
...
+		}
+	}
+
+	if len(dropinFiles) == 0 {
+		return nil
+	}
+
+	// sort the names in ascending order so that the highest priority items are applied
+	// last which allows for proper overriding
+	sort.Strings(dropinNames)
```
hexfusion, line 71: :+1:

hexfusion (approved):
lets fixup in followups, this should enable testing

packaging/selinux/flightctl_agent.te
keitwb: For some reason it is called `systemd_stop_systemd_services` in RHEL9 (and Fedora). It appears to be an older rule that uses the more specific `systemd_unit_file_t` domain instead of the newer attribute `systemd_unit_file_type` which could expand to multiple domains, but I doubt it matters for our uses.

I would either use that or switch this to use the `systemd_unit_file_type` attribute instead.

kkyrazis: Thanks. I'll test and open up a PR with the policy change (assuming everything works)

PR #1916, opened 2025-11-06
EDM-2460: Add local Prometheus metrics in the agent

asafbss:
- Implement a MetricsCollector in the agent that integrates with the existing `internal/instrumentation/metrics` package.
- Expose internal Prometheus metrics via a local, loopback-only endpoint (127.0.0.1).
- Collect existing RPC timing data through a shared hook (name, duration, error).
- Publish each RPC operation as a Prometheus histogram (*_duration_seconds) with a result label (success|error).
- Use Prometheus default latency buckets (sub-second through multi-second) to enable percentile (p95/p99) and SLO calculations.
- Require no code changes at RPC call sites—the collector attaches transparently to the existing hook.
- Example exported metrics:
  - create_enrollmentrequest_duration_seconds{result}
  - get_enrollmentrequest_duration_seconds{result}
  - get_rendered_device_spec_duration_seconds{result}
  - update_device_status_duration_seconds{result}
  - patch_device_status_duration_seconds{result}
  - create_certificate_signing_request_duration_seconds{result}
  - get_certificate_signing_request_duration_seconds{result}

## Summary by CodeRabbit

* **New Features**
  * Metrics enabled by default with a configurable Prometheus /metrics endpoint.
  * New RPC latency tracking with per-operation histograms.
  * Diagnostics now collect Prometheus metrics plus expanded profiling outputs.

* **Refactor**
  * Modernized metrics server with per-run configuration (address, timeouts, concurrency, handler wrapping).
  * Observability startup simplified to run metrics and profiling components in parallel.


internal/agent/instrumentation/instrumentation.go
```diff
@@ -2,19 +2,71 @@ package instrumentation
...
-	srv *coreinst.PprofServer
+	srv *instpprof.PprofServer
 }
 
-// Run blocks until ctx is canceled
+// NewPprofServer wires a core pprof server if profiling is enabled.
+// It exposes /debug/pprof on loopback.
+func NewPprofServer(l *log.PrefixLogger, cfg *agent_config.Config) *pprofServer {
+	ps := &pprofServer{log: l}
+	if cfg == nil || !cfg.ProfilingEnabled {
```
hexfusion, line 69: if config is nil that would be very unexpected should we tolerate this, can it happen?

internal/agent/instrumentation/instrumentation.go
```diff
@@ -29,14 +81,29 @@ func (p *pprofServer) Run(ctx context.Context) {
...
+		metrics: NewMetricsServer(l, cfg),
+		pprof:   NewPprofServer(l, cfg),
+		log:     l,
 	}
+}
 
-	ps.srv = coreinst.NewPprofServer(l)
-	return ps
+// Run starts all observability components (e.g., metrics, pprof) and blocks until ctx is canceled.
+func (ai *AgentInstrumentation) Run(ctx context.Context) {
```
hexfusion, line 108: can we log when this starts and when it shuts down please

internal/agent/instrumentation/instrumentation.go
```diff
@@ -29,14 +81,29 @@ func (p *pprofServer) Run(ctx context.Context) {
...
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go func() { defer wg.Done(); ai.metrics.Run(ctx) }()
+	go func() { defer wg.Done(); ai.pprof.Run(ctx) }()
```
hexfusion, line 113: what happens if one of these goroutines returns do we know why? should we shutdown the other goroutine?

internal/agent/instrumentation/instrumentation.go
```diff
@@ -2,19 +2,71 @@ package instrumentation
...
+	ms := &metricsServer{log: l}
+	if cfg == nil || !cfg.MetricsEnabled {
+		return ms
+	}
+
+	rpcCollector := agentmetrics.NewRPCCollector(l)
+
+	// Provide the Observe hook to producers (enrollment/management RPCs).
+	cfg.SetEnrollmentMetricsCallback(rpcCollector.Observe)
+	cfg.SetManagementMetricsCallback(rpcCollector.Observe) // <- was duplicated earlier
```
hexfusion: `// <- was duplicated earlier` can we remove :)

internal/agent/instrumentation/metrics/rpc_collector.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+// Observe is the drop-in hook for the agent’s RPC timing callback.
+// It matches the callback signature (operation, seconds, err) and records
+// latency under <operation>_duration_seconds with result="success|error".
+func (c *RPCCollector) Observe(operation string, seconds float64, err error) {
+	op := canonicalizeOp(operation)
+	if op == "" {
+		return
```
hexfusion, line 79: do we care if op is empty should we log?

internal/agent/instrumentation/metrics/rpc_collector.go
```diff
@@ -0,0 +1,129 @@
...
+	if !allowed || hv == nil {
+		c.log.WithFields(logrus.Fields{
+			"operation": operation,
+		}).Error("unknown RPC metric operation")
+		return
+	}
+
+	result := "success"
+	if err != nil {
+		result = "error"
```
hexfusion, line 97: do we care what the error is? should we log?

asafbss: I don’t think the collector should log per-error. Its job is just to record latency + result, not to interpret errors. If we want error details, the RPC caller should log them at the call site

internal/agent/instrumentation/instrumentation.go
asafbss: We don’t shut down the other goroutine. Metrics and pprof are independent. One may be disabled (returns immediately) while the other keeps serving—that’s valid.
Do we know why one returned? Yes—each component’s Run(ctx) already logs its reason (disabled vs. error). The wrapper doesn’t need to decide policy.

internal/agent/instrumentation/metrics/rpc_collector.go
asafbss: fixed

internal/agent/instrumentation/instrumentation.go
asafbss: fixed

internal/agent/instrumentation/instrumentation.go
asafbss: This shouldn’t happen, but we guard against it to avoid a panic. It effectively disables instrumentation until the underlying issue is fixed.

internal/agent/instrumentation/instrumentation.go
asafbss: @hexfusion When both are disabled, the suggestion to log “instrumentation starting,” followed immediately by “stopped" might look bad.

Proposal: rely on the metrics/pprof components to log their own state and make the wrapper silent.
Resulting logs when both enabled would be:

- pprof server listening on http://127.0.0.1:15689/debug/pprof/
- metrics server listening on http://127.0.0.1:15690/metrics/

WDYT?

packaging/sosreport/sos/report/plugins/flightctl.py
```diff
@@ -23,21 +23,27 @@ def setup(self):
...
+        # Prometheus metrics
+        self.add_cmd_output(
+            "curl -fsS --max-time 10 -H 'Accept: text/plain' 'http://127.0.0.1:15690/metrics'",
+            suggest_filename="flightctl-metrics.txt",
+        )
+
         # Goroutines
         self.add_cmd_output(
-            f"curl -fsS --max-time 10 'http://127.0.0.1:15689/debug/pprof/goroutine?debug=2'",
+            "curl -fsS --max-time 10 'http://127.0.0.1:15689/debug/pprof/goroutine?debug=2'",
```
ori-amizur, line 34: Will these commands fail if the endpoint 127.0.0.1:15689 doesn't exist? Will he other artifacts be produced (agent journal) ?

internal/agent/instrumentation/instrumentation.go
hexfusion: makes sense

hexfusion (approved):
looks good to make progress, thanks!

packaging/sosreport/sos/report/plugins/flightctl.py
asafbss: Yes — if the loopback endpoint isn’t running, curl exits non-zero and writes an error like
curl: (7) Failed to connect to 127.0.0.1:15689 into the sos artifact.
sos keeps going, so the other artifacts (including the agent journal) are still produced.

PR #1918, opened 2025-11-06
EDM-2267: Factor out systemd status into own field

fzdarsky:
## Summary by CodeRabbit

* **New Features**
  * Device status now includes systemd unit details (unit, description, enable/load/active/sub states).

* **Documentation**
  * OpenAPI/Swagger schemas and embedded API docs refreshed to expose new systemd types and DeviceStatus.systemd.

* **Bug Fixes**
  * Improved parsing and handling of diverse and edge-case systemd outputs.

* **Tests**
  * Expanded tests covering multiple unit types, varied states, malformed outputs, CLI edge cases and error paths.

### DIFF
--- a/internal/agent/client/systemd.go
+++ b/internal/agent/client/systemd.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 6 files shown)


api/v1alpha1/openapi.yaml
```diff
@@ -4527,6 +4532,70 @@ components:
       additionalProperties:
         type: string
         description: A user-defined custom property value.
+    SystemdUnitStatus:
+      type: object
+      required:
+        - unit
+        - load
+        - active
```
hexfusion: my only input is some of this status is lower level, do we want to abstract any of this or do we feel we can educate the users about these details. they are fairly intuitive.

fzdarsky: I agree it's low-level. No idea how to abstract it more, so it's more intuitive for some users while making it less intuitive for others (those who know systemd), though.

celdrake: I agree too.
I'm wondering if in the UI we should combine these into a handful of more directly understood statuses.

celdrake:
@fzdarsky Could you please rebase it so we can merge it please?

celdrake (approved):

PR #1922, opened 2025-11-06
EDM-2508: agent/device/application: prefetch image targets

hexfusion:
This PR adds the ability for the prefetch manager to prefetch dependencies for both embedded and image based applications.

## Summary by CodeRabbit

* **New Features**
  * Added Quadlet application type support alongside Compose.

* **Improvements**
  * More reliable multi-application handling with an upfront inspect-then-mount/unmount flow.
  * Two-phase OCI target collection with per-image nested-target caching to avoid redundant work.
  * Improved prefetch scheduling with readiness/requeue semantics for more robust dependency discovery and retries.

* **Tests**
  * Expanded tests for multi-app flows, caching behavior, and error/retry scenarios.

### DIFF
--- a/internal/agent/device/applications/manager.go
+++ b/internal/agent/device/applications/manager.go
@@ -14,6 +14,7 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/device/systeminfo"
 	"github.com/flightctl/flightctl/internal/agent/shutdown"
 	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/samber/lo"
 )
 
 const (
@@ -28,6 +29,12 @@ type manager struct {
 	systemdClient *client.Systemd
 	readWriter    fileio.ReadWriter
 	log           *log.PrefixLogger
+
+	// cache of extracted nested OCI targets
+	ociTargetCache *provider.OCITargetCache
+
+	// cache of temporary extracted app data
+	appDataCache map[string]*provider.AppData
 }
 
 func NewManager(
@@ -39,11 +46,13 @@ func NewManager(
 ) Manager {
 	bootTime := systemInfo.BootTime()
 	return &manager{
-		readWriter:    readWriter,
-		podmanMonitor: NewPodmanMonitor(log, podmanClient, systemdClient, bootTime, readWriter),
-		podmanClient:  podmanClient,
-		systemdClient: systemdClient,
-		log:           log,
+		readWriter:     readWriter,
+		podmanMonitor:  NewPodmanMonitor(log, podmanClient, systemdClient, bootTime, readWriter),
+		podmanClient:   podmanClient,
+		systemdClient:  systemdClient,
+		log:            log,
+		ociTargetCache: provider.NewOCITargetCache(),
+		appDataCache:   provider.NewAppDataCache(),
 	}
 }
 
@@ -100,7 +109,15 @@ func (m *manager) BeforeUpdate(ctx context.Context, desired *v1alpha1.DeviceSpec
 	m.log.Debug("Pre-checking application dependencies")
 	defer m.log.Debug("Finished pre-checking application dependencies")
 
-	providers, err := provider.FromDeviceSpec(ctx, m.log, m.podmanMonitor.client, m.readWriter, desired, provider.WithEmbedded())
+	providers, err := provider.FromDeviceSpec(
+		ctx,
+		m.log,
+		m.podmanMonitor.client,
+		m.readWriter,
+		desired,
+		provider.WithEmbedded(),
+		provider.WithAppDataCache(m.appDataCache),
+	)
 	if err != nil {
 		return fmt.Errorf("parsing apps: %w", err)
 	}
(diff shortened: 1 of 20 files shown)


internal/agent/device/applications/provider/image.go
```diff
@@ -110,37 +129,53 @@ func (p *imageProvider) Verify(ctx context.Context) error {
...
-			return fmt.Errorf("getting app path: %w", err)
-		}
-		p.spec.Path = path
 
 		// ensure the compose application content in tmp dir is valid
 		if err := ensureCompose(p.readWriter, tmpAppPath); err != nil {
 			return fmt.Errorf("ensuring compose: %w", err)
 		}
+	case v1alpha1.AppTypeQuadlet:
+		p.spec.ID = client.NewComposeID(p.spec.Name)
```
hexfusion, line 127: @kkyrazis is this accurate

kkyrazis: Yes, I just reused the same mechanism.

internal/agent/device/applications/manager.go
```diff
@@ -173,34 +165,57 @@ func (m *manager) Shutdown(ctx context.Context, state shutdown.State) error {
...
+		return nil, fmt.Errorf("collecting base OCI targets: %w", err)
+	}
+	allTargets = append(allTargets, baseTargets...)
+	m.log.Debugf("Collected %d base OCI targets", len(baseTargets))
+
+	// extract nested targets, cache results to avoid re-extraction
+	var cacheKey string
+	for _, target := range baseTargets {
+		if target.Type == dependency.OCITypeImage {
+			cacheKey += target.Reference + ","
```
hexfusion: reworking this

internal/agent/device/applications/provider/provider.go
```diff
@@ -24,7 +24,7 @@ import (
 // and lifecycle operations for installation to disk.
 type Provider interface {
 	// OCITargets returns the list of OCI images and artifacts required by the provider.
-	OCITargets(pullSecret *client.PullSecret) ([]dependency.OCIPullTarget, error)
+	OCITargets(ctx context.Context, pullSecret *client.PullSecret) ([]dependency.OCIPullTarget, error)
```
kkyrazis: I don't think we need this anymore since it just happens in CollectBaseOCITargets or the nested method.

internal/agent/device/applications/provider/provider.go
```diff
@@ -58,6 +58,246 @@ type ApplicationSpec struct {
...
+			}
+			if len(files) > 0 {
+				composeFound = true
+				break
+			}
+		}
+
+		if !composeFound {
+			continue
+		}
```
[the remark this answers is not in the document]
hexfusion: ifilepath.Glob is using readWriter.PathFor

internal/agent/device/applications/manager.go
```diff
@@ -173,34 +165,75 @@ func (m *manager) Shutdown(ctx context.Context, state shutdown.State) error {
...
+		allTargets = append(allTargets, cachedNestedTargets...)
+		return allTargets, nil
+	}
+
+	// Not fully cached, extract nested targets
+	nestedTargets, err := provider.CollectNestedOCITargets(ctx, m.log, m.podmanMonitor.client, m.readWriter, desired, secret)
 	if err != nil {
-		return nil, fmt.Errorf("collecting OCI targets: %w", err)
+		// collect on next iteration
+		m.log.Debugf("Nested target collection not ready yet: %v", err)
```
kkyrazis: should we introduce a retryable error here compared to hard failure? Until we're able to extract the labels from the image we should retry, but once we have the image fetched, any failures resulting from interaction with that image should bubble up an actual failure right?

hexfusion: good point will review

internal/agent/device/dependency/dependency.go
```diff
@@ -163,40 +169,60 @@ func (m *prefetchManager) isTargetsChanged(seenTargets map[string]struct{}) bool
...
-		targets, err := collector.CollectOCITargets(ctx, current, desired)
+		result, err := collector.CollectOCITargets(ctx, current, desired)
 		if err != nil {
 			return fmt.Errorf("prefetch collector %d failed: %w", i, err)
 		}
-		allTargets = append(allTargets, targets...)
+		allTargets = append(allTargets, result.Targets...)
+		if result.Requeue {
+			requeueNeeded = true
+		}
```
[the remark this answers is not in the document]
hexfusion: Keeping inline with golang error handling CollectOCITargets does not return nil result if err != nil

hexfusion:
reworking cache

internal/agent/device/applications/provider/provider.go
```diff
@@ -58,6 +56,306 @@ type ApplicationSpec struct {
...
+		readWriter,
+		imageSpec.Image,
+		appType,
+		pullSecret,
+	)
+	if err != nil {
+		return nil, nil, fmt.Errorf("extracting nested targets from %s: %w", imageSpec.Image, err)
+	}
+
+	return cachedAppData.Targets, cachedAppData, nil
```
hexfusion, line 356: @SiddarthR56 I think this is a bit wonky too ;) prob just need to return cachedAppData and let the caller use Targets

internal/agent/device/applications/provider/provider.go
```diff
@@ -284,6 +624,106 @@ func pathFromAppType(appType v1alpha1.AppType, name string, embedded bool) (stri
...
+			return nil, fmt.Errorf("copy image contents: %w (cleanup failed: %v)", err, rmErr)
+		}
+		return nil, fmt.Errorf("copy image contents: %w", err)
+	}
+
+	var targets []dependency.OCIPullTarget
+
+	switch appType {
+	case v1alpha1.AppTypeCompose:
+		// parse compose spec from tmpdir
```
hexfusion, line 677: parse and validate the images. we need to cleanly return the error identifying the app name involved.

PR #1925, opened 2025-11-07
NO-ISSUE: added harness helpers to use device simulator

Samudelacruz77:
Added functions:

- SetupDeviceSimulator
- RunDeviceSimulator
- StopDeviceSimulator

## Summary by CodeRabbit

* **New Features**
  * Added auth permissions API and public permission models; OpenShift and OAuth2 auth providers with OpenShift OAuth client/secret support.
  * Helm: OpenShift & multicluster extensions, per-component additional route/PVC labels, imagePullSecret support, UI on/off gating, gateway/gateway-port improvements, and cluster-scoped default roles.
* **Tests**
  * New device simulator e2e harness with auto-build/run/stop, agent config provisioning, resource cleanup, and fleet YAML generation/validation.
* **Documentation**
  * Expanded Helm README and simplified OpenShift auth guidance.

### DIFF
--- a/test/harness/e2e/harness_devicesim.go
+++ b/test/harness/e2e/harness_devicesim.go
@@ -255,59 +255,88 @@ func (h *Harness) SetupDeviceSimulatorAgentConfig(server string, logLevel string
 	return destConfigPath, nil
 }
 
-// DeleteAllResourcesFound deletes all fleets first and then all devices.
+// DeleteAllResourcesFound deletes only fleets and devices created by the current test (identified by test-id label).
 // Returns the names of deleted fleets and IDs of deleted devices, in order processed.
 // Partial results are returned even if an error occurs during deletion.
 func (h *Harness) DeleteAllResourcesFound() ([]string, []string, error) {
+	testID := h.GetTestIDFromContext()
+	logrus.Infof("Deleting test resources with test-id: %s", testID)
+
 	var deletedFleets []string
 	var deletedDevices []string
 	var deleteErr error
 
 	// Delete fleets first so they stop selecting devices
-	fleetsOut, err := h.CLI("get", "fleets", "-o", "name")
+	// Only get fleets with the test-id label
+	fleetsOut, err := h.CLI("get", util.Fleet, "-l", fmt.Sprintf("test-id=%s", testID), "-o", "name")
 	if err != nil {
-		return nil, nil, fmt.Errorf("listing fleets: %w", err)
+		// If no fleets found, that's fine
+		logrus.Debugf("No fleets found with test-id %s: %v", testID, err)
+	} else {
+		fleetsOut = strings.TrimSpace(fleetsOut)
+		if fleetsOut != "" {
+			// Parse fleet names from the output
+			// Output format: "fleet/name1\nfleet/name2\n..."
+			for _, line := range strings.Split(fleetsOut, "\n") {
+				name := strings.TrimSpace(line)
+				if name == "" {
+					continue
+				}
+				// Extract just the name part from "fleet/name"
+				name = strings.TrimPrefix(name, "fleet/")
+				if _, err := h.ManageResource("delete", "fleet/"+name); err != nil {
+					if deleteErr == nil {
+						deleteErr = fmt.Errorf("deleting fleet %s: %w", name, err)
+					} else {
+						deleteErr = fmt.Errorf("%v; deleting fleet %s: %w", deleteErr, name, err)
+					}
+					logrus.Warnf("failed to delete fleet %s: %v", name, err)
+					continue // Continue with remaining fleets
+				}
+				deletedFleets = append(deletedFleets, name)
+			}
+		}
 	}
 
-	for _, line := range strings.Split(strings.TrimSpace(fleetsOut), "\n") {
-		name := strings.TrimSpace(line)
-		if name == "" {
-			continue
-		}
-		name = strings.TrimPrefix(name, "fleet/")
-		if _, err := h.ManageResource("delete", "fleet/"+name); err != nil {
-			deleteErr = fmt.Errorf("deleting fleet %s: %w", name, err)
-			logrus.Warnf("failed to delete fleet %s: %v", name, err)
-			continue // Continue with remaining fleets
-		}
-		deletedFleets = append(deletedFleets, name)
-	}
-
-	// Then delete devices
-	devicesOut, err := h.CLI("get", "devices", "-o", "name")
+	// Then delete devices with the test-id label
+	devicesOut, err := h.CLI("get", util.Device, "-l", fmt.Sprintf("test-id=%s", testID), "-o", "name")
 	if err != nil {
 		if deleteErr != nil {
 			return deletedFleets, deletedDevices, fmt.Errorf("fleet deletion errors: %w; listing devices: %w", deleteErr, err)
 		}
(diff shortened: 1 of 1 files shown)


Samudelacruz77:
@coderabbit generate docstrings

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,361 @@
...
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+		if err := copyFile(filepath.Join(srcCertsDir, entry.Name()), filepath.Join(destCertsDir, entry.Name())); err != nil {
+			return "", fmt.Errorf("copying cert %s: %w", entry.Name(), err)
+		}
+	}
+
+	// Do not modify the generated/copied config file content here.
```
siserafin: is this comment wanted?

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,361 @@
...
+			return deleted, err
+		}
+		deleted = append(deleted, id)
+	}
+	return deleted, nil
+}
+
+// DeleteAllFleetsFound lists all fleets and deletes each one. It returns the
+// fleet names that were deleted, in the order processed.
+func (h *Harness) DeleteAllFleetsFound() ([]string, error) {
```
siserafin: DeleteAllFleetsFound() and DeleteAllDevicesFound() have the same flow..maybe they could be unified?

Samudelacruz77: sure, will do

test/harness/e2e/harness_devicesim.go
Samudelacruz77: will delete it

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+
+	logrus.Infof("device simulator agent config prepared at %s; certs in %s", destConfigPath, destCertsDir)
+	return destConfigPath, nil
+}
+
+// DeleteAllResourcesFound deletes all fleets first and then all devices.
+// Returns the names of deleted fleets and IDs of deleted devices, in order processed.
+func (h *Harness) DeleteAllResourcesFound() ([]string, []string, error) {
```
amalykhi, line 261: there are DeleteDevices(), DeleteFleets() and DeleteAll() functions in test/e2e/resources/resource_factory.go
Do you think it could be used here? Or may be modify them to return the list if necessary to avoid duplication?

Samudelacruz77: Nice, I wasnt aware, will use deleteAll() instead of my homemade functions

Samudelacruz77: Tried it but we can’t call test/e2e/resources/resource_factory.go’s DeleteAll from test/harness/e2e without creating an import cycle: resources already imports harness/e2e, so importing resources back inside harness/e2e forms a loop and gives a lint error

amalykhi: hmmm. they are helper functions using harness. I think they should be under   harness/e2e. WDYT @Samudelacruz77 ?

Samudelacruz77:
@coderabbitai generate docstrings

test/harness/e2e/harness_devicesim.go
siserafin: h.DeleteAll() doesn't work?

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,353 @@
...
+	ginkgo "github.com/onsi/ginkgo/v2"
+	"github.com/sirupsen/logrus"
+)
+
+// EnsureDeviceSimulatorBinary returns the path to the devicesimulator binary, building it if missing.
+func (h *Harness) EnsureDeviceSimulatorBinary() (string, error) {
+	// Compute the expected path under the repo top-level bin directory
+	expectedPath := filepath.Join(util.GetTopLevelDir(), "bin", "devicesimulator")
+
+	if st, err := os.Stat(expectedPath); err == nil && !st.IsDir() {
```
itzikb-redhat, line 24: Is it worth checking that it's executable?

test/harness/e2e/harness_devicesim.go
itzikb-redhat: A question: Should we use delete all resources? What about deleting just resources created by a test?

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	var b strings.Builder
+	for i := 0; i < fleetCount; i++ {
+		if i > 0 {
+			b.WriteString("\n---\n")
+		}
+		fleetName := fmt.Sprintf("%s-%02d", fleetBaseName, i)
+		// Minimal valid Fleet spec mirroring example, with selector/labels and an annotation to record devices-per-fleet
+		b.WriteString("apiVersion: flightctl.io/v1alpha1\n")
```
itzikb-redhat, line 369: What do you think about using a template?

Samudelacruz77: Yes we can use a template, for the deletion is better to delete all between tests, have in mind that this is not a real device or VM is a process simulating N devices so you cannot really interact with the devices, I think is better to delete all resources that get generated in the test to continue next testing with a pristine enviroment, like we do with VMs

test/harness/e2e/harness_devicesim.go
Samudelacruz77: Yes, will add it, thanks

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,353 @@
...
+		b.WriteString("            repository: flightctl-demos\n")
+		b.WriteString("            targetRevision: main\n")
+		b.WriteString("            path: /demos/basic-nginx-demo/configuration/\n")
+	}
+	return b.String(), nil
+}
+
+// ValidateFleetYAMLDevicesPerFleet performs a basic consistency check on the generated YAML,
+// ensuring it contains the expected number of Fleet documents and the devices-per-fleet annotation.
+func (h *Harness) ValidateFleetYAMLDevicesPerFleet(yamlContent string, expectedDevicesPerFleet int, expectedFleetCount int) error {
```
eldar101, line 399: isn't there some API to use to validate yamls?

test/harness/e2e/harness_devicesim.go
```diff
@@ -0,0 +1,353 @@
...
+	if _, ok := cmd.Stdout.(*os.File); !ok { // only attempt stderr fallback if stdout wasn't handled as file
+		if c, ok := cmd.Stderr.(interface{ Close() error }); ok {
+			_ = c.Close()
+		}
+	}
+
+	return stopErr
+}
+
+func copyFile(from, to string) error {
```
eldar101, line 142: Is this func related specificially to device simulator? i think generic ones cant be in another file

Samudelacruz77: Yes is generic, I will add to harness.go instead

siserafin: Or maybe to the utils

test/harness/e2e/harness_devicesim.go
Samudelacruz77: Will try to delete only the things the test creates better, what do you think?

PR #1929, opened 2025-11-07
EDM-2445: Add Artifact provider for Quadlet Apps

SiddarthR56:
Applied Spec:
```
spec:
  applications:
  - appType: quadlet
    image: 192.168.1.36:5000/sq-tar:v2
    name: my-app-2
```

Device Status:
```
status:
  applications:
  - name: my-app-2
    ready: 2/2
    restarts: 0
    status: Running
  applicationsSummary:
    info: Device's application workloads are healthy.
    status: Healthy
```

## Summary by CodeRabbit

* **New Features**
  * Support for OCI artifacts alongside container images: artifact digest & annotation inspection, OCI-type detection, and app-name resolution for image/artifact providers (requires Podman artifact support).

* **Bug Fixes / Validation**
  * Secure tar/gzip unpacking with path-traversal protection; stricter quadlet top-level validation; improved caching and requeue behavior when targets are missing.

* **Tests**
  * Expanded and updated tests/mocks for artifact/image flows, unpacking, quadlet checks, and command invocation patterns.

* **Documentation**
  * API docs updated to reference OCI artifacts alongside images.

* **Refactor**
  * Command invocation patterns standardized across code and tests.


api/v1alpha1/openapi.yaml
```diff
@@ -4461,8 +4461,14 @@ components:
             image:
               type: string
               description: Reference to the container image for the application package.
-          required:
-            - image
+            artifact:
```
hexfusion: we agreed to keep this image and discover the type with magic :)

api/v1alpha1/util.go
```diff
@@ -50,8 +50,9 @@ const (
 type ApplicationProviderType string
 
 const (
-	ImageApplicationProviderType  ApplicationProviderType = "image"
-	InlineApplicationProviderType ApplicationProviderType = "inline"
+	ImageApplicationProviderType    ApplicationProviderType = "image"
+	ArtifactApplicationProviderType ApplicationProviderType = "artifact"
```
celdrake: I also understood the same as previous  @hexfusion comment that we wouldn't have a new "artifact" type.
In fact it was added by @kkyrazis and dropped a bit later.

kkyrazis: We won't. Until we have auto discovery enabled this could be used for testing, but shouldn't be merged.

internal/agent/device/applications/provider/image.go
```diff
@@ -83,27 +89,49 @@ func (p *imageProvider) Verify(ctx context.Context) error {
...
 	}
 
+	ociType, err := detectOCIType(ctx, p.podman, p.spec.ImageProvider.Image)
+	if err != nil {
+		return fmt.Errorf("detecting OCI type: %w", err)
+	}
+
+	// Enforce OCI type and appType compatibility
+	p.isArtifact = ociType == dependency.OCITypeArtifact
+	if p.isArtifact && p.spec.AppType != v1alpha1.AppTypeQuadlet {
```
kkyrazis: With the original plan being that artifacts would be keyed off of `artifact` and images off of `image` it made more sense to restrict the app types as those would be processed by separate providers. Now that they've been bundled into the same provider, I think these checks are unnecessary. At this point the artifact/image is just the way to package the app, and we have to handle unpackaging it in both cases, so why not just allow any type?

internal/agent/device/applications/provider/provider.go
```diff
@@ -87,7 +87,7 @@ func CollectBaseOCITargets(
 			// Add base image with PullIfNotPresent policy
 			policy := v1alpha1.PullIfNotPresent
 			targets = append(targets, dependency.OCIPullTarget{
-				Type:       dependency.OCITypeAuto,
+				Type:       dependency.OCITypeImage,
```
kkyrazis: Why did this change back? If it's an image provider it could be an artifact or an image that we're pulling

internal/agent/device/applications/provider/provider.go
```diff
@@ -764,6 +778,85 @@ func ensureQuadlet(readWriter fileio.ReadWriter, appPath string) error {
...
+			if _, ok := common.SupportedQuadletExtensions[ext]; ok {
+				topLevelQuadletFiles = append(topLevelQuadletFiles, entry.Name())
+			}
+		}
+	}
+
+	// Validation rules:
+	// 1. Must have at least one quadlet file at top level OR in a single subdirectory
+	// 2. Cannot have quadlet files in multiple subdirectories
+	if len(subdirectoriesWithQuadlets) > 1 {
```
kkyrazis: lets make this 0.

The app packaging structure should be:

 ```
app.container
db.container
db.volume
app.volume
....
container.d/10-my-dropin.conf
other/config/stuff
```

File: internal/agent/device/fileio/writer.go [hunk 6]
```diff
@@ -668,3 +670,66 @@ func AppendFile(w Writer, name string, data []byte, perm fs.FileMode, opts ...Fi
 	_, err = file.Write(data)
 	return err
 }
+
+// UnpackTar unpacks a tar or tar.gz file to the destination directory.
+func UnpackTar(rw ReadWriter, tarPath, destDir string) error {
```
kkyrazis: is there a reason this isn't just 
```
rw.UnpackTar(tarPath, destDir)
```

internal/agent/device/fileio/writer.go
```diff
@@ -668,3 +670,66 @@ func AppendFile(w Writer, name string, data []byte, perm fs.FileMode, opts ...Fi
...
+
+	// Read the tar file
+	tarFile, err := rw.ReadFile(tarPath)
+	if err != nil {
+		return fmt.Errorf("reading tar file: %w", err)
+	}
+
+	var tarReader *tar.Reader
+	if isGzipped {
+		gzr, err := gzip.NewReader(strings.NewReader(string(tarFile)))
```
kkyrazis:
```suggestion
		gzr, err := gzip.NewReader(bytes.NewReader(tarFile))
```

internal/agent/device/fileio/writer.go
```diff
@@ -668,3 +670,66 @@ func AppendFile(w Writer, name string, data []byte, perm fs.FileMode, opts ...Fi
...
+		if err != nil {
+			return fmt.Errorf("creating gzip reader: %w", err)
+		}
+		defer gzr.Close()
+		tarReader = tar.NewReader(gzr)
+	} else {
+		tarReader = tar.NewReader(strings.NewReader(string(tarFile)))
```
kkyrazis:
```suggestion
		tarReader = tar.NewReader(bytes.NewReader(tarFile))
```

internal/agent/device/fileio/writer.go
kkyrazis: can a test be added for this?

internal/agent/device/applications/provider/image.go
SiddarthR56: I  kept the restriction as an early safety net, but you’re right—now that both formats share the same provider, failing early doesn’t buy us much. We still run the compose/quadlet validation after unpacking, so it’s actually better to let any packaging through and surface the concrete validation error if the contents don’t match the declared appType

internal/agent/device/applications/provider/provider.go
```diff
@@ -764,6 +778,74 @@ func ensureQuadlet(readWriter fileio.ReadWriter, appPath string) error {
...
+			errors.ErrInvalidSpec, subdirectoriesWithQuadlets)
+	}
+
+	if len(topLevelQuadletFiles) == 0 {
+		return fmt.Errorf("%w: no valid quadlet files found at top level of artifact", errors.ErrNoQuadletFile)
+	}
+
+	// Validate top-level quadlet files
+	return ensureQuadlet(readWriter, appPath)
+}
```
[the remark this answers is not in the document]
SiddarthR56: ensureQuadletArtifact is only called for packaged quadlet applications (OCI images/artifacts) where the top-level restriction ensures a consistent flat structure. Other quadlet providers (inline, embedded) use the more permissive ensureQuadlet function

internal/agent/client/podman.go
```diff
@@ -243,6 +256,60 @@ func (p *Podman) ArtifactExists(ctx context.Context, artifact string) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return annotations, nil
+}
+
+// extractArtifactAnnotations parses the podman artifact inspect JSON output and extracts annotations
+func extractArtifactAnnotations(stdout string) (map[string]string, error) {
+	var inspectResults []ArtifactInspect
```
kkyrazis: is this structure correct? Running locally I get something more similar to:

```json
kkyrazis:~/projects/flightctl (single-container-gen)$ podman artifact inspect localhost:5000/my-quadlet-raw-artifact:v1
{
     "Manifest": {
          "schemaVersion": 2,
          "mediaType": "application/vnd.oci.image.manifest.v1+json",
          "config": {
               "mediaType": "application/vnd.oci.empty.v1+json",
               "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
               "size": 2,
               "data": "e30="
          },
          "layers": [
               {
                    "mediaType": "text/plain; charset=utf-8",
                    "digest": "sha256:c559c6e429bbbf7bfb0f8e61583242a1fce5c8bac334443fd2a2ebd5387653da",
                    "size": 185,
                    "annotations": {
                         "appType": "quadlet",
                         "org.opencontainers.image.title": "nginx.container"
                    }
               },
               {
                    "mediaType": "text/plain; charset=utf-8",
                    "digest": "sha256:50e2f1db2a392035ff86306fb972f8389308aced5cda4333e1b95821f7e0ad48",
                    "size": 25,
                    "annotations": {
                         "org.opencontainers.image.title": "data.volume"
                    }
               },
               {
                    "mediaType": "text/plain; charset=utf-8",
                    "digest": "sha256:bd235f79040efed11ba7224cda135c7e22164c4d50a0e58332ca7b1e08dcb255",
                    "size": 26,
                    "annotations": {
                         "org.opencontainers.image.title": "app.network"
                    }
               }
          ]
     },
     "Name": "localhost:5000/my-quadlet-raw-artifact:v1",
     "Digest": "sha256:980f25021be167f5f4cd2514ef4faa104012f2072c7d8840a8752b9ab9ab5856"
}
```

which would require us to dig into the layers to get the annotations.

internal/agent/client/podman.go
```diff
@@ -243,6 +256,60 @@ func (p *Podman) ArtifactExists(ctx context.Context, artifact string) bool {
...
+	if err := json.Unmarshal([]byte(stdout), &inspectResults); err != nil {
+		return nil, fmt.Errorf("unmarshal artifact inspect output: %w", err)
+	}
+
+	if len(inspectResults) == 0 {
+		return nil, fmt.Errorf("empty artifact inspect results")
+	}
+
+	return inspectResults[0].Annotations, nil
```
kkyrazis: I am not against a returning the first set of annotations, but we'd have to document it. I'd prefer if we could return all sets of annotations though.

internal/agent/device/fileio/writer.go
```diff
@@ -668,3 +670,74 @@ func AppendFile(w Writer, name string, data []byte, perm fs.FileMode, opts ...Fi
...
+func (rw *readWriter) UnpackTar(tarPath, destDir string) error {
+	// Open the tar file for streaming
+	file, err := os.Open(rw.PathFor(tarPath))
+	if err != nil {
+		return fmt.Errorf("opening tar file: %w", err)
+	}
+	defer file.Close()
+
+	var tarReader *tar.Reader
+	if strings.HasSuffix(tarPath, ".gz") {
```
kkyrazis: the provider allows `.tar.gz` and `.tgz` but this only handles `.gz`

internal/agent/device/applications/provider/image.go
```diff
@@ -26,6 +29,9 @@ type imageProvider struct {
 
 	// AppData stores the extracted app data from OCITargets to reuse in Verify
 	AppData *AppData
+
+	// isArtifact indicates whether the image reference is an OCI artifact
+	isArtifact bool
 }
```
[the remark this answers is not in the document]
SiddarthR56: Verify() is invoked during FromDeviceSpec() (via WithVerify()), so any provider that fails verification is rejected before sync continues. Only verified providers are returned to syncProviders(), where Ensure()/Update() eventually call Install(). This guarantees Install() runs only on providers that have already passed verification.

internal/agent/device/applications/provider/image.go
```diff
@@ -26,6 +29,9 @@ type imageProvider struct {
 
 	// AppData stores the extracted app data from OCITargets to reuse in Verify
 	AppData *AppData
+
+	// isArtifact indicates whether the image reference is an OCI artifact
+	isArtifact bool
```
hexfusion: is this bool something dependency manager might be interested in?

internal/agent/device/applications/provider/image.go
```diff
@@ -137,6 +158,14 @@ func (p *imageProvider) Verify(ctx context.Context) error {
 		}
 	}()
 
+	// For artifacts, use stricter validation
+	if p.isArtifact && p.spec.AppType == v1alpha1.AppTypeQuadlet {
```
hexfusion: maybe someday we could have a helper. not blocking this PR
```go
	if p.isArtifact && p.spec.AppType.IsQuadlet() {
```

internal/agent/device/applications/provider/image_test.go
```diff
@@ -235,3 +258,7 @@ func mockPodmanInspect(labels map[string]string) []client.PodmanInspect {
 	}
 	return []client.PodmanInspect{inspect}
 }
+
+func TestImageProvider_ArtifactSupport(t *testing.T) {
+	t.Skip("artifact support covered by integration tests")
+}
```
hexfusion: do we need this?

internal/agent/device/fileio/fileio.go
```diff
@@ -63,6 +63,7 @@ type Reader interface {
 type ReadWriter interface {
 	Reader
 	Writer
+	UnpackTar(tarPath, destDir string) error
```
hexfusion: please remove from interface and create  a function for this. if it needs information pass via signature

internal/agent/device/applications/provider/image.go
```diff
@@ -137,6 +153,14 @@ func (p *imageProvider) Verify(ctx context.Context) error {
 		}
 	}()
 
+	// For artifacts, use stricter validation
+	if ociType == dependency.OCITypeArtifact && p.spec.AppType == v1alpha1.AppTypeQuadlet {
```
kkyrazis: I'm not sure I understand this check. Should the Verifys in the new handlers just always do this? My expectation at the end of the day would be that all formats: image, artifact, inline, and embedded would always setup a directory in the exact same way

i.e.
```
quadlet.container
quadlet.image
...
path/to/conf.conf
container.d/1-dropin.conf
```

What specifically about artifacts makes it need additional validation? I think the extra validations you added are more comprehensive and probably should just be applied to all types.

kkyrazis (approved):

PR #1931, opened 2025-11-09
EDM-2347: multi auth providers

asafbennatan:
1. separated auth and authz
2. added dynamic oidc/oauth providers
3. removed org extraction , auth providers must provide a full identity including roles and orgs
4. mapped identity - an identity object is mapped by the identity mapper into internal ids (if relevant)

## Summary by CodeRabbit

* **New Features**
  * Full AuthProvider CRUD API and runtime multi-provider auth (OIDC/OAuth2/K8s/AAP); dynamic provider loading and per-provider config surfaced.

* **Access Control**
  * New RBAC roles (admin, operator, viewer, installer) and flexible org/role assignment modes: static, dynamic, per-user.

* **Configuration**
  * Expanded OIDC config: clientId, enabled, issuer, organizationAssignment, usernameClaim (list), roleAssignment; default-org RBAC manifest added.

* **Security**
  * Certificate-based agent/enrollment auth, identity-to-DB mapping, provider secret redaction, and tokens now carry roles and organizations.


api/v1alpha1/openapi.yaml
```diff
@@ -144,6 +146,9 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  # NOTE: OIDC issuer endpoints (/auth/authorize, /auth/login, /auth/token, 
+  # /auth/userinfo, /auth/jwks, /.well-known/openid-configuration) have been 
+  # moved to the separate flightctl-pam-issuer service. See api/v1alpha1/pam-issuer/openapi.yaml
```
fzdarsky, line 151: Remove, as those endpoints weren't part of a release yet.

api/v1alpha1/openapi.yaml
```diff
@@ -5348,29 +5730,68 @@ components:
...
+          items:
+            $ref: '#/components/schemas/AuthProviderInfo'
+        defaultProvider:
           type: string
-          description: Auth type.
-        authURL:
+          description: Name of the default authentication provider.
+        organizationsEnabled:
+          type: boolean
+          description: Whether organizations are enabled for authentication.
```
fzdarsky, line 5836: Is there a reason _not_ to have orgs? IOW, do we need to have this configurable?

api/v1alpha1/openapi.yaml
```diff
@@ -5348,29 +5730,68 @@ components:
...
-          $ref: '#/components/schemas/AuthOrganizationsConfig'
+          description: Unique identifier for the provider.
+        displayName:
+          type: string
+          description: Human-readable display name.
+        type:
+          type: string
+          enum: [oidc, k8s, aap, oauth2]
+          description: Type of authentication provider.
+        issuer:
```
celdrake: `issuer` is still unconsistent for Oauth2 providers

- Upon creation: it's optional
- Upon PATCHING: once the value is set, it cannot be cleared.

```
[
  {
    "op": "remove",
    "path": "/spec/issuer"
  }
]

```
Error 400: Bad Request: request body has an error: doesn't match schema #/components/schemas/AuthProvider: Error at "/spec": doesn't match schema due to: Error at "/issuer": property "issuer" is missing Schema: { "description": "OAuth2ProviderSpec describes an OAuth2 provider configuration.", "properties": { "authorizationUrl": { "description": "The OAuth2 authorization endpoint URL.", "type": "string" }, "clientId": { "description": "The OAuth2 client ID.", "type": "string" }, "clientSecret": { "description": "The OAuth2 client secret.", "type": "string", "writeOnly": true }, "displayName": { "description": "Human-readable display name for the provider.", "type": "string" }, "enabled": { "default": true, "description": "Whether this OAuth2 provider is enabled.", "type": "boolean" }, "issuer": { "description": "The OAuth2 issuer identifier (used for issuer identification in tokens).", "type": "string" }, "organizationAssignment": { "$ref": "#/components/schemas/AuthOrganizationAssignment" }, "providerType": { "description": "The type of authentication provider.", "enum": [ "oauth2" ], "type": "string" }, "roleAssignment": { "$ref": "#/components/schemas/AuthRoleAssignment" }, "scopes": { "description": "List of OAuth2 scopes to request.", "items": { "type": "string" }, "type": "array" }, "tokenUrl": { "description": "The OAuth2 token endpoint URL.", "type": "string" }, "userinfoUrl": { "description": "The OAuth2 userinfo endpoint URL.", "type": "string" }, "usernameClaim": { "default": [ "preferred_username" ], "description": "JSON path to the username claim in the userinfo response as an array of path segments (e.g., [\"preferred_username\"], [\"email\"], [\"sub\"]).", "items": { "type": "string" }, "type": "array" } }, "required": [ "providerType", "issuer", "authorizationUrl", "tokenUrl", "userinfoUrl", "clientId", "clientSecret", "organizationAssignment", "roleAssignment" ], "type": "object" } Value: { "authorizationUrl": "https://github.com/login/oauth/authorize", "clientId": "CLIENT-ID", "clientSecret": "WHATEVER", "displayName": "github no issuer", "enabled": true, "organizationAssignment": { "organizationName": "default", "type": "static" }, "providerType": "oauth2", "roleAssignment": { "roles": [ "admin" ], "type": "static" }, "scopes": [ "read:user", "user:email" ], "tokenUrl": "https://github.com/login/oauth/access_token", "userinfoUrl": "https://api.github.com/user", "usernameClaim": [ "login" ] } 
```
```

(I redacted the secret as it's actually shown, I opened a separate thread for that)

api/v1alpha1/openapi.yaml
```diff
@@ -3639,12 +3644,389 @@ paths:
...
+              $ref: '#/components/schemas/PatchRequest'
+        required: true
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AuthProvider'
+        "400":
```
celdrake, line 3916: Patching a Provider with a BadRequest prints the entire provider details, including the `clientSecret`

celdrake (changes requested):

asafbennatan:
@coderabbitai pause

internal/auth/authn/jwt_generator.go
```diff
@@ -478,5 +484,33 @@ func (g *JWTGenerator) ValidateToken(tokenString string) (*JWTIdentity, error) {
...
+		if rolesList, ok := roles.([]interface{}); ok {
+			roleStrings := make([]string, 0, len(rolesList))
+			for _, role := range rolesList {
+				if roleStr, ok := role.(string); ok {
+					roleStrings = append(roleStrings, roleStr)
+				}
+			}
+			identity.SetRoles(roleStrings)
+		}
+	}
```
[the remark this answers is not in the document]
asafbennatan: not relevant we are using this only in the pam issuer , at that point we know what we return

api/v1alpha1/openapi.yaml

internal/service/identity_mapper.go
```diff
@@ -0,0 +1,192 @@
...
+
+// Start starts the cache background cleanup
+func (m *IdentityMapper) Start() {
+	m.cache.Start()
+}
+
+// Stop stops the cache background cleanup
+func (m *IdentityMapper) Stop() {
+	m.cache.Stop()
+}
```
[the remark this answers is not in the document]
asafbennatan: when we run this we use go xxx.start

api/v1alpha1/openapi.yaml
```diff
@@ -5348,29 +5730,68 @@ components:
...
+          description: OAuth2 scopes (for OAuth2 providers).
+        usernameClaim:
+          type: array
+          items:
+            type: string
+          description: JSON path to the username claim (for OIDC/OAuth2 providers) as an array of path segments.
+        isDefault:
+          type: boolean
+          description: Whether this is the default provider.
+          default: false
```
fzdarsky: Duplicate with `defaultProvider` in AuthConfig? Would allow setting multiple defaults?

api/v1alpha1/openapi.yaml
```diff
@@ -5348,29 +5730,68 @@ components:
...
+        name:
           type: string
-          description: Auth URL.
-        authOrganizationsConfig:
-          $ref: '#/components/schemas/AuthOrganizationsConfig'
+          description: Unique identifier for the provider.
+        displayName:
+          type: string
+          description: Human-readable display name.
+        type:
```
fzdarsky: Does not match name in `required` property (`authType`). Also, I think you've called it `providerType` in the `AuthProviderSpec`. It's not used as discriminator here, though, and instead mixes OIDC and OAuth2 properties, which are then cannot be marked as required, of course. Can this be better aligned / deduplicated with the AuthProviderSpec?

api/v1alpha1/openapi.yaml
```diff
@@ -5348,29 +5730,68 @@ components:
     AuthConfig:
       type: object
       properties:
-        authType:
+        providers:
+          type: array
+          description: List of all available authentication providers.
+          items:
+            $ref: '#/components/schemas/AuthProviderInfo'
+        defaultProvider:
```
fzdarsky, line 5831: We're breaking the API here. In particular as we cannot exclude the need for further changes in the future, we should probably version the new resource.

internal/api_server/middleware/agent_auth.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Start starts the cache background cleanup
+func (m *AgentAuthMiddleware) Start() {
+	m.cache.Start()
+}
+
+// Stop stops the cache background cleanup
+func (m *AgentAuthMiddleware) Stop() {
+	m.cache.Stop()
```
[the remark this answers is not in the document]
asafbennatan: we start agent auth middleware as goroutine

api/v1alpha1/validation.go
```diff
@@ -1246,3 +1249,309 @@ func hasPathField(rawJSON []byte) bool {
...
+
+	// Validate that all roles are non-empty strings
+	for i, role := range a.Roles {
+		if role == "" {
+			allErrs = append(allErrs, fmt.Errorf("role at index %d cannot be empty", i))
+		}
+	}
+
+	return allErrs
+}
```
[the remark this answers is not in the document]
asafbennatan: not anyone can create authproviders anyway...

api/v1alpha1/openapi.yaml
asafbennatan: no there is no - there are several places where this needs to be removed , question is should we tackle this now

api/v1alpha1/openapi.yaml
asafbennatan: done

api/v1alpha1/openapi.yaml
asafbennatan: removed isDefault

api/v1alpha1/openapi.yaml
asafbennatan: the problem is that we are getting this info from two places , the first one is from the config 
the second one is from the DB 
we could remove the config objects infavour of the DB objects and then replace AuthProviderInfo ( but it will require that we add aap and k8 auth spec to the api)
thoughts?

api/v1alpha1/openapi.yaml
asafbennatan: added

api/v1alpha1/openapi.yaml
```diff
@@ -5348,29 +5713,64 @@ components:
...
+          type: string
+          description: Name of the default authentication provider.
+        organizationsEnabled:
+          type: boolean
+          description: Whether organizations are enabled for authentication.
+    AuthProviderInfo:
```
ori-amizur: There is duplication between AuthProviderInfo and AuthProvider.Spec (for example OAuth2ProviderSpec).  Why do we need both ?

api/v1alpha1/roles.go
```diff
@@ -0,0 +1,9 @@
+package v1alpha1
+
+// Role constants for flightctl user roles
+const (
```
ori-amizur: These are constants.  Any reason why not to put them in consts.go ?

api/v1alpha1/openapi.yaml
asafbennatan: see the comment above from frank , the problem is that we have the data coming from the config and db each with different structure so this is another type which is filled from the other two

api/v1alpha1/roles.go
asafbennatan: done

api/v1alpha1/openapi.yaml
ori-amizur: > the problem is that we have the data coming from the config and db each with different structure so this is another type which is filled from the other two

So does it mean that the API structures are dictated by the fact that the data arrives from different sources ? This doesn't seem right.  Usually it should be the other way around.

internal/store/authprovider.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/google/uuid"
+	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+type AuthProvider interface {
+	InitialMigration(ctx context.Context) error
+
+	Create(ctx context.Context, orgId uuid.UUID, authProvider *api.AuthProvider, eventCallback EventCallback) (*api.AuthProvider, error)
```
hexfusion: question: does making the auth provider org scoped cause a circular dependency? caller needs an orgId to find  the provider to authenticate to then determine orgId.

api/v1alpha1/openapi.yaml
fzdarsky: I'd then lean towards removing, but let's check in channel and open a separate ticket for that.

api/v1alpha1/openapi.yaml
asafbennatan: have made the change , the problem was that it required a bit more changes to existing structures- but it was worth it

api/v1alpha1/openapi.yaml
asafbennatan: added

api/v1alpha1/openapi.yaml
```diff
@@ -6175,3 +6533,339 @@ components:
...
+      required:
+        - providerType
+        - authorizationUrl
+        - tokenUrl
+        - userinfoUrl
+        - clientId
+        - clientSecret
+        - organizationAssignment
+        - roleAssignment
+    AAPProviderSpec:
```
ori-amizur: Isn't AAP just oauth ?  If not, what is the difference ?

internal/api_server/middleware/agent_auth.go
```diff
@@ -0,0 +1,240 @@
...
+		// Extract organization ID from certificate extension
+		orgID, present, err := signer.GetOrgIDExtensionFromCert(peerCertificate)
+		if err != nil {
+			m.log.Errorf("Failed to extract organization ID from certificate: %v", err)
+			http.Error(w, fmt.Sprintf("Failed to extract organization ID: %v", err), http.StatusUnauthorized)
+			return
+		}
+
+		// Use default org ID if not present in certificate
+		if !present {
```
ori-amizur, line 117: Maybe I am missing something, but there should be some authentication logic here, but the only verification is that values can be extracted and the signer name is expected. Is this enough for authentication?  Where is the part that "asks" that this agent is actually authenticated? Do we verify the agent certificate signature?

api/v1alpha1/validation.go
```diff
@@ -1246,3 +1249,313 @@ func hasPathField(rawJSON []byte) bool {
...
+
+func (a AuthDynamicOrganizationAssignment) Validate(ctx context.Context) []error {
+	allErrs := []error{}
+
+	if len(a.ClaimPath) == 0 {
+		allErrs = append(allErrs, errors.New("claimPath is required for dynamic assignment"))
```
hexfusion: optional: is this worth adding as a const to allow unwrap validation in tests etc.  errors.Is(err, ErrSomething..)

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -130,66 +84,69 @@ func (p *AlertmanagerProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	// Check if user has permission to access alerts
-	allowed, err := p.authZ.CheckPermission(ctx, alertsResource, getAction)
-	if err != nil {
-		p.log.WithError(err).Error("Authorization check failed")
-		http.Error(w, "Authorization service unavailable", http.StatusServiceUnavailable)
-		return
-	}
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Skip auth for health and status endpoints
+			if r.URL.Path == healthPath || r.URL.Path == statusPath {
```
hexfusion, line 142: question: does this open up an attack surface? Is this a concern?

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -282,42 +239,55 @@ func main() {
...
+	authZ, err := auth.InitMultiAuthZ(cfg, logger)
 	if err != nil {
-		logger.Fatalf("Failed to initialize auth: %v", err)
+		logger.Fatalf("Failed to initialize authZ: %v", err)
 	}
 
+	// Create identity mapper for mapping identities to database objects
+	identityMapper := service.NewIdentityMapper(dataStore, logger)
+	go identityMapper.Start()
+	defer identityMapper.Stop()
```
hexfusion, line 278: I know you didnt start this patten but it feels like we should standardize on ctx, cancel context.WithCancel.. as the coordinated shutdown mechanism because if for any reason any of these goroutines should return should that  also be fatal for the process?

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -282,42 +239,55 @@ func main() {
...
-	proxy, err := NewAlertmanagerProxy(cfg, logger, authN, authZ)
+	proxy, err := NewAlertmanagerProxy(cfg, logger)
 	if err != nil {
 		logger.Fatalf("Failed to create alertmanager proxy: %v", err)
 	}
 
-	if proxy.authDisabled {
+	// Check if auth is disabled
+	authDisabled := false
+	value, exists := os.LookupEnv(auth.DisableAuthEnvKey)
```
hexfusion, line 293: disabling auth via ENV has tradeoffs, in terms of auditing would it be safer to enforce via config?

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -328,7 +298,7 @@ func main() {
 	router.Use(func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			// Skip logging for health checks to reduce noise
-			if r.URL.Path == "/health" {
+			if r.URL.Path == healthPath {
```
hexfusion, line 320: :+1:

internal/api_server/agentserver/server.go
```diff
@@ -208,21 +219,74 @@ func (s *AgentServer) prepareHTTPHandler(serviceHandler service.Service) (http.H
 		ErrorHandler: oapiErrorHandler,
 	}
 
+	// Create agent authentication middleware for device operations
+	agentAuthMiddleware := fcmiddleware.NewAgentAuthMiddleware(s.ca, s.log)
+	go agentAuthMiddleware.Start()
```
hexfusion, line 224: same comment as above how are we ensuring coordinated shutdown of these goroutines. since Run() has a context can we pass that to prepareHTTPHandler?

internal/api_server/middleware/agent_auth.go
```diff
@@ -0,0 +1,240 @@
...
+type AgentAuthMiddleware struct {
+	ca    *crypto.CAClient
+	log   logrus.FieldLogger
+	cache *ttlcache.Cache[string, *AgentIdentity]
+}
+
+// NewAgentAuthMiddleware creates a new device agent authentication middleware
+func NewAgentAuthMiddleware(ca *crypto.CAClient, log logrus.FieldLogger) *AgentAuthMiddleware {
+	cache := ttlcache.New[string, *AgentIdentity](
+		ttlcache.WithTTL[string, *AgentIdentity](10 * time.Minute),
```
hexfusion, line 31: const?

internal/api_server/middleware/agent_auth.go
```diff
@@ -0,0 +1,240 @@
...
+			m.log.Warnf("unexpected client certificate signer: expected %q, got %q", m.ca.Cfg.DeviceEnrollmentSignerName, s.Name())
+			http.Error(w, fmt.Sprintf("unexpected client certificate signer: expected %q, got %q", m.ca.Cfg.DeviceEnrollmentSignerName, s.Name()), http.StatusUnauthorized)
+			return
+		}
+
+		// Get peer certificate using the same pattern as handler.go
+		peerCertificate, err := m.ca.PeerCertificateFromCtx(ctx)
+		if err != nil {
+			m.log.Warnf("Agent certificate validation failed: %v", err)
+			http.Error(w, fmt.Sprintf("Certificate validation failed: %v", err), http.StatusUnauthorized)
```
hexfusion, line 96: nit: should this log be error?

internal/api_server/middleware/enrollment_auth.go
```diff
@@ -0,0 +1,233 @@
...
+		signer := m.ca.PeerCertificateSignerFromCtx(ctx)
+
+		got := "<nil>"
+		if signer != nil {
+			got = signer.Name()
+		}
+
+		if signer == nil || signer.Name() != m.ca.Cfg.ClientBootstrapSignerName {
+			m.log.Warnf("unexpected client certificate signer: expected %q, got %q", m.ca.Cfg.ClientBootstrapSignerName, got)
+			http.Error(w, fmt.Sprintf("unexpected client certificate signer: expected %q, got %q", m.ca.Cfg.ClientBootstrapSignerName, got), http.StatusUnauthorized)
```
hexfusion, line 94: verify log.Warn vs Error

internal/api_server/middleware/enrollment_auth.go
```diff
@@ -0,0 +1,233 @@
...
+			return
+		}
+
+		// Get peer certificate using the same pattern as handler.go
+		peerCertificate, err := m.ca.PeerCertificateFromCtx(ctx)
+		if err != nil {
+			m.log.Warnf("Enrollment certificate validation failed: %v", err)
```
hexfusion, line 101: verify log.Warn vs Error

internal/api_server/middleware/middleware.go
```diff
@@ -67,35 +63,36 @@ func AddEventMetadataToCtx(next http.Handler) http.Handler {
...
 				return
 			}
-
-			// Validate the organization ID
-			if err := resolver.EnsureExists(ctx, orgID); err != nil {
-				if errors.Is(err, flterrors.ErrResourceNotFound) {
-					http.Error(w, fmt.Sprintf("Organization not found: %s", orgID), http.StatusNotFound)
-					return
+			// Log the extracted organization ID
+			fmt.Printf("ExtractOrgIDToCtx: extracted orgID=%s from request\n", orgID.String())
```
DakCrowder: Should we pass a logger here rather than using fmt?

internal/auth/auth.go
```diff
@@ -65,100 +100,89 @@ func initK8sAuth(cfg *config.Config, log logrus.FieldLogger) (AuthNMiddleware, A
...
+		configuredAuthType = AuthTypeAAP
 	}
 
-	return nil, nil, errors.New("no auth configuration provided")
+	if !multiAuth.HasProviders() {
+		return nil, errors.New("no authentication providers configured")
+	}
+
+	// Start the cache background cleanup
+	multiAuth.Start()
```
hexfusion: is the comment here correct, it appears to be starting mutiAuth.

internal/auth/authn/aapgateway.go
```diff
@@ -104,5 +136,92 @@ func (a AapGatewayAuth) GetIdentity(ctx context.Context, token string) (common.I
 		return nil, fmt.Errorf("failed to get identity: %w", err)
 	}
 
+	// Get user organizations from AAP
+	organizations, err := a.getUserOrganizations(ctx, token, userInfo)
+	if err != nil {
+		// Log error but don't fail authentication - organizations are optional
+		// TODO: Add proper logging here
```
hexfusion, line 143: did you want to clear this before merge?

internal/auth/authn/k8s.go
```diff
@@ -75,12 +97,217 @@ func (o K8sAuthN) GetIdentity(ctx context.Context, token string) (common.Identit
...
+		// Keep roles that don't start with organizationPrefix
+		if !strings.HasPrefix(role, organizationPrefix) {
+			filteredRoles = append(filteredRoles, role)
+		}
 	}
+	return filteredRoles
+}
+
+// fetchRoleBindingsForUser fetches RoleBindings for a user in the configured RBAC namespace
+func (o K8sAuthN) fetchRoleBindingsForUser(ctx context.Context, username string) ([]string, error) {
```
hexfusion, line 175: why return error type if always nil

internal/auth/authn/k8s.go
```diff
@@ -75,12 +97,217 @@ func (o K8sAuthN) GetIdentity(ctx context.Context, token string) (common.Identit
...
 		return nil, err
 	}
-	return common.NewBaseIdentity(review.Status.User.Username, review.Status.User.UID, review.Status.User.Groups), nil
+
+	var organizations []string
+
+	// Token doesn't have scopes claim - this is a regular K8s token
+	// Get roles from RoleBindings instead of TokenReview groups
+	roles, _ := o.fetchRoleBindingsForUser(ctx, review.Status.User.Username)
+	scopes, err := o.parseScopesFromToken(token)
```
hexfusion, line 106: should we log this error if not nil?

internal/auth/middleware.go
```diff
@@ -163,6 +181,24 @@ func isAllowed(ctx context.Context, authZ AuthZMiddleware, log logrus.FieldLogge
 	return false
 }
 
+// isPublicAuthEndpoint checks if the given path is a public auth endpoint that doesn't require authentication
+func isPublicAuthEndpoint(path string) bool {
```
DakCrowder: It looks like there are two isPublicAuthEndpoint functions can they be consolidated?

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+
+	// Logger for authentication operations
+	log logrus.FieldLogger
+
+	// Dynamic OIDC providers - issuer+clientId -> provider mapping
+	dynamicProviders   map[AuthProviderCacheKey]common.AuthNMiddleware
+	dynamicProvidersMu sync.RWMutex
+
+	// Control for background loader goroutine
+	ctx    context.Context
```
hexfusion: nit: amost always context should be passed not stored cancelFn is enough to control the actual ctx.

internal/client/authorization.go
```diff
@@ -37,15 +37,15 @@ func CreateAuthProvider(authInfo AuthInfo, insecure bool) (login.AuthProvider, e
...
-	switch authInfo.AuthProvider.Name {
-	case common.AuthTypeK8s:
+	switch authInfo.AuthProvider.Type {
+	case string(api.K8s):
 		return login.NewK8sOAuth2Config(c[AuthCAFileKey], c[AuthClientIdKey], c[AuthUrlKey], insecure), nil
-	case common.AuthTypeOIDC:
+	case string(api.Oidc):
 		return login.NewOIDCConfig(c[AuthCAFileKey], c[AuthClientIdKey], c[AuthUrlKey], authInfo.OrganizationsEnabled, insecure), nil
-	case common.AuthTypeAAP:
+	case "aap":
```
DakCrowder: Can this string be changed to a ref to some existing const?

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+}
+
+// NewMultiAuth creates a new MultiAuth instance
+func NewMultiAuth(authProviderService AuthProviderService, tlsConfig *tls.Config, log logrus.FieldLogger) *MultiAuth {
+	ctx, cancel := context.WithCancel(context.Background())
```
hexfusion: creating ctx in constructor feels awkward.  can we use Run(ctx)  pattern or if you need off switch you can use `done <-chan struct{}`

```go

func (s *Shutdown) Close() {
    select {
    case <-s.done:
        return // already closed
    default:
        close(s.done)
    }
}
```

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+	m.staticProviders[issuer] = provider
+}
+
+// HasProviders returns true if any providers are configured
+func (m *MultiAuth) HasProviders() bool {
+	return len(m.staticProviders) > 0
+}
+
+// Start starts the background loader goroutine
+func (m *MultiAuth) Start() {
```
hexfusion: if you would like to use ctx which make sense you can pass it here and create a child and store the childs cancelFn in m.

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+}
+
+// Stop stops the background loader goroutine
+func (m *MultiAuth) Stop() {
+	m.cancel()
+	m.wg.Wait()
+}
+
+// periodicLoader runs in the background and reloads dynamic providers every 5 seconds
+func (m *MultiAuth) periodicLoader() {
```
hexfusion:
```suggestion
func (m *MultiAuth) periodicLoader(ctx context.Context) {
```

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+		m.log.Warnf("Failed to load auth providers on startup: %v", err)
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := m.LoadAllAuthProviders(m.ctx); err != nil {
+				m.log.Warnf("Failed to reload auth providers: %v", err)
+			}
+		case <-m.ctx.Done():
```
hexfusion:
```suggestion
		case <- ctx.Done():
```

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+				unchangedCount++
+			}
+		} else {
+			// New provider - create and add
+			_, authMiddleware, err := m.createAuthMiddlewareFromProvider(provider)
+			if err != nil {
+				m.log.Warnf("Failed to create auth provider %s: %v", lo.FromPtr(provider.Metadata.Name), err)
+				continue
+			}
+			m.dynamicProviders[providerKey] = authMiddleware
```
hexfusion, line 188: would it be possible that this providerKey already exists? should it always be overwritten

internal/auth/authn/multiauth.go
```diff
@@ -0,0 +1,796 @@
...
+
+// Organization assignment type constants
+const (
+	OrganizationAssignmentTypeStatic  = "static"
+	OrganizationAssignmentTypeDynamic = "dynamic"
+	OrganizationAssignmentTypePerUser = "perUser"
+)
+
+// contextKey is a custom type for context keys to avoid collisions
+type contextKey string
```
DakCrowder: contextKey is also defined in internal/auth/middleware.go can that be consolidated?

internal/auth/authn/role_extractor.go
```diff
@@ -0,0 +1,122 @@
...
+	return &RoleExtractor{
+		roleAssignment: roleAssignment,
+	}
+}
+
+// ExtractRolesFromMap extracts roles from a map of claims (for OAuth2 userinfo)
+func (r *RoleExtractor) ExtractRolesFromMap(claims map[string]interface{}) []string {
+	discriminator, err := r.roleAssignment.Discriminator()
+	if err != nil {
+		return []string{}
```
hexfusion: minor: go treats nil and empty slices the same
```suggestion
		return nil
```

internal/auth/common/common.go
```diff
@@ -6,49 +6,82 @@ import (
...
-
 type AuthOrganizationsConfig struct {
 	Enabled bool
+	// OrganizationAssignment defines how users are assigned to organizations
+	OrganizationAssignment *OrganizationAssignment
+}
+
+// OrganizationAssignment defines how users are assigned to organizations
+type OrganizationAssignment struct {
+	Type string `json:"type"` // "static", "dynamic", or "perUser"
```
hexfusion: should this be an enum?

internal/auth/auth.go
```diff
@@ -65,100 +100,89 @@ func initK8sAuth(cfg *config.Config, log logrus.FieldLogger) (AuthNMiddleware, A
...
-	authZProvider := K8sToK8sAuth{K8sAuthZ: authz.K8sAuthZ{K8sClient: k8sClient, Namespace: cfg.Auth.K8s.RBACNs}}
-	authNProvider, err := authn.NewK8sAuthN(k8sClient, externalOpenShiftApiUrl)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to create k8s AuthN: %w", err)
+		return nil, fmt.Errorf("failed to create k8s client: %w", err)
 	}
-	return authNProvider, authZProvider, nil
-}
 
-func getTlsConfig(cfg *config.Config) *tls.Config {
```
DakCrowder, line 78: the getTlsconfig previously used to setup a certpool / rootCAs but it no longer does - specifically for the AAP case i'm not sure if the current implementation allows for users to specify the `cfg.Auth.CACert` which I think we should continue to support.

api/v1alpha1/openapi.yaml
asafbennatan: it is , aap implementation includes some aap specific stuff , we can consolidate this later into oauth2 type if we want

cmd/flightctl-alertmanager-proxy/main.go
asafbennatan: this is what we had before - i just kept it as it was

internal/api_server/middleware/agent_auth.go
asafbennatan: the certificate is validated in the tls middleware layer , see 
`		err := tc.HandshakeContext(ctx)
`

in middleware/tls.go

cmd/flightctl-alertmanager-proxy/main.go
asafbennatan: there is a discussion about removing auth=true and orgs=true (and all other auth disabling) - i dont see proxy.authDisabled in config

cmd/flightctl-alertmanager-proxy/main.go
asafbennatan: adjusted

internal/auth/authn/multiauth.go
asafbennatan: done

internal/auth/authn/multiauth.go
asafbennatan: yes its possible , and it should be overwritten - it means it was updated , 
we should not have two different providers with the same key

internal/auth/authn/role_extractor.go
asafbennatan: adjusted

internal/auth/common/common.go
asafbennatan: remove this rouge OrganizationAssignment struct - we should be using the one from the generated types

internal/api_server/middleware/middleware.go
asafbennatan: fixed

internal/auth/middleware.go
asafbennatan: done

internal/auth/authn/multiauth.go
asafbennatan: done

internal/auth/authn/oauth2_auth.go
```diff
@@ -0,0 +1,248 @@
...
+	provider := api.AuthProvider{
+		ApiVersion: api.AuthProviderAPIVersion,
+		Kind:       api.AuthProviderKind,
+		Metadata:   o.metadata,
+		Spec:       api.AuthProviderSpec{},
+	}
+
+	// Create a copy of the spec with masked client secret
+	maskedSpec := o.spec
+	if maskedSpec.ClientSecret != nil && *maskedSpec.ClientSecret != "" {
```
celdrake: The way we're handling the clientSecret seems error-prone and that it could potentially lead to leaking the secrets.

I know other parts of the system use `SecureString` but IDK if it could be applied in this case, or if there's a better way that the current method.

EDIT: Asked Cursor, may not be accurate but what it suggested is
```
**The Problem:
/authproviders returns *api.AuthProvider or *api.AuthProviderList, which implement SensitiveDataHider. SetResponse in transport/common.go calls HideSensitiveData() before encoding, so secrets are masked.
/auth/config returns *api.AuthConfig, which does not implement SensitiveDataHider, so SetResponse does not call HideSensitiveData(). The manual masking in GetAuthConfig() is the only protection, but it has a bug.

The Bug in Manual Masking:
The code does maskedSpec := o.spec, which is a shallow copy. If ClientSecret is a *string, both maskedSpec.ClientSecret and o.spec.ClientSecret point to the same string. Setting maskedSpec.ClientSecret = &masked changes the pointer in the copy, but the original secret remains in o.spec.ClientSecret and could leak.
Solution:
Make AuthConfig implement SensitiveDataHider so it hides secrets in all nested providers. Add this to api/v1alpha1/util.go:**
```

internal/auth/auth.go
asafbennatan: fixed

internal/auth/authn/oauth2_auth.go
asafbennatan: nice , was not aware of securestring

asafbennatan: fixed

api/v1alpha1/validation.go
asafbennatan: done

deploy/helm/flightctl/README.md
```diff
@@ -302,7 +302,10 @@ For more detailed configuration options, see the [Values](#values) section below
...
-| global.auth.type | string | `"oidc"` | Type of the auth to use. Can be one of 'k8s', 'oidc', 'builtin', 'aap', or 'none' Note: 'builtin' is a legacy mode that translates to 'oidc' with PAM issuer automatically enabled For new deployments, explicitly set type to 'oidc' and configure pamOidcIssuer settings |
+| global.auth.oidc.organizationAssignment | object | `{"organizationName":"default","type":"static"}` | Organization assignment configuration |
+| global.auth.oidc.roleAssignment | object | `{"claimPath":["groups"],"type":"dynamic"}` | Role assignment configuration |
+| global.auth.oidc.usernameClaim | list | `["preferred_username"]` | Username claim to extract from OIDC token (default: "preferred_username") |
+| global.auth.type | string | `"k8s"` | Type of the auth to use. Can be one of 'k8s', 'oidc', 'builtin', 'aap', or 'none' Note: 'builtin' is a legacy mode that translates to 'oidc' with PAM issuer automatically enabled For new deployments, explicitly set type to 'oidc' and configure pamOidcIssuer settings |
```
fzdarsky, line 308: Do we still want `none` and `builtin`?

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-config.yaml
```diff
@@ -61,8 +61,22 @@ data:
         externalApiUrl: {{ .Values.global.auth.aap.externalApiUrl }}
       {{- else }}
       oidc:
-        oidcAuthority: {{ .Values.global.auth.oidc.issuer }}
+        clientId: {{ .Values.global.auth.oidc.clientId }}
+        {{- if .Values.global.auth.oidc.enabled }}
+        enabled: {{ .Values.global.auth.oidc.enabled }}
+        {{- end }}
```
fzdarsky, line 67: What is this parameter? It was present but not used anywhere before this PR and I'm wondering why one would configure `global.auth.oidc` but not enable?

api/v1alpha1/openapi.yaml
```diff
@@ -5455,29 +5820,22 @@ components:
...
-        - authURL
-        - authOrganizationsConfig
-      description: Auth config.
-    AuthOrganizationsConfig:
-      type: object
-      description: Auth related organizations configuration.
-      properties:
-        enabled:
+          description: Name of the default authentication provider.
+        organizationsEnabled:
```
fzdarsky, line 5834: Remove?

deploy/helm/flightctl/README.md
asafbennatan: i thought we should avoid breaking builtin- but we can remove it

api/v1alpha1/openapi.yaml
asafbennatan: lets do this in the removal of all org enable auth enable stuff

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-config.yaml
asafbennatan: if you had it configured but now want to disable it since you setup a dynamic provider

PR #1935, opened 2025-11-10
EDM-2266: CLI handle Multiple Org Selection

SiddarthR56:
## Summary by CodeRabbit

* **New Features**
  * Login now auto-selects and persists a single organization when present and shows a list when multiple organizations exist.

* **Improvements**
  * Client setup and login respect the server-side organizations flag.
  * Configuration is parsed earlier to surface errors sooner.
  * Validation can be configured to skip the organization requirement for applicable commands.
  * Clearer messages and warnings when an organization is required or cannot be fetched.

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -364,7 +364,7 @@ func (o *LoginOptions) getAuthProvider(ctx context.Context) (login.AuthProvider,
 			CAFile:             o.AuthCAFile,
 			InsecureSkipVerify: o.InsecureSkipVerify,
 		},
-		OrganizationsEnabled: true,
+		OrganizationsEnabled: o.organizationsEnabled(),
 	}
 	authProvider, err := client.CreateAuthProviderWithCredentials(authInfo, o.InsecureSkipVerify, o.clientConfig.Service.Server, o.CallbackPort, o.Username, o.Password, o.Web)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


fzdarsky:
We have the `-c/--context` flag to switch between FC instances and it may make sense to include the org ID into the context, so users can select both FC and org with a single flag.

SiddarthR56:
> We have the `-c/--context` flag to switch between FC instances and it may make sense to include the org ID into the context, so users can select both FC and org with a single flag.

This is supported; users can create context-specific config files with org-inclusive names, and each config stores its own organization
```
flightctl login url --context prod-org1
flightctl config set-organization org1 --context prod-org1

flightctl login url --context prod-org2
flightctl config set-organization org2 --context prod-org2

flightctl get devices --context prod-org1
flightctl get devices --context prod-org2 
```

internal/cli/login.go
```diff
@@ -528,31 +531,43 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
...
-
 	if err := o.clientConfig.Persist(o.ConfigFilePath); err != nil {
 		return fmt.Errorf("persisting client config: %w", err)
 	}
 
 	fmt.Println("Login successful.")
+
+	// Handle organization selection if enabled
+	if o.authConfig.OrganizationsEnabled != nil && *o.authConfig.OrganizationsEnabled {
+		if response, err := c.ListOrganizationsWithResponse(ctx, &v1beta1.ListOrganizationsParams{}); err == nil && response.StatusCode() == http.StatusOK && response.JSON200 != nil {
```
kkyrazis: If Organizations are enabled and the request failed, should we still be swallowing this err? Or would it be helpful to indicate that something went wrong with this call?

SiddarthR56: I decided swallowing this error might be better because login has already succeeded, and fetching orgs is just a best-effort convenience for auto-selection. The UX could be confusing if we surface an error/warning, making it look like login failed, even though the CLI works fine with --org explicitly.

Does that make sense, or do you think it would be better to add a warning log?

kkyrazis: Admittedly I am not a power user of organizations, but I think _I_ would prefer some indication. 

If 99.9% of the time I login and my org is set for me, but that .01% of the time it's not, I may have questions.

With Orgs enabled, the other flows - single org discovered, multiple orgs discovered, failing to persist config - result in additional information output.

PR #1938, opened 2025-11-11
EDM-2321: Add Agent Audit Log

galel12:
# Add audit logging for device spec transitions

## Summary
Implements structured audit logging for the flightctl agent to track all device specification changes, providing a comprehensive audit trail for troubleshooting, compliance, and operational visibility.

### Core Implementation
- **New audit package** (`internal/agent/device/spec/audit/`) with clean interface-based design
  - `Logger` interface for dependency injection and testability
  - `FileLogger` implementation using JSON Lines format
  - Four event types: `bootstrap`, `sync`, `upgrade`, `rollback`
  
- **Automatic log rotation** using `gopkg.in/natefinch/lumberjack.v2`
  - 2MB per file, 3 backup files (8MB total capacity)
  - Rotation parameters are intentionally hardcoded per project requirements

- **Integration with spec manager** lifecycle events
  - Bootstrap: First-time device spec creation
  - Sync: Receiving new desired spec from management API
  - Upgrade: Successfully applying a spec version
  - Rollback: Reverting to previous known-good state

### Event Format
```
{
  "ts": "2025-11-11T10:22:01Z",
  "device": "device-name",
  "old_version": "1",
  "new_version": "2",
  "result": "success",
  "type": "upgrade",
  "fleet_template_version": "my-fleet-1",
  "agent_version": "v1.0.0-main-260-g704cd6db"
}
```

### Key Features
- **Diagnostic integration**: Logs automatically collected by `must-gather` and `sosreport`
- **Configurable**: Enable/disable via agent configuration (`audit.enabled`)

## Testing
- ✅ Unit tests for all event types (bootstrap, sync, upgrade, rollback)
- ✅ Integration test for log rotation behavior
- ✅ End-to-end validation with real fleet management scenarios
- ✅ Verified diagnostic tool collection (`must-gather`, `sosreport`)

## Validation Example
Real audit log from testing showing complete lifecycle:
- Bootstrap event when device first enrolled
- Sync event when fleet template applied
- Failed upgrade attempt (authentication error)
- Automatic rollback to previous version
- Successful upgrade with corrected configuration

## Notes
- Log rotation parameters (2MB/file, 3 backups) are intentionally hardcoded as per project requirements
- The `context.Context` parameter in `LogEvent()` is currently unused but reserved for future extensibility (distributed tracing, cancellation)
- Bootstrap event detection logic handles edge case where `Ensure()` is called multiple times during initialization

## Summary by CodeRabbit

* **New Features**
  * Structured audit logging for device events (bootstrap, sync, upgrade, rollback) to /var/log/flightctl/audit.log with rotation, agent/fleet metadata, and runtime enable/disable.

* **Configuration**
  * Exposed audit logging configuration with sensible defaults; startup won’t fail if log target is absent.

* **Tests**
  * Comprehensive unit tests and generated mocks covering audit config and logger behavior.

* **Chores**
  * Packaging, SELinux and diagnostics updated to include audit logs; module dependency for log rotation added.

### DIFF
--- a/internal/agent/device/spec/audit/logger.go
+++ b/internal/agent/device/spec/audit/logger.go
@@ -91,6 +91,7 @@ func (f *FileLogger) LogEvent(ctx context.Context, info *EventInfo) error {
 	}
 
 	if f.config.Enabled == nil || !*f.config.Enabled {
+		f.log.Debug("Audit logging is disabled, skipping event")
 		return nil
 	}
 
--- a/internal/agent/device/spec/spec_test.go
+++ b/internal/agent/device/spec/spec_test.go
@@ -852,9 +852,9 @@ func TestRollback(t *testing.T) {
 
 			tc.setupMocks(mockPolicyManager)
 
-			err := s.write(ctx, Current, newVersionedDevice(tc.currentVersion), audit.ReasonBootstrap)
+			err := s.write(ctx, Current, newVersionedDevice(tc.currentVersion), audit.ReasonInitialization)
 			require.NoError(err)
-			err = s.write(ctx, Desired, newVersionedDevice(tc.desiredVersion), audit.ReasonBootstrap)
+			err = s.write(ctx, Desired, newVersionedDevice(tc.desiredVersion), audit.ReasonInitialization)
 			require.NoError(err)
 
 			opts := []RollbackOption{}
--- a/Makefile
+++ b/Makefile
@@ -367,6 +367,7 @@ clean: clean-agent-vm clean-e2e-agent-images clean-quadlets clean-swtpm-certs
 	- rm -rf $(shell uname -m)
 	- rm -rf obj-*-linux-gnu
 	- rm -rf debian
+	- rm -rf .output/stamps
 # Qcow2 disk depends on the touch file
 bin/output/qcow2/disk.qcow2: bin/.e2e-agent-images
 
@@ -388,14 +389,16 @@ LINT_CONTAINER := podman run --rm \
 	-v go-mod-cache:/go/pkg/mod \
 	-w /app --user 0 $(LINT_IMAGE)
 
-.PHONY: tools lint-image
+.PHONY: tools
 tools:
 
-lint-image:
+.output/stamps/lint-image: Containerfile.lint go.mod go.sum
+	@mkdir -p .output/stamps
 	podman build -f Containerfile.lint -t $(LINT_IMAGE)
+	@touch .output/stamps/lint-image
 
 .PHONY: lint
-lint: lint-image
+lint: .output/stamps/lint-image
 	$(LINT_CONTAINER) golangci-lint run -v
 
 .PHONY: rpmlint
@@ -412,15 +415,23 @@ rpmlint-ci:
 check-rpmlint:
 	@command -v rpmlint > /dev/null || (echo "rpmlint not found. Install with: sudo apt-get install rpmlint (Ubuntu/Debian) or sudo dnf install rpmlint (Fedora/RHEL)" && exit 1)
 
-.PHONY: lint-openapi
-lint-openapi:
+.output/stamps/lint-openapi: api/v1alpha1/openapi.yaml .spectral.yaml
+	@mkdir -p .output/stamps
 	@echo "Linting OpenAPI spec"
 	podman run --rm -it -v $(shell pwd):/workdir:Z docker.io/stoplight/spectral:6.14.2 lint --ruleset=/workdir/.spectral.yaml --fail-severity=warn /workdir/api/v1alpha1/openapi.yaml
+	@touch .output/stamps/lint-openapi
 
-.PHONY: lint-docs
-lint-docs:
+.PHONY: lint-openapi
+lint-openapi: .output/stamps/lint-openapi
+
+.output/stamps/lint-docs: $(wildcard docs/user/*.md)
+	@mkdir -p .output/stamps
 	@echo "Linting user documentation markdown files"
 	podman run --rm -v $(shell pwd):/workdir:Z docker.io/davidanson/markdownlint-cli2:v0.16.0 "docs/user/**/*.md"
+	@touch .output/stamps/lint-docs
+
+.PHONY: lint-docs
+lint-docs: .output/stamps/lint-docs
 
 .PHONY: lint-diagrams
 lint-diagrams:
@@ -436,10 +447,14 @@ lint-diagrams:
 		done ; \
 	done
 
-.PHONY: spellcheck-docs
-spellcheck-docs:
+.output/stamps/spellcheck-docs: $(wildcard docs/user/*.md)
+	@mkdir -p .output/stamps
 	@echo "Checking user documentation for spelling issues"
 	podman run --rm -v $(shell pwd):/workdir:Z docker.io/tmaier/markdown-spellcheck:latest --en-us --ignore-numbers --report "docs/user/**/*.md"
+	@touch .output/stamps/spellcheck-docs
+
+.PHONY: spellcheck-docs
+spellcheck-docs: .output/stamps/spellcheck-docs
 
 .PHONY: fix-spelling
 fix-spelling:


internal/agent/config/config.go
```diff
@@ -89,6 +90,9 @@ type Config struct {
 	// TPM holds all TPM-related configuration
 	TPM TPM `json:"tpm,omitempty"`
 
+	// Audit holds all audit logging configuration
+	Audit audit.AuditConfig `json:"audit,omitempty"`
```
hexfusion: We will eventually have file system adestation which may be considered audit level configuration. Should we make this config explicit as audit-log:

internal/agent/device/fileio/fileio.go
```diff
@@ -31,6 +31,8 @@ type Writer interface {
 	PathFor(filePath string) string
 	// WriteFile writes the provided data to the file at the path with the provided permissions and ownership information
 	WriteFile(name string, data []byte, perm fs.FileMode, opts ...FileOption) error
+	// AppendFile appends the provided data to the file at the path, creating the file if it doesn't exist
+	AppendFile(name string, data []byte, perm fs.FileMode, opts ...FileOption) error
```
hexfusion: This interface is getting quite fat could we break this out as a function that takes a Writer as input to isolate the functionality.

internal/agent/device/spec/audit/config.go
```diff
@@ -0,0 +1,66 @@
...
+
+const (
+	// DefaultLogPath is the hardcoded path for audit logs - not configurable by users
+	DefaultLogPath = "/var/log/flightctl/audit.log"
+	// DefaultEnabled is the default enabled state for audit logging
+	DefaultEnabled = true
+
+	// Hardcoded rotation defaults - non-configurable by users
+	// 2 MB per file, 3 backups = 8 MB total (≈27,960 records)
+	DefaultMaxSizeKB  = 2048 // 2MB per file (approximately 6,990 records)
```
hexfusion: should we reduce the default size here I think 1000 records is fine and we want to have a very minimal default footprint. but also 8mb isnt that much either :)
```
	DefaultMaxSizeKB  = 300
```

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return &FileLogger{
+		config:       config,
+		readWriter:   readWriter,
+		deviceID:     deviceID,
+		agentVersion: agentVersion,
+		log:          log,
+		rotatingLog:  rotatingLog,
+	}, nil
+}
```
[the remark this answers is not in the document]
galel12: I think functional options would add unnecessary complexity here since all 5 parameters are required with no defaults or optional behavior. The current positional approach is clearer for this use case. Functional options work well for CertManager because it has optional storage providers and configs, but our constructor is simpler. We can refactor if optional parameters are added in the future.

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,135 @@
...
+		// Test mode: use fileio AppendFile (allows mocking)
+		if err := f.readWriter.AppendFile(DefaultLogPath, eventBytes, fileio.DefaultFilePermissions); err != nil {
+			return fmt.Errorf("appending audit event to %q: %w", DefaultLogPath, err)
+		}
+	} else {
+		// Production mode: use lumberjack for rotation
+		if _, err := f.rotatingLog.Write(eventBytes); err != nil {
+			return fmt.Errorf("writing audit event to rotating log: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
galel12: This is an intentional and documented exception (see ARCHITECTURAL NOTE lines 127-133).

 Why bypass FileIO here:
 - ✅ Lumberjack provides battle-tested rotation (don't reinvent the wheel)
 - ✅ Append-only operations = low risk surface
 - ✅ Tests still use FileIO mocks via PathFor() detection
 - ❌ Extending FileIO adds complexity for a single use case (no other component needs rotation)

 Tradeoff: Direct lumberjack usage for specialized rotation vs. over-engineering FileIO for one use case.

 As noted in code: "This pattern should be reconsidered if rotation becomes needed elsewhere." Currently it's audit-only, so the exception is justified and documented for future maintainers.

internal/agent/device/spec/audit/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// AuditEvent represents a single audit log entry.
+type AuditEvent struct {
+	Ts                   string      `json:"ts"`                     // RFC3339 UTC format - when the attempt started
+	Device               string      `json:"device"`                 // device name
+	OldVersion           string      `json:"old_version"`            // current effective version before the attempt
+	NewVersion           string      `json:"new_version"`            // target version
+	Result               AuditResult `json:"result"`                 // success | failure
+	Type                 AuditType   `json:"type"`                   // bootstrap/sync/upgrade/rollback
```
hexfusion: is this `Reason`?

internal/agent/config/config.go
galel12: ✅ Done in 4d75a933. Renamed `Audit` → `AuditLog` in config struct.

internal/agent/device/fileio/fileio.go
galel12: ✅ Done in 4d75a933. Refactored `AppendFile` from Writer interface method to standalone function `fileio.AppendFile(w Writer, ...)`. Tests now use concrete `fileio.ReadWriter` with real file I/O instead of mocking for better integration testing.

internal/agent/device/spec/audit/config.go
galel12: ✅ Done in 4d75a933. Reduced from 2MB → 300KB per file (8MB → 1.2MB total capacity). This gives ~1,050 records per file, 4,200 total records across 4 files (1 active + 3 backups).

internal/agent/device/spec/audit/interfaces.go
galel12: ✅ Done in 4d75a933. Added strongly-typed `Reason` field (bootstrap/sync/upgrade/rollback) to track WHY state changes occur. Restructured `Type` field to track WHAT file operation (current/desired/rollback). Updated all 6 call sites in spec manager with correct reason/type mappings.

internal/agent/device/spec/audit/config.go
asafbss: Please update the design document if we eventually decide to change the default.

galel12: ✅ Will update the design doc. Important clarification:

**Actual behavior:** Lumberjack's `MaxSize` only accepts integer MB (no fractional values). So 300KB rounds up to **1MB per file** in practice.

**Final capacity:**
- 1MB per file (not 300KB due to lumberjack API limitation)
- 4MB total (1 active + 3 backups)
- ~14,000 audit events total (vs. ~4,200 at 300KB)

The 300KB constant (`DefaultMaxSizeKB = 300`) remains in the code for clarity, but the actual rotation happens at 1MB boundaries, so I think it's better to stick to luberjack limitations and change that to integer (i.e. 1/2/3MG) cause now it's a bit misleading.

@hexfusion  - want to confirm this is acceptable before I update the docs? 8MB→4MB is still a significant reduction.
@asafbss

internal/agent/config/config.go
```diff
@@ -409,6 +419,11 @@ func mergeConfigs(base, override *Config) {
 	overrideIfNotEmpty(&base.TPM.DevicePath, override.TPM.DevicePath)
 	overrideIfNotEmpty(&base.TPM.StorageFilePath, override.TPM.StorageFilePath)
 
+	// audit log
```
asafbennatan, line 422: not super falimiler with overrideifnotempty function but shouldnt we just use it here ?

internal/agent/device/spec/audit/interfaces.go
```diff
@@ -0,0 +1,78 @@
...
+
+const (
+	// AuditResultSuccess represents a successful operation
+	AuditResultSuccess AuditResult = "success"
+	// AuditResultFailure represents a failed operation
+	AuditResultFailure AuditResult = "failure"
+)
+
+// AuditEvent represents a single audit log entry.
+type AuditEvent struct {
```
asafbennatan: what will happen if we need to support additional events ? 
perheps we should have a descriminator 
something like 
AuditEvent{
//general auditing event fields
Ts
AgentVersion
//event data - specific event data
EventData

}
and then you can EventData as one of the event types 

also - could it make sense to add an event version here ?

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return fmt.Errorf("marshaling audit event: %w", err)
+	}
+
+	// Add newline for JSON lines format
+	eventBytes = append(eventBytes, '\n')
+
+	// Check if we're in test mode (fileio has PathFor method that indicates testing)
+	// In tests, use fileio AppendFile for mockability. In production, use lumberjack for rotation.
+	if testPath := f.readWriter.PathFor(""); testPath != "" {
```
asafbennatan: why do we need this distinction between test logic and production logic ?

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,148 @@
...
+// Uses lumberjack for rotation in production, fileio for testing.
+func (f *FileLogger) writeEvent(event AuditEvent) error {
+	// Marshal event to JSON
+	eventBytes, err := json.Marshal(event)
+	if err != nil {
+		return fmt.Errorf("marshaling audit event: %w", err)
+	}
+
+	// Add newline for JSON lines format
+	eventBytes = append(eventBytes, '\n')
```
asafbennatan, line 131: i think that if we want we can use logrus and create it with lumberjack:
https://medium.com/@danielmalagurti/efficient-logging-in-go-utilizing-logrus-with-lumberjack-b77713b972e0

should we do that ? 
and then use logurs structured logging feature ?

internal/agent/config/config.go
galel12: `overrideIfNotEmpty` only works for strings/pointers, but `Audit.Enabled` is a `*bool` (pointer to bool). Using `overrideIfNotEmpty` would treat `nil` and `false` the same, which breaks our requirement to distinguish between "unset" and "explicitly false".

The current pattern `if override.AuditLog.Enabled != nil` correctly handles all three states:
- nil → unset, don't override
- *false → explicitly disabled, override to false
- *true → explicitly enabled, override to true

This matches the pattern we'll need for other optional booleans (TPM.Enabled, etc.) once they're migrated to `*bool`.

internal/agent/device/spec/audit/interfaces.go
galel12: Valid concern. Since all audit events share the same schema (spec transitions), the flat structure keeps things simple for parsing. We can add a discriminator later if we need heterogeneous event types, not worth the complexity now in my opinion.

internal/agent/device/spec/audit/logger.go
galel12: The split enables real file I/O testing without mocking. Test mode uses concrete `fileio.ReadWriter` with `t.TempDir()` to catch actual bugs (permissions, ownership, etc.), while production uses lumberjack's battle-tested rotation directly. We can't unify them because lumberjack can't work with `t.TempDir()` isolation, and reimplementing rotation in FileIO would duplicate lumberjack's logic. It's a documented tradeoff (see ARCHITECTURAL NOTE lines 133-139).

internal/agent/device/spec/audit/logger.go
galel12: Thanks for the suggestion and the link! I went with manual marshaling to keep dependencies minimal, but logrus could make sense if we expand structured logging. Let me evaluate the trade-offs more carefully I want to make sure we're not reinventing the wheel if logrus is the better long-term approach.

internal/agent/config/config.go
```diff
@@ -89,6 +90,9 @@ type Config struct {
 	// TPM holds all TPM-related configuration
 	TPM TPM `json:"tpm,omitempty"`
 
+	// AuditLog holds all audit logging configuration
+	AuditLog audit.AuditConfig `json:"audit,omitempty"`
```
asafbss, line 94: Any reason for not using a pointer here?
Default can be true.

internal/agent/config/config.go
asafbss: Any reason for not using a pointer here?
Default can be true unless explicity override

internal/agent/device/spec/audit/config.go
```diff
@@ -0,0 +1,62 @@
+package audit
+
+import (
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+)
+
+const (
```
asafbss, line 7: Do we actually need a separate config file for the audit log?
Couldn’t we just use the general agent configuration instead?
The default values can be defined in logger.go.

packaging/must-gather/flightctl-must-gather
```diff
@@ -27,7 +27,10 @@ echo "### Collecting journal logs..."
 journalctl --since "24 hours ago" > "$temp_dir/journal.log"
 
 echo "### Flightctl Spec Files..."
-cp /var/lib/flightctl/*.json "$temp_dir/"
+cp /var/lib/flightctl/*.json "$temp_dir/" 2>/dev/null || true
+
+echo "### Flightctl Audit Log..."
+cp /var/log/flightctl/audit.log "$temp_dir/" 2>/dev/null || true
```
asafbss: Right now you’re copying only the current file audit.log.
If you use log rotation (e.g., lumberjack), you’ll also have files like audit.log.1, audit.log.2 etc.

internal/agent/device/spec/spec.go
```diff
@@ -137,6 +137,7 @@ func (c *cache) update(specType Type, device *v1alpha1.Device) {
 	}
 
 	renderedVersion := device.Version()
+
```
asafbss: remove

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,148 @@
...
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	"github.com/flightctl/flightctl/pkg/log"
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+var _ Logger = (*FileLogger)(nil)
```
asafbss, line 14: Word “Logger” is a bit of a confusing name. Can we use “audit” instead (e.g., audit.go, Audit.LogEvent)?

galel12:
@asafbss 
Changed based on CodeRabbit's feedback. 

**The problem:** Originally used plain `bool` for `Enabled`, which broke config overrides. Setting `enabled: false` was ignored because Go treats `false` as a zero value.

**The fix:** Changed to `*bool` so we can tell the difference:
- `nil` = not set, use default (true)
- `*false` = explicitly disabled
- `*true` = explicitly enabled

**Why not pointer for whole struct?** Don't need it. Making `*audit.AuditConfig` would just add nil checks everywhere without solving the core problem. The `*bool` inside handles what we need.

internal/agent/config/config.go
galel12: https://github.com/flightctl/flightctl/pull/1938#issuecomment-3522090611

packaging/must-gather/flightctl-must-gather
galel12: Totally agree, I'll update the must-gather script to use `audit.log*` to collect all rotated files for complete diagnostics.
Thanks for the comment! :+1:

internal/agent/device/spec/spec.go
galel12: Will do! :+1:

internal/agent/config/config.go
galel12: Ah, you're absolutely right I see it now! Since `Enabled` is already a `*bool`, `overrideIfNotEmpty` handles the nil vs non-nil check perfectly. Simplified it in the latest commit. dca1b17
Sorry for the confusion :heavy_check_mark:

internal/agent/device/spec/audit/interfaces.go
```diff
@@ -0,0 +1,78 @@
+package audit
```
hexfusion: existing agent pattern is to move all interfaces into the named file of the package. so in this case all interfaces should live in audit.go please

internal/agent/device/spec/audit/interfaces.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"context"
+	"time"
+)
+
+// AuditReason represents why the state change is happening (lifecycle phase).
+type AuditReason string
```
hexfusion: in terms of naming always think in the context of the package your in. So in the case you are in the audit package the name Audit is redundant. Also by removing this you have a more natural language usage.

```go
reason := audit.Reason
// vs
reason := audit.AuditReason
```

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,148 @@
...
+	maxSizeMB := (DefaultMaxSizeKB + 1023) / 1024
+	if maxSizeMB < 1 {
+		maxSizeMB = 1
+	}
+	rotatingLog := &lumberjack.Logger{
+		Filename:   DefaultLogPath,
+		MaxSize:    maxSizeMB, // MB (300KB rounds up to 1MB)
+		MaxBackups: DefaultMaxBackups,
+		MaxAge:     DefaultMaxAge,
+		Compress:   false, // Keep uncompressed for easier debugging
```
hexfusion: We are a steward of the device lets always lean towards having a smaller footprint when possible,

```suggestion
		Compress:   true, 
```

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,148 @@
...
+		rotatingLog:  rotatingLog,
+	}, nil
+}
+
+// LogEvent logs a complete audit event with all required fields.
+// The ctx parameter is currently unused but reserved for future extensibility
+// (e.g., context-aware logging, distributed tracing, cancellation support).
+func (f *FileLogger) LogEvent(ctx context.Context, info *AuditEventInfo) error {
+	if f.config.Enabled == nil || !*f.config.Enabled {
+		return nil
```
hexfusion, line 95: nit: if we are going to return early lets log why.

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,148 @@
...
+		agentVersion: agentVersion,
+		log:          log,
+		rotatingLog:  rotatingLog,
+	}, nil
+}
+
+// LogEvent logs a complete audit event with all required fields.
+// The ctx parameter is currently unused but reserved for future extensibility
+// (e.g., context-aware logging, distributed tracing, cancellation support).
+func (f *FileLogger) LogEvent(ctx context.Context, info *AuditEventInfo) error {
```
hexfusion: context is a powerful tool and I love to see it plumbed everwhere. saying that its important to use it :) see below as trick you can use to verify ctx before doing work 

```suggestion
     if err := ctx.Err(); err != nil {
           return ctx.Err()
     }
```

internal/agent/device/spec/audit/logger.go
hexfusion: I agree with @asafbennatan in this question my assumption is that lumberjack just cares about where it should look for things. the implementation details of fileio should enable this without bleeding that implementation into the code.

internal/agent/device/spec/manager.go
```diff
@@ -103,10 +109,21 @@ func (s *manager) Initialize(ctx context.Context) error {
 	}
 	// reconcile the initial spec even though its empty
 	s.queue.Add(ctx, newVersionedDevice("0"))
+
+	// Audit the bootstrap event
+	s.logAudit(ctx, nil, "", "0", audit.AuditResultSuccess, audit.AuditReasonBootstrap, audit.AuditTypeCurrent)
```
hexfusion: the audit log is enabling auditing of the state transitions of the on disk representations of device spec. what state change takes place here? if none should it be a log?

internal/agent/device/spec/manager.go
```diff
@@ -142,6 +163,12 @@ func (s *manager) Ensure() error {
 	// add the desired spec to the queue this ensures that the device will
 	// reconcile the desired spec on startup.
 	s.queue.Add(context.TODO(), desired)
+
+	// Log bootstrap event if this is the first time initializing (all 3 files were just created)
+	if isBootstrap {
+		s.logAudit(context.TODO(), nil, "", "0", audit.AuditResultSuccess, audit.AuditReasonBootstrap, audit.AuditTypeCurrent)
```
hexfusion: the auditing should happen at a lower level `s.write` level. the complexity of gathering the information here is not actually needed if all we care about is did the file change. If the file does not exist can we  assume we are bootstrapping it to version 0?

```go
   if err := s.write(specType, newVersionedDevice("0"), reason); err != nil {
```

internal/agent/device/spec/manager.go
```diff
@@ -207,6 +237,9 @@ func (s *manager) Upgrade(ctx context.Context) error {
 		}
 
 		s.log.Infof("Spec reconciliation complete: current version %s", desired.Version())
+
+		// Log successful spec application to audit log
+		s.logAudit(ctx, desired, currentVersion, desired.Version(), audit.AuditResultSuccess, audit.AuditReasonUpgrade, audit.AuditTypeCurrent)
```
hexfusion: the actual audit should happen inside s.write at line 231 

```go
		if err := s.write(Current, desired, audit.ReasonUpgrade); err != nil {
```

internal/agent/device/spec/manager.go
```diff
@@ -225,6 +258,13 @@ func (s *manager) SetUpgradeFailed(version string) error {
 		return err
 	}
 	s.queue.SetFailed(versionInt)
+
+	// Log failed spec application to audit log
+	currentVersion := s.cache.getRenderedVersion(Current)
+	// Try to get desired spec for fleet template version
+	desired, _ := s.Read(Desired)
+	s.logAudit(context.Background(), desired, currentVersion, version, audit.AuditResultFailure, audit.AuditReasonUpgrade, audit.AuditTypeCurrent)
```
hexfusion: where did the spec file change in this method?

internal/agent/device/spec/manager.go
```diff
@@ -300,6 +343,10 @@ func (s *manager) Rollback(ctx context.Context, opts ...RollbackOption) error {
 
 	// rollback in-memory cache current == desired
 	s.cache.update(Desired, current)
+
+	// Log successful rollback to audit log
+	s.logAudit(ctx, current, failedDesiredVersion, current.Version(), audit.AuditResultSuccess, audit.AuditReasonRollback, audit.AuditTypeRollback)
```
hexfusion: at line 332 we are copying the current to the desired so this is technically the desired state that has changed.
```suggestion
	s.logAudit(ctx, current, failedDesiredVersion, current.Version(), audit.ResultSuccess, audit.ReasonRollback, audit.TypeDesired)
```

internal/agent/device/spec/manager.go
```diff
@@ -417,9 +464,15 @@ func (s *manager) getDeviceFromQueue(ctx context.Context) (*v1alpha1.Device, boo
 			s.log.Infof("Writing new desired rendered spec to disk version: %s", desired.Version())
 		}
 
+		// Capture old desired version BEFORE writing to disk (for audit log)
+		oldDesiredVersion := s.cache.getRenderedVersion(Desired)
+
 		if err := s.write(Desired, desired); err != nil {
```
hexfusion:
```suggestion
		if err := s.write(Desired, desired, audit.ReasonSync); err != nil {
```

packaging/selinux/flightctl_agent.te
```diff
@@ -78,6 +80,12 @@ manage_files_pattern(flightctl_agent_t, flightctl_agent_var_lib_t, flightctl_age
 manage_lnk_files_pattern(flightctl_agent_t, flightctl_agent_var_lib_t, flightctl_agent_var_lib_t)
 files_var_lib_filetrans(flightctl_agent_t, flightctl_agent_var_lib_t, dir, "flightctl")
 
+# /var/log/flightctl management
+manage_dirs_pattern(flightctl_agent_t, flightctl_agent_var_log_t, flightctl_agent_var_log_t)
+manage_files_pattern(flightctl_agent_t, flightctl_agent_var_log_t, flightctl_agent_var_log_t)
+manage_lnk_files_pattern(flightctl_agent_t, flightctl_agent_var_log_t, flightctl_agent_var_log_t)
+logging_log_filetrans(flightctl_agent_t, flightctl_agent_var_log_t, dir, "flightctl")
```
hexfusion, line 87: cc @keitwb

internal/agent/config/config.go
asafbss: Hey, should we flatten the audit configuration to align with the same approach we use for profiling and metrics?
If we ever want to provide configuration for Lumberjack, we can handle that through environment variables.
@hexfusion , what do you think?

internal/agent/device/spec/manager.go
galel12: After implementing audit in write(), I realized we never log 'result: failure' events:
```
func write(ctx, specType, device, reason) error {
    err = writeToFile(...)    if err != nil {        
         return err  // Failed write - returns before audit   
 }    
s.logAudit(..., ResultSuccess, ...)  // Only audits successful writes
}
```
Previously SetUpgradeFailed() logged failures, but you asked me to remove it.
Questions:
Should we remove ResultFailure from the enum? (never used)
Or should failures be audited somewhere? If so, where?
My understanding: we only audit successful state transitions (disk writes that succeeded). Is that correct?

While it's true no state transition occurred, seeing only success audits creates ambiguity - operators can't distinguish between "no failures" vs "failures happened but weren't logged". If we're not using result: failure, we should remove it.

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: resolved in f74534b270b3825702f72989b34afb64cb057bdf

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: resolved in f74534b270b3825702f72989b34afb64cb057bdf

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: resolved in f74534b270b3825702f72989b34afb64cb057bdf

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: resolved in f74534b270b3825702f72989b34afb64cb057bdf

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/commit/f74534b270b3825702f72989b34afb64cb057bdf#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037R128

internal/agent/device/spec/audit/logger.go
galel12: :heavy_check_mark: I'm not sure if that's what you meant https://github.com/flightctl/flightctl/pull/1938/commits/aa02b7f1336987e829eabd13456e55a72bef3cfd#diff-42861db049a288c5466fbdac028c2d6a59da88043acf44fd49afb7a04c713e99R85

internal/agent/device/spec/audit/logger.go
galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/commit/aa02b7f1336987e829eabd13456e55a72bef3cfd#diff-42861db049a288c5466fbdac028c2d6a59da88043acf44fd49afb7a04c713e99R68

internal/agent/device/spec/audit/interfaces.go
galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/commit/aa02b7f1336987e829eabd13456e55a72bef3cfd#diff-0413f0bb804a6dc9d8cd35949de1b5e70a7b36918ada91a77d49ee6312a163ccR9

internal/agent/device/spec/audit/interfaces.go
galel12: :heavy_check_mark:  resolved in aa02b7f

packaging/must-gather/flightctl-must-gather
galel12: :heavy_check_mark:  resolved in https://github.com/flightctl/flightctl/commit/f74534b270b3825702f72989b34afb64cb057bdf#diff-7a056c370e12a83a94847be7cbe88d1a21aa0574dff24d772b316a143aa8eaffR33

internal/agent/device/spec/audit/logger.go
galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/commit/f74534b270b3825702f72989b34afb64cb057bdf#diff-42861db049a288c5466fbdac028c2d6a59da88043acf44fd49afb7a04c713e99R65

and:
https://github.com/flightctl/flightctl/commit/f74534b270b3825702f72989b34afb64cb057bdf#diff-42861db049a288c5466fbdac028c2d6a59da88043acf44fd49afb7a04c713e99R133

internal/agent/device/spec/audit/logger.go
```diff
@@ -0,0 +1,140 @@
...
+		return fmt.Errorf("marshaling audit event: %w", err)
+	}
+
+	// Add newline for JSON lines format
+	eventBytes = append(eventBytes, '\n')
+
+	// Write using lumberjack for automatic rotation (works in both test and production)
+	if _, err := f.rotatingLog.Write(eventBytes); err != nil {
+		return fmt.Errorf("writing audit event to rotating log: %w", err)
+	}
```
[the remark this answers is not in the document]
galel12: Good catch! This has been addressed in the latest commit (f74534b2).

 The refactoring now uses readWriter.PathFor(DefaultLogPath) to get the log file path for lumberjack initialization (line 65), which delegates the test vs. production path handling to the FileIO abstraction. Lumberjack then writes to that resolved path.

 This eliminates the conditional logic that was "bleeding implementation details" as Sam noted, while still leveraging lumberjack's battle-tested rotation.

internal/agent/device/spec/manager.go
```diff
@@ -478,19 +502,49 @@ func (s *manager) Publisher() Publisher {
...
+// an audit event if a reason is provided.
+//
+// Parameters:
+//   - ctx: Context for cancellation propagation (used in audit logging)
+//   - specType: Which spec file to write (Current, Desired, or Rollback)
+//   - device: The device spec to write
+//   - reason: Audit reason (bootstrap, sync, upgrade, rollback). Empty string ("") skips audit.
+//
+// Audit behavior:
+//   - If reason is empty, only writes to disk (no audit)
```
hexfusion: I think reason should be required what is the usecase where we would write to these files and not audit. even if the audit was disabled we would still require reason.

internal/agent/device/spec/manager.go
```diff
@@ -478,19 +502,49 @@ func (s *manager) Publisher() Publisher {
...
+
 	err = writeDeviceToFile(s.deviceReadWriter, device, filePath)
 	if err != nil {
 		return fmt.Errorf("writing %s: %w", specType, err)
 	}
 
 	s.cache.update(specType, device)
 
+	// Audit successful write (skip if reason is empty)
+	if reason != "" {
```
hexfusion: we should error if empty

internal/agent/device/spec/spec_test.go
```diff
@@ -800,9 +800,9 @@ func TestRollback(t *testing.T) {
 
 			tc.setupMocks(mockPolicyManager)
 
-			err := s.write(Current, newVersionedDevice(tc.currentVersion))
+			err := s.write(ctx, Current, newVersionedDevice(tc.currentVersion), "")
```
hexfusion: even in a test reason why is important metadata

internal/agent/device/spec/audit/logger.go
hexfusion: This makes sense to me to explore should not be a lot more work. Can be a followup, but we should try to stabalize by code freeze

internal/agent/device/spec/manager.go
```diff
@@ -89,24 +95,28 @@ func NewManager(
...
 	// reconcile the initial spec even though its empty
 	s.queue.Add(ctx, newVersionedDevice("0"))
+
 	return nil
 }
 
 func (s *manager) Ensure() error {
+	// Check and recreate missing spec files (recovery scenario)
+	// Note: This is NOT bootstrap - Initialize() handles that.
+	// This handles corruption/deletion of files and should not be audited.
```
hexfusion: not sure this comment is correct please verify

internal/agent/device/spec/manager.go
```diff
@@ -115,7 +125,8 @@ func (s *manager) Ensure() error {
 
 		if !exists {
 			s.log.Warnf("Spec file does not exist %s. Resetting state to empty...", specType)
-			if err := s.write(specType, newVersionedDevice("0")); err != nil {
+			// No audit - this is recovery, not a state transition
+			if err := s.write(context.TODO(), specType, newVersionedDevice("0"), ""); err != nil {
```
hexfusion: `Initializing` is a state transition isn't it?

internal/agent/device/spec/manager.go
```diff
@@ -198,7 +210,8 @@ func (s *manager) Upgrade(ctx context.Context) error {
 			return err
 		}
 
-		if err := s.write(Current, desired); err != nil {
+		// Write current spec and audit upgrade (audit happens inside write())
```
hexfusion: this comment is not necessary

internal/agent/device/spec/manager.go
```diff
@@ -258,14 +272,16 @@ func (s *manager) CreateRollback(ctx context.Context) error {
 		Os: &v1alpha1.DeviceOsSpec{Image: currentOSImage},
 	}
 
-	if err := s.write(Rollback, rollback); err != nil {
+	// No audit for creating rollback file
```
hexfusion: why? we are mutating one of the auditable files

internal/agent/device/spec/manager.go
```diff
@@ -276,8 +292,11 @@ func (s *manager) Rollback(ctx context.Context, opts ...RollbackOption) error {
 		opt(cfg)
 	}
 
+	// Capture the failed desired version BEFORE any disk operations (for audit log)
```
hexfusion: extra comments add cutter I think this is pretty obvious WDYT?

internal/agent/device/spec/audit/logger.go
galel12: Great minds think alike! 
I've already renamed interfaces.go to audit.go in the latest commit (f74534b2) following Sam's feedback to move all interfaces into the package-named file (standard agent pattern).
> existing agent pattern is to move all interfaces into the named file of the package. so in this case all interfaces should live in audit.go please

However, I kept this interface name as Logger rather than Audit to avoid redundancy. When used from other packages, audit.Logger reads clearly as "audit logger", while audit.Audit would be repetitive.

internal/agent/device/spec/manager.go
hexfusion: yeah I think failure is basically metadata for a state transition. I think if you added reason to for example the Rollback method of the interface you can add that. directly in device.go

- https://github.com/flightctl/flightctl/blob/3f1bc5ade6353e595a1283e9f92388536a12b70f/internal/agent/device/device.go#L244

- https://github.com/flightctl/flightctl/blob/3f1bc5ade6353e595a1283e9f92388536a12b70f/internal/agent/device/device.go#L544)

maybe we could use `errors.IsRetryable` as a isFailed bool 

```go
 if err := a.rollbackDevice(ctx, current, desired, a.sync, errors.IsRetryable(syncErr)); err != nil {
	a.log.Errorf("Rollback did not complete cleanly: %v", err)
}
```

hexfusion: but I am fine deferring this for MVP as we can conclude failed generally via logs

internal/agent/device/spec/spec.go
galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/commit/f74534b270b3825702f72989b34afb64cb057bdf#diff-974f7b52cff84304cf61ad9063d1c2cd3cea8e9d13c80c22b394dfb2f3d0af2aL140

internal/agent/device/spec/manager.go
```diff
@@ -586,6 +639,37 @@ func IsRollback(current *v1alpha1.Device, desired *v1alpha1.Device) bool {
...
+		Device:               s.deviceName,
+		OldVersion:           oldVersion,
+		NewVersion:           newVersion,
+		Result:               result,
+		Reason:               reason,
+		Type:                 auditType,
+		FleetTemplateVersion: fleetTemplateVersion,
+		StartTime:            time.Now(),
+	}
+
```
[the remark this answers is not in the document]
galel12: Since Metadata is a struct value (not *ObjectMeta pointer), it:

- Can never be nil - structs are always initialized (even if to their zero value)
- Always exists - Go automatically initializes struct fields to their zero value
- Only Metadata.Annotations (which is *map[string]string) can be nil

The Original Check Was Sufficient:
if device != nil && device.Metadata.Annotations != nil {
This is safe because:
1. device != nil ensures the device pointer exists
2. device.Metadata is always initialized (it's a struct, not a pointer)
3. device.Metadata.Annotations is a pointer to a map, so it needs the nil check

internal/agent/device/spec/audit/logger.go
galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/pull/1938/commits/4d494832565beee5461699eb8f284efcdc3152a9#diff-42861db049a288c5466fbdac028c2d6a59da88043acf44fd49afb7a04c713e99R94

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: you're right i removed a lot of obvious comments
https://github.com/flightctl/flightctl/pull/1938/commits/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037L295

internal/agent/device/spec/manager.go
galel12: https://github.com/flightctl/flightctl/pull/1938#discussion_r2532007751
:heavy_check_mark: resolved in https://github.com/flightctl/flightctl/pull/1938/commits/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037L213

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark:  In general i removed this doc of write func cause it felt a bit unnecessary, and also i updated it to require 'reason' anytime with validation check:
https://github.com/flightctl/flightctl/pull/1938/commits/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037R540

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: https://github.com/flightctl/flightctl/pull/1938#discussion_r2532010498
https://github.com/flightctl/flightctl/commit/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037R541

internal/agent/device/spec/spec_test.go
galel12: :heavy_check_mark: resolved in
https://github.com/flightctl/flightctl/pull/1938/commits/4d494832565beee5461699eb8f284efcdc3152a9#diff-8c67ee2191724bd44334c6cc0ee9caba26d4f7d0a58c979205a50c2c516794cbR855

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: you're right it's not accurate.
I updated it in:
https://github.com/flightctl/flightctl/commit/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037R116

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: Yes it is, now it's capturing **all** file mutations correctly:
https://github.com/flightctl/flightctl/commit/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037R139-R154

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: Absolutely correct!
now i'm catching rollback too:
https://github.com/flightctl/flightctl/commit/a2a6df3340084b60f949be76bf739b90d2938f96#diff-c8d23484096cc76398fab6c5a742999c1af8129cb0aff415ff2758c1c8243037R299

thanks for noticing!

PR #1947, opened 2025-11-12
EDM-2293: Quadlet Volumes

kkyrazis:
Adds support for showing Image based volumes (matching compose) in a device's status.
Note: has the same issue with podman volume events that previously existed: https://github.com/containers/podman/issues/26480 I've added some additional documentation to the issue.

Fixes an issue with systemd references being left on the device when the app is stopped.

Applied Spec:
```yaml
spec:
  applications:
    - name: my-app
      appType: quadlet
      inline:
        - content: |
            [Container]
            Image=192.168.68.73:5000/flightctl-tests/alpine:v1
            Volume=test.volume:/mnt/volume
            Exec=sh -c "echo 'Volume mounted. File count:' $(find /mnt/volume -type f | wc -l) && sleep infinity"
            
            [Install]
            WantedBy=default.target
          path: test.container
        - content: |
            [Volume]
            Driver=image
            Image=ghcr.io/homebrew/core/sqlite:3.50.2
          path: test.volume
```

Container's logs:
```text
Nov 12 20:37:31 localhost.localdomain systemd-my-app-250176-test[24839]: Volume mounted. File count: 10
Nov 12 20:37:31 localhost.localdomain my-app-250176-test[24741]: 550f3cf6b81c9dad0c8fdc752cc3db341bc085db44b83de40d37dd3e42524093
```

Device status:
```yaml
status:
  applications:
    - name: my-app
      ready: 1/1
      restarts: 0
      status: Running
      volumes:
        - name: test.volume
          reference: ghcr.io/homebrew/core/sqlite:3.50.2
```

## Summary by CodeRabbit

* **New Features**
  * Pluggable per-app-type handlers unify app path/ID resolution and lifecycle (verify/install/remove) across providers.
  * Volume manager can batch-add volumes and auto-register volumes extracted from quadlet sources.
  * Quadlet references accept an optional explicit volume name with a deterministic default when omitted.
  * Embedded app support unified for quadlet and compose deployments.

* **Bug Fixes / Improvements**
  * Consolidated install/verify flows and clearer quadlet parsing and error handling.

* **Tests**
  * New tests for quadlet volume extraction from specs and directories.


internal/agent/device/applications/provider/volume.go
```diff
@@ -244,3 +246,37 @@ func ToLifecycleVolumes(volumes []*Volume) []lifecycle.Volume {
...
+		if quad.Type != common.QuadletTypeVolume || quad.Image == nil {
+			continue
+		}
+
+		volumes = append(volumes, &Volume{
+			Name:      name,
+			ID:        quadlet.VolumeName(quad.Name, namespacedQuadlet(appID, name)),
+			Reference: *quad.Image,
+			Available: true, // TODO: event support is broken for volumes.  https://github.com/containers/podman/issues/26480
+		})
```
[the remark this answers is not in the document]
kkyrazis: @hexfusion what is your opinion on this. 
given the following yaml
```yaml
        - content: |
            [Volume]
            Driver=image
            Image=ghcr.io/homebrew/core/sqlite:3.50.2
          path: test.volume
```
podman will generate a volume with a name of `systemd-test`. With namespacing that becomes `systemd-my-app-12345-test`

if a user supplies an override via the VolumeName property, then the volume is just named whatever is there: 
```text
VolumeName=xyz
```

Right now I have the ID being generated as the final volume name so that once events are fixed, we can correlate properly, but for the Name, I have it set to the quadlet name:
```yaml
      volumes:
        - name: test.volume
          reference: ghcr.io/homebrew/core/sqlite:3.50.2
```

as that seems easier to reason about than:

```yaml
      volumes:
        - name: systemd-my-app-12345-test
          reference: ghcr.io/homebrew/core/sqlite:3.50.2
```

but it technically is not the name of the volume.

internal/agent/device/applications/provider/embedded.go
```diff
@@ -32,13 +32,24 @@ func newEmbedded(log *log.PrefixLogger, podman *client.Podman, readWriter fileio
 		return nil, fmt.Errorf("getting app path: %w", err)
 	}
 
+	id := client.NewComposeID(name)
+	if appType == v1alpha1.AppTypeQuadlet {
```
adalton: This is purely a matter of personal preference, so take this with a grain of salt.  It has nothing to do specifically with your current change.

Whenever I see repeated checks against an enumerated type, it makes me a bit nervous, because whenever I change something that uses them, I invariably miss something somewhere.

What I'd suggest doing is defining an interface that abstracts away the operations on the various app types, then pushing all the appType-specific implementation into functions that implement those interfaces.  Maybe something like:

```go
type embeddedApp interface {
    AddVolumesFromDir(...) error
    OCITargets(pullSecret *client.PullSecret) ([]dependency.OCIPullTarget, error)
    Verify(ctx context.Context) error
    Install(ctx context.Context) error
    Remove(ctx context.Context) error
}
```

Then, each concrete `appType` could provide a realization of that interface

```go
/* This might live in quadlet.go */
type embeddedQuadletApp struct {
    ...
}

func (qa *embeddedQuadletApp) AddVolumesFromDir(...) error {
    /* move implementation here */
}

...
```
And
```go
/* This might live in compose.go */
type embeddedComposeApp struct {
    ...
}

func (ca *embeddedComposeApp) AddVolumesFromDir(...) error {
    return nil
}
...
```

And
```go
/* This might live in unsupported.go */
type unsupportedComposeApp struct {
    ...
}

...
```

Then everything in `embedded.go` can operation on abstract `embeddedApp`s

```go
func newEmbedded(...) (Provider, error) {
...
app := newEmbeddedApp(...)
...
app.AddVolumesFromDir(...)

return &embeddedProvider{
    ...
    app: app,
    ...
}

func (p *embeddedProvider) OCITargets(pullSecret *client.PullSecret) ([]dependency.OCIPullTarget, error) {                          
    return p.app.OCITargets(pullSecret)                                                                                             
}

func (p *embeddedProvider) Spec() *ApplicationSpec {                                                                                
    return p.spec                                                                                                                   
}
                                                                                                                                    
func (p *embeddedProvider) Install(ctx context.Context) error {                                                                     
    return p.app.Install()                                                                                                          
}
                                                                                                                                    
func (p *embeddedProvider) Remove(ctx context.Context) error {                                                                      
    return p.app.Remove(ctx)                                                                                                        
}
```

This way, adding support for new app types turns into
* Creating a new file for that app type
* Creating a struct that models that specific app type
* Implementing the `embeddedApp` interface for that new app type
* Modify the factory function to support return instance of the new concrete type.

With this, `newEmbeddedApp()` is the only place where anything has to switch on the `appType`, and `embedded.go` has nothing about any concrete `appType`.  Also adding/changing something about one `appType` has less risk of breaking other `appType`s.  The interface makes it very clear what set of operations some has to implement, and if they miss one the compiler will let them know 😄

internal/agent/device/applications/provider/inline.go
```diff
@@ -55,8 +55,17 @@ func newInline(log *log.PrefixLogger, podman *client.Podman, spec *v1alpha1.Appl
 	p.spec.Path = path
 	p.spec.ID = client.NewComposeID(p.spec.Name)
 
-	return p, nil
+	if p.spec.AppType == v1alpha1.AppTypeQuadlet {
```
adalton: The same pattern would work here

internal/agent/device/applications/provider/embedded.go
keitwb: +1. Looking at the rest of this file shows this is ripe for interfacing.

kkyrazis: The third type `image` is in progress. When that lands I'll take a stab at refactoring this.

internal/agent/device/applications/provider/volume.go
```diff
@@ -244,3 +246,37 @@ func ToLifecycleVolumes(volumes []*Volume) []lifecycle.Volume {
 	}
 	return out
 }
+
+func extractQuadletVolumes(appID string, quadlets map[string]*common.QuadletReferences) []*Volume {
+	var volumes []*Volume
+	for name, quad := range quadlets {
+		// Only track volume's with images
```
DakCrowder, line 269: Why do we only care about volumes with images vs all images defined?  Is it because the pull can fail with these?  Just trying to think of some user who might have an image based volume and some directory mount both defined and if that leads to different reporting expectations.

kkyrazis: The quick answer is "feature parity with compose", which also allows for user defined volumes to be created and only the image backed ones are displayed. In that use case, the way those volumes are established is a little non-traditional and it definitely becomes worthwhile to show the status of those volumes. In this case they are a little more well defined, but still worthwhile to show that mounting was successful.

internal/agent/device/applications/provider/embedded.go
kkyrazis: Alright, sorry for the delay. If you could take another look @adalton that would be appreciated.

internal/agent/device/applications/provider/provider.go
```diff
@@ -587,23 +587,18 @@ func isEqual(a, b Provider) bool {
...
-		typePath = lifecycle.ComposeAppPath
-	case v1alpha1.AppTypeQuadlet:
-		// embedded quadlets must be moved to the default quadlet path (so that contents can be mutated)
-		// discovery checks EmbeddedQuadletAppPath, but the application's actual path is the default path
-		typePath = lifecycle.QuadletAppPath
-	default:
-		return "", fmt.Errorf("%w: %s", errors.ErrUnsupportedAppType, appType)
-	}
-	return filepath.Join(typePath, name), nil
+type appTypeHandler interface {
```
adalton: Even though the behaviors of app type handlers for `embeddedProvider` and `imageProvider` are almost the same, I'd suggest modeling them as two separate interfaces.  There are a couple of motivations:

* It seems like it's possible for the two to diverge over time.  For example, maybe someday the app type handler for an `embeddedProvider` will need some additional operation that doesn't make sense for an `imageProvider`'s app type handler (or vice versa).  (Or maybe one `Verify()` will need a directory and the other will not 😃).
* Implementing the same interface suggests a substitutability that I don't think exists here.  For example, right now, `embeddedProvider` is type compatible with `quadletBehavior` and `composeBehavior`.  (I realize that the factory functions don't currently return the wrong thing, but returning the wrong thing isn't prohibited by the type system).

Given that the interface is relatively small, I'd probably just duplicate it.  If that doesn't set well, you could extract the common behaviors into another interface that you embed in the two main ones.

internal/agent/device/applications/provider/embedded.go
```diff
@@ -72,36 +85,89 @@ func (p *embeddedProvider) Spec() *ApplicationSpec {
...
-		return fmt.Errorf("%w: %s", errors.ErrUnsupportedAppType, p.spec.AppType)
+	return p.handler.Install(ctx)
+}
+
+func (p *embeddedProvider) Remove(ctx context.Context) error {
+	return p.handler.Remove(ctx)
+}
+
+var _ appTypeHandler = (*embeddedQuadletBehavior)(nil)
+var _ appTypeHandler = (*embeddedComposeBehavior)(nil)
```
adalton: Don't you get this for free from `newEmbeddedHandler()`?

internal/agent/device/applications/provider/embedded.go
```diff
@@ -72,36 +85,89 @@ func (p *embeddedProvider) Spec() *ApplicationSpec {
...
+
+type embeddedQuadletBehavior struct {
```
adalton: Nit: You might consider putting the `appTypeHandler`-specific details in other files, and leave this file to handler only the higher-level stuff.  For example, you might separate files for each `xxxBehavior`

internal/agent/device/applications/provider/provider.go
kkyrazis: >  (Or maybe one Verify() will need a directory and the other will not 😃).

:shaking_face:

internal/agent/device/applications/provider/embedded.go
kkyrazis: Yes, it's just a personal preference of mine I guess. Especially after moving them to separate files, I'd rather see the errors in the file I'm working on, than in the file with the factory method.

adalton (approved):
LGTM -- thanks!

internal/agent/device/applications/provider/app_handler.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/flightctl/flightctl/internal/agent/client"
+	"github.com/flightctl/flightctl/internal/agent/device/applications/lifecycle"
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	"github.com/flightctl/flightctl/pkg/log"
+)
+
+type appTypeHandler interface {
+	// Dependencies returns the required applications required for the app to work
+	Dependencies() []string
```
hexfusion, line 16: since dependencies are only used in verify does the interface need to expose this

internal/agent/device/applications/provider/embedded_app_handler.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (e *embeddedComposeBehavior) Install(ctx context.Context) error {
+	// no-op
+	return nil
+}
+
+func (e *embeddedComposeBehavior) Remove(ctx context.Context) error {
+	return nil
+}
```
hexfusion, line 94: when I see these implementations empty I and or very slim wonder if this makes sense.  would it be reasonable to have a single implementation with a type enum? we can refactor later either way.

hexfusion (approved):

internal/agent/device/applications/provider/embedded_app_handler.go
kkyrazis: perhaps this would have been better to leave alone and wait to see how the next embedded application behaves before attempting this.

internal/agent/device/applications/provider/app_handler.go
kkyrazis: fair point. I'll adjust this.

internal/agent/device/applications/provider/embedded_app_handler.go
adalton: What's here is the product of an earlier suggestion I made to minimize the type enum-based approach, sorry!

PR #1948, opened 2025-11-13
EDM-2540: added Check Permission API

asafbennatan:
check permission api will be used by ui to hide /show ui elements 

## Summary by CodeRabbit

* **New Features**
  * New API endpoint to let authenticated users fetch their available permissions (returns structured permission lists).
  * Client libraries updated to support calling the new permissions endpoint and parsing responses.

* **Tests**
  * Added tests validating permission resolution across roles and identity scenarios.


api/v1alpha1/openapi.yaml
```diff
@@ -6978,3 +7027,55 @@ components:
...
         - items
       description: AuthProviderList is a list of auth providers.
+    PermissionCheckRequest:
+      type: object
+      description: Request to check multiple permissions.
+      properties:
+        permissions:
+          type: array
+          description: List of resource-operation pairs to check.
+          items:
```
avishayt, line 7293: Should we have a limit for the array length to prevent abuse?

celdrake: We discussed that ideally we'd do a single request in the UI to retrieve all possible permissions for the user.
If there's a limit we probably would need to do one request per page.

api/v1alpha1/openapi.yaml
```diff
@@ -146,6 +146,55 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/auth/checkpermission:
```
celdrake: (Not related to this line)
As mentioned, the current roles IMO need some changes.

1. All roles --> should have permission to list organizations. It's required to have the correct login flow for UI and CLI.
2. flightctl-viewer/flightctl-operator: can list resource syncs, but not repositories. Given that a RS requires a repository, that doesn't work well. At least in the UI, the RSs are nested in the repositories page, but I think the problem also happens in the CLI.

api/v1alpha1/openapi.yaml
avishayt: I meant to limit it to a number where normal use cases won't be affected - 100, 1000?

celdrake: Given that the endpoint now returns all user permissions, I think the comment does not apply any longer.

api/v1alpha1/openapi.yaml
```diff
@@ -146,6 +146,43 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /api/v1/auth/listpermissions:
```
celdrake: With the "list" prefix, we're implying we won't allow CRUD operations for permissions.
Not sure this is something we'd want to have in the future, and that would require breaking the API.

PR #1949, opened 2025-11-13
NO-ISSUE: Fix various console issues

ori-amizur:
- Remove from log input/output error when console is closed since this is normal behavior.
- Let the initial directory be the home directory of the calling process and not '/'.
- Eliminate CLI error strings when negative error codes are returned, since this is not supported by the client-go library.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stream error handling with distinct treatment for EOF, IO errors, and other failures to reduce noisy logs.
  * Corrected exit-status reporting so non-zero exits are represented as positive values in messages.
  * Session startup now detects and applies the user's home directory, setting the working directory and HOME when available.

* **Tests**
  * Updated CLI console end-to-end test to expect the current working directory output at /root.

### DIFF
--- a/.github/workflows/publish-containers.yaml
+++ b/.github/workflows/publish-containers.yaml
@@ -111,7 +111,7 @@ jobs:
         id: version-vars
         run: |
           set -euo pipefail
-          SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
+          SOURCE_GIT_TAG=$(./hack/current-version)
           if [ ! -d ".git" ]; then
             SOURCE_GIT_TREE_STATE=clean
           elif git diff --quiet; then
--- a/Makefile
+++ b/Makefile
@@ -29,34 +29,29 @@ CACHE_FLAGS_TEMPLATE := --cache-from=$(REGISTRY)/$(REGISTRY_OWNER)/%
 
 
 
-GOBASE=$(shell pwd)
-GOBIN=$(GOBASE)/bin/
-GO_BUILD_FLAGS := ${GO_BUILD_FLAGS}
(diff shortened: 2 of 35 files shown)


internal/agent/device/console/outgoing_streams.go
```diff
@@ -121,7 +124,11 @@ func (o *outgoingStdStream) run(parentWg, childWg *sync.WaitGroup) {
 			o.send(buffer[:n])
 		}
 		if err != nil {
-			if err != io.EOF {
+			switch {
+			case err == io.EOF:
+			case strings.Contains(err.Error(), "input/output error"):
```
avishayt: Any way other way to determine if it's an error or not?  This string might change based on OS and language?

internal/agent/device/console/session.go
```diff
@@ -24,6 +31,25 @@ type session struct {
...
+		if err == nil && homedir != "" {
+			return
+		}
+		var u *user.User
+		u, err = user.Current()
+		if err == nil && u.HomeDir != "" {
+			homedir = u.HomeDir
+			return
+		}
+		s.log.Warnf("unable to get user home directory for console session: %v", err)
```
avishayt: This logs err for user.Current() - do we also want to save and log err for os.UserHomeDir?

internal/agent/device/console/outgoing_streams.go
ori-amizur: I don't think it depends on the language (seems like system error).  I believe it is errno 5

internal/agent/device/console/session.go
ori-amizur: Done

internal/agent/device/console/outgoing_streams.go
ori-amizur: Changed

avishayt (approved):

PR #1954, opened 2025-11-13
EDM-2530: Flightctl version/tagging mismatch

adalton:
Currently, our GitHub action to get the current tag uses:

```bash
SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
```

Unfortunately, `actions/checkout@v4` tweaks the tag of interest in a way that makes it a lightweight "commit" tag instead of an annotated tag, for example:

```
git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules origin \
    +f147e4121aecc7729d86de8ee1056faa19abab36:refs/tags/v0.10.0
```

The `git describe` command gives preference to annotated tags over lightweight tags, so the above change results in `git describe` returning the wrong tag &ndash; an unexpected version.

This change updates the workflow to use the existing `hack/current-version` script, whose behavior does not depend on whether or not the tag is annotated or lightweight.

## Summary by CodeRabbit

* **Chores**
  * Enhanced container image versioning in the automated build pipeline for more reliable tag generation.

### DIFF
--- a/.github/workflows/publish-containers.yaml
+++ b/.github/workflows/publish-containers.yaml
@@ -111,7 +111,7 @@ jobs:
         id: version-vars
         run: |
           set -euo pipefail
-          SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
+          SOURCE_GIT_TAG=$(./hack/current-version)
           if [ ! -d ".git" ]; then
             SOURCE_GIT_TREE_STATE=clean
           elif git diff --quiet; then


.github/workflows/publish-containers.yaml
keitwb, line 34: Does anything up here need to change?

.github/workflows/publish-containers.yaml
```diff
@@ -111,7 +111,7 @@ jobs:
         id: version-vars
         run: |
           set -euo pipefail
-          SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
+          SOURCE_GIT_TAG=$(./hack/current-version)
```
keitwb, line 114: Since this is the default in the Makefile I would just omit this line altogether.

adalton: I agree; however, if I'm following the flow of this correctly, this value ends up getting passed as an `ARG` in a the container build process in the `Build` step.  I think if it's unset at that point, it'll end up being the empty string in the container build environment.  The `Makefile` uses conditional assignment, so it'll preserve the empty string value instead of running the script.

keitwb: ok, yeah looks like it is bypassing the Makefile altogether if I'm understanding it right.

.github/workflows/publish-containers.yaml
adalton: It's a bit of a Chesterton's Fence; I don't know the rationale for why this is the way it is, so I'm hesitant to change it.  I don't think that anything here contributed to the problem we're seeing with this issue.

PR #1958, opened 2025-11-14
EDM-2508: Prefetch

SiddarthR56:
## Summary by CodeRabbit

* **Refactor**
  * Two‑phase OCI collection with per‑app caching, deduplication, garbage collection and requeue signalling for more reliable, efficient updates.
  * Collectors now return richer results (including requeue) and targets include image digests.

* **New Features**
  * Image digest retrieval for images.
  * Added Quadlet support for image‑based apps and per‑app extracted data with cleanup.

* **Bug Fixes**
  * Improved retry semantics when collectors are not ready.

* **Tests**
  * New tests for cache behavior, nested‑target extraction, shared‑image transitions and error/requeue paths.

### DIFF
--- a/internal/agent/device/applications/manager.go
+++ b/internal/agent/device/applications/manager.go
@@ -287,7 +287,7 @@ func (m *manager) collectNestedTargets(
 		}
 
 		// cache miss or invalid - extract nested targets for this image
-		nestedTargets, appData, err := m.extractNestedTargetsForImage(ctx, appSpec, &imageSpec, secret)
+		appData, err := m.extractNestedTargetsForImage(ctx, appSpec, &imageSpec, secret)
 		if err != nil {
 			return nil, false, nil, fmt.Errorf("extracting nested targets for app %s: %w", appName, err)
 		}
@@ -303,12 +303,12 @@ func (m *manager) collectNestedTargets(
 				Reference: imageRef,
 				Digest:    digest,
 			},
-			Children: nestedTargets,
+			Children: appData.Targets,
 		}
 		m.ociTargetCache.Set(cacheEntry)
-		m.log.Debugf("Cached %d nested targets for app %s (digest: %s)", len(nestedTargets), appName, digest)
+		m.log.Debugf("Cached %d nested targets for app %s (digest: %s)", len(appData.Targets), appName, digest)
 
-		allNestedTargets = append(allNestedTargets, nestedTargets...)
+		allNestedTargets = append(allNestedTargets, appData.Targets...)
 	}
 
 	return allNestedTargets, needsRequeue, activeAppNames, nil
@@ -320,7 +320,7 @@ func (m *manager) extractNestedTargetsForImage(
 	appSpec v1alpha1.ApplicationProviderSpec,
 	imageSpec *v1alpha1.ImageApplicationProviderSpec,
 	secret *client.PullSecret,
-) ([]dependency.OCIPullTarget, *provider.AppData, error) {
+) (*provider.AppData, error) {
 	return provider.ExtractNestedTargetsFromImage(
 		ctx,
 		m.log,
--- a/internal/agent/device/applications/provider/image.go
+++ b/internal/agent/device/applications/provider/image.go
@@ -55,6 +55,7 @@ func newImage(log *log.PrefixLogger, podman *client.Podman, spec *v1alpha1.Appli
 		readWriter: readWriter,
 		spec: &ApplicationSpec{
 			Name:          appName,
+			ID:            client.NewComposeID(appName),
 			AppType:       appType,
 			Path:          path,
 			EnvVars:       lo.FromPtr(spec.EnvVars),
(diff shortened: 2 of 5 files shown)


internal/agent/device/applications/provider/image.go
```diff
@@ -110,37 +83,49 @@ func (p *imageProvider) Verify(ctx context.Context) error {
 		return fmt.Errorf("%w: ensuring volume dependencies: %w", errors.ErrNoRetry, err)
 	}
 
-	// create a temporary directory to copy the image contents
-	tmpAppPath, err := p.readWriter.MkdirTemp("app_temp")
-	if err != nil {
-		return fmt.Errorf("creating tmp dir: %w", err)
-	}
+	var tmpAppPath string
+	var shouldCleanup bool
```
kkyrazis, line 87: I think shouldCleanup is true in all cases once it reaches the defer statement. Is the intention for cleanup always to occur, or if the cached data already exists we shouldn't be cleaning up?

internal/agent/device/applications/provider/provider.go
```diff
@@ -58,6 +56,325 @@ type ApplicationSpec struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("collecting embedded OCI targets: %w", err)
+	}
+	targets = append(targets, embeddedTargets...)
+
+	return targets, nil
+}
+
+// collectEmbeddedOCITargets discovers embedded applications and extracts their OCI targets
+func collectEmbeddedOCITargets(ctx context.Context, readWriter fileio.ReadWriter, pullSecret *client.PullSecret) ([]dependency.OCIPullTarget, error) {
```
kkyrazis, line 166: embedded quadlets are missing. Should be realtively easy to adjust this to accommodate searching in the appropriate directories. Look at parseEmbeddedQuadlets

internal/agent/device/applications/provider/provider.go
```diff
@@ -58,6 +56,325 @@ type ApplicationSpec struct {
...
+		}
+	}
+
+	return targets, nil
+}
+
+// CollectNestedOCITargets collects nested OCI targets from image providers after base images
+// have been fetched, and aggregates them with the provided base targets.
+// This is used in phase 2 of prefetching.
+func CollectNestedOCITargets(
```
kkyrazis: Is this used in the actual flow? It seems like it's more like CollectBase and then ExtractNested rather than calling this.

hexfusion:
I would argue that the test actually passes we clearly fail and rollback its just a missmatch in the expected error string?
```
  [FAIL] VM Agent behavior during updates updates [It] Should rollback when updating to a broken image [82481, sanity]
```
We can make this message nicer right, and sooner. this means we dont error until app starts.

```
podman compose
        up: failed to resolve image short name: use the full name i.e registry/image:tag:
        code: 125: \e[4m>>>> Executing external compose provider \"/usr/bin/podman-c..
```

internal/agent/device/applications/manager.go
```diff
@@ -173,34 +190,144 @@ func (m *manager) Shutdown(ctx context.Context, state shutdown.State) error {
...
+			Children: appData.Targets,
+		}
+		m.ociTargetCache.Set(cacheEntry)
+		m.log.Debugf("Cached %d nested targets for app %s (digest: %s)", len(appData.Targets), appName, digest)
+
+		allNestedTargets = append(allNestedTargets, appData.Targets...)
+	}
+
+	return allNestedTargets, needsRequeue, activeAppNames, nil
+}
```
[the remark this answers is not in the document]
hexfusion: name should never be empty we have a pattern for ensuring that lets double check we do.

hexfusion:
LGTM

internal/agent/device/applications/provider/image.go
```diff
@@ -117,15 +118,11 @@ func (p *imageProvider) Verify(ctx context.Context) error {
 
 	switch p.spec.AppType {
 	case v1alpha1.AppTypeCompose:
-		p.spec.ID = client.NewComposeID(p.spec.Name)
```
hexfusion, line 133: why this change already exists?

internal/agent/device/applications/provider/provider.go
```diff
@@ -678,12 +661,20 @@ func extractAppDataFromOCITarget(
...
-				return nil, fmt.Errorf("parsing compose spec: %w (cleanup failed: %v)", err, rmErr)
+				return nil, fmt.Errorf("parsing compose spec for app %s (%s): %w (cleanup failed: %v)", appName, imageRef, err, rmErr)
 			}
-			return nil, fmt.Errorf("parsing compose spec: %w", err)
+			return nil, fmt.Errorf("parsing compose spec for app %s (%s): %w", appName, imageRef, err)
 		}
 
-		// extract images from compose services
+		// validate the compose spec
+		if errs := validation.ValidateComposeSpec(spec); len(errs) > 0 {
```
hexfusion, line 670: I think this makes sense honestly

internal/agent/device/applications/provider/image.go
SiddarthR56: I moved the ID assignment from Verify() to provider construction. This aligns with the pattern across all provider types.

kkyrazis (approved):

PR #1959, opened 2025-11-14
EDM-2304: Render quadlets and services config using go instead of bash

DakCrowder:
This PR encompasses adding a lightctl-standalone with two (current) capabilities:
- Renders quadlet files in our repo
  - Moves/installs quadlet files from within the repo to expected locations
  - Overlays image information using go templates and a config file, which should make downstreaming easier
- Templates service files on host once the services rpm is installed
  - In the future can contain functionality currently done in bash or elsewhere like podman secret generation, artifact cleanup, etc.
  - Additional per-service configuration files added to reduce reliance on flightctl-api config being widely used as a catchall

## Summary by CodeRabbit

* **New Features**
  * Adds a standalone CLI for rendering deployment templates and a build target to produce it; adds manifests for configurable container images/tags.

* **Refactor**
  * Startup now pre-renders configs/envs; many init scripts and init units simplified or removed and deploy flow moved to render-first.

* **Packaging**
  * Packaging updated to include the standalone tool and expanded service config/data layout.

* **Tests**
  * Adds tests for template rendering and manifest-driven install flow.

### DIFF
--- a/cmd/flightctl-standalone/render_quadlets.go
+++ b/cmd/flightctl-standalone/render_quadlets.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/flightctl/flightctl/internal/quadlet/renderer"
 	"github.com/flightctl/flightctl/pkg/log"
@@ -10,20 +9,12 @@ import (
 	"github.com/spf13/viper"
 )
 
-func main() {
-	command := NewRenderQuadletsCommand()
-	if err := command.Execute(); err != nil {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-		os.Exit(1)
-	}
-}
-
 func NewRenderQuadletsCommand() *cobra.Command {
 	var cfgFile string
 	config := renderer.NewRendererConfig()
 
 	cmd := &cobra.Command{
-		Use:   "render-quadlets",
+		Use:   "quadlets",
 		Short: "Render Flight Control service quadlet files",
 		Long:  "A tool to render Flight Control service quadlet files and systemd units from templates",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
--- a/deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
+++ b/deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
@@ -23,7 +23,7 @@ Restart=on-failure
 RestartSec=30
 TimeoutStartSec=10m
 
-ExecStartPre=/usr/bin/flightctl-ground-crew render --input-file /usr/share/flightctl/flightctl-db-migrate/config.yaml.template --output-file /etc/flightctl/flightctl-db-migrate/config.yaml
+ExecStartPre=/usr/bin/flightctl-standalone render template --input-file /usr/share/flightctl/flightctl-db-migrate/config.yaml.template --output-file /etc/flightctl/flightctl-db-migrate/config.yaml
 
 [Install]
 WantedBy=flightctl.target
--- a/deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
+++ b/deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
@@ -17,7 +17,7 @@ Volume=/etc/flightctl/flightctl-pam-issuer/config.yaml:/root/.flightctl/config.y
 Restart=always
 RestartSec=30
 
-ExecStartPre=/usr/bin/flightctl-ground-crew render --input-file /usr/share/flightctl/flightctl-pam-issuer/config.yaml.template --output-file /etc/flightctl/flightctl-pam-issuer/config.yaml
+ExecStartPre=/usr/bin/flightctl-standalone render template --input-file /usr/share/flightctl/flightctl-pam-issuer/config.yaml.template --output-file /etc/flightctl/flightctl-pam-issuer/config.yaml
 
 [Install]
 WantedBy=flightctl.target
--- a/internal/quadlet/renderer/manifest.go
+++ b/internal/quadlet/renderer/manifest.go
@@ -73,8 +73,8 @@ func servicesManifest(config *RendererConfig) []InstallAction {
 		{Action: ActionCopyFile, Source: "deploy/scripts/secrets.sh", Destination: filepath.Join(config.ReadOnlyConfigOutputDir, "secrets.sh"), Template: false, Mode: ExecutableFileMode},
 		{Action: ActionCopyFile, Source: "deploy/scripts/yaml_helpers.py", Destination: filepath.Join(config.ReadOnlyConfigOutputDir, "yaml_helpers.py"), Template: false, Mode: RegularFileMode},
 
-		// Ground crew binary
-		{Action: ActionCopyFile, Source: "bin/flightctl-ground-crew", Destination: filepath.Join(config.BinOutputDir, "flightctl-ground-crew"), Template: false, Mode: ExecutableFileMode},
+		// Standalone binary
+		{Action: ActionCopyFile, Source: "bin/flightctl-standalone", Destination: filepath.Join(config.BinOutputDir, "flightctl-standalone"), Template: false, Mode: ExecutableFileMode},
 
 		// Empty files
 		{Action: ActionCreateEmptyFile, Destination: filepath.Join(config.WriteableConfigOutputDir, "ssh", "known_hosts"), Mode: RegularFileMode},
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -428,9 +428,9 @@ echo "Flightctl Observability Stack uninstalled."
     SOURCE_GIT_COMMIT="%{?SOURCE_GIT_COMMIT:%{SOURCE_GIT_COMMIT}}%{!?SOURCE_GIT_COMMIT:%(echo %{version} | grep -o '[-~]g[0-9a-f]*' | sed 's/[-~]g//' || echo unknown)}" \
     SOURCE_GIT_TAG_NO_V="%{?SOURCE_GIT_TAG_NO_V:%{SOURCE_GIT_TAG_NO_V}}%{!?SOURCE_GIT_TAG_NO_V:%{version}}" \
     %if 0%{?rhel} == 9
-        %make_build build-cli build-agent build-restore build-render-quadlets build-ground-crew
+        %make_build build-cli build-agent build-restore build-standalone
     %else
-        DISABLE_FIPS="true" %make_build build-cli build-agent build-restore build-render-quadlets build-ground-crew
+        DISABLE_FIPS="true" %make_build build-cli build-agent build-restore build-standalone
     %endif
 
     # SELinux modules build
@@ -468,14 +468,14 @@ echo "Flightctl Observability Stack uninstalled."
     install -Dpm 0644 packaging/flightctl-services-install.conf %{buildroot}%{_sysconfdir}/flightctl/flightctl-services-install.conf
 
     # flightctl-services sub-package steps
-    # Use the render-quadlets binary to generate quadlet files with the correct image tags.
+    # Use the flightctl-standalone render quadlets command to generate quadlet files with the correct image tags.
     #
     # The IMAGE_TAG is derived from the RPM version, which may include tildes (~)
     # for proper version sorting (e.g., 0.5.1~rc1-1). However, the tagged images
     # always use hyphens (-) instead of tildes (~). To ensure valid image tags we need
     # to transform the version string by replacing tildes with hyphens.
     IMAGE_TAG=$(echo %{version} | tr '~' '-')
-    bin/render-quadlets \
+    bin/flightctl-standalone render quadlets \
         --config deploy/podman/images.yaml \
         --flightctl-services-tag-override "${IMAGE_TAG}" \
         --readonly-config-dir "%{buildroot}%{_datadir}/flightctl" \
@@ -680,7 +680,7 @@ rm -rf /usr/share/sosreport
 
     # Files mounted to bin dir
     %attr(0755,root,root) %{_bindir}/flightctl-services-must-gather
-    %attr(0755,root,root) %{_bindir}/flightctl-ground-crew
+    %attr(0755,root,root) %{_bindir}/flightctl-standalone
 
 # Optional pre-upgrade database migration dry-run
 %pre services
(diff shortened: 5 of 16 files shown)


packaging/rpm/flightctl.spec
```diff
@@ -466,18 +466,21 @@ echo "Flightctl Observability Stack uninstalled."
...
     # for proper version sorting (e.g., 0.5.1~rc1-1). However, the tagged images
     # always use hyphens (-) instead of tildes (~). To ensure valid image tags we need
     # to transform the version string by replacing tildes with hyphens.
-    CONFIG_READONLY_DIR="%{buildroot}%{_datadir}/flightctl" \
-    CONFIG_WRITEABLE_DIR="%{buildroot}%{_sysconfdir}/flightctl" \
-    QUADLET_FILES_OUTPUT_DIR="%{buildroot}%{_datadir}/containers/systemd" \
-    SYSTEMD_UNIT_OUTPUT_DIR="%{buildroot}/usr/lib/systemd/system" \
-    IMAGE_TAG=$(echo %{version} | tr '~' '-') \
-    deploy/scripts/install.sh
+    IMAGE_TAG=$(echo %{version} | tr '~' '-')
```
DakCrowder, line 477: @kkyrazis changes to the .spec here - some file diffs but I'm hoping the rendering will make downstreaming easier and we can just pass a configuration file with the different images for downstream rather than the sed scripting replacement I believe we have now.

kkyrazis: cc @gshilin-sdb 

Overall that change sounds much nicer to maintain.

deploy/podman/flightctl-cli-artifacts/flightctl-cli-artifacts-init.container
```diff
@@ -1,15 +1,16 @@
...
+PartOf=flightctl.target
 After=flightctl-api.service
 Wants=flightctl-api.service
-PartOf=flightctl.target
 
 [Container]
-Image=registry.access.redhat.com/ubi9/ubi-minimal
-ContainerName=flightctl-cli-artifacts-config
-Volume=/usr/share/flightctl/flightctl-cli-artifacts:/config-source:ro,z
+Image=registry.access.redhat.com/ubi9/ubi-minimal:9.6
```
hexfusion, line 8: smart pinning deps :+1:

internal/quadlet/renderer/renderer_test.go
```diff
@@ -0,0 +1,399 @@
...
+				verifyFileExists(t, filepath.Join(config.WriteableConfigOutputDir, "empty.txt"))
+				verifyDirExists(t, filepath.Join(config.WriteableConfigOutputDir, "emptydir"))
+			},
+		},
+		{
+			name: "create empty file that already exists",
+			setup: func(t *testing.T, sourceDir string, config *RendererConfig) []InstallAction {
+				destPath := filepath.Join(config.WriteableConfigOutputDir, "existing.txt")
+				if err := os.MkdirAll(filepath.Dir(destPath), 0755); err != nil {
+					t.Fatalf("Failed to create directory: %v", err)
```
hexfusion: in general can we align on require for unit testing.

internal/quadlet/renderer/renderer.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	if err := os.MkdirAll(destPath, mode); err != nil {
+		return fmt.Errorf("failed to create directory: %w", err)
+	}
+
+	return nil
+}
+
+func copyDir(src, dst string, mode os.FileMode) error {
```
hexfusion, line 184: feel like we have implemented this in a few places should we have a shared pkg somewhere? could be a followup

internal/quadlet/renderer/renderer.go
```diff
@@ -0,0 +1,266 @@
...
+const (
+	ActionCopyFile ActionType = iota
+	ActionCopyDir
+	ActionCreateEmptyFile
+	ActionCreateEmptyDir
+)
+
+const (
+	RegularFileMode    os.FileMode = 0644 // Regular files
+	ExecutableFileMode os.FileMode = 0755 // Executable files and directories
```
hexfusion, line 25: lovely consts for a shared pkg someday ;)

cmd/flightctl-ground-crew/main.go
```diff
@@ -0,0 +1,62 @@
...
+	if err := command.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func NewGroundCrewCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "flightctl-ground-crew [command]",
+		Short: "Flight Control utility functions for quadlet services",
```
keitwb: If this is meant for quadlets, then can it be rolled up into a subcommand of the render-quadlets command below?

go.mod
```diff
@@ -112,12 +112,14 @@ require (
...
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.130.0
 	github.com/prometheus/client_model v0.6.2
+	github.com/spf13/viper v1.20.1
 	go.opentelemetry.io/collector/consumer/consumererror v0.130.1
 	go.opentelemetry.io/otel/exporters/prometheus v0.59.1
 	go.opentelemetry.io/otel/sdk/metric v1.37.0
 	go.uber.org/zap v1.27.0
 	golang.org/x/net v0.46.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
+	gopkg.in/yaml.v2 v2.4.0
```
hexfusion: I thought we aligned on `sigs.k8s.io/yaml` not so concerned on which but we should standardize on one.

cmd/render-quadlets/main.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+func NewRenderQuadletsCommand() *cobra.Command {
+	var cfgFile string
+	config := renderer.NewRendererConfig()
+
+	cmd := &cobra.Command{
+		Use:   "render-quadlets",
+		Short: "Render FlightCtl service quadlet files",
```
hexfusion: are we Flight Control now?

internal/quadlet/renderer/renderer.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+func processInstallManifest(manifest []InstallAction, config *RendererConfig) error {
+	for _, action := range manifest {
+		switch action.Action {
+		case ActionCopyFile:
+			if err := processFile(action.Source, action.Destination, action.Template, action.Mode, config); err != nil {
+				return fmt.Errorf("failed to process file %s: %w", action.Source, err)
+			}
+			fmt.Printf("Processed file: %s -> %s (template=%t)\n", action.Source, action.Destination, action.Template)
```
hexfusion: should we use a logger?

deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
```diff
@@ -23,5 +23,7 @@ Restart=on-failure
 RestartSec=30
 TimeoutStartSec=10m
 
+ExecStartPre=/usr/bin/flightctl-ground-crew render --input-file /usr/share/flightctl/flightctl-api/config.yaml.template --output-file /etc/flightctl/flightctl-db/config.yaml
```
keitwb: Is the inconsistency between the input dir and output dir correct?

internal/quadlet/renderer/renderer.go
DakCrowder: Yes - I even reviewed @kkyrazis previous PR and he added a much nicer copyDir method :sweat_smile: but that needs to be pulled out of the agent and didn't want to tackle all of that here - but very much agree this should live in /pkg which would make the renderer quite thin

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
```diff
@@ -19,8 +17,7 @@ Volume=/etc/flightctl/flightctl-pam-issuer/config.yaml:/root/.flightctl/config.y
 Restart=always
 RestartSec=30
 
+ExecStartPre=/usr/bin/flightctl-ground-crew render --input-file /usr/share/flightctl/flightctl-pam-issuer/config.yaml.template --output-file /etc/flightctl/flightctl-pam-issuer/config.yaml
```
keitwb: Could "ground crew" just be rolled up into the cli package then? This seems consistent with kubectl having the kustomize rendering built into it.

deploy/podman/flightctl-cli-artifacts/flightctl-cli-artifacts-init.container
DakCrowder: Down with :latest !

go.mod
DakCrowder: Ahh good point will take a pass as aligning, `sigs.k8s.io/yaml` is widely used and i'll convert

deploy/podman/local-images.yaml
```diff
@@ -0,0 +1,36 @@
+api:
```
keitwb, line 1: Should this be committed to git?

internal/quadlet/renderer/manifest.go
```diff
@@ -0,0 +1,94 @@
+package renderer
+
+import "path/filepath"
+
+func servicesManifest(config *RendererConfig) []InstallAction {
+	return []InstallAction{
+		// API service
+		{Action: ActionCopyFile, Source: "deploy/podman/flightctl-api/flightctl-api.container", Destination: filepath.Join(config.QuadletFilesOutputDir, "flightctl-api.container"), Template: true, Mode: RegularFileMode},
```
keitwb, line 8: I'm a bit wary about this since it seems to overlap pretty hard with the functionality of the RPM `%files` section. What is the advantage of this?

cmd/flightctl-ground-crew/main.go
keitwb: Or rolled into the cli binary as a utility command

deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
DakCrowder: good catch no was an oversight

DakCrowder: good catch no was an oversight

deploy/podman/local-images.yaml
DakCrowder: As it stands now yes, it is being referenced in the local deployment option `make deploy-quadlets`

internal/quadlet/renderer/manifest.go
DakCrowder: We currently allow for non-rpm based installation for local dev using `make deploy-quadlets` and this is doing templating as well.

I could see  (and maybe we should) doing some form of rpm build/install in the future as `make deploy-quadlets` but there are some downsides like the build, still having to do some manual updates to the images and/or rpm build process to get what we want working locally.

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
DakCrowder: Right now this particular functionality isn't exposed to end users (and I'm not sure if the render functionality is desired in that way) - as well as if more options are added they might be fairly quadlets-specific, things like must-gather or cleaning up podman artifacts or other on host setup.  I see value in the isolation at least in the short term until the methods exposed get more sorted.

internal/quadlet/renderer/manifest.go
DakCrowder: Also I'd add that %files describes the end state of what is laid down not the copying/installation, another option would be to programatically write %files based on the manifest but that is also opening up a whole can of worms.

cmd/flightctl-standalone/render_quadlets.go
```diff
@@ -0,0 +1,68 @@
+package main
```
hexfusion, line 1: do we need a new cmd namespace or can do something like

```
flightctl-standalone render quadlets ....
flightctl-standalone render template ....
flightctl-standalone must-gather
flightctl-standalone status
```

DakCrowder: Seems clean to me - only real consideration is _currently_ the render quadlets would never be run by an end user but shouldn't be a big one.  Can take a pass at this

packaging/rpm/flightctl.spec
```diff
@@ -428,9 +428,9 @@ echo "Flightctl Observability Stack uninstalled."
     SOURCE_GIT_COMMIT="%{?SOURCE_GIT_COMMIT:%{SOURCE_GIT_COMMIT}}%{!?SOURCE_GIT_COMMIT:%(echo %{version} | grep -o '[-~]g[0-9a-f]*' | sed 's/[-~]g//' || echo unknown)}" \
     SOURCE_GIT_TAG_NO_V="%{?SOURCE_GIT_TAG_NO_V:%{SOURCE_GIT_TAG_NO_V}}%{!?SOURCE_GIT_TAG_NO_V:%{version}}" \
     %if 0%{?rhel} == 9
-        %make_build build-cli build-agent build-restore
+        %make_build build-cli build-agent build-restore build-standalone
     %else
-        DISABLE_FIPS="true" %make_build build-cli build-agent build-restore
+        DISABLE_FIPS="true" %make_build build-cli build-agent build-restore build-standalone
     %endif
```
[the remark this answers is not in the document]
DakCrowder: This happens in the invoked render quadlets cmd

DakCrowder:
@asafbennatan After building out the containerized templating (initial pr was [here](https://github.com/flightctl/flightctl/pull/1909)) it raised some additional questions with the approach - like is generic templating something that really needs to be shipped in every container?  It seems like solely a quadlet problem to me.  Quadlets are the wild west and we don't have niceties like helm and I'm not sure if users deploying in k8s environments would ever have these problems.  I  think there is value in potentially _some_ containerized operations **and** flightctl-standalone running on the host.

1.  There is value to flightctl-standalone replacing a number of current operations, such as must-gather, cleaning up artifacts, [init-host](https://github.com/flightctl/flightctl/blob/main/deploy/scripts/init_host.sh), some yaml_helpers.py usage and I would say templating as its (currently) only a quadlets concern and this feels simpler to get go templates out the door
2. There are operations such as conditionally creating an [aap oauth application](https://github.com/flightctl/flightctl/blob/main/deploy/podman/flightctl-api/flightctl-api-config/create_aap_application.sh), certs generation that is currently part of starting the api service, db wait etc. that I think are well suited to be in a container init process and should eventually be moved there

So I don't think they are mutually exclusive, and we could swap them later on as I don't think standalone templating is set in stone

go.mod
```diff
@@ -112,6 +112,7 @@ require (
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter v0.130.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver v0.130.0
 	github.com/prometheus/client_model v0.6.2
+	github.com/spf13/viper v1.20.1
```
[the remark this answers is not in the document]
DakCrowder: The mapstructure version specified in the indirect go.mod is 2.4 not the problematic version specified in the cve link

PR #1961, opened 2025-11-16
NO-ISSUE: make flightctl-admin service account as part of the helm deployment

asafbennatan:
previously  flightctl-admin (renamed from flightctl-user) was created as part of deploy_with_helm.sh , 
instead we are now creating it as part of flightctl chart - in addition this is the way to manage flightctl on openshift/k8,
using kubeadmin will not be supported out of the box

adds support for organization scoped roles

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added native OpenShift and OAuth2 authentication provider support
  * Implemented automatic authentication type detection for Kubernetes deployments
  * Auto-creation of admin service accounts during deployment

* **Configuration Updates**
  * Simplified authentication configuration structure
  * Removed legacy API URL field; now centralized configuration
  * Image pull secrets now require manual creation

* **Authorization**
  * Introduced organization-scoped role assignments
  * Added new admin and viewer role definitions
  * Improved permission enforcement across organizations


deploy/helm/flightctl/README.md
```diff
@@ -295,6 +295,7 @@ For more detailed configuration options, see the [Values](#values) section below
 | global.auth.caCert | string | `""` | The custom CA cert. |
 | global.auth.insecureSkipTlsVerify | bool | `false` | True if verification of authority TLS cert should be skipped. |
 | global.auth.k8s.apiUrl | string | `"https://kubernetes.default.svc"` | API URL of k8s cluster that will be used as authentication authority |
+| global.auth.k8s.createDefaultUser | bool | `true` | Create default flightctl-user ServiceAccount with admin access |
```
celdrake: Maybe rename to `createAdminUser` for clarity?

```suggestion
| global.auth.k8s.createAdminUser | bool | `true` | Create default flightctl-admin ServiceAccount |
```

api/v1alpha1/consts.go
```diff
@@ -101,8 +101,16 @@ const (
 	SystemComponentDB    = "database"
 	SystemComponentQueue = "queue"
 
-	// Role constants for flightctl user roles
+	// External role names - these come from authentication providers and are mapped to internal roles
+	ExternalRoleAdmin     = "flightctl-admin"
```
celdrake, line 105: Could you summarize the new roles, and whether they all can be used when defining new authentication providers with the static mapping?

api/v1alpha1/openapi.yaml
```diff
@@ -7112,6 +7112,10 @@ components:
           items:
             type: string
           description: The JSON path to the role/group claim (e.g., ["groups"], ["roles"], ["realm_access", "roles"]).
+        separator:
```
celdrake, line 7220: Not sure I understand this, could you please explain?

api/v1alpha1/consts.go
```diff
@@ -101,8 +101,16 @@ const (
...
+	ExternalRoleOrgAdmin  = "flightctl-org-admin"
+	ExternalRoleOperator  = "flightctl-operator"
+	ExternalRoleViewer    = "flightctl-viewer"
+	ExternalRoleInstaller = "flightctl-installer"
+
+	// Internal role constants - used within flightctl for authorization
 	RoleAdmin     = "admin"     // Full access to all resources
```
celdrake, line 112: Role definitions still don't allow users (other than admimns) to query `api/organizations`, which is a requirement for properly logging to the system.

https://github.com/asafbennatan/flightctl/blob/flightctl-user-as-part-of-chart/deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml#L26

deploy/helm/flightctl/templates/flightctl-default-user-k8.yaml
```diff
@@ -0,0 +1,47 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: org-default
+subjects:
+  - kind: ServiceAccount
+    name: flightctl-admin
+    namespace: {{ default .Release.Namespace .Values.global.auth.k8s.rbacNs }}
+{{- end }}
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/helm/flightctl/README.md
asafbennatan: fixed

api/v1alpha1/consts.go
asafbennatan: flightctl-admin -> flightctl super user , role is expected globally only ["flightctl-admin"] or ["*:flightctl-admin] , it will automatically add flightctl-org-admin to all of the user's orgs
flightctl-org-admin-> has access to everything in the org


in general we expect the roles array to be likeso:
["org1:"role1,.....]
where * and missing separator are global roles

api/v1alpha1/openapi.yaml
asafbennatan: see above

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-config.yaml
```diff
@@ -55,6 +55,15 @@ data:
...
         rbacNs: {{ default .Release.Namespace .Values.global.auth.k8s.rbacNs }}
+      {{- else if eq .Values.global.auth.type "openshift"}}
+      openshift:
+        clusterControlPlaneUrl: {{ .Values.global.auth.openshift.clusterControlPlaneUrl }}
+        authorizationUrl: {{ include "flightctl.getOpenShiftOAuthAuthorizationUrl" . }}
+        tokenUrl: {{ include "flightctl.getOpenShiftOAuthTokenUrl" . }}
+        issuer: {{ include "flightctl.getOpenShiftOAuthIssuer" . }}
+        clientId: {{ include "flightctl.getOpenShiftOAuthClientId" . }}
+        clientSecret: {{ .Values.global.auth.openshift.oauth.clientSecret | default (randAlphaNum 32) }}
+        rbacNs: {{ default .Release.Namespace .Values.global.auth.openshift.rbacNs }}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/authn/openshift_auth.go
```diff
@@ -0,0 +1,260 @@
...
+
+	_ = provider.Spec.FromOpenShiftProviderSpec(o.spec)
+
+	return &api.AuthConfig{
+		ApiVersion:           api.AuthConfigAPIVersion,
+		DefaultProvider:      o.metadata.Name,
+		OrganizationsEnabled: lo.ToPtr(true),
+		Providers:            &[]api.AuthProvider{provider},
+	}
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/authn/openshift_auth.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Create OpenShift identity
+	openshiftIdentity := common.NewOpenShiftIdentity(username, uid, reportedOrganizations, issuer, *o.spec.ClusterControlPlaneUrl)
+	openshiftIdentity.SetSuperAdmin(isSuperAdmin)
+
+	// Store identity in cache
+	o.identityCache.Set(tokenHash, openshiftIdentity, ttlcache.DefaultTTL)
+
+	return openshiftIdentity, nil
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/authz.go
```diff
@@ -5,22 +5,107 @@ import (
...
+		K8sClient: k8sClient,
+		Namespace: rbacNs,
+		Log:       m.log,
+	}
+
+	// Store in cache
+	m.k8sAuthZCache.Set(apiUrl, k8sAuthZ, ttlcache.DefaultTTL)
+
+	return k8sAuthZ, nil
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

api/v1alpha1/consts.go
celdrake: And we can have static role mapping to all the roles, including global admin?

api/v1alpha1/openapi.yaml
celdrake: That one is clear now, thanks.

api/v1alpha1/consts.go
asafbennatan: we have static mapping for all known roles

api/v1alpha1/consts.go
hexfusion, line 112: if this was something like superuser. then the rest of roles would self document to roles within an org? Or maybe I am not understanding. is RoleViewer on an org level or is this admin r/o. I think we could clarify this a bit.

api/v1alpha1/openapi.yaml
```diff
@@ -6889,13 +6889,15 @@ components:
       oneOf:
         - $ref: '#/components/schemas/OIDCProviderSpec'
         - $ref: '#/components/schemas/OAuth2ProviderSpec'
+        - $ref: '#/components/schemas/OpenShiftProviderSpec'
         - $ref: '#/components/schemas/AapProviderSpec'
```
hexfusion, line 6893: could `AAP` be on `OpenShift`? How would this map?

api/v1alpha1/openapi.yaml
```diff
@@ -7005,6 +7007,47 @@ components:
...
         - roleAssignment
+    OpenShiftProviderSpec:
+      type: object
+      description: OpenShiftProviderSpec describes an OpenShift OAuth provider configuration.
+      properties:
+        providerType:
+          type: string
+          enum: [openshift]
+          description: The type of authentication provider.
+        displayName:
```
hexfusion, line 7018: seems like these providers use shared fields instead or copy would it make sense to make these their own type shared among providers? i understand not the highest prio but it could reduce human error.

api/v1alpha1/openapi.yaml
```diff
@@ -7049,9 +7092,6 @@ components:
         apiUrl:
           type: string
           description: The internal Kubernetes API URL.
-        externalOpenShiftApiUrl:
-          type: string
-          description: The external OpenShift API URL (for external access).
```
hexfusion, line 7054: for the record  why is this removed

deploy/helm/flightctl/README.md
```diff
@@ -306,7 +304,19 @@ For more detailed configuration options, see the [Values](#values) section below
...
+| global.auth.openshift.oauth.authorizationUrl | string | `""` | OAuth authorization URL (leave empty to auto-detect from OpenShift cluster) |
+| global.auth.openshift.oauth.clientId | string | `""` | OAuth client ID (will be set to flightctl-{releaseName}) |
+| global.auth.openshift.oauth.clientSecret | string | `""` | OAuth client secret (leave empty for auto-generation) |
+| global.auth.openshift.oauth.enabled | bool | `true` | Enable OAuthClient and OAuth configuration |
+| global.auth.openshift.oauth.issuer | string | `""` | OAuth issuer URL (defaults to authorizationUrl if not specified) |
+| global.auth.openshift.oauth.tokenUrl | string | `""` | OAuth token URL (leave empty to auto-detect from OpenShift cluster) |
+| global.auth.openshift.rbacNs | string | `""` | Namespace that should be used for the RBAC checks |
+| global.auth.type | string | `"k8s"` | Type of the auth to use. Can be one of 'k8s', 'oidc', 'builtin', 'aap', 'openshift', or 'none' Note: 'builtin' is a legacy mode that translates to 'oidc' with PAM issuer automatically enabled For new deployments, explicitly set type to 'oidc' and configure pamOidcIssuer settings |
```
hexfusion: cc @rawagner

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
```diff
@@ -13,10 +12,9 @@ rules:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
   name: flightctl-viewer
```
hexfusion, line 17: back to my question above. If this is ClusterRole this means this is admin-readonly as the user has access to all namespaces? Is there a Role for a user to have org-readonly access?

celdrake (commented):
LGTM from the UI's perspective, delegating to backend team for proper code review.

internal/auth/authz/openshift.go
```diff
@@ -0,0 +1,135 @@
...
+		ttlcache.WithTTL[string, bool](5 * time.Minute),
+	)
+	go cache.Start()
+
+	return &OpenShiftAuthZ{
+		K8sClient: k8sClient,
+		Log:       log,
+		Cache:     cache,
+	}
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/auth.go
```diff
@@ -122,6 +134,33 @@ func initK8sAuth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMiddle
...
+	apiUrl := strings.TrimSuffix(*cfg.Auth.OpenShift.ClusterControlPlaneUrl, "/")
+	log.Infof("OpenShift auth enabled: %s", apiUrl)
+
+	var k8sClient k8sclient.K8SClient
+	var err error
+	if apiUrl == k8sApiService {
+		k8sClient, err = k8sclient.NewK8SClient()
+	} else {
+		k8sClient, err = k8sclient.NewK8SExternalClient(apiUrl, cfg.Auth.InsecureSkipTlsVerify, cfg.Auth.CACert)
+	}
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/helm/flightctl/templates/flightctl-oauthclient.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+    app: flightctl
+    component: oauth-client
+secrets:
+  - {{ include "flightctl.getOpenShiftOAuthClientId" . }}-secret
+redirectURIs:
+  - "{{ include "flightctl.getHttpScheme" . }}://ui.{{ include "flightctl.getBaseDomain" . }}/oauth2/callback"
+grantMethod: auto
+{{- end }}
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/helm/flightctl/templates/flightctl-default-user-openshift.yaml
```diff
@@ -0,0 +1,48 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: org-default
+subjects:
+  - kind: ServiceAccount
+    name: flightctl-admin
+    namespace: {{ .Release.Namespace }}
+{{- end }}
+
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/helm/flightctl/templates/flightctl-default-user-openshift.yaml
```diff
@@ -0,0 +1,48 @@
...
+    app: flightctl
+    component: default-user
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: org-default
+subjects:
+  - kind: ServiceAccount
+    name: flightctl-admin
+    namespace: {{ .Release.Namespace }}
```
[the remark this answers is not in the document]
asafbennatan: .Values.global.auth.k8s.rbacNs  is not relevant anymore and default-org is not relevant anymore

deploy/helm/flightctl/templates/api/flightctl-api-deployment.yaml
```diff
@@ -105,6 +105,11 @@ spec:
               name: flightctl-api-k8s-token
               readOnly: true
             {{- end}}
+            {{- if .Values.global.auth.openshift.externalApiToken}}
+            - mountPath: /var/flightctl/k8s
+              name: flightctl-api-openshift-token
+              readOnly: true
+            {{- end}}
```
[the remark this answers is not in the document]
asafbennatan: they are mutual exclusive - we have a type selector

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-deployment.yaml
```diff
@@ -77,6 +77,16 @@ spec:
...
+            {{- if .Values.global.auth.k8s.externalApiToken}}
+            - mountPath: /var/flightctl/k8s
+              name: flightctl-alertmanager-proxy-k8s-token
+              readOnly: true
+            {{- end}}
+            {{- if .Values.global.auth.openshift.externalApiToken}}
+            - mountPath: /var/flightctl/k8s
+              name: flightctl-alertmanager-proxy-openshift-token
+              readOnly: true
+            {{- end}}
```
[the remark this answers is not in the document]
asafbennatan: as they cannot both be set

api/v1alpha1/openapi.yaml
asafbennatan: we have a dedicated openshift type so there is no need to mix this into the k8 type

api/v1alpha1/openapi.yaml
asafbennatan: i am actually not sure - we need to discuss each deployment configuration and understand what we want to happen / what should be the default 

in that case would users receive a token from AAP or would they need to authenticate like they do in normal openshift ?

api/v1alpha1/consts.go
asafbennatan: all roles are org level , 
the only global one is "admin" - kept this naming so we dont break existing stuff (we would need to adjust couple of places to do that ) - should we do this now ?

internal/auth/authz.go
```diff
@@ -5,22 +5,135 @@ import (
...
+		return nil, fmt.Errorf("failed to create k8s client for OpenShift authZ: %w", err)
+	}
+
+	openshiftAuthZ := authz.NewOpenShiftAuthZ(m.ctx, k8sClient, m.log)
+
+	// Store in cache with TTL
+	m.openshiftAuthZCache.Set(apiUrl, openshiftAuthZ, ttlcache.DefaultTTL)
+
+	return openshiftAuthZ, nil
+}
```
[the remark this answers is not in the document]
asafbennatan: i dont think this can happen - we are checking permissions only after the server starts ,
and if a developer "forgets" to call Start - > i dont mind the code panicing

deploy/helm/flightctl/templates/flightctl-default-user-roles.yaml
asafbennatan: this does not imply access to all namespaces , the reason this is a `ClusterRole` is so we can provide sensible out-of-the-box roles across all namespaces 

so admins that want to grant view all permissions in a org "my-org" to a specific user- "bob" would :
1. make sure that `"bob"` has project `"my-org"`
2. create `RoleBindings` with the flightctl `ClusterRole` `"flightctl-viewer"` and `"bob"` in the namespace `"my-org"`

this is the case for openshift.

for vanilla k8 we still create cluster roles for simplicity even though we do not support multi org (since we cannot know what are the namespaces of a user since we dont have Openshifts Projects)
in  this case `RoleBindings` are created in the flightctl release namespace

as a note:
1.  the existence of these roles do not prevent the user from creating custom roles and binding them in the correct namespace
2. during the meeting today this was raised as a concern since this would prevent running multiple flightctl instances on the same cluster for example - but we have opted to do so for now since we do not have a use case like that , if we did want to avoid `ClusterRole` that would imply one of these two options:
 1. flightctl will not provide out of the box roles
 2. we would need a flightctl specific controller that would auto-create roles on annotated projects/namespaces -       this was dimmed out of scope for now

api/v1alpha1/openapi.yaml
asafbennatan: they dont have exactly the same fields there is variability - question is should we favor creating an intermediate object that would host these shared fields even though these fields are derived from the actual provider's capabilities and not the result of something we can control directly so if in the future we add another provider that does have some of these shared fields and not the others it would break that shared object

hexfusion: only makes sense if they are reused

deploy/helm/flightctl/values.yaml
```diff
@@ -62,6 +66,27 @@ global:
       apiUrl: ""
       # -- The URL of the AAP Gateway API endpoint that is reachable by clients
       externalApiUrl: ""
+      # -- Enabled AAP authentication
+      enabled: true
+    openshift:
+      # -- Create default flightctl-admin ServiceAccount with admin access
+      createAdminUser: true
+      # -- OpenShift cluster control plane API URL for RBAC checks (leave empty for auto-detection)
+      clusterControlPlaneUrl: ""
```
rawagner: dont we want to provide default here 
```suggestion
      clusterControlPlaneUrl: "https://kubernetes.default.svc"
```

which would also align with k8s option ? We dont actually do any detection, we just default to this value if none is provided.

deploy/helm/flightctl/values.yaml
```diff
@@ -62,6 +66,27 @@ global:
...
+      # -- OAuth client ID (will be set to flightctl-{releaseName})
+      clientId: ""
+      # -- OAuth client secret (leave empty for auto-generation)
+      clientSecret: ""
+      # -- OAuth authorization URL (leave empty to auto-detect from OpenShift cluster)
+      authorizationUrl: ""
+      # -- OAuth token URL (leave empty to auto-detect from OpenShift cluster)
+      tokenUrl: ""
+      # -- OAuth issuer URL (defaults to authorizationUrl if not specified)
+      issuer: ""
```
rawagner, line 79:
```suggestion
      issuerUrl: ""
```

deploy/helm/flightctl/values.yaml
```diff
@@ -13,29 +13,33 @@ global:
...
+  # -- External API URL for client access and authentication, used by the multiclusterhub operator
+  apiUrl: ""
   auth:
-    # -- Type of the auth to use. Can be one of 'k8s', 'oidc', 'builtin', 'aap', or 'none'
+    # -- Type of the auth to use. Can be one of 'k8s', 'oidc', 'builtin', 'aap', 'openshift', or 'none'
     # Note: 'builtin' is a legacy mode that translates to 'oidc' with PAM issuer automatically enabled
     # For new deployments, explicitly set type to 'oidc' and configure pamOidcIssuer settings
-    type: "k8s" # k8s, oidc, builtin (legacy), aap, none
+    # Leave empty (default) to auto-detect: 'openshift' on OpenShift clusters, 'k8s' otherwise
+    type: "" # Auto-detect (openshift on OpenShift, k8s otherwise), or explicitly set: k8s, oidc, builtin, aap, openshift, none
```
rawagner: can we get rid of `builtin` now ? Why keeping legacy option, when we dont have to ?
Also Im not entirely sure what the difference between auto and builtin option would be. It seems like unnecessary complication

deploy/helm/flightctl/values.yaml
```diff
@@ -13,29 +13,33 @@ global:
...
-      # -- API URL of OpenShift cluster that can be accessed by external client to retrieve auth token
-      externalOpenShiftApiUrl: ""
       # -- In case flightctl is not running within a cluster, you can provide api token
       externalApiToken: ""
       # -- Namespace that should be used for the RBAC checks
       rbacNs: ""
+      # -- Create default flightctl-admin ServiceAccount with admin access
+      createAdminUser: true
+      # -- Enabled Kubernetes authentication
+      enabled: true
```
rawagner: Having both `enabled` & auth `type` is a bit confusing. What if I set type=k8s and k8s.enabled=false, oidc=true ?
Maybe remove the `enabled` and instead make the `type` an array of enabled auths ?

docs/user/kubernetes-auth.md
```diff
@@ -15,7 +15,7 @@ If you want to use a different cluster as auth authority, the following variable
 * Set **global.auth.k8s.apiUrl** to API URL of the external k8s cluster
 * Set **global.auth.k8s.externalApiToken** to a token which has permission to CREATE `authentication.k8s.io/tokenreview` resource in the external k8s cluster
 
-If the k8s cluster is an OpenShift cluster, you can also set **global.auth.k8s.externalOpenShiftApiUrl** which enables interactive login for the CLI (`flightctl login <fctl_url> --web`)
+If the k8s cluster is an OpenShift cluster, you can also set **global.apiUrl** which enables interactive login for the CLI (`flightctl login <fctl_url> --web`)
```
rawagner: this option is specifically for ACM, which we are getting rid of.

Should we instead just mention that if openshift auth is enabled, interactive login will work ?

deploy/helm/flightctl/templates/flightctl-oauthclient.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: OAuthClient
+metadata:
+  name: {{ include "flightctl.getOpenShiftOAuthClientId" . }}
+  labels:
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+    app: flightctl
+    component: oauth-client
+secret: {{ include "flightctl.getOpenShiftOAuthClientSecret" . | quote }}
+redirectURIs:
+  - "{{ include "flightctl.getHttpScheme" . }}://ui.{{ include "flightctl.getBaseDomain" . }}/oauth2/callback"
```
rawagner: you can use `lightctl.getUIUrl` to get the  UI URL
```suggestion
  - "{{ include "flightctl.getUIUrl" . }}/oauth2/callback"
```

rawagner: @celdrake can you take a look at the callback URL so that UI is in sync ?

deploy/helm/flightctl/values.yaml
asafbennatan: can we do any meaningful detection ?

deploy/helm/flightctl/values.yaml
asafbennatan: the reason we have this is that we expect the usage to be something like :
1. admin user uses the defined config here to login
2.configures additional dynamic auth providers and disables this original one

setting this to enabled=false would achieve that 

note the currently using none would mean - disable authentication

deploy/helm/flightctl/values.yaml
asafbennatan: yes

docs/user/kubernetes-auth.md
asafbennatan: fixed

deploy/helm/flightctl/templates/flightctl-oauthclient.yaml
asafbennatan: fixed

deploy/helm/flightctl/values.yaml
asafbennatan: actually the field name is issuer , not issuerUrl

do you suggest a name change to the config ?

deploy/helm/flightctl/values.yaml
asafbennatan: fixed

pkg/k8sclient/k8s_client.go
```diff
@@ -85,6 +87,37 @@ func (k *k8sClient) ListRoleBindings(ctx context.Context, namespace string) (*rb
...
+	}
+
+	var roleNames []string
+	for _, binding := range roleBindings.Items {
+		for _, subject := range binding.Subjects {
+			if subject.Kind == "User" && subject.Name == username {
+				roleNames = append(roleNames, binding.RoleRef.Name)
+				break
+			}
+			if subject.Kind == "ServiceAccount" && fmt.Sprintf("system:serviceaccount:%s:%s", subject.Namespace, subject.Name) == username {
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/auth/authn/k8s.go
```diff
@@ -2,42 +2,40 @@ package authn
...
-		cache:     ttlcache.New(ttlcache.WithTTL[string, *k8sAuthenticationV1.TokenReview](5 * time.Second)),
+		metadata:      metadata,
+		spec:          spec,
+		k8sClient:     k8sClient,
+		cache:         ttlcache.New(ttlcache.WithTTL[string, *k8sAuthenticationV1.TokenReview](5 * time.Second)),
+		identityCache: ttlcache.New(ttlcache.WithTTL[string, common.Identity](30 * time.Second)),
 	}
 	go authN.cache.Start()
+	go authN.identityCache.Start()
 	return authN, nil
```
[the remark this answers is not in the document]
asafbennatan: fixed

deploy/helm/flightctl/templates/api/flightctl-api-openshift-token-secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{ if .Values.global.auth.openshift.externalApiToken }}
+kind: Secret
+apiVersion: v1
+metadata:
+  name: flightctl-api-openshift-token
+  namespace: {{ .Release.Namespace }}
+stringData:
+  token: {{ .Values.global.auth.openshift.externalApiToken }}
+type: Opaque
+{{ end }}
```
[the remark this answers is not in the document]
asafbennatan: this is not something consistent in the codebase

deploy/helm/flightctl/values.yaml
rawagner: Yeah, I propose to align the naming with the other fields (`authorizationUrl`, `tokenUrl`) which do have the `Url` suffix

deploy/helm/flightctl/values.yaml
asafbennatan: removed from values.yaml - after a discussion @rawagner suggested we should have it documented outside of the normal helm README

rawagner (approved):

PR #1970, opened 2025-11-17
EDM-2295: Clean up quadlet podman resources

kkyrazis:
* Adds behavior similar to compose that to clean up resources that may be left behind by podman when quadlets are removed. 
* Adds the ability to specify other filters for Podman calls to get networks or volumes
* Removes image based volumes from quadlet applications 
* Fixes bug with [Pod] labels.

Note:
Quadlets can be defined as:
```ini
#test.container
[Container]
...
Volume=name:/mnt/path
Volume=test.volume:/mnt/path2


#test.volume
[Volume]
Driver=image
Image=<>
```

get installed as:
```ini
#my-app-123-test.container
[Container]
...
Volume=my-app-123-name:/mnt/path #named volume is namespaced
Volume=my-app-123-test.volume:/mnt/path2 #quadlet reference is namespaced
Label=project=my-app-123 #tracking label is added.


#my-app-123-test.volume
[Volume]
Driver=image
Image=<>
Label=project=my-app-123 #tracking label is added.
```

Any resources (networks, volumes) that are created directly from a [Container] do **NOT** have the project label applied to them. To account for this, we allow filtering based on the project label OR the namespacing ID.

## Summary by CodeRabbit

* **New Features**
  * Broader Quadlet support for volumes, networks, pods, images and containers
  * Volume listing with label/filter support and volume driver/mount inspection
  * Pod-specific labeling support for Quadlets

* **Bug Fixes**
  * Improved Quadlet service name resolution with safer fallbacks
  * More reliable error messages during resource cleanup

* **Improvements**
  * Centralized Podman-backed resource cleanup with aggregated error reporting
  * Tests extended to cover Podman cleanup scenarios and argument matching

### DIFF
--- a/internal/agent/client/podman.go
+++ b/internal/agent/client/podman.go
@@ -376,6 +376,41 @@ func (p *Podman) CreateVolume(ctx context.Context, name string, labels []string)
 	return mountpoint, nil
 }
 
+type podmanVolume struct {
+	Name string `json:"Name"`
+}
+
+func (p *Podman) ListVolumes(ctx context.Context, labels []string, filters []string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, p.timeout)
+	defer cancel()
+
+	args := []string{
+		"volume",
+		"ls",
+		"--format",
+		"json",
+	}
+	args = applyFilters(args, labels, filters)
+	stdout, stderr, exitCode := p.exec.ExecuteWithContext(ctx, podmanCmd, args...)
+	if exitCode != 0 {
+		return nil, fmt.Errorf("list volumes: %w", errors.FromStderr(stderr, exitCode))
+	}
+	var podVols []podmanVolume
+	err := json.Unmarshal([]byte(strings.TrimSpace(stdout)), &podVols)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal volumes: %w", err)
+	}
+	volumesSeen := make(map[string]struct{})
+	volumes := make([]string, 0, len(podVols))
+	for _, volume := range podVols {
+		if _, ok := volumesSeen[volume.Name]; !ok {
+			volumesSeen[volume.Name] = struct{}{}
+			volumes = append(volumes, volume.Name)
+		}
+	}
+	return volumes, nil
+}
+
 func (p *Podman) VolumeExists(ctx context.Context, name string) bool {
 	ctx, cancel := context.WithTimeout(ctx, p.timeout)
 	defer cancel()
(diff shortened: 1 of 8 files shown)


internal/agent/client/podman.go
```diff
@@ -376,6 +376,39 @@ func (p *Podman) CreateVolume(ctx context.Context, name string, labels []string)
...
 
+func (p *Podman) ListVolumes(ctx context.Context, labels []string, filters []string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, p.timeout)
+	defer cancel()
+
+	args := []string{
+		"volume",
+		"ls",
+		"--format",
+		"{{.Name}}",
```
hexfusion: why not deserialize to "json" seems like it would be less error prone and simpler vs hacking list of strings

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -96,6 +98,50 @@ func (q *Quadlet) remove(ctx context.Context, action *Action) error {
...
+	}
+	var errs []error
+	if err := cleanPodmanResources(ctx, q.podman, labels, filters); err != nil {
+		errs = append(errs, fmt.Errorf("cleaning podman resources: %w", err))
+	}
+
+	// The agent currently doesn't distinguish between "stop for a graceful shutdown", and "remove application". Both elicit
+	// a "remove" operation. Most resources are fine to remove and recreate on startup, as they should be application specific,
+	// but volumes MUST survive restarts. Image backed volumes are removed to allow for updating to newer images if updated,
+	// and on restart, repopulating the volume from the already downloaded image is non-destructive
```
DakCrowder, line 121: If i understand this correctly, this implies we remove the volume but not the image the volume references correct?  So could we have danging images (or do we care)?

kkyrazis: > So could we have danging images (or do we care)?

Yes, and as a whole we do care, but there are features (ideas?, thoughts?, tickets?) to address that particular issue. As it exists, I don't think we could simply safely remove all associated images with an app here, as we'd need to be concerned about the possibility of a rollback and the assumption that all dependencies are already available.

hexfusion: pruning does need to be considered in the future but it is complicated in that the reference counter needs to consider applications embedded, and declared as part of current AND desired spec.

internal/agent/client/podman.go
kkyrazis: simply to have uniform CLI processing similar the other Network and Pods calls. No issues swapping over to json though

internal/agent/device/applications/lifecycle/quadlet.go
DakCrowder: Yeah makes sense definitely not straightforward

internal/agent/client/podman.go
hexfusion: your call just saw the nested for loops and my OCD kicked in :p

kkyrazis: json it is.

hexfusion (approved):

PR #1971, opened 2025-11-18
EDM-2584: Hide Quadlet Applications from Systemd Status

kkyrazis:
Quadlet applications will exclude their services from monitoring via the systemd status reporter. 

## Summary by CodeRabbit

* **New Features**
  * Ability to exclude specific systemd services from status reports at runtime (added/removed exclusions); excluded services are omitted from status/exported results.
  * Systemd status is now registered alongside other exporters for more consistent reporting.

* **Refactor**
  * Centralized creation of the systemd manager and threaded it into application lifecycle components to remove redundant initialization.

* **Tests**
  * Tests updated to use a mock systemd manager and verify exclusion and lifecycle behaviors.

### DIFF
--- a/test/harness/test_harness.go
+++ b/test/harness/test_harness.go
@@ -79,18 +79,22 @@ func WithAgentPprof() TestHarnessOption {
 	}
 }
 
+// WithAgentAudit enables the agent's audit logging when the harness starts the agent.
+// Note: Audit logging is enabled by default, so this option is primarily for test clarity.
+func WithAgentAudit() TestHarnessOption {
+	return func(h *TestHarness) {
+		if h.agentConfig != nil {
+			enabled := true
+			h.agentConfig.AuditLog.Enabled = &enabled
+		}
+	}
+}
+
 // NewTestHarness creates a new test harness and returns a new test harness
 // The test harness can be used from testing code to interact with a
 // set of agent/server/store instances.
 // It provides the necessary elements to perform tests against the agent and server.
-func NewTestHarness(ctx context.Context, testDirPath string, goRoutineErrorHandler func(error)) (*TestHarness, error) {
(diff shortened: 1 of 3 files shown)


internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -23,14 +24,16 @@ var _ ActionHandler = (*Quadlet)(nil)
...
 	log            *log.PrefixLogger
 	actionServices map[string][]string
 }
 
-func NewQuadlet(log *log.PrefixLogger, rw fileio.ReadWriter, systemd *client.Systemd) *Quadlet {
+func NewQuadlet(log *log.PrefixLogger, rw fileio.ReadWriter, systemdClient *client.Systemd, systemdManager systemd.Manager) *Quadlet {
 	return &Quadlet{
-		systemd:        systemd,
+		systemd:        systemdClient,
+		systemdManager: systemdManager,
```
hexfusion, line 35: hmm should manager expose what we need from client this feels a little off.

internal/agent/device/systemd/systemd.go
```diff
@@ -17,19 +18,28 @@ import (
...
 
 type manager struct {
-	patterns []string
-	client   *client.Systemd
-	log      *log.PrefixLogger
+	patterns         []string
+	client           *client.Systemd
+	log              *log.PrefixLogger
+	excludedServices map[string]struct{}
+	mu               sync.RWMutex
```
hexfusion: do we need the manager to be thread safe?

internal/agent/device/applications/lifecycle/quadlet.go
hexfusion: mainly because the manager has a client is why I ask

kkyrazis: Added all of the client methods currently used to the interface.

PR #1972, opened 2025-11-18
EDM-2595: Remove UI subchart and make it a component of a fctl helm chart

rawagner:
After removal of Keycloak subchart, the only one left is UI - and there is no good reason to not have UI as a regular fctl component. It also simplifies our chart.

## Summary by CodeRabbit

* **Chores**
  * Consolidated UI into the main Helm chart, removed the separate UI chart and simplified packaging; chart tarballs are now tracked.
  * Centralized UI configuration under a single UI block and gated UI resources on an enabled flag.

* **Documentation**
  * Expanded values documentation to expose granular UI image, auth, and branding options; removed the separate UI README.

* **Bug Fixes / Behavior**
  * Standardized URL, certificate, and feature-flag resolution for more predictable UI deployment.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -33,6 +33,13 @@ jobs:
             examples/**
             packaging/rpm/**
 
+      - name: Free Disk Space (Ubuntu)
+        if: ${{ steps.changed-files.outputs.any_changed == 'true' }}
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          large-packages: true # disabling this will result in ~4.6 GB not freed, but will save 2-3 minutes of exec time
+
       - name: Check commit message
         if: ${{ steps.changed-files.outputs.any_changed == 'true' }}
         run: hack/check-commits.sh
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -16,7 +16,7 @@ packages:
         message: "/packit build"
 actions:
   get-current-version:
-    - bash -c './hack/current-version --rpm | sed "s/^v//; s/-/~/g"'
+    - bash -c './hack/current-version | sed "s/^v//; s/-/~/g"'
 jobs:
   - job: copr_build
     trigger: pull_request
--- a/hack/current-version
+++ b/hack/current-version
@@ -13,21 +13,4 @@ if [[ -z "$tag" ]]; then
   tag=$(git describe --tags --exclude latest 2>/dev/null || echo -n "v0.0.0-unknown")
 fi
 
-# For RPM packaging (when called from packit), remove git commit hash but preserve main branch info
-# since COPR will add its own commit information, preventing duplication
-if [[ "${PACKIT_BUILD:-}" == "1" ]] || [[ "$1" == "--rpm" ]]; then
-  # For main branch builds, preserve the "-main" part but remove commit details
-  # Examples:
-  #   v1.0.0-main-211-g1234567 -> v1.0.0-main
-  #   v1.0.0-feature-123-g5678abc -> v1.0.0 (non-main branches get stripped completely)
-  #   v1.0.0 -> v1.0.0 (release tags stay as-is)
-  if [[ "$tag" =~ -main ]]; then
-    # Keep base version and "-main" but remove commit details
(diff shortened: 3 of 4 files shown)


deploy/helm/flightctl/templates/ui/flightctl-ui-cm.yaml
```diff
@@ -6,32 +6,27 @@ metadata:
...
-  {{- if .Values.baseURL }}
-  BASE_UI_URL: {{ .Values.baseURL | quote }}
-  {{- else }}
-  BASE_UI_URL: {{ include "flightctl.getUIUrl" . }}
   {{- end }}
+  BASE_UI_URL: {{ include "flightctl.getUIUrl" . }}
   {{- if eq (.Values.global).target "standalone" }}
   {{- if eq ((.Values.global).auth).type "oidc" }}
-  INTERNAL_AUTH_URL: {{ default (((.Values.global).auth).oidc).oidcAuthority .Values.auth.internalAuthUrl | quote }}
+  INTERNAL_AUTH_URL: {{ default (((.Values.global).auth).oidc).issuer .Values.ui.auth.internalAuthUrl | quote }}
```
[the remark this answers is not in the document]
rawagner: syntax is `{{ default <fallback> <value> }}`

meaning we already prefer ui.auth.internalAuthUrl over global setting

deploy/helm/flightctl/templates/ui/flightctl-ui-cm.yaml
```diff
@@ -6,32 +7,28 @@ metadata:
   labels:
     {{- include "flightctl.standardLabels" . | nindent 4 }}
 data:
-  FLIGHTCTL_SERVER: {{ .Values.api.url | quote }}
+  FLIGHTCTL_SERVER: https://flightctl-api:3443/
```
rawagner, line 10: the FCTL server is well-know and not configurable.

deploy/helm/flightctl/Chart.yaml
```diff
@@ -5,7 +5,3 @@ home: https://github.com/flightctl/flightctl
 type: application
 version: 0.1.0
 appVersion: "latest" # this is replaced at packaging with the hash of the image release
```
celdrake, line 7: This stays the same, right?
With a released version the UI images will be pinned to that particular version, otherwise we point to `latest`

deploy/helm/flightctl/README.md
```diff
@@ -353,9 +347,18 @@ For more detailed configuration options, see the [Values](#values) section below
 | prometheus.enabled | bool | `false` | Enable Prometheus deployment |
 | telemetryGateway | object | `{"enabled":false}` | Telemetry Gateway Configuration |
 | telemetryGateway.enabled | bool | `false` | Enable telemetry gateway service |
-| ui | object | `{"api":{"insecureSkipTlsVerify":true},"enabled":true}` | UI Configuration |
+| ui | object | `{"api":{"insecureSkipTlsVerify":true},"auth":{"caCert":"","clientId":"flightctl","insecureSkipTlsVerify":false,"internalAuthUrl":""},"enabled":true,"image":{"image":"quay.io/flightctl/flightctl-ui","pluginImage":"quay.io/flightctl/flightctl-ocp-ui","pullPolicy":"","tag":""},"isRHEM":false}` | UI Configuration |
```
celdrake, line 350: With the new Auth system, we don't need the `internalAuthUrl` or `clientId` as these would be retrieved from `auth/config`.

We can drop this later in my PR which removes those settings. WDYT?

https://github.com/flightctl/flightctl/compare/main...EDM-2373-remove-static-auth-ui-configuration

rawagner: Yes, I agree, lets remove them in your PR. I wanted to minimize changes unrelated to removal of the subchart :+1:

deploy/helm/flightctl/Chart.yaml
rawagner: Yes, correct :) We still pin to a specific version in our release job. And it would still work for UI - as the version for ui images is the same as others.

celdrake (approved):

PR #1974, opened 2025-11-18
EDM-2593: Remove global.target in favor of enableOpenShift/MCExtensions

rawagner:
## Summary by CodeRabbit

* **New Features**
  * New feature flags: enableOpenShiftExtensions and enableMulticlusterExtensions (auto/true/false). Exposure options now include auto/route/gateway (experimental)/none. Added imagePullPolicy and reorganized gateway/ports configuration.

* **Chores**
  * Removed legacy global.target, global.apiUrl, legacy nodePort/gatewayKeys and deleted configuration examples from the README.

* **Behavior Changes**
  * Flags now drive UI, routes, network policies, console plugin, CLI artifact rendering, auth selection and pre-upgrade scaling; exposure defaults auto-detect based on environment.

### DIFF
--- a/deploy/helm/flightctl/README.md
+++ b/deploy/helm/flightctl/README.md
@@ -289,22 +289,16 @@ For more detailed configuration options, see the [Values](#values) section below
 | global.clusterLevelSecretAccess | bool | `false` | Allow flightctl-worker to access secrets at the cluster level for embedding in device configs |
 | global.enableMulticlusterExtensions | string | `"auto"` | Enable MultiCluster Engine extensions - one of 'auto', 'true', 'false'. |
 | global.enableOpenShiftExtensions | string | `"auto"` | Enable OpenShift extensions - one of 'auto', 'true', 'false'. |
-| global.exposeServicesMethod | string | `""` | How the Flight Control services should be exposed. Can be one of 'nodePort', 'route', or 'gateway' |
-| global.gatewayClass | string | `""` | Gateway API class name for gateway exposure method |
-| global.gatewayPorts.http | int | `80` | HTTP port for Gateway API configuration |
-| global.gatewayPorts.tls | int | `443` | TLS port for Gateway API configuration |
+| global.exposeServicesMethod | string | `"auto"` | How the Flight Control services should be exposed. Can be either 'auto', 'route', 'gateway' (experimental) or 'none' |
+| global.gateway | object | `{"gatewayClassName":"","ports":{"http":80,"tls":443}}` | Configuration for 'gateway' service exposure method |
+| global.gateway.gatewayClassName | string | `""` | Gateway API class name for gateway exposure method |
+| global.gateway.ports.http | int | `80` | HTTP port for Gateway API configuration |
+| global.gateway.ports.tls | int | `443` | TLS port for Gateway API configuration |
 | global.generateSecrets | bool | `true` | Generate secrets when deploying Flight Control. This should be set to false if you want to provide your own secrets or when upgrading Flight Control to avoid overriding the existing secrets |
 | global.imagePullPolicy | string | `"IfNotPresent"` | Image pull policy for all containers |
 | global.imagePullSecretName | string | `""` | Name of the image pull secret for accessing private container registries You must create this secret manually before deployment |
 | global.internalNamespace | string | `""` | Namespace where internal components are deployed |
 | global.metrics.enabled | bool | `true` | Enable metrics exporting and service |
-| global.nodePorts.agent | int | `7443` | NodePort for agent communication service |
-| global.nodePorts.alertmanagerProxy | int | `8443` | NodePort for Alertmanager proxy service |
-| global.nodePorts.api | int | `3443` | NodePort for Flight Control API service |
-| global.nodePorts.cliArtifacts | int | `8090` | NodePort for CLI artifacts service |
-| global.nodePorts.telemetryGatewayOtlp | int | `4317` | NodePort for OTLP telemetry gateway |
-| global.nodePorts.telemetryGatewayProm | int | `9464` | NodePort for Prometheus telemetry gateway |
-| global.nodePorts.ui | int | `9000` | NodePort for web UI service |
 | global.organizations.enabled | bool | `false` | Enable IDP-provided organizations support |
 | global.rbac.create | bool | `true` | Create RBAC resources (roles, bindings, service accounts) |
 | global.sshKnownHosts.data | string | `""` | SSH known hosts file content for Git repository host key verification. |
--- a/deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager.yaml
+++ b/deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager.yaml
@@ -103,14 +103,14 @@ metadata:
   labels:
     flightctl.service: flightctl-alertmanager
 spec:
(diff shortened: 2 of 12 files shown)


deploy/helm/flightctl/templates/ui/flightctl-ui-route.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if and (not (eq .Values.global.target "acm")) (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
+{{ if and (eq (include "flightctl.enableMulticlusterExtensions" .) "false") (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
```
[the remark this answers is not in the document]
rawagner: the delimiter is not required

deploy/helm/flightctl/README.md
```diff
@@ -311,7 +280,9 @@ For more detailed configuration options, see the [Values](#values) section below
 | global.baseDomainTls.cert | string | `""` | Certificate for the base domain wildcard certificate, it should be valid for *.${baseDomain}. This certificate is only used for non mTLS endpoints, mTLS endpoints like agent-api, etc will use different certificates. |
 | global.baseDomainTls.key | string | `""` | Key for the base domain wildcard certificate. |
 | global.clusterLevelSecretAccess | bool | `false` | Allow flightctl-worker to access secrets at the cluster level for embedding in device configs |
-| global.exposeServicesMethod | string | `"route"` | How the Flight Control services should be exposed. Can be either nodePort or route |
+| global.enableMulticlusterExtensions | string | `"auto"` | Enable MultiCluster Engine extensions - one of 'auto', 'true', 'false'. |
+| global.enableOpenShiftExtensions | string | `"auto"` | Enable OpenShift extensions - one of 'auto', 'true', 'false'. |
+| global.exposeServicesMethod | string | `""` | How the Flight Control services should be exposed. Can be one of 'nodePort', 'route', or 'gateway' |
```
fzdarsky: `nodePort` is for developers only, right? So I'm wondering whether we should expose it here (user-facing API) rather than a (hidden) developer config?

Do we still need to support Ingress?

rawagner: right, `nodePort` is really just for devs.

Ingress -> do you mean `gateway` option ? That one, i'd say, is more of an experiment - not something we would want to support yet.
Should we then hide `exposeServicesMethod` from values.yaml ?

fzdarsky: Ok, so we do want to make it easy to configure routes correctly for FC, which is why we include Routes to the templates in the first place. But there needs to be at least a way to opt out and let the user to deploy their own Ingress/Route/Gateway instead (or the developer to use `nodePort` instead). So I do think we need an `exposeServiceMethod`. The question is what it defaults to? `""` or `"none"` (i.e. an opt-in) sounds reasonable, as it's probably least surprising to users and requires them to be explicit about wanting to expose their services.

_When_ the user opts-in to let us deploy, it'd be `route` on OpenShift today and `gateway` tomorrow. I'd be ok to keep the latter as "experimental" / not supported. What about vanilla K8s, though? If we keep `gateway` that'd also be covered. Do we also need to add `ingress`? No, we don't want that added complexity and users can chose `""` instead and deploy their own.

rawagner: what about `auto`, `route`, `none` options ? (the `gateway` and `nodePort` are still there, just not documented)

I'd try to avoid adding `ingress` now - so if they want to, they select `none` and configure their own.
On OpenShift, `auto` get translated to `route`
On other k8s, we dont expose anything.

You are mentioning having `none` as default, so that user has to opt-in explicitly - that is however not what we did  so far. We always defaulted to `route`. Im worried that having `none` as default will make getting started experience worse.

fzdarsky: Makes sense. So, yes, let's forget about opt-in and `ingress` and default to `auto`, with the logic for OpenShift you proposed.

What feels strange, though, is that `auto` on K8s means `none`, i.e. a successful deployment ends up without service exposure. Would it make sense to map `auto` to `gateway` instead and emit an error if no Gateway API exists (so users would need to explicitly choose `none`?

rawagner: OK, as we agreed on slack - I've put back the gateway option which, for non-openshift envs & if gateway CRDs are present, `auto` defaults to. If we cannot detect OpenShift, nor Gateway, user has to set `none`, otherwise the chart rendering fails.

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager.yaml
```diff
@@ -103,14 +103,14 @@ metadata:
   labels:
     flightctl.service: flightctl-alertmanager
 spec:
-  {{- if and .Values.global.nodePorts.alertmanager (eq (include "flightctl.getServiceExposeMethod" .) "nodePort") }}
+  {{- if and (eq (include "flightctl.getServiceExposeMethod" .) "nodePort") .Values.global.nodePorts.alertmanager }}
```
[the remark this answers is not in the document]
rawagner: the nodePort option is still available, but not documented - it serves mostly as a dev setting. If nodePort is selected, the ports must be provided.

deploy/helm/flightctl/templates/flightctl-gateway.yaml
```diff
@@ -7,41 +7,41 @@ metadata:
   labels:
     {{- include "flightctl.standardLabels" . | nindent 4 }}
 spec:
-  gatewayClassName: {{ .Values.global.gatewayClass }}
+  gatewayClassName: {{ .Values.global.gateway.gatewayClassName }}
```
celdrake, line 10: Nit, but could this be just 
```suggestion
  gatewayClassName: {{ .Values.global.gateway.className }} 
```

? since it's already in the scope of "gateway" and it would align with the other fields.

deploy/helm/flightctl/templates/ui/flightctl-ui-cm.yaml
```diff
@@ -17,7 +17,7 @@ data:
   {{- end }}
   FLIGHTCTL_SERVER_INSECURE_SKIP_VERIFY: {{ .Values.ui.api.insecureSkipTlsVerify | quote }}
   IS_RHEM: {{ .Values.ui.isRHEM | quote }}
-  ORGANIZATIONS_ENABLED: {{ if and (default false ((.Values.global).organizations).enabled) (not (eq (.Values.global).target "acm")) }}"true"{{ else }}"false"{{ end }}
+  ORGANIZATIONS_ENABLED: {{ if and (default false ((.Values.global).organizations).enabled) (eq (include "flightctl.enableMulticlusterExtensions" .) "false" ) }}"true"{{ else }}"false"{{ end }}
```
celdrake, line 20: I think we want to support Organizations for ACM, but we could do it separately.

deploy/helm/flightctl/templates/flightctl-gateway.yaml
rawagner: I wanted to align here with the gateway API naming

deploy/helm/flightctl/templates/ui/flightctl-ui-cm.yaml
rawagner: Ok, I did not want to change any logic here, just simply replace previous checks with the new one. We should keep an eye on it thought :+1:

celdrake: yes, there's a ticket for that (on me :sweat_smile: )

PR #1980, opened 2025-11-18
EDM-2092: Cleanup resources left after tests

eldar101:
Cleanup for ERS without labels as advised by @asafbennatan .
Currently running e2e tests leaves out only ERS.

## Summary by CodeRabbit

* **Tests**
  * Added a post-test cleanup step that re-runs cluster resource cleanup after main cleanup, logs success or non-fatal failures, and does not alter the test exit outcome.

* **Chores**
  * Minor formatting and whitespace adjustments in test cleanup scripts.

### DIFF
--- a/test/scripts/run_e2e_tests.sh
+++ b/test/scripts/run_e2e_tests.sh
@@ -148,5 +148,13 @@ if ! test/scripts/e2e_cleanup.sh; then
 fi
 echo "✅ [Test Execution] Cleanup completed"
 
+# Step 3b: Re-run the same cluster resource cleanup as startup
+echo "🔄 [Test Execution] Step 3b: Running post-test resource cleanup (same logic as e2e_startup.sh)..."
+if ! test/scripts/e2e_startup.sh; then
+    echo "⚠️  [Test Execution] Post-test resource cleanup failed, but continuing to report test result..."
+else
+    echo "✅ [Test Execution] Post-test resource cleanup completed successfully"
+fi
+
 # Exit with the test exit code
 exit $TEST_EXIT_CODE
--- a/test/scripts/e2e_cleanup.sh
+++ b/test/scripts/e2e_cleanup.sh
@@ -27,19 +27,19 @@ echo "🔍 [Cleanup] Found ${#flightctl_vms[@]} flightctl e2e VMs: ${flightctl_v
 # Clean up each VM
 for vm_name in "${flightctl_vms[@]}"; do
     echo "🔄 [Cleanup] Cleaning up VM: $vm_name"
-    
+
     # 1. Delete pristine snapshot (ignore errors)
     echo "🔄 [Cleanup] Deleting pristine snapshot for $vm_name"
     if ! virsh snapshot-delete "$vm_name" "pristine" --metadata 2>/dev/null; then
         echo "⚠️  [Cleanup] Failed to delete pristine snapshot for $vm_name (may not exist)"
     fi
-    
+
     # 2. Destroy the VM if it's running (ignore errors)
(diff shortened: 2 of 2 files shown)


test/scripts/run_e2e_tests.sh
```diff
@@ -2,6 +2,9 @@
 set -x -euo pipefail
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 source "${SCRIPT_DIR}"/functions
+# Resolve repo root and flightctl binary
```
siserafin: I would include the ER cleanup in test/scripts/e2e_cleanup.sh,which is called by run_e2e_test.sh in step 3, to keep this script clean

test/scripts/run_e2e_tests.sh
```diff
@@ -148,5 +151,48 @@ if ! test/scripts/e2e_cleanup.sh; then
...
+            if output="$(${FLIGHTCTL_BIN} delete er "${ername}" 2>&1)"; then
+                echo "Deleted EnrollmentRequest '${ername}'"
+            else
+                # If a race occurred and the server reports device exists, skip; otherwise surface the error.
+                if echo "${output}" | grep -qi "device exists"; then
+                    echo "Skipping EnrollmentRequest '${ername}': server reports device exists (race)"
+                    continue
+                fi
+                echo "Error deleting EnrollmentRequest '${ername}': ${output}"
+                # If you prefer cleanup to continue in presence of errors, replace `exit 1` with `continue`.
```
siserafin: maybe we can remove this comment

test/scripts/run_e2e_tests.sh
```diff
@@ -148,5 +151,48 @@ if ! test/scripts/e2e_cleanup.sh; then
...
+    if [ -n "${ERS}" ]; then
+        echo "Found EnrollmentRequests to evaluate:"
+        echo "${ERS}"
+
+        # Process each ER name one-by-one
+        echo "${ERS}" | sed 's!.*/!!' | while IFS= read -r ername; do
+            # Skip empty lines
+            [ -z "${ername}" ] && continue
+
+            # Check if a Device with the same name exists. If it does, skip ER deletion.
```
siserafin: If this is the final cleanup, should we also delete any devices and their associated ERs that might have been left over from tests?

test/scripts/run_e2e_tests.sh
eldar101: ok, moving now.

test/scripts/run_e2e_tests.sh
eldar101: everything gets deleted already besides some leftover pending ERs

asafbennatan (commented):
@eldar101 we are cleaning the resources on e2e_startup.sh

if we require a cleanup after we run the tests we should use the exact logic we have there

eldar101:
> @eldar101 we are cleaning the resources on e2e_startup.sh
> 
> if we require a cleanup after we run the tests we should use the exact logic we have there

I thought ERs were a different case since they can't be tied to test labels and also can't be deleted if device exists.
The leftovers after each test run is a bunch of Pending ERs.

asafbennatan:
@eldar101 i dont see a contradiction  , we are clearing all ers before we start the e2e, if having ers after the e2e test is problematic for some reason you can run the same logic at the end

siserafin:
> @eldar101 i dont see a contradiction , we are clearing all ers before we start the e2e, if having ers after the e2e test is problematic for some reason you can run the same logic at the end

I think this comment is correct, the code in e2e_startup.sh is exactly what we need. We could just re-run it as final step in run_e2e_test.sh. Maybe we can rename it as e2e_resources_cleanup.sh.

PR #1981, opened 2025-11-18
EDM-2604: Fix PAM issuer having the wrong redirect URI

celdrake:
The redirect URI for the flighctl UI is <ipURL>:PORT/callback
The PAM Issuer configuration had a mismatch, and used <ipURL>:PORT/auth/callback instead

When attempting to log in through the PAM issuer in the quadlets deployment, the API responded with the following error to the authorize request:
```
  
  {
    "apiVersion": "v1alpha1",
    "code": 400,
    "kind": "Status",
    "message": "invalid_request",
    "reason": "Error",
    "status": "Failure"
}
```

This PR attempts to fix the redirect URL for the UI. I'm not sure if additional redirect URIs are needed in cases where the UI may be disabled, so please check.

Finally, I made changes so that when "authorize" fails, it gives an indication of what caused the request being invalid. Currently it didn't log anything so I had to guess.

After the changes (using custom image for the UI deployment `quay.io/rh_ee_camadorg/fc-ui:auth-permissions-nov-18-1"`):
pam-issuer-with-quadlets

## Release Notes

* **Bug Fixes**
  * Enhanced error handling with context-aware validation feedback for authorization failures, including missing fields, redirect URI mismatches, PKCE validation issues, and state mismatches.
  * Improved security by masking sensitive data in error messages and logs.

* **Chores**
  * Updated default PAM OIDC redirect URI configuration.

### DIFF
--- a/internal/auth/issuer/pam/pam_issuer.go
+++ b/internal/auth/issuer/pam/pam_issuer.go
@@ -393,25 +393,46 @@ func (s *PAMOIDCProvider) handleAuthorizationCodeGrant(ctx context.Context, req 
 	return tokenResponse, nil
 }
 
+// maskSensitiveValue truncates a sensitive value for logging/error messages
+// Shows first 8 characters and last 4 characters, with "..." in between
+func maskSensitiveValue(value string) string {
+	if value == "" {
+		return "<empty>"
+	}
+	if len(value) <= 12 {
+		return "***" // Don't reveal short values at all
+	}
+	return value[:8] + "..." + value[len(value)-4:]
+}
+
 // Authorize handles the authorization endpoint for authorization code flow
 func (s *PAMOIDCProvider) Authorize(ctx context.Context, req *pamapi.AuthAuthorizeParams) (*AuthorizeResponse, error) {
+	stateValue := lo.FromPtrOr(req.State, "")
+	maskedState := maskSensitiveValue(stateValue)
 	s.log.Debugf("Authorize: starting authorization request - ClientId=%s, RedirectUri=%s, ResponseType=%s, Scope=%s, State=%s",
-		req.ClientId, req.RedirectUri, req.ResponseType, lo.FromPtrOr(req.Scope, ""), lo.FromPtrOr(req.State, ""))
+		req.ClientId, req.RedirectUri, req.ResponseType, lo.FromPtrOr(req.Scope, ""), maskedState)
 
 	// Validate required fields
-	if req.ClientId == "" || req.RedirectUri == "" {
-		s.log.Warnf("Authorize: missing required fields - ClientId=%s, RedirectUri=%s", req.ClientId, req.RedirectUri)
-		return nil, errors.New(ErrorInvalidRequest)
+	var missingFields []string
+	if req.ClientId == "" {
+		missingFields = append(missingFields, "client_id")
+	}
+	if req.RedirectUri == "" {
+		missingFields = append(missingFields, "redirect_uri")
+	}
+	if len(missingFields) > 0 {
+		s.log.Warnf("Authorize: missing required fields - %v", missingFields)
+		return nil, fmt.Errorf("%s: missing required parameter(s): %s", ErrorInvalidRequest, strings.Join(missingFields, ", "))
 	}
 
 	// Validate client ID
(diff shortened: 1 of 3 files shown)


internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -393,25 +393,46 @@ func (s *PAMOIDCProvider) handleAuthorizationCodeGrant(ctx context.Context, req
...
+	if len(missingFields) > 0 {
+		s.log.Warnf("Authorize: missing required fields - %v", missingFields)
+		return nil, fmt.Errorf("%s: missing required parameter(s): %s", ErrorInvalidRequest, strings.Join(missingFields, ", "))
 	}
 
 	// Validate client ID
 	if s.config == nil {
 		s.log.Errorf("Authorize: config is nil")
-		return nil, errors.New(ErrorInvalidClient)
+		return nil, fmt.Errorf("%s: PAM issuer configuration not initialized", ErrorInvalidClient)
```
hexfusion, line 431: ErrorInvalidClient I would expect to be an error type. I know you didnt add that but it feels awkward. 

```go 
const ErrInvalidClient =  errors.New("invalid client")

return fmt.Errorf("%w: PAM issuer configuration not initialized", ErrInvalidClient)
```

this allows golang level unwrapping in tests etc.
```go
if errors.Is(err, ErrInvalidClient) { // hangle error }

File: internal/auth/issuer/pam/pam_issuer.go [hunk 2]
```diff
@@ -428,14 +449,25 @@ func (s *PAMOIDCProvider) Authorize(ctx context.Context, req *pamapi.AuthAuthori
...
+		var allowedURIsForLog string
+		if s.config != nil && len(s.config.RedirectURIs) > 0 {
+			// For logs: show full list (server-side only)
+			allowedURIsForLog = fmt.Sprintf(" Allowed redirect URIs: %v", s.config.RedirectURIs)
+			// For client error: show count only to avoid revealing internal configuration
+			allowedURIs = fmt.Sprintf(" %d redirect URI(s) configured", len(s.config.RedirectURIs))
+		} else {
+			allowedURIs = " No redirect URIs configured"
+			allowedURIsForLog = allowedURIs
+		}
```
hexfusion, line 462: could we simplify this a bit with a single var, whenever I am in a function or method need to declare 2 vars I take a couple steps back :)

```go
var allowed string

if s.config != nil && len(s.config.RedirectURIs) > 0 {
	allowed = fmt.Sprintf("%d redirect URI(s) configured", len(s.config.RedirectURIs))
} else {
	allowed = "no redirect URIs configured"
}
...

File: internal/auth/issuer/pam/pam_issuer.go [hunk 3]
```diff
@@ -504,15 +536,19 @@ func (s *PAMOIDCProvider) Authorize(ctx context.Context, req *pamapi.AuthAuthori
...
-		return nil, errors.New(ErrorInvalidRequest)
+		return nil, fmt.Errorf("%s: redirect_uri mismatch. Session: %s, request: %s", ErrorInvalidRequest, sessionData.RedirectURI, req.RedirectUri)
 	}
 	if sessionData.State != lo.FromPtrOr(req.State, "") {
-		s.log.Warnf("Authorize: state mismatch - session=%s, request=%s", sessionData.State, lo.FromPtrOr(req.State, ""))
-		return nil, errors.New(ErrorInvalidRequest)
+		requestState := lo.FromPtrOr(req.State, "")
+		maskedSessionState := maskSensitiveValue(sessionData.State)
+		maskedRequestState := maskSensitiveValue(requestState)
+		s.log.Warnf("Authorize: state mismatch - session=%s, request=%s", maskedSessionState, maskedRequestState)
```
hexfusion, line 549: no need to allocate since your function is nicely self documenting
```go
s.log.Warnf(
	"Authorize: state mismatch - session=%s, request=%s",
	maskSensitiveValue(sessionData.State),
	maskSensitiveValue(requestState),
)
```

celdrake:
Discussed with @asafbennatan as I was not aware of the spec related to Errors. He'll fix it in the correct way.
Thanks for the reviews, also @hexfusion !

PR #1982, opened 2025-11-18
EDM-2511: Disallow editing ERs in CLI

keitwb:
Also prevent autocompleting resource names if they aren't allowed in the command in the first place

## Summary by CodeRabbit

* **Bug Fixes**
  * Autocomplete now validates resource kinds and only suggests allowed kinds, falling back to available options when needed.
  * Edit command no longer permits certain resource types and now returns a standardized "edit not allowed" error for unsupported edits.

* **Tests**
  * Updated tests to assert the new standardized error behavior.


internal/cli/edit_test.go
```diff
@@ -56,26 +56,26 @@ func TestEditOptions_Validate(t *testing.T) {
 			name:          "invalid - cannot edit events",
 			args:          []string{"event/test-event"},
 			expectError:   true,
-			errorContains: "you cannot edit events",
+			errorContains: "edit not permitted for event",
```
hexfusion: I would LOVE to see this become less brittle. I know this isnt your doing but we should not be matching strings for errors as its very brittle. 

```go
const ErrEditNotAllowed = errros.New("edit not permitted for event")
			
wantErr: ErrEditNotAllowed,

if errors.Is(err, ErrEditNotAllowed) { // handle
```

internal/cli/completion.go
```diff
@@ -174,7 +174,7 @@ func (kna KindNameAutocomplete) ValidArgsFunction(cmd *cobra.Command, args []str
 
 	if len(args) == 0 {
 		kindLike, _, _ := strings.Cut(toComplete, "/")
-		if kind, err := ResourceKindFromString(kindLike); err == nil {
+		if kind, err := ResourceKindFromString(kindLike); err == nil && slices.Contains(kna.AllowedKinds, kind) {
```
hexfusion: optional this would be a nice helper

```suggestion
		if kind, err := ResourceKindFromString(kindLike); err == nil && kna.isKindAllowed(kind) {
```

internal/cli/edit_test.go
keitwb: Ok, take a look now.

PR #1983, opened 2025-11-18
EDM-2507: Add OCI auto detection

kkyrazis:
Adds the ability for the dependency manager to auto detect the underlying OCI type and either pull the image or pull the artifact properly. Consumer's of these artifacts may need to check for the existence of the artifact or image to determine the real type.

## Summary by CodeRabbit

* **New Features**
  * Automatic detection and classification of OCI image types (Image vs Artifact) during prefetch, improving handling of multi-platform and artifact images.
  * Improved image prefetch reliability by adding manifest inspection to the fetch flow.
  * Enabled Quadlet-type inline applications to be used with image-based providers.

* **Bug Fixes**
  * Removed restrictive validation that blocked Quadlet-type applications with image-based specs.

* **Tests**
  * Added tests for manifest inspection and OCI type detection.

### DIFF
--- a/internal/agent/client/skopeo.go
+++ b/internal/agent/client/skopeo.go
@@ -1,0 +1,93 @@
+package client
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/flightctl/flightctl/internal/agent/device/errors"
+	"github.com/flightctl/flightctl/internal/agent/device/fileio"
+	"github.com/flightctl/flightctl/pkg/executer"
+	"github.com/flightctl/flightctl/pkg/log"
+)
+
+const (
+	skopeoCmd            = "skopeo"
+	defaultSkopeoTimeout = 2 * time.Minute
+)
+
+// OCIManifest represents the minimal OCI manifest structure needed for type detection
+type OCIManifest struct {
+	MediaType    string          `json:"mediaType"`
+	ArtifactType string          `json:"artifactType,omitempty"`
+	Config       *OCIDescriptor  `json:"config,omitempty"`
+	Manifests    json.RawMessage `json:"manifests,omitempty"`
+}
+
+// OCIDescriptor represents a content descriptor in an OCI manifest
+type OCIDescriptor struct {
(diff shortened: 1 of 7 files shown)


internal/agent/client/skopeo.go
```diff
@@ -0,0 +1,94 @@
...
+		exists, err := s.readWriter.PathExists(pullSecretPath)
+		if err != nil {
+			return nil, fmt.Errorf("check pull secret path: %w", err)
+		}
+		if !exists {
+			s.log.Errorf("Pull secret path %s does not exist", pullSecretPath)
+		} else {
+			args = append(args, "--authfile", pullSecretPath)
+		}
+	}
```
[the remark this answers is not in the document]
hexfusion: yeah I think lets terminate execution fast

internal/agent/device/applications/provider/provider.go
```diff
@@ -87,7 +87,7 @@ func CollectBaseOCITargets(
 			// Add base image with PullIfNotPresent policy
 			policy := v1alpha1.PullIfNotPresent
 			targets = append(targets, dependency.OCIPullTarget{
-				Type:       dependency.OCITypeImage,
+				Type:       dependency.OCITypeAuto,
```
hexfusion, line 90: :saluting_face:

internal/agent/device/dependency/dependency.go
```diff
@@ -23,14 +23,45 @@ const (
...
+
+	if manifest.ArtifactType != "" {
+		return OCITypeArtifact, nil
+	}
+
+	if manifest.Config != nil {
+		switch manifest.Config.MediaType {
+		case ociImageConfigMediaType:
+			return OCITypeImage, nil
+		default:
```
hexfusion, line 59: should default be an error case if its not set or unexpected or is that assumed artifact.

hexfusion (commented):
couple small things then lgtm

internal/agent/device/dependency/dependency.go
kkyrazis: If it's empty it should be an error

PR #1985, opened 2025-11-19
EDM-2604: fixed auto set redirect uri , fixed pam issuer errors (per spec)

asafbennatan:
spec ref:

https://datatracker.ietf.org/doc/html/rfc6749#section-5.2
https://openid.net/specs/openid-connect-core-1_0.html#AuthError

## Summary by CodeRabbit

* **New Features**
  * OAuth2 error responses now follow RFC 6749 standards with structured error codes and descriptions.
  * Added UserInfo endpoint support to retrieve authenticated user information.
  * Token and user info responses now enforce required fields for enhanced API compliance.

* **Refactor**
  * Updated redirect callback path from "/auth/callback" to "/callback".


internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -284,36 +314,54 @@ func (s *PAMOIDCProvider) handleAuthorizationCodeGrant(ctx context.Context, req
...
+			Code:             pamapi.InvalidRequest,
+			ErrorDescription: lo.ToPtr("Missing required parameter: redirect_uri"),
+		}
 	}
 	if codeData.RedirectURI != *req.RedirectUri {
 		s.log.Warnf("handleAuthorizationCodeGrant: redirect_uri mismatch - expected=%s, got=%s", codeData.RedirectURI, *req.RedirectUri)
-		return &pamapi.TokenResponse{Error: lo.ToPtr(ErrorInvalidGrant)}, nil
+		return nil, &pamapi.OAuth2Error{
+			Code:             pamapi.InvalidGrant,
+			ErrorDescription: lo.ToPtr("redirect_uri does not match the one used in the authorization request"),
```
celdrake: IIUC, this error throws when the redirect URI used by the client does not match the one the PAM issuer was configured with.

Maybe the message could be clearer for that case? This is the error I had.

asafbennatan: fixed

PR #1988, opened 2025-11-19
EDM-2368: CLI support for CRUD Auth providers

noga-magen:
## Summary by CodeRabbit

* **New Features**
  * Added support for managing Authentication Provider resources via the CLI
  * Users can now create, retrieve, delete, and list Authentication Providers using standard resource commands
  * Authentication Providers display in table format with NAME, TYPE, ISSUER, and CLIENT ID columns
  * Short alias "ap" available for Authentication Provider references

### DIFF
--- a/internal/cli/display/table.go
+++ b/internal/cli/display/table.go
@@ -74,6 +74,8 @@ func (f *TableFormatter) formatList(w *tabwriter.Writer, data interface{}, optio
 		return f.printCSRTable(w, data.(*apiclient.ListCertificateSigningRequestsResponse).JSON200.Items...)
 	case strings.EqualFold(options.Kind, api.EventKind):
 		return f.printEventsTable(w, data.(*apiclient.ListEventsResponse).JSON200.Items...)
+	case strings.EqualFold(options.Kind, api.AuthProviderKind):
+		return f.printAuthProvidersTable(w, data.(*apiclient.ListAuthProvidersResponse).JSON200.Items...)
 	default:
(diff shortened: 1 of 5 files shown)


internal/cli/display/table.go
```diff
@@ -395,3 +399,42 @@ func (f *TableFormatter) printEventsTable(w *tabwriter.Writer, events ...api.Eve
...
+		}
+
+		providerType := ""
+		issuer := ""
+		clientId := ""
+
+		// Extract type from the discriminator
+		if discriminator, err := ap.Spec.Discriminator(); err == nil {
+			providerType = discriminator
+		}
```
hexfusion, line 418: don't we want to know the error for debug? why return an error type if we never return an error?

```go
discriminator, err := ap.Spec.Discriminator(); 
if err != nil {
     retrun err
}

providerType := discriminator
```

internal/cli/display/table.go
```diff
@@ -395,3 +399,42 @@ func (f *TableFormatter) printEventsTable(w *tabwriter.Writer, events ...api.Eve
...
+
+		// Extract issuer and clientId based on type
+		switch providerType {
+		case "oidc":
+			if oidcSpec, err := ap.Spec.AsOIDCProviderSpec(); err == nil {
```
hexfusion, line 423: same here if there is an error we should fail silently?

asafbennatan:
comments will be fixed in
https://github.com/flightctl/flightctl/pull/1990/files

PR #1994, opened 2025-11-19
EDM-2617: More agent selinux systemd permissions

kkyrazis:
For more complex Quadlet applications we appear to need more permissions to properly stop services.

## Summary by CodeRabbit

* **Chores**
  * Expanded system service permissions to allow the agent to start and stop services in addition to checking their status.

### DIFF
--- a/internal/auth/authn/openshift_auth.go
+++ b/internal/auth/authn/openshift_auth.go
@@ -74,6 +74,11 @@ func hashToken(token string) string {
 	return hex.EncodeToString(hash[:])
 }
 
+// GetOpenShiftSpec returns the internal OpenShift spec with client secret intact (for internal use only)
+func (o *OpenShiftAuth) GetOpenShiftSpec() api.OpenShiftProviderSpec {
+	return o.spec
+}
+
 // GetAuthToken extracts the Bearer token from the HTTP request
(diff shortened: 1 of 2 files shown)


packaging/selinux/flightctl_agent.te
```diff
@@ -189,7 +189,7 @@ systemd_status_all_unit_files(flightctl_agent_t)
 systemd_reload_all_services(flightctl_agent_t)
 systemd_start_all_services(flightctl_agent_t)
 systemd_stop_systemd_services(flightctl_agent_t)
-allow flightctl_agent_t init_t:system status;
+allow flightctl_agent_t init_t:system { status stop start };
```
hexfusion, line 192: hmm do we not need the ones above then these I guess are not redundant

keitwb: @kkyrazis What command is being run that needs this? I'm trying to understand the difference between this and the systemd ones above.

keitwb: Ok, maybe it is something specific to how quadlets runs things.

kkyrazis: They are technically different domains One is `systemd_unit_file_t:service` vs `init_t:system`. I'm still not sure why sometimes it gets elevated to `init_t` and sometimes it doesn't.

PR #1999, opened 2025-11-19
EDM-2544: fix login pam issuer

noga-magen:
## Release Notes

* **New Features**
  * Added PKCE support to OAuth2 authentication flow
  * Introduced configuration option to allow public clients without PKCE for CLI web login
  * Enabled IdP-provided organizations feature

* **Improvements**
  * Enhanced localhost redirect callback validation for local development scenarios

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -271,19 +271,19 @@ func (o *LoginOptions) Validate(args []string) error {
 	return nil
 }
 
-// ensureClientID sets a default ClientId when not provided based on auth type and flags
-func (o *LoginOptions) ensureClientID() {
+// ensureClientID extracts the ClientId from the provider config based on auth type
+func (o *LoginOptions) ensureClientID() error {
 	if o.ClientId != "" {
-		return
+		return nil
 	}
 	provider := o.getDefaultProvider()
 	if provider == nil {
-		return
+		return fmt.Errorf("no authentication provider found")
 	}
 	// Get the provider type from the spec
 	providerType, err := provider.Spec.Discriminator()
 	if err != nil {
-		return
+		return fmt.Errorf("failed to determine provider type: %w", err)
 	}
 	switch providerType {
 	case string(v1alpha1.K8s):
@@ -292,9 +292,25 @@ func (o *LoginOptions) ensureClientID() {
 		} else {
 			o.ClientId = "openshift-cli-client"
 		}
+		return nil
 	case string(v1alpha1.Oidc):
-		o.ClientId = "flightctl"
+		if oidcSpec, err := provider.Spec.AsOIDCProviderSpec(); err == nil {
+			if oidcSpec.ClientId != "" {
+				o.ClientId = oidcSpec.ClientId
+				return nil
+			}
+		}
+		return fmt.Errorf("client ID is required for OIDC authentication but was not provided by the API")
+	case string(v1alpha1.Oauth2):
+		if oauth2Spec, err := provider.Spec.AsOAuth2ProviderSpec(); err == nil {
+			if oauth2Spec.ClientId != "" {
+				o.ClientId = oauth2Spec.ClientId
+				return nil
+			}
+		}
+		return fmt.Errorf("client ID is required for OAuth2 authentication but was not provided by the API")
 	}
+	return nil
 }
 
 // getDefaultProvider returns the default authentication provider
@@ -330,21 +346,37 @@ func (o *LoginOptions) createAuthProvider() error {
 	}
 
 	authURL := o.extractAuthURL(provider)
+	scopes := o.extractScopes(provider)
+	providerName := ""
+	if provider.Metadata.Name != nil {
+		providerName = *provider.Metadata.Name
+	}
 
 	orgEnabled := false
 	if o.authConfig.OrganizationsEnabled != nil {
 		orgEnabled = *o.authConfig.OrganizationsEnabled
 	}
 
+	configMap := map[string]string{
+		client.AuthUrlKey:      authURL,
+		client.AuthCAFileKey:   o.AuthCAFile,
+		client.AuthClientIdKey: o.ClientId,
+	}
+	if scopes != "" {
+		configMap[client.AuthScopesKey] = scopes
+	}
+	if o.clientConfig.Service.Server != "" {
+		configMap[client.AuthServerUrlKey] = o.clientConfig.Service.Server
+	}
+	if providerName != "" {
+		configMap[client.AuthProviderNameKey] = providerName
+	}
+
 	authProvider, err := client.CreateAuthProvider(client.AuthInfo{
 		AuthProvider: &client.AuthProviderConfig{
-			Name: *provider.Metadata.Name,
-			Type: providerType,
-			Config: map[string]string{
-				client.AuthUrlKey:      authURL,
-				client.AuthCAFileKey:   o.AuthCAFile,
-				client.AuthClientIdKey: o.ClientId,
-			},
+			Name:   *provider.Metadata.Name,
+			Type:   providerType,
+			Config: configMap,
 		},
 		OrganizationsEnabled: orgEnabled,
 	}, o.InsecureSkipVerify)
@@ -377,6 +409,33 @@ func (o *LoginOptions) extractAuthURL(provider *v1alpha1.AuthProvider) string {
 	case string(v1alpha1.Oauth2):
 		if oauth2Spec, err := provider.Spec.AsOAuth2ProviderSpec(); err == nil {
 			return oauth2Spec.AuthorizationUrl
+		}
+	}
+	return ""
+}
+
+// extractScopes extracts scopes from an AuthProvider as a space-delimited string
+// Returns empty string if scopes are not configured (provider will use default)
+func (o *LoginOptions) extractScopes(provider *v1alpha1.AuthProvider) string {
+	providerType, _ := provider.Spec.Discriminator()
+	switch providerType {
+	case string(v1alpha1.Oidc):
+		if oidcSpec, err := provider.Spec.AsOIDCProviderSpec(); err == nil {
+			if oidcSpec.Scopes != nil && len(*oidcSpec.Scopes) > 0 {
+				return strings.Join(*oidcSpec.Scopes, " ")
+			}
+		}
+	case string(v1alpha1.Oauth2):
+		if oauth2Spec, err := provider.Spec.AsOAuth2ProviderSpec(); err == nil {
+			if oauth2Spec.Scopes != nil && len(*oauth2Spec.Scopes) > 0 {
+				return strings.Join(*oauth2Spec.Scopes, " ")
+			}
+		}
+	case "openshift":
+		if openshiftSpec, err := provider.Spec.AsOpenShiftProviderSpec(); err == nil {
+			if openshiftSpec.Scopes != nil && len(*openshiftSpec.Scopes) > 0 {
+				return strings.Join(*openshiftSpec.Scopes, " ")
+			}
 		}
 	}
 	return ""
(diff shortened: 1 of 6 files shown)


api/v1alpha1/pam-issuer/openapi.yaml
```diff
@@ -162,6 +162,13 @@ paths:
                   type: string
                 state:
                   type: string
```
asafbennatan, line 164: @noga-magen why are we changing this ?

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/init.sh
```diff
@@ -37,10 +37,12 @@ PAM_OIDC_CLIENT_SECRET=$(extract_value "global.auth.pamOidcIssuer.clientSecret"
 PAM_OIDC_SCOPES=$(extract_value "global.auth.pamOidcIssuer.scopes" "$SERVICE_CONFIG_FILE")
 PAM_OIDC_REDIRECT_URIS=$(extract_value "global.auth.pamOidcIssuer.redirectUris" "$SERVICE_CONFIG_FILE")
 PAM_OIDC_SERVICE=$(extract_value "global.auth.pamOidcIssuer.pamService" "$SERVICE_CONFIG_FILE")
+PAM_OIDC_ALLOW_PUBLIC_WITHOUT_PKCE=$(grep -A 20 "pamOidcIssuer:" "$SERVICE_CONFIG_FILE" | grep "allowPublicClientWithoutPKCE:" | head -1 | sed 's/.*allowPublicClientWithoutPKCE:[[:space:]]*\(.*\)/\1/' | sed 's/[[:space:]]*$//')
```
asafbennatan: use extract value

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/init.sh
```diff
@@ -37,10 +37,12 @@ PAM_OIDC_CLIENT_SECRET=$(extract_value "global.auth.pamOidcIssuer.clientSecret"
...
 
 # Set defaults
 PAM_OIDC_CLIENT_ID=${PAM_OIDC_CLIENT_ID:-flightctl-client}
 PAM_OIDC_SERVICE=${PAM_OIDC_SERVICE:-flightctl}
+PAM_OIDC_ALLOW_PUBLIC_WITHOUT_PKCE=${PAM_OIDC_ALLOW_PUBLIC_WITHOUT_PKCE:-true}
```
asafbennatan: this should not be defaulted to true

deploy/podman/service-config.yaml
```diff
@@ -32,9 +32,10 @@ global:
       scopes: openid,profile,email,roles
       redirectUris:
       pamService: flightctl
+      allowPublicClientWithoutPKCE: true # Required for CLI web login without PKCE
```
asafbennatan: should be false

internal/auth/issuer/pam/pam_issuer.go
```diff
@@ -477,13 +477,29 @@ func (s *PAMOIDCProvider) Authorize(ctx context.Context, req *pamapi.AuthAuthori
...
 		s.log.Debugf("Authorize: checking redirect URI - requested=%s, allowed=%v", req.RedirectUri, s.config.RedirectURIs)
+		// Check exact matches first
 		for _, uri := range s.config.RedirectURIs {
 			if uri == req.RedirectUri {
 				validRedirect = true
 				s.log.Debugf("Authorize: redirect URI validation passed - %s", req.RedirectUri)
 				break
 			}
 		}
+		// If no exact match, allow localhost/127.0.0.1 callbacks for CLI usage (secure for local-only access)
```
asafbennatan: we shouldnt have this change 
just configure the redirect uri you need in pam issuers config

internal/cli/login/oauth.go
```diff
@@ -14,15 +14,23 @@ import (
...
 
-func getOAuth2AccessToken(client *osincli.Client, authorizeRequest *osincli.AuthorizeRequest, r *http.Request) (AuthInfo, error) {
+func getOAuth2AccessToken(client *osincli.Client, clientId string, authorizeRequest *osincli.AuthorizeRequest, r *http.Request) (AuthInfo, error) {
 	areqdata, err := authorizeRequest.HandleRequest(r)
 	if err != nil {
 		return AuthInfo{}, err
 	}
 
 	treq := client.NewAccessRequest(osincli.AUTHORIZATION_CODE, areqdata)
+
```
asafbennatan: why do we need this ? clientId is part of the common parameters for oauth2 request

internal/cli/login/oauth.go
```diff
@@ -109,13 +117,19 @@ func oauth2RefreshTokenFlow(refreshToken string, getClient GetClientFunc) (AuthI
 
 	// callback url is not used in refresh token flow, but has to be set, otherwise osincli fails
 	// the value should not matter
-	client, err := getClient("http://127.0.0.1")
+	client, clientId, err := getClient("http://127.0.0.1")
```
asafbennatan: prefer using localhost and probably something like :
http://localhost:8080/callback (or some other port)

internal/cli/login.go
```diff
@@ -285,15 +288,32 @@ func (o *LoginOptions) ensureClientID() {
...
+	// Extract client ID from provider config based on type
 	switch providerType {
+	case string(v1alpha1.Oidc):
+		// For OIDC, use the client ID from the provider config if available
+		if oidcSpec, err := provider.Spec.AsOIDCProviderSpec(); err == nil && oidcSpec.ClientId != "" {
+			o.ClientId = oidcSpec.ClientId
+			return
+		}
+		// Fallback to default if not in provider config
+		o.ClientId = "flightctl"
```
asafbennatan: dont default - use what you get from the api

internal/cli/login/aap.go
```diff
@@ -47,7 +47,7 @@ func (o AAPOAuth) getOAuth2Config() OauthServerResponse {
 	}
 }
 
-func (o AAPOAuth) getOAuth2Client(callback string) (*osincli.Client, error) {
+func (o AAPOAuth) getOAuth2Client(callback string) (*osincli.Client, string, error) {
 	oauthServerResponse := o.getOAuth2Config()
 
 	config := &osincli.ClientConfig{
```
asafbennatan, line 78: we should add the PKCE parameters to the client config - it supports PKCE

asafbennatan (changes requested):
there shouldnt be any changes to flightctl-pam-issuer as far as i know , 
we should oauth using PKCE , see my comments

internal/cli/login/aap.go
```diff
@@ -58,16 +62,19 @@ func (o AAPOAuth) getOAuth2Client(callback string) (*osincli.Client, error) {
 		SendClientSecretInParams: true,
 		RedirectUrl:              callback,
 		Scope:                    "read",
```
asafbennatan: we should take the scopes as a space delimited string from the output of /auth/config ( see providers[x].spec.scopes for oidc/oauth2/openshift

internal/cli/login/aap.go
asafbennatan, line 38: seems to me that the issue is that we need to change how it was originally designed .

server now reports proper structure for all auth types

`client.config.AuthProviderConfig` should change to hold `v1alpha.AuthProvider`
then if we need additional details that are not there like `CAFile/InsecureSkipVerify` add these as properties of `client.config.AuthProviderConfig`

this kind of change would allow us to simply provide `client.config.AuthProviderConfig `(that has `client.config.AuthProviderConfig.authProvider` of type AAP/OAuth2/OIDC/Openshift/K8) to this constructors , making sure that we have all the info we need ( for example the current way prevents you from getting the scopes  , the proper auth url /token url etc (we are currently constructing it instead of taking what we get from the server) 




```suggestion
func NewAAPOAuth2Config(config *client.config.AuthProviderConfig, providerConfig *v1alpha.AAPProviderSpec) AAPOAuth {

``` 

AAP provider config is provided directly so you dont have to keep type inferring it each time , so you just save the two as part of `AAPOAuth2Config` struct this allows very easy implementation of the `getOAuth2Config` and the `getOAuth2Client` methods

repeat this pattern for all providers

asafbennatan, line 81: we should override this with our token proxy endpoint (it will add the client secret to the request if necessery and send it to the issuer)

internal/cli/login/oauth.go
```diff
@@ -42,62 +59,218 @@ func getOAuth2AccessToken(client *osincli.Client, authorizeRequest *osincli.Auth
...
+	callbackURL := "http://localhost:8080/callback"
+	client, _, err := getClient(callbackURL)
 	if err != nil {
-		return ret, err
+		return AuthInfo{}, err
 	}
-	authorizeRequest := client.NewAuthorizeRequest(osincli.CODE)
+
+	authReq := client.NewAuthorizeRequest(osincli.CODE)
+	loginURL := authReq.GetAuthorizeUrl()
```
asafbennatan: should we use authReq.GetAuthorizeUrlWithParams(state) ?

internal/cli/login/oauth.go
```diff
@@ -42,62 +59,218 @@ func getOAuth2AccessToken(client *osincli.Client, authorizeRequest *osincli.Auth
...
-			fmt.Printf("failed to write response %s\n", err.Error())
+			fmt.Fprintf(w, "ERROR: %s", err)
+			done <- err
+			return
 		}
+		_, _ = w.Write([]byte("Login successful. Return to CLI."))
 		done <- nil
 	})
 
+	pkce := loginURL.Query().Get("code_challenge")
```
asafbennatan: why do we need to manually handle the request ? 

according to 

```golang
// Returns the authorize url
func (c *AuthorizeRequest) GetAuthorizeUrl() *url.URL {
	return c.GetAuthorizeUrlWithParams("")
}

// Returns the authorize url
func (c *AuthorizeRequest) GetAuthorizeUrlWithParams(state string) *url.URL {
	u := *c.client.configcache.authorizeUrl
	uq := u.Query()
	uq.Add("response_type", string(c.Type))
	uq.Add("client_id", c.client.config.ClientId)
	uq.Add("redirect_uri", c.client.config.RedirectUrl)
	if c.client.config.Scope != "" {
		uq.Add("scope", c.client.config.Scope)
	}
	if c.client.config.CodeChallenge != "" {
		uq.Add("code_challenge", c.client.config.CodeChallenge)
	}
	if c.client.config.CodeChallengeMethod != "" {
		uq.Add("code_challenge_method", c.client.config.CodeChallengeMethod)
	}
	if state != "" {
		uq.Add("state", state)
	}
	if c.CustomParameters != nil {
		for pn, pv := range c.CustomParameters {
			uq.Add(pn, pv)
		}
	}
	u.RawQuery = uq.Encode()
	return &u
}
```

this is already done for us - we just need to make sure that we are creating `client` correctly
(seems that we are creating it correctly for oidc atleast)

internal/cli/login/oauth.go
```diff
@@ -42,62 +59,218 @@ func getOAuth2AccessToken(client *osincli.Client, authorizeRequest *osincli.Auth
...
+	startURL := loginURL.String()
+
+	if pkce != "" {
```
asafbennatan: i think the following code is unnecessary we do not need a proxy or more code here

noga-magen: The server validates PKCE parameters in the wrong order:

[**Line 517** (`pam_issuer.go`)](https://github.com/flightctl/flightctl/blob/42757c2cec86ea2c10a533e750f8dbfbe2701737/internal/auth/issuer/pam/pam_issuer.go#L517): Server validates PKCE from the **URL parameters first** - if missing for a public client, it immediately returns `"invalid_request"` before checking the session.

**Line 530-545**: Session check happens **after** PKCE validation.

**Line 576-582**: Server uses PKCE from session, but this code is **never reached** if line 517 fails.

**Line 665**: After successful login, the server redirects back to `/api/v1/auth/authorize` **without PKCE parameters in the URL**.



The proxy intercepts the server redirect after login and injects the PKCE parameters into the authorize URL. This allows the request to pass the validation at line 517, after which the server correctly uses PKCE from the session (line 576-582).

asafbennatan: fixed

note that the solution is not to ignore the PKCE in the case the cookie existed but to actually fix the login endpoint to redirect to authorize with the PKCE parameters

PR #2005, opened 2025-11-20
EDM-2260: Graceful Shutdown and Signal Handling

gshilin-sdb:
This PR replaces https://github.com/flightctl/flightctl/pull/1962

A more minimal, linear shutdown model (OS signal → cancel ctx → wait for server to exit → run cleanup exactly once)

## Summary by CodeRabbit

* **Infrastructure**
  * Refactored server startup, shutdown, and error handling across all services with improved graceful shutdown coordination and configurable timeouts.

* **Bug Fixes**
  * Fixed unnecessary logging when cache monitor stops during normal shutdown.
  * Enhanced error visibility for device fleet reconciliation failures through status annotations.

* **Tests**
  * Added comprehensive test coverage for multi-server and single-server shutdown coordination.
  * Improved e2e test reliability using polling-based assertions with retry logic.

### DIFF
--- a/cmd/flightctl-alert-exporter/main.go
+++ b/cmd/flightctl-alert-exporter/main.go
@@ -2,10 +2,7 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"github.com/flightctl/flightctl/internal/alert_exporter"
@@ -20,6 +17,7 @@ import (
 	"github.com/flightctl/flightctl/internal/worker_client"
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/flightctl/flightctl/pkg/queues"
+	"github.com/flightctl/flightctl/pkg/shutdown"
 	"github.com/google/uuid"
 )
 
@@ -41,40 +39,22 @@ func runCmd(cfg *config.Config) error {
 	defer logger.Info("Alert exporter stopped")
 	logger.Infof("Using config: %s", cfg)
 
-	// Single context with signal handling - OS signal cancels context
-	ctx, cancel := signal.NotifyContext(context.Background(),
-		syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
-
-	// Build cleanup functions incrementally as resources are created
-	var cleanupFuncs []func() error
-	defer func() {
-		// First cancel context to signal all goroutines to stop
-		logger.Info("Cancelling context to stop all servers")
-		cancel()
-
-		// Then run cleanup in reverse order after goroutines have stopped
-		logger.Info("Starting cleanup")
-		for i := len(cleanupFuncs) - 1; i >= 0; i-- {
-			if err := cleanupFuncs[i](); err != nil {
-				logger.WithError(err).Error("Cleanup error")
-			}
-		}
-		logger.Info("Cleanup completed")
-	}()
-
 	tracerShutdown := tracing.InitTracer(logger, cfg, "flightctl-alert-exporter")
 	if tracerShutdown != nil {
-		cleanupFuncs = append(cleanupFuncs, func() error {
-			logger.Info("Shutting down tracer")
+		defer func() {
 			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 			defer cancel()
-			return tracerShutdown(ctx)
-		})
+			if err := tracerShutdown(ctx); err != nil {
+				logger.WithError(err).Error("Error shutting down tracer")
+			}
+		}()
 	}
 
-	ctx = context.WithValue(ctx, consts.EventSourceComponentCtxKey, "flightctl-alert-exporter")
-	ctx = context.WithValue(ctx, consts.EventActorCtxKey, "service:flightctl-alert-exporter")
-	ctx = context.WithValue(ctx, consts.InternalRequestCtxKey, true)
+	// Simple context for initialization phase
+	initCtx := context.Background()
+	initCtx = context.WithValue(initCtx, consts.EventSourceComponentCtxKey, "flightctl-alert-exporter")
+	initCtx = context.WithValue(initCtx, consts.EventActorCtxKey, "service:flightctl-alert-exporter")
+	initCtx = context.WithValue(initCtx, consts.InternalRequestCtxKey, true)
 
 	logger.Info("Initializing data store")
 	db, err := store.InitDB(cfg, logger)
@@ -83,77 +63,39 @@ func runCmd(cfg *config.Config) error {
 	}
 
 	store := store.NewStore(db, logger.WithField("pkg", "store"))
-	cleanupFuncs = append(cleanupFuncs, func() error {
-		logger.Info("Closing database connections")
-		return store.Close()
-	})
+	defer store.Close()
 
 	processID := fmt.Sprintf("alert-exporter-%s-%s", util.GetHostname(), uuid.New().String())
-	queuesProvider, err := queues.NewRedisProvider(ctx, logger, processID, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password, queues.DefaultRetryConfig())
+	queuesProvider, err := queues.NewRedisProvider(initCtx, logger, processID, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password, queues.DefaultRetryConfig())
 	if err != nil {
 		return fmt.Errorf("initializing queue provider: %w", err)
 	}
-	cleanupFuncs = append(cleanupFuncs, func() error {
-		logger.Info("Stopping queue provider")
-		queuesProvider.Stop()
-		queuesProvider.Wait()
-		return nil
-	})
+	defer queuesProvider.Wait()
+	defer queuesProvider.Stop()
 
-	kvStore, err := kvstore.NewKVStore(ctx, logger, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password)
+	kvStore, err := kvstore.NewKVStore(initCtx, logger, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password)
 	if err != nil {
 		return fmt.Errorf("initializing kv store: %w", err)
 	}
-	cleanupFuncs = append(cleanupFuncs, func() error {
-		logger.Info("Closing KV store connections")
-		kvStore.Close()
-		return nil
-	})
+	defer kvStore.Close()
 
-	publisher, err := worker_client.QueuePublisher(ctx, queuesProvider)
+	publisher, err := worker_client.QueuePublisher(initCtx, queuesProvider)
 	if err != nil {
 		return fmt.Errorf("initializing task queue publisher: %w", err)
 	}
-	cleanupFuncs = append(cleanupFuncs, func() error {
-		logger.Info("Closing publisher")
-		publisher.Close()
-		return nil
-	})
+	defer publisher.Close()
 
 	workerClient := worker_client.NewWorkerClient(publisher, logger)
 
 	orgCache := cache.NewOrganizationTTL(cache.DefaultTTL)
-	go func() {
-		orgCache.Start(ctx)
-		if ctx.Err() == nil {
-			logger.Warn("Organization cache stopped unexpectedly")
-		}
-	}()
-	cleanupFuncs = append(cleanupFuncs, func() error {
-		logger.Info("Stopping organization cache")
-		orgCache.Stop()
-		return nil
-	})
+	defer orgCache.Stop()
 
 	serviceHandler := service.WrapWithTracing(service.NewServiceHandler(store, workerClient, kvStore, nil, logger, "", "", []string{}))
 	server := alert_exporter.New(cfg, logger)
 
-	// Start server and wait for completion or signal
-	logger.Info("Starting alert exporter server")
-	err = server.Run(ctx, serviceHandler)
-	if err != nil {
-		err = fmt.Errorf("alert exporter server: %w", err)
-	}
-
-	// Handle shutdown reason
-	if errors.Is(err, context.Canceled) {
-		logger.Info("Server stopped due to shutdown signal")
-		return nil // Normal shutdown
-	} else if err != nil {
-		logger.WithError(err).Error("Server stopped with error")
-		return err // Error shutdown
-	}
-
-	logger.Info("Server stopped normally")
-	return nil // Normal completion
+	// Use single server shutdown coordination
+	singleServerConfig := shutdown.NewSingleServerConfig("alert exporter", logger)
+	return singleServerConfig.RunSingleServer(func(shutdownCtx context.Context) error {
+		return server.Run(shutdownCtx, serviceHandler)
+	})
 }
--- a/cmd/flightctl-alertmanager-proxy/main.go
+++ b/cmd/flightctl-alertmanager-proxy/main.go
@@ -13,17 +13,17 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
 	"os"
-	"os/signal"
-	"syscall"
 	"time"
 
 	"github.com/flightctl/flightctl/internal/api_server/middleware"
 	"github.com/flightctl/flightctl/internal/auth"
+	"github.com/flightctl/flightctl/internal/auth/authn"
 	"github.com/flightctl/flightctl/internal/auth/common"
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/crypto"
@@ -32,6 +32,7 @@ import (
 	"github.com/flightctl/flightctl/internal/service"
 	"github.com/flightctl/flightctl/internal/store"
 	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/shutdown"
 	"github.com/go-chi/chi/v5"
 	chimiddleware "github.com/go-chi/chi/v5/middleware"
 	"github.com/sirupsen/logrus"
@@ -149,25 +150,12 @@ func createConditionalAuthMiddleware(
 	}
 }
 
-func main() {
-	ctx := context.Background()
-
-	// Load configuration
-	cfg, err := config.LoadOrGenerate(config.ConfigFile())
-	if err != nil {
-		log.InitLogs().WithError(err).Fatal("Failed to load configuration")
-	}
-
-	// Initialize logging
-	logger := log.InitLogs(cfg.Service.LogLevel)
-	logger.Println("Starting Alertmanager Proxy service")
-	defer logger.Println("Alertmanager Proxy service stopped")
-	logger.Infof("Using config: %s", cfg)
-
+// setupCertificates initializes CA and TLS certificates following same pattern as API server
+func setupCertificates(ctx context.Context, cfg *config.Config, logger logrus.FieldLogger) (*crypto.CAClient, *tls.Config, error) {
 	// Initialize CA and TLS certificates (following same pattern as API server)
 	ca, _, err := crypto.EnsureCA(cfg.CA)
 	if err != nil {
-		logger.Fatalf("ensuring CA cert: %v", err)
+		return nil, nil, fmt.Errorf("ensuring CA cert: %w", err)
 	}
 
 	var serverCerts *crypto.TLSCertificateConfig
@@ -180,7 +168,7 @@ func main() {
 	if canReadCertAndKey, _ := crypto.CanReadCertAndKey(srvCertFile, srvKeyFile); canReadCertAndKey {
 		serverCerts, err = crypto.GetTLSCertificateConfig(srvCertFile, srvKeyFile)
 		if err != nil {
-			logger.Fatalf("failed to load existing certificate: %v", err)
+			return nil, nil, fmt.Errorf("failed to load existing certificate: %w", err)
 		}
 	} else {
 		// Create new certificate with same alt names as API server
@@ -191,7 +179,7 @@ func main() {
 
 		serverCerts, err = ca.MakeAndWriteServerCertificate(ctx, srvCertFile, srvKeyFile, altNames, cfg.CA.ServerCertValidityDays)
 		if err != nil {
-			logger.Fatalf("failed to create certificate: %v", err)
+			return nil, nil, fmt.Errorf("failed to create certificate: %w", err)
 		}
 	}
 
@@ -210,72 +198,63 @@ func main() {
 	// Create TLS config
 	tlsConfig, _, err := crypto.TLSConfigForServer(ca.GetCABundleX509(), serverCerts)
 	if err != nil {
-		logger.Fatalf("failed creating TLS config: %v", err)
+		return nil, nil, fmt.Errorf("failed creating TLS config: %w", err)
 	}
 
-	tracerShutdown := tracing.InitTracer(logger, cfg, "flightctl-alertmanager-proxy")
-	defer func() {
-		if err := tracerShutdown(ctx); err != nil {
-			logger.Fatalf("Failed to shut down tracer: %v", err)
-		}
-	}()
+	return ca, tlsConfig, nil
+}
 
-	// Initialize data store
-	db, err := store.InitDB(cfg, logger)
-	if err != nil {
-		logger.Fatalf("Initializing data store: %v", err)
+// createServers sets up all server specifications for multi-server coordination
+func createServers(cfg *config.Config, logger logrus.FieldLogger, dataStore store.Store, orgCache *cache.OrganizationTTLCache, authN common.MultiAuthNMiddleware, authZ auth.AuthZMiddleware, identityMapper *service.IdentityMapper, tlsConfig *tls.Config) []shutdown.ServerSpec {
+	var servers []shutdown.ServerSpec
+
+	// Add organization cache server
+	servers = append(servers, shutdown.ServerSpec{
+		Name: "organization cache",
+		Runner: func(shutdownCtx context.Context) error {
+			orgCache.Start(shutdownCtx)
+			return nil
+		},
+	})
+
+	// Add auth provider loader if MultiAuth is configured
+	if multiAuth, ok := authN.(*authn.MultiAuth); ok {
+		servers = append(servers, shutdown.ServerSpec{
+			Name: "auth provider loader",
+			Runner: func(shutdownCtx context.Context) error {
+				return multiAuth.Start(shutdownCtx)
+			},
+		})
 	}
 
-	dataStore := store.NewStore(db, logger.WithField("pkg", "store"))
-	defer dataStore.Close()
+	// Add identity mapper server
+	servers = append(servers, shutdown.ServerSpec{
+		Name: "identity mapper",
+		Runner: func(shutdownCtx context.Context) error {
+			identityMapper.Start(shutdownCtx)
+			return nil
+		},
+	})
 
-	// Handle graceful shutdown
-	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM, syscall.SIGQUIT)
-	defer cancel()
+	// Create main HTTP server
+	servers = append(servers, shutdown.ServerSpec{
+		Name: "alertmanager proxy server",
+		Runner: func(shutdownCtx context.Context) error {
+			return runAlertmanagerProxyServer(shutdownCtx, cfg, logger, dataStore, authN, authZ, identityMapper, tlsConfig)
+		},
+	})
 
-	orgCache := cache.NewOrganizationTTL(cache.DefaultTTL)
-	go func() {
-		orgCache.Start(ctx)
-		cancel() // Trigger coordinated shutdown if cache exits
-	}()
-	defer orgCache.Stop()
+	return servers
+}
 
-	// Create service handler for auth provider access
-	baseServiceHandler := service.NewServiceHandler(dataStore, nil, nil, nil, logger, "", "", nil)
-	serviceHandler := service.WrapWithTracing(baseServiceHandler)
-
-	// Initialize auth system
-	authN, err := auth.InitMultiAuth(cfg, logger, serviceHandler)
-	if err != nil {
-		logger.Fatalf("Failed to initialize auth: %v", err)
-	}
-
-	// Start auth provider loader
-	go func() {
-		if err := authN.Start(ctx); err != nil {
-			logger.Errorf("Failed to start auth provider loader: %v", err)
-		}
-		cancel() // Trigger coordinated shutdown if auth loader exits
-	}()
-
-	authZ, err := auth.InitMultiAuthZ(cfg, logger)
-	if err != nil {
-		logger.Fatalf("Failed to initialize authZ: %v", err)
-	}
-
-	// Start multiAuthZ to initialize cache lifecycle management
+// runAlertmanagerProxyServer handles the main HTTP server logic
+func runAlertmanagerProxyServer(shutdownCtx context.Context, cfg *config.Config, logger logrus.FieldLogger, dataStore store.Store, authN common.MultiAuthNMiddleware, authZ auth.AuthZMiddleware, identityMapper *service.IdentityMapper, tlsConfig *tls.Config) error {
+	// Start multiAuthZ synchronously (it's not a continuous service)
 	if multiAuthZ, ok := authZ.(*auth.MultiAuthZ); ok {
-		multiAuthZ.Start(ctx)
+		multiAuthZ.Start(shutdownCtx)
 		logger.Debug("Started MultiAuthZ with context-based cache lifecycle")
 	}
 
-	// Create identity mapper for mapping identities to database objects
-	identityMapper := service.NewIdentityMapper(dataStore, logger)
-	go func() {
-		identityMapper.Start(ctx)
-		cancel() // Trigger coordinated shutdown if identity mapper exits
-	}()
-	defer identityMapper.Stop()
 	identityMappingMiddleware := middleware.NewIdentityMappingMiddleware(identityMapper, logger)
 
 	// Create organization extraction and validation middlewares
@@ -285,7 +264,7 @@ func main() {
 	// Create proxy
 	proxy, err := NewAlertmanagerProxy(cfg, logger)
 	if err != nil {
-		logger.Fatalf("Failed to create alertmanager proxy: %v", err)
+		return fmt.Errorf("failed to create alertmanager proxy: %w", err)
 	}
 
 	// Check if auth is disabled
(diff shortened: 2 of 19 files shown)


cmd/flightctl-api/main.go
```diff
@@ -29,17 +30,71 @@ import (
...
+	defer func() {
+		// Always run cleanup exactly once, in reverse order
+		log.Info("Starting cleanup")
+		for i := len(cleanupFuncs) - 1; i >= 0; i-- {
+			if err := cleanupFuncs[i](); err != nil {
+				log.WithError(err).Error("Cleanup error")
+			}
+		}
+		log.Info("Cleanup completed")
+	}()
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-api/main.go
```diff
@@ -101,138 +177,182 @@ func main() {
...
-					if err := systemMetricsCollector.Shutdown(); err != nil {
-						log.Errorf("Failed to shutdown system metrics collector: %v", err)
-					}
-				}()
-			}
+	setupMetricsServer(ctx, log, apiCfg.cfg, services, errCh)
+
+	// Wait for any server to complete or signal
+	log.Info("API service started, waiting for shutdown signal...")
+	err = <-errCh
```
ori-amizur: It seems errCh is passed to all servers so every one write to it, but it is read only once.  Why ?

cmd/flightctl-api/main.go
```diff
@@ -101,138 +177,182 @@ func main() {
...
+
+	// Handle shutdown reason
+	if errors.Is(err, context.Canceled) {
```
ori-amizur: Maybe we can use errgroup (golang.org/x/sync/errgroup) ?

cmd/flightctl-api/main.go
gshilin-sdb: 1. Fail-fast principle: If ANY server stops (error or normal completion), we want to stop the entire service immediately
  2. First error wins: We only care about the first failure/completion - subsequent errors are less important
  3. Coordinated shutdown: After reading the first error:
    - We return from runServers()
    - The defer cleanup runs and calls cancel()
    - Context cancellation causes all other servers to stop
    - Those servers will also write to errCh, but we've already started shutdown

cmd/flightctl-api/main.go
gshilin-sdb: For this PR: No, the current approach works fine and keeps things simple.

In general: Yes, errgroup would be cleaner and more idiomatic Go.

gshilin-sdb: Added it to plan

cmd/flightctl-api/main.go
ori-amizur: Actually we need to get all errors from the channels just to verify that all goroutines terminated.  Otherwise, it is not graceful shutdown.

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +69,91 @@ func main() {
...
 	tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-alert-exporter")
-	defer func() {
-		if err := tracerShutdown(ctx); err != nil {
-			log.Fatalf("failed to shut down tracer: %v", err)
-		}
-	}()
+	if tracerShutdown != nil {
+		cleanupFuncs = append(cleanupFuncs, func() error {
+			log.Info("Shutting down tracer")
+			return tracerShutdown(context.Background())
```
asafbss: This can block indefinitely if the exporter is slow or stuck. It’d be safer to use a short timeout, e.g. context.WithTimeout(context.Background(), 5*time.Second).

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +69,91 @@ func main() {
...
+		publisher.Close()
+		return nil
+	})
+
 	workerClient := worker_client.NewWorkerClient(publisher, log)
 
 	orgCache := cache.NewOrganizationTTL(cache.DefaultTTL)
 	go func() {
 		orgCache.Start(ctx)
 		log.Warn("Organization cache stopped unexpectedly")
```
asafbss: On normal shutdown, ctx is canceled and Start(ctx) will return, but we still log a warning saying it “stopped unexpectedly”

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +69,91 @@ func main() {
 	log.SetLevel(logLvl)
 
 	tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-alert-exporter")
-	defer func() {
-		if err := tracerShutdown(ctx); err != nil {
-			log.Fatalf("failed to shut down tracer: %v", err)
-		}
-	}()
+	if tracerShutdown != nil {
+		cleanupFuncs = append(cleanupFuncs, func() error {
```
asafbss: (nit) You might consider adding a tiny helper for cleanupFuncs:
```
addCleanup := func(f func() error) {
    cleanupFuncs = append(cleanupFuncs, f)
}
```

cmd/flightctl-alert-exporter/main.go
```diff
@@ -21,14 +24,41 @@ import (
...
+	// Build cleanup functions incrementally as resources are created
+	var cleanupFuncs []func() error
+	defer func() {
+		// First cancel context to signal all goroutines to stop
+		log.Info("Cancelling context to stop all servers")
+		cancel()
+
+		// Then run cleanup in reverse order after goroutines have stopped
+		log.Info("Starting cleanup")
+		for i := len(cleanupFuncs) - 1; i >= 0; i-- {
```
asafbss: Two thoughts here:

- Why not rely on Go’s built-in defer LIFO semantics and just defer each resource’s cleanup right after it’s created? That gives you the same ordering without any custom stack logic.

- If you prefer a central cleanup function, we could build a chained cleanup instead of a slice + reverse loop

```
cleanup := func() {}
addCleanup := func(f func() error) {
    prev := cleanup
    cleanup = func() {
        if err := f(); err != nil {
            log.WithError(err).Error("Cleanup error")
        }
        prev()
    }
}
```

cmd/flightctl-worker/main.go
```diff
@@ -85,24 +124,43 @@ func main() {
...
 		}
 	}
 
 	server := workerserver.New(cfg, log, store, provider, k8sClient, workerCollector)
-	if err := server.Run(ctx); err != nil {
-		log.Fatalf("Error running server: %s", err)
+
+	go func() {
+		log.Info("Starting worker server")
+		errCh <- server.Run(ctx)
```
asafbss: The problem is that we lose the context of which server actually failed (worker or metrics). From the final log you can’t tell what triggered shutdown. It would be clearer if each goroutine wrapped its error before sending it
`errCh <- fmt.Errorf("worker server: %w", err)`

cmd/flightctl-api/main.go
asafbss: Also the problem is that we lose the context of which server actually failed (worker or metrics). From the final log you can’t tell what triggered shutdown. It would be clearer if each goroutine wrapped its error before sending it

`errCh <- fmt.Errorf("api server: %w", err)`

cmd/flightctl-periodic/main.go
```diff
@@ -30,23 +61,41 @@ func main() {
...
+		return nil
+	})
 
 	server := periodic.New(cfg, log, store)
-	if err := server.Run(ctx); err != nil {
-		log.Fatalf("Error running server: %s", err)
+
+	// Start server and wait for completion or signal
+	log.Info("Starting periodic server")
+	err = server.Run(ctx)
```
asafbss: Other places (like the API service) we’re already running multiple servers in parallel using an errCh + goroutines. That pattern is easy to drift from: if someone later adds a second server here and just copies the simple server.Run(ctx) style, it won’t behave consistently with the multi-server/errCh pattern we use elsewhere.

Might be worth introducing a tiny helper (a small “server group” / RunGroup-style abstraction)

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: added timeout

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: the next step will be to introduce shutdown package with all kinds of helpers

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: added verification whether the context was canceled

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: The native defer approach would probably be sufficient and clearer.
But there will be a **shutdown package** soon, so you'll not see these loops here.
It will enable us:
* Centralized error handling: The slice approach allows collecting and logging cleanup errors consistently in one place, rather than having each defer handle errors independently.
* Controlled context cancellation timing: The current pattern ensures cancel() runs first, then all cleanup functions. With individual defers, you'd need to carefully place the defer cancel() at the right point to ensure proper ordering.
* Atomic cleanup logging: Having "Starting cleanup" and "Cleanup completed" log messages that bracket all cleanup operations provides clearer operational visibility.
* Consistent error recovery: If one cleanup function panics, the slice approach with a loop can potentially continue cleaning up other resources, whereas individual defers might not.

cmd/flightctl-worker/main.go
gshilin-sdb: done

cmd/flightctl-periodic/main.go
gshilin-sdb: api - done
copy/paste problem - will be answered in the **shutdown package**

cmd/flightctl-api/main.go
gshilin-sdb: applied @asafbss 's change, so we do know now

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,95 @@ func main() {
...
+	if errors.Is(err, context.Canceled) {
+		log.Info("Server stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if err != nil {
+		log.WithError(err).Error("Server stopped with error")
+		return err // Error shutdown
 	}
+
+	log.Info("Server stopped normally")
+	return nil // Normal completion
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-periodic/main.go
```diff
@@ -30,23 +62,43 @@ func main() {
...
+	if errors.Is(err, context.Canceled) {
+		log.Info("Server stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if err != nil {
+		log.WithError(err).Error("Server stopped with error")
+		return err // Error shutdown
 	}
+
+	log.Info("Server stopped normally")
+	return nil // Normal completion
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-telemetry-gateway/main.go
```diff
@@ -32,25 +61,28 @@ func main() {
...
+	if errors.Is(err, context.Canceled) {
+		log.Info("Server stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if err != nil {
+		log.WithError(err).Error("Server stopped with error")
+		return err // Error shutdown
 	}
+
+	log.Info("Server stopped normally")
+	return nil // Normal completion
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-worker/main.go
```diff
@@ -85,24 +127,43 @@ func main() {
...
+		log.Info("Servers stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if err != nil {
+		log.WithError(err).Error("Server stopped with error")
+		return err // Error shutdown
 	}
-	cancel()
+
+	log.Info("Servers stopped normally")
+	return nil // Normal completion
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-api/main.go
```diff
@@ -101,138 +177,184 @@ func main() {
...
+		log.Info("Servers stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if err != nil {
+		log.WithError(err).Error("Server stopped with error")
+		return err // Error shutdown
+	}
+
+	log.Info("Servers stopped normally")
+	return nil // Normal completion
+}
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-api/main.go
```diff
@@ -101,138 +177,209 @@ func main() {
...
+	// Handle shutdown reason
+	if errors.Is(firstError, context.Canceled) {
+		log.Info("Servers stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if firstError != nil {
+		return firstError // Error shutdown
+	}
+
+	log.Info("Servers stopped normally")
+	return nil // Normal completion
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-alert-exporter/main.go
asafbss: The benefits you mention (centralized error handling, controlled cancel timing, and consistent logging) can all be achieved with plain defer and a small helper, without introducing a custom cleanup stack or reverse-ordering logic.

Given that I’m not blocking this PR and there’s a shutdown package coming anyway, I’m fine with keeping this function simple and idiomatic for now.

@hexfusion ?

gshilin-sdb: Immediately after this PR is merged I'll update https://github.com/flightctl/flightctl/pull/2030
There you can see what will happen with this code ASAP.

hexfusion: > Why not rely on Go’s built-in defer LIFO semantics and just defer each resource’s cleanup right after it’s created?

I agree, its important that we dont create new patterns to maintain when they exist for free.

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,97 @@ func main() {
...
-	}()
+		return nil
+	})
 
 	kvStore, err := kvstore.NewKVStore(ctx, log, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password)
 	if err != nil {
-		log.Fatalf("initializing kv store: %v", err)
+		return fmt.Errorf("initializing kv store: %w", err)
 	}
-	defer kvStore.Close()
```
hexfusion, line 68: as per above there is nothing wrong with this actually its simplicity makes it idiomatic and easy to maintain.

cmd/flightctl-alert-exporter/main.go
```diff
@@ -21,14 +25,41 @@ import (
...
+	}
+}
+
+func runCmd(log *logrus.Logger) error {
+	log.Info("Starting alert exporter")
+	defer log.Info("Alert exporter stopped")
+
+	// Single context with signal handling - OS signal cancels context
+	ctx, cancel := signal.NotifyContext(context.Background(),
+		syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
```
hexfusion: if we are cancelling the context  on signal arent we forcing the servers to exit abruptly is that graceful? i would of expected that this external parent signal handling would ensure shutdown (ctx cancel) after a global timeout. then internal to each server they would handle graceful shutdown directly.

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,97 @@ func main() {
...
-		if err := tracerShutdown(ctx); err != nil {
-			log.Fatalf("failed to shut down tracer: %v", err)
-		}
-	}()
+	if tracerShutdown != nil {
+		cleanupFuncs = append(cleanupFuncs, func() error {
+			log.Info("Shutting down tracer")
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			return tracerShutdown(ctx)
```
hexfusion: this didnt need any extra ctx or timeout. when this defer fired its exactly when it should
```suggestion
defer func() {
		if err := tracerShutdown(ctx); err != nil {
			log.Error("failed to shut down tracer: %v", err)
		}
	}()
```

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,97 @@ func main() {
...
 	db, err := store.InitDB(cfg, log)
 	if err != nil {
-		log.Fatalf("initializing data store: %v", err)
+		return fmt.Errorf("initializing data store: %w", err)
 	}
 
 	store := store.NewStore(db, log.WithField("pkg", "store"))
-	defer store.Close()
+	cleanupFuncs = append(cleanupFuncs, func() error {
+		log.Info("Closing database connections")
```
hexfusion: i would expect this to print as the result of store.Close() returning. the caller should be extremely boring

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,97 @@ func main() {
...
 
 	publisher, err := worker_client.QueuePublisher(ctx, queuesProvider)
 	if err != nil {
-		log.Fatalf("initializing task queue publisher: %v", err)
+		return fmt.Errorf("initializing task queue publisher: %w", err)
 	}
-	defer publisher.Close()
+	cleanupFuncs = append(cleanupFuncs, func() error {
+		log.Info("Closing publisher")
+		publisher.Close()
```
hexfusion: same as above the defer did what it needed to if you needed a log it should be internal to publisher

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,97 @@ func main() {
...
+
 	workerClient := worker_client.NewWorkerClient(publisher, log)
 
 	orgCache := cache.NewOrganizationTTL(cache.DefaultTTL)
 	go func() {
 		orgCache.Start(ctx)
-		log.Warn("Organization cache stopped unexpectedly")
+		if ctx.Err() == nil {
+			log.Warn("Organization cache stopped unexpectedly")
+		}
```
hexfusion: a ctx error here would be only for the parent ctx right even if the child context canceled or timed out you would not see it here.

https://go.dev/play/p/LuZd0vE_k9m

cmd/flightctl-alert-exporter/main.go
```diff
@@ -39,59 +70,97 @@ func main() {
...
 	}
+
+	// Handle shutdown reason
+	if errors.Is(err, context.Canceled) {
+		log.Info("Server stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if err != nil {
+		log.WithError(err).Error("Server stopped with error")
+		return err // Error shutdown
+	}
```
hexfusion: pretty sure this should be handeled internally to the server Run if it exists cleanly why would it return an error at all an make the caller unwrap it.

cmd/flightctl-worker/main.go
```diff
@@ -85,24 +127,71 @@ func main() {
...
+
+	var firstError error
+	for i := 0; i < serversStarted; i++ {
+		if err := <-errCh; err != nil {
+			if firstError == nil {
+				firstError = err
+				// Cancel context on first error to trigger shutdown of all other servers
+				if !errors.Is(err, context.Canceled) {
+					log.Info("Triggering shutdown of all servers due to error")
+					cancel()
```
hexfusion: if the cancel is defered where it is created and when a server errors it returns do we need channels or any of this?

cmd/flightctl-worker/main.go
```diff
@@ -85,24 +127,71 @@ func main() {
...
+	// Handle shutdown reason
+	if errors.Is(firstError, context.Canceled) {
+		log.Info("Servers stopped due to shutdown signal")
+		return nil // Normal shutdown
+	} else if firstError != nil {
+		return firstError // Error shutdown
+	}
+
+	log.Info("Servers stopped normally")
+	return nil // Normal completion
```
hexfusion: is that comment needed

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: Did it.

cmd/flightctl-worker/main.go
gshilin-sdb: deleted

cmd/flightctl-worker/main.go
gshilin-sdb: 1. Removed additional cancel, the deferred one is enough
2. We need these channels to catch **all errors** from all servers (tracer, metrics), not just the first one (Ori required it)

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: changed it everywhere, but the API and worker services because they coordinate multiple servers running concurrently

cmd/flightctl-alert-exporter/main.go
gshilin-sdb: Replace with a comment. There is no way to distinguish between cancellation/timeout vs unexpected errors. The server itself should do it.

The same problem was fixed in cmd/flightctl-restore/main.go (lines 108-112)

cmd/flightctl-alert-exporter/main.go
```diff
@@ -20,71 +24,125 @@ import (
...
 	if err != nil {
-		log.InitLogs().Fatalf("reading configuration: %v", err)
+		log.InitLogs().WithError(err).Fatal("reading configuration")
 	}
 
-	log := log.InitLogs(cfg.Service.LogLevel)
-	log.Println("Starting alert exporter")
-	log.Printf("Using config: %s", cfg)
+	if err = runCmd(cfg); err != nil {
+		log.InitLogs().WithError(err).Fatal("Alert exporter error")
```
hexfusion, line 31: why init logs twice should we pass logger to runCmd?

cmd/flightctl-alert-exporter/main.go
```diff
@@ -20,71 +24,125 @@ import (
...
+
+	// Single context with signal handling - OS signal cancels context
+	ctx, cancel := signal.NotifyContext(context.Background(),
+		syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
+
+	tracerShutdown := tracing.InitTracer(logger, cfg, "flightctl-alert-exporter")
+	if tracerShutdown != nil {
+		defer func() {
+			logger.Info("Shutting down tracer")
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
hexfusion, line 45: what is the 5 second wait for? when this defer fires it means graceful termination is over.

cmd/flightctl-periodic/main.go
```diff
@@ -11,35 +16,75 @@ import (
...
-	log.Println("Starting periodic")
-	log.Printf("Using config: %s", cfg)
+	logger.Info("Starting periodic service")
+	defer logger.Info("Periodic service stopped")
+	logger.Infof("Using config: %s", cfg)
 
-	tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-periodic")
+	// Single context with signal handling - OS signal cancels context
+	ctx, cancel := signal.NotifyContext(context.Background(),
+		syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
```
hexfusion: should we defer cancel here?

cmd/flightctl-worker/main.go
hexfusion: if we defer cancel right below where its created whenever we return from runCmd we cancel for free. so this means if any goroutine exits nil error or not ctx will be cancelled and everything will shutdown. if you need coordination before returning should you use a wait.Group?

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -362,23 +341,99 @@ func main() {
...
+		logger.Fatalf("Setting up certificates: %v", err)
+	}
+
+	// Initialize tracer
+	tracerShutdown := tracing.InitTracer(logger, cfg, "flightctl-alertmanager-proxy")
+	defer func() {
+		if err := tracerShutdown(ctx); err != nil {
+			logger.Fatalf("Failed to shut down tracer: %v", err)
+		}
+	}()
```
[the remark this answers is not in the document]
gshilin-sdb: done

internal/alert_exporter/server.go
```diff
@@ -35,41 +32,38 @@ func New(
...
 
-	// Start metrics server
-	s.startMetricsServer(ctx)
+	// Start metrics server with longer grace period
+	s.startMetricsServer(shutdownCtx, ctx)
 
 	// Start uptime tracking
 	go s.updateUptimeMetric(ctx)
 
 	alertExporter := NewAlertExporter(s.log, serviceHandler, s.cfg)
```
gshilin-sdb, line 48: done

pkg/shutdown/single_server.go
```diff
@@ -0,0 +1,147 @@
...
+
+		// Phase 2: Force cancellation after grace period
+		time.AfterFunc(config.GracefulTimeout, func() {
+			config.Logger.Warn("Graceful shutdown timeout exceeded, forcing cancellation")
+			cancel()
+		})
+	}()
+
+	// Wait for server completion
+	err := <-serverDone
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-api/main.go
```diff
@@ -24,40 +23,209 @@ import (
...
+						errCh <- nil
+					}
+				}()
+
+				// Wait for first error or both to complete
+				for i := 0; i < 2; i++ {
+					if err := <-errCh; err != nil && !errors.Is(err, context.Canceled) {
+						return err
+					}
+				}
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-worker/main.go
```diff
@@ -18,84 +16,119 @@ import (
...
-
 	processID := fmt.Sprintf("worker-%s-%s", util.GetHostname(), uuid.New().String())
-	provider, err := queues.NewRedisProvider(ctx, log, processID, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password, queues.DefaultRetryConfig())
+	provider, err := queues.NewRedisProvider(initCtx, logger, processID, cfg.KV.Hostname, cfg.KV.Port, cfg.KV.Password, queues.DefaultRetryConfig())
 	if err != nil {
-		log.Fatalf("failed connecting to Redis queue: %v", err)
+		return fmt.Errorf("failed connecting to Redis queue: %w", err)
 	}
+	defer provider.Wait()
+	defer provider.Stop()
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/single_server_test.go
```diff
@@ -0,0 +1,226 @@
...
+		// Should timeout and return an error
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "failed to stop within")
+		require.Contains(t, err.Error(), "50ms timeout")
+		// Should complete relatively quickly (much less than the 200ms hang)
+		require.Less(t, duration, 150*time.Millisecond)
+	case <-time.After(2 * time.Second):
+		t.Fatal("RunSingleServer did not complete within timeout - the timeout enforcement is not working!")
+	}
+}
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/single_server.go
```diff
@@ -0,0 +1,161 @@
...
+// - This ensures the application doesn't hang indefinitely
+// - Provides a hard deadline for shutdown completion
+//
+// Benefits of Two-Phase Shutdown:
+// - Prevents data corruption by allowing transactions to complete
+// - Improves user experience by avoiding connection drops mid-request
+// - Enables proper cleanup of resources (database connections, file handles, etc.)
+// - Allows metrics to be exported during shutdown for observability
+// - Provides predictable shutdown behavior with configurable timeouts
+//
```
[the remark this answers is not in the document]
gshilin-sdb: done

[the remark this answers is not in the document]
gshilin-sdb: done

internal/alert_exporter/server.go
```diff
@@ -35,41 +32,38 @@ func New(
...
 
-	ctx, cancel := context.WithCancel(ctx)
+	// Create child context for internal operations
+	ctx, cancel := context.WithCancel(shutdownCtx)
 	defer cancel()
 
-	// Start metrics server
-	s.startMetricsServer(ctx)
+	// Start metrics server with longer grace period
+	s.startMetricsServer(shutdownCtx, ctx)
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-api/main.go
```diff
@@ -94,138 +278,68 @@ func main() {
...
+			collectors = append(collectors, httpMetricsCollector)
 		}
-
-		go func() {
-			if err := tracing.RunMetricsServer(ctx, log, cfg.Metrics.Address, collectors...); err != nil {
-				log.Errorf("Error running metrics server: %s", err)
-			}
-			cancel()
-		}()
 	}
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

cmd/flightctl-userinfo-proxy/main.go
```diff
@@ -56,9 +68,29 @@ func main() {
...
+		}
+	}()
+
+	// Start server
+	if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		return fmt.Errorf("server error: %w", err)
 	}
+
+	// Wait for graceful shutdown completion
+	return <-shutdownDone
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/multi_server.go
```diff
@@ -0,0 +1,359 @@
...
+	go func() {
+		sig := <-sigCh
+		config.Logger.WithField("signal", sig.String()).Info("Shutdown signal received, initiating graceful shutdown")
+
+		// Phase 1: Immediately signal shutdown initiation (stop accepting new requests)
+		initiateShutdown()
+
+		// Signal that shutdown was received
+		close(shutdownReceived)
+	}()
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-alert-exporter/main.go
```diff
@@ -16,75 +17,85 @@ import (
...
-	workerClient := worker_client.NewWorkerClient(publisher, log)
+
+	workerClient := worker_client.NewWorkerClient(publisher, logger)
 
 	orgCache := cache.NewOrganizationTTL(cache.DefaultTTL)
-	go func() {
-		orgCache.Start(ctx)
-		log.Warn("Organization cache stopped unexpectedly")
-	}()
 	defer orgCache.Stop()
```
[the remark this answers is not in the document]
gshilin-sdb: line 78 is a goroutine start and line 79 is Start() call inside that goroutine. Exactly as in all other cases AFAIK.

cmd/flightctl-api/main.go
```diff
@@ -24,40 +23,240 @@ import (
...
+						// continue waiting for it in background to avoid shutdown race
+						if remainingServers > 0 {
+							go func() {
+								<-errCh // Wait for remaining server to prevent shutdown race
+							}()
+						}
+						return firstError
+					}
+				}
+				return firstError
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/multi_server_test.go
```diff
@@ -0,0 +1,286 @@
...
+	require.True(t, providerStopped.Load(), "Provider should have been stopped on first error")
+}
+
+func TestRunMultiServer_MetricsServerGracePeriod(t *testing.T) {
+	logger := logrus.New()
+	logger.SetLevel(logrus.WarnLevel) // Reduce noise
+
+	var metricsStarted, metricsShutdown, regularShutdown atomic.Bool
+	metricsShutdownTime := time.Time{}
+	regularShutdownTime := time.Time{}
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-api/main.go
```diff
@@ -24,40 +23,231 @@ import (
...
+					err := <-errCh
+					remainingServers--
+
+					// Record first non-context error
+					if err != nil && !errors.Is(err, context.Canceled) && firstError == nil {
+						firstError = err
+					}
+				}
+				return firstError
+			},
```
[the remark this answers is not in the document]
gshilin-sdb: fixed

pkg/shutdown/single_server.go
```diff
@@ -0,0 +1,171 @@
...
+		sig := <-sigCh
+		config.Logger.WithField("signal", sig.String()).Info("Shutdown signal received, initiating graceful shutdown")
+
+		// Phase 1: Immediately signal shutdown initiation (stop accepting new requests)
+		initiateShutdown()
+
+		// Signal that shutdown was received
+		close(shutdownReceived)
+	}()
+
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -364,23 +340,109 @@ func main() {
...
-	logger.Printf("Alertmanager proxy listening on port %s, proxying to %s", proxyPort[1:], proxy.target.String())
+	// Start server
 	if err := server.Serve(listener); err != nil && err != http.ErrServerClosed {
-		logger.Fatalf("Server error: %v", err)
+		close(serverErr) // Signal goroutine to exit
+		return fmt.Errorf("server error: %w", err)
+	}
+
+	// Wait for graceful shutdown completion
+	return <-shutdownDone
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/single_server_test.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.config.RunSingleServer(tt.runner)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), tt.errMsg)
+		})
+	}
+}
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/multi_server_test.go
```diff
@@ -0,0 +1,293 @@
...
+		config  *MultiServerConfig
+		servers []ServerSpec
+		errMsg  string
+	}{
+		{
+			name:    "nil config",
+			config:  nil,
+			servers: []ServerSpec{server},
+			errMsg:  "config is required",
+		},
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/multi_server.go
```diff
@@ -0,0 +1,360 @@
...
+			if hasMetricsServers {
+				// Metrics servers handle their own timeouts internally, so we don't impose
+				// an additional coordination timeout that would compete with their internal logic
+				config.Logger.Infof("Graceful shutdown initiated, waiting for %d remaining servers (metrics servers handle own timeouts)", serversStarted-completedCount)
+				return config.waitForCompletionWithoutTimeout(errCh, serversStarted, completedCount, firstError)
+			} else {
+				// No metrics servers - apply hard timeout for regular servers
+				config.Logger.Infof("Graceful shutdown initiated, waiting up to %v for %d remaining servers", config.GracefulTimeout, serversStarted-completedCount)
+				return config.waitForCompletionWithTimeout(errCh, serversStarted, completedCount, firstError, forceCancel, config.GracefulTimeout)
+			}
```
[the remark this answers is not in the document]
gshilin-sdb: done

pkg/shutdown/multi_server.go
```diff
@@ -0,0 +1,360 @@
...
+			// Phase 2: Hard timeout - force cancellation and return error immediately
+			remaining := serversStarted - completedCount
+			config.Logger.Errorf("Graceful shutdown timeout exceeded after %v, forcing %d remaining servers", effectiveTimeout, remaining)
+
+			// Force cancel metrics servers
+			forceCancel()
+
+			// Return timeout error immediately - this is the hard deadline
+			return fmt.Errorf("%s: %d servers failed to stop within %v timeout", config.ServiceName, remaining, effectiveTimeout)
+		}
```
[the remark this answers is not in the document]
gshilin-sdb: done

cmd/flightctl-alert-exporter/main.go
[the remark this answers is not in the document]
gshilin-sdb: I explained you this code is ok

PR #2010, opened 2025-11-20
EDM-2607: proper support for AP enabled flag

asafbennatan:
## Summary by CodeRabbit

* **New Features**
  * Providers now expose an enabled/disabled state and the auth config reflects provider Enabled flags.

* **Bug Fixes**
  * Auth config and token resolution consider only enabled providers.
  * Kubernetes and AAP providers are explicitly enabled by default when configured.
  * Auth config ordering is deterministic and no longer produces unintended defaults for fully disabled setups.
  * Provider listing API returns both enabled and disabled providers as expected.

* **Tests**
  * Test coverage updated to validate enabled/disabled provider behavior.

### DIFF
--- a/internal/auth/authn/multiauth.go
+++ b/internal/auth/authn/multiauth.go
@@ -85,6 +85,9 @@ func NewMultiAuth(authProviderService AuthProviderService, tlsConfig *tls.Config
 func (m *MultiAuth) AddStaticProvider(issuer string, provider common.AuthNMiddleware) {
 	m.staticProviders[issuer] = provider
 }
+func (m *MultiAuth) IsEnabled() bool {
+	return true
+}
 
 // HasProviders returns true if any providers are configured
 func (m *MultiAuth) HasProviders() bool {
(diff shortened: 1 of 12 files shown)


internal/auth/authn/multiauth.go
```diff
@@ -619,23 +627,83 @@ func (m *MultiAuth) GetAuthConfig() *api.AuthConfig {
...
+		// Static providers from config file may not have the enabled flag structure
+		// Consider them enabled by default if they don't have the full provider structure
+		return true
+	}
+
+	// Check all providers in the config (typically just one per middleware)
+	for _, provider := range *config.Providers {
+		// Check if this specific provider is enabled
+		if !m.isProviderEnabled(&provider) {
+			return false
```
keitwb: I know it says "typically just one per middleware" but what would it mean if there were two in a given middleware and one was enabled and the other disabled? This would return false but is that right?

internal/auth/authn/multiauth.go
```diff
@@ -619,23 +627,83 @@ func (m *MultiAuth) GetAuthConfig() *api.AuthConfig {
...
 	if err != nil || parsedToken.Issuer() == "" {
-		// Not a JWT token or JWT without issuer - return all possible providers for opaque tokens
+		// Not a JWT token or JWT without issuer - return all possible enabled providers for opaque tokens
 		providers := []common.AuthNMiddleware{}
 
-		// Add all static providers
+		// Add all enabled static providers
 		for _, provider := range m.staticProviders {
-			providers = append(providers, provider)
+			if m.isProviderMiddlewareEnabled(provider) {
```
keitwb: It's very confusing that this is called a provider (singular) here but then in the called function it is called a "middleware", and then within it, there are potentially multiple nested providers (plural).

internal/auth/authn/multiauth.go
```diff
@@ -619,23 +627,83 @@ func (m *MultiAuth) GetAuthConfig() *api.AuthConfig {
...
+		}
+	case string(api.K8s):
+		if spec, err := provider.Spec.AsK8sProviderSpec(); err == nil {
+			return spec.Enabled != nil && *spec.Enabled
+		}
+	}
+
+	// If no Enabled field found or if provider type is unknown, don't include it
+	return false
+}
```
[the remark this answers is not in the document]
asafbennatan: no

internal/auth/authn/multiauth.go
asafbennatan: fixed

internal/auth/authn/multiauth.go
asafbennatan: fixed

PR #2015, opened 2025-11-20
EDM-2624: Review db section of values.yaml

rawagner:
Fix external DB config - ssl certs
Passwords are passed as Secret reference only

## Summary by CodeRabbit

* **Refactor**
  * Restructured DB config to db.type with nested builtin/external blocks, added db.name and per-user secret-name fields (application/migration/master). Builtin now contains image, resources and storage; external holds hostname, port and TLS references. Port/hostname selection and secret resolution use templated helpers. SSL settings and mounts apply only for external DBs. KV password now uses a configurable secret name.

* **Documentation**
  * Updated README and external-database install docs/examples to the new nested, secret-based configuration.


fzdarsky:
Can we consider having all the DB connection parameters as part of the Secrets (i.e. not only username and password, but also hostname, port, and SSL parameters (which should be called TLS parameters these days)? That'd allow users to simply replace that Secret to point to an external DB and then disable deployment of the internal DB.

Another advantage would be we can move to mounting the Secret as a volume instead of passing stuff in through the environment variables, which is less secure.

rawagner:
> Can we consider having all the DB connection parameters as part of the Secrets (i.e. not only username and password, but also hostname, port, and SSL parameters (which should be called TLS parameters these days)? That'd allow users to simply replace that Secret to point to an external DB and then disable deployment of the internal DB.
> 


Putting hostname/port/.. to a secret means changes to our go code - which at this point, with limited time, I would like to avoid. Maybe as a separate PR, if we find some more time ?

> Another advantage would be we can move to mounting the Secret as a volume instead of passing stuff in through the environment variables, which is less secure.


The `sslmode` aligns with the pgsql param name. Do we want to change it to `tlsMode` in our helm chart values?
I agree with the others
```
sslConfigMapName -> tlsConfigMapName
sslSecretName -> tlsSecretName
```

PR #2017, opened 2025-11-21
NO-ISSUE: BREAKING: agent/device/status: add embedded and appType

hexfusion:
## Summary by CodeRabbit

* **New Features**
  * Device application status now exposes two new metadata properties:
    * appType — specifies the application type for better classification
    * embedded — boolean indicating whether the application is embedded

* **Documentation / API**
  * OpenAPI schema and embedded API spec updated to add and require these two fields in DeviceApplicationStatus.


api/v1alpha1/openapi.yaml
```diff
@@ -5189,6 +5189,11 @@ components:
           description: Number of restarts observed for the application.
         status:
           $ref: "#/components/schemas/ApplicationStatusType"
+        embedded:
+          type: boolean
+          description: Whether the application uses embedded inline content.
```
fzdarsky:
```suggestion
          description: Whether the application is embedded in the bootc image.
```

fzdarsky (approved):
Just a nit. Do we want to require the new fields?

hexfusion:
I did it to be backwards compatible but we can for 1.0 as upgrades from 0.10 are not supported and the change is for the greater good moving forward.

celdrake:
@hexfusion  @fzdarsky @pamparan Do we want to display these two fields in the UI.
The AI Prototype shows the appType, but what about the "embedded"?

hexfusion:
The intention is yes embedded is important for folks who will for example be using the software catalog. The signal will be useful in trying to understand early vs late binding applications.

PR #2018, opened 2025-11-21
EDM-2623: Add quadlet generator logs to error messages

kkyrazis:
Quadlets are generated after a `systemctl daemon-reload`, but quadlets can fail to generate for various reasons (i.e. a Key was added to the file that isn't supported in that particular version). This validation is incredibly difficult for us to do as it is extremely dependent on the podman version on the agent. 

This PR introduces behavior so that if we fail to start services after calling `systemctl daemon-reload` that it will check the journal for `quadlet-generator` logs and include those logs in the logs/error message.

```yaml
updated:
        info: 'Device has not been updated to the latest device spec: Failed to update to renderedVersion: 9: after update: error executing actions: quadlet generator: converting "my-app-4-198055-test-name.volume": unsupported key ''Name'' in group ''Volume'' in /etc/containers/systemd/my-app-4/my-app-4-198055-test-name.vo...'
```

Also fixes a bug in which adding apps and triggering an OS update in the same spec update will result in attempting to start applications during a shutdown operation. 

## Summary by CodeRabbit

* **New Features**
  * System journal and per-service logs are now fetched and attached to startup and service failure outputs for richer diagnostics.
  * Batch timestamps are propagated into action and lifecycle operations for clearer grouping and tracking.

* **Bug Fixes**
  * Improved propagation and handling of batch timestamps and last-success times to prevent missed or repeated actions.
  * Drain now ensures pending actions are processed before shutdown flows.

### DIFF
--- a/internal/agent/client/journalctl.go
+++ b/internal/agent/client/journalctl.go
@@ -24,10 +24,37 @@ type Journalctl struct {
 	exec executer.Executer
 }
 
-func (j *Journalctl) logsSince(ctx context.Context, baseArgs []string, since time.Time) ([]string, error) {
-	formattedTime := since.Format("2006-01-02 15:04:05")
-	args := append(baseArgs, "-o", "cat", "--since", formattedTime)
-	stdout, stderr, exitCode := j.exec.ExecuteWithContext(ctx, journalctlCommand, args...)
+type logOptions struct {
+	args []string
+}
+
+type LogOptions func(*logOptions)
+
+func WithLogUnit(unit string) LogOptions {
+	return func(o *logOptions) {
+		o.args = append(o.args, "-u", unit)
+	}
+}
+
+func WithLogTag(tag string) LogOptions {
+	return func(o *logOptions) {
+		o.args = append(o.args, "-t", tag)
+	}
+}
+
+func WithLogSince(t time.Time) LogOptions {
+	return func(o *logOptions) {
+		o.args = append(o.args, "--since", t.Format("2006-01-02 15:04:05"))
+	}
+}
+
+func (j *Journalctl) Logs(ctx context.Context, options ...LogOptions) ([]string, error) {
+	opts := logOptions{args: []string{"-o", "cat"}}
+	for _, option := range options {
+		option(&opts)
+	}
+
+	stdout, stderr, exitCode := j.exec.ExecuteWithContext(ctx, journalctlCommand, opts.args...)
 	if exitCode != 0 {
 		return nil, fmt.Errorf("journalctl logs: %w", errors.FromStderr(stderr, exitCode))
 	}
@@ -41,12 +68,4 @@ func (j *Journalctl) logsSince(ctx context.Context, baseArgs []string, since tim
 	}
 
 	return result, nil
-}
-
-func (j *Journalctl) LogsByTagSince(ctx context.Context, tag string, since time.Time) ([]string, error) {
-	return j.logsSince(ctx, []string{"-t", tag}, since)
-}
-
-func (j *Journalctl) LogsByUnitSince(ctx context.Context, unit string, since time.Time) ([]string, error) {
-	return j.logsSince(ctx, []string{"-u", unit}, since)
 }
--- a/internal/agent/device/applications/lifecycle/quadlet.go
+++ b/internal/agent/device/applications/lifecycle/quadlet.go
@@ -70,9 +70,9 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
 			// check the individual service logs and the quadlet generator logs for potential issues
 			err = fmt.Errorf("starting units: %w", err)
 			for _, service := range services {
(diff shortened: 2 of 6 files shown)


internal/agent/client/journalctl.go
```diff
@@ -0,0 +1,52 @@
...
+	return &Journalctl{
+		exec: exec,
+	}
+}
+
+type Journalctl struct {
+	exec executer.Executer
+}
+
+func (j *Journalctl) logsSince(ctx context.Context, baseArgs []string, since time.Time) ([]string, error) {
```
hexfusion:
```suggestion
func (j *Journalctl) since(ctx context.Context, baseArgs []string, since time.Time) ([]string, error) {
```

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -56,7 +66,29 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
...
+					q.log.Infof("Service: %q logs: %s", service, strings.Join(serviceLogs, "\n"))
+					err = fmt.Errorf("service: %q logs: %s: %w", service, strings.Join(serviceLogs, ","), err)
+				}
+			}
+			generatorLogs, logsErr := q.systemdManager.LogsByTagSince(ctx, "quadlet-generator", batchTime)
+			if logsErr != nil {
+				q.log.Warnf("Failed to fetch quadlet-generator logs: %v", logsErr)
+			}
+			if len(generatorLogs) > 0 {
+				q.log.Errorf("quadlet generator logs: %s", strings.Join(generatorLogs, "\n"))
```
hexfusion: what does generator mean for the user, just wonder if this can be more human friendly journal?

internal/agent/device/applications/podman_monitor.go
```diff
@@ -59,16 +60,23 @@ func NewPodmanMonitor(
 	bootTime string,
 	rw fileio.ReadWriter,
 ) *PodmanMonitor {
+	// don't fail for this. This is being parsed purely for informational reasons in the event something fails
+	startTime, err := time.Parse(time.RFC3339, bootTime)
+	if err != nil {
+		log.Warnf("Failed to parse bootTime %q: %v", bootTime, err)
```
hexfusion: sort of think this is an error while we would not stop execution its unexpected

internal/agent/device/systemd/systemd.go
```diff
@@ -33,20 +34,26 @@ type Manager interface {
 	ResetFailed(ctx context.Context, units ...string) error
 	// ListUnitsByMatchPattern lists systemd units matching the provided patterns.
 	ListUnitsByMatchPattern(ctx context.Context, matchPatterns []string) ([]client.SystemDUnitListEntry, error)
+	// LogsByTagSince retrieves journal logs for a specific tag since a given time.
+	LogsByTagSince(ctx context.Context, tag string, since time.Time) ([]string, error)
+	// LogsByUnitSince retrieves journal logs for a specific unit since a given time.
+	LogsByUnitSince(ctx context.Context, unit string, since time.Time) ([]string, error)
```
hexfusion:
```suggestion
	Logs(ctx context.Context, unit string, since time.Time opts ... SystemdOption) ([]string, error)
```
would it be useful to combine
```go
systemd.Logs(ctx, since, WithTag("something"), WithUnit("something-else"))
```

kkyrazis: I like this.

internal/agent/device/applications/lifecycle/quadlet.go
kkyrazis: changed the language to "Failed to generate services from the defined Quadlet. Check the syntax of the Quadlet files. [logs]"

internal/agent/device/applications/podman_monitor.go
```diff
@@ -59,16 +60,23 @@ func NewPodmanMonitor(
...
-		apps:          make(map[string]Application),
-		lastEventTime: bootTime,
-		log:           log,
-		rw:            rw,
+		apps:                   make(map[string]Application),
+		lastEventTime:          bootTime,
+		lastActionsSuccessTime: startTime,
+		log:                    log,
+		rw:                     rw,
 	}
```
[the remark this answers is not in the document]
kkyrazis: You are correct in that `uptime -s` is used to gather the boot time, however if you read that function a little more you'll see something very interesting:

```
	args := []string{"-s"}
	stdout, stderr, exitCode := exec.ExecuteWithContext(ctx, "uptime", args...)
	if exitCode != 0 {
		return "", fmt.Errorf("device uptime: %w", errors.FromStderr(stderr, exitCode))
	}

	// parse boot time in local timezone since uptime -s returns timestamp in local time
	bootTime, err := time.ParseInLocation("2006-01-02 15:04:05", strings.TrimSpace(stdout), time.Local)
	if err != nil {
		return "", err
	}

	return bootTime.UTC().Format(time.RFC3339), nil
```

It's returned formatted as RFC3339!

PR #2022, opened 2025-11-21
EDM-2624: Helm values.yaml cleanup

rawagner:
## Summary by CodeRabbit

* **Configuration Improvements**
  * Added global storage class and image pull policy options; clarified image pull secret naming.
  * Redis now references a password secret name instead of plaintext; memory settings exposed.
  * Many component enable flags removed; several components and probes now render by default.

* **Behavioral Changes**
  * Tracing/telemetry and metrics now depend on presence of config objects rather than explicit enabled flags.
  * UI organizations setting now always enabled.

* **Documentation**
  * Helm chart README updated to reflect new configuration layout.


deploy/helm/flightctl/values.yaml
```diff
@@ -104,14 +93,9 @@ global:
...
-  imagePullPolicy: IfNotPresent
-  rbac:
-    # -- Create RBAC resources (roles, bindings, service accounts)
-    create: true
-  organizations:
-    # -- Enable IDP-provided organizations support
-    enabled: false
+  metrics:
+    # -- Enable metrics exporting and service
+    enabled: true
```
rawagner: do we want to be able to disable metrics exporting ?

fzdarsky: I don't see a reason to.

fzdarsky (approved):

PR #2024, opened 2025-11-21
EDM-2631: fail upgrade if critical disk alert is firing

hexfusion:
This PR ensures that no new images are pulled by the dependency manager in the case that a critical alert if firing for Disk. This PR also allows for an escape hatch in the case there is a misconfiguration alert. For example if my accident the user declared the disk monitor for 1%. This would otherwise block progress for the agent. The user can update the monitor and it will be reflected `BeforeUpdate` this happens automatically. 

## Summary by CodeRabbit

* **New Features**
  * Resource configurations are applied before device updates; updates are blocked when critical resource alerts (disk/CPU/memory) are present.
  * Prefetch/scheduling checks resource alerts to prevent new downloads when storage is critical.

* **Bug Fixes**
  * Improved error reporting for critical resource conditions that block upgrades.
  * New explicit critical-resource error symbol for surfaced alerts.

* **Refactor**
  * Resource management API and wiring simplified: controller removed and manager interface extended and injected.

* **Tests**
  * Updated tests and mocks to cover pre-update sync, critical-alert checks, and prefetch behavior.

### DIFF
--- a/docs/user/references/alerts.md
+++ b/docs/user/references/alerts.md
@@ -36,6 +36,9 @@ Flight Control automatically generates alerts for the following conditions:
   - `DeviceDiskWarning`: Disk usage exceeds warning threshold
   - `DeviceDiskNormal`: Resolves disk alerts when usage returns to normal
 
+> [!NOTE]
+> When a critical disk alert is active, device upgrades that require downloading OCI images will automatically fail with an error message prompting the user to clear storage. This prevents upgrade failures due to insufficient disk space.
+
 ### Application Alerts
 
 - **Application Status**:
--- a/docs/user/using/managing-devices.md
+++ b/docs/user/using/managing-devices.md
@@ -850,6 +850,9 @@ spec:
 [...]
 ```
 
+> [!NOTE]
+> When a critical disk alert is active, device upgrades that require downloading OCI images will automatically fail to prevent upgrade failures due to insufficient disk space. The upgrade will fail with an error message prompting you to clear storage before attempting the upgrade again.
+
 ## Accessing Devices Remotely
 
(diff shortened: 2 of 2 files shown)


internal/agent/device/device.go
```diff
@@ -306,6 +306,11 @@ func (a *Agent) beforeUpdate(ctx context.Context, current, desired *v1alpha1.Dev
 		return fmt.Errorf("download policy: %w", err)
 	}
 
+	// check for critical disk storage alerts before proceeding with update
+	if a.resourceController.IsCriticalAlert(resource.DiskMonitorType) {
```
kkyrazis: How would we recover if someone accidentally defined critical as something absurdly low (1%) and effectively couldn't clear storage space?

hexfusion: aha great catch, yeah I was thinking that was covered but that is only in the policy will add for resource.

internal/agent/agent.go
```diff
@@ -339,12 +339,6 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log.Warnf("Failed to create gRPC client: %v", err)
 	}
 
-	// create resource controller
-	resourceController := resource.NewController(
-		a.log,
-		resourceManager,
-	)
-
```
hexfusion, line 347: note: since we are reconciling BeforeUpdate the controller is no longer needed

internal/agent/device/device.go
```diff
@@ -385,9 +389,7 @@ func (a *Agent) syncDevice(ctx context.Context, current, desired *v1alpha1.Devic
 		return fmt.Errorf("config: %w", err)
 	}
 
-	if err := a.resourceController.Sync(ctx, desired.Spec); err != nil {
-		return fmt.Errorf("resources: %w", err)
-	}
+	// resources are already synced in beforeUpdate, no need to sync again here
```
kkyrazis: thanks siri

kkyrazis (approved):

fzdarsky (approved):

PR #2025, opened 2025-11-22
NO-ISSUE: *: use math/rand/v2

hexfusion:
This PR updates all usage of math/rand to math/rand/v2 which has many improvements and optimizations as described in this blog[1] by Russ Cox

[1] https://go.dev/blog/randv2

## Summary by CodeRabbit

* **Chores**
  * Replaced legacy randomness implementation and added a linter rule enforcing the new RNG usage.
  * Changed config factory to return by value and improved defaults so RNG is safely auto-initialized when absent.

* **Tests**
  * Updated tests and helpers for the new RNG and added coverage verifying nil-RNG auto-initialization.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,6 +60,7 @@ issues:
 linters:
   enable:
     - copyloopvar
+    - depguard
     - errcheck
     - gci
     - gocyclo
@@ -78,6 +79,13 @@ linters:
 linters-settings:
   enable:
     - shadow
+
+  depguard:
+    rules:
+      main:
+        deny:
+          - pkg: "^math/rand$"
+            desc: "Use math/rand/v2 instead"
 
   govet:
     settings:
--- a/internal/agent/device/spec/spec.go
+++ b/internal/agent/device/spec/spec.go
@@ -2,7 +2,6 @@ package spec
 
 import (
 	"context"
-	"math/rand"
 	"time"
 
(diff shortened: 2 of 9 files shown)


.golangci.yml
```diff
@@ -79,6 +80,13 @@ linters-settings:
   enable:
     - shadow
 
+  depguard:
+    rules:
+      main:
+        deny:
+          - pkg: "^math/rand$"
+            desc: "Use math/rand/v2 instead"
+
```
fzdarsky, line 89: 👍 Good idea. I'm thinking we should review our deps and see whether there are other packages we may want to keep consistent. Thinking of YAML parsers and logging.

internal/agent/device/spec/spec.go
```diff
@@ -28,13 +27,14 @@ const (
 )
 
 // defaultSpecPollConfig is the default poll configuration for the spec priority queue.
-var defaultSpecPollConfig = poll.Config{
-	BaseDelay:    30 * time.Second,
-	Factor:       1.5,
-	MaxDelay:     5 * time.Minute,
-	JitterFactor: 0.1,
-	Rand:         rand.New(rand.NewSource(time.Now().UnixNano())), //nolint:gosec
```
fzdarsky, line 36: You probably want to initialize this?

hexfusion: we lazy init newRand now to save user needing to inline.

PR #2029, opened 2025-11-23
EDM-2544: CLI support for new Login Functionality

asafbennatan:
## Summary by CodeRabbit

* **New Features**
  * Provider-based login with selectable auth providers (OIDC, OAuth2, OpenShift, AAP) and callback-port support
  * PKCE and ID token support, direct token and no-op providers for flexible auth modes
  * Automatic token renewal and organization auto-selection after login

* **Bug Fixes**
  * Stronger CSRF/state validation and improved TLS/error handling during login flows

* **Removed Features**
  * Kubernetes OAuth2 login removed


internal/cli/login/oidc.go
```diff
@@ -59,96 +73,43 @@ func (o OIDC) getOIDCClient(callback string) (*osincli.Client, error) {
...
-		defer resp.Body.Close()
-		bodyBytes, err = io.ReadAll(resp.Body)
-		if err != nil {
-			return AuthInfo{}, fmt.Errorf("failed to read OIDC response: %w", err)
-		}
+	authInfo.TokenToUse = TokenToUseType(authInfo.TokenToUse)
+	if o.Spec.Scopes != nil && slices.Contains(*o.Spec.Scopes, "openid") {
+		authInfo.TokenToUse = TokenToUseIdToken
 	}
+	return authInfo, nil
```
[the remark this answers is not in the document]
asafbennatan: no

internal/cli/login/oauth2.go
```diff
@@ -0,0 +1,102 @@
...
+		AuthorizeUrl:             o.Spec.AuthorizationUrl,
+		TokenUrl:                 tokenProxyURL,
+		ErrorsInStatusCode:       true,
+		SendClientSecretInParams: true, // this makes sure we send the client id , the secret is not filled
+		RedirectUrl:              callback,
+		Scope:                    strings.Join(*o.Spec.Scopes, " "),
+		CodeVerifier:             codeVerifier,
+		CodeChallenge:            codeChallenge,
+		CodeChallengeMethod:      "S256",
+	}
```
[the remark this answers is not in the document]
asafbennatan: we validate stuff in the Validate function , there is no need to do that again

internal/cli/login/oauth.go
```diff
@@ -9,95 +14,138 @@ import (
...
 	err = <-done
-	return ret, err
+
+	// Shutdown the server gracefully with a timeout to allow the response to be sent
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	_ = server.Shutdown(ctx)
+
+	return result, err
 }
```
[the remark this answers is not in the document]
asafbennatan: this is how we had it before - i am not sure i want to change this now

internal/cli/login/aap.go
```diff
@@ -31,33 +34,44 @@ func (c *AAPRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		TokenUrl:                 fmt.Sprintf("%s/o/token/", authUrl),
 		ErrorsInStatusCode:       true,
-		SendClientSecretInParams: true,
+		SendClientSecretInParams: true, // this makes sure we send the client id , the secret is not filled
 		RedirectUrl:              callback,
 		Scope:                    "read",
+		CodeVerifier:             codeVerifier,
+		CodeChallenge:            codeChallenge,
+		CodeChallengeMethod:      "S256",
 	}
```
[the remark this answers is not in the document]
asafbennatan: proxy does not handle aap this is why

internal/cli/login.go
```diff
@@ -501,66 +502,49 @@ func (o *LoginOptions) Run(ctx context.Context, args []string) error {
...
+		authCAFile, err = o.getAbsAuthCAFile()
+		if err != nil {
+			return err
+		}
+		o.clientConfig.AuthInfo.AuthProvider.CAFile = authCAFile
+	}
+	token := authInfo.AccessToken
+	if authInfo.TokenToUse == login.TokenToUseIdToken {
+		token = authInfo.IdToken
 	}
```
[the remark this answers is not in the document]
asafbennatan: this is fine

internal/cli/login/aap.go
```diff
@@ -31,33 +34,48 @@ func (c *AAPRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	// Use the API server's token proxy endpoint instead of the AAP provider's token endpoint
+	if o.Metadata.Name == nil {
+		return nil, fmt.Errorf("provider name is required")
+	}
 
 	config := &osincli.ClientConfig{
 		ClientId:                 o.ClientId,
-		AuthorizeUrl:             oauthServerResponse.AuthEndpoint,
-		TokenUrl:                 oauthServerResponse.TokenEndpoint,
+		AuthorizeUrl:             fmt.Sprintf("%s/o/authorize/", authUrl),
```
ori-amizur, line 73: Is this format the only acceptable format ?
Why is the URL (in the spec) not in its final format ?

internal/cli/login.go
```diff
@@ -255,201 +247,100 @@ func (o *LoginOptions) Validate(args []string) error {
...
-			},
+			AuthProvider:       *provider,
+			CAFile:             o.AuthCAFile,
+			InsecureSkipVerify: o.InsecureSkipVerify,
 		},
-		OrganizationsEnabled: orgEnabled,
-	}, o.InsecureSkipVerify)
-	if err != nil {
-		return fmt.Errorf("creating auth provider: %w", err)
+		OrganizationsEnabled: true,
```
avishayt, line 337: Why always true?

internal/cli/login/aap.go
```diff
@@ -31,33 +34,48 @@ func (c *AAPRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		TokenUrl:                 fmt.Sprintf("%s/o/token/", authUrl),
```
avishayt, line 74: getTokenProxyURL?

internal/cli/login/oauth.go
```diff
@@ -9,95 +14,137 @@ import (
...
 	}
-	defer listener.Close()
 
-	port := listener.Addr().(*net.TCPAddr).Port
-	done := make(chan error, 1)
-	mux := http.NewServeMux()
-	server := &http.Server{Handler: mux} // #nosec G112
-	defer server.Close()
-	callback := fmt.Sprintf("http://127.0.0.1:%d/callback", port)
+	listener, err := net.Listen("tcp", "127.0.0.1:8080")
```
avishayt: Should the port be configurable?

internal/cli/login.go
```diff
@@ -255,201 +247,100 @@ func (o *LoginOptions) Validate(args []string) error {
...
+		return login.NewNilAuth(), client.AuthInfo{}, nil
 	}
-}
 
-// getDefaultProvider returns the default authentication provider
-func (o *LoginOptions) getDefaultProvider() *v1alpha1.AuthProvider {
-	if o.authConfig == nil || o.authConfig.Providers == nil {
-		return nil
+	// Create web-based auth provider from config
+	var providerName string = *o.authConfig.DefaultProvider
```
avishayt: check for nil pointers?

internal/client/config.go
```diff
@@ -86,10 +93,22 @@ type AuthInfo struct {
...
+	// +optional
+	AccessTokenExpiry string `json:"access-token-expiry,omitempty"`
+	// RefreshToken is the OAuth2/OIDC refresh token for obtaining new access tokens
 	// +optional
-	Token string `json:"token,omitempty"`
-	// The authentication provider (i.e. k8s, OIDC)
+	RefreshToken string `json:"refresh-token,omitempty"`
+	// IdToken is the OIDC ID token containing user identity information
+	// +optional
+	IdToken string `json:"id-token,omitempty"`
```
ori-amizur, line 107: Why do we need token types ? Only one of them is in use at a time?

internal/cli/login.go
```diff
@@ -216,13 +211,10 @@ func (o *LoginOptions) Bind(fs *pflag.FlagSet) {
 	o.GlobalOptions.Bind(fs)
 
 	fs.StringVarP(&o.AccessToken, "token", "t", o.AccessToken, "Bearer token for authentication to the API server")
-	fs.BoolVarP(&o.Web, "web", "w", o.Web, "Login via browser")
```
ori-amizur, line 219: Why was this removed?

internal/cli/login.go
```diff
@@ -216,13 +211,10 @@ func (o *LoginOptions) Bind(fs *pflag.FlagSet) {
...
-	fs.StringVarP(&o.ClientId, "client-id", "", o.ClientId, "ClientId to be used for OAuth2 requests")
+	fs.StringVarP(&o.Provider, "provider", "", o.Provider, "Name of the authentication provider to use")
 	fs.StringVarP(&o.CAFile, "certificate-authority", "", o.CAFile, "Path to a cert file for the certificate authority")
 	fs.StringVarP(&o.AuthCAFile, "auth-certificate-authority", "", o.AuthCAFile, "Path to a cert file for the auth certificate authority")
 	fs.BoolVarP(&o.InsecureSkipVerify, "insecure-skip-tls-verify", "k", o.InsecureSkipVerify, "If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure")
-	fs.StringVarP(&o.Username, "username", "u", o.Username, "Username for server")
```
ori-amizur, line 224: Even if we choose provider - can't we set username/password ?

internal/client/authorization.go
```diff
@@ -31,63 +31,101 @@ type accessTokenRefresher struct {
...
+		clientId := "aap-gateway-client"
+		return login.NewAAPOAuth2Config(provider.Metadata, aapSpec, clientId, caFile, insecure, apiServerURL), nil
+
+	case string(api.K8s):
+		return nil, fmt.Errorf("k8s auth requires providing --token flag")
 	default:
-		return nil, fmt.Errorf("unsupported auth provider: %s", authInfo.AuthProvider.Type)
+		return nil, fmt.Errorf("unsupported auth provider type: %s", providerType)
 	}
 }
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/cli/login/aap.go
asafbennatan: we do not support proxying aap url since we dont keep its client secret

internal/cli/login.go
asafbennatan: due to latest changes there is no org disabled - we always run with orgs , there is a ticket to even remove all of that functionality from flightctl (keep the api structure the same though)

internal/client/config.go
```diff
@@ -72,6 +72,13 @@ type Service struct {
 	InsecureSkipVerify       bool   `json:"insecureSkipVerify,omitempty"`
 }
 
+type TokenToUseType string
+
+const (
+	TokenToUseAccessToken TokenToUseType = "access"
+	TokenToUseIdToken     TokenToUseType = "id"
+)
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/client/config.go
asafbennatan: well - refresh token existed before , i only made a clear separation as for the idtoken - this is part of the oidc spec - > we need to use the id_token in that case , not the access token , in the config i keep all types so we have it if we want to use any of them later

internal/cli/login.go
asafbennatan: we allways login with web=true , there is no username password login case , and for the token case users use the --token flag

internal/cli/login.go
asafbennatan: we do not support username/password authentication - only oauth flow based and token - i believe the old username password was constructed to support keyclock specific authentication which is not necessary anymore since we have proper oauth2/oidc support

internal/cli/login/aap.go
asafbennatan: for aap this is the format - we did not do work to support this coming from the config - i suspect that we might want to do this on a separate PR

internal/cli/login.go
asafbennatan: fixed , note if we dont have providers we are already in the nilauth scenario and never reached here

internal/cli/login/oauth.go
asafbennatan: added , note this will require the user to setup thier oidc/oauth2 provider with the correct port in the callback ( they had to have localhost:8080 there anyway)

internal/cli/login/aap.go
ori-amizur: Any reason why the authUrl doesn't contain the full URL?  Even if this is the aap format

internal/client/config.go
ori-amizur: Refresh token has different use.  It seems both id_token and auth_token are used for the same purpose.

internal/cli/login/aap.go
asafbennatan: it just doesnt - we hold the aap api url not the authorize and token - > i agree that we can add all the details we have the same like oauth - but for that we will need a server change - i prefer this PR to stay focused on CLI changes

internal/cli/login/aap.go
```diff
@@ -79,29 +99,20 @@ func (o AAPOAuth) getOAuth2Client(callback string) (*osincli.Client, error) {
...
-	}
-
-	if args.Web && StrIsEmpty(args.ClientId) {
-		return fmt.Errorf("--client-id must be specified for AAP Gateway auth")
-	}
-
+func (o *AAPOAuth) Validate(args ValidateArgs) error {
+	// No validation needed - token is optional, web flow is default
 	return nil
 }
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/cli/login/nilauth.go
```diff
@@ -0,0 +1,27 @@
...
+	return AuthInfo{}, nil
+}
+
+// Validate performs no validation - authentication is disabled
+func (n *NilAuth) Validate(args ValidateArgs) error {
+	return nil
+}
+
+func (n *NilAuth) SetInsecureSkipVerify(insecureSkipVerify bool) {
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/cli/login/oauth.go
ori-amizur: There is a possibility to use anonymous port.  In this case it will be allocated automatically by the OS.  So no port configuration is needed.

ori-amizur: Having static port configuration may cause authentication requests to fail if there is another active authentication request

internal/cli/login/oauth.go
```diff
@@ -9,95 +14,137 @@ import (
...
 	}
-	defer listener.Close()
 
-	port := listener.Addr().(*net.TCPAddr).Port
-	done := make(chan error, 1)
-	mux := http.NewServeMux()
-	server := &http.Server{Handler: mux} // #nosec G112
-	defer server.Close()
-	callback := fmt.Sprintf("http://127.0.0.1:%d/callback", port)
+	listener, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", port))
```
ori-amizur, line 80: This will fail if couple of CLIs will attempt login at the same time (rare but possible).
Anonymous port is preferable for this

asafbennatan: this is not possible - you need redirect url to be set how will you do that ?
(redirect url must be configured at the issuer)

ori-amizur: Is this URL known to anyone else besides getClient (line 87) ?

internal/cli/login/oauth.go
asafbennatan: The point is that you need the redirect url to be allowed by the issuer so if it's random how will you allow it in the issuer ?

internal/client/config.go
asafbennatan: But Id token and access token are different, I don't see a reason to mix them

internal/client/authorization.go
```diff
@@ -29,65 +29,109 @@ type accessTokenRefresher struct {
...
 	if authInfo.ExpiresIn != nil {
-		c.config.AuthInfo.AuthProvider.Config[AuthAccessTokenExpiryKey] = time.Now().Add(time.Duration(*authInfo.ExpiresIn) * time.Second).Format(time.RFC3339Nano)
+		expiryTime := time.Now().Add(time.Duration(*authInfo.ExpiresIn) * time.Second)
+		c.config.AuthInfo.AccessTokenExpiry = expiryTime.Format(time.RFC3339Nano)
+	}
+	if authInfo.IdToken != "" {
+		c.config.AuthInfo.IdToken = authInfo.IdToken
 	}
 	return c.config.Persist(c.configFilePath)
 }
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/cli/login/openshift.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate PKCE parameters: %w", err)
+	}
+
+	// Use the API server's token proxy endpoint instead of the OpenShift provider's token endpoint
+	if o.Metadata.Name == nil {
+		return nil, fmt.Errorf("provider name is required")
+	}
+	tokenProxyURL := getTokenProxyURL(o.ApiServerURL, *o.Metadata.Name)
+
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/cli/login/oauth.go
asafbennatan: to the issuer , 
any oauth2/oidc provider will fail if you try to use redirect url that is not configured on their systems this is in accordance to the OAuth2 spec :
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.3

if you are referring to inside the code base then server listen address + port should match the host part of the callbackUrl

ori-amizur: OK.
In general it is expected that this listen address will be available permanently.  Therefore, a constant port is required.  In this implementation it is available  only for a short period of time.  
The only one who can detect if this listen address is not available is the same process (the CLI) - since it controls both the client and the server.
So here it is possible:
- Not create listener at all
- Create anonymous port listener and use it instead of the redirect URL
- Use the current implementation with a conflict risk

ori-amizur:
Can you fix the e2e tests ?

asafbennatan:
yes working on it

internal/cli/login/oauth.go
ori-amizur: OK

ori-amizur (approved):

PR #2044, opened 2025-11-25
EDM-2650: Add error messages to ResourceSync status conditions

avishayt:
- Set ResourceParsed condition before returning error to ensure error message is captured
- Add Synced condition updates in all error paths of SyncFleets (fleet name conflicts, list failures, delete failures)
- Change 'out of sync' condition to use 'NewHashDetected' reason instead of 'fail' when new commit is detected
- Include actual error messages in condition messages for better troubleshooting in UI
- Do not emit failure event on NewHashDetected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and clearer diagnostic messages during resource synchronization for easier troubleshooting.
  * Updated status updates to fail earlier on sync errors, ensuring better visibility when operations abort.
* **Behavior**
  * Refined event emission so failures aren’t reported for mere hash-detection; commit-hash change events remain distinct.
  * Introduced a distinct status reason for newly detected resource-hash changes to make out-of-sync states clearer.

### DIFF
--- a/deploy/helm/flightctl/templates/flightctl-mce-role.yaml
+++ b/deploy/helm/flightctl/templates/flightctl-mce-role.yaml
@@ -1,0 +1,34 @@
+{{- if eq (include "flightctl.enableMulticlusterExtensions" .) "true" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: flightctl-device-registration
+rules:
+  - verbs:
+      - get
+    apiGroups:
+      - flightctl.io
+    resources:
+      - devices
+  - verbs:
+      - create
+    apiGroups:
+      - flightctl.io
+    resources:
+      - repositories
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: managedcluster-import-controller-flightctl
+  namespace: multicluster-engine
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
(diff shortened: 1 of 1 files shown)


internal/tasks/resourcesync.go
```diff
@@ -89,7 +89,6 @@ func (r *ResourceSync) run(ctx context.Context, log logrus.FieldLogger, orgId uu
 	resources, err := r.parseAndValidateResources(rs, repo, CloneGitRepo)
 	if err != nil {
 		log.Errorf("resourcesync/%s: parsing failed. error: %s", *rs.Metadata.Name, err.Error())
```
gshilin-sdb, line 91: nit: should we start using .WithError(err)?

internal/tasks/resourcesync.go
```diff
@@ -99,14 +98,15 @@ func (r *ResourceSync) run(ctx context.Context, log logrus.FieldLogger, orgId uu
 
 	// Parse fleets from resources
 	fleets, err := r.ParseFleetsFromResources(resources, resourceName)
+
+	// Set the ResourceParsed condition based on the result
+	api.SetStatusConditionByError(&rs.Status.Conditions, api.ConditionTypeResourceSyncResourceParsed, "success", "fail", err)
+
 	if err != nil {
 		log.Errorf("%v", err)
```
gshilin-sdb, line 106: Should we add a format string to distinguish this error log?
If yes, then use .WithError(err) pls
if not, then where was it set? SetStatusConditionByError sets condition.Message, but here it's not printed

PR #2047, opened 2025-11-25
NO-ISSUE: Add standard labels to every resource

rawagner:
## Summary by CodeRabbit

* **Chores**
  * Standardized metadata labels added across Helm templates—injecting common standard labels and service-specific labels to Kubernetes resources for improved consistency and discoverability.

* **Documentation**
  * Installer NOTES updated: login instruction no longer uses the insecure TLS-skip flag and now directs to the web-based login flow.


deploy/helm/flightctl/templates/db/_db-migration-job.tpl
```diff
@@ -21,7 +21,7 @@ metadata:
   name: {{ $name }}
   namespace: {{ default $ctx.Release.Namespace $ctx.Values.global.internalNamespace }}
   labels:
-    app: flightctl-db-migration
+    flightctl.service: flightctl-db-migration
```
celdrake, line 24: There's a `spec.template.metadata.labels` later on this file that uses `app: flightctl-db-migration`, is that correct?

deploy/helm/flightctl/templates/ui/flightctl-ui-deployment.yaml
```diff
@@ -6,9 +6,9 @@ metadata:
   name: flightctl-ui
   namespace: {{ .Release.Namespace }}
   labels:
-    app: flightctl-ui
     role: frontend
     {{- include "flightctl.standardLabels" . | nindent 4 }}
+    flightctl.service: flightctl-ui
```
celdrake, line 11: Looking at this file as an example, shouldn't we also update the `spec.selector.matchLabels` and `spec.template.metadata.labels` to be "flightctl-service: flightctl-ui"?

deploy/helm/flightctl/templates/db/_db-migration-job.tpl
celdrake: Seems like in `deploy/helm/flightctl/templates/_helpers` the matcher needs to be updated too so it matches the job.

rawagner: great catch - Updated

deploy/helm/flightctl/templates/ui/flightctl-ui-deployment.yaml
rawagner: I've updated those as well. We are not using `app` label anywhere anymore

deploy/helm/flightctl/templates/db/_db-migration-job.tpl
celdrake: Didn't seem like it was pushed yet?

rawagner: it is now :)

PR #2052, opened 2025-11-25
EDM-2632: api/device/application/status: add NoApplications

hexfusion:
## Summary by CodeRabbit

* **New Features**
  * Added a NoApplications application-health status to represent devices with zero apps.

* **Behavior**
  * Devices with no applications now report NoApplications (previously reported Healthy); some events treat NoApplications like Healthy. Default summary and event messaging updated.

* **Documentation**
  * Updated device-status docs and state diagram; clarified Healthy definition and transitions.

* **Tests**
  * Updated tests and expectations to reflect NoApplications and adjusted message text.

* **Chores**
  * Replaced embedded API/OpenAPI payload used by the system.


celdrake (approved):
I'd just suggest that we change the "info" message when the "NoApplications" status is returned.

Currently it's "Device has no application workloads defined"
I'd suggest "Device has not reported /is not running any application workloads yet".

Given applications may have been defined but are not yet reporting, or some other edge cases.

docs/user/references/device-api-statuses.md
```diff
@@ -170,7 +170,8 @@ The `device.status.applicationSummary` field can have the following values:
 
 | Status | Description | Formal Definition<sup>1</sup> |
 | ------ | ----------- | ----------------------------- |
-| `Healthy` | All applications are reported to be in service or have successfully completed. | `!deviceIsDisconnected && ∀ a∈status.applications, status.applications[a]∈{Running, Completed}` |
+| `NoApplications` | No applications are defined for the device. | `!deviceIsDisconnected && len(status.applications) == 0` |
```
celdrake, line 173: Similar to my comment earlier, it's not the same "no application are defined" which would be related to the spec, to "no applications are reported yet". Applications may be defined and there can be no reported applications for a while.

kkyrazis (approved):

PR #2064, opened 2025-11-26
NO-ISSUE: detect docker images

sk-ilya:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced image format detection to properly support both Docker and OCI image configurations, improving compatibility across different container image types.

### DIFF
--- a/deploy/helm/flightctl/templates/worker/flightctl-worker-clusterrole.yaml
+++ b/deploy/helm/flightctl/templates/worker/flightctl-worker-clusterrole.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     {{- include "flightctl.standardLabels" . | nindent 4 }}
     flightctl.service: flightctl-worker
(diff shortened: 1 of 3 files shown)


internal/agent/device/dependency/dependency.go
```diff
@@ -25,7 +25,8 @@ const (
 )
 
 const (
-	ociImageConfigMediaType = "application/vnd.oci.image.config.v1+json"
+	ociImageConfigMediaType    = "application/vnd.oci.image.config.v1+json"
+	dockerImageConfigMediaType = "application/vnd.docker.container.image.v1+json"
```
hexfusion, line 29: cc @kkyrazis

hexfusion (approved):
This feels a bit strange that we have a docker type these days. We for sure should tolerate it but it might also be worth checking if it's a bug on podman eventually.

PR #2069, opened 2025-11-26
EDM-2344: Bug fix for config.yaml overwrite when using 'make deploy'

lkatalin:
The issue (as described in [EDM-2344](https://issues.redhat.com/browse/EDM-2344)) is that manual updates to `config.yaml` are wiped when running `make deploy`. Since TPM options are not enabled for `make deploy`, it's not possible to use `make deploy` to bring up a TPM-enabled agent test VM.

Note: I used this issue as my AI Goal for Emerging Tech with Claude Code, as is noted on each commit. Let me know if there are any concerns.

## Summary by CodeRabbit

* **New Features**
  * TPM support added to agent configuration (disabled by default, opt-in).
  * Deployment now runs a TPM certificate preparation step, including generation and installation of a test TPM CA and adding certs to deployments when TPM is enabled.

* **Chores**
  * Build/deploy scripts and VM provisioning flows updated to propagate TPM state and handle installation/cleanup of test CA artifacts.

### DIFF
--- a/test/scripts/agent-images/prepare_agent_config.sh
+++ b/test/scripts/agent-images/prepare_agent_config.sh
@@ -39,3 +39,14 @@ system-info-custom:
   - siteName
   - emptyValue
 EOF
+
+# Add TPM configuration if enabled via environment variable
+if [ "${TPM:-disabled}" = "enabled" ]; then
+cat <<EOF | tee -a bin/agent/etc/flightctl/config.yaml
+tpm:
+  enabled: true
+  device-path: /dev/tpm0
+  auth-enabled: false
+  storage-file-path: /var/lib/flightctl/tpm-blob.yaml
+EOF
+fi
--- a/test/scripts/agent-images/prepare_agent_config.sh
+++ b/test/scripts/agent-images/prepare_agent_config.sh
@@ -17,14 +17,16 @@ ensure_organization_set
 
 status_update_interval=0m2s
 spec_fetch_interval=0m2s
+tpm="${TPM:-disabled}"
 # Use external getopt for long options
-options=$(getopt -o h --long status-update-interval:,spec-fetch-interval:,help -n "$0" -- "$@")
+options=$(getopt -o h --long status-update-interval:,spec-fetch-interval:,tpm:,help -n "$0" -- "$@")
 eval set -- "$options"
 while true; do
   case "$1" in
-  -h|--help) echo "Usage: $0 --status-update-interval=0m2s"; exit 1 ;;
+  -h|--help) echo "Usage: $0 --status-update-interval=0m2s --tpm=enabled"; exit 1 ;;
   --status-update-interval) status_update_interval=$2; shift 2 ;;
   --spec-fetch-interval) spec_fetch_interval=$2; shift 2 ;;
+  --tpm) tpm=$2; shift 2 ;;
   --) shift; break ;;
   *) echo "Invalid option: $1" >&2; exit 1 ;;
   esac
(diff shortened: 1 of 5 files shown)


test/scripts/agent-images/prepare_agent_config.sh
```diff
@@ -17,14 +17,16 @@ ensure_organization_set
...
 spec_fetch_interval=0m2s
+tpm="${TPM:-disabled}"
 # Use external getopt for long options
-options=$(getopt -o h --long status-update-interval:,spec-fetch-interval:,help -n "$0" -- "$@")
+options=$(getopt -o h --long status-update-interval:,spec-fetch-interval:,tpm:,help -n "$0" -- "$@")
 eval set -- "$options"
 while true; do
   case "$1" in
-  -h|--help) echo "Usage: $0 --status-update-interval=0m2s"; exit 1 ;;
+  -h|--help) echo "Usage: $0 --status-update-interval=0m2s --spec-fetch-interval=0m2s --tpm=enabled"; exit 1 ;;
```
hexfusion: nit: spec fetch interval is deprecated

lkatalin: I ~~added a commit to~~ removed it

lkatalin:
@hexfusion This is ready for re-review now. I added a commit to set up the TPM certs (including swtpm) during testing, so that the TPM verification passes during enrollment.

lkatalin:
Keep open please

test/scripts/create-test-swtpm-ca.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# This script creates a test swtpm-localca CA for use with emulated TPMs.
```
kkyrazis, line 5: I can see why this would be useful on a fresh machine that has never run a swtpm before, but do we need to generate our own certs and take backups and what not? 

My understanding is that the libvirt will run `swtpm_setup` to ensure the localca is setup when agent-vm starts. Couldn't we also just take that approach? `TPM=enabled make deploy` -> some invocation of swtpm_setup, then adds the certs to the deployment? Why do we need our own config and everything?

lkatalin: It's a good question, as this wasn't the direction I was going in intially. Without this script, the test is brittle and not reliably reproducible since the swtpm's setup is different in different environments (the root cert location, certificate parameters, etc. seem to vary).

I think I remember also a bootstrapping problem where in this testing case we don't know about the swtpm root cert until the agent is started, but by then the server is already deployed with a specific trust store (not saying it can't be added after the server is running - but this made the test setup simpler).

So it may be counterintuitive but I concluded this is actually the least messy approach. However, I don't mind changing it if you have a better idea.

kkyrazis: I guess I'm still concerned that we're effectively redoing what `swtpm_setup` or `swtpm_localca` do for us. I believe that `swtpm-tools` is already a dependency of our project so I think the bootstrapping approach could rely on those tools to set everything up and it should be relatively straightforward to find the appropriate directories that need to fetch the created certs from. From the documentation it's defined as:

```
Path to configuration file containing the tool to use for creating certificates; see also swtpm_setup.conf

If this parameter is not provided, the default configuration file will be used. The search order for the default configuration file is as follows. If the environment variable XDG_CONFIG_HOME is set, ${XDG_CONFIG_HOME}/swtpm_setup.conf will be used if available, otherwise if the environment variable HOME is set, ${HOME}/.config/swtpm_setup.conf will be used if available. If none of the previous ones are available, /etc/swtpm_setup.conf will be used.
```

lkatalin: It's a reasonable point, I'll give it another try without that setup. In previous tests the cert chain check was always failing without this, but maybe I can get it working with a fresh brain since I haven't looked at this in a bit.

lkatalin:
Please keep open, I've not gotten back to this but I plan to when time permits

lkatalin:
WIP

PR #2078, opened 2025-11-27
EDM-2676: Fix telemetry gateway secrets

rawagner:
## Summary by CodeRabbit

* **Chores**
  * Updated telemetry gateway certificate handling: switched CA bundle mount from a secret to a configMap and updated the TLS secret reference used by the deployment.

### DIFF
--- a/deploy/helm/flightctl/templates/telemetry-gateway/flightctl-telemetry-gateway-deployment.yaml
+++ b/deploy/helm/flightctl/templates/telemetry-gateway/flightctl-telemetry-gateway-deployment.yaml
@@ -62,9 +62,9 @@ spec:
           readOnly: true
           subPath: config.yaml
         - mountPath: /etc/telemetry-gateway/certs/ca.crt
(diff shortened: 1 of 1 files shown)


deploy/helm/flightctl/templates/telemetry-gateway/flightctl-telemetry-gateway-deployment.yaml
```diff
@@ -81,7 +81,7 @@ spec:
         name: flightctl-telemetry-gateway-config
       - name: ca-secret
         secret:
-          secretName: flightctl-ca-secret
+          secretName: flightctl-ca
```
fzdarsky: Use `flightctl-ca-bundle` instead, to pass in the full cert chain and without key.

rawagner: thx, changed to use the CM instead

PR #2082, opened 2025-11-27
Remove flightctl-admin from AAP auth docs

itzikb-redhat:
There is not flightctl-admin role when using AAP, Only superadmin and flightctl-org-admin

## Summary by CodeRabbit

* **Documentation**
  * Removed the "flightctl-admin" role from the list of standard roles for AAP authentication configuration; remaining roles unchanged.


eldar101 (approved):

asafbennatan:
already fixed in a different PR

PR #2086, opened 2025-11-30
EDM-2648: flightctl version is not showing the client version in the latest helm chart, replacing hack/current-version script with git describe

gshilin-sdb:
## Summary by CodeRabbit

* **Chores**
  * Unified version resolution across CI, build and packaging to use git-based computation with improved fallback behavior (ensures a sensible default when git metadata is unavailable).
  * Container image and RPM builds now derive tags from this git-based approach for more consistent versioning.
  * Removed legacy local version script in favor of the new centralized git-describe method.


.github/workflows/publish-containers.yaml
```diff
@@ -106,7 +106,7 @@ jobs:
         id: version-vars
         run: |
           set -euo pipefail
-          SOURCE_GIT_TAG=$(./hack/current-version)
+          SOURCE_GIT_TAG=$(git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown)
```
avishayt: Missing quote?

gshilin-sdb: added it

PR #2088, opened 2025-11-30
EDM-2648: flightctl version is not showing the client version in the latest helm chart (#2086)

gshilin-sdb:


avishayt (approved):

keitwb:
@gshilin-sdb The point of the current-version script that was deleted here was to deal with the issue of having multiple tags on the same commit (e.g. the last RC version before release, and the final release tag). It would properly sort them by taking into account the rc suffix and prioritize the final releases. IIRC the `git describe` command does not do this on its own.

PR #2102, opened 2025-12-01
EDM-2705: fix role mapping to exclude internal role names coming from…

asafbennatan:
… auth provider , fixed auth provider cache invalidation for org assignment / role assignment

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication now reliably detects changes to organization and role assignment configurations so provider updates occur when those settings change.
  * Internal role names are now excluded from transformed role outputs so only external/mapped roles are returned.

### DIFF
--- a/internal/auth/authn/multiauth.go
+++ b/internal/auth/authn/multiauth.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"slices"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/auth/authn/multiauth.go
```diff
@@ -512,6 +528,152 @@ func equalStringSlices(a *[]string, b *[]string) bool {
...
+		}
+		return true
+
+	default:
+		return false
+	}
+}
+
+// equalNonPointerStringSlices compares two non-pointer string slices
+func equalNonPointerStringSlices(a, b []string) bool {
```
gshilin-sdb: There is a standard library (since 1.21) which does exactly this
```go
   import "slices"
   
   slices.Equal(a, b)
```

asafbennatan: fixed

PR #2089, opened 2025-11-30
EDM-2473: offline e2e cert injection

sk-ilya:
## Summary by CodeRabbit

* **Tests**
  * Simplified end-to-end container image setup to streamline test environments.
  * Added an automated agent-file injection tool to populate test disk images before runs.
  * Updated test preparation flow to run the injection step prior to e2e execution.
  * Added tmpfiles-based configuration to ensure required directories are created at runtime.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2105, opened 2025-12-01
NO-ISSUE: Add INJECT_CONFIG parameter in make agent-vm

galel12:
## Problem

After PR #2089 (offline e2e cert injection), `make agent-vm` boots a VM without the config file, causing the agent to fail with:
`Error loading config: reading config file: open /etc/flightctl/config.yaml: no such file or directory`

## Root Cause

The refactoring in #2089 moved config injection from build-time (baked into the bootc container) to a post-build injection step via `prepare-e2e-qcow-config`. The `e2e-test` target was updated to call this step, but `agent-vm` was not, resulting in a clean qcow2 image without configuration.

## Solution

Add `INJECT_CONFIG` parameter (defaults to `true`) that makes `agent-vm` conditionally depend on `prepare-e2e-qcow-config`:

- **Default:** `make agent-vm` → config injected automatically
- **Opt-out:** `INJECT_CONFIG=false make agent-vm` → clean image for late binding scenarios

## Summary by CodeRabbit

* **Chores**
  * Added a configurable build option to make E2E configuration injection optional during the VM agent build (enabled by default).
  * Build prerequisites now adjust conditionally: when injection is enabled the build also prepares the E2E config; when disabled it skips that step, preserving prior agent build behavior otherwise.

### DIFF
--- a/.github/actions/check-admin-override/action.yaml
+++ b/.github/actions/check-admin-override/action.yaml
@@ -1,0 +1,68 @@
+name: 'Check Admin Override Label'
+description: |
+  Checks if a specified override label exists on a pull request and validates
+  that it was added by a repository admin or maintainer.
+
+inputs:
+  label-name:
+    description: 'Name of the label to check for override'
+    required: true
+
+outputs:
+  override:
+    description: 'Whether the override is active (true/false)'
+    value: ${{ steps.check_override.outputs.override }}
+
+runs:
+  using: "composite"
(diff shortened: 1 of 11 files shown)


hexfusion:
How did #2089 pass Ci I want to ensure that does not pass in the future. cc @siserafin 

tracked https://issues.redhat.com/browse/EDM-2708

PR #2113, opened 2025-12-01
EDM-2159: Update flightctl-must-gather to exclude all confidential information

SiddarthR56:
## Summary by CodeRabbit

* **New Features**
  * Added --include-secrets flag and help; interactive prompts for storage choice and confirmation before collection.
  * Initial user-facing warning about potentially large output sizes.

* **Improvements**
  * Default redaction of sensitive fields with optional inclusion of secrets.
  * Collects diagnostics into a temporary workspace with automatic cleanup.
  * Expanded collection (journals, audit logs, spec files, system info) with user-visible warnings and continued operation on partial failures.

* **Chores**
  * Added jq runtime dependency.

### DIFF
--- a/packaging/must-gather/flightctl-must-gather
+++ b/packaging/must-gather/flightctl-must-gather
@@ -1,6 +1,25 @@
 #!/usr/bin/env bash
 
-set -x
+set -euo pipefail
+
+INCLUDE_SECRETS=false
+
+usage() {
+  cat <<'EOF'
+Usage: flightctl-must-gather [--include-secrets] [-h]
+
+  --include-secrets  Include sensitive files (specs unredacted, certs/keys, TPM blobs)
+  -h, --help         Show this help message and exit
+EOF
+}
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --include-secrets) INCLUDE_SECRETS=true; shift ;;
+    -h|--help) usage; exit 0 ;;
+    *) echo "Unknown argument: $1" >&2; usage; exit 1 ;;
+  esac
+done
 
 # Check if the script is run as root
 if [ "$EUID" -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


fzdarsky:
> Enabling `set -euo pipefail` and adding a small CLI parser for `--include-secrets` / `-h` is clean and straightforward. One side effect is that any non-zero exit from later commands (e.g., `journalctl`, `flightctl-agent version`, `bootc --version/status`) will now abort the entire must-gather instead of producing a partial archive. If you want to preserve “best-effort” behavior, it may be worth selectively adding `|| true` (or explicit error messaging) around non-critical probes while keeping strict mode for the rest.

Could you have another look whether some probes can fail but be optional, please? Better to have partial results than not being able to run.

hexfusion:
@SiddarthR56 to reiterate Franks concern here. This will exit early on some basic commands if they fail. please ensure `|| true` pattern is used consistently. We want the script to be resilient and give us what it can vs exit early.

packaging/must-gather/flightctl-must-gather
```diff
@@ -18,43 +37,125 @@ if [[ ! $REPLY =~ ^[Yy]$ ]]; then
...
+journalctl --since "24 hours ago" > "$temp_dir/journal.log" || {
+  echo "Warning: failed to collect journal logs; continuing." >&2
+  true
+}
 
-echo "### Flightctl Spec Files..."
-cp /var/lib/flightctl/*.json "$temp_dir/" 2>/dev/null || true
+collect_spec_files
 
 echo "### Flightctl Audit Log..."
```
fzdarsky, line 119: Could you please also add the missing `> "$output_file"` while you're at it?

packaging/must-gather/flightctl-must-gather
```diff
@@ -18,43 +37,127 @@ if [[ ! $REPLY =~ ^[Yy]$ ]]; then
...
 echo "### Collecting journal logs..."
-journalctl --since "24 hours ago" > "$temp_dir/journal.log"
+journalctl --since "24 hours ago" > "$temp_dir/journal.log" || {
+  echo "Warning: failed to collect journal logs; continuing." >&2
+  true
+}
 
-echo "### Flightctl Spec Files..."
-cp /var/lib/flightctl/*.json "$temp_dir/" 2>/dev/null || true
+collect_spec_files
```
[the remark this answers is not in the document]
SiddarthR56: collect_spec_files already handles its own errors. Each redact_json_file call is wrapped in an if ! …; then block that logs a warning and carries on, so the function never bubbles up a failure. Because of that, the plain collect_spec_files call is safe even with set -e

packaging/rpm/flightctl.spec
```diff
@@ -51,6 +51,7 @@ flightctl is the CLI for controlling the Flight Control service.
 Summary: Flight Control management agent
 
 Requires: flightctl-selinux = %{version}
+Requires: jq
```
SiddarthR56, line 54: cc @kkyrazis @gshilin-sdb

PR #2114, opened 2025-12-01
EDM-2681/2696 Quadlet applications management fixes

kkyrazis:
Fixes a few issues with quadlets

1. removes creating dropins for .image quadlets. Image quadlets are primarily used for customization of pulling images and don't support a Label key.
2. Fixes issue with idempotentcy of Quadlet artifact volumes. If the volume exists, its contents will be updated rather than erroring out for failing to create a new volume. (occurs after restart as an example).
3. Lifecycle Action Handlers now are called with all of their actions at once. This allows individual handlers to better group operations so that system wide operations can be coordinated carefully. 

## Summary by CodeRabbit

* **Refactor**
  * Lifecycle now processes multiple actions as batched sequences per app type, enabling coordinated add/remove/update ordering and centralized verification.
* **Bug Fixes / Reliability**
  * Improved sequencing, rollback and cleanup for failing add/remove/update flows; clearer logging when services or artifacts are missing.
* **Tests**
  * Expanded multi-action orchestration tests covering mixed action streams and error/cleanup paths.
* **Chores**
  * Skip generation of image-specific install drop-in.


internal/agent/device/applications/provider/quadlet.go
```diff
@@ -264,6 +264,12 @@ func namespaceDropInDirectory(readWriter fileio.ReadWriter, dirPath, appID strin
 // createQuadletDropIn creates a drop-in override directory and configuration file
 // for a specific quadlet type. It adds the project label and optionally the EnvironmentFile parameter.
 func createQuadletDropIn(readWriter fileio.ReadWriter, dirPath, appID, extension string, hasEnvFile bool) error {
+	// .image quadlets primarily allow for customization of pulling images. No, drop-ins are required
+	// for them
+	if extension == quadlet.ImageExtension {
+		return nil
+	}
```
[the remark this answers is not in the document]
kkyrazis: excellent usage of compute.

internal/agent/device/applications/lifecycle/lifecycle.go
```diff
@@ -21,8 +21,19 @@ const (
...
+
+// NilActionContinuation is a no-op continuation for actions that don't need deferred work.
+func NilActionContinuation(_ context.Context) error { return nil }
+
 type ActionHandler interface {
-	Execute(ctx context.Context, action *Action) error
+	// Execute performs the initial phase of an action and returns a continuation
+	// for work that should be deferred until after all initial phases complete.
+	// The continuation will be called after all Execute calls have finished.
+	Execute(ctx context.Context, action *Action) (ActionContinuation, error)
```
hexfusion: I wonder if we could simplify if we make Execute variadic?
```suggestion
	Execute(ctx context.Context, actions ...*Action) error
```

```go
// compose only supports single actions
if err := compose.Execute(ctx, action)..

// quadlet
if err := quadlet.Execute(ctx, actions...)..
```

Just an idea on how to encapsulate

```go
func (q *Quadlet) Execute(ctx context.Context, actions ...*Action) error {
	if len(actions) == 0 {
		return nil
	}

	var toStop, toCleanup, toStart []*Action
	for _, action := range actions {
		switch action.Type {
		case ActionRemove:
			if q.actionCache.hasAction(action.ID) {
				toStop = append(toStop, action)
				toCleanup = append(toCleanup, action)
			}
            [...]
		}
	}

	// stop
	for _, a := range toStop {
		if err := q.stopServices(ctx, a); err != nil {
			return err
		}
	}

	// reload
	[...]

	// cleanup
	[...]

	// start
    [...]

	return nil
}
```

kkyrazis: Sure, I can take this approach. The biggest downsides I see are the need to group all of the actions based on type before calling execute (not a huge deal) and the fact that we'll need to ensure that application types don't affect other application types. I.E. executing all of the quadlet actions first affects all of the systemd applications that haven't executed. However, as that isn't a real problem yet we can address that problem later.

hexfusion: yep just thinking of alternatives I have not dug that deeply. if the proposal causes more issues in the implementation than it solves we can revist later.

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -103,45 +103,59 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
...
+		return fmt.Errorf("listing loaded units: %w", err)
+	}
+
+	unitSet := make(map[string]struct{})
+	for _, u := range units {
+		unitSet[u.Unit] = struct{}{}
+	}
+
+	for _, service := range services {
+		if _, ok := unitSet[service]; !ok && filepath.Ext(service) != ".target" {
```
hexfusion: optional but this is getting a little hairy to read :)

```go
if isLoadedTarget(unitSet, service) {
```

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -256,28 +266,48 @@ func (q *Quadlet) remove(ctx context.Context, action *Action) error {
...
-		return q.update(ctx, action)
-	default:
-		return fmt.Errorf("unsupported action type: %s", action.Type)
+	for _, afterReload := range afterReloadFns {
+		if err := afterReload(ctx); err != nil {
+			return err
+		}
 	}
+	return nil
 }
```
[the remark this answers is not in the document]
kkyrazis: Is this an actual issue now? Or was this behavior changed in a recent version of golang?

hexfusion (approved):

PR #2130, opened 2025-12-02
Revert "EDM-2648: flightctl version is not showing the client version in the latest helm chart, replacing hack/current-version script with git describe"

keitwb:
Reverts flightctl/flightctl#2086.

We need to find a different way to solve this. The `current-version` script deals with the issue we ran into last release where multiple tags point to the same commit (e.g. the final rc release and the main release itself). That script has awareness of the rc suffix and prioritizes the main releases. If there is a problem with it we should fix that script but we still need the functionality it provides.

## Summary by CodeRabbit

* **Chores**
  * Centralized version resolution: replaced multiple git-based lookups with a single CI-friendly version script used by build, test, and publish workflows.
  * Added a helper script to reliably determine the repository version in CI, improving consistency and error handling for image/helm tagging and release-related tooling.

### DIFF
--- a/.github/actions/compute-tag/action.yaml
+++ b/.github/actions/compute-tag/action.yaml
@@ -28,7 +28,7 @@ runs:
       shell: bash
       run: |
         set -euo pipefail
-        TAG=$(git describe --tags --exclude latest 2>/dev/null || echo "v0.0.0-unknown")
+        TAG=$(./hack/current-version)
         echo "tag=$TAG" >> "$GITHUB_OUTPUT"
         echo "Computed tag: $TAG"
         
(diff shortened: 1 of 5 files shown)


hack/current-version
```diff
@@ -0,0 +1,21 @@
...
+git config --global --add safe.directory /__w/flightctl/flightctl 2>/dev/null || true
+git config --global --add safe.directory "$(pwd)" 2>/dev/null || true
+
+# Prefer proper release tags pointing at this exact commit first, prioritizing the highest version
+# tag. If there is no tag at this commit, then fallback to 'git describe' which will construct a
+# "human-readable" artificial tag.
+tag=$(git -c "versionsort.suffix=-rc" tag --points-at HEAD --sort version:refname | tail -n 1 2>/dev/null)
+
+if [[ -z "$tag" ]]; then
+  tag=$(git describe --tags --always --first-parent --exclude latest 2>/dev/null)
```
fzdarsky: Would it make sense to still only use the last version tag as base instead of any tag that's not "latest"?

```suggestion
  tag=$(git describe --tags --always --first-parent --match "v*" 2>/dev/null)
```

keitwb: Ok, both this and the line above where HEAD points to the same commit as a tag will only consider `v*` tags.

fzdarsky (approved):

PR #2131, opened 2025-12-03
NO-ISSUE: api: add volume reclaim policy

hexfusion:
The agent currently retains volumes which can be costly, this PR adds an explicit retention policy based on [persistentVolumeReclaimPolicy](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-v1/#PersistentVolumeSpec)

```yaml
  spec:
    applications:
      - name: my-app
        image: quay.io/hexfusion/app:latest
        volumes:
          - name: data-volume
            reclaimPolicy: Retain    # keep volume after app removal (default)
            image:
              reference: quay.io/hexfusion/data:v1
          - name: cache-volume
            reclaimPolicy: Delete    # remove volume when app is removed  
            mount:
              path: /cache
```

## Release Notes

* **New Features**
  * Added volume reclaim policy configuration for applications. Users can now specify whether volume data persists or is automatically deleted when an application is removed. Options: Retain (default) or Delete.


celdrake:
@hexfusion Is this meant to be included in 1.0?
And should there be a story to track this and handle it from the UI too? 
Currently since the UI would not know about this field, it could delete these new fields when added via the API/CLI.

hexfusion:
@celdrake this is meant to be a talking point for today's arch call there is no intention of merging support without UI ack but we feel even without UI integration the default would be a net positive in terms of documentation.

PR #2140, opened 2025-12-03
NO-ISSUE: Fix packit version selection

fzdarsky:
Fixes problem that the way the current version is obtained in  the `.packit.yaml` may pick tags from a different branch than the current and so get the wrong version.


fzdarsky:
Obsoleted by PR #2130.

PR #2161, opened 2025-12-04
EDM-2739: added validation for oauth2/oidc with static role assignment - only known external roles are allowed

asafbennatan:
added validation for oauth2/oidc with static role assignment - only known external roles are allowed
EDM-2606: additionally redacting client secrets from config log

## Summary by CodeRabbit

* **New Features**
  * Introduced a canonical list of known external roles.

* **Bug Fixes**
  * Tightened role validation for external (OIDC/OAuth2) assignments to reject unknown roles.
  * Configuration logging now redacts authentication client secrets and does not mutate the live config.

* **Tests**
  * Added and expanded tests covering role validation and redaction/handling of authentication secrets; updated tests to use the canonical role constant.

### DIFF
--- a/api/v1beta1/validation.go
+++ b/api/v1beta1/validation.go
@@ -1735,6 +1735,9 @@ func (a AuthStaticRoleAssignment) Validate(ctx context.Context) []error {
 		if role == "" {
 			allErrs = append(allErrs, fmt.Errorf("role at index %d cannot be empty", i))
 		}
+		if !slices.Contains(KnownExternalRoles, role) {
+			allErrs = append(allErrs, fmt.Errorf("role at index %d is not a valid role: %s", i, role))
+		}
 	}
 
 	return allErrs
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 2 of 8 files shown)


internal/config/config.go
```diff
@@ -710,13 +711,104 @@ func Validate(cfg *Config) error {
...
+	sanitized := cfg.sanitizeForLogging()
+	contents, err := json.Marshal(sanitized)
 	if err != nil {
 		return "<error>"
 	}
 	return string(contents)
 }
+
+// sanitizeForLogging creates a copy of the config with sensitive fields redacted
+func (cfg *Config) sanitizeForLogging() *Config {
```
sk-ilya, line 773: Can't `SecureString` solve this?

internal/config/config.go
```diff
@@ -710,13 +711,104 @@ func Validate(cfg *Config) error {
 			return fmt.Errorf("readinessPath and livenessPath must not be identical")
 		}
 	}
+
+	// Validate OIDC and OAuth2 provider role assignments
+	if cfg.Auth != nil {
+		if cfg.Auth.OIDC != nil {
+			if err := validateAuthProviderRoleAssignment(cfg.Auth.OIDC.RoleAssignment, "oidc"); err != nil {
```
sk-ilya: Nit: there are constants defined for "oidc" and "oauth2" - can we use them? 
See https://github.com/flightctl/flightctl/blob/main/internal/config/standalone/config.go#L4-L7 
Or https://github.com/flightctl/flightctl/blob/main/api/v1beta1/types.gen.go#L343-L351

internal/config/config.go
```diff
@@ -710,13 +711,104 @@ func Validate(cfg *Config) error {
...
+
+	staticAssignment, err := roleAssignment.AsAuthStaticRoleAssignment()
+	if err != nil {
+		return fmt.Errorf("%s provider: invalid static role assignment: %w", providerType, err)
+	}
+
+	// Validate that all roles are in KnownExternalRoles
+	for i, role := range staticAssignment.Roles {
+		if role == "" {
+			return fmt.Errorf("%s provider: role at index %d cannot be empty", providerType, i)
```
sk-ilya, line 752: Nit: As this would become a user-facing error, is "at index %d" actionable from the user perspective?

api/v1beta1/validation.go
```diff
@@ -1453,7 +1453,7 @@ func (o *OAuth2ProviderSpec) Validate(ctx context.Context, isUpdate bool) []erro
 	allErrs = append(allErrs, o.OrganizationAssignment.Validate(ctx)...)
 
 	// Validate role assignment
-	allErrs = append(allErrs, o.RoleAssignment.Validate(ctx)...)
+	allErrs = append(allErrs, o.RoleAssignment.Validate(ctx, true)...)
```
sk-ilya: If `RoleAssignment.Validate` is always called with `knownRolesOnly=true`, then perhaps this flag can be removed?

internal/config/config.go
asafbennatan: no since we save the save in the DB - secure string doesnt work properly

internal/config/config.go
asafbennatan: they would fix the role at index %d to be non empty / part of the known external roles

api/v1beta1/validation.go
asafbennatan: fixed

sk-ilya (approved):

PR #2171, opened 2025-12-05
NO-ISSUE: Consolidate api org setting middleware

DakCrowder:
We have a `orgID == uuid.Nil` conditoinal in the current api middleware.  This conditional evaluates to true when orgID is the default org ID since they are both "00000000-0000-0000-0000-000000000000", which can result in a situation where:

- A user passes the default org id as the org_id param
- We then might attempt to set the org id in context from the .GetOrganizations() call, which _could_ be a **different** organization than the default.

When taking a pass through the middleware it became simpler to just refactor the two extract and validate middlewares into one (they were always coupled).

Initial conditional problem raised by @ori-amizur a couple of days ago

## Summary by CodeRabbit

* **New Features**
  * Better automatic organization detection from certificates/identity when org is omitted.
  * Token exchange no longer requires providing an explicit org ID.

* **Bug Fixes**
  * More precise HTTP statuses and error responses for missing, ambiguous, invalid, or non-member organization cases.
  * Improved validation to reject conflicting or malformed organization identifiers.

* **Refactor**
  * Organization extraction and validation combined into a single middleware for simpler request flow.

### DIFF
--- a/internal/api_server/middleware/middleware.go
+++ b/internal/api_server/middleware/middleware.go
@@ -10,6 +10,8 @@ import (
 	"github.com/flightctl/flightctl/internal/consts"
 	"github.com/flightctl/flightctl/internal/contextutil"
 	"github.com/flightctl/flightctl/internal/crypto/signer"
+	"github.com/flightctl/flightctl/internal/flterrors"
+	"github.com/flightctl/flightctl/internal/identity"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/flightctl/flightctl/pkg/reqid"
@@ -77,7 +79,9 @@ func UserAgentLogger(logger logrus.FieldLogger) func(http.Handler) http.Handler 
 }
 
 // OrgIDExtractor extracts an organization ID from an HTTP request.
-type OrgIDExtractor func(context.Context, *http.Request) (uuid.UUID, error)
+// Returns (orgID, present, error) where present is true if the org ID was
+// explicitly specified in the request.
+type OrgIDExtractor func(context.Context, *http.Request) (uuid.UUID, bool, error)
 
 // QueryOrgIDExtractor is the default extractor that reads the org_id from the query string.
 var QueryOrgIDExtractor OrgIDExtractor = extractOrgIDFromRequestQuery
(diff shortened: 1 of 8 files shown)


internal/api_server/middleware/middleware.go
```diff
@@ -93,20 +97,29 @@ func ExtractOrgIDToCtx(extractor OrgIDExtractor, logger logrus.FieldLogger) func
...
-				if ok && len(mappedIdentity.GetOrganizations()) > 0 {
-					orgID = mappedIdentity.GetOrganizations()[0].ID
-					reqLogger.Debugf("ExtractOrgIDToCtx: extracted orgID=%s from mapped identity", orgID.String())
+				if ok {
+					switch len(mappedIdentity.GetOrganizations()) {
+					case 1:
+						orgID = mappedIdentity.GetOrganizations()[0].ID
+						reqLogger.Debugf("ExtractOrgIDToCtx: extracted orgID=%s from mapped identity", orgID.String())
+					case 0:
+						orgID = org.DefaultID
```
asafbennatan: @DakCrowder lets discuss this - this was not the intended behavior when i implemented , 
users should not be allowed to use flightctl if they dont have an organization

the original default org (as in org we auto give users access to )  doesn't makes sense any more since auth providers have the capability of defining their default , if they did - users will already arrive here with that org
(static organizationAssignment)

asafbennatan (changes requested):

internal/api_server/middleware/middleware.go
DakCrowder: Ok makes sense, can remove this case / fallback

internal/service/auth_token_proxy.go
```diff
@@ -77,7 +76,7 @@ func NewAuthTokenProxy(authN *authn.MultiAuth) *AuthTokenProxy {
 
 // ProxyTokenRequest handles OAuth2 token exchange requests
 // Returns TokenResponse and HTTP status code (200 for success, 400 for errors per OAuth2 spec)
-func (p *AuthTokenProxy) ProxyTokenRequest(ctx context.Context, orgId uuid.UUID, providerName string, tokenReq *api.TokenRequest) (*api.TokenResponse, int) {
```
DakCrowder, line 80: Note: orgId param was unused here

asafbennatan: this is fine

asafbennatan: i mean the parameter can be removed ,we do not need to special handle org here

PR #2172, opened 2025-12-05
EDM-2696: Fix embedded quadlet app reconciliation

kkyrazis:
Embedded compose applications exist only within the image that supplies them. Therefore, whenever a new OS image is deployed and application reconciliation occurs, enqueuing an `Add` to the application manager to effectively `podman compose up -d` is always appropriate. If updating to a new image, the old embedded application won't be added (as it doesn't exist) and therefore won't start. With quadlet's this is not the case.

Quadlets must be namespaced and as such, require moving the embedded quadlets into `/etc/containers/systemd/<app-name>`  This means that lifecycle of embedded applications can exist outside the scope of their image and must support natural `add`, `remove`, or `updates`.

To solve these issues the following was done:

* Instead of requiring the `lifecycle` handler to discover which applications to start/stop (by reading the installation directory), the act of installing a quadlet now generates an `<app-id>-flightctl.target` unit and marks all quadlets as PartOf the target.
* The `lifecycle` handler now calls `systemctl start <app-id>-flightctl.target`
* The `lifecycle` handler now calls `systemctl stop <app-id>-flightctl.target` (unfortunately this isn't synchronous)
*  The `lifecycle` handler now calls `systemctl list-dependencies <app-id>-flightctl.target` to determine the current state of the application and stops all of those services.
*  Embedded quadlet applications are installed with a marker file (`.flightctl-embedded`) containing the system's boot time in their application directory.
* When applications are synced, the `current` spec now includes searching for installed quadlets. (desired still scans for requested quadlets) Embedded quadlets have an unexported `boottime` added to their spec so that proper spec diffing can occur. If an embedded quadlet is installed and it's no longer in the image, it will be removed naturally. If the installed quadlet has a different installed boottime from current, an update will occur. In the steady state, embedded boottimes will match and `ensure` will be called. 

## Summary by CodeRabbit

* **New Features**
  * Discover and manage pre-installed embedded apps and their boot-time metadata.
  * Resolve and operate app dependencies via consolidated target units for more reliable lifecycle operations.
  * Expose dependency listing for systemd units to improve orchestration.

* **Refactor**
  * Thread device boot time through provisioning and embedded app handling.
  * Centralized installer flow and added structured logging for embedded/quadlet installs.

* **Bug Fixes**
  * Improved removal, cleanup and error propagation for app add/remove/update flows.

### DIFF
--- a/internal/agent/client/systemd.go
+++ b/internal/agent/client/systemd.go
@@ -190,6 +190,30 @@ func (s *Systemd) ShowByMatchPattern(ctx context.Context, matchPatterns []string
 	return units, nil
 }
 
+// ListDependencies returns the list of units that the specified unit depends on.
+// Uses `systemctl list-dependencies --plain` to get a flat list of dependencies.
+func (s *Systemd) ListDependencies(ctx context.Context, unit string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, defaultSystemctlTimeout)
+	defer cancel()
+
+	args := []string{"list-dependencies", "--plain", "--no-pager", unit}
+	stdout, stderr, exitCode := s.exec.ExecuteWithContext(ctx, systemctlCommand, args...)
+	if exitCode != 0 {
+		return nil, fmt.Errorf("list-dependencies for %s: %w", unit, errors.FromStderr(stderr, exitCode))
+	}
+
+	var deps []string
+	lines := strings.Split(strings.TrimSpace(stdout), "\n")
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		if line == "" || line == unit {
+			continue
+		}
+		deps = append(deps, line)
+	}
+	return deps, nil
+}
+
 // SystemdJob represents a systemd job from list-jobs
 type SystemdJob struct {
 	Job     string
--- a/internal/agent/device/applications/controller.go
+++ b/internal/agent/device/applications/controller.go
@@ -16,6 +16,7 @@ type Controller struct {
 	readWriter fileio.ReadWriter
 	manager    Manager
 	log        *log.PrefixLogger
+	bootTime   string
 }
 
 func NewController(
@@ -23,12 +24,14 @@ func NewController(
 	manager Manager,
 	readWriter fileio.ReadWriter,
 	log *log.PrefixLogger,
+	bootTime string,
 ) *Controller {
 	return &Controller{
 		log:        log,
 		manager:    manager,
 		podman:     podman,
 		readWriter: readWriter,
+		bootTime:   bootTime,
 	}
 }
 
@@ -42,6 +45,7 @@ func (c *Controller) Sync(ctx context.Context, current, desired *v1beta1.DeviceS
 		c.podman,
 		c.readWriter,
 		current,
+		provider.WithInstalledEmbedded(),
 	)
 	if err != nil {
 		return fmt.Errorf("current app providers: %w", err)
@@ -53,7 +57,7 @@ func (c *Controller) Sync(ctx context.Context, current, desired *v1beta1.DeviceS
 		c.podman,
 		c.readWriter,
 		desired,
-		provider.WithEmbedded(),
+		provider.WithEmbedded(c.bootTime),
 	)
 	if err != nil {
 		return fmt.Errorf("desired app providers: %w", err)
--- a/internal/agent/device/applications/lifecycle/quadlet.go
+++ b/internal/agent/device/applications/lifecycle/quadlet.go
@@ -19,6 +19,8 @@ import (
 const (
 	QuadletAppPath         = "/etc/containers/systemd"
 	EmbeddedQuadletAppPath = "/usr/local/etc/containers/systemd"
+	QuadletTargetPath      = "/etc/systemd/system/"
+	QuadletTargetName      = "flightctl-quadlet-app.target"
 )
 
 var _ ActionHandler = (*Quadlet)(nil)
@@ -32,7 +34,6 @@ type Quadlet struct {
 }
 
 type actionCacheEntry struct {
-	services        []string
 	artifactVolumes []string
 }
 
@@ -44,28 +45,6 @@ func newActionCache() *actionCache {
 	return &actionCache{
 		cache: make(map[string]*actionCacheEntry),
 	}
-}
-
-func (a *actionCache) hasAction(actionID string) bool {
-	_, ok := a.cache[actionID]
-	return ok
-}
-
-func (a *actionCache) addServices(actionID string, services []string) {
-	entry, ok := a.cache[actionID]
-	if !ok {
-		entry = &actionCacheEntry{}
-		a.cache[actionID] = entry
-	}
-	entry.services = append(entry.services, services...)
-}
-
-func (a *actionCache) services(actionID string) []string {
-	entry, ok := a.cache[actionID]
-	if !ok {
-		return nil
-	}
-	return entry.services
 }
 
 func (a *actionCache) clearAction(actionID string) {
@@ -100,12 +79,26 @@ func NewQuadlet(log *log.PrefixLogger, rw fileio.ReadWriter, systemdManager syst
 }
 
 func isServiceLoaded(unitSet map[string]struct{}, service string) bool {
-	// report targets as loaded
 	if filepath.Ext(service) == ".target" {
 		return true
 	}
 	_, exists := unitSet[service]
 	return exists
+}
+
+func (q *Quadlet) loadedUnits(ctx context.Context, services []string) (map[string]struct{}, error) {
+	units, err := q.systemdManager.ListUnitsByMatchPattern(ctx, services)
+	if err != nil {
+		return nil, fmt.Errorf("listing loaded units: %w", err)
+	}
+
+	unitSet := make(map[string]struct{}, len(units))
+	for _, u := range units {
+		if u.LoadState == string(v1beta1.SystemdLoadStateLoaded) {
+			unitSet[u.Unit] = struct{}{}
+		}
+	}
+	return unitSet, nil
 }
 
 func (q *Quadlet) add(ctx context.Context, action *Action) error {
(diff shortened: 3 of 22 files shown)


internal/agent/device/applications/controller.go
```diff
@@ -53,7 +57,7 @@ func (c *Controller) Sync(ctx context.Context, current, desired *v1beta1.DeviceS
 		c.podman,
 		c.readWriter,
 		desired,
-		provider.WithEmbedded(),
+		provider.WithEmbedded(c.bootTime),
```
hexfusion, line 60: note: one problem with embedded  in the controller is rollback never seem to be appropriate because technically they are not part of the spec. while this was my attempt to think smart, perhaps I should of thought less :)

kkyrazis: Will think through this a bit more. Embedded providers likely shouldn't also stop upgrades for failing to fetch their embedded images.

hexfusion: fair should all embedded images be LBI[1] to circumvent this nonsense.

also when I look again I wonder if we are parsing spec beforeUpdate do we need a controller at all

[1] https://bootc-dev.github.io/bootc/logically-bound-images.html

hexfusion: actually lets track that lets remind each other in 1:1

internal/agent/client/systemd.go
```diff
@@ -190,6 +190,30 @@ func (s *Systemd) ShowByMatchPattern(ctx context.Context, matchPatterns []string
 	return units, nil
 }
 
+// ListDependencies returns the list of units that the specified unit depends on.
+// Uses `systemctl list-dependencies --plain` to get a flat list of dependencies.
+func (s *Systemd) ListDependencies(ctx context.Context, unit string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, defaultSystemctlTimeout)
+	defer cancel()
+
+	args := []string{"list-dependencies", "--plain", "--no-pager", unit}
```
hexfusion, line 199: no json?

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -19,6 +19,8 @@ import (
 const (
 	QuadletAppPath         = "/etc/containers/systemd"
 	EmbeddedQuadletAppPath = "/usr/local/etc/containers/systemd"
+	QuadletTargetPath      = "/etc/systemd/system/"
+	QuadletTargetName      = "flightctl.target"
```
hexfusion: isnt this the same as the service can we be flightctl-agent.target or maybe I am missing something

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -157,65 +129,77 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
...
-					err = fmt.Errorf("service: %q logs: %s: %w", service, strings.Join(serviceLogs, ","), err)
-				}
+	q.log.Debugf("Starting quadlet: %s target: %s", appName, target)
+	if err := q.systemdManager.Start(ctx, target); err != nil {
+		err = fmt.Errorf("starting target %s: %w", target, err)
+		for _, service := range services {
+			serviceLogs, serviceErr := q.systemdManager.Logs(ctx, client.WithLogUnit(service), client.WithLogSince(startTime))
+			if serviceErr != nil {
+				q.log.Errorf("Failed to gather service: %q logs: %v", service, serviceErr)
+				continue
```
hexfusion, line 162: why not append this error

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -157,65 +129,77 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
...
-			return fmt.Errorf("stopping units: %w", err)
+			return fmt.Errorf("stopping services: %w", err)
 		}
 		failedServices := q.getFailedServices(ctx, services)
 		if len(failedServices) > 0 {
-			q.log.Debugf("Resetting failed state for services: %q", strings.Join(failedServices, ","))
+			q.log.Debugf("Resetting failed state for services: %q", strings.Join(failedServices, ", "))
 			if resetErr := q.systemdManager.ResetFailed(ctx, failedServices...); resetErr != nil {
-				q.log.Warnf("Failed to reset-failed for services %q: %v", strings.Join(failedServices, ","), resetErr)
+				q.log.Warnf("Failed to reset-failed for services %q: %v", strings.Join(failedServices, ", "), resetErr)
```
hexfusion: confirming this is not Error

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -456,3 +365,7 @@ func (q *Quadlet) ensureArtifactVolumes(ctx context.Context, action *Action) err
 
 	return nil
 }
+
+func targetName(appID string) string {
```
hexfusion: do we care if id is empty

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -157,65 +129,77 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
...
+	if err := q.systemdManager.Stop(ctx, target); err != nil {
+		return fmt.Errorf("stopping target %s: %w", target, err)
 	}
 
-	services := q.actionCache.services(action.ID)
 	if len(services) > 0 {
-		q.log.Debugf("Stopping quadlet: %s services: %q", appName, strings.Join(services, ","))
+		// stop and wait for all services to finish
+		q.log.Debugf("Stopping quadlet: %s services: %s", appName, strings.Join(services, ", "))
 		if err := q.systemdManager.Stop(ctx, services...); err != nil {
```
DakCrowder: Are these calls synchronous?  So stopping the target kicks off stops for the service but we verify each with a synchronous stop to each service?

internal/agent/device/applications/lifecycle/quadlet.go
kkyrazis: It all gets namespaced anyway, but I'll make the name more explicit.

internal/agent/client/systemd.go
kkyrazis: On my laptop and  test vm `--output json` outputs the same as `--plain` which, interestingly enough, doesn't appear to be valid json.

```text
sudo systemctl list-dependencies 
default.target
● ├─accounts-daemon.service
● ├─gdm.service
● ├─rtkit-daemon.service
● ├─switcheroo-control.service
● ├─tuned-ppd.service
● ├─udisks2.service
● ├─upower.service
● └─multi-user.target
●   ├─abrt-journal-core.service
●   ├─abrt-oops.service
```

```text
sudo systemctl list-dependencies --plain
default.target
  accounts-daemon.service
  gdm.service
  rtkit-daemon.service
  switcheroo-control.service
  tuned-ppd.service
  udisks2.service
  upower.service
  multi-user.target
  abrt-journal-core.service

```

```text
sudo systemctl list-dependencies --output json
default.target
  accounts-daemon.service
  gdm.service
  rtkit-daemon.service
  switcheroo-control.service
  tuned-ppd.service
  udisks2.service
  upower.service
  multi-user.target
  abrt-journal-core.service
```

internal/agent/device/applications/lifecycle/quadlet.go
DakCrowder: To answer my own question: yes.  Was trying to consider if this could cause any additional race issues but after reading some more docs none come to mind, could just be a bit more noisy on the systemd side.  But the alternative would be polling the target or services and this seems more straightforward.

internal/agent/client/systemd.go
hexfusion: lol sigh

internal/agent/device/applications/lifecycle/quadlet.go
kkyrazis: Do you have a target file setup that allows both synchronous start/stop? I experimented very briefly, but decided that the way that you had setup the service quadlets was good enough, even though it requires this extra call to stop as a sync point.

DakCrowder: I don't think its possible, what you did make sense sorry to cause any confusion ha

internal/agent/device/applications/provider/quadlet.go
```diff
@@ -303,6 +324,76 @@ func createQuadletDropIn(readWriter fileio.ReadWriter, dirPath, appID, extension
...
+	case quadlet.PodExtension:
+		return fmt.Sprintf("%s-pod.service", basename)
+	case quadlet.VolumeExtension:
+		return fmt.Sprintf("%s-volume.service", basename)
+	case quadlet.NetworkExtension:
+		return fmt.Sprintf("%s-network.service", basename)
+	case quadlet.ImageExtension:
+		return fmt.Sprintf("%s-image.service", basename)
+	default:
+		return ""
```
hexfusion: this isnt an error?

internal/agent/device/applications/provider/quadlet.go
```diff
@@ -303,6 +324,76 @@ func createQuadletDropIn(readWriter fileio.ReadWriter, dirPath, appID, extension
...
+		return "", fmt.Errorf("reading file: %w", err)
+	}
+	unit, err := quadlet.NewUnit(contents)
+	if err != nil {
+		return "", fmt.Errorf("parsing quadlet: %w", err)
+	}
+	name, err := unit.Lookup(quadlet.PodGroup, quadlet.ServiceNameKey)
+	if err != nil {
+		if errors.Is(err, quadlet.ErrKeyNotFound) {
+			return defaultName, nil
```
hexfusion, line 391: walk me through this

hexfusion: should we log?

internal/agent/device/applications/provider/quadlet.go
```diff
@@ -128,18 +141,22 @@ func installQuadlet(readWriter fileio.ReadWriter, path string, appID string) err
...
+		return fmt.Errorf("creating flightctl target: %w", err)
+	}
+
 	entries, err = readWriter.ReadDir(path)
 	if err != nil {
 		return fmt.Errorf("re-reading directory: %w", err)
 	}
 
-	// 2. Update any required references in quadlet files or in drop-in .conf files
+	// 3. Update any required references in quadlet files or in drop-in .conf files
```
hexfusion: do we need to count, math is hard?

kkyrazis: Siri is a genius counter

internal/agent/device/applications/provider/quadlet.go
kkyrazis: It is. It's just being handled as an error somewhere else. updating this to explicitly return the error.

internal/agent/device/applications/provider/quadlet.go
kkyrazis: will add comments, but this is just checking a Pod file for a `ServiceName` override and returning that service name if it exists

hexfusion: comment sounds good

internal/agent/device/applications/lifecycle/quadlet.go
kkyrazis: fixed.

hexfusion (approved):

PR #2174, opened 2025-12-06
EDM-2757: Add quadlet file cross referencing validation

kkyrazis:
Adds validation to ensure applications that have quadlets that reference other quadlets are all defined within their application.
The following spec will now produce an error like: 
```"apiVersion":"v1beta1","code":400,"kind":"Status","message":"quadlet file \"app.pod\" references \"app.network\" which is not defined in the application\nquadlet file \"app.container\" references \"app.network\" which is not defined in the application","reason":"Bad Request","status":"Failure"```
```yaml
      inline:
        - path: app.pod
          content: |
            [Pod]
            Network=app.network
        - path: app.container
          content: |
            [Unit]
            Description=Primary application container
            
            [Container]
            Image=quay.io/flightctl-tests/alpine:v1
            Network=app.network
            Pod=app.pod
            
            [Install]
            WantedBy=default.target
```

## Summary by CodeRabbit

* **New Features**
  * Stricter quadlet cross-reference validation: ensures images, volumes, networks and pods referenced in quadlets are defined within the application.
  * Expanded parsing of mount and volume references for more accurate validation.
  * Unified application validation flow that validates compose- and quadlet-based packages consistently.

* **Tests**
  * Expanded test coverage for quadlet references, mounts and cross-reference error scenarios.

### DIFF
--- a/api/v1beta1/validation.go
+++ b/api/v1beta1/validation.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/flightctl/flightctl/internal/api/common"
 	"github.com/flightctl/flightctl/internal/contextutil"
-	"github.com/flightctl/flightctl/internal/quadlet"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/internal/util/validation"
 	"github.com/robfig/cron/v3"
@@ -785,6 +784,16 @@ func (a ApplicationProviderSpec) Validate() []error {
 func (a InlineApplicationProviderSpec) Validate(appType AppType, fleetTemplate bool) []error {
 	allErrs := []error{}
 
+	var appValidator applicationValidator
+	switch appType {
+	case AppTypeCompose:
+		appValidator = &composeValidator{paths: make(map[string]struct{})}
+	case AppTypeQuadlet:
+		appValidator = &quadletValidator{quadlets: make(map[string]*common.QuadletReferences)}
+	default:
+		appValidator = &unknownAppTypeValidator{appType}
+	}
+
 	seenPath := make(map[string]struct{}, len(a.Inline))
 	for i := range a.Inline {
 		path := a.Inline[i].Path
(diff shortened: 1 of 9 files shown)


internal/api/common/quadlet.go
```diff
@@ -137,3 +171,22 @@ func ParseQuadletReferences(data []byte) (*QuadletReferences, error) {
 
 	return spec, nil
 }
+
+func quadletReferences(unit *quadlet.Unit, section string, key string, extension string) ([]string, error) {
```
hexfusion: optional: this is a powerful function can we hint its use a bit more in the naming

```suggestion
func lookupQuadletReferences(unit *quadlet.Unit, section string, key string, extension string) ([]string, error) {
```

api/v1beta1/validation.go
```diff
@@ -815,6 +816,33 @@ func (a InlineApplicationProviderSpec) Validate(appType AppType, fleetTemplate b
...
+
+	var errs []error
+	for _, c := range inline {
+		if !quadlet.IsQuadletFile(c.Path) {
+			continue
+		}
+
+		decoded, err := c.DecodeContent()
+		if err != nil {
+			// decoding content is already checked as part of Validations. If this fails, don't reinclude it as an error
```
hexfusion: not sure I am understanding this if its already checked why do we need to do it again. can we see if we can do this in another way. totally fine if its required.

internal/api/common/quadlet.go
```diff
@@ -2,6 +2,8 @@ package common
 
 import (
 	"fmt"
+	"path/filepath"
```
hexfusion: question: this feels like fileio territory should logic around the filesystem be part of internal/agent

kkyrazis: In this case, `filepath` is being used to discover what kind of quadlet we're dealing with. I think checking the extension - in the services or agent - should be an acceptable way to check the type. I can swap to `strings.HasSuffix` or roll my own if we'd rather not leak path into the api.

api/v1beta1/validation.go
kkyrazis: The current validation flow is to iterate through all of the contents defined, validate those individually and then move on to validating the spec as a whole. This continues on with that pattern, but it does require reparsing the content. I'll come up with a nicer way to do this that doesn't require multliple parsing passes.

internal/api/common/quadlet.go
hexfusion: all good if you feel its controlled

api/v1beta1/validation.go
hexfusion: not required but if the pattern emerges :)

internal/api/common/quadlet.go
kkyrazis: reworked to lean on the quadlet package for this.

api/v1beta1/validation.go
kkyrazis: reworked this section with specific app type validators.

PR #2178, opened 2025-12-07
EDM-2760: multi-auth listing AP's without org filter

asafbennatan:
EDM-2760: multi-auth listing AP's without org filter
NO-ISSUE: prevent non super-admins from setting static role = super admin , or by receiving it as a dynamic role value
NO-ISSUE: fix defaults not being set in validation.go or config.go (they were set in individual auth providers) which caused false detection of auth provider update

## Summary by CodeRabbit

* **New Features**
  * Added listing of all authentication providers across organizations.
  * Authentication providers created by super admins now support flightctl-admin role assignments.

* **Bug Fixes**
  * Fixed flightctl-admin role filtering: roles are now correctly preserved when providers are created by super admins.

* **Security**
  * Enforced admin-only restrictions for static role and organization mapping configurations.

### DIFF
--- a/api/v1beta1/validation.go
+++ b/api/v1beta1/validation.go
@@ -17,8 +17,7 @@ import (
 	"time"
 
 	"github.com/flightctl/flightctl/internal/api/common"
-	"github.com/flightctl/flightctl/internal/consts"
-	"github.com/flightctl/flightctl/internal/identity"
+	"github.com/flightctl/flightctl/internal/contextutil"
 	"github.com/flightctl/flightctl/internal/quadlet"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/internal/util/validation"
@@ -58,6 +57,7 @@ var (
 	ErrAapProviderConfigOnly                 = errors.New("aap provider type can only be created from configuration, not via API")
 	ErrDynamicOrgMappingAdminOnly            = errors.New("only flightctl-admin users are allowed to create auth providers with dynamic organization mapping")
 	ErrPerUserOrgMappingAdminOnly            = errors.New("only flightctl-admin users are allowed to create auth providers with per-user organization mapping")
+	ErrStaticRoleMappingAdminOnly            = errors.New("only flightctl-admin users are allowed to create static role mappings for flightctl-admin")
 )
 
 type Validator interface {
@@ -1431,20 +1431,9 @@ func (o *OAuth2ProviderSpec) Validate(ctx context.Context, isUpdate bool) []erro
 		allErrs = append(allErrs, ErrClientIdRequired)
 	}
 
-	// Handle introspection field validation
+	// Validate introspection field is present
 	if o.Introspection == nil {
-		if !isUpdate {
-			// During creation, try to infer introspection
-			introspection, err := InferOAuth2IntrospectionConfig(*o)
-			if err != nil {
-				allErrs = append(allErrs, fmt.Errorf("introspection field is required and could not be inferred: %w", err))
-			} else {
-				o.Introspection = introspection
-			}
-		} else {
-			// During update, introspection is required and cannot be nil
-			allErrs = append(allErrs, fmt.Errorf("introspection field is required for OAuth2 providers"))
-		}
+		allErrs = append(allErrs, fmt.Errorf("introspection field is required for OAuth2 providers"))
 	}
 
 	// Validate organization assignment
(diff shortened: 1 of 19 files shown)


internal/service/authprovider.go
```diff
@@ -35,6 +37,21 @@ func (h *ServiceHandler) CreateAuthProvider(ctx context.Context, orgId uuid.UUID
...
+		})
+		// Use fromAPI=false to preserve annotations
+		result, err := h.store.AuthProvider().CreateWithFromAPI(ctx, orgId, &authProvider, false, h.callbackAuthProviderUpdated)
+		return result, StoreErrorToApiStatus(err, true, api.AuthProviderKind, authProvider.Metadata.Name)
+	}
+
+	// For non-super-admin users, use regular Create (fromAPI=true, annotations cleared)
 	result, err := h.store.AuthProvider().Create(ctx, orgId, &authProvider, h.callbackAuthProviderUpdated)
 	return result, StoreErrorToApiStatus(err, true, api.AuthProviderKind, authProvider.Metadata.Name)
 }
```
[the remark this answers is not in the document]
asafbennatan: this is already done in generic.go when request is not internal

api/v1beta1/validation.go
```diff
@@ -1431,6 +1431,20 @@ func (o *OAuth2ProviderSpec) Validate(ctx context.Context, isUpdate bool) []erro
 		allErrs = append(allErrs, ErrClientIdRequired)
 	}
 
+	// Apply defaults (only during creation, not updates)
+	if !isUpdate {
+		// Default UsernameClaim to ["preferred_username"] if not provided
+		if o.UsernameClaim == nil || len(*o.UsernameClaim) == 0 {
+			defaultUsernameClaim := []string{"preferred_username"}
+			o.UsernameClaim = &defaultUsernameClaim
```
ori-amizur: Validate function should not modify the object

asafbennatan: moved it to a dedicated applyDefaults function that is called before validate in the relevant create cases

internal/store/authprovider.go
```diff
@@ -185,6 +206,115 @@ func (s *AuthProviderStore) CountByOrg(ctx context.Context, orgId *uuid.UUID) ([
...
+	if listParams.AnnotationSelector != nil {
+		q, p, err := listParams.AnnotationSelector.Parse(ctx, selector.NewHiddenSelectorName("metadata.annotations"), nil)
+		if err != nil {
+			return nil, err
+		}
+		query = query.Where(q, p...)
+	}
+
+	// Apply sorting
+	columns, order, _ := getSortColumns(listParams)
```
ori-amizur, line 245: Why arbitrary list? Shouldn't the sorting be based on the key ?  This way the the continue value will be more exact ?
If the sort columns are missing, is it possible to have a limit ? If not, it needs to be validated

asafbennatan: its not arbitrary list its ListParam object list we have for other List methods

api/v1beta1/validation.go
```diff
@@ -1738,6 +1699,28 @@ func (a AuthStaticRoleAssignment) Validate(ctx context.Context) []error {
...
+	hasAdminRole := false
+	for _, role := range a.Roles {
+		if role == ExternalRoleAdmin {
+			hasAdminRole = true
+			break
+		}
+	}
+
+	if hasAdminRole {
+		mappedIdentity, ok := contextutil.GetMappedIdentityFromContext(ctx)
```
sk-ilya, line 1712: `api/v1beta1/validation.go` should contain stateless API level validation and shouldn't depend on request context. It feels like these checks belong to service/middleware

asafbennatan: would you have the middleware handle specific apis , that would grow and would be unmaintainable at some point , in general the solution is to have something more of a ReBAC to control this - but we are not there yet so we must manually validate it

api/v1beta1/validation.go
```diff
@@ -1431,20 +1431,9 @@ func (o *OAuth2ProviderSpec) Validate(ctx context.Context, isUpdate bool) []erro
...
-			if err != nil {
-				allErrs = append(allErrs, fmt.Errorf("introspection field is required and could not be inferred: %w", err))
-			} else {
-				o.Introspection = introspection
-			}
-		} else {
-			// During update, introspection is required and cannot be nil
-			allErrs = append(allErrs, fmt.Errorf("introspection field is required for OAuth2 providers"))
-		}
+		allErrs = append(allErrs, fmt.Errorf("introspection field is required for OAuth2 providers"))
```
fzdarsky, line 1436: If this field is required, why not make it required in the OpenAPI spec?

asafbennatan: i wanted to set sensible defaults ,for that it must be optional

PR #2183, opened 2025-12-08
EDM-2734: Update Middleware HTTP errs

SiddarthR56:
## Summary by CodeRabbit

* **Refactor**
  * Standardized error handling across proxies, API server and middleware so failures are returned as consistent, structured API status payloads instead of raw HTTP error writes.

* **New Features**
  * Added several new standardized API status variants covering common HTTP error conditions (e.g., not found, service unavailable, method not allowed, request URI/header too large) to improve uniformity and clarity of error responses.

### DIFF
--- a/cmd/flightctl-alertmanager-proxy/main.go
+++ b/cmd/flightctl-alertmanager-proxy/main.go
@@ -24,6 +24,7 @@ import (
 	"syscall"
 	"time"
 
+	api "github.com/flightctl/flightctl/api/v1beta1"
 	"github.com/flightctl/flightctl/internal/api_server/middleware"
 	"github.com/flightctl/flightctl/internal/auth"
 	"github.com/flightctl/flightctl/internal/auth/common"
@@ -32,6 +33,7 @@ import (
 	"github.com/flightctl/flightctl/internal/org/cache"
 	"github.com/flightctl/flightctl/internal/service"
 	"github.com/flightctl/flightctl/internal/store"
+	"github.com/flightctl/flightctl/internal/transport"
 	fclog "github.com/flightctl/flightctl/pkg/log"
 	"github.com/go-chi/chi/v5"
 	chimiddleware "github.com/go-chi/chi/v5/middleware"
@@ -144,13 +146,13 @@ func createConditionalAuthMiddleware(
 						allowed, err := authZ.CheckPermission(r.Context(), alertsResource, getAction)
 						if err != nil {
 							logger.WithError(err).Error("Authorization check failed")
-							http.Error(w, "Authorization service unavailable", http.StatusServiceUnavailable)
+							transport.SetResponse(w, nil, api.StatusServiceUnavailable("Authorization service unavailable"))
 							return
 						}
 
 						if !allowed {
 							logger.Warn("User denied access to alerts")
-							http.Error(w, "Forbidden", http.StatusForbidden)
+							transport.SetResponse(w, nil, api.StatusForbidden("Forbidden"))
 							return
 						}
 
--- a/cmd/flightctl-userinfo-proxy/main.go
+++ b/cmd/flightctl-userinfo-proxy/main.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	api "github.com/flightctl/flightctl/api/v1beta1"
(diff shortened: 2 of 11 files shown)


internal/pam_issuer_server/server.go
```diff
@@ -45,7 +48,11 @@ func New(
 }
 
 func oapiErrorHandler(w http.ResponseWriter, message string, statusCode int) {
-	http.Error(w, fmt.Sprintf("API Error: %s", message), statusCode)
+	if statusCode < 0 || statusCode > math.MaxInt32 {
```
asafbennatan: i dont think this condition makes sense

internal/pam_issuer_server/server.go
```diff
@@ -45,7 +48,11 @@ func New(
...
+		statusCode = http.StatusInternalServerError
+	}
+	status := api.NewFailureStatus(int32(statusCode), http.StatusText(statusCode), fmt.Sprintf("API Error: %s", message)) // #nosec G115 -- statusCode validated above
+	transport.SetResponse(w, nil, status)
```
asafbennatan: in any case this shouldnt have effect - per OIDC spec errors are returned in a certain way

internal/api_server/middleware/identity_mapping.go
```diff
@@ -5,12 +5,13 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/flightctl/flightctl/api/v1beta1"
+	api "github.com/flightctl/flightctl/api/v1beta1"
```
asafbennatan: i dont think this renaming is necessary

internal/api_server/agentserver/server.go
```diff
@@ -123,7 +126,11 @@ func (s *AgentServer) GetGRPCServer() *AgentGrpcServer {
 }
 
 func oapiErrorHandler(w http.ResponseWriter, message string, statusCode int) {
-	http.Error(w, fmt.Sprintf("API Error: %s", message), statusCode)
+	if statusCode < 0 || statusCode > math.MaxInt32 {
```
asafbennatan: this condition doesnt make sense - status code is already an int

asafbennatan (changes requested):

cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -144,13 +146,29 @@ func createConditionalAuthMiddleware(
 						allowed, err := authZ.CheckPermission(r.Context(), alertsResource, getAction)
 						if err != nil {
 							logger.WithError(err).Error("Authorization check failed")
-							http.Error(w, "Authorization service unavailable", http.StatusServiceUnavailable)
+							transport.SetResponse(
+								w,
+								nil,
+								api.NewFailureStatus(
```
kkyrazis: could be useful to create something like `api.NewHttpFailure(http.StatusCode, message)`

cmd/flightctl-userinfo-proxy/main.go
```diff
@@ -120,30 +122,62 @@ func makeUserInfoHandler(config *Config, log *logrus.Logger) http.HandlerFunc {
...
 		if err != nil {
 			log.Printf("Error proxying to upstream AAP: %v", err)
-			http.Error(w, "Internal server error", http.StatusInternalServerError)
+			transport.SetResponse(
+				w,
+				nil,
+				api.NewFailureStatus(
+					int32(http.StatusInternalServerError),
+					http.StatusText(http.StatusInternalServerError),
+					"Internal server error",
```
kkyrazis: Is the intention to also provide more information? This is just going to come out as {"500", "Internal Server Error", "Internal server error"}, which isn't particularly useful

SiddarthR56: The main goal of this change is to ensure error responses conform to our OpenAPI spec by returning a proper Status object rather than plain text from http.Error().

You’re right that having both is redundant. The reason I didn’t add a more descriptive message is to avoid exposing too much information, consistent with the previous behavior. I could either keep the redundancy as-is or leave reason as an empty string when there’s no extra context. What would you suggest?

kkyrazis: I think your reasoning/decision is solid.

PR #2184, opened 2025-12-09
EDM-2764: cluster role suffix based on release

asafbennatan:
## Summary by CodeRabbit

* **New Features**
  * Optional configurable role suffix to allow ClusterRole names to be suffixed by release/namespace for multi-release coexistence.

* **Behavior**
  * Role lookups now normalize suffixed ClusterRole names so displayed and applied roles match standard role names.
  * Default admin service account name remains unsuffixed; RoleBindings must reference suffixed ClusterRole names when used.

* **Documentation**
  * Kubernetes and OpenShift auth guides updated with suffixed ClusterRole names and example commands.


deploy/helm/flightctl/templates/api/flightctl-api-config.yaml
```diff
@@ -78,6 +78,7 @@ data:
         k8s:
             apiUrl: {{ .Values.global.auth.k8s.apiUrl }}
             rbacNs: {{ default .Release.Namespace .Values.global.auth.k8s.rbacNs }}
+            roleSuffix: {{ .Release.Name }}
```
rawagner: we should use `.Release.Namespace` everywhere.
You can deploy only 1 instance of fctl into each ns. But all instances can have the same release name

rawagner (changes requested):

hexfusion:
@rawagner @asafbennatan what is the status of this PR

asafbennatan:
@hexfusion waiting for an admin to add the breaking helm changes , can you ?

PR #2194, opened 2025-12-10
NO-ISSUE: fix login form xss

asafbennatan:
using proper template to create the login form

## Summary by CodeRabbit

* **New Features**
  * Templated login UI with improved client-side behavior: loading state, inline error display, and smoother submit flow.
* **Bug Fixes**
  * Safer HTML rendering with proper escaping and clearer error responses when form generation fails.
* **Refactor**
  * Server now precompiles and serves the login form template for more reliable rendering and better performance.

### DIFF
--- a/internal/auth/oidc/pam/provider.go
+++ b/internal/auth/oidc/pam/provider.go
@@ -3,14 +3,16 @@
 package pam
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"crypto/sha256"
+	_ "embed"
 	"encoding/base64"
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"html"
+	"html/template"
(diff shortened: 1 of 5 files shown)


sk-ilya (approved):

internal/auth/oidc/pam/provider.go
```diff
@@ -24,6 +25,192 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const loginFormTemplate = `<!DOCTYPE html>
```
hexfusion: nit: can we  store the html as a file and use go:embed, I think it will save us from future formatting errors.

//go:embed templates/login_form.html
var loginFormTemplate string

asafbennatan: done

PR #2201, opened 2025-12-11
EDM-2783: agent/device/applications: ensure terminated containers

hexfusion:
The agent now automatically detects and restarts application containers that have unexpectedly stopped, died, or been removed. This enables self-healing for managed applications without manual intervention.
  

## Summary by CodeRabbit

* **New Features**
  * Applications now surface per-service restart policies; workloads include service identity and restart metadata; runtime inspection exposes host config/restart-policy info and compose service label.

* **Bug Fixes**
  * Reconcile/ensure detects stopped workloads and queues restarts during reconciliation; install flow now triggers post-install checks to enable self‑healing.

* **Tests**
  * Added suites validating stopped-workload detection and restart/reconciliation behavior.

* **Documentation**
  * Added "Application Self‑Healing" guidance.

* **Chores**
  * Reduced device version-change log verbosity; updated spelling dictionary.

### DIFF
--- a/internal/agent/device/applications/podman_monitor.go
+++ b/internal/agent/device/applications/podman_monitor.go
@@ -30,6 +30,10 @@ const (
 	stopMonitorWaitDuration       = 5 * time.Second
 	quadletSystemdLabel           = "PODMAN_SYSTEMD_UNIT"
 	maxAppSummaryInfoLength       = 256
+
+	RestartPolicyNo    = "no"
+	RestartPolicyNever = "never"
+	RestartPolicyFalse = "false"
 )
 
 type PodmanMonitor struct {
@@ -300,20 +304,25 @@ func (m *PodmanMonitor) Has(id string) bool {
 	return false
 }
 
-// Ensures that and application is added to the monitor. if the application
+// Ensures that an application is added to the monitor. If the application
 // is added for the first time an Add action is queued to be executed by the
-// lifecycle manager. so additional adds for the same app will be idempotent.
+// lifecycle manager. If the application already exists but has workloads that
+// are not running (died, stopped, or removed), the monitor will queue a restart.
 func (m *PodmanMonitor) Ensure(app Application) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
 	appID := app.ID()
-	_, ok := m.apps[appID]
+	existingApp, ok := m.apps[appID]
 	if ok {
-		// app already exists
-		return nil
+		stoppedWorkloads := m.getStoppedWorkloads(existingApp)
+		if len(stoppedWorkloads) == 0 {
+			return nil
+		}
+		m.log.Warnf("Application %s has stopped workloads %v, queueing restart", existingApp.Name(), stoppedWorkloads)
+	} else {
+		m.apps[appID] = app
 	}
-	m.apps[appID] = app
 
 	appName := app.Name()
 	action := lifecycle.Action{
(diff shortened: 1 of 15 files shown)


internal/agent/device/applications/podman_monitor.go
```diff
@@ -334,6 +339,19 @@ func (m *PodmanMonitor) canRemoveApp(app Application) bool {
 	return ok || app.IsEmbedded()
 }
 
+// getStoppedWorkloads returns a list of workload names that have been stopped,
+// died, or removed and need to be restarted.
+func (m *PodmanMonitor) getStoppedWorkloads(app Application) []string {
+	var stopped []string
+	for _, workload := range app.Workloads() {
+		switch workload.Status {
+		case StatusDied, StatusDie, StatusStop, StatusRemove:
```
kkyrazis, line 359: all of the app types have some built in mechanism for restarting. For compose it's `restart`, for quadlets it's `[Serivce]\nRestart=on-failure` and for container apps we manually add the Restart key.  Is the intention to make it simpler for the user so that they won't have to worry setting it up? My only real concern is that this makes it so that you can't define an app as "never restart"

hexfusion: that a good callout we should support those internal details will ensure that is considered

internal/agent/device/applications/podman_monitor.go
```diff
@@ -695,10 +728,20 @@ func (m *PodmanMonitor) updateQuadletContainerStatus(ctx context.Context, app Ap
 	if isFinishedStatus(status) && event.ContainerExitCode == 0 {
 		status = StatusExited
 	}
-	m.updateApplicationStatus(app, event, status, restartCount)
+
+	if status == StatusRemove {
```
kkyrazis, line 732: Quadlets are different in annoying ways. This is one of them. Containers don't live outside of the lifecycle of a single start/stop of the systemd unit that controls them. `systemctl stop <quadlet.service>` will generate a removal event, but we don't necessarily want to lose that workload. In this case, I purposely left out removing workloads.

internal/agent/device/applications/podman_monitor.go
```diff
@@ -695,10 +728,20 @@ func (m *PodmanMonitor) updateQuadletContainerStatus(ctx context.Context, app Ap
...
+		if removed := app.RemoveWorkload(event.Name); removed {
+			m.log.Debugf("Removed container: %s", event.Name)
+		}
+		return
+	}
+
+	m.updateApplicationStatus(app, event, "", status, restartCount)
```
kkyrazis, line 739:
```suggestion
	m.updateApplicationStatus(app, event, systemdUnit, status, restartCount)
```
is that appropriate in this case?

internal/agent/device/applications/podman_monitor.go
```diff
@@ -300,20 +304,25 @@ func (m *PodmanMonitor) Has(id string) bool {
...
-	_, ok := m.apps[appID]
+	existingApp, ok := m.apps[appID]
 	if ok {
-		// app already exists
-		return nil
+		stoppedWorkloads := m.getStoppedWorkloads(existingApp)
+		if len(stoppedWorkloads) == 0 {
+			return nil
+		}
+		m.log.Warnf("Application %s has stopped workloads %v, queueing restart", existingApp.Name(), stoppedWorkloads)
```
kkyrazis, line 322: where is the queueing happening?

hexfusion:
yeah lets hold this for 1.1 its not a critical thing and needs more testing

PR #2214, opened 2025-12-11
EDM-2786: Use auto instead of hardcoded ssl_ecdh_curve value

DakCrowder:
(cherry picked from commit 1b5c23b224e32a9522e25832f82f5a3e851b3ee9)


DakCrowder:
Note: I believe the Helm Upgrade test failure highlights an issue with the CI flow, currently `main` is always used as the basis for the Helm upgrade and since breaking changes have been merged to main and not release-1.0 it fails

DakCrowder:
Captured in https://github.com/flightctl/flightctl/pull/2217

PR #2220, opened 2025-12-12
EDM-2562: Apply override to UI tags only for releases

celdrake:
Done with the help of Cursor.

I believe this should fix the problem whereby for development branches etc the flightctl-ui image was tagged taking the tag formats in the flightctl repo. UI will always use `latest` in those cases.

For cut releases, the UI tag should behave as it does now (eg. for v1.0.0-rc2 it would have the tag 1.0.0-rc2).

## Summary by CodeRabbit

* **New Features**
  * Added a command-line option to explicitly enable a UI image tag override during rendering.

* **Improvements**
  * Build/release packaging now auto-enables the UI tag override for release-tagged builds.
  * Rendering logs now clearly indicate whether the UI tag override was applied or skipped.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -264,6 +264,7 @@ v1.0.0
 P-256
 SHA-256
 config.yaml
+FQDN
  - docs/user/references/api-resources.md
 namespacing
 approvedBy
--- a/cmd/flightctl-standalone/render_template.go
+++ b/cmd/flightctl-standalone/render_template.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/flightctl/flightctl/internal/config/standalone"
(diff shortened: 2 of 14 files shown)


internal/quadlet/renderer/renderer.go
```diff
@@ -47,6 +47,7 @@ type RendererConfig struct {
 	BinOutputDir             string `mapstructure:"bin-dir"`
 
 	FlightctlServicesTagOverride string `mapstructure:"flightctl-services-tag-override"`
+	ApplyUiTagOverride           bool   `mapstructure:"apply-ui-tag-override"`
```
kkyrazis: I think you'd still need to add a flag here: https://github.com/flightctl/flightctl/blob/main/cmd/flightctl-standalone/render_quadlets.go#L37

packaging/rpm/flightctl.spec
```diff
@@ -480,9 +480,22 @@ echo "Flightctl Observability Stack uninstalled."
...
     # to transform the version string by replacing tildes with hyphens.
     IMAGE_TAG=$(echo %{version} | tr '~' '-')
+
+    # Convert version to git tag format (add 'v' prefix) and check if it points at HEAD
+    GIT_TAG_VERSION="v${IMAGE_TAG}"
+
+    # If the version matches a git tag pointing at HEAD,
+    # it's a release and UI should use the same tag. Otherwise, it's a dev build and UI
+    # should keep its default tag (typically "latest").    
+    if git tag --points-at HEAD 2>/dev/null | grep -q "^${GIT_TAG_VERSION}$"; then
```
kkyrazis: I'm not sure how well `git tag` will work in all rpm building environments, and if it doesn't we'll default to latest. What about something that is functionally equivalent to IMAGE_TAG matches x.x.x or x.x.x-rcX. That's not great either, but just a thought.

internal/quadlet/renderer/renderer.go
celdrake: Good catch, thanks.

packaging/rpm/flightctl.spec
celdrake: That makes sense.
Used a regular expression to match the version to an release version or RC.

PR #2225, opened 2025-12-12
EDM-2794: add missing TLS SANs, validate FQDN

fzdarsky:
## Summary by CodeRabbit

* **New Features**
  * Expanded certificate SAN generation (multiple service types, hostnames, all IPs); rendering now auto-completes/validates config and supports rendering with provided data; OAuth client ID persisted to a file.

* **Bug Fixes**
  * Stricter hostname/FQDN validation with clear erroring and fallback to host FQDN when unset.

* **Configuration**
  * Added global.baseDomain and auth.insecureSkipTlsVerify; init captures HOST_FQDN at startup; updated volume bindings for alertproxy certs.

* **Tests**
  * Added comprehensive FQDN/hostname validation tests.

* **Documentation**
  * Consolidated installer guide and clarified base-domain and certificate guidance.

### DIFF
--- a/cmd/flightctl-standalone/render_template.go
+++ b/cmd/flightctl-standalone/render_template.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/flightctl/flightctl/internal/config/standalone"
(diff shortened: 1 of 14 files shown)


internal/util/validation/standalone.go
aekubacki, line 1: Do we need to enforce the "at least one dot" rule to validate its a FQDN?  If this is a machine not joined to a DNS, it would be the name of the server which does not enforce a domain being present. Do we want to dictate how a customer/user has their server named?

fzdarsky: Good catch, thank you. Yes, we should allow single hostnames.

fzdarsky: [x] done

deploy/scripts/init_certs.sh
```diff
@@ -13,6 +13,60 @@ else
...
-# Generate certificates
-BASE_DOMAIN=$(python3 "$YAML_HELPER" extract .global.baseDomain "$CONFIG_FILE")
-/usr/share/flightctl/generate-certificates.sh --cert-dir "$CERT_DIR" --api-san "$BASE_DOMAIN"
+# Get the system hostname FQDN for use as default and in certificate SANs
+hostname_fqdn=$(hostname -f || hostname)
+
+# Get the primary host IP address:
+# First try to get the IP address used to reach the internet.
+# If that fails (e.g. no internet access), use the first non-loopback IP.
+host_ip="$(
```
asafbennatan: why should we only extract the first / internet facing one ?
wouldnt just takeing hostname -I be a better choice ( this would make it work in more complicated scnearios where there are multiple ips etc)

deploy/scripts/init_certs.sh
```diff
@@ -13,6 +13,60 @@ else
     exit 0
 fi
 
-# Generate certificates
-BASE_DOMAIN=$(python3 "$YAML_HELPER" extract .global.baseDomain "$CONFIG_FILE")
-/usr/share/flightctl/generate-certificates.sh --cert-dir "$CERT_DIR" --api-san "$BASE_DOMAIN"
+# Get the system hostname FQDN for use as default and in certificate SANs
+hostname_fqdn=$(hostname -f || hostname)
```
asafbennatan, line 18: shouldnt we take both ?

deploy/scripts/init_certs.sh
```diff
@@ -13,6 +13,60 @@ else
...
+  ip route get 1.1.1.1 2>/dev/null \
+    | awk '{for (i=1;i<=NF;i++) if ($i=="src") {print $(i+1); exit}}' \
+    || true
+)"
+if [ -z "$host_ip" ]; then
+    host_ip=$(hostname -I | awk '{print $1}')
+fi
+
+# Validate the base domain from config, or default to hostname FQDN
+base_domain=$(python3 "$YAML_HELPER" extract .global.baseDomain "$CONFIG_FILE")
```
asafbennatan, line 34: should we replace this with extract_value like we do in init_host.sh?

docs/user/installing/installing-service-on-linux.md
```diff
@@ -28,53 +28,31 @@ sudo dnf config-manager addrepo --from-repofile=https://rpm.flightctl.io/flightc
 sudo dnf install -y flightctl-services
 ```
 
-### Installing a specific version
+Flight Control services can be configured through a central configuration file located at `/etc/flightctl/service-config.yaml`.
 
-Search for available versions:
+To spin up services quickly for testing or development purposes, you can leave this file's defaults. This sets the base domain of the services to the host's fully qualified domain name (FQDN) (from `hostname -f`) and generates a self-signed certificate authority (CA) from which required certificates are issued.
```
asafbennatan, line 33: if choose to adjust based on above comments - adjust the docs as well

deploy/scripts/init_host.sh
```diff
@@ -9,23 +9,22 @@ source "${SCRIPT_DIR}"/init_utils.sh
 
 SERVICE_CONFIG_FILE="/etc/flightctl/service-config.yaml"
 
-write_default_base_domain() {
-    # Write base domain to the config file
-    base_domain="$(ip route get 1.1.1.1 | grep -oP 'src \K\S+')"
-    echo "Setting base domain to: ${base_domain}"
-    sed -i "s/^\(\s*baseDomain\s*\):\s*.*$/\1: ${base_domain}/" "${SERVICE_CONFIG_FILE}"
```
asafbennatan, line 16: i think the removal of this breaks some assumptions down the line , see :

https://github.com/flightctl/flightctl/blob/7ddaf3135dc3150f96db14e0960a33bd2f6fb915/deploy/podman/flightctl-api/flightctl-api-config/config.yaml.template#L16


and seems like `baseUrl`  value is used in the code

i assume the correct approach would be to have the template set the correct value there based on the generated certificate ?

cc: @DakCrowder

deploy/scripts/init_certs.sh
fzdarsky: It's a slippery slope. That list would also contain, for example, the Podman bridge IP, so we'd need to filter that out reliably. It may also leak more informations about the internal networks than comfortable and include transient IP addresses...

deploy/scripts/init_certs.sh
fzdarsky: Makes sense, will do.

deploy/scripts/init_host.sh
fzdarsky: That's indeed a problem, yes. I've updated the rendering logic to also default the `baseDomain` to `$(hostname -f || hostname)`. However, there's one last instance to fix, which is in the `fligthctl-api-init` container.

fzdarsky: Ok, found a fix for the `flightctl-api-init` container, too. Only thing this file needs to do still is to ensure the secrets, so maybe we should give the file a clearer name.

cmd/flightctl-standalone/render_template.go
```diff
@@ -41,22 +42,111 @@ func NewRenderTemplateCommand() *cobra.Command {
...
+		}
+	}
+
+	return nil
+}
+
+func (o *RenderTemplateOptions) getHostnameFQDN() (string, error) {
+	// Try "hostname -f" first for FQDN
+	cmd := exec.Command("hostname", "-f")
+	output, err := cmd.Output()
```
hexfusion, line 130: question, is the assumption this should never error so we should not handle if err != nil?

cmd/flightctl-standalone/render_template.go
```diff
@@ -41,22 +42,111 @@ func NewRenderTemplateCommand() *cobra.Command {
...
+	auth, ok := global["auth"].(map[string]interface{})
+	if ok {
+		authType, _ := auth["type"].(string)
+		if authType == "aap" {
+			clientIDFile := "/etc/flightctl/pki/aap-client-id"
+			if clientIDData, err := os.ReadFile(clientIDFile); err == nil {
+				clientID := strings.TrimSpace(string(clientIDData))
+				if clientID != "" {
+					aap, ok := auth["aap"].(map[string]interface{})
+					if !ok {
```
hexfusion: optional: while logical this deep if nesting is getting a little hard to read. could we keep these if conditions on the root level and return early if condition is unexpected?

deploy/scripts/init_certs.sh
```diff
@@ -13,6 +13,85 @@ else
...
+# API certificate SANs
+api_sans=(
+    "api.$base_domain"
+    "agent-api.$base_domain"
+    "$base_domain"
+    "$hostname_short"
+    "$hostname_fqdn"
+    "flightctl-api"
+    "localhost"
+    "127.0.0.1"
```
hexfusion: there are security implications to allowing localhost in the SAN I suppose this is a non production self signed cert.

ref, https://letsencrypt.org/docs/certificates-for-localhost/

deploy/scripts/init_certs.sh
hexfusion: I agree this is a hard decision but it was the choice to make the system deterministic. first interface first ip.

deploy/scripts/init_certs.sh
hexfusion: we talked this over and are leaning towards usability for now and doing a more formal review in future.

deploy/scripts/init_certs.sh
```diff
@@ -13,6 +13,85 @@ else
...
+)
+api_sans+=("${host_ips[@]}")
+
+# Telemetry certificate SANs
```
DakCrowder, line 66: These SANs in isolation seem good to me but currently the telemetry gateway installation is largely "out of band" with the services install - I believe there is a [separate cert initialization process](https://github.com/flightctl/flightctl/blob/main/packaging/observability/flightctl-render-observability#L121) that we _should_ align on eventually but don't know if this needs to happen immediately or if we should/can follow up on this.

Additionally the alertmanager-proxy certs [volume](https://github.com/flightctl/flightctl/blob/main/deploy/podman/flightctl-alertmanager-proxy/flightctl-alertmanager-proxy.container#L15) can be updated to reference the proxy specific certs if we are now generating them

deploy/scripts/init_certs.sh
fzdarsky: Ok, I've implemented a compromise where the default (routable) interface is being used if it exists and if it doesn't, we fall back to `hostname -I` and add all returned interfaces (as we cannot assume the first to be the correct one). Either way, users are free to create their own certificates if needed.

deploy/scripts/init_certs.sh
fzdarsky: Yeah, feels unnecessary to have two ways for doing this. Looking at the code, either are used about 50:50.
I'd suggest cleaning this up separately.

deploy/scripts/init_certs.sh
fzdarsky: Thanks. Updated the alertmanager-proxy certs.

For the telemetry gateway, that process is probably no longer valid, as its goal was to work around the server cert generation by the `flightctl-api` itself, which is no longer the case. I'd need help understanding that previous process, though.

cmd/flightctl-standalone/render_template.go
fzdarsky: Idea was to have the equivalent of `$(hostname -f || hostname)`, i.e. if the first command fails for whatever reason, try the second one and only error out if that fails, too.

PR #2227, opened 2025-12-14
EDM-2795: Ensure consumer group is recreated on Redis restart 

sk-ilya:
Recreate consumer group if it is lost

## Summary by CodeRabbit

* **New Features**
  * Added configurable retry/backoff for queue operations with a public default and jitter.

* **Bug Fixes**
  * Improved recovery from missing consumer groups by creating them idempotently.
  * Added in-queue backoff retries for transient consume errors and more reliable handling of timed-out and in‑flight tasks.

* **Tests**
  * Added post-restart verification to ensure the task queue initializes after Redis recovery.

### DIFF
--- a/cmd/flightctl-agent/main.go
+++ b/cmd/flightctl-agent/main.go
@@ -160,7 +160,7 @@ func printAgentHelp() {
 
 func printVersion() {
 	versionInfo := version.Get()
-	fmt.Printf("Flightctl Agent Version: %s\n", versionInfo.String())
+	fmt.Printf("Agent Version: %s\n", versionInfo.String())
 	fmt.Printf("Git Commit: %s\n", versionInfo.GitCommit)
 }
 
--- a/deploy/helm/flightctl/README.md
+++ b/deploy/helm/flightctl/README.md
@@ -293,6 +293,7 @@ For more detailed configuration options, see the [Values](#values) section below
 | global.imagePullPolicy | string | `"IfNotPresent"` | Image pull policy for all containers |
 | global.imagePullSecretName | string | `""` | Name of the secret that holds image pull secret for accessing private container registries |
 | global.internalNamespace | string | `""` | A separate Namespace to which non-user-facing components should be deployed for increased security isolation. |
+| global.multiclusterEngineNamespace | string | `"multicluster-engine"` | Namespace where MultiCluster Engine is installed. Used for creating discovery ConfigMap and RBAC bindings. |
 | global.sshKnownHosts.data | string | `""` | SSH known hosts file content for Git repository host key verification. |
 | global.storageClassName | string | `""` | Storage class name for the PVCs. Keep empty to use the default storage class. |
 | kv | object | `{"fsGroup":"","image":{"image":"quay.io/sclorg/redis-7-c9s","pullPolicy":"","tag":"20250108"},"loglevel":"warning","maxmemory":"1gb","maxmemoryPolicy":"allkeys-lru","passwordSecretName":""}` | Key-Value Store Configuration |
--- a/deploy/helm/flightctl/templates/certs/flightctl-certs-from-certmanager.yaml
+++ b/deploy/helm/flightctl/templates/certs/flightctl-certs-from-certmanager.yaml
@@ -18,7 +18,7 @@ spec:
   selfSigned: {}
 
 ---
-# FlightCtl CA Certificate (only created if user hasn't provided flightctl-ca)
+# Flight Control CA Certificate (only created if user hasn't provided flightctl-ca)
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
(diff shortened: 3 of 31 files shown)


sk-ilya:
@coderabbitai review

pkg/queues/redis_provider.go
```diff
@@ -643,16 +661,27 @@ func (r *redisQueue) consumeOnce(ctx context.Context, handler ConsumeHandler) er
...
+					}
+					delay := calculateBackoff(attempt, infraRetryConfig)
+					time.Sleep(delay)
+					continue
 				}
-				time.Sleep(delay)
-				continue
+				// All retries exhausted, attempt to recreate the consumer group
+				r.log.WithError(err).Info("consumer group not found after retries, attempting to recreate")
+				if createErr := r.ensureConsumerGroup(ctx); createErr != nil {
```
ori-amizur: Why does consume perform group creation ? Doesn't seem correct

pkg/queues/redis_provider.go
```diff
@@ -483,6 +483,24 @@ type redisQueue struct {
 	retryConfig  RetryConfig
 }
 
+// ensureConsumerGroup recreates the consumer group if it was lost (e.g., after Redis restart).
+func (r *redisQueue) ensureConsumerGroup(ctx context.Context) error {
```
ori-amizur, line 490: Please verify that all places that perform group creation use this function

pkg/queues/redis_provider.go
sk-ilya: Since it is the one reading it (and detecting that the queue does not exist), it would make sense to self-heal

pkg/queues/redis_provider.go
sk-ilya: Done

pkg/queues/redis_provider.go
```diff
@@ -483,6 +482,30 @@ type redisQueue struct {
...
+	// This is idempotent - no need to synchronize
+	err := r.client.XGroupCreateMkStream(ctx, r.name, r.groupName, "0").Err()
+	if err == nil {
```
hexfusion: can we try to align with the standard golang error handling, eval err !=nil first and return early. it makes error handling more predictable and easier to maintain. 

```
	if err := r.client.XGroupCreateMkStream(ctx, r.name, r.groupName, "0").Err(); err != nil {
		if allowExists && strings.Contains(err.Error(), "BUSYGROUP") {
			return nil
		}
		return fmt.Errorf("failed to create consumer group: %w", err)
	}
	[...]
```

pkg/queues/redis_provider.go
```diff
@@ -986,12 +1032,23 @@ func (r *redisQueue) ProcessTimedOutMessages(ctx context.Context, timeout time.D
...
 	pending, err := r.client.XPending(ctx, r.name, r.groupName).Result()
 	if err != nil {
-		return 0, fmt.Errorf("failed to get pending messages: %w", err)
+		if strings.Contains(err.Error(), "NOGROUP") {
+			if ensureErr := r.ensureConsumerGroup(ctx); ensureErr != nil {
+				return 0, ensureErr
+			}
+			pending, err = r.client.XPending(ctx, r.name, r.groupName).Result()
+		}
+		if err != nil {
```
hexfusion, line 1036: nit: you have scope issue here the err at line 1040 can be replaced by an error on 1046 please revisit

pkg/queues/redis_provider.go
```diff
@@ -1006,7 +1063,22 @@ func (r *redisQueue) ProcessTimedOutMessages(ctx context.Context, timeout time.D
...
+			pendingMsgs, err = r.client.XPendingExt(ctx, &redis.XPendingExtArgs{
+				Stream: r.name,
+				Group:  r.groupName,
+				Start:  "-",
+				End:    "+",
+				Count:  pending.Count,
+				Idle:   timeout,
+			}).Result()
+		}
+		if err != nil {
```
hexfusion: when would this ever not be true

pkg/queues/redis_provider.go
```diff
@@ -483,6 +483,24 @@ type redisQueue struct {
...
+	}
+
+	// BUSYGROUP means it already exists - treat as success
+	if strings.HasPrefix(err.Error(), "BUSYGROUP") {
+		r.log.Debug("consumer group already exists")
+		return nil
+	}
+
+	return fmt.Errorf("failed to create consumer group: %w", err)
+}
```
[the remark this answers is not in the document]
avishayt: 0 makes sense.  After restart, all messages are wiped (no persistence).  We replay only messages that weren't processed.

pkg/queues/redis_provider.go
ori-amizur: It still doesn't make sense.  Same if I would try to read a file, and if doesn't exist I will create it.

pkg/queues/redis_provider.go
sk-ilya: Resolved

pkg/queues/redis_provider.go
avishayt: This isn't necessary anyway - it should be sufficient to recreate the consumer group in ProcessTimedOutMessages

sk-ilya: Done

PR #2233, opened 2025-12-15
EDM-2562: Apply override to UI tags only for releases (#2220)

kkyrazis:
(cherry picked from commit 098d7ea07d190ee5aca3492feb1ff77af9715065)


celdrake:
We'll do batch backports instead of individual ones.

PR #2235, opened 2025-12-16
EDM-2569: helm: flightctl api discovery

rawagner:
bp of https://github.com/flightctl/flightctl/pull/2218


rawagner:
Included in https://github.com/flightctl/flightctl/pull/2238

PR #2259, opened 2025-12-17
EDM-2820: Agent retries indenfinetly if invalid image is parsed by prefetch manager

kkyrazis:
The dependency manager now respects the poll config before retrying. 

Podman clients now return their last retriable errors if those fail

StdError's now return the reason that they were marked as retriable. 

Messages now look like:
```yaml
  updated:
    info: '0/2 images complete, retrying: 192.168.68.74:5002/flightctl-device:base: connection refused, and 1 more pending'
    status: Updating
```

### DIFF
--- a/deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
+++ b/deploy/podman/flightctl-db-migrate/flightctl-db-migrate.container
@@ -14,7 +14,6 @@ Environment=DB_MIGRATION_USER=flightctl_migrator
 Secret=flightctl-postgresql-migrator-password,type=env,target=DB_PASSWORD
 Secret=flightctl-postgresql-migrator-password,type=env,target=DB_MIGRATION_PASSWORD
 Volume=/etc/flightctl/flightctl-db-migrate/config.yaml:/root/.flightctl/config.yaml:ro,z
-Volume=/etc/flightctl/service-config.yaml:/etc/flightctl/service-config.yaml:ro,z
 Volume=/etc/flightctl/pki/db:/root/.flightctl/certs/db:ro,z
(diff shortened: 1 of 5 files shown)


internal/agent/device/dependency/dependency.go
```diff
@@ -17,11 +17,11 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/device/resource"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/poll"
 )
 
 const (
-	queueNextItemAfter = 5 * time.Second
```
hexfusion, line 23: :+1: note to self how does CI not fail on this

internal/agent/device/dependency/dependency.go
```diff
@@ -329,6 +335,10 @@ func (m *prefetchManager) processTarget(ctx context.Context, target string) {
 		}
 		if err := m.pull(ctx, target, task); err != nil {
 			if errors.IsRetryable(err) {
+				retries++
+				m.log.Warnf("Retrying prefetch for %s (attempt %d): %v", target, retries+1, err)
```
hexfusion, line 339: :+1:

internal/agent/device/errors/errors.go
```diff
@@ -102,8 +103,49 @@ var (
...
+	}
+
+	return err.Error()
+}
+
 // TODO: tighten up the retryable errors ideally all retryable errors should be explicitly defined
 func IsRetryable(err error) bool {
+	if err == nil {
+		return false
+	}
```
hexfusion, line 148: :sweat_smile:

PR #2265, opened 2025-12-18
EDM-2814: Use db-specific config, instead of service-config. Fix wait…

rawagner:
… script for the new format


hexfusion:
@rawagner did this miss backport?

hexfusion:
backported via https://github.com/flightctl/flightctl/pull/2270

PR #2272, opened 2025-12-18
EDM-2838: support branding for flightctl-pam-issuer

asafbennatan:
image

image

image

## Summary by CodeRabbit

* **New Features**
  * Customizable login branding: issuer name, light/dark logos, and per-theme colors.
  * Light/dark theming with a theme selector, persistent preference, and dynamic logo switching.

* **Documentation**
  * Updated auth configuration docs to document branding options, defaults, and theming behavior.

* **Chores**
  * Added built-in fallback branding assets and simplified login error handling (removed standalone static error page).


internal/auth/oidc/pam/templates/login_form.html
```diff
@@ -65,17 +96,202 @@
...
+            
+            // Update system preference listener
+            if (theme === 'system') {
+                window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', e => {
+                    if (getStoredTheme() === 'system') {
+                        applyTheme('system');
+                    }
+                });
+            }
+        }
```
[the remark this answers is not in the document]
asafbennatan: fixed

docs/user/installing/configuring-auth/auth-pam.md
```diff
@@ -103,6 +111,14 @@ auth:
...
 | `accessTokenExpiration` | Expiration duration for access tokens and ID tokens (e.g., `1h`, `30m`) | `1h` |
 | `refreshTokenExpiration` | Expiration duration for refresh tokens (e.g., `168h`, `720h`) | `168h` |
+| `branding` | Branding configuration for the login form | See defaults below |
+| `branding.issuerName` | Display name shown on the login form | `"Flight Control"` |
+| `branding.logoLight` | Logo image for light theme (URL or base64 data URI). If empty, uses default embedded logo | Empty (uses default) |
+| `branding.logoDark` | Logo image for dark theme (URL or base64 data URI). If empty, uses default embedded logo | Empty (uses default) |
+| `branding.lightPrimaryColor` | Primary color for light theme (hex color code) | `"#007bff"` |
+| `branding.lightSecondaryColor` | Secondary color for light theme (hex color code) | `"#666"` |
+| `branding.darkPrimaryColor` | Primary color for dark theme (hex color code) | `"#4a9eff"` |
+| `branding.darkSecondaryColor` | Secondary color for dark theme (hex color code) | `"#b0b0b0"` |
```
hexfusion, line 121: appreciate the PR! do we want all these knobs here I get we need more control but I wonder if @celdrake or @rawagner has ideas. when I first tagged this templating I didn't consider the loss of control/review by UI/UX team. I would really like to wait for additional feedback.

asafbennatan: yes , BTW as far as i know they are using a RHEM_UI env to control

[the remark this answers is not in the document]
asafbennatan: fixed

asafbennatan: the problem was that we didnt want to have RHEM reference in the code ,
this means that before RHEM meant :
logo=XXX
color=YYY
etc

now we cannot do that and we need these knobs so we can actually express RHEM theme without having the code mentioning it

internal/auth/oidc/pam/templates/login_form.html
```diff
@@ -131,9 +362,19 @@
...
+        <select id="theme-select" onchange="handleThemeChange(event)">
+            <option value="system">System</option>
+            <option value="light">Light</option>
+            <option value="dark">Dark</option>
+        </select>
+    </div>
     <div class="login-container">
         <div class="header">
-            <h1>Flight Control</h1>
+            {{if or .LogoLight .LogoDark}}
```
celdrake, line 374: I'd suggest that for the specific branding of Flight Control, we can use the same logo with the blue color for both light and dark modes.

Then if someone configures one of LogoLight/LogoDark, and the other is empty, fallback to the custom logo they configured. So that if the same logo can be used for both, there's no need to configure it twice (since they can be pretty large in size).

docs/user/installing/configuring-auth/auth-pam.md
fzdarsky: We should align on a common pattern across this and our UI. I don't see why this should be runtime-configurable. Why not make it a build-time selection?

fzdarsky:
Is this using Patternfly like the main UI? We wouldn't want this to be visually inconsistent.

docs/user/installing/configuring-auth/auth-pam.md
asafbennatan: In fc ui you can switch themes (at runtime) , to be consistent I did the same , also keep in mind that we are serving a single html , if we want something more complex we would need a dedicated nginx serving just pam issuer's UI

fzdarsky: True, you can switch between up- and downstream themes at runtime (I suspect to enable testing / demo theming). But the two themes are fully defined and embedded. Why make this user-configurable and thereby create API baggage without need?

asafbennatan: The switch is between light and dark , this is unrelated to up/down stream , for rhem theme you will need to static configure them dark and light the same for normal ui I assume (I assume that because fc light and dark themes both contain different versions of fc logo)

PR #2275, opened 2025-12-19
EDM-2805: agent: improve reconciliation speed

hexfusion:
we do a lot of waiting because we never cleaned up spec fetch via long poll implementation. plus reducing log level

## Summary by CodeRabbit

* **Deprecations**
  * `spec-fetch-interval` is deprecated; use `status-update-interval` instead.

* **Improvements**
  * Agent now uses long-polling for faster, near-immediate spec updates and status sync.

* **Documentation**
  * Updated install and device simulator docs to reflect the deprecation and long-polling behavior.
  * Clarified agent architecture to show new Publisher/Engine roles.

* **Tests**
  * Test suites updated to use the new poll-based timing configuration.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -474,6 +474,7 @@ PAM
 alice
 bob
 charlie
+MaxAge
  - docs/user/installing/configuring-auth/auth-oauth2.md
 GitHub
 github.com
--- a/api/v1beta1/application_validation.go
+++ b/api/v1beta1/application_validation.go
@@ -13,7 +13,8 @@ import (
 type applicationValidator interface {
 	// ValidateContents returns any errors for the supplied individual contents. It is expected for a caller
 	// to validate the contents before validating the application so that the validator can build state.
-	ValidateContents(path string, contents []byte) []error
+	// When fleetTemplate is true, template expressions like {{ .metadata.labels.x }} are allowed in image references.
+	ValidateContents(path string, contents []byte, fleetTemplate bool) []error
 	// Validate returns any errors for the application evaluated as a whole.
 	Validate() []error
 }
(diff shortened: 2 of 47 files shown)


internal/agent/device/spec/queue.go
```diff
@@ -125,12 +125,12 @@ func (m *queueManager) Next(ctx context.Context) (*v1beta1.Device, bool) {
 	}
 	version := item.Version
 
-	m.log.Debugf("Evaluating template version: %d", version)
+	m.log.Tracef("Evaluating template version: %d", version)
```
hexfusion, line 128: note: in debug these are like a firehose

internal/agent/device/spec/publisher_test.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/client"
 	"github.com/flightctl/flightctl/internal/agent/device/errors"
 	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/flightctl/flightctl/pkg/poll"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
-	"k8s.io/apimachinery/pkg/util/wait"
```
hexfusion, line 15: note: death to wait

internal/agent/device/dependency/dependency.go
```diff
@@ -552,11 +552,15 @@ func (m *prefetchManager) cleanupStaleTasks(seenTargets map[string]struct{}) {
...
-	m.log.Debug("Prefetch cleanup: canceling active task and draining queue")
-
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if len(m.tasks) == 0 {
+		return
+	}
+
+	m.log.Debugf("Prefetch cleanup: canceling %d active tasks", len(m.tasks))
```
hexfusion, line 562: note: lets really read our logs in debug make sure things like this get cleaned up. AKA do not log if nothing happens

PR #2283, opened 2025-12-22
EDM-2854: Add ttlSecondsAfterFinished to completed jobs in helm

itzikb-redhat:
Make sure there are no leftovers.

## Release Notes

* **New Features**
  * Added Time To Live (TTL) configuration for database migration jobs, enabling automatic cleanup of finished jobs after a configurable duration (disabled by default).


sk-ilya:
in case of the migration job it's [not as straightforward](https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/templates/_helpers.tpl#L459-L467). we need another implementation of the wait-for-migration  init container that will not depend on the cluster resources

PR #2284, opened 2025-12-22
EDM-2860: setting CreatedBySuperAdmin annotation when replace is call…

asafbennatan:
…ed without existing resource (as it is when we create from CLI)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replacement flow for authentication providers: stricter name-consistency validation and correct handling of non-existent resources by routing them through creation logic.
  * Fixed super-admin creation annotation behavior so it is applied only when appropriate.

* **Tests**
  * Added integration tests covering replacement and creation flows for admin and non-admin contexts to ensure annotation and validation behavior.

### DIFF
--- a/internal/service/authprovider.go
+++ b/internal/service/authprovider.go
@@ -91,6 +91,21 @@ func applyAuthProviderDefaults(spec *api.AuthProviderSpec) error {
 	return nil
 }
 
+// handleSuperAdminAnnotation checks if the request is from a super admin and sets the annotation if needed.
+// Returns true if the auth provider was created by a super admin, false otherwise.
+func (h *ServiceHandler) handleSuperAdminAnnotation(ctx context.Context, authProvider *api.AuthProvider) bool {
+	mappedIdentity, ok := contextutil.GetMappedIdentityFromContext(ctx)
+	createdBySuperAdmin := ok && mappedIdentity.IsSuperAdmin()
+
+	if createdBySuperAdmin {
+		// Clear user-provided annotations and set our annotation
+		authProvider.Metadata.Annotations = lo.ToPtr(map[string]string{
+			api.AuthProviderAnnotationCreatedBySuperAdmin: "true",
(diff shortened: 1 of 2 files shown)


internal/service/authprovider.go
```diff
@@ -196,6 +196,26 @@ func (h *ServiceHandler) ReplaceAuthProvider(ctx context.Context, orgId uuid.UUI
...
+			authProvider.Metadata.Annotations = lo.ToPtr(map[string]string{
+				api.AuthProviderAnnotationCreatedBySuperAdmin: "true",
+			})
+			// Use fromAPI=false to preserve annotations
+			if authProvider.Metadata.Name == nil {
+				return nil, api.StatusBadRequest("metadata.name is required")
+			}
+			if name != *authProvider.Metadata.Name {
+				return nil, api.StatusBadRequest("resource name specified in metadata does not match name in path")
+			}
```
avishayt: Aren't these checks always relevant?  (Not just when createdBySuperAdmin is true?)

internal/service/authprovider.go
```diff
@@ -196,6 +196,26 @@ func (h *ServiceHandler) ReplaceAuthProvider(ctx context.Context, orgId uuid.UUI
...
+			result, err := h.store.AuthProvider().CreateWithFromAPI(ctx, orgId, &authProvider, false, h.callbackAuthProviderUpdated)
+			return result, StoreErrorToApiStatus(err, true, api.AuthProviderKind, authProvider.Metadata.Name)
```
avishayt: You have the same code in CreateAuthProvider - factor out?

internal/service/authprovider.go
asafbennatan: fixed

internal/service/authprovider.go
asafbennatan: fixed

PR #2294, opened 2025-12-25
EDM-2878: EDM-2879 , support for OCI repositories including repotester (connectivity) implementation

asafbennatan:
## Summary by CodeRabbit

* **New Features**
  * OCI registry added as a repository type with registry URL, access modes, optional Docker-style credentials, and connectivity testing.

* **Behavior / Security**
  * Sensitive fields (passwords, tokens, private keys, passphrases) are masked for OCI, HTTP, and SSH repository specs.

* **Refactor**
  * Unified repo-type dispatch and generalized tester selection; selectors and listing extended to include OCI.

* **Validation**
  * Enhanced hostname/URL validation (host/FQDN with optional port) and explicit per-type validation for OCI/HTTP/Git.

* **Tests**
  * Expanded unit, integration, and e2e tests plus a mock OCI registry for auth scenarios.

### DIFF
--- a/internal/store/model/selectors_test.go
+++ b/internal/store/model/selectors_test.go
@@ -17,17 +17,6 @@ type selectorTest struct {
 	Selectors     selectorToTypeMap
 }
 
-// combineSelectorMaps merges multiple selector maps into one
-func combineSelectorMaps(maps ...selectorToTypeMap) selectorToTypeMap {
-	result := make(selectorToTypeMap)
-	for _, m := range maps {
-		for k, v := range m {
-			result[k] = v
-		}
-	}
-	return result
-}
-
 func TestModelSchemaSelectors(t *testing.T) {
 	testSelectors := []selectorTest{
 		{"status", &api.DeviceStatus{}, deviceStatusSelectors},


api/v1beta1/openapi.yaml
```diff
@@ -4317,13 +4317,47 @@ components:
...
+        mapping:
+          git: "#/components/schemas/GenericRepoSpec"
+          http: "#/components/schemas/HttpRepoSpec"
+          oci: "#/components/schemas/OciRepoSpec"
       oneOf:
         - $ref: "#/components/schemas/GenericRepoSpec"
         - $ref: "#/components/schemas/HttpRepoSpec"
         - $ref: "#/components/schemas/SshRepoSpec"
+        - $ref: "#/components/schemas/OciRepoSpec"
     RepositoryStatus:
```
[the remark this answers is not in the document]
asafbennatan: it's a transport mechanism for git repos. An SSH-accessed Git repo would correctly use type: git with sshConfig for the transport configuration.


they are structurally distinguished so its not a problem

internal/tasks/repotester.go
```diff
@@ -158,3 +195,168 @@ func (r *HttpRepoTester) TestAccess(repository *api.Repository) error {
...
+		return "", fmt.Errorf("failed to parse auth realm URL: %w", err)
+	}
+
+	// Build query parameters
+	query := authURL.Query()
+	if service != "" {
+		query.Set("service", service)
+	}
+
+	query.Set("scope", "repository:"+uuid.New().String()+":pull")
```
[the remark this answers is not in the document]
asafbennatan: tested registries allow puting any non-existent repo in the scope for testing read access to global repositories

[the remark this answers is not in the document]
asafbennatan: registry:catalog:* is not a thing , and not using any scope will fail in some registries such as quay.io

api/v1beta1/util.go
```diff
@@ -376,28 +376,44 @@ func (r *Repository) HideSensitiveData() error {
...
-	if err != nil {
-		gitHttpSpec, err := spec.GetHttpRepoSpec()
-		if err == nil {
-			hideValue(gitHttpSpec.HttpConfig.Password)
-			hideValue(gitHttpSpec.HttpConfig.TlsKey)
-			hideValue(gitHttpSpec.HttpConfig.TlsCrt)
-			if err := spec.FromHttpRepoSpec(gitHttpSpec); err != nil {
-				return err
-			}
+	// Try OCI spec first
```
sk-ilya: Any chance the discriminator can be used here?
e.g.:
```go
 	disc, err := spec.Discriminator()
	if err != nil {
		return err
	}

	switch disc {
	case "oci":
		oci, err := spec.AsOciRepoSpec()
		if err != nil {
			return err
		}
		hideValue(oci.Password)
		return spec.FromOciRepoSpec(oci)

	case "http":
   <...>
<...>
```

api/v1beta1/validation.go
```diff
@@ -674,7 +674,23 @@ func (r *Repository) Validate() []error {
 		allErrs = append(allErrs, validateSshConfig(&sshRepoSpec.SshConfig)...)
 	}
 
-	if genericErr != nil && httpErr != nil && sshErr != nil {
+	// Validate OciRepoSpec
+	ociRepoSpec, ociErr := r.Spec.GetOciRepoSpec()
+	if ociErr == nil {
+		allErrs = append(allErrs, validation.ValidateString(&ociRepoSpec.Url, "spec.url", 1, 2048, nil, "")...)
```
sk-ilya: These need to be reflected in the OpenAPI spec

api/v1beta1/validation.go
```diff
@@ -674,7 +674,23 @@ func (r *Repository) Validate() []error {
 		allErrs = append(allErrs, validateSshConfig(&sshRepoSpec.SshConfig)...)
 	}
 
-	if genericErr != nil && httpErr != nil && sshErr != nil {
+	// Validate OciRepoSpec
```
sk-ilya: Same as above, consider
```go
disc, err := r.Spec.Discriminator()
	if err != nil {
		return err
	}

	switch disc {
<...>

```

internal/tasks/repotester.go
```diff
@@ -158,3 +195,168 @@ func (r *HttpRepoTester) TestAccess(repository *api.Repository) error {
...
+	}
+	baseURL.Path = strings.TrimSuffix(baseURL.Path, "/")
+
+	v2URL := baseURL.JoinPath("/v2/").String()
+
+	client := &http.Client{
+		Timeout: 30 * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				MinVersion: tls.VersionTLS12,
```
sk-ilya: I guess we may want to allow custom `RootCAs` here... but this is out of this PR scope

internal/tasks/repotester.go
```diff
@@ -158,3 +195,168 @@ func (r *HttpRepoTester) TestAccess(repository *api.Repository) error {
...
+			},
+		},
+	}
+
+	// Step 1: Call /v2/ without auth
```
sk-ilya: What are these steps based on?


The [OCI spec](https://specs.opencontainers.org/distribution-spec/?v=v1.1.0) only mentions that `GET /v2/` **may** return 401, but the authentication flow is specific to the actual registry implementation. But docker is [more specific](https://distribution.github.io/distribution/spec/api/?api-version-check#api-version-check) here with

> if a 401 Unauthorized response is returned, the client should take action based on the contents of the “WWW-Authenticate” header and try the endpoint again.

 
Maybe the auth type should be supplied inside the `OciRepoSpec` so that it will be possible to know in advance without guessing? 

Something like:
```yaml
OciRepoSpec:
  url: ...
  auth:
    scheme: Docker-Distribution
    tokenRealm: https://quay.io/v2/auth
    username: ...
    password: ....

     
```

internal/tasks/repotester.go
sk-ilya, line 365: This doesn't feel right... Looks like a quay-specific quirk that can change in the future

api/v1beta1/validation.go
fzdarsky: Use [`ValidateHostnameOrFQDN()`](https://github.com/flightctl/flightctl/blob/main/internal/util/validation/formats.go#L33) as it's more specific. Potentially add `ValidateHostnameWithOptionalPort()` instead.

api/v1beta1/openapi.yaml
```diff
@@ -4317,13 +4317,47 @@ components:
...
         - url
         - type
+    OciRepoSpec:
+      type: object
+      description: OCI container registry specification.
+      additionalProperties: false
+      properties:
+        url:
+          type: string
+          description: 'The OCI registry URL (e.g., quay.io, registry.redhat.io).'
```
fzdarsky: This is not a URL but a hostname (plus optionally port).

internal/tasks/repotester.go
asafbennatan: OCI spec does not mention an auth type at all , it is based on docker spec though , in practice this is just the way these are implemented , if a registry is not implemented that way it is not going to work with docker , podman and other CLIs.

P.S - your suggestion assume that the entire mechanism is the same but the only issue is that the authWWW does not behave the same , but that mechanism itself is not described in OCI spec

i have added a discriminator type to the authtype so this can be extended in the future

internal/tasks/repotester.go
asafbennatan: this is not quay specific quirk , pulling from an unknown repo is allowed across all repos i have checked , there is no other way to get this working correctly , the only spec supported way would be to have the user to enter the repository they want to validate - IMO we shouldnt bog down the user with that

api/v1beta1/validation.go
asafbennatan: none of the other repo urls is validated via the openapi spec, wouldnt it be weird if only this specific one is validated differently ? or do you suggested changing how all of them are validated ? P.S this kind of way is done throughout validation.go (so there are other non repository related instances where this happens and not via openapi validation)

api/v1beta1/util.go
asafbennatan: done , although it is not as nice as it could be due to how SSHRepoSpec and GenericRepoSpec were implemented - they do not have a clear discriminator

api/v1beta1/validation.go
asafbennatan: same as above

internal/tasks/repotester.go
asafbennatan: added support

api/v1beta1/openapi.yaml
asafbennatan: done

api/v1beta1/validation.go
asafbennatan: done

internal/tasks/repotester.go
```diff
@@ -22,16 +29,50 @@ type API interface {
...
+	// Fall back to generic spec for git/http
+	repoSpec, _ := repository.Spec.GetGenericRepoSpec()
+	switch repoSpec.Type {
+	case api.Http:
+		return httpRepoTester
+	case api.Git:
+		return gitRepoTester
+	default:
+		return nil
 	}
```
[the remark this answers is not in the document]
asafbennatan: fixed

api/v1beta1/validation.go
sk-ilya: Ideally, we should validate all of them via OpenAPI for consistency, since it’s the API contract.

`validation.go` is still needed for cases that are too complex to express in the spec, and for re-validating after server-side changes (for example, applying a `PATCH`)

asafbennatan: due to Frank's comment this is no longer a url or something that can be validated in openapi anyway now , and i dont see a reason that this PR will change existing validation for other repo types

internal/tasks/repotester.go
sk-ilya: Maybe https://pkg.go.dev/github.com/containers/image/v5/docker#CheckAuth worth considering in this case?

PR #2305, opened 2025-12-30
EDM-2920: Add Pruning Configuration Support

ori-amizur:
## Summary by CodeRabbit

* **New Features**
  * Pruning configuration added: a configurable image-pruning setting is now supported and defaults to disabled.
  * Drop-in configuration files are applied in deterministic lexical order so overrides behave consistently.

* **Tests**
  * Added tests covering pruning loading, multiple drop-ins, override precedence, and handling of invalid drop-ins.

### DIFF
--- a/internal/agent/config/config.go
+++ b/internal/agent/config/config.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"sort"
 	"time"
 
 	"github.com/flightctl/flightctl/internal/agent/client"
@@ -16,6 +17,7 @@ import (
 	baseclient "github.com/flightctl/flightctl/internal/client"
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/util"
+	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
 )
@@ -136,6 +138,9 @@ type Config struct {
 	// ProfilingEnabled turns on the loopback-only pprof server for local debugging.
 	ProfilingEnabled bool `json:"profiling-enabled,omitempty"`
 
+	// ImagePruning holds all image/artifact pruning-related configuration
+	ImagePruning ImagePruning `json:"image-pruning,omitempty"`
+
 	readWriter fileio.ReadWriter
 }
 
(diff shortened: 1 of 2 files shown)


internal/agent/config/config.go
```diff
@@ -150,6 +154,12 @@ type TPM struct {
 	StorageFilePath string `json:"storage-file-path,omitempty"`
 }
 
+type Pruning struct {
+	// Enabled controls whether automatic pruning is enabled.
+	// Default: false
+	Enabled bool `json:"enabled,omitempty"`
```
hexfusion: what other configurations do we expect here in the future that would justify a root level key vs pruneEnabled. pruning is also somewhat vauge should this be more descriptive?

ori-amizur: Currently there is no need for other configurations - although I can think of some:
- Invocation threshold: When disk is occupied more than a specific threshold.
- Invocation time: Invoke only on 2PM

Is there a name which seems more suitable?

internal/agent/config/config.go
```diff
@@ -150,6 +154,12 @@ type TPM struct {
 	StorageFilePath string `json:"storage-file-path,omitempty"`
 }
 
+type Pruning struct {
+	// Enabled controls whether automatic pruning is enabled.
+	// Default: true
+	Enabled bool `json:"enabled,omitempty"`
+}
```
ori-amizur: Changed the default to false

internal/agent/config/config.go
```diff
@@ -150,6 +155,12 @@ type TPM struct {
 	StorageFilePath string `json:"storage-file-path,omitempty"`
 }
 
+type Pruning struct {
```
kkyrazis: Is the intention for this config to relate to all types of pruning? Currently we're only considering image/artifact pruning, but if in the future we needed to prune some other resource would we add a separate option for that, or just reuse this one?

ori-amizur: Currently, I am not aware of any additional required pruning types

hexfusion: what about `ImagePruning` I think it makes the intent much more clear

internal/agent/config/config_test.go
```diff
@@ -63,3 +65,215 @@ func TestParseConfigFile_NoFile(t *testing.T) {
...
+			setupFiles: func(t *testing.T, configDir string) string {
+				configFile := filepath.Join(configDir, "config.yaml")
+				content := `enrollment-service:
+  service:
+    server: https://enrollment.endpoint
+    certificate-authority-data: abcd
+  authentication:
+    client-certificate-data: efgh
+    client-key-data: ijkl
+spec-fetch-interval: 0m10s
```
hexfusion: minor: spec-fetch.. is deprecated we should not add to new tests

internal/agent/config/config_test.go
```diff
@@ -63,3 +65,215 @@ func TestParseConfigFile_NoFile(t *testing.T) {
...
```
hexfusion: same for all instances

internal/agent/config/config.go
ori-amizur: Changed

internal/agent/config/config_test.go
ori-amizur: Done

PR #2308, opened 2026-01-01
EDM-2872: Image Export Api

asafbennatan:
## Summary by CodeRabbit

* **New Features**
  * Added ImageExport resource for exporting container images in multiple formats (ISO, QCOW2, VMDK).
  * Implemented ImageExport API endpoints for list, create, retrieve, and delete operations.
  * Added CLI support for managing ImageExports (apply, delete, get).
  * Images can be sourced from existing ImageBuilds or external registries.

* **Documentation**
  * Added example YAML manifests demonstrating ImageExport usage.


internal/imagebuilder_api/service/imageexport.go
```diff
@@ -0,0 +1,178 @@
...
+				}
+			}
+		case string(api.ImageExportSourceTypeImageReference):
+			source, err := imageExport.Spec.Source.AsImageReferenceSource()
+			if err != nil {
+				errs = append(errs, errors.New("invalid imageReference source"))
+			} else {
+				if source.Repository == "" {
+					errs = append(errs, errors.New("spec.source.registryName is required for imageReference source type"))
+				}
```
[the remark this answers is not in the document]
asafbennatan: fixed

api/v1beta1/imagebuilder/openapi.yaml
```diff
@@ -392,3 +564,264 @@ components:
...
+    ImageExportSpec:
+      type: object
+      description: ImageExportSpec describes the specification for an image export.
+      properties:
+        source:
+          $ref: '#/components/schemas/ImageExportSource'
+        destination:
+          $ref: '#/components/schemas/ImageExportDestination'
+        formats:
+          type: array
```
avishayt: Would it be simpler to have a single format per imageExport task?  This way it's clearer to the user regarding what failed, what the error message refers to, and makes cleaning up easier (e.g., if we fail on the second one, do we need to clean the first one?)

api/v1beta1/imagebuilder/openapi.yaml
```diff
@@ -392,3 +564,264 @@ components:
...
+        conditions:
+          type: array
+          description: Current conditions of the ImageExport.
+          items:
+            $ref: '../openapi.yaml#/components/schemas/Condition'
+        phase:
+          $ref: '#/components/schemas/ImageExportPhase'
+        formats:
+          type: array
+          description: Per-format status tracking.
```
avishayt: If we have a single format per task then this also becomes simpler and clearer

api/v1beta1/imagebuilder/openapi.yaml
asafbennatan: at the api level this doesnt make sense IMO since the the bootc tool also receives multiple formats - i imagine that it does it faster then running each one individually

internal/cli/get.go
```diff
@@ -430,38 +430,70 @@ func (o *GetOptions) createMainAPIFetchers(ctx context.Context, kind ResourceKin
...
+			if err != nil {
+				return nil, err
+			}
+			if err := validateImageBuilderResponse(response); err != nil {
+				return nil, err
+			}
+			return response, nil
 		}
-		return response, nil
 	}
```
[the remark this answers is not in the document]
asafbennatan: fixed

api/v1beta1/imagebuilder/openapi.yaml
avishayt: Good job on checking the bootc-image-builder support - it definitely is faster to build multiple formats at once. The catch is that while efficient, exposing that array creates a "Split Brain" state where a resource can be half-failed/half-successful. That breaks standard K8s observability since clients can't just trust the top-level Status and have to parse nested logic. To keep our status reporting atomic and retry logic simple, we really need to stick to a "One Resource = One Artifact" model. Let's refactor spec.formats to a single spec.format enum; the operational simplicity outweighs the computational cost, especially since most users will likely just pick a single format anyway.

asafbennatan: fixed

api/v1beta1/imagebuilder/openapi.yaml
asafbennatan: fixed

api/v1beta1/imagebuilder/openapi.yaml
```diff
@@ -392,3 +564,228 @@ components:
...
+          description: Current conditions of the ImageExport.
+          items:
+            $ref: '../openapi.yaml#/components/schemas/Condition'
+        phase:
+          $ref: '#/components/schemas/ImageExportPhase'
+        format:
+          $ref: '#/components/schemas/ExportFormatType'
+        tagSuffix:
+          type: string
+          description: Optional suffix to append to the output tag for this format.
```
avishayt: Why do we need these in the status?

api/v1beta1/imagebuilder/openapi.yaml
```diff
@@ -392,3 +564,228 @@ components:
...
+          format: date-time
+          description: The time the export completed.
+        lastSeen:
+          type: string
+          format: date-time
+          description: The last time the export was seen (heartbeat).
+        nextRetryAt:
+          type: string
+          format: date-time
+          description: The next time to retry the export (for delayed processing).
```
avishayt: I don't remember discussing retries - I thought the task fails and a new one can be created?

api/v1beta1/imagebuilder/openapi.yaml
asafbennatan: removed

api/v1beta1/imagebuilder/openapi.yaml
asafbennatan: removed

internal/imagebuilder_api/store/imageexport.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	} else {
+		countQuery, err := flightctlstore.ListQuery(&ImageExport{}).Build(ctx, s.db.WithContext(ctx), orgId, listParams)
+		if err == nil {
+			numRemainingVal = flightctlstore.CountRemainingItems(countQuery, continueValues, listParams)
+		}
+	}
+
+	return flightctlstore.BuildContinueString(continueValues, numRemainingVal), &numRemainingVal
+}
```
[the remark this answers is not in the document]
asafbennatan: this is how we have it in other places as well

PR #2312, opened 2026-01-05
NO-ISSUE: refactor config

asafbennatan:
This PR re-organizes the config structure. as we are now working on the image builder feature i notice some places where adding the new standalone service requires a lot of changes and accommodations (or extending existing code in a manner that will dig further technical debt) , while this is not required for the image builder feature to move forward (as i have worked around that) i suspect that we will have more standalone services and in this case we should split the monolith config and set proper guidelines for multi-service architecture.

couple of points:
1. service-config.yaml and values.yaml remain unchanged and the only user-facing config files.
2. each service has its own config struct
3. common structs such as db/kv/auth have been added to common config and reused by individual service configs
4. we can also extract http config and tls config but i have not done that yet
5. internal services were changed to receive only the required configuration object 
6. the config.go monolith removed

additional refactor tasks that makes sense to do to for easier service addition (but are not the scope of this feature):
1. the monolith store and service files should be decomposed and removed , internal files should use only the services they really need - not the entire store/service monolith
2. entire code should be re-org to common,service1,service2..... packages , otherwise once we add more services the code base will start growing dependencies between files in an uncontrolled manner 
3. move to golang workspaces to create better separation between each service
4. cookie cutter like template for adding new services ?

## Summary by CodeRabbit

* **Refactor**
  * Configuration reorganized into service-specific configs with granular accessors (DB, KV, tracing, metrics, auth), and services updated to consume focused config pieces.

* **New Features**
  * New default configs, improved environment overrides, and centralized certificate loading/validation for safer TLS handling.
  * New dedicated config models for auth, KV, database, metrics and tracing with sensible defaults.

* **Tests**
  * Test suites updated to use the new config models and helpers.


internal/config/imagebuilderapi/config.go
```diff
@@ -0,0 +1,186 @@
...
+	if c == nil {
+		return nil
+	}
+
+	// Create a shallow copy
+	sanitized := *c
+	sanitized.Auth = c.Auth.SanitizeForLogging()
+
+	return &sanitized
+}
```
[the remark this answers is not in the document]
asafbennatan: they are sanitized by using secure string , auth doesnt do that since it uses the same objects that are saved in the DB

internal/config/common/service.go
```diff
@@ -0,0 +1,40 @@
...
+
+// NewDefaultBaseService returns a default base service configuration.
+func NewDefaultBaseService() *BaseServiceConfig {
+	return &BaseServiceConfig{
+		Address:               ":3443",
+		LogLevel:              "info",
+		HttpReadTimeout:       util.Duration(5 * time.Minute),
+		HttpReadHeaderTimeout: util.Duration(5 * time.Minute),
+		HttpWriteTimeout:      util.Duration(5 * time.Minute),
+		HttpIdleTimeout:       util.Duration(5 * time.Minute),
```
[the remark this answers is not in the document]
asafbennatan: defaults unchanged from previous code

internal/config/common/metrics.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ResourceSyncCollectorConfig holds resource sync metrics collector configuration.
+type ResourceSyncCollectorConfig struct {
+	PeriodicCollectorConfig
+}
+
+// WorkerCollectorConfig holds worker metrics collector configuration.
+type WorkerCollectorConfig struct {
+	CollectorConfig
+}
```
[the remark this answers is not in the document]
asafbennatan: same as we had before this pr

internal/config/api/config.go
```diff
@@ -0,0 +1,379 @@
...
+	Database      *common.DatabaseConfig     `json:"database,omitempty"`
+	Service       *ServiceConfig             `json:"service,omitempty"`
+	KV            *common.KVConfig           `json:"kv,omitempty"`
+	Auth          *common.AuthConfig         `json:"auth,omitempty"`
+	Metrics       *common.MetricsConfig      `json:"metrics,omitempty"`
+	Tracing       *common.TracingConfig      `json:"tracing,omitempty"`
+	CA            *ca.Config                 `json:"ca,omitempty"`
+	Organizations *OrganizationsConfig       `json:"organizations,omitempty"`
+	Alertmanager  *common.AlertmanagerConfig `json:"alertmanager,omitempty"`
+}
```
fzdarsky, line 36: Use composition?

```suggestion
type Config struct {
    common.Config `json:",inline"`

    Service       *ServiceConfig       `json:"service,omitempty"`
    CA            *ca.Config           `json:"ca,omitempty"`
    Organizations *OrganizationsConfig `json:"organizations,omitempty"`
}
```

fzdarsky: I'm actually wondering why we didn't make these fields required in the first place to avoid having to deal with pointers...

cmd/flightctl-alert-exporter/main.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 
 	"github.com/flightctl/flightctl/internal/alert_exporter"
-	"github.com/flightctl/flightctl/internal/config"
+	alertexportercfg "github.com/flightctl/flightctl/internal/config/alertexporter"
```
fzdarsky, line 8: Alias to `config` instead as the type of config is clear from context as doesn't lead to these long and unwieldy package names?

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/config.yaml.template
```diff
@@ -12,39 +11,37 @@ service:
...
+    - http://localhost:8080/callback
 {{- end}}
-    pamService: {{if .global.auth.pamOidcIssuer.pamService}}{{.global.auth.pamOidcIssuer.pamService}}{{else}}flightctl{{end}}
-    allowPublicClientWithoutPKCE: {{if .global.auth.pamOidcIssuer.allowPublicClientWithoutPKCE}}{{.global.auth.pamOidcIssuer.allowPublicClientWithoutPKCE}}{{else}}false{{end}}
-    accessTokenExpiration: {{if .global.auth.pamOidcIssuer.accessTokenExpiration}}{{.global.auth.pamOidcIssuer.accessTokenExpiration}}{{else}}1h{{end}}
-    refreshTokenExpiration: {{if .global.auth.pamOidcIssuer.refreshTokenExpiration}}{{.global.auth.pamOidcIssuer.refreshTokenExpiration}}{{else}}168h{{end}}
-    pendingSessionCookieMaxAge: {{if .global.auth.pamOidcIssuer.pendingSessionCookieMaxAge}}{{.global.auth.pamOidcIssuer.pendingSessionCookieMaxAge}}{{else}}10m{{end}}
-    authenticatedSessionCookieMaxAge: {{if .global.auth.pamOidcIssuer.authenticatedSessionCookieMaxAge}}{{.global.auth.pamOidcIssuer.authenticatedSessionCookieMaxAge}}{{else}}30m{{end}}
-    cookieHttpOnly: {{if .global.auth.pamOidcIssuer.cookieHttpOnly}}{{.global.auth.pamOidcIssuer.cookieHttpOnly}}{{else}}true{{end}}
```
fzdarsky, line 50: Dropped intentionally, I guess?

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/config.yaml.template
```diff
@@ -12,39 +11,37 @@ service:
       - {{.}}
 {{- end}}
     {{end}}
-auth:
-  pamOidcIssuer:
-    address: ":8444"
-    issuer: {{if .global.auth.pamOidcIssuer.issuer}}{{.global.auth.pamOidcIssuer.issuer}}{{else}}https://{{.global.baseDomain}}:8444/api/v1/auth{{end}}
-    clientId: {{if .global.auth.pamOidcIssuer.clientId}}{{.global.auth.pamOidcIssuer.clientId}}{{else}}flightctl-client{{end}}
+pamOidcIssuer:
+  address: ":8444"
```
fzdarsky, line 15: While semantically the same, should this be `0.0.0.0:8444` to make it more explicit this is the server's bind address? I've noticed some other services put the service name here (e.g. `fligthctl-api:3443`), which is probably something we should fix, too, but not necessarily in this PR.

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/config.yaml.template
```diff
@@ -12,39 +11,37 @@ service:
       - {{.}}
 {{- end}}
     {{end}}
-auth:
-  pamOidcIssuer:
-    address: ":8444"
-    issuer: {{if .global.auth.pamOidcIssuer.issuer}}{{.global.auth.pamOidcIssuer.issuer}}{{else}}https://{{.global.baseDomain}}:8444/api/v1/auth{{end}}
-    clientId: {{if .global.auth.pamOidcIssuer.clientId}}{{.global.auth.pamOidcIssuer.clientId}}{{else}}flightctl-client{{end}}
+pamOidcIssuer:
```
fzdarsky, line 14: We need to discuss and agree whether we want to consider config files to be APIs, too (even though in theory users should not modify them directly, there may be scenarios where users need an escape hatch), or just the service-config.yaml.

asafbennatan: i think it doesnt make sense -currently-  it cant even be done , if you manually change a config file the service restart will override your changes as the flightctl-standalone regenerates the config file on each restart

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/config.yaml.template
asafbennatan: IDK originally this was copied from other services behavior , i assume we can change this later

deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer-config/config.yaml.template
asafbennatan: changed structure a bit - > i hated that we had pam issuer under the auth config struct

cmd/flightctl-alert-exporter/main.go
asafbennatan: yep

internal/config/api/config.go
asafbennatan: i was actually not sure what should be the "common" config for example some services do not need DB

asafbennatan:
regarding this :
>One point of confusion I've seen with people new to the code is that they don't expect internal/config/ to be just for the service(s) and that the agent and CLI actually have their own config packages (under internal/agent/config/ and internal/cli/, respectively. I'm wondering whether we should make all components use a common pattern.

i would vote against - as what i want is even greater separation , IMO the codebase should be structured likeso:

common
flightctl-api
flightctl-worker
...
(all services/agent etc)

under common we will have all the common stuff organized how we see fit ( i wont mention the internal package structure at this time)
under each service(or agent) we will have only the stuff used by that service

then switch to golang workspaces

PR #2314, opened 2026-01-06
EDM-2887: ImageBuild with tasks API

asafbennatan:
adds create imagebuild with tasks api so imagebuild and imageexport can be created atomically - used by ui

## Summary by CodeRabbit

* **New Features**
  * Added ImagePipeline endpoint to atomically create an ImageBuild and optional ImageExport in one transaction.
* **Bug Fixes / Validation**
  * Improved validation and error handling to separate client (4xx) vs internal (5xx) errors and avoid partial creates on failure.
* **Tests**
  * Added comprehensive unit and integration tests for creation, validation failures, duplicate handling, rollback, and source-override behavior.


api/v1beta1/imagebuilder/openapi.yaml
```diff
@@ -240,6 +240,61 @@ paths:
             application/json:
               schema:
                 $ref: '../openapi.yaml#/components/schemas/Status'
+  /api/v1/imagebuilds/withtasks:
```
sk-ilya: Can `ImageExport` be a subresource of an `ImageBuild`? `ImageExports` are linked to `ImageBuilds` anyway.

Alternatively, the image export settings (format, destination) could be part of the `ImageBuild` spec so that, when defined, creating an `ImageBuild` also creates an `ImageExport`. It does not seem that doing this "atomically in a single transaction" is required for correctness, "eventually" will do as well.

Either option would align better with a resource-based API

internal/imagebuilder_api/service/imagebuild.go
```diff
@@ -12,11 +12,13 @@ import (
 	"github.com/google/uuid"
 	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
 )
 
 // ImageBuildService handles business logic for ImageBuild resources
 type ImageBuildService interface {
 	Create(ctx context.Context, orgId uuid.UUID, imageBuild api.ImageBuild) (*api.ImageBuild, v1beta1.Status)
+	CreateWithTx(ctx context.Context, tx *gorm.DB, orgId uuid.UUID, imageBuild api.ImageBuild) (*api.ImageBuild, v1beta1.Status)
```
sk-ilya: > tx *gorm.DB

This is a leaky abstraction requiring the caller to deal with lower-level database stuff. It’s a relatively minor issue since `ImageBuildWithTasksService` is the only caller, but it would still be better to expose cleaner methods.

Alternatively, consider the comment above about resource-based approach where the `ImageBuild`'s spec contains the fields needed to create an ImageExport + reconciliation mechanism.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -13,11 +13,13 @@ import (
 	"github.com/google/uuid"
 	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
 )
 
 // ImageExportService handles business logic for ImageExport resources
 type ImageExportService interface {
 	Create(ctx context.Context, orgId uuid.UUID, imageExport api.ImageExport) (*api.ImageExport, v1beta1.Status)
+	CreateWithTx(ctx context.Context, tx *gorm.DB, orgId uuid.UUID, imageExport api.ImageExport, skipSourceValidation bool) (*api.ImageExport, v1beta1.Status)
```
sk-ilya: > tx *gorm.DB,

Same as above. 

> skipSourceValidation bool 


is a similar issue.  Will it work if `skipSourceValidation` is removed and the validation is never skipped?

internal/imagebuilder_api/transport/handler.go
```diff
@@ -90,6 +90,18 @@ func (h *TransportHandler) DeleteImageBuild(w http.ResponseWriter, r *http.Reque
...
+func (h *TransportHandler) CreateImageBuildWithTasks(w http.ResponseWriter, r *http.Request) {
+	var req api.ImageBuildWithTasksRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		SetParseFailureResponse(w, err)
+		return
+	}
+
+	body, status := h.service.ImageBuildWithTasks().Create(r.Context(), OrgIDFromContext(r.Context()), req)
+	SetResponse(w, body, status)
+}
```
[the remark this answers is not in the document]
asafbennatan: fixed

internal/imagebuilder_api/store/store.go
```diff
@@ -14,6 +14,8 @@ type Store interface {
 	RunMigrations(ctx context.Context) error
 	Ping() error
 	Close() error
+	// DB returns the underlying database connection for transaction management
```
sk-ilya: Same as above. The gorm shouldn't exposed by the "Store" interface. 

Instead of the clear separation and dependency direction
```
ImageBuildService   ImageExportService
          |                    |
          V                    V
           Image Builder Store
                        |
                        V
                    Gorm DB
```

We have
```
           ImageExportService  ------------\
           |            |                  |
           V            V                  |
ImageBuildService   ImageExportService     |
   |       |            |          |       |
   |       V            V          |       |
   |     Image Builder Store       |       |
   |                 |             |       |
   |                 V             |       |
    \---------> Gorm DB  <---------/------/
```
Where everything depends on gorm

internal/imagebuilder_api/service/imagepipeline.go
```diff
@@ -0,0 +1,105 @@
...
+		// Override source to reference the ImageBuild being created
+		imageBuildName := *req.ImageBuild.Metadata.Name
+		source := api.ImageBuildRefSource{
+			Type:          api.ImageBuildRefSourceTypeImageBuild,
+			ImageBuildRef: imageBuildName,
+		}
+		if err := imageExport.Spec.Source.FromImageBuildRefSource(source); err != nil {
+			return nil, StatusInternalServerError("failed to set imageExport source: " + err.Error())
+		}
+	}
```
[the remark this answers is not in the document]
asafbennatan: done

go.mod
```diff
@@ -120,6 +120,7 @@ require (
 	go.uber.org/zap v1.27.0
 	golang.org/x/net v0.46.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
+	gorm.io/driver/sqlite v1.5.5
```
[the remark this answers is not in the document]
asafbennatan: removed change

api/v1beta1/imagebuilder/openapi.yaml
asafbennatan: > Can ImageExport be a subresource of an ImageBuild? ImageExports are linked to ImageBuilds anyway.

no , they are optionally linked to image builds , i would like to maintain a clear seperation between the two , it makes everything much easier to organize, 

additionally the idea of this api is to create both atomically so this being a subresource wont help as far as i understand :
since api/v1/imagebuild/{name}/imageexport
assumes that there is an image build that was previously created

these sort of create couple of things together simply dont work well with resource-based api but it is what we need here

sk-ilya: > so this being a subresource wont help

Right, my mistake.

Will it be a problem if a client first creates an  `ImageBuild`, and then an `ImageExport`? Does it _have_ to be atomic?  WDYT about the "Alternatively, " part of the original comment?

internal/imagebuilder_api/service/imagebuild.go
asafbennatan: fixed , by using the context to optionally pass a transaction

internal/imagebuilder_api/service/imageexport.go
asafbennatan: we need to skip the source validation since image export checks if the image build exist and it still doesnt as we are running it in the same transaction

internal/imagebuilder_api/store/store.go
asafbennatan: fixed , in general the pattern of services depending  on the main store is bad IMO - > we dont have it here which is good

api/v1beta1/imagebuilder/openapi.yaml
asafbennatan: >Will it be a problem if a client first creates an ImageBuild, and then an ImageExport? Does it have to be atomic?


it is , the ui/ux design implies these are created together in one of the possible flows , if two apis are called and one fails we have a split state the ui does not have an easy way to handle 


regarding this :

>Alternatively, the image export settings (format, destination) could be part of the ImageBuild spec so that, when defined, creating an ImageBuild also creates an ImageExport. It does not seem that doing this "atomically in a single transaction" is required for correctness, "eventually" will do as well.


there is a flow where image export is created without an image build so in that case it wouldn't make sense to duplicate the info/logic ,i would rather keep them separated for code organization purposes - the fact that imagebuild "knows" nothing about imageexport is a good thing

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -97,49 +97,52 @@ func (s *imageExportService) UpdateLastSeen(ctx context.Context, orgId uuid.UUID
...
-			source, err := imageExport.Spec.Source.AsImageBuildRefSource()
-			if err != nil {
-				errs = append(errs, errors.New("invalid imageBuild source"))
-			} else if source.ImageBuildRef == "" {
-				errs = append(errs, errors.New("spec.source.imageBuildRef is required for imageBuild source type"))
-			} else {
-				// Check that the referenced ImageBuild exists
-				_, err = s.imageBuildStore.Get(ctx, orgId, source.ImageBuildRef)
+	// Validate source - uses discriminator pattern (skip if source is auto-set)
+	if !skipSourceValidation {
```
sk-ilya: Is the validation skipped here only because of 

> _, err = s.imageBuildStore.Get(ctx, orgId, source.ImageBuildRef)?

If both `ImageBuild` and `ImageExport` are created within the same transaction, then it should work anyway?

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -97,49 +97,52 @@ func (s *imageExportService) UpdateLastSeen(ctx context.Context, orgId uuid.UUID
...
 				if err != nil {
-					errs = append(errs, fmt.Errorf("spec.source.imageBuildRef: ImageBuild %q not found", source.ImageBuildRef))
+					errs = append(errs, errors.New("invalid imageBuild source"))
+				} else if source.ImageBuildRef == "" {
+					errs = append(errs, errors.New("spec.source.imageBuildRef is required for imageBuild source type"))
+				} else {
+					// Check that the referenced ImageBuild exists
+					_, err = s.imageBuildStore.Get(ctx, orgId, source.ImageBuildRef)
+					if err != nil {
+						errs = append(errs, fmt.Errorf("spec.source.imageBuildRef: ImageBuild %q not found", source.ImageBuildRef))
```
sk-ilya:
```go
if errors.Is(err, flterrors.ErrResourceNotFound)
```

internal/imagebuilder_api/service/imageexport.go
asafbennatan: >Is the validation skipped here only because of

>_, err = s.imageBuildStore.Get(ctx, orgId, source.ImageBuildRef)?

yes , but only if we also change the get command to use the same transaction - fixed

internal/imagebuilder_api/service/imageexport.go
asafbennatan: fixed

internal/imagebuilder_api/service/imageexport.go
sk-ilya: So maybe worth renaming   `skipSourceValidation` to `ensureImageBuildExists` and have a private method with the same name to split resource validation from consistence checks.

 This way `Create()` can call `validate()` unconditionally, and then the new private method if `ensureImageBuildExists` is true.

Or even better since the validation is quite simple anyway - let the OpenAPI validator handle it and remove the validation code

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -119,7 +120,7 @@ func (s *imageExportService) validate(ctx context.Context, orgId uuid.UUID, imag
 			} else {
 				// Check that the referenced ImageBuild exists
 				_, err = s.imageBuildStore.Get(ctx, orgId, source.ImageBuildRef)
-				if err != nil {
+				if errors.Is(err, flterrors.ErrResourceNotFound) {
```
sk-ilya, line 126: Else it's an internal error?

The suggestion about extracting a `ensureImageBuildExists` private methods and calling both from `Create()` is still valid - this way it'll be more straightforward to propagate the error. 

But letting the middleware to handle the validation is the best option IMO, there will be no need to  keep the validation code here (with the exception of ensuring the build exists)

asafbennatan: fixed, i am not sure we should seperate the ensureimagebuildexists from Validate - i find that ensureimagebuildexists logic should belong to the Validate logic

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -119,8 +120,10 @@ func (s *imageExportService) validate(ctx context.Context, orgId uuid.UUID, imag
...
 					errs = append(errs, fmt.Errorf("spec.source.imageBuildRef: ImageBuild %q not found", source.ImageBuildRef))
+				} else if err != nil {
+					errs = append(errs, fmt.Errorf("spec.source.imageBuildRef: failed to get ImageBuild %q: %w", source.ImageBuildRef, err))
```
sk-ilya: If it's not an `ErrResourceNotFound`, then it should be either handled or propagated and result in a 5xx code. 

Now it's treated as a bad request as if it was a client's mistake

PR #2317, opened 2026-01-06
EDM-2261: Fix templating issue with inline application definitions (#…

kkyrazis:
…2285)

(cherry picked from commit 4c598f9c9676c0d819fc7fec29f6ce7864a67b55)

### DIFF
--- a/api/v1beta1/application_validation.go
+++ b/api/v1beta1/application_validation.go
@@ -23,13 +23,28 @@ type composeValidator struct {
 	paths map[string]struct{}
 }
 
+// fleets require additional processing to ensure templates are defined appropriately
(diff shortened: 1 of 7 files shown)


hexfusion:
@sk-ilya can you take a peak at this test please

sk-ilya:
@hexfusion  it's comparing with `main` instead of the target branch... I'd call it a pipeline bug

hexfusion:
> @hexfusion it's comparing with `main` instead of the target branch... I'd call it a pipeline bug

ok can we track that please :+1:

sk-ilya:
@hexfusion @kkyrazis should be fixed by #2321 

the fix needs to be backported first

PR #2324, opened 2026-01-07
EDM-2917: Implement Complete Pruning Manager

ori-amizur:
## Summary by CodeRabbit

* **New Features**
  * Added an image & artifact pruning manager with persistent reference tracking, safe eligibility determination, capability validation, and dynamic config reloads.

* **Tests**
  * Added a comprehensive test suite and generated mocks covering reference extraction, eligibility logic, pruning, validation, and removal workflows.

* **Documentation**
  * Corrected an explanatory comment to reference the proper pruning configuration key.


internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1053 @@
...
+		for _, pattern := range suffixPatterns {
+			files, err := filepath.Glob(m.readWriter.PathFor(filepath.Join(appPath, pattern)))
+			if err != nil {
+				continue
+			}
+			if len(files) > 0 {
+				composeFound = true
+				break
+			}
+		}
```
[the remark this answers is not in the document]
ori-amizur: The PathFor function inserts rootDir at the beginning of the returned path.  It is useful for testing.

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1030 @@
...
+	m.config = cfg.ImagePruning
+
+	// Update the config with the new values
+	if lo.FromPtr(cfg.ImagePruning.Enabled) && !lo.FromPtr(prev) {
+		m.prunePending.Store(true)
+	}
+
+	m.log.Infof("Image pruning config reloaded: enabled=%v", lo.FromPtr(m.config.Enabled))
+	return nil
+}
```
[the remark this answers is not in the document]
ori-amizur: What are the possible consequences of such race ?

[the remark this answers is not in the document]
ori-amizur: Does the recent change solve the race issue ?

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1051 @@
...
+		// Extract images from volume quadlets (only if it's an OCI image, not a quadlet file reference)
+		if quad.Type == common.QuadletTypeVolume && quad.Image != nil && !quadlet.IsImageReference(*quad.Image) {
+			refs = append(refs, *quad.Image)
+		}
+		// Extract mount images (only if they're OCI images, not quadlet file references)
+		for _, mountImage := range quad.MountImages {
+			if !quadlet.IsImageReference(mountImage) {
+				refs = append(refs, mountImage)
+			}
+		}
```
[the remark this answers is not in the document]
ori-amizur: @coderabbitai can you review this again? Seems the issue doesn't exist now

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1030 @@
...
+		specType spec.Type
+		name     string
+		required bool
+	}{
+		{spec.Current, "current", true},
+		{spec.Desired, "desired", false},
+	}
+
+	for _, st := range specTypes {
+		device, err := m.specManager.Read(st.specType)
```
hexfusion, line 403: question: why are we reading from disk if we have current and desired spec in memory?

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1030 @@
...
+
+// ImagePruningConfig holds configuration for image pruning operations.
+// This type is defined in internal/agent/config/config.go as config.ImagePruning
+// and is aliased here for backward compatibility and clarity.
+type ImagePruningConfig = config.ImagePruning
+
+// manager implements the Manager interface for image pruning operations.
+type manager struct {
+	podmanClient *client.Podman
+	specManager  spec.Manager
```
hexfusion, line 57: question: why do we need the spec manager vs consume current and desired spec in memory after update success?

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1030 @@
...
+	var removedImageRefs []string
+	var removedArtifacts int
+	var removedArtifactRefs []string
+
+	if totalEligible > 0 {
+		m.log.Infof("Starting pruning of %d eligible images and %d eligible artifacts", len(eligible.Images), len(eligible.Artifacts))
+
+		// Remove eligible images and artifacts separately, tracking which ones were successfully removed
+		var err error
+		removedImages, removedImageRefs, err = m.removeEligibleImages(ctx, eligible.Images)
```
hexfusion, line 125: it seems like we can reduce the scope of the error should we? it appears that if we error here we will shadow err for the next checks is that intentional?
```suggestion
		removedImages, removedImageRefs, err := m.removeEligibleImages(ctx, eligible.Images)
```

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1030 @@
...
+		log:          log,
+		dataDir:      dataDir,
+	}
+	ret.config.Store(&config)
+	return ret
+}
+
+// Prune removes unused container images and OCI artifacts after successful spec reconciliation.
+// It preserves images required for current and desired operations.
+func (m *manager) Prune(ctx context.Context) error {
```
hexfusion, line 95: is this 
```go
func (m *manager) AfterUpdate(ctx context.Context, current, desired *v1beta.Device) {
```

internal/agent/device/image_pruning/manager.go
```diff
@@ -0,0 +1,1030 @@
...
+	}
+
+	return nil
+}
+
+// RecordReferences records all image and artifact references from current and desired specs to a file.
+// This ensures the references file exists and is kept up to date even when pruning is disabled,
+// so that when pruning is later enabled, it has accurate historical data to work with.
+// This is called on every successful sync, regardless of pruning enabled status.
+func (m *manager) RecordReferences(ctx context.Context, current, desired *v1beta1.Device) error {
```
hexfusion, line 164: ideally the Manager interface can be standardized so can we use existing patterns?
```go
func (m *manager) Sync(ctx context.Context, current, desired *v1beta1.Device) error {
```

ori-amizur: We can, although it makes things less clear. There is no configuration here we are trying to apply - we just recording references.

internal/agent/device/image_pruning/manager.go
ori-amizur: OK

internal/agent/device/image_pruning/manager.go
ori-amizur: The current/desired in memory are read before upgrade.  So if we want to use this pattern, we need to read them before we call the function.
Anyway, I think that the current function name is more clear than just 'AfterUpdate' - but I can change.

internal/agent/device/image_pruning/manager.go
ori-amizur: They are not in memory after update success.  So we can use this pattern - but we will need to read them externally.

internal/agent/device/image_pruning/manager.go
ori-amizur: They are not in memory after [update](https://github.com/flightctl/flightctl/blob/e50b9d6f6892a69152bd8b0a38fd929b455f6a1a/internal/agent/device/device.go#L217), but we can read them and pass them if it is a preferred way.

internal/agent/device/image_pruning/manager.go
ori-amizur: The question: Should we name the function according to the calling context - or should we name it according to what it does?

PR #2328, opened 2026-01-08
EDM-2635: Structured Agent Error Messages

noga-magen:
extract element function and keywords are not completed yet. 

## Summary by CodeRabbit

* **New Features**
  * Device error messages are now structured with timestamp, phase, component, element, category and status for clearer diagnostics.

* **Documentation**
  * Added user guidance and examples for the new structured device error syntax.

* **Refactor**
  * Centralized error classification and mapping to status codes for consistent, user-facing messages.

* **Bug Fixes**
  * More consistent error logging and status messages across update/sync flows; tests adjusted for broader error messages.

* **Chores**
  * UpdateState constants aligned for API compatibility.

### DIFF
--- a/api/core/v1beta1/consts.go
+++ b/api/core/v1beta1/consts.go
@@ -3,6 +3,7 @@ package v1beta1
 import (
 	"time"
 
+	"github.com/flightctl/flightctl/internal/consts"
 	"github.com/google/uuid"
 )
 
@@ -122,41 +123,19 @@ const (
 
 var KnownExternalRoles = []string{ExternalRoleAdmin, ExternalRoleOrgAdmin, ExternalRoleOperator, ExternalRoleViewer, ExternalRoleInstaller}
 
-type UpdateState string
+// UpdateState is an alias to consts.UpdateState for API compatibility.
+type UpdateState = consts.UpdateState
 
 const (
-	// The agent is validating the desired device spec and downloading
-	// dependencies. No changes have been made to the device's configuration
-	// yet.
-	UpdateStatePreparing UpdateState = "Preparing"
-	//  The agent has validated the desired spec, downloaded all dependencies,
-	//  and is ready to update. No changes have been made to the device's
-	//  configuration yet.
-	UpdateStateReadyToUpdate UpdateState = "ReadyToUpdate"
-	// The agent has started the update transaction and is writing the update to
-	// disk.
-	UpdateStateApplyingUpdate UpdateState = "ApplyingUpdate"
-	// The agent initiated a reboot required to activate the new OS image and configuration.
-	UpdateStateRebooting UpdateState = "Rebooting"
-	// The agent has successfully completed the update and the device is
-	// conforming to its device spec. Note that the device's update status may
-	// still be reported as `OutOfDate` if the device spec is not yet at the
-	// same version as the fleet's device template
-	UpdateStateUpdated UpdateState = "Updated"
-	// The agent has canceled the update because the desired spec was reverted
-	// to the current spec before the update process started.
-	UpdateStateCanceled UpdateState = "Canceled"
-	// The agent failed to apply the desired spec and will not retry. The
-	// device's OS image and configuration have been rolled back to the
-	// pre-update version and have been activated
-	UpdateStateError UpdateState = "Error"
-	// The agent has detected an error and is rolling back to the pre-update OS
-	// image and configuration.
-	UpdateStateRollingBack UpdateState = "RollingBack"
-	// The agent failed to apply the desired spec and will retry. The device's
-	// OS image and configuration have been rolled back to the pre-update
-	// version and have been activated.
-	UpdateStateRetrying UpdateState = "Retrying"
+	UpdateStatePreparing      = consts.UpdateStatePreparing
+	UpdateStateReadyToUpdate  = consts.UpdateStateReadyToUpdate
+	UpdateStateApplyingUpdate = consts.UpdateStateApplyingUpdate
+	UpdateStateRebooting      = consts.UpdateStateRebooting
+	UpdateStateUpdated        = consts.UpdateStateUpdated
+	UpdateStateCanceled       = consts.UpdateStateCanceled
+	UpdateStateError          = consts.UpdateStateError
+	UpdateStateRollingBack    = consts.UpdateStateRollingBack
+	UpdateStateRetrying       = consts.UpdateStateRetrying
 )
 
 type DecommissionState string
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -120,16 +120,16 @@ func (a *Agent) Run(ctx context.Context) error {
 func (a *Agent) sync(ctx context.Context, current, desired *v1beta1.Device) error {
 	if !spec.IsRollback(current, desired) {
 		if err := a.beforeUpdate(ctx, current, desired); err != nil {
-			return fmt.Errorf("before update: %w", err)
+			return fmt.Errorf("%w: %w", errors.ErrPhasePreparing, err)
 		}
 	}
 
 	if err := a.syncDevice(ctx, current, desired); err != nil {
-		return fmt.Errorf("sync device: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrPhaseApplyingUpdate, err)
 	}
 
 	if err := a.afterUpdate(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("after update: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrPhaseActivatingConfig, err)
 	}
 
 	return nil
@@ -314,11 +314,11 @@ func (a *Agent) statusUpdate(ctx context.Context) {
 
 func (a *Agent) beforeUpdate(ctx context.Context, current, desired *v1beta1.Device) error {
 	if err := a.resourceManager.BeforeUpdate(ctx, desired.Spec); err != nil {
-		return fmt.Errorf("resources: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentResources, err)
 	}
 
 	if err := a.specManager.CheckPolicy(ctx, policy.Download, desired.Version()); err != nil {
-		return fmt.Errorf("download policy: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentDownloadPolicy, err)
 	}
 
 	// the agent is validating the desired device spec and downloading
@@ -352,19 +352,19 @@ func (a *Agent) beforeUpdate(ctx context.Context, current, desired *v1beta1.Devi
 	}
 
 	if err := a.prefetchManager.BeforeUpdate(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("prefetch: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentPrefetch, err)
 	}
 
 	if err := a.appManager.BeforeUpdate(ctx, desired.Spec); err != nil {
-		return fmt.Errorf("applications: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentApplications, err)
 	}
 
 	if err := a.hookManager.OnBeforeUpdating(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("hooks: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentHooks, err)
 	}
 
 	if err := a.specManager.CheckPolicy(ctx, policy.Update, desired.Version()); err != nil {
-		return fmt.Errorf("update policy: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentUpdatePolicy, err)
 	}
 
 	// the agent has validated the desired spec, downloaded all dependencies,
@@ -399,23 +399,23 @@ func (a *Agent) syncDevice(ctx context.Context, current, desired *v1beta1.Device
 	}
 
 	if err := a.applicationsController.Sync(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("applications: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentApplications, err)
 	}
 
 	if err := a.hookManager.Sync(current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("hooks: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentHooks, err)
 	}
 
 	if err := a.configController.Sync(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("config: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentConfig, err)
 	}
 
 	if err := a.systemdControllerSync(ctx, desired.Spec); err != nil {
-		return fmt.Errorf("systemd: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentSystemd, err)
 	}
 
 	if err := a.lifecycleManager.Sync(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("lifecycle: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrComponentLifecycle, err)
 	}
 
 	// NOTE: policy manager is reconciled early in sync() so that the agent
@@ -537,21 +537,22 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1beta1.Device, sy
 		Type: v1beta1.ConditionTypeDeviceUpdating,
 	}
 
+	se := errors.FormatError(syncErr)
 	if !errors.IsRetryable(syncErr) {
 		msg := fmt.Sprintf("Failed to update to renderedVersion: %s: %v", version, syncErr.Error())
 		conditionUpdate.Reason = string(v1beta1.UpdateStateError)
 		conditionUpdate.Message = log.Truncate(msg, status.MaxMessageLength)
 		conditionUpdate.Status = v1beta1.ConditionStatusFalse
 		a.prefetchManager.Cleanup()
-		a.log.Error(msg)
+		a.log.Error(msg, se.Timestamp)
 	} else {
 		msg := fmt.Sprintf("Failed to update to renderedVersion: %s: retrying: %v", version, syncErr.Error())
 		conditionUpdate.Reason = string(v1beta1.UpdateStateApplyingUpdate)
 		conditionUpdate.Message = log.Truncate(msg, status.MaxMessageLength)
 		conditionUpdate.Status = v1beta1.ConditionStatusTrue
-		a.log.Warn(msg)
+		a.log.Warn(msg, se.Timestamp)
 	}
-
+	conditionUpdate.Message = se.Message()
 	if err := a.statusManager.UpdateCondition(ctx, conditionUpdate); err != nil {
 		a.log.Warnf("Failed to update device status condition: %v", err)
 	}
--- a/internal/agent/device/errors/errors.go
+++ b/internal/agent/device/errors/errors.go
@@ -11,12 +11,31 @@ import (
 	"syscall"
 
 	"github.com/flightctl/flightctl/pkg/poll"
+	"google.golang.org/grpc/codes"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var (
 	ErrRetryable = errors.New("retryable error")
 	ErrNoRetry   = errors.New("no retry")
+
+	// phases - used to wrap errors indicating which sync phase failed
+	ErrPhasePreparing        = errors.New("before update")
+	ErrPhaseApplyingUpdate   = errors.New("sync device")
+	ErrPhaseActivatingConfig = errors.New("after update")
+
+	// components - used to wrap errors indicating which component failed
+	ErrComponentResources      = errors.New("resources")
+	ErrComponentDownloadPolicy = errors.New("download policy")
+	ErrComponentUpdatePolicy   = errors.New("update policy")
+	ErrComponentPrefetch       = errors.New("prefetch")
+	ErrComponentApplications   = errors.New("applications")
+	ErrComponentHooks          = errors.New("hooks")
+	ErrComponentConfig         = errors.New("config")
+	ErrComponentSystemd        = errors.New("systemd")
+	ErrComponentLifecycle      = errors.New("lifecycle")
+	ErrComponentOS             = errors.New("os")
+	ErrComponentOSReconciled   = errors.New("os reconciliation")
 
 	// bootstrap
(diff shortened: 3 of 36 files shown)


internal/agent/device/device.go
```diff
@@ -513,21 +513,22 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1beta1.Device, sy
...
 		a.prefetchManager.Cleanup()
-		a.log.Error(msg)
+		a.log.Error(msg, timestamp)
 	} else {
 		msg := fmt.Sprintf("Failed to update to renderedVersion: %s: retrying: %v", version, syncErr.Error())
 		conditionUpdate.Reason = string(v1beta1.UpdateStateApplyingUpdate)
 		conditionUpdate.Message = log.Truncate(msg, status.MaxMessageLength)
 		conditionUpdate.Status = v1beta1.ConditionStatusTrue
-		a.log.Warn(msg)
+		a.log.Warn(msg, timestamp)
```
hexfusion: question: the log generates timestamp by default I would like to find a way to not require a timestamp as a signature for the logger.

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+	ComponentDownloadPolicy Component = "DownloadPolicy"
+	ComponentUpdatePolicy   Component = "UpdatePolicy"
+	ComponentPrefetch       Component = "Prefetch"
+	ComponentApplications   Component = "Applications"
+	ComponentHooks          Component = "Hooks"
+	ComponentConfig         Component = "Config"
+	ComponentSystemd        Component = "Systemd"
+	ComponentLifecycle      Component = "Lifecycle"
+	ComponentOS             Component = "OS"
+	ComponentOSReconciled   Component = "OSReconciliation"
```
hexfusion: question: what is OSReconciliation

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+	"strings"
+	"time"
+
+	"google.golang.org/grpc/codes"
+)
+
+// FormatError converts a raw error into a StructuredError.
+func FormatError(err error, ts time.Time) string {
+	se := &StructuredError{
+		Identifier: ts,
```
hexfusion: question: when would ts not be now()

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+// StructuredError represents a sanitized, structured error for status display.
+type StructuredError struct {
+	Phase      Phase
+	Category   Category
+	Component  Component
+	Element    string
+	StatusCode codes.Code
+	Identifier time.Time
```
hexfusion: since this is time.Time is Identifier -> Timestamp?

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
+package errors
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"google.golang.org/grpc/codes"
```
hexfusion, line 7: question: do we want to bring codes locally so we can extend as we want?

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+// Phase represents the sync phase where the error occurred.
+type Phase string
+
+const (
+	PhasePreparing      Phase = "Preparing"
+	PhaseApplyingUpdate Phase = "ApplyingUpdate"
+	PhaseRebooting      Phase = "Rebooting"
+	PhaseUnknown        Phase = "Unknown"
```
hexfusion: question: would this enum also be useful to the service? if yes perhaps it should live somewhere shared or in api/?

internal/agent/device/errors/structured.go
hexfusion: question: are these types useful in other areas should they be shared

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+type Component string
+
+const (
+	ComponentResources      Component = "Resources"
+	ComponentDownloadPolicy Component = "DownloadPolicy"
+	ComponentUpdatePolicy   Component = "UpdatePolicy"
+	ComponentPrefetch       Component = "Prefetch"
+	ComponentApplications   Component = "Applications"
+	ComponentHooks          Component = "Hooks"
+	ComponentConfig         Component = "Config"
```
hexfusion: some of these mirror api root keys `Applications` others are functionality `Prefetch` I think we want very clear and concrete mapping here.

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+	ComponentOSReconciled   Component = "OSReconciliation"
+	ComponentUnknown        Component = "Unknown"
+)
+
+// Category represents the high-level functional area describing WHAT failed.
+type Category string
+
+const (
+	CategoryNetwork       Category = "Network"
+	CategoryConfiguration Category = "Configuration"
```
hexfusion, line 99: question: is this the agent config.yaml or config api. docs will help answer this in the future ;)

internal/agent/device/errors/errors.go
```diff
@@ -226,31 +250,7 @@ func IsTimeoutError(err error) bool {
 
 // FromStderr converts stderr output from a command into an error type.
 func FromStderr(stderr string, exitCode int) error {
```
hexfusion, line 545: since FromStderr converts any strings we are aware of to error types can we standardize on those types? my expectation is that any string that we find interesting becomes an error type. error types implement Stringer err.Error() if you ever needed a string. and in the same way we handle things like IsRetryable() you could do something like

```go
func ToCode(err error) Code
```

I think this might simplify your code

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+}{
+	{"before update: ", PhasePreparing},
+	{"sync device: ", PhaseApplyingUpdate},
+	{"after update: ", PhaseRebooting},
+}
+
+// phaseComponentPrefixes maps phases to their valid component prefixes.
+var phaseComponentPrefixes = map[Phase][]componentPrefixEntry{
+	PhasePreparing: {
+		{"resources: ", ComponentResources},
```
hexfusion: if ComponentResources was ErrComponentResources and and error type you can wrap them into the error itself 

```go
   // wrap
   if err := sycnResources(ctx); err != nil {
       return fmt.Errorf("%w, %w, ErrComponentResources, err)
   }
   
   // unwrap
   if errors.Is(err, ErrComponentResources {
       // do something interesting
   }
```

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+// afterUpdatePatterns maps error patterns to components for AfterUpdate phase.
+var afterUpdatePatterns = []struct {
+	pattern   string
+	component Component
+}{
+	{"failed to update status with decommission", ComponentLifecycle},
+	{"failed to wipe credentials", ComponentLifecycle},
+	{"failed to delete desired.json", ComponentLifecycle},
```
hexfusion: based on above you would not need this mapping I believe becase you would just wrap the component into the error

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+	{"copying image", codes.Unavailable},
+	// Generic catch-alls
+	{"unknown", codes.Internal},
+	{"unsupported", codes.Internal},
+	{"time", codes.Internal},
+	{"invalid", codes.InvalidArgument},
+}
+
+// errorTypeToCode maps error types from StderrKeywords to status codes.
+var errorTypeToCode = map[error]codes.Code{
```
hexfusion: +1 makes sense that only only error types should map to codes.

FromStderr  -> error ToCode(err) -> code

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+
+// extractPhaseAndComponent identifies the phase and component.
+func (se *StructuredError) extractPhaseAndComponent(errStr string) string {
+	errDetails := errStr
+
+	// if the sync process was changed without updating the formatting we default to unknown
+	se.Phase = PhaseUnknown
+	se.Component = ComponentUnknown
+
+	for _, p := range phasePrefix {
```
hexfusion: we want to try to avoid string matching where possible and instead try to unwrap errors. not sure it is possible here but lets see.

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+		if strings.HasPrefix(errDetails, p.prefix) {
+			se.Phase = p.phase
+			errDetails = errDetails[len(p.prefix):]
+			break
+		}
+	}
+
+	// after update has patterns not prefixes for components
+	if se.Phase == PhaseRebooting {
+		for _, p := range afterUpdatePatterns {
```
hexfusion: remember each of these for loops os worst case 0(n) can we achieve any of this without string matching?

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+	return errDetails
+}
+
+// determineStatusCode finds the error code using two-tier lookup.
+func (se *StructuredError) determineStatusCode(errDetails string) {
+	errDetailsLower := strings.ToLower(errDetails)
+
+	if keywords, ok := componentKeywords[se.Component]; ok {
+		for _, kw := range keywords {
+			if strings.Contains(errDetailsLower, kw.keyword) {
```
hexfusion: after FromStderr do we have/need any string matching?

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+// inferCategory infers the category from the status code and overrides the default category.
+func (se *StructuredError) inferCategory() {
+	if cat, ok := statusCategoryOverrides[se.StatusCode]; ok {
+		se.Category = cat
+		return
+	}
+	se.Category = CategoryUnknown
+}
+
+func extractElement(errStr string) string {
```
hexfusion: question: is this nessisary can this be done is any other way?

hexfusion (commented):
great start!

internal/agent/device/device.go
```diff
@@ -513,21 +513,22 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1beta1.Device, sy
...
 	}
-
+	conditionUpdate.Message = errors.FormatError(syncErr, timestamp)
```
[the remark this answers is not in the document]
noga-magen: The maximum length of the message can not be bigger than 250 chars. 
There is no combination that will exceed the max length.

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,604 @@
...
+		Identifier: ts,
+	}
+	errStr := err.Error()
+	errDetails := se.extractPhaseAndComponent(errStr)
+	se.determineStatusCode(errDetails)
+	se.inferCategory()
+	se.Element = extractElement(errDetails)
+	msg := se.buildMessage()
+	return msg
+}
```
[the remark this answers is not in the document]
noga-magen: not need because of this line: 
https://github.com/flightctl/flightctl/blob/5ab823132e228cfa2b595e50688007230f5d9753/internal/agent/device/device.go#L507

internal/agent/device/errors/structured.go
noga-magen: Now instead of defining the time in the handleSyncError and receiving the ts in  FormatError we return it and add it to the structure.

se := &StructuredError{
		Timestamp: time.Now(),
}
..........
return msg, se.Timestamp
 
 
 The ts must be either received or returned so both the agent and the API will have the same exact timestamp.
 Otherwise it's not a very good identifier.

internal/agent/device/device.go
noga-magen: see two comments below.

internal/agent/device/errors/structured.go
noga-magen: ComponentOSReconciled - Check if OS is at desired state	(pattern: "getting current bootc status:, reading current")
ComponentOS - Apply OS image update reboot (pattern: "stage image:, apply image:, bootc, unable to parse image reference")

It could be useful to use the consts for the og components
Example: 
https://github.com/flightctl/flightctl/blob/5ab823132e228cfa2b595e50688007230f5d9753/internal/agent/device/device.go#L394

internal/agent/device/errors/structured.go
noga-magen: I don't think so; at the moment, there is no reason to. 
If in the future we see that error status codes are being classified as UNKNOWN due to a lack of specific errors, we could change that.

internal/agent/device/errors/structured.go
noga-magen: Yes definitely,
Since we are making these changes, we should also update the component in the 'after-update' sync phase. This will ensure consistency with the 'before' and 'device-sync' components.

internal/agent/device/errors/structured.go
noga-magen: I put those in internal/consts.
Expect for Unknown which in only used in error/structure.go

internal/agent/device/errors/structured.go
noga-magen: Sure, I Merged the components like this: 
DownloadPolicy→UpdatePolicy
Prefetch→Applications
Hooks→Lifecycle
OSReconciled→OS

internal/agent/device/errors/structured.go
noga-magen: So it's possible if all the formatting are done by %w. 
I'll change those that don't.  And make a map from things that needs  a string matching.

internal/agent/device/errors/structured.go
noga-magen: see the comment above

internal/agent/device/errors/structured.go
noga-magen: see two comments  above

internal/agent/device/errors/structured.go
noga-magen: I'll try another way

internal/agent/device/errors/structured.go
hexfusion: ok we can look at how to improve this but that makes sense, thanks

internal/agent/device/errors/structured.go
hexfusion: great consistency is critical sounds like your on the right path

internal/agent/device/errors/structured.go
noga-magen: The way I see it there are three options: 
1. Use regex with the risk of matching irrelevant strings - risky 
2. Save within the function that logs the element and err separately - out of feature scope.    
3. keep it as it is - not optimal


I think we have to keep in the way it is ):

api/core/v1beta1/consts.go
```diff
@@ -122,41 +123,19 @@ const (
...
-	UpdateStateRetrying UpdateState = "Retrying"
+	UpdateStatePreparing      = consts.UpdateStatePreparing
+	UpdateStateReadyToUpdate  = consts.UpdateStateReadyToUpdate
+	UpdateStateApplyingUpdate = consts.UpdateStateApplyingUpdate
+	UpdateStateRebooting      = consts.UpdateStateRebooting
+	UpdateStateUpdated        = consts.UpdateStateUpdated
+	UpdateStateCanceled       = consts.UpdateStateCanceled
+	UpdateStateError          = consts.UpdateStateError
+	UpdateStateRollingBack    = consts.UpdateStateRollingBack
+	UpdateStateRetrying       = consts.UpdateStateRetrying
```
hexfusion, line 138: question: can we keep the comments?

internal/agent/device/device.go
```diff
@@ -116,16 +116,16 @@ func (a *Agent) Run(ctx context.Context) error {
...
 	}
 
 	if err := a.syncDevice(ctx, current, desired); err != nil {
-		return fmt.Errorf("sync device: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrPhaseApplyingUpdate, err)
 	}
 
 	if err := a.afterUpdate(ctx, current.Spec, desired.Spec); err != nil {
-		return fmt.Errorf("after update: %w", err)
+		return fmt.Errorf("%w: %w", errors.ErrPhaseRebooting, err)
```
hexfusion: `Rebooting` would be explicit to OS upgrade right this is technically `ActivatingConfig` in the docs it says this is transient and not reported which is true from status perspective but we can use it here.

```suggestion
		return fmt.Errorf("%w: %w", errors.ErrPhaseActivatingConfig, err)
```

ref. https://github.com/flightctl/flightctl/blob/main/docs/user/references/device-api-statuses.md

internal/agent/device/errors/errors.go
```diff
@@ -101,6 +120,101 @@ var (
...
+		ErrNoQuadletWorkload:  codes.InvalidArgument,
+		ErrInvalidTokenFormat: codes.InvalidArgument,
+		ErrTokenNotSupported:  codes.InvalidArgument,
+		ErrRunActionInvalid:   codes.InvalidArgument,
+		ErrPathIsDir:          codes.InvalidArgument,
+		ErrInvalidPath:        codes.InvalidArgument,
+		ErrInvalidSpec:        codes.InvalidArgument,
+
+		// internal errors
+		ErrAppDependency:            codes.Internal,
```
hexfusion: we should think about this one and this is an example of where using grpc codes sort of breaks. this is a dependency error and I think we want to be explicit.  not saying we need to block this PR but we need a followup to migrate away from gRPC codes. its not appropriate.

internal/agent/device/errors/errors.go
```diff
@@ -226,31 +340,7 @@ func IsTimeoutError(err error) bool {
...
-		// context
-		"context canceled":          context.Canceled,
-		"context deadline exceeded": context.DeadlineExceeded,
-		// container image resolution
-		"short-name resolution enforced": ErrImageShortName,
-		// no such object
-		"no such object": ErrNotFound,
-	}
-	for check, err := range errMap {
+	for check, err := range StderrKeywords {
```
hexfusion: I dont have a problem with breaking this out and exporting but you only use it in one place :)

internal/agent/device/errors/structured.go
hexfusion: ok can be a followup

api/core/v1beta1/consts.go
noga-magen: The same exact comments are in internal/consts/consts.go.
here was an import cycle when those consts were in the API directory so I moved them to the mentioned directory. 

So adding the comments again would be a redundant duplicity.

internal/agent/device/errors/errors.go
noga-magen: Yeah I thought it would be useful but the is no need to do that if it's only used once. I'll change that.

internal/agent/device/errors/errors.go
```diff
@@ -11,13 +11,32 @@ import (
...
+	ErrComponentDownloadPolicy = errors.New("download policy")
+	ErrComponentUpdatePolicy   = errors.New("update policy")
+	ErrComponentPrefetch       = errors.New("prefetch")
+	ErrComponentApplications   = errors.New("applications")
+	ErrComponentHooks          = errors.New("hooks")
+	ErrComponentConfig         = errors.New("config")
+	ErrComponentSystemd        = errors.New("systemd")
+	ErrComponentLifecycle      = errors.New("lifecycle")
+	ErrComponentOS             = errors.New("os")
+	ErrComponentOSReconciled   = errors.New("os reconciliation")
```
hexfusion, line 38: still not sure what ErrComponentOSReconciled is I see its only used in test so maybe can be removed

noga-magen: It's only used in the [after update](https://github.com/flightctl/flightctl/blob/b4b4236d16a4b009003763146935c9b49034a9cf/internal/agent/device/device.go#L452
) function
I can remove it and use os instead - but it is how the error was originally written.

hexfusion: ok it can change after if need

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,492 @@
...
+
+// Components map directly to DeviceSpec API keys.
+var componentKeywords = map[string][]keywordToCode{
+	ErrComponentDownloadPolicy.Error(): {
+		{"convert string to int64", codes.Internal},
+		{"cannot be negative", codes.Internal},
+		{"policy check failed", codes.Internal},
+	},
+	ErrComponentUpdatePolicy.Error(): {
+		{"convert string to int64", codes.Internal},
```
kkyrazis: are there plans to enhance this behavior so that it's less brittle? I think it's going to be difficult to ensure that everyone adding logic is also updating these maps and not changing error messages.

I really like the non-invasive nature of this, but yeah I'm just concerned about maintenance.

hexfusion: yeah we really do not need this layer can just remove this and all converting error to code. the code I think should be something we can infer by unwrapping it. In this case if we wanted a fine grained error

```
ErrInt64Convertion = errors.New("convert string to int64")
```
But maintaining both errors and this conversion will be expensive. please remove and then we can merge

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,390 @@
...
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	"github.com/flightctl/flightctl/internal/consts"
+	"google.golang.org/grpc/codes"
+)
+
+// FormatError converts a raw error into a StructuredError.
+func FormatError(err error) (string, time.Time) {
```
asafbennatan: nit:
consider making formatError return structured error ,
additionally i find the inference logic to be more appropriate as a function rather then a method - this way StructuredError itself doesn't have any extraction logic , it is simply a container and the code becomes clearer IMO:
```
	phase,component:=extractPhaseAndComponent(err)
	statusCode:= determineStatusCode(err)
	category:=inferCategory()
	element = extractElement(err.Error())
return &StructuredError{
		Timestamp: time.Now(),
        Phase: phase,
        Component:component.
        StatusCode: statusCode,
        Element: element
	} // message is built internally
```

internal/agent/device/errors/structured.go
```diff
@@ -0,0 +1,390 @@
...
+	ErrComponentOSReconciled:   ErrComponentOS,
+}
+
+// globalKeywords are patterns not covered by typed errors.
+type keywordToCode struct {
+	keyword string
+	code    codes.Code
+}
+
+var globalKeywords = []keywordToCode{
```
asafbennatan: i wouldnt about the key maps we have here like globalKeywords , elementPatterns etc.
i would ideally keep the mapping logic between a sentinal error and a structured error very simple .
i think that the having the error thrower explicitly infer them as sentinals would make that simpler

just a thought - i am aware we are very close to code freeze so feel free to ignore

PR #2331, opened 2026-01-08
EDM-701: Clarify Usages for CSR

lkatalin:
This PR:

- Addresses EDM-701 by adding further description of mandatory Usages in openapi.yaml
- Adds validation to disallow contradictory Usages
- Updates the management signer name to match previous changes in #2215 

## Summary by CodeRabbit

* **Bug Fixes**
  - Enforced stricter validation for Certificate Signing Requests: certain usages are now disallowed and required usages must be present.

* **Documentation**
  - Updated OpenAPI and generated type documentation to reflect the CSR usage requirements.
  - Refreshed embedded API specification content to align docs and tooling with the changes.


lkatalin:
We do still want to review this; not stale

lkatalin:
Please keep the PR open and I will do a rebase next week.

internal/util/validation/validation.go
```diff
@@ -349,7 +352,7 @@ func ValidateSignerName(s string) []error {
 
 	validSigners := map[string]struct{}{
 		"flightctl.io/enrollment":                {},
-		"flightctl.io/device-enrollment":         {},
+		"flightctl.io/device-management":         {},
```
celdrake: I don't know exactly how the signers are used, but it seems like this change should be aligned with https://github.com/flightctl/flightctl/blob/main/internal/config/ca/ca.go#L50

Is there any change any existing configuration could be rejected after this change if the configuration had the `flightctl.io/device-enrollment` setting which now becomes invalid?

lkatalin: Thanks @celdrake , this is a good point and I will look into this.

lkatalin: I took a look, this change is actually correct to match with the changes in https://github.com/flightctl/flightctl/pull/2215/changes, which changed the name for the signer that is in charge of issuing permanent identity certificates for devices that have approved enrollment from `DeviceEnrollmentSignerName` to `DeviceManagementSignerName` to differentiate it from the Client Bootstrap / Enrollment service signer.

However, I am not sure whether it is a breaking change, so someone else will need to evaluate that.

asafbss: There is no such signer named `flightctl.io/device-management`.
The signer `flightctl.io/device-enrollment` is the first signer for the initial management certificate.
The signer `flightctl.io/device-management-renewal` is used when the agent requests a renewal

lkatalin: There actually is a `SignerDeviceManagement`: https://github.com/flightctl/flightctl/blob/main/internal/crypto/signer/signer_device_management.go - this used to be called `SignerDeviceEnrollment`. It is the signer that signs the generated Management Certificate that is specific to the device after its enrollment is approved.

The `SignerDeviceEnrollment` signs the initial "bootstrap" certificate, otherwise known as the enrollment certificate.

I think `SignerDeviceManagementRenewal` is new and should be compatible with the others.

For more context, see https://github.com/flightctl/flightctl/pull/2215

asafbss: Ok, I think I understand the confusion now.

The signer `SignerDeviceManagement` uses the value `flightctl.io/device-enrollment` and not `flightctl.io/device-management`; that value doesn’t exist.

The variable name doesn’t match its value, which is the main source of the confusion:
https://github.com/flightctl/flightctl/blob/main/internal/config/ca/ca.go#L50

I wouldn’t recommend changing the value at this stage, as it could impact many existing cases.

We currently rely on this value to verify whether a certificate is a valid device management certificate as part of the authentication flow:

https://github.com/flightctl/flightctl/blob/efc6499a13810ecda787ec34c2a4ba798e7337b3/internal/api_server/middleware/agent_auth.go#L95
https://github.com/flightctl/flightctl/blob/main/internal/crypto/signer/common.go#L162

The reason I haven’t changed it is to avoid breaking existing devices that were issued certificates with this signer value in previous releases. Changing it now could prevent those devices from authenticating with the service.

lkatalin: Ah, got it, yes that makes sense as we don't want to break anything. It's too bad that the names don't match but better to leave it this way. I'll drop this commit and re-push without it.

lkatalin:
@asafbss I dropped the commit renaming the signer, could you please take a look again at this PR?

PR #2337, opened 2026-01-11
EDM-2637: [POC] Header-based API routing

sk-ilya:
The POC's goal: Route HTTP requests based on an API version header to different `oapi-codegen` generated servers/handlers, even when those servers expose overlapping HTTP paths - both define "/api/v1/**".

In this demo `/api/v1/devices` and `/api/v1/devices/{name}` are served for both `v1beta1` and `v1`. 

**Scope**
In scope: Routing infra and architectural approach, primarily in:
- `internal/api_server/server.go`: wiring the outer router, building per-version routers, and mounting the dispatcher under the `/api/*` subtree
- `internal/api_server/versioning/dispatcher.go`: selecting the per-version router based on version stored in request context
- `internal/api_server/versioning/router.go`: creating per-version chi routers with per-version OpenAPI validation and registering generated routes
- `internal/api_server/versioning/version.go`: middleware + request context storage for the negotiated version
- API spec layout under the "/api" dir, removal of "/api/v1" prefixes from paths (stored in `servers.url` instead)

**Out of scope**: Version negotiation, API conversion logic, the content of the v1 OpenAPI spec - a dummy implementation to support the demo

### DIFF
--- a/internal/agent/identity/tpm.go
+++ b/internal/agent/identity/tpm.go
@@ -415,8 +415,8 @@ func (t *tpmProvider) CreateManagementClient(config *base_client.Config, metrics
 		}
 	}
 
-	// Append /api/v1 to server URL for client requests since OpenAPI spec uses servers.url
-	serverURL := strings.TrimSuffix(configCopy.Service.Server, "/") + "/api/v1"
+	// Agent API OpenAPI paths already include the /api/v1 prefix.
+	serverURL := strings.TrimSuffix(configCopy.Service.Server, "/")
 	clientWithResponses, err := agent_client.NewClientWithResponses(serverURL, agent_client.WithHTTPClient(httpClient))
 	if err != nil {
 		return nil, fmt.Errorf("creating client: %w", err)
--- a/internal/api_server/server.go
+++ b/internal/api_server/server.go
@@ -158,12 +158,13 @@ func (s *Server) Run(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed loading swagger spec: %w", err)
 	}
-	// Skip server name validation
-	swagger.Servers = nil
+	// Keep swagger.Servers intact. It may carry a base path (e.g. /api/v1) used by the
+	// request validator for operation matching.
 
 	oapiOpts := oapimiddleware.Options{
-		ErrorHandler:      oapiErrorHandler,
-		MultiErrorHandler: oapiMultiErrorHandler,
+		ErrorHandler:          oapiErrorHandler,
+		MultiErrorHandler:     oapiMultiErrorHandler,
+		SilenceServersWarning: true,
 	}
 
 	// Create service handler and wrap with tracing
@@ -289,6 +290,45 @@ func (s *Server) Run(ctx context.Context) error {
 		// Version middleware parses header and stores version in context
 		r.Use(versioning.WithAPIVersion(versioning.NewDefaultRegistry()))
 
+		// Register auth validate endpoint with stricter rate limiting under /api/v1.
+		// Note: with /api/v1 mounted as a subrouter, the path seen here is "/auth/validate".
+		r.Group(func(r chi.Router) {
+			// Add conditional middleware
+			r.Use(oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapiOpts))
+			r.Use(authMiddewares...)
+			r.Use(identityMappingMiddleware.MapIdentityToDB) // Map identity to DB objects AFTER authentication
+
+			// Add auth-specific rate limiting (only if configured and enabled)
+			if s.cfg.Service.RateLimit != nil && s.cfg.Service.RateLimit.Enabled {
+				trustedProxies := s.cfg.Service.RateLimit.TrustedProxies
+				authRequests := 20      // Default auth requests limit
+				authWindow := time.Hour // Default auth window
+				if s.cfg.Service.RateLimit.AuthRequests > 0 {
+					authRequests = s.cfg.Service.RateLimit.AuthRequests
+				}
+				if s.cfg.Service.RateLimit.AuthWindow > 0 {
+					authWindow = time.Duration(s.cfg.Service.RateLimit.AuthWindow)
+				}
+				fcmiddleware.InstallRateLimiter(r, fcmiddleware.RateLimitOptions{
+					Requests:       authRequests,
+					Window:         authWindow,
+					Message:        "Login rate limit exceeded, please try again later",
+					TrustedProxies: trustedProxies,
+				})
+			}
+
+			h := transport.NewTransportHandler(serviceHandler, s.authN, authTokenProxy, authUserInfoProxy, s.authZ)
+			// Use the wrapper to handle the AuthValidate method signature
+			wrapper := &server.ServerInterfaceWrapper{
+				Handler:            h,
+				HandlerMiddlewares: nil,
+				ErrorHandlerFunc: func(w http.ResponseWriter, r *http.Request, err error) {
+					oapiErrorHandler(w, err.Error(), http.StatusBadRequest)
+				},
+			}
+			r.Get("/auth/validate", wrapper.AuthValidate)
+		})
+
 		// Dispatcher routes to appropriate version-specific router
 		// Each inner router has its own OpenAPI validation
 		r.Mount("/", versioning.VersionDispatcher{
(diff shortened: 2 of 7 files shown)


api/v1beta1/openapi.yaml
```diff
@@ -12,7 +12,7 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
 servers:
-  - url: /
+  - url: /api/v1
```
sk-ilya, line 15: Note :arrow_up: 
This is actually the intended way based on the [swagger docs](https://swagger.io/docs/specification/v3_0/api-host-and-base-path/)

internal/api_server/server.go
```diff
@@ -260,12 +283,18 @@ func (s *Server) Run(ctx context.Context) error {
...
+		// Version middleware parses header and stores version in context
+		r.Use(versioning.WithAPIVersion(versioning.NewDefaultRegistry()))
 
-		// Register all other endpoints with general rate limiting (already applied at router level)
-		// Create a custom handler that excludes the auth validate endpoint
-		customHandler := &customTransportHandler{h}
-		server.HandlerFromMux(customHandler, r)
+		// Dispatcher routes to appropriate version-specific router
+		// Each inner router has its own OpenAPI validation
+		r.Mount("/", versioning.VersionDispatcher{
```
sk-ilya, line 334: - "/" is relative to "/api/v1/"
- `VersionDispatcher` is where the request will be handled by either `v1` or `v1beta1` handler

internal/api_server/server.go
```diff
@@ -234,13 +236,34 @@ func (s *Server) Run(ctx context.Context) error {
...
+		return fmt.Errorf("failed to create v1 router: %w", err)
+	}
+
+	rV1Beta1, err := versioning.CreateV1Beta1Router(v1beta1Handler, oapiVersionOpts)
+	if err != nil {
+		return fmt.Errorf("failed to create v1beta1 router: %w", err)
+	}
+
+	// Backward-compatible /api/version endpoint (redirects to same handler as /api/v1/version)
+	router.Get("/api/version", v1beta1Handler.GetVersion)
```
sk-ilya, line 263: It’s a workaround to keep the `/api/version` endpoint while moving the `/api/v1` prefix from individual paths entries into `servers.url`. 

`/api/version` was the only path in the spec that wasn’t under `/api/v1`, so with a centralized `servers.url: /api/v1` it would  get served as `/api/v1/version` as well

internal/service/authprovider.go
```diff
@@ -223,7 +223,7 @@ func (h *ServiceHandler) PatchAuthProvider(ctx context.Context, orgId uuid.UUID,
 	}
 
 	newObj := &api.AuthProvider{}
-	err = ApplyJSONPatch(ctx, currentObj, newObj, patch, "/api/v1/authproviders/"+name)
+	err = ApplyJSONPatch(ctx, currentObj, newObj, patch, "/authproviders/"+name)
```
sk-ilya, line 226: Note

api/v1beta1/openapi.yaml
```diff
@@ -1808,7 +1808,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/Status'
 
-  /api/v1/enrollmentrequests:
```
sk-ilya, line 1811: Note: repeats for all paths

internal/agent/identity/tpm.go
```diff
@@ -415,7 +415,9 @@ func (t *tpmProvider) CreateManagementClient(config *base_client.Config, metrics
 		}
 	}
 
-	clientWithResponses, err := agent_client.NewClientWithResponses(configCopy.Service.Server, agent_client.WithHTTPClient(httpClient))
+	// Append /api/v1 to server URL for client requests since OpenAPI spec uses servers.url
+	serverURL := strings.TrimSuffix(configCopy.Service.Server, "/") + "/api/v1"
```
sk-ilya: There is a cleaner way to achieve this in `oapi-codegen` v2.5.1 (we are using 2.3) - [ server-urls: true](https://github.com/oapi-codegen/oapi-codegen/tree/v2.5.1?tab=readme-ov-file#with-server-urls)

internal/client/config.go
```diff
@@ -331,6 +331,27 @@ func WithUserAgentHeader(component string) client.ClientOption {
...
+// APIBaseURL returns the base URL for the main HTTP API client.
+//
+// The OpenAPI spec uses servers.url = /api/v1, so callers may provide either:
+//   - https://host:port
+//   - https://host:port/api/v1
+//
+// This helper ensures we only add /api/v1 once.
+func APIBaseURL(server string) string {
+	base := strings.TrimSuffix(server, "/")
+	if strings.HasSuffix(base, "/api/v1") {
```
sk-ilya, line 349: There is a cleaner way in `oapi-codegen` v2.5.1 (we are using 2.3) - [ server-urls: true](https://github.com/oapi-codegen/oapi-codegen/tree/v2.5.1?tab=readme-ov-file#with-server-urls) to generate such [utils](https://github.com/oapi-codegen/oapi-codegen/blob/main/examples/generate/serverurls/gen.go#L60)

api/v1/openapi.yaml
```diff
@@ -0,0 +1,667 @@
...
+    This is the v1 API version.
+  contact:
+    name: The Flight Control Team
+    url: https://flightctl.io
+    email: team@flightctl.io
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+servers:
+  - url: /
```
asafbennatan, line 16: use the same intended way you have changed in v1beta1?

internal/api/conversion/device.go
```diff
@@ -0,0 +1,75 @@
+package conversion
```
asafbennatan, line 1: i thought conversion was out of scope for this PR.
note - depending  on the number of supported versions this will get messy very fast (n*n-1 per type where n is the number of supported versions ) - so its important this is very organized

internal/api_server/versioning/dispatcher.go
```diff
@@ -0,0 +1,22 @@
+package versioning
+
+import (
+	"net/http"
+)
+
+// VersionDispatcher routes requests to version-specific handlers based on
+// the API version stored in the request context by WithAPIVersion middleware.
+type VersionDispatcher struct {
+	V1      http.Handler
```
asafbennatan, line 10: should this be a map ? so we dont need to add more and more fields as we support additional versions ?

internal/api_server/server.go
```diff
@@ -234,13 +237,34 @@ func (s *Server) Run(ctx context.Context) error {
...
 	)
 
-	// a group is a new mux copy, with its own copy of the middleware stack
-	// this one handles the OpenAPI handling of the service (excluding auth validate endpoint)
-	router.Group(func(r chi.Router) {
-		//NOTE(majopela): keeping metrics middleware separate from the rest of the middleware stack
-		// to avoid issues with websocket connections
-		r.Use(oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapiOpts))
-		r.Use(authMiddewares...)
+	// Create transport handlers for each API version
```
asafbennatan, line 240: consider using generics + loop to initialize the api - this will get long very quickly

internal/transport/v1/device.go
```diff
@@ -0,0 +1,143 @@
...
+
+	v1 "github.com/flightctl/flightctl/api/v1"
+	"github.com/flightctl/flightctl/internal/api/conversion"
+	"github.com/flightctl/flightctl/internal/transport"
+)
+
+// ListDevices handles GET /api/v1/devices for v1 API.
+func (h *TransportHandler) ListDevices(w http.ResponseWriter, r *http.Request, params v1.ListDevicesParams) {
+	// Convert v1 params to v1beta1
+	v1beta1Params := conversion.ListDevicesParamsV1ToV1Beta1(params)
```
asafbennatan, line 15: i assume this is because the internal representation is v1beta?
a seperate internal representation is out of scope for this PR ?

api/v1/openapi.yaml
sk-ilya: right, should've done it this way

internal/api/conversion/device.go
sk-ilya: Yes, but since it's a POC (not going to merge), it's done in a dummy way

sk-ilya: The idea is to will be to convert all API versions to a single internal representation and vice versa, but, as mentioned, it's not in scope

internal/transport/v1/device.go
sk-ilya: Yes, to both

internal/cli/global.go
```diff
@@ -48,6 +49,7 @@ func (o *GlobalOptions) Bind(fs *pflag.FlagSet) {
 	fs.StringVarP(&o.Context, "context", "c", o.Context, "Read client config from 'client_<context>.yaml' instead of 'client.yaml'.")
 	fs.StringVarP(&o.ConfigDir, "config-dir", "", o.ConfigDir, "Specify the directory for client configuration files.")
 	fs.IntVar(&o.RequestTimeout, "request-timeout", o.RequestTimeout, "Request Timeout in seconds (0 - use default OS timeout)")
+	fs.StringVar(&o.APIVersion, "api-ver", o.APIVersion, "API version to use (e.g., 'v1' or 'v1beta1'). Sets the Flightctl-API-Version header.")
```
hexfusion, line 52: today the apis are versioned together, but that won’t always be true. the client should ideally make this decision based on what it discovers from the service. exposing this as a knob is probably something we’d want to avoid long term, but for a poc it’s acceptable as long as we’re clear on the intent/experiment. like could this just be a const?

internal/client/config.go
hexfusion: we should look at the cost of updating to support this and possible trade offs and track.

internal/cli/global.go
```diff
@@ -98,7 +100,11 @@ func (o *GlobalOptions) ValidateCmd(args []string) error {
 // from the global options (config file path, organization override, etc.).
 func (o *GlobalOptions) BuildClient() (*apiclient.ClientWithResponses, error) {
 	organization := o.GetEffectiveOrganization()
-	return client.NewFromConfigFile(o.ConfigFilePath, client.WithOrganization(organization), client.WithUserAgentHeader("flightctl-cli"))
+	return client.NewFromConfigFile(o.ConfigFilePath,
+		client.WithOrganization(organization),
+		client.WithUserAgentHeader("flightctl-cli"),
+		client.WithHeader("Flightctl-API-Version", o.APIVersion),
```
hexfusion, line 106: note: we could potentially leverage the `Vary` header for ui/cli requests in environments that support cdn or corporate proxy caching. for now, let’s track this as a future optimization?

ref. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary

hexfusion (commented):
overall this makes sense given our current state. the key consideration is ensuring we have a clear plan for how we move to path-based routing. as discussed in our 1:1, we’ll want to define the observability options for the project. the sooner we can move to path-based routing, the sooner we can deprecate this interim state.

internal/cli/global.go
sk-ilya: Yep, it was mentioned in the design. Any reason not to just add `Vary: Flightctl-API-Version` to any response?

internal/cli/global.go
sk-ilya: It was just to run a quick test :)

But in general, this can be achieved without a knob. The k8s-ish way is:
```
flightctl get devices.flightctl.io          # get server-preferred ver
flightctl get devices.v1beta1.flightctl.io  # get  specific ver
```
Then the CLI can use discovery to find the route to the versioned resource (path + header).

PR #2341, opened 2026-01-12
NO-ISSUE: Check that the test-vm has IPs before provisioning.

itzikb-redhat:
Add a loop to check that the VM has the IPs before provisioning.

## Release Notes

* **Improvements**
  * VM startup now intelligently waits for IP acquisition on network interfaces instead of fixed delays, improving provisioning reliability.
  * Extended timeout window from 30 to 60 seconds with periodic polling for IP availability.
  * Enhanced interface discovery and IPv4 address tracking for both default and secondary network interfaces.


siserafin (approved):
lgtm

PR #2345, opened 2026-01-12
EDM-2918: Integrate Pruning with Agent Lifecycle

ori-amizur:
## Release Notes

* **New Features**
  * Integrated automatic image pruning into the device management lifecycle, executing after successful upgrades.

* **Improvements**
  * Image pruning now operates non-blocking, preventing failures from halting reconciliation.
  * Enhanced error handling for upgrade and pruning operations improves system resilience.

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -15,6 +15,7 @@ import (
 	"github.com/flightctl/flightctl/internal/agent/device/errors"
 	"github.com/flightctl/flightctl/internal/agent/device/fileio"
 	"github.com/flightctl/flightctl/internal/agent/device/hook"
+	imagepruning "github.com/flightctl/flightctl/internal/agent/device/image_pruning"
 	"github.com/flightctl/flightctl/internal/agent/device/lifecycle"
 	"github.com/flightctl/flightctl/internal/agent/device/os"
 	"github.com/flightctl/flightctl/internal/agent/device/policy"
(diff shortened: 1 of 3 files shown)


internal/agent/device/device.go
```diff
@@ -208,23 +212,33 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
...
-		return
 	}
 
-	if err := a.updatedStatus(ctx, desired); err != nil {
-		a.log.Warnf("Failed updating status: %v", err)
+	// execute pruning after successful spec application and update
+	// All managers have read and applied the spec changes, and the spec files have been updated
+	// Pruning errors are logged but don't block reconciliation
+	if err := a.pruningManager.Prune(ctx); err != nil {
+		a.log.Warnf("Pruning completed with errors: %v", err)
```
kkyrazis, line 240: I noticed in the previous PR that we were leaning heavily on "Don't block reconciliation" within the `Prune` implementation itself. I think the end behavior is appropriate, but I think it would be nice to have the reconciliation loop make the decision whether or not it should block vs the implementation itself. I think currently it's not possible to even have an error returned.

ori-amizur: This function is called by the engine.  The engine just schedules the calls.  It doesn't make sense to handle this logic in the engine.  So the question is: where should the handling be deferred to?

ori-amizur: Since this is the last statement in the function, there aren't any other handling options in case of error.

kkyrazis: I think this is the appropriate place to handle it. What I am suggesting is that .Prune is problematic. If every function swallows its own errors because contextually, reconciliation shouldn't fail, then I don't see the point of even returning an error. 

Regardless, I don't think this PR has any logic issues and will tag it. It was just something I noticed.

PR #2346, opened 2026-01-13
EDM-3067: Document image pruning

ori-amizur:
## Summary by CodeRabbit

* **Documentation**
  * Added Image and Artifact Pruning guidance: pruning model, reference tracking before/after updates, safe-pruning criteria, preservation rules, monitoring guidance, example scenarios, and behavior notes
  * Added configuration guidance including base and drop-in options and automatic SIGHUP-triggered reload behavior
  * Updated device lifecycle hook guidance to note drop-in reload path and precedence
  * Minor wording and formatting tweaks around drop-in configuration and tables

### DIFF
--- a/docs/user/using/managing-devices.md
+++ b/docs/user/using/managing-devices.md
@@ -687,6 +687,115 @@ spec:
 [...]
 ```
 
+## Image and Artifact Pruning
+
+The Flight Control agent can automatically remove unused container images and OCI artifacts from devices to free up disk space. This feature helps prevent storage exhaustion on edge devices with limited capacity.
+
+### How Pruning Works
+
+Image and artifact pruning operates on a "lost reference" model:
+
+1. **Reference Tracking**: Before each device update, the agent records all images and artifacts referenced in the current and desired device specifications to a file (`/var/lib/flightctl/image-artifact-references.json`).
+
+2. **Pruning Execution**: After a successful spec update, the agent compares the previously recorded references with the current device specifications. Any images or artifacts that were previously referenced but are no longer needed are eligible for removal.
+
+3. **Safe Pruning**: The agent only removes items that:
+   * Were previously recorded in the references file
+   * Are no longer referenced in the current or desired device specifications
+   * Exist locally on the device
+
+4. **Preservation**: Images and artifacts required for current operations or potential rollback (desired state) are always preserved, even if they appear in the references file.
+
+> [!NOTE]
+> Pruning runs automatically after successful device spec updates. The references file is maintained even when pruning is disabled, ensuring accurate pruning behavior when the feature is later enabled.
+
+### What Gets Pruned
+
+The pruning process considers the following types of OCI targets:
+
+* **Application Images**: Container images referenced by Compose, Quadlet, or container applications
+* **OS Images**: Operating system images specified in the device's OS configuration
+
+### Configuring Pruning
+
+Image and artifact pruning is **disabled by default**. To enable it, configure the agent's pruning settings.
+
+#### Base Configuration
+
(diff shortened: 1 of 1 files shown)


docs/user/using/managing-devices.md
```diff
@@ -687,6 +687,114 @@ spec:
...
+> Pruning runs automatically after successful device spec updates. The references file is maintained even when pruning is disabled, ensuring accurate pruning behavior when the feature is later enabled.
+
+### What Gets Pruned
+
+The pruning process considers the following types of OCI targets:
+
+* **Application Images**: Container images referenced by Compose, Quadlet, or container applications
+* **OS Images**: Operating system images specified in the device's OS configuration
+* **OCI Artifacts**: Artifacts used for volume population or other application needs
+* **Nested Targets**: Images and artifacts referenced within other images (e.g., a Compose file inside an image referencing another artifact)
```
kkyrazis: does this implementation detail need to be documented? I feel like "Application Images" covers this portion.

docs/user/using/managing-devices.md
```diff
@@ -687,6 +687,114 @@ spec:
...
+
+3. **Safe Pruning**: The agent only removes items that:
+   * Were previously recorded in the references file
+   * Are no longer referenced in the current or desired device specifications
+   * Exist locally on the device
+
+4. **Preservation**: Images and artifacts required for current operations or potential rollback (desired state) are always preserved, even if they appear in the references file.
+
+> [!NOTE]
+> Pruning runs automatically after successful device spec updates. The references file is maintained even when pruning is disabled, ensuring accurate pruning behavior when the feature is later enabled.
```
kkyrazis, line 710: Is there a max file size? I guess it probably takes a while to reach any substantial size.

docs/user/using/managing-devices.md
```diff
@@ -687,6 +687,114 @@ spec:
...
+
+```yaml
+image-pruning:
+  enabled: false
+```
+
+> [!NOTE]
+> Configuration changes take effect after the agent reloads its configuration. The agent automatically reloads configuration when it receives a SIGHUP signal. When files in `/etc/flightctl/conf.d/` are created, updated, or removed, the agent automatically sends itself a SIGHUP signal to reload configuration. You can also manually restart the agent service to apply changes.
+
+### Pruning Behavior
```
kkyrazis, line 756: it may be useful to note that as only current image references are kept, it may be useful to disable pruning if experimenting with many images across many specs.

docs/user/using/managing-devices.md
ori-amizur: No there isn't.  Initially pruning was invoked when critical disk alert was triggered, but it was removed during kickoff.

docs/user/using/managing-devices.md
ori-amizur: Do you think that "Application Images" is as clear also for an end user?

docs/user/using/managing-devices.md
ori-amizur: Isn't the explanation from L696 - L707 enough ?
Why is the second part of the question (" it may be useful to disable pruning if experimenting with many images across many specs") necessary ?

docs/user/using/managing-devices.md
kkyrazis: In my opinion, yes.

docs/user/using/managing-devices.md
kkyrazis: It's not necessary, but makes it clear that enabling pruning may cause heavy network usage and increased update times when quickly changing application images.

PR #2348, opened 2026-01-13
EDM-2981: Internal domain model

sk-ilya:
This PR introduces the first step toward separating internal code from API types by establishing a canonical internal/domain package and redirecting internal dependencies to it.

What changed

- Added internal/domain as type aliases to api/core/v1beta1.
- Updated internal code to use internal/domain instead of importing api/core/v1beta1 directly (mostly import swaps + api.X -> domain.X).
- Added lint rules to prevent new direct API imports in internal/service and internal/store

Next steps:

- Add explicit API <-> domain conversions at the transport boundary (initially passthrough).
- Clean up re-exported helpers like `var MatchExpressionsToString = v1beta1.MatchExpressionsToString`  and `domain.PatchRequest`  by moving API-agnostic utils to the domain layer
- Change type aliases (`type X = Y`) to distinct types (`type X Y`) to ensure types are not used interchangeably without conversion

Changed files (switched to domain):

```● .
  ├── .golangci.yml
  └── internal/
      ├── alert_exporter/
      │   ├── event_processor.go
      │   └── event_processor_test.go
      ├── config/
      │   ├── ca/
      │   │   └── ca.go
      │   ├── config.go
      │   └── config_test.go
      ├── console/
      │   └── console.go
      ├── domain/
      │   └── *.go (17 files)              <--- new types
      ├── instrumentation/metrics/domain/
      │   └── *.go (4 files)
      ├── kvstore/
      │   └── kvstore.go
      ├── periodic_checker/
      │   ├── publisher.go
      │   └── publisher_test.go
      ├── rollout/
      │   ├── device_selection/
      │   │   └── *.go (4 files)
      │   ├── disruption_budget/
      │   │   └── reconciler.go
      │   └── util.go
      ├── service/
      │   ├── common/
      │   │   ├── device.go
      │   │   ├── device_test.go
      │   │   └── events.go
      │   └── *.go (31 files)
      ├── store/
      │   ├── model/
      │   │   └── *.go (11 files)
      │   └── *.go (12 files)
      ├── tasks/
      │   └── *.go (20 files)
      └── worker_client/
          ├── mock_worker_client.go
          └── worker_client.go
```

## Summary by CodeRabbit

* **Refactor**
  * Unified internal data model: services, stores, tasks, events and consoles now use a consolidated domain model — behavior unchanged for end users.

* **Chores**
  * Added lint rules to prevent importing external API types in service/store layers.
  * Switched config/storage to use secure-string domain types.

* **Tests**
  * Updated tests and mocks to the new domain types to preserve coverage.

### DIFF
--- a/internal/service/enrollment_credential.go
+++ b/internal/service/enrollment_credential.go
@@ -10,9 +10,9 @@ import (
 	"encoding/pem"
 	"fmt"
 
-	api "github.com/flightctl/flightctl/api/core/v1beta1"
 	"github.com/flightctl/flightctl/internal/consts"
 	"github.com/flightctl/flightctl/internal/crypto"
+	"github.com/flightctl/flightctl/internal/domain"
 	"github.com/flightctl/flightctl/internal/util"
 	fccrypto "github.com/flightctl/flightctl/pkg/crypto"
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


hexfusion:
holding for review thanks

internal/alert_exporter/event_processor.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	api "github.com/flightctl/flightctl/api/core/v1beta1"
+	"github.com/flightctl/flightctl/internal/domain"
```
hexfusion, line 11: why do we make the version implicit vs explicit?
```suggestion
   corev1beta "github.com/flightctl/flightctl/api/core/v1beta1
```

hexfusion (changes requested):
i know this is a transition PR but the manual type enumeration is way too brittle. also I can not understand breaking from the kube naming semantics.

why `domain.Device` instead of explicit `corev1beta1.Device` and future `corev1.Device`? I agree the current usage of `api` did not make sense but I am failing to be convinced this is required. as the product moves to api versions the diff is explicit and deliberate.

hexfusion:
Overall technical direction seems reasonable, still open questions about maintenance and documentation that can be followups.

sk-ilya:
I've double checked the usage of `domain.*` in `console/console.go` and `internal/alert_exporter/event_processor.go`

1.  `console/console.go`  uses `domain.Device`  returned by service to check the device state, which is intended
2. `internal/alert_exporter/event_processor.go` - similar, uses `domain.Event*` returned by the `event.go` service

PR #2359, opened 2026-01-16
EDM-2636: fileio.Writer: Add ownership handling

keitwb:
Allow the writer to be configured with a particular uid/gid that will own all of the files and leaf dirs created by the writer.

This is going to be needed for doing rootless containers.

## Summary by CodeRabbit

* **New Features**
  * Secret mounts can specify user and group ownership; file writers now support per-instance UID/GID and explicit file creation with ownership.

* **Bug Fixes / Robustness**
  * Ownership is applied consistently for atomic writes, directory creation, temp files, copies, and unpacking.

* **Breaking Changes**
  * File I/O now uses explicit reader/writer construction with option-based root-dir configuration; prior SetRootDir/Test-root wiring and legacy append API removed.

* **Tests**
  * Tests and fixtures updated to use the new constructors and cover ownership behaviors.


internal/agent/device/fileio/writer.go
```diff
@@ -167,8 +168,24 @@ func (w *writer) RemoveContents(path string) error {
 	return nil
 }
 
+func (w *writer) OpenFile(path string, flag int, perm fs.FileMode) (*os.File, error) {
+	f, err := os.OpenFile(w.PathFor(path), flag, perm)
+	if err != nil {
+		return nil, fmt.Errorf("opening file %s: %w", path, err)
+	}
+	if err := f.Chown(w.uid, w.gid); err != nil {
+		return nil, fmt.Errorf("setting owner of %s: %w", path, err)
```
kkyrazis, line 201: I think the rabbit is still right here. This should close the file before returning.

internal/agent/device/fileio/writer.go
```diff
@@ -105,31 +120,17 @@ func (w *writer) PathFor(filePath string) string {
...
-	fopts := &fileOptions{uid: -1, gid: -1}
+	fopts := &fileOptions{uid: w.uid, gid: w.gid}
 	for _, opt := range opts {
 		opt(fopts)
 	}
 
-	var uid, gid int
-	// if rootDir is set use the default UID and GID
-	if w.rootDir != "" {
-		defaultUID, defaultGID, err := getUserIdentity()
```
kkyrazis, line 116: the rootDir is set in tests and the device simulator. Will removing the functionality affect those?

internal/agent/device/fileio/writer.go
```diff
@@ -180,6 +197,9 @@ func (w *writer) MkdirTemp(prefix string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	if err := os.Chown(path, w.uid, w.gid); err != nil {
+		return "", err
```
kkyrazis: should we clean up the tempdir on failure?

internal/agent/device/fileio/writer.go
keitwb: My goal is to make the tests not depend on that hack by updating them to specify users somehow. We'll see...

internal/agent/device/fileio/writer.go
keitwb: Fixed.

internal/agent/device/fileio/writer.go
keitwb: I'll just return it to the caller and make it responsible for the cleanup if they want. Failure of chown indicates something very wrong with the filesystem anyway so cleaning up will probably fail too.

api/core/v1beta1/openapi.yaml
```diff
@@ -4545,6 +4545,12 @@ components:
             mountPath:
               type: string
               description: Path in the device's file system at which the secret should be mounted.
+            user:
+              type: string
+              description: The file's owner, specified either as a name or numeric ID. Defaults to "root".
+            group:
+              type: string
+              description: The file's group, specified either as a name or numeric ID. Defaults to "root".
```
hexfusion, line 4553: I wonder if we want/need this much flexibility on the first iteration. Do we have an explicit ask for workloads run as many users or just root vs rootless?

keitwb: This is just to make the integration tests work without the hacks where it does special user handling when using the rootdir feature in the file writer implementation.

keitwb: See https://github.com/flightctl/flightctl/pull/2359/changes/BASE..9a716accddc16938bc1045f23830dc10147ea5a0#diff-caa1b512d4773983ae6f722aaa301e0c09a1e54d7d882786153a9af2f8836accR35

internal/agent/device/fileio/fileio.go
```diff
@@ -38,6 +38,8 @@ type Writer interface {
 	// RemoveContents removes all files and subdirectories within the given path,
 	// but leaves the directory itself intact. It is a no-op if the path does not exist.
 	RemoveContents(path string) error
+	// OpenFile opens the file at the given path, ensuring that it is owned by the correct user.
+	OpenFile(path string, flag int, perm fs.FileMode) (*os.File, error)
```
hexfusion: can we talk about the usecase of `OpenFile` in Writer interface vs `ReadFile` that can understand uid gid.
```go
u, err := user.Lookup("sam")
// get uid gid
bytes, err := something.ReadFile(filePath, WithPerms(uid, gid))
```

api/core/v1beta1/openapi.yaml
hexfusion: api should be extended to solve user requirements vs testing. there is no other way to solve this?

internal/agent/device/fileio/fileio.go
keitwb: I changed the name of this to `CreateFile` which better reflects the intent of the only place it is used.

api/core/v1beta1/openapi.yaml
keitwb: As long as we run the integration tests as non-root there is no way to set the secrets owned by root (which is an important part of the security of writing the secrets so they don't get exposed) under the rules of chown. 

The better way would probably be to run the integration tests as root until we were to have a rootless agent mode but that feels like a much more complex undertaking before feature freeze.

Also conceivably with the rootless application feature, users could want to write secrets as non-root so that rootless applications can use them, so this isn't necessarily exclusively for testing use. It is consistent with the ability to set user/group on the inline config provider.

keitwb:
I made the Reader/Writer accept an option factory which resulted in a lot of places where it had to change, but gives it a lot more flexibility when actually using this with rootless apps.

api/core/v1beta1/openapi.yaml
hexfusion: Ok so secret resource is pulled from cluster and translated into FileSpec. But then the mechanism to bring a file into an application is a mount. I guess we can trust the user to ensure the path is reasonable given the user declared. We should update docs to reflect this.

internal/agent/device/fileio/fileio.go
```diff
@@ -66,39 +65,19 @@ type ReadWriter interface {
...
 }
 
-func NewReadWriter(opts ...Option) ReadWriter {
-	rw := &readWriter{
-		reader: NewReader(),
-		writer: NewWriter(),
-	}
-	for _, opt := range opts {
-		opt(rw)
+func NewReadWriter(reader Reader, writer Writer) ReadWriter {
```
hexfusion, line 72: since its a constructor why not `NewReadWriter(WithRootDir(tmpDir))` but I get you just updated everything

kkyrazis: Overall I agree with this comment, and would have preferred not to change so much, but I get the desire for separate Read/Write options. Since NewReadWriters should only actually be created in a handful of places (for real usage), I imagine it would be simple enough to create a test util function to create ReadWriters more easily.

keitwb: Mainly because it follows the Go stdlib convention in these composite types of accepting the underlying types used rather than constructing them internally (e.g. https://pkg.go.dev/io).

keitwb: Yeah the tests are the vast majority of the usage.

PR #2360, opened 2026-01-16
EDM-2636: Executer user options

keitwb:
This will be useful when we need to setup quadlets for a non-root
user.

It currently supports setting the uid/gid and homedir. More things can
be added if needed later (e.g. the USER envvar).

## Summary by CodeRabbit

* **New Features**
  * Execute commands as a specified user (UID/GID) and set HOME on Linux

* **Refactor**
  * Executor initialization moved to a configurable constructor with functional options
  * Executor implementation made internal and public surface simplified (removed legacy helper methods)

* **Tests**
  * Added unit tests covering environment and user-handling behavior

### DIFF
--- a/pkg/executer/executer_linux.go
+++ b/pkg/executer/executer_linux.go
@@ -6,17 +6,18 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"os"
 	"os/exec"
 	"syscall"
 )
 
-func (e *CommonExecuter) CommandContext(ctx context.Context, command string, args ...string) *exec.Cmd {
+func (e *commonExecuter) CommandContext(ctx context.Context, command string, args ...string) *exec.Cmd {
 	cmd := exec.CommandContext(ctx, command, args...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Pdeathsig: syscall.SIGTERM}
 	return cmd
 }
 
-func (e *CommonExecuter) execute(ctx context.Context, cmd *exec.Cmd) (stdout string, stderr string, exitCode int) {
+func (e *commonExecuter) execute(ctx context.Context, cmd *exec.Cmd) (stdout string, stderr string, exitCode int) {
 	var stdoutBytes, stderrBytes bytes.Buffer
 	cmd.Stdout = &stdoutBytes
 	cmd.Stderr = &stderrBytes
@@ -26,6 +27,22 @@ func (e *CommonExecuter) execute(ctx context.Context, cmd *exec.Cmd) (stdout str
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid:   true,
 		Pdeathsig: syscall.SIGTERM,
+	}
+
+	// Inherit the current env if none was already specified in the cmd.
+	if len(cmd.Env) == 0 {
+		cmd.Env = os.Environ()
+	}
+	if e.homeDir != "" {
+		// Forcefully override any existing HOME envvar if homeDir is set.
+		cmd.Env = append(cmd.Env, "HOME="+e.homeDir)
+	}
+
+	if e.uid >= 0 {
+		cmd.SysProcAttr.Credential = &syscall.Credential{
+			Uid: uint32(e.uid), //nolint:gosec
+			Gid: uint32(e.gid), //nolint:gosec
+		}
 	}
 	// if context is canceled, kill the entire process group
 	cmd.Cancel = func() error {
--- a/pkg/executer/executer_other.go
+++ b/pkg/executer/executer_other.go
@@ -9,15 +9,18 @@ import (
 	"os/exec"
 )
 
-func (e *CommonExecuter) CommandContext(ctx context.Context, command string, args ...string) *exec.Cmd {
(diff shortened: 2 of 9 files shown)


pkg/executer/user.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type userExecuter struct {
+	executer Executer
+	user     string
+}
+
+var _ Executer = &userExecuter{}
+
+func NewUserExecutor(user string) *userExecuter {
```
hexfusion: optional but the constructor could be more general to accommodate further needs if yes can user support be added to the other exec impl?
```golang
func NewExecuter(opts... Option) Executer {

exec := NewExecuter(WithUser("foo"))
```

pkg/executer/user.go
```diff
@@ -0,0 +1,81 @@
...
+	command, args = u.commandArgsWithUser(command, args)
+	return u.executer.ExecuteWithContext(ctx, command, args...)
+}
+
+func (u *userExecuter) ExecuteWithContextFromDir(ctx context.Context, workingDir string, command string, args []string, env ...string) (stdout string, stderr string, exitCode int) {
+	command, args = u.commandArgsWithUser(command, args)
+	return u.executer.ExecuteWithContextFromDir(ctx, workingDir, command, args, env...)
+}
+
+func (u *userExecuter) TempFile(dir, pattern string) (f *os.File, err error) {
```
hexfusion: this feels like I/0 should exec return a user.User Object that can be fed to fileio etc?

pkg/executer/user.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+
+	if err := f.Chown(uid, gid); err != nil {
+		return nil, err
+	}
+	return f, nil
+}
+
+func (u *userExecuter) LookPath(file string) (string, error) {
```
hexfusion: same here

pkg/executer/user.go
keitwb: I changed the way it is implemented to use the proc credential attrs that does the setuid more directly. This makes it much more appropriate to use the options now so I did that.

pkg/executer/user.go
keitwb: This and LookPath weren't used anywhere so I cleaned them up.

pkg/executer/user.go
keitwb: Cleaned up.

pkg/executer/executer_other.go
```diff
@@ -9,15 +9,18 @@ import (
...
 	return cmd
 }
 
-func (e *CommonExecuter) execute(ctx context.Context, cmd *exec.Cmd) (stdout string, stderr string, exitCode int) {
+func (e *commonExecuter) execute(ctx context.Context, cmd *exec.Cmd) (stdout string, stderr string, exitCode int) {
 	var stdoutBytes, stderrBytes bytes.Buffer
 	cmd.Stdout = &stdoutBytes
 	cmd.Stderr = &stderrBytes
+	if e.uid >= 0 {
+		panic("executing under a different user is only supported on Linux")
```
hexfusion, line 22: do we really want to panic?

keitwb: Panics make it far easier to discover the issues. With the interface of the `execute` method we would have to put an error message in `stderr` and return `-1` as the exit code.

Also we don't support non-linux envs so this will also reinforce this.

pkg/executer/executer_linux.go
```diff
@@ -27,6 +28,19 @@ func (e *CommonExecuter) execute(ctx context.Context, cmd *exec.Cmd) (stdout str
 		Setpgid:   true,
 		Pdeathsig: syscall.SIGTERM,
 	}
+
+	// Inherit the current env.
+	cmd.Env = os.Environ()
+	if e.homeDir != "" {
+		cmd.Env = append(cmd.Env, "HOME="+e.homeDir)
+	}
```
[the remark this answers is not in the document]
keitwb: homeDir takes precidence.

pkg/executer/executer_other.go
```diff
@@ -9,15 +9,18 @@ import (
...
+	}
```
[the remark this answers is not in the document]
keitwb: Setting this is a programmer mistake that should be discovered immediately.

pkg/executer/executer_linux.go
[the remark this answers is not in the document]
kkyrazis: What about cases where cmd.Env has already been modified with other non os.Environ variables? Is that possible to do in this flow?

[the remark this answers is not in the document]
keitwb: @kkyrazis I changed it to only set cmd.Env if it isn't already.

pkg/executer/executer_other.go
kkyrazis: I doubt anyone external is importing our pkg/executer, but could this break their usage? Probably not since the options to configure Ids were just added correct?

keitwb: The fact that I made the type unexported now and it has to be created through a constructor is going to break any external things depending on this. If they updated that part of it to still work it would default to -1 for the user which means the current process user and would not trigger a panic.

PR #2362, opened 2026-01-18
EDM-2326: Implement/Ensure greenboot rollback flow

galel12:
### Problem
Edge devices need robust health checks during boot to detect failed OS upgrades and trigger automatic rollback via greenboot.

### Solution
Add a `flightctl-agent health` subcommand that greenboot calls to verify the agent is running properly after an OS update. This replaces the previous pure-bash health check with a hybrid bash + Go approach for better testability and maintainability.

### Changes

**Health Check (`flightctl-agent health`)**
- Verifies the agent service is enabled and active
- Reports the agent's self-reported connectivity status (informational only - does not cause rollback)

**Connectivity Status Reporting**
- Agent reports "Connected"/"Disconnected" via `sd_notify(STATUS=...)`
- Visible in `systemctl status flightctl-agent`
- User-configurable via `status-notify` config option (default: true) for deployments with expected intermittent connectivity

**Greenboot Scripts**
- Health check script calls `flightctl-agent health`
- Pre-rollback script collects diagnostics when rollback is imminent
- Shared functions library for consistent logging

**Rollback Detection and Loop Prevention**
- Fixed a bug in `consumeLatest()` where permanently failed spec versions caused a tight polling loop
- When a spec version is marked as permanently failed (e.g., invalid image URL), the agent would continuously attempt to requeue it
- The fix adds a check in `consumeLatest()` to skip requeuing versions that are in the `failedVersions` set
- This prevents unnecessary CPU usage and log spam while waiting for a new valid spec from the server

### Testing
- [x] Unit tests for health package
- [x] Manual testing on VM with greenboot
- [x] Verified `systemctl status` shows connectivity status
- [x] Verified failed spec versions no longer cause tight polling loop

### Not Included (Part 2)
- API/UI exposure of rollback events

## Summary by CodeRabbit

* **New Features**
  * Added a health command with configurable timeout, stability window, verbosity, and clearer output.
  * Added Greenboot helper scripts and a pre-rollback diagnostic hook.

* **Bug Fixes / Improvements**
  * Simplified greenboot health scripts to invoke the health command; unified logging and exit handling.
  * Improved systemd notifications and lifecycle banner flows to report connected/ready status.

* **Tests**
  * Added comprehensive unit tests for the health checker and device spec manager; updated bootstrap tests.

### DIFF
--- a/internal/agent/device/bootstrap.go
+++ b/internal/agent/device/bootstrap.go
@@ -132,7 +132,7 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
 	// This is visible via `systemctl status flightctl-agent` as StatusText.
 	if b.systemdClient != nil {
 		if err := b.systemdClient.SdNotify(ctx, "STATUS=Connected"); err != nil {
-			b.log.Debugf("Failed to notify systemd of connectivity status: %v", err)
+			b.log.Errorf("Failed to notify systemd of connectivity status: %v", err)
 		}
 	}
 
--- a/internal/agent/device/lifecycle/manager.go
+++ b/internal/agent/device/lifecycle/manager.go
@@ -219,9 +219,7 @@ func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
 	errs = m.deleteSpec(errs)
 
 	// TODO: incorporate before-reboot hooks
-	if m.systemdClient == nil {
-		m.log.Warnf("Systemd client not available, skipping reboot")
-	} else if err := m.systemdClient.Reboot(ctx); err != nil {
+	if err := m.systemdClient.Reboot(ctx); err != nil {
 		errs = append(errs, fmt.Errorf("failed to initiate system reboot: %w", err))
 	}
 	if len(errs) > 0 {
@@ -366,10 +364,8 @@ func (m *LifecycleManager) writeQRBanner(ctx context.Context, message, url strin
 		return fmt.Errorf("failed to write banner to disk: %w", err)
 	}
 
-	if m.systemdClient != nil {
-		if err := m.systemdClient.SdNotify(ctx, "READY=1"); err != nil {
-			m.log.Warnf("Failed to notify systemd: %v", err)
-		}
+	if err := m.systemdClient.SdNotify(ctx, "READY=1"); err != nil {
+		m.log.Warnf("Failed to notify systemd: %v", err)
 	}
 
 	value := os.Getenv("FLIGHTCTL_DISABLE_CONSOLE_BANNER")
--- a/internal/agent/device/spec/manager.go
+++ b/internal/agent/device/spec/manager.go
@@ -400,8 +400,10 @@ func (s *manager) consumeLatest(ctx context.Context) (bool, error) {
 		}
 		if s.lastConsumedDevice.Version() != version {
 			// Check if this version is permanently failed
-			lastVersion, _ := stringToInt64(s.lastConsumedDevice.Version())
-			if s.queue.IsFailed(lastVersion) {
+			lastVersion, err := stringToInt64(s.lastConsumedDevice.Version())
+			if err != nil {
+				s.log.Warnf("Failed to parse last consumed device version %q: %v", s.lastConsumedDevice.Version(), err)
+			} else if s.queue.IsFailed(lastVersion) {
 				s.log.Debugf("Last consumed device version %s is permanently failed, skipping requeue", s.lastConsumedDevice.Version())
 				return false, nil
 			}
--- a/internal/agent/device/spec/manager_test.go
+++ b/internal/agent/device/spec/manager_test.go
@@ -1,0 +1,88 @@
+package spec
+
+import (
+	"context"
+	"testing"
+
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func TestConsumeLatestSkipsFailedVersion(t *testing.T) {
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockPriorityQueue := NewMockPriorityQueue(ctrl)
+	mockWatcher := NewMockWatcher(ctrl)
+	log := log.NewPrefixLogger("test")
+
+	cache := newCache(log)
+	// Set current to "3" and desired to "4"
+	// getRenderedVersion() will check if desired (4) is failed
+	// If not failed, it returns current version ("3")
+	// Then consumeLatest compares lastConsumedDevice ("5") != "3", triggering requeue check
+	cache.current.renderedVersion = "3"
+	cache.desired.renderedVersion = "4"
+
+	s := &manager{
+		log:                log,
+		queue:              mockPriorityQueue,
+		cache:              cache,
+		watcher:            mockWatcher,
+		lastConsumedDevice: newVersionedDevice("5"),
+	}
+
+	ctx := context.Background()
+
+	// Mock: no new messages from watcher
+	mockWatcher.EXPECT().TryPop().Return(nil, false, nil)
+	// Mock: getRenderedVersion() checks if desired (4) is failed - return false
+	mockPriorityQueue.EXPECT().IsFailed(int64(4)).Return(false)
+	// Mock: consumeLatest checks if lastConsumed (5) is failed - return true to skip requeue
+	mockPriorityQueue.EXPECT().IsFailed(int64(5)).Return(true)
+	// Expect NO call to Add() since version 5 is failed
+
+	consumed, err := s.consumeLatest(ctx)
+	require.NoError(err)
+	require.False(consumed, "consumeLatest should return false for failed versions")
+}
+
+func TestConsumeLatestRequeuesNonFailedVersion(t *testing.T) {
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockPriorityQueue := NewMockPriorityQueue(ctrl)
+	mockWatcher := NewMockWatcher(ctrl)
+	log := log.NewPrefixLogger("test")
+
+	cache := newCache(log)
+	// Set current to "3" and desired to "4"
+	cache.current.renderedVersion = "3"
+	cache.desired.renderedVersion = "4"
+
+	lastConsumedDevice := newVersionedDevice("5")
+	s := &manager{
+		log:                log,
+		queue:              mockPriorityQueue,
+		cache:              cache,
+		watcher:            mockWatcher,
+		lastConsumedDevice: lastConsumedDevice,
+	}
+
+	ctx := context.Background()
+
+	// Mock: no new messages from watcher
+	mockWatcher.EXPECT().TryPop().Return(nil, false, nil)
+	// Mock: getRenderedVersion() checks if desired (4) is failed - return false
+	mockPriorityQueue.EXPECT().IsFailed(int64(4)).Return(false)
+	// Mock: consumeLatest checks if lastConsumed (5) is failed - return false to allow requeue
+	mockPriorityQueue.EXPECT().IsFailed(int64(5)).Return(false)
+	mockPriorityQueue.EXPECT().Add(ctx, lastConsumedDevice)
+
+	consumed, err := s.consumeLatest(ctx)
+	require.NoError(err)
+	require.True(consumed, "consumeLatest should return true when requeuing non-failed version")
+}
--- a/internal/agent/health/health_test.go
+++ b/internal/agent/health/health_test.go
@@ -1,6 +1,3 @@
-// Copyright 2025 Red Hat, Inc.
-// SPDX-License-Identifier: Apache-2.0
-
 package health
 
 import (
--- a/packaging/greenboot/flightctl-agent-pre-rollback.sh
+++ b/packaging/greenboot/flightctl-agent-pre-rollback.sh
@@ -4,6 +4,13 @@
 # Runs AFTER health checks fail and BEFORE system rolls back.
 # Installed to: /usr/lib/greenboot/red.d/40_flightctl_agent_pre_rollback.sh
 #
+# Exit codes:
+#   0 - Success (does not affect rollback)
+#   non-zero - Failure (logged but does not prevent rollback)
+#
+# Note: This script's exit code does NOT affect the rollback process.
+# Greenboot will proceed with rollback regardless of this script's outcome.
+#
 set -x -e -o pipefail
 
 # shellcheck source=packaging/greenboot/functions.sh
@@ -19,12 +26,11 @@ if [ "$(id -u)" -ne 0 ]; then
 fi
 
 log_info "=== flightctl-agent pre-rollback script started ==="
-log_info "This script runs after greenboot health checks fail, before system rollback"
 print_boot_status
 
 # Only collect debug info if rollback is imminent
 if ! grub2-editenv - list 2>/dev/null | grep -q '^boot_counter=0'; then
-    log_info "System is not scheduled to roll back"
+    log_info "System is not scheduled to rollback"
     exit 0
 fi
 


internal/agent/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+
+// WithSystemdClient sets a custom systemd client (for testing).
+func WithSystemdClient(systemd *client.Systemd) Option {
+	return func(c *checker) {
+		c.systemd = systemd
+	}
+}
+
+// NewChecker creates a new health checker with the given options.
+func NewChecker(log *log.PrefixLogger, opts ...Option) *checker {
```
hexfusion, line 77: question: should executor be on the constructor signature for testing?

internal/agent/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	c := &checker{
+		log:     log,
+		timeout: 30 * time.Second,
```
hexfusion: can we make the default timeout a const?

internal/agent/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+		output:  os.Stdout,
+	}
+	for _, opt := range opts {
+		opt(c)
+	}
+	// Default systemd client if not provided
+	if c.systemd == nil {
+		c.systemd = client.NewSystemd(&executer.CommonExecuter{})
```
hexfusion: should systemd client be on the signature of the constructor?

internal/agent/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	if !ok || statusText == "" {
+		c.printInfo("Connectivity: unknown (agent has not reported status)")
+		return
+	}
+	c.printInfo("Agent status: %s", statusText)
+}
+
+func (c *checker) printInfo(format string, args ...any) {
+	if c.output != nil && c.verbose {
+		fmt.Fprintf(c.output, "[health] "+format+"\n", args...)
```
hexfusion: question: can our logger handle this?

packaging/greenboot/flightctl-agent-pre-rollback.sh
```diff
@@ -0,0 +1,31 @@
...
+# Exit handler (preserve original exit code)
+trap 'rc=$?; [ "$rc" -ne 0 ] && echo "FAILURE" || echo "FINISHED"; exit $rc' EXIT
+
+# Exit if not root
+if [ "$(id -u)" -ne 0 ]; then
+    echo "The '${SCRIPT_NAME}' script must be run with root privileges"
+    exit 1
+fi
+
+echo "STARTED"
```
hexfusion: started what :)

packaging/greenboot/flightctl-agent-pre-rollback.sh
```diff
@@ -0,0 +1,31 @@
...
+print_boot_status
+
+# Only collect debug info if rollback is imminent
+if ! grub2-editenv - list 2>/dev/null | grep -q '^boot_counter=0'; then
+    log_info "System is not scheduled to roll back"
+    exit 0
+fi
+
+log_info "Rollback imminent - collecting debug info"
```
hexfusion: lets try to provide as much context as possible 
```suggestion
log_info "Greenboot health check failed rollback imminent - collecting debug info"
```

internal/agent/health/health.go
```diff
@@ -0,0 +1,161 @@
+// Package health provides health check functionality for the flightctl-agent.
+// Used by greenboot to verify agent health during boot.
+package health
```
hexfusion, line 3: while this is technically health its very opinionated in the context of greenboot/systemd is that an issue

packaging/greenboot/functions.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
hexfusion:
```suggestion
#!/usr/bin/env bash
```

packaging/greenboot/flightctl-agent-running-check.sh
```diff
@@ -1,99 +1,30 @@
...
+print_boot_status
 
-# Wait for flightctl-agent service to be active (failed status terminates the script)
-echo "Waiting ${TIMEOUT}s for flightctl-agent service to be active..."
-if ! wait_for_success "${TIMEOUT}" "${SUCCESS_DURATION}" verify_flightctl_agent_status ; then
-  echo "Error: flightctl-agent.service did not become active within ${TIMEOUT}s. Aborting..."
-  exit 1
+# Run health check via Go binary
+# Checks: service status + reports agent's self-reported connectivity (informational)
+if ! flightctl-agent health --timeout="${FLIGHTCTL_HEALTH_CHECK_TIMEOUT}s" --verbose; then
```
hexfusion, line 26: in the previous logic the agent had 60s to become active and then it ensured remained active for another 60s. idea being there could be logic in the agent that panics causing a restart but can not make progress. how does this work now?

galel12: You're right, I removed it initially to save time during reboots and e2e testing, but you're correct that ignoring this scenario is a real gap.

Two options:
1. **Restore the 60s stability window** like before
2. **Check `NRestarts` property** - this is a native systemd counter that tracks how many times the service has crashed and been restarted since boot. Checking `NRestarts != 0` achieves the same goal without the 60s wait.

I can implement option 2 in `health.go`:
```go
nRestarts, ok := props["NRestarts"]
if ok && nRestarts != "" && nRestarts != "0" {
    return fmt.Errorf("service has restarted %s time(s) since boot (unstable)", nRestarts)
}
```
What do you think?

hexfusion: we need validation on "is the process stable", that should remain simple. adding anything new risks regression. these are just general principles, how is sort of your problem but this logic should exist and time is simple :)

internal/agent/client/systemd.go
```diff
@@ -24,6 +26,8 @@ var (
 func NewSystemd(exec executer.Executer) *Systemd {
 	return &Systemd{
 		exec: exec,
+		// Capture NOTIFY_SOCKET early, before bootstrap unsets it
+		notifySocket: os.Getenv("NOTIFY_SOCKET"),
 	}
```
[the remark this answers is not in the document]
galel12: This is a refactoring of existing code (commit ada33d3f). The socket handling logic existed previously - I moved it into a centralized SdNotify method on the Systemd client. The abstract socket handling was not changed and works in our deployment environment. This is not a regression introduced by this PR.
The second concern (constructor pattern) is also invalid - NewSystemd existed before your changes too. You're just adding a field and method, not rewriting constructors.

cmd/flightctl-agent/main.go
```diff
@@ -50,6 +51,14 @@ func main() {
 		if err := command.Execute(); err != nil {
 			os.Exit(1)
 		}
+
+	case "health":
+		flag.CommandLine = flag.NewFlagSet("health", flag.ExitOnError)
+		command := NewHealthCommand()
+		if err := command.Execute(); err != nil {
+			os.Exit(1)
```
hexfusion, line 71: should we print the error at all or is that already done

cmd/flightctl-agent/main.go
```diff
@@ -138,11 +147,62 @@ func (s *systemInfoCmd) Execute() error {
...
+	)
+	return checker.Run(context.Background())
+}
+
 func printUsage() {
 	fmt.Printf("Usage of %s:\n", os.Args[0])
 	fmt.Println("commands:")
 	fmt.Println("  version      Display version information")
 	fmt.Println("  system-info  Display system information")
+	fmt.Println("  health       Perform health checks on agent configuration and state")
```
hexfusion: this is just a simple health check right we offer no guarantees on state or config other than it starts.

```suggestion
	fmt.Println("  health       Perform health checks on agent")
```

internal/agent/client/systemd.go
```diff
@@ -317,3 +323,33 @@ func (s *Systemd) Show(ctx context.Context, unit string, opts ...SystemdShowOpti
...
+	if s.notifySocket == "" {
+		return nil // Not running under systemd with Type=notify
+	}
+
+	socketAddr := &net.UnixAddr{
+		Name: s.notifySocket,
+		Net:  "unixgram",
+	}
+
+	conn, err := net.DialUnix(socketAddr.Net, nil, socketAddr)
```
hexfusion: to be safe we probably need a timeout here in case the socket is misbehaving 
```go
ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
defer cancel()
conn, err := d.DialContext(ctx, socketAddr.Net, socketAddr.String())
```

internal/agent/config/config.go
```diff
@@ -141,6 +143,9 @@ type Config struct {
 	// ImagePruning holds all image/artifact pruning-related configuration
 	ImagePruning ImagePruning `json:"image-pruning,omitempty"`
 
+	// StatusNotify enables connectivity status reporting to systemd via sd_notify.
+	StatusNotify bool `json:"status-notify,omitempty"`
```
hexfusion: as long as we ship as a systemd unit this will always be true. I really would like to not add a flag unless we need it. if we have a feature request we can revisit.

internal/agent/device/device_test.go
```diff
@@ -310,9 +310,11 @@ func TestRollbackDevice(t *testing.T) {
 				policyManager,
 				readWriter,
 				mockOSClient,
+				nil, // systemdClient not needed in tests
```
hexfusion: we have mocks we should use them in these cases. but for the test to be accurate we should pass either real or mocked.

internal/agent/device/device_test.go
```diff
@@ -310,9 +310,11 @@ func TestRollbackDevice(t *testing.T) {
...
 				poll.NewConfig(time.Second, 1.5),
 				func() error { return nil },
 				mockAuditLogger,
+				false, // reportConnectivityStatus disabled in tests
```
hexfusion: why? the reason we test is to ensure the actual flow. if we dont we are not testing the actual flow.

internal/agent/device/spec/manager.go
```diff
@@ -51,9 +52,11 @@ func NewManager(
 	policyManager policy.Manager,
 	deviceReadWriter fileio.ReadWriter,
 	osClient os.Client,
+	systemdClient *client.Systemd,
 	pollConfig poll.Config,
 	deviceNotFoundHandler func() error,
 	auditLogger audit.Logger,
+	statusNotify bool,
```
hexfusion: statusNotify is now a systemd concern why should spec manager care? can we move this into the systemd client.

internal/agent/device/spec/publisher.go
```diff
@@ -163,13 +170,19 @@ func (n *publisher) pollAndPublish(ctx context.Context) {
 		}
 
 		if errors.Is(err, errors.ErrNoContent) || errors.IsTimeoutError(err) {
+			// Server was reachable, just no new content
 			n.log.Debug("No new template version from management service")
+			n.notifyConnectivityStatus("Connected")
```
hexfusion: if all we need is to check connectivity during start can't we use status reporting for that? this is critical code and I can't see the reason to adjust the notify status for the lifetime of the process.

could we perform this one here? https://github.com/flightctl/flightctl/blob/main/internal/agent/device/bootstrap.go#L126

internal/agent/config/config.go
galel12: The purpose of this flag is to let operators control connectivity status reporting based on their deployment model - e.g., 'always-online' factory cameras want to see live Connected/Disconnected status in `systemctl status`, while 'expected-offline' drones don't want constant Disconnected noise.

Maybe my naming choice `StatusNotify` is confusing? It does sound like it controls the systemd notify type rather than connectivity reporting. Happy to rename to `ReportConnectivityStatus` or similar.

Or am I misunderstanding your concern?

internal/agent/device/spec/publisher.go
galel12: You're right, the connectivity status is only consumed during the greenboot sequence at boot. There's no point updating it continuously throughout the agent's lifetime.

I'll move the status reporting to bootstrap (one-time after enrollment succeeds) and remove it from the publisher.

cmd/flightctl-agent/main.go
galel12: It's already done, the health checker calls `printError()` before returning the error

internal/agent/device/device_test.go
galel12: you're right.
However, this becomes moot with the change discussed in https://github.com/flightctl/flightctl/pull/2362#discussion_r2714244696. 
Moving connectivity reporting to bootstrap means spec.Manager won't need these parameters, and the test becomes simpler.

internal/agent/device/device_test.go
galel12: > you're right. However, this becomes moot with the change discussed in [#2362 (comment)](https://github.com/flightctl/flightctl/pull/2362#discussion_r2714244696). Moving connectivity reporting to bootstrap means spec.Manager won't need these parameters, and the test becomes simpler.

internal/agent/device/spec/publisher.go
galel12: :heavy_check_mark: https://github.com/flightctl/flightctl/pull/2362/files#diff-2ff32d9394aea7f7b74c684abee6fe707a6ee4fcf0e8670f8ac398e13c039e91R134-R140

cmd/flightctl-agent/main.go
galel12: :heavy_check_mark: resolved https://github.com/flightctl/flightctl/pull/2362/changes#diff-a9b7aa45ab77404f4f04e574f1b59018faecd590a7ceac839c1d24843d5f208dR205

internal/agent/client/systemd.go
galel12: :heavy_check_mark: resolved https://github.com/flightctl/flightctl/pull/2362/changes#diff-b5d8e98b83ff127601b2de0f8bce64f8820728188d92ee2b2cd1de3ca25de09fR339-R342

internal/agent/config/config.go
galel12: @hexfusion  changed it to:

```GO
ReportConnectivityStatus bool `json:"report-connectivity-status,omitempty"`
```
But i'm not sure that's what you meant 
see https://github.com/flightctl/flightctl/pull/2362/changes#diff-cf4baf0eefa6c1dcdaafa9a397273d19e0d83b655dfdee8c3aaa3dd002d0659bR146-R149

internal/agent/device/spec/manager.go
```diff
@@ -399,6 +399,12 @@ func (s *manager) consumeLatest(ctx context.Context) (bool, error) {
 			return false, fmt.Errorf("getting rendered version: %w", err)
 		}
 		if s.lastConsumedDevice.Version() != version {
+			// Check if this version is permanently failed
+			lastVersion, _ := stringToInt64(s.lastConsumedDevice.Version())
+			if s.queue.IsFailed(lastVersion) {
+				s.log.Debugf("Last consumed device version %s is permanently failed, skipping requeue", s.lastConsumedDevice.Version())
+				return false, nil
+			}
```
galel12, line 409: Hey @hexfusion, added the fix for the tight polling loop we discussed: 7fa7df9
**The bug:** consumeLatest() was setting consumed=true after queue.Add() even when the add was silently rejected (because the version was in failedVersions). This caused a tight loop where the agent 
kept trying to requeue the same failed version.

**The fix:** Added an IsFailed() check before attempting to requeue, so we skip permanently failed versions entirely.
Current behavior:
1. Failed versions are correctly rejected and won't be reconciled
2. After reboot, the in-memory failedVersions map is empty, so a failed version will be tried once, fail again, get added back to the map, and then be correctly rejected

I think this is preferable to persisting a blocklist on disk - the one-time retry after reboot has negligible overhead and avoids disk space/cleanup complexity.

internal/agent/config/config.go
hexfusion: as per our conversation we are going to remove this flag and you can decide on the subcommand.

galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/pull/2362/changes/daf6679fa737968a252e52c649f7d67d239d35bc
@hexfusion

galel12: > as per our conversation we are going to remove this flag and you can decide on the subcommand.

I'd prefer to keep `health` - it aligns with MicroShift's `microshift healthcheck` and greenboot's own naming conventions (`greenboot-healthcheck.service`, `greenboot-default-health-checks`). The `--help` output clearly documents what it checks.

internal/agent/device/bootstrap.go
```diff
@@ -125,6 +128,14 @@ func (b *Bootstrap) Initialize(ctx context.Context) error {
 
 	b.updateStatus(ctx)
 
+	// Report connectivity status to systemd.
+	// This is visible via `systemctl status flightctl-agent` as StatusText.
+	if b.systemdClient != nil {
+		if err := b.systemdClient.SdNotify(ctx, "STATUS=Connected"); err != nil {
+			b.log.Debugf("Failed to notify systemd of connectivity status: %v", err)
```
hexfusion:
```suggestion
			b.log.Errorf("Failed to notify systemd of connectivity status: %v", err)
```

internal/agent/device/lifecycle/manager.go
```diff
@@ -220,7 +219,9 @@ func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
 	errs = m.deleteSpec(errs)
 
 	// TODO: incorporate before-reboot hooks
-	if err := m.systemdClient.Reboot(ctx); err != nil {
+	if m.systemdClient == nil {
```
hexfusion: nit: i am for being defensive but is this actually possible

internal/agent/device/lifecycle/manager.go
```diff
@@ -319,32 +320,32 @@ func (m *LifecycleManager) verifyEnrollment(ctx context.Context) (bool, error) {
...
 	url := fmt.Sprintf("%s/manage/%s", m.enrollmentUIEndpoint, m.deviceName)
-	if err := m.writeQRBanner("\nYour device is enrolled to flightctl,\nyou can manage your device scanning the above QR. or following this URL:\n%s\n\n", url); err != nil {
+	if err := m.writeQRBanner(ctx, "\nYour device is enrolled to flightctl,\nyou can manage your device scanning the above QR. or following this URL:\n%s\n\n", url); err != nil {
 		return fmt.Errorf("failed to write device management banner: %w", err)
 	}
 	return nil
 }
 
-func (m *LifecycleManager) writeQRBanner(message, url string) error {
+func (m *LifecycleManager) writeQRBanner(ctx context.Context, message, url string) error {
```
hexfusion, line 346: nit: always for plumbing ctx but are we using this?

internal/agent/device/lifecycle/manager.go
```diff
@@ -365,8 +366,10 @@ func (m *LifecycleManager) writeQRBanner(message, url string) error {
 		return fmt.Errorf("failed to write banner to disk: %w", err)
 	}
 
-	if err := m.sdNotify(); err != nil {
-		m.log.Warnf("Failed to notify systemd: %v", err)
+	if m.systemdClient != nil {
```
hexfusion: again this only adds value if its possible in practice, given the way the agent.go is declared I don't  believe it is. someday if we support non systemd modes we can reconsider?

internal/agent/device/spec/manager.go
```diff
@@ -399,6 +399,12 @@ func (s *manager) consumeLatest(ctx context.Context) (bool, error) {
 			return false, fmt.Errorf("getting rendered version: %w", err)
 		}
 		if s.lastConsumedDevice.Version() != version {
+			// Check if this version is permanently failed
+			lastVersion, _ := stringToInt64(s.lastConsumedDevice.Version())
```
hexfusion: we should handle the err case right
```suggestion
			lastVersion, err := stringToInt64(s.lastConsumedDevice.Version())
			if err....
```

internal/agent/device/spec/manager.go
hexfusion: good catch can we add a unit test to gate regression here?

internal/agent/health/health_test.go
```diff
@@ -0,0 +1,392 @@
+// Copyright 2025 Red Hat, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
hexfusion: ? :)

packaging/greenboot/flightctl-agent-pre-rollback.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+fi
+
+log_info "=== flightctl-agent pre-rollback script started ==="
+log_info "This script runs after greenboot health checks fail, before system rollback"
+print_boot_status
+
+# Only collect debug info if rollback is imminent
+if ! grub2-editenv - list 2>/dev/null | grep -q '^boot_counter=0'; then
+    log_info "System is not scheduled to roll back"
```
hexfusion:
```suggestion
    log_info "System is not scheduled to rollback"
```

packaging/greenboot/flightctl-agent-pre-rollback.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Pre-rollback script for flightctl-agent.
+# Runs AFTER health checks fail and BEFORE system rolls back.
+# Installed to: /usr/lib/greenboot/red.d/40_flightctl_agent_pre_rollback.sh
+#
+set -x -e -o pipefail
```
hexfusion, line 14: what happens if this script returns non 0 will it affect rollback. whatever the answer is please add details to this script

packaging/greenboot/flightctl-agent-pre-rollback.sh
```diff
@@ -0,0 +1,32 @@
...
+trap 'rc=$?; [ "$rc" -ne 0 ] && echo "[pre-rollback] FAILED (exit code: $rc)" || echo "[pre-rollback] FINISHED successfully"; exit $rc' EXIT
+
+# Exit if not root
+if [ "$(id -u)" -ne 0 ]; then
+    echo "The '${SCRIPT_NAME}' script must be run with root privileges"
+    exit 1
+fi
+
+log_info "=== flightctl-agent pre-rollback script started ==="
+log_info "This script runs after greenboot health checks fail, before system rollback"
```
hexfusion: question: is the second log line useful?

packaging/greenboot/functions.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
hexfusion, line 1: I am all for improvements but do we need more files to deliver via rpm. we need to keep @gshilin-sdb and @kkyrazis in the loop

packaging/rpm/flightctl.spec
```diff
@@ -626,7 +630,9 @@ fi
     /usr/lib/systemd/system/flightctl-agent.service
     /usr/lib/tmpfiles.d/flightctl.conf
     /usr/lib/tmpfiles.d/centos-buildinfo.conf
+    /usr/share/flightctl/functions/greenboot.sh
     /usr/lib/greenboot/check/required.d/20_check_flightctl_agent.sh
+    /usr/lib/greenboot/red.d/40_flightctl_agent_pre_rollback.sh
```
hexfusion, line 635: cc  @gshilin-sdb and @kkyrazis

internal/agent/device/lifecycle/manager.go
galel12: Because we're calling sd_notify in here:
```GO
if err := m.systemdClient.SdNotify(ctx, "READY=1")
```
and sd_notify() requires a ctx for the timeout we talked about i.e.:
```GO
ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
defer cancel()
d := net.Dialer{}
conn, err := d.DialContext(ctx, socketAddr.Net, socketAddr.String())
```

packaging/greenboot/flightctl-agent-pre-rollback.sh
galel12: :heavy_check_mark: resolved in 44e65c23fe34985360c45e99354c262212bb06f1

packaging/greenboot/flightctl-agent-pre-rollback.sh
galel12: :heavy_check_mark: resolved in 44e65c23fe34985360c45e99354c262212bb06f1

internal/agent/device/spec/manager.go
galel12: :heavy_check_mark: resolved in 44e65c23fe34985360c45e99354c262212bb06f1

internal/agent/device/lifecycle/manager.go
galel12: :heavy_check_mark: resolved in 44e65c23fe34985360c45e99354c262212bb06f1

internal/agent/device/lifecycle/manager.go
galel12: :heavy_check_mark: resolved in 44e65c23fe34985360c45e99354c262212bb06f1

internal/agent/device/bootstrap.go
galel12: :heavy_check_mark: resolved in 44e65c23fe34985360c45e99354c262212bb06f1

internal/agent/client/systemd.go
```diff
@@ -317,3 +323,36 @@ func (s *Systemd) Show(ctx context.Context, unit string, opts ...SystemdShowOpti
 
 	return lines, nil
 }
+
+// SdNotify sends notification messages to systemd via the NOTIFY_SOCKET.
+// The socket path is captured when NewSystemd is called, before bootstrap unsets it.
+func (s *Systemd) SdNotify(ctx context.Context, messages ...string) error {
+	if s.notifySocket == "" {
+		return nil // Not running under systemd with Type=notify
```
hexfusion, line 331: question: should we log this? do we support it?

internal/agent/health/health_test.go
galel12: :open_hands: sorry for that haha
resolved in 44e65c23fe34985360c45e99354c262212bb06f1

packaging/greenboot/flightctl-agent-pre-rollback.sh
galel12: not really
deleted in 44e65c23fe34985360c45e99354c262212bb06f1

internal/agent/device/spec/manager_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+func TestConsumeLatestSkipsFailedVersion(t *testing.T) {
```
galel12, line 12: @hexfusion Added this file to unit test the bug

packaging/greenboot/functions.sh
galel12: I'll keep them updated next time i'm changing things in the rpm!
Thanks for this feedback!

PR #2365, opened 2026-01-19
EDM-2994: API version negotiation via Flightctl-API-Version header

sk-ilya:
## Summary by CodeRabbit

* **New Features**
  * API version negotiation with explicit version headers and clear 406 responses for unsupported versions.
  * Per-version routing mounted under /api/v1 with per-version OpenAPI validation.
  * Configurable, scope-based rate limiting.

* **Chores**
  * Public endpoints now exposed under shorter paths (e.g., /devices, /repositories); clients consistently target the API v1 base path.
  * Simplified patch payload handling and codegen/tooling upgrades.

* **Tests**
  * Added version-negotiation unit/integration tests and improved E2E failure logging.

### DIFF
--- a/.spectral.yaml
+++ b/.spectral.yaml
@@ -4,13 +4,13 @@ extends:
 
 aliases:
   PathToResourceCollection:
-    - '$.paths[?(@property.match( "^/api/v1/[a-z]+$" ))]'
+    - '$.paths[?(@property.match( "^(/api/v[0-9]+[a-z0-9]*)?/[a-z]+$" ))]'
   PathToResource:
-    - '$.paths[?(@property.match( "^/api/v1/[a-z]+/{name}$" ))]'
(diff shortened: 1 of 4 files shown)


internal/api_server/versioning/router.go
```diff
@@ -0,0 +1,91 @@
...
+	routers        map[Version]chi.Router
+	fallback       Version
+	fallbackRouter chi.Router
+}
+
+// newDispatcher creates a dispatcher with the given version routers.
+// It panics if the fallback version is empty or has no corresponding router.
+func newDispatcher(routers map[Version]chi.Router, fallback Version) *dispatcher {
+	if fallback == "" {
+		panic("versioning: fallback version cannot be empty")
```
hexfusion: while we could argue this is justified do we want to rethink if this should be an error? if panic is OK should we add notes on why.

sk-ilya: I’d prefer returning an error rather than adding a note explaining why a panic was chosen :)

hexfusion (approved):

PR #2367, opened 2026-01-19
EDM-3136: token refresh is now initilized on its own and does not cause issues where multiple instance are used

asafbennatan:
Adding a bit of context as to why this was required - > QE were trying to test RBAC but they discovered that they are not able to reliabliy use differetn login tokens (representing different users and different permissions) -> the refresher structure was the root cause -> this PR fixes it

## Summary by CodeRabbit

* **New Features**
  * CLI and tools now automatically start/stop API and image-builder clients.
  * Token refresher available via client wrappers to manage auth tokens during runtime.

* **Refactor**
  * Unified client wrappers with Start/Stop lifecycle used across CLI, console, completion, simulator, and tests.
  * Fetching and apply flows updated to use the new client surface for consistent behavior and cleanup.

* **Tests**
  * Test harnesses updated to use and clean up the client lifecycle.

### DIFF
--- a/internal/cli/console.go
+++ b/internal/cli/console.go
@@ -148,7 +148,10 @@ func (o *ConsoleOptions) Run(ctx context.Context, flagArgs, passThroughArgs []st
 		return err
 	}
 
-	o.analyzeResponseAndExit(ctx, name, o.connectViaWS(ctx, config, name, client.GetAccessToken(config, o.ConfigFilePath), passThroughArgs))
+	refresher := client.NewAccessTokenRefresher(config, o.ConfigFilePath, 8080)
+	refresher.Start(ctx)
+	accessToken := refresher.GetAccessToken()
+	o.analyzeResponseAndExit(ctx, name, o.connectViaWS(ctx, config, name, accessToken, passThroughArgs))
 
 	// unreachable
 	return nil
@@ -383,6 +386,8 @@ func (o *ConsoleOptions) emitUpgradeFailureError(ctx context.Context, name strin
 		fmt.Fprintf(os.Stderr, "Error for device %s: %v\n", name, origErr)
 		return
 	}
+	c.Start(ctx)
+	defer c.Stop()
 	response, err := c.GetDeviceWithResponse(ctx, name)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error for device %s: %v\n", name, origErr)
--- a/internal/cli/get.go
+++ b/internal/cli/get.go
@@ -13,6 +13,7 @@ import (
 	imagebuilderapi "github.com/flightctl/flightctl/api/imagebuilder/v1beta1"
 	apiclient "github.com/flightctl/flightctl/internal/api/client"
 	"github.com/flightctl/flightctl/internal/cli/display"
+	"github.com/flightctl/flightctl/internal/client"
 	"github.com/flightctl/flightctl/internal/util"
 	"github.com/flightctl/flightctl/internal/util/validation"
 	"github.com/samber/lo"
@@ -378,9 +379,12 @@ func (o *GetOptions) Run(ctx context.Context, args []string) error {
 	formatter := display.NewFormatter(display.OutputFormat(o.Output))
 
 	// Create resource fetchers based on kind
-	listFetcher, singleFetcher, err := o.createFetchers(ctx, kind)
+	listFetcher, singleFetcher, stopFn, err := o.createFetchers(ctx, kind)
 	if err != nil {
 		return err
+	}
+	if stopFn != nil {
+		defer stopFn()
 	}
 
 	// Handle list case (no specific names)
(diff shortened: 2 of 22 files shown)


itzikb-redhat:
LGTM

internal/client/authorization.go
```diff
@@ -87,118 +97,127 @@ func CreateAuthProviderWithCredentials(authInfo AuthInfo, insecure bool, apiServ
...
-func (c *accessTokenRefresher) accessToken() string {
-	if c.config.AuthInfo.TokenToUse == TokenToUseIdToken {
-		return c.config.AuthInfo.IdToken
+// Stop stops the token refresh loop gracefully
+func (r *AccessTokenRefresher) Stop() {
+	r.cancelMu.Lock()
+	defer r.cancelMu.Unlock()
+	if r.cancel != nil {
+		r.cancel()
+		r.cancel = nil
```
hexfusion, line 203: ~~if your goal is to only fire cancel once sync.Once will save you the lock~~ nmd I see your writing as well.

internal/client/authorization.go
```diff
@@ -87,118 +97,127 @@ func CreateAuthProviderWithCredentials(authInfo AuthInfo, insecure bool, apiServ
...
 	var err error
-	c.provider, err = CreateAuthProvider(c.config.AuthInfo, c.config.Service.InsecureSkipVerify, c.config.Service.Server, c.callbackPort)
+	r.provider, err = CreateAuthProvider(r.config.AuthInfo, r.config.Service.InsecureSkipVerify, r.config.Service.Server, r.callbackPort)
 	return err
 }
 
-func (c *accessTokenRefresher) parseExpireTime() (time.Time, error) {
-	if c.config.AuthInfo.AccessTokenExpiry == "" {
+func (r *AccessTokenRefresher) parseExpireTime() (time.Time, error) {
+	if r.config.AuthInfo.AccessTokenExpiry == "" {
```
hexfusion, line 106: just a general state question, I see your guarding the cancelFn so I assume this can be accessed concurrently. if yes do you also need to guard the config?

internal/cli/get.go
```diff
@@ -404,9 +405,11 @@ func (o *GetOptions) createMainAPIFetchers(ctx context.Context, kind ResourceKin
 	if err != nil {
 		return nil, nil, fmt.Errorf("creating client: %w", err)
 	}
+	c.Start(ctx)
+	defer c.Stop()
```
hexfusion: can you add a note so folks understand why the client should stop before the fetcher funcs are fired.

internal/client/authorization.go
asafbennatan: mutex removed - there is no intention of multiple threads using the refresher

internal/cli/get.go
asafbennatan: this was a bug - fixed

internal/cli/console.go
```diff
@@ -148,7 +148,10 @@ func (o *ConsoleOptions) Run(ctx context.Context, flagArgs, passThroughArgs []st
 		return err
 	}
 
-	o.analyzeResponseAndExit(ctx, name, o.connectViaWS(ctx, config, name, client.GetAccessToken(config, o.ConfigFilePath), passThroughArgs))
+	refresher := client.NewAccessTokenRefresher(config, o.ConfigFilePath, 8080)
+	refresher.Start(ctx)
+	accessToken := refresher.GetAccessToken()
+	o.analyzeResponseAndExit(ctx, name, o.connectViaWS(ctx, config, name, accessToken, passThroughArgs))
```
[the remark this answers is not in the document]
asafbennatan: should be stopped by the caller

internal/client/authorization.go
```diff
@@ -87,118 +97,127 @@ func CreateAuthProviderWithCredentials(authInfo AuthInfo, insecure bool, apiServ
...
 			}
 		}
-		go c.refreshLoop(context.TODO())
+		ctx, cancel := context.WithCancel(ctx)
+		r.cancelMu.Lock()
+		r.cancel = cancel
+		r.cancelMu.Unlock()
+		go r.refreshLoop(ctx)
 	})
 }
```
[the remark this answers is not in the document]
asafbennatan: we dont want this

internal/client/config.go
```diff
@@ -331,30 +332,163 @@ func WithUserAgentHeader(component string) client.ClientOption {
...
 	if err != nil {
-		return nil, fmt.Errorf("NewFromConfig: creating HTTP client %w", err)
+		return nil, fmt.Errorf("NewImageBuilderClientFromConfig: creating HTTP client %w", err)
+	}
+
+	var refresher *AccessTokenRefresher
+	var ref imagebuilderclient.ClientOption
+
+	if config.AuthInfo.RefreshToken != "" {
+		refresher = NewAccessTokenRefresher(config, configFilePath, 8080)
```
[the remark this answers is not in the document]
asafbennatan: logic of refresher is not changed

PR #2368, opened 2026-01-19
EDM-2636: Systemd user client

keitwb:
This adds an option to the systemd client to run with the --user flag which will be necessary when managing rootless quadlets.

Nothing uses it yet.

## Summary by CodeRabbit

* **New Features**
  * Added user-mode systemd support so the app can perform service operations in both user and system contexts.
  * Public API now exposes construction of user-mode systemd clients.

* **Behavioral Changes**
  * Reboot operations are now disallowed for user-mode instances.
  * Existing system-wide behavior remains backward compatible.

* **Tests**
  * Added tests verifying correct command construction and execution for user vs system modes.

### DIFF
--- a/internal/agent/client/systemd.go
+++ b/internal/agent/client/systemd.go
@@ -27,13 +27,32 @@ func NewSystemd(exec executer.Executer) *Systemd {
 	}
 }
 
+// NewUserSystemd creates an systemd client that will use the user instance of systemd instead of
+// the system-wide instance. The user is determined by the executer -- whichever user it executes
+// commands under by default.
+func NewUserSystemd(exec executer.Executer) *Systemd {
+	return &Systemd{
+		exec:           exec,
+		isUserInstance: true,
+	}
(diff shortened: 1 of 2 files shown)


internal/agent/client/systemd.go
```diff
@@ -27,13 +27,28 @@ func NewSystemd(exec executer.Executer) *Systemd {
 	}
 }
 
+func NewUserSystemd(exec executer.Executer) *Systemd {
+	return &Systemd{
+		exec:           exec,
+		isUserInstance: true,
```
hexfusion, line 36: can we add a short doc above this what it means

hexfusion (approved):

internal/agent/client/systemd.go
keitwb: @hexfusion ok it should be clearer now how it works.

Also added two unit tests to verify.

PR #2369, opened 2026-01-19
EDM-3125: add helm application type with OpenAPI and validation support

kkyrazis:
The following test device passes validation. Note: it's not a real chart.
```yaml
spec:
    applications:
      - name: my-helm-app
        appType: helm
        image: quay.io/flightctl/nginx-chart:1.0.0
        namespace: my-app-namespace
        values:
          replicaCount: 3
          image:
            repository: nginx
            tag: "1.21"
            pullPolicy: IfNotPresent
          service:
            type: ClusterIP
            port: 80
            annotations:
              prometheus.io/scrape: "true"
          ingress:
            enabled: true
            className: nginx
            hosts:
              - host: example.com
                paths:
                  - path: /
                    pathType: Prefix
            tls:
              - secretName: example-tls
                hosts:
                  - example.com
          resources:
            limits:
              cpu: 100m
              memory: 128Mi
            requests:
              cpu: 50m
              memory: 64Mi
          autoscaling:
            enabled: true
            minReplicas: 2
            maxReplicas: 10
            targetCPUUtilizationPercentage: 80
          nodeSelector:
            kubernetes.io/os: linux
          tolerations:
            - key: "node-role.kubernetes.io/master"
              operator: "Exists"
              effect: "NoSchedule"
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                  - matchExpressions:
                      - key: topology.kubernetes.io/zone
                        operator: In
                        values:
                          - us-east-1a
                          - us-east-1b
        valuesFiles:
          - values-production.yaml
          - configs/values-overrides.yml
```

## Summary by CodeRabbit

* **New Features**
  * Added Compose, Quadlet, Container and Helm application types; Helm supports namespace, values and valuesFiles.

* **Validation / Bug Fixes**
  * Centralized per-type validation with clearer field-scoped errors; Helm valuesFiles restricted to .yaml/.yml.

* **Platform / Provider**
  * Unified provider flows and improved nested OCI/image handling across app types.

* **API / Spec**
  * OpenAPI/Swagger updated to expose new app schemas and Helm enum.

* **Tests**
  * Expanded tests covering multi-type behaviors and validation.


hexfusion:
@kkyrazis please add example usage patterns to the description

internal/util/validation/validation.go
```diff
@@ -411,6 +411,22 @@ func isTCGCSRFormat(csr []byte) bool {
 	return false
 }
 
+func ValidateHelmValuesFile(s *string, path string, maxLength int) []error {
+	if s == nil {
+		return []error{}
```
hexfusion:
```suggestion
		return nil
```

api/core/v1beta1/validation.go
```diff
@@ -996,6 +998,39 @@ func ValidateContainerImageApplicationSpec(appName string, spec *ImageApplicatio
 	return errs
 }
 
+func validateHelmApplicationFields(appName string, appType AppType, spec *ImageApplicationProviderSpec) []error {
```
hexfusion: do we we need something like `validateHelmInlineValues`

api/core/v1beta1/openapi.yaml
```diff
@@ -4678,6 +4678,37 @@ components:
...
               type: string
               description: Reference to the OCI image or artifact for the application package.
+            namespace:
+              type: string
+              description: Kubernetes namespace for helm chart installation. Only applicable when appType is 'helm'.
+              example: "my-app-namespace"
+            values:
+              type: object
+              additionalProperties: true
+              description: Helm values to pass during install/upgrade. Supports arbitrarily nested YAML structures. Only applicable when appType is 'helm'.
```
hexfusion: what does `Supports arbitrarily nested YAML structures.` means exactly like if I input invalid yaml does it fail preflight?

```yaml
values:
          replicaCount: 3
          image:
             repository: nginx
            tag: "1.21"
       pullPolicy: IfNotPresent
```

kkyrazis: I'll update the verbiage as yaml shouldn't be called out here. But the intention is an arbitrary "document" to be defined for users to override values at the application level. https://swagger.io/docs/specification/v3_0/data-models/dictionaries/#free-form-objects

@celdrake is this something that is feasible to support from the UI?

api/core/v1beta1/validation.go
kkyrazis: TBD. 

As defined no. Values is an arbitrary object, that is considered valid as long as the `values` property is well formatted json. Verification as to the validity of the contents of the values would require the agent to lint/template/apply.

api/core/v1beta1/openapi.yaml
celdrake: Should be. This is expressed in Typescript as `values?: Record<string, any>` which effectively would allow arbitrary data to be sent.

api/core/v1beta1/openapi.yaml
```diff
@@ -4826,10 +4857,12 @@ components:
         - "compose"
         - "quadlet"
         - "container"
+        - "helm"
```
celdrake, line 4947: @kkyrazis Could you clarify the following for Helm applications:

1. Can they define volumes?
2. Can they define environment variables (I assume not)?
3. Is the name mandatory in this case? I know in the past we allowed for applications with an image to not have the name defined.

A note that we're extending `ImageApplicationProviderSpec` which is combining fields for different application types. For example, `values, valuesFiles` is only valid for HelmApplication, whereas `ports,resources` would only be valid for SingleContainerApps.

I'm wondering if we may have problems with scaling this model as we add more functionality.

kkyrazis: @keitwb recently proposed this to me:
```yaml
    ApplicationProviderSpec:
      type: object
      discriminator:
        propertyName: appType
        mapping:
          compose: '#/components/schemas/ComposeApplication'
          quadlet: '#/components/schemas/QuadletApplication'
          container: '#/components/schemas/ContainerApplication'
      allOf:
        - type: object
          properties:
            appType:
              $ref: '#/components/schemas/AppType'
            name:
              type: string
              description: The application name must be 1–253 characters long, start with a letter or number, and contain no whitespace.
          required:
            - appType
        - oneOf:
            - $ref: '#/components/schemas/ComposeApplication'
            - $ref: '#/components/schemas/QuadletApplication'
            - $ref: '#/components/schemas/ContainerApplication'

    ComposeApplication:
      type: object
      allOf:
        - type: object
          properties:
            ports:
              type: array
              description: Port mappings.
              items:
                $ref: '#/components/schemas/ApplicationPort'
            resources:
              $ref: '#/components/schemas/ApplicationResources'
        - $ref: '#/components/schemas/ApplicationEnvVars'
        - $ref: '#/components/schemas/ApplicationVolumeProviderSpec'
        - oneOf:
          - type: object
            properties:
              inline:
                type: array
                description: The compose yaml file specified directly in one or more files.
                items:
            required:
              - inline
          - type: object
            properties:
              image:
                type: string
                description: Reference to the OCI artifact that contains the compose yaml file.
            required:
              - image

    QuadletApplication:
      type: object
      allOf:
        - $ref: '#/components/schemas/PodmanRuntimeSpec'
        - $ref: '#/components/schemas/ApplicationEnvVars'
        - $ref: '#/components/schemas/ApplicationVolumeProviderSpec'
        - oneOf:
          - type: object
            properties:
              inline:
                type: array
                description: A list of quadlet unit file content.
                items:
            required:
              - inline
          - type: object
            properties:
              image:
                type: string
                description: Reference to the OCI artifact that contains the quadlet unit files.
            required:
              - image

    ContainerApplication:
      type: object
      allOf:
        - $ref: '#/components/schemas/PodmanRuntimeSpec'
        - $ref: '#/components/schemas/ApplicationEnvVars'
        - $ref: '#/components/schemas/ApplicationVolumeProviderSpec'
        - type: object
          properties:
            image:
              type: string
              description: Reference to the image for this container.
            ports:
              type: array
              description: Port mappings.
              items:
                $ref: '#/components/schemas/ApplicationPort'
            resources:
              $ref: '#/components/schemas/ApplicationResources'
          required:
            - image
```

which I think is a great idea to try out. It should be compatible with the current API but  it is _a lot_ of code changes. This wasn't really possible to do until `AppType` was made required at the end of 1.0. I would like to revisit this once we have the e2e tests written out for containers, quadlets, and helm apps.

kkyrazis: And as for: 

> Can they define volumes?
> Can they define environment variables (I assume not)?
> Is the name mandatory in this case? I know in the past we allowed for applications with an image to not have the name defined.

no for all.

keitwb: Yeah we are moving farther and farther away from the image/inline provider concept with each new app type and overloading the 'image' part of it to mean more than just an OCI artifact with content that is semantically equivalent to what could be provided 'inline'.

It is a ton of rework to do this though so trying to maintain conformity to the image/inline thing as much as possible with validation preventing improper config seems to be the easiest approach until it can be thought about more.

celdrake: Totally agree, let's think this better.
In the meantime, documentation could be used to bridge the gap as to what is possible for each application type.

api/core/v1beta1/types.gen.go
```diff
@@ -21,6 +21,7 @@ const (
 const (
 	AppTypeCompose   AppType = "compose"
 	AppTypeContainer AppType = "container"
+	AppTypeHelm      AppType = "helm"
 	AppTypeQuadlet   AppType = "quadlet"
```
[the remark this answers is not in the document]
kkyrazis: This is fine for now.

api/core/v1beta1/openapi.yaml
fzdarsky: For context: The pattern we originally wanted to establish is one where the same versioned content set would either be stored unbundled in Git or bundled (tar'ed) in an OCI registry. Content would be either host configuration files (config providers) or application packages (app package providers). The "Inline" provider was later added as an ad-hoc way of specifying content (for demos, for service-generated content), but it should be functionally equivalent, i.e. a map of file names/paths to file content.

So "git" or "image" or "inline" should be different ways of transporting the same content and should be agnostic of what the content is (Helm chart, compose, etc.).

Instead, what we're talking about is how to
a) parameterize templates inside the "git" or "image" or "inline" content, whereby the mechanism (env var, Helm --set parameter, ...) is app type specific.
b) define external, pre-provisioned resources (e.g. volumes, ports) that the app has access to.

kkyrazis: I think the concern about the increasing number of options on the image provider is valid, but if the desire is to continue to use `image` as our key for "artifact reference", there isn't a nice way to do it with what currently exists. Ben's approach is theoretically compatible with the current API definition and _should_ be considered non-breaking, but it's a very large task with non-zero risk of breaking applications due to our current lack of automation surround the applications. I could introduce a `HelmApplicationProvider` in the spirit of what Ben proposed:

```yaml
   HelmApplicationProviderSpec:
      type: object
      properties:
        chart:
          type: string
        namespace:
          type: string
        values:
          type: object
          additionalProperties: true
        valuesFiles:
          type: array
          items:
            type: string
```

but I would have to use a new key for chart reference (in the example it's `chart`). I can't reuse `image` because openapi would complain about the ambiguity of the following request:

```yaml
appType: helm
image: <some-chart>
```

kkyrazis: I tried experimenting a bit with some different configurations with `oneOf` in the image application provider to select either container, or helm properties but there are issues there as all properties on both structs are optional and it can't select one without a discriminator. 

The options therefore become:

1. ApplicationSpecific providers everywhere (huge refactor and pretty risky)
2. HelmSpecific provider and keep the Image and Inline as they are. (requires not using `image`)
3. Keep extending image provider for now.

I've at least moved the container and helm specific configurations into their own schemas to make it more clear.

fzdarsky: The pattern seems to be fairly consistent to me:

```
# all app provider specs
    - name: someapp
      appType: container|compose|quadlet|helm

# image app provider for all types
      image: quay.io/org/app-package:v1 # or single app image
# inline app provider for all types except "container" (where it could be nil/empty)
      inline:
        - path: path/to/file
          content: |
            some content

# namespacing:
## helm (all others rely on app name + randomization)
      namespace: mine

# template paramteers:
## container|quadlet|compose
      envVars:
        LOG_MESSAGE: "Artifact"
## helm
      values:
        some:
          nested: objects
      valuesFiles:
        - somefile.yaml          

# volumes
## container|quadlet|compose|helm
      volumes:
        - name: model-data
          image:
            reference: quay.io/org/model-artifact:v1

# extra parameters
## container
      ports:
        - "8080:8080"
      resources:
        limits:
          cpu: "0.5"
          memory: "256m"
```

If we ignore the `ports` and `resources` params that exist only for the `container` type, then the main thing that stands out to me is the template parameters that are different for `helm`. Arguably a consequence of choosing an implementation-specific name (`envVar` vs `values`) instead of a neutral `parameters:` name.

I'm not clear why we'd need a new HelmApplicationProvider? Is that to support legacy Helm repos? Maybe we can ignore these? Or use the `image:` field but optionally allow prefixing a schema (`https://` for legacy repos)?

kkyrazis: I think the argument is that:
```
# extra parameters
## container
      ports:
      resources:
        limits:
```

could be expanded to
```
# extra parameters
## container
      ports:
        - "8080:8080"
      resources:
        limits:
          cpu: "0.5"
          memory: "256m"
## helm
      namespace:
      values:
      valuesFiles:
```
depending on perspective and the concern was that this would keep expanding.

`HelmAppProvider` isn't a requirement, it was just a suggestion for an alternative declaration that doesn't keep expanding the scope of `ImageProvider`; with the trade off being that we **can't** use the `image` key (because of openapi validation / ambiguity concerns).

I'm ok with any approach, but I think the desire to have this be valid
```yaml

appType: compose
image: quay.io/compose/package:v1

appType: quadlet
image: quay.io/quadlet/package:v1

appType: container
image: redis:v1

appType: helm
image: quay.io/helm/chart:v1
```

compared to

```yaml
appType: helm
chart: quay.io/helm/chart:v1
```

probably leans us towards what is currently implemented.

api/core/v1beta1/types.gen.go
```diff
@@ -4019,6 +4121,154 @@ func (t *Batch_Limit) UnmarshalJSON(b []byte) error {
...
+
+	if raw, found := object["volumes"]; found {
+		err = json.Unmarshal(raw, &t.Volumes)
+		if err != nil {
+			return fmt.Errorf("error reading 'volumes': %w", err)
+		}
+	}
+
+	return err
+}
```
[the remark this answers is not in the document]
kkyrazis: this is generated code, I'm not touching it.

internal/agent/device/applications/provider/utils.go
```diff
@@ -0,0 +1,182 @@
...
+		contentPath := content.Path
+		if len(contentPath) == 0 {
+			return fmt.Errorf("application content path is empty")
+		}
+		if err := rw.WriteFile(filepath.Join(appPath, contentPath), contentBytes, fileio.DefaultFilePermissions); err != nil {
+			return fmt.Errorf("writing application content: %w", err)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
kkyrazis: It's all validated prior to reaching here.

internal/agent/device/applications/provider/container.go
```diff
@@ -0,0 +1,244 @@
...
+			}
+
+			if podman.ImageExists(ctx, imageMountSpec.Image.Reference) {
+				if err := createVolumeQuadlet(rw, dir, vol.Name, imageMountSpec.Image.Reference); err != nil {
+					return fmt.Errorf("creating volume quadlet for %s: %w", vol.Name, err)
+				}
+				unit.Add(quadlet.ContainerGroup, quadlet.VolumeKey, fmt.Sprintf("%s.volume:%s", vol.Name, imageMountSpec.Mount.Path))
+			} else {
+				unit.Add(quadlet.ContainerGroup, quadlet.VolumeKey, fmt.Sprintf("%s:%s", vol.Name, imageMountSpec.Mount.Path))
+			}
```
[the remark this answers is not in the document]
kkyrazis: Mount.Path is defined as the `container` path in this usage.

internal/agent/device/applications/provider/provider.go
```diff
@@ -477,6 +689,34 @@ func FromDeviceSpec(
...
+	case v1beta1.AppTypeCompose:
+		return newComposeProvider(ctx, log, podman, providerSpec, rwFactory, cfg)
+
+	case v1beta1.AppTypeQuadlet:
+		return newQuadletProvider(ctx, log, podman, providerSpec, rwFactory, cfg)
+
+	default:
+		return nil, fmt.Errorf("%w: %s", errors.ErrUnsupportedAppType, appType)
+	}
+}
```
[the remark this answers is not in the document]
kkyrazis: this is intentional

internal/agent/device/applications/provider/container.go
```diff
@@ -0,0 +1,244 @@
...
+		return fmt.Errorf("creating app directory: %w", err)
+	}
+
+	if err := writeENVFile(p.spec.Path, p.readWriter, p.spec.EnvVars); err != nil {
+		return fmt.Errorf("writing env file: %w", err)
+	}
+
+	if err := generateQuadlet(ctx, p.podman, p.readWriter, p.spec.Path, p.spec.ContainerApp); err != nil {
+		return fmt.Errorf("generating quadlet: %w", err)
+	}
```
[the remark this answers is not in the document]
kkyrazis: installQuadlet handles this with dropins

PR #2374, opened 2026-01-20
EDM-2985: Add API version to HTTP request logs

sk-ilya:
The requested version is appended as `(v1beta1)` after the request url:

```
2026/01/20 17:08:50 [iskornya/zdOvzCLVNr] "GET https://hostname:3443/api/v1/devices&summaryOnly=false HTTP/1.1" (v1beta1) from 10.244.0.1:62635 - 200 3629B in 21.222134ms
```

## Summary by CodeRabbit

* **Chores**
  * Request logs now include an API version tag for each request to improve observability; existing logging thresholds and behavior are unchanged.

* **Documentation**
  * Clarified inline middleware notes describing version-specific middleware configuration.

* **Tests**
  * Added tests to verify API version tagging appears correctly in request logs.

### DIFF
--- a/internal/api_server/middleware/chi_logger.go
+++ b/internal/api_server/middleware/chi_logger.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"strings"
 
+	apiversioning "github.com/flightctl/flightctl/api/versioning"
 	"github.com/flightctl/flightctl/internal/api_server/versioning"
 	chimw "github.com/go-chi/chi/v5/middleware"
 )
@@ -60,7 +61,7 @@ type apiVersionLogFormatter struct {
 }
 
 func (f *apiVersionLogFormatter) NewLogEntry(r *http.Request) chimw.LogEntry {
-	apiVersion := apiVersionTag(r.Header.Get(versioning.HeaderAPIVersion))
+	apiVersion := apiVersionTag(r.Header.Get(apiversioning.HeaderAPIVersion))
 	logger := apiVersionTaggingLogger{base: f.Logger, apiVersion: apiVersion}
 
 	df := &chimw.DefaultLogFormatter{Logger: logger, NoColor: f.NoColor}
--- a/internal/api_server/middleware/chi_logger_test.go
+++ b/internal/api_server/middleware/chi_logger_test.go
@@ -6,7 +6,7 @@ import (
 	"regexp"
 	"testing"
 
-	"github.com/flightctl/flightctl/internal/api_server/versioning"
+	"github.com/flightctl/flightctl/api/versioning"
 	chimw "github.com/go-chi/chi/v5/middleware"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"


internal/api_server/middleware/chi_logger.go
```diff
@@ -0,0 +1,70 @@
...
+func (l apiVersionTaggingLogger) Print(v ...interface{}) {
+	if l.apiVersion == "" || len(v) != 1 {
+		l.base.Print(v...)
+		return
+	}
+
+	// Chi logs format: "HTTP/1.1 200 OK from 127.0.0.1". This finds " from "
+	// and injects the version before it: "HTTP/1.1 200 OK (v1alpha1) from 127.0.0.1"
+
+	// It's hacky, but allows reusing the existing logger
```
asafbss, line 30: I’m a bit concerned about mutating chi’s default access-log message by doing string injection.
This changes the log format in a way that can break existing log parsers or downstream tooling that assumes the current structure. Also, the value comes from a client-controlled header, so logging it raw into the main message feels risky unless we strictly validate/normalize it first.

Before we go down this path, can we clarify the actual goal here? If the intent is observability (e.g., understanding usage or errors by API version), could HTTP metrics be a better fit for this instead?

sk-ilya: The goal is to extend request logging to indicate the actual API version being used. Normally, just seeing the path is enough, but in our case the version is carried in the `Flightctl-API-Version` header. HTTP metrics would be nice to have, but they’re optional and won’t help troubleshoot request-level issues.

> existing log parsers or downstream tooling that assumes the current structure.

I don’t think chi’s default logger is intended to be parsed by tooling that assumes a strict, stable format.


> value comes from a client-controlled header, so logging it raw into the main message feels risky 

Good point, I'll update

sk-ilya: Done

asafbss: I’m not an expert on chi logging internals, but given the trade-offs here I think it’s worth getting more feedback from others as well (maybe @avishayt  and @fzdarsky  PTAL) before we lock into modifying the default access-log format.

Would it be possible to log the API version as a separate structured field instead of injecting it into the message itself?
That would avoid relying on string formatting and reduce the risk of breaking any downstream log consumers.

sk-ilya: Chi's log format is not structured, it's just a plain text (the example is in the PR's description). We can consider implementing our own structured logger or using an off-the-shelf one, but this out of this PR's scope

asafbss: We might consider:
https://github.com/go-chi/httplog?tab=readme-ov-file#usage

https://pkg.go.dev/github.com/go-chi/chi/middleware#Logger
_"Alternatively, look at https://github.com/goware/httplog for a more in-depth http logger with structured logging support."_

internal/api_server/middleware/chi_logger.go
```diff
@@ -0,0 +1,82 @@
+package middleware
```
hexfusion, line 1: maybe some tests? :)

internal/api_server/middleware/chi_logger.go
```diff
@@ -0,0 +1,82 @@
...
+	apiVersion := apiVersionTag(r.Header.Get(versioning.HeaderAPIVersion))
+	logger := apiVersionTaggingLogger{base: f.Logger, apiVersion: apiVersion}
+
+	df := &chimw.DefaultLogFormatter{Logger: logger, NoColor: f.NoColor}
+	return df.NewLogEntry(r)
+}
+
+func ChiLoggerWithAPIVersionTag() func(http.Handler) http.Handler {
+	stdLogger := log.New(os.Stdout, "", log.LstdFlags)
+	noColor := runtime.GOOS == "windows"
```
hexfusion, line 73: no color on windows :sob:

internal/api_server/middleware/chi_logger.go
sk-ilya: Done

PR #2376, opened 2026-01-20
EDM-3127: Add Helm OCI chart prefetching support

kkyrazis:
Wires up the dependency to manager to handle CRI vs podman based images and to handle pull configs properly.

## Summary by CodeRabbit

* **New Features**
  * Helm chart dependency support.
  * Broader runtime support: Podman and CRI images are detected and handled.
  * Per-target pull/configuration support for fetching container artifacts.

* **Refactor**
  * Centralized CLI client handling for image/chart operations.
  * Reworked pull/config flow to use a provider abstraction and wired into prefetch/pull logic.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -157,6 +157,22 @@ func (a *Agent) Run(ctx context.Context) error {
 	// create skopeo client
 	skopeoClient := client.NewSkopeo(a.log, executer, deviceReadWriter)
 
+	// create kube client
+	kubeClient := client.NewKube(a.log, executer, deviceReadWriter)
+
+	// create helm client
+	helmClient := client.NewHelm(a.log, executer, deviceReadWriter, a.config.DataDir)
+
+	// create CRI client
+	criClient := client.NewCRI(a.log, executer, deviceReadWriter)
+
+	// create CLI clients
+	cliClients := client.NewCLIClients(
+		client.WithKubeClient(kubeClient),
+		client.WithHelmClient(helmClient),
+		client.WithCRIClient(criClient),
+	)
+
 	// create systemd client
 	systemdClient := client.NewSystemd(executer)
 
@@ -255,8 +271,10 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	// create prefetch manager
 	prefetchManager := dependency.NewPrefetchManager(
-		a.log, podmanClient,
+		a.log,
+		podmanClient,
 		skopeoClient,
+		cliClients,
 		deviceReadWriter,
 		a.config.PullTimeout,
 		resourceManager,
--- a/internal/agent/device/dependency/dependency.go
+++ b/internal/agent/device/dependency/dependency.go
@@ -33,39 +33,42 @@ const (
 type OCIType string
 
 const (
-	OCITypeImage    OCIType = "Image"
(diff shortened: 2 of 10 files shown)


internal/agent/device/dependency/dependency.go
```diff
@@ -33,39 +33,42 @@ const (
 type OCIType string
 
 const (
-	OCITypeImage    OCIType = "Image"
-	OCITypeArtifact OCIType = "Artifact"
-	OCITypeAuto     OCIType = "Auto"
+	OCITypePodmanImage    OCIType = "PodmanImage"
```
kkyrazis, line 36: request: I'm not thrilled with this approach as `CRIImage` and `PodmanImage` are technically the same "thing" and all that's really different is how the image is fetched. 

I didn't feel it was appropriate to pass along a "ImageFetcher" (podman or crictl) as I would like the keep the separation of defining what the PullTarget is and letting the manager figure out the how. 

I toyed with a separate Runtime property, but that didn't seem any better either.

internal/agent/device/dependency/dependency.go
```diff
@@ -444,6 +448,35 @@ func (m *prefetchManager) Schedule(ctx context.Context, targets []OCIPullTarget)
...
+	case OCITypeCRIImage:
+		if cfg := target.Configs.Get(client.ConfigTypeContainerSecret); cfg != nil {
+			opts = append(opts, client.WithPullSecret(cfg.Path))
+		}
+		if cfg := target.Configs.Get(client.ConfigTypeCRIConfig); cfg != nil {
+			opts = append(opts, client.WithCRIConfig(cfg.Path))
+		}
+	case OCITypePodmanImage, OCITypePodmanArtifact, OCITypeAuto:
+		if cfg := target.Configs.Get(client.ConfigTypeContainerSecret); cfg != nil {
+			opts = append(opts, client.WithPullSecret(cfg.Path))
```
hexfusion: does auth change per pull or per syncDevice

internal/agent/agent.go
```diff
@@ -157,6 +157,24 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	criClient := client.NewCRI(a.log, executer, deviceReadWriter)
+
+	// create CLI clients
+	cliClients := client.NewCLIClients(
+		client.WithPodmanClient(podmanClient),
+		client.WithSkopeoClient(skopeoClient),
+		client.WithKubeClient(kubeClient),
+		client.WithHelmClient(helmClient),
+		client.WithCRIClient(criClient),
+	)
```
hexfusion, line 174: neat

internal/agent/device/dependency/dependency.go
kkyrazis: It _should_ be per syncDevice

internal/agent/agent.go
kkyrazis: updated to avoid conflicts with: https://github.com/flightctl/flightctl/pull/2377 
will evaluate again in the near future whether it makes sense to include the new changes

internal/agent/device/dependency/dependency.go
keitwb: In the context of this dependency fetcher it makes sense since it is "an image for podman" and "an image for CRI" which are very distinct.

PR #2377, opened 2026-01-20
EDM-2636: User aware clients + image pruner

keitwb:
This makes the podman, skopeo, journalctl, and systemd clients capable of being instantiated for a specific user on the system via a factory.

Everything still makes use of the process user (root) as of now, but later with rootless applications, we will run things as specific users.

Executer user handling is still pending PR and has not been hooked up yet.

The image pruner is also made user-aware in a way that is backwards compatible with the json state file on disk.

There are several more things that need to be user-aware but this is a start.

## Summary by CodeRabbit

* **New Features**
  * Per-user client support and ownership-aware image/artifact management (pruning/operations now track owners).

* **Refactor**
  * Centralized factory-based wiring for system services and container/image clients to improve reuse and testability.

* **Tools**
  * Added a user lookup utility for resolving system user IDs and home directories.

* **Documentation**
  * Updated embedded API/OpenAPI spec and metadata to reflect the new Username type.

* **Public API**
  * Several public fields now use the new Username type (affects JSON typing/serialization).


internal/agent/device/image_pruning/manager.go
```diff
@@ -230,8 +234,26 @@ func (m *manager) getImageReferencesFromSpecs(ctx context.Context, device *v1bet
 // References are stored without categorization during recording.
 // Categorization happens during pruning when we can check if they exist as images or artifacts.
 type ImageArtifactReferences struct {
-	Timestamp  string   `json:"timestamp"`
+	Timestamp string `json:"timestamp"`
+	// Deprecated in favor of ReferencesWithOwner. Old data will still be read from this field but
```
kkyrazis: image pruning is a 1.1 feature so it should be acceptable to just change the format to whatever is currently required.

internal/agent/device/image_pruning/manager.go
```diff
@@ -230,8 +234,26 @@ func (m *manager) getImageReferencesFromSpecs(ctx context.Context, device *v1bet
...
+	// Replacement for References that includes user ownership information.
+	ReferencesWithOwner []ImageRef `json:"referencesWithOwner"`
+}
+
+type ImageRef struct {
+	Image string `json:"image"`
+	Owner string `json:"owner"`
+}
+
+// GetReferences allows the
```
kkyrazis: incomplete comment

internal/agent/device/image_pruning/manager.go
```diff
@@ -810,13 +836,20 @@ func (m *manager) extractNestedTargetsFromSpec(ctx context.Context, device *v1be
 			continue
 		}
 
-		imageRef := imageSpec.Image
+		imageName := imageSpec.Image
+
+		// TODO: Update this to use specific users when available
+		podmanClient, err := m.podmanClientFactory("")
```
kkyrazis, line 828: This is because users aren't defined on the app spec yet?

pkg/executer/executer.go
```diff
@@ -46,15 +49,15 @@ func LookupUserOptions(username string) ([]ExecuterOption, error) {
 	}
 
 	return []ExecuterOption{
-		WithUIDAndGID(uid, gid),
+		WithUIDAndGID(uint32(uid), uint32(gid)), //nolint:gosec
```
kkyrazis: nit: but can you add comments why you're disabling linting?

keitwb: Added.

internal/agent/device/image_pruning/manager.go
keitwb: Removed this function.

internal/agent/device/image_pruning/manager.go
keitwb: Ok, I just replaced the old type then. Hopefully it won't trip up anybody using dev builds.

internal/agent/device/image_pruning/manager.go
keitwb: Yes, two PRs after this one will have it. Blank user means the current process user.

internal/agent/device/image_pruning/manager.go
```diff
@@ -958,25 +987,29 @@ func (m *manager) validateCapability(ctx context.Context) error {
...
-func (m *manager) removeEligibleImages(ctx context.Context, eligibleImages []string) (int, []string, error) {
+func (m *manager) removeEligibleImages(ctx context.Context, eligibleImages []ImageRef) (int, []ImageRef, error) {
 	var removedCount int
-	var removedRefs []string
+	var removedRefs []ImageRef
 	var removalErrors []error
 
-	for _, imageRef := range eligibleImages {
+	for _, ref := range eligibleImages {
+		podmanClient, err := m.podmanClientFactory(ref.Owner)
```
kkyrazis, line 996: I imagine the answer is no, but is there much overhead in having to resolve the user info for every factory call? I would guess that the overall number of owners is pretty small.

keitwb: My plan is to memoize the lookup process at the function in `userutil` so that it doesn't require any syscalls to retrieve. That will be in a subsequent PR.

pkg/executer/executer.go
```diff
@@ -30,6 +30,9 @@ type ExecuterOption func(e *commonExecuter)
 // LookupUserOptions generates a set of options to NewCommonExecuter used to execute commands as a
 // different user.
 func LookupUserOptions(username string) ([]ExecuterOption, error) {
+	if username == "" {
+		return nil, nil
```
hexfusion, line 34: question: why isnt this an error? if the user is empty why call lookup?

internal/agent/client/skopeo.go
```diff
@@ -40,6 +42,42 @@ type Skopeo struct {
 	readWriter fileio.ReadWriter
 }
 
+// SkopeoFactory creates a skopeo client. A blank username means to use the process user.
+type SkopeoFactory func(v1beta1.Username) (*Skopeo, error)
+
+func NewSkopeoFactory(log *log.PrefixLogger, testRootDir string) SkopeoFactory {
```
hexfusion, line 48: feels strange that all these factories need to reason about a testRootDir. then in general why do clients need to reason about users at all vs just the executor?

pkg/executer/executer.go
keitwb: The semantics are: 

- blank username == current process user
- "Options for the current process user" -> No options needed
- "Options for an explicit user" -> A set of options.

In a subsequent PR I'm making the blank string a const for current process user to make it more explicit.

internal/agent/client/skopeo.go
keitwb: I've already changed this to accept an executor factory instead of the testRootDir. Will be in the next PR in this series.

PR #2379, opened 2026-01-20
EDM-3144: Bug fix for files remaining in /var/lib/flightctl after decommissioning

lkatalin:
This PR updates the decommissioning implementation to delete all files in the `/var/lib/flightctl` directory and adds an integration test to check that this occurs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Decommissioning now performs a full data-directory cleanup with clearer logging and consolidated error reporting during wipe-and-reboot.

* **Tests**
  * Added integration tests verifying decommissioning removes all data files (leaving only empty directories).
  * Added unit tests for a safe executer that blocks destructive commands in test runs.

* **Chores**
  * Test harness updated to inject the safe executer to prevent destructive commands during tests.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -46,19 +46,39 @@ const (
 	gracefulShutdownTimeout = 2 * time.Minute
 )
 
+// AgentOption is a function that configures an Agent
+type AgentOption func(*Agent)
+
+// WithExecuter sets a custom executer for the agent (e.g., for testing).
+// This allows tests to inject a safe executer that prevents dangerous commands from executing.
+func WithExecuter(exec executer.Executer) AgentOption {
+	return func(a *Agent) {
+		a.executer = exec
+	}
+}
+
 // New creates a new agent.
-func New(log *log.PrefixLogger, config *agent_config.Config, configFile string) *Agent {
-	return &Agent{
+func New(log *log.PrefixLogger, config *agent_config.Config, configFile string, opts ...AgentOption) *Agent {
+	a := &Agent{
 		config:     config,
 		configFile: configFile,
 		log:        log,
+		executer:   executer.NewCommonExecuter(), // default executer
 	}
+
+	// Apply all options
+	for _, opt := range opts {
+		opt(a)
+	}
+
+	return a
 }
 
 type Agent struct {
 	config     *agent_config.Config
 	configFile string
 	log        *log.PrefixLogger
+	executer   executer.Executer
 }
 
 func (a *Agent) GetLogPrefix() string {
@@ -137,7 +157,8 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log.Infof("Using persisted CSR for enrollment")
 	}
 
-	rootExecuter := executer.NewCommonExecuter()
+	// Use the executer configured during agent construction
+	exec := a.executer
 
 	// create enrollment client
 	enrollmentClient, err := newEnrollmentClient(a.config, a.log)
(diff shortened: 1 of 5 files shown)


internal/agent/agent.go
```diff
@@ -181,13 +198,10 @@ func (a *Agent) Run(ctx context.Context) error {
 		client.WithCRIClient(criClient),
 	)
 
-	// create systemd client
-	rootSystemdClient := client.NewSystemd(rootExecuter)
-
```
lkatalin, line 195: Note: This moved above to line 182

lkatalin:
Keep open please

lkatalin:
Reopening

lkatalin:
@Samudelacruz77 @hexfusion please review when you get a chance, this is a bug I fixed a while back (decommissioning was not wiping all relevant files before reboot) but the PR has been open.

Quick context:

- The PR wipes all files instead of just spec
- Added an integration test to make sure this happens correctly. As part of that test, realized we need a `test Executor` because the integration test calling reboot as part of decom was rebooting my host machine
- Integration test also exposed a problem where a decommission would result in `409 conflict` - IIUC this is because in the original `DecommissionDevice` function there was a race condition such that the resource version would sometimes update between reading from the DB and writing changes to the DB (lmk if this part is a reasonable change)

internal/agent/agent.go
```diff
@@ -137,7 +144,14 @@ func (a *Agent) Run(ctx context.Context) error {
 		a.log.Infof("Using persisted CSR for enrollment")
 	}
 
-	rootExecuter := executer.NewCommonExecuter()
+	// Use test executer if set (for integration tests), otherwise use real executer
+	var exec executer.Executer
```
kkyrazis: would we be able to apply the options pattern here? We tend to prefer that over Set overrides if possible.

```go
a := agent.New(/*elided*/, agent.WithExecutor(myExec))
a.Run(ctx) 
```

internal/agent/device/lifecycle/manager.go
```diff
@@ -228,20 +233,33 @@ func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
...
-func (m *LifecycleManager) deleteSpec(errs []error) []error {
-	if err := m.deviceReadWriter.RemoveFile(filepath.Join(m.dataDir, "desired.json")); err != nil {
-		m.log.Errorf("Failed to delete desired.json: %v", err)
-		errs = append(errs, fmt.Errorf("failed to delete desired.json: %w", err))
+// deleteAllDataDirFiles removes all files in the data directory recursively
+func (m *LifecycleManager) deleteAllDataDirFiles() error {
+	m.log.Infof("Deleting all files in data directory: %s", m.dataDir)
+
+	// Read all top-level entries in the data directory
+	entries, err := m.deviceReadWriter.ReadDir(m.dataDir)
```
kkyrazis, line 241: is it just that we want to preserve the dataDir itself while allowing for the removal of all nested directories?

internal/agent/device/lifecycle/manager.go
```diff
@@ -201,7 +201,7 @@ func (m *LifecycleManager) updateWithErrorCondition(ctx context.Context, errs []
 }
 
 // point of no return - wipes management cert and keys
-func (m *LifecycleManager) wipeAndReboot(ctx context.Context) error {
+func (m *LifecycleManager) wipeAndReboot(ctx context.Context, decommissioning *v1beta1.DeviceDecommission) error {
```
kkyrazis: was this parameter change necessary?

lkatalin: Strictly, no - we would need it later once we've got a few types of decommissioning that we support. I'm going to remove it for simplicity.

internal/agent/device/lifecycle/manager.go
lkatalin: That was my impression, but if there's no downside to wiping the directory itself then we could do that instead.

kkyrazis: It's probably safer to take your approach.

kkyrazis (approved):

PR #2390, opened 2026-01-22
EDM-2636: User-aware lifecycle handler + podman monitor

keitwb:
This makes the compose and quadlet app lifecycle handlers mostly aware of users (though everything is still treated as root for now).

For the quadlets lifecycle handler, now that there are multiple systemd instances involved, we need to do all of the daemon reloading on a per-action basis instead of once for everything.

Also the podman monitor has the beginnings of user handling, but still needs more to fire up multiple event watchers for each user.

## Summary by CodeRabbit

* **New Features**
  * Actions now include an owning user and are grouped per-user.
  * Per-user executer support enables operations to run with user-specific permissions.

* **Refactor**
  * Centralized file I/O via a root ReadWriter factory and wired client factories for consistent, safer access.
  * Lifecycle and container managers create per-action resources on demand for better isolation and error handling.

* **Tests**
  * Test suites updated to use factory-based dependency injection.

### DIFF
--- a/internal/agent/device/applications/lifecycle/compose.go
+++ b/internal/agent/device/applications/lifecycle/compose.go
@@ -18,16 +18,16 @@ const (
 var _ ActionHandler = (*Compose)(nil)
 
 type Compose struct {
-	podman *client.Podman
-	writer fileio.Writer
-	log    *log.PrefixLogger
+	podmanFactory client.PodmanFactory
+	writerFactory fileio.ReadWriterFactory
+	log           *log.PrefixLogger
 }
 
-func NewCompose(log *log.PrefixLogger, writer fileio.Writer, podman *client.Podman) *Compose {
+func NewCompose(log *log.PrefixLogger, rwFactory fileio.ReadWriterFactory, podmanFactory client.PodmanFactory) *Compose {
 	return &Compose{
-		podman: podman,
-		writer: writer,
-		log:    log,
+		podmanFactory: podmanFactory,
+		writerFactory: rwFactory,
+		log:           log,
 	}
 }
 
@@ -36,12 +36,22 @@ func (c *Compose) add(ctx context.Context, action *Action) error {
 	projectName := action.ID
 	c.log.Debugf("Starting application: %s projectName: %s path: %s", appName, projectName, action.Path)
 
-	if err := c.ensurePodmanVolumes(ctx, action.Volumes, appName); err != nil {
+	podman, err := c.podmanFactory(action.User)
+	if err != nil {
+		return fmt.Errorf("creating podman client: %w", err)
+	}
+
+	writer, err := c.writerFactory(action.User)
+	if err != nil {
+		return fmt.Errorf("creating writer: %w", err)
+	}
+
+	if err := c.ensurePodmanVolumes(ctx, action.Volumes, appName, podman, writer); err != nil {
 		return fmt.Errorf("creating volumes: %w", err)
 	}
 
 	noRecreate := true
-	if err := c.podman.Compose().UpFromWorkDir(ctx, action.Path, projectName, noRecreate); err != nil {
+	if err := podman.Compose().UpFromWorkDir(ctx, action.Path, projectName, noRecreate); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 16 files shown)


internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -78,20 +78,30 @@ func (q *Quadlet) add(ctx context.Context, action *Action) error {
...
 		return fmt.Errorf("target name: %w", err)
 	}
-	services, err := q.systemdManager.ListDependencies(ctx, target)
+
+	systemctl, err := q.systemdFactory(action.User)
+	if err != nil {
+		return fmt.Errorf("creating systemd client: %w", err)
+	}
+
+	if err := systemctl.DaemonReload(ctx); err != nil {
```
kkyrazis: I _think_ it would be safer to gather users observed and then run daemon-reload in the same place that it was previously, but for all observed users. 

The initial problem as I remember it is that all disk modifications have occurred for all applications prior to invoking any action. So once a single `daemon-reload` occurs, all other application's states are affected. 

In cases where you're adding an app and upgrading/removing another, that could be problematic.

Assuming I have have a service that starts like this:
```
sudo systemctl list-dependencies demo.target
demo.target
● ├─demo-sleeper-one.service
● └─demo-sleeper-two.service
```

then add a new quadlet application and update the initial quadlet to remove `demo-sleeper-two` (the equivalent of `rm /etc/systemd/system/demo-sleeper-two.service`)


If the new quadlet application is added first, then it will call `daemon-reload` and the initial `demo.target` will appear as:
```
sudo systemctl list-dependencies demo.target
demo.target
● └─demo-sleeper-one.service
```

which would mean we fail to stop the service that was removed on the upgrade step.

keitwb: Ok, this should be in the right order now and broken down by user.

internal/agent/device/applications/lifecycle/compose.go
```diff
@@ -36,12 +36,22 @@ func (c *Compose) add(ctx context.Context, action *Action) error {
...
+		return fmt.Errorf("creating writer: %w", err)
+	}
+
+	if err := c.ensurePodmanVolumes(ctx, action.Volumes, appName, podman, writer); err != nil {
 		return fmt.Errorf("creating volumes: %w", err)
 	}
 
 	noRecreate := true
-	if err := c.podman.Compose().UpFromWorkDir(ctx, action.Path, projectName, noRecreate); err != nil {
+	if err := podman.Compose().UpFromWorkDir(ctx, action.Path, projectName, noRecreate); err != nil {
```
kkyrazis, line 54: asked offline, but confirmed that path changing is coming in a future PR so all of this still should work with no user.

internal/agent/device/applications/lifecycle/quadlet.go
```diff
@@ -206,92 +215,89 @@ func (q *Quadlet) cleanResources(ctx context.Context, action *Action) error {
...
+			return fmt.Errorf("systemd daemon reload: %w", err)
+		}
+
+		// Add requires daemon reload to be called prior to performing any service starting
+		for _, a := range slices.Concat(byType.Adds, byType.Updates) {
+			if err := q.add(ctx, a, systemctl); err != nil {
+				return fmt.Errorf("adding: %w", err)
+			}
+		}
+		if len(byType.Unknown) > 0 {
```
kkyrazis: might as well do this validation prior to attempting any real work.

internal/agent/device/applications/manager.go
```diff
@@ -129,7 +141,11 @@ func (m *manager) BeforeUpdate(ctx context.Context, desired *v1beta1.DeviceSpec)
 }
 
 func (m *manager) resolvePullSecret(desired *v1beta1.DeviceSpec) (*client.PullConfig, error) {
-	secret, found, err := client.ResolvePullConfig(m.log, m.readWriter, desired, pullAuthPath)
+	rootRW, err := m.rwFactory("")
```
kkyrazis, line 144: will we need to be resolving secrets/configs per user?

internal/agent/device/applications/lifecycle/quadlet.go
keitwb: Done.

internal/agent/device/applications/manager.go
keitwb: Yet to be determined. I'm thinking probably yes but I'm going to come back to it one the rootless quadlets are working in general.

PR #2393, opened 2026-01-22
EDM-3100: formalize system-info key registry

asafbss:
It partially reverts the approach introduced in [117f47e](https://github.com/flightctl/flightctl/pull/2364/commits/117f47ef020ddd8c32f4a60e299ac06f801d6235), replacing ad-hoc key handling with a shared source of truth for built-in and runtime (conditional) keys.

## Summary by CodeRabbit

* **Improvements**
  * Better validation and error reporting for system information requests; unknown keys are reported while collection continues.
  * Newly added collectors are detected and trigger fresh collection so new data is picked up promptly.

* **Refactor**
  * Centralized management of system-info keys and dynamic default key composition for more consistent behavior.

* **Tests**
  * Tests updated to use centralized keys and cover new validation and error scenarios.


internal/agent/device/systeminfo/common/types.go
```diff
@@ -0,0 +1,135 @@
+package common
+
+import "sort"
+
+const (
+	// Info key constants for system information collection
+	ArchitectureKey = "architecture"
```
kkyrazis, line 7: What about making these all 
```
type infoKey string
```

and then only exporting the consts? Theoretically you won't need the runtime check at that point, but you would have to move this back into the systeminfo package though and perhaps that can't be done because the circular dependency in config?

asafbss:
> I just want to point out that the design of systemInfo was intentionally to post a static representation of the system ONCE when the agent starts. This was done to reduce abuse by folks trying to use the flightctl db for state for their application. I would be very careful to change this. But a backdoor is it will update status on reload.
> 
> Saying that if we want to make a special case for this it might make sense short term. but then I wonder should the management cert have its own status?

Ack — but I don’t want to change the systemInfo contract this close to release.
This was a deliberate design choice from the original proposal and discussed in the design meeting (@avishayt  suggestion), and we aligned on moving the management cert info into systemInfo at that time.
Happy to revisit post-release if we still think a dedicated cert status block makes more sense.

internal/agent/device/systeminfo/common/types.go
asafbss: Agreed a typed key would be nicer long-term.
However, system-info keys originate both from config ([]string) and from runtime-registered collectors, so we still need runtime validation anyway — a type alone can be bypassed via casting.
Moving this back into systeminfo also  introducing a config ↔ systeminfo cycle, which is why it lives in a small common pkg.
Introducing a new key type would require wider changes across the systeminfo package itself, which I wanted to avoid here in favor of a minimal, contained fix.

hexfusion:
> > I just want to point out that the design of systemInfo was intentionally to post a static representation of the system ONCE when the agent starts. This was done to reduce abuse by folks trying to use the flightctl db for state for their application. I would be very careful to change this. But a backdoor is it will update status on reload.
> > Saying that if we want to make a special case for this it might make sense short term. but then I wonder should the management cert have its own status?
> 
> Ack — but I don’t want to change the systemInfo contract this close to release. This was a deliberate design choice from the original proposal and discussed in the design meeting (@avishayt suggestion), and we aligned on moving the management cert info into systemInfo at that time. Happy to revisit post-release if we still think a dedicated cert status block makes more sense.

My main point was the status reports once when the agent starts is that an issue and does this PR change that.

internal/agent/device/systeminfo/common/types.go
```diff
@@ -0,0 +1,135 @@
+package common
```
hexfusion, line 1: I see the circular dependency but would it be better to just fix it? I believe this all stems from systeminfo depending on config. but we can get around that in agent.go. can be a followup if it feels like too much

```go
	// register reloader with reload manager
	reloadManager.Register(agent.ReloadConfig)
	reloadManager.Register(systemInfoManager.ReloadConfig) 

    // let annon func take the dep not systeminfo
    reloadManager.Register(func(ctx context.Context, cfg *agent_config.Config) error {                                                                                                                   
      return systemInfoManager.ReloadConfig(ctx, systeminfo.Config{                                                                                                                              
          InfoKeys:          cfg.SystemInfo,                                                                                                                                                     
          CustomKeys:        cfg.SystemInfoCustom,                                                                                                                                               
          CollectionTimeout: cfg.SystemInfoTimeout,                                                                                                                                              
      })                                                                                                                                                                                         
    })  
```

asafbss:
> > > I just want to point out that the design of systemInfo was intentionally to post a static representation of the system ONCE when the agent starts. This was done to reduce abuse by folks trying to use the flightctl db for state for their application. I would be very careful to change this. But a backdoor is it will update status on reload.
> > > Saying that if we want to make a special case for this it might make sense short term. but then I wonder should the management cert have its own status?
> > 
> > 
> > Ack — but I don’t want to change the systemInfo contract this close to release. This was a deliberate design choice from the original proposal and discussed in the design meeting (@avishayt suggestion), and we aligned on moving the management cert info into systemInfo at that time. Happy to revisit post-release if we still think a dedicated cert status block makes more sense.
> 
> My main point was the status reports once when the agent starts is that an issue and does this PR change that.

This PR doesn’t change that behavior. The only time systemInfo is updated post-start is on config reload / cert rotation, which was introduced in https://github.com/flightctl/flightctl/pull/2364

internal/agent/device/systeminfo/common/types.go
asafbss: This isn’t fully isolated to config / ReloadConfig.

There’s additional coupling inside systeminfo itself (e.g., system_info.go around here:
https://github.com/flightctl/flightctl/blob/8c4b1f06f9241f63ae7fc9d67f41a7ef652de9d1/internal/agent/device/systeminfo/system_info.go#L739).

We can restructure things to break the cycle, but it would be fairly easy to make it fragile over time. I’m a bit concerned it’ll be hard to maintain if we push too much logic reshaping here.

Your call

hexfusion: tagged the PR this can all be internal to systeminfo in the future

PR #2398, opened 2026-01-23
EDM-2636: Add application user config

keitwb:
This adds a single config option to the ApplicationProviderSpec for a user. If it is not specified it will default to the current agent process user which is generally root.

If it is specified, the application will be run underneath that user. For quadlets this eventually means rootless podman containers managed by the user instance of systemd (not implemented yet).

The user has been hooked up in several places where it would be needed but many things still have no awareness of it and will be changed in later PRs.

## Summary by CodeRabbit

* **New Features**
  * Apps can specify a system user to run as (runAs), with defaults when unspecified.
  * Added ResourceMonitor types for CPU, memory and disk monitoring and alerting.
  * OCI image pulls and prefetching are tracked per-user for isolated pulls and cleanup.

* **Changes**
  * User-aware defaults applied across app management and runtime lookups.


api/core/v1beta1/openapi.yaml
```diff
@@ -4672,6 +4672,11 @@ components:
               description: The application name must be 1–253 characters long, start with a letter or number, and contain no whitespace.
             appType:
               $ref: '#/components/schemas/AppType'
+            user:
```
hexfusion: can we bring this up in the Monday meeting. the feature does not describe a requirement for a user per application. but instead privileged vs non or rootful vs rootless. while it makes sense for the agent to be able able to declare what the rootless user on the agent config level this might not be what we want.

keitwb: The primary issue is that quadlets/podman has far too much nuance around the notion of "privilege" to reduce it down to a single flag. The main thing that the agent is incapable of doing right now is run them under a specific user other than the same user as the agent (root).

hexfusion: yeah agreed

api/core/v1beta1/openapi.yaml
```diff
@@ -4672,6 +4672,11 @@ components:
               description: The application name must be 1–253 characters long, start with a letter or number, and contain no whitespace.
             appType:
               $ref: '#/components/schemas/AppType'
+            runAs:
```
hexfusion, line 4675: :+1:

hexfusion: question: do we need any validation on this string?

internal/agent/device/dependency/dependency.go
```diff
@@ -627,21 +686,19 @@ func (m *prefetchManager) Cleanup() {
...
 
-	if tmpDir == "" {
-		var err error
-		tmpDir, err = m.podmanClient.GetImageCopyTmpDir(ctx)
-		if err != nil {
-			m.log.Warnf("Failed to get image copy tmpdir: %v", err)
-			return nil
-		}
-		m.tmpDir = tmpDir
+	podman, err := m.podmanFactory(user)
```
hexfusion, line 693: do we cache these clients at all? probably fairly light but just curious.

keitwb: That is definitely going to be a future PR. I've tried to centralize everything under the `useruril` package so that it will be easy to memoize lookups there since that info is effectively static on a running host.

api/core/v1beta1/openapi.yaml
keitwb: I think it is probably reasonable to look up any users specified and make sure they exist during the validation stage. Applications will eventually error out when it tries to do things with it but earlier seems better. Will be a follow up PR.

api/core/v1beta1/openapi.yaml
```diff
@@ -4672,6 +4672,11 @@ components:
...
+              type: string
+              description: The username of the system user this application should be run under. This is not the same as the user within any containers of the application (if applicable). Defaults to the user that the agent runs as (generally root) if not specified.
```
celdrake, line 4677: @keitwb according to our discussion yesterday, the description is now outdated relative to what the default value is.

keitwb: Yes, this is going to be changed in a PR probably today.

PR #2418, opened 2026-01-26
NO-ISSUE: api: add discriminator field to application provider specs

hexfusion:


celdrake:
@hexfusion I see that the discriminator is optional. Is that for backwards compatibility?
Would the client (Eg. the UI) be expected to fill it in for new apps? The UI could work without changes otherwise.

hexfusion:
yeah I am experimenting on if we can do this type of thing and not break.

PR #2426, opened 2026-01-27
EDM-2636: Fix up PodmanMonitor

keitwb:
Make it operate multiple instances for each podman instance required.

Also make Username a newtype to make passing it as an arg less risky.

## Summary by CodeRabbit

* **Bug Fixes**
  * Skip unnecessary user lookup for the current process user, reducing errors and overhead.

* **New Features / Refactor**
  * Per-user application event monitoring with queued lifecycle actions for improved isolation and reliability.

* **Other**
  * Consolidated explicit user identity handling across APIs, rendering, file ownership, image references, and validation (typed username usage).

* **Tests**
  * Updated tests for per-user scenarios, boot-time event replay, and adjusted fixtures.


test/e2e/configuration/inline_config_test.go
```diff
@@ -279,7 +280,7 @@ func newFileSpec(path string, mode *int, user *string, group *string, content st
 	return v1beta1.FileSpec{
 		Path:    path,
 		Mode:    mode,
-		User:    user,
+		User:    lo.EmptyableToPtr(v1beta1.Username(lo.FromPtr(user))),
```
hexfusion: boring is good, idiomatic golang is simple to reason about. do we need this?

```go
if user != nil && *user != "" {
    u := v1beta1.Username(*user)
    User = &u
}
```

keitwb: I checked every use of it and there is not a single place we are distinguishing between nil and empty string so I just removed the pointer type altogether and it is even simpler now. Made group the same way for consistency.

internal/agent/device/applications/podman_monitor.go
kkyrazis, line 1: In https://github.com/flightctl/flightctl/pull/2416 I tried to combine the kube and podman monitors into a single base monitor so that I could share logic.  should I just don't do that?

keitwb: If it doesn't do user handling, no. Running it for separate podman instances per user completely changes the structure and nature of it.

keitwb: We can probably reuse some of it after feature freeze when we have time to refactor things.

kkyrazis (approved):

PR #2444, opened 2026-01-30
EDM-2636: Update docs for rootless quadlets

keitwb:
## Summary by CodeRabbit

* **Documentation**
  * Clarified the RunAs field for Quadlet and container apps (optional; default behavior) and standardized appType usage across examples.
  * Expanded examples and guidance for running workloads rootless with the flightctl user, including notes on linger/loginctl and when rootless execution is recommended or unnecessary.
  * Updated public property descriptions to document RunAs behavior.
* **Chores**
  * Added RunAs, cgroup(s), cpu/CPU, and non-root to the global spelling dictionary.

### DIFF
--- a/docs/user/using/managing-devices.md
+++ b/docs/user/using/managing-devices.md
@@ -692,6 +692,8 @@ spec:
   applications:
   - name: wordpress
     image: quay.io/flightctl-demos/wordpress-app:v1.2.3
+    appType: container
+    runAs: flightctl
     envVars:
       WORDPRESS_DB_HOST: "mysql"
       WORDPRESS_DB_USER: "user"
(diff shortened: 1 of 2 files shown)


docs/user/using/managing-devices.md
```diff
@@ -1099,6 +1102,23 @@ spec:
 
 Quadlet applications use [Podman Quadlet](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html) to manage containers as native systemd services. This allows full integration with systemd's dependency management, restart policies, resource limits, and logging.
 
+We recommend running regular workloads under a rootless podman instance under the provisioned
+`flightctl` user's systemd instance on the agent machine. You do this by specifying the `runAs:
```
kkyrazis: I would just make it very explicit that `flightctl` is a user that we automatically create for non root application usage. 
Optionally mention that other users can be used, but they must have the linger option (and whatever else we need) enabled.

keitwb: Take a look now.

PR #2445, opened 2026-01-30
EDM-2636: Rootless Quadlets Merge to 1.1

keitwb:
Respect the new runAs flag for quadlet applications.

Non-root quadlets get written to the user's ~/.config dir instead of /etc. Otherwise they are managed and run basically the same as root quadlets are.

The RPM package installation will create a new regular user called 'flightctl' that will have a home dir at '/home/flightctl'. Reusing the /var/lib/flightctl dir was too tricky due to the selinux labels on that dir. Eventually this user will serve as the default user for applications to run as if not specified.

(cherry picked from commit d6dba1d35e36d4451c7046451e8fd87efcc5e692)


celdrake:
@keitwb This PR can be closed. The commit was backport to release-1.1 via https://github.com/flightctl/flightctl/pull/2490

PR #2446, opened 2026-01-30
NO-ISSUE: fix agent update e2e test after structured error message ch…

DakCrowder:
This test has become flaky due to non-deterministic errors returned by the agent.  Basically what can happen is deeply wrapped errors can return different values in an inconsistent way, resulting in a flaky test.  This change modifies the test to reflect consistent parts of the error message to get things passing.

## Summary by CodeRabbit

* **Tests**
  * Simplified update-test assertions to focus on core failure messages instead of many legacy substrings.
  * Removed extensive legacy checks and commented workaround code.
  * Added explicit rollback verification by confirming the device returns to its previous rendered version.

### DIFF
--- a/test/e2e/agent/agent_update_test.go
+++ b/test/e2e/agent/agent_update_test.go
@@ -243,10 +243,6 @@ var _ = Describe("VM Agent behavior during updates", func() {
 				return device.Spec.Os.Image != initialImage
 			}, TIMEOUT)
 
-			// There is currently a bug https://issues.redhat.com/browse/EDM-1365
-			// that prevents the device from rolling back to the initial image
-			// When that bug is fixed, the following assertions will need to change.
-
 			harness.WaitForDeviceContents(deviceId, "device status should indicate updating failure", func(device *v1beta1.Device) bool {
 				return e2e.ConditionExists(device, v1beta1.ConditionTypeDeviceUpdating, v1beta1.ConditionStatusFalse, string(v1beta1.UpdateStateError))
 			}, LONGTIMEOUT)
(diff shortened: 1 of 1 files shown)


test/e2e/agent/agent_update_test.go
```diff
@@ -276,20 +272,14 @@ var _ = Describe("VM Agent behavior during updates", func() {
...
-					return device.Spec.Os.Image == initialImage
-				}, TIMEOUT)
-			*/
+				ContainSubstring("While ApplyingUpdate"),
+				ContainSubstring("applications failed"),
+				ContainSubstring("internal error occurred"),
+			))
+
+			harness.WaitForDeviceContents(deviceId, "device image should be reverted to the old image", func(device *v1beta1.Device) bool {
+				return device.Spec.Os.Image == initialImage
```
hexfusion: application level failure should not result in OS rollback. is that what is happening? or is this greenboot rollback flow?

DakCrowder: I had uncommented an older check that was incorrect / had assumptions that were maybe changed, updated the condition to just look at rendered version

siserafin (approved):

test/e2e/agent/agent_update_test.go
siserafin: @hexfusion The device here is attempting to upgrade to v8, which has an invalid embedded application compose (bad service image and command). What would be the expected behavior in this case?

hexfusion: embedded applications might need to be best effort in the context of OS upgrade. OS rollback is a dangerous operation and probably be reserved for hard failure (agent cant start). so I believe what we should consider an embedded application status as Error but the version/upgrade a success.

saying that I don't love having embedded apps treated differently. so does this mean we should revisit how application level errors result in rollback? should an application just be an error status?

cc @keitwb @kkyrazis

DakCrowder:
Closing as test has been disabled until root cause of non-deterministic test are addressed

test/e2e/agent/agent_update_test.go
galel12: @hexfusion  as far as i understand this scenario tests internal rollback - no OS reboot, no greenboot involved.

Agent boots into v8, compose validation fails, agent marks version as failed. OS stays v8 but agent reports OutOfDate because the spec couldn't be fully applied.
Greenboot doesn't trigger because agent is still running - only the application layer failed.

PR #2447, opened 2026-01-30
EDM-3224: Add deferred application dependencies support

kkyrazis:
This PR adds the ability to defer application resolution until OS update **IF** an OS update is scheduled. To do this, I refactored the provider's to attempt to gather as many artifacts as possible (e.g. Quadlet's are valid, but Helm apps aren't) so that as much pre-validation can be done prior to OS updates. After reboot, the update continues and will install any applications that are now considered valid. If they are still missing dependencies after the update, it's considered a failed spec. 

## Summary by CodeRabbit

* **New Features**
  * Runtime dependency checks for apps (container/compose/helm/quadlet/embedded) with deferral when an OS update is pending.
  * Improved OCI collection/prefetch options to control behavior during updates and avoid failing pulls unnecessarily.
  * Backoff/retry for Helm and container image pulls; Helm can extract images for prefetch.
  * Kubeconfig and Kubernetes client resolution deferred until first use.
  * App removal made idempotent (cleanup proceeds even if not previously tracked).

* **Documentation**
  * Guidance added about deferring dependency validation during OS updates.


internal/agent/device/spec/manager.go
```diff
@@ -430,6 +430,9 @@ func (s *manager) GetDesired(ctx context.Context) (*v1beta1.Device, bool, error)
 		}
 		s.log.Tracef("Requeuing current desired spec from disk version: %s", desired.Version())
 		s.queue.Add(ctx, desired)
+		if s.lastConsumedDevice == nil {
```
kkyrazis, line 433: Note: this was causing weird behavior on reboot where the agent wouldn't recognize the 'lastConsumedDevice' and any retryable error that would cause a rollback wouldn't restart from the correct spec.

internal/agent/device/applications/manager.go
```diff
@@ -336,211 +340,45 @@ func (m *manager) Shutdown(ctx context.Context, state shutdown.State) error {
...
-			needsRequeue = true
-		}
-		allNestedTargets = allNestedTargets.Add(targetUser, targets...)
-	}
-
-	return allNestedTargets, needsRequeue, activeAppNames, nil
-}
-
-// collectNestedTargetsForApp extracts nested OCI targets from a single image-based application.
-func (m *manager) collectNestedTargetsForApp(
```
kkyrazis, line 367: all of this was moved into the individual providers that we have now.

docs/user/using/managing-devices.md
```diff
@@ -716,6 +716,20 @@ The following dependencies must be installed on the device:
 > [!NOTE]
 > The agent automatically detects whether `kubectl` or `oc` is available (with preference for `kubectl`) and uses the first one found.
 
+#### Dependency Resolution with OS Updates
+
+When updating a device's OS image, you may include new dependencies (such as `helm`, `kubectl`, or `crictl`) that are required by applications in the device specification. The Flight Control agent handles this gracefully by deferring dependency validation when an OS update is pending.
```
hexfusion: I sort of feel like this section be condensed into a sentence. the assumption is this just works.

internal/agent/client/kube.go
```diff
@@ -34,6 +36,14 @@ func WithBinary(binary string) KubernetesOption {
 	}
 }
 
+// WithKubeKubeconfigPath pre-configures the kubeconfig path, skipping resolution.
+func WithKubeKubeconfigPath(path string) KubernetesOption {
+	return func(opts *kubernetesOptions) {
+		opts.kubeconfigPath = path
+		opts.kubeconfigSet = true
```
hexfusion, line 43: question: is the fact that the path != "" mean its set or do we need set explicitly

internal/agent/client/kube.go
```diff
@@ -237,7 +256,8 @@ func (k *Kube) ResolveKubeconfig() (string, error) {
 				return "", fmt.Errorf("check KUBECONFIG path: %w (checked: %s)", err, strings.Join(checkedPaths, ", "))
 			}
 			if exists {
-				return path, nil
+				// KUBECONFIG is set and valid - return empty to let tools use the env var
+				return "", nil
```
hexfusion: i sometimes pause when this is the return of function or method. its not an error and not a usable response. not saying it needs changed.

internal/agent/device/applications/controller.go
```diff
@@ -43,6 +47,11 @@ func (c *Controller) Sync(ctx context.Context, current, desired *v1beta1.DeviceS
 	c.log.Debug("Syncing device applications")
 	defer c.log.Debug("Finished syncing device applications")
 
+	osUpdatePending, err := c.specManager.IsOSUpdatePending(ctx)
```
hexfusion: I think we have a method in specManager that can check this based on the in memory spec. passing around the spec manager should be the last resort imo.

internal/agent/device/applications/controller.go
```diff
@@ -69,37 +78,52 @@ func (c *Controller) Sync(ctx context.Context, current, desired *v1beta1.DeviceS
...
 		}
 	}
 
-	for _, provider := range diff.Ensure {
-		log.Debugf("Ensuring application: %s", provider.Name())
-		if err := manager.Ensure(ctx, provider); err != nil {
+	for _, p := range diff.Ensure {
+		if err := p.EnsureDependencies(ctx); err != nil {
+			if isDeferredError(osUpdatePending, err) {
+				log.Infof("Deferring ensuring app %s until after OS update: %v", p.Name(), err)
```
hexfusion:
```suggestion
				log.Infof("Deferring ensuring app %s dependencies until after OS update: %v", p.Name(), err)
```

docs/user/using/managing-devices.md
```diff
@@ -716,6 +716,9 @@ The following dependencies must be installed on the device:
 > [!NOTE]
 > The agent automatically detects whether `kubectl` or `oc` is available (with preference for `kubectl`) and uses the first one found.
 
+> [!NOTE]
+> When an OS update is pending, the agent defers dependency validation, allowing you to deploy new application types alongside an OS image that includes the required dependencies. The agent will first apply the OS update and reboot, then validate and deploy applications after booting into the new OS.
+
```
hexfusion, line 721: :tada:

internal/agent/device/applications/manager.go
```diff
@@ -53,14 +56,36 @@ func NewManager(
...
 		appDataCache:       provider.NewAppDataCache(),
 	}
 }
 
+func isDeferrableAppError(osUpdatePending bool, err error) bool {
+	return osUpdatePending && errors.Is(err, errors.ErrAppDependency)
+}
+
+func (m *manager) validateProviderDeps(ctx context.Context, p provider.Provider) error {
+	osUpdatePending, err := m.specManager.IsOSUpdatePending(ctx)
```
hexfusion: could we learn about this `BeforeUpdate`?

internal/agent/device/applications/controller.go
kkyrazis: The spec manager currently has two methods:
 `IsOSUpdate` which simply does the in memory version checks and
`CheckOsReconciliation` which returns whether or not we're in the desired image.

The second function is all that is really needed (`if notReconciled { defer } else { error }`), but I liked checking `IsOSUpdate` first to avoid unnecessary client calls. `IsOSUpdatePending` simply combines the two and short circuits if there is no update. I've managed to reduce the usage to just the applications manager, so I'll see if I can just inject this information via `BeforeUpdate` and completely remove the dependency.

internal/agent/device/applications/provider/helm.go
```diff
@@ -19,20 +21,73 @@ import (
...
+			OptionFn: client.WithPullSecret,
+		},
+	)
+}
+
+var helmBinaryDeps = []dependencyBins{
+	{variants: []string{"helm"}},
+	{variants: []string{"kubectl", "oc"}},
+	{variants: []string{"crictl"}},
+}
```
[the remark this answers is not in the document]
kkyrazis: The expectation is that images will be needed for a vast majority of charts.

PR #2449, opened 2026-02-01
EDM-3005: Add running cs10-bootc e2e tests. Fix certificate handling

gshilin-sdb:
- Expand cs10-bootc e2e tests to run full test suite (sanity && !slow + sanity && slow) matching the comprehensive testing that cs9-bootc already has
- Fix certificate file handling in agent image build process - gracefully handle missing certificate files instead of failing the build
- Disable SELinux for all lint containers

## Summary by CodeRabbit

* **Chores**
  * Reworked E2E workflow to separate non-agent vs agent runs with slow/no-slow variants and adjusted node allocations; updated workflow discovery.  
  * Disabled SELinux relabeling and simplified container volume mounts for linting/docs commands.  
  * Made image build certificate copy conditional to skip when optional certs are absent.  
* **Tests**
  * Added "agent" label to multiple E2E tests and widened a rollback error expectation to include an additional internal error message.

### DIFF
--- a/.github/workflows/pr-e2e-testing.yaml
+++ b/.github/workflows/pr-e2e-testing.yaml
@@ -191,7 +191,7 @@ jobs:
           - os_id: cs9-bootc
             backend_type: helm
             label_filter: 'sanity && !agent'
-            ginkgo_total_nodes: 2
+            ginkgo_total_nodes: 1
           # Agent metrics
           ## CS9 Agent metrics
           - os_id: cs9-bootc
--- a/Makefile
+++ b/Makefile
@@ -461,7 +461,7 @@ check-rpmlint:
 .output/stamps/lint-openapi: api/core/v1beta1/openapi.yaml  api/imagebuilder/v1alpha1/openapi.yaml .spectral.yaml
 	@mkdir -p .output/stamps
 	@echo "Linting OpenAPI specs"
-	podman run --rm -it -v $(shell pwd):/workdir:Z docker.io/stoplight/spectral:6.14.2 lint --ruleset=/workdir/.spectral.yaml --fail-severity=warn \
+	podman run --rm -it --security-opt label=disable -v $(shell pwd):/workdir docker.io/stoplight/spectral:6.14.2 lint --ruleset=/workdir/.spectral.yaml --fail-severity=warn \
 		/workdir/api/core/v1beta1/openapi.yaml \
 		/workdir/api/imagebuilder/v1alpha1/openapi.yaml
 	@touch .output/stamps/lint-openapi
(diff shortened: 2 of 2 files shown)


.github/workflows/pr-e2e-testing.yaml
```diff
@@ -194,10 +194,14 @@ jobs:
             backend_type: helm
             ginkgo_total_nodes: 1
             label_filter: 'sanity && slow'
+          - os_id: cs10-bootc
+            backend_type: helm
+            label_filter: 'sanity && !slow'
+            ginkgo_total_nodes: 4
```
sk-ilya: Do we want to run **all** of the e2e tests for cs9 and cs10?

- There are quite a few tests that are not touching the agent, or agent is not the main component being tested (e.g., cli tests) - these will run twice as well
-  I'm not sure we want to run all of the agent tests on every PR for both cs9 and cs10 as part of sanity

gshilin-sdb: Yes, to be sure nothing breaks it. In any case they're all running simultaneously, so there is no time penalty.

sk-ilya: > Yes, to be sure nothing breaks it.

Yet it would make sense to label a subset of tests that will run in the agent test matrix. I don’t think we get additional signal from running tests that don’t touch the agent  for each agent OS - those can run once.

>  In any case they're all running simultaneously, so there is no time penalty.

There is, to some extent. The pipeline will take as long as the longest job takes, and the more tests we have, the higher the chance of one of them taking an unusually long time (or flaking) and delaying the whole PR.

gshilin-sdb: Made changes according to our agreement.

Makefile
```diff
@@ -425,8 +425,8 @@ clean-quadlets:
 
 # Use custom golangci-lint container with libvirt support
 LINT_IMAGE := flightctl-lint:latest
-LINT_CONTAINER := podman run --rm \
-	-v $(GOBASE):/app:Z \
+LINT_CONTAINER := podman run --rm --security-opt label=disable \
+	-v $(GOBASE):/app \
```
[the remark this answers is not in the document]
gshilin-sdb: done

.github/workflows/pr-e2e-testing.yaml
```diff
@@ -186,18 +186,31 @@ jobs:
...
-            label_filter: 'sanity && slow'
+          ## CS10 Agent metrics
+          - os_id: cs10-bootc
+            backend_type: helm
+            label_filter: 'sanity && agent && !slow'
+            ginkgo_total_nodes: 2
           - os_id: cs10-bootc
             backend_type: helm
+            label_filter: 'sanity && agent && slow'
             ginkgo_total_nodes: 1
```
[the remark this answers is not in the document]
gshilin-sdb: that's the original desire

Makefile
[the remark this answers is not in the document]
gshilin-sdb: done

PR #2452, opened 2026-02-02
NO-ISSUE: imagebuild api set to v1alpha1

asafbennatan:
## Summary by CodeRabbit

* **API Updates**
  * ImageBuild and ImageExport moved to v1alpha1 (new alpha version available); v1beta1 remains supported.
* **Documentation**
  * User docs and resource reference updated to show v1alpha1, with a note about alpha status and expanded resource listings.
* **Examples**
  * All example manifests updated to use flightctl.io/v1alpha1 for ImageBuild/ImageExport.
* **CLI & Tools**
  * CLI, OpenAPI spec, and tooling now target the v1alpha1 ImageBuilder API.


celdrake:
@asafbennatan I've noticed that without any UI changes, we are sending POST etc request with `apiVersion: v1beta1` (as well as having that version in the `Flightctl-Api-Version` header) and the API accepts it (and it responds by creating the entity with `v1alpha1`). 

Should the system throw an error?

Update: Seems like something's not quite right in the PR, since if the UI sends `v1alpha1`, then the imageBuilder API throws error 406 indicating that the supported version is `v1beta1`

PR #2457, opened 2026-02-03
EDM-3262: fixed cancel not working properly for pending states

asafbennatan:
moving pending state ib/ie directly to cancel state - relying on db optimistic locking with resource_version to avoid race conditions between api and worker

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable cancellation with retry-on-conflict and clearer cancel vs canceling transitions
  * Pending/no-status resources now cancel immediately; in-progress items transition to canceling and complete properly
  * Improved cleanup and signaling so cancellation events are emitted and removed predictably

* **Tests**
  * Updated tests to expect immediate cancellation for pending/no-status cases

### DIFF
--- a/internal/imagebuilder_api/service/imagebuild.go
+++ b/internal/imagebuilder_api/service/imagebuild.go
@@ -84,7 +84,17 @@ func (s *imageBuildService) Create(ctx context.Context, orgId uuid.UUID, imageBu
 	if err != nil {
 		return result, StoreErrorToApiStatus(err, true, string(domain.ResourceKindImageBuild), imageBuild.Metadata.Name)
 	}
-
+	// Clear any stale Redis keys from a previous resource with the same name
+	// This prevents old cancellation signals from affecting the new resource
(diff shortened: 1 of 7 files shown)


internal/imagebuilder_api/service/imagebuild.go
```diff
@@ -80,6 +80,18 @@ func (s *imageBuildService) Create(ctx context.Context, orgId uuid.UUID, imageBu
 		return nil, StatusBadRequest(errors.Join(errs...).Error())
 	}
 
+	// Clear any stale Redis keys from a previous resource with the same name
+	// This prevents old cancellation signals from affecting the new resource
+	if s.kvStore != nil && imageBuild.Metadata.Name != nil {
```
avishayt, line 89: What if I create while the resource still exists?  We'll delete the Redis keys and then return error

asafbennatan: fixed

PR #2459, opened 2026-02-03
EDM-3265: build cs9 microshift image only if the repo is accessible

sk-ilya:
## Summary by CodeRabbit

* **Tests**
  * Increased VM memory allocation for test environments from 512 MiB to 2048 MiB.
  * Enhanced test infrastructure with improved auto-detection logic for variant builds based on system accessibility.
  * Added detailed logging output for better visibility into test build configuration and variant selection.
  * Updated test configuration handling to remove hardcoded defaults.

### DIFF
--- a/test/scripts/agent-images/create_agent_images.sh
+++ b/test/scripts/agent-images/create_agent_images.sh
@@ -31,11 +31,19 @@ fi
 
 export JOBS="${PARALLEL_JOBS}"
 
-# Handle ACM detection - enable v7 variant and increase VM memory
-if is_acm_installed; then
-    export EXCLUDE_VARIANTS=""
-    sed -i 's|<memory unit="MiB">512</memory>|<memory unit="MiB">2048</memory>|' test/harness/e2e/vm/domain-template.xml
-    echo "ACM detected, enabling v7 variant"
+# Handle v7 variant based on OS and RHOCP access
+# Only auto-detect if user hasn't explicitly set EXCLUDE_VARIANTS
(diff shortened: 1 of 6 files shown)


test/scripts/agent-images/create_agent_images.sh
```diff
@@ -170,9 +178,9 @@ build_variants_and_qcow2() {
         mv "${QCOW_SRC}" "${QCOW_DST}"
         echo "Moved qcow2 to ${QCOW_DST}"
 
-        # Resize disk for ACM if installed
-        if is_acm_installed; then
-            echo "ACM detected, resizing qcow2 disk +5G"
+        # Resize disk if v7 is enabled (for MicroShift)
+        if [ -z "${EXCLUDE_VARIANTS:-}" ] || ! echo "${EXCLUDE_VARIANTS}" | grep -qw "v7"; then
+            echo "v7 enabled, resizing qcow2 disk +5G"
             sudo qemu-img resize "${QCOW_DST}" +5G
```
siserafin, line 184: Can we re-size by +10G? I noticed node pressure with just +5G when I also deploy many applications

sk-ilya: Many apps seem like an edge case. If this is for manual testing, maybe the resize command can be run manually as well? GH runners have very limited disk space, and in general it’s preferable for test artifacts to have a smaller footprint.

siserafin: ok, we can keep 5G, in case we can increase later if needed.

siserafin (approved):
lgtm

PR #2463, opened 2026-02-03
EDM-2932: Separate user for console sessions

keitwb:
The user that will be used for console sessions is right now hardcoded
to a 'flightctl-console' user by default. This default user is created
upon RPM installation (if it doesn't exist) and is added to the 'wheel'
supplimental group so that it gets passwordless sudo access.

The SELinux domain for the console shell is transitioned to the normal `unconfined_t` domain which is used by normal shell users so that things are more consistent with a normal shell session (e.g. via ssh).

## Summary by CodeRabbit

* **New Features**
  * Configurable console user for shell sessions (default "console"); interactive shells run as that user and use its home (/var/lib/flightctl).

* **Chores**
  * Added sudo as a runtime dependency and automated system-user provisioning via sysusers.
  * Updated SELinux policy to allow required shell transitions and permissions for console sessions.

* **Tests**
  * Updated tests for new user, home dir, prompt and file paths; added failure log dumping.

### DIFF
--- a/internal/agent/device/console/console_test.go
+++ b/internal/agent/device/console/console_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"os/user"
 	"strings"
 	"sync"
 	"testing"
@@ -76,6 +77,7 @@ func setupVars(t *testing.T) *vars {
 		controller: NewManager(
 			mockGrpcClient,
 			"mydevice",
+			lo.Must(user.Current()).Username,
 			executor,
 			mockWatcher,
 			logger),
@@ -207,9 +209,10 @@ func TestConsole(t *testing.T) {
 		sessionID := uuid.New().String()
 		consoleDef := deviceConsole(sessionID, sessionMetadata(t, "xterm", nil,
 			&v1beta1.DeviceCommand{
-				Command: "exit",
+				Command: "/bin/bash",
 				Args: []string{
-					"11",
+					"-c",
+					`exit 11`,
 				}},
 			false))
 
@@ -258,12 +261,10 @@ func TestConsole(t *testing.T) {
 		v := setupVars(t)
 		sessionID := uuid.New().String()
 		consoleDef := deviceConsole(sessionID, sessionMetadata(t, "xterm", nil, &v1beta1.DeviceCommand{
-			Command: "echo",
-			Args: []string{"stdout",
-				";",
-				"echo",
-				"stderr",
-				">&2",
+			Command: "/bin/bash",
+			Args: []string{
+				"-c",
+				`echo stdout; echo stderr >&2`,
 			},
 		}, false))
 
--- a/internal/agent/device/console/session.go
+++ b/internal/agent/device/console/session.go
@@ -2,13 +2,10 @@ package console
 
 import (
 	"context"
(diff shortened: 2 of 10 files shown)


packaging/rpm/sysusers.d/flightctl.conf
```diff
@@ -0,0 +1,6 @@
+u console - "Flightctl Agent Console" /var/lib/flightctl  /bin/bash
+g console -
+# Map the user to the group
+m console console
+# Give the console user sudo access
+m console wheel
```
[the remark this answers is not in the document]
keitwb: This is presumed on RHEL based systems.

internal/agent/device/console/console_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"os/user"
 	"strings"
```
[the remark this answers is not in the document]
keitwb: The tests aren't intended for such environments.

internal/agent/device/console/session.go
```diff
@@ -70,24 +50,29 @@ func (s *session) close() error {
...
-	ret := s.executor.CommandContext(ctx, "/bin/bash", args...)
+	ret := s.executor.CommandContext(ctx, args[0], args[1:]...)
 
 	if metadata.Term != nil {
 		ret.Env = append(ret.Env, "TERM="+*metadata.Term)
 	}
 
-	h := s.getHomedir()
-	if h != "" {
+	h, err := s.getHomedir()
```
kkyrazis, line 76: is it ok to ignore the error in this case?

keitwb: I don't see why it should cause the console session to fail. If it errors, it just doesn't set the `HOME` envvar and start up the shell in that dir. Instead, whatever dir the agent process is running in would be used.

internal/agent/config/config.go
```diff
@@ -146,6 +148,8 @@ type Config struct {
 	// ImagePruning holds all image/artifact pruning-related configuration
 	ImagePruning ImagePruning `json:"image-pruning,omitempty"`
 
+	ConsoleUser string `json:"console-user,omitempty"`
```
hexfusion: every config has a cost, do we need to expose this or can we ensure sane defaults + docs and let the user tell us why they need it? this feels like a possible exploit if an OS image overwrote this with a user unexpected by the agent? I think this could happen without the cluster operators knowledge

keitwb: I think the main concern I had in mind with this was if an os image already used a `console` user for something and it wasn't appropriate to overload it with the agent console functionality. I thought about calling it `flightctl-console` but that seems clunky, but would namespace it enough to avoid any chance of overlap. As discussed before, reusing the `flightctl` user is another option but giving that sudo access seems to work against the whole privilege separation goal.

I'm open to whatever you think is best long-term.

hexfusion: my main concern is only about exposing the flag. with regards to the name, making it unique seems to solve multiple problems.

keitwb: Ok, check it out now. The user is called `flightctl-console` which should definitely avoid name clashing.

hexfusion (approved):

PR #2464, opened 2026-02-04
EDM-3260: A bootc rollback flow should not start when microshift fails healthcheck

galel12:
## Problem

When a device updates to an OS image containing third-party applications with greenboot health checks (e.g., MicroShift), those applications can trigger OS rollbacks independently of flightctl. This causes two issues:

1. **EDM-3260**: Third-party applications like MicroShift can trigger greenboot rollbacks, violating the architectural principle that only the flightctl agent should control rollback decisions.

2. **EDM-3261**: After such a rollback, the device reports "UpToDate" status even though it rolled back to a previous OS version, because the agent wasn't aware of the externally-triggered rollback.

## Root Cause

Greenboot executes all health check scripts in `/usr/lib/greenboot/check/required.d/` and `/etc/greenboot/check/required.d/`. When applications like MicroShift are installed, they add their own health checks (e.g., `40_microshift_running_check.sh`) that can fail and trigger rollbacks without the flightctl agent's involvement.

## Solution

Add a systemd oneshot service (`flightctl-configure-greenboot.service`) that runs before `greenboot-healthcheck.service` on every boot. This service:

1. Scans for third-party health check scripts in the `required.d` directories
2. Preserves core greenboot scripts (`00_required_scripts_start.sh`, `01_repository_dns_check.sh`, `02_watchdog.sh`) and flightctl scripts (`*flightctl*`)
3. Disables all other third-party application scripts by updating `DISABLED_HEALTHCHECKS` in `/etc/greenboot/greenboot.conf`

This ensures only the flightctl agent's health check can trigger rollbacks, maintaining the agent's awareness of device state.

## RPM Changes

cc @gshilin-sdb @kkyrazis - This PR adds new files to the `flightctl-agent` RPM:

| File | Install Path |
|------|--------------|
| `packaging/greenboot/flightctl-configure-greenboot.sh` | `/usr/libexec/flightctl/configure-greenboot.sh` |
| `packaging/systemd/flightctl-configure-greenboot.service` | `/usr/lib/systemd/system/flightctl-configure-greenboot.service` |

The service is enabled automatically via `%post agent` in the RPM spec.

## Testing

Verified by updating a device from base image to v7 (with MicroShift):
- MicroShift's `40_microshift_running_check.sh` was automatically disabled
- Device booted successfully with `boot_success=1`
- No rollback occurred
- Pre-rollback script was not triggered
- Device correctly reports "UpToDate" on v7

## Summary by CodeRabbit

* **New Features**
  * Boot-time integration that disables third-party health checks so only flightctl checks can trigger rollbacks.

* **Chores**
  * Agent package now installs and enables a boot-time configuration script and systemd unit to manage the integration.

* **Documentation**
  * Added guidance on third-party health check management and warning not to edit the generated configuration.

### DIFF
--- a/docs/user/installing/configuring-device-greenboot.md
+++ b/docs/user/installing/configuring-device-greenboot.md
@@ -34,6 +34,13 @@ Exiting the health check with a non-zero status declares the boot as failed. The
 
 > If the system is not booted using `bootc`, the health check still runs, but no rollback is possible.
 
+### Third-Party Health Check Management
+
+Flight Control is designed to be the sole controller of OS rollback decisions. The `flightctl-configure-greenboot.service` runs before `greenboot-healthcheck.service` on every boot, automatically disabling third-party health checks (e.g., MicroShift) by setting `DISABLED_HEALTHCHECKS` in `/etc/greenboot/greenboot.conf`. Core greenboot scripts and Flight Control's own health checks are preserved.
+
+> [!WARNING]
+> Do not manually edit `DISABLED_HEALTHCHECKS` in `/etc/greenboot/greenboot.conf` it is replaced on every boot by `flightctl-configure-greenboot.service`.
+
 ## The `systemd` Journal Service Configuration
 
 The default `systemd` journal configuration stores data in volatile `/run/log/journal`, which does not persist across reboots. To monitor greenboot activities across boots, enable persistent storage:


hexfusion:
> Overall looks ok, but if it's not already, should we document that the agent disables other greenboot checks?

Good callout this should be inline with how we disable the bootc auto update timer.

docs/user/installing/configuring-device-greenboot.md
```diff
@@ -34,6 +34,10 @@ Exiting the health check with a non-zero status declares the boot as failed. The
 
 > If the system is not booted using `bootc`, the health check still runs, but no rollback is possible.
 
+### Third-Party Health Check Management
+
+Flight Control is designed to be the sole controller of OS rollback decisions. The `flightctl-configure-greenboot.service` runs before `greenboot-healthcheck.service` on every boot, automatically disabling third-party health checks (e.g., MicroShift) by adding them to `DISABLED_HEALTHCHECKS` in `/etc/greenboot/greenboot.conf`. Core greenboot scripts and Flight Control's own health checks are preserved.
+
```
galel12, line 43: @hexfusion @kkyrazis  resolved ✔️

hexfusion: @galel12 @kkyrazis this is a failing healthcheck can the agent observe this information during runtime and report status. clearly microshift is failing but just like any application this is runtime status?

PR #2467, opened 2026-02-04
EDM-3267: Improve application rollback and kube client configuration

kkyrazis:
Fixes a few bugs:
1. Kube config / binary resolution is now lazily loaded rather than immediately on agent startup. Depending on the order of operations, sometimes the microshift kubeconfig wouldn't exist yet.
2. Depending on ordering, applications could be installed and fail to uninstall on rollback. (app1 installs, app2 fails to install and triggers a rollback, on rollback app2 is attempted to remove but that also fails. App1 never experiences the remove command)

Adds some behavior so that applications that are added and then immediately removed (from rollback) no longer attempt to add themselves first. When ExecuteActions runs, only the last action for an application is run.

## Summary by CodeRabbit

* **New Features**
  * On-demand detection of kubectl/oc with a new option to supply a kubeconfig path and runtime kubeconfig resolution.

* **Bug Fixes**
  * Continue syncing providers on individual errors and report aggregated failures.
  * Helm removals safely skip when Kubernetes is unavailable.

* **Performance**
  * Deterministic ordering and deduplication of application actions to avoid redundant work.

* **Tests**
  * Expanded table-driven tests for kubeconfig/binary resolution, action reduction, provider sync, and helm flows.

### DIFF
--- a/internal/agent/client/kube.go
+++ b/internal/agent/client/kube.go
@@ -24,13 +24,21 @@ const (
 type KubernetesOption func(*kubernetesOptions)
 
 type kubernetesOptions struct {
-	binary string
+	binary         string
+	kubeconfigPath string
 }
 
 // WithBinary sets a specific kubectl/oc binary path instead of auto-discovering.
 func WithBinary(binary string) KubernetesOption {
 	return func(opts *kubernetesOptions) {
 		opts.binary = binary
+	}
+}
+
+// WithKubeconfigPath sets the kubeconfig path instead of auto-discovering.
+func WithKubeconfigPath(path string) KubernetesOption {
+	return func(opts *kubernetesOptions) {
+		opts.kubeconfigPath = path
 	}
 }
 
@@ -59,14 +67,20 @@ func WithKubeLabels(labels []string) KubeOption {
 
 // Kube provides a client for executing kubectl/oc CLI commands.
 type Kube struct {
-	exec       executer.Executer
-	log        *log.PrefixLogger
-	binary     string
-	readWriter fileio.ReadWriter
+	exec             executer.Executer
+	log              *log.PrefixLogger
+	readWriter       fileio.ReadWriter
+	commandAvailable func(string) bool
+
+	binary             string
+	kubeconfigPath     string
+	kubeconfigResolved bool
 }
 
-// NewKube creates a new Kube client. It auto-discovers kubectl or oc if no binary is specified.
-// Use IsAvailable to check if a kubernetes CLI binary was found.
+// NewKube creates a new Kube client with lazy initialization.
+// The kubernetes binary and kubeconfig are discovered on first use.
+// Use WithBinary option to set a specific binary path instead of auto-discovering.
+// Use WithKubeconfigPath option to set a specific kubeconfig path.
 func NewKube(
 	log *log.PrefixLogger,
 	exec executer.Executer,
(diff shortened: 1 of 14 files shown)


internal/agent/client/kube.go
```diff
@@ -78,54 +92,123 @@ func NewKube(
...
+			}
+			checkedPaths = append(checkedPaths, fmt.Sprintf("%s (KUBECONFIG env)", path))
+			exists, err := k.readWriter.PathExists(path)
+			if err != nil {
+				return "", fmt.Errorf("check KUBECONFIG path: %w (checked: %s)", err, strings.Join(checkedPaths, ", "))
+			}
+			if exists {
+				return "", nil
+			}
+		}
```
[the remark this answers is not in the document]
kkyrazis: In the case that the KUBECONFIG environment variable is set, all invocations that rely on `--kubeconfig` will still work as KUBECONFIG will be forwarded to those calls.

internal/agent/client/kube_test.go
```diff
@@ -12,82 +12,150 @@ import (
...
-	require.True(k8s.IsAvailable())
+			k8s := NewKube(logger, mockExec, mockReadWriter,
+				WithBinary(tc.binary),
+				WithKubeconfigPath("/tmp/kubeconfig"),
+			)
+			require.NotNil(k8s)
+			require.Equal(tc.binary, k8s.Binary())
+			require.True(k8s.IsAvailable())
+		})
+	}
```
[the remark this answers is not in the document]
kkyrazis: specifying the binary/config path should resolve that problem

internal/agent/device/applications/controller.go
```diff
@@ -83,26 +83,34 @@ func syncProviders(
...
-			return err
+	for _, p := range diff.Changed {
+		log.Debugf("Updating application: %s", p.Name())
+		if err := manager.Update(ctx, p); err != nil {
+			log.Warnf("Failed to update application %s: %v", p.Name(), err)
+			errs = append(errs, fmt.Errorf("updating: %w: %w", errors.WithElement(p.Name()), err))
 		}
 	}
 
+	if len(errs) > 0 {
```
keitwb: `errors.Join(nil) == nil` and `errors.Join([]error{}) == nil` so this conditional is not necessary.

internal/agent/device/applications/controller.go
```diff
@@ -83,26 +83,34 @@ func syncProviders(
 		return err
 	}
 
-	for _, provider := range diff.Removed {
-		log.Debugf("Removing application: %s", provider.Name())
-		if err := manager.Remove(ctx, provider); err != nil {
-			return err
+	var errs []error
```
keitwb, line 86: I think before when I was messing with this I asked about collecting errors vs immediate return and you said we wanted to return quickly since it was all going to be rolled back anyway. Is that still the case?

kkyrazis: There was an instance where two applications were applied, the third one failed and triggered a rollback. The rollback failed for the same reason, but the ordering of the applications was different. So only the failed application attempted rollback, reported a failure, and then the rollback aborted. This left the "good" applications running when they should have rolled back and been removed. I fixed the underlying bug that caused this issue, but without deterministic ordering or storing exactly what has been applied and only undoing those changes (i.e. on upgrade we do adds, updates, removes; on rollback we do removes, updates, adds) it seemed like it would be better to attempt everything on upgrade / rollback to ensure cleanliness.

PR #2468, opened 2026-02-04
EDM-3288: Fix config provider cleanup race condition with centralized PullConfig resolver

kkyrazis:
Moves pull dependency management (the client options) into a centralized place with a more well defined lifecycle. This allows for better sharing of any temporary files that may need to be generated for pull secrets used amongst the providers / os puller.

## Summary by CodeRabbit

* **Refactor**
  * Introduces a shared pull-config resolver injected across components, standardizing pull-secret/options handling and resolving options at pull time; replaces per-target cleanup with lazy option evaluation and a centralized prepare/schedule flow.

* **Tests**
  * Adds comprehensive tests for the new resolver (resolution, caching, inline vs on-disk fallbacks, cleanup, cache invalidation) and removes legacy inline-config test scaffolding and related mocks.


internal/agent/client/client.go
```diff
@@ -100,147 +96,6 @@ func IsComposeAvailable() bool {
 	return false
 }
 
-// PullConfig holds the path to a configuration file and a cleanup function for temporary files.
-type PullConfig struct {
```
kkyrazis, line 104: moved all of this into dependency

internal/agent/device/dependency/dependency.go
```diff
@@ -489,119 +499,94 @@ func (m *prefetchManager) setError(target imageRef, err error) {
...
-		opts = m.buildClientOptions(target, opts)
-
-		var cleanupFns []func()
-		if target.Configs != nil {
-			cleanupFns = append(cleanupFns, target.Configs.Cleanup)
-		}
-
 		ref := imageRef{image: target.Reference, owner: user}
-		if err := m.schedule(ctx, ref, target.Type, cleanupFns, opts...); err != nil {
+		if err := m.schedule(ctx, ref, target.Type, target.ClientOptsFn); err != nil {
```
keitwb, line 503: Is there really any advantage to passing a no-args function into this instead of just calling it here and making the args more explicit? Is it an expensive funciton?

kkyrazis: It's not nothing, but I wouldn't say it's the end of the world. Some disk i/o / binary comparison. 

I think the biggest advantage of this is that in the steady state nothing will be done as `podman image exists` and other similar checks will result in early returns and the client options (and temp files) won't get built.

PR #2470, opened 2026-02-05
EDM-3261: Unexplained rollback service should show out of date

galel12:
## Problem

After a greenboot-triggered OS rollback (e.g., caused by MicroShift health check failure), the device incorrectly reports `UpToDate` status even though it's running a previous OS version. The service has no visibility that a rollback occurred.

## Root Cause

The agent calls `Upgrade()` to commit the spec **before** greenboot finishes validating the boot:

1. Device boots into new OS (v2), agent calls `Upgrade()` → commits spec and clears `rollback.json`
2. Greenboot runs third-party health checks → one fails → triggers rollback to v1
3. On reboot, spec files still say v2, but device is running v1
4. Agent sees `current == desired` (both v2) → reports `UpToDate` despite running wrong OS

## Solution

Three-layer fix:

### 1. Agent: Wait for greenboot before committing spec
- Check `systemctl is-active boot-complete.target` before calling `Upgrade()`
- This target is reached after all greenboot health checks in `required.d/` pass
- If not yet active, skip upgrading and wait for next sync cycle

### 2. Service: Detect OS image mismatch
- Add OS image comparison in `updateServerSideDeviceUpdatedStatus`
- If `status.os.image != spec.os.image` while rendered versions match, override status to `OutOfDate`
- The service owns the UpToDate/OutOfDate determination

### 3. Detection: Log critical error if version goes backwards
- Safety net in `write()` function to detect if rendered version ever decreases during upgrade
- If primary protections fail, this provides forensic evidence in logs/audit

## Summary by CodeRabbit

* **New Features**
  * Upgrades now gate on successful boot completion to avoid partial/failed upgrades.
  * Upgrades detect version regressions during an upgrade and emit critical alerts on downgrades.
  * Service status now flags and reports OS image mismatches between desired and active configurations.

* **Tests**
  * Added unit tests covering boot-gating behavior and OS image mismatch detection.

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -221,8 +221,8 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 	// also ensures previous failed status is not overwritten.
 	if a.specManager.IsUpgrading() {
 		// Wait for greenboot to mark the boot as successful before committing the spec.
-		// This prevents committing a spec that could be rolled back by greenboot.
-		if !a.isBootSuccessful(ctx) {
+		// This only applies to OS updates — greenboot validates the new OS image, not config changes.
+		if a.specManager.IsOSUpdate() && !a.isBootSuccessful(ctx) {
 			a.log.Debug("Waiting for greenboot to mark boot as successful before upgrading spec")
 			return
 		}


internal/agent/device/device.go
```diff
@@ -591,3 +616,22 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
 	}
 	return nil
 }
+
+// isBootSuccessful checks if greenboot has marked the current boot as successful.
+// Returns true if boot_success=1 is set in the GRUB environment, or if greenboot
+// is not installed (grub2-editenv not available).
+func (a *Agent) isBootSuccessful() bool {
+	stdout, stderr, exitCode := a.executer.Execute("grub2-editenv", "-", "list")
```
hexfusion: question: my assumption is that greenboot would talk to bootc directly here can we confirm.

hexfusion (changes requested):
@galel12 I believe outofdate status is owned by the service. can the service understand that the current version and desired version are not aligned?

internal/agent/device/device.go
galel12: Can you clarify what you mean?
Currently greenboot sets `boot_success=1` in GRUB env when all health checks pass.

hexfusion: Gotcha I though it was trying to understand boot order. Touching grub still feels a little heavy can we use systemd? The greenboot README says.

```
If all scripts in required.d folder succeeded:
boot-complete.target is reached.
```

```
systemctl is-active boot-complete.target
```

can we explore using systemd as an alternative

galel12: :heavy_check_mark: resolved in 73db34aa2fb11b8ed8476e8c2812a8aa833c1fb7
 yeah you're right about the service owning the outofdate status so I moved the OS image mismatch detection there - `updateServerSideDeviceUpdatedStatus` now overrides to `OutOfDate`:
```go
if status.os.image != spec.os.image 
```
Also switched to `systemctl is-active boot-complete.target` instead of `grub2-editenv`

galel12:
@coderabbitai review

internal/agent/device/spec/manager.go
```diff
@@ -550,6 +550,18 @@ func (s *manager) write(ctx context.Context, specType Type, device *v1beta1.Devi
...
+	newVersion := device.Version()
+
+	// CRITICAL: Detect if rendered version is going backwards during upgrade.
+	// This should never happen and indicates a serious bug (e.g., unexpected rollback
+	// after spec was committed). The audit log will capture this transition.
+	if specType == Current && reason == audit.ReasonUpgrade {
+		oldV, errOld := stringToInt64(oldVersion)
+		newV, errNew := stringToInt64(newVersion)
+		if errOld == nil && errNew == nil && newV < oldV {
+			s.log.Errorf("CRITICAL: Rendered version going backwards during upgrade! %s -> %s - this should never happen", oldVersion, newVersion)
```
hexfusion, line 562: I believe this is possible in disaster recovery but I generally agree the log makes sense

internal/agent/device/device.go
```diff
@@ -606,3 +617,21 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
 	}
 	return nil
 }
+
+// isBootSuccessful checks if greenboot has marked the current boot as successful
+// by checking if boot-complete.target has been reached. This target is activated
+// by greenboot after all required health checks in required.d/ pass.
+// Returns true if boot-complete.target is active, or if greenboot is not installed.
+func (a *Agent) isBootSuccessful(ctx context.Context) bool {
+	_, stderr, exitCode := a.executer.ExecuteWithContext(ctx, "systemctl", "is-active", "boot-complete.target")
```
hexfusion: we have  systemd client can we extend/call that please vs inject raw executor.
```
func (s *Systemd) IsActive(ctx context.Context, name string) (bool, error) {
```
ref. https://github.com/flightctl/flightctl/blob/main/internal/agent/client/systemd.go

galel12: :heavy_check_mark: resolved in f703b68cb7fa10a6af3a94df133833fd9b3042a1

internal/agent/device/device.go
```diff
@@ -606,3 +616,17 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
...
+
+// isBootSuccessful checks if greenboot has marked the current boot as successful
+// by checking if boot-complete.target has been reached. This target is activated
+// by greenboot after all required health checks in required.d/ pass.
+// Returns true if boot-complete.target is active, or if greenboot is not installed.
+func (a *Agent) isBootSuccessful(ctx context.Context) bool {
+	active, err := a.systemdClient.IsActive(ctx, "boot-complete.target")
+	if err != nil {
+		// Unit not found or other error means greenboot is likely not installed
+		a.log.Debugf("Failed to check boot-complete.target: %v", err)
```
hexfusion: I think this should be a warn or error because its really not expected today. in the future if we support non systemd then it could change

galel12: :heavy_check_mark: resolved in https://github.com/flightctl/flightctl/compare/f703b68cb7fa10a6af3a94df133833fd9b3042a1..db4ada1f66377cc6d5f79c77a3cbfdf94603b692

internal/agent/device/device.go
```diff
@@ -606,3 +616,17 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
...
 	return nil
 }
+
+// isBootSuccessful checks if greenboot has marked the current boot as successful
+// by checking if boot-complete.target has been reached. This target is activated
+// by greenboot after all required health checks in required.d/ pass.
+// Returns true if boot-complete.target is active, or if greenboot is not installed.
+func (a *Agent) isBootSuccessful(ctx context.Context) bool {
+	active, err := a.systemdClient.IsActive(ctx, "boot-complete.target")
+	if err != nil {
```
hexfusion, line 626: in the future it would be nice to be more perscriptive, IE errors.Is(err, errors.UnitNotFound) ... but not blocking now.

galel12: Noted, thanks! I'll apply that from now on

internal/agent/device/device.go
```diff
@@ -217,6 +220,13 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 	// skip status update if the device is in a steady state and not upgrading.
 	// also ensures previous failed status is not overwritten.
 	if a.specManager.IsUpgrading() {
+		// Wait for greenboot to mark the boot as successful before committing the spec.
+		// This only applies to OS updates — greenboot validates the new OS image, not config changes.
+		if a.specManager.IsOSUpdate() && !a.isBootSuccessful(ctx) {
```
galel12, line 225: @hexfusion the merge queue e2e was failing because the `boot-complete.target` check was gated on IsUpgrading(), which is true for any spec change (config, apps, etc.), not just OS updates. So config-only updates would block forever waiting for greenboot. 
Fixed by adding IsOSUpdate() guard - the greenboot wait now only applies during actual OS image changes.

PR #2471, opened 2026-02-05
EDM-2636: Skopeo rootless credential handling

keitwb:
Skopeo tries to access the file /run/containers/1002/auth.json to check for default registry credentials, which causes an error when running against a non-root skopeo process due to env setup and systemd session quirks.

Also made the executor env handling more secure by not automatically inheriting the env for non-root processes, which could cause things to break or be a security risk.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credentials handling for container manifest inspection: explicitly disables use of default credentials when no credentials are provided, ensuring consistent unauthenticated behavior.
  * Fixed environment handling for non-root execution: prevents unintended inheritance of parent process environment variables while still honoring an explicitly set HOME, improving isolation and predictability.

### DIFF
--- a/internal/agent/client/skopeo.go
+++ b/internal/agent/client/skopeo.go
@@ -101,6 +101,11 @@ func (s *Skopeo) InspectManifest(ctx context.Context, image string, opts ...Clie
 			return nil, fmt.Errorf("pull secret path %s does not exist", pullSecretPath)
 		}
 		args = append(args, "--authfile", pullSecretPath)
+	} else {
+		// Skopeo does not behave well when looking up default credentials as a non-root user without a proper systemd session
+		// running, so disable default credentials when none were explicitly provided. This
+		// means any credentials required have to be specified in the options.
+		args = append(args, "--no-creds")
 	}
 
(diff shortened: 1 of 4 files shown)


internal/agent/client/skopeo.go
```diff
@@ -101,6 +101,11 @@ func (s *Skopeo) InspectManifest(ctx context.Context, image string, opts ...Clie
 			return nil, fmt.Errorf("pull secret path %s does not exist", pullSecretPath)
 		}
 		args = append(args, "--authfile", pullSecretPath)
+	} else {
+		// Skopeo does not behave well when looking up default credentials as a non-root user without a proper systemd session
+		// running, so disable default credentials when none were explicitly provided. This
+		// means any credentials required have to be specified in the options.
+		args = append(args, "--no-creds")
```
kkyrazis, line 108: I'm assuming there are still issues with rootless skopeo attempting to use the root level credentials?

keitwb: If they are owned and only readable by root yes, but you can configure k8s secrets to be owned by specific users now as well as other files so it should be possible to use credentials with rootless.

PR #2474, opened 2026-02-06
NO-ISSUE: Fix podman monitor test race conditions and StatusCreate constant

kkyrazis:
## Summary by CodeRabbit

* **Refactor**
  * Updated application status handling and improved internal code organization.
  * Removed unused internal methods.
  * Enhanced test reliability with improved event streaming synchronization and error handling.

### DIFF
--- a/internal/agent/device/applications/applications.go
+++ b/internal/agent/device/applications/applications.go
@@ -22,7 +22,7 @@ const (
 type StatusType string
 
 const (
-	StatusCreated StatusType = "created"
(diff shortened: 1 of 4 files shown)


internal/agent/device/applications/applications.go
```diff
@@ -22,7 +22,7 @@ const (
 type StatusType string
 
 const (
-	StatusCreated StatusType = "created"
```
kkyrazis, line 25: Note: The event is `create` not `created`. The tests were right in this case.

PR #2475, opened 2026-02-08
EDM-3304: canceling dependent image exports automatically

asafbennatan:
## Release Notes

**Bug Fixes**
- Enhanced the image build cancellation workflow to automatically cancel all associated image export operations when an image build is canceled. This ensures thorough cleanup of related resources and prevents orphaned export tasks from remaining active in the system, improving overall resource management and system reliability.


internal/imagebuilder_api/service/imagebuild.go
sk-ilya, line 191: Nit: duplicated, and perhaps it will be better to keep this selector string as a constant or encapsulate this logic a new  `imageExportService`'s method like `ListForImageBuild(buildRef string)`

PR #2481, opened 2026-02-08
EDM-3291: reject requests with apiVersion mismatch

sk-ilya:
## Summary by CodeRabbit

* **New Features**
  * Standardized Status response object for error and non-object responses across the API.
  * Canonical ApiVersion identifiers added (including qualified group-prefixed values) for apiVersion fields.

* **Improvements**
  * More consistent error payloads with structured fields: code, message, reason, status, kind, and apiVersion.
  * Unified apiVersion usage across resources for clearer versioning and compatibility.


internal/transport/v1beta1/response.go
```diff
@@ -0,0 +1,22 @@
...
+// For 2xx responses, body is encoded; for non-2xx, the converted status is encoded.
+func (h *TransportHandler) SetResponse(w http.ResponseWriter, body any, status domain.Status) {
+	apiStatus := h.converter.Common().StatusFromDomain(status)
+	transport.WriteJSONResponse(w, body, apiStatus, int(status.Code))
+}
+
+// SetParseFailureResponse writes a 500 response for JSON decode failures.
+func (h *TransportHandler) SetParseFailureResponse(w http.ResponseWriter, err error) {
+	h.SetResponse(w, nil, domain.StatusInternalServerError(fmt.Sprintf("can't decode JSON body: %v", err)))
+}
```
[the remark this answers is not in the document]
sk-ilya: It is called on requests that are already validated by openapi. If we can't decode a valid request it's 500

internal/api_server/versioning/apiversion_validator.go
```diff
@@ -0,0 +1,83 @@
...
+	switch req.Method {
+	case http.MethodGet, http.MethodDelete, http.MethodHead, http.MethodOptions:
+		return true
+	}
+
+	if req.Body == nil || req.Body == http.NoBody {
+		return true
+	}
+
+	body, err := io.ReadAll(req.Body)
```
asafbennatan: Should we use LimitReader to avoid reading requests with huge body ?

internal/api_server/versioning/apiversion_validator.go
```diff
@@ -0,0 +1,83 @@
...
+// Transport handlers implement this interface via their SetResponse method.
+type StatusResponder interface {
+	SetResponse(http.ResponseWriter, any, domain.Status)
+}
+
+// ValidateBodyAPIVersion returns middleware that checks the body's apiVersion
+// field matches the expected version. It accepts both bare ("v1beta1") and
+// prefixed ("flightctl.io/v1beta1") formats by extracting the version part
+// after the last slash. Unrecognized versions are rejected with 400.
+func ValidateBodyAPIVersion(expected Version, responder StatusResponder) Middleware {
```
asafbennatan: instead of doing that why cant we use the enum approach you have suggested ? 
i.e - make imagebuild.apiVersion = enum of [v1alpha1,flightctl.io/v1alpha1] and device.apiVersion enum of [v1beta1,flightctl.io/v1beta1]

internal/api_server/versioning/apiversion_validator.go
sk-ilya: Technically the field can be at any place in the JSON, and the request body size is  limited globally anyway. 
To make this safer I think I can re-order the middleware:

```go
		Middlewares: []versioning.Middleware{
            //  validate first
			v1beta1OapiMiddleware, 
            //  then check the field
			versioning.ValidateBodyAPIVersion(versioning.V1Beta1, handlerV1Beta1),
		},

```

PR #2483, opened 2026-02-09
EDM-3316: Fixed pam issuer startup/ user management issue caused by #2448

asafbennatan:
fixes issue introduced in #2448 
1. missing env for running flightctl-standalone
2.  individual mounts do not allow managing as we did before according to auth-pam.md , this is because normal exec's used to manage group/passwd/shadow/gshadow replace the file which is not allowed when you mount individual files

the first one is fixed by simply adding to env , the second one is fixed by :
1. mounting entire userdb folder to /userdb
2. on startup copy the mentioned files from userdb to container's /etc
3. watch file changes on container's etc dir -> copy relevant files back to userdb (so changes are retained across runs)

## Summary by CodeRabbit

* **New Features**
  * Configurable user DB and system directories via environment variables with sensible defaults
  * Bidirectional user database sync with initial copy-in and continuous monitoring and events

* **Chores**
  * Simplified container deployment: single named volume for user DB, preserved config bind, added HOME env; removed legacy migration step

* **Tests**
  * Added unit tests covering sync startup, copy-in/copy-back flows, error cases, and watcher behavior

* **Documentation**
  * Updated deployment docs to reflect named-volume approach and removed migration instructions


deploy/podman/flightctl-pam-issuer/flightctl-pam-issuer.container
```diff
@@ -11,17 +11,13 @@ Pull=missing
...
 Volume=/etc/flightctl/pki/flightctl-pam-issuer/token-signer.key:/root/.flightctl/certs/client-signer.key:ro,z
-# User database files - persist local users created inside the container.
-# Other /etc files (PAM configs, nsswitch.conf, etc.) come from the container image
-# and are updated when the image is updated.
-Volume=/etc/flightctl/flightctl-pam-issuer/userdb/passwd:/etc/passwd:z
-Volume=/etc/flightctl/flightctl-pam-issuer/userdb/shadow:/etc/shadow:z
-Volume=/etc/flightctl/flightctl-pam-issuer/userdb/group:/etc/group:z
-Volume=/etc/flightctl/flightctl-pam-issuer/userdb/gshadow:/etc/gshadow:z
+# User database dir: entrypoint copies to /etc on start; wrappers copy back after groupadd/useradd/etc.
+Volume=/etc/flightctl/flightctl-pam-issuer/userdb:/userdb:z
```
sk-ilya: Is mounting the host's DB considered a valid use case? If yes, then aren't we loosing this ability?

asafbennatan: its possible - we do not want to mount sensitive files as default - this was considered way back when flightctl-pam-issuer was designed

sk-ilya: In that case, I guess all of `/etc` will be mounted?

There are a few concerns:
- "too much of the host" will be exposed to the container, which is a security and isolation risk
- race conditions if both DBs are modified at the same time

asafbennatan: we have explicitly tried to avoid mounting /etc , host users should be handled by configuring SSSD pipe correctly

internal/pam_issuer_server/userdb_sync.go
```diff
@@ -0,0 +1,158 @@
...
+		src := filepath.Join(userdbDir, f)
+		dst := filepath.Join(etcDir, f)
+		if err := copyFile(src, dst); err != nil {
+			continue // e.g. userdb/passwd missing: skip, no panic
+		}
+		log.Debugf("userdb sync: copied %s -> %s", src, dst)
+	}
+}
+
+func copyBack(log logrus.FieldLogger, etcDir, userdbDir string) {
```
sk-ilya:
```go
func copyIn(log logrus.FieldLogger, userdbDir, etcDir string) 
func copyBack(log logrus.FieldLogger, userdbDir, etcDir string) 
```

Can it just be `copy(log, src, dst)`?

asafbennatan: done , i am now actually copying only the changed file so this is used only for initial copy

internal/pam_issuer_server/userdb_sync.go
```diff
@@ -0,0 +1,155 @@
...
+		return
+	}
+	defer watcher.Close()
+	if err := watcher.Add(etcDir); err != nil {
+		log.Warnf("userdb watch %s: %v (sync disabled)", etcDir, err)
+		sendSyncEvent(events, SyncEvent{Kind: SyncError, Err: err})
+		return
+	}
+	sendSyncEvent(events, SyncEvent{Kind: SyncWatcherStarted})
+	log.Infof("userdb sync: watching %s, persisting to %s", etcDir, userdbDir)
```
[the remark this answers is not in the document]
asafbennatan: its redundant

cmd/flightctl-pam-issuer/main.go
```diff
@@ -66,6 +71,20 @@ func main() {
 	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT)
 	defer cancel()
 
+	userdbDir := os.Getenv("USERDB_DIR")
+	if userdbDir == "" {
+		userdbDir = defaultUserDBDir
+	}
+	etcDir := os.Getenv("ETC_DIR")
```
hexfusion, line 78: question: is this more for testing or what would be the general usecase for redefining etc

internal/pam_issuer_server/userdb_sync.go
```diff
@@ -0,0 +1,155 @@
...
+// RunUserDBSync copies userdb from userdbDir into etcDir on start, then watches etcDir
+// and copies the four userdb files back to userdbDir whenever they change (so changes
+// from groupadd/useradd/usermod/chpasswd persist). Run until ctx is done.
+// Initial validation (empty dirs, userdbDir missing or not a directory) is done
+// synchronously and returns an error; runtime failures (e.g. watcher) are sent as
+// SyncError on the channel. On success returns (events, nil); the sync runs in a
+// goroutine and the channel is closed when it stops.
+func RunUserDBSync(ctx context.Context, log logrus.FieldLogger, userdbDir, etcDir string) (<-chan SyncEvent, error) {
+	if userdbDir == "" || etcDir == "" {
+		return nil, nil
```
hexfusion: we have a logger this seems unexpected should we at least log as an error? then also a non nil error but a channel that is nil sounds like a foot gun. ensuring that pattern never happens makes the code easy to reason about. "If func or method returns err == nil then the return is usable and safe."

cmd/flightctl-pam-issuer/main.go
asafbennatan: could be used for testing - it doesnt make sense for production i think ? although some linux user management commands accept a home dir parameter which implies the home dir could be somewhere else if we really wanted to

internal/pam_issuer_server/userdb_sync.go
asafbennatan: fixed

PR #2485, opened 2026-02-09
EDM-3308: add upstream target cs10 to container build

gshilin-sdb:


.github/actions/deploy-backend-with-helm/action.yaml
```diff
@@ -40,6 +40,10 @@ inputs:
     description: 'Wait for deployment to complete (adds --wait --debug to helm install)'
     required: false
     default: 'true'
+  flavor:
```
hexfusion, line 43: just for my own understanding what is the difference between `flavor` and `os_id` that I see being used by the agent.

gshilin-sdb: 1. Scope:
    - flavor = Backend services (control plane)
    - os_id = Agent images (edge devices)
  2. Specificity:
    - flavor = Just the EL version (el9)
    - os_id = EL version + boot method (el9-bootc)
  3. Why the difference?:
    - Backend containers just need to know the base OS for compatibility
    - Agent images need to know the exact boot mechanism since they run on physical edge devices where the boot method matters

hack/publish_containers.sh
```diff
@@ -1,14 +1,136 @@
...
+
+            echo "Publishing ${local_image}..."
+
+            # Tag and push to registry with EL naming only
+            podman tag "${local_image}" "quay.io/flightctl/flightctl-${service}-${EL_FLAVOR}:latest"
+            podman tag "${local_image}" "quay.io/flightctl/flightctl-${service}-${EL_FLAVOR}:${GIT_REF}"
+
+            # Push EL naming only
+            podman push "quay.io/flightctl/flightctl-${service}-${EL_FLAVOR}:latest"
+            podman push "quay.io/flightctl/flightctl-${service}-${EL_FLAVOR}:${GIT_REF}"
```
asafbennatan: i think it makes more sense to have the version in the tag rather than the image ref

images/Containerfile.imagebuilder-api
```diff
@@ -0,0 +1,69 @@
...
+# Build EL9: podman build --build-arg BUILD_IMAGE=registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 --build-arg RUNTIME_IMAGE=quay.io/flightctl/flightctl-base:el9-9.7-1762965531 --build-arg EL_VERSION=9 -f images/Containerfile.imagebuilder-api -t flightctl-imagebuilder-api-el9
+# Build EL10: podman build --build-arg BUILD_IMAGE=registry.access.redhat.com/ubi10/go-toolset:10.1-1770279878 --build-arg RUNTIME_IMAGE=quay.io/flightctl/flightctl-base:el10-10.1-1769518576 --build-arg EL_VERSION=10 -f images/Containerfile.imagebuilder-api -t flightctl-imagebuilder-api-el10
+
+
+ARG BUILD_IMAGE
+ARG RUNTIME_IMAGE
+ARG EL_VERSION
+ARG RHEM=false
+
+ARG RHEM=false
```
asafbennatan: duplicate arg?  , see the same in other container files

hack/publish_containers.sh
```diff
@@ -1,14 +1,136 @@
...
+        echo "Publishing FlightCtl containers for ${EL_FLAVOR} to registry..."
+
+        for service in $CONTAINER_SERVICES; do
+            local_image="flightctl-${service}-${EL_FLAVOR}:latest"
+
+            echo "Publishing ${local_image}..."
+
+            # Tag and push to registry with EL naming only
+            podman tag "${local_image}" "quay.io/flightctl/flightctl-${service}-${EL_FLAVOR}:latest"
+            podman tag "${local_image}" "quay.io/flightctl/flightctl-${service}-${EL_FLAVOR}:${GIT_REF}"
```
asafbennatan: el flavor should be in tag

images/Containerfile.userinfo-proxy
```diff
@@ -39,17 +51,19 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
     make build-userinfo-proxy
 
-FROM quay.io/flightctl/flightctl-base:9.7-1762965531
+FROM ${RUNTIME_IMAGE}
 WORKDIR /app
 LABEL \
   com.redhat.component="flightctl-userinfo-proxy-container" \
-  description="Flight Control UserInfo proxy for Grafana authentication" \
-  io.k8s.description="Flight Control UserInfo proxy for Grafana authentication" \
+  description="Flight Control userinfo proxy service" \
```
asafbennatan: should we change description

images/Containerfile.imagebuilder-worker
```diff
@@ -41,27 +51,19 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
-COPY --from=build /runtime-root/ /
-
+FROM ${RUNTIME_IMAGE}
 WORKDIR /app
-
 LABEL \
   com.redhat.component="flightctl-imagebuilder-worker-container" \
-  description="Flight Control ImageBuilder Worker for processing image build jobs" \
-  io.k8s.description="Flight Control ImageBuilder Worker for processing image build jobs" \
+  description="Flight Control image builder worker service" \
```
asafbennatan: should we change description ?

test/test.mk
```diff
@@ -158,18 +160,22 @@ E2E_AGENT_IMAGES_SENTINEL := $(ROOT_DIR)/bin/.e2e-agent-images-$(AGENT_OS_ID)
...
 
 # Run e2e tests with optional parallel execution
+# Set FLAVOR to control backend container flavor (el9 or el10, defaults to el9)
 # Set GINKGO_PROCS to control number of parallel processes (defaults to number of CPU cores)
 # Set GINKGO_OUTPUT_INTERCEPTOR_MODE to control parallel output (defaults to "dup" for full output)
-# Example: make run-e2e-test GO_E2E_DIRS=test/e2e/agent GINKGO_PROCS=4
-# Example: make run-e2e-test GO_E2E_DIRS=test/e2e/agent GINKGO_OUTPUT_INTERCEPTOR_MODE=swap
+# Example: make run-e2e-test FLAVOR=el10 GO_E2E_DIRS=test/e2e/agent GINKGO_PROCS=4
+# Example: make run-e2e-test FLAVOR=el9 GO_E2E_DIRS=test/e2e/agent GINKGO_OUTPUT_INTERCEPTOR_MODE=swap
+run-e2e-test: FLAVOR ?= el9
```
asafbennatan: nit: why do we need to redefine flavor here ? cant we use the value we have in top-level make ?

Makefile
```diff
@@ -131,7 +140,9 @@ help:
 
 .PHONY: publish
 publish: build-containers
```
asafbennatan, line 142: i think we should decide what to publish based on ARG and then on github we can use a parallel matrix so these can run together

Makefile
```diff
@@ -211,93 +222,19 @@ build-devicesimulator: bin
...
-flightctl-imagebuilder-worker-container: Containerfile.imagebuilder-worker go.mod go.sum $(GO_FILES)
-	podman build $(call CACHE_FLAGS_FOR_IMAGE,flightctl-imagebuilder-worker) \
-		--build-arg SOURCE_GIT_TAG=${SOURCE_GIT_TAG} \
-		--build-arg SOURCE_GIT_TREE_STATE=${SOURCE_GIT_TREE_STATE} \
-		--build-arg SOURCE_GIT_COMMIT=${SOURCE_GIT_COMMIT} \
-		-f Containerfile.imagebuilder-worker -t flightctl-imagebuilder-worker:latest -t quay.io/flightctl/flightctl-imagebuilder-worker:$(SOURCE_GIT_TAG)
-
-.PHONY: flightctl-api-container flightctl-pam-issuer-container flightctl-db-setup-container flightctl-worker-container flightctl-periodic-container flightctl-alert-exporter-container flightctl-alertmanager-proxy-container flightctl-multiarch-cli-container flightctl-userinfo-proxy-container flightctl-telemetry-gateway-container flightctl-imagebuilder-api-container flightctl-imagebuilder-worker-container
+# Build all containers for both CS9 and CS10
+build-containers: go.mod go.sum $(GO_FILES)
```
asafbennatan, line 227: i think we should decide what to publish based on ARG and then on github we can use a parallel matrix so these can run together

hack/publish_containers.sh
gshilin-sdb: done

images/Containerfile.imagebuilder-api
gshilin-sdb: removed

hack/publish_containers.sh
gshilin-sdb: done

images/Containerfile.userinfo-proxy
gshilin-sdb: reverted

images/Containerfile.imagebuilder-worker
gshilin-sdb: reverted

test/test.mk
gshilin-sdb: fixed

.github/actions/deploy-backend-with-helm/action.yaml
sk-ilya: In the case of agent images, flavors are a set of image parameters, and `os_id` is one of them. See: https://github.com/flightctl/flightctl/tree/main/test/scripts/agent-images/flavors
These are fed into a single parameterized base [Containerfile](https://github.com/flightctl/flightctl/blob/main/test/scripts/agent-images/base/Containerfile) to produce various agent images. 

By looking at the service Containerfiles in this PR, it seems that only the PAM issuer requires multiple parameters, and all other images differ only by the base image. @gshilin-sdb, maybe it would be simpler to have two Containerfiles (EL9 and EL10) just for the PAM issuer, and have the rest parameterized only by the base image?

.github/workflows/build-images-and-charts.yaml
```diff
@@ -65,15 +70,66 @@ jobs:
           REGISTRY_OWNER: flightctl
         run: |
           set -euo pipefail
-          
+
           echo "::group::Building backend containers"
```
sk-ilya: The pipeline should be as simple as possible, with steps that can be easily replicated locally, especially for an operation as common as `build-containers`.  Do we need both `make build-containers-ci` and `make build-containers`?  Can it be just something like `make -j4 build-containers FLAVOR=$flavor` (a single line as it was before)?

.github/workflows/build-images-and-charts.yaml
```diff
@@ -83,11 +139,22 @@ jobs:
...
           else
             echo "bundle-images target not found, falling back to direct save"
-            echo "Getting list of loaded flightctl images with tag: $TAG"
-            mapfile -t loaded_images < <(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${TAG}")
+            echo "Getting list of loaded flightctl images with flavor $FLAVOR and tag: $TAG"
+            # Try flavor-in-tag images first (new naming pattern)
+            mapfile -t loaded_images < <(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${FLAVOR}-${TAG}" || true)
+            # Fallback to non-flavor images (legacy naming) if no flavor-specific images found
+            if [ ${#loaded_images[@]} -eq 0 ]; then
+              echo "No flavor-in-tag images found, trying legacy naming pattern..."
```
sk-ilya: How can they not be found if they were built in the previous step?

.github/workflows/build-images-and-charts.yaml
```diff
@@ -98,26 +165,38 @@ jobs:
...
           BUNDLE_FILE: ${{ steps.bundle-images.outputs.bundle_file }}
         run: |
           set -euo pipefail
           echo "::group::Listing images in bundle"
-          IMAGES=$(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${TAG}" | sort)
+          # Try flavor-in-tag images first (new naming pattern)
+          IMAGES=$(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${FLAVOR}-${TAG}" | sort || true)
+          # Fallback to non-flavor images (legacy naming) if no flavor-specific images found
+          if [ -z "$IMAGES" ]; then
+            echo "No flavor-in-tag images found, listing legacy naming pattern..."
```
sk-ilya: How can they not be found if they were built in the previous step?

.github/workflows/pr-e2e-testing.yaml
```diff
@@ -175,7 +186,8 @@ jobs:
   e2e-tests:
     needs:
       - compute-tag
-      - build-images-and-charts
+      - build-images-and-charts-el9
```
sk-ilya: Consider building these in a matrix, same as rpms and agent images

.github/workflows/pr-e2e-testing.yaml
```diff
@@ -187,30 +199,35 @@ jobs:
...
             backend_type: helm
             label_filter: 'sanity && agent && !slow'
             ginkgo_total_nodes: 2
-          - os_id: cs10-bootc
+            flavor: el10
+          - os_id: el10-bootc
             backend_type: helm
             label_filter: 'sanity && agent && slow'
             ginkgo_total_nodes: 1
+            flavor: el10
```
sk-ilya: Can we specify the backend image OS and the agent’s `os_id` independently?

.github/workflows/pr-helm-upgrade-testing.yaml
```diff
@@ -223,15 +224,40 @@ jobs:
...
           CHART_FILE="artifacts-pr/flightctl-chart.tgz"
-          # Use values files with image overrides
+          # Use flavor-in-tag approach (matching base image pattern)
+          echo "=== Starting helm upgrade ==="
           helm upgrade flightctl "${CHART_FILE}" \
             --namespace flightctl-external \
             --reuse-values \
+            --set upgradeHooks.databaseMigrationDryRun=false \
+            --set dbSetup.image.tag="el9-${PR_GIT_TAG}" \
+            --timeout 20m \
```
sk-ilya: The default timeout (5 min IIRC) should be just fine

.github/workflows/pr-helm-upgrade-testing.yaml
```diff
@@ -213,25 +225,68 @@ jobs:
...
+            ls -la artifacts-pr/
+            exit 1
+          fi
+          echo "Using chart file: $CHART_FILE"
+          # Use flavor-in-tag approach (matching base image pattern)
+          echo "=== Starting helm upgrade ==="
           helm upgrade flightctl "${CHART_FILE}" \
             --namespace flightctl-external \
             --reuse-values \
+            --set upgradeHooks.databaseMigrationDryRun=false \
```
sk-ilya: `upgradeHooks.databaseMigrationDryRun=false` why is it disabled? it should run normally as part of an upgrade

deploy/helm/flightctl/values.yaml
```diff
@@ -394,7 +394,7 @@ dbSetup:
     # -- Number of retries for the migration Job on failure 
     backoffLimit: 2147483647
     # -- Maximum runtime in seconds for the migration Job (0 = no deadline)
-    activeDeadlineSeconds: 0
+    activeDeadlineSeconds: 1500  # 25 minutes - must be longer than helm timeout (20m) to avoid conflicts
```
sk-ilya: The default behavior here was to keep retrying indefinitely or until the helm timeout is reached (if deployed with `--wait`).  Why is this changed? 
Also the default helm timeout is 5 min, not 20

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: reverted

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: removed

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: removed

.github/workflows/pr-e2e-testing.yaml
gshilin-sdb: done

.github/workflows/pr-e2e-testing.yaml
gshilin-sdb: done

.github/workflows/pr-helm-upgrade-testing.yaml
gshilin-sdb: removed

.github/workflows/pr-helm-upgrade-testing.yaml
gshilin-sdb: removed

deploy/helm/flightctl/values.yaml
gshilin-sdb: reverted

images/Containerfile.alert-exporter
```diff
@@ -1,11 +1,23 @@
...
+# Build EL9: podman build --build-arg BUILD_IMAGE=registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 --build-arg RUNTIME_IMAGE=quay.io/flightctl/flightctl-base:el9-9.7-1762965531 --build-arg EL_VERSION=9 -f images/Containerfile.alert-exporter -t flightctl-alert-exporter-el9
+# Build EL10: podman build --build-arg BUILD_IMAGE=registry.access.redhat.com/ubi10/go-toolset:10.1-1770279878 --build-arg RUNTIME_IMAGE=quay.io/flightctl/flightctl-base:el10-10.1-1769518576 --build-arg EL_VERSION=10 -f images/Containerfile.alert-exporter -t flightctl-alert-exporter-el10
+
+
+ARG BUILD_IMAGE
+ARG RUNTIME_IMAGE
+ARG EL_VERSION
+ARG RHEM=false
+
+ARG RHEM=false
```
asafbss: You are declaring a duplicate ARG with the same name (RHEM).

Could you clarify why we need the RHEM build argument and what it is used for? I don’t see it being referenced in this stage.

images/Containerfile.alert-exporter
```diff
@@ -1,11 +1,23 @@
...
+
+
+FROM ${BUILD_IMAGE} as build
 WORKDIR /app
 
 # Switch to root for build operations
 USER 0
 
 # Copy dependencies first (most stable)
-COPY go.* ./
```
asafbss, line 8: Why did we remove this?

images/Containerfile.alert-exporter
```diff
@@ -39,14 +51,19 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
-  description="Flight Control Edge management service, alert exporter" \
-  io.k8s.description="Flight Control Edge management service, alert exporter" \
+  description="Flight Control alert exporter service" \
+  io.k8s.description="Flight Control alert exporter service" \
   io.k8s.display-name="Flight Control Alert Exporter" \
-  name="flightctl-alert-exporter" \
-  summary="Flight Control Edge management service, alert exporter"
+  name="flightctl-alert-exporter-el${EL_VERSION}" \
+  summary="Flight Control alert exporter service (el${EL_VERSION})" \
+  os.id="el${EL_VERSION}"
```
asafbss: Since this image inherits from UBI, which already provides OS metadata and labels, could you clarify why we need to introduce this custom os.id="el${EL_VERSION}" label?

images/Containerfile.alert-exporter
```diff
@@ -39,14 +51,19 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
 
-FROM quay.io/flightctl/flightctl-base:9.7-1762965531
+FROM ${RUNTIME_IMAGE}
 WORKDIR /app
 LABEL \
   com.redhat.component="flightctl-alert-exporter-container" \
-  description="Flight Control Edge management service, alert exporter" \
-  io.k8s.description="Flight Control Edge management service, alert exporter" \
+  description="Flight Control alert exporter service" \
+  io.k8s.description="Flight Control alert exporter service" \
```
asafbss, line 50: If this change is not strictly necessary, I would prefer to avoid modifying existing labels in this PR, since it appears unrelated to the main scope of the change.

images/Containerfile.alert-exporter
```diff
@@ -39,14 +51,19 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
-  name="flightctl-alert-exporter" \
-  summary="Flight Control Edge management service, alert exporter"
+  name="flightctl-alert-exporter-el${EL_VERSION}" \
+  summary="Flight Control alert exporter service (el${EL_VERSION})" \
+  os.id="el${EL_VERSION}"
 COPY --from=build /app/bin/flightctl-alert-exporter .
-CMD ./flightctl-alert-exporter
+
+# Ensure binary has execute permissions
+RUN chmod +x ./flightctl-alert-exporter
```
asafbss: The image was building and running correctly before, so I’m wondering what issue this is addressing.

If the binary is already built with executable permissions, this chmod +x should not be necessary. Could you clarify the reason for adding it?

images/Containerfile.alert-exporter
```diff
@@ -1,11 +1,23 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 as build
+# FlightCtl Alert Exporter
+# Build EL9: podman build --build-arg BUILD_IMAGE=registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 --build-arg RUNTIME_IMAGE=quay.io/flightctl/flightctl-base:el9-9.7-1762965531 --build-arg EL_VERSION=9 -f images/Containerfile.alert-exporter -t flightctl-alert-exporter-el9
```
asafbss: Keep the container file clean.

images/Containerfile.alert-exporter
```diff
@@ -1,11 +1,23 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 as build
+# FlightCtl Alert Exporter
```
asafbss: The following comments apply to all Containerfiles.

asafbss: If we are not introducing additional version-specific Containerfiles, we should consider moving this back to the main repository directory to avoid unnecessary structural changes.

internal/auth/oidc/pam/templates/.excalidraw-ignore
```diff
@@ -0,0 +1 @@
+flight-control-logo.svg
```
asafbss: Why?

test/scripts/agent-images/scripts/build_and_qcow2.sh
```diff
@@ -6,13 +6,13 @@ set -euo pipefail
...
 #       IMAGE_REPO     (quay.io/flightctl/flightctl-device by default in callers)
-#       OS_ID          (flavor id, e.g., cs9-bootc or cs10-bootc)
+#       OS_ID          (flavor id, e.g., el9-bootc or el10-bootc)
 #   - Tools:
 #       ./build.sh, ./bundle.sh, ./qcow2.sh, ./upload-images.sh in the same directory
 #
 # Usage:
-#   OS_ID=cs9-bootc TAG=vX IMAGE_REPO=... ./build_and_qcow2.sh
-#   ./build_and_qcow2.sh --os-id cs10-bootc --push
+#   OS_ID=el9-bootc TAG=vX IMAGE_REPO=... ./build_and_qcow2.sh
```
asafbss, line 14: I can still see references to cs9 and cs10 when searching across the repository (e.g., in the docs and Makefile).
Could you please update those as well for consistency?

docs/user/installing/installing-service-on-linux.md
```diff
@@ -52,6 +52,23 @@ or by checking the running containers with
...
+Flight Control containers are built for different CentOS Stream versions to ensure compatibility with your host environment:
+
+- **CS9 containers**: Default, compatible with RHEL 9, CentOS Stream 9, and Fedora
+- **CS10 containers**: Available for RHEL 10, CentOS Stream 10 environments
+
+The RPM packages automatically select appropriate container versions based on your host OS. When building from source, you can choose container flavors using:
+
+```bash
+# Build CS9 containers (default)
+make build-containers
```
asafbss: I’m not sure that recommending _make build-containers_ in the user documentation is the right approach.
@avishayt @fzdarsky  - PTAL

docs/user/installing/installing-service-on-kubernetes.md
```diff
@@ -121,6 +121,41 @@ kubectl get pods -n flightctl
...
+## Container OS Compatibility
+
+Flight Control supports container images built for different CentOS Stream versions to ensure compatibility across diverse environments:
+
+- **CS9 (CentOS Stream 9)**: Default and recommended for most deployments
+- **CS10 (CentOS Stream 10)**: For environments requiring latest OS version
+
+### Selecting Container Flavors
+
+By default, Helm charts use CS9 container images. For specific OS compatibility requirements, you can configure the deployment to use CS10 images:
```
asafbss: We should think about how this scales for future versions (e.g., el11). Hardcoding version numbers in documentation and scripts may require updates in multiple places for each new release.

Makefile
```diff
@@ -41,6 +41,10 @@ RPM_MOCK_ROOT_DEFAULT = centos-stream+epel-next-9-x86_64
 
 VERBOSE ?= false
 
+# Multi-OS container support
+OS_ID ?= el9
```
asafbss: This looks misleading.

Both OS_ID and FLAVOR default to el9, and it’s not immediately clear why we need both variables.

Could you provide an example that demonstrates why we need to keep them separate? If they always move together, it might make sense to simplify this to a single variable.

Makefile
```diff
@@ -347,8 +278,58 @@ clean-containers:
 	- podman rmi flightctl-telemetry-gateway:latest || true
 	- podman rmi flightctl-imagebuilder-api:latest || true
 	- podman rmi flightctl-imagebuilder-worker:latest || true
-
-build-containers: flightctl-api-container flightctl-pam-issuer-container flightctl-db-setup-container flightctl-worker-container flightctl-periodic-container flightctl-alert-exporter-container flightctl-alertmanager-proxy-container flightctl-multiarch-cli-container flightctl-userinfo-proxy-container flightctl-telemetry-gateway-container flightctl-imagebuilder-api-container flightctl-imagebuilder-worker-container
+	# Old flavor-in-name approach (for cleanup during transition)
```
asafbss, line 343: Could you please explain the rationale behind supporting all three naming approaches (legacy, flavor-in-name, and flavor-in-tag)?

Is the "marked" one the one we use today?

Makefile
```diff
@@ -347,8 +278,58 @@ clean-containers:
 	- podman rmi flightctl-telemetry-gateway:latest || true
 	- podman rmi flightctl-imagebuilder-api:latest || true
 	- podman rmi flightctl-imagebuilder-worker:latest || true
-
-build-containers: flightctl-api-container flightctl-pam-issuer-container flightctl-db-setup-container flightctl-worker-container flightctl-periodic-container flightctl-alert-exporter-container flightctl-alertmanager-proxy-container flightctl-multiarch-cli-container flightctl-userinfo-proxy-container flightctl-telemetry-gateway-container flightctl-imagebuilder-api-container flightctl-imagebuilder-worker-container
```
asafbss, line 351: What if I want something specific?

.github/actions/collect-logs/action.yml
```diff
@@ -42,11 +42,16 @@ runs:
...
+          if kubectl get deployment/$deployment -n $namespace >/dev/null 2>&1; then
+            if kubectl logs -n $namespace deployment/$deployment --all-containers=true \
+               > ${{ inputs.log-directory }}/${deployment}_${timestamp}.log 2>/dev/null; then
+              log_files_collected=true
+            else
+              echo "Error collecting logs for $deployment"
+            fi
           else
-            echo "Error collecting logs for $deployment"
+            echo "Deployment $deployment not found in namespace $namespace, skipping"
```
asafbss: I get the intent (making the script more robust), but this PR is already quite large.
Could we avoid introducing additional non-essential changes like this, and keep the scope focused on the main goal of the PR?

.github/workflows/build-images-and-charts.yaml
```diff
@@ -63,17 +68,29 @@ jobs:
...
+          FLAVOR: ${{ inputs.flavor || 'el9' }}
         run: |
           set -euo pipefail
-          
-          echo "::group::Building backend containers"
-          make -j4 build-containers
+          echo "::group::Building backend containers for flavor: $FLAVOR"
+
+          # Use appropriate build method: build-containers-ci only supports el9
+          if [ "$FLAVOR" = "el9" ] && make -n build-containers-ci >/dev/null 2>&1; then
```
asafbss: This is a hardcoding version-specific logic (FLAVOR = el10, 9) and special-casing script behavior. That’s likely to bite us as soon as we add another version (e.g., el11).

Can we refactor this to be version-agnostic (or data-driven) instead of adding more blocks per release?

.github/actions/wait-for-api-readiness/action.yaml
```diff
@@ -21,26 +21,80 @@ runs:
         BASE_DOMAIN: ${{ inputs.base-domain }}
         RETRIES: ${{ inputs.retries }}
       run: |
+        echo "=== WAITING FOR API READINESS ==="
+        echo "Endpoint: https://api.${BASE_DOMAIN}:3443/readyz"
+        echo "Max retries: $RETRIES"
+        echo ""
+
         for i in $(seq 1 $RETRIES); do
+            # Show detailed debugging every 10 seconds
```
asafbss: I get the intent (making the script more robust), but this PR is already quite large.
Could we avoid introducing additional non-essential changes like this, and keep the scope focused on the main goal of the PR?

.github/actions/wait-for-api-readiness/action.yaml
```diff
@@ -21,26 +21,80 @@ runs:
...
-            printf '%s\n' "$body"
+            # Print response body (only if not empty and not a connection error)
+            if [ -n "$body" ] && [ "$body" != "curl: (7) Failed to connect"* ]; then
+                printf '[%d] API Response: %s\n' "$i" "$body"
+            fi
 
             if [ "$http_code" -ge 200 ] && [ "$http_code" -lt 300 ]; then
-                echo "OK: API responded with $http_code on attempt $i" >&2
+                echo ""
+                echo "✅ SUCCESS: API responded with $http_code on attempt $i" >&2
```
asafbss: AVOID AI SYMBOLS like ✅ and ❌

deploy/helm/cmd/charttmpl/main.go
```diff
@@ -59,9 +59,25 @@ func runTemplate(in string, out string, templateData templateContext) {
...
+	// Get FLAVOR, default to el9 for consistency
+	flavor := os.Getenv("FLAVOR")
+	if flavor == "" {
+		flavor = "el9"
+	}
+
+	// Validate flavor is el9 or el10
+	profileFlavor := flavor
+	switch flavor {
+	case "el9", "el10":
```
asafbss: Hardcoding "el9" and "el10" here does not scale well.

When we introduce a new version (e.g., el11), we’ll have to keep updating this switch. Could we make this version-agnostic?

images/Containerfile.alert-exporter
gshilin-sdb: All removed

images/Containerfile.alert-exporter
gshilin-sdb: restored

images/Containerfile.alert-exporter
gshilin-sdb: removed

images/Containerfile.alert-exporter
gshilin-sdb: reverted

images/Containerfile.alert-exporter
gshilin-sdb: This was a request of Frank

images/Containerfile.alert-exporter
gshilin-sdb:
```
  Target database: flightctl                                                                                                                            
  Error: crun: open executable: Permission denied: OCI permission denied                                                                                
  make[1]: *** [test/test.mk:87: _run_template_migration] Error 126                                                                                     
  make[1]: Leaving directory '/home/runner/work/flightctl/flightctl'                                                                                    
  make[1]: Entering directory '/home/runner/work/flightctl/flightctl'                                                                                   
  sudo systemctl stop flightctl-alertmanager.service                                                                                                    
  sudo systemctl stop flightctl-kv.service                                                                                                              
  sudo systemctl stop flightctl-db.service                                                                                                              
  make[1]: Leaving directory '/home/runner/work/flightctl/flightctl'                                                                                    
  make: *** [test/test.mk:61: integration-test] Error 2                                                                                                 
  Error: Process completed with exit code 2.                                                                                                            

```

The error indicates:

```
Error: crun: open executable: Permission denied: OCI permission denied                                                                                    

```
   
This suggests the flightctl-db-setup binary inside the container doesn't have execute permissions.                                                        
                                                                                                                                                          
When we removed these lines from all Containerfiles:

```
  RUN chmod +x /usr/local/bin/flightctl-*
```
We assumed Go builds always create executables with proper permissions, but in containerized environments with strict security contexts, the execute permissions might not be preserved during the COPY operation.

This is the proper approach:
```
COPY --chmod=0755 --from=build /app/bin/flightctl-alert-exporter .
```

images/Containerfile.alert-exporter
gshilin-sdb: removed

internal/auth/oidc/pam/templates/.excalidraw-ignore
gshilin-sdb: removed

test/scripts/agent-images/scripts/build_and_qcow2.sh
gshilin-sdb: removed

docs/user/installing/installing-service-on-kubernetes.md
gshilin-sdb: There are some places where you have to mention the existing values. Probably we can add "etc." for the future ones, but I'd prefer it to be a full list....

Makefile
gshilin-sdb: replaced OS_ID with FLAVOR everywhere

Makefile
gshilin-sdb: The clean-containers target in the Makefile supports all three because:
  1. Cleanup during transition: Remove old images from previous naming schemes
  2. Developer environments: Developers might have built images using old approaches
  3. CI/CD: Ensure clean state regardless of what was built before

Makefile
gshilin-sdb: oops... reverted specific targets + made them to use FLAVOR

.github/actions/collect-logs/action.yml
gshilin-sdb: i'm trying to understand why tests are failing, so added some debugging... we can clean it later, though I'd keep it

.github/actions/wait-for-api-readiness/action.yaml
gshilin-sdb: i'm trying to understand why tests are failing, so added some debugging... this one I'll remove, but later

.github/actions/wait-for-api-readiness/action.yaml
gshilin-sdb: These symbols are used by ginkgo, so you'll see them all over the tests. This script is part of the tests, so it's quite logical to see these symbols here.

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: done

deploy/helm/cmd/charttmpl/main.go
gshilin-sdb: done

docs/user/installing/installing-service-on-linux.md
gshilin-sdb: replaced

hack/build_flightctl-base.sh
```diff
@@ -1,26 +1,66 @@
...
+
+# Set base image and version info based on FLAVOR (matching publish_containers.sh)
+case "$FLAVOR" in
+    el9)
+        BASE_IMAGE="registry.access.redhat.com/ubi9/ubi-minimal:9.7-1763362218"
+        EL_VERSION="9"
+        BASE_TAG="9.7-1763362218"
+        ;;
+    el10)
+        BASE_IMAGE="registry.access.redhat.com/ubi10/ubi-minimal:10.1-1769677092"
```
asafbss: The base image was intentionally chosen to be micro.

images/Containerfile.alertmanager-proxy
```diff
@@ -1,11 +1,14 @@
...
+ARG EL_VERSION
+
+FROM ${BUILD_IMAGE} as build
 WORKDIR /app
 
 # Switch to root for build operations
 USER 0
 
 # Copy dependencies first (most stable)
-COPY go.* ./
```
asafbss, line 8: Was not reverted

images/Containerfile.alertmanager-proxy
```diff
@@ -39,15 +42,15 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
     make build-alertmanager-proxy
 
-FROM quay.io/flightctl/flightctl-base:9.7-1762965531
+FROM ${RUNTIME_IMAGE}
 WORKDIR /app
 LABEL \
   com.redhat.component="flightctl-alertmanager-proxy-container" \
-  description="Flight Control Edge management service, alertmanager proxy" \
-  io.k8s.description="Flight Control Edge management service, alertmanager proxy" \
+  description="Flight Control alertmanager proxy service" \
```
asafbss: Was not reverted

images/Containerfile.alertmanager-proxy
```diff
@@ -39,15 +42,15 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
+  io.k8s.description="Flight Control alertmanager proxy service" \
   io.k8s.display-name="Flight Control Alertmanager Proxy" \
-  name="flightctl-alertmanager-proxy" \
-  summary="Flight Control Edge management service, alertmanager proxy"
-COPY --from=build /app/bin/flightctl-alertmanager-proxy .
-EXPOSE 8443
```
asafbss, line 52: Why?

hack/build_flightctl-base.sh
gshilin-sdb: done

images/Containerfile.alertmanager-proxy
gshilin-sdb: done

images/Containerfile.alertmanager-proxy
gshilin-sdb: restored

hack/build_flightctl-base.sh
gshilin-sdb: pls note: cli artifacts and db-setup require microdnf to install postgresql and something else, so they require minimal base image

.github/actions/collect-logs/action.yml
```diff
@@ -34,19 +34,24 @@ runs:
         kubectl get all,configmaps,secrets --all-namespaces 
         echo "::endgroup::"
         
-        for deployment in flightctl-api flightctl-db flightctl-periodic flightctl-worker flightctl-alert-exporter flightctl-imagebuilder-api flightctl-kv; do
+        for deployment in flightctl-api flightctl-db flightctl-periodic flightctl-worker flightctl-alert-exporter flightctl-imagebuilder-api; do
```
fzdarsky: The changes in this file look sensible, but they're not strictly necessary to enable el10 support, or?

images/Containerfile.imagebuilder-worker
```diff
@@ -41,27 +42,15 @@ RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
...
-    echo 'driver = "overlay"' >> /etc/containers/storage.conf
-# Copy the worker binary
-COPY --from=build /app/bin/flightctl-imagebuilder-worker .
-
+  name="flightctl-imagebuilder-worker-el${EL_VERSION}" \
+  summary="Flight Control image builder worker service (el${EL_VERSION})"
+COPY --chmod=0755 --from=build /app/bin/flightctl-imagebuilder-worker .
 
-CMD ["./flightctl-imagebuilder-worker"]
+CMD ./flightctl-imagebuilder-worker
```
fzdarsky: Please do not change the exec form of `CMD` to the shell form of `CMD`, because it changes the behavior: The shell form runs the command in a Bash shell instead of as PID 1, which impliess the process no longer receives graceful shutdown signals. I understand most other Containerfiles do the same, that needs wider discussion.

docs/user/installing/installing-service-on-linux.md
```diff
@@ -52,6 +52,23 @@ or by checking the running containers with
...
+The RPM packages automatically select appropriate container versions based on your host OS. When building from source, you can choose container flavors using:
+
+```bash
+# Build EL9 containers (default)
+make build-containers
+
+# Build specific flavor
+FLAVOR=el10 make build-containers
+```
+
```
fzdarsky, line 63: This is the user documentation. Users should never use `make` or even have to know about EL9 vs EL10 (the repo file should handle that for them).

docs/user/installing/installing-service-on-kubernetes.md
```diff
@@ -315,4 +315,348 @@ If you need to access your service using the `flightctl` CLI, find and click the
 
 ## Installing with Advanced Cluster Management
 
-## Installing with Ansible Automation Platform
+Flight Control can be deployed alongside Advanced Cluster Management (ACM) for enhanced multi-cluster management capabilities.
```
fzdarsky: I just removed all this text during a reorganisation+rewrite of the installation docs (rewrite of ACM and AAP is still pending).

deploy/podman/images.yaml.gotmpl
```diff
@@ -0,0 +1,70 @@
...
+  tag: latest
+alertmanager-proxy:
+  image: quay.io/flightctl/flightctl-alertmanager-proxy-{{.Flavor}}
+  tag: latest
+pam-issuer:
+  image: quay.io/flightctl/flightctl-pam-issuer-{{.Flavor}}
+  tag: latest
+ui:
+  image: quay.io/flightctl/flightctl-ui
+  tag: latest
```
fzdarsky, line 24: Shouldn't the UI also have different flavors?

.github/actions/collect-logs/action.yml
gshilin-sdb: cleaned

images/Containerfile.imagebuilder-worker
gshilin-sdb: fixed

docs/user/installing/installing-service-on-linux.md
gshilin-sdb: left a minimum just for info

docs/user/installing/installing-service-on-kubernetes.md
gshilin-sdb: removed. bad rebase ;(

deploy/podman/images.yaml.gotmpl
gshilin-sdb: probably, but it was not part of this epic AFAIK

docs/user/introduction.md
```diff
@@ -16,8 +16,9 @@ Features and use cases Flight Control aims to support:
 * Container or VM workloads on Podman using docker-compose or Quadlets, Kubernetes services on MicroShift using kustomize or Helm.
 * Agent-based architecture, allowing:
   * scalable and robust management under adverse networking conditions (agent "calls home" when the device has connectivity),
-  * autonomous and safe OTA updates (agent downloads and verifies assets before updating, then transactionally updates or rolls back), and
-  * timely yet resource-unintensive monitoring of devices and applications (agent notifies service on update progress and alarms).
+  * autonomous and safe OTA updates (agent downloads and verifies assets before updating, then transactionally updates or rolls back),
+  * timely yet resource-unintensive monitoring of devices and applications (agent notifies service on update progress and alarms), and
+  * cross-version compatibility (service supports CentOS Stream 9 and 10 containers for diverse deployment environments).
```
fzdarsky: An agent-based architecture allows for "cross-version compatibility"?

README.md
```diff
@@ -14,3 +14,4 @@ For more information, please refer to:
 
 [![Watch the demo](docs/images/demo-thumbnail.png)](https://www.youtube.com/watch?v=WzNG_uWnmzk)
 
+# Trigger CI after cache cleanup
```
fzdarsky: ?

docs/user/introduction.md
gshilin-sdb: Probably is just bad wording and its meaning is that we support both cats and dogs. Give me an good sentence in English please.

Actually, but secretly, we support cross-breading as well.

README.md
gshilin-sdb: oops, removed

.github/actions/collect-logs/action.yml
gshilin-sdb: cleaned

images/Containerfile.imagebuilder-worker
gshilin-sdb: reverted

docs/user/installing/installing-service-on-linux.md
gshilin-sdb: removed the "make" part

docs/user/installing/installing-service-on-kubernetes.md
gshilin-sdb: removed it too, bad rebase :(

deploy/podman/images.yaml.gotmpl
gshilin-sdb: it was not part of the epic (or at least I didn't think in this direction...)

.github/workflows/build-images-and-charts.yaml
```diff
@@ -17,19 +17,30 @@ on:
         required: false
         type: string
         default: ""
+      flavor:
+        description: "Container flavor to build (el9 or el10)"
+        required: false
```
asafbennatan: i would say this should be required

.github/workflows/build-images-and-charts.yaml
```diff
@@ -17,19 +17,30 @@ on:
...
         value: ${{ jobs.build-cli.outputs.git_tag }}
+      helm_tag:
+        description: "Computed helm tag (without v prefix)"
+        value: ${{ jobs.build-helm-charts.outputs.helm_tag }}
 
 jobs:
   build-backend-containers:
     runs-on: ubuntu-24.04
+    env:
+      FLAVOR: ${{ inputs.flavor || 'el9' }}
```
asafbennatan: should be required so no need for ||

.github/workflows/build-images-and-charts.yaml
```diff
@@ -65,9 +79,24 @@ jobs:
           REGISTRY_OWNER: flightctl
         run: |
           set -euo pipefail
-          
-          echo "::group::Building backend containers"
-          make -j4 build-containers
+          echo "::group::Building backend containers for flavor: $FLAVOR"
```
asafbennatan, line 81: we dont need these fallbacks , there should be exactly one successful way , if it fails everything fails

.github/workflows/build-images-and-charts.yaml
```diff
@@ -83,11 +112,41 @@ jobs:
             make bundle-images
           else
             echo "bundle-images target not found, falling back to direct save"
-            echo "Getting list of loaded flightctl images with tag: $TAG"
-            mapfile -t loaded_images < <(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${TAG}")
+            echo "Getting list of loaded flightctl images with flavor $FLAVOR and tag: $TAG"
```
asafbennatan: i dont understand why we need this fallback mechanism , is there a reason 
this is not enough:

```
          # Try flavor-in-tag images first (new naming pattern)
            mapfile -t loaded_images < <(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${FLAVOR}-${TAG}" || true)
```

.github/workflows/build-images-and-charts.yaml
```diff
@@ -102,22 +161,40 @@ jobs:
         run: |
           set -euo pipefail
           echo "::group::Listing images in bundle"
-          IMAGES=$(podman images --format "{{.Repository}}:{{.Tag}}" | grep "quay\.io/flightctl/flightctl-.*:${TAG}" | sort)
+          # Try exact flavor-tag match first
```
asafbennatan: no fallbacks

.github/workflows/build-images-and-charts.yaml
```diff
@@ -155,6 +232,11 @@ jobs:
 
   build-helm-charts:
     runs-on: ubuntu-24.04
+    env:
+      FLAVOR: ${{ inputs.flavor || 'el9' }}
```
asafbennatan: would be required no need for ||

.github/workflows/build-images-and-charts.yaml
```diff
@@ -192,16 +284,19 @@ jobs:
...
+          name: helm-chart-${{ steps.set-flavor.outputs.flavor }}-${{ steps.compute-tag.outputs.git_tag }}
           path: ${{ steps.build-charts.outputs.chart_file }}
           compression-level: 0
           if-no-files-found: error
           retention-days: 1
 
   build-cli:
     runs-on: ubuntu-24.04
+    env:
+      FLAVOR: ${{ inputs.flavor || 'el9' }}
```
asafbennatan: same as above

.github/workflows/build-images-and-charts.yaml
```diff
@@ -253,6 +352,10 @@ jobs:
 
   build-simulator:
     runs-on: ubuntu-24.04
+    env:
+      FLAVOR: ${{ inputs.flavor || 'el9' }}
```
asafbennatan: same as above

.github/workflows/integration-tests.yaml
```diff
@@ -47,11 +47,13 @@ jobs:
           COMMIT_SHA="${{ github.sha }}"
           echo "Building db-migrator image tagged with commit: ${COMMIT_SHA}"
 
-          # Build the db-setup container with commit tag
-          sudo make flightctl-db-setup-container
+          # Build the db-setup container with configurable FLAVOR (defaults to el9)
+          FLAVOR=${FLAVOR:-el9}
```
asafbennatan: dont allow flavor to be optional

.github/workflows/run-e2e-tests.yaml
```diff
@@ -26,6 +26,11 @@ on:
         description: "Image tag"
         required: true
         type: string
+      flavor:
```
asafbennatan, line 29: should not be optional

.github/workflows/run-e2e-tests.yaml
```diff
@@ -89,20 +138,45 @@ jobs:
...
+      - name: Debug E2E artifact download parameters
+        run: |
+          echo "os_id: '${{ inputs.os_id }}'"
+          echo "tag: '${{ inputs.tag }}'"
+          echo "pattern: 'agent-*-${{ inputs.os_id }}-${{ inputs.tag }}'"
+
       - name: Download test artifacts
+        id: download-test-artifacts
         uses: actions/download-artifact@v4
+        continue-on-error: true
```
asafbennatan: why ?

.github/workflows/run-e2e-tests.yaml
```diff
@@ -89,20 +138,45 @@ jobs:
...
         with:
           pattern: agent-*-${{ inputs.os_id }}-${{ inputs.tag }}
           path: test-artifacts
           merge-multiple: true
 
       - name: Stage test artifacts
         run: |
+          if [ "${{ steps.download-test-artifacts.outcome }}" != "success" ]; then
```
asafbennatan: this was added to supported the continue-on-error which is unnecessary

.github/workflows/run-e2e-tests.yaml
```diff
@@ -130,10 +204,20 @@ jobs:
           df -h
 
       # ========== PREP E2E ENV ==========
+      - name: Load API readiness timeout configuration
```
asafbennatan: i think this is a lot of extract code just to control the timeout, 
in general if RHEL10 takes 240 seconds more to start this is a very critical bug by the RHEL team that would be fixed eventually there is no need to make that configurable , 
meanwhile while that is fixed -> just change the global timeout to 300 across the board

.github/workflows/run-e2e-tests.yaml
```diff
@@ -148,6 +232,19 @@ jobs:
           make prepare-e2e-test
 
       # ========== RUN E2E TESTS ==========
+      - name: Pre-test validation
```
asafbennatan: remove once PR is ready

.github/workflows/run-e2e-tests.yaml
```diff
@@ -163,8 +260,6 @@ jobs:
         run: |
           HASH=$(printf '%s' "${{ inputs.ginkgo_label_filter }}" | sha256sum | cut -c1-8)
           echo "hash=${HASH}" >> "$GITHUB_OUTPUT"
-          echo "Label filter: ${{ inputs.ginkgo_label_filter }}"
```
asafbennatan, line 166: why remove ?

deploy/helm/flightctl/templates/flightctl-preupgrade-scale-jobs.yaml
```diff
@@ -111,21 +111,90 @@ spec:
               set -euo pipefail
               NAMESPACE={{ default .Release.Namespace .Values.global.internalNamespace }}
               TIMEOUT={{ default 120 .Values.upgradeHooks.scaleDown.timeoutSeconds }}
-              echo "Pre-upgrade scale to zero started in namespace: ${NAMESPACE} (timeout=${TIMEOUT}s)"
+              echo "=== PRE-UPGRADE SCALE-DOWN DEBUG ==="
```
asafbennatan: cleanup this debugging code

deploy/helm/flightctl/values.yaml.gotmpl
```diff
@@ -105,6 +105,12 @@ global:
   additionalRouteLabels: null
   # -- Additional labels for PVCs.
   additionalPVCLabels: null
+  # -- Minimal container image configuration for init containers and test pods
```
asafbennatan: i dont think this should be user facing , the chart has already generates proper values.yaml ,
additionally for upgrades using the el9 by default probably doesnt matter , we should instruct users to do 
--reset-then-reuse-values
which would solve that entirely

.github/workflows/pr-helm-upgrade-testing.yaml
```diff
@@ -219,6 +221,8 @@ jobs:
           helm diff upgrade flightctl "${CHART_FILE}" \
             --namespace flightctl-external \
             --reuse-values \
+            --set global.minimalImage.image=registry.access.redhat.com/ubi9-micro \
```
asafbennatan: should not be userfacing

.github/workflows/pr-helm-upgrade-testing.yaml
```diff
@@ -230,6 +234,8 @@ jobs:
           helm upgrade flightctl "${CHART_FILE}" \
             --namespace flightctl-external \
             --reuse-values \
+            --set global.minimalImage.image=registry.access.redhat.com/ubi9-micro \
```
asafbennatan: should not be userfacing

.github/workflows/pr-helm-upgrade-testing.yaml
```diff
@@ -302,6 +308,8 @@ jobs:
...
```
asafbennatan: should not be userfacing

deploy/deploy.mk
```diff
@@ -77,25 +87,58 @@ deploy-alertmanager-proxy:
...
+		sudo podman tag flightctl-periodic-$(FLAVOR):latest flightctl-periodic:latest; \
+		sudo podman tag flightctl-alert-exporter-$(FLAVOR):latest flightctl-alert-exporter:latest; \
+		sudo podman tag flightctl-cli-artifacts-$(FLAVOR):latest flightctl-cli-artifacts:latest; \
+		sudo podman tag flightctl-alertmanager-proxy-$(FLAVOR):latest flightctl-alertmanager-proxy:latest; \
+		sudo podman tag flightctl-pam-issuer-$(FLAVOR):latest flightctl-pam-issuer:latest; \
+		sudo podman tag flightctl-imagebuilder-api-$(FLAVOR):latest flightctl-imagebuilder-api:latest; \
+		sudo podman tag flightctl-imagebuilder-worker-$(FLAVOR):latest flightctl-imagebuilder-worker:latest; \
+		sudo podman tag flightctl-userinfo-proxy-$(FLAVOR):latest flightctl-userinfo-proxy:latest; \
+		sudo podman tag flightctl-telemetry-gateway-$(FLAVOR):latest flightctl-telemetry-gateway:latest; \
+	else \
```
asafbennatan: why fallback

deploy/deploy.mk
```diff
@@ -18,42 +18,52 @@ cluster: bin/e2e-certs/ca.pem
...
 	test/scripts/redeploy.sh api
 
-redeploy-worker: flightctl-worker-container
+redeploy-worker: FLAVOR ?= el9
+redeploy-worker: build-containers
 	test/scripts/redeploy.sh worker
 
-redeploy-periodic: flightctl-periodic-container
+redeploy-periodic: FLAVOR ?= el9
+redeploy-periodic: build-containers
```
asafbennatan: previously when we redeployed we built only the specific container , the change now requires us to build all containers

docs/developer/pam-issuer-architecture.md
```diff
@@ -45,7 +45,7 @@ The PAM issuer service has been separated into its own independent microservice
 **Build**:
 ```bash
 make build-pam-issuer          # Build binary
-make flightctl-pam-issuer-container  # Build container
+make build-containers             # Build all containers (el9 and el10)
```
asafbennatan: i shouldnt need to build all containers

hack/build_containers_ci.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
asafbennatan: why do we need a separate build script for CI ? CI should use the existing build system , having multiple build scripts will be hard to maintain

hack/build_containers_ci.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+# Check if hack/publish_containers.sh exists and supports build command
+if [ -f "hack/publish_containers.sh" ] && grep -q "build)" "hack/publish_containers.sh" 2>/dev/null; then
+    echo "Using hack/publish_containers.sh build for $FLAVOR"
+    exec hack/publish_containers.sh build "$FLAVOR"
+fi
+
+# Fallback to legacy build-containers (builds all flavors)
+if make -n build-containers >/dev/null 2>&1; then
```
asafbennatan: no fallbacks

hack/container-flavors.conf
```diff
@@ -0,0 +1,6 @@
+# Container flavor configuration
```
asafbennatan: this acts as another source of truth just for podman deployment 
we have another one in helmchart confgs

hack/publish_containers.sh
```diff
@@ -1,14 +1,128 @@
 #!/usr/bin/env bash
-set -x -e
+set -euo pipefail
```
asafbennatan, line 2: shouldnt this just call build single container ? seems like it duplicates this logic

images/Containerfile.cli-artifacts
```diff
@@ -1,11 +1,13 @@
...
+ARG RUNTIME_IMAGE
+
+FROM ${BUILD_IMAGE} as builder
 WORKDIR /app
 
 # Switch to root for build operations
 USER 0
 
 # Copy dependencies first (most stable)
-COPY go.* ./
```
asafbennatan, line 8: why removed ?

test/scripts/deploy_with_helm.sh
```diff
@@ -6,10 +6,28 @@ ONLY_DB=
 DB_SIZE_PARAMS=
 # If using images from a private registry, specify a path to a Kubernetes Secret yaml for your pull secret (in the flightctl-internal namespace)
 # IMAGE_PULL_SECRET_PATH=
-SQL_VERSION=${SQL_VERSION:-"latest"}
-SQL_IMAGE=${SQL_IMAGE:-"quay.io/sclorg/postgresql-16-c9s"}
-KV_VERSION=${KV_VERSION:-"7.4.1"}
-KV_IMAGE=${KV_IMAGE:-"docker.io/redis"}
+# Database image selection based on FLAVOR
+FLAVOR=${FLAVOR:-el9}
```
asafbennatan: shouldn't this be just a part of the flavor ? why are we setting flavor specific config here again ?

test/scripts/deploy_with_helm.sh
```diff
@@ -56,11 +74,27 @@ kubectl create namespace flightctl-e2e      --context kind-kind 2>/dev/null || t
 # if we are only deploying the database, we don't need inject the server container
 if [ -z "$ONLY_DB" ]; then
 
+  # Load required flavor-in-tag images (new naming approach)
   for suffix in periodic api worker alert-exporter alertmanager-proxy cli-artifacts db-setup telemetry-gateway imagebuilder-api imagebuilder-worker ; do
```
asafbennatan, line 115: why do we need fallback , assume the new naming

test/test.mk
```diff
@@ -101,8 +101,8 @@ _run_template_migration:
 	    echo "##################################################"; \
 	    echo "No MIGRATION_IMAGE provided; building a fresh one ..."; \
 	    echo "##################################################"; \
-	    $(MAKE) --no-print-directory -B flightctl-db-setup-container; \
-	    img="flightctl-db-setup:latest"; \
+	    FLAVOR=$$FLAVOR $(MAKE) --no-print-directory build-containers; \
```
asafbennatan: why do we need to build all containers ?

test/test.mk
```diff
@@ -137,7 +137,8 @@ bin/.e2e-agent-injected: bin/output/qcow2/disk.qcow2 bin/.e2e-agent-certs
 
 prepare-e2e-qcow-config: bin/.e2e-agent-injected
 
-prepare-e2e-test: RPM_MOCK_ROOT=centos-stream+epel-next-9-x86_64
+prepare-e2e-test: RPM_MOCK_ROOT=$(shell if [ "$(FLAVOR)" = "el10" ]; then echo "epel-10-x86_64"; else echo "centos-stream+epel-next-9-x86_64"; fi)
```
asafbennatan: this should be part of flavor configuration

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: done

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: removed all of them

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: yep, removed

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: removed

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: removed

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: removed

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: replaced with verification

.github/workflows/build-images-and-charts.yaml
gshilin-sdb: yes

.github/workflows/integration-tests.yaml
gshilin-sdb: done

.github/workflows/run-e2e-tests.yaml
gshilin-sdb: done

.github/workflows/run-e2e-tests.yaml
gshilin-sdb: removed

.github/workflows/run-e2e-tests.yaml
gshilin-sdb: removed

.github/workflows/run-e2e-tests.yaml
gshilin-sdb: removed

.github/workflows/run-e2e-tests.yaml
gshilin-sdb: removed

.github/workflows/run-e2e-tests.yaml
gshilin-sdb: reverted

deploy/helm/flightctl/templates/flightctl-preupgrade-scale-jobs.yaml
gshilin-sdb: done

deploy/helm/flightctl/values.yaml.gotmpl
gshilin-sdb: removed here and added the new flag do doc

.github/workflows/pr-helm-upgrade-testing.yaml
gshilin-sdb: removed

.github/workflows/pr-helm-upgrade-testing.yaml
gshilin-sdb: removed

.github/workflows/pr-helm-upgrade-testing.yaml
gshilin-sdb: removed

deploy/deploy.mk
gshilin-sdb: removed

deploy/deploy.mk
gshilin-sdb: fixed

docs/developer/pam-issuer-architecture.md
gshilin-sdb: reverted

hack/build_containers_ci.sh
gshilin-sdb: removed

hack/build_containers_ci.sh
gshilin-sdb: removed

.github/actions/check-simulator-logs/action.yml
```diff
@@ -65,14 +65,14 @@ runs:
           fi
         done <<< "${error_lines}"
 
-        echo "=== Simulator Log Error Summary ==="
+        echo "Simulator Log Error Summary:"
```
avishayt: Not relevant

.github/actions/setup-dependencies/action.yaml
```diff
@@ -37,7 +37,7 @@ runs:
           sudo apt install -y make python3-pip podman podman-compose libvirt-dev libpam0g-dev
 
           curl -Lo protoc.zip "https://github.com/protocolbuffers/protobuf/releases/download/v3.14.0/protoc-3.14.0-linux-x86_64.zip"
-          sudo unzip -q protoc.zip bin/protoc -d /usr/local
+          sudo unzip -qo protoc.zip bin/protoc -d /usr/local
```
avishayt: Not relevant

README.md
```diff
@@ -13,4 +13,3 @@ For more information, please refer to:
 <br><br>
 
 [![Watch the demo](docs/images/demo-thumbnail.png)](https://www.youtube.com/watch?v=WzNG_uWnmzk)
-
```
avishayt, line 16: Not relevant

hack/container-flavors.conf
gshilin-sdb: updated, now deploy/helm/helm-chart-opts.yaml is a single source of truth

hack/publish_containers.sh
gshilin-sdb: done

images/Containerfile.cli-artifacts
gshilin-sdb: was replaced with COPY go.mod go.sum ./ to be more specific
reverted

test/scripts/deploy_with_helm.sh
gshilin-sdb: yes, done

test/test.mk
gshilin-sdb: fixed

test/scripts/deploy_with_helm.sh
gshilin-sdb: fixed

test/test.mk
gshilin-sdb: added

Containerfile.db-setup
```diff
@@ -1,70 +0,0 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 as build
```
avishayt, line 1: Why is this file no longer needed?

docs/user/introduction.md
```diff
@@ -16,8 +16,9 @@ Features and use cases Flight Control aims to support:
 * Container or VM workloads on Podman using docker-compose or Quadlets, Kubernetes services on MicroShift using kustomize or Helm.
 * Agent-based architecture, allowing:
   * scalable and robust management under adverse networking conditions (agent "calls home" when the device has connectivity),
-  * autonomous and safe OTA updates (agent downloads and verifies assets before updating, then transactionally updates or rolls back), and
-  * timely yet resource-unintensive monitoring of devices and applications (agent notifies service on update progress and alarms).
+  * autonomous and safe OTA updates (agent downloads and verifies assets before updating, then transactionally updates or rolls back),
```
avishayt: Not needed - please revert

.github/actions/setup-dependencies/action.yaml
gshilin-sdb: it fails on being unable to override existing files

.github/actions/check-simulator-logs/action.yml
gshilin-sdb: reverted

README.md
gshilin-sdb: reverted

deploy/deploy.mk
```diff
@@ -77,25 +87,58 @@ deploy-alertmanager-proxy:
...
-	podman save flightctl-pam-issuer:latest | sudo podman load
-	podman save flightctl-imagebuilder-api:latest | sudo podman load
-	podman save flightctl-imagebuilder-worker:latest | sudo podman load
-	podman save flightctl-userinfo-proxy:latest | sudo podman load
-	podman save flightctl-telemetry-gateway:latest | sudo podman load
+	@echo "Copying containers from user to root context for systemd services (FLAVOR=$(FLAVOR))..."
+	# Check if flavor-specific images exist (new system) or fall back to old naming
+	@if podman image exists flightctl-api-$(FLAVOR):latest; then \
+		echo "Using flavor-specific container images..."; \
+		podman save flightctl-api-$(FLAVOR):latest | sudo podman load; \
```
avishayt: I thought the flavor was in the tag?

Containerfile.db-setup
gshilin-sdb: see line 5
```
FROM ${BUILD_IMAGE} as build
```

docs/user/introduction.md
gshilin-sdb: done

deploy/deploy.mk
```diff
@@ -18,42 +18,52 @@ cluster: bin/e2e-certs/ca.pem
...
 deploy: cluster build-containers build-cli deploy-helm prepare-agent-config
 else
 deploy: cluster deploy-helm prepare-agent-config
 	@echo "Skipping container and CLI builds (SKIP_BUILD is set)"
 endif
 
-redeploy-api: flightctl-api-container
+redeploy-api: FLAVOR ?= el9
+redeploy-api: build-containers
 	test/scripts/redeploy.sh api
```
avishayt: Will this work?
The redeploy script has "latest" hardcoded

deploy/deploy.mk
```diff
@@ -18,42 +18,52 @@ cluster: bin/e2e-certs/ca.pem
...
+redeploy-telemetry-gateway: FLAVOR ?= el9
+redeploy-telemetry-gateway: build-containers
 	test/scripts/redeploy.sh telemetry-gateway
 
-redeploy-imagebuilder-worker: flightctl-imagebuilder-worker-container
+redeploy-imagebuilder-worker: FLAVOR ?= el9
+redeploy-imagebuilder-worker: build-containers
 	test/scripts/redeploy.sh imagebuilder-worker
 
 redeploy-imagebuilder-api: flightctl-imagebuilder-api-container
```
avishayt, line 58: Why is this container different than the others?

deploy/deploy.mk
gshilin-sdb: fixed

deploy/helm/helm-chart-opts.yaml
```diff
@@ -44,7 +44,77 @@ community:
...
+  annotations:
+    name: FlightControl
+    provider: FlightControl
+    supportURL: https://github.com/flightctl/flightctl
+  images:
+    db:
+      image: quay.io/sclorg/postgresql-16-c10s
+      tag: "20250214"
+    kv:
+      image: quay.io/sclorg/valkey-8-c10s
```
avishayt, line 80: Why valkey?

deploy/deploy.mk
gshilin-sdb: fixed

deploy/deploy.mk
gshilin-sdb: already fixed, not pushed yet

deploy/helm/helm-chart-opts.yaml
gshilin-sdb: that's the new official name starting from RHEL10

avishayt: quadlets is using redis?

{{- else if eq .Flavor "el10" }}
  image: quay.io/sclorg/redis-7-c10s
  tag: "20250108"

PR #2492, opened 2026-02-10
EDM-3291: Add varnames to API versions

celdrake:
An empty Api version was added in https://github.com/flightctl/flightctl/pull/2481/changes#diff-828788da5d0c2a00bd1b5c830702bfa4345441463b60af325f01a6ece9ff9b3eR4172.

As it's unnamed, it generates an invalid `ApiVersion` type in the typescript generator used by the UI.
```
export enum ApiVersion {
  V1BETA1 = 'v1beta1',
  FLIGHTCTL_IO_V1BETA1 = 'flightctl.io/v1beta1',
  = '',
}
```

Adding `x-enum-varnames` to these values fixes the issue. Also adding them to the `imagebuilder` openAPI schema so that UI can reference those values rather than having to hardcode constants.

## Summary by CodeRabbit

* **Chores**
  * Added enum alias metadata to API schemas to provide labeled enum names for downstream code generation and tooling.
  * Refreshed embedded API specification payloads used for runtime documentation and client generation.


api/core/v1beta1/openapi.yaml
```diff
@@ -4170,6 +4170,10 @@ components:
         - v1beta1
         - flightctl.io/v1beta1
         - "" # to provide backward compatibility for older clients that are not setting the apiVersion field
+      x-enum-varnames:
+        - ApiVersionV1beta1
```
celdrake, line 4174: Perhaps the names could be generic rather than reference the actual version?

sk-ilya: Could you please elaborate what do you mean by "more generic"?

celdrake: I meant to name them with a stable name like `shortApiVersion/longApiVersion` so that the references don't change even if the values change.

celdrake: As discussed, since updating to a newer version requires code changes anyway, seems appropriate that the refernces to the versions also change.

PR #2501, opened 2026-02-11
NO-ISSUE:  Fix pruning CI flaky tests

eldar101:
Fixing Failure in CI:

`msg="Failed to update to renderedVersion: 6: before update: prefetch: pulling oci target quay.io/my-org/my-compose-app:v1.0: inspecting manifest for auto-detection: inspect manifest: authentication failed: code: 1: time="2026-02-13T13:00:56Z" level=fatal msg="Error parsing image name \"docker://quay.io/my-org/my-compose-app:v1.0\": reading manifest v1.0 in quay.io/my-org/my-compose-app: unauthorized: access to the requested resource is not authorized"2026-02-13 13:01:28.424450149 +0000 UTC m=+76725.051582546" file="agent/device/device.go:572"`

What I changed in agent_pruning_test.go:

Added explicit device-spec YAML evidence generation with the enrolled Flightctl device ID.
Each YAML uses metadata.name: <deviceID from EnrollAndWaitForOnlineStatus()>.
Kept applying changes via harness update functions

## Summary by CodeRabbit

* **Tests**
  * Updated E2E agent pruning tests to use a compose-based artifact reference flow, add YAML-based device evidence at key steps, wait for update completion, and exercise re-pruning triggers for more reliable validation.

* **Refactor**
  * Simplified artifact handling and introduced test utilities to build inline compose specs and manage image-reference-based assertions.

### DIFF
--- a/test/e2e/agent/agent_pruning_test.go
+++ b/test/e2e/agent/agent_pruning_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/flightctl/flightctl/test/harness/e2e"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"sigs.k8s.io/yaml"
 )
 
 var _ = Describe("Agent image and artifact pruning", func() {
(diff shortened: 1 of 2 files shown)


test/e2e/agent/agent_pruning_test.go
```diff
@@ -798,6 +798,38 @@ func waitForReferencePresence(harness *e2e.Harness, reference string, imageRefMo
 		}))
 }
 
+func resolveAgentPodmanCommand(harness *e2e.Harness, artifactDir string) string {
```
siserafin: Can you please explain this function and why we need it?

eldar101: In In CI, the failing test was creating andlisting the artifact in a different podman context than the agent uses for pruning

siserafin: So I think this is also part of a bug I open...there is an application setting called "runAs". It can be set to root or flightctl. I think that pruning should work in both case, instead it works just in case it is set to root.

eldar101:
@siserafin now with your latest suggestions

test/e2e/agent/agent_pruning_test.go
```diff
@@ -202,33 +203,50 @@ var _ = Describe("Agent image and artifact pruning", func() {
 		})
 
 		It("OCI Artifacts: Unreferenced Artifacts Are Pruned", Label("87089", "agent"), func() {
+			if composeArtifactImageRef == "" {
+				Skip("PRUNING_COMPOSE_ARTIFACT_IMAGE must be set to a pullable OCI artifact/image reference for this test")
```
siserafin: This test will always be skipped, since we are not setting this env variable. Can you try to use quay.io/flightctl-tests/models/gpt2?
This could be an example of application refering it:
```
apiVersion: flightctl.io/v1beta1
kind: Device
metadata:
  name: some_device_name
[...]
spec:
  applications:
    - name: my-inline
      appType: compose
      inline:
        - path: docker-compose.yaml
          content: |
            version: "3.8"
            services:
              service1:
                image: quay.io/flightctl-tests/alpine:v1
                command: ["sleep", "infinity"]
                volumes:
                  - my-data:/data
            volumes:
              my-data:
                external: true
      volumes:
        - name: my-data
          image:
            reference: quay.io/flightctl-tests/models/gpt2
            pullPolicy: IfNotPresent
```

Anyway I think this test is failing for this bug https://issues.redhat.com/browse/EDM-3352 (if the artifact is already in the device should not be pulled)

eldar101: @siserafin do you think I should close this PR for now?

test/e2e/agent/agent_pruning_test.go
```diff
@@ -770,6 +792,42 @@ func buildNginxSpec(containerContent string) (v1beta1.ApplicationProviderSpec, e
 	})
 }
 
+func buildComposeWithImageVolumeSpec(imageRef string) (v1beta1.ApplicationProviderSpec, error) {
```
siserafin: This function should be in harness_application, If we don't have one already

PR #2503, opened 2026-02-11
EDM-2838: Pam issuer UI branding

asafbennatan:
adds branding support for pam issuer login 

## New Features
- Added login page branding customization with support for custom display names, separate light and dark theme logos, custom favicons, and color overrides for brand elements, backgrounds, and text colors.

### DIFF
--- a/Containerfile.pam-issuer
+++ b/Containerfile.pam-issuer
@@ -6,8 +6,12 @@ WORKDIR /app
 # Switch to root for build operations
 USER 0
 
-# Install C development tools needed for CGO
-RUN dnf install -y gcc glibc-devel
+# Install/upgrade C development tools needed for CGO.
+# --nobest: UBI repos may temporarily have inconsistent glibc versions
+# (e.g. new glibc published before glibc-common syncs), causing unresolvable
+# dependency errors. --nobest falls back to the installed versions when the
+# best candidates can't be resolved.
+RUN dnf install -y --nobest gcc glibc-devel
 
 # Install matching PAM and pam-devel from CentOS mirror for build
 RUN dnf install -y https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/pam-1.5.1-24.el9.x86_64.rpm && \


deploy/podman/service-config.yaml
```diff
@@ -52,6 +52,31 @@ global:
...
+          textColorRegular:    # --pf-t--global--text--color--regular
+        # PatternFly v6 semantic token overrides for dark theme
+        darkTheme:
+          logoDataUri:         # Logo data URI for dark mode
+          brandDefault:        # --pf-t--global--color--brand--default
+          brandHover:          # --pf-t--global--color--brand--hover
+          brandClicked:        # --pf-t--global--color--brand--clicked
+          backgroundPrimary:   # --pf-t--global--background--color--primary--default
+          backgroundSecondary: # --pf-t--global--background--color--secondary--default
+          textColorRegular:    # --pf-t--global--text--color--regular
```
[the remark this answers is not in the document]
asafbennatan: we do set defaults in config and other places - the way it is now is the way i want it as it documents the available fields

internal/config/config.go
```diff
@@ -228,6 +228,45 @@ type PAMOIDCIssuer struct {
...
+	// LogoDataUri is a data URI for a custom logo (e.g. "data:image/svg+xml;base64,...").
+	// Use different logos per theme (e.g. dark text on light backgrounds, white text on dark backgrounds).
+	// If not set, the default embedded Flight Control logo is used.
+	LogoDataUri string `json:"logoDataUri,omitempty"`
+	// BrandDefault overrides --pf-t--global--color--brand--default (primary brand color for buttons, links)
+	BrandDefault string `json:"brandDefault,omitempty"`
+	// BrandHover overrides --pf-t--global--color--brand--hover (brand color on hover/focus)
+	BrandHover string `json:"brandHover,omitempty"`
+	// BrandClicked overrides --pf-t--global--color--brand--clicked (brand color on click/active)
+	BrandClicked string `json:"brandClicked,omitempty"`
```
sk-ilya, line 263: My concern is that it leaks the UI styling implementation details into the main service config.

Can we get away with a flag (e.g., "isRHEM") or with drop-in html/css/favicon files that can be different/patched for downstream?

asafbennatan: >My concern is that it leaks the UI styling implementation details 
i would argue that this specific configuration options are quite generic and you would have equivalents for  any css bundle assuming we would replace css/html at some point (which i dont know we will)

per previous discussion we want to avoid mentioning downstream versions , as for using drop in html/css it provides grater customization then intended - not sure if that's desirable , and what are the implications from security standpoint + UX

internal/pam_issuer_server/server.go
```diff
@@ -162,6 +164,16 @@ func (s *Server) Run(ctx context.Context) error {
 	router.Get("/auth/assets/flight-control-logo.svg", handler.ServeFlightControlLogo)
 	router.Get("/auth/assets/login.js", handler.ServeLoginJS)
 	router.Get("/auth/assets/login.css", handler.ServeLoginCSS)
+	router.Get("/auth/assets/favicon.png", handler.ServeFavicon)
```
sk-ilya, line 167: nit: can these also be served with `http.FileServer(http.FS(...))` ?

PR #2508, opened 2026-02-12
EDM-3323: Element exceed the 64 chars limit

noga-magen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Error element paths are now truncated in error messages so overly long paths display a shortened tail, preventing excessively long error strings.
* **Tests**
  * Unit tests updated to expect truncated element paths.
  * End-to-end test flow made delay-tolerant by waiting for the device to report ready/System info before asserting.

### DIFF
--- a/internal/agent/device/errors/structured_test.go
+++ b/internal/agent/device/errors/structured_test.go
@@ -100,8 +100,8 @@ func TestGetElement(t *testing.T) {
 	}{
 		{
 			name:     "extracts element from direct wrap",
-			err:      fmt.Errorf("creating directory %w: %w", WithElement("/var/lib/myapp"), ErrPermissionDenied),
-			expected: "/var/lib/myapp",
+			err:      fmt.Errorf("creating directory %w: %w", WithElement("/var/lib/myapp/making/sure/the/string/is/longer/than/64/characters"), ErrPermissionDenied),
+			expected: "/var/lib/myapp/making/sure/the/string/is/longer/than/64/characters",
 		},
 		{
 			name:     "extracts element from nested chain",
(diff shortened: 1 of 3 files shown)


internal/agent/device/errors/errors.go
```diff
@@ -438,6 +438,10 @@ func (e *Element) Error() string {
 func GetElement(err error) string {
 	var elem *Element
 	if errors.As(err, &elem) {
+		const maxLength = 64
+		if len(elem.Value) > maxLength {
+			return "..." + elem.Value[len(elem.Value)-maxLength:]
+		}
```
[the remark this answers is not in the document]
noga-magen: Yes I aware.

internal/agent/device/errors/structured_test.go
```diff
@@ -100,8 +100,8 @@ func TestGetElement(t *testing.T) {
 	}{
 		{
 			name:     "extracts element from direct wrap",
-			err:      fmt.Errorf("creating directory %w: %w", WithElement("/var/lib/myapp"), ErrPermissionDenied),
-			expected: "/var/lib/myapp",
+			err:      fmt.Errorf("creating directory %w: %w", WithElement("/var/lib/myapp/making/sure/the/string/is/longer/than/64/characters"), ErrPermissionDenied),
```
hexfusion: question: what is the problem we are trying to solve? the whole idea of using element is to add additional context. do you have a working example of the problem?

internal/agent/device/errors/structured_test.go
```diff
@@ -100,8 +100,8 @@ func TestGetElement(t *testing.T) {
...
+			expected: "..." + "ar/lib/myapp/making/sure/the/string/is/longer/than/64/characters",
```
gshilin-sdb: I was sure GetElement truncates the end of string, not the beginning

noga-magen: we start at the index that is the len - max ( like in here 66 - 64 = 2)   until the end (which would be 64 chars later)

eldar101 (approved):

gshilin-sdb (approved):

PR #2510, opened 2026-02-12
EDM-2846: Apply Security HTTP Headers

sk-ilya:
# HTTP Security Headers

| Server | HSTS | X-Content-Type-Options | X-Frame-Options | Content-Security-Policy |
|---|---|---|---|---|
| Main API | yes | yes | yes | `default-src 'none'; frame-ancestors 'none'` |
| Agent Endpoint | yes | yes | yes | `default-src 'none'; frame-ancestors 'none'` |
| ImageBuilder API | yes | yes | yes | `default-src 'none'; frame-ancestors 'none'` |
| PAM Issuer | yes | yes | yes | `default-src 'none'; script-src 'self'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; form-action 'self'; frame-ancestors 'none'; base-uri 'self' `|
| Alertmanager Proxy | yes | yes | yes | none (reverse proxy, upstream controls content) |
| UserInfo Proxy | yes | yes | yes | `default-src 'none'; frame-ancestors 'none'` |
| CLI Artifacts (nginx, TLS) | yes | yes | yes | `default-src 'none'; frame-ancestors 'none'` |
| CLI Artifacts (nginx, no TLS) | no | yes | yes | `default-src 'none'; frame-ancestors 'none'` |

## Summary by CodeRabbit

* **New Features**
  * Global security headers and Content-Security-Policy applied across servers, proxies, and middleware.
  * Externalized login UI assets with client-side script for theming and improved form submission; new endpoints to serve these assets.

* **Tests**
  * Added tests validating security headers and CSP behavior.

* **Chores**
  * Updated server and nginx configurations and adjusted server timeout settings.

### DIFF
--- a/internal/api_server/middleware/middleware.go
+++ b/internal/api_server/middleware/middleware.go
@@ -199,8 +199,8 @@ func extractOrgIDFromRequestCert(ctx context.Context, r *http.Request) (uuid.UUI
 const (
 	// StrictCSP is the strictest CSP for JSON-only API servers.
 	StrictCSP = "default-src 'none'; frame-ancestors 'none'"
-	// PAMIssuerCSP allows the inline scripts/styles and assets needed by the login page.
-	PAMIssuerCSP = "default-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self'; connect-src 'self'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'"
+	// PAMIssuerCSP allows the scripts, styles, and assets needed by the login page.
+	PAMIssuerCSP = "default-src 'none'; script-src 'self'; style-src 'self'; font-src 'self'; img-src 'self'; connect-src 'self'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'"
 )
 
 // SecurityHeaders adds security headers to all HTTP responses.
--- a/internal/auth/oidc/pam/provider.go
+++ b/internal/auth/oidc/pam/provider.go
@@ -44,6 +44,12 @@ var PatternFlyCSS string
 //go:embed templates/flight-control-logo.svg
 var FlightControlLogo string
 
+//go:embed templates/login.js
+var LoginJS string
+
+//go:embed templates/login.css
+var LoginCSS string
+
 // LoginFormData represents the data used to populate the login form template
 // Currently empty as all authorization parameters are stored in encrypted cookie
 type LoginFormData struct {
--- a/internal/auth/oidc/pam/templates/login.css
+++ b/internal/auth/oidc/pam/templates/login.css
@@ -1,0 +1,34 @@
+html, body {
+    height: 100%;
+    margin: 0;
+    padding: 0;
+}
+#login-page {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    min-height: 100vh;
+    background-color: var(--pf-t--global--background--color--secondary--default);
+}
+#login-page > .pf-v6-l-stack {
+    align-items: center;
+}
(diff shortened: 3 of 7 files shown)


cmd/flightctl-alertmanager-proxy/main.go
```diff
@@ -295,6 +295,7 @@ func main() {
 	router := chi.NewRouter()
 	router.Use(chimiddleware.RequestSize(int64(cfg.Service.HttpMaxRequestSize)))
 	router.Use(middleware.RequestSizeLimiter(cfg.Service.HttpMaxUrlLength, cfg.Service.HttpMaxNumHeaders))
+	router.Use(middleware.SecurityHeaders)
```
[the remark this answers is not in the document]
sk-ilya: It's a proxy so the server behind it should decide and set the CSP

internal/api_server/middleware/middleware.go
```diff
@@ -195,6 +195,14 @@ func extractOrgIDFromRequestCert(ctx context.Context, r *http.Request) (uuid.UUI
 	return orgID, true, nil
 }
 
+// Content-Security-Policy constants for different server types.
+const (
+	// StrictCSP is the strictest CSP for JSON-only API servers.
+	StrictCSP = "default-src 'none'; frame-ancestors 'none'"
+	// PAMIssuerCSP allows the inline scripts/styles and assets needed by the login page.
+	PAMIssuerCSP = "default-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self'; connect-src 'self'; form-action 'self'; frame-ancestors 'none'; base-uri 'self'"
```
asafbennatan: @sk-ilya not sure about this , i think self is too restrictive :
we redirect back to caller - (ui/CLI)
according to what i read assuming pam is on pam.xxx.com (or even xxx.com:8444) this means different origin aka not 'self' aka we wil be blocked from returning the user back to the main ui / cli

sk-ilya: But it's a redirect, so it shouldn't be affected by the policies?

asafbennatan: why not ?

sk-ilya: If my understanding is correct, since nothing is _fetched_ from this different origin, then it should be OK.

asafbennatan: im not a CSP expert but from what i read this still an issue due to :
`form-action: self
`
see [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/form-action)

i think the critical part is : 

`Warning: Whether form-action should block redirects after a form submission is [debated](https://github.com/w3c/webappsec-csp/issues/8) and browser implementations of this aspect are inconsistent (e.g., Firefox 57 doesn't block the redirects whereas Chrome 63 does).`

due to this since we do not want to block redirect we should set the appropriate value instead of self or remove it (to avoid any inconsistencies) 

note - form-action  applies since the original action is a POST ( POST /login ( the xhr called by the html returned by the original /authorize) -> /authorize -> redirect back to caller )

according to what i read chrome considers this redirection chain as part of the original form-action

sk-ilya: The form's native submission is prevented via `event.preventDefault()`, and the request is made with `fetch()`. It seems that `form-action` is not even relevant here

asafbennatan: i have just tested this locally and you are right , the reason is that login returns 200 and the js calls authorize.
had the login itself redirected to /authorize it wouldn't work

PR #2512, opened 2026-02-12
EDM-2966: Capital Letters in hostname cause service startup failure

galel12:
## Problem

On RHEL systems with uppercase letters in the hostname (e.g. `rhel9-GA`), starting Flight Control services via `systemctl start flightctl.target` fails. The `flightctl-db-wait.service` exits with a validation error, causing all dependent services to stall:
```bash
global.baseDomain: Invalid value: "rhel9-GA": Invalid pattern
(regex used for validation is '(a-z0-9?\\.)a-z?')
```
This occurs when `baseDomain` is explicitly set with uppercase characters in `service-config.yaml`. The case where `baseDomain` is left blank (defaulting to hostname) was already normalized in `getHostnameFQDN()`.

## Root Cause

The `baseDomain` value is validated against a lowercase-only DNS regex, but normalization to lowercase was only applied in some code paths (e.g. when defaulting from hostname via `hostname -f`). When a user explicitly sets an uppercase `baseDomain` in `service-config.yaml`, it passes through to validation unnormalized and is rejected.

DNS hostnames are case-insensitive per RFC 1123, so uppercase letters are valid but need to be canonicalized to lowercase before validation and use in downstream contexts (certificate SANs, service URLs, TLS verification).

## Solution

- Added `NormalizeStandaloneConfig()` to canonicalize `baseDomain` to lowercase before validation, covering both explicit and hostname-defaulted values.
- Updated bash init scripts (`init_certs.sh`, `init.sh`) to normalize `baseDomain` unconditionally before the regex check, not only when defaulting from hostname.
- Improved error messages to include remediation hints (`hostnamectl set-hostname` or setting `baseDomain` in config).
- Added a "Prerequisites" section to the Linux installation docs documenting hostname requirements.
- Added unit tests for the normalization behavior and uppercase validation contract.

### DIFF
--- a/cmd/flightctl-standalone/render_template.go
+++ b/cmd/flightctl-standalone/render_template.go
@@ -91,6 +91,15 @@ func (o *RenderTemplateOptions) completeConfig(data map[string]interface{}) erro
 		fmt.Fprintf(os.Stderr, "global.baseDomain not set, defaulting to system hostname FQDN (%s)\n", hostname)
 	}
 
+	// Normalize baseDomain to lowercase (DNS is case-insensitive per RFC 1123).
+	// This covers both explicit values from service-config.yaml and hostname defaults.
+	baseDomain, _ = global["baseDomain"].(string)
+	normalized := strings.ToLower(baseDomain)
+	if normalized != baseDomain {
+		global["baseDomain"] = normalized
+		fmt.Fprintf(os.Stderr, "global.baseDomain normalized to lowercase: %s -> %s\n", baseDomain, normalized)
+	}
+
 	// Default UI forwarded-header settings.
 	ui, ok := global["ui"].(map[string]interface{})
 	if !ok {
@@ -162,6 +171,7 @@ func (o *RenderTemplateOptions) validateConfig(data map[string]interface{}) erro
 		return fmt.Errorf("failed to unmarshal config for validation: %w", err)
 	}
 
+	validation.NormalizeStandaloneConfig(&config)
 	if errs := validation.ValidateStandaloneConfig(&config); len(errs) > 0 {
 		errMsgs := make([]string, len(errs))
 		for i, err := range errs {
--- a/deploy/scripts/init_certs.sh
+++ b/deploy/scripts/init_certs.sh
@@ -37,14 +37,19 @@ host_ips+=("127.0.0.1")
 # Validate the base domain from config, or default to hostname FQDN
 base_domain=$(python3 "$YAML_HELPER" extract .global.baseDomain "$CONFIG_FILE")
 if [[ -z "$base_domain" ]]; then
-    # Normalize to lowercase (DNS is case-insensitive per RFC 1123)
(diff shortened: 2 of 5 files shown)


galel12:
@coderabbitai review

asafbennatan (changes requested):
@galel12 i think we did not want to allow upper case in hostname as we are following k8 , we have already fixed that automatically generated baseDomain (when empty)
takes lowercase hostname:

https://github.com/flightctl/flightctl/pull/2458

if you think we should do this please verify with @fzdarsky as he had the opinion (and i agree) that we should reject uppercase basedomains that were used explicitly

galel12 (commented):
Thanks for the review @asafbennatan . I see your point and I'm aware of #2458 , that fix covers the blank/defaulted case.

The question is what to do when a user explicitly sets `baseDomain: rhel9-GA` in `service-config.yaml`. @hexfusion raised this in the [Jira discussion](https://issues.redhat.com/browse/EDM-2966):

> "in the context of kubernetes the regex probably makes sense as they have a lowercase requirement that is not part of the RFC itself. but in this context we are in RHEL and failing on upper case feels like it will cause friction."

He also pointed out the severity: "a user could fail to install the product because of hostname."

Since this is standalone RHEL (not K8s), I think normalizing with a log message is a better UX than rejecting a technically valid RFC 1123 hostname. The user sees:
 `global.baseDomain normalized to lowercase: rhel9-GA -> rhel9-ga` 
 in the logs, so it's not silent, but it doesn't block their install.

That said, Asaf mentioned you had an opinion on this @fzdarsky. 
Would appreciate your input on whether we should normalize or reject explicit uppercase `baseDomains` in the standalone deployment context.

fzdarsky:
> The question is what to do when a user explicitly sets `baseDomain: rhel9-GA` in `service-config.yaml`. @hexfusion raised this in the [Jira discussion](https://issues.redhat.com/browse/EDM-2966): [...] 
> He also pointed out the severity: "a user could fail to install the product because of hostname."

I would argue the main issue here is that we do not have proper input/schema validation on the `service-config.yaml` that'd allow us to fail fast and with clear errors. Something we should address irrespective of the issue this PR wants to address.

> Since this is standalone RHEL (not K8s), I think normalizing with a log message is a better UX than rejecting a technically valid RFC 1123 hostname. The user sees: `global.baseDomain normalized to lowercase: rhel9-GA -> rhel9-ga` in the logs, so it's not silent, but it doesn't block their install.

It's not only about K8s's conventions, the problem is that Linux tools are not consistent in how they do case sensitive or insensitive comparisons, whether they filter or normalize, etc. - and I'm not 100% sure it's consistent even within FC. Because that can lead to subtle bugs / unexpected errors, there's the guideline that hostnames should always be lower case (https://www.freedesktop.org/software/systemd/man/latest/hostname.html).

Without input validation, I think normalizing and logging is a pragmatic approach.

asafbennatan (approved):
LGTM

PR #2513, opened 2026-02-12
Fix RBAC tests

itzikb-redhat:
Make sure we get the default admin k8s context
Before doing any flightctl action set the current organization to the one under test

## Summary by CodeRabbit

* **Tests**
  * Stricter detection of an admin Kubernetes context for end-to-end RBAC tests — now requires a context clearly identified as both default and admin to avoid ambiguous targets.
  * When switching namespaces during tests, the suite now resolves the corresponding organization and switches the test client to that organization automatically.
  * Improved diagnostics and clearer error messages when admin context or organization matching fails.

### DIFF
--- a/test/harness/e2e/harness.go
+++ b/test/harness/e2e/harness.go
@@ -2044,22 +2044,28 @@ func (h *Harness) CreateResource(resourceType string) (string, string, []byte, e
 	}
 }
 
-// GetDefaultK8sContext returns the a K8s context with default in its name
-func (h *Harness) GetDefaultK8sContext() (string, error) {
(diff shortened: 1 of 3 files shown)


test/harness/e2e/harness.go
```diff
@@ -1111,6 +1133,29 @@ func (h *Harness) GetOrganizationID() (string, error) {
 	return *name, nil
 }
 
+// GetOrganizationIDForNamespace returns the organization ID whose Spec.ExternalId matches the given namespace (e.g. OpenShift project).
+// If none match, returns an error so callers can fall back to GetOrganizationID() if desired.
+func (h *Harness) GetOrganizationIDForNamespace(namespace string) (string, error) {
```
siserafin: Can we move these organizations functions to harness_rbac or create harness_organizations file? I think that Samuel as well will use it

PR #2520, opened 2026-02-15
EDM-3313: Fix to CI failure in test case [83029]

eldar101:
Why it suddenly started failing:

The test is old, but dependencies changed recently , especially update/status behavior in service/harness code.

After recent backend/status changes, devices can recover from pull errors but still spend longer in transient states where that exact boolean stays false, causing your 600s timeout.
So the test became flaky without the test body itself being heavily modified.

Why this fix solves it:
The scenario’s real success criterion is: after network is restored, the device reaches the expected new rendered version.
WaitForDeviceNewRenderedVersion is already designed for current update semantics and handles transient states more robustly.
This makes the test validate recovery behavior directly, instead of depending on a brittle status boolean at one specific moment.

### DIFF
--- a/test/e2e/cli/cli_console_test.go
+++ b/test/e2e/cli/cli_console_test.go
@@ -2,6 +2,7 @@ package cli_test
 
 import (
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/flightctl/flightctl/api/core/v1beta1"
@@ -163,9 +164,12 @@ var _ = Describe("CLI - device console", func() {
 		// Get harness directly - no shared package-level variable
 		harness := e2e.GetWorkerHarness()
 
+		expectedRenderedVersion, err := harness.PrepareNextDeviceVersion(deviceID)
+		Expect(err).ToNot(HaveOccurred())
+
 		GinkgoWriter.Printf("Simulating network failure\n")
 		DeferCleanup(func() { _ = harness.FixNetworkFailure() })
-		err := harness.SimulateNetworkFailure()
+		err = harness.SimulateNetworkFailure()
 		Expect(err).ToNot(HaveOccurred())
 
 		_, _, err = harness.WaitForBootstrapAndUpdateToVersion(deviceID, util.DeviceTags.V4)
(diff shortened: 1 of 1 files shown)


test/e2e/cli/cli_console_test.go
```diff
@@ -196,10 +200,19 @@ var _ = Describe("CLI - device console", func() {
 		err = harness.FixNetworkFailure()
 		Expect(err).ToNot(HaveOccurred())
 
-		GinkgoWriter.Printf("Waiting for the device to finish updating\n")
+		GinkgoWriter.Printf("Waiting for the device to finish updating (renderedVersion >= %d)\n", expectedRenderedVersion)
```
siserafin, line 203: The expected RenderedVersion should be a specific #, so here should be just "="

test/e2e/cli/cli_console_test.go
```diff
@@ -2,6 +2,7 @@ package cli_test
 
 import (
```
siserafin, line 3: The ticket EDM-3313 is open for acm-d profile...is this test failing in main?

test/e2e/cli/cli_console_test.go
```diff
@@ -196,10 +200,19 @@ var _ = Describe("CLI - device console", func() {
 		err = harness.FixNetworkFailure()
 		Expect(err).ToNot(HaveOccurred())
 
-		GinkgoWriter.Printf("Waiting for the device to finish updating\n")
+		GinkgoWriter.Printf("Waiting for the device to finish updating (renderedVersion >= %d)\n", expectedRenderedVersion)
 		util.EventuallySlow(resources.GetJSONByName[*v1beta1.Device]).
 			WithArguments(harness, resources.Devices, deviceID).
-			Should(WithTransform((*v1beta1.Device).IsUpdatedToDeviceSpec, BeTrue()))
+			Should(WithTransform(func(device *v1beta1.Device) bool {
```
siserafin, line 206: Here you are getting the version information from the device status config instead of from the device's annotations. 
Re-using harness functions you could get the same result with:
`device, err := harness.GetDevice(deviceId)
Expect(device.Status.Config.RenderedVersion).To(Equal(strconv.Itoa(expectedVersion)))`

The two versions (in annotations and device status) should be the same, I am not sure this will improve the test stability. 

If this test fails in main I think we should clarify with devs if it is a bug, or what is the expected behavior.

test/e2e/cli/cli_console_test.go
eldar101: no, only acm-d.

PR #2521, opened 2026-02-15
EDM-3310: Fix to CI failure in test case [77672]

eldar101:
Fixed flaky CI failure  by replacing manual per-step update polling with harness-managed device spec updates.
The test now applie config via PrepareNextDeviceVersion + UpdateDeviceConfigWithRetries, matching automation behavior and using the enrolled Flightctl-assigned device ID flow.
Final validation was simplified to assert the rendered version reaches initial + applied updates, removing fragile transition timing checks.

## Summary by CodeRabbit

* **Tests**
  * Improved device configuration update testing by implementing progressive batching for spec application, enhancing test reliability and efficiency. Updated test assertions to validate final rendered versions more accurately, replacing incremental verification steps with consolidated endpoint checks.

### DIFF
--- a/test/e2e/agent/agent_update_test.go
+++ b/test/e2e/agent/agent_update_test.go
@@ -201,30 +201,28 @@ var _ = Describe("VM Agent behavior during updates", func() {
 				},
 			}
 
-			// Apply each spec in quick succession, just waiting for the device to register that it
-			// has acknowledged it should update
-			currentVersion := initialVersion
+			// Build the device spec progressively and update through harness device-spec helpers,
+			// matching automation behavior.
+			deviceSpecConfig := make([]v1beta1.ConfigProviderSpec, 0, len(configFactories))
 			for i, factory := range configFactories {
 				specVersion := i + 1
-				By(fmt.Sprintf("Applying spec: %d", specVersion))
+				By(fmt.Sprintf("Applying spec update: %d", specVersion))
 				var configProviderSpec v1beta1.ConfigProviderSpec
 				err := factory(&configProviderSpec)
 				Expect(err).ToNot(HaveOccurred())
-				err = harness.AddConfigToDeviceWithRetries(deviceId, configProviderSpec)
+				deviceSpecConfig = append(deviceSpecConfig, configProviderSpec)
+
+				nextRenderedVersion, err := harness.PrepareNextDeviceVersion(deviceId)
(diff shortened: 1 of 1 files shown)


test/e2e/agent/agent_update_test.go
```diff
@@ -201,30 +201,28 @@ var _ = Describe("VM Agent behavior during updates", func() {
...
+			deviceSpecConfig := make([]v1beta1.ConfigProviderSpec, 0, len(configFactories))
 			for i, factory := range configFactories {
 				specVersion := i + 1
-				By(fmt.Sprintf("Applying spec: %d", specVersion))
+				By(fmt.Sprintf("Applying spec update: %d", specVersion))
 				var configProviderSpec v1beta1.ConfigProviderSpec
 				err := factory(&configProviderSpec)
 				Expect(err).ToNot(HaveOccurred())
-				err = harness.AddConfigToDeviceWithRetries(deviceId, configProviderSpec)
+				deviceSpecConfig = append(deviceSpecConfig, configProviderSpec)
```
siserafin, line 213: I think the function is doing the same steps

eldar101:
@siserafin I'm removing reportportal links from here (Per Itzik) but you can check 
edge-management-flightctl-acm #310
Yes, the test is flaky there too.

test/e2e/agent/agent_update_test.go
eldar101: New flow (PrepareNextDeviceVersion + UpdateDeviceConfigWithRetries): updates spec and blocks until that exact rendered version is reached.

siserafin: Yes, but also  AddConfigToDeviceWithRetries and WaitForDeviceContents. Anyway this solution re-uses harness functions, which is better.

siserafin: We might consider to increase the harness TIMEOUT from 5 to 7 minutes, since many tests fails during device reboot or in updating status.

siserafin: Actually Noga created a reboot aware function that could be used here https://github.com/flightctl/flightctl/pull/2532 instead of increasing the timeout

eldar101: Thanks

siserafin (approved):

PR #2535, opened 2026-02-16
EDM-3346: Fix OS rollback issues in agent

kkyrazis:
## Summary by CodeRabbit

* **Bug Fixes**
  * Rollback failures now record and surface both failed version and reason; status reporting and upgrade-failed messages improved.
  * OS rollback and post-update activation are more resilient, avoiding premature failures and aggregating component errors when appropriate.

* **Refactor**
  * Application install/update/remove monitoring centralized into staged monitor orchestration to reduce duplicated logic and improve reliability.
  * Device sync flow reorganized for clearer rollback-aware cleanup and error aggregation.

* **Tests**
  * Expanded tests to cover rollback reporting and dependency/error-handling scenarios.

### DIFF
--- a/internal/agent/device/applications/controller.go
+++ b/internal/agent/device/applications/controller.go
@@ -89,7 +89,11 @@ func syncProviders(
 		log.Debugf("Removing application: %s", p.Name())
 		if err := manager.Remove(ctx, p); err != nil {
 			log.Warnf("Failed to remove application %s: %v", p.Name(), err)
-			errs = append(errs, fmt.Errorf("removing: %w: %w", errors.WithElement(p.Name()), err))
+			// Missing dependencies is acceptable for removed applications. If the spec defines removing a helm application
+			// at the same time that it changes OSs to one that does not have helm app support, then we accept it without failure
+			if !errors.Is(err, errors.ErrAppDependency) {
+				errs = append(errs, fmt.Errorf("removing: %w: %w", errors.WithElement(p.Name()), err))
+			}
 		}
 	}
 
--- a/internal/agent/device/applications/controller_test.go
+++ b/internal/agent/device/applications/controller_test.go
@@ -391,6 +391,62 @@ func TestSyncProviders_ContinueOnError(t *testing.T) {
 			},
 			expectedErrors: 2,
 		},
+		{
+			name: "remove with ErrAppDependency is ignored",
+			currentProvs: func(ctrl *gomock.Controller) []provider.Provider {
+				return []provider.Provider{
+					newMockProviderForSync(ctrl, "helm-app", "Helm App"),
+				}
+			},
+			desiredProvs: func(ctrl *gomock.Controller) []provider.Provider { return nil },
+			setupMocks: func(mockManager *MockManager) {
+				mockManager.EXPECT().Remove(gomock.Any(), gomock.Any()).Return(errors.ErrAppDependency)
+			},
+			expectedErrors: 0,
+		},
+		{
+			name: "remove with wrapped ErrAppDependency is ignored",
+			currentProvs: func(ctrl *gomock.Controller) []provider.Provider {
+				return []provider.Provider{
+					newMockProviderForSync(ctrl, "helm-app", "Helm App"),
+				}
+			},
+			desiredProvs: func(ctrl *gomock.Controller) []provider.Provider { return nil },
+			setupMocks: func(mockManager *MockManager) {
+				mockManager.EXPECT().Remove(gomock.Any(), gomock.Any()).Return(fmt.Errorf("remove failed: %w", errors.ErrAppDependency))
+			},
+			expectedErrors: 0,
+		},
+		{
+			name:         "ensure with ErrAppDependency still fails",
+			currentProvs: func(ctrl *gomock.Controller) []provider.Provider { return nil },
+			desiredProvs: func(ctrl *gomock.Controller) []provider.Provider {
+				return []provider.Provider{
+					newMockProviderForSync(ctrl, "helm-app", "Helm App"),
+				}
+			},
+			setupMocks: func(mockManager *MockManager) {
+				mockManager.EXPECT().Ensure(gomock.Any(), gomock.Any()).Return(errors.ErrAppDependency)
+			},
+			expectedErrors: 1,
+		},
+		{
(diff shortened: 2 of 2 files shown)


internal/agent/device/applications/manager.go
```diff
@@ -77,81 +77,82 @@ func (m *manager) validateProviderDeps(ctx context.Context, p provider.Provider)
 	return nil
 }
 
-func (m *manager) Ensure(ctx context.Context, provider provider.Provider) error {
-	if err := m.validateProviderDeps(ctx, provider); err != nil {
-		return err
+func (m *manager) Ensure(ctx context.Context, p provider.Provider) error {
+	if m.isMonitored(p) {
+		return nil
 	}
```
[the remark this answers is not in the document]
kkyrazis: A non retryable error triggers rollback. Rollback performs the opposite action, which in this case, is cleanup

internal/agent/device/spec/spec.go
```diff
@@ -194,3 +198,22 @@ func newVersionedDevice(version string) *v1beta1.Device {
...
+func IsOSRollback(current, desired *v1beta1.Device) bool {
+	if !IsRollback(current, desired) {
+		return false
+	}
+	if current.Spec == nil || desired.Spec == nil {
+		return false
+	}
+	if current.Spec.Os == nil || desired.Spec.Os == nil {
+		return false
+	}
```
keitwb:
```suggestion
	if !HasOSImage(current.Spec) || !HasOSImage(desired.Spec) {
		return false
	}
```

internal/agent/device/bootstrap.go
```diff
@@ -175,13 +175,20 @@ func (b *Bootstrap) updateStatus(ctx context.Context) {
 		Type: v1beta1.ConditionTypeDeviceUpdating,
 	}
 
-	if b.specManager.IsUpgrading() {
+	switch {
```
keitwb: Why the change to `switch`?

internal/agent/device/device.go
```diff
@@ -422,30 +444,49 @@ func (a *Agent) syncDevice(ctx context.Context, current, desired *v1beta1.Device
...
+	var errs []error
+
+	handle := func(errType, err error) error {
+		if err == nil {
+			return nil
+		}
+		wrapped := fmt.Errorf("%w: %w", errType, err)
+		if cfg.isRollbackCleanup {
+			errs = append(errs, fmt.Errorf("%w: %w", errors.ErrRollbackCleanup, wrapped))
+			return nil
```
keitwb, line 442: Is this the fix for the rollback not working properly -- because the rollback was erroring out early and not fully completing? What was the rollback error in the case of the missing helm command upgrade failure?

internal/agent/device/device.go
```diff
@@ -409,7 +414,24 @@ func (a *Agent) beforeUpdate(ctx context.Context, current, desired *v1beta1.Devi
...
+
+type syncDeviceOpt func(*syncDeviceConfig)
+
+func withRollbackCleanup(enabled bool) syncDeviceOpt {
+	return func(c *syncDeviceConfig) {
+		c.isRollbackCleanup = enabled
+	}
+}
+
+func (a *Agent) syncDevice(ctx context.Context, current, desired *v1beta1.Device, opts ...syncDeviceOpt) error {
```
keitwb: Adding func options to a method that has one call site and is not a public API seems like overkill. Based on how different the error handling is below, it seems like this needs to be two separate methods (`syncDevice` and `rollbackDevice`, with perhaps a helper method) that clearly shows the difference in error handling.

internal/agent/device/device.go
kkyrazis: There were 3 problems that I found.
1. `specManager.IsOsUpdate` never returned true on a rollback. The way the calls work, on rollback, resetting desired to current would happen first so this always returned false.
2. some operations during the sync could fail on rollback. Mostly because we don't differentiate between a rollback and a real remove or whatever.
3. The "failedVersions" is only stored in memory. When rebooting into the previous OS, we'd immediately attempt to upgrade again.

internal/agent/device/device.go
kkyrazis: I cleaned up the overengineered portion, and experimenting with separating sync and rollback sync, but ultimately didn't love having the duplicated methods that differ simply from how they handle errors. Anything that kept the two methods in sync, felt just as overengineered.

internal/agent/device/applications/manager.go
```diff
@@ -77,81 +77,82 @@ func (m *manager) validateProviderDeps(ctx context.Context, p provider.Provider)
...
-		return nil
+func (m *manager) Update(ctx context.Context, p provider.Provider) error {
+	if err := p.Remove(ctx); err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrRemovingApplication, err)
+	}
+	if err := p.Install(ctx); err != nil {
+		return fmt.Errorf("%w: %w", errors.ErrInstallingApplication, err)
 	}
+	return m.queueMonitorUpdate(ctx, p)
+}
```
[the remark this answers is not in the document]
kkyrazis: No, the caller should supply the errors.WithElement(p.Name()) if it's relevant to distinguish what occured. There is no ambiguity as to which provider failed with a call to Update or Install or Remove.

internal/agent/device/applications/manager.go
```diff
@@ -77,81 +77,82 @@ func (m *manager) validateProviderDeps(ctx context.Context, p provider.Provider)
...
+	switch p.Spec().AppType {
+	case v1beta1.AppTypeCompose, v1beta1.AppTypeQuadlet, v1beta1.AppTypeContainer:
+		return m.podmanMonitor.QueueRemove(NewApplication(p))
+	case v1beta1.AppTypeHelm:
+		return m.kubernetesMonitor.Remove(NewHelmApplication(p))
 	default:
-		return fmt.Errorf("%w: %s", errors.ErrUnsupportedAppType, appType)
+		return fmt.Errorf("%w: %s", errors.ErrUnsupportedAppType, p.Spec().AppType)
 	}
 }
```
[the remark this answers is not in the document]
kkyrazis: You can't have a monitor entry if the dependencies are missing.

PR #2537, opened 2026-02-16
EDM-3365: Skip quadlet actions during system shutdown drain

kkyrazis:
When a system shutdown is pending, systemd calls (list-dependencies, .etc) report errors. Since shutting down will stop the applications anyway, we skip those actions and just kill any monitors.

## Release Notes

* **New Features**
  * Improved system shutdown and drain operations with enhanced action management for better stability and controlled execution flow.
  * Refined action execution logic with improved state and monitor lifecycle management.

* **Tests**
  * Added comprehensive test coverage for system drain and shutdown operations.

### DIFF
--- a/internal/agent/device/applications/podman_monitor.go
+++ b/internal/agent/device/applications/podman_monitor.go
@@ -166,7 +166,7 @@ func (m *PodmanMonitor) drain(ctx context.Context) error {
 		}
 	}
 
-	if err := m.ExecuteActions(ctx); err != nil {
+	if err := m.executeActions(ctx, true); err != nil {
 		errs = append(errs, err)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/agent/device/applications/podman_monitor.go
```diff
@@ -285,14 +285,25 @@ func normalizeActionAppType(appType v1beta1.AppType) v1beta1.AppType {
...
+func (m *PodmanMonitor) executeActions(ctx context.Context, systemShutdown bool) error {
 	ctx = m.addBatchTimeToCtx(ctx)
 	actions := m.drainActions()
 
 	groupedActions := make(map[v1beta1.AppType][]lifecycle.Action)
 	for i := range actions {
 		action := actions[i]
 		appType := normalizeActionAppType(action.AppType)
+
+		if systemShutdown && appType == v1beta1.AppTypeQuadlet {
```
keitwb, line 302: Does the same thing apply to `AppTypeContainer` or not?

kkyrazis: normalizeActionAppType converts `container` to `quadlet` since they have the same lifecycle type and needed to ensure that daemon reload is called at the right time.

PR #2543, opened 2026-02-17
EDM-1057: Agent shows Online but stops updating after spec files are deleted

galel12:
## Problem

When `/var/lib/flightctl` is deleted while the agent is running, the device gets stuck. It stays `Online` but stops applying updates. The reconciliation loop silently fails on every tick with `ErrMissingRenderedSpec` from `GetDesired()` or `Read(Current)`, logs the error, and returns indefinitely. Only an agent restart recovers the device.

## Root Cause

`syncDeviceSpec()` in `device.go` treats `ErrMissingRenderedSpec` as a generic error: it logs and returns without any recovery action. The existing `Ensure()` method in the spec manager already knows how to detect and recreate missing spec files (with proper logging and auditing), but it is only called during bootstrap. Never during the runtime reconciliation loop.

## Solution

When a spec file goes missing at runtime, the agent pushes an error status to the server and exits non-zero, letting systemd restart it.
On restart, `Ensure()` detects the missing file(s) and resets ALL spec files to version "0" - the agent does not try to reason about partial state. The publisher then fetches the latest spec from the server and
reconciles from scratch.

Key changes:
- `handleMissingSpec()` in `syncDeviceSpec()` pushes `DeviceSummaryStatusError` and calls `log.Fatalf()`
- `Ensure()` resets all three spec files when any single file is missing (not just the missing one)
- `ResetVersion()` on the publisher aligns the in-memory version after the on-disk reset so the next poll fetches the latest spec
- Audit trail entries use `reason: "recovery"` (or `"bootstrap"` on first startup)

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -151,6 +151,9 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 
 	desired, requeue, err := a.specManager.GetDesired(ctx)
 	if err != nil {
+		if a.recoverMissingSpec(err) {
+			return
+		}
 		a.log.Errorf("Failed to get desired spec: %v", err)
 		return
 	}
@@ -163,6 +166,9 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 
 	current, err := a.specManager.Read(spec.Current)
 	if err != nil {
+		if a.recoverMissingSpec(err) {
+			return
+		}
 		a.log.Errorf("Failed to get current spec: %v", err)
 		return
 	}
@@ -255,6 +261,20 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 			// Don't return error - pruning failures must not block reconciliation
 		}
 	}
+}
+
+// recoverMissingSpec checks if the error is due to a missing spec file and
+// attempts to recover by calling Ensure(). Returns true if recovery was
+// attempted, signaling the caller to return and let the next tick proceed.
+func (a *Agent) recoverMissingSpec(err error) bool {
+	if !errors.Is(err, errors.ErrMissingRenderedSpec) {
+		return false
+	}
+	a.log.Warnf("Spec file missing, recovering: %v", err)
+	if ensureErr := a.specManager.Ensure(); ensureErr != nil {
+		a.log.Errorf("Failed to recover spec files: %v", ensureErr)
+	}
+	return true
 }
 
 func (a *Agent) rollbackDevice(ctx context.Context, current, desired *v1beta1.Device, syncErr error, syncFn func(context.Context, *v1beta1.Device, *v1beta1.Device) error) error {
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -151,7 +151,7 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 
 	desired, requeue, err := a.specManager.GetDesired(ctx)
 	if err != nil {
-		if a.recoverMissingSpec(err) {
+		if a.handleMissingSpec(ctx, err) {
 			return
 		}
 		a.log.Errorf("Failed to get desired spec: %v", err)
@@ -166,7 +166,7 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 
 	current, err := a.specManager.Read(spec.Current)
 	if err != nil {
-		if a.recoverMissingSpec(err) {
+		if a.handleMissingSpec(ctx, err) {
 			return
 		}
 		a.log.Errorf("Failed to get current spec: %v", err)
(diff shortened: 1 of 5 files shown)


keitwb:
It would be simpler to just treat the missing spec files as a fatal unexpected error and shutdown the agent with a non-zero exit code. Systemd ensures the agent will be restarted very quickly. Upon restart, the disk files would be created but with the full startup routine, which may be more comprehensive and robust in handling the reset state than simply calling `Ensure` (with all of the cached state I'm not really confident `Ensure` by itself doesn't have unknown edge cases). A restart means there is a single path for doing the reset in such cases vs two.

kkyrazis:
> It would be simpler to just treat the missing spec files as a fatal unexpected error and shutdown the agent with a non-zero exit code. Systemd ensures the agent will be restarted very quickly. Upon restart, the disk files would be created but with the full startup routine, which may be more comprehensive and robust in handling the reset state than simply calling `Ensure` (with all of the cached state I'm not really confident `Ensure` by itself doesn't have unknown edge cases). A restart means there is a single path for doing the reset in such cases vs two.

I know that Ensure doesn't do it now, but a benefit of doing it without restarting is that we could potentially handle recovering from our `lastConsumedDevice` rather than always starting with the empty spec.

keitwb:
> > It would be simpler to just treat the missing spec files as a fatal unexpected error and shutdown the agent with a non-zero exit code. Systemd ensures the agent will be restarted very quickly. Upon restart, the disk files would be created but with the full startup routine, which may be more comprehensive and robust in handling the reset state than simply calling `Ensure` (with all of the cached state I'm not really confident `Ensure` by itself doesn't have unknown edge cases). A restart means there is a single path for doing the reset in such cases vs two.
> 
> I know that Ensure doesn't do it now, but a benefit of doing it without restarting is that we could potentially handle recovering from our `lastConsumedDevice` rather than always starting with the empty spec.

Another way to deal with this is to just keep the file open and reuse the file handler within the spec manager so that even if it is deleted, the specManager can always read it since Linux won't remove a file until the last handler is closed after an unlink call (which is what `rm` does). The files would get recreated upon an update without any errors, or reset as expected upon agent restart if never recreated.

hexfusion:
the agent must be able to recover from this type of observable condition. while in general I believe it makes sense to try and continue operation during error. in this case the error is extremely unlikely and unexpected. because of the nature of the issue extreme caution should be taken. saying that I agree with @keitwb that the simplest resolution would be to restart the process. but we must also consider the below. 

- if the filesystem is full we must ensure that the signal received from an I/O operation does not mimic the condition used to restart. (should be part of e2e)
- the agent should not try to reason about the correct current of desired state. by design the agent should recover all specs to 0 and then recover be reconciling the latest version.
- proper status is critical if we are to exit non zero and reset/recreate some/all of the spec files to empty this should be oberservible in the status. 
- if one file is missing we should reset all to zero

internal/agent/device/device.go
```diff
@@ -257,6 +263,26 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
...
+	a.log.Errorf("%s", msg)
+	_, updateErr := a.statusManager.Update(ctx, status.SetDeviceSummary(v1beta1.DeviceSummaryStatus{
+		Status: v1beta1.DeviceSummaryStatusError,
+		Info:   lo.ToPtr(msg),
+	}))
+	if updateErr != nil {
+		a.log.Warnf("Failed to push error status before exit: %v", updateErr)
+	}
+	a.log.Fatalf("Exiting due to missing spec file; systemd will restart the agent")
+	return true
```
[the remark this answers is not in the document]
keitwb: It should not go through normal shutdown handling because it is an undefined state and we want to minimize the risk of device spec corruption and quit as fast as possible.

internal/agent/device/device.go
```diff
@@ -241,20 +247,40 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
...
+	msg := fmt.Sprintf("Spec file missing at runtime: %v", err)
+	a.log.Errorf("%s", msg)
+	_, updateErr := a.statusManager.Update(ctx, status.SetDeviceSummary(v1beta1.DeviceSummaryStatus{
+		Status: v1beta1.DeviceSummaryStatusError,
+		Info:   lo.ToPtr(msg),
+	}))
+	if updateErr != nil {
+		a.log.Warnf("Failed to push error status before exit: %v", updateErr)
 	}
+	a.log.Fatalf("Exiting due to missing spec file; systemd will restart the agent")
```
keitwb, line 282:
```suggestion
	a.log.Fatalf("Exiting due to missing spec file")
```

The agent doesn't need to say or know anything about how it is run.

internal/agent/device/spec/manager.go
asafbennatan, line 127: can  we consolidate this into a single loop where the output is allmissing and anymissing ?

asafbennatan (approved):
see nit comment

asafbennatan (approved):
LGTM

PR #2555, opened 2026-02-18
EDM-3075: Fleet creation allows 253-char name but fleet immediately becomes Invalid due to internal name derivation

galel12:
## Problem

Creating a Fleet with a 253-character name (the DNS subdomain maximum) succeeds at the API, but the Fleet immediately transitions to **Invalid** status with an error saying `metadata.name` exceeds 253 characters, even though the Fleet name itself is within the limit.

## Root Cause

A background task derives `TemplateVersion` names from the Fleet name:

```go
templateVersionName = fmt.Sprintf("%s-%d", fleetName, generation)
```

1. Fleet with 253-char name passes `Validate()` (standard DNS subdomain check)
2. Background task generates TemplateVersion name: `{253 chars}-1` = 255 chars
3. `TemplateVersion.Validate()` rejects it - exceeds 253-char DNS subdomain limit
4. Fleet is marked **Invalid** asynchronously, with a confusing error

The Fleet's `Validate()` only enforced the generic 253-char resource name limit, without accounting for the `-{generation}` suffix appended during TemplateVersion derivation.

## Solution

The fleet name is already a primary key - TemplateVersions are always scoped to a fleet, so embedding the fleet name in the TemplateVersion name is redundant. The fix simplifies name derivation to:

```go
func generateTemplateVersionName(fleet *domain.Fleet) string {    
        return "v" + strconv.FormatInt(*fleet.Metadata.Generation, 10)}
```

This can never exceed the DNS limit (max 20 chars), requires no truncation or hashing logic, and keeps Fleet name validation unchanged at 253 chars.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures generated template version names are DNS-compliant: uses the short form when it fits and a truncated+hash fallback plus generation suffix when needed to avoid overly long names.

* **Tests**
  * Added comprehensive tests validating name-generation behavior across short and very long names, large generation values, uniqueness, and stability.

### DIFF
--- a/internal/tasks/fleet_validate.go
+++ b/internal/tasks/fleet_validate.go
@@ -2,8 +2,6 @@ package tasks
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -12,7 +10,6 @@ import (
 	"github.com/flightctl/flightctl/internal/domain"
 	"github.com/flightctl/flightctl/internal/service"
 	"github.com/flightctl/flightctl/internal/util"
-	"github.com/flightctl/flightctl/internal/util/validation"
 	"github.com/flightctl/flightctl/pkg/k8sclient"
 	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
@@ -263,27 +260,6 @@ func (t *FleetValidateLogic) validateHttpProviderConfig(ctx context.Context, con
 	return &httpConfigProviderSpec.Name, &httpConfigProviderSpec.HttpRef.Repository, nil
 }
 
-const (
-	hashLen          = 8
-	maxGenerationLen = 19 // max decimal digits in int64
-)
-
-// generateTemplateVersionName returns {name}-{gen} for short fleet names, or
-// {truncatedName}-{hash}-{gen} for long ones to stay within DNS1123MaxLength.
-// Reserves 20 chars for the generation suffix so the format is stable.
 func generateTemplateVersionName(fleet *domain.Fleet) string {
-	name := *fleet.Metadata.Name
-	genStr := strconv.FormatInt(*fleet.Metadata.Generation, 10)
-
-	if len(name)+1+maxGenerationLen <= validation.DNS1123MaxLength {
-		return name + "-" + genStr
-	}
-
-	hash := sha256.Sum256([]byte(name))
-	hashStr := hex.EncodeToString(hash[:])[:hashLen]
-
-	maxPrefix := validation.DNS1123MaxLength - 1 - hashLen - 1 - maxGenerationLen
-	prefix := strings.TrimRight(name[:maxPrefix], "-.")
-
-	return prefix + "-" + hashStr + "-" + genStr
+	return "v" + strconv.FormatInt(*fleet.Metadata.Generation, 10)
 }
--- a/internal/tasks/fleet_validate_test.go
+++ b/internal/tasks/fleet_validate_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/flightctl/flightctl/internal/domain"
 	"github.com/flightctl/flightctl/internal/service"
-	"github.com/flightctl/flightctl/internal/util/validation"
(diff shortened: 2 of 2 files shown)


avishayt:
The current fix (capping fleet names at 233 chars) works but has tradeoffs. Here are the options as I see them:

**Option 1 (current PR – cap fleet name at 233)**
This is a breaking change: fleet names that were previously valid (234–253 chars under the generic DNS subdomain limit) would become invalid. Anyone relying on the documented 253-char resource name limit would be affected.

**Option 2 (allow TemplateVersion names up to 273)**
That would break the DNS subdomain limit (253) used everywhere else for resource names. TemplateVersion would be the only resource with a different, higher limit, which is inconsistent and goes against the existing convention.

**Option 3 (derive a name that always fits, without changing fleet validation)**
Keep fleet name validation at 253 chars. When deriving the TemplateVersion name:
If len(fleetName) + 1 + len(str(generation)) ≤ 253, use the simple form:
<fleet-name>-<generation> (current behavior).
Otherwise, use a long-name form that stays within 253 and remains unique:
<fleet-name-prefix>-<fleet-name-hash>-<generation>
where the prefix is the start of the fleet name and its length is 253 − 20 − len(hash) (reserving space for - and up to 19-digit generation). The hash is over the full fleet name (e.g. first 8 hex chars of SHA-256) so that two names that only differ after the prefix still get different TemplateVersion names.
No change to fleet name limits, TemplateVersion names stay within 253 chars, and the long-name form avoids collisions when truncating the fleet name alone would not be unique.

**I recommend option 3**

galel12:
Thanks for the detailed breakdown, @avishayt . I wanted to share a thought on Option 3 before implementing.

On the breaking change concern with Option 1: The 234-253 range isn't really valid today. As the bug describes, fleet names in that range are accepted by the API but immediately fail when the TemplateVersion is derived - the fleet goes Invalid with a confusing error. So in practice, there are no working fleets with names over 233 chars. Capping at 233 doesn't break anything that was actually working - it just moves the error from an async confusing failure to a clear upfront rejection.

On Option 3: the TemplateVersion name would silently diverge from what the user chose as their fleet name. Unlike an internal detail (e.g. hashing a password), the derived name is user-visible - they'll see it in `flightctl get templateversion`, in fleet annotations, etc. Accepting a name and then producing a different derived name feels like it could be a confusing experience.

Option 1 is more restrictive, but it's also more transparent - the user knows upfront that their name is too long, they adjust it, and everything downstream stays predictable.

That said, I see your point about the breaking change for names in the 234-253 range. Happy to go with Option 3 if you think backward compatibility outweighs the UX concern here - just wanted to raise it.

avishayt:
> Thanks for the detailed breakdown, @avishayt . I wanted to share a thought on Option 3 before implementing.
> 
> On the breaking change concern with Option 1: The 234-253 range isn't really valid today. As the bug describes, fleet names in that range are accepted by the API but immediately fail when the TemplateVersion is derived - the fleet goes Invalid with a confusing error. So in practice, there are no working fleets with names over 233 chars. Capping at 233 doesn't break anything that was actually working - it just moves the error from an async confusing failure to a clear upfront rejection.

Understood, but I'd rather have uniform validations for resource names.
 
> On Option 3: the TemplateVersion name would silently diverge from what the user chose as their fleet name. Unlike an internal detail (e.g. hashing a password), the derived name is user-visible - they'll see it in `flightctl get templateversion`, in fleet annotations, etc. Accepting a name and then producing a different derived name feels like it could be a confusing experience.

Sure, but we're not guaranteeing any name.  Actually, since template versions are always listed in the context of a fleet, the fleet name prefix isn't even necessary.

> Option 1 is more restrictive, but it's also more transparent - the user knows upfront that their name is too long, they adjust it, and everything downstream stays predictable.
> 
> That said, I see your point about the breaking change for names in the 234-253 range. Happy to go with Option 3 if you think backward compatibility outweighs the UX concern here - just wanted to raise it.

I think option 3 has the minimum impact.

galel12:
@coderabbitai full review

avishayt (approved):

PR #2556, opened 2026-02-18
EDM-2258: Defer updates on critical CPU/Memory alerts

kkyrazis:
Error messages now look like:
```yaml
updated:                                                                                                                                           
      info: '[2026-02-18 17:45:06] While Preparing: resources failed for CPU: resource limit exceeded (CPU, memory, or disk)'                          
      status: Updating  
 ```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updates are now deferred when critical CPU or memory resource alerts are active, preventing deployments during high system stress.
  * Critical resource alert errors are now treated as retryable so deferred updates are requeued appropriately.
  * Resource-exhausted messaging clarified to indicate CPU, memory, or disk limits.

* **Tests**
  * Added test cases covering deferred updates when CPU or memory critical alerts are firing.

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -199,8 +199,9 @@ func (a *Agent) syncDeviceSpec(ctx context.Context) {
 			return
 		}
 
-		// Policy may defer update; in all cases, warn and roll back to the previous renderedVersion.
-		if errors.Is(syncErr, errors.ErrUpdatePolicyNotReady) || errors.Is(syncErr, errors.ErrDownloadPolicyNotReady) {
+		// Policy or critical resource alerts may defer update; in all cases, warn and roll back to the previous renderedVersion.
+		if errors.Is(syncErr, errors.ErrUpdatePolicyNotReady) || errors.Is(syncErr, errors.ErrDownloadPolicyNotReady) ||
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: internal/agent/device/device.go [hunk 1]
```diff
@@ -331,6 +332,15 @@ func (a *Agent) beforeUpdate(ctx context.Context, current, desired *v1beta1.Devi
...
 
+	if a.resourceManager.IsCriticalAlert(resource.CPUMonitorType) {
+		return fmt.Errorf("%w: %w", errors.ErrComponentResources,
+			fmt.Errorf("%w: %w", errors.WithElement("CPU"), errors.ErrCriticalResourceAlert))
+	}
+	if a.resourceManager.IsCriticalAlert(resource.MemoryMonitorType) {
+		return fmt.Errorf("%w: %w", errors.ErrComponentResources,
+			fmt.Errorf("%w: %w", errors.WithElement("Memory"), errors.ErrCriticalResourceAlert))
+	}
+
```
keitwb, line 343: Do disk errors actually block upgrades? I don't see it in this function unless it is checked somewhere else.

kkyrazis: It exists here: https://github.com/flightctl/flightctl/blob/main/internal/agent/device/dependency/dependency.go#L343

PR #2567, opened 2026-02-19
EDM-3379: Port aap creation script to go, reference correct vars

DakCrowder:
Previously the bash script used to make a request to create an AAP oauth application was referencing old variables (oAuthToken and oAuthApplicationClientId) which were removed before GA.  This causes issues as the way the services ingest the client id for example has diverged from downstream docs which tell users to set a now deprecated oAuthToken var in service-config.yaml.

The goals here are to:
1.  Get the script working with the clientId and new token var specified in service-config.yaml
2. Since we're changing the script, update it to go from the nasty bash request logic.  Part of this allows us to make the api init process a .service that just calls the go binary

## Summary by CodeRabbit

* **New Features**
  * New CLI command to create AAP OAuth applications and persist client IDs.
  * Service config gains an AAP token field and CA cert option.

* **Chores**
  * Replaced init scripts with a systemd oneshot service to run OAuth creation at startup.
  * Added TLS/CA handling, hostname FQDN discovery, and idempotent OAuth provisioning.

* **Tests**
  * Added unit tests for AAP client and OAuth application provisioning.

* **Documentation**
  * Updated AAP authentication guide with automated and manual setup flows.

### DIFF
--- a/internal/standalone/aap_application.go
+++ b/internal/standalone/aap_application.go
@@ -117,6 +117,10 @@ func createOAuthApplication(ctx context.Context, client OAuthApplicationCreator,
 		return "", err
 	}
 
+	if response == nil {
+		return "", fmt.Errorf("AAP returned empty response")
+	}
+
 	if response.ClientID == "" {
 		return "", fmt.Errorf("AAP returned empty client_id")
 	}
--- a/.spelling
+++ b/.spelling
@@ -311,6 +311,8 @@ Datadog
 CRDs
 wildcard
 Istio
+delimited
+URIs
 subdomain
  - docs/user/installing/configuring-imagebuilder.md
  entitlement
--- a/docs/user/installing/configuring-auth/auth-aap.md
+++ b/docs/user/installing/configuring-auth/auth-aap.md
@@ -44,28 +44,157 @@ Flight Control uses role-based authorization with organization-scoped access:
 
 ## Configuration
 
-### Static Configuration
+### Prerequisites
 
-AAP authentication is configured via configuration files.
+Before configuring AAP authentication, ensure you have:
 
-Configuration values need to be set in `config.yaml`:
+1. AAP Gateway Access: Administrative credentials and API access to our AAP gateway instance.  Flight Control must have network access to your AAP Gateway
+2. AAP Permissions: You must have an admin user with OAuth application creation permissions in AAP Gateway
+
+### TLS Certs
+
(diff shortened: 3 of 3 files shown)


internal/standalone/aap_application.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := writeClientIDToFile(clientID, opts.OutputFile); err != nil {
+		return fmt.Errorf("failed to write client_id to %s: %w", opts.OutputFile, err)
+	}
+
+	opts.Logger.Infof("AAP OAuth client_id saved to %s", opts.OutputFile)
+	return nil
+}
+
+func buildOAuthApplicationRequest(baseDomain string, appName string, organization int) *aap.AAPOAuthApplicationRequest {
+	redirectURIs := fmt.Sprintf("https://%s:443/callback http://127.0.0.1/callback", baseDomain)
```
asafbennatan: consider using net/url instead of sprintf

internal/standalone/aap_application.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type OAuthApplicationCreator interface {
+	CreateOAuthApplication(ctx context.Context, token string, req *aap.AAPOAuthApplicationRequest) (*aap.AAPOAuthApplicationResponse, error)
+}
+
+type CreateAAPClientOptions struct {
+	AAPConfig       *standaloneconfig.AAPConfig
+	BaseDomain      string
```
kkyrazis: potentially minor, but I'm not seeing `BaseDomain` being used in the create client flow. Is this a miss, or is it truly not needed.

DakCrowder: Not needed, was left over accidentally from refactoring - will remove

internal/standalone/aap_application.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return &aap.AAPOAuthApplicationRequest{
+		Name:                   appName,
+		Organization:           organization,
+		AuthorizationGrantType: "authorization-code",
+		ClientType:             "public",
+		RedirectURIs:           redirectURIs,
+		AppURL:                 appURL.String(),
+	}
+}
```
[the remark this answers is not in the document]
DakCrowder: Per this RFC https://datatracker.ietf.org/doc/html/rfc8252#section-7.3 - the loopback address has a special case to use randomly assigned ports and should be handled by the authorization serer.  AAP uses django oauth toolkit under the hood which mentions loopback addresses can be configured without explicit port configuration - https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#allowed-redirect-uri-schemes

kkyrazis (approved):

PR #2574, opened 2026-02-23
EDM-2859: Use only FIPS crypto for SSH when on FIPS machine

fzdarsky:
This PR addresses the issue that `golang.org/x/crypto/ssh` is not FIPS-aware and therefore defaults to non-FIPS-compliant crypto even when running on a machine in FIPS mode. This prevents accessing Git repositories using SSH on these machines.

The PR adds the following changes:

- Detects whether the service is running on a FIPS-enabled host using multiple methods for robustness (in `fips.go`).

- Adds a function that returns the SSH crypto settings to be used based on this detection or on user policies (in `ssh_config.go`).

- Wraps the Git Auth settings with the returned SSH crypto settings (in `git_helpers.go`).

- Allows users to configure whether to auto-detect, force on, or force off FIPS crypto as well as to override the used crypto to further restrict it (in `config.go`). Note this isn't exposed through the Helm chart / `service-config.yaml` yet, so it just an escape hatch at the moment.

- The rest of the file changes is just wiring through the configuration.

## Summary by CodeRabbit

* **New Features**
  * Added FIPS mode support with global and SSH-specific cryptographic policy configuration.
  * Added SSH algorithm, cipher, and key exchange configuration for enhanced control over cryptographic settings.
  * Added FIPS mode detection capabilities with intelligent caching.
  * Added integration tests validating SSH operations under FIPS and non-FIPS environments.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -33,9 +33,52 @@ type Config struct {
 	CA                  *ca.Config                 `json:"ca,omitempty"`
 	Tracing             *tracingConfig             `json:"tracing,omitempty"`
 	GitOps              *gitOpsConfig              `json:"gitOps,omitempty"`
+	CryptoPolicy        *CryptoPolicyConfig        `json:"cryptoPolicy,omitempty"`
 	Periodic            *periodicConfig            `json:"periodic,omitempty"`
 	Organizations       *organizationsConfig       `json:"organizations,omitempty"`
 	TelemetryGateway    *telemetryGatewayConfig    `json:"telemetrygateway,omitempty"`
+}
+
+// CryptoPolicyConfig contains cryptographic policy configuration for all protocols.
+type CryptoPolicyConfig struct {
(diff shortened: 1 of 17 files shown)


internal/tasks/ssh_config.go
```diff
@@ -0,0 +1,161 @@
+package tasks
```
hexfusion: just a general observation/question should this be a more global ssh/crypto.Config vs tasks specific?

fzdarsky: You're referring to the choice/location of package, right? I think you're right. Which location would make more sense, e.g. `internal/ssh/`?

fzdarsky: Then again, it's currently not used by anything but the git_helpers.go in `internal/tasks/`, so it could also make sense to keep it where it is until we have a more global need.

hexfusion: no strong feelings

fzdarsky: [x] done

PR #2633, opened 2026-03-02
Add Appliance Containerfile for flightctl

aekubacki:
Upstream containerfile for appliance install of CentOS + flightctl

## New Features

* **New Appliance Container Image**: Introduced a pre-configured container image for appliance deployment with GNOME GUI support, enabling graphical desktop functionality.
* **Pre-installed Components**: Image includes Flightctl for system management and Podman container utilities for containerized application support, ready for immediate use.

### DIFF
--- a/packaging/images/Containerfile.appliance
+++ b/packaging/images/Containerfile.appliance
@@ -1,0 +1,26 @@
+FROM quay.io/centos-bootc/centos-bootc:stream9
+
+# 1. Install Flightctl and system updates
+RUN dnf config-manager --add-repo https://rpm.flightctl.io/flightctl-epel.repo && \
+    dnf -y install flightctl-services && \
+    dnf -y clean all && \
+    systemctl enable flightctl.target && \
+    systemctl mask bootc-fetch-apply-updates.timer
+
+# 2. Install EPEL and Podman utilities
+RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
+    dnf -y install podman-compose && \
+    dnf -y clean all && \
+    systemctl enable podman.service
+    
+ # 3. Install the GUI environment (GNOME)
+# This installs the full desktop environment and windowing system
+RUN dnf -y groupinstall "Server with GUI" && \
+    dnf clean all
+
+# 4. Set the boot target to graphical and enable the display manager
+RUN systemctl set-default graphical.target && \
+    systemctl enable gdm
+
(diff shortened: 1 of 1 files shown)


keitwb:
@avishayt @fzdarsky What are your thoughts on this and whether/where it belongs in the repo? It is a convenience for users to quickly spin up the backend services as an appliance ISO (which would need to be built by Github actions and stuck somewhere).

packaging/images/Containerfile.appliance
```diff
@@ -0,0 +1,26 @@
+FROM quay.io/centos-bootc/centos-bootc:stream9
+
+# 1. Install Flightctl and system updates
+RUN dnf config-manager --add-repo https://rpm.flightctl.io/flightctl-epel.repo && \
+    dnf -y install flightctl-services && \
```
fzdarsky, line 5: Shall we also by default install the `flightctl-observability` stack for a more integrated experience?

packaging/images/Containerfile.appliance
```diff
@@ -0,0 +1,26 @@
...
+# 2. Install EPEL and Podman utilities
+RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
+    dnf -y install podman-compose && \
+    dnf -y clean all && \
+    systemctl enable podman.service
+    
+ # 3. Install the GUI environment (GNOME)
+# This installs the full desktop environment and windowing system
+RUN dnf -y groupinstall "Server with GUI" && \
+    dnf clean all
```
fzdarsky, line 19: Do we expect users to access RHEM through the host's console or a web browser? If we do want them to use the console (or use it, too), then let's consider configuring RHEL in "kiosk mode" (see @avishayt's demo).

packaging/images/Containerfile.appliance
```diff
@@ -0,0 +1,26 @@
...
+
+# 4. Set the boot target to graphical and enable the display manager
+RUN systemctl set-default graphical.target && \
+    systemctl enable gdm
+
+# Set root password - DEMO PURPOSES ONLY
+RUN echo root:redhat123 | chpasswd
```
fzdarsky, line 26: User accounts on the host should be exclusively assigned through `cloud-init` or `ignition` (both of which would need to be installed).

Can be a separate PR, but... can we configure the host for `sssd` and set up the PAM issuer to use the host's user DB? This way, the user provisioned through `cloud-init` would also be the default RHEM super admin.

PR #2685, opened 2026-03-14
NO-ISSUE: Preventing helm chart install failure when exposeServiceMethod cannot be auto detected

rprakashg:
When installing helm chart on xK8 clusters (EKS for ex.) where Gateway controllers are not installed chart fails to install when it cannot auto detect exposeServiceMethod.  Expected behavior would be chart should install all the flightctl microservices and Ingress can be setup manually. For ex. on EKS would like to leverage the Ingress via AWS ALB

## Summary by CodeRabbit

* **Bug Fixes**
  * Service exposure auto-detection now falls back to a "none" exposure mode when required resources are not found, preventing deployment aborts and allowing setup to continue gracefully, improving resilience of deployment flows.


fzdarsky:
It's possible to set the `exposeServiceMethod: none` to skip detection and make the deployment succeed. So the question is about the appropriate default behavior.

The thinking behind failing was to be explicit, so the installer does not report success when the result is not in a working state.

Maybe we should at least print a warning (in NOTES.txt)?

PR #2701, opened 2026-03-16
EDM-3491: fixed oidc url normalization

asafbennatan:
replaces string concat with proper url building when creating derived urls from issuer

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -12,6 +12,7 @@ import (
 	"github.com/flightctl/flightctl/internal/auth/authn"
 	"github.com/flightctl/flightctl/internal/auth/authz"
 	"github.com/flightctl/flightctl/internal/auth/common"
+	authprovider "github.com/flightctl/flightctl/internal/auth/provider"
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/consts"
 	"github.com/flightctl/flightctl/pkg/k8sclient"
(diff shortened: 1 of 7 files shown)


internal/auth/auth.go
```diff
@@ -223,7 +224,12 @@ func InitMultiAuth(cfg *config.Config, log logrus.FieldLogger,
 			}
 
 			// Add OpenShift auth with issuer:clientId key
-			openshiftIssuer := strings.TrimSuffix(*cfg.Auth.OpenShift.AuthorizationUrl, "/")
+			openshiftIssuer := *cfg.Auth.OpenShift.AuthorizationUrl
+			if normalized, err := authprovider.NormalizeIssuerURL(openshiftIssuer); err == nil {
+				openshiftIssuer = normalized
+			} else {
+				openshiftIssuer = strings.TrimSuffix(openshiftIssuer, "/")
```
DakCrowder: Is this is not returning the err from .NormalizeIssuerURL so initialization call sites (like the services) won't hard fail on trying to start up if the normalize call fails?  If we fall back to the trim from an error with normalize might be a good idea to log a warning

asafbennatan: fixed

internal/auth/provider/introspection.go
```diff
@@ -109,3 +110,35 @@ func buildIntrospectionURL(baseURL string) string {
...
+	}
+	parsed, err := url.Parse(issuer)
+	if err != nil {
+		return "", fmt.Errorf("invalid issuer URL: %w", err)
+	}
+	if parsed.Scheme == "" || parsed.Host == "" {
+		return "", errors.New("issuer URL must have scheme and host")
+	}
+	parsed.Path = strings.TrimSuffix(parsed.Path, "/")
+	return parsed.String(), nil
```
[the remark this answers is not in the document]
asafbennatan: this is unchanged from previous behaviour

gshilin-sdb (approved):

PR #2722, opened 2026-03-20
EDM-5889: ImageBuilder worker fails on Kubernetes when nested podman build cannot raise RLIMIT_NOFILE

lucamaf:
## Summary by CodeRabbit

* **New Features**
  * Worker containers now automatically detect and apply file descriptor limits from the host system during initialization, with graceful fallback to default settings if detection fails.

### DIFF
--- a/internal/imagebuilder_worker/tasks/imagebuild.go
+++ b/internal/imagebuilder_worker/tasks/imagebuild.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/flightctl/flightctl/internal/config"
@@ -702,6 +703,18 @@ ignore_chown_errors = "true"
 	if c.cfg.ImageBuilderWorker.EffectivePodmanSkipTLSVerify() {
 		startArgs = append(startArgs, "--tls-verify=false")
 	}
+
+	// Pass the current RLIMIT_NOFILE to the worker container so nested
+	// podman builds don't attempt to raise the limit beyond what Kubernetes
+	// permits for this pod, which would cause "operation not permitted".
+	var rLimit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit); err == nil {
+		startArgs = append(startArgs, "--ulimit", fmt.Sprintf("nofile=%d:%d", rLimit.Cur, rLimit.Max))
+		log.Debugf("Passing ulimit nofile=%d:%d to worker container", rLimit.Cur, rLimit.Max)
+	} else {
+		log.WithError(err).Warn("Could not read RLIMIT_NOFILE; worker container will use default ulimits")
+	}
+
 	startArgs = append(startArgs,
 		"--cap-add=SYS_ADMIN",
 		podmanImage,


eldar101 (approved):

galel12 (dismissed):
#### 1. Update user docs

`docs/user/installing/configuring-imagebuilder.md` (~lines 773–774) still says:

> The ImageBuilder Worker sets `--ulimit nofile=1048576:1048576` on the nested podman worker container

That was never true on `main`, and this PR makes it dynamically detected. The troubleshooting section should describe the new behavior and when node-level `MachineConfig` / systemd tuning is still needed if buildah insists on `1048576` internally.

#### 2. Commit message prefix

Per `CONTRIBUTING.md`, squash-merge should use `EDM-XXXX:` or `NO-ISSUE:`.

#### 3. Tests (optional)

`startPodmanWorker` is untested. Extracting ulimit arg construction into a small helper would make a table-driven unit test proportionate; mocking the syscall is probably not worth it for this size change.

galel12 (commented):
see below

internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
```
galel12, line 15: Should we use `golang.org/x/sys/unix` here instead of `syscall`? It is already a direct dep and matches usage elsewhere in the repo (e.g. `internal/agent/device/console/manager.go`). (chai-bot raised the same point on the thread — skip this if you would rather keep discussion there.)

```suggestion
	"golang.org/x/sys/unix"
```

internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -702,6 +703,18 @@ ignore_chown_errors = "true"
 	if c.cfg.ImageBuilderWorker.EffectivePodmanSkipTLSVerify() {
 		startArgs = append(startArgs, "--tls-verify=false")
 	}
+
+	// Pass the current RLIMIT_NOFILE to the worker container so nested
+	// podman builds don't attempt to raise the limit beyond what Kubernetes
+	// permits for this pod, which would cause "operation not permitted".
+	var rLimit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit); err == nil {
+		startArgs = append(startArgs, "--ulimit", fmt.Sprintf("nofile=%d:%d", rLimit.Cur, rLimit.Max))
```
galel12, line 712: When `rLimit.Max` is `RLIM_INFINITY`, should we cap the hard limit to `rLimit.Cur` before passing it to podman? On bare-metal/quadlet hosts that can differ from the typical K8s case where soft and hard are equal.

```suggestion
			max := rLimit.Max
			if max == unix.RLIM_INFINITY {
				max = rLimit.Cur
			}
			startArgs = append(startArgs, "--ulimit", fmt.Sprintf("nofile=%d:%d", rLimit.Cur, max))
```

galel12:
Opened [EDM-5889](https://redhat.atlassian.net/browse/EDM-5889) to track this.

@lucamaf — could you update the commit message to use the `EDM-5889:` prefix per CONTRIBUTING.md? For example:

`EDM-5889: pass current rlimit_nofile to imagebuilder-worker container`

galel12 (changes requested):
Withdrawing my earlier approval — please address the open review comments before merge:

- `golang.org/x/sys/unix` instead of `syscall`
- `RLIM_INFINITY` guard on `--ulimit`
- Update `configuring-imagebuilder.md` troubleshooting section
- Commit message: `EDM-5889: …` prefix

lucamaf:
moving discussion here https://github.com/flightctl/flightctl/pull/3547 since a lot has changed since my last PR

PR #2767, opened 2026-03-27
NO-ISSUE: Automatic refresh of base images

fzdarsky:
## Summary by CodeRabbit

* **Chores**
  * Added a scheduled (daily) and manual workflow to detect and prepare updates for EL9/EL10 base container images.
  * Produces a human-readable summary of detected tag changes and affected categories/versions.
  * Automates multi-architecture image builds and multi-arch manifest publishing when updates are found.
  * Opens or updates a PR with only the changed base image references; skips committing if nothing changed.
  * Made base-image build tooling configurable by target release version and more robust with improved quoting and cleanup.

### DIFF
--- a/.github/workflows/refresh-base-images.yaml
+++ b/.github/workflows/refresh-base-images.yaml
@@ -1,0 +1,315 @@
+name: Refresh Base Images
+
+on:
+  schedule:
+    # Run daily at 03:00 UTC (staggered from RPM builder at 02:00)
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: refresh-base-images
+  cancel-in-progress: false
+
+env:
+  IMAGE_REGISTRY: quay.io/flightctl
+  UBI_REGISTRY: registry.access.redhat.com
+
+jobs:
+  detect-updates:
+    runs-on: ubuntu-24.04
+    outputs:
+      build_matrix: ${{ steps.check.outputs.build_matrix }}
+      manifest_matrix: ${{ steps.check.outputs.manifest_matrix }}
+      updates_json: ${{ steps.check.outputs.updates_json }}
+      any_updated: ${{ steps.check.outputs.any_updated }}
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/refresh-base-images.yaml
```diff
@@ -0,0 +1,305 @@
...
+          IMAGE_REPO: ${{ env.IMAGE_REGISTRY }}/flightctl-base-el${{ matrix.relver }}
+        run: |
+          sudo ./hack/build_flightctl-base.sh
+
+      - name: Push arch-specific image
+        env:
+          IMAGE_TAG: ${{ matrix.micro_tag }}
+          IMAGE_REPO: ${{ env.IMAGE_REGISTRY }}/flightctl-base-el${{ matrix.relver }}
+        run: |
+          sudo buildah login -u "${{ secrets.QUAY_FLIGHTCTL_INFRA_ROBOT_USERNAME }}" \
```
sk-ilya: Why not `redhat-actions/push-to-registry`? 
Same for the "Create and push manifest list" below

.github/workflows/refresh-base-images.yaml
```diff
@@ -0,0 +1,305 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y skopeo jq
```
sk-ilya: Both packages come pre-installed, no need

fzdarsky: [x] removed

.github/workflows/refresh-base-images.yaml
fzdarsky: The action runs as regular user, so wouldn't see our image that's in root's container store. We need to run buildah as root because of `buildah mount`.

For the manifest list we could use it, but TBH I have a bit of a distaste for GH actions that are trivial and do what can be done with 1-2 commands more explicitly.

PR #2810, opened 2026-04-08
NO-ISSUE: Consolidate Service Setup

keitwb:
This just takes the common setup code that all the services use and puts it into a common function that gets called.

This also brings a few services into more alignment with the rest.

## Summary by CodeRabbit

* **Refactor**
  * Centralized service startup across commands: unified context, config, logger, and a single coordinated shutdown handler; startup/shutdown lifecycle simplified.

* **Chores**
  * Standardized logging initialization and levels via a typed log-level configuration with safe defaults and simplified logger setup.

### DIFF
--- a/cmd/flightctl-imagebuilder-api/main.go
+++ b/cmd/flightctl-imagebuilder-api/main.go
@@ -3,41 +3,22 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
-	"os/signal"
-	"syscall"
 
-	"github.com/flightctl/flightctl/internal/config"
+	"github.com/flightctl/flightctl/internal/cmdsetup"
 	imagebuilderapi "github.com/flightctl/flightctl/internal/imagebuilder_api"
 	imagebuilderstore "github.com/flightctl/flightctl/internal/imagebuilder_api/store"
-	"github.com/flightctl/flightctl/internal/instrumentation/tracing"
 	"github.com/flightctl/flightctl/internal/kvstore"
 	"github.com/flightctl/flightctl/internal/store"
 	"github.com/flightctl/flightctl/internal/util"
-	"github.com/flightctl/flightctl/pkg/log"
 	"github.com/flightctl/flightctl/pkg/queues"
 	"github.com/google/uuid"
 )
 
 func main() {
-	ctx := context.Background()
+	ctx, cfg, log, shutdown := cmdsetup.InitService(context.Background(), "imagebuilder-api")
+	defer shutdown()
 
-	cfg, err := config.LoadOrGenerate(config.ConfigFile())
-	if err != nil {
-		log.InitLogs().Fatalf("reading configuration: %v", err)
-	}
-
-	log := log.InitLogs(cfg.ImageBuilderService.LogLevel)
-	log.Println("Starting ImageBuilder API service")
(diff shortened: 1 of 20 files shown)


cmd/flightctl-imagebuilder-api/main.go
```diff
@@ -3,41 +3,20 @@ package main
...
 
 func main() {
-	ctx := context.Background()
-
-	cfg, err := config.LoadOrGenerate(config.ConfigFile())
-	if err != nil {
-		log.InitLogs().Fatalf("reading configuration: %v", err)
-	}
-
-	log := log.InitLogs(cfg.ImageBuilderService.LogLevel)
```
kkyrazis, line 30: I see a couple of services use specific service levels rather than the general Service.LogLevel. Is this behavior that we want to remove?

keitwb: No, I reimplemented those in the service-specific main code with a few refactors of the logging setup as well.

internal/cmdsetup/common.go
```diff
@@ -0,0 +1,45 @@
...
+	logger.Printf("Initializing %s", serviceName)
+	logger.Printf("Using config: %s", cfg)
+
+	tracerShutdown := tracing.InitTracer(logger, cfg, serviceName)
+
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT)
+
+	ctx = context.WithValue(ctx, consts.EventSourceComponentCtxKey, serviceName)
+	ctx = context.WithValue(ctx, consts.EventActorCtxKey, fmt.Sprintf("service:%s", serviceName))
+	ctx = context.WithValue(ctx, consts.InternalRequestCtxKey, true)
```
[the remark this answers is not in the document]
kkyrazis: Evaluate if this is actually a concern or not. Does this context propagate to anywhere it shouldn't? Will api requests inherit this context?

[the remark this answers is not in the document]
keitwb: Yeah I don't really trust that just because it doesn't flow to external API request handling that it is going to result in completely consistent logic so I'm just going to make it match the same behavior as before.

test/harness/e2e/harness.go
```diff
@@ -368,7 +368,7 @@ func (h *Harness) PrintAgentLogsIfFailed() {
 // Registered via DeferCleanup in SetTestContext so it runs for every test.
 func (h *Harness) checkLogsForPanicAVC(sinceTime time.Time) {
 	var findings []string
-	sinceArg := sinceTime.UTC().Format(time.RFC3339)
+	sinceArg := sinceTime.UTC().Format(time.DateTime)
```
keitwb, line 371: This is apparently the only [acceptable date string format to journalctl.](https://www.freedesktop.org/software/systemd/man/latest/journalctl.html?__goaway_challenge=meta-refresh&__goaway_id=81bf51b3d8f062a07fc879a0a36aeb9c&__goaway_referer=https%3A%2F%2Fwww.google.com%2F#-S)

keitwb:
@asafbennatan @asafbss can you all take a look? This is just cleaning up some redundant service init code.

PR #2817, opened 2026-04-09
EDM-3626: Agent stuck in "updating" forever when greenboot is not installed but boot-complete.target exists

galel12:
## Problem

On devices where greenboot is not installed but `boot-complete.target` exists as a systemd unit, the flightctl-agent gets permanently stuck in the "Updating" state after an OS image update. The agent never calls `Upgrade()` and the device cannot progress.

## Root Cause

`isBootSuccessful()` checks `systemctl is-active boot-complete.target` to determine whether greenboot has validated the boot. On systems without greenboot, `boot-complete.target` exists (shipped by systemd v256+) but is never activated — so it stays `inactive` forever. The function cannot distinguish "greenboot is pending" from "greenboot is not installed", causing an infinite wait.

## Fix

Before checking `boot-complete.target`, the agent now checks whether `greenboot-healthcheck.service` is enabled via `systemctl is-enabled`. If greenboot is not installed or not enabled, `isBootSuccessful()` returns `true` immediately, allowing the OS update to proceed.

Changes:
1. Added `IsEnabled()` method to the systemd client (`internal/agent/client/systemd.go`)
2. Updated `isBootSuccessful()` to gate on greenboot presence (`internal/agent/device/device.go`)
3. Added regression tests covering both greenboot-absent and greenboot-present-but-pending scenarios (`internal/agent/device/device_test.go`)

## Testing

- **Unit tests**: Two new table-driven test cases verify correct behavior when greenboot is absent (update proceeds) and when greenboot is present but pending (update waits). Existing tests updated with greenboot mocks.
- **E2E VM test**: Built a qcow2 image without greenboot, deployed to a kind cluster, triggered an OS image update. Verified the agent completed the full update cycle (prefetch → bootc switch → reboot → UpToDate). Also confirmed the unfixed RPM agent gets stuck in the same scenario.

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -680,9 +680,20 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
 // by greenboot after all required health checks in required.d/ pass.
 // Returns true if boot-complete.target is active, or if greenboot is not installed.
 func (a *Agent) isBootSuccessful(ctx context.Context) bool {
+	// Check if greenboot is actually managing boot validation.
+	// On systems where boot-complete.target exists but greenboot is not installed,
+	// the target stays inactive forever — we must not block on it.
+	enabled, err := a.systemdClient.IsEnabled(ctx, "greenboot-healthcheck.service")
+	if err != nil {
+		a.log.Warnf("Failed to check greenboot-healthcheck.service: %v", err)
+		return true
+	}
+	if !enabled {
+		return true
+	}
+
 	active, err := a.systemdClient.IsActive(ctx, "boot-complete.target")
 	if err != nil {
-		// Unit not found or other error — unexpected on systemd-based systems
 		a.log.Warnf("Failed to check boot-complete.target: %v", err)
 		return true
 	}
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -680,9 +680,9 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
 // by greenboot after all required health checks in required.d/ pass.
 // Returns true if boot-complete.target is active, or if greenboot is not installed.
 func (a *Agent) isBootSuccessful(ctx context.Context) bool {
-	// Check if greenboot is actually managing boot validation.
-	// On systems where boot-complete.target exists but greenboot is not installed,
-	// the target stays inactive forever — we must not block on it.
+	// Defensive check: greenboot is a hard RPM dependency, but the agent binary
+	// may be deployed without RPM (e.g., dev/test environments). If greenboot is
+	// absent, boot-complete.target stays inactive forever — don't block on it.
 	enabled, err := a.systemdClient.IsEnabled(ctx, "greenboot-healthcheck.service")
 	if err != nil {
 		a.log.Warnf("Failed to check greenboot-healthcheck.service: %v", err)
--- a/internal/agent/device/device_test.go
+++ b/internal/agent/device/device_test.go
@@ -121,7 +121,8 @@ func TestSync(t *testing.T) {
 				mockLifecycleManager.EXPECT().Sync(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
 				mockLifecycleManager.EXPECT().AfterUpdate(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
 				mockSpecManager.EXPECT().CheckOsReconciliation(gomock.Any()).Return("", true, nil).AnyTimes()
-				// Mock systemctl for boot success check (via systemd client)
+				// Mock systemctl for greenboot presence and boot success check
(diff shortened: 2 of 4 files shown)


galel12:
@hexfusion — this fix highlights a broader design question worth discussing.

**Context:** Before EDM-3511, the agent RPM had `Requires: greenboot >= 0.15.0`. That was dropped because a version pin broke builds on `rhel-bootc:9.7` (greenboot 0.16+ in repos, 0.15.x unavailable). PR #2762 removed the entire `Requires:` line rather than just relaxing the version constraint, making greenboot fully optional. The consequence was this bug: on systems without greenboot, `boot-complete.target` exists but never activates, so the agent gets stuck in "Updating" forever after an OS update.

This PR fixes the immediate symptom by checking `greenboot-healthcheck.service` is-enabled before blocking on `boot-complete.target`. But **should we require greenboot as a hard dependency for the agent again?**

Arguments for restoring `Requires: greenboot` (unpinned):
- The agent's OS update lifecycle fundamentally depends on boot validation — without greenboot, there's no rollback safety net on failed updates
- `boot-complete.target` semantics only make sense when something (greenboot) is actually gating it
- Every workaround for "greenboot missing" adds defensive complexity and increases the surface for bugs like this one
- All our supported bootc images already ship greenboot
- The greenboot-rs 0.16.x issues that motivated the original pin have been fixed upstream

**Precedent — MicroShift requires greenboot as a hard dependency:**

MicroShift uses a dedicated subpackage (`microshift-greenboot`) that the main package hard-requires, which in turn hard-requires `greenboot` (unpinned). Their spec: [`packaging/rpm/microshift.spec`](https://github.com/openshift/microshift/blob/main/packaging/rpm/microshift.spec#L137)

```
Requires: microshift-greenboot = %{version}   # main package
```
```
%package greenboot
Requires: microshift = %{version}
Requires: greenboot                            # hard dep, no version pin
```

What do you think, should we restore `Requires: greenboot`?

hexfusion:
> What do you think, should we restore Requires: greenboot?

we can not assume greenboot and not ensure it

packaging/rpm/flightctl.spec
```diff
@@ -48,6 +48,7 @@ flightctl is the CLI for controlling the Flight Control service.
 Summary: Flight Control management agent
 
 Requires: flightctl-selinux = %{version}
+Requires: greenboot
```
galel12, line 51: @hexfusion ✔️ restored the pin

internal/agent/device/device.go
```diff
@@ -632,9 +632,20 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
 // by greenboot after all required health checks in required.d/ pass.
 // Returns true if boot-complete.target is active, or if greenboot is not installed.
 func (a *Agent) isBootSuccessful(ctx context.Context) bool {
+	// Check if greenboot is actually managing boot validation.
+	// On systems where boot-complete.target exists but greenboot is not installed,
+	// the target stays inactive forever — we must not block on it.
```
asafbennatan: given that now our RPM requires greenboot - is this comment still correct ?

galel12: Good catch — updated the comment. Greenboot is now a hard RPM dep, so the `IsEnabled` check is a defensive fallback for non-RPM deployments (e.g., dev/test where the binary is pushed directly). See latest commit.

internal/agent/device/device.go
```diff
@@ -680,9 +680,20 @@ func (a *Agent) ReloadConfig(ctx context.Context, config *agent_config.Config) e
...
+	enabled, err := a.systemdClient.IsEnabled(ctx, "greenboot-healthcheck.service")
+	if err != nil {
+		a.log.Warnf("Failed to check greenboot-healthcheck.service: %v", err)
+		return true
+	}
```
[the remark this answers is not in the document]
galel12: Intentionally fail-open. This function runs in a polling loop during reconciliation — fail-closed on a transient `systemctl` error (e.g., dbus timeout) would make the agent think boot hasn't completed and keep waiting, potentially forever. That's the same class of hang this PR fixes. The existing `IsActive` error handler on line 697 uses the same fail-open pattern for the same reason.

internal/agent/device/device_test.go
```diff
@@ -134,6 +135,124 @@ func TestSync(t *testing.T) {
...
+
+				mockSpecManager.EXPECT().IsUpgrading().Return(true).AnyTimes()
+				mockSpecManager.EXPECT().IsOSUpdate().Return(true).AnyTimes()
+
+				// greenboot-healthcheck.service is not enabled (exit 1) — greenboot not installed
+				mockExec.EXPECT().ExecuteWithContext(gomock.Any(), "/usr/bin/systemctl", "is-enabled", "greenboot-healthcheck.service").Return("not-found\n", "", 1).AnyTimes()
+
+				// Upgrade() should be called — greenboot is not installed, so boot is considered successful
+				mockSpecManager.EXPECT().Upgrade(gomock.Any()).Return(nil).AnyTimes()
+			},
```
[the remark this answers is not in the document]
galel12: Done — changed to `MinTimes(1)` in the latest commit.

asafbennatan (approved):
LGTM

kkyrazis (approved):

PR #1126, opened 2025-05-01
EDM-1492: device/systemInfo: add reload support 

hexfusion:
This pr adds drop-in reload support via `<configDir>/conf.d` for the following agent configs.
```
system-info:
system-info-custom:
system-info-timeout:
```
It also ensures the agent refreshes systemInfo and posts a status update on the next sync when a SIGHUP is received via for example, systemctl reload flightctl.service). This enables updating these configurations without restarting the agent process.

### usage examples
```
    systemInfo:
      agentVersion: 0.7.0-main-47-gbda61b95
      architecture: amd64
      bootID: b9edea97-d761-44cb-a3f4-bcccfc84bfdd
      distroName: CentOS Stream
      distroVersion: "9"
      hostname: localhost.localdomain
      kernel: 5.14.0-515.el9.x86_64
      netInterfaceDefault: enp1s0
      netIpDefault: 192.168.124.67/24
      netMacDefault: 52:54:00:af:12:4b
      operatingSystem: linux
      productName: Standard PC (Q35 + ICH9, 2009)
      productSerial: ""
      productUuid: 745563f8-0e9f-4178-921d-b14192fd0316
```

updated config to only show hostname.

```
system-info:
- hostname
```

```
systemctl reload flightctl-agent.service

flightctl-agent[1925]: time="2025-05-05T19:39:18.558236Z" level=info msg="Agent received SIGHUP signal"
flightctl-agent[1925]: time="2025-05-05T19:39:18.559417Z" level=info msg="Reloading agent config"
flightctl-agent[1925]: time="2025-05-05T19:39:18.562062Z" level=info msg="Updating log level to info"
flightctl-agent[1925]: time="2025-05-05T19:39:18.563542Z" level=info msg="Reloading system info config"
flightctl-agent[1925]: time="2025-05-05T19:39:18.563934Z" level=info msg="Updating system info keys: [hostname kernel distroName distroVersion productName productUuid productSerial netInterfaceDefault netIpDefault netMacDefault] -> [hostname]"
flightctl-agent[1925]: time="2025-05-05T19:39:18.839639Z" level=info msg="Detected default route: 192.168.124.1 via enp1s0"
flightctl-agent[1925]: time="2025-05-05T19:39:18.845074Z" level=info msg="No custom info keys provided, skipping custom info collection"
flightctl-agent[1925]: time="2025-05-05T19:39:18.845952Z" level=info msg="Agent config reloaded"
```
```
    systemInfo:
[...]

PR #2823, opened 2026-04-12
EDM-3454: Skip invalid config drop-ins instead of crashing

galel12:
## Problem

When a device spec delivers an invalid YAML file to `/etc/flightctl/conf.d/`, the agent crashes at startup with `log.Fatalf` before reaching bootstrap/rollback logic. Combined with an OS update, this permanently bricks the device: greenboot rolls back the OS, but `/etc` is a persistent overlay in bootc systems, so the bad config file survives and the agent crash-loops indefinitely.

## Root Cause

`LoadWithOverrides()` in `internal/agent/config/config.go` treated YAML parse errors in individual `conf.d/` drop-in files as fatal errors. The error propagated to `main.go:97` where `log.Fatalf` terminated the process. Introduced in PR #1126 (EDM-1492) when `conf.d/` support was added without distinguishing recoverable drop-in failures from fatal base config failures.

## Fix

Change the drop-in loop in `LoadWithOverrides` to log a warning and `continue` to the next file when a drop-in is unreadable or contains invalid YAML. Base config parse errors, `Complete()`, and `Validate()` remain fatal.

This matches the resilience model already used in `reload.go:63-66`, where `config.Load()` errors are handled gracefully.

## Testing

- **Unit tests**: 3 new test cases covering invalid drop-in skip, base config still fatal, and all-invalid-dropins scenario. All config package tests pass.
- **VM verification**: Deployed fixed binary to the bricked agent VM. Agent starts, logs warning for the invalid drop-in, detects OS mismatch, performs spec rollback, and completes bootstrap.

## Risk Assessment

LOW — The change is 2 lines of production code (error return → warn + continue), scoped to the drop-in iteration loop only. Base config errors are unaffected. The `mergeConfigs` function limits which fields drop-ins can override, so semantic validation via `Validate()` is not impacted by the fields drop-ins touch.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -340,6 +340,9 @@ func (a *Agent) Run(ctx context.Context) error {
 	statusManager.RegisterStatusExporter(osManager)
 	statusManager.RegisterStatusExporter(specManager)
 	statusManager.RegisterStatusExporter(systemInfoManager)
+	if len(a.config.Warnings) > 0 {
+		statusManager.RegisterStatusExporter(&configWarningExporter{warnings: a.config.Warnings})
+	}
 
 	// create config controller
 	configController := config.NewController(
@@ -367,17 +370,6 @@ func (a *Agent) Run(ctx context.Context) error {
 	// bootstrap
 	if err := bootstrap.Initialize(ctx); err != nil {
 		return fmt.Errorf("bootstrap failed: %w", err)
-	}
-
-	if len(a.config.Warnings) > 0 {
-		msg := strings.Join(a.config.Warnings, "; ")
-		_, updateErr := statusManager.Update(ctx, status.SetDeviceSummary(v1beta1.DeviceSummaryStatus{
-			Status: v1beta1.DeviceSummaryStatusDegraded,
-			Info:   &msg,
-		}))
-		if updateErr != nil {
-			a.log.Warnf("Failed to report config warnings to status: %v", updateErr)
-		}
 	}
 
 	// Initialize certificate manager
@@ -539,5 +531,23 @@ func wipeCertificateAndRestart(ctx context.Context, identityProvider identity.Pr
 	}
 
 	log.Info("Successfully wiped certificate and restarted flightctl-agent service")
+	return nil
+}
+
+// configWarningExporter surfaces config loading warnings (e.g. skipped drop-ins)
+// in the device summary on every status collection cycle.
+type configWarningExporter struct {
+	warnings []string
+}
+
+func (e *configWarningExporter) Status(_ context.Context, s *v1beta1.DeviceStatus, _ ...status.CollectorOpt) error {
+	if len(e.warnings) == 0 {
+		return nil
+	}
+	msg := log.Truncate(strings.Join(e.warnings, "; "), status.MaxMessageLength)
+	if s.Summary.Status == v1beta1.DeviceSummaryStatusOnline {
+		s.Summary.Status = v1beta1.DeviceSummaryStatusDegraded
+	}
+	s.Summary.Info = &msg
 	return nil
 }
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -341,7 +341,7 @@ func (a *Agent) Run(ctx context.Context) error {
 	statusManager.RegisterStatusExporter(specManager)
 	statusManager.RegisterStatusExporter(systemInfoManager)
 	if len(a.config.Warnings) > 0 {
-		statusManager.RegisterStatusExporter(&configWarningExporter{warnings: a.config.Warnings})
+		statusManager.RegisterStatusExporter(newConfigWarningExporter(a.config.Warnings))
 	}
 
 	// create config controller
@@ -537,17 +537,20 @@ func wipeCertificateAndRestart(ctx context.Context, identityProvider identity.Pr
 // configWarningExporter surfaces config loading warnings (e.g. skipped drop-ins)
 // in the device summary on every status collection cycle.
 type configWarningExporter struct {
-	warnings []string
+	msg string
+}
+
+func newConfigWarningExporter(warnings []string) *configWarningExporter {
+	return &configWarningExporter{
+		msg: log.Truncate(strings.Join(warnings, "; "), status.MaxMessageLength),
+	}
 }
 
 func (e *configWarningExporter) Status(_ context.Context, s *v1beta1.DeviceStatus, _ ...status.CollectorOpt) error {
-	if len(e.warnings) == 0 {
+	if s.Summary.Status != v1beta1.DeviceSummaryStatusOnline {
 		return nil
 	}
-	msg := log.Truncate(strings.Join(e.warnings, "; "), status.MaxMessageLength)
-	if s.Summary.Status == v1beta1.DeviceSummaryStatusOnline {
-		s.Summary.Status = v1beta1.DeviceSummaryStatusDegraded
-	}
-	s.Summary.Info = &msg
+	s.Summary.Status = v1beta1.DeviceSummaryStatusDegraded
+	s.Summary.Info = &e.msg
 	return nil
 }
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -366,6 +367,17 @@ func (a *Agent) Run(ctx context.Context) error {
 	// bootstrap
 	if err := bootstrap.Initialize(ctx); err != nil {
 		return fmt.Errorf("bootstrap failed: %w", err)
+	}
+
+	if len(a.config.Warnings) > 0 {
+		msg := strings.Join(a.config.Warnings, "; ")
+		_, updateErr := statusManager.Update(ctx, status.SetDeviceSummary(v1beta1.DeviceSummaryStatus{
+			Status: v1beta1.DeviceSummaryStatusDegraded,
+			Info:   &msg,
+		}))
+		if updateErr != nil {
+			a.log.Warnf("Failed to report config warnings to status: %v", updateErr)
+		}
 	}
 
 	// Initialize certificate manager
--- a/internal/agent/config/config.go
+++ b/internal/agent/config/config.go
@@ -421,16 +421,20 @@ func (cfg *Config) LoadWithOverrides(configFile string) error {
 	}
 	sort.Strings(yamlFiles)
 
-	// Apply drop-ins in order (later files override earlier ones)
+	// Apply drop-ins in order (later files override earlier ones).
+	// Invalid drop-ins are skipped with a warning so the agent can still
+	// start and perform spec rollback if needed.
 	for _, filename := range yamlFiles {
 		overrideCfg := &Config{}
 		overridePath := filepath.Join(confSubdir, filename)
 		contents, err := cfg.readWriter.ReadFile(overridePath)
 		if err != nil {
-			return fmt.Errorf("reading override config %s: %w", overridePath, err)
+			logrus.Warnf("Skipping unreadable override config %s: %v", overridePath, err)
+			continue
 		}
 		if err := yaml.Unmarshal(contents, overrideCfg); err != nil {
-			return fmt.Errorf("unmarshalling override config %s: %w", overridePath, err)
+			logrus.Warnf("Skipping invalid override config %s: %v", overridePath, err)
+			continue
 		}
 		mergeConfigs(cfg, overrideCfg)
 	}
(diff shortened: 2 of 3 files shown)


internal/agent/config/config.go
```diff
@@ -421,16 +421,20 @@ func (cfg *Config) LoadWithOverrides(configFile string) error {
...
+	// Apply drop-ins in order (later files override earlier ones).
+	// Invalid drop-ins are skipped with a warning so the agent can still
+	// start and perform spec rollback if needed.
 	for _, filename := range yamlFiles {
 		overrideCfg := &Config{}
 		overridePath := filepath.Join(confSubdir, filename)
 		contents, err := cfg.readWriter.ReadFile(overridePath)
 		if err != nil {
-			return fmt.Errorf("reading override config %s: %w", overridePath, err)
+			logrus.Warnf("Skipping unreadable override config %s: %v", overridePath, err)
```
kkyrazis, line 432: I'm not against this approach, and perhaps as a follow up it could be possible make this more visible that a drop-in is malformed, but what about cases in which the drop-in is parsed, but results in an invalid config (Validate fails). Wouldn't we be in the same spot?

asafbennatan: agreed - we should make sure this is propagated to agent status ?

internal/agent/config/config_test.go
```diff
@@ -271,3 +271,104 @@ image-pruning:
...
+		[]byte("{{{not yaml at all\n"),
+		0o600,
+	))
+
+	err := cfg.LoadWithOverrides(configFile)
+	require.NoError(err, "all invalid dropins should be skipped, agent should still start")
+
+	// Base config values should be intact
+	require.Equal("https://enrollment.endpoint", cfg.EnrollmentService.Service.Server)
+}
```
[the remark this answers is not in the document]
galel12: Good catch — added `TestLoadWithOverrides_UnreadableDropinIsSkipped` which creates a drop-in with `0o000` permissions and verifies the `ReadFile` failure branch is exercised (warning logged, dropin skipped, subsequent valid dropin still applied).

internal/agent/config/config.go
galel12: Addressed both points:

Regarding Kyle's (@kkyrazis ) `Validate()` question - `Validate()` runs on the *merged* config (base + all valid drop-ins), not on individual drop-ins. The merge is additive via `mergeConfigs`, which only overrides fields that the drop-in explicitly sets. So a drop-in that parses but has semantically invalid values for fields that `Validate()` checks (like enrollment/management endpoints) would need to explicitly override those fields — which is a much narrower scenario than the syntactically-broken YAML that triggers this bug.

Regarding propagating to agent status (@asafbennatan ) - done. Skipped-dropin warnings are now collected in `Config.Warnings` during loading and surfaced as a `Degraded` device summary status after bootstrap. This makes malformed drop-ins visible from the control plane, not just local logs.

kkyrazis: My main concern is that the agent is still able to enter a state in which it can't recover without manual intervention via a simple spec. I agree that it is far less likely now without a drop-in directly targeting an invalid config value, but I think it should still be fixed at some point.

galel12: Agreed — the Validate() path is still vulnerable. The fix for that is more involved (need to either validate incrementally after each drop-in merge and skip offenders, or validate the final result and fall back to base-only config), so I'll file a follow-up to keep this PR focused on the common case. The remaining risk surface is narrow since it requires a drop-in to explicitly target a validated field with an invalid value rather than an accidental tab/space issue.

internal/agent/agent.go
```diff
@@ -368,6 +369,17 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	if len(a.config.Warnings) > 0 {
+		msg := strings.Join(a.config.Warnings, "; ")
+		_, updateErr := statusManager.Update(ctx, status.SetDeviceSummary(v1beta1.DeviceSummaryStatus{
+			Status: v1beta1.DeviceSummaryStatusDegraded,
+			Info:   &msg,
+		}))
+		if updateErr != nil {
+			a.log.Warnf("Failed to report config warnings to status: %v", updateErr)
+		}
+	}
```
asafbennatan: **[Finding 1 — HIGH]** This one-shot `SetDeviceSummary(Degraded)` is overwritten by the first periodic `Sync()` call. `Sync()` calls `collect()` which runs `ResourceManager.Status()`, and that exporter unconditionally sets `Summary.Status = Online` when no resource alerts are firing. The warning disappears within one sync interval.

Suggested fix: register a status exporter that checks `cfg.Warnings` on every `collect()` cycle so the degraded summary persists as long as the warnings are present, mirroring how `ResourceManager` keeps resource degradation visible across syncs.

asafbennatan: This one-shot `SetDeviceSummary(Degraded)` pushes the degraded summary correctly on the first call, but the warning won't survive subsequent sync cycles.

`statusManager.Update` calls `collect()` before applying the update functions — so the sequence is:
1. `collect()` runs `ResourceManager.Status()`, which **unconditionally** sets `Summary.Status = Online` when no resource alerts are firing.
2. `SetDeviceSummary(Degraded)` then overwrites it → first push is `Degraded` ✓
3. On every subsequent `Sync()`, `collect()` runs again → `ResourceManager` resets `Summary.Status = Online` → warning is gone ✗

Suggested fix: register a status exporter instead of a one-shot update, so the degraded summary participates in every `collect()` cycle:

```go
type configWarningExporter struct{ warnings []string }

func (e *configWarningExporter) Status(_ context.Context, s *v1beta1.DeviceStatus, _ ...status.CollectorOpt) error {
    if len(e.warnings) > 0 {
        msg := strings.Join(e.warnings, "; ")
        s.Summary.Status = v1beta1.DeviceSummaryStatusDegraded
        s.Summary.Info = &msg
    }
    return nil
}
```

Then replace the one-time `Update` block with:
```go
statusManager.RegisterStatusExporter(&configWarningExporter{warnings: a.config.Warnings})
```

internal/agent/config/config_test.go
```diff
@@ -271,3 +271,153 @@ image-pruning:
...
+
+	configFile := filepath.Join(configDir, "config.yaml")
+	require.NoError(os.WriteFile(configFile, []byte(yamlConfig), 0o600))
+
+	dropinDir := filepath.Join(configDir, "conf.d")
+	require.NoError(os.MkdirAll(dropinDir, 0o755))
+
+	// Create a dropin file with no read permissions
+	unreadablePath := filepath.Join(dropinDir, "01-unreadable.yaml")
+	require.NoError(os.WriteFile(unreadablePath, []byte("log-level: debug\n"), 0o000))
```
asafbennatan: This relies on `0o000` permissions to simulate an unreadable file, but on Linux `root` bypasses DAC permissions entirely (`CAP_DAC_OVERRIDE`). Many CI containers run as root — in that case `ReadFile` succeeds, no warning is recorded, and the `require.Len(cfg.Warnings, 1)` assertion below will fail.

Suggested fix — add this guard right after `require := require.New(t)`:
```go
if os.Getuid() == 0 {
    t.Skip("permission test is not meaningful when running as root")
}
```

internal/agent/config/config_test.go
```diff
@@ -271,3 +271,153 @@ image-pruning:
 	require.NotNil(cfg.ImagePruning.Enabled, "Enabled should not be nil when dropin enables pruning")
 	require.True(*cfg.ImagePruning.Enabled, "pruning dropin should override config setting")
 }
+
+func TestLoadWithOverrides_InvalidDropinIsSkipped(t *testing.T) {
```
asafbennatan: The four new `TestLoadWithOverrides_*` functions share identical boilerplate (temp-dir creation, `Config` struct init, config-file write). `internal/agent/AGENTS.md` requires table-driven tests — these would be a good candidate to consolidate into a single `TestLoadWithOverrides_DropinErrorHandling` with a `testCases` table, where each case provides the drop-in files to write and the expected outcome.

internal/agent/agent.go
galel12: Fixed in c4ffc4f0 — replaced the one-shot Update with a `configWarningExporter` registered after resourceManager. It runs on every collect() cycle so the Degraded summary persists. Only escalates Online → Degraded, never downgrades an existing Error. Also truncates via `log.Truncate(msg, MaxMessageLength)`.

internal/agent/config/config_test.go
galel12: Fixed in c4ffc4f0 — added `os.Getuid() == 0` skip guard.

internal/agent/config/config_test.go
galel12: Fixed in c4ffc4f0 — consolidated into a single table-driven `TestLoadWithOverrides_DropinErrorHandling` with "When ... it should ..." naming.

internal/agent/agent.go
galel12: Fixed in c4ffc4f0 — replaced with a registered `configWarningExporter` that runs on every collect() cycle. Only escalates Online → Degraded, never downgrades existing Error/Degraded from resources.

internal/agent/agent.go
```diff
@@ -529,3 +533,24 @@ func wipeCertificateAndRestart(ctx context.Context, identityProvider identity.Pr
...
+	msg string
+}
+
+func newConfigWarningExporter(warnings []string) *configWarningExporter {
+	return &configWarningExporter{
+		msg: log.Truncate(strings.Join(warnings, "; "), status.MaxMessageLength),
+	}
+}
+
+func (e *configWarningExporter) Status(_ context.Context, s *v1beta1.DeviceStatus, _ ...status.CollectorOpt) error {
```
asafbennatan, line 549: i think it might be a good time to adjust this , the Exporter interface does too much :
`// Status collects status information and updates the device status.`

it shouldnt do that , each exporter should return the status part it manages without touching the device.
the status manager should aggregate the data (collect individual status parts) , calculate the aggregate summary status

otherwise the exporter status is very ambiguous and brittle 
WDYT ? @galel12

asafbennatan: concrete suggestion

```
type ExporterResult struct {
    // Applies this exporter's domain fields onto the device status
    Apply   func(*v1beta1.DeviceStatus)
    // The exporter's opinion on device-level summary (nil = no opinion)
    Summary *v1beta1.DeviceSummaryStatus
}
type Exporter interface {
    Status(context.Context, ...CollectorOpt) (ExporterResult, error)
}


// status manager

func (m *StatusManager) collect(ctx context.Context, opts ...CollectorOpt) error {
    m.reset()
    var votes []v1beta1.DeviceSummaryStatus
    var errs []error
    for _, exporter := range m.exporters {
        result, err := exporter.Status(ctx, opts...)
        if errors.Is(err, context.Canceled) {
            return nil
        }
        if err != nil {
            errs = append(errs, err)
            continue
        }
        result.Apply(m.device.Status)
        if result.Summary != nil {
            votes = append(votes, *result.Summary)
        }
    }
    m.device.Status.Summary = aggregateSummary(votes)
    return errors.Join(errs...)
}

//how individual exporters look like 

// resourceManager
func (m *ResourceManager) Status(ctx context.Context, _ ...status.CollectorOpt) (status.ExporterResult, error) {
    cpu, disk, memory, summaryVote := m.computeResourceStatus()
    return status.ExporterResult{
        Apply: func(s *v1beta1.DeviceStatus) {
            s.Resources.Cpu    = cpu
            s.Resources.Disk   = disk
            s.Resources.Memory = memory
        },
        Summary: &summaryVote,
    }, nil
}

// osManager — no summary opinion
func (m *manager) Status(ctx context.Context, _ ...status.CollectorOpt) (status.ExporterResult, error) {
    image, digest, err := m.getOSStatus(ctx)
    return status.ExporterResult{
        Apply: func(s *v1beta1.DeviceStatus) {
            s.Os.Image       = image
            s.Os.ImageDigest = digest
        },
        Summary: nil,
    }, err
}

// new integrityManager — has a summary opinion or whatever - adjust to your exporter
func (m *manager) Status(ctx context.Context, _ ...status.CollectorOpt) (status.ExporterResult, error) {
    integrity, summaryVote, err := m.check(ctx)
    return status.ExporterResult{
        Apply:   func(s *v1beta1.DeviceStatus) { s.Integrity = integrity },
        Summary: &summaryVote,
    }, err
}



func aggregateSummary(votes []v1beta1.DeviceSummaryStatus) v1beta1.DeviceSummaryStatus {
    // decide how to do that based on the "votes" from exporters
}
```

galel12: Addressed in follow-up PR: https://github.com/flightctl/flightctl/pull/3137

Refactored the `Exporter` interface so each exporter returns a `StatusContribution` struct (with typed fields per domain) instead of mutating a shared `*DeviceStatus`. The `StatusManager` now merges contributions and picks the highest-severity summary via a priority map. Config warnings are also handled inside the manager constructor rather than a separate exporter.

asafbennatan (approved):
LGTM

PR #2825, opened 2026-04-13
EDM-2854: Cleanup job should remove other jobs too

rawagner:
## Summary by CodeRabbit

* **Chores**
  * Enhanced upgrade testing workflow to uninstall the release and verify no pods, jobs, PVCs or matching PVs remain; the job fails if leftover resources are detected.
  * Expanded post-delete cleanup to remove additional job resources and ensure orphaned PVCs/PVs are cleaned up during and after upgrades.


.github/workflows/pr-helm-upgrade-testing.yaml
```diff
@@ -329,6 +329,73 @@ jobs:
...
+            JOBS=$(kubectl get jobs -n "$NS" --no-headers 2>/dev/null | wc -l)
+            if [ "$JOBS" -ne 0 ]; then
+              echo "FAIL: $JOBS job(s) still exist in $NS:"
+              kubectl get jobs -n "$NS"
+              FAILED=true
+            else
+              echo "✓ No jobs in $NS"
+            fi
+
+            PVCS=$(kubectl get pvc -n "$NS" --no-headers 2>/dev/null | wc -l)
```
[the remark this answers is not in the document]
rawagner: helm uninstall does not remove namespace. So the additional guardrail is not needed.

deploy/helm/flightctl/templates/flightctl-cleanup-job.yaml
```diff
@@ -36,6 +37,10 @@ spec:
                 | grep "flightctl-alertmanager-data-flightctl-alertmanager-" \
                 | xargs -r oc delete -n {{ $ns }} --ignore-not-found
               {{- end }}
+              echo "Deleting Jobs..."
+              {{- range $ns := $namespaces }}
+              oc delete job -l 'flightctl.service,flightctl.service!=cleanup' -n {{ $ns }} --ignore-not-found
```
celdrake: Hi! I'm not overly familiar with the existing clean-up requirements.
From Cursor, it would seem like this deletion is based on job metadata labels, not the pod template labels. Based on this, it seems like we'd be missing to remove the `flightctl-preupgrade-scale-to-zero` job.

The suggestion would be to delete all jobs except the clean-up one.
```
oc delete job -l app.kubernetes.io/instance={{ .Release.Name }} \
  --field-selector metadata.name!=flightctl-cleanup \
  -n {{ $ns }} --ignore-not-found
```

It also raised a concern that the exclusion of deleting the `clean-up` job itself is not explicit, and it could be wrongly targeted if someone updated the job metadata.

deploy/helm/flightctl/templates/flightctl-cleanup-job.yaml
```diff
@@ -5,18 +5,19 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: flightctl-cleanup-pvcs
+  name: flightctl-cleanup
```
celdrake, line 9: By renaming the job, and the old "cleanup-pvcs" not having the `flightctl-service` job metadata labels, they would not be cleaned (See comment in line 42).

The suggestion below would also delete the existing `flightctl-cleanup-pvcs` job AFAIAA

deploy/helm/flightctl/templates/flightctl-cleanup-job.yaml
rawagner: ok, now targeting the jobs using the `app.kubernetes.io/instance` selector.

i have also added additional info into `echo` msg about cleanup job exception

deploy/helm/flightctl/templates/flightctl-cleanup-job.yaml
rawagner: you are right, there is a corner-case where the old `-pvcs` job may still exist and the new code would omit it. Updated `field-selector` as suggested.

celdrake (approved):
Tested in OCP (install --> upgrade --> uninstall).
No leftovers were left after the uninstall, all jobs including the new cleanup job were deleted in the end.

PR #2826, opened 2026-04-13
NO-ISSUE: mandate backporting selection

asafbennatan:
## Summary by CodeRabbit

* **Documentation**
  * Added backport labeling requirements documentation and guidance
  * Updated pull request template with mandatory backport intent section

* **Chores**
  * Added automated validation for backport labels on pull requests to the main branch

### DIFF
--- a/.github/BACKPORT_LABELS.md
+++ b/.github/BACKPORT_LABELS.md
@@ -1,0 +1,38 @@
+# Backport labels (Phase 1)
+
+Pull requests targeting **`main`** must declare backport intent using repository labels. CI enforces this via the **PR backport labels** workflow (job **Backport labels**).
+
+## Rules
+
+- **`backport: none`** — the change is not intended for backport to release branches.
+- **`backport: release-<branch-suffix>`** — one or more labels, each naming a **`release-*` branch** on the remote. The label text after `backport: ` must be exactly the branch name (e.g. branch `release-1.1` → label `backport: release-1.1`).
+- **Invalid:** no qualifying labels, or **`backport: none`** together with any **`backport: release-*`** label.
+
+## Creating labels
+
+Create labels in the GitHub UI (**Issues → Labels**) or with the CLI:
+
+```bash
+# Required opt-out label (create once)
+gh label create "backport: none" --color "BFD4F2" --description "PR is not intended for backport to release branches"
+
+# For each active release branch (repeat when a new release line is cut)
+BRANCH=release-1.1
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-backport-labels.yaml
```diff
@@ -0,0 +1,50 @@
...
+  backport-labels:
+    name: Backport labels
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Require backport intent labels
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const noneLabel = 'backport: none';
```
amir-yogev-gh, line 24: The format should be "backport-to-*" for backward compatibility

.github/BACKPORT_LABELS.md
```diff
@@ -0,0 +1,38 @@
+# Backport labels (Phase 1)
+
+Pull requests targeting **`main`** must declare backport intent using repository labels. CI enforces this via the **PR backport labels** workflow (job **Backport labels**).
+
+## Rules
+
+- **`backport: none`** — the change is not intended for backport to release branches.
+- **`backport: release-<branch-suffix>`** — one or more labels, each naming a **`release-*` branch** on the remote. The label text after `backport: ` must be exactly the branch name (e.g. branch `release-1.1` → label `backport: release-1.1`).
```
amir-yogev-gh, line 8: See previous comment on label template

.github/workflows/pr-backport-labels.yaml
asafbennatan: not sure if we need to retain backwards compatibility , since its relevant only from when we start using it . 
the benefit of this is that we have the exact release branch name after the : or 'none'

keitwb:
As discussed, requiring the labels does seem like the only way to avoid missing things in backporting (assuming everybody labels it correctly), but I think the labeling should be more generic to indicate the intended release(s), using the existing `release-X.Y` labels that were kind of used informally before.

This scheme gives us information on all PRs about the intended release, not just those made after backporting has begun after feature freeze. We can use it for backporting verification as well and not have to worry about the edge case of a PR that was open before feature freeze that was intended to be merged before the release branch was cut and assumed to not need backporting, but got delayed and then has the wrong label.

Also we could automate the labeling with some kind of Jira integration since we have the target versions on Jira issues and not have to worry about whether it is in the "backporting" phase of the dev cycle or not.

keitwb (approved):

PR #2827, opened 2026-04-13
EDM-2839: Suppress misleading worker warning logs

kkyrazis:
## Summary
- Distinguishes between missing Kubernetes environment variables (`rest.ErrNotInCluster`) and actual k8s client errors during worker startup. Missing env vars log at `Info` level; other errors log at `Error` level. Both allow the worker to continue with k8s features disabled.
- Downgrades the checkpoint parse log in `queue_maintenance.go` for fresh systems: empty checkpoint logs `Info`, malformed non-empty checkpoint logs `Error` with the value included.
- Fixes flaky `TestPeriodicTaskConsumer_ContextCanceled` test by waiting for all consumer goroutines to exit before publishing a second task, eliminating a `select` race condition.

Fixes: [EDM-2839](https://redhat.atlassian.net/browse/EDM-2839), [EDM-2415](https://redhat.atlassian.net/browse/EDM-2415)

## Test plan
- [x] `make unit-test` passes
- [x] `make lint` passes
- [x] Deploy via Helm (`make deploy`) — worker starts cleanly, no misleading warnings
- [x] Run worker container outside k8s (`podman run`) — logs `"Kubernetes environment not detected, kubernetes features are disabled"` at info level
- [x] Flaky test fix verified with 50 consecutive passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes cluster detection to distinguish between running outside a cluster and other initialization failures, with appropriate logging for each scenario.
  * Enhanced checkpoint recovery error logging to provide more detailed diagnostic information, including the raw checkpoint value, for better troubleshooting.
  * Improved consumer shutdown handling with more reliable signal verification.

### DIFF
--- a/cmd/flightctl-worker/main.go
+++ b/cmd/flightctl-worker/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -20,6 +21,7 @@ import (
 	"github.com/flightctl/flightctl/pkg/queues"
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/rest"
 )
 
 func main() {
@@ -64,8 +66,11 @@ func main() {
 
 	k8sClient, err := k8sclient.NewK8SClient()
 	if err != nil {
-		log.WithError(err).Warning("initializing k8s client, assuming k8s is not supported")
-		k8sClient = nil
+		if errors.Is(err, rest.ErrNotInCluster) {
+			log.Info("Kubernetes environment not detected, kubernetes features are disabled")
+		} else {
+			log.WithError(err).Error("initializing k8s client, kubernetes features are disabled")
+		}
 	}
 
 	// Initialize metrics collectors
--- a/internal/tasks/queue_maintenance.go
+++ b/internal/tasks/queue_maintenance.go
@@ -239,13 +239,14 @@ func (t *QueueMaintenanceTask) recoverFromMissingCheckpoint(ctx context.Context,
 			return fmt.Errorf("failed to get checkpoint from database: %s", status.Message)
 		}
 	} else {
-		// Success case - parse the database checkpoint
-		if parsedTime, err := time.Parse(time.RFC3339Nano, string(dbCheckpointBytes)); err == nil {
+		rawCheckpoint := string(dbCheckpointBytes)
+		if rawCheckpoint == "" {
+			log.Info("Database checkpoint is empty, treating as fresh system")
+		} else if parsedTime, err := time.Parse(time.RFC3339Nano, rawCheckpoint); err != nil {
+			log.WithError(err).WithField("checkpoint", rawCheckpoint).Error("Failed to parse database checkpoint, treating as fresh system")
+		} else {
 			lastCheckpointTime = parsedTime
 			log.WithField("lastCheckpoint", lastCheckpointTime.Format(time.RFC3339Nano)).Info("Found last checkpoint in database")
-		} else {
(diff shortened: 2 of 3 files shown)


internal/tasks/queue_maintenance.go
```diff
@@ -239,13 +239,14 @@ func (t *QueueMaintenanceTask) recoverFromMissingCheckpoint(ctx context.Context,
 			return fmt.Errorf("failed to get checkpoint from database: %s", status.Message)
 		}
 	} else {
-		// Success case - parse the database checkpoint
-		if parsedTime, err := time.Parse(time.RFC3339Nano, string(dbCheckpointBytes)); err == nil {
+		rawCheckpoint := string(dbCheckpointBytes)
```
keitwb, line 242: What does this have to do with this PR?

deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
```diff
@@ -26,6 +26,7 @@ data:
         {{- end }}
         {{- end }}
     service: {}
+    target: kubernetes
```
keitwb: Having more config just to suppress a warning log message is pretty heavy-handed.

Did you consider just detecting it automatically with the k8s-specific envvars that are pretty much guaranteed to be there (the k8s client init looks for them too)?https://kubernetes.io/docs/tutorials/services/connect-applications-service/#environment-variables

internal/tasks/queue_maintenance.go
kkyrazis: The jira tickets call out two logs that add noise. 
1. Quadlets warning about the lack of k8s clients
2. Fresh deployments warning about a lack of db checkpoints

To fix the first "issue", this adds a deployment type so that we can properly log rather than guess.

This fixes the second issue. I can open a separate PR if desired, but it was such a small change that I left it in.

deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
kkyrazis: I mean that's whats already happening. A nil k8s client implies quadlets which is why it doesn't stop the service. Can just drop the respective log to Info / reword it, but I also don't think we need to be in the position where we have to guess what environment type we're in.

internal/tasks/queue_maintenance.go
keitwb: You can leave it in but make the PR title reflect the fact that this is a PR for suppressing warnings.

deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
keitwb: It isn't really guessing though, it is directly deducing it from an envvar (`KUBERNETES_SERVICE_HOST`) that is documented to be there in k8s envs, without exception. It fulfills the same purpose as having the user specify a "target" without any risk of error, and without bloating the configs with more options.

cmd/flightctl-worker/main.go
```diff
@@ -62,10 +63,9 @@ func main() {
 		log.Fatalf("failed connecting to Redis queue: %v", err)
 	}
 
-	k8sClient, err := k8sclient.NewK8SClient()
+	k8sClient, err := initK8sClient(log, cfg)
 	if err != nil {
-		log.WithError(err).Warning("initializing k8s client, assuming k8s is not supported")
-		k8sClient = nil
+		log.Fatalf("initializing k8s client: %v", err)
```
keitwb: This is debatable whether the service should straight up blow up if the k8s client can't be created. It is only used to access secrets as far as I can tell, which are an optional config component. Doing this makes the services slightly more fragile to environmental factors so I would make sure there is a good reason for it.

keitwb (approved):

PR #2828, opened 2026-04-13
EDM-3684: Allow stopping quadlet-generated services on Fedora 43.6+

kkyrazis:
## Summary

- Fedora 43.6+ (`selinux-policy-targeted >= 43.6`) introduced `systemd_generator_unit_file_t` for quadlet-generated units in `/run/systemd/generator/` ([fedora-selinux/selinux-policy#2764](https://github.com/fedora-selinux/selinux-policy/pull/2764)). The existing `systemd_stop_systemd_services` macro only covers `systemd_unit_file_t`, causing AVC denials when the agent tries to stop quadlet services.
- Adds an `optional_policy` block granting `stop` on `systemd_generator_unit_file_t`. The block is silently skipped at compile time on platforms where this type does not exist (EL9, older Fedora).
- CentOS/RHEL is unaffected since their policy still labels generator output as `systemd_unit_file_t`.

## Test plan

- [x] Policy compiles on Fedora 43 (`selinux-policy-devel-43.3`)
- [x] Policy compiles on CentOS Stream 9 (`selinux-policy-devel-38.1.76`)
- [x] Policy compiles on CentOS Stream 10 (`selinux-policy-devel-42.1.19`)
- [x] Policy compiles on RHEL 9 (`selinux-policy-devel-38.1.65`)
- [x] Policy compiles on RHEL 10 (`selinux-policy-devel-42.1.7`)
- [x] On Fedora 43.6, verified compiled policy contains `allow flightctl_agent_t systemd_generator_unit_file_t:service stop` rule

Jira: [EDM-3684](https://redhat.atlassian.net/browse/EDM-3684)

## Summary by CodeRabbit

* **Chores**
  * Enhanced agent permissions to stop a broader set of system services on platforms that support newer service types, improving its ability to manage generated or non-standard units.
  * Maintains existing behavior and compatibility on older platforms where the newer service type support is unavailable, avoiding changes to current deployments.

### DIFF
--- a/packaging/selinux/flightctl_agent.te
+++ b/packaging/selinux/flightctl_agent.te
@@ -111,7 +111,13 @@ systemd_start_all_unit_files(flightctl_agent_t)
 systemd_status_all_unit_files(flightctl_agent_t)
 systemd_reload_all_services(flightctl_agent_t)
 systemd_start_all_services(flightctl_agent_t)
-systemd_stop_systemd_services(flightctl_agent_t)
+# Use systemd_stop_all_services to cover systemd_generator_unit_file_t (Fedora 43.6+),
+# falling back to systemd_stop_systemd_services on EL9 where the macro doesn't exist.
+ifdef(`systemd_stop_all_services',`
+    systemd_stop_all_services(flightctl_agent_t)
+',`
+    systemd_stop_systemd_services(flightctl_agent_t)
+')
 allow flightctl_agent_t init_t:system { status stop start };
 
 # sshd config/service permissions


packaging/selinux/flightctl_agent.te
```diff
@@ -112,6 +112,16 @@ systemd_status_all_unit_files(flightctl_agent_t)
...
 systemd_stop_systemd_services(flightctl_agent_t)
+# Fedora 43.6+ (selinux-policy PR #2764) labels quadlet-generated units in /run/systemd/generator/
+# as systemd_generator_unit_file_t rather than systemd_unit_file_t. Without this rule, the agent
+# is denied permission to stop quadlet services. The optional_policy block is silently skipped on
+# platforms where this type does not exist (EL9, older Fedora).
+optional_policy(`
+    gen_require(`
+        type systemd_generator_unit_file_t;
+    ')
+    allow flightctl_agent_t systemd_generator_unit_file_t:service stop;
```
keitwb: Do we really only need "stop" and no other actions?

kkyrazis: I need to fully test it in these systems, but it's the fact that `systemd_start_all_services` didn't originally have a `systemd_stop_all_services` counterpart, just  the `systemd_stop_systemd_services`. So start_all expands correctly while stop_systemd doesn't

keitwb: Looks like they did [add the one for stop all services in newer versions](https://github.com/fedora-selinux/selinux-policy/blob/7f644e53bc14de65d4546db012efe21c98244ef4/policy/modules/system/systemd.if#L1563) so I would just reference that in the optional policy instead of the explicit rule. The `systemd_unit_file_type` includes the generated type so it should work.

keitwb (approved):

PR #2829, opened 2026-04-13
EDM-3683: Allow dots in configuration provider names

galel12:
## Problem

Configuration names are validated using DNS 1123 label format (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`), which only allows lowercase letters, numbers, and hyphens. Names like `example.json`, `config.yaml`, or `example.env` are rejected, despite the UI indicating dots are allowed.

## Root Cause

All 4 config provider validators (`GitConfigProviderSpec`, `InlineConfigProviderSpec`, `HttpConfigProviderSpec`, `KubernetesSecretProviderSpec`) use `ValidateGenericName` for the `Name` field, which enforces DNS 1123 label format — no dots.

## Fix

1. Added `ValidateConfigName` with regex `[a-z0-9]([-a-z0-9.]*[a-z0-9])?` that allows dots while still requiring alphanumeric start/end characters.
2. Updated the 4 config provider `Validate` methods to use `ValidateConfigName` for the `Name` field only.
3. `ValidateGenericName` remains unchanged — it is still used for Kubernetes secret names, namespaces, Helm namespaces, and volume names where DNS 1123 label format is correct.

## Testing

- Unit tests added for `ValidateConfigName` covering valid names (`example.json`, `config.yaml`, `my.config.v2`) and invalid names (leading/trailing dots, dashes, uppercase, underscores, spaces, length overflow).
- All existing tests pass with zero regressions.
- E2E verified: applied a fleet with `name: example.json` config — API returned `201 Created`, fleet status `Valid`.

## Summary by CodeRabbit

* **Improvements**
  - Enhanced configuration name validation:
    - Now supports dot-separated naming patterns (e.g., "example.json", "my-config")
    - Maximum name length enforced at 253 characters
    - Names must start and end with alphanumeric characters

### DIFF
--- a/internal/util/validation/formats.go
+++ b/internal/util/validation/formats.go
@@ -16,6 +16,12 @@ const (
 	DNS1123MaxLength      int    = 253
 	envVarNameFmt         string = `[A-Za-z_][A-Za-z0-9_]*`
 
+	// ConfigNameFmt extends the DNS 1123 label format to also allow dots,
+	// supporting config names like "example.json" or "config.yaml".
+	// Must start and end with an alphanumeric character.
+	ConfigNameFmt       string = `[a-z0-9]([-a-z0-9.]*[a-z0-9])?`
+	configNameMaxLength int    = 253
+
 	// HostnameOrFQDNFmt validates a hostname or FQDN (Fully Qualified Domain Name).
 	// A hostname is a single DNS label, an FQDN is one or more labels separated by dots.
 	// Each label follows DNS rules: lowercase alphanumerics and hyphens, start and end with alphanumeric.
@@ -44,6 +50,7 @@ const (
 
 var (
 	GenericNameRegexp                    = regexp.MustCompile("^" + Dns1123LabelFmt + "$")
+	ConfigNameRegexp                     = regexp.MustCompile("^" + ConfigNameFmt + "$")
 	EnvVarNameRegexp                     = regexp.MustCompile("^" + envVarNameFmt + "$")
 	HostnameOrFQDNRegexp                 = regexp.MustCompile("^" + HostnameOrFQDNFmt + "$")
 	HostnameOrFQDNWithOptionalPortRegexp = regexp.MustCompile("^" + HostnameOrFQDNWithOptionalPortFmt + "$")
@@ -52,6 +59,13 @@ var (
 
 func ValidateGenericName(name *string, path string) []error {
 	return ValidateString(name, path, 1, dns1123LabelMaxLength, GenericNameRegexp, Dns1123LabelFmt)
+}
+
+// ValidateConfigName validates a configuration provider name.
+// It allows dots in addition to the characters permitted by ValidateGenericName,
+// supporting names like "example.json" or "config.yaml".
+func ValidateConfigName(name *string, path string) []error {
+	return ValidateString(name, path, 1, configNameMaxLength, ConfigNameRegexp, ConfigNameFmt, "example.json", "my-config")
 }
 
 func ValidateHostnameOrFQDN(name *string, path string) []error {
--- a/api/core/v1beta1/validation.go
+++ b/api/core/v1beta1/validation.go
@@ -342,7 +342,7 @@ func (r ResourceAlertRule) Validate(specSampleInterval string) []error {
 
 func (c GitConfigProviderSpec) Validate(fleetTemplate bool) []error {
 	allErrs := []error{}
-	allErrs = append(allErrs, validation.ValidateGenericName(&c.Name, "spec.config[].name")...)
+	allErrs = append(allErrs, validation.ValidateConfigName(&c.Name, "spec.config[].name")...)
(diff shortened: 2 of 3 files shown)


avishayt (approved):
Approved, but the UI validation needs to be updated to match (@celdrake). Currently the UI uses `GENERIC_NAME_REGEXP` in `/libs/ui-components/src/components/form/validations.ts` which rejects dots. We should add a         
  `CONFIG_NAME_REGEXP` pattern (matching the backend's `ConfigNameFmt`) and use it in `validConfigTemplatesSchema` so users don't get blocked in the UI.

celdrake:
> Approved, but the UI validation needs to be updated to match (@celdrake). Currently the UI uses `GENERIC_NAME_REGEXP` in `/libs/ui-components/src/components/form/validations.ts` which rejects dots. We should add a `CONFIG_NAME_REGEXP` pattern (matching the backend's `ConfigNameFmt`) and use it in `validConfigTemplatesSchema` so users don't get blocked in the UI.

We just changed this behaviour in the UI after EDM-3544. So we can align either way. Maybe disabling the dot it better so that users don't think it's the file name instead of the config name?

**To clarify**:  after the UI PR that's already merged, the UI and Backend are aligned prior to this PR).

avishayt:
> > Approved, but the UI validation needs to be updated to match (@celdrake). Currently the UI uses `GENERIC_NAME_REGEXP` in `/libs/ui-components/src/components/form/validations.ts` which rejects dots. We should add a `CONFIG_NAME_REGEXP` pattern (matching the backend's `ConfigNameFmt`) and use it in `validConfigTemplatesSchema` so users don't get blocked in the UI.
> 
> We just changed this behaviour in the UI after EDM-3544. So we can align either way. Maybe disabling the dot it better so that users don't think it's the file name instead of the config name?
> 
> **To clarify**: after the UI PR that's already merged, the UI and Backend are aligned prior to this PR).

I understand the concern but I'm OK making it more flexible if users are asking for it.  Thanks for the clarification!

celdrake:
> > > Approved, but the UI validation needs to be updated to match (@celdrake). Currently the UI uses `GENERIC_NAME_REGEXP` in `/libs/ui-components/src/components/form/validations.ts` which rejects dots. We should add a `CONFIG_NAME_REGEXP` pattern (matching the backend's `ConfigNameFmt`) and use it in `validConfigTemplatesSchema` so users don't get blocked in the UI.
> > 
> > 
> > We just changed this behaviour in the UI after EDM-3544. So we can align either way. Maybe disabling the dot it better so that users don't think it's the file name instead of the config name?
> > **To clarify**: after the UI PR that's already merged, the UI and Backend are aligned prior to this PR).
> 
> I understand the concern but I'm OK making it more flexible if users are asking for it. Thanks for the clarification!

Sound reasonable. I'll do a follow-up on EDM-3544 to align back accepting the dot.

eldar101 (approved):

internal/util/validation/formats.go
```diff
@@ -16,6 +16,12 @@ const (
 	DNS1123MaxLength      int    = 253
 	envVarNameFmt         string = `[A-Za-z_][A-Za-z0-9_]*`
 
+	// ConfigNameFmt extends the DNS 1123 label format to also allow dots,
+	// supporting config names like "example.json" or "config.yaml".
+	// Must start and end with an alphanumeric character.
+	ConfigNameFmt       string = `[a-z0-9]([-a-z0-9.]*[a-z0-9])?`
+	configNameMaxLength int    = 253
```
celdrake, line 23: Was it intentional to increase from the previous 63 max allowed characters for config names to 253?

galel12 (commented):
@celdrake — Yes, it was intentional. The reasoning: since config names now allow dots, the format is structurally closer to an FQDN (multiple dot-separated segments) than a single DNS label, so 253 (the FQDN max) felt like the natural upper bound instead of 63 (the single-label max).

That said, config names aren't DNS names — they're just human-readable identifiers. 63 characters is already generous for realistic names like `example.json` or `my-app.config.yaml`. So there's an argument for keeping it at 63 to minimize the scope of this change.

What's your preference — keep 253 for the FQDN analogy, or revert to 63 to stay conservative?

internal/util/validation/formats.go
galel12: @celdrake — Yes, it was intentional. The reasoning: since config names now allow dots, the format is structurally closer to an FQDN (multiple dot-separated segments) than a single DNS label, so 253 (the FQDN max) felt like the natural upper bound instead of 63 (the single-label max).

That said, config names aren't DNS names — they're just human-readable identifiers. 63 characters is already generous for realistic names like `example.json` or `my-app.config.yaml`. So there's an argument for keeping it at 63 to minimize the scope of this change.

What's your preference — keep 253 for the FQDN analogy, or revert to 63 to stay conservative?

celdrake: Thanks for the explanation.

Perhaps 253 may be too generous, since in this particular case it refers to a file name so segments are not needed in the sense of "file paths", but rather more like "base name"."extension".

PR #2837, opened 2026-04-15
EDM-3695: Validate persisted CSR CN matches device identity on startup

kkyrazis:
## Summary
- Adds `ValidateCSRIdentity` to detect when a persisted CSR's Subject CN does not match the current device name, indicating inconsistent identity state (e.g., keys regenerated but stale CSR left on disk)
- On mismatch the agent fails with a clear error instead of silently submitting a doomed enrollment request
- Handles both PEM (file-based) and TCG-CSR-IDEVID (TPM) CSR formats

Note: This was found with e2e tests which do a lot of manual cleaning. It appears that occasionally a test would fail to clear the CSR. When a new identity was generated, the old CSR would be used causing a mismatch. We now crash the agent if this behavior is detected, rather than submit an invalid ER that can't be approved.

## Test plan
- [x] Unit tests for `ValidateCSRIdentity` covering matching CN, mismatched CN, empty CN, empty/invalid CSR bytes
- [x] Manual VM test: file-based identity — deleted `agent.key`, agent correctly refused to start with stale CSR
- [x] Manual VM test: TPM identity — deleted `tpm-blob.yaml`, agent correctly refused to start with stale CSR
- [x] Verified agent starts normally after removing the stale CSR

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -154,6 +154,9 @@ func (a *Agent) Run(ctx context.Context) error {
 		}
 		a.log.Infof("CSR generated and persisted successfully")
 	} else {
+		if err := identity.ValidateCSRIdentity(csr, deviceName); err != nil {
+			return fmt.Errorf("persisted CSR identity mismatch: %w", err)
+		}
 		a.log.Infof("Using persisted CSR for enrollment")
 	}
 
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -134,30 +134,9 @@ func (a *Agent) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to get device name: %w", err)
 	}
 
-	clientCSRPath := identity.GetCSRPath(a.config.DataDir)
-
-	// Try to load persisted CSR first, generate a new one only if not found
-	csr, found, err := identity.LoadCSR(rootReadWriter, clientCSRPath)
+	csr, err := identity.ResolveCSR(rootReadWriter, a.config.DataDir, identityProvider, deviceName, a.log)
 	if err != nil {
-		return fmt.Errorf("failed to load CSR: %w", err)
-	}
-
-	if !found {
-		a.log.Infof("No persisted CSR found, generating new CSR for enrollment")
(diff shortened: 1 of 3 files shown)


internal/agent/identity/identity.go
```diff
@@ -150,6 +150,27 @@ func StoreCSR(rw fileio.ReadWriter, csrPath string, csr []byte) error {
...
+	}
+
+	parsed, err := fccrypto.ParseCSR(standardCSR)
+	if err != nil {
+		return fmt.Errorf("parsing CSR: %w", err)
+	}
+
+	if parsed.Subject.CommonName != deviceName {
+		return fmt.Errorf("persisted CSR CN %q does not match device name %q", parsed.Subject.CommonName, deviceName)
+	}
```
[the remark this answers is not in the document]
kkyrazis: this isn't a device status condition, it's a fatal bootstrap error logged to the journal

internal/agent/agent.go
```diff
@@ -154,6 +154,9 @@ func (a *Agent) Run(ctx context.Context) error {
 		}
 		a.log.Infof("CSR generated and persisted successfully")
 	} else {
+		if err := identity.ValidateCSRIdentity(csr, deviceName); err != nil {
```
asafbennatan: @kkyrazis what happens to the device , will it require manual intervention or will the offending CSR be deleted somewhere , so the device can create a new ER ?

kkyrazis: I left this as requiring manual intervention. To get into this scenario the user is likely manipulating the device in an unsupported way (like the e2e tests do).

asafbennatan: in the agent code we delete both agent.csr and tpm-blob.yaml , there could be a disk corruption situation which causes one to be deleted and not the other , in this case the device is stuck and requires manual intervention - i think we should attempt to delete the offending CSR (even though the scenario i have mentioned is unlikely - we should be careful about moving the agent into an unrecoverable state without a good reason)

asafbennatan (approved):
LGTM

PR #2839, opened 2026-04-16
EDM-3692: Fix helm upgrade regenerating CA certs, breaking agent trust

asafbss:
The cert-generator Job mounts /tmp/certs as an emptyDir, which is always empty on helm upgrade. This caused generate-certificates.sh to regenerate all certificates including the root CA, even though its "skip if exists" checks were designed to prevent this.

Fix by restoring existing certificates from K8s secrets into the working directory before generation runs, so the skip checks find the existing certs and preserve them. The restore logic is guarded by --create-k8s-secrets and has no effect on Quadlet deployments.

Secret existence is checked separately from data fetching so that "not found" (expected on fresh install) is handled gracefully while transient errors (network, RBAC) surface via set -e instead of silently causing CA regeneration.


asafbss:
Duplicate https://github.com/flightctl/flightctl/pull/2834

PR #2851, opened 2026-04-20
EDM-3512: Add tool field to cli download artifacts

celdrake:
Adds a `tool` field to the response of cliArtifacts. Since this is a new field, the result is backwards-compatible.

I restructured the script to make the flow more explicit so that when new CLI artifacts are added the `tool` will be populated automatically for consumers.

Co-authored by `Cursor`

Example output:

```
  {
      "os": "windows",
      "arch": "arm64",
      "filename": "flightctl-windows-arm64.zip",
      "sha256": "e9087be01f5f880e5894921c48e776c6ede4611a016cb13ce934dbb1a05a693f",
      "tool": "flightctl"
    },
    {
      "os": "linux",
      "arch": "amd64",
      "filename": "flightctl-restore-linux-amd64.tar.gz",
      "sha256": "999eb3f880dd0bb76f809e77b4042d57d2f1169b5ffe1f630038b96f18085d40",
      "tool": "flightctl-restore"
    },
```

## Summary by CodeRabbit

* **Chores**
  * Optimized the multi-architecture CLI build and artifact generation flow for more reliable, incremental builds and manifest output.
* **Bug Fixes**
  * Updated Console CLI download listings to add ARM64 Linux/macOS packages and standardize ARM64 label formatting (including Windows).

### DIFF
--- a/deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-console.yaml
+++ b/deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-console.yaml
@@ -14,16 +14,16 @@ spec:
   links:
     - href: '{{ $url }}/amd64/linux/flightctl-linux-amd64.tar.gz'
       text: Download flightctl for Linux for x86_64
+    - href: '{{ $url }}/arm64/linux/flightctl-linux-arm64.tar.gz'
+      text: Download flightctl for Linux for ARM64
     - href: '{{ $url }}/amd64/mac/flightctl-darwin-amd64.zip'
       text: Download flightctl for Mac for x86_64
+    - href: '{{ $url }}/arm64/mac/flightctl-darwin-arm64.zip'
+      text: Download flightctl for Mac for ARM64
     - href: '{{ $url }}/amd64/windows/flightctl-windows-amd64.zip'
       text: Download flightctl for Windows for x86_64
-    - href: '{{ $url }}/arm64/linux/flightctl-linux-arm64.tar.gz'
-      text: Download flightctl for Linux for ARM 64
-    - href: '{{ $url }}/arm64/mac/flightctl-darwin-arm64.zip'
-      text: Download flightctl for Mac for ARM 64
     - href: '{{ $url }}/arm64/windows/flightctl-windows-arm64.zip'
-      text: Download flightctl for Windows for ARM 64
+      text: Download flightctl for Windows for ARM64
 ---
 apiVersion: console.openshift.io/v1
 kind: ConsoleCLIDownload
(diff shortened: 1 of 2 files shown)


celdrake:
I'm going to make some additional changes to align the wording and other patterns between the UI and the Console CRD

deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-console.yaml
```diff
@@ -14,16 +14,16 @@ spec:
   links:
     - href: '{{ $url }}/amd64/linux/flightctl-linux-amd64.tar.gz'
       text: Download flightctl for Linux for x86_64
+    - href: '{{ $url }}/arm64/linux/flightctl-linux-arm64.tar.gz'
```
celdrake, line 17: We're unifying how the Links appear in the OCP console, with the rest of the tools.

The OpenShift console is sorting the links alphabetically, using case-sensitive comparison.

Therefore "ARM64" will appear before "x86_64", and the OS are in the order: "Linux, Mac, Windows".
Renaming "Mac" to "macOS" would make "macOS" appear as the last OS.

asafbennatan (approved):
LGTM

hack/build_multiarch_clis.sh
```diff
@@ -69,12 +74,29 @@ set -euo pipefail
...
-  local GOOS=$2
+# Build and archive one CLI tool for a specific platform
+build_platform_cli() {
+  local CLI=$1
+  local GOARCH=$2
+  local GOOS=$3
+  local make_target=$4
 
-  DISABLE_FIPS=true GOARCH="${GOARCH}" GOOS="${GOOS}" make build-cli build-restore
+  DISABLE_FIPS=true GOARCH="${GOARCH}" GOOS="${GOOS}" make "${make_target}"
```
Samudelacruz77, line 99: Is possible to make the cli for fips enabled servers?

celdrake: IMO the scope for this bug should not modify the current behaviour relative to FIPS enablement.

My understanding if that this is only possible for certain combinations of deployment types and OS systems, so we should clarify the requirements before doing any changes.

In any case, the CLI will [dynamically check](https://github.com/flightctl/flightctl/blob/main/internal/ssh/crypto.go#L32) if FIPS is enabled when it's used to generate crypto keys etc.

celdrake: cc @fzdarsky What do you think?

celdrake: Merged, we can have a follow-up if needed but it may not be low risk so best not to include it for 1.2

celdrake:
@amalykhi FYI we merged this change. 
Please follow the same pattern for the new CLI tools.
When the new CLI tools are ready, please let me know so I make sure these are displayed correctly in the UI.

PR #2860, opened 2026-04-21
NO-ISSUE: use testcontainers to run integration tests

asafbennatan:
1. uses testcontainers to control necessary services (db,kv,alert manager) to run integration testing (centralizes existing e2e testcontainers control logic to do that )
2. centralizes migration logic
3. integration tests use production migration logic (contrary to before where they were using alternative methods)
4. enabled integration-test parallelization 

## Summary by CodeRabbit

* **New Features**
  * Integration tests now start/stop Postgres, Redis and Alertmanager via dedicated make targets and a preflight CLI; help text and make targets updated.

* **Tests**
  * Integration tests run with an integration build tag and now use environment-aware helpers to discover published host/ports, hosts, ports and credentials.

* **Documentation**
  * Docs updated to describe the testcontainers-based workflow, runtime detection, and published-port endpoint discovery.

* **Chores**
  * Tooling and scripts detect Podman/Docker sockets, auto-configure the runtime, and migration scripts adjust DB role/privilege handling.


test/integration/integrationstack/stack.go
```diff
@@ -0,0 +1,237 @@
...
+		Cmd:          []string{"--config.file=/etc/alertmanager/alertmanager.yml", "--storage.path=/tmp/am"},
+		Files: []testcontainers.ContainerFile{
+			{HostFilePath: amPath, ContainerFilePath: "/etc/alertmanager/alertmanager.yml", FileMode: 0644},
+		},
+		WaitingFor: wait.ForHTTP("/-/ready").WithPort("9093/tcp").WithStartupTimeout(60 * time.Second),
+		SkipReaper: reuse,
+	}
+	if _, err := containers.GenericStart(ctx, amReq, reuse, containers.WithNetwork(network), containers.WithHostAccess()); err != nil {
+		return fmt.Errorf("alertmanager container: %w", err)
+	}
```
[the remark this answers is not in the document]
asafbennatan: once it becomes dynamic we will handle it , the addition of code to handle this non existent usecase is redundent

galel12 (commented):
Great refactor. 
few things I noticed below.

test/util/testdb/testdb.go
```diff
@@ -0,0 +1,283 @@
...
+	adminCfg.Database.Password = IntegrationPostgresAdminPassword()
+
+	adminDB, err := OpenMinimalAdminDB(adminCfg, log)
+	if err != nil {
+		return nil, fmt.Errorf("admin connection: %w", err)
+	}
+	defer CloseDB(adminDB)
+
+	log.Debugf("Creating test database from template: flightctl")
+	res := adminDB.WithContext(ctx).Exec(fmt.Sprintf(`CREATE DATABASE "%s" TEMPLATE flightctl;`, dbName))
```
galel12, line 195: `CREATE DATABASE ... TEMPLATE flightctl` requires zero active connections to the source database. With `--procs=4` (or 8 in CI), parallel suites will call `cloneFromMigratedDB` concurrently — if any suite still has a connection to `flightctl` (even briefly from `initDB` plugin init), another suite's clone will fail with `source database "flightctl" is being accessed by other users`.

Easiest fix: after migrations, mark it as a proper template so no one can connect to it:
```sql
ALTER DATABASE flightctl IS_TEMPLATE true;
UPDATE pg_database SET datallowconn = false WHERE datname = 'flightctl';
```
Or create a separate `flightctl_tmpl` and clone from that instead.

test/util/testdb/testdb.go
```diff
@@ -0,0 +1,283 @@
...
+		return "", 0, "", nil, fmt.Errorf("starting Redis container: %w", err)
+	}
+
+	mapped, err := c.MappedPort(ctx, "6379/tcp")
+	if err != nil {
+		_ = c.Terminate(ctx)
+		return "", 0, "", nil, fmt.Errorf("getting Redis mapped port: %w", err)
+	}
+
+	host = "127.0.0.1"
```
galel12, line 55: `host` is hardcoded to `127.0.0.1` — this won't work on Docker-in-Docker, remote daemons, etc. You're already using `c.MappedPort()`, might as well use `c.Host(ctx)` for the host too:
```go
host, err = c.Host(ctx)
if err != nil {
    _ = c.Terminate(ctx)
    return "", 0, "", nil, fmt.Errorf("getting Redis host: %%w", err)
}
```

test/integration/preflight/main.go
```diff
@@ -0,0 +1,45 @@
...
+		if err := integrationstack.EnsureContainersOnly(ctx); err != nil {
+			logrus.Fatalf("preflight start: %v", err)
+		}
+	case "stop":
+		if err := integrationstack.Stop(ctx); err != nil {
+			logrus.Fatalf("preflight stop: %v", err)
+		}
+	case "migrate":
+		// Run migrations using flightctl-db-migrate binary (same as production)
+		if err := integrationstack.RunMigrations(ctx); err != nil {
```
galel12, line 38: This calls `RunMigrations` directly, which skips the file lock. `RunMigrationsWithLock` exists specifically to prevent parallel suites from deadlocking on `CREATE INDEX`. If someone runs `preflight migrate` while tests are starting, they'll race.

```go
case "migrate":
    if err := integrationstack.RunMigrationsWithLock(ctx); err != nil {
```

test/test.mk
```diff
@@ -7,52 +7,50 @@ GO_UNITTEST_DIRS 		= ./internal/... ./api/... ./pkg/...
...
-	    run_integration || { $(MAKE) _collect_junit; exit 1; }; \
-	  fi'
-	$(MAKE) _collect_junit
+	@go install github.com/onsi/ginkgo/v2/ginkgo
+	@GOBIN=$$(go env GOBIN); \
+	if [ -z "$$GOBIN" ]; then GOBIN=$$(go env GOPATH)/bin; fi; \
+	count="$(INTEGRATION_TEST_COUNT)"; \
+	repeat_flag=""; \
+	if [ "$${count:-1}" -gt 1 ] 2>/dev/null; then repeat_flag="--repeat=$$((count - 1))"; fi; \
+	$$GOBIN/ginkgo run \
```
galel12, line 41: The old `gotestsum` invocation had `--coverprofile`. The new ginkgo invocation drops coverage collection entirely. Intentional? Ginkgo supports `--cover --coverprofile=integration-coverage.out` if you want to keep it.

test/util/testdb/testdb.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	// Terminate any remaining connections to the test database.
+	// As postgres admin, we can terminate any connection.
+	adminDB.WithContext(ctx).Exec(fmt.Sprintf(
+		"SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = '%s' AND pid <> pg_backend_pid();",
+		dbName))
+
+	// Drop the test database
+	adminDB = adminDB.WithContext(ctx).Exec(fmt.Sprintf("DROP DATABASE IF EXISTS %s;", dbName))
```
galel12, line 260: Nit: `cloneFromMigratedDB` quotes the DB name (`CREATE DATABASE "%s" TEMPLATE ...`) but this `DROP` doesn't. Safe today since names are UUID-based, but might as well be consistent:
```go
fmt.Sprintf(`DROP DATABASE IF EXISTS "%s";`, dbName)
```

galel12 (commented):
Great refactor. A few things I noticed below.

test/util/testdb/testdb.go
asafbennatan: Each test is connected to its own db , the flightctl db is cloned to achieve that , cloning is protected with a file lock (and a migration marker flag) so this can't happen

test/integration/preflight/main.go
asafbennatan: This is not a real scenario, no one should be running preflight while tests are running , the reason we are not using the lock is that for preflighting want this to always happen (run with lock also checks marker)

test/test.mk
asafbennatan: We do not need coverage for integration testing the metric is relevant only for unit testing

test/util/testdb/testdb.go
asafbennatan: Not relevant this is a test code , not production code receiving it from unknown source

test/util/testdb/testdb.go
galel12: AFAIK the file lock protects migrations (`RunMigrationsWithLock`), not the `CREATE DATABASE ... TEMPLATE` call here in `cloneFromMigratedDB`. Each parallel ginkgo proc calls `CreateTestDB` → `cloneFromMigratedDB` → `CREATE DATABASE "%s" TEMPLATE flightctl` without any lock. The PostgreSQL constraint is that `CREATE DATABASE ... TEMPLATE X` fails if there are *any* active connections to database X at that moment — even a brief connection from another proc's `initDB` would cause it. The file lock won't help here since it's not acquired in this code path.

Easy to verify: run with `--procs=8` a few times and see if you get `source database "flightctl" is being accessed by other users`.

asafbennatan: got you , i still think this is not something that happens in practice ( P.S integration tests are ran with proc 8 and i have constantly tested it with proc 8) reason being :
1. flightctldb is created and migrated once - this is promised by the flock
2. each test creates a clone of flightctl -> this is not an access to flightctl so this will be fine when ran concurrently

test/util/testdb/testdb.go
asafbennatan: c.Host(ctx) often returns "localhost", which can resolve to ::1 (IPv6) on some systems, causing connection failures when the container only binds to IPv4.

we have taken the same approach in our e2e testing following actual issues i have seen with the normal approach

test/harness/test_harness.go
```diff
@@ -324,6 +388,10 @@ func (h *TestHarness) Cleanup() {
 	h.StopAgent()
 	// stop any pending API requests
 	h.cancelCtx()
+	// wait for all servers to finish gracefully
+	if h.serversFinished != nil {
+		<-h.serversFinished
```
galel12: This blocks forever if any server goroutine hangs during shutdown. Might be worth adding a `select` with a timeout (say 30s) and a warning log so tests don't freeze silently.

Also related — the old code called `cancel()` at the end of each server goroutine, so if one server crashed mid-test the context cancellation would cascade to the others and you'd get a fast failure. Now that's removed, so a crashed server just logs via `goRoutineErrorHandler` (which doesn't actually fail the test — `GinkgoRecover()` is a no-op without a preceding panic) and the other servers keep running until `Cleanup()`. Could lead to confusing test timeouts instead of clear crash reports.

test/scripts/run_migration.sh
```diff
@@ -11,22 +11,137 @@ MIGRATION_IMAGE=${MIGRATION_IMAGE:-localhost/flightctl-db-setup:latest}
...
 DB_MIGRATION_PASSWORD=${FLIGHTCTL_POSTGRESQL_MIGRATOR_PASSWORD:-adminpass}
 
+# When DOCKER_HOST is unset, detect the container socket (same order as
+# test/harness/containers.detectContainerSocket) and set DOCKER_HOST so docker|podman
+# CLI matches where testcontainers started integration Postgres.
+is_container_socket() {
+    [[ -e "$1" ]] && [[ -S "$1" ]]
+}
+
+detect_container_socket() {
```
galel12: This is a bash reimplementation of `test/harness/containers.detectContainerSocket()` — same ordered probe list (XDG, /run/user, ~/.local, docker.sock, system podman). If the Go code ever changes the probe order, this won't follow. Might be worth having the preflight binary emit the detected socket/CLI, or at least a prominent comment cross-referencing the Go source so they stay in sync.

test/integration/integrationstack/stack.go
```diff
@@ -0,0 +1,560 @@
...
+// The process is idempotent - users are created only if they don't exist, and migrations use
+// the schema_migrations table to skip already-applied migrations.
+// Requires: psql (postgresql-client) and envsubst.
+func RunMigrations(ctx context.Context) error {
+	h, p, ok := PublishedTCPPort(PostgresContainerName, "5432/tcp")
+	if !ok {
+		return fmt.Errorf("postgres container %q is not running or has no published port 5432/tcp", PostgresContainerName)
+	}
+
+	if _, err := exec.LookPath("psql"); err != nil {
```
galel12: Previously migrations ran inside the `flightctl-db-setup` container (which bundles psql). Now they shell out to `psql` and `envsubst` on the host, which adds implicit system-level dependencies that aren't in `go.mod`. The CI setup action (`setup-dependencies/action.yaml`) doesn't explicitly install `postgresql-client` or `gettext` — it works because GHA Ubuntu runners happen to ship them, but that's fragile. Might surprise devs running `make integration-test` for the first time on a fresh machine too.

test/integration/tasks/repotester_test.go
```diff
@@ -42,78 +42,39 @@ func newOciAuth(username, password string) *api.OciAuth {
...
-	require.NoError(err)
-}
-
-func startSshGitRepo(pubKey gossh.PublicKey, require *require.Assertions) {
+func startSshGitRepo(pubKey gossh.PublicKey) {
 	ssh.Handle(func(s ssh.Session) {
 		path, err := filepath.Abs("./git/base")
-		require.NoError(err)
+		if err != nil {
+			return
```
galel12: The old `require.NoError` would fail the test immediately and point straight at the root cause. Now errors are silently swallowed — if the test data file is missing or unreadable, the SSH session just closes with no data and the test fails downstream with a confusing EOF or similar. The HTTPS counterpart at least returns `http.Error(w, err.Error(), 500)`. Consider `GinkgoWriter.Printf` or `Fail()` here so broken test infra is visible.

test/integration/integrationstack/stack.go
asafbennatan: you are right , the reason i removed it is because i didnt want stale flightctl-db-setup to fail the tests .
i will solve this by running psql from the db container itself

galel12:
@asafbennatan One thing I noticed: a few docs outside `test/AGENTS.md` and `deploy/AGENTS.md` still reference the old integration test flow and would be stale after this merges:

- **`test/README.md`** — has sections on `FLIGHTCTL_TEST_DB_STRATEGY=template|local`, `MIGRATION_IMAGE`, and the old `make deploy-db`/`deploy-kv` flow. Also mentions merging unit + integration coverage which is removed in this PR.
- **`AGENTS.md`** (root) — lines 25 and 54 say `make integration-test` "starts DB/KV/alertmanager via deploy targets" — it's testcontainers now.
- **`docs/developer/service-quadlets.md`** — mentions `make deploy-db` / `make deploy-kv` being used for integration testing (lines 208-214).
- **`docs/developer/README.md`** — mentions unit tests but has no mention of integration tests at all. A short section with `make integration-test` + the new `psql`/`gettext` prereqs would help devs setting up a fresh machine.

`test/README.md` is the most impactful one since it has full paragraphs that no longer apply.

asafbennatan:
> @asafbennatan One thing I noticed: a few docs outside `test/AGENTS.md` and `deploy/AGENTS.md` still reference the old integration test flow and would be stale after this merges:
> 
> * **`test/README.md`** — has sections on `FLIGHTCTL_TEST_DB_STRATEGY=template|local`, `MIGRATION_IMAGE`, and the old `make deploy-db`/`deploy-kv` flow. Also mentions merging unit + integration coverage which is removed in this PR.
> * **`AGENTS.md`** (root) — lines 25 and 54 say `make integration-test` "starts DB/KV/alertmanager via deploy targets" — it's testcontainers now.
> * **`docs/developer/service-quadlets.md`** — mentions `make deploy-db` / `make deploy-kv` being used for integration testing (lines 208-214).
> * **`docs/developer/README.md`** — mentions unit tests but has no mention of integration tests at all. A short section with `make integration-test` + the new `psql`/`gettext` prereqs would help devs setting up a fresh machine.
> 
> `test/README.md` is the most impactful one since it has full paragraphs that no longer apply.

fixed

test/integration/integrationstack/stack.go
```diff
@@ -0,0 +1,546 @@
...
+		return "", false
+	}
+
+	masterPW := envOrDefault("FLIGHTCTL_POSTGRESQL_MASTER_PASSWORD", defaultIntegrationPassword)
+
+	sub, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	//nolint:gosec // G204: arguments are from controlled integration test environment
+	cmd := exec.CommandContext(sub, "psql",
```
galel12: The migration itself no longer needs host psql (nice fix with podman exec), but this function still shells out to host-side `psql` for the sentinel check. On CI `psql` isn't installed, so `getFlightctlDatabaseID` silently fails every time → sentinel is never created → every parallel suite re-runs migrations through the lock, defeating the optimization.

Could use the same `podman exec` approach as `runSetupDatabaseUsersInContainer`, or a quick Go DB connection since GORM is already in the package.

test/util/testdb/testdb.go
```diff
@@ -0,0 +1,283 @@
...
+	defer span.End()
+
+	cfg := config.NewDefault()
+	ApplyIntegrationConnectionOverrides(cfg)
+	randomDBName := generateRandomDBName(prefix)
+	log.Debugf("Test DB name: %s", randomDBName)
+
+	gormDb, err := cloneFromMigratedDB(ctx, cfg, randomDBName, log, initDB)
+	if err != nil {
+		log.Fatal(err)
```
galel12: Several functions here use `log.Fatal`/`Fatalf` (this line, plus lines 110, 114, 128, 149, 225, 262). Since these run inside Ginkgo, `os.Exit(1)` skips all deferred cleanup — ephemeral Redis containers won't terminate, DB connections stay open. For example this `CreateTestDB` is called from `NewTestHarness` which is designed to return errors, but `log.Fatal` short-circuits that.

Returning errors or using Ginkgo's `Fail()` would let the framework handle cleanup properly.

galel12 (approved):
LGTM

PR #2863, opened 2026-04-23
EDM-3766: Fix for 'Alertmanager proxy denies authenticated user without alerts' standalone failure

eldar101:
- Updated the alertmanager proxy authz-denied e2e test to stop relying on pre-seeded non-admin org membership.
- The test now creates a temporary organization and grants demouser1 access to it before attempting flightctl login.
- Added a retry loop around non-admin login plus explicit selection of the temporary organization in the client config.
- Added cleanup to restore admin login and remove the temporary organization after the test (like @siserafin requested in previous PR)

This makes the test validate alert permission denial instead of failing earlier with “no organizations” in standalone/OCP standalone profiles.

## Summary by CodeRabbit

* **Tests**
  * Enhanced RBAC authorization testing for non-admin user scenarios with improved retry logic and organization provisioning.


eldar101:
Updating with successful run soon.

siserafin (approved):
lgtm

PR #2864, opened 2026-04-23
EDM-3767: CI fix - VM Agent behavior Resources Alert Validation Rules [78853]

eldar101:
The failure was in the agent e2e test, not in validation itself: standalone OCP returned the 400 error as a JSON Status body, and Harness.UpdateFleet() turned that raw JSON into the error string.

Because of that, regex text like \d appeared escaped as \\d, so the test’s substring assertion failed only in standalone even though the API was correctly rejecting the invalid config.

Changes:
1) I changed test/harness/e2e/harness_fleet.go so UpdateFleet() now uses the parsed API status message (JSON400.Message, etc.) when available instead of the raw response body.
2) I then simplified the two affected assertions in test/e2e/agent/agent_test.go back to matching the plain validation message.

The test now checks the actual API validation message consistently across environments, instead of depending on whether the transport body was JSON-escaped.

## Summary by CodeRabbit

* **Tests**
  * Updated test assertions to properly validate sampling interval error messages
  * Improved test infrastructure error handling to use structured error messages from HTTP responses, replacing generic response body fallbacks

### DIFF
--- a/test/harness/e2e/harness_fleet.go
+++ b/test/harness/e2e/harness_fleet.go
@@ -272,7 +272,28 @@ func (h *Harness) UpdateFleet(fleetName string, updateFunc func(*v1beta1.Fleet))
 	if replaceResp.StatusCode() != 200 {
 		logrus.Errorf("Unexpected http status code received: %d", replaceResp.StatusCode())
 		logrus.Errorf("Unexpected http response: %s", string(replaceResp.Body))
-		return fmt.Errorf("unexpected status code %d: %s", replaceResp.StatusCode(), string(replaceResp.Body))
+
+		statusMessage := ""
+		switch {
+		case replaceResp.JSON400 != nil:
+			statusMessage = replaceResp.JSON400.Message
+		case replaceResp.JSON401 != nil:
+			statusMessage = replaceResp.JSON401.Message
+		case replaceResp.JSON403 != nil:
+			statusMessage = replaceResp.JSON403.Message
+		case replaceResp.JSON404 != nil:
+			statusMessage = replaceResp.JSON404.Message
+		case replaceResp.JSON409 != nil:
+			statusMessage = replaceResp.JSON409.Message
+		case replaceResp.JSON429 != nil:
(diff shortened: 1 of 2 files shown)


test/harness/e2e/harness_fleet.go
```diff
@@ -272,6 +272,28 @@ func (h *Harness) UpdateFleet(fleetName string, updateFunc func(*v1beta1.Fleet))
...
+			statusMessage = replaceResp.JSON429.Message
+		case replaceResp.JSON503 != nil:
+			statusMessage = replaceResp.JSON503.Message
+		}
+
+		if statusMessage != "" {
+			return fmt.Errorf("unexpected status code %d: %s", replaceResp.StatusCode(), statusMessage)
+		}
 		return fmt.Errorf("unexpected status code %d: %s", replaceResp.StatusCode(), string(replaceResp.Body))
 	}
```
amalykhi, line 297:
```suggestion
default:
		statusMessage = string(resp.Body)
	}

	return fmt.Errorf("unexpected status code %d: %s", code, statusMessage
```

siserafin (approved):
lgtm.

test/harness/e2e/harness_fleet.go
```diff
@@ -272,9 +272,29 @@ func (h *Harness) UpdateFleet(fleetName string, updateFunc func(*v1beta1.Fleet))
...
+		case replaceResp.JSON404 != nil:
+			statusMessage = replaceResp.JSON404.Message
+		case replaceResp.JSON409 != nil:
+			statusMessage = replaceResp.JSON409.Message
+		case replaceResp.JSON429 != nil:
+			statusMessage = replaceResp.JSON429.Message
+		case replaceResp.JSON503 != nil:
+			statusMessage = replaceResp.JSON503.Message
+default:
+		statusMessage = string(resp.Body)
```
amalykhi:
```suggestion
		default:
			statusMessage = string(resp.Body)
```

amalykhi (approved):

PR #2868, opened 2026-04-23
EDM-3768: UI connects to alertmanager-proxy via internal service URL

celdrake:
Instead of modifying `getAlertManagerProxyUrl`, I chose the same pattern that for other services where we have a "getInternal...Url".

Now `getAlertManagerProxyUrl` would not be used anymore, we can either keep it for future reference, or delete it and bring it back if needed.

## Release Notes

* **Chores**
  * Helm deployment templates updated to configure Alertmanager proxy routing through internal infrastructure endpoints instead of external proxy services
  * Proxy endpoint configuration modified to route monitoring and alerting service communications through internal infrastructure components
  * Deployment behavior changed for how alert management integrations establish connections across the platform


asafbennatan (approved):
LGTM

PR #2874, opened 2026-04-26
NO-ISSUE: Fix flaky greenboot rollback e2e test

galel12:
## Problem
The `Should trigger greenboot rollback when agent fails to start` e2e test flakes intermittently with two distinct failures.

## Root Cause
1. **`version: 0: invalid rendered version`** — After greenboot rolls back from v11, the agent re-enrolls from scratch (new CSR, fresh bootstrap + spec rollback). `PrepareNextDeviceVersionFromCurrentStatus` did a single API read of the rendered version, which races with re-enrollment — the server hasn't assigned a rendered spec yet so the version is still `0`.
2. **`resource not updated`** — The boot ID captured during rollback detection can go stale if greenboot-rs (0.16.x) does extra boot cycles after the initial rollback. `EnsureDeviceContents` then fails because the device has a different boot ID than expected.

## Solution
1. Replace the one-shot read in `PrepareNextDeviceVersionFromCurrentStatus` with a `WaitForDeviceContents` polling loop that retries until the rendered version becomes valid (> 0).
2. Wait for Online + correct image without pinning a boot ID, then capture a fresh boot ID before the 2-minute stability assertion.

## Test Plan
- [x] `go build ./test/e2e/agent/...` compiles cleanly
- [x] `go vet ./test/e2e/agent/...` passes
- [ ] CI e2e sanity agent suite passes (greenboot-rollback label)

## Summary by CodeRabbit

* **Tests**
  * Improved test harness device version preparation to use polling-based detection instead of a single API read for more reliable version confirmation
  * Enhanced rollback stabilization test to wait for device online status before capturing stable boot state for validation

### DIFF
--- a/test/e2e/agent/agent_update_test.go
+++ b/test/e2e/agent/agent_update_test.go
@@ -275,16 +275,23 @@ var _ = Describe("VM Agent behavior during updates", Label("agent-update"), func
 			initialStatusImage, postRollbackBootID := waitForGreenbootOSRollbackFromV11BrokenAgent(harness, deviceId, false)
 
 			By("Verifying device does NOT retry the failed v11 image")
-			// Wait for the device to stabilize as Online after rollback. With greenboot-rs
-			// (0.16.x), the device may do an extra boot cycle before rolling back, causing
-			// a longer disconnect that temporarily shows as Unknown.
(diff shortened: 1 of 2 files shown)


asafbennatan (approved):
LGTM

asafbennatan (approved):
Lgtm

PR #2878, opened 2026-04-26
EDM-3781: Fix devices stuck in AwaitingReconnect after backup/restore

galel12:
## Problem

After a database backup and restore, devices whose rendered version matches the server (the common case — most devices are not mid-update) can get stuck in `AwaitingReconnect` status instead of transitioning to `Online`.

## Root Cause

`processAwaitingReconnectIfNeeded` returned `wasConflictPaused` (false for non-conflicted devices) instead of `true` (annotation was processed). This caused `GetRenderedDevice` to enter the `WaitForNewVersion` long-poll (120s → HTTP 204) instead of immediately returning HTTP 200. Without a 200 response, a concurrent agent status heartbeat could race with the annotation removal and overwrite the device status back to `AwaitingReconnect`, leaving it stuck.

## Solution

Return `true` from `processAwaitingReconnectIfNeeded` whenever the annotation was successfully processed, regardless of whether the device ended up ConflictPaused or Online. This skips the long-poll for all reconnecting devices so the agent gets a 200 and re-pushes its status.

## Test Plan

- [x] Integration test updated (was asserting old buggy 204 behavior, now correctly expects 200)
- [x] `make integration-test` passes (service suite including AwaitingReconnect tests)
- [x] E2e backup_restore test (label `84938`) should no longer flake

### DIFF
--- a/internal/service/device.go
+++ b/internal/service/device.go
@@ -328,11 +328,11 @@ func (h *ServiceHandler) PatchDeviceStatus(ctx context.Context, orgId uuid.UUID,
 
 func (h *ServiceHandler) GetRenderedDevice(ctx context.Context, orgId uuid.UUID, name string, params domain.GetRenderedDeviceParams) (*domain.Device, domain.Status) {
 	var (
-		isNew                 bool
-		kvRenderedVersion     string
-		err                   error
-		isAgent               bool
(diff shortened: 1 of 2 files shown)


asafbennatan (approved):
LGTM

PR #2882, opened 2026-04-27
EDM-3785: Service port/TLS config flexibility

keitwb:
This allows us to fully configure the listen address and TLS usage among all of the services we want to put behind a gateway service. We won't need TLS listening for those services in that scenario and we also shouldn't use tls-like ports (e.g. 8443) in those situations either since it is misleading.

This is backwards compatible in that the default will still be to enable TLS.


asafbennatan:
what about imagebuilder-api ?

keitwb:
> what about imagebuilder-api ?

That service already allows configuring the port with the `imageBuilderService.address` field and it will automatically run without TLS if you just don't specify the cert/key so it should be good to go.

keitwb:
This got included in #2962

PR #2889, opened 2026-04-29
EDM-3724: additional vulnerabilities apis

asafbennatan:
## Summary by CodeRabbit

* **New Features**
  * Org vulnerability summary with device counts by severity.
  * New endpoints: vulnerability summary, list vulnerabilities, and CVE impact.
  * CVE impact now reports per-digest findings and includes worst-severity, max CVSS score, and max published-at timestamps.
  * List endpoints use cursor-based pagination with continue tokens.
  * Field selector updated to support owner-based filtering (owner, = / == / !=, in / notin).

* **Bug Fixes**
  * Unknown CVEs return HTTP 200 with an empty affected-resources list (404 removed).
  * Disabled vulnerability feature returns HTTP 501 instead of 503.


internal/service/vulnerability_test.go
```diff
@@ -284,6 +284,90 @@ func TestGetDeviceVulnerabilitySummary(t *testing.T) {
...
+			expectedStatus: http.StatusOK,
+			assertResp: func(t *testing.T, resp *domain.VulnerabilityGroupList) {
+				require.Len(t, resp.Items, 2)
+				g := resp.Items[0]
+				require.Equal(t, "CVE-2024-0001", g.CveId)
+				require.NotNil(t, g.AffectedDevices)
+				// Two devices on digest2 (fleet + fleetless) + one on digest1 => 3 devices for this CVE.
+				require.Equal(t, int64(3), *g.AffectedDevices)
+				require.Len(t, g.Findings, 2)
+			},
```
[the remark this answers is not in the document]
asafbennatan: After reviewing the code, this finding is not valid and no fix is needed.

The test is not flaky because the ordering is deterministic:

Default sorting is applied: The DummyVulnerabilityFinding.GetVulnerabilityGroups method (lines 927-930 in teststore_framework_test.go) sorts the groups using sort.SliceStable with compareVulnerabilityGroupsBySortColumn.

Severity-based ordering is deterministic: The default sort is by severity descending (lines 1117-1127). The comparison function at lines 1185-1197 ensures:

Critical (rank 1) comes before Low (rank 4) when sorting DESC
CVE ID is used as a tiebreaker when severities are equal
Test data has different severities: Looking at the test setup (lines 330-335):

CVE-2024-0001 has severity "Critical" (rank 1)
CVE-2024-0002 has severity "Low" (rank 4)
Since Critical (rank 1) always sorts before Low (rank 4) in descending order, resp.Items[0] will always be CVE-2024-0001. The assertion at line 342 is correct and stable.

The real store implementation (GetVulnerabilityGroups in vulnerability_finding.go) also applies the same default sorting (lines 711-731), and the test store accurately mirrors this behavior.

api/core/v1alpha1/openapi.yaml
```diff
@@ -1546,7 +1546,7 @@ paths:
       parameters:
         - name: cveId
           in: path
-          description: The CVE identifier (for example, CVE-2024-1234).
+          description: The CVE identifier (for example, CVE-2024-1234). Unknown identifiers still return 200 with an empty blast-radius list when there is no matching data.
```
asafbss:
```suggestion
          description: The CVE identifier (for example, CVE-2024-1234). Unknown identifiers still return 200 with an empty list of affected resources (blast radius) when no matching data is found.
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
           description: Resource kind; always VulnerabilitySummary.
         cvesBySeverity:
           $ref: '#/components/schemas/CveCountsBySeverity'
+    DeviceCountsBySeverity:
+      type: object
+      description: Counts of distinct devices affected in the org, bucketed by finding severity (join of devices to vulnerability_findings).
```
asafbss: that description is too implementation-heavy and leaks internal details

```suggestion
      description: Counts of distinct devices affected in the organization, grouped by vulnerability severity.
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        - high
+        - medium
+        - low
+        - none
+        - unknown
+      properties:
+        total:
+          type: integer
+          format: int64
+          description: Distinct devices with at least one non-not_affected finding in scope.
```
asafbss:
```suggestion
          description: Number of distinct devices with at least one vulnerability finding in scope (excluding "not_affected").
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        critical:
+          type: integer
+          format: int64
+          description: Devices whose highest severity finding is critical.
```
asafbss:
```suggestion
          description: Number of devices whose highest severity finding is "critical".
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        high:
+          type: integer
+          format: int64
+          description: Devices whose highest severity finding is high.
```
asafbss:
```suggestion
          description: Number of devices whose highest severity finding is "high".
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        medium:
+          type: integer
+          format: int64
+          description: Devices whose highest severity finding is medium.
```
asafbss:
```suggestion
          description: Number of devices whose highest severity finding is "medium".
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        low:
+          type: integer
+          format: int64
+          description: Devices whose highest severity finding is low.
```
asafbss:
```suggestion
          description: Number of devices whose highest severity finding is "low".
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        none:
+          type: integer
+          format: int64
+          description: Devices whose highest severity finding is none.
```
asafbss:
```suggestion
          description: Number of devices with findings, but none with an assigned severity.
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2078,6 +2072,47 @@ components:
...
+        unknown:
+          type: integer
+          format: int64
+          description: Devices whose highest severity finding is unknown.
```
asafbss:
```suggestion
          description: Number of devices whose highest severity finding is "unknown".
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2427,17 +2462,18 @@ components:
         cveId:
           type: string
           description: CVE identifier for this blast radius response.
-        advisoryId:
-          type: string
-          description: Vendor advisory identifier when available.
         severity:
           type: string
-          description: Highest normalized severity for this CVE in the org.
+          description: Worst severity across all affected digests.
```
asafbss:
```suggestion
          description: Highest severity of this CVE across all affected resources in the organization.
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2427,17 +2462,18 @@ components:
...
           enum: [Critical, High, Medium, Low, None, Unknown]
-        cvssScore:
+        maxCvssScore:
           type: number
           format: float
-          description: CVSS base score when available.
+          description: Highest CVSS base score across all affected digests.
```
asafbss:
```suggestion
          description: Highest CVSS base score observed for this CVE across affected resources in the organization.
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2427,17 +2462,18 @@ components:
...
+        maxPublishedAt:
+          type: string
+          format: date-time
+          description: Latest advisory publish time across all affected digests.
```
asafbss:
```suggestion
          description: Most recent advisory publication time for this CVE across affected resources in the organization.
```

api/core/v1alpha1/openapi.yaml
```diff
@@ -2463,8 +2499,8 @@ components:
           type: integer
           format: int64
           description: Devices in this fleet or fleetless group affected by the CVE.
-        images:
+        findings:
```
asafbss, line 2504: images felt more user-facing, while “per-digest findings” seems to expose more implementation details and may be less clear to API consumers. Is the intention here to reflect a change in the underlying model, or just naming?

internal/service/vulnerability.go
```diff
@@ -18,6 +20,275 @@ func (h *ServiceHandler) featureDisabledStatus() domain.Status {
...
+		}
+	}
+
+	sortCol := store.SortByDeviceAffected
+	if params.SortBy != nil && *params.SortBy != "" {
+		switch *params.SortBy {
+		case "affectedDevices":
+			sortCol = store.SortByDeviceAffected
+		case "fleetName":
+			sortCol = store.SortByOwner
```
asafbss, line 83: One concern here around pagination + sorting:

Since we currently support sorting by a single field only (e.g. affectedDevices or fleetName), and these fields are not guaranteed to be unique—and may also contain NULL values—pagination with the continue token may become unstable.

If multiple rows share the same value (or NULL), the ordering between them is not deterministic, which could lead to duplicated or skipped results across pages.

@celdrake do you remember the problem we had with sorting?

internal/service/vulnerability.go
```diff
@@ -18,6 +20,275 @@ func (h *ServiceHandler) featureDisabledStatus() domain.Status {
...
+	sel, err := fieldselector.ParseSelector(strings.TrimSpace(*fieldSelectorStr))
+	if err != nil {
+		return nil, domain.StatusBadRequest(fmt.Sprintf("failed to parse fieldSelector: %v", err))
+	}
+	reqs, ok := sel.Requirements()
+	if !ok {
+		return fieldSelectorStr, domain.StatusOK()
+	}
+	for _, r := range reqs {
+		if r.Key().String() != "fleetName" {
```
asafbss: Do we need this custom validation here?
We already have selector infrastructure that handles supported field enforcement and validation.
https://github.com/flightctl/flightctl/blob/a9ea82103e980cfa9485275b6b70c22e9697cd21/internal/store/model/resource.go#L28

internal/service/vulnerability.go
```diff
@@ -18,6 +20,275 @@ func (h *ServiceHandler) featureDisabledStatus() domain.Status {
...
+			h.log.WithError(err).Error("failed to aggregate vulnerability impact")
+			return nil, domain.StatusInternalServerError("failed to aggregate vulnerability impact")
+		}
+	}
+
+	var details []store.ImpactDigestDetail
+	if len(pageResult.Rows) > 0 {
+		owners, hasFleetless := extractOwnersFromPage(pageResult.Rows)
+		var derr error
+		details, derr = h.store.VulnerabilityFinding().GetImpactDigestDetails(ctx, orgId, cveId, owners, hasFleetless)
```
asafbss, line 143: We first fetch the page (GetVulnerabilityImpactPage) and then issue a second query (GetImpactDigestDetails) to retrieve the details.

Is there a reason we’re not fetching this in a single query (e.g. via a join)?

internal/service/vulnerability.go
```diff
@@ -18,6 +20,275 @@ func (h *ServiceHandler) featureDisabledStatus() domain.Status {
...
+		if rows[i].Owner == nil {
+			hasFleetless = true
+			continue
+		}
+		owners = append(owners, *rows[i].Owner)
+	}
+	return owners, hasFleetless
+}
+
+func groupDetailsByOwner(details []store.ImpactDigestDetail) map[string][]*store.ImpactDigestDetail {
```
asafbss, line 201: Can we have the DB query return the impact rows already joined/aggregated with their digest details, instead of fetching details separately and grouping them by owner in Go?

asafbss (commented):
Looks good overall.
A few comments/questions below around some design and implementation details.

api/core/v1alpha1/openapi.yaml
asafbennatan: the structure changes , the problem with just having images that each digest could have different CVE details potentially

internal/service/vulnerability.go
asafbennatan: fleet name is unique in this context ( the data is grouped by fleetName )

internal/service/vulnerability.go
asafbennatan: i think this applies only for stuff extending Resource ? 
VulnerabilityFinding is not a resource since it is not org scoped (org scoping comes from the relevant devices we are looking at)

internal/service/vulnerability.go
asafbennatan: you will need two queries anyway as you need a count per group (number of effected devices)

note that we have two groups here :

top level group is the fleet
then for each entry we group by the digest ( CVE can have different details per digest and we also care how many devices are affected per digest in that fleet)

now , if try to use one query and group by fleet+digest we will get multiple top level entries per fleet - causing a problem with pagination . 

we could use a nested subquery or CTE but this query is more complicated and still does two logical passes of the data

internal/service/vulnerability.go
asafbennatan: we cannot we do the double query (reason mentioned above) , then we construct the page here

internal/service/vulnerability.go
asafbss: It’s not limited to types extending Resource.

For example, Checkpoint is not a Resource, but still uses selector tags:
https://github.com/flightctl/flightctl/blob/a9ea82103e980cfa9485275b6b70c22e9697cd21/internal/store/model/checkpoint.go#L15

internal/service/vulnerability.go
asafbss: Got it. So the expectation is 2 queries per request, not one query per row, right?

If so, I’m fine with the approach. I just want to make sure that we stay efficient at scale.

internal/service/vulnerability.go
asafbss: Got it for fleetName.
What about sorting by affectedDevices? Since this is just a count, multiple groups can have the same value.

internal/service/vulnerability.go
asafbennatan: correct - 2 per request - not per row

internal/service/vulnerability.go
asafbennatan: then we auto add the owner as an additional sort after it when creating the continue token , this way the continue is unique for a specific page entry

api/core/v1alpha1/openapi.yaml
asafbennatan: your suggestion is not 100% accurate as this is the first time it was observed in this specific fleet/fleetless entry - not the entire org

internal/store/vulnerability_finding.go
```diff
@@ -644,6 +644,8 @@ func (s *VulnerabilityFindingStore) GetImpactDigestDetails(
 }
 
 // vulnerabilityImpactResolver maps field selector keys to SQL for the devices alias `d` in impact aggregation.
+// The impact API fieldSelector supports only `owner` (for example owner=Fleet/my-fleet), not a separate fleetName key;
```
asafbss, line 647: FieldSelector already has built-in validation for unsupported fields (you can see this behavior with devices, for example).

I think we should try to rely on that mechanism and keep the integration consistent, rather than introducing custom validation logic here, if possible.

asafbennatan: ok i am now returning null for this, like in other places

asafbss (approved):

PR #2900, opened 2026-05-04
EDM-3818: fix 77672 CI failure in disconnected

eldar101:
The failure was that this test depended on flightctl-demos being reachable over HTTP. In disconnected envs it isn’t, so the first update never got picked up and the test just timed out waiting for a rendered-version change.

The fix is to make the test check whether that repo is actually accessible. If yes, it uses the original repo-backed update. If not, it uses an inline config instead, so the test still verifies “multiple updates, latest wins” without depending on internet access.

## Summary by CodeRabbit

* **Tests**
  * Enhanced the multi-update resolution test to dynamically handle repository accessibility by gracefully switching between HTTP and inline configuration fallback paths based on runtime conditions.

### DIFF
--- a/test/e2e/agent/agent_update_test.go
+++ b/test/e2e/agent/agent_update_test.go
@@ -184,17 +184,43 @@ var _ = Describe("VM Agent behavior during updates", Label("agent-update"), func
 				}
 			}()
 
-			// Add more factories here if desired. The first spec applied will add a repo spec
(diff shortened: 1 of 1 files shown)


siserafin (approved):
lgtm, in this test case we can use this workaround, another possibility would be to use the local git service and create a local repository.

PR #2901, opened 2026-05-04
EDM-3724: Add CVE id filter for device

asafbennatan:


internal/store/device.go
```diff
@@ -420,8 +426,84 @@ func (s *DeviceStore) GetWithTimestamp(ctx context.Context, orgId uuid.UUID, nam
 	return s.getWithTimestamp(ctx, orgId, name)
 }
 
-func (s *DeviceStore) List(ctx context.Context, orgId uuid.UUID, listParams ListParams) (*domain.DeviceList, error) {
-	return s.genericStore.List(ctx, orgId, listParams)
+func (s *DeviceStore) List(ctx context.Context, orgId uuid.UUID, listParams DeviceListParams) (*domain.DeviceList, error) {
```
avishayt, line 429: It's a shame to have to copy the implementation just for one filter that diverges.  Consider refactoring to allow the generic store to accept custom query modifiers

api/core/v1beta1/openapi.yaml
```diff
@@ -1083,6 +1083,12 @@ paths:
           required: false
           schema:
             type: boolean
+        - name: cveId
+          in: query
+          description: Filter devices by CVE ID. Only returns devices whose OS image digest has the specified vulnerability.
+          required: false
+          schema:
+            type: string
```
avishayt, line 1091: Should we add any validation here?  For example CVE-YYYY-NNNNN?

internal/store/device.go
asafbennatan: the problem is that we need to add joins and filter by a joined field column , would you now add a generic mechanisem to join and filter for generic ? IMO the solution should be - avoid using generic at all .
unfortunately the way we build the continue token is also highly query specific

avishayt: OK, fine as is

api/core/v1beta1/openapi.yaml
asafbennatan: done

avishayt (approved):

PR #2918, opened 2026-05-07
EDM-3841: Git change detection and propagation

galel12:
## EDM-3841: Git change detection and propagation

**Jira:** [EDM-3841](https://issues.redhat.com/browse/EDM-3841)

### Summary

Adds the git change detection sync controller, building on the infrastructure from EDM-3840:

* **`DependencySyncGit` periodic task** — polls git repositories via `GitLsRemote`, compares SHAs against stored fingerprints, emits `DependencyChangeDetected` events on change
* **`DependencyChangeDetected` event** — new event reason with `resourceKey` and `fingerprint` in details; consumed by `fleetValidate` (fleet-owned) and device re-render (standalone)
* **Sync-triggered TV naming** — `v{generation}-{short-hash}` for sync-triggered TemplateVersions, `v{generation}` unchanged for spec-driven
* **`dependency_refs` population** — fleet validation now collects git/HTTP/secret dependencies from template config into `dependency_refs` (the sync controller's polling work list)
* **Design update** — removed per-fleet `DependenciesSync` API field; polling interval moved to global server config (`dependenciesSync.pollInterval`, default 15m)

### Breaking Changes

* **Removed `DependenciesSync` from `FleetSpec` and `DeviceSpec`** — per-fleet/device polling configuration replaced by a global server-side setting (`dependenciesSync.pollInterval`)
* **Added `DependencyChangeDetected` to `EventDetails` oneOf** — new event reason for the sync controller

### Testing

* Unit tests for change detection, no-change, probe errors, fan-out, TV naming, fingerprint extraction, dependency ref collection/population
* Integration tests for `sync_state` and `dependency_refs` store CRUD


api/core/v1beta1/openapi.yaml
```diff
@@ -7188,6 +7172,7 @@ components:
             - ResourceSyncParsingFailed
             - ResourceSyncSynced
             - ResourceSyncSyncFailed
+            - DependencyChangeDetected
```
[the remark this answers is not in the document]
galel12: This is an intentional addition — `DependencyChangeDetected` is a new event reason for the sync controller. The `breaking-change-approved` label has been applied.

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,222 @@
...
+			d.serviceHandler.CreateEvent(ctx, orgId, event)
+		}
+	}
+}
+
+func containsTemplateParam(s string) bool {
+	return strings.Contains(s, "{{") && strings.Contains(s, "}}")
+}
+
+func deref(s *string) string {
```
kkyrazis: nit: we have lo.FromPtr for this.

internal/worker_server/server.go
```diff
@@ -77,7 +77,7 @@ func (s *Server) Run(ctx context.Context) error {
 	serviceHandler := service.WrapWithTracing(
 		service.NewServiceHandler(s.store, workerClient, kvStore, nil, s.log, "", "", []string{}, false))
 
-	if err = tasks.LaunchConsumers(ctx, s.queuesProvider, serviceHandler, s.k8sClient, kvStore, s.cfg, 1, 1, s.workerMetrics); err != nil {
+	if err = tasks.LaunchConsumers(ctx, s.queuesProvider, serviceHandler, s.k8sClient, kvStore, s.cfg, s.store.DependencyRef(), 1, 1, s.workerMetrics); err != nil {
```
kkyrazis: For both the periodic checker and the worker we've been accessing our resources via the services rather than directly via stores. I just want to confirm that this divergence is intentional and that we don't want to continue the existing pattern.

internal/tasks/dependency_sync_git.go
galel12: Good call — replaced with `lo.FromPtr` in [0424fe2](https://github.com/flightctl/flightctl/pull/2918/commits/0424fe24186c24309d23d3a2026584e6e58e242f).

internal/worker_server/server.go
galel12: You're right — refactored to route through the service layer in [0424fe2](https://github.com/flightctl/flightctl/pull/2918/commits/0424fe24186c24309d23d3a2026584e6e58e242f). Added `UpsertDependencyRef` and `DeleteDependencyRefsByFleet` to the `Service` interface and removed the direct store dependency from the consumer/fleetValidate chain.

galel12: Refactored both paths to go through the service layer. The `DependencySyncGit` periodic task no longer holds `syncStore` `depRefStore` fields; all access routes through `serviceHandler` via new `ListDependencyRefsByRefType`, `GetSyncState`, `SetSyncState`, and `SetSyncStateLastCheckedAt` methods.

51533060

internal/tasks/fleet_validate.go
```diff
@@ -260,6 +268,101 @@ func (t *FleetValidateLogic) validateHttpProviderConfig(ctx context.Context, con
...
+	return base + "-" + shortHash
+}
+
+// populateDependencyRefs replaces all dependency_refs rows for the fleet
+// with the current set extracted from the template config. This gives the
+// sync controller an up-to-date polling work list.
+func (t *FleetValidateLogic) populateDependencyRefs(ctx context.Context, fleetName string) error {
+	refs := t.collectDependencyRefs(fleetName)
+
+	if status := t.serviceHandler.DeleteDependencyRefsByFleet(ctx, t.orgId, fleetName); status.Code != http.StatusOK {
```
kkyrazis: Is it ok that these operations are non-transactional? Can we appropriately recover if all refs are deleted and only a subset of Upserts succeed?

internal/periodic_checker/task_definitions.go
```diff
@@ -82,6 +84,12 @@ func MergeTasksWithConfig(cfg *config.Config) map[PeriodicTaskType]PeriodicTaskM
 		merged[PeriodicTaskTypeVulnerabilitySync] = meta
 	}
 
+	if cfg.DependenciesSync != nil && cfg.DependenciesSync.PollInterval > 0 {
+		meta := merged[PeriodicTaskTypeDependencySyncGit]
+		meta.Interval = time.Duration(cfg.DependenciesSync.PollInterval)
```
asafbennatan: i think the task itself should run at higher frequency , the data queries should represent the relevant poll interval - 
so for example this task runs every 3 minutes.
when you fetch the data you only get results which should be processed based on time .
SQL query to make it clear (use gorm though):

`select * from dependenciesRef where lastChecked + DependndenciesSync.PollInterval > NOW`

this also allows us to recover in the case of partial faliures

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,210 @@
...
+type probeTarget struct {
+	repoName string
+	revision string
+	// fleets/devices that will receive events if this target changes
+	fleetNames  []string
+	deviceNames []string
+}
+
+func (d *DependencySyncGit) Poll(ctx context.Context, orgId uuid.UUID) {
+	refs, status := d.serviceHandler.ListDependencyRefsByRefType(ctx, orgId, "git")
```
asafbennatan: handling by type is a good idea .
the query itself is too broad , we should optimize:

**selection:**

1. join with repository so you dont need to get this data later
2. join with SyncState so you dont need to get this data later

**filtering:**

1. see the above regarding the lastChecked interval filtering .
2. does not contain template parameters

**group by:**

1. do a group by query , do not group in memory.additionally you can filter by template parameters in the SQL as well.

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,210 @@
...
+	for _, t := range targetMap {
+		result = append(result, *t)
+	}
+	return result
+}
+
+func (d *DependencySyncGit) probeAndReconcile(ctx context.Context, orgId uuid.UUID, target *probeTarget, pollInterval time.Duration) {
+	resourceKey := fmt.Sprintf("git:%s/%s", target.repoName, target.revision)
+
+	existing, status := d.serviceHandler.GetSyncState(ctx, orgId, resourceKey)
```
asafbennatan: see above regarding the query

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,210 @@
...
+	if status.Code != http.StatusOK {
+		d.log.Errorf("failed reading sync state for %s: %s", resourceKey, status.Message)
+		return
+	}
+
+	if existing != nil && time.Since(existing.LastCheckedAt) < pollInterval {
```
asafbennatan: see above regarding the query

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,210 @@
...
+		return
+	}
+
+	repo, status := d.serviceHandler.GetRepository(ctx, orgId, target.repoName)
```
asafbennatan: see above regarding the query

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,210 @@
...
+		return
+	}
+
+	auth, err := GetAuth(repo, d.cfg)
+	if err != nil {
+		d.log.WithError(err).Warnf("failed getting auth for repository %s", target.repoName)
+		return
+	}
+
+	resolved, err := d.lsRemote(ctx, repoURL, []string{target.revision}, auth)
```
asafbennatan: see above regarding the query , since you are using ls remote , you can group by the repository url , and lsRemote all refs at the same time ( this is implied by the group by)

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,210 @@
...
+	newSHA, found := resolved[target.revision]
+	if !found {
+		d.log.Warnf("ref %s not found in repository %s", target.revision, target.repoName)
+		return
+	}
+
+	now := time.Now().UTC()
+
+	if existing == nil {
+		if st := d.serviceHandler.SetSyncState(ctx, orgId, &model.SyncState{
```
asafbennatan: since you are potentially handling multiple refs (due to group by) , you should massUpdate (do not update one ref at a time , update all of them)

note - update all the relevant fields in the same query

internal/tasks/fleet_validate.go
```diff
@@ -260,6 +268,101 @@ func (t *FleetValidateLogic) validateHttpProviderConfig(ctx context.Context, con
...
+		shortHash = shortHash[:8]
+	}
+	return base + "-" + shortHash
+}
+
+// populateDependencyRefs replaces all dependency_refs rows for the fleet
+// with the current set extracted from the template config. This gives the
+// sync controller an up-to-date polling work list.
+func (t *FleetValidateLogic) populateDependencyRefs(ctx context.Context, fleetName string) error {
+	refs := t.collectDependencyRefs(fleetName)
```
asafbennatan: not sure i follow - in the case of templating only - we should create the dependency ref per device in the fleet after we have calculated the resulting value , as far as i understand this code it is not templating specific and it creates it for the fleet , not the devices in the fleet

internal/tasks/fleet_validate.go
asafbennatan: see my question above regarding the entire logic of this function , i assume that even if i am correct you will require similar logic , in this case make sure that:
1. you run only two queries (one for deleting all , one for upserting all)
2. these two are run inside a DB transaction or using postgresql CTE to avoid a case where a fail corrupts the data or a select query reads incorrect data

asafbennatan (changes requested):

internal/tasks/fleet_validate.go
galel12: Good point — replaced the non-transactional delete+upsert loop with a single `ReplaceByFleet` store method that wraps both the DELETE and a bulk INSERT inside a DB transaction. If any insert fails, the whole thing rolls back and old refs stay intact.

403a1d65

internal/periodic_checker/task_definitions.go
galel12: Addressed — decoupled the two concerns:

1. **Task frequency** now defaults to 3m (`DefaultDependencySyncTaskInterval`), and the `MergeTasksWithConfig` override that set it to the poll interval has been removed.
2. **Poll interval** (default 15m) is only used in the SQL time-gate: `ss.last_checked_at + ? * INTERVAL '1 second' < NOW()`. Most runs return 0 rows and exit immediately.

This gives faster recovery from partial failures and faster first-poll for newly added refs.

d0f61258

internal/tasks/dependency_sync_git.go
galel12: Addressed — rewrote to a single GORM query that LEFT JOINs `sync_states`, filters by poll interval + template params in SQL, and groups by `(repository_name, revision)` with `array_agg` for fan-out. `GetRepository` is called once per unique repo (not per revision), and all revisions for a repo are passed to a single `lsRemote` call.

0aecf94a

internal/tasks/dependency_sync_git.go
galel12: Addressed in the query optimization — see reply above. 0aecf94a

internal/tasks/dependency_sync_git.go
galel12: Addressed in the query optimization — see reply above. 0aecf94a

internal/tasks/dependency_sync_git.go
galel12: Addressed in the query optimization — see reply above. 0aecf94a

internal/tasks/dependency_sync_git.go
galel12: Addressed — `reconcile()` now calls `BulkUpsertSyncState` for changed entries and `BulkUpdateLastCheckedAt` for unchanged ones, so it's at most 2 DB round-trips per poll cycle regardless of probe count.

0aecf94a

internal/tasks/dependency_sync_git.go
galel12: Addressed — `reconcile()` now calls `BulkUpsertSyncState` (single `INSERT ... ON CONFLICT DO UPDATE` for changed entries) and `BulkUpdateLastCheckedAt` (single `UPDATE ... WHERE resource_key IN (...)` for unchanged entries). At most 2 DB round-trips per poll cycle regardless of probe count, and all relevant fields are updated in the same query.

0aecf94a

internal/tasks/fleet_validate.go
galel12: Addressed — `collectDependencyRefs` now detects parameterized revisions and calls `resolveParameterizedGitRefs`, which lists all fleet devices, resolves the template per device, and creates per-device `DependencyRef` entries. The SQL query filters out unresolved templates via `dr.revision NOT LIKE '%{{%'`, and the `fleet_names` aggregation uses `FILTER (WHERE dr.device_name = '')` to avoid double-counting device-level refs as fleet events.

01a0c43c

internal/tasks/fleet_validate.go
```diff
@@ -75,6 +77,12 @@ func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Co
 		return fmt.Errorf("setting repository references: %s", status.Message)
 	}
 
+	// Populate dependency_refs for the sync controller's polling work list.
+	// Done even if validation failed — same rationale as OverwriteFleetRepositoryRefs.
+	if err := t.populateDependencyRefs(ctx, *fleet.Metadata.Name); err != nil {
+		return fmt.Errorf("failed populating dependency refs for fleet %s: %w", *fleet.Metadata.Name, err)
+	}
```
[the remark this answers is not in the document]
galel12: Addressed — `populateDependencyRefs` now uses `ReplaceByFleet` (transactional delete+insert) and propagates errors from both the replace and the delete path.

internal/tasks/dependency_sync_git.go
asafbennatan: the correct approch runs n+1 query where n is the number of unique repo names , 
there is no need for this .
join with repository in this query and get the relevant info.

internal/store/dependencyref.go
```diff
@@ -32,3 +38,80 @@ func (s *DependencyRefStore) getDB(ctx context.Context) *gorm.DB {
...
+func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID uuid.UUID, pollInterval time.Duration) ([]model.GitDependencyProbe, error) {
+	var probes []model.GitDependencyProbe
+	err := s.getDB(ctx).
+		Table("dependency_refs dr").
+		Select(
+			"dr.repository_name, dr.revision, ss.fingerprint, "+
+				"array_agg(DISTINCT dr.fleet_name) FILTER (WHERE dr.fleet_name <> '' AND dr.device_name = '') AS fleet_names, "+
+				"array_agg(DISTINCT dr.device_name) FILTER (WHERE dr.device_name <> '') AS device_names",
+		).
+		Joins("LEFT JOIN sync_states ss ON ss.org_id = dr.org_id AND ss.resource_key = 'git:' || dr.repository_name || '/' || dr.revision").
```
asafbennatan: the join doesnt makes sense to me , shouldnt it just be :
LEFT JOIN sync_states on dr.resource_key = sync_startes.resource_key ?

internal/tasks/fleet_validate.go
```diff
@@ -260,6 +269,152 @@ func (t *FleetValidateLogic) validateHttpProviderConfig(ctx context.Context, con
...
+	fleetName string, gitSpec *domain.GitConfigProviderSpec) []model.DependencyRef {
+
+	var refs []model.DependencyRef
+	listParams := domain.ListDevicesParams{
+		FieldSelector: lo.ToPtr(fmt.Sprintf("metadata.owner=%s", *util.SetResourceOwner(domain.FleetKind, fleetName))),
+		Limit:         lo.ToPtr(int32(ItemsPerPage)),
+	}
+
+	for {
+		devices, status := t.serviceHandler.ListDevices(ctx, t.orgId, listParams, nil)
```
asafbennatan: i think this should be done in the fleet_rollout task , not the fleet_validate task , the reason is that :
1. we are already resolving templating vars in fleet_rollout , 
2.we already list the fleet devices

the current approch adds many ListDevices calls

internal/store/dependencyref.go
galel12: Agreed — added `resource_key` to the `DependencyRef` model so the JOIN is now a direct column comparison:

```
LEFT JOIN sync_states ss ON ss.org_id = dr.org_id AND ss.resource_key = dr.resource_key
```

Also joined with `repositories` in the same query to eliminate the N+1 `GetRepository` calls — repo spec is now carried in the probe result.

c4d7f7c1

internal/tasks/fleet_validate.go
galel12: Agreed — moved parameterized ref population from `fleet_validate` to `fleet_rollout`. Now:

- **`fleet_validate`** only handles non-parameterized refs (e.g. `targetRevision: main`) — no `ListDevices` call needed.
- **`fleet_rollout`** collects device-level `DependencyRef`s as a side effect of the template resolution it already does during `replaceGitConfigParameters`, then batch-upserts them after each page.
- **`ReplaceByFleet`** now only deletes fleet-level refs (`device_name = ''`), so device-level refs managed by rollout are preserved.

Zero extra queries — parameterized ref population piggybacks on the device iteration that rollout already performs.

c4d7f7c1

internal/tasks/dependency_sync_git.go
galel12: Addressed — the query now also JOINs with `repositories` to fetch `r.spec` inline, eliminating the per-repo `GetRepository` calls entirely. The repo spec is carried in `GitDependencyProbe.RepoSpec` and deserialized in `probeRepo` without any additional DB round-trips.

c4d7f7c1

internal/store/model/dependencyref.go
```diff
@@ -12,14 +51,28 @@ type DependencyRef struct {
 	FleetName       *string   `gorm:"primaryKey;default:''"`
 	DeviceName      *string   `gorm:"primaryKey;default:''"`
 	RefType         string    `gorm:"primaryKey"` // "git", "http", "secret"
+	ResourceKey     string    // Stable key matching sync_state, e.g. "git:repo/ref"
```
asafbennatan: the primary key should be ResourceKey,FleetName,DeviceName - thats it

internal/store/dependencyref.go
galel12: The org_id is needed in the JOIN to prevent cross-tenant matching. Two different orgs could have the same resource_key value (git:my-repo/main), and without org_id in the JOIN condition, org A's dependency_ref would match org B's sync_state.

internal/tasks/consumer.go
```diff
@@ -236,6 +236,11 @@ func shouldValidateFleet(ctx context.Context, event domain.Event, log logrus.Fie
 		return true
 	}
 
+	// If a dependency change was detected for a fleet, return true
```
asafbennatan, line 246: why for validate ?

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,215 @@
...
+	var (
+		mu      sync.Mutex
+		results []probeResult
+	)
+
+	sem := make(chan struct{}, d.maxConcurrent)
+	var wg sync.WaitGroup
+
+	for repoName, group := range repoGroups {
+		wg.Add(1)
```
asafbennatan, line 79: doesnt that means we are only probing one at a time ? i think this should be a configurable value , default should be a sensible value like 5

internal/tasks/dependency_sync_git.go
```diff
@@ -0,0 +1,215 @@
...
+		sem <- struct{}{}
+		go func() {
+			defer wg.Done()
+			defer func() { <-sem }()
+			res := d.probeRepo(ctx, repoName, group)
+			mu.Lock()
```
asafbennatan, line 85: remove locks and move reconsile into the goroutine - yes we will have more events but they will be independent

internal/store/model/dependencyref.go
```diff
@@ -12,14 +51,28 @@ type DependencyRef struct {
...
+// and DeviceNames carry the fan-out targets collected via array_agg.
+// RepoSpec carries the repository's JSONB spec so callers can extract the
+// URL and auth without a separate GetRepository round-trip.
+type GitDependencyProbe struct {
+	RepositoryName string
+	Revision       string
+	Fingerprint    *string
+	FleetNames     StringArray
+	DeviceNames    StringArray
+	RepoSpec       json.RawMessage
```
asafbennatan: can we have it typed and not a rawMessage?

internal/store/model/dependencyref.go
galel12: Agreed — simplified the PK to `(org_id, resource_key, fleet_name, device_name)`. The other columns (`ref_type`, `repository_name`, `secret_name`, `secret_namespace`) are now regular data columns since `resource_key` already encodes the identity.

549e1fa2

internal/tasks/fleet_rollout.go
```diff
@@ -182,13 +183,22 @@ func (f FleetRolloutsLogic) rolloutFleetPage(
...
+		refs, updateErr := f.updateDeviceToFleetTemplate(iterCtx, device, templateVersion, delayDeviceRender)
 		if updateErr != nil {
 			f.log.Errorf("failed to update target generation for device %s (fleet %s): %v", *device.Metadata.Name, f.event.InvolvedObject.Name, updateErr)
 			pageFailures++
 		}
+		pageRefs = append(pageRefs, refs...)
+	}
+
+	if len(pageRefs) > 0 {
+		st := f.serviceHandler.BulkUpsertDeviceDependencyRefs(iterCtx, f.orgId, pageRefs)
```
asafbennatan: what about deleted refs ? (lets say that device resolved ref to git.com/X and now to git.com/Y , we should delete git.com/X , same issue if the fleet spec itself was removed a config ref).
check how to do this in a single transaction like before

internal/tasks/consumer.go
galel12: `DependencyChangeDetected` triggers fleet validation because that's where the new TemplateVersion gets created — with the sync-triggered `v{gen}-{hash}` naming. This follows the same pattern as `ReferencedRepositoryUpdated` (line 235): an external dependency changed, so the fleet template needs re-validation, a new TV, and rollout.

internal/tasks/dependency_sync_git.go
galel12: `maxConcurrent` is already set to 10 (not 1) — the semaphore at line 76 limits to `d.maxConcurrent` concurrent goroutines across repos. Each repo gets its own goroutine, so up to 10 repos are probed in parallel. Happy to make it configurable if needed.

internal/tasks/dependency_sync_git.go
galel12: There's a tradeoff here with the batched DB operations from your earlier feedback — `reconcile()` currently calls `BulkUpsertSyncState` and `BulkUpdateLastCheckedAt` as single bulk queries across all probes. Moving reconcile into each goroutine would eliminate the mutex but replace the 2 bulk DB calls with N individual calls (one per repo).

Which direction do you prefer — independent goroutines with per-repo DB writes, or the current bulk approach?

internal/store/model/dependencyref.go
galel12: `RepoSpec` is `json.RawMessage` because it's a GORM scan target from a JSONB column. The domain type (`RepositorySpec`) is a discriminated union (generic vs HTTP git repos with different auth) that requires `FromRepositorySpec()` conversion — GORM can't scan directly into it. The raw JSON is deserialized in `probeRepo()` via the existing helper immediately after the query.

internal/tasks/fleet_rollout.go
galel12: Good catch — added `DeleteDeviceRefsByFleet` which deletes all device-level refs (`device_name <> ''`) for the fleet at the start of `RolloutFleet`, before the page loop. Each page then re-populates current refs via `BulkUpsertDeviceRefs`. This handles both cases: revision changes (git:repo/v1 → git:repo/v2) and removed config items.

Fleet-level refs are already handled transactionally by `ReplaceByFleet` in `fleet_validate`.

e945e578

internal/store/model/dependencyref.go
asafbennatan: RepoSpec *JSONField[domain.RepositorySpec] `gorm:"type:jsonb"`
?

galel12: Good call — switched to `*JSONField[domain.RepositorySpec]` with `gorm:"type:jsonb"`. GORM now scans directly into the typed struct, and `probeRepo` uses `group[0].RepoSpec.Data` without any manual unmarshaling.

b8fa2f25

internal/tasks/fleet_rollout.go
asafbennatan: i might be missing something but i dont see :
DeleteDeviceRefsByFleet
being called

additionally even if it was - we must make sure we replace the refs in a way that does not allow a select query that runs between delete and upsert to see empty data
aka in a db transaction

internal/tasks/dependency_sync_git.go
asafbennatan: lets keep it as it is now , if there will be any scaling issue we can do what i suggested , there is the in between option of batching these , anyway for now the current approach is fine

internal/tasks/fleet_rollout.go
galel12: `DeleteDeviceRefsByFleet` was being called at the top of `RolloutFleet`, but you're right about the transactional concern there was a window between delete and upsert where readers could see empty data. Fixed in a5e2c4ac: instead of delete-then-per-page-upsert, rolloutFleetPage now returns refs, `RolloutFleet` accumulates them across all pages, then calls `ReplaceDeviceRefsByFleet` which does the delete + insert atomically inside a `gorm.DB.Transaction()`. 
The old non-transactional `DeleteDeviceRefsByFleet` is removed

galel12:
@asafbennatan — Addressed your review feedback in the last 3 commits:

**41772efe — Extract populate_dependency_refs handler, add standalone device support**
- Extracted dependency ref population into a dedicated event-driven handler (`populate_dependency_refs.go`) with its own predicate, decoupled from fleet validation.
- Added standalone device support (AC-6).

**37bea68e — Harden dependency ref lifecycle and consolidate store methods**
1. Fleet-owned devices now clean up leftover standalone refs on adoption (prevents stale entries from the device's standalone era).
2. `ResourceDeleted` events for fleets/devices now trigger dependency ref cleanup.
3. Consolidated `ReplaceByFleet`, `ReplaceDeviceRefsByFleet`, and `ReplaceByStandaloneDevice` into a single `transactionalReplace` helper — eliminates triplicated delete+insert logic.
4. Each `DependencyRef` gets its own `FleetName`/`DeviceName` pointer per loop iteration — no shared-pointer aliasing.

**978982b7 — Fix device deletion cleanup and event emission ordering**
1. Device deletion now removes ALL refs (`WHERE device_name = ?`) — not just standalone refs. Previously, fleet-rollout parameterized refs were orphaned when a device was deleted.
2. Reordered `reconcile()` in `DependencySyncGit`: events are emitted *before* sync state fingerprints are persisted, ensuring consumers always see the event before the fingerprint advances.


Test coverage: 23 unit tests covering all new paths (fleet/device deletion, adoption cleanup, pointer independence, predicate logic).

internal/tasks/fleet_rollout.go
```diff
@@ -240,10 +251,24 @@ func (f FleetRolloutsLogic) RolloutDevice(ctx context.Context) error {
 		return nil
 	}
 	delayDeviceRender := fleet.Spec.RolloutPolicy != nil && fleet.Spec.RolloutPolicy.DisruptionBudget != nil
-	return f.updateDeviceToFleetTemplate(ctx, device, templateVersion, delayDeviceRender)
+	refs, err := f.updateDeviceToFleetTemplate(ctx, device, templateVersion, delayDeviceRender)
+	if err != nil {
+		return err
+	}
+	deviceName := f.event.InvolvedObject.Name
+	if st := f.serviceHandler.DeleteDependencyRefsByDevice(ctx, f.orgId, deviceName); st.Code != http.StatusOK {
```
asafbennatan: this should be in the same transaction as when you insert the dependency refs .

additionally we are deleteing all the dependency refs of a device . 
but in populate_dependency_refs we also delete refs if the device is fleet owned  why both ?
https://github.com/flightctl/flightctl/pull/2918/changes#diff-239592cf35e715dcd7f1d4fa0c74608263bda73fe6cb9581a34338a9354a9ab7R63

galel12:
## Bugs found during local E2E testing

I deployed the full stack to a local Kind cluster with a Podman-based SSH git server and ran through the complete dependency sync lifecycle. Found and fixed three bugs:

### 1. GORM `StringArray` scanning failure (`52b68f48`)

**Symptom:** `dependency-sync-git` correctly resolved refs via `git ls-remote` and updated `sync_states`, but never emitted `DependencyChangeDetected` events. Logs showed `unsupported data type: &[]`.

**Root cause:** `model.StringArray` (used for `fleet_names`/`device_names` in the `ListDueGitDependencies` query) implements `sql.Scanner`/`driver.Valuer` but GORM didn't know the PostgreSQL column type. It fell back to generic scanning which silently returned empty slices.

**Fix:** Added `GormDataType() string` returning `"text[]"` to `model.StringArray`.

### 2. Wrong fleet name in device-level `DependencyRef` (`9d024bda`)

**Symptom:** Device-level refs created during fleet rollout had `fleet_name` set to the **device** name instead of the fleet name.

**Root cause:** `replaceGitConfigParameters` used `f.event.InvolvedObject.Name` as the fleet name. In the fleet-event path that's correct, but in the device-event rollout path, `InvolvedObject.Name` is the device name.

**Fix:** Extract fleet name from `device.Metadata.Owner` via `getOwnerFleet(device)`.

### 3. Stale device-level refs on branch switch or fleet move (`9d024bda`)

**Symptom:** When a device's label changes (e.g., `branch: main` → `branch: feature-x`), the old `git:test-sync/main` device-level ref was left behind alongside the new `git:test-sync/feature-x` ref. Same issue when a device moves between fleets.

**Root cause:** The device-event rollout path used `BulkUpsertDeviceDependencyRefs` which only inserts/updates by primary key — it never deletes rows whose `resource_key` changed. Similarly, when scoped to the new fleet, old-fleet refs were untouched.

**Fix:** Changed to delete-then-insert: `DeleteDependencyRefsByDevice` (wipes all device refs regardless of fleet) followed by `BulkUpsertDeviceDependencyRefs` for the new set.

---

### E2E scenarios validated

| # | Scenario | Result |
|---|----------|--------|
| 1 | Non-parameterized fleet gitRef → dependency_refs populated → periodic sync → change detection → event emitted | ✅ |
| 2 | Parameterized fleet gitRef (`{{ .metadata.labels.branch }}`) skipped at fleet level | ✅ |
| 3 | Device joins fleet → rollout resolves template → device-level ref created | ✅ |
| 4 | Device label change (branch switch) → old ref replaced, no stale rows | ✅ |
| 5 | Device moves between fleets → old fleet refs cleaned | ✅ |
| 6 | Fleet config updated (add/remove gitRef) → refs update atomically | ✅ |
| 7 | Standalone device with gitRef → ref created/deleted correctly | ✅ |
| 8 | Standalone device adopted by fleet → standalone ref cleaned | ✅ |
| 9 | Device/fleet deletion → all associated refs removed | ✅ |

internal/tasks/fleet_rollout_test.go
```diff
@@ -973,6 +976,164 @@ func TestFleetRolloutsLogic_ReplaceComposeInlineApplicationParameters(t *testing
 	}
 }
 
+func TestReplaceGitConfigParameters_DeviceLevelRefs(t *testing.T) {
```
asafbennatan, line 979: we need more tests here around fleet owned devioce dependency refs 
(device is standalone ) 
-- check  only device level dep ref exist
(device is owned by fleet A )
-- check only fleet A dep refs exists
(fleet A got a ref template for an existing config)
check that the matching ref is now at the device level
(fleet A parameterized ref is now non-parametrized)
check that matching ref is now device level 
update labels on device , make sure parameterized ref changes value
update labels on device that are not used in parameterized dep ref - make sure no change 
etc

internal/tasks/fleet_rollout.go
galel12: Addressed in 511b33b7 — two changes:

1. **Transaction:** Replaced the separate `DeleteDependencyRefsByDevice` + `BulkUpsertDeviceDependencyRefs` with a single `ReplaceFleetDeviceDependencyRefs` call, which routes to `ReplaceByFleetDevice` in the store — that uses `transactionalReplace` (DELETE + INSERT inside a `gorm.DB.Transaction()`).

2. **Redundant deletion:** The old `DeleteDependencyRefsByDevice` deleted ALL refs for the device (including standalone refs that `populate_dependency_refs` manages). Now `ReplaceByFleetDevice` is scoped to `WHERE fleet_name = ? AND device_name = ?`, so it only touches refs for this device within its owning fleet. The two code paths are orthogonal:
   - `fleet_rollout` → `ReplaceFleetDeviceDependencyRefs(fleet, device, refs)` — replaces fleet-scoped device refs
   - `populate_dependency_refs` → `ReplaceStandaloneDeviceDependencyRefs(device, refs)` — replaces standalone refs
   - `populate_dependency_refs` (deletion) → `DeleteDependencyRefsByDevice(device)` — wipes everything on device deletion

internal/tasks/fleet_rollout_test.go
galel12: Addressed in 511b33b7 — added `TestDeviceDependencyRefLifecycle` with 7 subtests covering:

1. **Standalone device** → non-parameterized revision produces no device-level refs
2. **Device owned by fleet A** → parameterized ref scoped to fleet A with correct device/fleet names
3. **Parameterized ref resolution** → `{{ .metadata.labels.branch }}` resolves to device-level ref with correct value
4. **Parameterized → non-parameterized** → non-parameterized revision produces no device-level refs (fleet-level only)
5. **Label change (relevant)** → parameterized ref value updates (`main` → `feature-x`)
6. **Label change (irrelevant)** → changing `env` label doesn't affect `branch`-based ref
7. **RolloutDevice integration** → mock verifies `ReplaceFleetDeviceDependencyRefs` is called (not the old delete+upsert)

asafbennatan (approved):
LGTM

asafbennatan (approved):
LGTM

kkyrazis (approved):

PR #2920, opened 2026-05-10
EDM-3854: Fix flaky agent update test 82481

eldar101:
the failure was a test flake around SSH/log collection during post-rollback VM settling, not a product failure.

I fixed it by changing the rollback log assertion in test/e2e/agent/agent_update_test.go so a temporary journal/SSH read failure does not fail the test immediately. It now retries log reads for up to TENMINTIMEOUT, polling every TENSECTIMEOUT, and still requires the positive proof.

So the test is less flaky, but it still verifies rollback behavior instead of silently passing

Link to failures:
https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-flightctl-gotests/1474/testReport/

https://reportportal-ecosystem-qe.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#edge-management-qe/launches/all/36939/1941223/1941266/log?item0Params=filter.eq.hasStats%3Dtrue%26filter.eq.hasChildren%3Dfalse%26filter.in.issueType%3Dti001

### DIFF
--- a/test/e2e/agent/agent_update_test.go
+++ b/test/e2e/agent/agent_update_test.go
@@ -273,17 +273,9 @@ var _ = Describe("VM Agent behavior during updates", Label("agent-update"), func
 			}, LONGTIMEOUT)
 
 			By("Verifying agent logs show rollback attempt")
-			dur, err := time.ParseDuration(TIMEOUT)
-			Expect(err).ToNot(HaveOccurred())
-			Eventually(func() string {
-				GinkgoWriter.Printf("Checking console output for rollback logs\n")
-				logs, err := harness.ReadPrimaryVMAgentLogs("", util.FLIGHTCTL_AGENT_SERVICE)
-				Expect(err).NotTo(HaveOccurred())
-				return logs
-			}).
+			Eventually(readAgentLogsForRollbackAssertion, TIMEOUT, TENSECTIMEOUT).
+				WithArguments(harness).
 				WithContext(harness.GetTestContext()).
-				WithTimeout(dur).
-				WithPolling(time.Second * 10).
 				Should(ContainSubstring(fmt.Sprintf("Attempting to rollback to previous renderedVersion: %d", expectedVersion)))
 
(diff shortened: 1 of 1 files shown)


itzikb-redhat (commented):
lgtm

itzikb-redhat (approved):

test/e2e/agent/agent_update_test.go
```diff
@@ -273,17 +273,11 @@ var _ = Describe("VM Agent behavior during updates", Label("agent-update"), func
...
-				logs, err := harness.ReadPrimaryVMAgentLogs("", util.FLIGHTCTL_AGENT_SERVICE)
-				Expect(err).NotTo(HaveOccurred())
-				return logs
-			}).
+			Eventually(readAgentLogsForRollbackAssertion).
+				WithArguments(harness).
 				WithContext(harness.GetTestContext()).
-				WithTimeout(dur).
-				WithPolling(time.Second * 10).
+				WithTimeout(TENMINTIMEOUT).
```
siserafin: I see that this timeout was previously set to 5 minutes. I would not increase it to 10 minutes, since the device reboot should already be completed by this stage of the test.
This log should appear before the reboot, and I think waiting 5 minutes to SSH into the device is sufficient. If the device is still not reachable via SSH after 5 minutes, that likely indicates an issue with either the device or the network.

eldar101: np, restoring 5 min.

siserafin: We also often see this issue in random tests beforeEach, and I think that a good solution would be to just re-create the device

siserafin (approved):
lgtm as a refactoring. I think this is the same issue we occasionally see in the random beforeEach test failures. Sometimes something goes wrong after the device is restored or rebooted, and the device becomes unreachable via SSH.
In the beforeEach case, one possible solution could be to recreate the device when ssh connectivity issues are detected. In this case, though, it seems more difficult to eliminate the flakiness completely.

PR #2922, opened 2026-05-10
EDM-3084: Fix console websocket 500 error for nonexistent devices

gshilin-sdb:
The console websocket handler was returning 500 Internal Server Error instead of 404 Not Found for nonexistent devices due to a logical flaw in the StartSession function.

Root cause:
- modifyAnnotations() calls could fail with 500 errors that masked the proper 404 device not found errors
- Race conditions between device existence checks and annotation operations could lead to inconsistent error responses

Fix using guard pattern:
- Check device existence once at the beginning
- Validate device state upfront before any annotation operations
- Add fallback device existence check if annotation operations fail
- Ensure 404 errors are never masked by 500 errors

Added comprehensive unit tests covering:
- Device not found scenarios (EDM-3084 regression test)
- Decommissioned device handling
- Missing session metadata validation


asafbennatan (approved):
LGTM

PR #2924, opened 2026-05-11
EDM-3761: Improve status reporting for critical resource alerts

gshilin-sdb:
Critical resource alerts were showing as update failures instead of proper deferral messages. This fix:

- Preserves original blocking behavior for safety
- Improves status reporting to show deferred updates clearly
- Handles critical resource alerts separately in error handling
- Shows "Update deferred due to critical resource alerts" instead of confusing error messages

The fix maintains all existing safety mechanisms while providing better UX.

### DIFF
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -620,10 +620,18 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1beta1.Device, sy
 	}
 
 	se := errors.FormatError(syncErr)
-	if !errors.IsRetryable(syncErr) {
+
+	// Handle critical resource alerts differently from update failures
+	if errors.Is(syncErr, errors.ErrCriticalResourceAlert) {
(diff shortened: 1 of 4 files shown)


internal/agent/device/device.go
```diff
@@ -620,22 +620,31 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1beta1.Device, sy
 	}
 
 	se := errors.FormatError(syncErr)
-	if !errors.IsRetryable(syncErr) {
+
+	// Handle critical resource alerts differently from update failures
+	if errors.Is(syncErr, errors.ErrCriticalResourceAlert) {
+		msg := fmt.Sprintf("Update to renderedVersion %s deferred due to critical resource alerts. Will retry when resources improve.", version)
```
asafbennatan: > resources improve
maybe resources are available?
additionally is there a way to report what resource was the one causing it ? (CPU/MEM/DISK) etc

gshilin-sdb: updated text. added resource name

asafbennatan (approved):
LGTM

PR #2932, opened 2026-05-12
EDM-3887: taking AAP team role assignment into account

asafbennatan:


pkg/aap/roles.go
```diff
@@ -60,3 +83,15 @@ func (a *AAPGatewayClient) ListUserRoleAssignments(ctx context.Context, token st
 	endpoint := a.buildEndpoint("/api/controller/v2/role_user_assignments/", query)
 	return getWithPagination[AAPRoleUserAssignment](a, ctx, endpoint, token)
 }
+
+// GET /api/controller/v2/role_team_assignments/?team__id={team_id}
+func (a *AAPGatewayClient) ListTeamRoleAssignments(ctx context.Context, token string, teamID string) ([]*AAPRoleTeamAssignment, error) {
+	query := url.Values{}
+	query.Set("team__id", teamID)
+	if a.maxPageSize != nil {
+		query.Set("page_size", fmt.Sprintf("%d", *a.maxPageSize))
```
gshilin-sdb: Nit: strconv.Itoa(*a.maxPageSize)

pkg/aap/roles.go
```diff
@@ -60,3 +83,15 @@ func (a *AAPGatewayClient) ListUserRoleAssignments(ctx context.Context, token st
...
+	}
+
+	endpoint := a.buildEndpoint("/api/controller/v2/role_team_assignments/", query)
```
gshilin-sdb, line 95: Are we going to keep v2 forever?

internal/auth/authn/aapgateway.go
gshilin-sdb: in case we'll move this initialization after
```go
orgSpecificRoles := a.mapRoleAssignmentsToOrgRoles(roleAssignments)
```
then we'll be able to initialize it with
```go
// 1 for orgRoles["*"] = globalRoles
orgRoles := make(map[string][]string, 0, len(orgSpecificRoles) + len(teamOrgRoles) + 1) 
```

pkg/aap/roles.go
asafbennatan: for now yes , this how all other aap apis are modeled

internal/auth/authn/aapgateway.go
asafbennatan: are you taking duplicate roles into account with this ? (team org roles and org specific roles could share roles)

gshilin-sdb: The underlying array can be longer that slice uses (it can be even re-allocated to take more space). 0 - means zero size, the 3rd argument is capacity. This way all appends will work without any change. Just no reallocations and copying.

gshilin-sdb (approved):

PR #2934, opened 2026-05-12
EDM-3842: Secret change detection

galel12:
## EDM-3842: Secret change detection

**Jira:** [EDM-3842](https://redhat.atlassian.net/browse/EDM-3842)

### Summary

Adds secret change detection to `flightctl-periodic`, building on the sync infrastructure from EDM-3840/EDM-3841:

* **`DependencySyncSecret` informer** — K8s SharedInformer watches secrets labeled `flightctl.io/sync=true`, computes SHA-256 fingerprints, compares against `sync_state`, and emits `DependencyChangeDetected` events on change
* **Secret fan-out query** — cross-org store query resolves affected fleets/devices from `dependency_refs` joined with `sync_state` for fingerprint comparison
* **Parameterized secret ref handling** — fleet-level refs with template parameters in namespace/name are skipped during collection and resolved to device-level refs during rollout (same pattern as git)
* **Helm RBAC** — opt-in `ClusterRole` + `ClusterRoleBinding` for `flightctl-periodic` to `list`/`watch` secrets cluster-wide, gated by `periodic.clusterLevelSecretAccess` (default `false`)

### Testing

* Unit tests for change detection, no-op, first-seen, multi-org fan-out, fleet vs device event routing, parameterized ref guards, device-level ref creation, deterministic hashing, context cancellation
* Integration tests for the secret fan-out store query (cross-org, parameterized exclusion, sync_state join)


internal/periodic_checker/server.go
```diff
@@ -85,6 +88,19 @@ func (s *Server) Run(ctx context.Context) error {
...
+		s.log.Debug("Not running in-cluster, secret informer will not start")
+	} else {
+		clientset, err := kubernetes.NewForConfig(restConfig)
+		if err != nil {
+			s.log.WithError(err).Error("Failed to create K8s clientset for secret informer")
+		} else {
+			secretInformerClientset = clientset
+		}
+	}
+
```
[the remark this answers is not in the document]
galel12: Good catch — the informer was starting unconditionally whenever in-cluster config was available, regardless of whether the RBAC was provisioned. Fixed: added `clusterLevelSecretAccess` to the Go `periodicConfig` struct, propagated it through the Helm config template, and gated the informer startup on `s.cfg.Periodic.ClusterLevelSecretAccess`. The informer now only starts when both the config flag is enabled and in-cluster config is available.

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,176 @@
...
+			d.handleSecretEvent(ctx, newObj)
+		},
+	}); err != nil {
+		d.log.WithError(err).Error("Failed to add secret informer event handler")
+		return
+	}
+
+	factory.Start(ctx.Done())
+	factory.WaitForCacheSync(ctx.Done())
+
```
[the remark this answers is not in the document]
galel12: Fixed — now checking the return map from `WaitForCacheSync` and bailing out with an error log if any informer fails to sync.

internal/tasks/populate_dependency_refs_test.go
```diff
@@ -153,6 +153,122 @@ func TestPopulateDependencyRefs_Fleet(t *testing.T) {
...
+				return okStatus
+			},
+		)
+
+		logic := NewPopulateDependencyRefsLogic(logrus.New(), mockSvc, orgId)
+		err := logic.PopulateForFleet(context.Background(), fleetName)
+
+		require.NoError(t, err)
+		assert.Empty(t, capturedRefs)
+	})
```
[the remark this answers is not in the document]
galel12: `makeSecretConfigItem` is defined in `fleet_validate_test.go` which is in the same `package tasks` test scope. Go test files within the same package share all unexported symbols — there's no need to duplicate the helper. The tests compile and pass as-is.

internal/store/dependencyref.go
```diff
@@ -183,3 +184,25 @@ func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID u
...
+// fingerprint) for all non-parameterized dependency_refs matching the given secret.
+// The query is cross-org (no org_id filter) because the K8s informer has no org
+// context — it only receives (namespace, name) from the watch event. Rows with
+// parameterized ResourceKeys are excluded since they are resolved at device level.
+func (s *DependencyRefStore) ListSecretDependencyTargets(ctx context.Context, secretNamespace, secretName string) ([]model.SecretDependencyRef, error) {
+	var refs []model.SecretDependencyRef
+	err := s.getDB(ctx).
+		Table("dependency_refs dr").
+		Select("dr.org_id, dr.fleet_name, dr.device_name, ss.fingerprint").
+		Joins("LEFT JOIN sync_states ss ON ss.org_id = dr.org_id AND ss.resource_key = dr.resource_key").
```
asafbennatan: why do we need the org id here , join just by resource key

internal/store/dependencyref.go
```diff
@@ -183,3 +184,25 @@ func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID u
...
+		Where("dr.ref_type = 'secret'").
+		Where("dr.secret_namespace = ?", secretNamespace).
+		Where("dr.secret_name = ?", secretName).
+		Where("dr.resource_key NOT LIKE ?", "%{{%").
```
asafbennatan: why?

		Where("dr.resource_key NOT LIKE ?", "%{{%").

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, ref := range refs {
+		orgRefs[ref.OrgID] = append(orgRefs[ref.OrgID], ref)
+		if _, seen := orgFingerprints[ref.OrgID]; !seen {
+			orgFingerprints[ref.OrgID] = ref.Fingerprint
+		}
+	}
+
+	now := time.Now().UTC()
+
+	for orgID, orgRefList := range orgRefs {
```
asafbennatan: im not following this logic here , 
wouldnt we just need to:
1. list all dependency targets for this secret -> this returns org_id,fleetName,deviceName -> we can already filter by fingerprint!= the one we got
2. group by org_id,fleetName,deviceName -> now we got a list of devices/fleets we need to send events for 
3.update sync state

in general this comment : 
	// Group refs by org — different orgs may have different stored fingerprints.
i dont get , the fingerprint is per resourceKey which is unique , not per org

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,182 @@
...
+		return
+	}
+
+	secret, ok := obj.(*corev1.Secret)
+	if !ok {
+		d.log.Warn("Received non-Secret object from informer")
+		return
+	}
+
+	fingerprint := d.hashFunc(secret.Data)
```
asafbennatan: why dont we use the resource version ?

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+const secretSyncLabel = "flightctl.io/sync"
```
asafbennatan: i wonder if we should make this release namespace specific , WDYT?

internal/store/dependencyref.go
galel12: The `sync_state` table is keyed by `(org_id, resource_key)` — same schema as git. Without `org_id` in the JOIN, a single `dependency_ref` row could match multiple `sync_state` rows (one per org) and produce cross-joined duplicates. That said, the fingerprint of a K8s secret doesn't vary by org, so the per-org distinction is arguably unnecessary for secrets. Keeping it for now for consistency with the git sync schema — happy to revisit if you think we should simplify `sync_state` keying for secrets.

internal/store/dependencyref.go
galel12: You're right — `collectConfigRefs` already skips parameterized secret refs at the fleet level, so these rows can't exist. Removed.

Fixed in 2df26405981094aa889dfad28c50c6dc2c5f81a0

internal/tasks/dependency_sync_secret.go
galel12: You're right — the fingerprint is per `resource_key`, not per org. Simplified: pushed the fingerprint comparison into the SQL query (`WHERE ss.fingerprint IS NULL OR ss.fingerprint != ?`), so the query only returns targets that actually need events. The Go-side grouping and per-org comparison are gone — reconcile now just iterates the returned rows, emits events for changed ones (non-nil fingerprint), and updates `sync_state` per org.

Fixed in 2df26405.

internal/tasks/dependency_sync_secret.go
galel12: `resourceVersion` changes on any mutation including metadata-only changes (e.g., adding a label or annotation). SHA-256 of `.data` gives us content-based comparison — only actual secret data changes trigger rollouts. The false positive from `resourceVersion` would be mostly harmless (TV would be identical, no device re-roll), but it adds unnecessary churn. WDYT — is the simplicity of `resourceVersion` worth the occasional no-op TV creation?

internal/tasks/dependency_sync_secret.go
galel12: If we scope to the release namespace, we can't detect changes to secrets in other namespaces — and users reference secrets from arbitrary namespaces in fleet specs. The current approach (cluster-wide, gated behind `clusterLevelSecretAccess: false`) keeps it opt-in with explicit RBAC. We could add namespace-scoped as a middle ground — watch only specific namespaces from config — but that adds complexity. Do you see namespace-scoped as the default deployment, with cluster-wide as the exception?

asafbennatan: i think there is a misunderstanding here , the namespace here will mean that a secret can decide to watched by a specific flightctl deployment assuming there could be several in the same cluster - this is the reason we include the ns as part of the role definition for example

internal/tasks/dependency_sync_secret.go
asafbennatan: i think we should prefer using the resourceVersion , there is no reason to hold this sensitive info in flightctl (consider rainbow tables , the lack of salt etc)

internal/store/dependencyref.go
asafbennatan: i think in general sync state should not be keyed by org_id.
its actually the other way around a single sync_state could have multiple dependency_ref as it should

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	fingerprint := d.hashFunc(secret.Data)
+	d.reconcile(ctx, secret.Namespace, secret.Name, fingerprint)
+}
+
+// reconcile queries for dependency targets whose stored fingerprint differs from
+// newFingerprint (or has no fingerprint yet), emits DependencyChangeDetected
+// events for changed targets, and updates sync_state.
+func (d *DependencySyncSecret) reconcile(ctx context.Context, namespace, name, newFingerprint string) {
```
asafbennatan, line 94: given the above , i dont see why this should be handled per org . please revisit it once we address the comment about orgs above

internal/store/dependencyref.go
galel12: For secrets I agree - secret:prod/db-creds is globally unique in the cluster, so sync_state doesn't need org scoping. 
But for git, the resource_key is `git:<repo-name>/<revision>` and repo-name is per-org. Two orgs could have a Repository called my-repo pointing to different git URLs so without org_id they'd collide. Should we scope this change to secrets only?

internal/tasks/dependency_sync_secret.go
galel12: Done — switched to a namespace-specific label: `flightctl.io/sync-<releaseNamespace>=true`. The release namespace is injected from Helm (`{{ .Release.Namespace }}`) into the periodic config and passed to the informer constructor. Commit: 5770b47e

internal/tasks/dependency_sync_secret.go
galel12: Agreed — switched to `resourceVersion`. Removed the SHA-256 function entirely, the handler now passes `secret.ResourceVersion` directly to reconcile. Commit: 5770b47e

internal/store/dependencyref.go
galel12: Done — using `uuid.Nil` (`00000000-0000-0000-0000-000000000000`) as the sentinel `org_id` in `sync_state` for secrets. The SQL JOIN now matches on `ss.resource_key = dr.resource_key AND ss.org_id = '00000000-...'`, and reconcile writes a single sync_state row with `uuid.Nil` instead of per-org. No schema change, git sync untouched. Commit: 5770b47e

internal/tasks/dependency_sync_secret.go
galel12: Fixed — reconcile no longer groups by org. It iterates the returned refs (already filtered by SQL), emits events using each ref's real `org_id`, and writes a single `sync_state` row with `uuid.Nil`. Commit: 5770b47e

internal/store/dependencyref.go
```diff
@@ -183,3 +184,25 @@ func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID u
...
+func (s *DependencyRefStore) ListSecretDependencyTargets(ctx context.Context, secretNamespace, secretName, newFingerprint string) ([]model.SecretDependencyRef, error) {
+	var refs []model.SecretDependencyRef
+	err := s.getDB(ctx).
+		Table("dependency_refs dr").
+		Select("dr.org_id, dr.fleet_name, dr.device_name, ss.fingerprint").
+		Joins("LEFT JOIN sync_states ss ON ss.resource_key = dr.resource_key AND ss.org_id = '00000000-0000-0000-0000-000000000000'").
+		Where("dr.ref_type = 'secret'").
+		Where("dr.secret_namespace = ?", secretNamespace).
+		Where("dr.secret_name = ?", secretName).
+		Where("ss.fingerprint IS NULL OR ss.fingerprint != ?", newFingerprint).
```
asafbennatan, line 200: do we need this OR ? if its null it will be different from the provided value , no ?

internal/store/dependencyref.go
```diff
@@ -183,3 +184,25 @@ func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID u
...
+// for all dependency_refs matching the given secret where the stored fingerprint
+// differs from newFingerprint (or no fingerprint exists yet). Secret sync_state
+// rows use uuid.Nil as org_id (secrets are globally unique, not per-org), so the
+// join matches on resource_key with the sentinel org_id.
+func (s *DependencyRefStore) ListSecretDependencyTargets(ctx context.Context, secretNamespace, secretName, newFingerprint string) ([]model.SecretDependencyRef, error) {
+	var refs []model.SecretDependencyRef
+	err := s.getDB(ctx).
+		Table("dependency_refs dr").
+		Select("dr.org_id, dr.fleet_name, dr.device_name, ss.fingerprint").
+		Joins("LEFT JOIN sync_states ss ON ss.resource_key = dr.resource_key AND ss.org_id = '00000000-0000-0000-0000-000000000000'").
```
asafbennatan: there is just no need to mention org id here at all

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if len(refs) == 0 {
+		return
+	}
+
+	resourceKey := fmt.Sprintf("secret:%s/%s", namespace, name)
+	now := time.Now().UTC()
+
+	for _, ref := range refs {
+		if ref.Fingerprint != nil {
```
asafbennatan: i dont understand why do we care about fingerprint!=null ?

internal/tasks/dependency_sync_secret.go
```diff
@@ -0,0 +1,135 @@
...
+		AddFunc: func(obj interface{}) {
+			d.handleSecretEvent(ctx, obj)
+		},
+		UpdateFunc: func(_, newObj interface{}) {
+			d.handleSecretEvent(ctx, newObj)
+		},
+	}); err != nil {
+		d.log.WithError(err).Error("Failed to add secret informer event handler")
+		return
+	}
```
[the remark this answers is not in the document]
galel12: deliberately omitted for consistency with git sync which also doesn't handle deletes. the worker will surface a clear error at render time if the secret is gone

internal/store/dependencyref.go
galel12: in SQL `NULL != 'value'` evaluates to NULL, not TRUE — so the row gets excluded. the IS NULL catches the first-seen case where no sync_state row exists yet (LEFT JOIN produces NULL)

internal/store/dependencyref.go
galel12: removed, resource_key prefix (`secret:`) prevents collision with git entries
f143870d84dcd4caf1ce58c00d6562110f3929a6

internal/tasks/dependency_sync_secret.go
galel12: when fingerprint is nil it means no sync_state exists yet — first time we see this secret. without this check every informer restart would emit DependencyChangeDetected for all watched secrets during initial cache sync

asafbennatan: we are already getting all the defs that :

have different fingerprint (ref.fingerprint!=null)
dont have any sync state (ref.fingerprint==null)


in both cases i expect an emit event
in both cases we upsert the syncstate data

whats the difference ?

galel12: you're right, the SQL already filters — if a row came back it needs an event regardless. removed the distinction. a372f2d8

asafbennatan (approved):
LGTM

PR #2936, opened 2026-05-12
EDM-3870: Remove db migration job before a new one is created

rawagner:


deploy/helm/flightctl/templates/db/flightctl-db-migration-job.yaml
```diff
@@ -22,20 +22,12 @@
 ---
 {{- end }}
 
-{{- if .Release.IsUpgrade }}
+{{- $hookPhase := ternary "pre-upgrade" "post-install" .Release.IsUpgrade }}
```
sk-ilya, line 25: Won't it cause issues with
https://github.com/flightctl/flightctl/blob/1840251fcc03e42d2b1917fa843fe7887861c577/deploy/helm/flightctl/templates/_helpers.tpl#L451-L464
?

rawagner: You are right, this can cause issues.

I've updated the helper to check job's revision against the expected revision (IOW expected revision is the pod's revision).

the job name is now static, so the querying is simpler (not by label selector, but by job name)

`job_revision == expected` or `job not found` -> wait for the job
`job_revision > expected`  -> This is a corner case where user started an upgrade and didnt enable pod scale-down and before the migration job finished, an old pod was restarted. When the old pod starts it looks up the migration job and if the job_revision is higher than expected revision it means that the upgrade is in progress, the old migration was completed and the pod can start.

sk-ilya: There is still a chicken-and-egg problem on install. The pods wait for the job, but the job runs post-install, so the deployment will fail when using `--wait`.
Perhaps it would be cleaner for the migration job to record the schema (=app) version in the DB itself? and for the init containers to just wait until that version is >= the required version, instead of relying on hooks ordering / presence of jobs / other fragile things... The migration  job can be launched as a regular job in this case, not as a hook.

rawagner: `--wait` is not new - we always had this problem. If we want to fix that, we should have a separate JIRA/PR for it.

sk-ilya (approved):

PR #2939, opened 2026-05-13
NO-ISSUE: Fix console flakiness

eldar101:
Make StartFlightCtlAgent restart the agent so test config changes are picked up even when the service is already running.

The console E2E setup now resets enrollment config before enrolling, preventing stale agent configuration from making the test fail before it reaches the console assertion.


Samudelacruz77 (approved):
lgtm

hferber1 (approved):
lgtm

PR #2940, opened 2026-05-13
EDM-3480: dnf retry/timeout/skip parameters

asafbennatan:
adds retry/timeout/skip defaults so air-gaped env works (otherwise dnf waits for too long trying to connect to CentOS public repo which is not available) the defaults are suitable for non air-gaped as well and can be controlled if necessary


internal/imagebuilder_worker/tasks/templates/Containerfile.tmpl
```diff
@@ -14,10 +14,18 @@ ARG AGENT_CONFIG_DEST_PATH=/etc/flightctl/config.yaml
...
 RUN if [ "${RPM_REPO_ADD}" = "true" ]; then \
         dnf -y config-manager --add-repo ${RPM_REPO_ADD_URL}; \
     fi && \
-    dnf install -y ${RPM_REPO_ENABLE:+--enablerepo=${RPM_REPO_ENABLE}} flightctl-agent && \
+    dnf install -y \
+        --setopt=timeout=${DNF_TIMEOUT} \
+        --setopt=retries=${DNF_RETRIES} \
+        --setopt=skip_if_unavailable=${DNF_SKIP_UNAVAILABLE} \
+        ${RPM_REPO_ENABLE:+--enablerepo=${RPM_REPO_ENABLE}} \
+        flightctl-agent && \
```
itzikb-redhat, line 28: The dnf settings are needed for open-vm-tools (below) as well.

deploy/podman/flightctl-imagebuilder-worker/flightctl-imagebuilder-worker-config/config.yaml.template
```diff
@@ -39,6 +39,15 @@ imageBuilderWorker:
   {{- if .imagebuilderWorker.rpmRepoEnable}}
   rpmRepoEnable: {{.imagebuilderWorker.rpmRepoEnable}}
   {{- end}}
+  {{- if .imagebuilderWorker.dnfTimeout}}
```
galel12: 0 is falsy, so this conditional drops `dnfTimeout: 0` (which means "no timeout" in dnf). The dnfRetries conditional below correctly handles zero with the `or (eq (printf "%v" ...) "0")` pattern. This should match:

```bash
{{- if or (eq (printf "%v" .imagebuilderWorker.dnfTimeout) "0") .imagebuilderWorker.dnfTimeout}}
```

asafbennatan: fixed

galel12 (approved):
LGTM

PR #2942, opened 2026-05-13
EDM-3856: added RBAC for vulnerabilities

asafbennatan:


celdrake (approved):

internal/service/vulnerability.go
```diff
@@ -349,7 +349,7 @@ func (h *ServiceHandler) GetDeviceVulnerabilities(
 		result, err = h.store.VulnerabilityFinding().GetVulnerabilities(ctx, imageDigest, listParams)
 		if err != nil {
 			h.log.WithError(err).Error("failed to query vulnerability findings")
-			return nil, domain.StatusInternalServerError("failed to query vulnerability findings")
+			return nil, StoreErrorToApiStatus(err, false, domain.VulnerabilityKind, nil)
```
galel12, line 352: The old code returned an opaque "failed to query vulnerability findings" to the client. StoreErrorToApiStatus's default path does `StatusInternalServerError(err.Error())`, which forwards the raw store/Postgres error to the API response. If the query fails for a reason that's not `ErrResourceNotFound` or a known bad-request/conflict, the 500 body could leak table names, column names, or query fragments.

Consider keeping the generic message for the 500 path - e.g. only use `StoreErrorToApiStatus` for the specific error types it distinguishes (not-found, bad-request, conflict) and fall back to the original opaque string for the default case. Same applies to the other 4 sites changed in this PR.

internal/service/vulnerability.go
```diff
@@ -398,7 +398,7 @@ func (h *ServiceHandler) GetDeviceVulnerabilitySummary(
 		counts, err = h.store.VulnerabilityFinding().GetVulnerabilitySummary(ctx, imageDigest, fs)
 		if err != nil {
 			h.log.WithError(err).Error("failed to compute device vulnerability summary")
-			return nil, domain.StatusInternalServerError("failed to compute device vulnerability summary")
+			return nil, StoreErrorToApiStatus(err, false, domain.DeviceVulnerabilitySummaryKind, nil)
```
galel12, line 401: nit: name is nil in all 5 call sites, so the `ErrResourceNotFound` path would produce `"Vulnerability 'none' not found"` - a confusing 404 for what are list/digest-based queries with no single resource name. Might be worth passing a more descriptive identifier (e.g. the image digest or device name) or confirming that the store layer can't return ErrResourceNotFound for these code paths.

internal/service/vulnerability.go
asafbennatan: this is the common pattern in the codebase , if we want to address this we should do codebase wide in a separate PR probably

internal/service/vulnerability.go
asafbennatan: it cannot return 404 these are listing apis

PR #2945, opened 2026-05-13
EDM-3783: Relative PAM Issuer paths

keitwb:
Instead of a fixed absolute path for the login and authorize paths, use
relative paths.

Also use relative paths for the web assets.

This will allow the login to work in different deployment models where the
PAM Issuer is not at the root path of the URL.

### DIFF
--- a/internal/auth/oidc/pam/provider.go
+++ b/internal/auth/oidc/pam/provider.go
@@ -1063,7 +1063,7 @@ func (s *PAMOIDCProvider) Login(ctx context.Context, username, password, encrypt
 	s.log.Debugf("Login: created encrypted session cookie for %s", username)
 
 	// Redirect back to authorization endpoint with all parameters
-	authURL := fmt.Sprintf("/api/v1/auth/authorize?response_type=code&client_id=%s&redirect_uri=%s",
+	authURL := fmt.Sprintf("authorize?response_type=code&client_id=%s&redirect_uri=%s",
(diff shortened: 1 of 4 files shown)


asafbennatan:
@keitwb not sure i follow , will we now always serve the pam issuer ui from / and the api will be at /login ?
otherwise not sure hows thats compatible

keitwb:
> @keitwb not sure i follow , will we now always serve the pam issuer ui from / and the api will be at /login ? otherwise not sure hows thats compatible

The page where this gets called is at `/api/v1/auth/authorize` on the PAM issuer server. With a relative path this will result in the fetch posting to `/api/v1/auth/login`. This means that the PAM Issuer service can be run as a namespaced subpath off the main hostname for the services deployment, which is what we're going to do with the gateway (e.g. something like `/_/pam-issuer/api/v1/auth/login`.

But it still has to work for the helm deployment model so we can't just hardcode the new path there so this seems to be the best way to make it work for all deployments without requiring special config all over the place.

keitwb:
I also realized that the authorize path used as the redirect URI after login needed to be relative as well to be consistent. It works the same way.

asafbennatan:
note that currently pam issuer is available only on RHEL

keitwb:
> note that currently pam issuer is available only on RHEL

Ok, this is still the simplest way to do this since it decouples the deployment model of the pam issuer from the paths embedded in the service so that the service doesn't have to know or care, just be consistent in the relative pathing of how it refers to resources.

asafbennatan (approved):
LGTM

PR #2948, opened 2026-05-13
EDM-3843: HTTP change detection

galel12:
## EDM-3843: HTTP change detection

**Jira:** [EDM-3843](https://redhat.atlassian.net/browse/EDM-3843)

### Summary

Adds HTTP change detection to `flightctl-periodic`, building on the sync infrastructure from EDM-3840/EDM-3841:

* **`DependencySyncHttp` periodic task** — sends conditional GETs (ETag / Last-Modified) against HTTP-type repositories, compares fingerprints against `sync_state`, emits `DependencyChangeDetected` events on change
* **Conditional header dispatch** — sends `If-None-Match` for ETags, `If-Modified-Since` for HTTP-dates (RFC 7232 compliant)
* **Parameterized suffix handling** — fleet-level refs with template parameters in suffix are skipped during collection and resolved to device-level refs during rollout (same pattern as git/secret)
* **Skip-probe rate limiting** — endpoints that don't support conditional requests (HttpNotConditional) or return errors still update `last_checked_at` to avoid re-polling every cycle

### Testing

* Unit tests for change detection, 304 no-change, first-seen, HttpNotConditional, HTTP errors, multi-fleet/device fan-out, parameterized suffix lifecycle, nil RepoSpec guard
* Integration tests for `ListDueHttpDependencies` store query (poll-interval gate, multi-fleet aggregation, org isolation, repo spec JOIN)

### DIFF
--- a/internal/store/dependencyref.go
+++ b/internal/store/dependencyref.go
@@ -25,6 +25,7 @@ type DependencyRef interface {
 	ReplaceByStandaloneDevice(ctx context.Context, orgID uuid.UUID, deviceName string, refs []model.DependencyRef) error
 	BulkUpsertDeviceRefs(ctx context.Context, orgID uuid.UUID, refs []model.DependencyRef) error
 	ListDueGitDependencies(ctx context.Context, orgID uuid.UUID, pollInterval time.Duration) ([]model.GitDependencyProbe, error)
+	ListDueHttpDependencies(ctx context.Context, orgID uuid.UUID, pollInterval time.Duration) ([]model.HttpDependencyProbe, error)
 	ListSecretDependencyTargets(ctx context.Context, secretNamespace, secretName, newFingerprint string) ([]model.SecretDependencyRef, error)
 }
 
@@ -178,6 +179,31 @@ func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID u
 		Where("dr.revision NOT LIKE ?", "%{{%").
 		Where("ss.last_checked_at IS NULL OR ss.last_checked_at + ? * INTERVAL '1 second' < NOW()", pollInterval.Seconds()).
 		Group("dr.repository_name, dr.revision, ss.fingerprint, r.spec").
+		Scan(&probes).Error
+	if err != nil {
+		return nil, ErrorFromGormError(err)
+	}
+	return probes, nil
+}
+
+// ListDueHttpDependencies returns HTTP dependency probes that are due for
+// polling. Mirrors ListDueGitDependencies but filters by ref_type='http' and
+// groups by (repository_name, http_suffix).
+func (s *DependencyRefStore) ListDueHttpDependencies(ctx context.Context, orgID uuid.UUID, pollInterval time.Duration) ([]model.HttpDependencyProbe, error) {
+	var probes []model.HttpDependencyProbe
+	err := s.getDB(ctx).
+		Table("dependency_refs dr").
+		Select(
+			"dr.repository_name, dr.http_suffix, ss.fingerprint, r.spec AS repo_spec, "+
+				"array_agg(DISTINCT dr.fleet_name) FILTER (WHERE dr.fleet_name <> '' AND dr.device_name = '') AS fleet_names, "+
+				"array_agg(DISTINCT dr.device_name) FILTER (WHERE dr.device_name <> '') AS device_names",
+		).
+		Joins("LEFT JOIN sync_states ss ON ss.org_id = dr.org_id AND ss.resource_key = dr.resource_key").
+		Joins("LEFT JOIN repositories r ON r.org_id = dr.org_id AND r.name = dr.repository_name").
+		Where("dr.org_id = ?", orgID).
+		Where("dr.ref_type = 'http'").
+		Where("ss.last_checked_at IS NULL OR ss.last_checked_at + ? * INTERVAL '1 second' < NOW()", pollInterval.Seconds()).
+		Group("dr.repository_name, dr.http_suffix, ss.fingerprint, r.spec").
 		Scan(&probes).Error
 	if err != nil {
 		return nil, ErrorFromGormError(err)
--- a/internal/tasks/dependency_sync_http.go
+++ b/internal/tasks/dependency_sync_http.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
@@ -104,7 +105,7 @@ func (d *DependencySyncHttp) probeEndpoint(ctx context.Context, probe *model.Htt
 	}
 
 	repoURL := httpSpec.Url + probe.HTTPSuffix
-	rk := fmt.Sprintf("http:%s/%s", probe.RepositoryName, probe.HTTPSuffix)
+	rk := httpResourceKey(probe.RepositoryName, probe.HTTPSuffix)
 
 	storedFP := ""
 	if probe.Fingerprint != nil {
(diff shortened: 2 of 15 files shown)


internal/store/dependencyref.go
```diff
@@ -185,6 +186,31 @@ func (s *DependencyRefStore) ListDueGitDependencies(ctx context.Context, orgID u
...
+		Where("dr.org_id = ?", orgID).
+		Where("dr.ref_type = 'http'").
+		Where("ss.last_checked_at IS NULL OR ss.last_checked_at + ? * INTERVAL '1 second' < NOW()", pollInterval.Seconds()).
+		Group("dr.repository_name, dr.http_suffix, ss.fingerprint, r.spec").
+		Scan(&probes).Error
+	if err != nil {
+		return nil, ErrorFromGormError(err)
+	}
+	return probes, nil
+}
```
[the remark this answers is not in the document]
galel12: By design, `collectConfigRefs` already skips parameterized HTTP suffixes at the fleet level (same guard as git/secret), so `dependency_refs` never contains `{{ }}` values. The matching filter in the git query is redundant defense-in-depth. Skipping this to keep the queries consistent and avoid implying the table can contain them.

internal/tasks/dependency_sync_http.go
```diff
@@ -0,0 +1,254 @@
...
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", 0, fmt.Errorf("sending request: %w", err)
+	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
galel12: Tied to the parameterized suffix filter discussion above — since `collectConfigRefs` prevents these rows from entering `dependency_refs` in the first place, there's nothing to test at the SQL level.

test/integration/store/dependencyref_test.go
```diff
@@ -312,4 +315,219 @@ var _ = Describe("DependencyRefStore", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(probes).To(BeEmpty())
+		})
+
+		It("should return empty when no HTTP refs exist", func() {
+			probes, err := storeInst.DependencyRef().ListDueHttpDependencies(ctx, orgId, 15*time.Minute)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(probes).To(BeEmpty())
+		})
+	})
```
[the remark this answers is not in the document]
galel12: Same reasoning as [above](https://github.com/flightctl/flightctl/pull/2948#discussion_r3235822231) — parameterized suffixes never reach `dependency_refs` so there's nothing for the SQL to filter. Adding such a test would be misleading.

internal/tasks/dependency_sync_http.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	spec := probe.RepoSpec.Data
+	httpSpec, err := spec.AsHttpRepoSpec()
+	if err != nil {
+		d.log.WithError(err).Warnf("failed decoding HTTP spec for repository %s", probe.RepositoryName)
+		return httpProbeResult{probe: probe, skip: true}
+	}
+
+	repoURL := httpSpec.Url + probe.HTTPSuffix
```
asafbennatan: should we use proper URL joining here ?

internal/tasks/dependency_sync_http.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+}
+
+func httpResourceKey(repoName, suffix string) string {
+	return fmt.Sprintf("http:%s/%s", repoName, strings.TrimPrefix(suffix, "/"))
+}
+
+// httpConditionalGet sends a conditional GET to the given URL using stored
+// fingerprint for If-None-Match/If-Modified-Since headers.
+func httpConditionalGet(ctx context.Context, repoURL string,
```
asafbennatan: you are doing a conditional get which returns the data and then later you discard it , instead just get the fingerprint with HEAD

internal/tasks/dependency_sync_http.go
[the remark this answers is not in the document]
galel12: Apologies — this reply landed on the wrong thread. The ctx/timeout fix was addressed in 8c7b961: `http.NewRequestWithContext(ctx, ...)` + `Timeout: 30s` on the client.

Follow-up in 6c62470: switched from GET to HEAD since we only need headers, making the body drain moot.

internal/tasks/dependency_sync_http.go
galel12: Good catch — switched to `url.JoinPath` in 6c62470.

internal/tasks/dependency_sync_http.go
galel12: Great point — switched to HEAD in 6c62470. We only need the ETag/Last-Modified headers, so there's no reason to transfer the body. Conditional semantics (If-None-Match/If-Modified-Since → 304) work identically with HEAD per RFC 7232.

internal/tasks/dependency_sync_http_test.go
```diff
@@ -0,0 +1,367 @@
...
+	"github.com/flightctl/flightctl/internal/store/model"
+	"github.com/google/uuid"
+	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	gomock "go.uber.org/mock/gomock"
+)
+
+func httpRepoSpec(t *testing.T, url string) *model.JSONField[api.RepositorySpec] {
```
asafbennatan, line 25: use proper http server for testing this i would expect the following tests :
1. endpoint with etag
2. endpoint with etag without support for conditional head
3. endpoint with last modified
4. endpoint with last modified without support for conditional head
5.endpoint with no HEAD support (405)
6. other 4xx and 5xx responses

galel12: Done in 01a8000 — added `TestHttpConditionalHead` with `httptest.NewServer` covering all 6 cases.

internal/tasks/dependency_sync_http.go
```diff
@@ -0,0 +1,261 @@
...
+	)
+
+	sem := make(chan struct{}, d.maxConcurrent)
+	var wg sync.WaitGroup
+
+	for i := range probes {
+		probe := &probes[i]
+		wg.Add(1)
+		sem <- struct{}{}
+		go func() {
```
asafbennatan, line 84: should we set proper context timeout ?

internal/tasks/dependency_sync_http.go
```diff
@@ -0,0 +1,261 @@
...
+		// (or vice versa) is a protocol violation that may cause servers to
+		// ignore the condition or return 400.
+		if strings.HasPrefix(storedFingerprint, `"`) || strings.HasPrefix(storedFingerprint, `W/"`) {
+			req.Header.Set("If-None-Match", storedFingerprint)
+		} else {
+			req.Header.Set("If-Modified-Since", storedFingerprint)
+		}
+	}
+
+	client := &http.Client{
```
asafbennatan: should we create a client for each probe?

internal/tasks/dependency_sync_http.go
galel12: Added per-probe `context.WithTimeout(ctx, 30s)` in e6a7454.

internal/tasks/dependency_sync_http.go
galel12: Refactored in e6a7454 — probes are now grouped by repo, one `http.Client` is built per repo group and shared across all probes in the group.

internal/tasks/dependency_sync_http.go
galel12: fixed also for the git case d261bc56f4f92cf6c1cdcd75c643d98cd7c4b215

asafbennatan (approved):
LGTM

PR #2955, opened 2026-05-14
EDM-3919: Provision default catalog

rawagner:


internal/service/org_provisioner.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/flightctl/flightctl/internal/store"
+	"github.com/flightctl/flightctl/internal/store/model"
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+)
+
+// OrgProvisionerInterface is implemented by anything that can provision default resources for new orgs.
+type OrgProvisionerInterface interface {
+	EnsureDefaults(ctx context.Context, orgs []*model.Organization)
+}
```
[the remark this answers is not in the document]
rawagner: failure to provision default catalog wont fail the whole request - this is per design doc.

internal/store/store.go
```diff
@@ -247,7 +256,41 @@ func (s *DataStore) customizeMigration(ctx context.Context) error {
...
+		}
+
+		displayName := domain.DefaultCatalogDisplayName
+		name := domain.DefaultCatalogName
+		return tx.Exec(`
+			INSERT INTO catalogs (org_id, name, spec, labels, annotations, created_at, updated_at)
+			SELECT o.id, ?, ?, '{}'::jsonb, '{}'::jsonb, NOW(), NOW()
+			FROM organizations o
+			WHERE NOT EXISTS (
+				SELECT 1 FROM catalogs c WHERE c.org_id = o.id
```
asafbennatan: when we migrate an existing org without default catalog but with other catalogs - > should we create the default catalog or not ? (current code does not as far as i understand ) 
is that the intended behavior - i suspect that its not

rawagner: just for the reference, we agreed to provision default catalog only if the org doesnt already have any other that doesnt have metadata.owner.

asafbennatan (approved):
LGTM

PR #2957, opened 2026-05-14
EDM-2017: Modernize RPM spec for FIPS, versioning, and multi-RHEL support

kkyrazis:
## Summary

Fixes and modernizes the upstream RPM spec:
- Enables FIPS on all RHEL versions (was hardcoded to RHEL 9 only)
- Fixes version injection for tarball builds using git-metadata with export-subst
- Replaces hardcoded RHEL 10 conditionals with dynamic macro construction (`el%{rhel}`)
- Adds vendored license collection for the CLI subpackage
- Removes redundant grafana UID chown (handled by the container)
- Correctly pins the rhel9 quadlet images

## Test plan
- [x] Local `make rpm` build succeeds, all 6 sub-packages produced
- [ ] Packit COPR build validates Fedora builds
- [ ] `flightctl version` shows correct commit when built from tarball


packaging/rpm/flightctl.spec
```diff
@@ -15,23 +43,25 @@ Version:        0.6.0
...
+%gometa -L -f
 
 License:        Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND ISC AND MIT
 URL:            %{gourl}
 
-Source0:        1%{?dist}
+Source:         %{gosource}
+Source1:        source-patches.inc
+
+%include %{S:1}
```
keitwb: What does this do?

packaging/rpm/flightctl.spec
```diff
@@ -3,10 +3,38 @@
...
+%global images_config     packaging/images/%{?rhel:el%{rhel}}%{!?rhel:el9}/images.yaml
 %global selinuxtype targeted
 %define selinux_policyver 3.14.3-67
+%global flightctl_target flightctl.target
+
+# Downstream override hooks
+%global extra_ldflags %{nil}
+%global rpm_build_flags %{nil}
+%global rpm_ld_flags %{nil}
+%global set_proxy \
```
keitwb: Can the downstream build process just set the right values in go.env in the go root before building the RPM spec? Even if not, it seems like there is a simpler way to do this, just just making the value of the GOPROXY var be the rpm var instead of this code block.

packaging/rpm/flightctl.spec
kkyrazis: https://rpm.org/docs/4.20.x/manual/macros#spec-specific-macros

Includes sources as part of the spec. Downstream it would like something like:

```
P0: something.patch
P1: something_else.patch
```

As we get more patches this just makes it easier to add patches without having to directly update the spec (and probably running into conflicts)

packaging/rpm/flightctl.spec
```diff
@@ -427,10 +466,9 @@ fi
 # File listings
 # No %%files section for the main package, so it won't be built
 
-%files cli
+%files cli -f licenses.list
```
keitwb, line 458: Does upstream need to have a licenses.list file or does something autogenerate this?

packaging/rpm/flightctl.spec
kkyrazis: Honestly this is purely an upstream thing. Downstream doesn't adjust proxies directly. I'm setting this to nil downstream

packaging/rpm/flightctl.spec
kkyrazis: It's part of this PR:
```
    # Collect license files (top-level + vendored dependencies if present)
    rm -f licenses.list
    find -type f -name LICENSE -or -name License | while read LICENSE_FILE; do
        install -Dv -m0644 "${LICENSE_FILE}" "%{buildroot}%{_datadir}/licenses/%{NAME}/${LICENSE_FILE}"
        echo "%%license %{_datadir}/licenses/%{NAME}/${LICENSE_FILE}" >> licenses.list
    done
    touch licenses.list
```

packaging/rpm/flightctl.spec
```diff
@@ -250,41 +280,39 @@ if [ -f /run/flightctl-observability-was-active ]; then
...
-        export GOPROXY='https://proxy.golang.org,direct'
-    fi
+    %{?set_proxy}
 
-    # Prefer values injected by Makefile/CI; fall back to RPM macros when unset
-    SOURCE_GIT_TAG="%{?SOURCE_GIT_TAG:%{SOURCE_GIT_TAG}}%{!?SOURCE_GIT_TAG:%(echo "v%{version}" | tr '~' '-')}" \
-    SOURCE_GIT_TREE_STATE="%{?SOURCE_GIT_TREE_STATE:%{SOURCE_GIT_TREE_STATE}}%{!?SOURCE_GIT_TREE_STATE:clean}" \
-    SOURCE_GIT_COMMIT="%{?SOURCE_GIT_COMMIT:%{SOURCE_GIT_COMMIT}}%{!?SOURCE_GIT_COMMIT:%(
-        commit=$(git rev-parse --short HEAD 2>/dev/null);
+    SOURCE_GIT_TAG="%(echo "v%{version}" | tr '~' '-')" \
```
keitwb: This needs to be checked to ensure it works right with the RC versions still (that the version gets properly relayed to the binaries so that `flightctl version` shows the right things.

https://github.com/flightctl/flightctl/pull/2086/changes#diff-69dae8b92ae5b5df364c7fc6cd5c885983542c8606b939da07b692af6cf5cd85

which was reverted in:

https://github.com/flightctl/flightctl/pull/2130

keitwb: I seem to remember us being dependent on that SOURCE_GIT_TAG var getting set in the Makefile.

keitwb: Maybe `${version}` will work but it definitely needs to be confirmed as the issue was very subtle and was not obvious at all during the RPM build process.

kkyrazis: Going through now to triple check this. When I was looking at it I couldn't find anywhere that actually invoked the rpm build with these values defined and to the best of my knowledge, we were always using the defaulted values. I'm trying to see if I can just remove these and let the make scripts, which calls the hack, handle everything.

kkyrazis: I've made changes so that we prefer calling the hack script, but for upstream it didn't _really_ matter. The `.packit.yaml` file sets the version to the result of the hack script, so the previous behavior worked as expected. I also found that upstream releases use git archives, so we've been releasing artifacts with unkown commits:
```+ SOURCE_GIT_TAG=v1.1.2
+ SOURCE_GIT_TREE_STATE=clean
+ SOURCE_GIT_COMMIT=unknown
```

I've pushed a fix for that too, but downstream will still utilize git commands.

keitwb (approved):

packaging/rpm/flightctl.spec
```diff
@@ -3,10 +3,38 @@
...
+%global fips_enabled 0%{?rhel}
+%global fips_validator %{go_bin_dir}/fips-validator
+%if %{fips_enabled}
+%global disable_fips %{nil}
+%else
+%global disable_fips DISABLE_FIPS="true"
+%endif
+
+# Package metadata
+%global grafana_uid 472
```
keitwb: This seems redundant to https://github.com/flightctl/flightctl/blob/main/deploy/podman/flightctl-grafana/flightctl-grafana.container#L37 and hopefully we could just remove it from the RPM spec.

packaging/rpm/flightctl.spec
```diff
@@ -3,10 +3,38 @@
 
 # Define the Go Import Path
 %global goipath github.com/flightctl/flightctl
```
keitwb, line 4: Does this change for downstream?

kkyrazis: No, this was originally defined by upstream and actually only used downstream. I assume this reference was made unnecessary once upstream transitioned to using make_build, but I don't know for sure.

packaging/rpm/flightctl.spec
kkyrazis: The world will never know... I'll test this out and see what I can find.

packaging/rpm/flightctl.spec
```diff
@@ -645,21 +692,13 @@ if [ "$1" -eq 2 ]; then
             TMPSCRIPT=$(mktemp /tmp/flightctl-dry-run.XXXXXX.sh)
             cp "$SCRIPT" "$TMPSCRIPT"
             chmod +x "$TMPSCRIPT"
-%if 0%{?rhel} == 10
```
keitwb, line 648: Given that this is already defined in the `images.yaml` files for each distro version in up and down stream, can't we read that dynamically at build time to figure out the right image instead of having to macro it?

packaging/rpm/flightctl.spec
kkyrazis: Confirmed that this is redundant and can be removed from the spec. All downstream images resolve to 1001 (whats hardcoded there) and the upstream ones resolve to 472. Regardless, whatever happens in the StartPre takes precedence over what we do in the container at install time, so I'll make sure to remove these lines.

.packit.yaml
```diff
@@ -5,6 +5,7 @@ packages:
     # add or remove files that should be synced
     files_to_sync:
       - packaging/rpm/flightctl.spec
+      - packaging/rpm/source-patches.inc
```
galel12: This file doesn't exist in the repo and isn't created by this PR. 
The spec uses `Source1: overrides.inc` instead, this looks like a stale reference from a prior iteration

Should this be removed, or does `source-patches.inc` need to be created as a separate placeholder alongside `overrides.inc`?

packaging/rpm/flightctl.spec
```diff
@@ -323,6 +360,14 @@ fi
 
     install -Dpm 0440 packaging/rpm/sudoers.d/flightctl %{buildroot}/etc/sudoers.d/flightctl
 
+    # Collect license files (top-level + vendored dependencies if present)
+    rm -f licenses.list
+    find -type f -name LICENSE -or -name License | while read LICENSE_FILE; do
```
galel12: `-and` (implicit) binds tighter than `-or`, so this parses as `(-type f -name LICENSE) -or (-name License)`. 

the second branch matches any filesystem entry named "License" regardless of type (including directories).

maybe go for this instead for explicit grouping:
```bash
find -type f \( -name LICENSE -o -name License \) | while read LICENSE_FILE; do
```

.packit.yaml
kkyrazis: based on our discussion earlier, I will be reverting a large percentage of this. The "automatically" update downstream spec wasn't well liked and won't really be necessary as the intention is that everyone will be making their own changes. 

To answer the question though, overrides.inc was originally source-patches.inc so this should have been renamed.

packaging/rpm/flightctl.spec
kkyrazis: Nice find. This was copied directly from downstream so I'll ensure we update it there too.

packaging/rpm/flightctl.spec
kkyrazis: Ok, turns out this a requirement for %gometa 
```specfile.exceptions.RPMException: Please set the Go import path in the %{goipath} variable before calling %gometa```

File: packaging/rpm/flightctl.spec (hunk 7 above)
Comment (kkyrazis) [reply packaging/rpm/flightctl.spec:648]: Ideally we could, but as this is a `%pre` script, the `images.yaml` aren't available the the rpm at this point. Even if the `images.yaml` file was bundled with the rpm (currently it's not) it's still not available for consuming because the current version's files aren't installed at this point.

File: packaging/rpm/flightctl.spec [hunk 10]
```diff
@@ -254,32 +250,47 @@ fi
...
-    %else
-        DISABLE_FIPS="true" %make_build build-cli build-agent build-restore build-standalone
-    %endif
+    )" \
+    %{?disable_fips} %make_build build-cli build-agent build-restore build-standalone
 
     # SELinux modules build
     %make_build --directory packaging/selinux
 
+    GOFLAGS='' GOBIN="$PWD/bin" go install github.com/flightctl/fips-validator@v0.0.0-20250930084220-ceca2caa6e48
```
galel12, line 295: This runs on all platforms but `fips-validator` is only used in `%check` under `%if %{fips_enabled}`. 
On Fedora this just wastes build time and adds an unnecessary network fetch. 

Should be wrapped:
```bash
%if %{fips_enabled}
    GOFLAGS='' GOBIN="$PWD/bin" go install github.com/flightctl/fips-validator@v0.0.0-20250930084220-ceca2caa6e48
%endif
```

galel12 (approved):
LGTM

galel12:
@kkyrazis just wanted to say I updated the PR description to reflect the current state of the changes (the old one still referenced the overridable macros, Makefile changes, etc. that were removed). Should make it easier to understand if we come back to this later.

galel12 (approved):
LGTM

PR #2958, opened 2026-05-14
EDM-3939: Fix SyncStateStore.Set() upsert for uuid.Nil org_id

galel12:
## Problem

`SyncStateStore.Set()` fails on the second write when `org_id` is `uuid.Nil`, returning:

```
a resource with this name already exists
```

This breaks secret sync state persistence — the first write succeeds but every subsequent secret change fails to update the stored fingerprint.

## Root Cause

GORM's `Save()` uses `reflect.IsZero()` to decide between INSERT and UPDATE. Since `uuid.Nil` is the zero value for `uuid.UUID`, GORM always chooses INSERT, hitting the primary key constraint on the second call.

## Fix

Replace `Save()` with an explicit `ON CONFLICT (org_id, resource_key) DO UPDATE` clause — the same pattern already used by `BulkUpsert()` in the same file.

## Testing

- Regression test added in `test/integration/store/syncstate_test.go` ("When upserting with uuid.Nil org_id")
- Confirmed test **fails** without fix (duplicate key error) and **passes** with fix
- All 7 SyncStateStore integration tests pass
- All 7 DependencySyncSecret unit tests pass

### DIFF
--- a/internal/store/syncstate.go
+++ b/internal/store/syncstate.go
@@ -60,7 +60,7 @@ func (s *SyncStateStore) Set(ctx context.Context, orgID uuid.UUID, state *model.
 	state.OrgID = orgID
 	result := s.getDB(ctx).Clauses(clause.OnConflict{
 		Columns:   []clause.Column{{Name: "org_id"}, {Name: "resource_key"}},
-		DoUpdates: clause.AssignmentColumns([]string{"fingerprint", "last_checked_at", "last_change_at"}),
+		UpdateAll: true,
 	}).Create(state)
 	if result.Error != nil {
 		return ErrorFromGormError(result.Error)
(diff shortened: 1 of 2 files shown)


internal/store/syncstate.go
```diff
@@ -58,7 +58,10 @@ func (s *SyncStateStore) Set(ctx context.Context, orgID uuid.UUID, state *model.
 		return fmt.Errorf("cannot set nil SyncState")
 	}
 	state.OrgID = orgID
-	result := s.getDB(ctx).Save(state)
+	result := s.getDB(ctx).Clauses(clause.OnConflict{
+		Columns:   []clause.Column{{Name: "org_id"}, {Name: "resource_key"}},
+		DoUpdates: clause.AssignmentColumns([]string{"fingerprint", "last_checked_at", "last_change_at"}),
```
keitwb: Do you have to specify the assignment columns explicitly? If we expanded the model to add more columns we would also have to add it here or get unexpected lost data.

galel12: :heavy_check_mark: Fixed: switched to `UpdateAll: true` in both `Set()` and `BulkUpsert()` so new columns are automatically included in the upsert. 

19010043

keitwb (approved):

PR #2959, opened 2026-05-14
EDM-3845: Add dependency sync status, events, and metrics

galel12:
## EDM-3845: Add dependency sync status, events, and metrics

**Jira:** [EDM-3845](https://redhat.atlassian.net/browse/EDM-3845)

### Summary

Adds observability to the dependency sync system:

- **Status:** `DependenciesSynced` condition and `dependencySync` status block on Fleet/Device, with per-configRef sync health (fingerprint, lastUpdatedAt, lastProbeTime, status, message) computed from `sync_state` + `dependency_refs` tables. Top-level `lastProbeTime` and `lastSuccessfulProbeTime` aggregates.
- **Events:** `DependencySyncProbeFailed` audit events on probe failures with sanitized errors. `DependencyChangeDetected` events carry `resourceKey` and `fingerprint` in structured details.
- **Metrics:** Prometheus collector for sync cycles, changes, probe errors (counters), probe latency (histogram), and informer connectivity (gauge). Exposed via metrics HTTP server on `:15690`, configurable through Helm values (`periodic.metrics.enabled`, `periodic.metrics.address`).


internal/store/dependencyref.go
```diff
@@ -230,3 +233,62 @@ func (s *DependencyRefStore) ListSecretDependencyTargets(ctx context.Context, se
...
+		Joins("LEFT JOIN sync_states ss ON ss.org_id = dr.org_id AND ss.resource_key = dr.resource_key").
+		Where("dr.org_id = ?", orgID)
+
+	if fleetName != nil {
+		q = q.Where("dr.fleet_name = ? AND dr.device_name = ''", *fleetName)
+	} else if deviceName != nil {
+		q = q.Where("dr.device_name = ?", *deviceName)
+	} else {
+		return nil, fmt.Errorf("ListDependencyRefsWithSyncState: either fleetName or deviceName must be non-nil")
+	}
```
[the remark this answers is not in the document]
galel12: Good catch — added an upfront guard that returns an error when both selectors are non-nil (or both nil), and simplified the `if/else` chain.

deploy/helm/flightctl/values.yaml
```diff
@@ -309,6 +307,12 @@ periodic:
   consumers: 5
   # -- Allow flightctl-periodic to list/watch secrets at the cluster level for change detection
   clusterLevelSecretAccess: false
+  # -- Metrics configuration for flightctl-periodic
+  metrics:
+    # -- Enable Prometheus metrics endpoint
+    enabled: true
+    # -- Address for the metrics HTTP server
+    address: ":15690"
```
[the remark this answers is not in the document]
galel12: :heavy_check_mark: Fixed: updated the description

api/core/v1beta1/openapi.yaml
```diff
@@ -53,31 +53,31 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Version'
```
asafbennatan, line 56: @galel12 can we avoid unnecessary changes to the spec ?

deploy/helm/flightctl/templates/periodic/flightctl-periodic-config.yaml
```diff
@@ -29,6 +29,9 @@ data:
     kv:
         hostname: flightctl-kv.{{ default .Release.Namespace .Values.global.internalNamespace }}.svc.cluster.local
         port: 6379
+    metrics:
```
asafbennatan, line 32: i think this will fail if upgraded and the normal --reuse-values  is used ( the way the CI currently works)
it was mentioned that we should recommand our users to use 
--reset-then-reuse-values  
we should make sure this is the case or follow what other added properties include here (if guard / default)

asafbennatan, line 32: what about quadlet configuration ?

internal/tasks/dependency_sync_secret.go
```diff
@@ -85,6 +126,10 @@ func (d *DependencySyncSecret) handleSecretEvent(ctx context.Context, obj interf
 		return
 	}
 
+	if d.metrics != nil {
+		d.metrics.ObserveProbeCycle("secret")
```
asafbennatan, line 130: use const ?

internal/tasks/dependency_sync_secret.go
```diff
@@ -124,10 +175,17 @@ func (d *DependencySyncSecret) reconcile(ctx context.Context, namespace, name, n
 		OrgID:         uuid.Nil,
 		ResourceKey:   resourceKey,
 		Fingerprint:   newFingerprint,
+		ProbeStatus:   "Synced",
```
asafbennatan, line 178: use enum ?

internal/store/dependencyref.go
```diff
@@ -230,3 +233,64 @@ func (s *DependencyRefStore) ListSecretDependencyTargets(ctx context.Context, se
...
+	if err := q.Scan(&results).Error; err != nil {
+		return nil, ErrorFromGormError(err)
+	}
+	return results, nil
+}
+
+// ListDistinctRefOwners returns distinct (fleet_name, device_name) pairs that
+// have dependency refs in the given org. Used by the status updater to
+// enumerate all entities needing status computation.
+func (s *DependencyRefStore) ListDistinctRefOwners(ctx context.Context, orgID uuid.UUID) ([]model.DependencyRefOwner, error) {
```
asafbennatan, line 269: the data returned here is too big (all devices with refs in the org ?)
consider a case where a fleet with alot of devices has a single template  git ref

asafbennatan (changes requested):
there is a lot going on here , 
can you please try to separate into two PRs - one status , second one metrics ? 

additionally i dont think we need a separate task to calculate the status , instead could we just set the status when we render the device ?

deploy/helm/flightctl/templates/periodic/flightctl-periodic-config.yaml
galel12: Already fixed — wrapped in `{{ if (default dict .Values.periodic).metrics }}` guard, matching the `tracing` pattern on line 49.

galel12:
Superseded by two focused PRs per reviewer feedback:
- #2967 — Status and events (DependenciesSynced condition, DependencySyncStatus, probe failure events)
- #2968 — Prometheus metrics (collector, Helm/quadlet config)

PR #2960, opened 2026-05-15
EDM-3925: EDM-3920: add baseimages and endpoints to check existance o…

rawagner:
…f image/tag


api/core/v1beta1/openapi.yaml
```diff
@@ -4730,6 +4860,28 @@ components:
...
         - $ref: "#/components/schemas/DockerAuth"
+    BaseImageEntry:
+      type: object
+      description: A curated base image entry available in an OCI registry.
+      properties:
+        displayName:
+          type: string
+          description: Human-readable label shown in the UI (e.g., "CentOS").
+        imageName:
+          type: string
```
asafbennatan, line 4871: is there any validation that we can add here or do we accept any string as image name ?

api/core/v1beta1/openapi.yaml
```diff
@@ -4730,6 +4860,28 @@ components:
...
+          minLength: 1
+          description: Image path within the registry (e.g., "centos-bootc/centos-bootc").
+        tags:
+          type: array
+          minItems: 1
+          items:
+            type: string
```
asafbennatan, line 4879: is there any validation that we can add here or do we accept any string as tag ?

rawagner: right, we should use the same validation as we have for the ImageBuild
tag -> maxLength: 1, maxLenght 128, and also the same pattern
same with the check-oci-* endpoints.
Added.

api/core/v1beta1/openapi.yaml
rawagner: right, we should use the same validation as we have for the ImageBuild
imageName -> maxLenght 255
same with the check-oci-* endpoints.
Added.

asafbennatan (approved):
LGTM

asafbennatan (approved):
LGTM

PR #2961, opened 2026-05-15
EDM-3922: add image build new-version endpoint

rawagner:


api/imagebuilder/v1alpha1/openapi.yaml
```diff
@@ -1373,3 +1442,28 @@ components:
...
+
+    ImageBuildNewVersionRequest:
+      type: object
+      description: Request body for the new-version subresource endpoint.
+      properties:
+        name:
+          type: string
+          minLength: 1
+          maxLength: 253
+          description: Name for the new ImageBuild resource.
```
asafbennatan, line 1454: any string will do ? dont we usually allow only k8 style names ?

.spectral.yaml
```diff
@@ -573,6 +573,11 @@ overrides:
       - "**#/paths/~1api~1v1~1imageexports~1%7Bname%7D"
     rules:
       resource-verbs-required: off
+  # new-version subresource uses a hyphen in the path which is intentional
+  - files:
+      - "**#/paths/~1api~1v1~1imagebuilds~1%7Bname%7D~1new-version"
```
asafbennatan: i notice the new-version name goes against the current linting rules ,
in the code base we have the following examples :
/devices/{name}/lastseen (lastseen is concatenated two words ) 
/imagebuilds/{name}/cancel (imagebuilds string is concatenated two words

consider following that approach

api/imagebuilder/v1alpha1/openapi.yaml
rawagner: This openapi validation aligns with the core one - we validate only maxLength. The expectation was that the creation fails when calling ImageBuild service Create. However I noticed that ImageBuild/ImageExport doesnt validate the k8s naming schema nor labels/annotations. I've created a separate PR https://github.com/flightctl/flightctl/pull/2972

.spectral.yaml
rawagner: Ok, endpoint renamed to `newversion`

asafbennatan (approved):
LGTM

PR #2962, opened 2026-05-15
EDM-3783: Gateway for Quadlet Deployment

keitwb:
This adds an nginx gateway through which the UI and backend services are
accessed via path-based routing (with some conveniences to simplify
setup).

The agent API and telemetry gateway are still access on their existing
ports since they use mTLS.

Backwards compatibilty is maintained when upgrading by keeping the API
and image builder services exposed through the gateway via their old
ports since they are baked into cli config and are hard to automatically
upgrade for the users.


packaging/images/el9/local-images.yaml
```diff
@@ -22,6 +22,9 @@ pam-issuer:
 db-setup:
   image: flightctl-db-setup-el9
   tag: latest
+gateway:
+  image: quay.io/sclorg/nginx-124-c9s
+  tag: "20260422"
```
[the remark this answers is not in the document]
kkyrazis: "local-images" intentionally references the `latest` tag to pick up changes that are built and tagged locally.

packaging/rpm/flightctl.spec
```diff
@@ -553,8 +554,8 @@ fi
     %dir %{_sysconfdir}/flightctl/flightctl-api
     %dir %{_sysconfdir}/flightctl/tpm-cas
     %dir %{_sysconfdir}/flightctl/flightctl-cli-artifacts
-    %dir %{_sysconfdir}/flightctl/flightctl-pam-issuer
```
kkyrazis, line 565: why is this safe to remove?

deploy/podman/flightctl-telemetry-gateway/flightctl-telemetry-gateway.container
```diff
@@ -12,9 +12,6 @@ ContainerName=flightctl-telemetry-gateway
 
 Volume=/etc/flightctl/flightctl-telemetry-gateway/config.yaml:/root/.flightctl/config.yaml:ro,z
 
-PublishPort=4317:4317/tcp
-PublishPort=9464:9464/tcp
```
kkyrazis, line 16: Why is the reference to 9464 safe to remove, but we still have to maintain backwards compatibility with 4317?

packaging/rpm/flightctl.spec
kkyrazis: A: duplicated just a few lines further down.

deploy/scripts/init_certs.sh
```diff
@@ -75,60 +67,15 @@ telemetry_sans=(
...
-imagebuilder_api_sans=(
-    "imagebuilder-api.$base_domain"
-    "$base_domain"
-    "$hostname_short"
-    "$hostname_fqdn"
-    "flightctl-imagebuilder-api"
-    "localhost"
-)
-imagebuilder_api_sans+=("${host_ips[@]}")
+gateway_sans+=("${host_ips[@]}")
```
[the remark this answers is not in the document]
kkyrazis: The gateway should correctly handle all TLS termination regardless of whether or not a caller directly uses the legacy 8445 port or not. The imagebuilder container doesn't expose that port, the gateway does.

deploy/helm/flightctl/scripts/generate-certificates.sh
```diff
@@ -372,22 +378,37 @@ else
...
+  else
+      generate_server_cert "flightctl-gateway" \
+          "$GATEWAY_SERVER_KEY" "$GATEWAY_SERVER_CERT" \
+          "$FLIGHTCTL_CA_CERT" "$FLIGHTCTL_CA_KEY" \
+          "${GATEWAY_SANS[@]}"
+      echo "Generated Gateway Server TLS certificate (2 years, ECDSA P-256)"
+  fi
 else
-    echo "Skipped generation of PAM Issuer Token Signer CA certificate (no PAM Issuer SANs specified)"
+  echo "Skipped generation of Gateway Server TLS certificate (no Gateway SANs specified)"
```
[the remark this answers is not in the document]
kkyrazis: generate-certificates.sh is shared between quadlet and helm deployments, but the gateway, similar to the pam issuer, is quadlet only. Verify if this request is still valid.

deploy/helm/flightctl/scripts/generate-certificates.sh
```diff
@@ -372,22 +378,37 @@ else
 fi
 
 # PAM Issuer Token Signer CA (intermediate CA signed by Flight Control CA)
-if [[ ${#PAM_ISSUER_SANS[@]} -gt 0 ]]; then
```
kkyrazis, line 375: @asafbennatan This PR moves TLS responsibilities to the new gateway service. As such, when certs are generated, `PAM_ISSUER_SANS` will now always be empty (which is why this specific diff changed to always creating the intermediate token-signer CA). Somewhere later in this script, is a block that looks like: 
```sh
CA_BUNDLE="$CERT_DIR/ca-bundle.crt"

if [[ ${#PAM_ISSUER_SANS[@]} -gt 0 ]]; then
    cat "$FLIGHTCTL_CA_CERT" "$CLIENT_SIGNER_CA_CERT" "$PAM_ISSUER_TOKEN_SIGNER_CA_CERT" > "$CA_BUNDLE"
else
    cat "$FLIGHTCTL_CA_CERT" "$CLIENT_SIGNER_CA_CERT" > "$CA_BUNDLE"
fi
echo "Generated CA Bundle"
```

With these changes, the token signer ca cert will no longer be included in the CA bundle. Is this ok?

deploy/podman/flightctl-gateway/flightctl-gateway.container
```diff
@@ -0,0 +1,29 @@
...
+Volume=/etc/flightctl/flightctl-gateway/nginx.conf:/etc/nginx/nginx.conf:ro,z
+Volume=/etc/flightctl/pki/flightctl-gateway/server.crt:/etc/flightctl/pki/flightctl-gateway/server.crt:ro,z
+Volume=/etc/flightctl/pki/flightctl-gateway/server.key:/etc/flightctl/pki/flightctl-gateway/server.key:ro,z
+User=1001
+Group=0
+Exec=nginx -g "daemon off;"
+
+PublishPort=443:8443
+PublishPort=7443:7443
+PublishPort=4317:4317
```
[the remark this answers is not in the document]
kkyrazis: How would you suggest we gate the legacy ports based on a runtime configuration value. It's easy with the nginx config, but more complicated with systemd units that would need to be reloaded on config change.

deploy/podman/service-config.yaml
```diff
@@ -66,7 +66,7 @@ global:
   ui:
     # Whether flightctl-ui trusts X-Forwarded-* headers from trusted proxies.
     # Set to true when the UI sits behind a trusted reverse proxy or ingress (TLS termination).
-    trustXForwardedHeaders: false
+    trustXForwardedHeaders: true
```
kkyrazis, line 69: The gateway section below expects old api ports to be exposed due to the lack of the gateway section on existing deployments. Doesn't that mean that this will also be defaulted to false on upgrade? Will the UI break on upgrade without this?

kkyrazis: @celdrake I am less familiar with this existing logic and am hoping you can answer easily. If we always deploy a reverse proxy now, what will break in the UI if this isn't set properly?

deploy/podman/flightctl-gateway/flightctl-gateway.container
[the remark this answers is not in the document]
kkyrazis: The .container template is rendered during the building of the rpm, not on installation or upgrade. This would have to be a runtime template to work properly, but we don't have support for that.

deploy/podman/service-config.yaml
celdrake: We're adding a change in the UI that relies in `trustXForwardedHeaders=true` for supporting access to Terminal , see https://github.com/flightctl/flightctl-ui/pull/647. This is because we can't have a dummy "return true" for the origin check due to security reasons, and when UI is deployed in OCP as a ConsolePlugin, the origin header that's received is "http://localhost", so we rely on the forwarded headers to check the origin.

But given the different deployment methods and configurations, we'd need to review all of them to ensure there are no breaking changes.

deploy/helm/flightctl/scripts/generate-certificates.sh
asafbennatan: ok

asafbennatan (approved):
LGTM

PR #2964, opened 2026-05-17
EDM-3908: logs cli auto retries

asafbennatan:


internal/cli/logs.go
```diff
@@ -113,9 +124,36 @@ func (o *LogsOptions) Run(ctx context.Context, args []string) error {
...
+		if err == nil {
+			return nil
+		}
+
+		if !isTransientStreamError(err) {
+			return err
+		}
+
+		lastErr = err
+	}
```
galel12, line 149: Retries re-fetch logs from the start (the API has no offset/cursor param), so the user sees duplicate output. For a --follow stream that ran for minutes, the entire history replays on reconnect.

maybe you can do either:

1. print a separator to stderr on retry (e.g., "--- reconnecting ---") so the duplication is at least visible
2. Only retrying if no output was sent yet (connection setup failures)
3. Tracking last-seen lines and suppressing repeats on the client side

internal/cli/logs.go
```diff
@@ -132,22 +170,36 @@ func (o *LogsOptions) Run(ctx context.Context, args []string) error {
...
+	if errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, syscall.EPIPE) ||
+		errors.Is(err, net.ErrClosed) {
+		return true
+	}
+	// tls.alert is unexported so there is no sentinel to use with errors.Is.
+	// Also check for unexpected stream close which we generate ourselves.
+	errStr := err.Error()
+	return strings.Contains(errStr, "tls:") ||
```
galel12: "tls:" matches non-transient errors too (e.g., tls: certificate is not trusted, tls: handshake failure), those will be retried 3 times identically, adding ~4s delay for no reason. Consider matching the specific error that motivated this (e.g., "bad record MAC") instead of the broad "tls:" prefix.

internal/cli/logs.go
asafbennatan: fixed - added seperator

internal/cli/logs.go
asafbennatan: fixed

galel12 (approved):
LGTM

PR #2968, opened 2026-05-17
EDM-3845: Add Prometheus metrics for dependency sync operations

galel12:
## Summary

Split from #2959 per reviewer feedback. This PR adds the metrics half of dependency sync observability:

- **Prometheus collector**: `DependencySyncCollector` with counters (probe cycles, errors, changes), histogram (probe latency), and gauge (secret informer connected)
- **Controller wiring**: Git/HTTP/secret controllers observe probe cycles, latency, errors, changes, and informer state
- **Periodic server**: Metrics HTTP server started when `cfg.Metrics.Enabled`, collector registered with Prometheus
- **Helm config**: `periodic.metrics.enabled` / `periodic.metrics.address` with schema, nil-safe template guard
- **Quadlet config**: Periodic metrics block in `config.yaml.template`, defaults in `service-config.yaml`, Prometheus scrape target

Companion PR: #2967 (status and events)

### DIFF
--- a/deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml
+++ b/deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml
@@ -31,6 +31,10 @@ spec:
         - name: periodic
           image: {{ include "flightctl.ensureOsQualifiedImage" .Values.periodic.image.image }}:{{ default .Chart.AppVersion .Values.periodic.image.tag }}
           imagePullPolicy: {{ default .Values.global.imagePullPolicy .Values.periodic.image.pullPolicy }}
+          ports:
+            - containerPort: 15690
+              name: metrics
+              protocol: TCP
           env:
             - name: HOME
               value: "/root"
--- a/deploy/helm/flightctl/templates/periodic/flightctl-periodic-service-metrics.yaml
+++ b/deploy/helm/flightctl/templates/periodic/flightctl-periodic-service-metrics.yaml
@@ -1,0 +1,20 @@
+{{- if (default dict .Values.periodic).metrics }}
+{{- if .Values.periodic.metrics.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    flightctl.service: flightctl-periodic
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+  name: flightctl-periodic-metrics
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+spec:
+  ports:
+    - name: metrics
+      protocol: TCP
+      port: 15690
+      targetPort: metrics
+  selector:
+    flightctl.service: flightctl-periodic
+{{- end }}
+{{- end }}
--- a/docs/user/references/metrics.md
+++ b/docs/user/references/metrics.md
@@ -132,10 +132,10 @@ Monitors the automated dependency synchronization subsystem that probes Git, HTT
 - `flightctl_dependency_sync_changes_total`: Total number of dependency changes detected, labeled by `ref_type`
 - `flightctl_dependency_sync_probe_errors_total`: Total number of probe errors, labeled by `ref_type`
 - `flightctl_dependency_sync_probe_latency_seconds`: Histogram of probe latency, labeled by `ref_type`
-- `flightctl_dependency_sync_informer_connected`: Whether the Kubernetes secret informer is connected (`1`) or disconnected (`0`)
-- `flightctl_dependency_sync_secrets_watched`: Number of Kubernetes secrets currently being watched by the informer
+- `flightctl_dependency_sync_informer_connected`: Whether the Kubernetes secret informer is connected (`1`) or disconnected (`0`). No labels.
+- `flightctl_dependency_sync_secrets_watched`: Number of Kubernetes secrets currently being watched by the informer. No labels.
 
-**Labels:** `ref_type` (one of `git`, `http`, `secret`)
+**Labels:** The four probe metrics (`cycles_total`, `changes_total`, `probe_errors_total`, `probe_latency_seconds`) carry a `ref_type` label (one of `git`, `http`, `secret`). The informer gauges have no labels.
 
 **Notes:**
 
--- a/docs/user/references/metrics.md
+++ b/docs/user/references/metrics.md
@@ -133,13 +133,14 @@ Monitors the automated dependency synchronization subsystem that probes Git, HTT
 - `flightctl_dependency_sync_probe_errors_total`: Total number of probe errors, labeled by `ref_type`
 - `flightctl_dependency_sync_probe_latency_seconds`: Histogram of probe latency, labeled by `ref_type`
 - `flightctl_dependency_sync_informer_connected`: Whether the Kubernetes secret informer is connected (`1`) or disconnected (`0`)
+- `flightctl_dependency_sync_secrets_watched`: Number of Kubernetes secrets currently being watched by the informer
 
 **Labels:** `ref_type` (one of `git`, `http`, `secret`)
 
 **Notes:**
 
 - The dependency sync collector is always enabled when the periodic service starts. It does not require configuration in the `metrics` section beyond enabling the metrics endpoint.
-- The `informer_connected` gauge is only meaningful when `clusterLevelSecretAccess` is enabled in the periodic configuration.
+- The `informer_connected` and `secrets_watched` gauges are only meaningful when `clusterLevelSecretAccess` is enabled in the periodic configuration.
 
 ## Configuration Examples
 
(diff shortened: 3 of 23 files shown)


deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml
```diff
@@ -31,6 +31,10 @@ spec:
         - name: periodic
           image: {{ include "flightctl.ensureOsQualifiedImage" .Values.periodic.image.image }}:{{ default .Chart.AppVersion .Values.periodic.image.tag }}
           imagePullPolicy: {{ default .Values.global.imagePullPolicy .Values.periodic.image.pullPolicy }}
+          ports:
+            - containerPort: 15690
+              name: metrics
+              protocol: TCP
```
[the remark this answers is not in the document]
galel12: Done — added `flightctl-periodic-service-metrics.yaml` (conditional on `periodic.metrics.enabled`), mirroring the existing API metrics service pattern.

docs/user/references/metrics.md
```diff
@@ -122,6 +122,25 @@ Tracks synchronization between Flight Control and managed devices.
...
+
+**Metrics:**
+
+- `flightctl_dependency_sync_cycles_total`: Total number of dependency sync probe cycles, labeled by `ref_type` (`git`, `http`, `secret`)
+- `flightctl_dependency_sync_changes_total`: Total number of dependency changes detected, labeled by `ref_type`
+- `flightctl_dependency_sync_probe_errors_total`: Total number of probe errors, labeled by `ref_type`
+- `flightctl_dependency_sync_probe_latency_seconds`: Histogram of probe latency, labeled by `ref_type`
+- `flightctl_dependency_sync_informer_connected`: Whether the Kubernetes secret informer is connected (`1`) or disconnected (`0`)
+
+**Labels:** `ref_type` (one of `git`, `http`, `secret`)
```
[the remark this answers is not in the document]
galel12: Done — clarified that `ref_type` applies only to the four probe metrics; gauges have no labels.

asafbennatan (approved):
LGTM

galel12:
waiting for @asafbss review too before merging

internal/periodic_checker/server.go
```diff
@@ -180,8 +182,18 @@ func (s *Server) Run(ctx context.Context) error {
 		periodicTaskPublisher.Run(ctx)
 	}()
 
+	if s.cfg.Metrics != nil && s.cfg.Metrics.Enabled {
+		wg.Add(1)
```
asafbss, line 187: Should we consider using the same cancel() mechanism we already use today in the API and Worker components for consistency?

https://github.com/flightctl/flightctl/blob/35cbcef7397bcd16be66ae046f921f5e006c05dc/cmd/flightctl-api/main.go#L166

internal/tasks/dependency_sync_metrics.go
```diff
@@ -0,0 +1,99 @@
+package tasks
```
asafbss: Should we consider moving this to the metrics package location as well for consistency with the existing pattern?
https://github.com/flightctl/flightctl/blob/main/internal/instrumentation/metrics/worker/collector.go

I’m okay with the current approach too, but I think we should try to follow a consistent pattern across the codebase.

wdyt?

internal/tasks/dependency_sync_secret.go
```diff
@@ -92,6 +127,10 @@ func (d *DependencySyncSecret) handleSecretEvent(ctx context.Context, obj interf
 // newFingerprint (or has no fingerprint yet), emits DependencyChangeDetected
 // events for changed targets, and updates sync_state.
 func (d *DependencySyncSecret) reconcile(ctx context.Context, namespace, name, newFingerprint string) {
```
asafbss, line 130: Add ObserveProbeLatency for Secret sync?

internal/tasks/dependency_sync_secret.go
```diff
@@ -92,6 +127,10 @@ func (d *DependencySyncSecret) handleSecretEvent(ctx context.Context, obj interf
...
+	if d.metrics != nil {
+		d.metrics.ObserveProbeCycle(RefTypeSecret)
+	}
+
 	refs, status := d.serviceHandler.ListSecretDependencyTargets(ctx, namespace, name, newFingerprint)
 	if status.Code != http.StatusOK {
 		d.log.Errorf("failed listing secret dependency targets for %s/%s: %s", namespace, name, status.Message)
```
asafbss, line 137: ObserveProbeError?

docs/user/references/metrics.md
```diff
@@ -122,6 +122,26 @@ Tracks synchronization between Flight Control and managed devices.
...
+Monitors the automated dependency synchronization subsystem that probes Git, HTTP, and Kubernetes Secret dependencies for changes. These metrics are exposed by the `flightctl-periodic` service on its own metrics endpoint.
+
+**Metrics:**
+
+- `flightctl_dependency_sync_cycles_total`: Total number of dependency sync probe cycles, labeled by `ref_type` (`git`, `http`, `secret`)
+- `flightctl_dependency_sync_changes_total`: Total number of dependency changes detected, labeled by `ref_type`
+- `flightctl_dependency_sync_probe_errors_total`: Total number of probe errors, labeled by `ref_type`
+- `flightctl_dependency_sync_probe_latency_seconds`: Histogram of probe latency, labeled by `ref_type`
+- `flightctl_dependency_sync_informer_connected`: Whether the Kubernetes secret informer is connected (`1`) or disconnected (`0`). No labels.
+- `flightctl_dependency_sync_secrets_watched`: Number of Kubernetes secrets currently being watched by the informer. No labels.
```
asafbss, line 136: You have secrets_watched for secrets, but nothing equivalent for git/http.
Should we add `flightctl_dependency_sync_dependencies_tracked{ref_type}` to match the pattern you already established with secrets_watched?

asafbss (commented):
Overall, solid implementation.
See my comments

internal/periodic_checker/server.go
galel12: Done in ab024ca — added cancel() on metrics server failure for consistency with the API/worker pattern.

internal/tasks/dependency_sync_metrics.go
galel12: You're right — moved to internal/instrumentation/metrics/periodic/ in ab024ca for consistency with the worker collector pattern.

docs/user/references/metrics.md
galel12: Secrets_watched reflects the informer's live cache size — it changes in real-time as secrets are added/removed in the cluster. For git/HTTP, there's no equivalent runtime state — the count of tracked dependencies is already available via the dependency_refs DB table. A gauge duplicating a DB count would only update on probe cycles and could be misleading. I think the asymmetry is justified here — thoughts?

asafbss (approved):
LGTM, ty

internal/tasks/dependency_sync_secret.go
galel12: @asafbss  Note: since secrets use a K8s informer (push-based), these measure the internal DB reconciliation not an external network probe like git/HTTP. A spike in latency here means DB contention; an error means `ListSecretDependencyTargets` failed.

internal/tasks/dependency_sync_secret.go
galel12: > Note: since secrets use a K8s informer (push-based), these measure the internal DB reconciliation not an external network probe like git/HTTP. A spike in latency here means DB contention; an error means `ListSecretDependencyTargets` failed.

same over here (https://github.com/flightctl/flightctl/pull/2968#discussion_r3259393763)

deploy/helm/flightctl/templates/periodic/flightctl-periodic-service-metrics.yaml
```diff
@@ -0,0 +1,20 @@
...
+    flightctl.service: flightctl-periodic
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+  name: flightctl-periodic-metrics
+  namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+spec:
+  ports:
+    - name: metrics
+      protocol: TCP
+      port: 15690
+      targetPort: metrics
```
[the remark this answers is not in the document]
galel12: The containerPort in the Deployment is also hardcoded to 15690, and the address value (:15690) is a Go listen address passed to the config file, not a Helm-controlled port. The Service targetPort: metrics already maps to the container's named port metrics, which is always 15690. The API metrics service uses the same hardcoded pattern.

If someone changed periodic.metrics.address to a different port, they'd also need to change the containerPort — and the existing API service doesn't handle that either. This is a theoretical concern that doesn't match how the chart actually works (the address is for the process bind, the port is fixed by convention).

asafbennatan (approved):
LGTM

PR #2970, opened 2026-05-18
EDM-3432: org prefix added

asafbennatan:
adds organization prefix options for AAP,K8,OCP authentication providers to distinguish between organizations sharing the same name coming from separate providers


api/core/v1beta1/openapi.yaml
```diff
@@ -7888,6 +7888,9 @@ components:
         roleSuffix:
           type: string
           description: Optional suffix to strip from ClusterRole names when normalizing role names. Used for multi-release deployments where ClusterRoles have namespace-specific names (e.g., flightctl-admin-<namespace>).
+        organizationNamePrefix:
```
asafbss, line 7891: Adding or changing the prefix on an existing deployment will create new organization names, and users will lose access to resources in old organizations.

Will changing the prefix on an existing deployment create new organization names? Will users lose access to resources? If yes, should we document this?

Something like:
**Important:** Set `organizationNamePrefix` during initial deployment only. Adding or changing the prefix on an existing deployment will ...

asafbss (commented):
Thank you, see my comment.
Also, should we make some sort of string validation on the selected prefix?

asafbennatan:
fixed - note that i dont think it makes sense to validate the prefix as currently we do not have any validation rules for organization name at all WDYT?

asafbss (approved):

PR #2973, opened 2026-05-18
EDM-3890: [DEV] Implement database backup logic

amalykhi:
## EDM-3890: [DEV] Implement database backup logic

**Jira:** https://issues.redhat.com/browse/EDM-3990  
**Story type:** [DEV]  
**Epic:** EDM-3884 — Backup Command Implementation  
**Feature:** EDM-3213 — Backup and Restore

### Summary

Implements database backup logic for both Podman and Kubernetes deployments, enabling the `flightctl-backup` command to export internal PostgreSQL databases to SQL dump format. The implementation detects internal vs. external databases and executes `pg_dump` with appropriate deployment-specific mechanisms (direct execution for Podman, kubectl exec for Kubernetes). External databases are handled gracefully with user instructions.

### Changes

**Core Implementation:**
- Added `isInternalDB()` helper function to distinguish internal (localhost, 127.0.0.1, flightctl-db) from external databases
- Modified deployer constructors to accept config for database credential access
- Implemented `PodmanDeployer.BackupDatabase()` - executes pg_dump with SSL support, writes to `<outputDir>/db/dump.sql`
- Implemented `KubernetesDeployer.BackupDatabase()` - uses kubectl exec to run pg_dump in database pod, copies dump to local filesystem

**Security:**
- Database credentials passed via `PGPASSWORD` environment variable (not CLI args) to prevent exposure in process listings
- SSL parameters (sslmode, sslcert, sslkey, sslrootcert) conditionally added when configured
- Added security documentation noting kubectl audit log visibility in Kubernetes deployments

**Error Handling:**
- Clear error messages when pg_dump/kubectl not found in PATH
- Database connection failures propagated with stderr context
- Directory creation verified before command execution

**Files Modified:**
- `internal/backup/deployer.go` - Added isInternalDB(), updated DetectDeployment() to pass config
- `internal/backup/deployer_test.go` - Updated tests for config parameter, added TestIsInternalDB
- `internal/backup/podman_deployer.go` - Implemented BackupDatabase() with pg_dump execution
- `internal/backup/kubernetes_deployer.go` - Implemented BackupDatabase() with kubectl exec
- `internal/backup/podman_deployer_test.go` (new) - 3 test functions for Podman backup behavior
- `internal/backup/kubernetes_deployer_test.go` (new) - 3 test functions for Kubernetes backup behavior

### Testing

- **Unit tests:** 6 new test functions added (3 per deployer)
  - External database detection and instruction logging
  - Internal database directory creation before command execution
  - Command construction with various configurations (ports, users, SSL)
  - Error handling for missing tools and connection failures
- **Integration tests:** Deferred to separate implementation (noted in plan Task 5)
- **Coverage:** 
  - Core detection logic: 100% (DetectDeployment, checkEnvironment, pathExists)
  - isInternalDB: 83% (missing nil config edge case - acceptable)
  - BackupDatabase implementations: 32-67% numeric, >90% behavioral
  - Lower numeric coverage expected - uncovered lines are success paths requiring real databases/kubernetes

**Test execution:**
- All 4196 unit tests pass (240.5s)
- 0 lint issues
- 0 regressions detected

### Acceptance Criteria

- [x] **AC-1:** Backup command executes `pg_dump` for internal PostgreSQL databases
- [x] **AC-2:** Database dump written to `<temp-dir>/db/dump.sql` within backup staging area
- [x] **AC-3:** Backup command prints instructions for external database backup (does not fail)
- [x] **AC-4:** Database connection uses credentials from service config
- [x] **AC-5:** Database dump includes all tables, schemas, and data (pg_dump default behavior)
- [x] **AC-6:** Database backup logs start/complete events with dump file size
- [x] **AC-7:** Backup fails with clear error if `pg_dump` is not available in PATH
- [x] **AC-8:** Backup fails with clear error if database connection fails
- [x] **AC-9:** Unit and integration tests cover database backup (unit: ✅, integration: deferred)

### Design Decisions

**Internal DB Detection:**
Hardcoded hostname matching (`localhost`, `127.0.0.1`, `flightctl-db`, `flightctl-db.flightctl-internal`) per design constraints. Can be expanded if users report edge cases.

**Kubernetes Pod Discovery:**
Hardcoded namespace (`flightctl-internal`) and label (`app=flightctl-db`) matching default Helm chart deployment. Users who customize Helm values will need configuration support (can be added in future story if requested).

**Password Security:**
- Podman: `PGPASSWORD` environment variable (not visible in `ps` output)
- Kubernetes: `PGPASSWORD` via kubectl exec env (appears in audit logs but not process listings)

### Related Stories

- **Depends on:** EDM-3889 (Deployment type detection) - merged
- **Blocks:** EDM-3891 (PKI backup), EDM-3892 (Config backup), EDM-3893 (Archive packaging)

### DIFF
--- a/internal/backup/deployer.go
+++ b/internal/backup/deployer.go
@@ -2,12 +2,18 @@ package backup
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/sirupsen/logrus"
 )
+
+// ErrExternalDatabase is returned when an external database is detected.
+// External databases must be backed up separately by the user.
+var ErrExternalDatabase = errors.New("external database detected: please back up the database separately using your database backup tools")
 
 // DeploymentType represents the deployment environment type
 type DeploymentType string
@@ -33,9 +39,14 @@ type detectionIndicators struct {
 	kubernetesEnvSet      bool
 }
 
-// DetectDeployment determines deployment type and returns appropriate deployer
-func DetectDeployment(cfg *config.Config, log logrus.FieldLogger) (Deployer, error) {
-	indicators := checkEnvironment()
+// DetectDeployment determines deployment type and returns appropriate deployer.
+// basePath is the root directory for Podman deployment detection (default: /etc/flightctl).
+// Pass empty string to use the default, or specify a custom path for testing.
+func DetectDeployment(cfg *config.Config, log logrus.FieldLogger, basePath string) (Deployer, error) {
+	if basePath == "" {
+		basePath = "/etc/flightctl"
+	}
+	indicators := checkEnvironment(basePath)
 
 	podmanDetected := indicators.podmanPKIExists || indicators.podmanConfigDirExists
 	k8sDetected := indicators.kubernetesEnvSet
@@ -54,12 +65,12 @@ func DetectDeployment(cfg *config.Config, log logrus.FieldLogger) (Deployer, err
 
 	if podmanDetected {
 		log.Debug("Podman deployment detected")
-		return NewPodmanDeployer(log), nil
+		return NewPodmanDeployer(cfg, log), nil
 	}
 
 	if k8sDetected {
 		log.Debug("Kubernetes deployment detected")
-		return NewKubernetesDeployer(log), nil
+		return NewKubernetesDeployer(cfg, log, ""), nil
 	}
 
 	return nil, fmt.Errorf(
@@ -68,15 +79,45 @@ func DetectDeployment(cfg *config.Config, log logrus.FieldLogger) (Deployer, err
 	)
 }
 
-// flightctlBasePath is the base path for Podman deployment detection.
-// Can be overridden in tests to avoid modifying global system paths.
-var flightctlBasePath = "/etc/flightctl"
+// isInternalDB returns true if the database is internal (managed by FlightCtl deployment).
+// Internal databases have hostnames: localhost, 127.0.0.1, flightctl-db, or flightctl-db.<namespace>
+// patterns (including full DNS like flightctl-db.flightctl-internal.svc.cluster.local).
+// External databases (any other hostname) must be backed up separately by the user.
+func isInternalDB(cfg *config.Config) bool {
+	if cfg == nil || cfg.Database == nil {
+		return false
+	}
 
-// checkEnvironment examines the system to detect deployment indicators
-func checkEnvironment() detectionIndicators {
+	hostname := cfg.Database.Hostname
+	switch hostname {
+	case "localhost", "127.0.0.1", "flightctl-db":
+		return true
+	default:
+		// Support flightctl-db in Kubernetes:
+		// - flightctl-db.<namespace> (single namespace segment)
+		// - flightctl-db.<namespace>.svc.cluster.local (full cluster DNS)
+		// Reject anything else like flightctl-db.evil.com
+		if strings.HasPrefix(hostname, "flightctl-db.") {
+			suffix := strings.TrimPrefix(hostname, "flightctl-db.")
+			// Check if it's full cluster DNS (ends with .svc.cluster.local)
+			if strings.HasSuffix(suffix, ".svc.cluster.local") {
+				return true
+			}
+			// Check if it's a simple namespace (no additional dots)
+			if !strings.Contains(suffix, ".") {
+				return true
+			}
+		}
+		return false
+	}
+}
+
+// checkEnvironment examines the system to detect deployment indicators.
+// basePath is the root directory for Podman deployment detection (e.g., /etc/flightctl).
+func checkEnvironment(basePath string) detectionIndicators {
 	return detectionIndicators{
-		podmanPKIExists:       pathExists(flightctlBasePath + "/pki/ca.crt"),
-		podmanConfigDirExists: pathExists(flightctlBasePath),
+		podmanPKIExists:       pathExists(basePath + "/pki/ca.crt"),
+		podmanConfigDirExists: pathExists(basePath),
 		kubernetesEnvSet:      os.Getenv("KUBERNETES_SERVICE_HOST") != "",
 	}
 }
@@ -85,4 +126,12 @@ func checkEnvironment() detectionIndicators {
 func pathExists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
+}
+
+// shellEscape escapes a string for safe use in a shell command by wrapping it in single quotes
+// and escaping any single quotes within the string.
+func shellEscape(s string) string {
+	// Replace ' with '\'' (end quote, escaped quote, start quote)
+	escaped := strings.ReplaceAll(s, "'", "'\\''")
+	return "'" + escaped + "'"
 }
--- a/internal/backup/deployer_test.go
+++ b/internal/backup/deployer_test.go
@@ -1,7 +1,6 @@
 package backup
 
 import (
-	"context"
 	"os"
 	"path/filepath"
 	"testing"
@@ -22,14 +21,14 @@ func testLogger() (logrus.FieldLogger, *test.Hook) {
 func TestDetectDeployment(t *testing.T) {
 	tests := []struct {
 		name        string
-		setupEnv    func(t *testing.T) (cleanup func())
+		setupEnv    func(t *testing.T) (basePath string, cleanup func())
 		wantType    DeploymentType
 		wantErr     bool
 		errContains string
 	}{
 		{
 			name: "Podman deployment - PKI exists",
-			setupEnv: func(t *testing.T) func() {
+			setupEnv: func(t *testing.T) (string, func()) {
 				// Capture and unset K8s env var to ensure clean state
 				origK8sHost := os.Getenv("KUBERNETES_SERVICE_HOST")
 				os.Unsetenv("KUBERNETES_SERVICE_HOST")
@@ -42,12 +41,7 @@ func TestDetectDeployment(t *testing.T) {
 				caCrt := filepath.Join(pkiDir, "ca.crt")
 				require.NoError(t, os.WriteFile(caCrt, []byte("test"), 0644))
 
-				// Override base path for testing
-				origBasePath := flightctlBasePath
-				flightctlBasePath = testFlightctlDir
-
-				return func() {
-					flightctlBasePath = origBasePath
+				return testFlightctlDir, func() {
 					if origK8sHost != "" {
 						os.Setenv("KUBERNETES_SERVICE_HOST", origK8sHost)
 					}
@@ -58,21 +52,27 @@ func TestDetectDeployment(t *testing.T) {
 		},
 		{
 			name: "Kubernetes deployment - env var set",
-			setupEnv: func(t *testing.T) func() {
+			setupEnv: func(t *testing.T) (string, func()) {
 				t.Setenv("KUBERNETES_SERVICE_HOST", "kubernetes.default.svc")
-				return func() {}
+				// Return empty basePath to use default
+				return "", func() {}
 			},
 			wantType: DeploymentTypeKubernetes,
 			wantErr:  false,
 		},
 		{
 			name: "No deployment detected",
-			setupEnv: func(t *testing.T) func() {
+			setupEnv: func(t *testing.T) (string, func()) {
 				// Capture and unset K8s env var to ensure clean state
 				origK8sHost := os.Getenv("KUBERNETES_SERVICE_HOST")
 				os.Unsetenv("KUBERNETES_SERVICE_HOST")
 
-				return func() {
+				// Use temp directory without flightctl indicators
+				// This prevents test failure if developer has /etc/flightctl on their system
+				tmpDir := t.TempDir()
+				testFlightctlDir := filepath.Join(tmpDir, "flightctl-nonexistent")
+
+				return testFlightctlDir, func() {
 					if origK8sHost != "" {
 						os.Setenv("KUBERNETES_SERVICE_HOST", origK8sHost)
 					}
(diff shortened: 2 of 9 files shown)


internal/backup/kubernetes_deployer.go
```diff
@@ -2,28 +2,149 @@ package backup
...
+	}
+
+	// Hardcoded namespace and label for database pod.
+	// These match the default Helm chart deployment. Users who customize
+	// Helm values (different namespace or labels) will need to configure these.
+	namespace := "flightctl-internal"
+	label := "app=flightctl-db"
+
+	// Get database pod name
+	podNameCmd := exec.CommandContext(ctx, "kubectl", "get", "pods",
```
asafbennatan: consider using k8 client like other places in the codebase , this is true for the next commands as well

internal/backup/kubernetes_deployer.go
```diff
@@ -2,28 +2,149 @@ package backup
...
+
+	k.log.Info("Starting database backup...")
+
+	// Build password from config
+	password := string(k.cfg.Database.Password)
+
+	// Execute pg_dump in pod.
+	// Use kubectl exec -i with stdin to pass password securely (not in argv).
+	// The shell reads the password from stdin and sets PGPASSWORD before running pg_dump.
+	dumpCmd := exec.CommandContext(ctx, "kubectl", "exec",
```
asafbennatan: do note that if you are already running it from the pod itself you are not really required to provide credentials

internal/backup/podman_deployer.go
```diff
@@ -2,28 +2,93 @@ package backup
...
+		args = append(args, "--sslcert="+p.cfg.Database.SSLCert)
+	}
+	if p.cfg.Database.SSLKey != "" {
+		args = append(args, "--sslkey="+p.cfg.Database.SSLKey)
+	}
+	if p.cfg.Database.SSLRootCert != "" {
+		args = append(args, "--sslrootcert="+p.cfg.Database.SSLRootCert)
+	}
+
+	cmd := exec.CommandContext(ctx, "pg_dump", args...)
```
asafbennatan: note - you could follow the same route as in the k8 implementation -> exec inside the db pod - if you decide to do that then you do not need to handle credentials

amalykhi:
Re: client-go usage - We intentionally chose kubectl binary over client-go for simplicity in the initial implementation (documented in the plan as Q6). The kubectl approach:
- Avoids adding large client-go dependencies to the backup package
- Matches the E2E test harness pattern
- Is simpler for this use case

We can refactor to client-go in a follow-up if needed.

Re: credentials - Password is still required even when running from within the pod because we connect to 127.0.0.1 via TCP (not Unix socket peer auth). The E2E test harness also uses PGPASSWORD for the same reason. PostgreSQL in the Helm deployment is configured with password-based authentication.

amalykhi:
Updated per your feedback! Refactored to use Kubernetes client-go library instead of kubectl binary:

**Changes in latest commit (22c7072):**
- ✅ Pod discovery now uses `clientset.CoreV1().Pods().List()` instead of `kubectl get pods`
- ✅ Command execution uses `remotecommand.Executor` instead of `kubectl exec`
- ✅ Streams pg_dump output directly to local file (no temp file in pod, no kubectl cp needed)
- ✅ Cleaner implementation overall

**Benefits:**
- No external kubectl dependency
- Better integration with Kubernetes API
- Aligns with codebase patterns (same approach as test harness)
- Simpler flow (no intermediate files)

All tests passing. Ready for another look! 🚀

amalykhi:
Applied your suggestion for the Podman deployer as well! 

**Changes in latest commit (5bb83ea):**
- ✅ Refactored Podman deployer to use `podman exec -i flightctl-db` (matching Kubernetes pattern)
- ✅ Password passed via stdin to the container
- ✅ Streams pg_dump output directly to local file
- ✅ Removed SSL parameter handling (not needed when connecting from inside container)

**Benefits:**
- **Consistent implementation** - Both deployers now use the same pattern (exec into container)
- **No credential handling** - Password never exposed in backup process
- **Simpler code** - No SSL configuration, no external pg_dump requirement on host
- **Better security** - Credentials only exist inside the container

Both Podman and Kubernetes deployers now follow the same approach! All tests passing. 🎉

amalykhi:
Re: https://github.com/flightctl/flightctl/pull/2973#discussion_r3259775191

Good point about credentials! I investigated and found:

**Podman container** has PGPASSWORD set via secrets:
```
Secret=flightctl-postgresql-password,type=env,target=PGPASSWORD
```

**Kubernetes pod** has POSTGRESQL_PASSWORD set (not PGPASSWORD).

**Current implementation** passes the password explicitly via stdin, which:
- ✅ Works regardless of container environment setup
- ✅ Matches E2E test harness pattern (test/harness/e2e/harness_backup_restore.go also passes PGPASSWORD)
- ✅ Is secure (via stdin, not argv)

**Question:** Would you prefer we rely on the container environment variables instead? For Podman we could skip passing the password since PGPASSWORD is already set. For Kubernetes we'd need to verify if the existing env vars work.

Happy to refactor if you think using container env vars is cleaner!

internal/backup/deployer.go
```diff
@@ -68,6 +68,23 @@ func DetectDeployment(cfg *config.Config, log logrus.FieldLogger) (Deployer, err
...
+// Internal databases have hostnames: localhost, 127.0.0.1, flightctl-db, flightctl-db.flightctl-internal.
+// External databases (any other hostname) must be backed up separately by the user.
+func isInternalDB(cfg *config.Config) bool {
+	if cfg == nil || cfg.Database == nil {
+		return false
+	}
+
+	hostname := cfg.Database.Hostname
+	switch hostname {
+	case "localhost", "127.0.0.1", "flightctl-db", "flightctl-db.flightctl-internal":
```
asafbennatan: we should support other namespaces as FC support it ,
additionally we are usually configuring flightctl-db.flightctl-internal.svc.cluster.local.
flightctl-db.flightctl-internal would technically work but i dont see its every configured by us - we can keep it but we must add the above as well

internal/backup/deployer_test.go
asafbennatan, line 56: not part of this PR but i just noticed this will fail if the developer has /etc/flightctl folder , better point to a temp folder so the test is isolated

internal/backup/podman_deployer.go
```diff
@@ -1,29 +1,90 @@
...
+		p.cfg.Database.Name)
+
+	cmd := exec.CommandContext(ctx, "podman", "exec", "-i", containerName, "sh", "-c", pgDumpCmd)
+
+	// Pass password via stdin to avoid exposing it in process argv
+	cmd.Stdin = bytes.NewReader([]byte(password))
+
+	// Capture stdout (SQL dump) and stderr
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
```
asafbennatan: this could cause out of memory as we are holding the entire dump in memory , best to directly stream to file :

```
dumpFile := filepath.Join(dbDir, "dump.sql")
file, err := os.OpenFile(dumpFile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
if err != nil {
    return fmt.Errorf("failed to create dump file: %w", err)
}
defer file.Close()
cmd := exec.CommandContext(ctx, "podman", "exec", "-i", containerName, "sh", "-c", pgDumpCmd)
cmd.Stdin = bytes.NewReader([]byte(password))
cmd.Stdout = file  // Stream directly to file
cmd.Stderr = &stderr
```

(follow the same for K8)

internal/backup/kubernetes_deployer.go
```diff
@@ -1,29 +1,145 @@
...
+	}
+
+	// Prepare buffers for stdin/stdout/stderr
+	stdin := strings.NewReader(password)
+	var stdout, stderr bytes.Buffer
+
+	// Execute command
+	if err := executor.StreamWithContext(ctx, remotecommand.StreamOptions{
+		Stdin:  stdin,
+		Stdout: &stdout,
```
asafbennatan: output to file directly see more extensive example in the podman file

internal/backup/kubernetes_deployer.go
```diff
@@ -1,29 +1,145 @@
...
+	// Create db directory
+	dbDir := filepath.Join(outputDir, "db")
+	if err := os.MkdirAll(dbDir, 0755); err != nil {
+		return fmt.Errorf("failed to create db directory: %w", err)
+	}
+
+	// Hardcoded namespace and label for database pod.
+	// These match the default Helm chart deployment. Users who customize
+	// Helm values (different namespace or labels) will need to configure these.
+	namespace := "flightctl-internal"
```
asafbennatan: FC supports deploying for different namespaces - we should honor that and get the namespace as input

internal/backup/kubernetes_deployer_test.go
```diff
@@ -0,0 +1,137 @@
...
+	err := deployer.BackupDatabase(ctx, outputDir)
+
+	// Should succeed without error
+	require.NoError(t, err)
+
+	// Should log INFO message about external database
+	require.NotEmpty(t, hook.Entries)
+	found := false
+	for _, entry := range hook.Entries {
+		if entry.Level == logrus.InfoLevel && entry.Message == "External database detected. Please back up the database separately using your database backup tools." {
```
asafbennatan: the fact that we got an external DB should be returned as an error or as another output from the function .
we shouldnt need to check logs to get that info.

additionally it might makes sense to return the DB dump file as output as well

internal/backup/kubernetes_deployer_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+			// Verify db directory was created (this happens before Kubernetes client operations)
+			dbDir := filepath.Join(outputDir, "db")
+			stat, statErr := os.Stat(dbDir)
+			require.NoError(t, statErr, "db directory should be created")
+			require.True(t, stat.IsDir())
+
+			// Error expected since Kubernetes cluster is not available or pod doesn't exist
+			// The actual command execution and success path will be tested in integration tests
+			if err != nil {
```
asafbennatan: i dont think we need this assertion , we are basically asserting k8 go client which is unnecessary , in any case we shouldn't be relaying on sting contains in errors ,assuming there was any meaningful way to handle this different situation the proper approach would be to return a different error representing that case , but as far as i understand it is not meaningful - hence this should be checked

internal/backup/podman_deployer_test.go
```diff
@@ -0,0 +1,141 @@
...
+	err := deployer.BackupDatabase(ctx, outputDir)
+
+	// Should succeed without error
+	require.NoError(t, err)
+
+	// Should log INFO message about external database
+	require.NotEmpty(t, hook.Entries)
+	found := false
+	for _, entry := range hook.Entries {
+		if entry.Level == logrus.InfoLevel && entry.Message == "External database detected. Please back up the database separately using your database backup tools." {
```
asafbennatan: see above ( this shouldnt be checked in the logs but returned as output)

internal/backup/podman_deployer_test.go
```diff
@@ -0,0 +1,141 @@
...
+			err := deployer.BackupDatabase(ctx, outputDir)
+
+			// Verify db directory was created (this happens before podman exec execution)
+			dbDir := filepath.Join(outputDir, "db")
+			stat, statErr := os.Stat(dbDir)
+			require.NoError(t, statErr, "db directory should be created")
+			require.True(t, stat.IsDir())
+
+			// Error expected since flightctl-db container is not running in test environment
+			// The actual command execution and success path will be tested in integration tests
```
asafbennatan, line 122: lets add integration tests in this PR so we are sure this works

internal/backup/kubernetes_deployer_test.go
```diff
@@ -0,0 +1,137 @@
...
+			err := deployer.BackupDatabase(ctx, outputDir)
+
+			// Verify db directory was created (this happens before Kubernetes client operations)
+			dbDir := filepath.Join(outputDir, "db")
+			stat, statErr := os.Stat(dbDir)
+			require.NoError(t, statErr, "db directory should be created")
+			require.True(t, stat.IsDir())
+
+			// Error expected since Kubernetes cluster is not available or pod doesn't exist
+			// The actual command execution and success path will be tested in integration tests
```
asafbennatan: this probably can only be tested in E2E (since you will need a kind cluster)

internal/backup/kubernetes_deployer.go
```diff
@@ -1,29 +1,157 @@
...
+
+	// Create db directory
+	dbDir := filepath.Join(outputDir, "db")
+	if err := os.MkdirAll(dbDir, 0755); err != nil {
+		return fmt.Errorf("failed to create db directory: %w", err)
+	}
+
+	// Get namespace from service account token mount (standard Kubernetes mechanism).
+	// Falls back to "flightctl-internal" if detection fails.
+	namespace, err := getCurrentNamespace()
```
asafbennatan: current namespace as the one the user has set to default on his local machine ? 
why would that necessarily be where flightctl is installed ?
i think this should be an optional input to the command

internal/backup/deployer_test.go
```diff
@@ -72,7 +71,15 @@ func TestDetectDeployment(t *testing.T) {
 				origK8sHost := os.Getenv("KUBERNETES_SERVICE_HOST")
 				os.Unsetenv("KUBERNETES_SERVICE_HOST")
 
+				// Override base path to a temp directory without flightctl indicators
+				// This prevents test failure if developer has /etc/flightctl on their system
+				tmpDir := t.TempDir()
```
asafbennatan, line 72: im not sure how this works , i suggest that  deployer should receive flightctl base path as optional input , then  for testing pass some temp path

internal/backup/deployer.go
```diff
@@ -68,15 +79,45 @@ func DetectDeployment(cfg *config.Config, log logrus.FieldLogger) (Deployer, err
...
 
-// checkEnvironment examines the system to detect deployment indicators
-func checkEnvironment() detectionIndicators {
+// checkEnvironment examines the system to detect deployment indicators.
+// basePath is the root directory for Podman deployment detection (e.g., /etc/flightctl).
+func checkEnvironment(basePath string) detectionIndicators {
 	return detectionIndicators{
-		podmanPKIExists:       pathExists(flightctlBasePath + "/pki/ca.crt"),
-		podmanConfigDirExists: pathExists(flightctlBasePath),
+		podmanPKIExists:       pathExists(basePath + "/pki/ca.crt"),
```
asafbennatan, line 119: nit: use path join

asafbennatan (approved):

PR #2976, opened 2026-05-19
EDM-3874: fix flaky CI issue in template label

eldar101:
In this branch I changed the parametrisable template e2e test to stop using Docker Hub for the templated container app image.

The fix changes the test image source to the existing e2e image namespace already used by other application tests.

This keeps the same coverage: the test still verifies that a missing template label causes rollout failure, then adding the label renders the application spec and lets the device reconcile. The only behavior removed is the flaky dependency on pulling nginx:alpine from Docker Hub during standalone CI.


itzikb-redhat:
Thanks! Please consider using a constant and changing the tests that use docker images to quay.io instead

Samudelacruz77 (approved):
LGTM

siserafin (approved):
lgtm

PR #2983, opened 2026-05-19
EDM-3980: Enforce unique config provider names in validation

galel12:
## Problem

The `validateConfigs()` function did not enforce unique config provider names within a fleet or device spec. This allowed multiple configurations to share the same name, causing ambiguous `dependencySync.configRefs` references and silent data loss in fingerprint tracking.

## Root Cause

`validateConfigs()` tracked unique file paths via a `seenPath` map but had no equivalent check for config provider names. The dependency sync system (`buildDependencySyncStatus` in `device.go`) uses `ConfigProviderName` as a map key — duplicate names cause silent overwrite of fingerprint data.

## Fix

Added a `seenNames` map to `validateConfigs()` that rejects duplicate config provider names across all four provider types (git, HTTP, inline, K8s secret). This mirrors the existing `seenAppNames` pattern in `validateApplications()`.

## Testing

- 6 new regression test cases covering same-type and cross-type name collisions
- Verified tests fail without the fix and pass with it
- Full unit test suite passes (4249 tests, 0 regressions)

### DIFF
--- a/api/core/v1beta1/validation.go
+++ b/api/core/v1beta1/validation.go
@@ -119,6 +119,7 @@ func (r DeviceSpec) Validate(fleetTemplate bool) []error {
 func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
 	allErrs := []error{}
 	seenPath := make(map[string]struct{}, len(configs))
+	seenNames := make(map[string]struct{}, len(configs))
 	for i, config := range configs {
 		t, err := config.Type()
 		if err != nil {
@@ -126,6 +127,7 @@ func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
 			return allErrs
 		}
 
+		var configName string
 		switch t {
 		case GitConfigProviderType:
 			provider, err := config.AsGitConfigProviderSpec()
@@ -133,6 +135,7 @@ func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
 				allErrs = append(allErrs, err)
 				break
 			}
+			configName = provider.Name
 			allErrs = append(allErrs, provider.Validate(fleetTemplate)...)
 		case HttpConfigProviderType:
 			provider, err := config.AsHttpConfigProviderSpec()
@@ -140,6 +143,7 @@ func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
 				allErrs = append(allErrs, err)
 				break
 			}
+			configName = provider.Name
 			path := provider.HttpRef.FilePath
 			if _, exists := seenPath[path]; exists {
 				allErrs = append(allErrs, fmt.Errorf("spec.config[%d].httpRef, device path must be unique for all config providers: %s", i, path))
@@ -153,6 +157,7 @@ func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
 				allErrs = append(allErrs, err)
 				break
 			}
+			configName = provider.Name
 
 			for j, inline := range provider.Inline {
 				path := inline.Path
@@ -169,10 +174,20 @@ func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
 				allErrs = append(allErrs, err)
 				break
 			}
+			configName = provider.Name
 			allErrs = append(allErrs, provider.Validate(fleetTemplate)...)
 		default:
-			// if we hit this case, it means that the type should be added to the switch statement above
 			allErrs = append(allErrs, fmt.Errorf("unknown config provider type: %s", t))
+		}
+
+		// configName is empty only when type-specific decoding failed (break above).
+		// Empty names are separately rejected by each provider's Validate() (ValidateConfigName enforces minLen=1).
+		if configName != "" {
+			if _, exists := seenNames[configName]; exists {
+				allErrs = append(allErrs, fmt.Errorf("spec.config[%d]: duplicate config provider name: %s", i, configName))
+			} else {
+				seenNames[configName] = struct{}{}
+			}
 		}
 	}
 	return allErrs
--- a/api/core/v1beta1/validation_test.go
+++ b/api/core/v1beta1/validation_test.go
@@ -735,6 +735,10 @@ func TestValidateConfigs(t *testing.T) {
 	const (
 		dupName    = "my-config"
 		sharedName = "shared-name"
+		pathA      = "/path/a"
+		pathB      = "/path/b"
+		pathC      = "/path/c"
+		mntSecret  = "/mnt/secret"
 	)
 
 	require := require.New(t)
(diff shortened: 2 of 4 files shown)


api/core/v1beta1/validation.go
```diff
@@ -169,11 +174,19 @@ func validateConfigs(configs []ConfigProviderSpec, fleetTemplate bool) []error {
...
 				break
 			}
+			configName = provider.Name
 			allErrs = append(allErrs, provider.Validate(fleetTemplate)...)
 		default:
-			// if we hit this case, it means that the type should be added to the switch statement above
 			allErrs = append(allErrs, fmt.Errorf("unknown config provider type: %s", t))
 		}
+
+		if configName != "" {
```
kkyrazis, line 185: Why do we not need to validate empty config names? 

From the description:
> The dependency sync system (buildDependencySyncStatus in device.go) uses ConfigProviderName as a map key — duplicate names cause silent overwrite of fingerprint data.

Aren't multiple empty config names still duplicate names?

galel12: Good catch I added a comment clarifying. The `configName != ""` guard is for when type-specific decoding fails (`As*` returns error → `break` → configName stays zero-value). 

Empty names themselves are already rejected by each provider's `Validate()` which calls `ValidateConfigName` with minLen=1, so they can never pass validation.

kkyrazis (commented):
Overall I think the risk is low, but I did want to point out that this only fixes the issue for new specs. For specs that were already defined with duplicate names, this issue persists. Should the worker add some extra logging or something to make this issue more visible in those specific cases?

galel12:
@kkyrazis Good call, fixed in the 8ae41d7a389db09b9199f09758aa7f22464de8fa. 
`collectConfigRefs` now tracks seen names and emits a warning on the first duplicate it encounters (once per unique name, to avoid log spam). Extracted the logic into a `warnDuplicateConfigName` helper so it stays DRY across provider types. 
Tests updated to cover it.

internal/tasks/populate_dependency_refs_test.go
```diff
@@ -620,6 +620,78 @@ func TestShouldPopulateDependencyRefs(t *testing.T) {
 	})
 }
```
eldar101, line 622: docstring please.

internal/tasks/populate_dependency_refs_test.go
```diff
@@ -620,6 +620,78 @@ func TestShouldPopulateDependencyRefs(t *testing.T) {
...
+	})
+
+	t.Run("When three configs share a name it should log exactly one warning", func(t *testing.T) {
+		logger := logrus.New()
+		logger.SetLevel(logrus.WarnLevel)
+		hook := &logHook{}
+		logger.AddHook(hook)
+
+		config := &[]domain.ConfigProviderSpec{
+			makeGitConfigItem(t, "dup", "repo-a", "main"),
```
eldar101: think how to minimize this repetitive code please.

internal/tasks/populate_dependency_refs_test.go
```diff
@@ -620,6 +620,78 @@ func TestShouldPopulateDependencyRefs(t *testing.T) {
...
+
+	t.Run("When configs have unique names it should not log any warning", func(t *testing.T) {
+		logger := logrus.New()
+		logger.SetLevel(logrus.WarnLevel)
+		hook := &logHook{}
+		logger.AddHook(hook)
+
+		suffix := "/config.yaml"
+		config := &[]domain.ConfigProviderSpec{
+			makeGitConfigItem(t, "git-cfg", "repo-a", "main"),
```
eldar101: some strings here appear more than twice so please put them in vars.

api/core/v1beta1/validation_test.go
```diff
@@ -768,6 +768,56 @@ func TestValidateConfigs(t *testing.T) {
 			configs: []ConfigProviderSpec{},
 			wantErr: false,
 		},
+		{
+			name: "When duplicate http config names are used it should reject",
+			configs: []ConfigProviderSpec{
+				newNamedHttpConfigProviderSpec("my-config", "/path/a"),
```
eldar101: put in vars please.

internal/tasks/populate_dependency_refs_test.go
galel12: Done — added a docstring to the test function and extracted the logger/hook setup into a `newTestLogger` helper.

3b48a48a880faae28ba96ef6acc0d4512d459025

internal/tasks/populate_dependency_refs_test.go
galel12: Done — extracted repeated strings into constants (`testFleet`, `testBranch`, `dupName`, `httpSuffix`).

3b48a48a880faae28ba96ef6acc0d4512d459025

internal/tasks/populate_dependency_refs_test.go
galel12: Done — extracted the logger+hook setup into a `newTestLogger()` helper that all three sub-tests now share.

3b48a48a880faae28ba96ef6acc0d4512d459025

api/core/v1beta1/validation_test.go
galel12: Done — extracted `"my-config"` and `"shared-name"` into `dupName` and `sharedName` constants at the top of the test.

3b48a48a880faae28ba96ef6acc0d4512d459025

api/core/v1beta1/validation_test.go
```diff
@@ -768,6 +773,56 @@ func TestValidateConfigs(t *testing.T) {
 			configs: []ConfigProviderSpec{},
 			wantErr: false,
 		},
+		{
+			name: "When duplicate http config names are used it should reject",
+			configs: []ConfigProviderSpec{
+				newNamedHttpConfigProviderSpec(dupName, "/path/a"),
```
eldar101: move paths into vars please.

internal/tasks/populate_dependency_refs_test.go
```diff
@@ -620,6 +620,85 @@ func TestShouldPopulateDependencyRefs(t *testing.T) {
...
+
+// logHook captures logrus entries for test assertions.
+type logHook struct {
+	entries []logrus.Entry
+}
+
+func (h *logHook) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
```
eldar101, line 702: docstrings for all please

api/core/v1beta1/validation_test.go
galel12: Fixed in 8a5fcf53baafba6b5bb716b67ce2f6172a5ed0c2.

internal/tasks/populate_dependency_refs_test.go
galel12: Fixed in 8a5fcf53baafba6b5bb716b67ce2f6172a5ed0c2.

eldar101 (approved):

PR #2987, opened 2026-05-19
EDM-3953: add scripts/air-gap/mirror-images.sh for air-gapped artifact mirroring

aekubacki:
## Summary

- Implements `scripts/air-gap/mirror-images.sh` covering four dev stories under EDM-3953 (Image Mirroring and Artifact Preparation)
- Enumerates all RHEM container images for a given variant and generates `skopeo copy` commands ready to run in an air-gapped environment
- Writes a machine-readable `artifact-manifest-<variant>.yaml` listing images and RPM runtime dependencies

## Stories covered

| Story | Jira | Description |
|---|---|---|
| 1.01 | EDM-3957 | CLI scaffold (`--variant`, `--dest-registry`, `--execute`, `--help`) |
| 1.02 | EDM-3958 | Parse `deploy/helm/helm-chart-opts.yaml`; fall back to `Chart.yaml` `appVersion` for untagged images |
| 1.03 | EDM-3959 | Parse `packaging/images/el{9,10}/images.yaml` for observability images installed by RPMs |
| 1.04 | EDM-3960 | Generate `artifact-manifest-<variant>.yaml` with metadata, images[], rpms[], catalogs[] |

## How it works

```
./scripts/air-gap/mirror-images.sh \
    --variant community-el9 \
    --dest-registry local-registry.example.com:5000
```

- Reads image references from two sources and deduplicates across them before generating commands
- Outputs one `skopeo copy --all docker://<src> docker://<dest>` command per line to stdout (stderr carries `[INFO]`/`[WARN]`/`[ERROR]` logs — clean for piping)
- Without `--execute` the script is always a safe dry-run
- Parses `packaging/rpm/flightctl.spec` `Requires:` lines to populate the RPM list in the manifest

## Variants supported

`community-el9` · `community-el10` · `redhat-el9` · `redhat-el10`

## Test plan

- [x] `--help` output correct for all flags
- [x] `--variant community-el9` dry-run: 21 unique images, deduped across helm-chart-opts + observability files
- [x] Manifest `images` array length matches skopeo command count
- [x] `--variant redhat-el9` emits `registry.redhat.io` source URLs for downstream images
- [x] `--variant bad-variant` exits non-zero with clear error message
- [ ] QE: E2E validation against a live local registry (EDM-3961)

## Dependencies

- `yq` v4+ (mikefarah) — YAML parsing; script errors with install instructions if missing
- `skopeo` — only required when `--execute` is passed; script warns but still prints commands if absent


scripts/air-gap/mirror-images.sh
```diff
@@ -0,0 +1,691 @@
...
+# expansion leaves the variable unchanged if already set in the environment.
+HELM_CHART_OPTS="${HELM_CHART_OPTS:-${REPO_ROOT}/deploy/helm/helm-chart-opts.yaml}"
+
+# Helm chart metadata — used to read appVersion as the fallback tag for
+# flightctl component images that have no explicit tag in helm-chart-opts.yaml.
+CHART_YAML="${CHART_YAML:-${REPO_ROOT}/deploy/helm/flightctl/Chart.yaml}"
+
+# Secondary image source: observability images installed by RPMs.
+# These are NOT listed in helm-chart-opts.yaml so they must be sourced here.
+OBS_IMAGES_EL9="${OBS_IMAGES_EL9:-${REPO_ROOT}/packaging/images/el9/images.yaml}"
```
kkyrazis: while the helm-chart-opts contains variants for both upstream and downstream, there is no equivalent for a handful of downstream containers that a referenced only by the rpms. Off the top of my head: pam-issuer, userinfo-proxy, grafana, and prometheus. So this script is unable to appropriately pull down those images.

scripts/air-gap/mirror-images.sh
```diff
@@ -0,0 +1,691 @@
+#!/usr/bin/env bash
```
asafbennatan: should we consider doing this in golang? there is an ongoing effort in the codebase to avoid bash scripts for maintainability sake.
added benefits are :
1. integrates with existing testing mechanisms 
2. could avoid jq and skopeo dependencies (if uses golang libraries) - although not a must

scripts/air-gap/mirror-images/mirror.go
```diff
@@ -0,0 +1,121 @@
...
+// GenerateCommands prints one skopeo copy command per ImagePair to stdout and,
+// when execute is true, also runs the command immediately.
+//
+// stdout carries only the skopeo commands (pipe-safe); all progress logs go to
+// stderr via logInfo/logWarn/logError so they do not pollute captured output.
+//
+// All images are attempted even when one fails so the operator gets a complete
+// picture of what succeeded and what did not.  If any copy fails, an error is
+// returned after the loop so callers (and CI) receive a non-zero exit code.
+func GenerateCommands(ctx context.Context, pairs []ImagePair, execute bool, exec executer.Executer) error {
```
dudyas6: The generated skopeo commands fail against HTTP registries (e.g. plain `registry:2` for lab/testing):

> Get "https://host:5000/v2/": http: server gave HTTP response to HTTPS client

packaging/images/rhel10/images.yaml
kkyrazis, line 1: I don't think we should be publishing this upstream. I'm sure there is a good reason for the helm charts having it, but I imagine in the perfect world there would be no references to the downstream images in the charts either. 

Either way, the grafana and prometheus images are incorrect. These are the current references.

```yaml
  grafana:
    image: registry.redhat.io/rhel9/grafana
    tag: "9.7-1766395319"
  prometheus:
    image: registry.redhat.io/rhacm2/prometheus-rhel9
    tag: "v2.14.2-1"
```

asafbennatan (approved):
LGTM , 
i will add that i think this README.md got a bit out of focus - additionally documentation shouid be in the docs dir (perhaps we have something more focused there not sure)

packaging/images/rhel10/images.yaml
```diff
@@ -0,0 +1,23 @@
+# RPM-only images for the redhat-el10 (RHEM) variant.
```
kkyrazis, line 1: whats the point of el9/10 and rhel9/10 image files?

packaging/images/el10/images.yaml
```diff
@@ -1,58 +1,29 @@
-api:
```
kkyrazis, line 1: The rendering of the quadlet files depends on these files - probably why the rpm building tasks are failing - if we want to trigger everything based on some combo of the helm chart opts and these, we'd need to make some changes to the renderer

packaging/images/rhel10/images.yaml
aekubacki: should be fixed now.

scripts/air-gap/mirror-images/mirror.go
aekubacki: This should be resolved — import.sh defaults to  --dest-tls-verify=false since the bundle mode was added, and --execute mode has --insecure for HTTP registries.

aekubacki:
> LGTM , i will add that i think this README.md got a bit out of focus - additionally documentation shouid be in the docs dir (perhaps we have something more focused there not sure)

@asafbennatan I cleaned up the README.md please let me know if it looks better to you now.  All the actual documentation should be in the docs/user/installing section.

asafbennatan:
> > LGTM , i will add that i think this README.md got a bit out of focus - additionally documentation shouid be in the docs dir (perhaps we have something more focused there not sure)
> 
> @asafbennatan I cleaned up the README.md please let me know if it looks better to you now. All the actual documentation should be in the docs/user/installing section.

ok make sure tests pass and lets merge this (try rebasing first)

packaging/images/rhel10/images.yaml
aekubacki: Good question — the two tiers serve different purposes:

**`packaging/images/el9/` and `packaging/images/el10/`** are the *complete* image lists for the community variants. These are the authoritative config for the quadlet renderer: `flightctl.spec` passes them to `flightctl-standalone render quadlets --config` to generate the systemd unit files during the RPM build. The mirror tool also reads them, but `Dedup()` drops anything already covered by `helm-chart-opts.yaml`, so there's no duplication at runtime.

**`packaging/images/rhel9/` and `packaging/images/rhel10/`** are a *supplement* for the downstream (RHEM) variants only. They contain only the images that differ from community — specifically the `registry.redhat.io` equivalents for grafana, prometheus, pam-issuer, and userinfo-proxy. These files are NOT read by the RPM build for quadlet rendering; they exist only so the mirror tool can enumerate those extra downstream images when `--variant redhat-el9` or `--variant redhat-el10` is specified.

The RPM build failure you spotted was caused by a prior commit that inadvertently stripped `el9/el10/images.yaml` down to just the RPM-supplement subset. That's been restored in commit `34a3e4b` with the full image set.

Grafana and prometheus refs in both `rhel9/` and `rhel10/images.yaml` have been corrected to `registry.redhat.io/rhel9/grafana:9.7-1766395319` and `registry.redhat.io/rhacm2/prometheus-rhel9:v2.14.2-1` per your comment.

deploy/helm/flightctl/templates/tests/api-test.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   {{- end }}
   containers:
     - name: "{{ .Release.Name }}-connection-test"
-      image: "registry.access.redhat.com/ubi9/ubi-minimal:9.7-1763362218"
+      image: "{{ if .Values.ubiMinimal }}{{ .Values.ubiMinimal.image }}:{{ .Values.ubiMinimal.tag }}{{ else }}registry.access.redhat.com/ubi9/ubi-minimal:9.7-1763362218{{ end }}"
```
kkyrazis, line 16: just curious why the ubiminimal value never got passed along to require this?

packaging/images/rhel10/images.yaml
kkyrazis, line 1: referencing our discussion today, these rhel10/9 images.yaml files should be removed now no?

scripts/air-gap/mirror-images/manifest.go
```diff
@@ -0,0 +1,157 @@
...
+	"fmt"
+	"os"
+	"time"
+
+	"gopkg.in/yaml.v3"
+)
+
+// toolVersion is embedded in every manifest so operators can correlate a manifest
+// with the tool version that produced it.  Increment when parsing or output logic changes.
+const toolVersion = "2.0.0"
```
kkyrazis, line 13: 2.0.0?! this evolved very quickly.

packaging/images/rhel10/images.yaml
aekubacki: Should be fixed.

keitwb (approved):

keitwb (approved):

keitwb (approved):

PR #2990, opened 2026-05-21
EDM-3984: Fleet-owned devices never re-render after dependency sync change

galel12:
## Problem

Fleet-owned devices did not re-render their configuration after a dependency sync (git repo change, secret rotation, etc.) triggered a new template version rollout. The device appeared stuck on stale configuration until a manual spec change forced a re-render.

## Root Cause

Two issues prevented fleet-owned devices from updating after a dependency change:

1. **Spec-hash bypass** (`internal/tasks/device_render.go`): The spec-hash optimization short-circuits rendering when the device spec hash hasn't changed. This bypass only applied to `DependencyChangeDetected` events (standalone devices). Fleet-owned devices receive `FleetRolloutDeviceSelected` events — this event was not bypassed, causing the optimization to skip the re-render.

2. **Store fast-path** (`internal/store/device.go`): The `updateRendered` function had a `specUnchanged` fast-path that only updated `service_conditions` (fingerprints) without updating `rendered_config` or bumping `renderedVersion`. Even after bypassing the spec-hash check, the rendered content was never persisted and the agent never saw a new version to consume.

## Fix

1. Add `FleetRolloutDeviceSelected` to the spec-hash bypass condition alongside `DependencyChangeDetected`. Both event types indicate that external dependencies changed and the device must re-render regardless of spec hash.

2. Remove the `specUnchanged` fast-path in `updateRendered` so that dependency-change-triggered renders always persist the full rendered state, ensuring `renderedVersion` increments and the agent picks up the new configuration.

## Testing

- Added integration test in `test/integration/tasks/device_render_test.go` that:
  1. Creates a fleet-owned device with a K8s secret config provider
  2. Performs an initial render (ResourceCreated) — verifies spec hash is stored
  3. Simulates a dependency-change rollout by updating the template version annotation
  4. Triggers a second render with `FleetRolloutDeviceSelected`
  5. Asserts that `renderedVersion` bumps to "2" (proving the full update path ran)
  6. Asserts that `dependencySync.ConfigRefs` fingerprints are populated
- Test fails without either fix, passes with both

### DIFF
--- a/internal/store/device.go
+++ b/internal/store/device.go
@@ -1110,25 +1110,6 @@ func (s *DeviceStore) updateRendered(ctx context.Context, orgId uuid.UUID, name,
 		return false, "", nil
 	}
 
-	// Spec unchanged but fingerprints provided (dependency-change-triggered render):
-	// only update service_conditions with new fingerprints, skip the full re-render.
-	if specUnchanged {
-		if updatedServiceConditions != nil {
-			result = s.getDB(ctx).Model(existingRecord).Where("resource_version = ?", lo.FromPtr(existingRecord.ResourceVersion)).Updates(map[string]interface{}{
-				"service_conditions": updatedServiceConditions,
-				"resource_version":   gorm.Expr("resource_version + 1"),
-			})
-			err = ErrorFromGormError(result.Error)
-			if err != nil {
-				return strings.Contains(err.Error(), "deadlock"), "", err
-			}
-			if result.RowsAffected == 0 {
-				return true, "", flterrors.ErrNoRowsUpdated
-			}
-		}
-		return false, "", nil
-	}
-
 	existingAnnotations[domain.DeviceAnnotationRenderedSpecHash] = hash
 
 	renderedApplicationsJSON := renderedApplications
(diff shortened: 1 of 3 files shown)


internal/store/device.go
```diff
@@ -1110,25 +1110,6 @@ func (s *DeviceStore) updateRendered(ctx context.Context, orgId uuid.UUID, name,
 		return false, "", nil
 	}
 
-	// Spec unchanged but fingerprints provided (dependency-change-triggered render):
```
asafbennatan, line 1113: why was this removed ?

galel12: This block only updated `service_conditions` (the dependency fingerprints) but skipped updating `rendered_config` and bumping `renderedVersion`. So even after the spec-hash bypass in `device_render.go` correctly allowed re-rendering to proceed, the store would discard the rendered output — the agent never saw a new `renderedVersion` to consume and the device stayed `OutOfDate` indefinitely.

The guard at line 1109 (`specUnchanged && len(configFingerprints) == 0`) still provides the true no-op short-circuit for when nothing changed at all. Removing this fast-path just means that when fingerprints *do* change, we persist the full rendered state (including the new `renderedVersion`), which is the correct behavior.

asafbennatan (approved):
LGTM

PR #2991, opened 2026-05-21
EDM-3987: Defining two configurations from the same repository+target leads to "InternalTaskFailed" event for a fleet.

galel12:
## Problem

Defining two configurations from the same repository+target (e.g., same git repo and branch but different paths) causes an `InternalTaskFailed` warning event with "a resource with this name already exists".

## Root Cause

The `DependencyRef` composite PK is `(org_id, resource_key, fleet_name, device_name)`. The `ResourceKey` for git configs is `git:<repository>/<targetRevision>`, which is identical for two configs sharing repo+target but different paths. When `collectConfigRefs` produces multiple refs with the same key, `transactionalReplace` batch-inserts them and PostgreSQL rejects the duplicate.

## Fix

Deduplicate refs by `resource_key` before insertion. This aligns with the original design intent — the `dependency_refs` table is a polling work list, and the fingerprint (commit SHA) is per repo+revision, not per path. Multiple configs from the same repo+branch are one external dependency to poll.

Deduplication is applied in:
- `collectConfigRefs` — fleet/device creation and update path
- `getDeviceConfig` (via `appendUniqueRefs`) — fleet rollout path for parameterized revisions

No schema changes or migrations needed.

## Testing

- Added `TestCollectConfigRefs_DeduplicatesByResourceKey` with sub-tests for git, HTTP, and secret config types
- All existing unit tests pass

Fixes EDM-3987

### DIFF
--- a/internal/tasks/fleet_rollout.go
+++ b/internal/tasks/fleet_rollout.go
@@ -711,7 +711,7 @@ func (f FleetRolloutsLogic) getDeviceConfig(device *domain.Device, templateVersi
 	}
 
 	deviceConfig := []domain.ConfigProviderSpec{}
-	var depRefs []model.DependencyRef
+	depRefs := make(map[string]model.DependencyRef)
 	configErrs := []error{}
 	for _, configItem := range *templateVersion.Status.Config {
 		var newConfigItem *domain.ConfigProviderSpec
@@ -727,17 +727,23 @@ func (f FleetRolloutsLogic) getDeviceConfig(device *domain.Device, templateVersi
 		case domain.GitConfigProviderType:
 			var refs []model.DependencyRef
 			newConfigItem, refs, errs = f.replaceGitConfigParameters(device, configItem)
-			depRefs = append(depRefs, refs...)
+			for _, ref := range refs {
+				depRefs[ref.ResourceKey] = ref
+			}
 		case domain.KubernetesSecretProviderType:
 			var refs []model.DependencyRef
 			newConfigItem, refs, errs = f.replaceKubeSecretConfigParameters(device, configItem)
-			depRefs = append(depRefs, refs...)
+			for _, ref := range refs {
+				depRefs[ref.ResourceKey] = ref
+			}
 		case domain.InlineConfigProviderType:
 			newConfigItem, errs = f.replaceInlineConfigParameters(device, configItem)
 		case domain.HttpConfigProviderType:
 			var refs []model.DependencyRef
 			newConfigItem, refs, errs = f.replaceHTTPConfigParameters(device, configItem)
-			depRefs = append(depRefs, refs...)
+			for _, ref := range refs {
+				depRefs[ref.ResourceKey] = ref
+			}
 		default:
 			errs = append(errs, fmt.Errorf("%w: unsupported config type %q", ErrUnknownConfigName, configType))
 		}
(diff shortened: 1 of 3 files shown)


internal/store/model/dependencyref.go
```diff
@@ -51,18 +51,18 @@ func (StringArray) GormDataType() string {
...
 type DependencyRef struct {
-	OrgID       uuid.UUID `gorm:"type:uuid;primaryKey"`
-	ResourceKey string    `gorm:"primaryKey"` // e.g. "git:repo/ref", "http:repo/path", "secret:ns/name"
-	FleetName   *string   `gorm:"primaryKey;default:''"`
-	DeviceName  *string   `gorm:"primaryKey;default:''"`
+	OrgID              uuid.UUID `gorm:"type:uuid;primaryKey"`
+	ResourceKey        string    `gorm:"primaryKey"` // e.g. "git:repo/ref", "http:repo/path", "secret:ns/name"
+	FleetName          *string   `gorm:"primaryKey;default:''"`
+	DeviceName         *string   `gorm:"primaryKey;default:''"`
+	ConfigProviderName string    `gorm:"primaryKey;default:''"`
```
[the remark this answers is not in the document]
galel12: Addressed in b08c555d:

1. **Duplicate config names**: EDM-3980 (already ON_QA) enforces unique config names at the API validation layer, so the PK collision scenario for duplicate names is blocked before persistence. The `warnDuplicateConfigName` log is a safety net for pre-existing data only.

2. **Missing DB migration**: Added a `SchemaMigration`-guarded DDL step that drops the old 4-column PK and recreates it with `config_provider_name` included. Uses the same pattern as `backfillDefaultCatalogs` — idempotent, safe for concurrent replicas.

internal/store/dependencyref.go
```diff
@@ -45,15 +45,44 @@ func (s *DependencyRefStore) getDB(ctx context.Context) *gorm.DB {
...
+	if result.RowsAffected == 0 {
+		return nil
+	}
+
+	return db.Exec(`
+		ALTER TABLE dependency_refs
+		  DROP CONSTRAINT IF EXISTS dependency_refs_pkey,
+		  ADD PRIMARY KEY (org_id, resource_key, fleet_name, device_name, config_provider_name)
+	`).Error
 }
```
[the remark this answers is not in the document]
galel12: fixed in d70f2013772379d9a24fb3586a5ed1005dc212e4

internal/tasks/fleet_rollout.go
```diff
@@ -712,6 +712,7 @@ func (f FleetRolloutsLogic) getDeviceConfig(device *domain.Device, templateVersi
 
 	deviceConfig := []domain.ConfigProviderSpec{}
 	var depRefs []model.DependencyRef
```
asafbennatan: shouldn't we just make depRefs a map ? 
currently we have it as an array and have a seenkeys next to it which feels redundant 

same comment for additional locations below

internal/tasks/populate_dependency_refs.go
```diff
@@ -183,13 +194,18 @@ func collectConfigRefs(log logrus.FieldLogger, config *[]domain.ConfigProviderSp
 				continue
 			}
 			warnDuplicateConfigName(log, seenNames, warnedNames, k8sSpec.Name, i, fleetName, deviceName)
+			key := fmt.Sprintf("secret:%s/%s", k8sSpec.SecretRef.Namespace, k8sSpec.SecretRef.Name)
```
asafbennatan, line 194: could there be a bug here ? previously we inserted two rows for (resourceKey,deviceA) (resourceKey,deviceB)

now we will insert only one . 

instead just take the primary key  as the depRefs map key (have it as a map )

internal/tasks/fleet_rollout.go
galel12: Refactored to use a map in 76149cab — removed the separate seenKeys + appendUniqueRefs.

internal/tasks/populate_dependency_refs.go
galel12: No bug here — collectConfigRefs is always called for a single (fleet, device) pair, so resource_key is the correct dedup key within one invocation. But switched to a map anyway in 76149cab for clarity.

asafbennatan (approved):
LGTM

PR #2992, opened 2026-05-21
EDM-3986: Fix incomplete message in DeviceOSImageChanged event

galel12:
## Problem

When a device first reports its OS image after enrollment, the `DeviceOSImageChanged` event message reads "OS image changed from  to sha256:..." — the empty "from" field is confusing for operators querying events.

## Root Cause

`ComputeDeviceStatusChanges` emits the event whenever `oldDigest != newDigest && newDigest != ""`. For a newly enrolled device, `oldDigest` is `""` (no prior image), so the format string produces a blank "from" value.

## Fix

Add a conditional: when `oldDigest == ""`, emit `"Initial OS image detected: <digest>"` instead. The event is still emitted (required by CVE lifecycle processing in `vulnerability_sync.go`) — only the message text is fixed.

## Testing

- Added `TestComputeDeviceStatusChanges_OSImageChanged_EDM3986` with 4 table-driven sub-tests covering: initial image, image change, no change, empty new digest
- All existing unit tests pass

### DIFF
--- a/internal/service/common/device_test.go
+++ b/internal/service/common/device_test.go
@@ -76,6 +76,79 @@ func TestComputeDeviceStatusChanges_DeviceUpdateFailed(t *testing.T) {
 	assert.Contains(t, updates[0].Details, "has not been updated")
 }
 
+func TestComputeDeviceStatusChanges_OSImageChanged_EDM3986(t *testing.T) {
+	ctx := context.Background()
+	orgId := uuid.New()
+
+	tests := []struct {
+		name            string
+		oldDigest       string
+		newDigest       string
+		expectEvent     bool
+		expectedDetails string
+	}{
+		{
+			name:            "When initial OS image is reported it should emit event with 'Initial OS image detected' message",
+			oldDigest:       "",
+			newDigest:       "sha256:abc123",
+			expectEvent:     true,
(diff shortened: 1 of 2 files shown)


internal/service/common/device_test.go
```diff
@@ -76,6 +76,79 @@ func TestComputeDeviceStatusChanges_DeviceUpdateFailed(t *testing.T) {
...
+			expectedDetails: "OS image changed from sha256:old111 to sha256:new222",
+		},
+		{
+			name:        "When OS image is unchanged it should not emit event",
+			oldDigest:   "sha256:same",
+			newDigest:   "sha256:same",
+			expectEvent: false,
+		},
+		{
+			name:        "When new digest is empty it should not emit event",
```
celdrake, line 111: I know this is existing behaviour, but we probably should emit an EVent in this case.

Something like "the OS image in this device will stop being managed" as it's an important change to the device.

celdrake (approved):

internal/service/common/device_test.go
galel12: Good point, an "OS image unset" event would be useful for operators. This is new behavior though (the existing code never emitted for empty newDigest), so I'll track it as a follow-up rather than expanding this bug-fix PR. Filed as part of the observability epic.

PR #2998, opened 2026-05-25
NO-ISSUE: Fix dependency ref test ordering

eldar101:
The fix moves the dependency-ref test to order-independent assertions.

The failing test expected git at capturedRefs[0] and http at capturedRefs[1], but the refs can arrive in either order. The updated test builds a map keyed by ResourceKey, then verifies both expected refs exist and have the correct RefType.


galel12 (approved):
LGTM

asafbennatan (approved):
LGTM

PR #3003, opened 2026-05-26
NO-ISSUE: Fix flaky dependency ref tests assuming map iteration order

galel12:
## Summary
- `getDeviceConfig` and `collectConfigRefs` both use a map for deduplication then convert to a slice, making output order non-deterministic across Go runtime invocations.
- `TestReplaceGitConfigParameters_DeviceLevelRefs` and `TestCollectConfigRefs_DeduplicatesByResourceKey` were asserting by slice index, causing intermittent failures in the merge queue.
- Switched both tests to map-based lookups (by `ResourceKey`) so assertions are order-independent.

## Verification
- Audited all multi-ref test assertions across `internal/tasks/` and `internal/store/` — no other tests assert positional order on map-derived slices.
- Ran affected tests with `-count=10` — all pass consistently.

### DIFF
--- a/internal/tasks/fleet_rollout_test.go
+++ b/internal/tasks/fleet_rollout_test.go
@@ -1045,11 +1045,16 @@ func TestReplaceGitConfigParameters_DeviceLevelRefs(t *testing.T) {
 		require.Empty(t, errs)
 		require.Len(t, refs, 2)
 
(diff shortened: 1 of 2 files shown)


asafbennatan (approved):
LGTM

PR #3006, opened 2026-05-26
EDM-3893: Implement service configuration collection

amalykhi:
## EDM-3893: Implement service configuration collection

**Jira:** https://issues.redhat.com/browse/EDM-3893  
**Story type:** [DEV]  
**Epic:** Epic 1 — Backup Command Implementation (EDM-3884)

### Summary

Extends `BackupConfig()` methods in PodmanDeployer and KubernetesDeployer to collect service configuration files for the backup workflow. Podman deployments back up service.yaml and PAM Issuer volume; Kubernetes deployments export the API server ConfigMap and Helm values. All components handle missing optional files gracefully (log warning, continue).

Additionally fixes two security/reliability issues in the merged EDM-3892 code: (1) adds context cancellation support in archive creation for large file transfers, and (2) implements all-or-nothing security for staging directory cleanup.

### Changes

**Podman deployer (`internal/backup/podman_deployer.go`):**
- Implements `BackupConfig()` to copy `/etc/flightctl/service.yaml` to `<outputDir>/config/service.yaml`
- Backs up PAM Issuer volume via `podman volume export` to `<outputDir>/volumes/pam-issuer-etc.tar`
- Treats service.yaml as required (errors if missing), PAM volume as optional (warns if unavailable)
- Includes context cancellation check before volume export

**Kubernetes deployer (`internal/backup/kubernetes_deployer.go`):**
- Implements `BackupConfig()` to export `flightctl-api-config` ConfigMap to YAML
- Extracts user-supplied Helm values via `helm get values` command
- Tries common Helm release names in order (flightctl, flightctl-api)
- Treats ConfigMap as required, Helm values as optional (warns if helm CLI unavailable)
- Includes context cancellation check in helm release loop
- Fixes resource leak (file descriptors not closed on retry) and early return preventing retries

**Archive creation (`internal/backup/backup.go`):**
- Replaces `io.Copy` with chunked copy loop (32KB buffer) that checks context cancellation after each chunk
- Handles partial writes with explicit short-write detection
- Implements all-or-nothing staging cleanup: removes archive and checksum if staging cleanup fails (prevents sensitive data from remaining on disk when cleanup fails)

### Testing

**Unit tests:**
- 7 new tests for BackupConfig() methods (Podman: 3, Kubernetes: 4)
- Tests verify error handling, directory permissions (0700), file permissions (0600), graceful degradation for optional components
- Uses fake Kubernetes clientset and temporary directories
- All 36 tests in internal/backup package passing

**Coverage:**
- Behavioral contracts comprehensively tested through public API
- Lower line coverage for infrastructure code expected (requires actual podman/kubectl/helm CLI)
- Full command execution tested in E2E tests (Story 1.07/EDM-3892)

**Quality review:**
- Initial validation review: 3 findings fixed (resource leak, context cancellation checks)
- Cross-cutting review: 3 findings fixed (Helm file descriptor leak, early return fix, staging cleanup all-or-nothing)

### Acceptance Criteria

- [x] AC-1: Podman backup copies `/etc/flightctl/service.yaml` to `<temp-dir>/config/service.yaml`
- [x] AC-2: Podman backup backs up PAM Issuer volume to `<temp-dir>/volumes/pam-issuer-etc.tar`
- [x] AC-3: Kubernetes backup exports service ConfigMap to `<temp-dir>/config/flightctl-api-config.yaml`
- [x] AC-4: Kubernetes backup captures Helm values to `<temp-dir>/config/helm-values.yaml`
- [x] AC-5: Backup logs configuration files collected
- [x] AC-6: Backup handles missing optional files gracefully (log warning, continue)
- [x] AC-7: Unit tests cover config collection for both deployment types

### Additional Fixes (EDM-3892 follow-up)

- [x] Context cancellation in `createTarGz` honors cancellation during large file transfers
- [x] Staging cleanup failure triggers archive removal (all-or-nothing security)
- [x] Helm values extraction fixes resource leak and retry logic


internal/backup/backup.go
```diff
@@ -350,13 +375,15 @@ func PerformBackup(ctx context.Context, deployer Deployer, outputDir string, log
...
-	} else {
-		stagingCleaned = true
-		log.Debugf("Staging directory cleaned up: %s", stagingDir)
+		// Cleanup failed - remove archive and checksum for all-or-nothing security
+		os.Remove(archivePath)
+		os.Remove(checksumPath)
+		return "", fmt.Errorf("backup failed: unable to clean up staging directory %s (contains sensitive data): %w", stagingDir, err)
 	}
+	stagingCleaned = true
+	log.Debugf("Staging directory cleaned up: %s", stagingDir)
```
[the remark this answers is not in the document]
amalykhi: fixed

internal/backup/backup.go
```diff
@@ -350,13 +375,15 @@ func PerformBackup(ctx context.Context, deployer Deployer, outputDir string, log
...
+	// Clean up staging directory (contains sensitive data - must succeed)
 	if err := os.RemoveAll(stagingDir); err != nil {
-		log.Warnf("Failed to clean up staging directory %s: %v", stagingDir, err)
-	} else {
-		stagingCleaned = true
-		log.Debugf("Staging directory cleaned up: %s", stagingDir)
+		// Cleanup failed - remove archive and checksum for all-or-nothing security
+		os.Remove(archivePath)
+		os.Remove(checksumPath)
+		return "", fmt.Errorf("backup failed: unable to clean up staging directory %s (contains sensitive data): %w", stagingDir, err)
```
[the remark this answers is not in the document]
amalykhi: fixed

internal/backup/kubernetes_deployer_test.go
```diff
@@ -137,3 +139,114 @@ func TestKubernetesDeployer_BackupPKI_NoDirectoryOnValidationFailure(t *testing.
...
+	// Use fake clientset with the ConfigMap
+	fakeClient := fake.NewSimpleClientset(configMap)
+	deployer := NewKubernetesDeployer(cfg, log, "flightctl", "", fakeClient)
+	ctx := context.Background()
+	outputDir := t.TempDir()
+
+	// BackupConfig should succeed even if helm command fails (helm values are optional)
+	// In test environment, helm CLI is likely not available, but this should not cause an error
+	err := deployer.BackupConfig(ctx, outputDir)
+	require.NoError(t, err, "BackupConfig should succeed even if helm values extraction fails")
```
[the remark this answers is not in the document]
amalykhi: fixed

internal/backup/podman_deployer.go
```diff
@@ -216,8 +216,64 @@ func (p *PodmanDeployer) BackupPKI(ctx context.Context, outputDir string) error
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("service configuration file not found: %s", serviceConfigSrc)
+		}
+		return fmt.Errorf("failed to stat service configuration file: %w", err)
+	}
+
+	// Copy file preserving permissions
+	if err := copyFilePreserveMode(serviceConfigSrc, serviceConfigDst, srcInfo.Mode().Perm(), p.log); err != nil {
+		return fmt.Errorf("failed to copy service configuration: %w", err)
```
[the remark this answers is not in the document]
amalykhi: fixed

internal/backup/podman_deployer.go
```diff
@@ -216,8 +216,64 @@ func (p *PodmanDeployer) BackupPKI(ctx context.Context, outputDir string) error
...
+
+	// Execute podman volume export
+	cmd := exec.CommandContext(ctx, "podman", "volume", "export", volumeName, "-o", volumeArchive)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		// Log warning but don't fail - PAM Issuer volume is optional
+		p.log.Warnf("PAM Issuer volume export failed (volume may not exist or podman unavailable): %v (output: %s)", err, string(output))
+		p.log.Warn("Continuing backup without PAM Issuer volume")
+		return nil
+	}
```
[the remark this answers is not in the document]
amalykhi: fixed

internal/backup/backup.go
```diff
@@ -133,12 +133,37 @@ func createTarGz(ctx context.Context, stagingDir string, archivePath string, log
...
+						return fmt.Errorf("failed to write file content for %s: short write", path)
+					}
+				}
+				if err != nil {
+					if err == io.EOF {
+						break
+					}
+					return fmt.Errorf("failed to read file %s: %w", path, err)
+				}
 			}
```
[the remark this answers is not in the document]
amalykhi: already fixed. Recheck

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +286,108 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+
+		// Success! Helm values extracted
+		k.log.Infof("Helm values backed up from release: %s", releaseName)
+		return nil
+	}
+
+	// All release names failed - log warning but don't error (Helm values are optional)
+	k.log.Warnf("Helm values extraction failed for all release names: %v (helm CLI may not be available)", releaseNames)
+	k.log.Warn("Continuing backup without Helm values")
 	return nil
```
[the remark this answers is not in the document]
amalykhi: already fixed. recheck

internal/backup/podman_deployer.go
```diff
@@ -216,8 +216,68 @@ func (p *PodmanDeployer) BackupPKI(ctx context.Context, outputDir string) error
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		// Clean up any partial archive before returning
+		if removeErr := os.Remove(volumeArchive); removeErr != nil && !os.IsNotExist(removeErr) {
+			p.log.Warnf("Failed to remove partial PAM Issuer archive %s: %v - manual cleanup may be required", volumeArchive, removeErr)
+		}
+		// Log warning but don't fail - PAM Issuer volume is optional
+		p.log.Warnf("PAM Issuer volume export failed (volume may not exist or podman unavailable): %v (output: %s)", err, string(output))
+		p.log.Warn("Continuing backup without PAM Issuer volume")
+		return nil
```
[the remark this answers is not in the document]
amalykhi: fixed

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +286,115 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+			}
+			// Helm command failed - remove empty file and try next release name
+			os.Remove(helmValuesPath)
+			k.log.Debugf("Helm values extraction failed for release %s: %v (stderr: %s)", releaseName, cmdErr, stderr.String())
+			continue
+		}
+
+		// Success! Helm values extracted
+		k.log.Infof("Helm values backed up from release: %s", releaseName)
+		return nil
```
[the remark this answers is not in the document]
amalykhi: fixed. recheck

internal/backup/podman_deployer.go
```diff
@@ -216,8 +216,75 @@ func (p *PodmanDeployer) BackupPKI(ctx context.Context, outputDir string) error
...
-	p.log.Debug("BackupConfig called (stub implementation)")
+	// Create config directory
+	configDir := filepath.Join(outputDir, "config")
+	if err := os.MkdirAll(configDir, pkiDirMode); err != nil {
+		return fmt.Errorf("failed to create config directory: %w", err)
+	}
+
+	// Copy service.yaml (required file)
+	// Service config path for Podman deployments (standard installation location)
+	serviceConfigSrc := "/etc/flightctl/service.yaml"
```
asafbennatan: i think this is incorrect ? 
the file is /etc/flightctl/service-config.yaml

asafbennatan (changes requested):

internal/backup/kubernetes_deployer.go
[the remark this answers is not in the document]
amalykhi: Fixed. recheck

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +286,125 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+	yamlPath := filepath.Join(configDir, configMapName+".yaml")
+	if err := os.WriteFile(yamlPath, yamlBytes, pkiFileMode); err != nil {
+		return fmt.Errorf("failed to write ConfigMap YAML %s: %w", configMapName, err)
+	}
+
+	k.log.Info("Service ConfigMap backed up")
+
+	// Backup Helm values (optional component)
+	// Try common Helm release names in order
+	releaseNames := []string{"flightctl", "flightctl-api"}
```
asafbennatan: flightctl-api is an actual possible release name ? 
get this as input for k8deployer

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +286,125 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+	// Create config directory
+	configDir := filepath.Join(outputDir, "config")
+	if err := os.MkdirAll(configDir, pkiDirMode); err != nil {
+		return fmt.Errorf("failed to create config directory: %w", err)
+	}
+
+	// Use namespace (set in constructor, defaults to "flightctl")
+	namespace := k.namespace
+	// ConfigMap name must match Helm chart (deploy/helm/flightctl/templates/api/flightctl-api-config.yaml)
+	configMapName := "flightctl-api-config"
```
asafbennatan: what about other config maps ?

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +286,125 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+	}
+
+	k.log.Info("Service ConfigMap backed up")
+
+	// Backup Helm values (optional component)
+	// Try common Helm release names in order
+	releaseNames := []string{"flightctl", "flightctl-api"}
+	helmValuesPath := filepath.Join(configDir, "helm-values.yaml")
+
+	for _, releaseName := range releaseNames {
```
asafbennatan: no need to iterate over possibilities , get release name as input , have sensible default - if it fails , fail the backup

internal/backup/podman_deployer.go
```diff
@@ -216,8 +216,75 @@ func (p *PodmanDeployer) BackupPKI(ctx context.Context, outputDir string) error
...
-	p.log.Debug("BackupConfig called (stub implementation)")
+	// Create config directory
+	configDir := filepath.Join(outputDir, "config")
+	if err := os.MkdirAll(configDir, pkiDirMode); err != nil {
+		return fmt.Errorf("failed to create config directory: %w", err)
+	}
+
+	// Copy service-config.yaml (required file)
+	// Service config path for Podman deployments (standard installation location)
+	serviceConfigSrc := "/etc/flightctl/service-config.yaml"
```
asafbennatan: get this as input (set default in constructor)

internal/backup/podman_deployer.go
amalykhi: Good catch! Fixed in edbb398a - changed to `/etc/flightctl/service-config.yaml` and added as constructor parameter following the same pattern as `pkiPath`.

internal/backup/podman_deployer.go
amalykhi: Done in edbb398a - added `serviceConfigPath` as constructor parameter with default `/etc/flightctl/service-config.yaml`. Follows the same pattern as `pkiPath` for easier unit testing.

internal/backup/kubernetes_deployer.go
amalykhi: Good point - added `helmReleaseName` as constructor parameter with "flightctl" as default in 9346c3d8. Removed the hardcoded list and iteration - now uses single release name and fails backup if Helm extraction fails (instead of silently continuing).

internal/backup/kubernetes_deployer.go
amalykhi: Agreed - removed iteration in 9346c3d8. Now uses `helmReleaseName` from constructor (default "flightctl") and fails backup if Helm extraction fails instead of silently continuing.

internal/backup/kubernetes_deployer.go
amalykhi: `flightctl-api-config` is the service configuration ConfigMap (per acceptance criteria AC-3). Other ConfigMaps like `flightctl-db-config`, `flightctl-prometheus-config` are component-specific configs, not part of the "service configuration" backup scope defined in this story. Should we expand scope to include them?

asafbennatan: nothing special about flightctl-api-config , if we have decided we should keep a snapshot of the actual config maps we should backup all of them . 
what we could do is backup only helm values -but this requires that we have the chart at restore time - this means i dont see any good reason backing up helm values at all as they are not restorable without the original chart

third option would be to backup the chart as well , but i think this is just more complex - backing up individual config maps seems the easy choice here

amalykhi: Great point about the architectural inconsistency. You're absolutely right - there's nothing special about `flightctl-api-config`.

I've expanded the implementation in 456b5f64 to backup **all FlightCtl-owned ConfigMaps** using label selector `app.kubernetes.io/name=flightctl`. This:
- Backs up all ConfigMaps from the Helm chart (api, worker, alertmanager, etc.)
- Filters out system/unrelated ConfigMaps automatically
- Provides comprehensive configuration backup for restore

Regarding Helm values: I kept them in the backup for now since they document user customizations and are required by acceptance criteria AC-4. You're right they need the chart at restore time - we can revisit this in a future story if needed.

Changes:
- List ConfigMaps with label selector instead of Get single ConfigMap
- Iterate through all matches and export each to YAML
- Added test for label filtering
- All 41 unit tests passing

asafbennatan: following up a discussion i had with @rawagner  - we should not backup any config maps , instead we should backup the helm values and the chart - and when restoring the backup we should use helm upgrade

amalykhi:
After discussing with Rasti, we've changed the architectural approach for Kubernetes config backup:

**Problem with ConfigMaps backup:**
- ConfigMaps are generated artifacts from chart + values
- If we restore ConfigMaps directly, Helm doesn't know about them
- On next `helm upgrade`, Helm regenerates ConfigMaps from its stored values
- Result: restored configuration gets overwritten ❌

**New approach - Helm Secret backup:**
- Helm 3 stores complete release state in Secrets: `sh.helm.release.v1.<name>.v<revision>`
- Secret's `data.release` field contains (base64 + gzip + JSON):
  - `chart` - full chart (templates, Chart.yaml, values, files)
  - `config` - user-supplied values  
  - `manifest` - rendered output
- Backup the Helm Secret → self-contained, includes everything needed
- Restore: apply Secret → Helm sees existing release → `helm upgrade` works correctly ✅

**Changes in 04ff0a2c:**
- Removed ConfigMaps List/export code
- Removed `helm get values` command execution
- Added Helm Secret backup with label selector `owner=helm,name=<release>`
- Automatically selects latest revision (highest `.v<N>`)
- Updated all tests (5 tests for Helm Secret behavior)
- All 41 tests passing, lint clean

This is the only bulletproof way to backup/restore Helm-managed configuration.

internal/backup/kubernetes_deployer.go
amalykhi: Implemented as suggested via extracting secrets # 1. Find the secret name
kubectl get secrets -n <namespace> | grep sh.helm.release

# 2. Extract, decode, and unzip it (Requires 'jq')
kubectl get secret sh.helm.release.v1.<release-name>.v1 -n <namespace> -o jsonpath='{.data.release}' | base64 -d | base64 -d | gzip -d > release.json

should work

asafbennatan:
> After discussing with Rasti, we've changed the architectural approach for Kubernetes config backup:
> 
> **Problem with ConfigMaps backup:**
> 
> * ConfigMaps are generated artifacts from chart + values
> * If we restore ConfigMaps directly, Helm doesn't know about them
> * On next `helm upgrade`, Helm regenerates ConfigMaps from its stored values
> * Result: restored configuration gets overwritten ❌
> 
> **New approach - Helm Secret backup:**
> 
> * Helm 3 stores complete release state in Secrets: `sh.helm.release.v1.<name>.v<revision>`
> * Secret's `data.release` field contains (base64 + gzip + JSON):
>   
>   * `chart` - full chart (templates, Chart.yaml, values, files)
>   * `config` - user-supplied values
>   * `manifest` - rendered output
> * Backup the Helm Secret → self-contained, includes everything needed
> * Restore: apply Secret → Helm sees existing release → `helm upgrade` works correctly ✅
> 
> **Changes in [04ff0a2](https://github.com/flightctl/flightctl/commit/04ff0a2c157515ff12632d2f21cb92e25304888b):**
> 
> * Removed ConfigMaps List/export code
> * Removed `helm get values` command execution
> * Added Helm Secret backup with label selector `owner=helm,name=<release>`
> * Automatically selects latest revision (highest `.v<N>`)
> * Updated all tests (5 tests for Helm Secret behavior)
> * All 41 tests passing, lint clean
> 
> This is the only bulletproof way to backup/restore Helm-managed configuration.

note - backing up the full secret is fine , but when restoring we should use helm upgrade - applying the secret is not enough as we need helm to reeval the values which regenerates the config maps - thats the reason we dont need to back them up

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +292,124 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+// Restore process:
+// 1. Apply the Secret: kubectl apply -f helm-release-<name>-v<revision>.yaml
+// 2. Run helm upgrade: helm upgrade <release> <chart> -n <namespace>
+//    - Helm sees the existing release (from the restored Secret)
+//    - Re-evaluates values and regenerates ConfigMaps automatically
+//    - ConfigMaps are recreated with correct configuration
+// 3. ConfigMaps backup is not needed - they are regenerated by Helm
+//
+// Returns error if no Helm release Secrets are found.
 func (k *KubernetesDeployer) BackupConfig(ctx context.Context, outputDir string) error {
```
asafbennatan, line 314: Helm 3 already stores the revision number as a label on each Secret — secret.Labels["version"]. No name parsing needed. The current approach breaks for any release name containing .v followed by digits (e.g., a release named v2-app would produce name sh.helm.release.v1.v2-app.v5 → split gives ["sh.helm.release.", "1.", "2-app.", "5"] which happens to still work, but this is accidental).

Problem 2: Finding the max revision by iterating is also unnecessary

Helm marks the currently-deployed revision with the label status=deployed. Superseded revisions get status=superseded. So you can get exactly the right Secret with a single query — no loop, no comparison:

labelSelector := fmt.Sprintf("owner=helm,name=%s,status=deployed", releaseName)
That returns one Secret directly. Combined with reading the revision from the label:

revision, err := strconv.Atoi(secret.Labels["version"])
The whole BackupConfig simplifies to: list with owner=helm,name=X,status=deployed → take the single result → marshal → write. Roughly 40 lines fewer, no string parsing, and semantically correct (you get the deployed state, not just the numerically highest revision which could be a failed or pending-upgrade state).

The one edge case to handle: if the status label is missing or the list returns 0 items (e.g., a failed initial install), the current max-revision fallback is more defensive. You could handle that with a two-step query: try status=deployed first, fall back to max revision by label if empty.

amalykhi: Excellent catch on the name parsing bug and the unnecessary iteration! You're absolutely right:

1. The `.v` split breaks for release names containing `.v` followed by digits
2. Using `status=deployed` is semantically correct and simpler
3. Reading revision from `secret.Labels["version"]` is more robust

Implementing the two-step approach:
1. Try `owner=helm,name=X,status=deployed` first
2. Fall back to max revision by `version` label if empty

This eliminates ~40 lines of string parsing and iteration. Fixing now.

internal/backup/kubernetes_deployer.go
```diff
@@ -285,9 +292,148 @@ func (k *KubernetesDeployer) BackupPKI(ctx context.Context, outputDir string) er
...
+			latestRevision, parseErr = strconv.Atoi(versionLabel)
+			if parseErr != nil {
+				return fmt.Errorf("failed to parse version label %q from Secret %s: %w", versionLabel, latestSecret.Name, parseErr)
+			}
+		} else {
+			return fmt.Errorf("Helm Secret %s missing version label", latestSecret.Name)
+		}
+		k.log.Infof("Found deployed Helm release Secret: %s (revision %d)", latestSecret.Name, latestRevision)
+	} else {
+		// Fallback: no deployed release found, find highest revision by version label
```
asafbennatan: dont fallback to anything , the whole idea was to simplify the logic

asafbennatan: P.S not sure why we need version at all - dont we care only for the secret ?

amalykhi: You're right - the fallback defeats the simplification goal. Removing the entire else block. If there's no deployed release, we should fail explicitly rather than trying to guess which Secret to use.

amalykhi: Good point - we don't need the version at all. Removed version extraction and simplified the filename to just `helm-release-<name>.yaml`. The Secret itself contains all the version information anyway.

asafbennatan (approved):
LGTM

PR #3009, opened 2026-05-26
EDM-3895: EDM-3896,EDM-3897

asafbennatan:
adds restore command


internal/restore/restore.go
```diff
@@ -0,0 +1,111 @@
...
+	kv kvstore.KVStore,
+	log logrus.FieldLogger,
+) (retErr error) {
+	log.Infof("Verifying checksum of archive %q", archivePath)
+	if err := verifyChecksum(archivePath); err != nil {
+		return fmt.Errorf("archive integrity check failed: %w", err)
+	}
+
+	log.Infof("Extracting archive %q", archivePath)
+	extractDir, err := ExtractArchive(ctx, archivePath)
```
amalykhi, line 45: I would suggest to extract and verify in one step instead of 
1. Read once for checksum verification
  2. Closed
  3. Read again for extraction

cmd/flightctl-restore/main.go
```diff
@@ -93,19 +118,12 @@ func runRestore(ctx context.Context) error {
 		}
 	}()
 
-	log.Println("Initializing KV store connection for restore operations")
+	log.Println("Initializing KV store connection")
```
amalykhi: The database connection is opened before restore starts, held open while services are stopped, and
   only used after database restore completes (in PrepareDevices). This is problematic - Connection may be invalid or block shut down. May be move DB initialization to after database restore?

internal/restore/deployer.go
```diff
@@ -0,0 +1,445 @@
...
+		return fmt.Errorf("failed to open database dump %q: %w", dumpPath, err)
+	}
+	defer dumpFile.Close()
+
+	p.log.Infof("Starting database restore from %s into container %s", dumpPath, p.containerName)
+
+	// Pass the password as the first stdin line so it never appears in argv.
+	// The shell reads it with `read -r PGPASSWORD`; the remaining stdin (the SQL
+	// dump) is then consumed by `psql -f -`. Mirrors the backup pattern.
+	psqlCmd := fmt.Sprintf("IFS= read -r PGPASSWORD; export PGPASSWORD; psql -h 127.0.0.1 -p %s -U %s -d %s -f -",
```
amalykhi: The database restore runs without --single-transaction flag. If the SQL dump is large and fails
  mid-way (OOM, context cancellation, network issue), the database is left in a partially restored state
  with no rollback.

internal/restore/deployer.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+	defer dumpFile.Close()
+
+	k.log.Infof("Starting database restore from %s into pod %s", dumpPath, podName)
+
+	// Same stdin-injection pattern as the Podman deployer: password on the first
+	// line, SQL dump on the rest. Keeps the password out of the command argv.
+	command := []string{
+		"sh", "-c",
+		fmt.Sprintf("IFS= read -r PGPASSWORD; export PGPASSWORD; psql -h 127.0.0.1 -p %s -U %s -d %s -f -",
```
amalykhi: The database restore runs without --single-transaction flag. If the SQL dump is large and fails
  mid-way (OOM, context cancellation, network issue), the database is left in a partially restored state
  with no rollback.

internal/restore/deployer.go
```diff
@@ -0,0 +1,445 @@
...
+				return fmt.Errorf("context cancelled waiting for deployment %q to scale down: %w", name, err)
+			}
+			if time.Now().After(deadline) {
+				return fmt.Errorf("timed out after %s waiting for deployment %q to reach zero replicas", stopWaitTimeout, name)
+			}
+			current, err := clientset.AppsV1().Deployments(k.namespace).Get(ctx, name, metav1.GetOptions{})
+			if err != nil {
+				return fmt.Errorf("failed to poll deployment %q status: %w", name, err)
+			}
+			if current.Status.Replicas == 0 {
```
amalykhi, line 702: No Validation That Pods Are Actually Terminated. pods might still be in
  Terminating state with graceful shutdown while Replicas==0

internal/restore/restore.go
asafbennatan: i have considered this while implementing , but this does not worth the complexity overhead it introduces

internal/restore/deployer.go
asafbennatan: Deployment.Status.Replicas counts all pods with matching labels that have not yet been deleted from the API server. A pod in Terminating state still has its labels and is still counted in Status.Replicas until the pod object is fully removed from etcd. So Status.Replicas == 0 genuinely means no pods remain, including any that were gracefully shutting down. The existing polling loop is correct.

cmd/flightctl-restore/main.go
asafbennatan: gorm has a connection pool that handles this - we dont need to do that , if we did try to do that that would split the restore function across different functions (as now it has to also initialize db connection instead of getting one ) 
or include a complex db init function input with cleanup logic 

the current approach is good balance IMO

internal/restore/deployer.go
asafbennatan: fixed

internal/restore/deployer.go
asafbennatan: fixed

internal/restore/restore.go
amalykhi: ACK

amalykhi (approved):

amalykhi (approved):

amalykhi (approved):

PR #3017, opened 2026-05-28
EDM-2708: Add agent VM PR check

eldar101:
Adds a PR smoke workflow that exercises `make deploy agent-vm` on the VM-capable CI runner. The check sets up KVM/libvirt, deploys the backend, boots the agent VM, and verifies the VM reaches the `running` state.

### DIFF
--- a/.github/workflows/pr-agent-vm-testing.yaml
+++ b/.github/workflows/pr-agent-vm-testing.yaml
@@ -1,0 +1,201 @@
+name: "Agent VM smoke test"
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - 'release-*'
+  pull_request:
+  merge_group:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  # Enable BuildKit for cache mounts
+  DOCKER_BUILDKIT: 1
+  BUILDKIT_PROGRESS: plain
+
+  QUAY_ORG: quay.io/flightctl
+  REGISTRY: quay.io
+  REGISTRY_OWNER: flightctl
+  REGISTRY_OWNER_TESTS: flightctl-tests
+  GITHUB_ACTIONS: true
+
+jobs:
+  agent-vm:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-agent-images-from-rpms.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v46
```
asafbennatan, line 29: lets use a proper release version , or avoid the changte entirely if its unnecessery

asafbennatan (commented):
are we not already checking that with our e2e - why do we need another workflow for that ?

eldar101:
> are we not already checking that with our e2e - why do we need another workflow for that ?

The existing E2E workflow does boot VMs, but it does not run the actual `make deploy agent-vm` command. This workflow is intended to cover that specific developer path.

.github/workflows/build-agent-images-from-rpms.yaml
eldar101: Done thanks.

.github/workflows/pr-agent-vm-testing.yaml
```diff
@@ -0,0 +1,147 @@
...
+        with:
+          path: |
+            bin/dnf-cache
+            bin/osbuild-cache
+          key: ${{ runner.os }}-bootc-cache-agent-vm-${{ hashFiles('test/scripts/agent-images/containerfiles/*/Containerfile') }}
+          restore-keys: |
+            ${{ runner.os }}-bootc-cache-agent-vm-
+            ${{ runner.os }}-bootc-cache-e2e-
+
+      # ========== DEPLOY BACKEND AND AGENT VM ==========
```
Samudelacruz77, line 104: No timeout on the job. If the VM hangs or make deploy stalls, the workflow will run until GitHub's default 6-hour timeout

asafbennatan:
> > are we not already checking that with our e2e - why do we need another workflow for that ?
> 
> The existing E2E workflow does boot VMs, but it does not run the actual `make deploy agent-vm` command. This workflow is intended to cover that specific developer path.

workflows should check user flows not developer ones - i dont mind we check this as long as this does not run automatically or delays the other PR checks

Samudelacruz77 (approved):
LGTM

dudyas6:
LGTM

Samudelacruz77 (approved):
LGTM

PR #3022, opened 2026-05-29
EDM-4045: E2E Vulnerability test 89034 fix

amalykhi:


test/e2e/vulnerability/vulnerability_suite_test.go
```diff
@@ -321,6 +321,11 @@ func configurePeriodicServiceForTrustify(ctx context.Context, trustify *auxiliar
 			return fmt.Errorf("API service not ready after restart: %w", err)
 		}
 
+		// Give API service time to fully initialize connections after restart
+		// This prevents EOF errors on the first API request after restart in ACM/OpenShift environments
+		logrus.Info("Waiting for API service to stabilize connections...")
```
asafbennatan, line 326: we shouldnt need to do this - the lifecycle providers should already check that the pods are ready

amalykhi: there was connection failure in one test run. I see in others it is succeeded. May be some connection general issue, will monitor

PR #3026, opened 2026-05-31
EDM-3936: fix image build failing with RLIMIT_NOFILE operation not permitted

asafbennatan:
## Problem

Image builds fail during the `dnf install flightctl-agent` step with:

```
Error: building at STEP "RUN … dnf install -y … flightctl-agent …":
setting "RLIMIT_NOFILE" limit to soft=1048576,hard=1048576
(was soft=524288,hard=524288): operation not permitted
```

This occurs on OpenShift clusters where the default `DefaultLimitNOFILE` is 524288:524288.

## Root Cause

The ImageBuilder Worker starts a nested Podman worker container (`startPodmanWorker()`) without setting any `--ulimit` flag. The worker container inherits the host's hard limit (524288 on default OpenShift). The inner `podman build` (buildah) then tries to raise `RLIMIT_NOFILE` to 1048576 but cannot, because its hard limit is 524288 and it lacks `CAP_SYS_RESOURCE`.

## Fix

Add `--ulimit nofile=1048576:1048576` to the outer worker container start args in `startPodmanWorker()`. The imagebuilder worker always runs as a privileged container (`CAP_SYS_RESOURCE`), which allows it to set limits above the inherited host value. Once the outer container has a hard limit of 1048576, the inner buildah can raise to that value without needing its own special capabilities.

Also add a troubleshooting section to `docs/user/installing/configuring-imagebuilder.md` documenting the root cause and the OpenShift MachineConfig fallback workaround for environments with further restrictions.

## Testing

- `go build ./internal/imagebuilder_worker/...` — clean
- `make lint-docs` — 0 errors
- `make spellcheck-docs` — 56 files free from spelling errors

## Confidence

Medium-High — the privileged pod `CAP_SYS_RESOURCE` analysis is based on kernel semantics (`man 2 setrlimit`) and the Helm chart / Quadlet configuration; the fix cannot be exercised in a unit test without a real nested-Podman environment.

## Rollback

Revert the `--ulimit` line from `startPodmanWorker()`. No data migration needed.

## Risk Assessment

Low — the change adds a single flag to the outer `podman run` command and is scoped to the imagebuilder worker path only.

### DIFF
--- a/docs/user/installing/configuring-imagebuilder.md
+++ b/docs/user/installing/configuring-imagebuilder.md
@@ -333,3 +333,61 @@ To use a custom CA for the registry that serves the Podman or bootc-image-builde
 - Expired entitlement certificates
 - Incorrect secret format (certificates must be PEM encoded)
 - Missing certificate key file
+
+### Build Fails with "setting RLIMIT\_NOFILE limit … operation not permitted"
+
+**Problem**: Image build fails with an error similar to:
+
+```text
+Error: building at STEP "RUN … dnf install -y … flightctl-agent …":
+setting "RLIMIT_NOFILE" limit to soft=1048576,hard=1048576
+(was soft=524288,hard=524288): operation not permitted
+```
+
+**Cause**: The build container (buildah) tries to raise the open-file-descriptor limit to
+1048576. This fails when the host's hard limit is lower (for example, 524288, which is the
+OpenShift default) and the container runtime does not grant `CAP_SYS_RESOURCE` to raise it.
+
+**Solution**: The ImageBuilder Worker sets `--ulimit nofile=1048576:1048576` on the nested
+podman worker container, relying on the privileged pod's `CAP_SYS_RESOURCE` capability.
+If you still encounter this error, the host itself must be configured to allow the higher limit.
+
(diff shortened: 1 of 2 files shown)


docs/user/installing/configuring-imagebuilder.md
```diff
@@ -605,3 +333,61 @@ reference ready to use in a Fleet template.
...
+          path: /etc/systemd/system.conf.d/60-nofile.conf
+```
+
+This sets `DefaultLimitNOFILE=1048576:1048576` in systemd, which raises the limit for all
+processes on the node, including container runtimes.
+
+On a Podman Quadlet (Linux host), add the following file instead:
+
+```ini
+# /etc/systemd/system.conf.d/60-nofile.conf
```
itzikb-redhat, line 383: Can we limit it to only the worker service?

asafbennatan: no i dont think so , this issue tackles a situation where the host config is not enough (not the specific pods)

itzikb-redhat: But specifically the issue is with the ib worker

itzikb-redhat (approved):

PR #3030, opened 2026-06-01
imagebuild: auto-detect disconnected and fall back to OCP mirror registry

dudyas6:
## Summary
- ImageBuild E2E tests hardcoded `quay.io` as the source registry, which fails in disconnected environments where the imagebuilder-worker cannot reach external registries
- Add `resolveSourceRegistry()` that probes quay.io accessibility via `ConditionTypeRepositoryAccessible` (same pattern as `agent_update_test.go`), and falls back to the OCP mirror registry (discovered via `ImageTagMirrorSet`) when unreachable
- The check runs once via `sync.Once` and caches the result for all tests
- When using a local mirror, source repositories are created with `skipTlsVerify: true` for self-signed certs
- `E2E_SOURCE_REGISTRY` env var available as manual override

## Test plan
- [ ] Connected mode: tests use quay.io as before (no behavior change)
- [ ] Disconnected mode: tests auto-detect unreachable quay.io, switch to OCP mirror
- [ ] E2E_SOURCE_REGISTRY override works when set explicitly


dudyas6:
Recreating with clean branch (no unrelated EDM-3953 commits)

PR #3062, opened 2026-06-07
[EDM-4075]: add maxLength:63 to catalogItemName in imagebuilder API

asafbennatan:
## Problem

The `catalogItemName` field in the imagebuilder API (`api/imagebuilder/v1alpha1/openapi.yaml`) had `minLength: 1` but **no `maxLength`**. A user could create an `ImagePromotion` with a name longer than 63 characters. The resulting catalog item would be accepted at every step (promotion creation, completion, catalog item creation) but would be **permanently undeployable**.

At fleet deploy time the UI sets:
```
metadata.labels["os.catalog.flightctl.io/item"] = <catalogItemName>
```
The Fleet API enforces `maxLength: 63` on all label values per Kubernetes convention, so the deployment fails with a cryptic `400 Bad Request`:
```
Error at "/metadata/labels/os.catalog.flightctl.io/item": maximum string length is 63
```
The user sees "Failed to deploy" with no indication that the catalog item name is the root cause. Recovery requires deleting and recreating the promotion, catalog item, and all associated exports.

## Root Cause

Missing `maxLength: 63` on `catalogItemName` in `ImagePromotionTargetBase`, `NewCatalogItemTarget`, and `ExistingCatalogItemTarget` in the imagebuilder OpenAPI spec.

## Fix

Add `maxLength: 63` to `catalogItemName` in all three schema locations in `api/imagebuilder/v1alpha1/openapi.yaml`. Regenerated `spec.gen.go` via `make generate` so the constraint is embedded in the spec blob used by the `oapimiddleware` request validator at runtime.

The middleware now rejects promotion creation requests with a name longer than 63 characters at the HTTP boundary, before any handler code runs, with a clear schema validation error.

## Testing

- `make generate` — clean
- `make lint-openapi` — no warnings or errors
- `go build ./api/imagebuilder/...` — clean
- `go build ./internal/imagebuilder_api/...` — clean

## Confidence

High — one-line schema change per affected schema, minimal blast radius, enforced by the existing request validation middleware with no handler code changes required.

## Rollback

Revert this PR. The change only tightens input validation; no data migration or state changes are involved.

## Risk Assessment

Low — only affects requests that create `ImagePromotion` resources with `catalogItemName` longer than 63 characters. Such requests were silently accepted before and will now be rejected at creation time with a clear error. No existing valid data is affected.

## Related

- EDM-4074 — UI-side consequence: long names cause layout breakage in catalog cards (separate fix needed in the frontend).


celdrake:
@asafbennatan Since the current implementation based on labels will soon change, should we bring this restriction permanently to the system?
For other items, we keep length to 253 characters, and currently it's also the same limit that the UI will accept for catalog item names.

IMO this bug could be added to EDMRFE-114 as  a "can't fix".

asafbennatan:
@celdrake ok

PR #3063, opened 2026-06-07
EDM-4069: Fail ImagePromotion when required exports reach terminal non-completed state

asafbennatan:
## Summary

- `ImagePromotion` was stuck in `WaitingForArtifacts` indefinitely when required `ImageExport`s reached a terminal non-completed state (`Failed` or `Canceled`), with no indication to the user of why it was blocked.
- Added `ListTerminalNonCompletedForBuild` to `ImageExportService` that queries for `Failed` or `Canceled` exports for a given build ref and format (two sequential field-selector queries, one per reason).
- In `computeArtifactsStatus`, when no completed export exists for a format, the new check is made and affected formats collected in `terminalFailedFormats`.
- In `evaluateAndTransition`, if `terminalFailedFormats` is non-empty, the promotion transitions to `Failed` immediately with a message naming the affected format(s) and the build ref (e.g. _"Export(s) for format(s) [iso, vmdk] failed or were canceled for build "2222". Please create new exports to retry."_).

## Test plan

- [ ] `TestEvaluator_TerminalExportFailsPromotion` — 5 new sub-tests covering: single canceled export, single failed export, multiple terminal formats, one terminal + one pending (fails immediately for the terminal one), completed export (regression guard — no false positives)
- [ ] All existing `imagebuilder_worker/tasks` and `imagebuilder_api/service` tests continue to pass

Fixes: EDM-4069

### DIFF
--- a/internal/imagebuilder_api/service/imageexport.go
+++ b/internal/imagebuilder_api/service/imageexport.go
@@ -72,6 +72,9 @@ type ImageExportService interface {
 	// ListCompletedForBuild returns the most recently completed ImageExport for the given
 	// imageBuildRef and export format, or nil if none exists yet.
 	ListCompletedForBuild(ctx context.Context, orgId uuid.UUID, imageBuildRef string, format domain.ExportFormatType) (*domain.ImageExport, error)
+	// ListTerminalNonCompletedForBuild returns a terminal non-completed ImageExport (Failed or
+	// Canceled) for the given imageBuildRef and export format, or nil if none exists.
+	ListTerminalNonCompletedForBuild(ctx context.Context, orgId uuid.UUID, imageBuildRef string, format domain.ExportFormatType) (*domain.ImageExport, error)
 }
 
 // ImageExportDownload contains information for downloading an ImageExport artifact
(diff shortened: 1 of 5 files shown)


rawagner:
Is this really the behavior we want ?

The current implementation follows what was agreed in the design doc
```
WaitingForArtifacts — the ImageBuild has not yet reached Completed state, or at least one format in exportFormats with published: false does not yet have a completed ImageExport. Individual format readiness is tracked in status.artifactStatuses. A failed ImageExport does not transition the publication to Failed — the user can create a new ImageExport of the same format and the publication will pick it up automatically. PATCH may be applied in this state to add new formats; the publication stays in WaitingForArtifacts until all formats (including newly added ones) are satisfied. Also re-entered from Completed when new formats are added via PATCH and no completed export exists yet for those formats.
```

ImagePromotion is not referencing any exact image export, Only the image export type. When image export fails/gets cancelled, it should not be a reason to fail the promotion as well. The status of the artifacts is exposed in the ImagePromotion state - UI just doesnt show it yet.

asafbennatan:
it was decided not to implement this

PR #3067, opened 2026-06-08
EDM-3697: fix copy of flightctl-backup in E2E tests and start services at flightctl-restore

amalykhi:
fixes expose port logic for flightctl-restore - for quadlets


internal/restore/deployer.go
```diff
@@ -130,6 +130,9 @@ type Deployer interface {
 	// StartServices starts the FlightCtl services previously stopped by StopServices.
 	// Must be called after all restore steps complete, whether they succeed or fail.
 	StartServices(ctx context.Context) error
+	// StartInfraServices ensures database and KV services are running.
+	// Must be called before ExposeService to ensure infrastructure is accessible.
+	StartInfraServices(ctx context.Context) error
```
asafbennatan: why do we need that - we already have start services and that is already called by defer function in restore , the problem is not a missing start services problem

amalykhi: Good question! The distinction is important:

**`StartServices()` vs `StartInfraServices()`** operate on different service sets:

1. **`StopServices()` / `StartServices()`** (lines 305-314) operate on **application services** via the ServiceHandler:
   - `flightctl-api`
   - `flightctl-worker`
   - `flightctl-periodic`
   - imagebuilder services
   - alert services
   
   These services **intentionally exclude** infrastructure (see line 85 comment: "Units that do not connect to the DB (e.g. flightctl-db, flightctl-kv, flightctl-ui, flightctl-cli-artifacts) are intentionally omitted.")

2. **`StartInfraServices()`** (lines 317-353) operates on **infrastructure services**:
   - `flightctl-db.service`
   - `flightctl-kv.service`

**The problem this fixes:**

After `RestoreConfig()` (line 112), the systemd configuration is restored. This can cause systemd to reload or restart services, potentially stopping the database. The code then attempts to:
- Call `ExposeService()` for the database (line 129)
- Connect to the database (line 148)

Without `StartInfraServices()`, these operations fail with "connection refused" because the database service may not be running or ready to accept connections.

The defer at line 79 calls `StartServices()` which only starts the **application** services, not the infrastructure. We need infrastructure running **before** we try to connect at line 148.

**Flow:**
```
StopServices()          → stops app services (DB/KV intentionally left running)
RestoreConfig()         → may affect DB/KV service state
StartInfraServices()    → explicitly ensures DB/KV are running + ready (NEW)
ExposeService()         → needs DB/KV to be running
InitDB()                → needs DB to accept connections
```

The bug was discovered in E2E tests where restore consistently failed with "connection refused" errors. See the bug report: `BUG_REPORT_RESTORE.md` in the repo root.

asafbennatan: but these services are not stopped by the stop command

amalykhi: You're absolutely right - that's a critical design flaw. The restore command should only manage services it explicitly stopped.

I need to reconsider the root cause. Looking at the code:

1. `RestoreDatabase()` (line 376-460) already executes `podman exec` commands against the database container, which proves the DB **is** running during restore
2. `StopServices()` intentionally excludes DB/KV (line 85-95)
3. Therefore, adding `StartInfraServices()` to start services we never stopped violates the principle

**The real question is: what's actually causing the "connection refused" errors in the E2E tests?**

Let me investigate the actual E2E test environment. The test runs restore **inside a VM** (see `test/harness/e2e/harness_backup_restore.go:254`), where systemd manages the database container. 

**Hypothesis:** The issue might be specific to the E2E test infrastructure, not the restore logic itself. In the VM environment:
- Database might be accessible via container exec (used by `RestoreDatabase`)
- But not accessible via TCP localhost:5432 (used by `InitDB` at line 148)

If that's the case, the fix belongs in the **E2E test infrastructure**, not in the restore command itself.

Should I:
1. Revert the `StartInfraServices()` changes
2. Investigate why TCP connectivity fails in the VM environment
3. Fix the actual root cause (which might be test infrastructure configuration)

amalykhi: I've reverted the `StartInfraServices()` changes in commit `f4db34ed`. You were correct - restore should not manage services it didn't stop.

**Current state (EDM-3697 branch):**
- ✅ Binary transfer fix (stdin piping to avoid argument length limits) - **KEPT**
- ✅ Checksum file transfer (.sha256) - **KEPT**
- ✅ Security fixes (SQL injection + sensitive data exposure) - **KEPT**
- ❌ `StartInfraServices()` interface and implementations - **REMOVED**

**What's left:**
Only the actual bugs are fixed:
1. Binary/archive transfers now use stdin instead of command arguments (fixes "argument list too long")
2. SQL injection vulnerability fixed (database name escaping)
3. Sensitive data exposure fixed (systemctl output removed from errors)

The infrastructure services (DB/KV) remain untouched by the restore command, as they should be.

I'll let the E2E tests determine if there are any remaining connectivity issues. If the binary transfer was the root cause (tests failing before restore could properly execute), then this should be sufficient. If there's still a "connection refused" error, we'll need to investigate the actual network configuration in the test VM environment.

amalykhi: ## Root Cause of "Connection Refused" Error

I've identified and fixed the actual root cause in commit `c2bc23de`.

### The Problem

**Restore command runs on the HOST, not in a container.**

In Quadlet deployments:
1. Database container (`flightctl-db`) runs on the `flightctl.network` 
2. Application containers (API, worker) connect via container network using hostname `flightctl-db:5432`
3. The database container **did not publish ports to the host** (no `PublishPort` directive)
4. Restore command runs on the VM host (see `test/harness/e2e/harness_backup_restore.go`)
5. When restore calls `ExposeService("flightctl-db")`, it returns `cfg.Database.Hostname` = `"flightctl-db"`
6. Host tries to connect to hostname `flightctl-db:5432` which:
   - ❌ Doesn't resolve (container name only works inside the network)
   - ❌ Port not published to host anyway
7. Result: **"connection refused"**

### Why RestoreDatabase() Worked

`RestoreDatabase()` (line 376-460) uses `podman exec` to run `psql` commands **inside the database container**, so it never needed host network access:

```go
exec.CommandContext(ctx, p.containerCLI, "exec", p.containerName, 
    "psql", "-U", "postgres", "-d", dbName, "-c", sql)
```

But `InitDB()` (line 143) tries to connect via TCP from the **host**, which fails.

### The Fix

Added `PublishPort=127.0.0.1:5432:5432` to:
- `deploy/podman/flightctl-db/flightctl-db.container`
- `deploy/podman/flightctl-kv/flightctl-kv.container`

This:
- ✅ Allows host-based tools (restore, CLI) to connect via `localhost:5432`
- ✅ Maintains security (only bound to 127.0.0.1, not externally accessible)
- ✅ Matches the standalone deployment configuration

### Why StartInfraServices Was Wrong

The original `StartInfraServices()` attempted to **start** the database service with `systemctl start`, but:
1. The database was never stopped (intentionally excluded from `StopServices`)
2. The database was already running (proven by `RestoreDatabase` working)
3. Starting wouldn't fix the connectivity issue (problem was missing published ports, not service state)

The real issue was **network accessibility from the host**, not **service availability**.

amalykhi: Given your preference for minimal changes, I've kept only the essential fixes:

**Current state (commit 5f009805):**
- ✅ Binary transfer fix (stdin piping)
- ✅ Security fixes (SQL injection + sensitive data exposure)
- ❌ No deployment changes
- ❌ No product behavior changes

**About the "connection refused" root cause:**

I identified that it occurs because restore runs on the **host** but containers don't publish ports - the host can't reach `flightctl-db:5432` (container network hostname).

**Fixing it would require ONE of:**
1. **Product change**: Modify `ExposeService()` to resolve container IPs
2. **Deployment change**: Add `PublishPort` to DB/KV container files  
3. **E2E change**: Run restore inside a container on the network (not on host)

**However** - the binary transfer issue may have been masking this. With binaries now transferring correctly, restore might work. Let's let the E2E tests determine if additional fixes are needed.

If tests still fail with connection refused, I recommend option #1 (product change to ExposeService) as it's the most general solution for host-based restore operations.

amalykhi: @asafbennatan
 Added PublishPort=127.0.0.1:5432:5432 to flightctl-db.container and PublishPort=127.0.0.1:6379:6379 to flightctl-kv.container.
 please review again

asafbennatan (approved):
LGTM , 
thanks for looking into it

PR #3069, opened 2026-06-08
EDM-4120: Unix socket for standalone agent API

keitwb:
This will bypass any concerns about ephimeral ports that could be problematic at scale. We can do the same thing for other services as necessary.

Also increase the worker connection limits.

### DIFF
--- a/deploy/podman/flightctl-api/flightctl-api-config/config.yaml.template
+++ b/deploy/podman/flightctl-api/flightctl-api-config/config.yaml.template
@@ -22,7 +22,7 @@ database:
 {{- end}}
 service:
   address: flightctl-api:3080
-  agentEndpointAddress: flightctl-api:7443
+  agentEndpointAddress: unix:///var/run/flightctl-listeners/api-agent.sock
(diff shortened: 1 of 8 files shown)


deploy/podman/flightctl-listeners.volume
kkyrazis, line 1: does this need to be added to to spec?

keitwb: Yes, added.

kkyrazis (approved):

deploy/podman/flightctl-api/flightctl-api-config/config.yaml.template
```diff
@@ -22,7 +22,7 @@ database:
 {{- end}}
 service:
   address: flightctl-api:3080
-  agentEndpointAddress: flightctl-api:7443
+  agentEndpointAddress: unix:///var/run/flightctl-listeners/api-agent.sock
```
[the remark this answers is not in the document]
keitwb: This cannot be correct since the previous value was also unusable by the agents. Check again.

kkyrazis (approved):

PR #3072, opened 2026-06-09
EDM-4125: fixed device taking stale data on dependency change

asafbennatan:
fixes stale cache data on template revisioned dependency refs


keitwb (approved):
Seems generally ok.

PR #3083, opened 2026-06-10
NO-ISSUE: Fix authprovider e2e setup

eldar101:
Fixes the authprovider e2e setup so tests don’t lose admin auth while managing providers. It also makes the PAM/OpenShift/AAP browser login flows pick the intended provider explicitly and hardens the Cypress helper so it won’t patch the wrong OAuthClient.

The tests were failing on the RHEL9 test-vm because they assumed the environment behaves like the Keycloak/OCP setup. On quadlet/test-vm, the default auth provider is PAM, so the OAuth2 test now uses the existing PAM issuer instead of trying to force the Keycloak OAuth2 flow.
The steps we now do:

restores admin login before creating/deleting authproviders
waits for dynamic providers to really appear/disappear inlogin --show-providers
avoids picking the wrong OpenShift OAuthClient when more than one exists
validates the callback port before patching JSON
skips only the browser-login Cypress test when Cypress/npm are missing on the test host

### DIFF
--- a/test/e2e/authprovider/authprovider_suite_test.go
+++ b/test/e2e/authprovider/authprovider_suite_test.go
@@ -59,7 +59,7 @@ var _ = BeforeSuite(func() {
 	authProviderYAML := buildOIDCAuthProviderYAML(
 		keycloakAuthProviderName,
 		auxSvcs.Keycloak.IssuerURL(),
-		"flightctl-client",
+		keycloakClientID,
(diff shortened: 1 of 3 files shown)


eldar101:
@itzikb-redhat I already showed on your host that it works, kindly approve.

itzikb-redhat (approved):

PR #3096, opened 2026-06-11
EDM-3953: fix --tag-override not pinning RPM version, causing bundle mismatch

aekubacki:
## Summary

When `--tag-override 1.2.0-rc3` is set and `--rpm-repo-url` points to a repo containing multiple versions (e.g. a COPR repo with both `1.2.0~rc3` and `1.3.0~main`), dnf downloads the **latest** RPM regardless of the image tag override. The RPM's baked-in quadlet files reference image tags from the RPM version (`1.3.0-main`), while the bundled images are tagged `1.2.0-rc3`. All services fail to start on the target machine.

## Fix

When `--tag-override` is set and `--rpm-packages` was not explicitly overridden by the user, automatically pin flightctl RPM packages to the corresponding version. Image tags use hyphens for pre-release suffixes (`1.2.0-rc3`) while RPM versions use tildes (`1.2.0~rc3`); the tool converts automatically.

Also adds `--rpm-version` for cases where the RPM version differs from the image tag.

## Behaviour

| Command | Result |
|---------|--------|
| `--tag-override 1.2.0-rc3` | RPMs pinned to `1.2.0~rc3` automatically |
| `--tag-override 1.2.0-rc3 --rpm-version 1.2.0~rc2` | RPMs pinned to `1.2.0~rc2` (explicit override wins) |
| `--tag-override 1.2.0-rc3 --rpm-packages flightctl-services-1.2.0~rc3` | User-specified packages respected, no auto-pin |
| `--tag-override 1.2.0` | RPMs pinned to `1.2.0` |

## Test plan

- [ ] Run with `--tag-override 1.2.0-rc3 --rpm-repo-url <copr-with-multiple-versions>` and verify downloaded RPM is `1.2.0~rc3` not the latest
- [ ] Verify `1.2.0-rc3` → `1.2.0~rc3` tilde conversion
- [ ] Verify `--rpm-version` explicit override works
- [ ] Verify `--rpm-packages` explicit override skips auto-pin

### DIFF
--- a/scripts/air-gap/mirror-images/main.go
+++ b/scripts/air-gap/mirror-images/main.go
@@ -173,22 +173,15 @@ func pinRPMPackages(packages []string, version string) []string {
 	return pinned
 }
 
-// resolveRPMVersion pins flightctl RPM packages to a specific version when
-// bundling RPMs so the downloaded RPM version matches the bundled image tags.
-// --rpm-version takes precedence; when absent and --tag-override is set and the
-// caller did not explicitly set --rpm-packages, the version is derived from the
-// tag override (converting image-tag hyphens to RPM-version tildes).
-func resolveRPMVersion(packages []string, rpmVersion, tagOverride string, shouldPin, packagesExplicitlySet bool) []string {
-	if !shouldPin {
+// resolveRPMVersion pins flightctl RPM packages to the version derived from
+// tagOverride when bundling RPMs, so the downloaded RPM version matches the
+// bundled image tags. Only applies when tagOverride is set and the caller did
+// not explicitly set --rpm-packages.
+func resolveRPMVersion(packages []string, tagOverride string, shouldPin, packagesExplicitlySet bool) []string {
+	if !shouldPin || tagOverride == "" || packagesExplicitlySet {
 		return packages
 	}
-	effective := rpmVersion
-	if effective == "" && tagOverride != "" && !packagesExplicitlySet {
-		effective = imageTagToRPMVersion(tagOverride)
-	}
-	if effective == "" {
-		return packages
-	}
+	effective := imageTagToRPMVersion(tagOverride)
 	pinned := pinRPMPackages(packages, effective)
 	logInfo("  RPM version pin:  %s", effective)
 	return pinned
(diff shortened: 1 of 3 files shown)


scripts/air-gap/mirror-images/main.go
```diff
@@ -140,6 +140,60 @@ func validateFlags(variant, bundle string, execute, bundleRPMs, rpmReposync, rpm
...
+		if tag[i] == '-' {
+			return tag[:i] + "~" + tag[i+1:]
+		}
+	}
+	return tag
+}
+
+// pinRPMPackages appends the given version to each flightctl package name so
+// that dnf downloads exactly that version rather than the latest available.
+// Third-party packages (those that do not start with "flightctl-") are left
```
kkyrazis, line 162: just curious what the usecase is for third party packages? Are we bundling rpms that aren't flightctl-*?

aekubacki: Here is the current break down: 

Standard bundle (--bundle-rpms): Only flightctl-* packages by default — flightctl-services, flightctl-cli, flightctl-agent. No third-party  RPMs.

  Agent-only bundle (--agent-only): This is where the third-party packages come in. The defaults are:
  - open-vm-tools — VMware guest tools, needed on VMs
  - ignition — provisioning tool for edge device first-boot config
  - afterburn — fetches instance metadata on first boot (used with ignition)
  - cloud-init — cloud provisioning, needed on some edge platforms

  Plus skopeo is silently added to the download list (but not install) whenever --bundle-rpms is used — it's needed on the air-gapped target  to bootstrap the local registry.

  The pinRPMPackages function deliberately skips non-flightctl-* packages precisely because these third-party tools (open-vm-tools,  ignition, etc.) have their own independent versioning that has nothing to do with the flightctl release version. Only the flightctl-*  packages need to be pinned to match the image tags.

kkyrazis: Thanks claude

kkyrazis (commented):
> Also adds --rpm-version for cases where the RPM version differs from the image tag.

I'm curious why this is a necessary case? If I --rpm-version 1.2.0 and --tag-override 1.1.0, it'll pull the 1.2.0 rpms which inherently reference the 1.2.0 images right?

aekubacki:
> > Also adds --rpm-version for cases where the RPM version differs from the image tag.
> 
> I'm curious why this is a necessary case? If I --rpm-version 1.2.0 and --tag-override 1.1.0, it'll pull the 1.2.0 rpms which inherently reference the 1.2.0 images right?

Removed the new tag so this question of if is no longer necessary.

kkyrazis (approved):

PR #3097, opened 2026-06-11
EDM-3953: update agent offline install doc to use flightctl-mirror-images tool

aekubacki:
## Summary

Doc-only change. Rewrites `docs/user/installing/installing-agent-offline.md` to use the `flightctl-mirror-images --agent-only` workflow as the primary installation approach, replacing the manual `dnf download` procedure.

Key changes:
- Primary path now uses `flightctl-mirror-images --agent-only --rpm-createrepo --tag-override <version>` which handles version pinning, repo configuration, and generates `install-rpms.sh` automatically
- Added COPR repo example for testing versions from the COPR repository
- Kept Part 2 (container images for managed workloads) unchanged
- Removed manual `dnf config-manager` / `dnf download` steps

## Test plan

- [ ] Verify the `--agent-only` command produces a working bundle on RHEL 9

### DIFF
--- a/docs/user/installing/installing-agent-offline.md
+++ b/docs/user/installing/installing-agent-offline.md
@@ -1,9 +1,9 @@
 # Installing the Flight Control agent offline on RHEL
 
 This document describes how to install `flightctl-agent` and `flightctl-cli` on a RHEL
-machine that has no internet access. The procedure uses a connected prep machine to
-download the required packages, transfers them using portable media, and installs them
-locally on the target.
+machine that has no internet access. The `flightctl-mirror-images` tool running on a
+connected prep machine creates a portable RPM bundle. You then transfer the bundle to
+the target device and run the included install script.
 
 The agent itself does not require container images to run. However, if the agent manages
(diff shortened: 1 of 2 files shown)


docs/user/installing/installing-agent-offline.md
```diff
@@ -1,112 +1,99 @@
...
 ```
 
-> [!NOTE]
-> If you need a full offline repo that lets the target machine resolve packages with
-> `dnf install` instead of installing from flat `.rpm` files, use `dnf reposync`
-> instead. See [Setting up a local RPM repository](offline-rpm-repository.md) for both
-> approaches and when to choose each.
-
-Verify that the downloaded files include the expected packages:
+To use a version from the COPR repository:
```
kkyrazis, line 52: what about rpm.flightctl.io, we tend to prefer that over copr for our upstream artifacts.

aekubacki: Yeah we default to that one, but this gives us an option to pull from copr for dev builds.

kkyrazis (approved):

PR #3102, opened 2026-06-12
EDM-3953: fix install-rpms.sh to handle RHEL minor version mismatches between prep and target

aekubacki:
## Summary

When a bundle is built on a prep machine running a newer RHEL minor version than the target (e.g. bundle built on RHEL 9.8, installed on RHEL 9.5), `dnf download --alldeps` pulls system packages at the prep machine's version. If the target has tightly-pinned packages (e.g. `python3-librepo` pins `librepo` to an exact version), the install fails with a dependency conflict.

**Error seen:**
```
cannot install both librepo-1.19.0-1.el9.x86_64 from flightctl-local
and librepo-1.14.5-2.el9.x86_64 from @System
```

**Fix:** Add `--nobest` to all `dnf install` calls in `install-rpms.sh`. This tells dnf it doesn't need to install the absolute newest version of system packages — if an older version already on the target satisfies the requirements, dnf will use it. Bundles built on newer RHEL minor versions now install cleanly on older targets.

## Test plan

- [ ] Build bundle on RHEL 9.8, install on RHEL 9.5 — verify no librepo conflict
- [ ] Verify flightctl packages still install at the correct pinned version


aekubacki:
I don't think this is a show stopper but came up in David's testing when the prep/staging machine RHEL version differed from the target machine.  Should be an option for the future.

keitwb (approved):

PR #3109, opened 2026-06-15
EDM-4106: agent serial console handler

asafbennatan:
## Summary

- Adds `applications/console` package: `Manager` (session orchestrator), `Session` interface, `ExecStreamer` interface, `vmSerialSession` (Unix-socket-to-gRPC bridge), `AppConsoleResolver`/`ResolverFunc` interfaces
- `vmSerialSession` owns `vmSerialSocketPath` and dials via `ExecStreamer` (production: `*client.Podman`; tests: `mockExecStreamer`)
- `PodmanMonitor` implements `resolveConsole` as a factory — creates a podman client and returns the appropriate session type; no socket-path or dial knowledge
- `applications.manager` owns `AppConsoleManager` and wires it via `WithConsole` / `RunConsole` / `resolveConsole`
- `agent.go` wires a separate gRPC client to `flightctl-remote-access` and calls `applicationsManager.WithConsole`
- Removed `VMWorkloadHelper` (`IsVMWorkload`, `ComputeContainerName`) from the public `applications.Manager` interface

## Design highlights

- One-way import: `applications/console` defines narrow interfaces (`Session`, `AppConsoleResolver`, `ExecStreamer`); `applications` imports `applications/console`, not vice-versa — no circular dependency
- `ResolverFunc` adapter (like `http.HandlerFunc`) keeps the unexported `manager.resolveConsole` method out of the public API
- Session tracking (active/inactive dedup) mirrors existing `internal/agent/device/console` patterns
- `Closed` flag checked on `StreamResponse.Recv()` to match existing device-console protocol
- `bridgeConn` calls `CloseSend()` when the conn→gRPC goroutine exits, ensuring the gRPC→conn goroutine's `Recv()` is unblocked and `wg.Wait()` does not hang

## Test plan

- [ ] Unit tests: `TestAppConsoleManager` — resolver error path, gRPC stream failure, dedup, empty annotation, end-to-end VM session
- [ ] Unit tests: `TestSessionBridgeErrorPaths` — Send failure, Write failure, gRPC EOF
- [ ] Unit tests: `TestPodmanMonitorResolveConsole` — not-found, non-VM, wrong console type, tracked workload, untracked workload
- [ ] Unit tests: `TestManagerResolveConsole` — delegation to podmanMonitor, not-found-in-any-monitor
- [ ] `make unit-test` — 4655 tests pass, 0 failures
- [ ] `make lint` — 0 issues


internal/agent/device/applications/console/session.go
```diff
@@ -0,0 +1,164 @@
...
+			if err == io.EOF || (msg != nil && msg.Closed) {
+				return
+			}
+			if err != nil {
+				s.log.Debugf("recv from gRPC stream failed: %v", err)
+				return
+			}
+			if len(msg.Payload) > 0 {
+				if _, writeErr := conn.Write(msg.Payload); writeErr != nil {
+					s.log.Debugf("write to serial socket failed: %v", writeErr)
```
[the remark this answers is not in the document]
asafbennatan: Fixed — `bridgeConn` now loops over the payload until all bytes are written.

internal/agent/device/applications/console/manager.go
```diff
@@ -0,0 +1,205 @@
...
+	deviceName string,
+	resolver AppConsoleResolver,
+	log *log.PrefixLogger,
+) *Manager {
+	return &Manager{
+		grpcClient: grpcClient,
+		deviceName: deviceName,
+		resolver:   resolver,
+		log:        log,
+	}
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added a nil guard in `Start` that logs an error and returns early, consistent with the existing nil guard for `grpcClient`.

internal/console/app_console.go
```diff
@@ -0,0 +1,266 @@
...
+		_, err = m.svc.UpdateDevice(context.WithValue(ctx, consts.InternalRequestCtxKey, true), orgId, deviceName, *device, nil)
+		if !errors.Is(err, flterrors.ErrResourceVersionConflict) {
+			break
+		}
+	}
+	if err == nil {
+		if pubErr := m.publisher.Publish(ctx, orgId, deviceName, nextRenderedVersion); pubErr != nil {
+			m.log.WithError(pubErr).Errorf("annotation for device %s persisted but rendered-version notification failed", deviceName)
+			return domain.StatusInternalServerError(pubErr.Error())
+		}
```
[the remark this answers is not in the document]
asafbennatan: Fixed — `StoreAndNotify` failure after a successful `UpdateDevice` is now logged but no longer returns an error status, preventing orphaned remote-session annotations.

deploy/podman/flightctl-remote-access/flightctl-remote-access-config/config.yaml.template
```diff
@@ -1,4 +1,114 @@
...
+    {{end}}
+auth:
+  insecureSkipTlsVerify: {{.global.auth.insecureSkipTlsVerify}}
+  {{if not .global.auth.insecureSkipTlsVerify}}caCert: /root/.flightctl/certs/auth/ca.crt{{end}}
+{{- if eq .global.auth.type "aap"}}
+  aap:
+    apiUrl: {{.global.auth.aap.apiUrl}}
+    issuer: {{.global.auth.aap.issuer}}
+    clientId: {{.global.auth.aap.clientId}}
+    clientSecret: {{if .global.auth.aap.clientSecret}}{{.global.auth.aap.clientSecret}}{{end}}
```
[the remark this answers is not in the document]
asafbennatan: Fixed — all three `clientSecret` fields now use `printf "%q"` to safely quote values containing YAML special characters.

test/integration/tasks/vmrender/suite_test.go
```diff
@@ -0,0 +1,145 @@
...
+	c, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ProviderType:     containers.GetProviderType(),
+		ContainerRequest: req,
+		Started:          true,
+	})
+	if err != nil {
+		return "", nil, fmt.Errorf("start kubevirt-vm-to-pod container: %w", err)
+	}
+	cleanup := func() { _ = c.Terminate(ctx) }
+
```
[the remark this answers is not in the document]
asafbennatan: Fixed — the cleanup closure now uses `context.WithTimeout(context.Background(), 30*time.Second)` for `Terminate`, independent of the `BeforeSuite` context.

internal/remote_access_server/ws_handler.go
```diff
@@ -0,0 +1,144 @@
...
+		http.Error(w,
+			fmt.Sprintf("timed out waiting for protocol for device: %s app: %s", deviceName, appName),
+			http.StatusGatewayTimeout)
+		return
+	}
+
+	upgrader := websocket.Upgrader{
+		CheckOrigin: func(r *http.Request) bool {
+			return true
+		},
```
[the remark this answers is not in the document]
asafbennatan: This matches the existing behavior on the device console WebSocket endpoint (`/ws/v1/devices/{name}/console`). Changing the origin policy is a separate concern that should be aligned across all console handlers — leaving open as a follow-up.

internal/remote_access_server/ws_handler.go
```diff
@@ -0,0 +1,144 @@
...
+		}()
+		for {
+			msgType, message, err := conn.ReadMessage()
+			if err != nil {
+				h.log.Infof("websocket app console session %s closed for device %s app %s: %v", session.UUID, deviceName, appName, err)
+				break
+			}
+			if msgType == websocket.BinaryMessage {
+				session.SendCh <- message
+			} else {
```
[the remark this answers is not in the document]
asafbennatan: Console sessions are intended to be long-lived (interactive terminal). Applying a fixed write deadline (e.g. 10s) would disconnect sessions during normal quiet periods. The right fix is context-cancellation only, not a hard timeout. Leaving open for a focused follow-up that targets the correct liveness signal (e.g. WebSocket ping/pong) rather than a write deadline.

internal/agent/device/applications/vm_status.go
```diff
@@ -0,0 +1,117 @@
...
+	domain := fmt.Sprintf("%s_%s", virtLauncherDomainNamespace, p.appName)
+	stdout, stderr, exitCode := p.exec.ExecuteWithContext(ctx, "podman", "exec", container, "virsh", "domstate", domain)
+	if exitCode != 0 {
+		p.consecutiveFailures++
+		p.log.Debugf("virsh domstate for %q exited %d (failure %d/%d): %s", p.appName, exitCode, p.consecutiveFailures, p.maxFailures, strings.TrimSpace(stderr))
+		if p.consecutiveFailures >= p.maxFailures {
+			return v1beta1.ApplicationStatusError
+		}
+		return v1beta1.ApplicationStatusStarting
+	}
```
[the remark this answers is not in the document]
asafbennatan: Not in scope for this PR — `vm_status.go` is not part of the changes introduced here. Will be tracked as a follow-up.

internal/agent/device/applications/console/session.go
```diff
@@ -0,0 +1,164 @@
...
+	defer s.log.Debugf("vm serial console session finished for container %s", s.containerName)
+
+	pid, err := s.getContainerPID(ctx)
+	if err != nil {
+		sendErrorOverStream(streamClient, fmt.Sprintf("failed to get container PID for %s: %v", s.containerName, err))
+		return
+	}
+
+	socketPath := fmt.Sprintf("/proc/%s/root%s", pid, serialSocketPath)
+	conn, err := s.dialWithRetry(ctx, socketPath)
```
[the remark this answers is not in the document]
asafbennatan: Not applicable to the current implementation. The serial console now bridges via `podman exec -i nc -U <socket>` inside the compute container rather than directly opening the socket path via the container PID. There is no PID lookup to retry.

internal/kvstore/keys.go
```diff
@@ -109,3 +109,15 @@ type AwaitingReconnectionKey struct {
...
+// of the kubevirt-vm-to-pod conversion. The key is global (not scoped to
+// org/fleet/templateVersion) because the conversion is deterministic: the same
+// vm.yaml input always produces the same pod.yaml output.
+type VmPodYamlKey struct {
+	Sha256 string // hex-encoded SHA-256 of the vm.yaml content
+}
+
+func (k *VmPodYamlKey) ComposeKey() string {
+	return fmt.Sprintf("v1/vm-pod-yaml/%s", k.Sha256)
+}
```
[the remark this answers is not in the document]
asafbennatan: Out of scope for this PR. Versioning the VM render cache key is a valid concern but requires a broader design decision. Tracking as a follow-up.

internal/agent/client/podman.go
```diff
@@ -252,6 +253,23 @@ func (p *Podman) ExtractArtifact(ctx context.Context, artifact, destination stri
 
 // Inspect returns the JSON output of the image inspection. The expectation is
 // that the image exists in local container storage.
+// ContainerPID returns the host PID of the container's init process.
```
kkyrazis: minor, but it looks like the documentation for Inspect was moved to the wrong spot

Also, did this end up being used somewhere? I probably missed it, but didn't see it.

internal/agent/client/podman.go
```diff
@@ -475,6 +493,46 @@ func (p *Podman) ExecInContainer(ctx context.Context, container string, args ...
...
+	}
+	_ = c.cmd.Wait()
+	return nil
+}
+
+// ExecStream starts `podman exec -i <containerName> <cmd...>` and returns the
+// process's stdin/stdout as an io.ReadWriteCloser. Caller must Close() to release resources.
+func (p *Podman) ExecStream(containerName string, cmd ...string) (io.ReadWriteCloser, error) {
+	args := append([]string{"exec", "-i", containerName}, cmd...)
+	c := p.exec.CommandContext(context.Background(), podmanCmd, args...)
```
kkyrazis: we should probably allow a context to be supplied rather than always spawning a background context

internal/agent/client/podman.go
```diff
@@ -475,6 +493,46 @@ func (p *Podman) ExecInContainer(ctx context.Context, container string, args ...
...
+type pipeConn struct {
+	cmd    *exec.Cmd
+	stdin  io.WriteCloser
+	stdout io.ReadCloser
+}
+
+func (c *pipeConn) Read(p []byte) (int, error)  { return c.stdout.Read(p) }
+func (c *pipeConn) Write(p []byte) (int, error) { return c.stdin.Write(p) }
+func (c *pipeConn) Close() error {
+	_ = c.stdin.Close()
```
kkyrazis: should we be ignoring all of these errors? probably at least need to worry about cmd.Wait right?

internal/agent/config/config.go
```diff
@@ -299,9 +302,34 @@ func (cfg *Config) Complete() error {
...
+		cfg.RemoteAccessService.Config = *cfg.EnrollmentService.Config.DeepCopy()
+		cfg.RemoteAccessService.Config.AuthInfo = baseclient.AuthInfo{}
+		cfg.RemoteAccessService.Config.Service.Server = remoteAccessServer
+	}
 	return nil
 }
 
+// deriveRemoteAccessServer replaces the port in the enrollment service URL with the
+// remote-access gRPC port (7444). Falls back to DefaultRemoteAccessEndpoint if the
+// enrollment URL cannot be parsed or has no port.
```
kkyrazis: the fallback comment doesn't look to be what is implemented

internal/agent/device/applications/console/manager.go
```diff
@@ -0,0 +1,209 @@
...
+// Session is the per-console-type I/O interface. Each app type (VM, container, …)
+// provides its own implementation. Manager knows nothing about the specific
+// mechanism — it only calls Run.
+type Session interface {
+	Run(ctx context.Context, streamClient grpc_v1.RouterService_StreamClient)
+}
+
+// AppConsoleResolver resolves an app name and console type into a ready-to-run
+// Session, or returns an error if the combination is not supported.
+// Defined here so this package does not need to import the applications package.
```
kkyrazis: this comment seems unnecessary

internal/agent/device/applications/console/manager.go
```diff
@@ -0,0 +1,209 @@
...
+	grpc_v1 "github.com/flightctl/flightctl/api/grpc/v1"
+	"github.com/flightctl/flightctl/internal/consts"
+	"github.com/flightctl/flightctl/pkg/log"
+	"github.com/samber/lo"
+	"google.golang.org/grpc/metadata"
+)
+
+// Session is the per-console-type I/O interface. Each app type (VM, container, …)
+// provides its own implementation. Manager knows nothing about the specific
+// mechanism — it only calls Run.
```
kkyrazis: the comments starting with 'Manager knows nothing' doesn't seem all that useful

internal/agent/device/applications/console/session.go
```diff
@@ -0,0 +1,132 @@
...
+	go func() {
+		if _, err := conn.Write([]byte("\r")); err != nil {
+			s.log.Debugf("failed to send initial CR to serial console: %v", err)
+		}
+	}()
+
+	bridgeConn(ctx, "serial", conn, streamClient, s.log)
+}
+
+// Ensure net.Conn satisfies io.ReadWriteCloser so existing test helpers that
```
kkyrazis: these are both types that are out of our control. Is this assertion something useful to us?

internal/agent/device/applications/console/session.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	defer conn.Close()
+
+	// Send an initial CR to wake up agetty, which waits for the first character
+	// before displaying the login prompt (baud-rate detection on real hardware).
+	// Done asynchronously so that a slow or synchronous connection (e.g. net.Pipe
+	// in tests) does not block bridge startup.
+	go func() {
+		if _, err := conn.Write([]byte("\r")); err != nil {
+			s.log.Debugf("failed to send initial CR to serial console: %v", err)
```
kkyrazis: if this doesn't succeed can we still progress with the bridge? Also, there _could_ be a race condition in which the bridge attempts to write to the conn first right?

internal/agent/client/podman.go
asafbennatan: you are right this if from an earlier iteration - removed

internal/agent/client/podman.go
asafbennatan: done

internal/agent/client/podman.go
asafbennatan: well this is from close - what will we do if it errors out ? its a best effort teardown

internal/agent/config/config.go
asafbennatan: fixed

internal/agent/device/applications/console/manager.go
asafbennatan: removed

internal/agent/device/applications/console/manager.go
asafbennatan: changed so its clear that the intention here is that each session type provides its own implementation ( only serial for now , vnc will be introduced on a later PR)

internal/agent/device/applications/console/session.go
asafbennatan: removed

internal/agent/device/applications/console/session.go
asafbennatan: its intentionally fire and forget  , the only reason its there is to wake up the connection - otherwise the user has to press enter before they see data (login prompt for example) 
in the case it fails the connection is probably closed and will close anyway - it wouldn't make sense to control the connection lifecycle from there . even if there is a theoretical race condition (which probably couldn't happen since the user would have to somehow connect and send bytes through the channel) then nothing critical will happen

internal/agent/client/podman.go
kkyrazis: I don't disagree but the interface explicitly allows for returning an error. Shouldn't that be up to the caller to decide to handle that error?

asafbennatan: adjusted

internal/agent/device/applications/console/session.go
```diff
@@ -0,0 +1,83 @@
...
+						logger.Debugf("write to %s connection failed: %v", label, writeErr)
+						return
+					}
+					remaining = remaining[n:]
+				}
+			}
+		}
+	}()
+
+	wg.Wait()
```
galel12, line 85: if the socket read goroutine exits first, the other side can still be stuck in `Recv()` — `cancel()` closes the conn but doesn't unblock the gRPC stream. The device console calls `CloseSend()` from the outgoing side for exactly this reason (`outgoing_streams.go`). Should we do something similar here so `wg.Wait()` doesn't hang (and `RunConsole` → `Wait()` doesn't block shutdown)?

internal/agent/device/applications/console/manager.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+	m.inactiveSessions = result
+}
+
+func (m *Manager) add(s *managedSession) bool {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.cleanup()
+	if _, exists := lo.Find(append(m.activeSessions, m.inactiveSessions...), func(ms *managedSession) bool {
```
galel12, line 90: `add()` dedupes against both active and inactive sessions, and inactive ones stick around for `cleanupDuration` (5 min). If the remote-session annotation is still on the device after disconnect, we won't start a new session with the same ID until that window expires. Is that intentional? If the server clears the annotation on close this is probably fine, but worth confirming the lifecycle lines up.

internal/agent/device/applications/podman_monitor.go
```diff
@@ -729,3 +731,65 @@ func (e *podmanEventWatcher) Stop() error {
...
+		return nil, fmt.Errorf("app %q: unsupported console type %q for VM (supported: serial)", appName, consoleType)
+	}
+
+	// The container name comes from podman events via workload tracking — it is the authoritative
+	// runtime name and requires no spec-side derivation.
+	// KubeVirt VM pods have two containers: an infra/pause container (suffix "-infra") and the
+	// actual virt-launcher compute container (suffix "-compute"). Exec must target the compute
+	// container; the infra container has no tools or sockets.
+	for _, w := range found.Workloads() {
+		m.log.Infof("console: workload %q status=%s", w.Name, w.Status)
```
galel12: minor, but this logs every workload at Info on each resolve. `Sync` runs on every device update while the annotation is present, so this could get noisy — maybe Debug, or just log the container we end up selecting?

internal/agent/config/config.go
```diff
@@ -299,9 +302,33 @@ func (cfg *Config) Complete() error {
...
 }
 
+// deriveRemoteAccessServer replaces the port in the enrollment service URL with the
+// remote-access gRPC port (7444). Returns an error if the enrollment URL cannot be parsed.
+func deriveRemoteAccessServer(enrollmentServer string) (string, error) {
+	u, err := url.Parse(enrollmentServer)
+	if err != nil {
+		return "", fmt.Errorf("parsing enrollment server URL %q: %w", enrollmentServer, err)
+	}
+	u.Host = u.Hostname() + ":7444"
```
galel12: minor, but `u.Hostname() + ":7444"` won't produce a valid URL for IPv6 enrollment endpoints. `net.JoinHostPort(u.Hostname(), "7444")` would handle that if we ever need it.

galel12 (commented):
Agent-side serial console looks good overall — clean layering and the podman exec approach makes sense. A few things below; the bridge teardown one is the main one I'd want addressed before merge.

Also, the PR description still mentions `TestVMSerialSessionDialWithRetry` but I don't see that test in the diff anymore — worth updating the test plan.

internal/agent/config/config.go
asafbennatan: fixed although not an established pattern in the codebase

internal/agent/device/applications/podman_monitor.go
asafbennatan: fixed

internal/agent/device/applications/console/manager.go
asafbennatan: session is random so no reason for dedups

internal/agent/device/applications/console/session.go
asafbennatan: fixed

asafbennatan:
> Agent-side serial console looks good overall — clean layering and the podman exec approach makes sense. A few things below; the bridge teardown one is the main one I'd want addressed before merge.
> 
> Also, the PR description still mentions `TestVMSerialSessionDialWithRetry` but I don't see that test in the diff anymore — worth updating the test plan.

adjusted PR description

galel12 (approved):
LGTM

PR #3112, opened 2026-06-16
EDM-4184: multi-org cleanup to fix regression in parameterizable templates

eldar101:
## Summary

Reset the quadlet multiorg suite back to the default org during cleanup.

## Why

The quadlet job has a shared client config across suites in the same Jenkins workspace/test VM. The multiorg suite switches that config to the PAM org, for example `multiorg-test` / `b2be...`.

Later suites, such as `parametrisable_templates`, reuse the same harness/client config. However, quadlet agent enrollment creates EnrollmentRequests in the default org:

`00000000-0000-0000-0000-000000000000`

So the ER exists, but the next suite polls for it in the leaked multiorg context and gets `404` until `WaitForEnrollmentRequest()` times out.

This is why the issue shows up in quadlets: it combines PAM multiorg org switching, shared persisted client config, and default-org agent enrollment.

## Change

During quadlet multiorg `AfterSuite`, restore the persisted current organization to the default org before cleaning up the multiorg users/org. The PAM org cleanup now has bounded timeout handling and reports partial cleanup failures.

https://reportportal-ecosystem-qe.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#edge-management-qe/launches/all/38139/2198131

### DIFF
--- a/test/e2e/multiorg/multiorg_suite_test.go
+++ b/test/e2e/multiorg/multiorg_suite_test.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"os"
 	"testing"
+	"time"
 
+	"github.com/flightctl/flightctl/internal/org"
 	"github.com/flightctl/flightctl/test/e2e/infra"
 	"github.com/flightctl/flightctl/test/e2e/infra/auxiliary"
 	"github.com/flightctl/flightctl/test/e2e/infra/quadlet"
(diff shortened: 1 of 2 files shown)


test/e2e/multiorg/multiorg_suite_test.go
```diff
@@ -17,7 +19,10 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-const quadletOrgName = "multiorg-test"
+const (
+	quadletOrgName           = "multiorg-test"
+	quadletOrgCleanupTimeout = 30 * time.Second
```
itzikb-redhat: I would raise it , other than that lgtm

itzikb-redhat (approved):

PR #3120, opened 2026-06-18
NO-ISSUE: Deny operator enrollment request access in static auth

itzikb-redhat:
Align quadlet/PAM operator permissions with the K8s ClusterRole by explicitly blocking enrollmentrequests, overriding the read wildcard.


eldar101:
Fix coderabbit and warnings and I'll approve.

itzikb-redhat:
There is a bug for the gaps between OCP and quadlets that should be addressed so closing this PR

PR #3122, opened 2026-06-18
EDM-3853: Expose remote Quadlet services through SSH tunnels

eldar101:
Redis restart tests failed because Redis port 6379 is not exposed after a remote Quadlet upgrade.

The tests now create an SSH tunnel to the Redis container instead of connecting directly to the VM port.

The tunnel is automatically recreated after Redis restarts.

https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-flightctl-gotests/1813/testReport/(root)/Redis%20Restart%20E2E%20Suite/

### DIFF
--- a/test/e2e/infra/quadlet/infra.go
+++ b/test/e2e/infra/quadlet/infra.go
@@ -4,6 +4,7 @@ package quadlet
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -13,10 +14,17 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/flightctl/flightctl/test/e2e/infra"
 	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
+)
+
+const (
+	remoteForwardReadyTimeout = 10 * time.Second
+	remoteForwardPollInterval = 100 * time.Millisecond
+	remoteForwardStopTimeout  = time.Second
 )
 
 // ServiceInfo holds metadata about a Quadlet service.
@@ -160,7 +168,8 @@ func (p *InfraProvider) runCommandWithOptionalStdinContext(ctx context.Context, 
 
 	if p.isRemote() {
 		sshArgs := []string{"-o", "StrictHostKeyChecking=no"}
-		usePassword := p.sshKeyPath == "" && getSSHPassword() != ""
+		password := getSSHPassword()
+		usePassword := p.sshKeyPath == "" && password != ""
 		if p.sshKeyPath != "" {
 			sshArgs = append(sshArgs, "-o", "BatchMode=yes", "-i", p.sshKeyPath)
 		} else if !usePassword {
@@ -179,8 +188,14 @@ func (p *InfraProvider) runCommandWithOptionalStdinContext(ctx context.Context, 
 		sshArgs = append(sshArgs, strings.Join(remoteParts, " "))
 
 		if usePassword {
-			cmd = exec.CommandContext(ctx, "sshpass", append([]string{"-e", "ssh"}, sshArgs...)...) //nolint:gosec // G204: sshArgs from trusted config
-			cmd.Env = append(os.Environ(), "SSHPASS="+getSSHPassword())
+			passwordFile, cleanupPasswordFile, err := sshpassPasswordFile(password)
+			if err != nil {
+				return "", fmt.Errorf("prepare SSH password for remote command: %w", err)
+			}
+			defer cleanupPasswordFile()
+
+			cmd = exec.CommandContext(ctx, "sshpass", append([]string{"-d", "3", "ssh"}, sshArgs...)...) //nolint:gosec // G204: sshArgs from trusted config
+			cmd.ExtraFiles = []*os.File{passwordFile}
 		} else {
 			cmd = exec.CommandContext(ctx, "ssh", sshArgs...) //nolint:gosec // G204: sshArgs from trusted config
 		}
@@ -221,6 +236,13 @@ func (p *InfraProvider) runCommandWithStdin(stdin io.Reader, command ...string) 
 
 // GetConfigValue retrieves a configuration value from config files or environment.
 func (p *InfraProvider) GetConfigValue(name, key string) (string, error) {
+	if err := validatePathComponent("config name", name); err != nil {
+		return "", err
+	}
+	if err := validatePathComponent("config key", key); err != nil {
+		return "", err
+	}
+
 	// Try environment variable first (uppercase, with underscores)
 	envKey := strings.ToUpper(strings.ReplaceAll(name+"_"+key, "-", "_"))
 	if value := os.Getenv(envKey); value != "" {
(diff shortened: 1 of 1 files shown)


test/e2e/infra/quadlet/infra.go
```diff
@@ -13,12 +14,19 @@ import (
...
 	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
 )
 
+const (
+	remoteForwardReadyTimeout = 10 * time.Second
+	remoteForwardPollInterval = 100 * time.Millisecond
+	remoteForwardStopTimeout  = time.Second
+)
+
```
Samudelacruz77, line 29: Why not to use HOST_PORT:CONTAINER_PORT in container start to avoid the issue?

eldar101: I’d avoid it for this PR.

If Redis isn’t already published, adding HOST_PORT:CONTAINER_PORT changes the actual quadlet deployment and exposes Redis on the host. That feels like a separate deploy/security decision.

The tunnel keeps this as a test-infra fix: Redis stays private, and CI can still reach it when needed.

asafbennatan (commented):
do we need to adjust getPublishedPort ?

asafbennatan (commented):
Code review findings (4 items, reviewed against the latest commit ).

test/e2e/infra/quadlet/infra.go
```diff
@@ -667,12 +728,173 @@ func (p *InfraProvider) mergeAndWriteServiceConfig(updates map[string]interface{
...
+		sshArgs = append(sshArgs, "-o", "BatchMode=yes", "-i", p.sshKeyPath)
+	} else if !usePassword {
+		sshArgs = append(sshArgs, "-o", "BatchMode=yes")
+	}
+	sshArgs = append(sshArgs, fmt.Sprintf("%s@%s", p.sshUser, p.host))
+
+	var cmd *exec.Cmd
+	if usePassword {
+		cmd = exec.CommandContext(ctx, "sshpass", append([]string{"-d", "0", "ssh"}, sshArgs...)...) //nolint:gosec // G204: sshArgs from trusted config
+		cmd.Stdin = strings.NewReader(password)
```
asafbennatan: **MEDIUM — `startRemotePortForward` should use `sshpassPasswordFile` for consistency**

`runCommandWithOptionalStdinContext` was just updated to use the `sshpassPasswordFile` pipe/fd-3 approach, but `startRemotePortForward` still passes the password via `sshpass -d 0` + `cmd.Stdin`:

```go
cmd = exec.CommandContext(ctx, "sshpass", append([]string{"-d", "0", "ssh"}, sshArgs...)...)
cmd.Stdin = strings.NewReader(password)
```

Two different fd-based mechanisms now coexist for the same authentication path. Suggestion — reuse the new helper here too:

```go
passwordFile, cleanupPasswordFile, err := sshpassPasswordFile(password)
if err != nil {
    cancel()
    return "", nil, fmt.Errorf("prepare SSH password for %s forward: %w", service, err)
}
defer cleanupPasswordFile() // fd is dup'd into child before Start returns, so this is safe
cmd = exec.CommandContext(ctx, "sshpass", append([]string{"-d", "3", "ssh"}, sshArgs...)...)
cmd.ExtraFiles = []*os.File{passwordFile}
```

Note: `startRemotePortForward` uses `cmd.Start()` not `cmd.CombinedOutput()`, but the fd is `dup2`'d into the child before `Start()` returns, so closing the parent's copy via the deferred cleanup immediately after `Start()` is safe.

test/e2e/infra/quadlet/infra.go
```diff
@@ -667,12 +728,173 @@ func (p *InfraProvider) mergeAndWriteServiceConfig(updates map[string]interface{
...
 	if _, err := p.runCommand("sh", "-c", script); err != nil {
-		return err
+		return fmt.Errorf("write remote file %s: %w", path, err)
 	}
 	return nil
 }
+
+// startRemotePortForward creates an SSH tunnel from a local port to an unpublished service container port.
+func (p *InfraProvider) startRemotePortForward(containerIP string, containerPort int, protocol string, service infra.ServiceName) (string, func(), error) {
+	if p == nil {
```
asafbennatan: **LOW — Nil receiver check is unreachable**

```go
if p == nil {
    return "", nil, fmt.Errorf("start remote port-forward: provider is nil")
}
```

`startRemotePortForward` is only called from `ExposeService`, where `p` is guaranteed non-nil — a nil receiver would have panicked on `p.exposeCacheMu.Lock()` long before reaching this function. This guard is dead code and non-idiomatic Go. Suggest removing it.

test/e2e/infra/quadlet/infra.go
```diff
@@ -479,13 +520,17 @@ func (p *InfraProvider) getContainerIP(containerName string) (string, error) {
...
 func getFreePort() (int, error) {
 	listener, err := net.Listen("tcp", "127.0.0.1:0")
 	if err != nil {
 		return 0, err
 	}
-	defer listener.Close()
-	return listener.Addr().(*net.TCPAddr).Port, nil
+	port := listener.Addr().(*net.TCPAddr).Port
+	if err := listener.Close(); err != nil {
+		return 0, fmt.Errorf("close temporary port listener: %w", err)
```
asafbennatan: **LOW — Probe-listener close error is overly fatal**

```go
if err := listener.Close(); err != nil {
    return 0, fmt.Errorf("close temporary port listener: %w", err)
}
```

A failure to close the probe listener doesn't mean the port is unavailable — the port number was captured on the previous line and is valid. Returning an error here discards a working port on an extremely rare OS-level event. Suggest logging the close failure as a warning and returning the port regardless:

```go
if err := listener.Close(); err != nil {
    logrus.Warnf("getFreePort: close probe listener: %v", err)
}
return port, nil
```

test/e2e/infra/quadlet/infra.go
```diff
@@ -496,15 +541,31 @@ func runTCPForward(listener net.Listener, target string, service infra.ServiceNa
...
+				if _, err := io.Copy(client, backend); err != nil && !errors.Is(err, net.ErrClosed) {
+					logrus.Debugf("Quadlet forward %s: copy backend to client: %v", service, err)
+				}
+				copyDone <- struct{}{}
+			}()
+
+			<-copyDone
+			closeForwardConnection(client, service, "client")
+			closeForwardConnection(backend, service, "backend")
+			<-copyDone
```
asafbennatan, line 569: **LOW — Double-close in `runTCPForward` teardown deserves a comment**

```go
<-copyDone
closeForwardConnection(client, service, "client")
closeForwardConnection(backend, service, "backend")
<-copyDone
```

Closing both connections after the first copy direction finishes is intentional — it unblocks the still-running goroutine in the other direction (which will get `net.ErrClosed`, suppressed at Debug level). This is correct but non-obvious. A short comment would help future readers:

```go
<-copyDone
// Close both ends to unblock the other copy goroutine; ErrClosed is suppressed in both goroutines.
closeForwardConnection(client, service, "client")
closeForwardConnection(backend, service, "backend")
<-copyDone
```

asafbennatan (approved):
LGTM

PR #3134, opened 2026-06-21
EDM-4104: Remote access scaffold

asafbennatan:
adds boilerplate for a new service - flightctl-remote-access service,
fixes the upgrade process to use the --reset-then-reuse flag as discussed multiple times already , the issue mostly resurfaces for new services requiring to add boiler plate templating , this fixes it

### DIFF
--- a/cmd/flightctl-remote-access/main.go
+++ b/cmd/flightctl-remote-access/main.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/flightctl/flightctl/internal/config"
+	"github.com/flightctl/flightctl/internal/crypto"
+	"github.com/flightctl/flightctl/internal/instrumentation/tracing"
+	remoteaccessserver "github.com/flightctl/flightctl/internal/remote_access_server"
+	"github.com/flightctl/flightctl/pkg/log"
+)
+
+func main() {
+	ctx := context.Background()
+
+	cfg, err := config.LoadOrGenerate(config.ConfigFile())
+	if err != nil {
+		log.InitLogs().Fatalf("reading configuration: %v", err)
+	}
+
+	log := log.InitLogs(cfg.Service.LogLevel)
+	log.Println("Starting remote-access service")
+	defer log.Println("Remote-access service stopped")
+	log.Printf("Using config: %s", cfg)
+
+	ca, err := crypto.LoadInternalCA(cfg.CA)
+	if err != nil {
+		log.Fatalf("loading CA certificates: %v", err)
+	}
+	caClient := crypto.NewCAClient(cfg.CA, ca)
+
+	serverCerts, err := config.LoadServerCertificates(cfg, log)
+	if err != nil {
+		log.Fatalf("loading server certificates: %v", err)
+	}
+
+	tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-remote-access")
+	defer func() {
+		if err := tracerShutdown(ctx); err != nil {
+			log.Fatalf("failed to shut down tracer: %v", err)
+		}
+	}()
+
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT)
+	defer cancel()
+
+	server, err := remoteaccessserver.New(log, cfg, caClient, serverCerts)
+	if err != nil {
+		log.Fatalf("initializing remote-access server: %v", err)
+	}
+
+	if err := server.Run(ctx); err != nil {
+		log.Fatalf("Error running remote-access server: %v", err)
+	}
+}
--- a/cmd/flightctl-remote-access/main.go
+++ b/cmd/flightctl-remote-access/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/flightctl/flightctl/internal/config"
 	"github.com/flightctl/flightctl/internal/crypto"
@@ -39,8 +40,10 @@ func main() {
 
 	tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-remote-access")
 	defer func() {
-		if err := tracerShutdown(ctx); err != nil {
-			log.Fatalf("failed to shut down tracer: %v", err)
+		shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer shutdownCancel()
+		if err := tracerShutdown(shutdownCtx); err != nil {
+			log.Errorf("failed to shut down tracer: %v", err)
 		}
 	}()
 
--- a/deploy/helm/flightctl/templates/_helpers.tpl
+++ b/deploy/helm/flightctl/templates/_helpers.tpl
@@ -729,6 +729,23 @@ Usage: {{- $result := include "flightctl.getImagebuilderApiDNSSans" . | fromJson
 {{- end }}
 
 {{- /*
+Get DNS SANs for remote-access agent server certificate
+Usage: {{- $result := include "flightctl.getRemoteAccessDNSSans" . | fromJson }}{{ $remoteAccessDNSSans := $result.sans }}
+*/}}
+{{- define "flightctl.getRemoteAccessDNSSans" }}
+  {{- $sans := list }}
+  {{- $baseDomain := include "flightctl.getBaseDomain" . }}
+  {{- $sans = append $sans (printf "agent-remote-access.%s" $baseDomain) }}
+  {{- $sans = append $sans "flightctl-remote-access" }}
+  {{- $sans = append $sans "flightctl-remote-access-agent" }}
+  {{- $sans = append $sans (printf "flightctl-remote-access.%s" .Release.Namespace) }}
+  {{- $sans = append $sans (printf "flightctl-remote-access-agent.%s" .Release.Namespace) }}
+  {{- $sans = append $sans (printf "flightctl-remote-access.%s.svc.cluster.local" .Release.Namespace) }}
+  {{- $sans = append $sans (printf "flightctl-remote-access-agent.%s.svc.cluster.local" .Release.Namespace) }}
+  {{- dict "sans" $sans | toJson -}}
+{{- end }}
+
+{{- /*
 Auth configuration block helper.
 Outputs the auth configuration section for service config files.
 Usage: {{- include "flightctl.authConfig" . | nindent 4 }}
--- a/deploy/helm/flightctl/templates/_helpers.tpl
+++ b/deploy/helm/flightctl/templates/_helpers.tpl
@@ -735,6 +735,7 @@ Usage: {{- $result := include "flightctl.getRemoteAccessDNSSans" . | fromJson }}
 {{- define "flightctl.getRemoteAccessDNSSans" }}
   {{- $sans := list }}
   {{- $baseDomain := include "flightctl.getBaseDomain" . }}
+  {{- $sans = append $sans (printf "remote-access.%s" $baseDomain) }}
   {{- $sans = append $sans (printf "agent-remote-access.%s" $baseDomain) }}
   {{- $sans = append $sans "flightctl-remote-access" }}
   {{- $sans = append $sans "flightctl-remote-access-agent" }}
--- a/deploy/helm/flightctl/templates/remote-access/flightctl-remote-access-deployment.yaml
+++ b/deploy/helm/flightctl/templates/remote-access/flightctl-remote-access-deployment.yaml
@@ -1,0 +1,84 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    flightctl.service: flightctl-remote-access
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+  name: flightctl-remote-access
+  namespace: {{ .Release.Namespace }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      flightctl.service: flightctl-remote-access
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        flightctl.service: flightctl-remote-access
+        {{- include "flightctl.standardLabels" . | nindent 8 }}
+    spec:
+      {{- if .Values.global.imagePullSecretName }}
+      imagePullSecrets:
+        - name: {{ .Values.global.imagePullSecretName }}
+      {{- end }}
+      containers:
+        - name: flightctl-remote-access
+          image: {{ include "flightctl.ensureOsQualifiedImage" (dict "root" . "imageName" .Values.remoteAccess.image.image) }}:{{ default .Chart.AppVersion .Values.remoteAccess.image.tag }}
+          imagePullPolicy: {{ default .Values.global.imagePullPolicy .Values.remoteAccess.image.pullPolicy }}
+          ports:
+            - containerPort: 3444
+              name: service-api
+              protocol: TCP
+            - containerPort: 7444
+              name: agent-api
+              protocol: TCP
+          env:
+            - name: HOME
+              value: "/root"
+            {{- if .Values.remoteAccess.env }}
+            {{- range $key, $value := .Values.remoteAccess.env }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
+            {{- end }}
+          volumeMounts:
+            - mountPath: /root/.flightctl/config.yaml
+              name: flightctl-remote-access-config
+              subPath: config.yaml
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/client-signer.crt
+              name: flightctl-client-signer
+              subPath: tls.crt
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/client-signer.key
+              name: flightctl-client-signer
+              subPath: tls.key
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/server.crt
+              name: flightctl-remote-access-server-tls
+              subPath: tls.crt
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/server.key
+              name: flightctl-remote-access-server-tls
+              subPath: tls.key
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/ca-bundle.crt
+              name: flightctl-ca-bundle
+              subPath: ca-bundle.crt
+              readOnly: true
+      restartPolicy: Always
+      volumes:
+        - name: flightctl-remote-access-config
+          configMap:
+            name: flightctl-remote-access-config
+        - name: flightctl-client-signer
+          secret:
+            secretName: flightctl-client-signer-ca
+        - name: flightctl-remote-access-server-tls
+          secret:
+            secretName: flightctl-remote-access-server-tls
+        - name: flightctl-ca-bundle
+          secret:
+            secretName: flightctl-ca-bundle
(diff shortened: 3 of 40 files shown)


packaging/images/el9/images.yaml
```diff
@@ -64,4 +64,7 @@ telemetry-gateway:
   tag: latest
 userinfo-proxy:
   image: quay.io/flightctl/flightctl-userinfo-proxy-el9
+  tag: latest
```
kkyrazis, line 67: there should be corresponding el10 changes

cmd/flightctl-remote-access/main.go
```diff
@@ -0,0 +1,58 @@
...
+
+	tracerShutdown := tracing.InitTracer(log, cfg, "flightctl-remote-access")
+	defer func() {
+		if err := tracerShutdown(ctx); err != nil {
+			log.Fatalf("failed to shut down tracer: %v", err)
+		}
+	}()
+
+	ctx, cancel := signal.NotifyContext(ctx, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGQUIT)
+	defer cancel()
```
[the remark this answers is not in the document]
asafbennatan: Fixed — using a fresh `context.WithTimeout(context.Background(), 5*time.Second)` for tracer shutdown so the flush is not affected by the already-cancelled signal context. Also downgraded `Fatalf` → `Errorf` since a tracer flush failure on clean shutdown should not obscure the exit code.

deploy/helm/flightctl/templates/remote-access/flightctl-remote-access-deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/server.key
+              name: flightctl-remote-access-server-tls
+              subPath: tls.key
+              readOnly: true
+            - mountPath: /root/.flightctl/certs/ca-bundle.crt
+              name: flightctl-ca-bundle
+              subPath: ca-bundle.crt
+              readOnly: true
+      restartPolicy: Always
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added `automountServiceAccountToken: false` at the pod spec level and a container `securityContext` with `runAsNonRoot: true`, `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, and `capabilities.drop: ["ALL"]`. All volume mounts are already read-only, so the root filesystem restriction is safe for the current stub implementation.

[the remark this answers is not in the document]
asafbennatan: Fixed — added TCP liveness and readiness probes on the `agent-api` port (`:7444`). HTTP probes are not applicable since the service-api port (`:3444`) is a stub that returns 501 for all requests; a `/healthz` endpoint will be added in a future story. Resource requests and limits are added via `values.yaml` (cpu: 100m/500m, memory: 128Mi/256Mi) with schema coverage in `values.schema.json`.

deploy/podman/flightctl-gateway/flightctl-gateway-config/nginx.conf.template
```diff
@@ -143,4 +143,12 @@ stream {
...
     proxy_pass flightctl-telemetry-gateway:4317;
   }
+
+  # Passthrough to the Remote Access Agent service
+  server {
+    listen 7444;
+    listen [::]:7444 ipv6only=on;
+
+    proxy_pass flightctl-remote-access:7444;
+  }
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added `PublishPort=7444:7444` to `deploy/podman/flightctl-gateway/flightctl-gateway.container` alongside the existing port mappings.

deploy/podman/flightctl-remote-access/flightctl-remote-access.container
```diff
@@ -0,0 +1,29 @@
...
+AddHost=%H:host-gateway
+Environment=HOME=/root
+
+Volume=/etc/flightctl/pki/flightctl-remote-access/server.crt:/root/.flightctl/certs/server.crt:ro,z
+Volume=/etc/flightctl/pki/flightctl-remote-access/server.key:/root/.flightctl/certs/server.key:ro,z
+Volume=/etc/flightctl/pki/flightctl-remote-access/client-signer.crt:/root/.flightctl/certs/client-signer.crt:ro,z
+Volume=/etc/flightctl/pki/flightctl-remote-access/client-signer.key:/root/.flightctl/certs/client-signer.key:ro,z
+Volume=/etc/flightctl/pki/ca-bundle.crt:/root/.flightctl/certs/ca-bundle.crt:ro,z
+Volume=/etc/flightctl/flightctl-remote-access/config.yaml:/root/.flightctl/config.yaml:ro,z
+
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added `NoNewPrivileges=true` and `DropCapability=all` to the `[Container]` section.

internal/remote_access_server/server_test.go
```diff
@@ -0,0 +1,67 @@
...
+func TestServerStream(t *testing.T) {
+	t.Parallel()
+	srv := &Server{}
+	stream := &stubStreamServer{}
+
+	err := srv.Stream(stream)
+	if err != nil {
+		t.Errorf("When Stream is called it should return nil, got %v", err)
+	}
+}
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added `TestGrpcMuxHandlerFunc` with table-driven cases covering `application/grpc`, `application/grpc+proto`, `application/grpc; charset=utf-8` (all routed to gRPC), and `application/json` / HTTP/1 with gRPC content-type (routed to the HTTP stub). Tests use the actual `grpcMuxHandlerFunc` function.

internal/remote_access_server/server.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}()
+
+	<-ctx.Done()
+	s.log.Println("Shutdown signal received:", ctx.Err())
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), apiserver.GracefulShutdownTimeout)
+	defer cancel()
+	_ = agentSrv.Shutdown(ctxTimeout)
+	_ = httpSrv.Shutdown(ctxTimeout)
+	return nil
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added a `serveErrCh` error channel (buffered, capacity 2) and replaced `<-ctx.Done()` with a `select` over `ctx.Done()` and `serveErrCh`. Unexpected listener errors are now propagated via the channel, trigger graceful shutdown of both servers, and are returned from `Run()` so the caller can respond appropriately.

internal/remote_access_server/server.go
```diff
@@ -0,0 +1,151 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		http.Error(w, http.StatusText(http.StatusNotImplemented), http.StatusNotImplemented)
+	})
+}
+
+// grpcMuxHandlerFunc routes incoming requests to grpcServer (gRPC) or
+// httpHandler (HTTP) based on the Content-Type header.
+func grpcMuxHandlerFunc(grpcServer *grpc.Server, httpHandler http.Handler, log logrus.FieldLogger) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.ProtoMajor == 2 && r.Header.Get("Content-Type") == "application/grpc" {
```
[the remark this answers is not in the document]
asafbennatan: Fixed — replaced the exact equality check `== "application/grpc"` with `strings.HasPrefix(..., "application/grpc")` so that `application/grpc+proto` and `application/grpc; charset=utf-8` are correctly routed to the gRPC server. Added `"strings"` import accordingly.

packaging/images/el9/images.yaml
asafbennatan: Added — `packaging/images/el10/Containerfile.remote-access`, `packaging/images/el10/images.yaml` (`quay.io/flightctl/flightctl-remote-access-el10`), and `packaging/images/el10/local-images.yaml` (`flightctl-remote-access-el10`).

deploy/podman/flightctl-remote-access/flightctl-remote-access-config/config.yaml.template
```diff
@@ -0,0 +1,3 @@
+service:
+  address: :3444
```
kkyrazis, line 2: I see the 7444 port added to the gateway config, but I'm missing where the 3444 port is setup to proxy appropriately. Is it needed?

deploy/helm/flightctl/values.schema.json
```diff
@@ -528,6 +528,46 @@
...
+              }
+            },
+            "limits": {
+              "type": "object",
+              "properties": {
+                "cpu": { "type": "string" },
+                "memory": { "type": "string" }
+              }
+            }
+          }
```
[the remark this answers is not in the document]
asafbennatan: Fixed — added `additionalProperties: false` to the `resources`, `requests`, and `limits` objects, matching the pattern already used by `db.resources` in the schema.

deploy/helm/flightctl/values.yaml
```diff
@@ -322,6 +322,26 @@ periodic:
...
+  # -- Additional environment variables for the remote access container
+  env: {}
+  # -- Resource requests and limits for the remote access container
+  resources:
+    requests:
+      cpu: "100m"
+      memory: "128Mi"
+    limits:
+      cpu: "500m"
+      memory: "256Mi"
```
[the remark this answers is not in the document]
asafbennatan: The `values.yaml` entry already carries the `# --` helm-docs annotation (`# -- Resource requests and limits for the remote access container`) so the README will pick it up on the next `helm-docs` regeneration run. Skipping a manual README edit to avoid drift with the generated output.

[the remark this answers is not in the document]
asafbennatan: Fixed — updated the README parameter table to include `remoteAccess.resources` and its sub-keys (`requests.cpu`, `requests.memory`, `limits.cpu`, `limits.memory`) with their default values. `helm-docs` is not wired into the CI pipeline for this chart, so the README is maintained manually alongside values.yaml changes.

deploy/podman/flightctl-remote-access/flightctl-remote-access-config/config.yaml.template
asafbennatan: Good catch — both ports are now wired up (479e586ad):

- **3444 (service API — CLI/UI WebSocket console):** exposed as `location /_/remote-access/` inside the main 8443 server block, following the same pattern as `/_/imagebuilder/` and other sub-services. TLS is terminated at nginx; the service accepts plain HTTP internally. No dedicated external port needed.

- **7444 (agent gRPC — mTLS):** remains a TCP stream passthrough so the remote-access service can read client certificates directly, same reason the existing agent API (7443) uses passthrough. Published as `PublishPort=7444:7444` on the gateway container.

kkyrazis: Does the corresponding route exist in the helm deployment too? I'm assuming if we need to expose 3444 here we need a route there too. If so, I believe that we'll need to conditionally start a tls enabled listener or a raw tcp listener similar to how the flightctl-api server behaves.

asafbennatan: added the route , 
as for conditional tls listener i dont think we do .
as far as i understand the tls listeners for existing services are there for backwards compatibility before we had the flightctl-gateway.
for ocp the tls terminates at the route
for quadlets we terminate at flightctl-gatway . 

am i missing anything ?

keitwb: Ideally for the standalone quadlets deployment it should be configured to use the `/_/remote-access` endpoint and not port 3444. Ideally we explicitly configure it everywhere since there isn't really any good default given that helm uses host-based routing and standalone quadlets uses path based without regard for service port.

deploy/podman/flightctl-gateway/flightctl-gateway.container
kkyrazis: The new service should be added to this list too.

deploy/podman/flightctl-remote-access/flightctl-remote-access-config/config.yaml.template
kkyrazis: All of these routes do tls passthrough rather than edge termination. 
https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/templates/api/flightctl-api-route.yaml
https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-route.yaml#L23
https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/templates/imagebuilder-api/flightctl-imagebuilder-api-route.yaml#L23
https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/templates/telemetry-gateway/flightctl-telemetry-gateway-route.yaml#L21

I just want to confirm that the current edge termination is correct

asafbennatan: So the current approach is fine as far as I understand? 
@keitwb @kkyrazis

keitwb: > as far as i understand the tls listeners for existing services are there for backwards compatibility before we had the flightctl-gateway.

The gateway eliminates the need for TLS on listeners that don't use mTLS for auth.

There is a node port deployment that I think QE uses in places that might still need the ability to listen with TLS but this isn't where they would configure it.

So this is fine as far as gateway is concerned.

asafbennatan: what about ocp ?

deploy/helm/flightctl/templates/remote-access/flightctl-remote-access-route.yaml
```diff
@@ -0,0 +1,71 @@
+{{ if and .Values.remoteAccess.enabled (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
+{{- $useExtCert := eq (include "flightctl.useRouteExternalCertificate" (dict "root" . "routeName" "flightctl-remote-access-route")) "true" }}
```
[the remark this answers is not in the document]
asafbennatan: Fixed — `routeName` now passes `"flightctl-remote-access"` to match the Route's metadata name.

internal/remote_access_server/server.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Plain TCP listener — TLS is terminated by the nginx gateway for CLI/UI
+	// WebSocket connections; the service only handles plain HTTP internally.
+	httpListener, err := net.Listen("tcp", cfg.Service.Address)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
asafbennatan: Fixed — all three `New()` failure paths now wrap the error with `fmt.Errorf` context: `"building agent TLS config: %w"`, `"listening on service address %q: %w"`, and `"listening on agent endpoint address %q: %w"`.

deploy/podman/flightctl-gateway/flightctl-gateway.container
asafbennatan: Fixed — added `flightctl-remote-access.service` to both `After=` and `Requires=` in the gateway unit.

deploy/podman/flightctl-remote-access/flightctl-remote-access-config/config.yaml.template
asafbennatan: Good point — the explicit external URL config will be wired up in the follow-up story that implements the actual session API, at which point both the Helm (host-based) and quadlets (path-based `/_/remote-access`) variants can be configured together with a real consumer of the value.

asafbennatan:
@kkyrazis adjusted to follow other services routing

internal/remote_access_server/server.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}()
+
+	go func() {
+		s.log.Printf("Remote-access HTTP listener on %s (disableTLS=%v)", s.httpListener.Addr(), s.cfg.Service.DisableTLS)
+		var err error
+		if s.cfg.Service.DisableTLS {
+			err = httpSrv.Serve(s.httpListener)
+		} else {
+			httpSrv.TLSConfig = s.agentTLSConfig
```
kkyrazis: do we want to use the agent's TLS config for this listener? Won't that require mTLS? Shouldn't this be using the first result from `crypto.TLSConfigForServer`

asafbennatan: fixed

deploy/helm/flightctl/templates/_helpers.tpl
```diff
@@ -728,6 +728,23 @@ Usage: {{- $result := include "flightctl.getImagebuilderApiDNSSans" . | fromJson
...
+  {{- $sans := list }}
+  {{- $baseDomain := include "flightctl.getBaseDomain" . }}
+  {{- $sans = append $sans (printf "agent-remote-access.%s" $baseDomain) }}
+  {{- $sans = append $sans "flightctl-remote-access" }}
+  {{- $sans = append $sans "flightctl-remote-access-agent" }}
+  {{- $sans = append $sans (printf "flightctl-remote-access.%s" .Release.Namespace) }}
+  {{- $sans = append $sans (printf "flightctl-remote-access-agent.%s" .Release.Namespace) }}
+  {{- $sans = append $sans (printf "flightctl-remote-access.%s.svc.cluster.local" .Release.Namespace) }}
+  {{- $sans = append $sans (printf "flightctl-remote-access-agent.%s.svc.cluster.local" .Release.Namespace) }}
+  {{- dict "sans" $sans | toJson -}}
```
galel12, line 746: The SAN helper has `agent-remote-access.{baseDomain}` but not `remote-access.{baseDomain}`. Since both routes use passthrough, the cert needs to cover both hostnames, otherwise clients hitting the service-api route will get a TLS mismatch.

asafbennatan: Fixed — added `remote-access.{baseDomain}` to `getRemoteAccessDNSSans` so the cert covers both the service-api route (`remote-access.<baseDomain>`) and the agent route (`agent-remote-access.<baseDomain>`).

galel12 (commented):
The remote-access container is built by `build-containers` (updated in this PR) but never loaded into the kind cluster, so `make deploy` will leave the remote-access pod in `ImagePullBackOff`.

Two files need a small update:

**`test/scripts/deploy_with_helm.sh`** line 101 -- add `remote-access` to the `kind_load_image` loop:
```bash
for suffix in periodic api worker alert-exporter alertmanager-proxy cli-artifacts db-setup telemetry-gateway imagebuilder-api imagebuilder-worker remote-access ; do
```

**`deploy/deploy.mk`** line 54 -- add `flightctl-remote-access-container` to the `deploy-helm` prerequisites so that running `make deploy-helm` directly also builds it:
```makefile
deploy-helm: ... flightctl-telemetry-gateway-container flightctl-remote-access-container
```

asafbennatan:
> The remote-access container is built by `build-containers` (updated in this PR) but never loaded into the kind cluster, so `make deploy` will leave the remote-access pod in `ImagePullBackOff`.
> 
> Two files need a small update:
> 
> **`test/scripts/deploy_with_helm.sh`** line 101 -- add `remote-access` to the `kind_load_image` loop:
> 
> ```shell
> for suffix in periodic api worker alert-exporter alertmanager-proxy cli-artifacts db-setup telemetry-gateway imagebuilder-api imagebuilder-worker remote-access ; do
> ```
> 
> **`deploy/deploy.mk`** line 54 -- add `flightctl-remote-access-container` to the `deploy-helm` prerequisites so that running `make deploy-helm` directly also builds it:
> 
> ```makefile
> deploy-helm: ... flightctl-telemetry-gateway-container flightctl-remote-access-container
> ```

fixed

kkyrazis (approved):

PR #3142, opened 2026-06-23
NO-ISSUE: Fix fleet rollout skipping re-render when renderedTemplateVersion is stale

### LINKED ISSUE #3141: Fleet rollout: device stuck OutOfDate when renderedTemplateVersion is stale after fleet reassignment
**Summary**

When a device is assigned to a fleet whose templateVersion matches the device's existing fleet-controller/templateVersion annotation and the fleet spec is identical to the device's current spec, updateDeviceToFleetTemplate silently skips the rollout. If the device's fleet-controller/renderedTemplateVersion annotation is stale (from a previous fleet assignment), the device is left permanently OutOfDate.

**Root cause**

`updateDeviceToFleetTemplate (internal/tasks/fleet_rollout.go)` contains this optimization:
```
if currentVersion == *templateVersion.Metadata.Name && domain.DeviceSpecsAreEqual(newDeviceSpec, *device.Spec) {
    // skip – nothing to do
    return depRefs, nil
}
```

The condition checks `fleet-controller/templateVersion` but ignores `fleet-controller/renderedTemplateVersion`. `IsUpdatedToFleetSpec` (used for the UI `OutOfDate` status) compares `renderedTemplateVersion` against the fleet's `templateVersion`. If these don't match, the device is `OutOfDate` regardless of what `templateVersion` says.

**Steps to reproduce**

- Device is in fleet A (templateVersion `v2`) → `templateVersion: v2`, `renderedTemplateVersion: v2`
- Device is removed from fleet A and assigned to fleet B (templateVersion `v1`, identical spec)
- `RolloutDevice` fires: `currentVersion (v1) == fleet.templateVersion (v1)` AND `DeviceSpecsAreEqual → true` → skip
- `renderedTemplateVersion` stays at v2
- `IsUpdatedToFleetSpec: v2 ≠ v1` → device shows OutOfDate forever

The stale `templateVersion: v1` / `renderedTemplateVersion: v2` state can arise whenever a device transitions through fleets with the same spec but different version counters (e.g. after a fleet is deleted and recreated).

**Expected behaviour**

If `renderedTemplateVersion` does not match the fleet's `templateVersion`, the rollout must proceed and trigger device-render, which will update `renderedTemplateVersion`.

mirokolar:
… Fixes #3141

- [ ] `release-1.2`
- [ ] `release-1.3`

**Fix**

Add `renderedTemplateVersion` to the skip condition. A rollout is skipped only when all three agree:

```
if currentVersion == *templateVersion.Metadata.Name &&
    currentRenderedVersion == *templateVersion.Metadata.Name &&
    domain.DeviceSpecsAreEqual(newDeviceSpec, *device.Spec) {
    return depRefs, nil
}
```

When `renderedTemplateVersion` is stale, the rollout proceeds, `ReplaceDevice` is called, device-render re-runs and updates `renderedTemplateVersion` to the correct value.

**Testing**

- Added table-driven unit test TestFleetRolloutsLogic_updateDeviceToFleetTemplate_SkipCondition covering:
  - Skip when templateVersion, renderedTemplateVersion, and spec all match
  - No skip when renderedTemplateVersion differs (stale) — new case
  - No skip when templateVersion differs
  - No skip when spec differs

### DIFF
--- a/internal/tasks/fleet_rollout.go
+++ b/internal/tasks/fleet_rollout.go
@@ -264,10 +264,13 @@ func (f FleetRolloutsLogic) RolloutDevice(ctx context.Context) error {
 
 func (f FleetRolloutsLogic) updateDeviceToFleetTemplate(ctx context.Context, device *domain.Device, templateVersion *domain.TemplateVersion, delayDeviceRender bool) ([]model.DependencyRef, error) {
 	currentVersion := ""
+	currentRenderedVersion := ""
 	if device.Metadata.Annotations != nil {
-		v, ok := (*device.Metadata.Annotations)[domain.DeviceAnnotationTemplateVersion]
-		if ok {
+		if v, ok := (*device.Metadata.Annotations)[domain.DeviceAnnotationTemplateVersion]; ok {
 			currentVersion = v
+		}
+		if v, ok := (*device.Metadata.Annotations)[domain.DeviceAnnotationRenderedTemplateVersion]; ok {
+			currentRenderedVersion = v
 		}
 	}
 	errs := []error{}
@@ -313,7 +316,7 @@ func (f FleetRolloutsLogic) updateDeviceToFleetTemplate(ctx context.Context, dev
 		return nil, fmt.Errorf("failed validating device spec for %s/%s: %w", f.orgId, *device.Metadata.Name, errors.Join(errs...))
 	}
 
-	if currentVersion == *templateVersion.Metadata.Name && domain.DeviceSpecsAreEqual(newDeviceSpec, *device.Spec) {
+	if currentVersion == *templateVersion.Metadata.Name && currentRenderedVersion == *templateVersion.Metadata.Name && domain.DeviceSpecsAreEqual(newDeviceSpec, *device.Spec) {
 		f.log.Debugf("Not rolling out device %s/%s because it is already at templateVersion %s", f.orgId, *device.Metadata.Name, *templateVersion.Metadata.Name)
 		return depRefs, nil
 	}
--- a/internal/tasks/fleet_rollout_test.go
+++ b/internal/tasks/fleet_rollout_test.go
@@ -395,6 +395,99 @@ func createTestDeviceWithLabels(name string, owner string, labels map[string]str
 	}
 }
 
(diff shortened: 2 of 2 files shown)


asafbennatan:
Thanks for the PR. One question: what happens if `device_render` fails with a permanent error? With the old code, fleet rollout would skip the device (zero extra work). With this fix, the stale `renderedTemplateVersion` causes fleet rollout to proceed on every sweep, which re-triggers `device_render`, which fails again — indefinitely. Should this distinguish between retryable and permanent render failures before deciding to proceed?

mirokolar:
You raise a valid concern. Looking at the code, device_render sets ConditionTypeDeviceSpecValid=False on failure but does not distinguish between transient errors (e.g. network timeout) and permanent ones (e.g. invalid spec). This makes it difficult to short-circuit retries in fleet_rollout without risking that a genuinely transient failure is never retried.
The trade-off is: the old behavior caused devices to be permanently stuck in OutOfDate with no recovery path even after the root cause was resolved. The new behavior triggers a re-render on each fleet rollout sweep (event-driven, not a tight polling loop), and the failure remains visible via DeviceSpecValid=False.
A proper fix would require device_render to either: (a) update renderedTemplateVersion even on failure to signal "already attempted this version", or (b) introduce a permanent-failure marker. Both are larger changes that go beyond the scope of this PR and would be better addressed separately.

asafbennatan:
> You raise a valid concern. Looking at the code, device_render sets ConditionTypeDeviceSpecValid=False on failure but does not distinguish between transient errors (e.g. network timeout) and permanent ones (e.g. invalid spec). This makes it difficult to short-circuit retries in fleet_rollout without risking that a genuinely transient failure is never retried. The trade-off is: the old behavior caused devices to be permanently stuck in OutOfDate with no recovery path even after the root cause was resolved. The new behavior triggers a re-render on each fleet rollout sweep (event-driven, not a tight polling loop), and the failure remains visible via DeviceSpecValid=False. A proper fix would require device_render to either: (a) update renderedTemplateVersion even on failure to signal "already attempted this version", or (b) introduce a permanent-failure marker. Both are larger changes that go beyond the scope of this PR and would be better addressed separately.

ok , lets move forward with this and address the churn later

asafbennatan (approved):
LGTM , we will need to address persistent rendering errors (EDM-4688)

PR #3143, opened 2026-06-23
EDM-3778: VM Agent Helm Application Tests SSH fix

Samudelacruz77:
The fix adds an SSH readiness check (30s timeout) at the start of each performRevertOperation attempt, before the createRevertVerificationData() call that was failing. If SSH isn't available, the attempt fails gracefully and the outer RevertToSnapshot retry loop (with exponential backoff) retries. This prevents the immediate "Connection reset by peer" failures that exhausted all 5 attempts in ~11 seconds.

### DIFF
--- a/test/e2e/applications/helm/helm_application_test.go
+++ b/test/e2e/applications/helm/helm_application_test.go
@@ -545,6 +545,20 @@ image-pruning:
 			})
 			Expect(err).ToNot(HaveOccurred())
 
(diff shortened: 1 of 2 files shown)


Samudelacruz77:
https://jenkins-csb-kniqe-auto.dno.corp.redhat.com/job/Private_Folders/job/sdelacru/job/job-runner-sdelacru/229/console

eldar101 (approved):

PR #3151, opened 2026-06-24
EDM-3530: Fix OIDC/OAuth2 auth provider URL normalization across all paths

asafbennatan:
## Summary

- **Root cause (EDM-3530)**: CLI OIDC login used `fmt.Sprintf` to build the discovery URL, producing double-slash paths (`https://idp.example.com//.well-known/...`) when the configured issuer had a trailing slash. Server-side code already used the correct `authprovider.DiscoveryURL` helper.
- **Additional normalization gaps** found during audit and fixed end-to-end:
  - `getTokenProxyURL` (all CLI providers) was using `fmt.Sprintf` for URL construction — replaced with `url.JoinPath` + error return
  - `introspectJWT` compared JWT issuer claims without normalizing either side — could reject valid tokens
  - `initOIDCAuth`/`initAAPAuth` computed a trimmed URL for logging but passed the raw spec (with trailing slash) to the auth constructor — dead code fixed
  - Dynamic AP resource write paths (`Create`, `Replace`, `Patch`) stored URL fields raw — trailing slashes in issuer would produce duplicate DB entries for the same IdP and confuse the DB unique index
  - `hasProviderChanged` compared raw issuer strings — a cosmetically different trailing slash caused unnecessary provider teardown/restart

## Changes

| Area | Change |
|---|---|
| `internal/cli/login/oidc.go` | Use `authprovider.DiscoveryURL` instead of `fmt.Sprintf` |
| `internal/cli/login/common.go` | `getTokenProxyURL` uses `url.JoinPath`, returns error |
| `internal/cli/login/{aap,oauth2,openshift}.go` | Handle new error from `getTokenProxyURL` |
| `internal/auth/authn/oauth2_auth.go` | Normalize both sides of JWT issuer comparison |
| `internal/auth/auth.go` | Apply `NormalizeIssuerURL`/`TrimSuffix` to spec copy before passing to constructors |
| `internal/service/authprovider.go` | Add `normalizeAuthProviderURLs`; call from all write paths |
| `api/core/v1beta1/validation.go` | Add `validateHTTPSURL`; validate URL fields in OIDC/OAuth2 `Validate()` |
| `internal/auth/authn/multiauth.go` | Normalize both sides of issuer comparison in `hasProviderChanged` |

## Test plan

- [x] `make unit-test` passes (2 pre-existing unrelated failures in console/Redis)
- [x] `internal/auth/provider` coverage: 26% → 96% (new `introspection_test.go`)
- [x] New tests: CLI login URL construction for all providers, server-side OIDC/OAuth2 auth, token proxy discovery, introspection config inference, AP validation
- [x] Existing tests: no regressions in `internal/auth/...`, `internal/service/...`, `api/core/v1beta1/...`

### DIFF
--- a/api/core/v1beta1/validation.go
+++ b/api/core/v1beta1/validation.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"reflect"
 	"regexp"
 	"slices"
@@ -1760,11 +1761,22 @@ func (a *AuthProvider) ValidateUpdate(ctx context.Context, oldObj *AuthProvider)
 	return allErrs
 }
 
+// validateHTTPSURL returns an error if the value is not a valid URL with a scheme and host.
+func validateHTTPSURL(fieldName, value string) error {
+	parsed, err := url.Parse(value)
+	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
+		return fmt.Errorf("%s must be a valid URL with scheme and host", fieldName)
+	}
+	return nil
+}
+
 func (o *OIDCProviderSpec) Validate(ctx context.Context, isUpdate bool) []error {
 	allErrs := []error{}
 
 	if o.Issuer == "" {
 		allErrs = append(allErrs, ErrIssuerRequired)
+	} else if err := validateHTTPSURL("issuer", o.Issuer); err != nil {
+		allErrs = append(allErrs, err)
 	}
 	if o.ClientId == "" {
 		allErrs = append(allErrs, ErrClientIdRequired)
@@ -1784,15 +1796,26 @@ func (o *OAuth2ProviderSpec) Validate(ctx context.Context, isUpdate bool) []erro
 
 	if o.AuthorizationUrl == "" {
 		allErrs = append(allErrs, ErrAuthorizationUrlRequired)
+	} else if err := validateHTTPSURL("authorizationUrl", o.AuthorizationUrl); err != nil {
+		allErrs = append(allErrs, err)
 	}
 	if o.TokenUrl == "" {
 		allErrs = append(allErrs, ErrTokenUrlRequired)
+	} else if err := validateHTTPSURL("tokenUrl", o.TokenUrl); err != nil {
+		allErrs = append(allErrs, err)
 	}
 	if o.UserinfoUrl == "" {
 		allErrs = append(allErrs, ErrUserinfoUrlRequired)
+	} else if err := validateHTTPSURL("userinfoUrl", o.UserinfoUrl); err != nil {
+		allErrs = append(allErrs, err)
 	}
 	if o.ClientId == "" {
 		allErrs = append(allErrs, ErrClientIdRequired)
+	}
+	if o.Issuer != nil && *o.Issuer != "" {
+		if err := validateHTTPSURL("issuer", *o.Issuer); err != nil {
+			allErrs = append(allErrs, err)
+		}
 	}
 
 	// Validate introspection field is present
--- a/api/core/v1beta1/validation.go
+++ b/api/core/v1beta1/validation.go
@@ -1761,8 +1761,8 @@ func (a *AuthProvider) ValidateUpdate(ctx context.Context, oldObj *AuthProvider)
 	return allErrs
 }
 
-// validateHTTPSURL returns an error if the value is not a valid URL with a scheme and host.
-func validateHTTPSURL(fieldName, value string) error {
+// validateAbsoluteURL returns an error if the value is not a valid absolute URL with a scheme and host.
+func validateAbsoluteURL(fieldName, value string) error {
 	parsed, err := url.Parse(value)
 	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
 		return fmt.Errorf("%s must be a valid URL with scheme and host", fieldName)
@@ -1775,7 +1775,7 @@ func (o *OIDCProviderSpec) Validate(ctx context.Context, isUpdate bool) []error 
 
 	if o.Issuer == "" {
 		allErrs = append(allErrs, ErrIssuerRequired)
-	} else if err := validateHTTPSURL("issuer", o.Issuer); err != nil {
+	} else if err := validateAbsoluteURL("issuer", o.Issuer); err != nil {
 		allErrs = append(allErrs, err)
 	}
 	if o.ClientId == "" {
@@ -1796,24 +1796,24 @@ func (o *OAuth2ProviderSpec) Validate(ctx context.Context, isUpdate bool) []erro
 
 	if o.AuthorizationUrl == "" {
 		allErrs = append(allErrs, ErrAuthorizationUrlRequired)
-	} else if err := validateHTTPSURL("authorizationUrl", o.AuthorizationUrl); err != nil {
+	} else if err := validateAbsoluteURL("authorizationUrl", o.AuthorizationUrl); err != nil {
 		allErrs = append(allErrs, err)
 	}
 	if o.TokenUrl == "" {
 		allErrs = append(allErrs, ErrTokenUrlRequired)
-	} else if err := validateHTTPSURL("tokenUrl", o.TokenUrl); err != nil {
+	} else if err := validateAbsoluteURL("tokenUrl", o.TokenUrl); err != nil {
 		allErrs = append(allErrs, err)
 	}
 	if o.UserinfoUrl == "" {
 		allErrs = append(allErrs, ErrUserinfoUrlRequired)
-	} else if err := validateHTTPSURL("userinfoUrl", o.UserinfoUrl); err != nil {
+	} else if err := validateAbsoluteURL("userinfoUrl", o.UserinfoUrl); err != nil {
 		allErrs = append(allErrs, err)
 	}
 	if o.ClientId == "" {
 		allErrs = append(allErrs, ErrClientIdRequired)
 	}
 	if o.Issuer != nil && *o.Issuer != "" {
-		if err := validateHTTPSURL("issuer", *o.Issuer); err != nil {
+		if err := validateAbsoluteURL("issuer", *o.Issuer); err != nil {
 			allErrs = append(allErrs, err)
 		}
 	}
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -85,8 +85,11 @@ func initOAuth2Auth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMid
 	return authNProvider, nil
 }
 func initOIDCAuth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMiddleware, error) {
-	oidcUrl := strings.TrimSuffix(cfg.Auth.OIDC.Issuer, "/")
-	log.Infof("OIDC auth enabled: %s", oidcUrl)
+	normalizedIssuer, err := authprovider.NormalizeIssuerURL(cfg.Auth.OIDC.Issuer)
+	if err != nil {
+		return nil, fmt.Errorf("invalid OIDC issuer URL: %w", err)
+	}
+	log.Infof("OIDC auth enabled: %s", normalizedIssuer)
 
 	providerName := "oidc"
 	metadata := api.ObjectMeta{
@@ -96,7 +99,9 @@ func initOIDCAuth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMiddl
 		},
 	}
 
-	authNProvider, err := authn.NewOIDCAuth(metadata, *cfg.Auth.OIDC, getTlsConfig(cfg), log)
+	spec := *cfg.Auth.OIDC
+	spec.Issuer = normalizedIssuer
+	authNProvider, err := authn.NewOIDCAuth(metadata, spec, getTlsConfig(cfg), log)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create OIDC AuthN: %w", err)
 	}
@@ -112,7 +117,9 @@ func initAAPAuth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMiddle
 		Name: &providerName,
 	}
 
-	authNProvider, err := authn.NewAapGatewayAuth(metadata, *cfg.Auth.AAP, getTlsConfig(cfg))
+	spec := *cfg.Auth.AAP
+	spec.ApiUrl = gatewayUrl
+	authNProvider, err := authn.NewAapGatewayAuth(metadata, spec, getTlsConfig(cfg))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create AAP Gateway AuthN: %w", err)
 	}
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
-	"strings"
 
 	api "github.com/flightctl/flightctl/api/core/v1beta1"
 	"github.com/flightctl/flightctl/internal/auth/authn"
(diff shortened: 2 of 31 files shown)


api/core/v1beta1/validation.go
```diff
@@ -1736,11 +1737,22 @@ func (a *AuthProvider) ValidateUpdate(ctx context.Context, oldObj *AuthProvider)
 	return allErrs
 }
 
+// validateHTTPSURL returns an error if the value is not a valid URL with a scheme and host.
+func validateHTTPSURL(fieldName, value string) error {
+	parsed, err := url.Parse(value)
+	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
+		return fmt.Errorf("%s must be a valid URL with scheme and host", fieldName)
+	}
```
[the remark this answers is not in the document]
asafbennatan: Deliberate choice — `validateAbsoluteURL` accepts any absolute URL (any scheme + host) rather than enforcing `https://` specifically. The rename from `validateHTTPSURL` to `validateAbsoluteURL` was intentional and was part of an earlier code review round to name the function accurately for what it actually does.

The rationale: auth providers may legitimately use `http://` in development, test, or airgapped environments. Enforcing HTTPS at the API validation layer would break these use cases and is a broader policy decision beyond the scope of this bug fix. If the project wants to add HTTPS enforcement in the future, it can be done as a separate, targeted change with a migration story.

internal/auth/authn/multiauth.go
```diff
@@ -383,7 +383,9 @@ func (m *MultiAuth) hasProviderChanged(existingMiddleware common.AuthNMiddleware
 		}
 
 		// Compare all fields including client secret
-		if existingOidcSpec.Issuer != newOidcSpec.Issuer {
+		existingNormalizedIssuer, _ := authprovider.NormalizeIssuerURL(existingOidcSpec.Issuer)
+		newNormalizedIssuer, _ := authprovider.NormalizeIssuerURL(newOidcSpec.Issuer)
+		if existingNormalizedIssuer != newNormalizedIssuer {
```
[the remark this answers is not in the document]
asafbennatan: This was already addressed in a subsequent commit (`53789bb`). The current code explicitly captures both normalization errors and uses a fail-open approach: if either normalization fails, the provider is treated as changed (triggering a reload). This is safer than the raw-string fallback because two distinct malformed URLs that happen to normalize to the same empty string would still be correctly treated as different. The fix is on the branch now.

internal/cli/login/common.go
```diff
@@ -72,9 +73,13 @@ type ValidateArgs struct {
...
 // getTokenProxyURL returns the Flight Control API server's token proxy endpoint
-// for the given provider name
-func getTokenProxyURL(apiServerURL, providerName string) string {
-	return fmt.Sprintf("%s/api/v1/auth/%s/token", apiServerURL, providerName)
+// for the given provider name. Returns an error if apiServerURL is not a valid URL.
+func getTokenProxyURL(apiServerURL, providerName string) (string, error) {
+	u, err := url.Parse(apiServerURL)
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return "", fmt.Errorf("invalid API server URL %q: %w", apiServerURL, err)
+	}
```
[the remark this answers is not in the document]
asafbennatan: Fixed in commit `53789bb`. The guard is now split into two separate returns: one wrapping the actual `url.Parse` error with `%w`, and a second using a plain `fmt.Errorf` string (no `%w`) for the missing scheme/host case. The CLI now emits a clear actionable message.

internal/cli/login/oidc_test.go
```diff
@@ -0,0 +1,181 @@
...
+		if validPaths[r.URL.Path] {
+			w.Header().Set("Content-Type", "application/json")
+			discovery := OIDCDiscoveryResponse{
+				Issuer:                           server.URL,
+				AuthorizationEndpoint:            server.URL + "/authorize",
+				TokenEndpoint:                    server.URL + "/token",
+				JwksUri:                          server.URL + "/jwks",
+				SubjectTypesSupported:            []string{"public"},
+				ResponseTypesSupported:           []string{"code"},
+				IdTokenSigningAlgValuesSupported:  []string{"RS256"},
```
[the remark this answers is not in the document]
asafbennatan: Fixed in commit `de3841a` — `gci write` was run on the file and formatting is now correct.

internal/cli/login/providers_url_test.go
```diff
@@ -0,0 +1,274 @@
...
+			}
+
+			_, err := o.getOAuth2Client("http://localhost:8080/callback")
+			require.NoError(t, err, "getOAuth2Client should succeed for API server URL %q", apiServerURL)
+
+			mu.Lock()
+			paths := append([]string{}, *requestedPaths...)
+			mu.Unlock()
+
+			assertNoDoubleSlash(t, paths, "OAuth2 token proxy URL")
```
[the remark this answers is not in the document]
asafbennatan: Fixed in commit `6807e0c`. The tests now call `getTokenProxyURL(apiServerURL, providerName)` directly and assert that the returned URL's _path_ component (not the full URL string, to avoid matching `://`) contains no double slashes. The server setup is kept for `osincli.NewClient` to have a valid `AuthorizeUrl`, but the vacuous path-recording infrastructure has been removed.

internal/service/authprovider/normalize_test.go
```diff
@@ -0,0 +1,316 @@
...
+			}
+		})
+	}
+}
+
+func TestNormalizeAuthProviderURLs_NonURLTypes(t *testing.T) {
+	t.Run("When AAP provider type it should be a no-op", func(t *testing.T) {
+		spec := domain.AuthProviderSpec{}
+		_ = spec.FromAapProviderSpec(domain.AapProviderSpec{
+			ProviderType:     domain.Aap,
```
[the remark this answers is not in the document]
asafbennatan: Fixed in commit `6807e0c`. All `_ = spec.From*ProviderSpec(...)` and `_ = i.FromRfc7662IntrospectionSpec(...)` calls have been replaced with `require.NoError(t, ...)`. The helper closure also now takes `t *testing.T` so it can call `require.NoError` directly.

internal/service/authprovider/normalize_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+func TestApplyAuthProviderDefaults_OAuth2(t *testing.T) {
+	rfc7662 := func(t *testing.T) *domain.OAuth2Introspection {
+		i := &domain.OAuth2Introspection{}
+		require.NoError(t, i.FromRfc7662IntrospectionSpec(domain.Rfc7662IntrospectionSpec{
+			Type: domain.IntrospectionTypeRfc7662,
+			Url:  "https://idp.example.com/introspect",
+		}))
+		return i
+	}
```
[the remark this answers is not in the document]
asafbennatan: Fixed in commit `c6bb6ddd`. `t.Helper()` added as the first statement in the `rfc7662` closure so failures are reported at the calling subtest rather than inside the helper.

internal/service/authprovider/handler.go
```diff
@@ -47,6 +47,61 @@ func sanitizeSchemaError(err error) string {
...
+		oidcSpec, err := spec.AsOIDCProviderSpec()
+		if err != nil {
+			return fmt.Errorf("invalid OIDC provider spec: %w", err)
+		}
+		if oidcSpec.Issuer != "" {
+			normalized, err := provider.NormalizeIssuerURL(oidcSpec.Issuer)
+			if err != nil {
+				return fmt.Errorf("invalid OIDC issuer URL: %w", err)
+			}
+			oidcSpec.Issuer = normalized
```
galel12: Do we need a one-shot migration / startup rewrite of issuer/authorizationUrl (and maybe userinfoUrl)? Or at least a documented upgrade note + admin query?

Cause legacy DB rows can still duplicate logical IdPs. Unique index is exact string match, so `https://idp.example.com/` and `https://idp.example.com` won't collide.

internal/service/authprovider/handler.go
```diff
@@ -47,6 +47,61 @@ func sanitizeSchemaError(err error) string {
...
+			}
+			oauth2Spec.AuthorizationUrl = normalized
+		}
+		if oauth2Spec.Issuer != nil && *oauth2Spec.Issuer != "" {
+			normalized, err := provider.NormalizeIssuerURL(*oauth2Spec.Issuer)
+			if err != nil {
+				return fmt.Errorf("invalid OAuth2 issuer URL: %w", err)
+			}
+			oauth2Spec.Issuer = &normalized
+		}
```
galel12: We normalize authorizationUrl/issuer here but not tokenUrl/userinfoUrl, and the OAuth2 unique index is on (userinfoUrl, clientId). Same trailing-slash duplicate class stays open. Should canonicalize cover those too?

internal/auth/auth.go
```diff
@@ -96,7 +115,9 @@ func initOIDCAuth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMiddl
 		},
 	}
 
-	authNProvider, err := authn.NewOIDCAuth(metadata, *cfg.Auth.OIDC, getTlsConfig(cfg), log)
+	spec := *cfg.Auth.OIDC
+	spec.Issuer = normalizedIssuer
```
galel12: Static config mutates the spec to the normalized form before constructing middleware. Dynamic path (createOIDCAuthFromProvider / createOAuth2AuthFromProvider) still passes the DB spec raw. Should that path canonicalize once too so stored form == runtime form?

api/core/v1beta1/validation.go
```diff
@@ -1760,11 +1761,22 @@ func (a *AuthProvider) ValidateUpdate(ctx context.Context, oldObj *AuthProvider)
 	return allErrs
 }
 
+// validateAbsoluteURL returns an error if the value is not a valid absolute URL with a scheme and host.
+func validateAbsoluteURL(fieldName, value string) error {
+	parsed, err := url.Parse(value)
+	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
+		return fmt.Errorf("%s must be a valid URL with scheme and host", fieldName)
+	}
+	return nil
```
galel12, line 1770: Saw that accepting any scheme was deliberate. Just checking it's an explicit product call. These endpoints get fetched by the server, so http:// widens SSRF surface for whoever can create AuthProviders. If HTTP is intentional for air-gapped/dev, worth documenting. Nested jwksUrl / introspection issuer also skip this check.

internal/service/authprovider/handler.go
```diff
@@ -47,6 +47,61 @@ func sanitizeSchemaError(err error) string {
...
+	}
+
+	switch discriminator {
+	case string(domain.Oidc):
+		oidcSpec, err := spec.AsOIDCProviderSpec()
+		if err != nil {
+			return fmt.Errorf("invalid OIDC provider spec: %w", err)
+		}
+		if oidcSpec.Issuer != "" {
+			normalized, err := provider.NormalizeIssuerURL(oidcSpec.Issuer)
```
galel12: NormalizeIssuerURL only strips a trailing path slash. Host case etc. are left alone, so `https://IdP.Example.com` and `https://idp.example.com` still count as different for the unique index. Fine if the policy is trailing-slash only, but uniqueness/docs should say that.

internal/auth/auth.go
```diff
@@ -112,7 +133,9 @@ func initAAPAuth(cfg *config.Config, log logrus.FieldLogger) (common.AuthNMiddle
 		Name: &providerName,
 	}
 
-	authNProvider, err := authn.NewAapGatewayAuth(metadata, *cfg.Auth.AAP, getTlsConfig(cfg))
+	spec := *cfg.Auth.AAP
+	spec.ApiUrl = gatewayUrl
```
galel12: nit: OIDC/OAuth2 moved onto NormalizeIssuerURL, but AAP (and K8s/OpenShift nearby) still use TrimSuffix. Close enough for now, just another almost-the-same path.

galel12 (commented):
Discovery fix looks right. Main open question is whether we also need to clean up existing DB rows so the unique indexes match what we normalize at runtime. Details inline.

internal/service/authprovider/handler.go
asafbennatan: Added a one-shot DB migration (`normalize_auth_provider_urls_v1` in `flightctl-db-migrate`) that rewrites stored OIDC/OAuth2 URL fields to the canonical form. If two rows would collide after normalization (slash or case variants for the same uniqueness key), the migration fails with a clear error so an admin can resolve the duplicate before upgrade completes.

internal/service/authprovider/handler.go
asafbennatan: Done — `normalizeAuthProviderURLs` / `NormalizeOAuth2ProviderSpecURLs` now canonicalize `tokenUrl` and `userinfoUrl` as well (same rules as `authorizationUrl`/`issuer`), including in the one-shot migration so the `(userinfoUrl, clientId)` unique index matches write-path normalization.

internal/auth/auth.go
asafbennatan: Agreed — `createOIDCAuthFromProvider` / `createOAuth2AuthFromProvider` now normalize the spec before constructing middleware, so dynamic DB-loaded providers match the static init path (and legacy rows still get a canonical runtime form even before migration).

api/core/v1beta1/validation.go
asafbennatan: Confirmed we did not introduce non-HTTPS acceptance: on `main` these fields only had required/non-empty checks (no scheme restriction), so `http://` was already allowed. This PR only adds absolute-URL shape validation (scheme + host). Leaving that behavior unchanged for airgapped/dev. Nested `jwksUrl` / introspection URL validation is a follow-up.

internal/service/authprovider/handler.go
asafbennatan: Agreed — `NormalizeIssuerURL` now lowercases scheme/host/path/query/fragment in addition to stripping a trailing path slash. The one-shot migration applies the same rules, so case-only issuer variants collapse under the unique indexes.

internal/auth/auth.go
asafbennatan: Fixed — AAP `ApiUrl`, K8s `ApiUrl`, and OpenShift `ClusterControlPlaneUrl`/`AuthorizationUrl` static init now go through `NormalizeIssuerURL` as well.

internal/auth/provider/introspection.go
```diff
@@ -124,7 +126,12 @@ func NormalizeIssuerURL(issuer string) (string, error) {
 	if parsed.Scheme == "" || parsed.Host == "" {
 		return "", errors.New("issuer URL must have scheme and host")
 	}
-	parsed.Path = strings.TrimSuffix(parsed.Path, "/")
+	parsed.Scheme = strings.ToLower(parsed.Scheme)
+	parsed.Host = strings.ToLower(parsed.Host)
+	parsed.Path = strings.ToLower(strings.TrimSuffix(parsed.Path, "/"))
```
galel12: Lowercasing scheme/host is right, but path/query/fragment shouldn't be. OIDC issuer paths are case-sensitive (`/realms/MyRealm` vs `/realms/myrealm`), and this also rewrites stored rows via the migration + discovery fetches. Can we keep path/query/fragment as-is and only canonicalize scheme + host + trailing slash?

internal/auth/authn/multiauth.go
```diff
@@ -456,10 +459,18 @@ func (m *MultiAuth) hasProviderChanged(existingMiddleware common.AuthNMiddleware
...
+			newNormalizedIssuer, newNormErr := authprovider.NormalizeIssuerURL(*newOauth2Spec.Issuer)
+			if existingNormErr != nil || newNormErr != nil || existingNormalizedIssuer != newNormalizedIssuer {
+				return true, nil
+			}
 		}
-		if existingOauth2Spec.AuthorizationUrl != newOauth2Spec.AuthorizationUrl {
+		// If normalization fails for either URL (malformed), treat as changed.
+		existingNormalizedAuthzURL, existingAuthzNormErr := authprovider.NormalizeIssuerURL(existingOauth2Spec.AuthorizationUrl)
+		newNormalizedAuthzURL, newAuthzNormErr := authprovider.NormalizeIssuerURL(newOauth2Spec.AuthorizationUrl)
+		if existingAuthzNormErr != nil || newAuthzNormErr != nil || existingNormalizedAuthzURL != newNormalizedAuthzURL {
```
galel12, line 473: nit: authorizationUrl/issuer compare via NormalizeIssuerURL, but TokenUrl/UserinfoUrl just below are still raw string compare. Write path normalizes them now, so this is mostly consistency. Same normalize-or-treat-as-changed pattern as authz would keep the change detector aligned.

galel12 (commented):
Two leftover notes from the broader normalize change.

internal/auth/provider/introspection.go
asafbennatan: Done — `NormalizeIssuerURL` now only lowercases scheme + host, preserves path/query/fragment case, and still strips a trailing path slash. Migration + tests updated accordingly (e.g. `/Realm/Master` stays as-is).

internal/auth/authn/multiauth.go
asafbennatan: Done — `TokenUrl` / `UserinfoUrl` now use the same normalize-or-treat-as-changed pattern as `AuthorizationUrl`. Added unit coverage in `has_provider_changed_test.go` for host-case/trailing-slash equivalence and for path-case being significant.

galel12 (approved):
LGTM

galel12 (approved):

PR #3166, opened 2026-06-25
NO-ISSUE: fix: flightctl-mirror-images fails when installed via RPM

aekubacki:
## Summary

- `flightctl-mirror-images` fails immediately when run from an RPM installation with `Error: required file not found: /usr/deploy/helm/helm-chart-opts.yaml` because `repoRoot()` assumed the binary lived under a source checkout rather than `/usr/bin`
- RPM spec now installs the required data files (`helm-chart-opts.yaml`, `Chart.yaml`, `images.yaml` files, `flightctl.spec`) to `/usr/share/flightctl/`
- `repoRoot()` now checks `/usr/share/flightctl` first (RPM install path) and falls back to the binary-relative path for development source checkouts

## Workaround for existing installs

```bash
git clone --depth 1 --branch v1.2.0 https://github.com/flightctl/flightctl.git /tmp/flightctl-src

HELM_CHART_OPTS=/tmp/flightctl-src/deploy/helm/helm-chart-opts.yaml \
CHART_YAML=/tmp/flightctl-src/deploy/helm/flightctl/Chart.yaml \
RPM_SPEC=/tmp/flightctl-src/packaging/rpm/flightctl.spec \
OBS_IMAGES_RHEL9=/tmp/flightctl-src/packaging/images/rhel9/images.yaml \
OBS_IMAGES_RHEL10=/tmp/flightctl-src/packaging/images/rhel10/images.yaml \
flightctl-mirror-images --variant rhem-el9 --bundle ~/flightctl-bundle.tar.gz
```

## Test plan

- [ ] Build RPM from this branch and install `flightctl-cli`
- [ ] Verify `/usr/share/flightctl/deploy/helm/helm-chart-opts.yaml` is present after install
- [ ] Run `flightctl-mirror-images --variant rhem-el9 --bundle /tmp/test.tar.gz` without setting any env vars — should succeed
- [ ] Verify development workflow still works: `make build-mirror-images && ./bin/flightctl-mirror-images --variant rhem-el9 --dest-registry localhost:5000`


packaging/rpm/flightctl.spec
```diff
@@ -303,6 +303,19 @@ fi
     cp bin/flightctl-backup %{buildroot}/usr/bin
     cp bin/flightctl-restore %{buildroot}/usr/bin
     cp bin/flightctl-mirror-images %{buildroot}/usr/bin
+    mkdir -p %{buildroot}/usr/share/flightctl/deploy/helm/flightctl
```
kkyrazis: I don't think we should be including these as part of the rpm package. I thought the intention of the tool was that it had to be used from the flightctl repo?

aekubacki: I think they should be included because the intended workflow for the disconnected install is that the prep/staging machine (the internet-connected one) installs flightctl-cli  via RPM and then runs flightctl-mirror-images directly.  There's no expectation that they clone the source repo.

Happy to discuss more

kkyrazis: the previous comment in the code:
```
// repoRoot attempts to resolve the flightctl repository root relative to the
// running binary's location.
//
// The binary is expected to be built from scripts/air-gap/flightctl-mirror-images/, so
// the repo root is three directories above the binary:
//
//	bin/flightctl-mirror-images → bin/ → repo root      (when run via `go build -o bin/`)
//
// If os.Executable() fails we fall back to the current working directory so
// that `go run ./scripts/air-gap/flightctl-mirror-images` still works from the repo root.
```

This seems like a pretty large change from that assumption. 

I just don't think we should be shipping a bunch of build artifacts as part of the rpms.

kkyrazis: I've been thinking about this more and I think there's a cleaner approach than shipping the raw repo files in the RPM.

The core observation is that all 7 files the tool reads at runtime (`helm-chart-opts.yaml`, `Chart.yaml`, 4x `images.yaml`, `flightctl.spec`) are hand-maintained source-of-truth configs that are inherently version-tied. An RPM built from `v1.2.0` should always use the data from that same tag — so resolving these at runtime from a source checkout (or from shipped copies of source files) is doing work that could be done once at build time.

Two options that would make the binary self-contained without shipping repo artifacts:

**Option A: Build-time manifest generation.** A preprocessing step during `make build-mirror-images` reads the 7 source files, resolves the per-variant image lists (merging helm-chart-opts + obs images, filtering obs-only entries, extracting RPM `Requires:`), and writes a single `mirror-manifest.json` with a clean schema. That manifest gets either `go:embed`'d into the binary or installed as a single well-known file. The tool's 4 parsing functions (`ParseHelmChartOpts`, `ParseObsImages`, `ReadAppVersion`, `ParseRPMRequires`) are all deterministic transforms over static files — none require runtime state, so this is straightforward.

**Option B: `go:embed` the raw files.** Simpler mechanically — embed the 7 files directly into the binary via `//go:embed` directives. The parsing logic stays as-is, but reads from the embedded FS instead of disk. Less architectural cleanup than option A, but still eliminates the runtime file dependency and the `repoRoot()` heuristic entirely.

Both approaches mean: no repo checkout needed for RPM users, no `repoRoot()` priority bugs for developers, no repo directory structure leaking into the RPM, and version-matching by construction. `--tag-override` and the env-var overrides can remain as escape hatches. Option A is more work but produces a cleaner separation; option B is a smaller diff.

kkyrazis:
## Review feedback

The build-time manifest generation approach is the right architecture — the binary is now self-contained, version-matched by construction, and no repo artifacts leak into the RPM. Here are some findings from the updated diff:

---

### 1. No CI staleness check for `embedded_data_generated.go`

**Makefile** — The `generate` target (used by CI lint via `make generate && git diff --exit-code`) runs only `go generate` — it does **not** include `generate-mirror-embed`. A contributor can modify any of the 7 source YAML files (helm-chart-opts, Chart.yaml, images.yaml, flightctl.spec), commit the change, and CI will merge a stale embedded manifest without complaint. The RPM build is fine (it calls `build-mirror-images` → `generate-mirror-embed`), but the committed file can silently drift.

**Fix:** Add `generate-mirror-embed` as a prerequisite of the `generate` target, or add a dedicated CI step.

---

### 2. Stdout redirect truncates output file on generator failure

**Makefile** — The rule `go run ... > $(MIRROR_EMBED_OUT)` truncates the output file *before* the generator starts. If the generator fails mid-run (malformed YAML, missing dependency), the file is left empty or partial, and the subsequent `go build` fails with a confusing compile error rather than the actual generator error.

**Fix:** Use a temp-file-then-rename pattern:
```make
go run ./scripts/air-gap/generate-embed/ $(MIRROR_EMBED_SOURCES) > $(MIRROR_EMBED_OUT).tmp && mv $(MIRROR_EMBED_OUT).tmp $(MIRROR_EMBED_OUT)
```

---

### 3. `generate-mirror-embed` is `.PHONY`, defeating timestamp-based prerequisites

**Makefile** — The target lists `$(MIRROR_EMBED_SOURCES)` as prerequisites (suggesting it should only re-run when sources change), but being `.PHONY` means Make always runs the generator regardless of timestamps. Either remove it from `.PHONY` and use a proper file target (with `$(MIRROR_EMBED_OUT)` as the target name), or accept the always-run cost. If it's always-run, the atomic write pattern from finding #2 matters more.

---

### 4. `--tag-override` doesn't apply to `"latest"` tags in custom `MIRROR_MANIFEST` files

**`build_manifest.go:42`** — The old code applied the effective tag when `tag == "" || tag == "latest"`. The new `resolveVariant` only checks `tag == ""`. The build-time generator normalizes `"latest"` to `""` via `canonicalTag()`, so the *embedded* manifest is fine. But a user-provided `MIRROR_MANIFEST` JSON file could have `"tag": "latest"` — in that case `--tag-override` would not apply, and the image would be mirrored as `:latest` instead of the overridden tag.

**Fix:** Either add `|| tag == "latest"` to the condition in `resolveVariant`, or document that custom manifests must use empty-string tags for images that should track the effective version.

---

### 5. `parseRPMRequires` doesn't check `scanner.Err()` after the scan loop

**`generate-embed/main.go:305`** — The old `ParseRPMRequires` in `parser.go` checked `scanner.Err()` and returned an error on scanner failures (e.g. lines exceeding the 64KB token limit). The new build-time version silently drops truncated input, potentially producing an incomplete RPM list baked into the manifest.

---

### 6. registry.redhat.io warning for community variants dropped

**`generate-embed/main.go:252`** — The old `ParseObsImages` warned when community variants referenced `registry.redhat.io` images (`"RPM image requires downstream registry access"`). The new generator's `resolveImages` has no equivalent check. If a community variant's `images.yaml` mistakenly references a downstream registry image, users would get opaque skopeo failures at mirror time with no diagnostic hint. Worth preserving this guard rail in the generator.

kkyrazis (approved):

PR #3170, opened 2026-06-26
NO-ISSUE: increase e2e VM memory from 512 to 2048 MiB

amir-yogev-gh:
## Summary
- Increases QEMU VM memory in `domain-template.xml` from 512 MiB to 2048 MiB
- Fixes all e2e test failures on `release-1.0` caused by GRUB being unable to allocate the initrd with only 512 MiB of RAM
- This was already fixed on `main` (via EDM-2265) with a templated/dynamic approach defaulting to 2048 MiB, but never backported to `release-1.0`

## Root cause
GRUB error: `../../grub-core/loader/i386/efi/linux.c:159:can't allocate initrd.`
The VM enters a boot loop, SSH never becomes ready, and the 1-minute SSH timeout is hit — causing the BeforeSuite to fail and all 19 tests to be skipped on every node.

## Test plan
- [ ] All 4 e2e-tests jobs pass in CI

Assisted-by: Claude <noreply@anthropic.com>

### DIFF
--- a/test/e2e/parametrisable_templates/parametrisable_templates_test.go
+++ b/test/e2e/parametrisable_templates/parametrisable_templates_test.go
@@ -341,7 +341,7 @@ var (
 	repoTestUrl           = "https://github.com/flightctl/flightctl-demos"
 	deviceAlias           = "base"
 	branchTargetRevision  = "demo"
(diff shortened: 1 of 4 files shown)


asafbennatan (approved):
LGTM

PR #3171, opened 2026-06-26
EDM-4429: flightctl-backup fails on quadlet/podman deployments with external database

Samudelacruz77:
Files changed:

  1. internal/backup/deployer.go — Two changes:
    - Added serviceConfigDBType() — parses db.type directly from the raw service-config.yaml (which uses db: key,
  not database:)
    - Updated isInternalDB() — added early return false when cfg.Database.Type == "external" (defense-in-depth for
  per-service configs where the JSON tag maps correctly)
  2. internal/backup/podman_deployer.go — Updated BackupDatabase() to read the raw service-config.yaml and check
  serviceConfigDBType() before falling through to config.Load/isInternalDB
  3. internal/backup/deployer_test.go — Added tests for serviceConfigDBType() (7 cases) and isInternalDB with
  db.type field (5 new cases)
  4. internal/backup/podman_deployer_test.go — Added TestPodmanDeployer_BackupDatabase_ExternalDB_RawServiceConfig
  that writes a real service-config.yaml with db.type: "external" and verifies ErrExternalDatabase is returned.
  Updated existing missing-config test for new error message.


internal/backup/deployer.go
```diff
@@ -138,15 +139,36 @@ func ValidateDeploymentType(s string) error {
...
+func serviceConfigDBType(data []byte) string {
+	var root struct {
+		DB struct {
+			Type string `json:"type"`
+		} `json:"db"`
+	}
+	if err := yaml.Unmarshal(data, &root); err != nil {
+		return ""
+	}
+	return strings.ToLower(strings.TrimSpace(root.DB.Type))
```
eldar101: Can you add a log line so we know what is being returned? for debugging purposes.

internal/backup/podman_deployer.go
```diff
@@ -108,12 +108,28 @@ func (p *PodmanDeployer) Type() DeploymentType {
...
+
+	cfg, err := config.LoadFromBytes(rawCfg)
+	if err != nil {
+		return fmt.Errorf("failed to parse service configuration from %s: %w", p.serviceConfigPath, err)
+	}
+
+	// Defense-in-depth for per-service configs where the "database" key maps
+	// correctly to Config.Database.
 	if !isInternalDB(cfg) {
 		return ErrExternalDatabase
```
eldar101, line 137: Makes sense. I’d log the normalized db.type at the BackupDatabase caller after serviceConfigDBType(rawCfg), rather than inside the helper, since that function has no logger and we should avoid logging the raw config.

internal/config/config.go
```diff
@@ -973,8 +973,12 @@ func Load(cfgFile string) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("reading config file: %v", err)
 	}
+	return LoadFromBytes(contents)
```
asafbennatan: this is unnecessary change , this file should not have any changes

internal/backup/deployer.go
```diff
@@ -138,15 +139,36 @@ func ValidateDeploymentType(s string) error {
...
+}
+
 // isInternalDB returns true if the database is internal (managed by FlightCtl deployment).
-// Internal databases have hostnames: localhost, 127.0.0.1, flightctl-db, or flightctl-db.<namespace>
+// When db.type is explicitly set to "external" in the config, returns false regardless of hostname.
+// Otherwise checks hostnames: localhost, 127.0.0.1, flightctl-db, or flightctl-db.<namespace>
 // patterns (including full DNS like flightctl-db.flightctl-internal.svc.cluster.local).
 // External databases (any other hostname) must be backed up separately by the user.
 func isInternalDB(cfg *config.Config) bool {
```
asafbennatan: is internal is redundant as you are calling the serviceconfigdbtype from podman_deployer directly - remove it - we dont want this as a fallback logic

internal/backup/podman_deployer.go
asafbennatan: cfg is loaded from service-config.yaml ,.it does not contain a database key , so how is this used exactly ?

internal/backup/podman_deployer.go
```diff
@@ -142,18 +155,26 @@ func (p *PodmanDeployer) BackupDatabase(ctx context.Context, outputDir string) e
...
-	// Execute pg_dump inside the container with password from stdin and safely escaped parameters.
-	// Output streams directly to dump file.
-	// Use shell escaping to prevent injection attacks from user/database names.
-	pgDumpCmd := fmt.Sprintf("PGPASSWORD=$(cat -) pg_dump --clean --if-exists -h 127.0.0.1 -p %s -U %s -d %s",
-		ShellEscape(strconv.Itoa(int(cfg.Database.Port))),
-		ShellEscape(cfg.Database.User),
-		ShellEscape(dbName))
-
-	cmd := exec.CommandContext(ctx, p.containerCLI, "exec", "-i", p.dbContainerName, "sh", "-c", pgDumpCmd)
+	// Build pg_dump command. When explicit credentials are provided (e.g. integration
```
asafbennatan: changing the original approach here is unnecessary

internal/backup/podman_deployer.go
```diff
@@ -144,10 +157,14 @@ func (p *PodmanDeployer) BackupDatabase(ctx context.Context, outputDir string) e
 
 	// Execute pg_dump inside the container with password from stdin and safely escaped parameters.
 	// Output streams directly to dump file.
-	// Use shell escaping to prevent injection attacks from user/database names.
-	pgDumpCmd := fmt.Sprintf("PGPASSWORD=$(cat -) pg_dump --clean --if-exists -h 127.0.0.1 -p %s -U %s -d %s",
-		ShellEscape(strconv.Itoa(int(cfg.Database.Port))),
-		ShellEscape(cfg.Database.User),
+	password := p.dbPassword
+	user := p.dbUser
+	if user == "" {
```
asafbennatan: is this the correct location to apply defaults ?

internal/backup/podman_deployer_test.go
```diff
@@ -38,13 +30,17 @@ func internalDBServiceConfig() *config.Config {
...
 	return cfg
 }
 
 func TestPodmanDeployer_BackupDatabase_ExternalDB(t *testing.T) {
 	log, _ := test.NewNullLogger()
-	cfgPath := writeServiceConfig(t, externalDBServiceConfig())
+
+	dir := t.TempDir()
+	cfgPath := filepath.Join(dir, "service-config.yaml")
+	rawYAML := "db:\n  type: \"external\"\n  name: flightctl\n"
```
asafbennatan: can we use multiline string instead ?

asafbennatan (approved):
LGTM

PR #3175, opened 2026-06-28
EDM-2392: Adjust observability tests to run on quadlets and OCP

itzikb-redhat:
Adjustments to have the observability tests work with quadlets and OCP


test/e2e/observability/observability_test.go
```diff
@@ -34,14 +34,26 @@ func getPrometheusURL() (string, error) {
...
+		if cleanup == nil {
+			cleanup = noopCleanup
+		}
+		return url, cleanup, nil
+	}
+	url, err := getPrometheusURL()
+	return url, noopCleanup, err
+}
 
+var _ = Describe("Device observability", func() {
```
eldar101, line 25: Please keep the device observability block scoped to KIND || Quadlet. This PR removed the old KIND-only guard, so these VM/TPM specs now also run on OCP even though the change only validates quadlet support

test/e2e/infra/infra_provider.go
```diff
@@ -120,6 +121,7 @@ var DeploymentServiceNames = map[string]ServiceName{
 	"flightctl-imagebuilder-api":    ServiceImageBuilderAPI,
 	"flightctl-imagebuilder-worker": ServiceImageBuilderWorker,
 	"flightctl-alert-exporter":      ServiceAlertExporter,
+	"flightctl-prometheus":          ServicePrometheus,
```
eldar101: Adding flightctl-prometheus to the global deployment-name map changes existing StartServiceAccess("flightctl-prometheus", ...) behavior. 

The alertmanager proxy suite currently relies on the fallback path with explicit namespaces/port.

this new mapping sends it through Infra.ExposeService(ServicePrometheus), but K8s has no ServicePrometheus registry entry.

test/e2e/observability/observability_suite_test.go
```diff
@@ -41,6 +42,9 @@ var _ = AfterSuite(func() {
 })
 
 var _ = BeforeEach(func() {
```
eldar101, line 64: Move the quadlet observability skip after harness.SetTestContext(ctx). 

If it skips before the test ID is initialized, AfterEach cleanup/log capture can fail while trying to read the missing test ID.

itzikb-redhat: Removed from the suite

test/e2e/infra/infra_provider.go
itzikb-redhat: Removed

test/e2e/observability/observability_test.go
itzikb-redhat: Done

test/e2e/observability/observability_test.go
```diff
@@ -238,19 +263,20 @@ var _ = Describe("Device observability", func() {
...
 	Context("service level prometheus metrics", func() {
 		It("should expose service level metrics via the prometheus server", Label("88170"), func() {
 			harness := e2e.GetWorkerHarness()
 
-			By("getting Prometheus URL from aux (testcontainer)")
-			promURL, err := getPrometheusURL()
+			By("getting Prometheus URL")
+			promURL, promCleanup, err := getServiceObservabilityPrometheusURL()
 			Expect(err).ToNot(HaveOccurred())
+			defer promCleanup()
```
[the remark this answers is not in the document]
itzikb-redhat: It's in the tests

test/e2e/infra/infra_provider.go
```diff
@@ -28,6 +28,7 @@ const (
 	ServiceImageBuilderAPI    ServiceName = "imagebuilder-api"
 	ServiceImageBuilderWorker ServiceName = "imagebuilder-worker"
 	ServiceAlertExporter      ServiceName = "alert-exporter"
+	ServicePrometheus         ServiceName = "prometheus"
```
eldar101, line 34: Do we need to start aux services for Quadlet too? Quadlet uses flightctl-prometheus, so this adds extra setup and failure surface.

test/e2e/observability/observability_test.go
```diff
@@ -34,12 +33,33 @@ func getPrometheusURL() (string, error) {
 	return auxSvcs.Prometheus.URL, nil
 }
 
+// getServiceObservabilityPrometheusURL returns the Prometheus URL for observability tests.
```
eldar101: Can we move this helper to the bottom of the file with the other helpers?

test/e2e/observability/observability_test.go
```diff
@@ -260,10 +284,11 @@ var _ = Describe("Service observability", func() {
 				"http_server_request_duration_seconds_bucket",
 				"flightctl_repositories_total",
 			}
-
 			for _, query := range metrics {
 				By(fmt.Sprintf("verifying metric %s", query))
 				Eventually(harness.PromQueryResultCount(promURL, query), TIMEOUT, POLLING).Should(BeNumerically(">", 0))
```
eldar101: Can we print these before the assertion? Right now they only show up after success, not when the query fails.

test/scripts/agent-images/prepare_otel_config.sh
```diff
@@ -68,7 +68,11 @@ if [ -z "$ip" ]; then
   dns_suffix="nip.io"
 fi
 
-[ -n "$ip" ] || die "Could not extract IP from server URL: $server_url"
+# Hostname fallback (quadlet and other deployments using global.baseDomain)
```
eldar101: This logic is duplicated in the v10 script too.

itzikb-redhat: It should also be in v10. Maybe there is a way to not have a duplicate script but this is what exists right now

test/e2e/observability/observability_test.go
itzikb-redhat: Done

test/e2e/observability/observability_test.go
itzikb-redhat: Done

test/e2e/infra/infra_provider.go
itzikb-redhat: It's skipped now if it's quadlets

eldar101 (approved):

cmd/aux-service/main.go
```diff
@@ -163,7 +163,7 @@ func parseServices(args []string) ([]auxiliary.Service, error) {
 	for _, arg := range args {
 		switch arg {
 		case "all":
-			return auxiliary.AllServices, nil
+			return auxiliary.DefaultServices(), nil
```
asafbennatan: why ? shouldn't you just use the default instead? the original code has an ALL and a default , seems that this change simply makes default = all ?

test/e2e/infra/auxiliary/auxiliary.go
```diff
@@ -47,16 +49,34 @@ const (
...
+// Skips aux Prometheus on Quadlet (uses flightctl-prometheus from the observability stack).
+func DefaultServices() []Service {
+	if isQuadletE2EEnvironment() {
+		logrus.Info("Skipping aux Prometheus on Quadlet (uses flightctl-prometheus)")
+		return []Service{ServiceRegistry, ServiceGitServer}
+	}
+	return AllServices
+}
+
+func isQuadletE2EEnvironment() bool {
```
asafbennatan: why not use infra.isQuadlet?

test/scripts/agent-images/variants/v10/prepare_otel_config.sh
```diff
@@ -68,7 +68,11 @@ if [ -z "$ip" ]; then
   dns_suffix="nip.io"
```
asafbennatan, line 68: can we consolidate this and use the regular prepare_otel_config ?

test/e2e/observability/observability_test.go
```diff
@@ -400,3 +410,49 @@ func buildOTELRestartHook() v1beta1.FileSpec {
...
+		return "", fmt.Errorf("Prometheus not started")
+	}
+	return auxSvcs.Prometheus.URL, nil
+}
+
+// getServiceObservabilityPrometheusURL returns the Prometheus URL for observability tests.
+// On Quadlet uses flightctl-prometheus; on OCP uses the COO MonitoringStack; on KIND uses the aux testcontainer.
+func getServiceObservabilityPrometheusURL() (string, func(), error) {
+	noopCleanup := func() {}
+	if infra.IsQuadletEnvironment() {
```
asafbennatan: having this helper function to say , if its ocp or quadlets get it from infra but if its k8 get it from aux is fine , 
but we shouldnt handle quadlets or ocp logic here , this should just be infra.getservice/exposeservice whatever

we can even avoid infra types entirely - get from infra , if null get it from aux

test/e2e/observability/observability_test.go
```diff
@@ -21,25 +20,31 @@ const (
...
+
+	ocpMonitoringStackPrometheusService = "flightctl-monitoring-stack-prometheus"
+	ocpMonitoringStackPrometheusPort    = 9090
 )
 
-// getPrometheusURL returns the Prometheus URL from auxiliary.Services.
-func getPrometheusURL() (string, error) {
-	if auxSvcs == nil {
-		return "", fmt.Errorf("aux services not initialized")
+func ocpPrometheusBackends(infraProvider infra.InfraProvider) ([]e2e.ServiceAccessBackend, error) {
```
asafbennatan: service location should be part of the infra package , callers should not need to infra check the env type

test/e2e/infra/skip.go
```diff
@@ -39,6 +39,36 @@ func SkipIfRBACNotSupported(rbacProvider RBACProvider) {
...
 
+// quadletObservabilityServices are systemd services from the flightctl-observability stack
+// required by observability e2e tests on Quadlet (Prometheus scrapes API and gateway metrics).
+var quadletObservabilityServices = []ServiceName{
+	ServicePrometheus,
+}
+
+// SkipIfQuadletObservabilityNotRunning skips when on Quadlet and the observability stack
+// is not active. Requires flightctl-observability RPM and flightctl-observability.target.
+func SkipIfQuadletObservabilityNotRunning(lifecycle ServiceLifecycleProvider) {
```
asafbennatan: do we really need that , if we are saying observability tests are optional , just use infra the service url , if its null skip it - in all cases

test/scripts/agent-images/variants/v10/prepare_otel_config.sh
itzikb-redhat: I need to check with Ilya whether the non v10 is needed.

cmd/aux-service/main.go
itzikb-redhat: I'll revert and start the needed services from the test suite

test/e2e/infra/auxiliary/auxiliary.go
itzikb-redhat: Will remove

test/e2e/infra/skip.go
itzikb-redhat: Moving to test suite

cmd/aux-service/main.go
itzikb-redhat: Will remove

test/e2e/infra/auxiliary/auxiliary.go
itzikb-redhat: Removed

test/e2e/infra/infra_provider.go
eldar101: @itzikb-redhat 
I mean this changes how existing tests resolve `flightctl-prometheus`. Today that name falls back to the namespace/port passed by the test. With this map entry, it gets routed through `Infra.ExposeService(ServicePrometheus)` instead. That works for Quadlet, but KIND/K8s does not register `ServicePrometheus`, so existing KIND tests can fail.

test/e2e/infra/k8s/infra.go
```diff
@@ -516,6 +536,25 @@ func (p *InfraProvider) GetExternalNamespace() string {
 	return p.externalNamespace
 }
 
+// ServiceExists reports whether the backing Kubernetes Service for a logical service is present.
+func (p *InfraProvider) ServiceExists(service infra.ServiceName) (bool, error) {
```
asafbennatan: should we get context from the caller ?

test/e2e/infra/k8s/infra.go
```diff
@@ -516,6 +536,25 @@ func (p *InfraProvider) GetExternalNamespace() string {
...
+	info, ns, err := p.getServiceInfo(service)
+	if err != nil {
+		if service == infra.ServicePrometheus {
```
asafbennatan, line 543: i wonder if we can avoid singling out promethues and using the service map we have today (adapting it to support other namespaces as well )

test/e2e/infra/k8s/infra.go
itzikb-redhat: Done

test/e2e/infra/k8s/infra.go
itzikb-redhat: IIUC Prometheus is different and the move is not easy. Please let me know if you have a way.

asafbennatan (approved):
LGTM

eldar101 (approved):

PR #3133, opened 2026-06-21
EDM-4100: VM status reporting

asafbennatan:
adds vm status reporting that relies on virsh domstate inside the kubevirt-launcher container

Depends on: #3100

PR #3182, opened 2026-06-30
EDM-4095: switch VM render to vm-to-quadlet, expose publishPorts, revert .kube support

asafbennatan:
## Summary

- Switches VM application rendering from `kubevirt-vm-to-pod` to `flightctl/vm-to-quadlet`, which outputs a TAR stream of native Quadlet unit files
- Adds `publishPorts` as a direct field on `VmApplication` (removes the `.kube` inline input mechanism)
- Reverts agent-side `.kube` unit handling; restores `VMSpec` struct for VM metadata; fixes relative volume namespacing


internal/agent/device/applications/provider/quadlet.go
```diff
@@ -937,7 +922,9 @@ func namespaceVolumeName(value, appID string) string {
 	}
 
 	volumePart := parts[0]
-	if strings.HasPrefix(volumePart, "/") {
+	// Skip host bind mounts: absolute paths (/...) and relative paths (./... or ../).
+	// Named volumes never start with a path separator or dot.
+	if strings.HasPrefix(volumePart, "/") || strings.HasPrefix(volumePart, ".") {
```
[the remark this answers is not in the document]
asafbennatan: no - this was preexisting behavior which we even mitigated in this PR , adding what you asked for here is ambiguous to detect

test/integration/tasks/vmrender/Containerfile.vm-to-quadlet
```diff
@@ -0,0 +1,12 @@
...
+FROM registry.access.redhat.com/ubi9/go-toolset:1.25.9-1778675823 AS build
+USER 0
+ARG VM_TO_QUADLET_COMMIT=982eacb71ce62755a9dfd347933f27e0d12cd845
+RUN git clone https://github.com/flightctl/vm-to-quadlet /src && \
+    cd /src && git checkout ${VM_TO_QUADLET_COMMIT} && \
+    CGO_ENABLED=0 GOOS=linux go build -buildvcs=false -trimpath -ldflags="-s -w" -o /out/vm-to-quadlet ./cmd/vm-to-quadlet
+
+# Final stage: copy only the binary; no toolchain, non-root user.
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1763362218
+COPY --from=build /out/vm-to-quadlet /usr/local/bin/vm-to-quadlet
```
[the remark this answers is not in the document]
asafbennatan: this follows existing patterns in the codebase

api/core/v1beta1/validation.go
```diff
@@ -1358,21 +1360,20 @@ func validateContainerPorts(ports *[]ApplicationPort, path string) []error {
 	}
 
 	var allErrs []error
-	portPattern := regexp.MustCompile(`^[0-9]+:[0-9]+$`)
+	portPattern := regexp.MustCompile(`^[0-9]+:[0-9]+(/(tcp|udp|sctp))?$`)
```
galel12: `validateContainerPorts` is shared by both `ContainerApplication.ports` and `VmApplication.publishPorts`, but the regex now accepts `/protocol` for both. Container apps still use `ApplicationPort` with OpenAPI pattern `^\d+:\d+$` (no protocol suffix).

So something like `"8080:80/tcp"` on a container app would pass this Go validator but fail the OpenAPI schema check. Worth splitting this, maybe an `allowProtocol` flag or two thin wrappers, so container and VM keep their own contracts.

internal/tasks/device_render_vm.go
```diff
@@ -52,56 +57,59 @@ func NewVmConverter(binaryPath string) VmConverterFn {
...
-// always set. When kubeContent is non-nil it is parsed and any existing Yaml=
-// directive is overridden; otherwise the minimal default unit is used.
-// Yaml= is always forced to pod.yaml because renderVmApplication always emits
-// the converted pod YAML under that name.
-func buildKubeUnit(kubeContent *string) (string, error) {
-	var u *quadlet.Unit
-	if kubeContent != nil {
-		var err error
-		u, err = quadlet.NewUnit([]byte(*kubeContent))
+		files, err := parseTarFiles(stdoutBuf.Bytes())
```
galel12: Everything from vm-to-quadlet lands in `stdoutBuf` with no size cap before we parse it here. A bad vm.yaml or a noisy converter could produce a huge amount of output and OOM the worker, which handles rendering for many devices at once.

Might be worth wrapping stdout/stderr in a limited writer (something like 16MiB) so `cmd.Run()` fails fast instead of taking down the process.

internal/agent/device/applications/manager.go
```diff
@@ -78,18 +78,8 @@ func (m *manager) validateProviderDeps(ctx context.Context, p provider.Provider)
...
-func (m *manager) newAppFromProvider(p provider.Provider) (Application, error) {
-	if p.Spec().VM == nil {
-		return NewApplication(p), nil
-	}
-	podman, err := m.podmanFactory(p.Spec().User)
-	if err != nil {
-		return nil, fmt.Errorf("creating podman client for VM application %q: %w", p.Name(), err)
-	}
-	return NewVMApplication(p, podman, m.log), nil
+func (m *manager) newAppFromProvider(p provider.Provider) Application {
```
galel12, line 81: This drops the VM-specific path (`NewVMApplication` + `vmStatusPoller` / virsh domstate from #3133). Rendered VM apps now go through generic workload-count status in `Status()`.

That means we lose Stopped/Starting/Stopping and the ready=`0/1` when stopped semantics from #3133. The PR summary mentions HealthCmd from vm-to-quadlet as the replacement, but `podman_monitor` also ignores `health_status` events now.

E2e only checks Running so it might still pass, but worth confirming this is intentional for 1.3 and that we don't need a follow-up for VM-specific status.

galel12:
`test/harness/e2e/harness_application.go` wasn't touched in this PR but still builds VM specs with a `.kube` inline file and `PublishPort=2222:2222` inside it (`NewVmApplicationSpec`, lines 287-298).

This PR rejects any inline file other than `vm.yaml`, so `test/e2e/vm` will fail at validation/apply time.

Should drop the `.kube` entry, remove `vmKubeUnitTemplate`, and move the port mapping to `publishPorts: ["2222:2222"]` on the VmApplication instead.

galel12:
A few smaller things I noticed:

- `publishPorts` in OpenAPI has no pattern constraint, only Go validation enforces the format. Consider adding a pattern (container port format + optional `/protocol`) so schema and business validation stay aligned.

- No `publishPorts` test cases in `validation_test.go` for VmApplication. Render/inject is covered in `device_render_vm_test.go`, but API-level validation for VM ports isn't.

- `parseTarFiles` uses basic `../` checks on TAR entry names. `filepath.IsLocal` would catch more traversal forms (`dir/..`, backslashes). Also `io.ReadAll` per entry has no size cap, same OOM class as the unbounded stdout capture.

- `regexp.MustCompile` inside `validateContainerPorts` runs on every validation call. Could move to package scope.

- `"When inline has two .kube files it should fail"` in `validation_test.go` still expects `.kube` in the error message. The new message is `"only vm.yaml is allowed"`. Test probably still passes on substring match, but worth updating to match the new behavior.

api/core/v1beta1/validation.go
asafbennatan: whats the problem the go validator happens after the openapi validation - so any original input keeps working

internal/tasks/device_render_vm.go
asafbennatan: fixed

internal/agent/device/applications/manager.go
asafbennatan: this is the intended behavior for now - i will address in another PR,possibly with collapsing this into the normal app status - well see

asafbennatan:
> A few smaller things I noticed:
> 
> * `publishPorts` in OpenAPI has no pattern constraint, only Go validation enforces the format. Consider adding a pattern (container port format + optional `/protocol`) so schema and business validation stay aligned.
> * No `publishPorts` test cases in `validation_test.go` for VmApplication. Render/inject is covered in `device_render_vm_test.go`, but API-level validation for VM ports isn't.
> * `parseTarFiles` uses basic `../` checks on TAR entry names. `filepath.IsLocal` would catch more traversal forms (`dir/..`, backslashes). Also `io.ReadAll` per entry has no size cap, same OOM class as the unbounded stdout capture.
> * `regexp.MustCompile` inside `validateContainerPorts` runs on every validation call. Could move to package scope.
> * `"When inline has two .kube files it should fail"` in `validation_test.go` still expects `.kube` in the error message. The new message is `"only vm.yaml is allowed"`. Test probably still passes on substring match, but worth updating to match the new behavior.

fixed

packaging/images/el9/Containerfile.worker
```diff
@@ -1,14 +1,14 @@
...
+FROM registry.access.redhat.com/ubi9/go-toolset:1.25.9-1778675823 as vm-to-quadlet-build
 USER 0
-ARG KUBEVIRT_VM_TO_POD_COMMIT=44941fd0c7a8073c90b592bbc7534bb3278065ce
+ARG VM_TO_QUADLET_COMMIT=982eacb71ce62755a9dfd347933f27e0d12cd845
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/root/go/pkg/mod \
-    git clone https://github.com/vladikr/kubevirt-vm-to-pod /src && \
-    cd /src && git checkout ${KUBEVIRT_VM_TO_POD_COMMIT} && \
-    CGO_ENABLED=1 CGO_CFLAGS=-flto GOEXPERIMENT=strictfipsruntime GOOS=linux go build -o /out/kubevirt-vm-to-pod ./cmd
+    git clone https://github.com/flightctl/vm-to-quadlet /src && \
```
asafbss, line 9: 👍

api/core/v1beta1/validation.go
asafbss, line 1357: Should we also validate that the same host port is not defined more than once for the same protocol?

For example, duplicate 8080/tcp mappings would conflict, while using the same host port once for TCP and once for UDP should still be valid.

asafbennatan: this is not VM specific thing - we should have it at the application level - we currently do not - i propose we leave it for future work - once i get this entire feature merged i will be free to work on these sort of stuff

packaging/images/el9/Containerfile.worker
```diff
@@ -1,14 +1,14 @@
-# ── kubevirt-vm-to-pod build stage ──────────────────────────────────────────
+# ── vm-to-quadlet build stage ─────────────────────────────────────────────────
 # Isolated from the main build to avoid module-cache contamination — the tool
 # pins k8s.io at a different version than flightctl.
-FROM registry.access.redhat.com/ubi9/go-toolset:1.25.9-1778675823 as kubevirt-vm-to-pod-build
+FROM registry.access.redhat.com/ubi9/go-toolset:1.25.9-1778675823 as vm-to-quadlet-build
 USER 0
-ARG KUBEVIRT_VM_TO_POD_COMMIT=44941fd0c7a8073c90b592bbc7534bb3278065ce
+ARG VM_TO_QUADLET_COMMIT=982eacb71ce62755a9dfd347933f27e0d12cd845
```
asafbss, line 6: Now that vm-to-quadlet is maintained under the Flightctl organization, should we introduce versioned releases and reference a release tag here rather than only a raw commit SHA? Even if the tool changes infrequently, this would make compatibility and dependency tracking clearer.

asafbss (approved):

PR #3190, opened 2026-07-05
EDM-4161/EDM-4101: user documentation for VM applications and remote-…

asafbennatan:
## Summary

User-facing documentation for VM applications and related day-2 operations:

* **VM applications** in `managing-devices.md`: requirements (`/dev/kvm`, Podman 5.0), specification parameters, Fedora and Windows examples, status values (including `Stopped`/`Stopping`), restart behavior, and a clear boundary that Flight Control manages VM lifecycle but not content inside the guest.
* **Application lifecycle**: document `flightctl app start` / `stop` / `restart` (device and fleet where applicable), with fleet-vs-device precedence, and CLI reference entries that the procedural docs link to.
* **VM application console**: document `flightctl app console` (serial/VNC, `--force`, single session per app) under the VM applications section, with a full CLI reference entry.
* **References**: `flightctl-remote-access` in the components diagram, remote-access cert paths, and the `devices/applications/console` auth resource.

## Test plan

- [x] `make lint-docs`
- [x] `make spellcheck-docs`
- [ ] Spot-check rendered docs / anchors from `README.md` and cross-links between `managing-devices.md` and `cli-commands.md`

### DIFF
--- a/docs/user/references/cli-commands.md
+++ b/docs/user/references/cli-commands.md
@@ -317,9 +317,190 @@ flightctl get devices --cve-id CVE-2023-44487 --selector region=us-west
 
 ---
 
+## flightctl app start
+
+Start an application running on a device, or on every device owned by a fleet.
+
+### Synopsis
+
+```shell
+flightctl app start (device/NAME | fleet/NAME) --name APP [flags]
+```
+
+### Arguments
+
+* `device/NAME` or `fleet/NAME` - Target device or fleet
+
+### Flags
+
+* `--name <app_name>` - Application name to control (required)
+* `-y, --yes` - Skip the confirmation prompt
+
+### Description
+
+Starting an application on a fleet sets a fleet-wide default that is applied on top of the rendered application spec of every device currently owned by the fleet, and is automatically inherited by devices that join the fleet later. If a specific device also has its own override for the same application, from a previous `flightctl app start` or `flightctl app stop` issued directly against that device, whichever of the two actions was issued most recently takes effect for that device.
+
+### Examples
+
+```shell
+# Start an application on a single device
+flightctl app start device/my-device --name my-app
+
+# Start an application on every device in a fleet
+flightctl app start fleet/my-fleet --name my-app
+
+# Skip the confirmation prompt
+flightctl app start device/my-device --name my-app --yes
+```
+
+### Exit Status
+
+* `0` - Success
+* Non-zero - Error
+
+---
+
+## flightctl app stop
+
+Stop an application running on a device, or on every device owned by a fleet.
+
+### Synopsis
+
+```shell
+flightctl app stop (device/NAME | fleet/NAME) --name APP [flags]
+```
+
+### Arguments
+
+* `device/NAME` or `fleet/NAME` - Target device or fleet
+
+### Flags
+
+* `--name <app_name>` - Application name to control (required)
+* `-y, --yes` - Skip the confirmation prompt
+
+### Description
+
+Stopping an application on a fleet sets a fleet-wide default that is applied on top of the rendered application spec of every device currently owned by the fleet, and is automatically inherited by devices that join the fleet later. If a specific device also has its own override for the same application, from a previous `flightctl app start` or `flightctl app stop` issued directly against that device, whichever of the two actions was issued most recently takes effect for that device.
+
+### Examples
+
+```shell
+# Stop an application on a single device
+flightctl app stop device/my-device --name my-app
+
+# Stop an application on every device in a fleet
+flightctl app stop fleet/my-fleet --name my-app
+
+# Skip the confirmation prompt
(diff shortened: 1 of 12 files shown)


docs/user/references/cli-commands.md
```diff
@@ -317,9 +317,190 @@ flightctl get devices --cve-id CVE-2023-44487 --selector region=us-west
...
 ---
 
+## flightctl app start
+
+Start an application running on a device, or on every device owned by a fleet.
+
+### Synopsis
+
+```shell
+flightctl app start (device/NAME | fleet/NAME) --name APP [flags]
```
celdrake, line 327: Perhaps it should be described what happens if the application is already in a "started" state.

docs/user/references/cli-commands.md
```diff
@@ -317,9 +317,190 @@ flightctl get devices --cve-id CVE-2023-44487 --selector region=us-west
...
+* `device/NAME` or `fleet/NAME` - Target device or fleet
+
+### Flags
+
+* `--name <app_name>` - Application name to control (required)
+* `-y, --yes` - Skip the confirmation prompt
+
+### Description
+
+Starting an application on a fleet sets a fleet-wide default that is applied on top of the rendered application spec of every device currently owned by the fleet, and is automatically inherited by devices that join the fleet later. If a specific device also has its own override for the same application, from a previous `flightctl app start` or `flightctl app stop` issued directly against that device, whichever of the two actions was issued most recently takes effect for that device.
```
celdrake: For user documentation, I think this description goes into a little too much implementation detail.

docs/user/references/cli-commands.md
```diff
@@ -317,9 +317,190 @@ flightctl get devices --cve-id CVE-2023-44487 --selector region=us-west
...
+### Exit Status
+
+* `0` - Success
+* Non-zero - Error
+
+---
+
+## flightctl app stop
+
+Stop an application running on a device, or on every device owned by a fleet.
```
celdrake, line 367: ATM it's possible to run this command for any applciation, whether they are Running or not.
Ideally the CLI should acknowledge if nothing is going to happen, and that behaviour could also be described here.

docs/user/references/cli-commands.md
asafbennatan: Addressed: the start description now says that if the application is already started, the request succeeds and no restart is performed.

docs/user/references/cli-commands.md
asafbennatan: Addressed: rewrote the start/stop descriptions in user terms (fleet-wide vs device precedence) without the rendered-spec/override implementation detail.

docs/user/references/cli-commands.md
asafbennatan: Documented current behavior for restart (and the same idea for start/stop): the CLI/API request still succeeds when nothing will happen, and a restart only takes effect while the application is supposed to be running.

Having the CLI explicitly acknowledge a no-op is a good follow-up for the CLI itself; happy to track that separately if you want.

celdrake: Agree, can be a follow-up.

docs/user/using/managing-devices.md
```diff
@@ -713,6 +724,252 @@ spec:
...
+                        memory: 1Gi
+                  volumes:
+                    - name: containerdisk
+                      containerDisk:
+                        image: quay.io/containerdisks/fedora:40
+```
+
+##### Windows
+
+This example installs a Windows guest from ISOs staged on the device (`hostDisk`), with a blank root disk, VirtIO drivers media, UEFI firmware, Hyper-V features, and a TPM. You can also supply those ISOs from an OCI registry (for example as a `containerDisk`) instead; this example uses `hostDisk` for illustration. The `hostDisk` paths must already exist on the target device before applying the specification, and RDP is published on host port `3389`:
```
asafbss: The Windows example mentions “Hyper-V features”, which may be confusing because the device is still running the VM through KubeVirt/KVM, not Microsoft Hyper-V.

Maybe clarify that these are Hyper-V enlightenments exposed to the Windows guest for compatibility/performance, and that this does not imply nested virtualization or running Hyper-V as the underlying hypervisor.

asafbss: Since this requires the ISO files to be present on each target device, should we clarify that this `hostDisk` example is mainly intended for manual/illustrative setups? For production or larger-scale use, pulling the ISOs from OCI/containerDisk may be a more scalable approach.

asafbss (commented):
Looks solid, couple of comments

docs/user/using/managing-devices.md
asafbennatan: fixed

asafbennatan: fixed

asafbss (approved):
LGTM

PR #3183, opened 2026-07-01
NO-ISSUE: Fix authprovider browser login env

eldar101:
## Summary
This PR fixes the authprovider browser-login e2e tests so they work consistently across quadlet and OCP-style environments.

Main changes:

- Replaced the Keycloak browser-login path with a direct HTTP form submission helper instead of relying on browser automation.
- Added dynamic callback ports for OAuth browser-login tests so they do not depend on fixed port `8080`, which can already be used on quadlet hosts.
- Updated the Keycloak test realm to allow localhost callback ports and to include the expected audience in access tokens.
- Fixed the PAM browser-login test flow:
  - temporarily adds the selected callback URI to the bundled PAM issuer config
  - restarts the affected services
  - restores the original config after the test
  - follows the PAM login redirect the same way browser JavaScript does
- Improved auth test reliability by preserving the admin client config between specs instead of repeatedly creating new login attempts.
- Added a small retry path for API auth rate-limit failures by restarting the API once and retrying the login.
- Added helper coverage for PAM-vs-Keycloak OAuth2 selection, PAM secret resolution, URL redaction, login pipe cleanup, and PAM form redirect behavior.

## Why
The authprovider tests were failing because the test assumptions did not match the real environments:

- Keycloak/OAuth2 tokens were rejected with `401 failed to validate token` because the access token did not include the expected audience.
- Browser-login tests used fixed callback behavior that broke when `localhost:8080` was already occupied.
- PAM login is JavaScript-driven: the HTML form does not submit directly to the final callback, so the test helper needed to mimic the browser flow.
- Repeated login attempts could hit the API auth rate limiter and make later specs fail for unrelated reasons.

## Release target
Target release: release-1.3

PR #3192, opened 2026-07-06
NO-ISSUE: Add Keycloak client audiences for authprovider tests

eldar101:
## Summary
- Adds Keycloak audience mappers for the e2e `flightctl-client` and `flightctl-oauth2-client` clients.
- Ensures Keycloak access tokens include the client ID in the `aud` claim.
- Fixes authprovider browser-login flows where Flight Control could not route/validate the returned token against the configured provider.

## Description
The issue was that Keycloak login succeeded, but Flight Control rejected the returned token with 401 failed to validate token.
Why: 
Flight Control matches dynamic auth providers by token issuer + audience.
The Keycloak test clients were not putting their own client ID in the access token aud claim.
So Flight Control could not match the token to flightctl-client / flightctl-oauth2-client.
The fix adds Keycloak audience mappers for those test clients, so their access tokens include the expected client ID audience. Then Flight Control can route and validate the token correctly, and the browser login tests should pass.

## Release target
Target release: release-1.3


eldar101:
Closing this as a duplicate. The Keycloak audience fixture fix is now included in #3183 with the rest of the authprovider browser-login fixes.

PR #3187, opened 2026-07-02
EDM-4560: add VNC support for VM application console

asafbennatan:
## Summary

Adds VNC as a second supported console type for VM applications, alongside the existing serial console:

- **API (`api/core/v1beta1/openapi.yaml`)** — add `vnc` to the `consoleType` enum
- **Agent (`internal/agent/device/applications/console/vm_vnc_session.go`)** — new `vmVNCSession`, mirroring the existing `vmSerialSession`, that bridges a VM's VNC socket over the console gRPC stream
- **Agent (`internal/agent/device/applications/podman_monitor.go`)** — `resolveConsole` now accepts `consoleType=vnc` and returns a `vmVNCSession` for VM apps
- **Remote-access server (`internal/remote_access_server/ws_handler.go`) / session manager (`internal/console/app_console.go`)** — accept and forward `consoleType=vnc`
- **CLI (`internal/cli/console.go`)** — `flightctl console --app <name> --remote-type vnc [--exposed-port PORT]` opens a local TCP listener that port-forwards a VNC viewer connection through the persistent WebSocket session to the agent

## Test plan

- `make unit-test` — all tests pass
- `make lint` — 0 issues

Made with [Cursor](https://cursor.com)

PR #3193, opened 2026-07-06
EDM-4727: fix console trigger to avoid bumping renderedVersion

asafbennatan:
Depends on: #3187

## Summary

Opening a console session (device or VM app) previously triggered a rendered-version bump (via `GetNextDeviceRenderedVersion` + annotation update) purely to wake up the agent's long-polling `GetRenderedDevice` call. This conflated two unrelated concerns — "the device spec changed" and "the agent should reconnect for a console session" — and caused an unnecessary rendered-version increment (and downstream spec reconciliation) on every console open.

This decouples the two by introducing a separate, dedicated notification channel:

- **`internal/rendered` (`pubsub.go`, `manager.go`)** — replace the `string` rendered-version pub/sub payload with a typed `Notification{Type, RenderedVersion}` (`NotificationTypeSpecUpdated` / `NotificationTypeConsole`). Add a `console-pending` KV flag (`NotifyConsole` / `ClearConsoleNotification`) so a console wake-up survives even if no long-poll is currently subscribed. `WaitForNotification` (renamed from `WaitForNewVersion`) now returns whichever notification type fires first.
- **`internal/service/device.go`** — `GetRenderedDevice` switches on the notification type: `SpecUpdated` updates `kvRenderedVersion` as before; `Console` just clears the pending flag and returns the current device, without incrementing anything.
- **`internal/console/{console.go,app_console.go}`** — `modifyAnnotations` no longer computes/persists a next rendered version on session start/stop; `StartSession` now calls the new `ConsoleEventNotifier.NotifyConsole` directly instead of going through `RenderedVersionPublisher.StoreAndNotify`.
- **`internal/remote_access_server/server.go`, `internal/api_server/server.go`** — wire the renamed `ConsoleEventNotifier` (backed by `rendered.Bus.Instance()`) into both session managers.

## Test plan

- `make unit-test` — all tests pass
- `make lint` — 0 issues

### DIFF
--- a/internal/console/app_console.go
+++ b/internal/console/app_console.go
@@ -41,10 +41,11 @@ type AppConsoleSessionRegistration interface {
 	CloseSession(session *AppConsoleSession) error
 }
 
-// RenderedVersionPublisher stores and broadcasts rendered version change notifications so that
-// waiting GetRenderedDevice long-polls on the API server see the change immediately.
-type RenderedVersionPublisher interface {
-	StoreAndNotify(ctx context.Context, orgId uuid.UUID, name string, renderedVersion string) error
+// ConsoleEventNotifier signals the device agent that a console session needs handling.
+// It is deliberately decoupled from rendered-spec versioning.
+type ConsoleEventNotifier interface {
+	NotifyConsole(ctx context.Context, orgId uuid.UUID, name string) error
+	ClearConsoleNotification(ctx context.Context, orgId uuid.UUID, name string) error
 }
 
 // AppConsoleSessionManager manages application console sessions using device annotations.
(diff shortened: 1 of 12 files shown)


avishayt:
This is a big change for NO-ISSUE. Open a Jira please (there may already be one open).

internal/console/app_console.go
```diff
@@ -282,6 +270,11 @@ func (m *AppConsoleSessionManager) StartSession(ctx context.Context, orgId uuid.
 		}
 		return nil, domain.StatusInternalServerError(err.Error())
 	}
+
+	if err := m.notifier.NotifyConsole(ctx, orgId, deviceName); err != nil {
```
kkyrazis, line 274: Previously, failure to publish was an error, but now we're making the update in the database without notifying the api that a change was committed. Since we're no longer updating the RenderedVersion with console updates, is there any way for the device to become aware of these update if we fail to publish?

kkyrazis (commented):
There is a log message here: https://github.com/flightctl/flightctl/blob/main/internal/agent/device/spec/publisher.go#L226 that I think should be reworked as part of this. It was already incorrectly saying that it was skipping the version, but now it's even more incorrect.

internal/console/app_console.go
asafbennatan: we are publishing and keeping in KV , so the device will know it has to open a console on the next connection . 
note that this will fail in the case of kv failure , but you could make the same argument for db in the previous implementation.

asafbennatan:
> There is a log message here: https://github.com/flightctl/flightctl/blob/main/internal/agent/device/spec/publisher.go#L226 that I think should be reworked as part of this. It was already incorrectly saying that it was skipping the version, but now it's even more incorrect.

fixed

kkyrazis (approved):

PR #3200, opened 2026-07-07
NO-ISSUE: fix import.sh including failed images from BundleImages

aekubacki:
## Summary

- `BundleImages` returned an error for failed images but the caller passed the full original pair list to `WriteImportScript`, so `import.sh` referenced images never written to the bundle directory.
- Changed `BundleImages` signature from `error` to `([]ImagePair, error)`, returning only the successfully bundled pairs. The caller now passes this subset to `WriteImportScript`.

## Root Cause

`flightctl-remote-access-el9:1.2.0-rc3` was never published to quay.io. This caused `BundleImages` to fail for that image while silently including it in `import.sh` anyway, resulting in a broken bundle that referenced a non-existent local image path.

## Test plan

- [ ] Run `flightctl-mirror-images --bundle` against a variant that includes an image known to be missing or unreachable
- [ ] Verify `import.sh` in the resulting bundle does **not** reference the failed image
- [ ] Verify the warning for the failed image is still printed to stderr
- [ ] Verify all successfully bundled images are still present in `import.sh`

### DIFF
--- a/scripts/air-gap/mirror-images/main.go
+++ b/scripts/air-gap/mirror-images/main.go
@@ -205,10 +205,11 @@ func runBundleMode(ctx context.Context, unique []ImagePair, bundle, variant stri
 		Dest:   destRegistry + "/library/registry:2",
 	})
 
-	if err := BundleImages(ctx, bundleImages, tmpDir, exec); err != nil {
+	bundled, err := BundleImages(ctx, bundleImages, tmpDir, exec)
+	if err != nil {
 		logWarn("Some images failed to bundle: %v", err)
 	}
-	if err := WriteImportScript(tmpDir, bundleImages, variant); err != nil {
+	if err := WriteImportScript(tmpDir, bundled, variant); err != nil {
 		return fmt.Errorf("write import script: %w", err)
 	}
 	if bundleRPMs {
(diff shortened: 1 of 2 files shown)


scripts/air-gap/mirror-images/main.go
```diff
@@ -205,10 +205,11 @@ func runBundleMode(ctx context.Context, unique []ImagePair, bundle, variant stri
 		Dest:   destRegistry + "/library/registry:2",
 	})
 
-	if err := BundleImages(ctx, bundleImages, tmpDir, exec); err != nil {
+	bundled, err := BundleImages(ctx, bundleImages, tmpDir, exec)
+	if err != nil {
 		logWarn("Some images failed to bundle: %v", err)
```
kkyrazis, line 210: why is it just a warning if some images failed to bundle?

aekubacki: I dont think these are critical, and I dont know if it should fail out right.  Also if these images aren't available for us to pull as expected that would result in the tool always failing if true.  I would like it to pull what it can and warn when its missing something.  Hopefully we can address remote-access not being found separately.

aekubacki: Good catch on the CodeRabbit comment too — just pushed a follow-up that fixes `WriteManifest` to also use `bundled` instead of the original `bundleImages` list, so the artifact manifest stays in sync with the archive contents.

On the warning vs error: the intent is resilience — the tool should produce a usable bundle with whatever images are available and warn loudly about what's missing, rather than hard-failing and producing nothing. In our case, `flightctl-remote-access-el9:1.2.0-rc3` was never published to quay.io (a separate publishing pipeline gap), so a hard error would have blocked every bundle run until that's resolved.

kkyrazis (approved):

PR #3205, opened 2026-07-08
EDM-4561: add registry auth support for imagebuilder-worker service images

aekubacki:
## Summary

- Adds a new Helm value `imageBuilderWorker.serviceImages.pullSecretName` that mounts a Kubernetes secret (containing an `auth.json` key) at `/root/.config/containers/auth.json` inside the imagebuilder-worker pod
- Adds a commented `Volume=` line to the Podman quadlet container unit for the same purpose
- Documents both paths in `configuring-imagebuilder.md` with a new "Authenticating to a private registry for builder service images" section

## Context

In air-gapped environments, users mirror the builder service images (podman, bootc-image-builder, Syft) to an internal registry and already override `serviceImages.*` in Helm. However, if that registry requires authentication, builds fail at the `podman run` call with `unauthorized: authentication required` and there was no supported way to supply credentials.

`imagePullSecrets` and `ImageTagMirrorSet` only affect CRI-O pulls of the outer worker pod — they do **not** apply to the inner `podman run` calls the worker makes at build time. The fix is to mount a credential file directly.

## Test plan

- [ ] Deploy with `serviceImages.pullSecretName` set pointing to a valid secret; confirm builds succeed against an authenticated internal registry
- [ ] Deploy without the value set; confirm no regression for unauthenticated registries
- [ ] Verify the new section renders correctly in the documentation site


docs/user/installing/configuring-imagebuilder.md
```diff
@@ -274,6 +275,88 @@ To use a custom CA for the registry that serves the Podman or bootc-image-builde
...
+```bash
+sudo mkdir -p /etc/flightctl
+sudo cp /path/to/auth.json /etc/flightctl/service-images-auth.json
+sudo chmod 600 /etc/flightctl/service-images-auth.json
+```
+
+Edit `/usr/share/containers/systemd/flightctl-imagebuilder-worker.container` and uncomment (or add) the Volume line:
+
+```ini
+[Container]
```
kkyrazis, line 351: we should recommend creating dropins over manually editing the container files for this. If we modify the direct quadlet file then on upgrade we'll lose the changes while dropins will persist.

mletalie:
Edge Manager docs (MR 194) now recommend Quadlet drop-ins under `/etc/containers/systemd/flightctl-imagebuilder-worker.container.d/` instead of editing the package-shipped unit file, per @kkyrazis feedback above. Worth updating the Quadlet section in this PR to match before merge so both doc sets stay aligned.

asafbennatan (commented):
mostly LGTM , please mention drop in files in comments and docs and not original file editing

kkyrazis (approved):

PR #3208, opened 2026-07-08
EDM-4667: isolated store subpackages

asafbennatan:
Isolated store and services sub packages


internal/store/event/store.go
```diff
@@ -0,0 +1,92 @@
+package event
```
keitwb, line 1: `/internal/store/event.go` that this is from wasn't deleted.

asafbennatan: Good catch — `internal/store/event.go` was an orphaned leftover from the split; deleted it in 3d2d99b15.

keitwb (approved):

internal/service/dependencyref/service.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	"time"
+
+	"github.com/flightctl/flightctl/internal/domain"
+	"github.com/flightctl/flightctl/internal/store/model"
+	"github.com/google/uuid"
+)
+
+// Service is the focused DependencyRef service interface, extracted from the monolithic
+// internal/service.Service. It covers the 12 DependencyRef methods defined in the old
```
kkyrazis: I don't want to hold this PR up because overall I think it's a great improvement want to ensure you don't have to deal with conflicts all day, but I think many of the workflow's comments on the individual service interfaces aren't very good. Why is the interface's documentation referencing implementation details? Is the refactor documentation necessary? Are our comments supposed to be directly referencing design doc items?

kkyrazis (approved):

internal/service/dependencyref/service.go
asafbennatan: thanks will adjust

galel12 (approved):
LGTM

internal/service/dependencyref/service.go
asafbennatan: Went through internal/service/**/service.go and handler.go and stripped the refactor-narration doc comments (monolith references, EDM ticket numbers, design-doc section numbers, exhaustive-inspection audit notes) — kept only what actually documents current behavior. Also caught and fixed two comments that had gone stale relative to the code they were describing. In d9c7f1fb4.

galel12 (approved):

galel12 (approved):

PR #3209, opened 2026-07-09
EDM-4696: add NetworkPolicy to allow OCP Router access to cli-artifacts

talhil-rh:
## Summary

- The Helm chart deploys a `allow-from-flightctl` NetworkPolicy that blocks all external ingress to every pod in the `flightctl` namespace.
- All other services with an OpenShift Route (e.g. `flightctl-api`, `flightctl-ui`) have a corresponding NetworkPolicy that explicitly allows ingress from the `openshift-ingress` namespace (OCP Router).
- `flightctl-cli-artifacts` was missing this policy, so the OCP Router's requests were silently dropped by OVN-Kubernetes, returning HTTP 503 for all CLI artifact downloads in standalone OCP deployments.

## Fix

Add `flightctl-cli-artifacts-networkpolicy.yaml` — modeled on the existing `flightctl-api-networkpolicy.yaml` — which allows ingress from the `openshift-ingress` namespace to pods labelled `flightctl.service=flightctl-cli-artifacts`. The policy is gated on the same conditions as the Route itself (`enableMulticlusterExtensions=false` and `exposeServicesMethod=route`).

## Verification

Manually applied the policy to `ocp-edge124` (cluster `ocp3m0w-ic4s20`). CLI artifact URLs that previously returned 503 now return HTTP 200.

Fixes: [EDM-4696](https://redhat.atlassian.net/browse/EDM-4696)
Related: [EDM-3771](https://redhat.atlassian.net/browse/EDM-3771)

## Test plan

- [ ] Deploy to a standalone OCP cluster with `exposeServicesMethod=route`.
- [ ] Confirm that all CLI artifact download URLs are accessible from outside the cluster (HTTP 200).
- [ ] Confirm that ACM/multicluster deployments (`enableMulticlusterExtensions=true`) are not affected (policy not rendered).

### DIFF
--- a/deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-networkpolicy.yaml
+++ b/deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-networkpolicy.yaml
@@ -1,0 +1,23 @@
+{{ if and (.Values.cliArtifacts.enabled) (eq (include "flightctl.enableMulticlusterExtensions" .) "false") (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: flightctl-cli-artifacts-from-ingress
+  namespace:  {{ .Release.Namespace }}
+  labels:
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+    flightctl.service: flightctl-cli-artifacts
+spec:
+  ingress:
+  - from:
+    - podSelector: {}
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          network.openshift.io/policy-group: ingress
+  podSelector:
+    matchLabels:
+      flightctl.service: flightctl-cli-artifacts
+  policyTypes:
+  - Ingress
+{{ end }}


eldar101:
Fix issues and I'll approve.

deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-networkpolicy.yaml
```diff
@@ -0,0 +1,23 @@
+{{ if and (eq (include "flightctl.enableMulticlusterExtensions" .) "false") (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
```
asafbennatan: Values.cliArtifacts.enabled ? 
see other network policies

talhil-rh:
Thanks @asafbennatan — fixed. Added `.Values.cliArtifacts.enabled` to the condition (matching the guard used in `flightctl-cli-artifacts-route.yaml` and all other cli-artifacts templates) so the NetworkPolicy is only rendered when the service itself is enabled.

---

Addressing the three CodeRabbit threads:

**Trim markers (`{{-` / `{{- end }}`)** — skipping intentionally. The two peer NetworkPolicy files in this chart (`flightctl-api-networkpolicy.yaml`, `flightctl-imagebuilder-api-networkpolicy.yaml`) both use `{{ if` / `{{ end }}` without trim markers. Changing only this file would be inconsistent; a cleanup of all three together is a separate concern.

**Restrict ingress to port 8090** — skipping intentionally. Neither `flightctl-api-networkpolicy.yaml` nor `flightctl-imagebuilder-api-networkpolicy.yaml` restrict to a specific port. Adding port restrictions is a worthwhile hardening task but should be applied uniformly across all NetworkPolicies in a dedicated follow-up.

**Chart version bump** — skipping. `Chart.yaml` is pinned to `0.1.0` across the repo and is not bumped per-template PR.

deploy/helm/flightctl/templates/cli-artifacts/flightctl-cli-artifacts-networkpolicy.yaml
talhil-rh: fixed
 Added .Values.cliArtifacts.enabled to the condition — the template now reads:

{{ if and (.Values.cliArtifacts.enabled) (eq ... "false") (eq ... "route") }}
This matches the guard in flightctl-cli-artifacts-route.yaml and all other cli-artifacts templates.

eldar101 (approved):

itzikb-redhat:
@talhil-rh https://github.com/flightctl/flightctl/pull/3237

talhil-rh:
Closing — superseded by #3237 which already added the cli-artifacts NetworkPolicy to main.

PR #3215, opened 2026-07-09
NO-ISSUE: overlap E2E aux service setup with VM boot, stream image bundle via skopeo

asafbennatan:
## Summary

Two independent optimizations to the one-time E2E shard setup cost, found while investigating why E2E wall time didn't drop as much as expected after the buildah cache fix (#3186 / #3212).

Traced raw BeforeSuite timestamps from real runs and found that the ~4.3-4.4 min "cold start" tax paid once per shard splits almost 50/50 between two things that were previously lumped together as "VM management overhead":

- **Aux service setup** (registry, agent image bundle upload, git server, prometheus): ~145-150s, dominated by uploading a 3.24GB/6-image bootc bundle into the local registry (~96-100s).
- **VM cold boot + first pristine snapshot**: ~110-115s (boot+SSH ~40s, greenboot wait ~60s, snapshot create ~10s).

Both fixes are behavioral no-ops for test correctness; only the setup mechanics change.

### Commit 1 — overlap aux setup with VM boot

Every suite's `BeforeSuite` called `auxiliary.Get(ctx)` and then `SetupWorkerHarness*()` sequentially, even though the two are independent (aux uses podman networking, VM setup uses libvirt/qemu). Added a `StartAuxServicesAsync`/`AuxServicesFuture` helper that kicks off aux setup in the background, and switched ~20 VM-based suites to `Wait()` on it right after VM/harness setup instead of before.

Since every shard's first package pays this tax exactly once, and ~15 shards run in parallel, this overlaps ~100-110s of independent work per shard.

`dependency_sync_suite_test.go` is intentionally left alone — it uses `SynchronizedBeforeSuite`, where Ginkgo blocks the per-process phase on the return value of the controller phase, so the two phases can't overlap there.

### Commit 2 — stream image bundle directly to registry via skopeo

`uploadBundle` ran `podman load -i bundle.tar` (extracting every image's full layers into local containers storage) and then looped over each tag doing a digest-check plus `podman push` from storage.

`uploadBundle` is only invoked when the registry container was just created (see `StartServices`' `Reused` check), so the registry is always empty and every image always needs pushing — the local/registry digest comparison never actually skipped anything, it just cost two `skopeo inspect` calls per tag.

Replaced `podman load`+push with `skopeo copy docker-archive:bundle:<ref> docker://...`, run concurrently (bounded to 4 at a time) across image refs. This streams layer blobs straight from the tar to the registry without ever extracting a full OS root filesystem to local storage just to immediately read it back out for push.

Verified locally end-to-end: built a multi-image docker-archive from real images, copied every ref through the new code path into a throwaway registry, and confirmed each pushed image's config digest and image ID match the source exactly after pulling it back down.

## Test plan

- [x] `go build ./...` and `go vet ./test/e2e/... ./test/harness/...` pass clean
- [x] Manually verified `skopeo copy docker-archive:path:<ref>` correctly disambiguates distinct images in a multi-image archive (digest-verified)
- [x] Manually verified end-to-end upload correctness: pushed images pulled back down match source image ID and config digest exactly
- [ ] Watch the next few merge-queue E2E runs / weekly health report to confirm the real-world wall-time delta (couldn't reproduce the ~100s CI bottleneck locally; this sandbox's disk is too fast for a representative before/after timing)

### DIFF
--- a/test/e2e/agent/agent_suite_test.go
+++ b/test/e2e/agent/agent_suite_test.go
@@ -48,9 +48,10 @@ func TestAgent(t *testing.T) {
 var auxSvcs *auxiliary.Services
 
 var _ = BeforeSuite(func() {
-	auxSvcs = auxiliary.Get(context.Background())
+	auxFuture := e2e.StartAuxServicesAsync(context.Background())
 	Expect(setup.EnsureDefaultProviders(nil)).To(Succeed())
 	e2e.SetupWorkerHarnessOrAbort()
+	auxSvcs = auxFuture.Wait()
 })
 
 var _ = AfterSuite(func() {
--- a/test/e2e/applications/containers/containers_suite_test.go
+++ b/test/e2e/applications/containers/containers_suite_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/flightctl/flightctl/test/e2e/infra/auxiliary"
 	"github.com/flightctl/flightctl/test/e2e/infra/setup"
 	"github.com/flightctl/flightctl/test/harness/e2e"
 	"github.com/flightctl/flightctl/test/util"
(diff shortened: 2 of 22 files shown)


itzikb-redhat:
lgtm

itzikb-redhat (approved):

PR #3218, opened 2026-07-12
EDM-4520: Fix for e2e backup-restore ers cleanup

itzikb-redhat:
Verify API is accessible before cleanup
Only start a VM for tests that need it

PR #3221, opened 2026-07-12
NO-ISSUE: reduce E2E LPT makespan via configurable greenboot/dependency-sync timeouts and backup_restore parallelization

asafbennatan:
E2E LPT makespan reductions and related CI hardening, following up on the E2E health report's optimization matrix.

## Timing / config (test-accelerated, production defaults unchanged unless noted)

- **Configurable greenboot health timeouts**: `flightctl-agent health`'s `--stability-window` and `--poll-interval` are CLI flags (were hardcoded 150s/60s), sourced from `/etc/greenboot/greenboot.conf`. Rollback E2E overrides use Phase 1=90s (>= systemd `TimeoutStartSec` for Type=notify, since the override survives ostree merge into later good boots), Phase 2=10s, poll=2s.
- **Configurable dependency-sync + repository-tester intervals**: Helm `periodic.tasks.dependencySync` / `repositoryTester` schedule intervals; e2e sets 5s (dev keeps production defaults).
- **Configurable enrollment-verify backoff**: `EnrollmentVerifyInterval` / `Cap` / `Steps` on the agent; e2e uses a short interval with widened Cap/Steps so VM-pool bootstrap does not restart-loop the agent. Production Cap/Steps stay at the prior hardcoded values.

## E2E harness / image path

- **Parallelize backup_restore's 2nd/3rd VM setup** via `errgroup` (~one VM-boot off the longest spec).
- **Mirror `quay.io/flightctl-tests/*` into the local registry** at aux startup (with retries) and remap on devices, so specs stop paying per-pull quay latency.
- **Rootless suite starts aux services itself** (was assuming another suite had already created the e2e registry; LPT could schedule it first and then hit connection-refused for the whole run).
- Shrink post-rollback / update-schedule no-op stability windows to 45s.

## CI cache hygiene

- Gate real e2e bootc-image-builder cache save to main-only; prune stale e2e-unified bootc, lychee, and setup-go cache entries; drop redundant setup-go cache in `publish-cli-bins`.

## Test plan

- [x] CI passes on this PR (E2E)
- [ ] Confirm greenboot rollback tests still correctly detect rollback/stability with Phase 1=90s / Phase 2=10s overrides
- [ ] Confirm backup_restore spec still passes with parallel VM setup
- [ ] Confirm dependency-sync / repository-tester / enrollment-verify are faster in e2e only
- [ ] Confirm rootless + containers/helm suites pull `flightctl-tests` images via the local registry


asafbennatan:
Rebased onto `main` (resolved a conflict in `backup_restore_suite_test.go` against the newer `SetupWorkerHarnessWithoutVM` behavior from #3218) and addressed all 3 actionable review comments:

- `backup_restore_test.go`: cleanup for `harness2`/`harness3` is now keyed on the harness being non-nil rather than a separate `Ready` flag, so a harness left partially initialized by a failed `SetupVMFromPoolAndStartAgent` still gets its VM/resources cleaned up.
- `images.go`: threaded the `StartServices` context through `UploadImages`/`uploadBundle`/`copyImageFromBundle` and switched to `exec.CommandContext` with a 5-minute per-copy timeout, so a hung `skopeo copy` can no longer pin an upload semaphore slot indefinitely.
- `task_definitions_test.go`: added table-driven coverage asserting a configured `DependencySync` interval applies to both `dependency-sync-git` and `dependency-sync-http`, and that both keep their default when unconfigured.

Also replaced the `Co-authored-by: Cursor` trailers with `Made-with: Cursor` across the commits that had them, to fix the failing Ai-Attribution check.

Made-with: Cursor

test/e2e/agent/agent_update_test.go
```diff
@@ -757,13 +768,38 @@ func readAgentLogsForRollbackAssertion(harness *e2e.Harness) string {
...
+// Production defaults are 150s/60s, calibrated for real hardware; these tests only need
+// to observe that rollback triggers and that stability is checked, not the exact
+// durations, so there's no fidelity lost by shortening them here.
+//
+// /etc is carried across bootc deployments via ostree's 3-way merge, so writing this to
+// the currently-booted deployment before triggering the v11 update also takes effect on
+// the v11 boot and the rollback boot that follows it.
+const fastGreenbootOverrideScript = `sudo mkdir -p /etc/greenboot
+cat <<'EOF' | sudo tee -a /etc/greenboot/greenboot.conf >/dev/null
+FLIGHTCTL_HEALTH_CHECK_TIMEOUT=10
```
galel12: Phase 1 = 10s is great for killing the broken-agent wait, but this lands in `/etc` and survives ostree merge into later good boots in the same flow. Script comment over in flightctl-agent-running-check.sh says Phase 1 should be >= systemd TimeoutStartSec (~90s for Type=notify). If a healthy agent is slow to hit active under load, greenboot can fail a good boot on the leftover 10s timer. CI passed, so maybe fine in practice, but this feels like the main flakiness risk. Intentional?

test/e2e/agent/agent_update_test.go
```diff
@@ -757,13 +768,38 @@ func readAgentLogsForRollbackAssertion(harness *e2e.Harness) string {
...
+// via /etc/greenboot/greenboot.conf, the same file used for GREENBOOT_MAX_BOOT_ATTEMPTS.
+// Production defaults are 150s/60s, calibrated for real hardware; these tests only need
+// to observe that rollback triggers and that stability is checked, not the exact
+// durations, so there's no fidelity lost by shortening them here.
+//
+// /etc is carried across bootc deployments via ostree's 3-way merge, so writing this to
+// the currently-booted deployment before triggering the v11 update also takes effect on
+// the v11 boot and the rollback boot that follows it.
+const fastGreenbootOverrideScript = `sudo mkdir -p /etc/greenboot
+cat <<'EOF' | sudo tee -a /etc/greenboot/greenboot.conf >/dev/null
```
galel12, line 798: nit: `tee -a` appends every call. If the VM isn't restored pristine between specs that hit this, greenboot.conf just keeps growing (last assignment wins, but messy). Overwrite / dedupe?

galel12 (commented):
Looks directionally good, and e2e is green. A few leftovers:

PR body is stale. Original 3 bullets don't cover enrollment-verify backoff, flightctl-tests mirroring, rootless aux-registry fix, or the cache prune work. Worth refreshing so reviewers know what's in the bag.

test/e2e/agent/agent_update_test.go
asafbennatan: Good catch — intentional for the broken-agent wait, but not safe for later good boots that inherit the `/etc` override via ostree merge (rollback + recovery in the same flow).

Raised Phase 1 to **90s** to match the `TimeoutStartSec` floor called out in `flightctl-agent-running-check.sh`. Phase 2/poll stay short (10s/2s); those only run after the agent is already active. See `2f438ed43`.

test/e2e/agent/agent_update_test.go
asafbennatan: Agreed. Now strips any existing `FLIGHTCTL_HEALTH_*` keys before appending, so repeated calls / a non-pristine VM don't grow `greenboot.conf`. Same commit as above.

asafbennatan:
Addressed @galel12's review leftovers:

1. **PR body** refreshed to cover enrollment-verify backoff, flightctl-tests mirroring, rootless aux-registry fix, and cache prune work.
2. **Phase 1 timeout**: raised from 10s → 90s (`TimeoutStartSec` floor) so leftover `/etc` overrides can't flake healthy post-rollback/recovery boots; Phase 2/poll stay fast.
3. **`tee -a`**: strip prior `FLIGHTCTL_HEALTH_*` keys before append so the conf doesn't accumulate duplicates.

galel12 (approved):
LGTM

PR #3222, opened 2026-07-13
NO-ISSUE: stop go/lint/dnf/bootc caches from thrashing the repo's 10GB cache budget

asafbennatan:
## Summary

Follow-up to #3214/#3219's buildah cache pruning: the repo's total Actions cache usage was still over the 10GB budget (~11GB), which was evicting small, unrelated caches out from under other workflows - most notably `e2e-test-timings` (causing the E2E LPT scheduler to silently fall back to unweighted round-robin assignment) and `Linux-apt-kvm-*` (causing severe `apt install` slowdowns from cold, unthrottled-cache network pulls).

Root cause this time: `Linux-go-*` and `lint-cache-Linux-*` (and, latent but not yet triggered, `Linux-dnf-*` / `Linux-bootc-cache-smoke-*`) use a content-hash key that's identical across every PR/branch, but `actions/cache` save/restore is scoped per-ref. Their save steps were unconditional, so every single PR/branch run left its own throwaway ~1-1.2GB copy behind that nothing else could ever restore from and nothing ever deleted - 6 `Linux-go-*` copies (6.33GB) and 2 `lint-cache-Linux-*` copies (2.1GB) were alive at once as of this PR.

- **Prevention:** gate every such save to `github.ref == 'refs/heads/main'`, matching the existing buildah cache pattern. Every other ref still benefits from restoring main's copy via `restore-keys` fallback. Also skip the save when the paired restore already got an exact `cache-hit`, avoiding a wasted archive-then-reject round trip (`actions/cache/save` archives the directory *before* checking whether the key already exists server-side).
- **Cleanup/safety net:** extend `prune-ci-caches.yaml`'s branch-scoped pruning to keep only the single newest `refs/heads/main` entry per prefix (covers both existing PR-scoped duplicates and any future stale entry left behind by a since-changed content hash), and add the `Linux-dnf-`/`Linux-bootc-cache-smoke-` prefixes for completeness. Also trigger the prune workflow after `Code Quality` completes on main, since that's what saves the lint-cache entry.
- **Lint-cache freshness:** unlike the standalone Go module/build cache (fully determined by `go.sum`), the lint-cache volume set also holds golangci-lint's own analysis cache, which is invalidated by source content and lint config, not `go.sum` - keying it on `go.sum` left it frozen at whatever it looked like right after the last `go.sum` change, however much source changed in between. Switched it to the same pattern as buildah: a fresh `run_id`-keyed save on every main push, with `restore-keys` falling back to the newest entry, and moved it to the plain "keep newest" prune job.

Expected effect: steady-state total cache usage drops from ~11GB (over budget) to ~4.6GB, giving enough headroom that this class of cascading eviction shouldn't recur from the same cause.

## Test plan

- [x] Validated all modified workflow YAML parses correctly
- [x] Validated the refined prune jq logic against sample restore/main-ref data (keeps newest main entry, deletes the rest; leaves everything alone when no main entry exists yet)
- [ ] After merge: confirm on the next few main-branch runs that `Save Go cache`/`Save golangci-lint volumes cache`/etc. only actually save once (first run after a `go.sum`/lockfile change, or every run for lint-cache), and that the prune jobs clean up the currently-existing PR-scoped duplicates
- [ ] Confirm total repo cache usage (`gh cache list`) settles near the ~4.6GB estimate

### DIFF
--- a/.github/workflows/build-images-and-charts.yaml
+++ b/.github/workflows/build-images-and-charts.yaml
@@ -223,6 +223,7 @@ jobs:
         uses: ./.github/actions/compute-tag
 
       - name: Restore Go cache
+        id: cache-go-restore
         uses: actions/cache/restore@v4
         with:
           path: |
@@ -251,7 +252,14 @@ jobs:
 
       - name: Save Go cache
         uses: actions/cache/save@v4
-        if: always()
+        # Same key across every PR/branch (hashFiles('**/go.sum') rarely changes), but Actions
+        # cache save is scoped per-ref, so an unconditional save here leaves one throwaway
+        # ~1GB copy behind per PR/branch that nothing else can ever restore from and nothing
+        # ever deletes - pure eviction pressure on the repo's shared 10GB budget. Restrict
+        # saves to main, matching the buildah cache above; every other ref still benefits via
+        # restore-keys fallback to main's copy. cache-hit skip avoids racing this job's save
+        # against build-restore's identically-keyed save when both run on main concurrently.
+        if: always() && github.ref == 'refs/heads/main' && steps.cache-go-restore.outputs.cache-hit != 'true'
         with:
           path: |
             ~/.cache/go-build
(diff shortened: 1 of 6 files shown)


EfratIfergan (approved):
LGTM

PR #3224, opened 2026-07-13
NO-ISSUE: fix VM app console on Kubernetes (RBAC, UI wiring, annotation leaks, CLI errors)

asafbennatan:
## Summary

Several VM app console (serial/VNC) issues were only reproducible on Kubernetes (kind) deployments, not on quadlets:

- **Missing RBAC for `flightctl-remote-access`**: it had no `ServiceAccount`, so token review (`tokenreviews:create`) and rolebinding lookups (`rolebindings:get,list`) failed. Adds a `ServiceAccount`, `ClusterRole`, and `ClusterRoleBinding`, and wires auth config into its Helm templates.
- **UI not configured to reach remote-access**: wires `FLIGHTCTL_REMOTE_ACCESS_SERVER` into the UI config for both the Helm chart and the Podman/Quadlet deployment, and NodePort-exposes the remote-access service and its agent gRPC endpoint for kind/dev deployments (`values.dev.yaml`, `values.nodeport.yaml`).
- **Wrong remote-access hostname derivation**: fixes `deriveRemoteAccessServer` in the agent to correctly substitute `agent-api.` → `agent-remote-access.`.
- **kind cluster port mapping**: maps host ports 3444/7444 so local testing can reach the NodePort-exposed services.
- **CLI silently succeeded on agent-reported session failures** (e.g. requesting a console for an app that doesn't exist) instead of raising an error. Adds an explicit `error` field to the console gRPC `StreamRequest` so the agent can signal this distinctly from console payload data, routes it through a dedicated WebSocket close code (`AppConsoleErrorCloseCode`) in the remote-access server, and surfaces it in the CLI as a `ConsoleSessionError` with a non-zero exit code — for both the serial and VNC console paths.
- **Stale console annotation blocking new sessions**: the device/app console session annotation was being left behind as a literal `"[]"` instead of being deleted once the last session closed, permanently blocking new console sessions. `removeSession`/`removeAppSession` now delete the annotation key entirely when no sessions remain, and self-heal any pre-existing empty-array value.

## Test plan

- [x] `go build ./...`
- [x] `go vet` across affected packages
- [x] `go test -race` across `internal/cli`, `internal/console`, `internal/remote_access_server`, `internal/agent/...`, `api/...` — all passing
- [x] `make lint-openapi` — passing
- [x] `helm lint deploy/helm/flightctl` — passing
- [x] Added unit test coverage for annotation cleanup/self-heal, gRPC error routing, WebSocket close-code handling, and CLI session-error handling (including the VNC path)

- [x] **`release-1.3`**


internal/cli/console.go
```diff
@@ -87,6 +88,17 @@ type ConsoleOptions struct {
 	exposedPort int
 }
 
+// ConsoleSessionError indicates the server or agent reported a session-level failure
+// (e.g. the requested application does not exist) over an already-established
+// WebSocket connection, as opposed to a transport/handshake-level error.
+type ConsoleSessionError struct {
```
celdrake: The error for an unknown application is not working as expected, as the annotation for the console session is written before it's validated. See initially the CLI indicates that the connection was established, and then that the app was not found. The annotation was written nevertheless.


Tested with `205b1ac593fa09a73c5818d92510dd43084d5c74`

```
flightctl console dev/$dev --app=vm-app-not-real --remote-type serial 
Connected to vm-app-not-real console. Use ~. to exit.
Error for device daai3p2dmjuiecfh8ds87a8nh9u9mke86a9l43c4cl4nqoj8tqbg: app "vm-app-not-real" not found in any monitor
```

Then
```
flightctl get dev/$dev -o yaml | yq '.metadata.annotations'       
device-controller/remote-session: '[{"sessionID":"bf44eb91-5903-4d4f-962e-f00648f6e5ca","appName":"vm-app-not-real","consoleType":"serial"}]'
```

celdrake: Update: with the last commit "fix stale remote-session annotation after VM app console disconnect", the annotation is no longer written though the misleading "Connected to <app> console" is still printed.

internal/agent/device/applications/console/manager.go
```diff
@@ -146,12 +150,20 @@ func (m *Manager) Start(ctx context.Context, entry v1beta1.DeviceRemoteSession)
...
 		consts.GrpcClientNameKey, m.deviceName,
 		consts.GrpcAppNameKey, entry.AppName,
-		consts.GrpcSelectedProtocolKey, entry.ConsoleType,
-	)
+	}
+	// Report resolve failures via metadata (read by the server before any message
+	// exchange) rather than over the stream body, so the server can fail the session
+	// before the client's connection is upgraded instead of only after.
+	if resolveErr != nil {
+		metadataPairs = append(metadataPairs, consts.GrpcSessionErrorKey, resolveErr.Error())
```
asafbss: gRPC-Go validates metadata values before sending them on the wire.
https://github.com/grpc/grpc-go/blob/v1.71.1/internal/metadata/metadata.go#L106

[hasNotPrintable(...)?](https://github.com/grpc/grpc-go/blob/cdbdb759dd67c89544f9081f854c284493b5461c/internal/metadata/metadata.go#L127)

If resolveErr.Error() contains a newline will fail with error

internal/remote_access_server/ws_handler.go
```diff
@@ -98,6 +115,15 @@ func (h *AppConsoleHandler) HandleApplicationConsole(w http.ResponseWriter, r *h
...
 			return
 		}
+	case agentErr := <-session.ErrCh:
+		// The agent reported a session-level failure (e.g. the app does not exist) before
+		// selecting a protocol. Fail the request now, before upgrading to WebSocket, so the
+		// client never sees a false "connected" state for a session that never started.
+		close(session.SendCh)
+		h.log.Infof("app console session for device %s app %s failed before protocol selection with an agent-reported error", deviceName, appName)
+		h.log.Debugf("app console session %s failure detail: %s", session.UUID, agentErr)
+		http.Error(w, agentErr, http.StatusNotFound)
```
asafbss: truncateWSCloseReason?

api/grpc/v1/router_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.2.0
-// - protoc             v3.19.6
+// - protoc             v3.14.0
```
asafbss: Regenerated the proto with an older protoc?

asafbss (commented):
Looks solid overall. I left a couple of comments.

asafbss (approved):

PR #3232, opened 2026-07-14
EDM-4741: Fix edit no-args e2e expectation

eldar101:
Changed one e2e test expectation in test/e2e/cli/cli_test.go.

**Why** 
flightctl edit with no args now returns its custom validation error:
`Error: you must specify a resource to edit (TYPE NAME or TYPE/NAME)`
The test was still expecting Cobra’s generic arg-count error:
`Error: accepts between 1 and 2 arg(s), received 0`

**How**: 
Updated only that assertion to match the actual edit command behavior. No production code or other failures were changed.

## Release target
Target release: release-1.3


hferber1 (approved):
LGTM

PR #3234, opened 2026-07-15
EDM-4772: Remove unnecessary VM setup from selector and RBAC e2e suites

eldar101:
## Summary
- switch selectors, label_selectors, field_selectors, and rbac suites to SetupWorkerHarnessWithoutVM
- remove per-spec SetupVMFromPoolAndStartAgent calls from those suites
- keep per-spec tracing, cleanup, provider setup, and RBAC login/context handling intact

## Why
These suites create and validate API resources directly and do not require a live agent VM for their assertions. Removing the VM setup avoids expensive neutral-device provisioning while preserving per-test cleanup labels and contexts.

The main change was removing VM setup from suites that only validate API-created resources and do not need a live agent. That made the suites much faster, but it also removed the natural delay that VM restore/start used to add between tests.

Once the tests became API-only, resource creation and cleanup happened much more densely. That exposed transient API 429 Rate limit exceeded responses during cleanup, which caused new branch-only AfterEach failures even though the test assertions themselves were unchanged.

The additional harness changes are there to preserve the same result profile as main:
**retry transient CLI/API 429 responses instead of failing cleanup immediately
make cleanup more robust when bulk deletes partially succeed
keep cleanup tied to the per-test test-id
add a focused unit test for the new rate-limit detection behavior
After those fixes, the branch has the same results as main, but the runtime drops from about 36m13s to 11m22s.**

## Release target
Target release: release-1.3

### DIFF
--- a/test/e2e/field_selectors/field_selectors_suite_test.go
+++ b/test/e2e/field_selectors/field_selectors_suite_test.go
@@ -19,8 +19,8 @@ func TestFieldSelector(t *testing.T) {
 var _ = BeforeSuite(func() {
 	auxFuture := e2e.StartAuxServicesAsync(context.Background())
 	Expect(setup.EnsureDefaultProviders(nil)).To(Succeed())
-	// Setup VM and harness for this worker
-	e2e.SetupWorkerHarnessOrAbort()
+	_, _, err := e2e.SetupWorkerHarnessWithoutVM()
+	Expect(err).ToNot(HaveOccurred())
 	auxFuture.Wait()
 })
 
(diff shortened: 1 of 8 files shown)


eldar101:
Ran 88 of 88 Specs in 572.718 seconds
88 Passed | 0 Failed | 0 Pending | 0 Skipped
=============================================================
Random Seed: 1784109874

itzikb-redhat:
I wonder whether we shouldn't set rate limit false when testing

eldar101:
> I wonder whether we shouldn't set rate limit false when testing


I would avoid disabling it.

Because the e2e env should behave like the real deployed product and maybe it has it's own realistic rate limit to API calls.

itzikb-redhat:
Other than that lgtm

hferber1 (approved):
LGTM

PR #3237, opened 2026-07-15
Add flightctl telemetry gateway network policy

itzikb-redhat:
Add a policy to telemetry gateway otherwise a device can't reach the telemetry unless NP is configured
Also add network policies for:
- flightctl-remote-access-networkpolicy
- flightctl-cli-artifacts-networkpolicy
- flightctl-alertmanager-proxy-networkpolicy

### DIFF
--- a/deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-networkpolicy.yaml
+++ b/deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-networkpolicy.yaml
@@ -1,0 +1,25 @@
+{{- if and .Values.alertManager .Values.alertManager.enabled }}
+{{ if and (eq (include "flightctl.enableMulticlusterExtensions" .) "false") (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: flightctl-alertmanager-proxy-from-ingress
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+    flightctl.service: flightctl-alertmanager-proxy
+spec:
+  ingress:
+  - from:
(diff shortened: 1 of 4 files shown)


fzdarsky:
This should only be the case for OpenShift clusters with an installed deny all NetworkPolicy. However, this is common practice, so we definitely need to fix this, thanks!

Please check the other externally facing components, too, I think others need that same NetworkPolicy.

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-networkpolicy.yaml
```diff
@@ -0,0 +1,23 @@
...
+    {{- include "flightctl.standardLabels" . | nindent 4 }}
+    flightctl.service: flightctl-alertmanager-proxy
+spec:
+  ingress:
+  - from:
+    - podSelector: {}
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          network.openshift.io/policy-group: ingress
```
asafbennatan: New files copy the legacy OpenShift network.openshift.io/policy-group: ingress label instead of the current-preferred policy-group.network.openshift.io/ingress: "" already used by the chart's own UI policy

deploy/helm/flightctl/templates/alertmanager/flightctl-alertmanager-proxy-networkpolicy.yaml
```diff
@@ -0,0 +1,23 @@
+{{ if and (eq (include "flightctl.enableMulticlusterExtensions" .) "false") (eq (include "flightctl.getServiceExposeMethod" .) "route") }}
```
asafbennatan, line 2: 3 of 4 new policies (alertmanager-proxy, cli-artifacts, remote-access) omit the .Values.<service>.enabled guard that every other template in their directories applies

asafbennatan (approved):
LGTM

eldar101 (approved):

PR #3242, opened 2026-07-17
EDM-4774: Implement encryption-at-rest handlers for Repository and Au…

asafbss:
…thProvider fields

Register GORM beforeSave encryption handlers for Repository and AuthProvider models so sensitive fields are encrypted before being persisted to PostgreSQL.

Repository handler encrypts: httpConfig.password, httpConfig.token, httpConfig.tls.key, httpConfig.tls.crt, sshConfig.sshPrivateKey, sshConfig.privateKeyPassphrase, ociAuth.password.

AuthProvider handler encrypts: clientSecret (OIDC, OAuth2, OpenShift, AAP).

### DIFF
--- a/internal/auth/authn/oauth2_auth.go
+++ b/internal/auth/authn/oauth2_auth.go
@@ -18,6 +18,7 @@ import (
 	api "github.com/flightctl/flightctl/api/core/v1beta1"
 	"github.com/flightctl/flightctl/internal/auth/common"
 	"github.com/flightctl/flightctl/internal/identity"
+	"github.com/flightctl/flightctl/internal/instrumentation/encryption"
 	"github.com/jellydator/ttlcache/v3"
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jwt"
@@ -60,6 +61,12 @@ func NewOAuth2Auth(metadata api.ObjectMeta, spec api.OAuth2ProviderSpec, tlsConf
 	if spec.ClientSecret == "" {
 		return nil, fmt.Errorf("clientSecret is required")
 	}
+
+	decryptedSecret, _, err := encryption.Decrypt(context.Background(), encryption.Ciphertext(spec.ClientSecret))
+	if err != nil {
+		return nil, fmt.Errorf("decrypt clientSecret: %w", err)
+	}
+	spec.ClientSecret = string(decryptedSecret)
 
 	// Convert organization assignment to org config
 	orgConfig := convertOrganizationAssignmentToOrgConfig(spec.OrganizationAssignment)
--- a/internal/auth/authn/oauth2_auth.go
+++ b/internal/auth/authn/oauth2_auth.go
@@ -62,12 +62,6 @@ func NewOAuth2Auth(metadata api.ObjectMeta, spec api.OAuth2ProviderSpec, tlsConf
 		return nil, fmt.Errorf("clientSecret is required")
 	}
 
-	decryptedSecret, _, err := encryption.Decrypt(context.Background(), encryption.Ciphertext(spec.ClientSecret))
-	if err != nil {
-		return nil, fmt.Errorf("decrypt clientSecret: %w", err)
-	}
-	spec.ClientSecret = string(decryptedSecret)
-
 	// Convert organization assignment to org config
 	orgConfig := convertOrganizationAssignmentToOrgConfig(spec.OrganizationAssignment)
 
@@ -136,6 +130,14 @@ func NewOAuth2Auth(metadata api.ObjectMeta, spec api.OAuth2ProviderSpec, tlsConf
 	}
 
 	return oauth2Auth, nil
+}
+
+func (o *OAuth2Auth) decryptClientSecret(ctx context.Context) (string, error) {
+	plaintext, _, err := encryption.Decrypt(ctx, encryption.Ciphertext(o.spec.ClientSecret))
+	if err != nil {
+		return "", fmt.Errorf("decrypt clientSecret: %w", err)
+	}
+	return string(plaintext), nil
 }
 
 func (o *OAuth2Auth) IsEnabled() bool {
@@ -431,7 +433,11 @@ func (o *OAuth2Auth) introspectRFC7662(ctx context.Context, token string, spec a
 	}
 
 	// RFC 7662 requires client authentication via Basic Auth
-	req.SetBasicAuth(o.spec.ClientId, o.spec.ClientSecret)
+	plainSecret, err := o.decryptClientSecret(ctx)
+	if err != nil {
+		return err
+	}
+	req.SetBasicAuth(o.spec.ClientId, plainSecret)
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 	req.Header.Set("Accept", "application/json")
 
@@ -485,7 +491,11 @@ func (o *OAuth2Auth) introspectGitHub(ctx context.Context, token string, spec ap
 	}
 
 	// GitHub requires Basic Auth with client ID and secret
-	req.SetBasicAuth(o.spec.ClientId, o.spec.ClientSecret)
+	plainSecret, err := o.decryptClientSecret(ctx)
+	if err != nil {
+		return err
+	}
+	req.SetBasicAuth(o.spec.ClientId, plainSecret)
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Accept", "application/vnd.github+json")
 
--- a/internal/store/gorm.go
+++ b/internal/store/gorm.go
@@ -13,6 +13,7 @@ import (
 	"github.com/flightctl/flightctl/internal/domain"
 	"github.com/flightctl/flightctl/internal/flterrors"
 	"github.com/flightctl/flightctl/internal/instrumentation/encryption"
+	"github.com/flightctl/flightctl/internal/store/model"
 	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/sirupsen/logrus"
 	"go.opentelemetry.io/otel/trace"
@@ -160,7 +161,7 @@ func initDBWithUser(cfg *config.Config, log *logrus.Logger, user string, passwor
 
 	// Register encryption plugin if encryption is initialized
 	if encryption.GlobalManager() != nil {
-		if err = newDB.Use(encryption.NewPlugin(encryption.GlobalManager(), map[string]encryption.ModelEncryptHandler{})); err != nil {
+		if err = newDB.Use(encryption.NewPlugin(encryption.GlobalManager(), model.EncryptionHandlers())); err != nil {
 			sqlDB.Close()
 			log.Errorf("failed to register encryption plugin: %v", err)
 			return nil, fmt.Errorf("failed to register encryption plugin: %w", err)
--- a/internal/store/model/encryption.go
+++ b/internal/store/model/encryption.go
@@ -13,14 +13,13 @@ import (
 // This registry is used by the GORM encryption plugin to know how to encrypt each model.
 func EncryptionHandlers() map[string]encryption.ModelEncryptHandler {
 	return map[string]encryption.ModelEncryptHandler{
-		"Repository":   encryptRepository,
-		"AuthProvider": encryptAuthProvider,
+		domain.RepositoryKind:   encryptRepository,
+		domain.AuthProviderKind: encryptAuthProvider,
 	}
 }
 
 // repositoryEncryptPaths lists the sensitive fields in the Repository Spec JSONB
-// that must be encrypted at rest. These match the fields redacted by HideSensitiveData
-// in api/core/v1beta1/util.go.
+// that must be encrypted at rest.
 // Each entry is a slice of JSON key segments (e.g. {"httpConfig", "tls.key"}).
 var repositoryEncryptPaths = [][]string{
 	{"httpConfig", "password"},
--- a/internal/store/model/encryption.go
+++ b/internal/store/model/encryption.go
@@ -1,0 +1,149 @@
+package model
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/flightctl/flightctl/internal/instrumentation/encryption"
+)
+
+// EncryptionHandlers maps model type name to its encryption handler.
+// This registry is used by the GORM encryption plugin to know how to encrypt each model.
+func EncryptionHandlers() map[string]encryption.ModelEncryptHandler {
+	return map[string]encryption.ModelEncryptHandler{
+		"Repository":   encryptRepository,
+		"AuthProvider": encryptAuthProvider,
+	}
+}
+
+// repositoryEncryptPaths lists the sensitive fields in the Repository Spec JSONB
+// that must be encrypted at rest. These match the fields redacted by HideSensitiveData
+// in api/core/v1beta1/util.go.
+// Each entry is a slice of JSON key segments (e.g. {"httpConfig", "tls.key"}).
+var repositoryEncryptPaths = [][]string{
+	{"httpConfig", "password"},
+	{"httpConfig", "token"},
+	{"httpConfig", "tls.key"},
+	{"httpConfig", "tls.crt"},
+	{"sshConfig", "sshPrivateKey"},
+	{"sshConfig", "privateKeyPassphrase"},
+	{"ociAuth", "password"},
+}
+
+// authProviderEncryptPaths lists the sensitive fields in the AuthProvider Spec JSONB.
+// clientSecret appears at the top level of all provider type variants (OIDC, OAuth2,
+// OpenShift, AAP).
+var authProviderEncryptPaths = [][]string{
+	{"clientSecret"},
+}
+
+func encryptRepository(ctx context.Context, v interface{}, encrypt encryption.EncryptFunc) error {
+	repo, ok := v.(*Repository)
+	if !ok {
+		return fmt.Errorf("expected *Repository, got %T", v)
+	}
+
+	if repo.Spec == nil {
+		return nil
+	}
+
+	return encryptJSONField(ctx, &repo.Spec.Data, repositoryEncryptPaths, encrypt)
+}
+
+func encryptAuthProvider(ctx context.Context, v interface{}, encrypt encryption.EncryptFunc) error {
+	ap, ok := v.(*AuthProvider)
+	if !ok {
+		return fmt.Errorf("expected *AuthProvider, got %T", v)
+	}
+
+	if ap.Spec == nil {
+		return nil
+	}
+
+	return encryptJSONField(ctx, &ap.Spec.Data, authProviderEncryptPaths, encrypt)
+}
+
+// encryptJSONField encrypts string values at the given paths within a JSONB field.
+// It marshals the data to a generic map, encrypts matching paths, and unmarshals back.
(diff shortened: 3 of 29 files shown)


internal/auth/authn/oauth2_auth.go
```diff
@@ -61,6 +62,12 @@ func NewOAuth2Auth(metadata api.ObjectMeta, spec api.OAuth2ProviderSpec, tlsConf
 		return nil, fmt.Errorf("clientSecret is required")
 	}
 
+	decryptedSecret, _, err := encryption.Decrypt(context.Background(), encryption.Ciphertext(spec.ClientSecret))
+	if err != nil {
+		return nil, fmt.Errorf("decrypt clientSecret: %w", err)
+	}
+	spec.ClientSecret = string(decryptedSecret)
```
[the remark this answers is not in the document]
asafbss: These are init-time constructors called during service startup — there is no request context, cancellation, or tracing spans to propagate at that point. Using context.Background() is intentional here. Threading a ctx parameter through the init chain would add complexity without benefit since these run once at startup, not on the request path.

test/integration/store/repository_test.go
```diff
@@ -25,6 +26,13 @@ import (
 	"gorm.io/gorm"
 )
 
+// decryptString decrypts an encrypted string field value, passing through plaintext unchanged.
+func decryptString(val string) string {
+	plaintext, _, err := encryption.Decrypt(context.Background(), encryption.Ciphertext(val))
+	Expect(err).ToNot(HaveOccurred())
+	return string(plaintext)
+}
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/tasks/repotester.go
```diff
@@ -307,13 +311,13 @@ func (r *OciRepoTester) authenticateDocker(client *http.Client, v2URL string, in
 	}
 
 	// Get token from auth endpoint
-	token, err := r.getToken(client, realm, service, username, password)
+	token, err := r.getToken(ctx, client, realm, service, username, password)
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/tasks/git_helpers.go
```diff
@@ -37,9 +38,9 @@ import (
 var scpLikeUrlRegExp = regexp.MustCompile(`^(?:(?P<user>[^@]+)@)?(?P<host>[^:\s]+):(?:(?P<port>[0-9]{1,5}):)?(?P<path>[^\\].*)$`)
 
 // a function to clone a git repo, for mockable unit testing
-type cloneGitRepoFunc func(repo *domain.Repository, revision *string, depth *int, cfg *config.Config) (billy.Filesystem, string, error)
+type cloneGitRepoFunc func(ctx context.Context, repo *domain.Repository, revision *string, depth *int, cfg *config.Config) (billy.Filesystem, string, error)
 
-func CloneGitRepo(repo *domain.Repository, revision *string, depth *int, cfg *config.Config) (billy.Filesystem, string, error) {
+func CloneGitRepo(ctx context.Context, repo *domain.Repository, revision *string, depth *int, cfg *config.Config) (billy.Filesystem, string, error) {
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/store/model/encryption.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := json.Unmarshal(jsonBytes, &jsonData); err != nil {
+		return fmt.Errorf("unmarshal spec to map: %w", err)
+	}
+
+	modified := false
+	for _, path := range paths {
+		encrypted, err := encryptJSONPath(ctx, jsonData, path, encrypt)
+		if err != nil {
+			continue
+		}
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/store/model/encryption.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/flightctl/flightctl/internal/instrumentation/encryption"
+)
+
+// EncryptionHandlers maps model type name to its encryption handler.
+// This registry is used by the GORM encryption plugin to know how to encrypt each model.
+var EncryptionHandlers = map[string]encryption.ModelEncryptHandler{
+	"Repository":   encryptRepository,
+	"AuthProvider": encryptAuthProvider,
+}
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/tasks/git_helpers.go
```diff
@@ -315,19 +337,27 @@ func GetAuth(repository *domain.Repository, cfg *config.Config) (transport.AuthM
...
 		if err != nil {
 			return err
 		}
 
-		key, err := base64.StdEncoding.DecodeString(*httpConfig.TlsKey)
+		decryptedKey, _, err := encryption.Decrypt(ctx, encryption.Ciphertext(*httpConfig.TlsKey))
+		if err != nil {
+			return fmt.Errorf("decrypt TLS key: %w", err)
+		}
+		key, err := base64.StdEncoding.DecodeString(string(decryptedKey))
```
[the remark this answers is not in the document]
asafbss: CaCrt is not encrypted at rest — it's not in repositoryEncryptPaths. Only secret/sensitive fields are encrypted: private keys (tls.key, sshPrivateKey), passwords, tokens, passphrases, and tls.crt (client certificate). CA certificates are public trust anchors used to verify server identity, not secrets, so they don't need encrypt/decrypt.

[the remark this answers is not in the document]
asafbss: Thanks you, check again please

internal/store/model/encryption_test.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			modelRepo, err := NewRepositoryFromApiResource(tc.apiRepo)
+			require.NoError(t, err)
+
+			handler := EncryptionHandlers()["Repository"]
+			err = handler(context.Background(), modelRepo, mgr.ProcessEncryption)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
asafbss: Thanks you, check again please

internal/tasks/git_helpers.go
```diff
@@ -235,14 +236,25 @@ func GetAuth(repository *domain.Repository, cfg *config.Config) (transport.AuthM
...
+			// Decrypt SSH private key (passthrough if plaintext for BC)
+			decryptedKey, _, err := encryption.Decrypt(ctx, encryption.Ciphertext(*gitSpec.SshConfig.SshPrivateKey))
+			if err != nil {
+				return nil, fmt.Errorf("decrypt SSH private key: %w", err)
+			}
+
+			sshPrivateKey, err := base64.StdEncoding.DecodeString(string(decryptedKey))
 			if err != nil {
 				return nil, err
 			}
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/tasks/http_helpers.go
```diff
@@ -56,18 +59,34 @@ func buildHttpRepoRequestAuth(repoHttpSpec domain.HttpRepoSpec, req *http.Reques
...
 		if err != nil {
 			return nil, tlsConfig, err
 		}
 
-		key, err := base64.StdEncoding.DecodeString(*repoHttpSpec.HttpConfig.TlsKey)
+		decryptedKey, _, err := encryption.Decrypt(ctx, encryption.Ciphertext(*repoHttpSpec.HttpConfig.TlsKey))
+		if err != nil {
+			return nil, tlsConfig, fmt.Errorf("decrypt TLS key: %w", err)
+		}
+		key, err := base64.StdEncoding.DecodeString(string(decryptedKey))
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/tasks/repotester_test.go
```diff
@@ -1,17 +1,56 @@
...
+
+	logger := logrus.New()
+	logger.SetLevel(logrus.FatalLevel)
+
+	if err := encryption.InitGlobalEncryption(logger, cfg); err != nil {
+		panic(err)
+	}
+
+	os.Exit(m.Run())
+}
```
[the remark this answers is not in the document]
asafbss: Thank you, check again please

internal/tasks/repotester.go
```diff
@@ -302,18 +307,23 @@ func (r *OciRepoTester) authenticateDocker(client *http.Client, v2URL string, in
...
+		}
+		pw := string(decryptedPassword)
 		username = &dockerAuth.Username
-		password = &dockerAuth.Password
+		password = &pw
 	}
 
 	// Get token from auth endpoint
-	token, err := r.getToken(client, realm, service, username, password)
+	token, err := r.getToken(ctx, client, realm, service, username, password)
```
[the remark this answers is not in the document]
asafbss: Valid concern, but this is a pre-existing behavior unrelated to the encryption changes in this PR. The HTTPS-to-HTTP downgrade risk exists regardless of whether the password is encrypted at rest. I'd recommend tracking this as a separate issue to avoid scope creep — it needs its own design consideration around proxy setups, development environments, and SkipServerVerification semantics.

internal/tasks/git_helpers.go
```diff
@@ -315,21 +337,29 @@ func GetAuth(repository *domain.Repository, cfg *config.Config) (transport.AuthM
...
 
-		key, err := base64.StdEncoding.DecodeString(*httpConfig.TlsKey)
+		decryptedKey, _, err := encryption.Decrypt(ctx, encryption.Ciphertext(*httpConfig.TlsKey))
 		if err != nil {
-			return err
+			return fmt.Errorf("decrypt TLS key: %w", err)
+		}
+		key, err := base64.StdEncoding.DecodeString(string(decryptedKey))
+		if err != nil {
+			return fmt.Errorf("decode TLS key: %w", err)
```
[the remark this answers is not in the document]
asafbss: This is a pre-existing design concern unrelated to the encryption changes. The InstallProtocol global mutation existed before this PR — we only added encryption.Decrypt calls around the existing TLS configuration logic. Scoping the transport per-clone would be a worthwhile improvement but belongs in a separate issue to keep this PR focused on encryption at rest.

internal/tasks/http_helpers.go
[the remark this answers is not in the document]
asafbss: okay added, check again

internal/auth/authn/oauth2_auth.go
```diff
@@ -61,6 +62,12 @@ func NewOAuth2Auth(metadata api.ObjectMeta, spec api.OAuth2ProviderSpec, tlsConf
...
+
```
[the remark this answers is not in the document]
asafbss: Encryption is not optional — it's a required component that must always be initialized. If GlobalManager() is nil, that's a configuration/startup error that should fail loudly,

internal/store/model/encryption.go
```diff
@@ -0,0 +1,173 @@
...
+	{"httpConfig", "tls.crt"},
+	{"sshConfig", "sshPrivateKey"},
+	{"sshConfig", "privateKeyPassphrase"},
+	{"ociAuth", "password"},
+}
+
+// authProviderEncryptPaths lists the sensitive fields in the AuthProvider Spec JSONB.
+// clientSecret appears at the top level of all provider type variants (OIDC, OAuth2,
+// OpenShift, AAP).
+var authProviderEncryptPaths = [][]string{
```
asafbennatan, line 37: Should we also decrypt `clientSecret` for OIDC (and AAP if applicable) before use?

internal/auth/authn/oauth2_auth.go
asafbennatan: Should we keep ciphertext on the compare path (or decrypt the DB copy before comparison) so MultiAuth change detection stays stable?

internal/store/model/encryption.go
```diff
@@ -0,0 +1,173 @@
...
+// in api/core/v1beta1/util.go.
+// Each entry is a slice of JSON key segments (e.g. {"httpConfig", "tls.key"}).
+var repositoryEncryptPaths = [][]string{
+	{"httpConfig", "password"},
+	{"httpConfig", "token"},
+	{"httpConfig", "tls.key"},
+	{"httpConfig", "tls.crt"},
+	{"sshConfig", "sshPrivateKey"},
+	{"sshConfig", "privateKeyPassphrase"},
+	{"ociAuth", "password"},
```
asafbennatan, line 31: Should we also decrypt `ociAuth.password` at the other OCI consumers (not only RepoTester)?

internal/store/gorm.go
```diff
@@ -160,7 +161,7 @@ func initDBWithUser(cfg *config.Config, log *logrus.Logger, user string, passwor
 
 	// Register encryption plugin if encryption is initialized
 	if encryption.GlobalManager() != nil {
```
asafbennatan, line 163: Should we fail closed here when the encryption manager is nil, now that real handlers are wired?

internal/auth/authn/oauth2_auth.go
```diff
@@ -61,6 +62,12 @@ func NewOAuth2Auth(metadata api.ObjectMeta, spec api.OAuth2ProviderSpec, tlsConf
 		return nil, fmt.Errorf("clientSecret is required")
 	}
 
+	decryptedSecret, _, err := encryption.Decrypt(context.Background(), encryption.Ciphertext(spec.ClientSecret))
```
asafbennatan: Should we pick one AuthProvider decrypt contract and apply it everywhere?

Today OAuth2/OpenShift decrypt in the constructor (so `Get*Spec()` returns plaintext), while OIDC does not decrypt at all and the DB still holds ciphertext. That makes MultiAuth comparison and `auth_token_proxy` see different shapes of the secret.

Options:
1. Decrypt once at load into the runtime object, and make change detection compare ciphertext-to-ciphertext (or a dedicated field).
2. Keep ciphertext in specs/`Get*Spec()`, and decrypt only at token-proxy / introspection call sites.

internal/store/model/encryption.go
```diff
@@ -0,0 +1,173 @@
...
+// This registry is used by the GORM encryption plugin to know how to encrypt each model.
+func EncryptionHandlers() map[string]encryption.ModelEncryptHandler {
+	return map[string]encryption.ModelEncryptHandler{
+		"Repository":   encryptRepository,
+		"AuthProvider": encryptAuthProvider,
+	}
+}
+
+// repositoryEncryptPaths lists the sensitive fields in the Repository Spec JSONB
+// that must be encrypted at rest. These match the fields redacted by HideSensitiveData
```
asafbennatan: Should we rephrase this so it doesn't imply exact set equality with `HideSensitiveData` (e.g. note that `ca.crt` is intentionally excluded)?

internal/store/model/encryption.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/flightctl/flightctl/internal/domain"
+	"github.com/flightctl/flightctl/internal/instrumentation/encryption"
+)
+
+// EncryptionHandlers maps model type name to its encryption handler.
+// This registry is used by the GORM encryption plugin to know how to encrypt each model.
+func EncryptionHandlers() map[string]encryption.ModelEncryptHandler {
+	return map[string]encryption.ModelEncryptHandler{
+		"Repository":   encryptRepository,
```
asafbennatan: Should we key these handlers with `domain.RepositoryKind` / `domain.AuthProviderKind` instead of string literals?

internal/store/model/encryption.go
```diff
@@ -0,0 +1,173 @@
...
+		"AuthProvider": encryptAuthProvider,
+	}
+}
+
+// repositoryEncryptPaths lists the sensitive fields in the Repository Spec JSONB
+// that must be encrypted at rest. These match the fields redacted by HideSensitiveData
+// in api/core/v1beta1/util.go.
+// Each entry is a slice of JSON key segments (e.g. {"httpConfig", "tls.key"}).
+var repositoryEncryptPaths = [][]string{
```
asafbennatan, line 24: Should we drive these paths from struct tags + reflection (e.g. `sensitive:"true"` on the generated types) so encrypt and HideSensitiveData can't diverge?

asafbennatan (commented):
Encryption-at-rest wiring looks solid overall. Left a few questions on decrypt-at-use coverage and AuthProvider boundary consistency.

internal/store/model/encryption.go
asafbss: Agreed, the two lists shouldn't diverge. A reflection-based approach with struct tags would unify the source of truth, but it adds runtime overhead and complexity, especially since some of these types are generated from OpenAPI. We can evaluate it as a follow-up but not sure it's worth the tradeoff for the current number of fields.

internal/store/model/encryption.go
asafbss: Good catch

internal/store/gorm.go
asafbss: Yes

internal/store/model/encryption.go
asafbss: Good catch, ty

internal/auth/authn/oauth2_auth.go
asafbss: Fixed. We moved decrypt out of the constructors. specs now hold ciphertext. Both sides of hasProviderChanged compare ciphertext-to-ciphertext, so change detection is stable.

internal/store/model/encryption.go
asafbss: Thank you, fixed.
All four providers' clientSecret is decrypted at the point of use in auth_token_proxy.go:proxyTokenRequest, which handles token exchange for OIDC, OAuth2, OpenShift, and AAP. OAuth2 also decrypts in its introspection methods (introspectRFC7662, introspectGitHub).

internal/auth/authn/oauth2_auth.go
asafbss: https://github.com/flightctl/flightctl/pull/3242#discussion_r3613277718

internal/store/gorm.go
```diff
@@ -158,13 +159,16 @@ func initDBWithUser(cfg *config.Config, log *logrus.Logger, user string, passwor
...
-		if err = newDB.Use(encryption.NewPlugin(encryption.GlobalManager(), map[string]encryption.ModelEncryptHandler{})); err != nil {
-			sqlDB.Close()
-			log.Errorf("failed to register encryption plugin: %v", err)
-			return nil, fmt.Errorf("failed to register encryption plugin: %w", err)
-		}
+	// Register encryption plugin — fail closed if encryption is not initialized,
+	// since decrypt-at-use call sites depend on it at runtime.
+	if encryption.GlobalManager() == nil {
+		sqlDB.Close()
+		return nil, fmt.Errorf("encryption manager not initialized: encryption is required for database operations")
```
asafbennatan: Should we also call `InitGlobalEncryption` in the restore path before `store.InitDB`?

`flightctl-restore` restores encryption keys, then calls `InitDB` without initializing the encryption manager — with this fail-closed check, post-restore device prep would fail.

asafbennatan (commented):
One follow-up question after the decrypt-at-use updates.

internal/store/gorm.go
asafbss: Reverted this check — it also breaks integration test migrations and the restore path, both of which call InitDB/InitMigrationDB without encryption initialized. The guard already exists in each service's main.go where InitGlobalEncryption fatals on failure, so encryption is always initialized before any application DB operations.

asafbennatan (approved):
LGTM

PR #3243, opened 2026-07-19
EDM-3605: Show enrollment QR code only after successful enrollment request creation

EfratIfergan:
## Problem
During device enrollment, the agent displayed a QR code and enrollment URL to the user **before** the enrollment request was created on the server. If the enrollment request creation failed (network issues, TLS errors, server rejection), the user saw a QR code pointing to a non-existent enrollment.

## Root Cause
In `internal/agent/device/lifecycle/manager.go`, the `Initialize` method called `writeEnrollmentBanner()` before `enrollmentRequest()`. The banner was always written regardless of whether the enrollment request succeeded.

## Fix
Swapped the order of `writeEnrollmentBanner()` and `enrollmentRequest()` in the `Initialize` method so the QR code is only displayed after the enrollment request is successfully created on the server.

## Testing
- Added regression test `TestLifecycleManager_Initialize_NoBannerOnEnrollmentFailure` with two cases:
  - When enrollment request creation fails, the banner file is never written
  - When enrollment request succeeds, the banner file is written
- All lifecycle package tests pass
- Full unit test suite passes (5164 tests, 1 pre-existing failure in unrelated `internal/backup` package)
- Manually verified on agent VM:
  - Broken enrollment URL → no QR code shown
  - Correct enrollment URL → QR code appears after successful enrollment request

## Confidence
HIGH — minimal change (reordering two function calls), well-tested, manually verified end-to-end.

## Risk Assessment
LOW — no new logic introduced, only execution order changed within the same error-handling block.

### DIFF
--- a/internal/agent/device/lifecycle/manager.go
+++ b/internal/agent/device/lifecycle/manager.go
@@ -103,8 +103,6 @@ func (m *LifecycleManager) Initialize(ctx context.Context, status *v1beta1.Devic
 			return err
 		}
 
-		// EDM-3605: write the enrollment banner only after the enrollment request
-		// has been successfully created on the server.
 		if err := m.writeEnrollmentBanner(ctx); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


asafbennatan (approved):
LGTM

PR #3245, opened 2026-07-19
EDM-4842: Fix edit argument validation e2e assertions

eldar101:
## Summary
- reuse the edit command-specific validation error in the CLI e2e test
- update the too-many-arguments edit assertion to match current edit command behavior

## Why
`flightctl edit` uses a custom Args validator for both missing and excessive arguments, so both cases return `Error: you must specify a resource to edit (TYPE NAME or TYPE/NAME)`. The logs from the failing profiles show the stale Cobra arg-count expectation only on the too-many-args edit check after the no-args assertion passes.

This time we fixed the full edit arg-count case, not just the first failure.
Before, we only updated the no-args assertion. The test then moved forward and failed on the too-many-args assertion.
Now we added one shared expected error for flightctl edit and use it for both:
flightctl edit
flightctl edit 1 2 3 4
That matches the actual edit command code, where both cases return the same custom validation error.

## Release target
Target release: release-1.3

## AI attribution
This change was assisted by OpenAI Codex and reviewed by Eldar Weiss.

### DIFF
--- a/test/e2e/cli/cli_test.go
+++ b/test/e2e/cli/cli_test.go
@@ -25,6 +25,7 @@ var (
 	unspecifiedResource  = "Error: name must be specified when deleting"
 	resourceCreated      = `(200 OK|201 Created)`
 	invalidResource      = "invalid resource kind"
+	editResourceRequired = "Error: you must specify a resource to edit (TYPE NAME or TYPE/NAME)"
 	strictfipsruntimeTag = "X:strictfipsruntime"
 	applyOperation       = "apply"
 )
@@ -979,7 +980,7 @@ var _ = Describe("cli login", func() {
 		By("failing when no arguments are provided")
 		out, err = harness.CLI("edit")
 		Expect(err).To(HaveOccurred())
-		Expect(out).To(ContainSubstring("Error: you must specify a resource to edit (TYPE NAME or TYPE/NAME)"))
+		Expect(out).To(ContainSubstring(editResourceRequired))
 
 		By("failing on invalid resource kind (numeric)")
 		out, err = harness.CLI("edit", "1234")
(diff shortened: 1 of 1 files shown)


eldar101:
@EfratIfergan 

I looked deeper into the flightctl edit e2e failure. The test originally expected Cobra’s arg-count errors, for example accepts between 1 and 2 arg(s), received 4.
The reason it started failing is that internal/cli/edit.go now has a custom Args validator:
```
if len(args) < 1 || len(args) > 2 {
    return fmt.Errorf("you must specify a resource to edit (TYPE NAME or TYPE/NAME)")
}
```
That means both cases now return the same message:
```
flightctl edit
flightctl edit 1 2 3 4
```
My PR only updates the e2e test to match that current behavior, but I agree with your point that the product message is misleading for the too-many-args case.

Could you consider fixing the CLI validation instead, either by restoring Cobra’s RangeArgs(1, 2) behavior or by splitting the custom validation into two clearer errors? For example, missing args can keep you must specify a resource to edit, while too many args should say something like you must specify only one resource to edit.
If we fix the CLI behavior directly, I can close my test-only PR or adjust it to match your final CLI output.

eldar101:
it passed anyway:
[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report
autogenerated by Ginkgo
  > Enter [ReportAfterSuite] TOP-LEVEL - autogenerated by Ginkgo @ 07/19/26 14:16:44.195
  < Exit [ReportAfterSuite] TOP-LEVEL - autogenerated by Ginkgo @ 07/19/26 14:16:44.229 (34ms)
[ReportAfterSuite] PASSED [0.034 seconds]
------------------------------

Ran 2 of 28 Specs in 142.743 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 26 Skipped
PASS

Ginkgo ran 1 suite in 2m58.690571359s
Test Suite Passed

EfratIfergan:
@eldar101 I agree that separating the validation cases would provide a better user experience, with a more accurate error message for the "too many arguments" case. I think it would make sense to open a dedicated issue for that, since it is outside the scope of this PR.

the current tests look good and this PR fixes the original issue by aligning the e2e test with the current CLI behavior. We can merge this PR and address the validation improvement in a separate PR

eldar101:
Opened bug waiting for fix 
[EDM-4846](https://redhat.atlassian.net/browse/EDM-4846)

eldar101:
Closing as 
https://github.com/flightctl/flightctl/pull/3279
will resolve the need for this PR.

PR #3246, opened 2026-07-19
EDM-4754: Verify package-mode and image-mode RPM install paths

galel12:
**Jira:** https://redhat.atlassian.net/browse/EDM-4754
**Story type:** [DEV]
**Depends on:** #3244 (EDM-4752 - greenboot subpackage split)

### Summary

Fixes the agent systemd unit so it starts on package-mode installs (where `flightctl-greenboot` is absent), and adds automated container-based verification of both RPM install paths via a Go test.

### Changes

**Service unit fix:**
- `packaging/systemd/flightctl-agent.service` - Fail-closed `ExecStartPre` wrapper: skips silently when `mask-bootc-timer.sh` is absent (package-mode), fails hard when present but exits non-zero (image-mode).

**Go-based packaging verification:**
- `test/packaging/rpm_install_test.go` - Container-based RPM install verification with 4 test cases:
  1. Package-mode install (`install_weak_deps=False`) - agent installs without greenboot, `mask-bootc-timer.sh` is absent
  2. Image-mode install (default weak deps) - greenboot pulled via Recommends
  3. Upgrade path mechanism - RPM metadata proves Recommends + version equality lock
  4. Agent unit smoke test - fail-closed ExecStartPre wrapper + `systemd-analyze verify`
- `Makefile` - `verify-rpm-install` target invokes `go test ./test/packaging/...`

### Testing

- **Packaging verification:** 4 Go test cases (all pass locally)
- **Coverage:** All acceptance criteria exercised by the test
- **Note:** `systemd-analyze verify` reports warnings about `RestartMaxDelaySec`/`RestartSteps` on CS9 (systemd 252 < 253) - pre-existing and benign

### How to run locally

```bash
make rpm                    # build RPMs (includes flightctl-greenboot)
make verify-rpm-install     # run the Go test (requires podman)
```

### Acceptance Criteria

- [x] AC-1: Automated test verifies package-mode install (`install_weak_deps=False`) succeeds without greenboot
- [x] AC-2: Automated test verifies image-mode install pulls `flightctl-greenboot` when available
- [x] AC-3: Upgrade path from existing monolithic install installs greenboot subpackage when greenboot was present (mechanism verification via RPM metadata)
- [x] AC-4: Agent systemd unit starts after package-mode install (smoke test)

- [ ] **`release-X.Y`** - e.g., `release-1.2`, `release-1.3`


packaging/systemd/flightctl-agent.service
```diff
@@ -11,7 +11,9 @@ StartLimitBurst=1000
 
 [Service]
 # Mask bootc auto-update timer before agent runs (bootc/composefs may not persist %post/oneshot enable).
-ExecStartPre=/usr/libexec/flightctl/mask-bootc-timer.sh
+# The "-" prefix makes this non-fatal: on package-mode installs without flightctl-greenboot,
+# mask-bootc-timer.sh is absent and the agent must still start.
+ExecStartPre=-/usr/libexec/flightctl/mask-bootc-timer.sh
```
asafbennatan: The `-` prefix makes *all* non-zero exits from `mask-bootc-timer.sh` non-fatal, including when the script exists on image-mode installs. That is broader than “script missing on package-mode.”

If we want fail-closed when the script is present, something like:

```systemd
ExecStartPre=/bin/bash -c 'test ! -x /usr/libexec/flightctl/mask-bootc-timer.sh || exec /usr/libexec/flightctl/mask-bootc-timer.sh'
```

If fail-open on image-mode is intentional, worth calling out explicitly in the PR description.

.github/workflows/verify-rpm-install.yaml
```diff
@@ -0,0 +1,49 @@
+name: "Verify RPM Install Paths"
```
asafbennatan: We already have a lot of workflows. I’d rather not add a dedicated one for this.

If CI coverage is needed, fold a verify job into the existing packaging workflow (`rpmlint.yaml`) and reuse `build-rpms.yaml`, instead of introducing `verify-rpm-install.yaml`. The Makefile/local target can still be the primary entrypoint.

test/scripts/verify-rpm-install.sh
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env bash
```
asafbennatan: I’d prefer we don’t add a new shell script here — we should be moving packaging/install verification toward Go (e.g. under `test/packaging/` or similar), even if it still shells out to `podman`/`dnf`/`rpm`.

When rewriting in Go, please keep the AC coverage, and tighten a few things the bash version currently under-asserts:

- **AC-1:** also assert `/usr/libexec/flightctl/mask-bootc-timer.sh` is absent in package-mode.
- **AC-3:** assert a real version equality lock (`flightctl-agent = …`), not just a `flightctl-agent` substring in Requires.
- **AC-4:** either actually exercise unit start (systemd-in-container), or rename the AC so it matches static checks (`ExecStartPre` soft-fail + `systemd-analyze verify`) rather than claiming “unit starts.”

Makefile
```diff
@@ -486,6 +487,10 @@ rpmlint-ci:
 check-rpmlint:
 	@command -v rpmlint > /dev/null || (echo "rpmlint not found. Install with: sudo apt-get install rpmlint (Ubuntu/Debian) or sudo dnf install rpmlint (Fedora/RHEL)" && exit 1)
 
+.PHONY: verify-rpm-install
+verify-rpm-install: bin/.rpm
+	$(ROOT_DIR)/test/scripts/verify-rpm-install.sh
```
asafbennatan: If verification moves to Go, this target should invoke that test (e.g. `go test ./test/packaging/...`) instead of a shell script.

asafbennatan (commented):
1. Keep (and possibly refine) the systemd unit fix.
2. Prefer not adding another standalone workflow; fold packaging verification into existing packaging CI (e.g. `rpmlint.yaml`) if/when needed.
3. Prefer a Go-based packaging test over a new shell script under `test/scripts/`.

Inline comments have more detail. Leaving this as a comment review (not requesting changes) so you can choose how to reshape the verification.

packaging/systemd/flightctl-agent.service
galel12: Done. Switched to the fail-closed bash wrapper:
```
ExecStartPre=/bin/bash -c 'test ! -x /usr/libexec/flightctl/mask-bootc-timer.sh || exec /usr/libexec/flightctl/mask-bootc-timer.sh'
```

.github/workflows/verify-rpm-install.yaml
galel12: Done. Removed the standalone workflow. The Makefile target (`make verify-rpm-install`) is the primary entrypoint now. Can fold into `rpmlint.yaml` later if CI coverage is needed.

test/scripts/verify-rpm-install.sh
galel12: Done. Rewrote as Go test at `test/packaging/rpm_install_test.go`. All AC coverage kept. Tightened assertions:
- AC-1: added `assertFileAbsent` for `mask-bootc-timer.sh` in package-mode.
- AC-3: now checks for version equality lock (`flightctl-agent` + `=` on the same requires line).
- AC-4: renamed to "smoke test" to match what it does (static ExecStartPre check + `systemd-analyze verify`).

Makefile
galel12: Done. Now runs `go test -v -count=1 -timeout 15m ./test/packaging/...`

Makefile
```diff
@@ -486,6 +487,10 @@ rpmlint-ci:
 check-rpmlint:
 	@command -v rpmlint > /dev/null || (echo "rpmlint not found. Install with: sudo apt-get install rpmlint (Ubuntu/Debian) or sudo dnf install rpmlint (Fedora/RHEL)" && exit 1)
 
+.PHONY: verify-rpm-install
+verify-rpm-install: bin/.rpm
+	go test -v -count=1 -timeout 15m ./test/packaging/...
```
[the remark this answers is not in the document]
galel12: Skipping. The Makefile target depends on `bin/.rpm` which runs the full RPM build. The Go test's `t.Skipf` is the standard Go idiom for missing prerequisites when running outside of `make`.

packaging/systemd/flightctl-agent.service
```diff
@@ -11,7 +11,8 @@ StartLimitBurst=1000
 
 [Service]
 # Mask bootc auto-update timer before agent runs (bootc/composefs may not persist %post/oneshot enable).
-ExecStartPre=/usr/libexec/flightctl/mask-bootc-timer.sh
+# Fail-closed: runs the script when present (image-mode), skips silently when absent (package-mode).
+ExecStartPre=/bin/bash -c 'test ! -x /usr/libexec/flightctl/mask-bootc-timer.sh || exec /usr/libexec/flightctl/mask-bootc-timer.sh'
```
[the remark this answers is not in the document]
galel12: Out of scope for this PR. The agent manages the entire host (bootc, systemd units, device config). Hardening directives like `ProtectSystem` or `ProtectHome` would break core functionality.

asafbennatan (approved):
LGTM

PR #3249, opened 2026-07-19
EDM-3677: Surface fleet rollout errors in device status

EfratIfergan:
## Problem

When a fleet template references a device label that doesn't exist (e.g. `{{ .metadata.labels.domainname }}`), the template rendering error is stored in the `fleet-controller/lastRolloutError` annotation but `status.updated.info` shows a generic "Device has not yet been scheduled for update to the fleet's latest spec." message instead of the actual error.

## Root Cause

Three issues combined to hide the error:

1. **Missing status recomputation:** After writing the `lastRolloutError` annotation in `fleet_rollout.go`, `UpdateServerSideDeviceStatus` was never called, so `status.updated.info` was not recomputed.

2. **Wrong error priority:** In `updateServerSideDeviceUpdatedStatus`, the `DeviceUpdating` condition (agent-side) was checked first via an `else if` chain, shadowing the `lastRolloutError` annotation (server-side). Any existing `DeviceUpdating` condition — even a stale or non-error one — prevented the rollout error from being shown.

3. **Info-only changes not detected:** `updateServerSideDeviceUpdatedStatus` only checked whether `status.updated.status` changed (e.g. `UpToDate` → `OutOfDate`). When the status stayed `OutOfDate` but the info message changed from generic to specific, the function returned `false` and the update was never persisted.

## Fix

- **`fleet_rollout.go`:** Call `UpdateServerSideDeviceStatus` after writing the `lastRolloutError` annotation so the error is immediately visible in device status.
- **`device.go`:** Check `lastRolloutError` annotation before the `DeviceUpdating` condition so server-side rollout errors take priority over stale agent-side errors.
- **`device.go`:** Track `lastUpdateInfo` alongside `lastUpdateStatus` and include info changes in all return value comparisons so info-only updates are persisted.

## Testing

- **`fleet_rollout_test.go`:** Added `TestFleetRolloutsLogic_updateDeviceToFleetTemplate_TemplateRenderingFailure` — verifies the annotation is written and `UpdateServerSideDeviceStatus` is called after a template rendering failure.
- **`device_test.go`:** Added `TestUpdateServerSideDeviceUpdatedStatus_ManagedDeviceErrorPriority` — verifies error priority with 4 cases: rollout error only, both errors (rollout wins), agent error only, neither (generic fallback).
- Locally verified with `make deploy-quadlets` using a fleet with `{{ .metadata.labels.domainname }}` and a device missing the label. Confirmed `status.updated.info` now shows the actual template rendering error.

## Confidence

High — the fix is minimal and well-scoped. All three changes are in the server-side status computation path and do not affect agent behavior.

## Risk Assessment

Low — changes only affect the display of error messages in `status.updated.info` for devices with rollout failures. No changes to rollout logic, device specs, or agent communication.

### DIFF
--- a/internal/service/common/device.go
+++ b/internal/service/common/device.go
@@ -247,25 +247,7 @@ func updateServerSideDeviceUpdatedStatus(device *domain.Device, ctx context.Cont
 			device.Status.Updated.Info = lo.ToPtr("Device was updated to the fleet's latest device spec.")
 		} else {
 			device.Status.Updated.Status = domain.DeviceUpdatedStatusOutOfDate
-
-			var errorMessage string
-			baseMessage := "Device could not be updated to the fleet's latest device spec"
-			if device.Metadata.Annotations != nil {
-				if lastRolloutError, ok := (*device.Metadata.Annotations)[domain.DeviceAnnotationLastRolloutError]; ok && lastRolloutError != "" {
-					errorMessage = fmt.Sprintf("%s: %s", baseMessage, lastRolloutError)
-				}
-			}
-			if errorMessage == "" {
-				if updateCondition := domain.FindStatusCondition(device.Status.Conditions, domain.ConditionTypeDeviceUpdating); updateCondition != nil {
-					if updateCondition.Reason == string(domain.UpdateStateError) {
-						errorMessage = fmt.Sprintf("%s: %s", baseMessage, updateCondition.Message)
-					}
-				}
-			}
-			if errorMessage == "" {
-				errorMessage = domain.DeviceOutOfSyncWithFleetText
-			}
-			device.Status.Updated.Info = lo.ToPtr(errorMessage)
+			device.Status.Updated.Info = lo.ToPtr(managedDeviceOutOfDateMessage(device))
 		}
 	} else {
 		device.Status.Updated.Status = domain.DeviceUpdatedStatusUpToDate
@@ -281,6 +263,21 @@ func updateServerSideDeviceUpdatedStatus(device *domain.Device, ctx context.Cont
 	}
 
 	return device.Status.Updated.Status != lastUpdateStatus || lo.FromPtr(device.Status.Updated.Info) != lastUpdateInfo
+}
+
+func managedDeviceOutOfDateMessage(device *domain.Device) string {
+	baseMessage := "Device could not be updated to the fleet's latest device spec"
+	if device.Metadata.Annotations != nil {
+		if lastRolloutError, ok := (*device.Metadata.Annotations)[domain.DeviceAnnotationLastRolloutError]; ok && lastRolloutError != "" {
+			return fmt.Sprintf("%s: %s", baseMessage, lastRolloutError)
+		}
+	}
+	if updateCondition := domain.FindStatusCondition(device.Status.Conditions, domain.ConditionTypeDeviceUpdating); updateCondition != nil {
+		if updateCondition.Reason == string(domain.UpdateStateError) {
+			return fmt.Sprintf("%s: %s", baseMessage, updateCondition.Message)
+		}
+	}
+	return domain.DeviceOutOfSyncWithFleetText
 }
 
 func updateServerSideApplicationStatus(device *domain.Device) bool {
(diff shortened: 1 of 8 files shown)


asafbennatan:
@coderabbitai review

internal/service/common/device.go
```diff
@@ -196,10 +196,11 @@ func updateServerSideLifecycleStatus(device *domain.Device) bool {
 
 func updateServerSideDeviceUpdatedStatus(device *domain.Device, ctx context.Context, fleetStore fleetstore.Store, log logrus.FieldLogger, orgId uuid.UUID) bool {
 	lastUpdateStatus := device.Status.Updated.Status
+	lastUpdateInfo := lo.FromPtr(device.Status.Updated.Info)
```
asafbennatan: Worth a closer look: if we return `true` whenever `info` changes, what other call paths recompute status while the enum stays the same?

One example to try: the disconnected branch builds info with a relative time (`humanize.Time`). Does that string change across recomputes even when status stays `Unknown`? What would that mean for write frequency?

I’m not saying “revert” yet — more: can we confirm the original status-only gate was intentional under a “status and info always move together” model, and whether this bug is a case that model doesn’t cover anymore?

internal/tasks/fleet_rollout.go
```diff
@@ -333,6 +333,9 @@ func (f FleetRolloutsLogic) updateDeviceToFleetTemplate(ctx context.Context, dev
 		if status.Code != http.StatusOK {
 			errs = append(errs, common.ApiStatusToErr(status))
 		}
+		if err := f.deviceSvc.UpdateServerSideDeviceStatus(ctx, f.orgId, *device.Metadata.Name); err != nil {
```
asafbennatan: Good instinct to recompute after writing the annotation.

A few things to verify locally:

1. If the device is **already** `OutOfDate` and only `info` would change, does `UpdateServerSideDeviceStatus` actually write to the DB with today’s (pre–info-diff) logic?
2. If the annotation update fails, what does the following status recompute see?
3. On the success path nearby, how do we clear `lastRolloutError` — does that suggest anything about when status should be updated vs when annotation-only is enough?

If (1) is “no write,” what’s another way to get the error into `status.updated.info` on **this** path only, without changing the global changed-detection?

internal/service/common/device.go
```diff
@@ -276,7 +262,22 @@ func updateServerSideDeviceUpdatedStatus(device *domain.Device, ctx context.Cont
...
+
+func managedDeviceOutOfDateMessage(device *domain.Device) string {
+	baseMessage := "Device could not be updated to the fleet's latest device spec"
+	if device.Metadata.Annotations != nil {
+		if lastRolloutError, ok := (*device.Metadata.Annotations)[domain.DeviceAnnotationLastRolloutError]; ok && lastRolloutError != "" {
+			return fmt.Sprintf("%s: %s", baseMessage, lastRolloutError)
+		}
+	}
+	if updateCondition := domain.FindStatusCondition(device.Status.Conditions, domain.ConditionTypeDeviceUpdating); updateCondition != nil {
+		if updateCondition.Reason == string(domain.UpdateStateError) {
```
asafbennatan: Small consistency check: compare this helper’s `DeviceUpdating` handling with the unmanaged `OutOfDate` branch a few lines above. Do they treat an empty `Message` the same way? If not, which behavior do we want?

internal/service/common/device_test.go
```diff
@@ -469,3 +479,99 @@ func TestUpdateServerSideDeviceUpdatedStatus_OsImageMismatch(t *testing.T) {
 		})
 	}
 }
+
+func TestUpdateServerSideDeviceUpdatedStatus_ManagedDeviceErrorPriority(t *testing.T) {
```
asafbennatan, line 483: The priority table is a solid start.

For the persistence behavior: what would a regression test look like that fails if the error never makes it into **stored** status (not only into the in-memory struct or a mock “was called”)? That might clarify whether the right contract is “recompute returned true” vs “this failure path updated status.”

asafbennatan (changes requested):
Nice work on diagnosing this — splitting it into annotation write, message priority, and status persistence is a clear write-up.

One design question before we merge the persistence change: historically `updateServerSideDeviceUpdatedStatus` only returns “changed” when the **status enum** changes, even though `info` has always been in the API. It may be worth checking *why* that was written that way (look at the original EDM-417-era code and how other `updateServerSide*` helpers decide what to persist) before we treat info diffs like status diffs.

Could you dig into whether the EDM-3677 failure case can be fixed **without** widening that shared “changed” check? Happy to discuss what you find.

The priority change (`lastRolloutError` before the agent condition) looks directionally right — please keep that investigation separate from the persistence question.

asafbennatan (approved):
LGTM

PR #3254, opened 2026-07-20
EDM-4180: Fix AAP role-assignment lookups to use ansible_id (Controller API)

asafbennatan:
## Summary
- **Update:** the first commit in this PR switched role-assignment lookups from the Controller API to the Gateway API per initial AAP team guidance. Live testing against a real AAP 2.5 environment showed this breaks custom-role-based RBAC entirely: custom role definitions (e.g. `flightctl-org-admin`) only exist in the Controller's role registry and are invisible through the Gateway's `role_user_assignments`/`role_team_assignments` endpoints (confirmed live: Gateway only returns AAP's 5 built-in managed roles; Gateway's `role_definitions` endpoint doesn't even support creating custom roles).
- The second commit reverts those two lookups back to the Controller API, and separately fixes a real (if latent) bug along the way: the code was assuming the Gateway numeric user/team ID is the same as the Controller's numeric ID for the same resource. That's not guaranteed. Lookups now filter by `ansible_id` (`user__resource__ansible_id` / `team__resource__ansible_id`), which is meant to stay stable across AAP components.
- Confirmed live against a real AAP instance: filtering by `ansible_id` returns the same (correct) set of role assignments as the old numeric-ID filter, including the custom `flightctl-org-admin` assignment that the Gateway-only endpoint was silently dropping.

## Test plan
- [x] `go build ./pkg/aap/... ./internal/auth/...`
- [x] `go test ./pkg/aap/... ./internal/auth/...`
- [x] Verified live against a real AAP 2.5 instance: `GET /api/controller/v2/role_user_assignments/?user__resource__ansible_id=<id>` returns both the custom `flightctl-org-admin` assignment and the built-in `Organization Member` assignment for a test user, matching what the numeric-ID filter returned
- [ ] Team-role-assignment path (`team__resource__ansible_id`) implemented symmetrically but not live-tested — the test AAP environment has no teams configured

### DIFF
--- a/internal/auth/authn/aapgateway.go
+++ b/internal/auth/authn/aapgateway.go
@@ -28,6 +28,7 @@ type AAPIdentity struct {
 	common.BaseIdentity
 	superUser       bool
 	platformAuditor bool
+	ansibleID       string
 }
 
 // Ensure AAPIdentity implements AAPGatewayUserIdentity
@@ -131,6 +132,7 @@ func (a *AapGatewayAuth) loadUserInfo(ctx context.Context, token string) (*AAPId
 		BaseIdentity:    *common.NewBaseIdentityWithIssuer(aapUserInfo.Username, strconv.Itoa(aapUserInfo.ID), []common.ReportedOrganization{}, identity.NewIssuer(identity.AuthTypeAAP, a.spec.ApiUrl)),
 		superUser:       aapUserInfo.IsSuperuser,
 		platformAuditor: aapUserInfo.IsPlatformAuditor,
+		ansibleID:       aapUserInfo.SummaryFields.Resource.AnsibleID,
 	}
 
 	return userInfo, nil
@@ -205,7 +207,7 @@ func (a *AapGatewayAuth) GetIdentity(ctx context.Context, token string) (common.
 		orgRoles["*"] = globalRoles
 	}
 
-	roleAssignments, err := a.getUserRoleAssignments(ctx, token, userInfo.GetUID())
+	roleAssignments, err := a.getUserRoleAssignments(ctx, token, userInfo.ansibleID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get role assignments: %w", err)
 	}
@@ -310,8 +312,8 @@ func (a *AapGatewayAuth) getUserScopedOrganizations(ctx context.Context, token s
 }
 
 // getUserRoleAssignments gets role assignments for a specific user from AAP
-func (a *AapGatewayAuth) getUserRoleAssignments(ctx context.Context, token string, userID string) ([]*aap.AAPRoleUserAssignment, error) {
-	roleAssignments, err := a.aapClient.ListUserRoleAssignments(ctx, token, userID)
+func (a *AapGatewayAuth) getUserRoleAssignments(ctx context.Context, token string, ansibleID string) ([]*aap.AAPRoleUserAssignment, error) {
+	roleAssignments, err := a.aapClient.ListUserRoleAssignments(ctx, token, ansibleID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


aekubacki:
Tested this against a real AAP 2.5 environment (containerized install, build `2.5-25`) to check whether the underlying issue was present that far back — it is, and this fix directly resolves it.

**Setup**: fresh AAP 2.5 all-in-one, created a throwaway org ("Test Org") and a plain non-superuser user (`demo`) with both "Organization Member" and "Organization Admin" roles assigned. Built `flightctl-api` from this PR cherry-picked onto `release-1.2` and deployed via `make deploy-quadlets`.

**Direct API comparison** (same user, same role assignments, both queried immediately after each other):

- `GET /api/controller/v2/role_user_assignments/?user__id=3` (current/unpatched code path) → `200 OK`, **`count: 0`** — returns nothing at all, despite `demo` genuinely holding both role assignments.
- `GET /api/gateway/v1/role_user_assignments/?user__id=3` (this PR's target) → `200 OK` with both role assignments returned correctly (`Organization Member` and `Organization Admin` on `Test Org`).

So on AAP 2.5, the current code doesn't just handle this suboptimally — it silently gets **zero data back** from the Controller endpoint for Gateway-created role assignments. This PR's switch to the Gateway-native endpoint fixes that cleanly.

**End-to-end**: logged into RHEM as `demo` with the patched build — still landed on "Restricted Access" (screenshot attached). That's expected and unrelated to this PR: it's the separate, already-tracked role-name-mapping bug (`mapRoleAssignmentsToOrgRoles` never translates AAP's literal role definition names like `"Organization Admin"` into Flight Control's internal permission constants) — a different bug this PR doesn't touch. Both fixes are needed together before org-scoped RBAC actually works end-to-end, but this PR's specific fix is confirmed necessary and correct on AAP 2.5.

Nice fix — LGTM pending the still-unchecked smoke-test item in the PR description, which this comment should cover.

aekubacki:
Follow-up finding from further testing on the same AAP 2.5 environment — this one's independent of the earlier verification and worth flagging before merge.

**Setup**: created a custom AAP role named literally `flightctl-org-admin` (Content Type: Organization, `managed: false`), assigned to the same test user (`demo`, id 3) on `Test Org`, alongside the existing managed "Organization Member"/"Organization Admin" roles from the earlier test.

**Finding**: managed and custom roles are visible through **different, mutually exclusive** API endpoints on this AAP 2.5 environment:

- `GET /api/gateway/v1/role_user_assignments/?user__id=3` → returns the two **managed** roles ("Organization Member", "Organization Admin"), `count: 2`. The custom `flightctl-org-admin` role does **not** appear here at all.
- `GET /api/controller/v2/role_user_assignments/?user__id=3` → returns **only** the custom role (`flightctl-org-admin`, `managed: false`), `count: 1`. Neither managed role appears here (consistent with the earlier finding — this endpoint returned `count: 0` before the custom role existed).

So this isn't one endpoint being a superset of the other — it's a real split: managed roles live on the Gateway side, custom/local roles live on the Controller side, and each endpoint is blind to the other's roles.

**Why this matters for this PR**: switching entirely to the Gateway endpoint fixes visibility for managed roles (confirmed in the earlier comment), but it trades that for a new blind spot — any customer using **custom roles** (which are common for org-scoped RBAC customization) would go from "visible via the current Controller-based code" to "invisible via this PR's Gateway-based code." Today's code has the opposite gap (blind to managed roles, sees custom roles); this PR flips which gap exists rather than closing it.

Given that, it looks like the correct fix needs to query **both** endpoints and merge the results (deduping as needed), rather than switching from one to the other. Happy to help verify a merged approach against this same environment if useful.

aekubacki (approved):
Tested the revised commit (`52ab18f2`, ansible_id matching) against both AAP 2.6 and AAP 2.7 — confirmed working on both, full org-scoped RBAC restored.

**Note on the commit history**: `52ab18f2` is a standalone commit, not built on top of the earlier `76b917c5` (`git merge-base --is-ancestor 76b917c5 <new-head>` returns false) — cherry-pick only the new commit if reproducing this, not both.

**Setup**: cherry-picked `52ab18f2` onto `release-1.2`, built `flightctl-api`, and swapped it into the existing RPM-installed RHEM instances already paired with our AAP 2.6/2.7 labs (image swap via `podman save`/`load`/`tag` against the quadlet's `Pull=missing` behavior — no need to touch or rebuild the rest of the stack). Then created a custom AAP role named `flightctl-org-admin` (Organization resource type, single `Can view organization` permission — deliberately minimal, matching Flight Control's expected external role string exactly) and assigned it to our existing org-scoped test users (`test-admin`/`test-org-27` on AAP 2.7, `orga-admin`/`org-a` on AAP 2.6) alongside their existing "Organization Admin" role.

**Result**: full working RHEM access on both environments — Overview page loads completely normally, no "Restricted Access." This confirms:
- The ansible_id + Controller-only approach in this revision works correctly on both AAP 2.6 and 2.7, not just AAP 2.5
- The custom-role-naming path (naming a custom AAP role to exactly match Flight Control's `ExternalRole*` constants) is a real, functional mechanism once role assignments are correctly retrieved — not just a theoretical possibility from the earlier bifurcation finding

This closes the loop on both things we'd flagged: the AAP 2.5 data-retrieval gap (confirmed fixed) and the managed/custom role endpoint bifurcation (confirmed fixed by staying on Controller with the more robust ansible_id lookup, rather than switching to Gateway).

PR #3255, opened 2026-07-20
NO-ISSUE: Sync air-gap embed after flightctl-greenboot Requires agent

galel12:
## Summary

- Regenerates `scripts/air-gap/mirror-images/embedded_data_generated.go` via `make generate-mirror-embed`
- Adds `flightctl-agent` to each variant's RPM list

## Why

`generate-embed` builds the air-gap RPM list from bare `Requires:` lines in `packaging/rpm/flightctl.spec`. After the `flightctl-greenboot` subpackage split, the spec includes `Requires: flightctl-agent`, but the embedded manifest was not regenerated. This syncs generated output with the current spec.

Spotted while working on EDM-4755; kept out of that PR as unrelated scope.

## Test plan

- [x] Confirmed diff is only the four `flightctl-agent` RPM entries
- [ ] CI passes

### DIFF
--- a/scripts/air-gap/mirror-images/embedded_data_generated.go
+++ b/scripts/air-gap/mirror-images/embedded_data_generated.go
@@ -109,6 +109,7 @@ var embeddedBuildManifest = []byte(`
       "rpms": [
         "bash",
         "container-selinux",
+        "flightctl-agent",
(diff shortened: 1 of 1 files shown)


asafbennatan (approved):
LGTM

PR #3256, opened 2026-07-20
EDM-4688: Separate temporary and permanent spec rendering errors

EfratIfergan:
## Problem

After PR #3142, devices with permanent render errors (bad git path, forbidden config path, unknown config type) get re-rendered indefinitely. The fleet rollout skip condition at `fleet_rollout.go:356` requires `renderedTemplateVersion` to match, but this annotation is only written on successful render via `UpdateRenderedDevice`. Failed renders leave it empty, so the skip condition never fires and the device is re-processed every rollout cycle.

## Root Cause

`renderedSpecHash` and `renderedTemplateVersion` annotations are only written inside `UpdateRenderedDevice` (`store/device/store.go`), which only runs on **successful** render. When render fails permanently, these annotations stay empty/stale, causing:

1. Fleet rollout skip condition fails (`"" != "v3"`) → device re-rolled-out
2. `FleetRolloutDeviceSelected` events bypass the spec-hash guard → forced re-render
3. Same permanent error repeats indefinitely

## Fix

Classify render errors as **permanent vs retryable** using an allowlist-based classifier. On permanent failure, write `renderedSpecHash` and `renderedTemplateVersion` annotations via `UpdateDeviceAnnotations` to stop re-processing.

- **`isPermanentRenderError`** — defaults to **retryable** (non-permanent), only whitelists known immutable validation errors as permanent:
  - `ErrUnknownConfigName` — unsupported or unparseable config type
  - `ErrUnknownApplicationType` — unsupported or unparseable app type / malformed app shape
  - `ErrForbiddenDevicePath` — forbidden device paths (all `DenyForbiddenDevicePath` rejections: denied roots, agent config files, relative paths, colon-containing paths)
- **External/infrastructure errors remain retryable** — DNS, connection reset, EOF, context cancellation, repo not found, K8s API unavailable, HTTP 5xx, secret NotFound. These do NOT write annotations, so repository-update recovery and retry continue working.
- **`markPermanentRenderFailure`** — writes hash/version annotations on permanent failure via `UpdateDeviceAnnotations` (no agent notification, no event emission).
- **Both error paths covered** — config render error and app render error in `RenderDevice`.
- **App-shape sentinel wrapping** — `renderApplication` now wraps `GetAppType`, `As*Application` parse, and `GetName` failures with `ErrUnknownApplicationType` (matching the config-side pattern). VM infra/conversion errors are intentionally not wrapped.
- **Path validation sentinel wrapping** — `DenyForbiddenDevicePath` early rejects (empty, relative, colon-containing paths) now wrap `ErrForbiddenDevicePath`, matching the denied-root and agent-config branches.
- **Recovery**: when user fixes fleet template → new template version → new hash → hash guard passes → render succeeds.

## Testing

- 23 unit tests for `isPermanentRenderError` covering all error types including relative-path rejection
- 5 `RenderDevice` tests for permanent/retryable behavior:
  - Fleet-owned device + permanent config error → both annotations written
  - Standalone device + permanent config error → only specHash written
  - Retryable error (K8s EOF) → no annotations written
  - External error (K8s unavailable, fleet-owned) → no annotations written
  - Permanent app error (empty app spec) → annotations written, `ErrorIs(ErrUnknownApplicationType)`
- `DenyForbiddenDevicePath` tests verify all rejections wrap `ErrForbiddenDevicePath`
- Regression validated: reverting fix causes permanent error tests to fail

## Confidence

High (95%) — minimal change, well-tested, no side effects on existing code paths.

## Risk Assessment

Low — only adds annotation writes on a previously-unhandled failure path. No changes to fleet rollout, store layer, or agent notification.

## Rollback

Revert the commit. Devices with permanent errors will resume the retry loop (pre-existing behavior).

### DIFF
--- a/internal/tasks/device_render.go
+++ b/internal/tasks/device_render.go
@@ -407,7 +407,7 @@ func (t *DeviceRenderLogic) renderConfigItem(ctx context.Context, configItem *do
 func renderApplication(ctx context.Context, app *domain.ApplicationProviderSpec, vmConverter VmConverterFn, kvStore kvstore.KVStore) (*string, *domain.ApplicationProviderSpec, error) {
 	appType, err := (*app).GetAppType()
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get app type: %w", err)
+		return nil, nil, fmt.Errorf("%w: failed to get app type: %w", ErrUnknownApplicationType, err)
 	}
 
 	switch appType {
@@ -422,7 +422,7 @@ func renderApplication(ctx context.Context, app *domain.ApplicationProviderSpec,
 	case domain.AppTypeVm:
 		vmApp, parseErr := (*app).AsVmApplication()
 		if parseErr != nil {
-			return nil, nil, fmt.Errorf("failed to parse application spec for type %s: %w", appType, parseErr)
+			return nil, nil, fmt.Errorf("%w: failed to parse application spec for type %s: %w", ErrUnknownApplicationType, appType, parseErr)
 		}
 		appName := vmApp.Name
 		converted, renderErr := renderVmApplication(ctx, vmApp, vmConverter, kvStore)
@@ -434,11 +434,11 @@ func renderApplication(ctx context.Context, app *domain.ApplicationProviderSpec,
 		return nil, nil, fmt.Errorf("%w: unsupported application type: %q", ErrUnknownApplicationType, appType)
 	}
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to parse application spec for type %s: %w", appType, err)
+		return nil, nil, fmt.Errorf("%w: failed to parse application spec for type %s: %w", ErrUnknownApplicationType, appType, err)
 	}
 	appName, err := (*app).GetName()
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get app name: %w", err)
+		return nil, nil, fmt.Errorf("%w: failed to get app name: %w", ErrUnknownApplicationType, err)
 	}
 	return appName, app, nil
 }
--- a/internal/tasks/device_render.go
+++ b/internal/tasks/device_render.go
@@ -209,11 +209,17 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 	}
 
 	if renderErr != nil {
+		if !isRetryableRenderError(renderErr) {
+			t.markPermanentRenderFailure(ctx, specHash)
+		}
 		return t.setStatus(ctx, renderErr)
 	}
 
 	renderedApplications, err := t.renderApplications(ctx)
 	if err != nil {
+		if !isRetryableRenderError(err) {
+			t.markPermanentRenderFailure(ctx, specHash)
+		}
 		return t.setStatus(ctx, err)
 	}
 
@@ -230,6 +236,20 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 	// even though specHash is unchanged (see bypassHashCheck above).
 	status = t.deviceSvc.UpdateRenderedDevice(ctx, t.orgId, t.event.InvolvedObject.Name, string(renderedConfig), string(renderedApplications), specHash, syncRefs, bypassHashCheck)
 	return t.setStatus(ctx, common.ApiStatusToErr(status))
+}
+
+func (t *DeviceRenderLogic) markPermanentRenderFailure(ctx context.Context, specHash string) {
+	annotations := map[string]string{
+		domain.DeviceAnnotationRenderedSpecHash: specHash,
+	}
+	if t.templateVersion != nil {
+		annotations[domain.DeviceAnnotationRenderedTemplateVersion] = *t.templateVersion
+	}
+
+	status := t.deviceSvc.UpdateDeviceAnnotations(ctx, t.orgId, t.event.InvolvedObject.Name, annotations, nil)
+	if status.Code != http.StatusOK {
+		t.log.Errorf("Failed writing render-failure annotations for device %s/%s: %s", t.orgId, t.event.InvolvedObject.Name, status.Message)
+	}
 }
 
 func (t *DeviceRenderLogic) setStatus(ctx context.Context, renderErr error) error {
--- a/internal/tasks/device_render.go
+++ b/internal/tasks/device_render.go
@@ -209,7 +209,7 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 	}
 
 	if renderErr != nil {
-		if !isRetryableRenderError(renderErr) {
+		if isPermanentRenderError(renderErr) {
 			t.markPermanentRenderFailure(ctx, specHash)
 		}
 		return t.setStatus(ctx, renderErr)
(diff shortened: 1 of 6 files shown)


internal/tasks/render_errors.go
```diff
@@ -0,0 +1,54 @@
...
+		"i/o timeout",
+		"unexpected EOF",
+	} {
+		if strings.Contains(msg, substr) {
+			return true
+		}
+	}
+
+	return false
+}
```
asafbennatan, line 24: **Critical / High — invert the classifier (explicit permanent errors)**

Defaulting to permanent and whitelisting only a few network/context errors will:

1. **Break standalone repo-update recovery.** `markPermanentRenderFailure` writes `renderedSpecHash`, but `EventReasonReferencedRepositoryUpdated` is *not* in `bypassHashCheck`. After a "permanent" failure (including repo-not-found, which this PR's tests treat as non-retryable), a later repo fix re-triggers render with the same spec hash, the guard early-returns, and the device stays `SpecValid=False`. That contradicts the existing comment in `RenderDevice` that repo refs are kept specifically so a later repository update can re-render and maybe fix the device.

2. **Freeze previously self-healing external failures** that never match the whitelist, e.g.:
   - HTTP `unexpected status code %d` (including 5xx)
   - `failed fetching specified Repository definition ...`
   - K8s secret NotFound / non-EOF client errors
   - `kubernetes API is not available`

**Suggestion:** invert the policy for control-plane render — default to **retryable**, and only treat known immutable validation errors as permanent:
- `ErrUnknownConfigName` / unsupported config type
- `ErrUnknownApplicationType` / invalid app shape
- forbidden / invalid device paths (git, secret-derived, inline)
- other pure-spec validation failures that cannot change without a new spec/TV

Those are the EDM-4688 cases. For those, writing the hash/version annotations is correct. Prefer this over adding `ReferencedRepositoryUpdated` to `bypassHashCheck` alone (that only papers over the symptom).

internal/tasks/device_render_test.go
```diff
@@ -691,6 +693,183 @@ func TestRenderApplication_PreservesLifecycleFields(t *testing.T) {
...
+
+	mockSvc.EXPECT().SetDeviceServiceConditions(gomock.Any(), orgId, deviceName, gomock.Any()).Return(statusOK)
+	mockSvc.EXPECT().UpdateServerSideDeviceStatus(gomock.Any(), orgId, deviceName).Return(nil)
+
+	event := createTestEvent(domain.DeviceKind, domain.EventReasonResourceUpdated, deviceName)
+	logic := NewDeviceRenderLogic(logrus.New(), mockSvc, nil, nil, newTestKVStore(), &config.Config{}, orgId, event)
+
+	err := logic.RenderDevice(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid application")
```
asafbennatan: **Medium — add regression coverage after the classifier invert**

Current tests cover annotation write/skip for permanent vs retryable paths well. Please also add:

1. After a permanent validation failure (annotations written), an event with `EventReasonReferencedRepositoryUpdated` must still re-enter render if the failure class is external — or, with the inverted classifier, assert that external failures do **not** write `renderedSpecHash` so repo-update recovery keeps working.
2. Classifier cases for HTTP `unexpected status code` and secret NotFound once the permanent allowlist is explicit (they should remain retryable).

Optional: `UpdateDeviceAnnotations` non-OK in `markPermanentRenderFailure` (fail-open / log-only) — lower priority.

asafbennatan (changes requested):
see below

internal/tasks/device_render.go
```diff
@@ -209,11 +209,17 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
...
 	if renderErr != nil {
+		if isPermanentRenderError(renderErr) {
+			t.markPermanentRenderFailure(ctx, specHash)
+		}
 		return t.setStatus(ctx, renderErr)
 	}
 
 	renderedApplications, err := t.renderApplications(ctx)
 	if err != nil {
+		if isPermanentRenderError(err) {
```
asafbennatan, line 220: **Medium — wrap app-shape failures with `ErrUnknownApplicationType`**

This path marks permanent only when the error wraps a sentinel. Empty/malformed apps fail in `renderApplication` with `failed to get app type` / `failed to parse application spec` **without** wrapping `ErrUnknownApplicationType`, so they stay retryable. Empty configs already wrap `ErrUnknownConfigName` in `renderConfigItem` — apps should match that pattern, e.g.:

```go
return nil, nil, fmt.Errorf("%w: failed to get app type: %w", ErrUnknownApplicationType, err)
```

Same for the parse-failure returns in `renderApplication`.

Note: for fleets this is mostly defense-in-depth today — API `validateApplications` rejects typeless apps on apply, and fleet rollout `getDeviceApps` fails before updating the device. Still worth returning the proper sentinel so any path that does reach render stops reprocessing (and so the app test below can expect annotations again). Keep VM/infra failures non-permanent.

internal/tasks/device_render_test.go
```diff
@@ -691,6 +693,221 @@ func TestRenderApplication_PreservesLifecycleFields(t *testing.T) {
...
+
+	event := createTestEvent(domain.DeviceKind, domain.EventReasonFleetRolloutDeviceSelected, deviceName)
+	logic := NewDeviceRenderLogic(logrus.New(), mockSvc, nil, nil, newTestKVStore(), &config.Config{}, orgId, event)
+
+	err := logic.RenderDevice(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "kubernetes API is not available")
+}
+
+func TestRenderDevice_NonPermanentAppError(t *testing.T) {
```
asafbennatan: **Related to the app-shape comment** — this test currently locks in non-permanent behavior for an empty `ApplicationProviderSpec`. After wrapping `GetAppType`/parse failures with `ErrUnknownApplicationType`, restore a permanent-app `RenderDevice` test that expects `UpdateDeviceAnnotations` (mirror the config permanent cases).

internal/tasks/render_errors.go
```diff
@@ -0,0 +1,24 @@
...
+		return false
+	}
+
+	switch {
+	case errors.Is(err, ErrUnknownConfigName):
+		return true
+	case errors.Is(err, ErrUnknownApplicationType):
+		return true
+	case errors.Is(err, validation.ErrForbiddenDevicePath):
+		return true
```
asafbennatan, line 20: **Medium — `DenyForbiddenDevicePath` early rejects are not covered by this sentinel**

This correctly catches forbidden-root / agent-config denials (those already `%w`-wrap `ErrForbiddenDevicePath`, including via git render).

But in `internal/util/validation/validation.go`, the early branches for empty, relative, or `:`-containing paths return plain `fmt.Errorf("invalid device path …")` **without** the sentinel. Those are immutable when they come from git content (fleet validate does not clone/scan git paths — only device render does), so they will keep retrying under this classifier.

Prefer fixing at the source — wrap those rejection branches with `ErrForbiddenDevicePath` (same as the denied-root branches), rather than special-casing strings here. Add a classifier/`RenderDevice` case for a relative path if practical.

internal/tasks/render_errors.go
```diff
@@ -0,0 +1,24 @@
...
+	}
+
+	return false
+}
```
asafbennatan, line 24: **Low — PR description is stale**

The Fix/Testing sections still describe `isRetryableRenderError`, “defaults to permanent”, and “Permanent app error → annotations written”. Please update the PR body to match this allowlist + default-retryable design and the current test matrix (including external errors not writing annotations).

asafbennatan (changes requested):
see below

asafbennatan (approved):
LGTM , thank you for addressing the comments!

PR #3260, opened 2026-07-20
EDM-4829: Catalog non-empty delete (service rule + TX)

asafbennatan:
Jira: EDM-4829

Moves the catalog non-empty delete product rule to the service call site and locks the contract with unit + integration tests.

## Summary

- Documents that `DeleteCatalog` chooses the store delete path that returns `ErrResourceNotEmpty` when items still exist, and maps that error to HTTP 409
- Clarifies the store `Delete` TX as a persistence contract (count+delete), not a separate policy API
- Adds service-layer unit coverage: non-empty delete → conflict / catalog retained; empty delete → success
- Tightens the existing integration case to assert the `ErrResourceNotEmpty` message

Behavior is unchanged: deleting a catalog that still has items continues to return conflict.

- [x] **`release-1.3`**

---
**Stack:** 3/12 in the EDM-4806 (Technical Debt: Store Contains Business Logic) closeout stack.
Base: `main` (after #3259)

### DIFF
--- a/internal/service/catalog/handler.go
+++ b/internal/service/catalog/handler.go
@@ -162,9 +162,10 @@ func (h *ServiceHandler) DeleteCatalog(ctx context.Context, orgId uuid.UUID, nam
 		return nil
 	}
 
+	// Product rule: refuse deleting a non-empty catalog. The service chooses store.Delete
+	// (TX primitive that returns ErrResourceNotEmpty when items exist) and maps the error.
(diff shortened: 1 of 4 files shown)


asafbennatan:
Closing as docs/tests-only closeout with no production behavior change. Folding EDM-4829 into the next stack PR (#3261 / EDM-4830), which already carries these commits.

PR #3261, opened 2026-07-20
EDM-4830: Device decommission decisions in service

asafbennatan:
Jira: EDM-4830

Moves device decommission decisions into the service layer, using the store-native `Mutate` CAS RMW introduced on `main` (EDM-5028).

## Summary

- Service `DecommissionDevice` applies Spec/Status/Owner/Labels changes inside a `Mutate` callback against the CAS-fresh device
- Refuses a second decommission with `ErrResourceVersionConflict` and does not emit a success event
- Removes store-owned `DecommissionDevice` (precondition + mutation + event callback)
- Unit + integration coverage for success, already-decommissioning conflict, and not-found

Already covered on `main` (not re-done here): Replace/Patch/Update reject already-decommissioning devices inside their `Mutate` callbacks.

- [x] **`release-1.3`**

---
**Stack:** bottom of the EDM-4806 store business-logic closeout stack.  
Base: `main` (includes Mutate CAS).

### DIFF
--- a/internal/service/device/handler.go
+++ b/internal/service/device/handler.go
@@ -215,9 +215,10 @@ func (h *DeviceServiceHandler) GetDevice(ctx context.Context, orgId uuid.UUID, n
 	return result, common.StoreErrorToApiStatus(err, false, domain.DeviceKind, &name)
 }
 
-// DeviceVerificationCallback ensures the device wasn't decommissioned before an update proceeds.
-func DeviceVerificationCallback(ctx context.Context, before, after *domain.Device) error {
-	if before != nil && before.Spec != nil && before.Spec.Decommissioning != nil {
+// rejectIfAlreadyDecommissioning returns flterrors.ErrDecommission when the existing
+// device already has Spec.Decommissioning set.
+func rejectIfAlreadyDecommissioning(existing *domain.Device) error {
+	if existing != nil && existing.Spec != nil && existing.Spec.Decommissioning != nil {
 		return flterrors.ErrDecommission
 	}
 	return nil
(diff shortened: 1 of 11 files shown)


internal/service/device/handler.go
```diff
@@ -749,8 +749,42 @@ func snapshotDeviceForStatusUpdate(device *domain.Device) *domain.Device {
...
 
 func (h *DeviceServiceHandler) DecommissionDevice(ctx context.Context, orgId uuid.UUID, name string, decom domain.DeviceDecommission) (*domain.Device, domain.Status) {
-	result, err := h.deviceStore.DecommissionDevice(ctx, orgId, name, decom, h.callbackDeviceDecommission)
-	return result, common.StoreErrorToApiStatus(err, false, domain.DeviceKind, &name)
+	result, before, _, err := h.deviceStore.Mutate(ctx, orgId, name, nil, func(m *devicestore.DeviceMutation) error {
+		if err := m.RequireExisting(); err != nil {
+			return err
+		}
+		// Product rule: refuse a second decommission without emitting a success event.
+		if m.Device.Spec != nil && m.Device.Spec.Decommissioning != nil {
```
galel12: The "already decommissioning" check duplicates `rejectDecommissionedDevice` used elsewhere in this package. Should we call that helper here and map its error to `ErrResourceVersionConflict`, so the guard lives in one place if the rule ever expands beyond `Spec.Decommissioning`?

```go
if err := rejectDecommissionedDevice(m.Device); err != nil {
    return flterrors.ErrResourceVersionConflict
}
```

internal/service/device/handler.go
```diff
@@ -749,8 +749,42 @@ func snapshotDeviceForStatusUpdate(device *domain.Device) *domain.Device {
...
+			return flterrors.ErrResourceVersionConflict
+		}
+		applyDeviceDecommission(m.Device, decom)
```
galel12, line 761: The old store `DecommissionDevice` path didn't bump `generation`; this now goes through `Mutate`/`Update`, which will increment it when spec changes. Is that intentional?

internal/service/device/handler.go
asafbennatan: Good call — switched to `rejectDecommissionedDevice` and map to `ErrResourceVersionConflict` so the API stays the same while the guard lives in one place.

internal/service/device/handler.go
asafbennatan: Not intentional as a behavior change. The old store path updated spec/status/owner/labels without bumping `generation`; Mutate’s `Update` would have incremented it on the Spec delta.

Pinned via `DeviceMutation.PreserveGeneration` so decommission keeps the previous generation contract (covered by unit + integration asserts).

asafbennatan:
Closing to recreate as a fork-based stacked PR from asafbennatan/flightctl → flightctl/flightctl for the gh-stack demo workflow.

PR #3279, opened 2026-07-22
EDM-4846: Distinguish error messages for missing vs excess args in edit command

EfratIfergan:
## Problem

`flightctl edit` returns the same error message for both missing arguments and too many arguments:

```
Error: you must specify a resource to edit (TYPE NAME or TYPE/NAME)
```

This is misleading in the too-many-arguments case because the user *did* specify resources.

## Root Cause

The `Args` validator in `internal/cli/edit.go` combined both conditions into a single check with one error message:

```go
if len(args) < 1 || len(args) > 2 {
    return fmt.Errorf("you must specify a resource to edit (TYPE NAME or TYPE/NAME)")
}
```

## Fix

Split the single condition into two separate checks with distinct messages:

- `flightctl edit` (no args) → `"you must specify a resource to edit (TYPE NAME or TYPE/NAME)"`
- `flightctl edit 1 2 3 4` (too many args) → `"too many arguments: edit accepts at most 2 arguments (TYPE NAME or TYPE/NAME)"`

Updated the test in `TestEditCommand_CobraValidation` to assert on error message content for both cases.

## Testing

- Regression test verified bidirectionally (fails without fix, passes with fix)
- All CLI package tests pass
- Full unit test suite passes (1 pre-existing unrelated failure in `internal/backup`)
- Lint passes

## Confidence

High (95%) — minimal change to error message text only, no behavioral changes.

## Risk Assessment

Low — only error message strings changed. No API, behavioral, or security impact.

Fixes: https://redhat.atlassian.net/browse/EDM-4846


eldar101:
LGTM, ping me for approval once done and out of draft.

asafbennatan (changes requested):
## Code review finding

**Severity:** HIGH — Test coverage

The Args split and unit-test updates look correct, but this PR does not update the e2e assertion that pins the old Cobra too-many-args message.

In `test/e2e/cli/cli_test.go` (around lines 994–997):

```go
By("failing when too many arguments are provided")
out, err = harness.CLI("edit", "1", "2", "3", "4")
Expect(out).To(ContainSubstring("Error: accepts between 1 and 2 arg(s), received 4"))
```

After this change, `flightctl edit 1 2 3 4` returns:

```text
Error: too many arguments: edit accepts at most 2 arguments (TYPE NAME or TYPE/NAME)
```

That e2e expectation matches neither the new message nor the previous combined custom Args message. Closed PR #3245 already showed this assertion needs updating when edit Args messaging changes.

### Suggested fix

Update the too-many-args e2e expectation in the same PR, for example:

```go
Expect(out).To(ContainSubstring("Error: too many arguments: edit accepts at most 2 arguments (TYPE NAME or TYPE/NAME)"))
```

Keep the no-args assertion as-is so the two cases stay distinct. Unit coverage in `TestEditCommand_CobraValidation` is helpful but not sufficient while e2e still pins the obsolete string.

asafbennatan (approved):
LGTM

eldar101 (approved):

PR #3281, opened 2026-07-22
EDM-4858: Add version mismatch diagnostics

eldar101:
## Summary

Add failure-time diagnostics to the rpm-sanity version check for label `79621`.

The test still validates that the enrolled agent version matches the deployed server version. When that comparison fails, it now logs the agent version details, deployed `flightctl-api` image, relevant env inputs, namespaces, and runtime environment so we can tell whether the mismatch came from the agent RPM, server deployment image, or build metadata.

## Release target

- [x] release-1.3

Target release: release-1.3

### DIFF
--- a/test/e2e/authprovider/authprovider_helpers_test.go
+++ b/test/e2e/authprovider/authprovider_helpers_test.go
@@ -503,6 +503,11 @@ func (f fakeInfraProvider) GetServiceEndpoint(infra.ServiceName) (string, int, e
 	return "", 0, errors.New("not implemented")
 }
 
+// GetServiceImage is unused by these tests and satisfies infra.InfraProvider.
+func (f fakeInfraProvider) GetServiceImage(infra.ServiceName) (string, error) {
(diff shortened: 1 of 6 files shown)


eldar101:
The 79621  (Agent version check) failure isn’t caused by the test logic.
CI deploys the flightctl backend from the latest Quay chart tag, but in custom runs we point ocp-flightctl-gotests at a feature branch. That means the server comes from latest main, while the agent RPM is built from the branch.
So the test compares:
server = latest Quay/main
vs
agent = branch RPM
and fails correctly.
To fix it properly, rpm-sanity should only run when backend and agent are built from the same source, or custom branch runs should exclude rpm-sanity unless they also deploy a matching backend build.

eldar101:
closing this:
The proper fix is to make CI use one resolved FlightCtl version/tag for both. For this PR branch run, the failure is expected unless we also deploy a matching backend tag or skip rpm-sanity.

PR #3282, opened 2026-07-22
NO-ISSUE: Wait for API endpoint after PAM auth restart

eldar101:
Fixes a quadlet externaldb authprovider flake where the PAM browser login starts immediately after restarting PAM/API services, but the public gateway endpoint is not reachable yet.

The test now waits for the external API /readyz endpoint before running flightctl login, so it does not fail with connect: connection refused while the gateway/API exposure is still settling.

Test passes:
https://jenkins-csb-kniqe-ci.dno.corp.redhat.com/job/ocp-flightctl-gotests/2027/testReport/

## Release target

- [x] release-1.3

Target release: release-1.3


eldar101:
Will test on quadlets externaldb and update.

itzikb-redhat (approved):

PR #3286, opened 2026-07-23
EDM-4843: Retry imagebuild/export streams in case of EOF error

itzikb-redhat:


internal/cli/logs.go
```diff
@@ -193,15 +193,18 @@ func isTransientStreamError(err error) bool {
 	}
 	if errors.Is(err, syscall.ECONNRESET) ||
 		errors.Is(err, syscall.EPIPE) ||
-		errors.Is(err, net.ErrClosed) {
+		errors.Is(err, net.ErrClosed) ||
+		errors.Is(err, io.EOF) ||
```
asafbennatan, line 197: do we need to retry on EOF ? isnt that the proper way the log is marked as finished ? ( i dont fully recall )

itzikb-redhat: I think it's https://github.com/flightctl/flightctl/blob/main/api/imagebuilder/v1alpha1/consts.go#L19

asafbennatan (approved):
LGTM

PR #3287, opened 2026-07-23
EDM-4876: Added capabilities to devices summary

celdrake:
Add a breakdown of the counts of devices for available capabilities.
Currently it only handles `osMode`

CLI:
```
flightctl fc get dev --summary-only                    
DEVICES
5

STATUS TYPE	STATUS		COUNT
SYSTEM		Online		2
SYSTEM		Unknown		3
UPDATED		OutOfDate	4
UPDATED		UpToDate	1
APPLICATIONS	NoApplications	2
APPLICATIONS	Unknown		3

CAPABILITY	VALUE	COUNT
OS MODE		image	2
OS MODE		package	1
OS MODE		unknown	2
```


celdrake:
I'll review the linting and Coderabbit comments.
FYI we should ensure this aligns well with the future requirements in https://redhat.atlassian.net/browse/EDM-4865

asafbennatan (dismissed):
see below

internal/store/model/devices_summary_test.go
```diff
@@ -0,0 +1,93 @@
+package model
+
+import (
+	"testing"
+
+	"github.com/flightctl/flightctl/api/core/v1beta1"
```
asafbennatan: **[HIGH / Conventions]** This import violates `store-no-api` depguard (CI lint is failing on it). Store tests should use `internal/domain`, same as sibling model tests.

Root cause: `internal/domain/device.go` aliases `DevicesSummaryCapabilities` but not the types needed to build test values.

Suggested fix — add to `internal/domain/device.go`:
```go
type DeviceCapabilities = v1beta1.DeviceCapabilities
type OsModeType = v1beta1.OsModeType

const (
	OsModeImage   = v1beta1.OsModeImage
	OsModePackage = v1beta1.OsModePackage
)
```
Then construct via `domain.DeviceCapabilities{OsMode: &image}` (etc.) and drop the `api/core/v1beta1` import.

internal/store/model/devices_summary.go
```diff
@@ -0,0 +1,36 @@
...
+// paths) into CapabilityCountUnknown. Other keys are left unchanged.
+func NormalizeCapabilityCounts(counts map[string]int64) map[string]int64 {
+	if counts == nil {
+		counts = make(map[string]int64)
+	}
+	if n, ok := counts[""]; ok {
+		counts[CapabilityCountUnknown] += n
+		delete(counts, "")
+	}
+	return counts
```
asafbennatan: **[MEDIUM / Design]** `NormalizeCapabilityCounts` mutates the input map in place while also returning it. Current call sites discard the original immediately, so no live bug — but the dual return+mutate contract is a footgun if a future caller retains the map.

Prefer building and returning a new map (copy keys, move `""` → `CapabilityCountUnknown`) instead of `delete`/in-place writes on the argument.

internal/cli/display/table.go
```diff
@@ -215,6 +215,13 @@ func (f *TableFormatter) printDevicesSummaryTable(w *tabwriter.Writer, summary *
 	for k, v := range summary.ApplicationStatus {
 		f.printTableRowLn(w, "APPLICATIONS", k, fmt.Sprintf("%d", v))
 	}
+
+	if summary.Capabilities != nil && summary.Capabilities.OsMode != nil && len(*summary.Capabilities.OsMode) > 0 {
+		fmt.Fprintln(w)
```
asafbennatan, line 225: **[MEDIUM / Correctness]** This blank line starts a new `text/tabwriter` block, so OS MODE columns align independently from SYSTEM/UPDATED/APPLICATIONS above (verified with `tabwriter`).

Drop this `fmt.Fprintln(w)` so the rows share one aligned block, **or** `Flush()` and treat capabilities as a separate section with its own header.

asafbennatan (commented):
see below

internal/cli/display/table.go
```diff
@@ -215,6 +215,13 @@ func (f *TableFormatter) printDevicesSummaryTable(w *tabwriter.Writer, summary *
...
+		for k, v := range *summary.Capabilities.OsMode {
+			f.printTableRowLn(w, "OS MODE (capability)", k, fmt.Sprintf("%d", v))
+		}
```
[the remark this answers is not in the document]
celdrake: Added sorting for all the elements.

internal/store/model/devices_summary_test.go
celdrake: Done, thanks

internal/store/model/devices_summary.go
celdrake: Done

internal/cli/display/table.go
celdrake: I updated the description with the new output.
I think adding a new section is the best option, given that we'll likely extend the "capabilities" system soon.

asafbennatan (approved):
LGTM

PR #3288, opened 2026-07-23
EDM-4335: status update path optimizations

asafbennatan:
## Summary
- Avoid re-fetching the device in `UpdateStatus` when callers already have a pre-update snapshot (`ReplaceDeviceStatus`, server-side status updates).
- Persist rendered config and derived status atomically via an `UpdateRendered` status mutator: apply `SpecValid=Valid`, compute server-side status against that committed condition set, and write both in the same CAS update.
- Skip no-op `SetServiceConditions` DB writes and callbacks when conditions are unchanged (common on worker render when SpecValid is already Valid).
- Treat `StoreAndNotify` as best-effort after a successful persist so a notify failure cannot flip freshly committed `SpecValid=Valid` back to False via `setErrorStatus`.
- Keep `setErrorStatus` for actual render/validation failures only (renamed from the old catch-all `setStatus` path).

## Release target
- release-1.3

## Test plan
- [x] `go test ./internal/store/device/ ./internal/service/device/ ./internal/tasks/ -count=1`
- [x] Compile `./test/integration/store/` and `./test/integration/tasks/`
- [x] Store integration: no-op `SetServiceConditions` leaves `resourceVersion`/callback unchanged
- [x] Integration coverage for `UpdateRenderedDevice` success path (rendered version, events, notify)
- [ ] Broader e2e as needed for status update and device-render paths

## Measured improvement
vs `v1.3.0-main-335-g94098fd2`:

| | Before | After |
|---|---|---|
| Duration | 22m (1370s) | 9m (587s) |
| Rollout rate | 7.3 devices/s | 17.04 devices/s |

### DIFF
--- a/internal/service/device/handler.go
+++ b/internal/service/device/handler.go
@@ -562,10 +562,7 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
 		h.log.Debugf("Rendered device %s/%s: no change in rendered version", orgId, name)
 		return domain.StatusOK()
 	}
-	err = h.UpdateServerSideDeviceStatus(ctx, orgId, name)
-	if err != nil {
-		return common.StoreErrorToApiStatus(err, false, domain.DeviceKind, &name)
-	}
+	// Status refresh is owned by device-render setStatus (avoids a duplicate write).
 
 	err = rendered.Bus.Instance().StoreAndNotify(ctx, orgId, name, renderedVersion)
 	if err != nil {
--- a/internal/service/device/handler.go
+++ b/internal/service/device/handler.go
@@ -567,13 +567,7 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
 			updated = nil
 			if device.Status != nil {
 				oldConditions = append([]domain.Condition(nil), device.Status.Conditions...)
-			}
-			// device.Status.Conditions still holds the pre-render SpecValid at this point
-			// (oldConditions above is the diff baseline for event emission). Apply the
-			// SpecValid=Valid that this render is about to commit before computing derived
-			// status, so e.g. Status.Updated (via IsUpdatedToDeviceSpec) isn't computed
-			// against a stale Invalid condition from before this render.
-			if device.Status != nil {
+				// Apply the SpecValid this render is about to commit before computing derived status.
 				domain.SetStatusCondition(&device.Status.Conditions, specValid)
 			}
 			if !common.UpdateServiceSideStatus(ctx, orgId, device, h.fleetStore, h.log) {
@@ -588,8 +582,7 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
 	}
 	if renderedVersion == "" {
 		h.log.Debugf("Rendered device %s/%s: no change in rendered version", orgId, name)
-		// No rendered write; still ensure SpecValid=Valid (e.g. recovered from Invalid), and
-		// refresh derived status since it may have been computed against the stale condition.
+		// No rendered write; still ensure SpecValid=Valid and refresh derived status.
 		status := h.SetDeviceServiceConditions(ctx, orgId, name, []domain.Condition{specValid})
 		if status.Code != http.StatusOK {
 			return status
@@ -612,10 +605,7 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
 		h.diffAndEmitConditionEvents(ctx, orgId, deviceForEvents, oldConditions, newConditions)
 	}
 
-	// StoreAndNotify only wakes up agents already long-polling for a new version; it is not
-	// the source of truth (the render above already persisted successfully). Treat failure
-	// here as best-effort: agents still pick up the new rendered version on their next poll.
-	// Do not fail the call or touch SpecValid over a notify-only failure.
+	// Best-effort: StoreAndNotify only wakes up already-polling agents, it's not the source of truth.
 	if err := rendered.Bus.Instance().StoreAndNotify(ctx, orgId, name, renderedVersion); err != nil {
 		h.log.Errorf("Failed to publish rendered device %s/%s: %v", orgId, name, err)
 	}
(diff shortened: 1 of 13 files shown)


internal/service/device/handler.go
```diff
@@ -545,10 +561,7 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
 		h.log.Debugf("Rendered device %s/%s: no change in rendered version", orgId, name)
 		return domain.StatusOK()
 	}
-	err = h.UpdateServerSideDeviceStatus(ctx, orgId, name)
-	if err != nil {
-		return common.StoreErrorToApiStatus(err, false, domain.DeviceKind, &name)
-	}
+	// Status refresh is owned by device-render setStatus (avoids a duplicate write).
 
 	err = rendered.Bus.Instance().StoreAndNotify(ctx, orgId, name, renderedVersion)
```
galel12: nit: ordering change — we used to refresh server-side status before StoreAndNotify, now notify happens first and setStatus does the refresh after. Small window where agents/UI can see the new rendered version with stale Updated/Summary. Probably fine since setStatus is right there in device-render, just want to confirm it's intentional.

(Also means UpdateServerSideDeviceStatus failure no longer fails this call / risks SpecValid going Invalid via setStatus(renderErr) — that part seems better.)

internal/service/device/handler_test.go
```diff
@@ -829,12 +829,20 @@ func TestSetDeviceServiceConditions(t *testing.T) {
...
+	}
+	status := svc.SetDeviceServiceConditions(ctx, orgId, "foo", []domain.Condition{condition})
 	require.Equal(t, int32(http.StatusOK), status.Code)
 	// SpecValid transitioning from absent to invalid emits a DeviceSpecInvalid event.
 	require.Len(t, ev.created, 1)
+
+	status = svc.SetDeviceServiceConditions(ctx, orgId, "foo", []domain.Condition{condition})
+	require.Equal(t, int32(http.StatusOK), status.Code)
+	// Unchanged conditions must not write or emit another event.
+	require.Len(t, ev.created, 1)
```
galel12, line 845: nit: this proves no second event via the fake, but can't catch a real resource_version bump on a no-op. A tiny store integration assert on identical re-apply would lock the DB-write claim. Non-blocking.

galel12 (commented):
Looks good overall — solid hot-path cuts. Two notes below, neither blocking.

internal/service/device/handler.go
asafbennatan: i think this is fine , the event is published to wake up long polling agents who do not care about the status.

api(called directly or via cli/ui) is not exposed to the publish event anyway so from its perspective it shouldn't care about its ordering . 

IMO i think its fine - the event says that the spec changed not the status

internal/service/device/handler_test.go
asafbennatan: Good call — added a store integration assert in `test/integration/store/device_test.go` (`SetServiceConditions skips write and callback when conditions are unchanged`): first apply fires the callback and advances `resourceVersion`; an identical re-apply leaves both unchanged.

internal/tasks/device_render.go
```diff
@@ -204,23 +204,23 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 	if device.Metadata.Owner == nil || *device.Metadata.Owner == "" {
 		status = t.deviceSvc.OverwriteDeviceRepositoryRefs(ctx, t.orgId, *device.Metadata.Name, referencedRepos...)
 		if status.Code != http.StatusOK {
-			return t.setStatus(ctx, fmt.Errorf("setting repository references: %s", status.Message))
+			return t.setErrorStatus(ctx, fmt.Errorf("setting repository references: %s", status.Message))
```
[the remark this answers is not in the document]
asafbennatan: The repository-reference failure path (line 207) is intentionally left as-is here — it already called `setStatus`/set `SpecValid=False` on `main` before this PR (this branch only renamed `setStatus` to `setErrorStatus`, no behavior change at that call site). It's a legitimate finding, but a pre-existing one, so we're not folding it into this PR's scope.

The other part of this finding — `UpdateRenderedDevice` overwriting a just-persisted `SpecValid=Valid` with `False` when the post-persistence `StoreAndNotify` publish fails — is a real regression introduced by this PR's atomic-write refactor (on `main`, `SpecValid` was never written as part of the render's DB write, so this couldn't happen). Fixed by having `UpdateRenderedDevice` treat `StoreAndNotify` failure as best-effort (log and return OK) instead of propagating it as a failure, since it only affects the wake-up notification for already-long-polling agents, not the durably persisted render.

internal/store/device/store.go
```diff
@@ -1175,10 +1183,29 @@ func (s *DeviceStore) updateRendered(ctx context.Context, orgId uuid.UUID, name,
...
+			deviceOnlyRecord, convertErr := model.NewDeviceFromApiResource(apiDevice)
+			if convertErr != nil {
+				return false, "", convertErr
+			}
+			updates["status"] = deviceOnlyRecord.Status
+		}
 	}
 
+	// SpecValid=Valid is part of a successful render write (same CAS UPDATE).
+	updates["service_conditions"] = withSpecValidCondition(updatedServiceConditions, existingRecord.ServiceConditions)
```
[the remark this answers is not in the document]
asafbennatan: Confirmed — `updateServerSideDeviceUpdatedStatus`'s `IsUpdatedToDeviceSpec()`/`isSpecValidConditionUpdated()` does read `SpecValid` off `apiDevice.Status.Conditions`, and that snapshot was built from the stale pre-render `service_conditions`.

Rather than reordering `updateRendered` (which would also shift the `oldConditions` baseline that `UpdateRenderedDevice` uses for event-diffing, since that baseline is captured from the same merged snapshot), I applied the fix inside the `UpdateRenderedDevice` mutator itself: after capturing `oldConditions` for event emission, it now applies `SpecValid=Valid` to `device.Status.Conditions` before calling `UpdateServiceSideStatus`, so derived status is computed against the value actually being committed. `NewDeviceFromApiResource` still splits it back out into `service_conditions` afterward, so nothing changes about where it's persisted.

Also extended the `renderedVersion == ""` recovery branch to call `UpdateServerSideDeviceStatus` after correcting `SpecValid`, for the same reason.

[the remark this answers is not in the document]
asafbennatan: The fix is on the branch now:

- `UpdateRenderedDevice` mutator applies `SpecValid=Valid` to `device.Status.Conditions` before `UpdateServiceSideStatus`
- empty `renderedVersion` path calls `UpdateServerSideDeviceStatus` after `SetDeviceServiceConditions`

Resolving this thread.

galel12 (approved):
Looks good — approving.

Prior notes closed: status+SpecValid land in the same CAS write; store no-op assert is in place. StoreAndNotify best-effort is the right durability tradeoff.

nit: PR summary still describes the intermediate design (drop status refresh from `UpdateRenderedDevice`; rely on post-`setStatus`). HEAD is atomic mutator + `setErrorStatus`-only + best-effort notify. One-line update would help reviewers/backport.

PR #3292, opened 2026-07-24
EDM-4780: [QE] E2E tests for encryption-at-rest

amalykhi:
## EDM-4780: [QE] E2E tests for encryption-at-rest 

**Jira:** https://redhat.atlassian.net/browse/EDM-4780
**Story type:** [QE]

### Summary

Adds a dedicated `test/e2e/encryption/` suite that validates the encryption-at-rest Phase 1 feature end-to-end. Tests verify that sensitive credential fields (OIDC `clientSecret`, SSH private keys, HTTP passwords) are encrypted in the database using the `enc:v1:<key-id>:...` ciphertext format, that API GET responses always redact these fields as `*****`, and that encryption-specific behaviors (key rotation, backup key material, Prometheus metrics) work correctly across deployment environments.

### Test Infrastructure

- **Suite location:** `test/e2e/encryption/`
- **Reference suite:** `test/e2e/authprovider/` (primary — API-only, Keycloak, no VM); `test/e2e/backup_restore/` (secondary — backup helper and archive inspection patterns)
- **Auxiliary services:** Keycloak (AuthProvider OIDC tests), Git server (SSH/HTTP Repository tests), Prometheus (metrics tests)
- **Test infrastructure used:** `e2e.SetupWorkerHarnessWithoutVM`, `harness.CreateRepositoryWithSSHCredentials`, `harness.CreateHTTPRepository`, `harness.WaitForRepositoryAccessible`, `harness.WaitForRepositoryNotAccessible`, `harness.NewBackupRestore`, `p.Infra.ExecInService` (psql queries), `p.Infra.GetServiceConfig`/`SetServiceConfig` (key rotation), `p.Lifecycle.Restart`/`WaitForReady`, `p.Infra.ExposeService` (Prometheus), `harness.PromQueryResultCount`, `harness.PromQueryHasLabels`

### DIFF
--- a/test/e2e/authprovider/authprovider_test.go
+++ b/test/e2e/authprovider/authprovider_test.go
@@ -30,6 +30,7 @@ import (
 
 	api "github.com/flightctl/flightctl/api/core/v1beta1"
 	"github.com/flightctl/flightctl/internal/quadlet/renderer"
+	authproviderhelpers "github.com/flightctl/flightctl/test/e2e/authprovider/helpers"
 	"github.com/flightctl/flightctl/test/e2e/infra"
 	"github.com/flightctl/flightctl/test/e2e/infra/auxiliary"
 	quadletinfra "github.com/flightctl/flightctl/test/e2e/infra/quadlet"
@@ -602,21 +603,14 @@ func checkAPIEndpointReady(ctx context.Context, apiEndpoint string) error {
 				MinVersion:         tls.VersionTLS12,
 			},
 		},
-		CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
-			return http.ErrUseLastResponse
-		},
 		Timeout: authConfigHTTPTimeout,
 	}
 	resp, err := client.Do(req)
 	if err != nil {
 		return fmt.Errorf("GET %s/readyz: %w", strings.TrimRight(apiEndpoint, "/"), err)
 	}
-	defer func() {
-		if err := resp.Body.Close(); err != nil {
-			logrus.Warnf("[authprovider] failed to close API readiness response body: %v", err)
-		}
-	}()
-	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
+	defer resp.Body.Close()
+	if resp.StatusCode >= http.StatusBadRequest {
 		return fmt.Errorf("GET %s/readyz returned %d", strings.TrimRight(apiEndpoint, "/"), resp.StatusCode)
 	}
 	return nil
@@ -938,6 +932,7 @@ func writeAndApplyAuthProviderManifest(harness *e2e.Harness, providerPath, provi
 }
 
 // deleteAuthProviderByName deletes a dynamic authprovider through the harness resource API.
+// It restores admin login before deletion to handle cases where an OIDC provider is currently active.
 func deleteAuthProviderByName(harness *e2e.Harness, providerName string) (string, error) {
 	if harness == nil {
 		return "", fmt.Errorf("worker harness is required")
@@ -948,7 +943,7 @@ func deleteAuthProviderByName(harness *e2e.Harness, providerName string) (string
 	if err := restoreAdminLoginForResourceManagement(harness); err != nil {
 		return "", err
 	}
-	return harness.ManageResource("delete", "authprovider", providerName)
+	return authproviderhelpers.DeleteAuthProvider(harness, providerName)
 }
 
 // restoreAdminLoginForResourceManagement resets the CLI session to an admin user before authprovider mutations.
@@ -1800,31 +1795,7 @@ func authProviderPackageDir() string {
 
 // buildOIDCAuthProviderYAML renders a dynamic OIDC authprovider manifest for this suite.
 func buildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret string, enabled bool) string {
-	return fmt.Sprintf(`apiVersion: flightctl.io/v1beta1
-kind: AuthProvider
-metadata:
-  name: %s
-spec:
-  providerType: oidc
-  displayName: %s
-  issuer: %s
-  clientId: %s
-  clientSecret: %s
-  enabled: %t
-  scopes:
-    - openid
-    - profile
-    - email
-  usernameClaim:
-    - preferred_username
-  organizationAssignment:
-    type: static
-    organizationName: %s
-  roleAssignment:
-    type: static
-    roles:
-      - %s
-`, name, name, issuerURL, clientID, clientSecret, enabled, defaultOrganizationName, defaultAdminRole)
+	return authproviderhelpers.BuildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret, enabled)
 }
 
 // buildKeycloakOAuth2AuthProviderYAML renders a Keycloak-backed dynamic OAuth2 authprovider manifest for this suite.
--- a/test/e2e/authprovider/helpers/helpers.go
+++ b/test/e2e/authprovider/helpers/helpers.go
@@ -1,0 +1,87 @@
+package authproviderhelpers
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+// BuildOIDCAuthProviderYAML renders a minimal OIDC AuthProvider manifest string.
+// When enabled is false the provider is created in disabled state (safe for e2e tests
+// that only need to exercise encryption or credential storage, not live authentication).
+func BuildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret string, enabled bool) string {
+	return fmt.Sprintf(`apiVersion: flightctl.io/v1beta1
+kind: AuthProvider
+metadata:
+  name: %q
+spec:
+  providerType: oidc
+  displayName: %q
+  issuer: %q
+  clientId: %q
+  clientSecret: %q
+  enabled: %t
+  scopes:
+    - openid
+    - profile
+    - email
+  usernameClaim:
+    - preferred_username
+  organizationAssignment:
+    type: static
+    organizationName: default
+  roleAssignment:
+    type: static
+    roles:
+      - flightctl-admin
+`, name, name, issuerURL, clientID, clientSecret, enabled)
+}
+
+// ApplyManifest writes manifestYAML to a temp file and applies it via the harness CLI.
+// Login must be established by the caller before invoking this function.
+func ApplyManifest(harness *e2e.Harness, manifestYAML string) (string, error) {
+	if harness == nil {
+		return "", fmt.Errorf("ApplyManifest: harness is required")
+	}
+	if manifestYAML == "" {
+		return "", fmt.Errorf("ApplyManifest: manifestYAML is empty")
+	}
+	tmp, err := os.CreateTemp("", "authprovider-manifest-*.yaml")
+	if err != nil {
+		GinkgoWriter.Printf("ApplyManifest: create temp file failed: %v\n", err)
+		return "", fmt.Errorf("create temp manifest file: %w", err)
+	}
+	defer os.Remove(tmp.Name())
+	defer tmp.Close() //nolint:errcheck
+	if _, err := tmp.WriteString(manifestYAML); err != nil {
+		GinkgoWriter.Printf("ApplyManifest: write temp file failed: %v\n", err)
+		return "", fmt.Errorf("write manifest: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return "", fmt.Errorf("close temp file: %w", err)
+	}
+	out, err := harness.ApplyResource(tmp.Name())
+	if err != nil {
+		GinkgoWriter.Printf("ApplyManifest: ApplyResource failed: %v\n", err)
+	}
+	return out, err
+}
+
+// DeleteAuthProvider deletes an AuthProvider CR by name via the harness CLI.
+// Login must be established by the caller before invoking this function.
+// Returns the CLI output (e.g. "authprovider/name deleted") so callers can assert on it.
+func DeleteAuthProvider(harness *e2e.Harness, name string) (string, error) {
+	if harness == nil {
+		return "", fmt.Errorf("DeleteAuthProvider: harness is required")
+	}
+	if name == "" {
+		return "", fmt.Errorf("DeleteAuthProvider: name is required")
+	}
+	out, err := harness.ManageResource("delete", "authprovider", name)
+	if err != nil {
+		GinkgoWriter.Printf("DeleteAuthProvider: ManageResource failed for %q: %v\n", name, err)
+	}
+	return out, err
+}
--- a/test/e2e/encryption/backup_restore_test.go
+++ b/test/e2e/encryption/backup_restore_test.go
@@ -1,0 +1,193 @@
+package encryption_test
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/flightctl/flightctl/test/e2e/infra"
+	"github.com/flightctl/flightctl/test/e2e/infra/setup"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/login"
+	testutil "github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Encryption at rest — Backup", Label("encryption", "backup-restore"), func() {
+	var (
+		harness *e2e.Harness
+		br      *e2e.BackupRestore
+	)
+
+	BeforeEach(func() {
+		if reason := backupRestoreExternalDBSkipReason(); reason != "" {
+			Skip(reason)
+		}
+		harness = e2e.GetWorkerHarness()
+		p := setup.GetDefaultProviders()
+		br = harness.NewBackupRestore(p)
+	})
+
+	Context("When a backup is taken with encryption at rest enabled", func() {
+		It("S3: backup archive should include encryption key material", Label("90094"), func() {
+			By("running flightctl-backup to create an archive")
+			outputDir := GinkgoT().TempDir()
+			archivePath, _, err := br.RunFlightCtlBackup(outputDir)
+			Expect(err).ToNot(HaveOccurred(), "backup must succeed")
+			Expect(archivePath).ToNot(BeEmpty(), "backup must return an archive path")
+			Expect(archivePath).To(BeAnExistingFile(), "backup archive must exist on disk")
+			GinkgoWriter.Printf("Backup archive path: %s\n", archivePath)
+
+			By("listing archive entries")
+			entries, err := listTarGzEntries(archivePath)
+			Expect(err).ToNot(HaveOccurred(), "must be able to read archive entries")
+			Expect(entries).ToNot(BeEmpty(), "archive must have entries")
+
+			By("verifying at least one encryption entry is present in the archive")
+			prefix := encryptionKeyArchiveEntryPrefix()
+			var encryptionEntries []string
+			for _, e := range entries {
+				if strings.HasPrefix(e, prefix) {
+					encryptionEntries = append(encryptionEntries, e)
+				}
+			}
+			Expect(encryptionEntries).ToNot(BeEmpty(),
+				"backup archive must contain encryption key material under %q; entries: %v",
+				prefix, entries)
+
+			By("logging found encryption archive entries")
+			GinkgoWriter.Printf("Encryption entries in backup: %v\n", encryptionEntries)
+		})
+	})
+
+	Context("When restoring from a backup taken before a key rotation", Serial, func() {
+		var (
+			providers   *infra.Providers
+			savedConfig string
+		)
+
+		BeforeEach(func() {
+			providers = setup.GetDefaultProviders()
+			savedConfig = ""
+		})
+
+		// Safety net: if the test mutated the encryption config and did not clean up (e.g. due to
+		// a mid-test failure), restore the original config so subsequent specs can start cleanly.
+		AfterEach(func() {
+			if savedConfig != "" {
+				// Remove stale canary rows before restarting. If S4 failed after rotating the key
+				// (line 123) but before the restore completed, the DB contains a canary for the
+				// rotated key (post-backup-key). Restoring the original config (which only has
+				// "default") without deleting that canary causes ValidateCanaries to crash-loop
+				// on startup (cannot decrypt the rotated-key canary with the restored config).
+				deleteStaleCanaries(defaultKeyID)
+				// Reset migration checkpoints so the restored default config does not inherit a
+				// Complete=true checkpoint from this run's key rotation.
+				resetMigrationCheckpoints()
+				if err := restoreEncryptionConfig(savedConfig); err != nil {
+					GinkgoWriter.Printf("Warning: restore encryption config failed: %v\n", err)
+				}
+				if err := restartServicesAndWait(); err != nil {
+					GinkgoWriter.Printf("Warning: service restart after config restore failed: %v\n", err)
+				}
+				if _, err := login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout); err != nil {
+					GinkgoWriter.Printf("Warning: re-login after config restore failed: %v\n", err)
+				}
+			}
+		})
+
+		It("S4: restore should revert DB ciphertexts to the backup-time key and services should decrypt correctly", Label("90151"), func() {
+			const rotatedKeyID = "post-backup-key"
+
+			By("creating a pre-backup AuthProvider with a clientSecret encrypted under the default key")
+			Expect(auxSvcs).ToNot(BeNil(), "auxiliary services must be initialized")
+			apName := "enc-restore-ap-" + harness.GetTestIDFromContext()
+			secretBytes, err := generateAESKey()
+			Expect(err).ToNot(HaveOccurred(), "generate AuthProvider client secret")
+			clientSecret := fmt.Sprintf("pre-backup-secret-%x", secretBytes[:8])
+			manifest := buildOIDCAuthProviderYAML(apName, auxSvcs.Keycloak.IssuerURL(), "flightctl-client", clientSecret)
+			_, err = applyManifest(harness, manifest)
+			Expect(err).ToNot(HaveOccurred(), "create pre-backup AuthProvider")
+			DeferCleanup(func() { _ = deleteAuthProvider(harness, apName) })
+
+			By("confirming pre-backup DB ciphertext uses the default key")
+			apCipherBefore, err := queryDB(providers, fmt.Sprintf(
+				"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", apName,
+			))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(apCipherBefore, defaultKeyID)).To(BeTrue(),
+				"pre-backup AuthProvider ciphertext must use default key; got: %s", apCipherBefore)
+
+			By("running flightctl-backup to capture the pre-rotation state")
+			outputDir := GinkgoT().TempDir()
+			archivePath, _, err := br.RunFlightCtlBackup(outputDir)
+			Expect(err).ToNot(HaveOccurred(), "backup must succeed")
+			Expect(archivePath).ToNot(BeEmpty(), "backup must return an archive path")
+			Expect(archivePath).To(BeAnExistingFile(), "backup archive must exist on disk")
+			GinkgoWriter.Printf("Backup archive: %s\n", archivePath)
+
+			By("rotating the active encryption key to mutate state post-backup")
+			newKeyBytes, err := generateAESKey()
+			Expect(err).ToNot(HaveOccurred(), "generate new AES key for post-backup rotation")
+			savedConfig, err = rotateEncryptionKey(rotatedKeyID, newKeyBytes)
+			Expect(err).ToNot(HaveOccurred(), "rotate encryption key post-backup")
+			Expect(restartServicesAndWait()).To(Succeed(), "restart services after key rotation")
+			_, err = login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout)
+			Expect(err).ToNot(HaveOccurred(), "re-login after key rotation restart")
+
+			By("triggering re-encryption so the DB ciphertext uses the rotated key")
+			updatedSecret := fmt.Sprintf("post-rotation-secret-%x", newKeyBytes[:8])
+			updatedManifest := buildOIDCAuthProviderYAML(apName, auxSvcs.Keycloak.IssuerURL(), "flightctl-client", updatedSecret)
+			_, err = applyManifest(harness, updatedManifest)
+			Expect(err).ToNot(HaveOccurred(), "update AuthProvider to trigger re-encryption")
+
+			By("confirming DB ciphertext now uses the rotated key")
+			apCipherRotated, err := queryDB(providers, fmt.Sprintf(
+				"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", apName,
+			))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(apCipherRotated, rotatedKeyID)).To(BeTrue(),
+				"post-rotation AuthProvider ciphertext must use rotated key; got: %s", apCipherRotated)
+			GinkgoWriter.Printf("Post-rotation ciphertext prefix confirmed: %s\n", apCipherRotated[:min(40, len(apCipherRotated))])
+
+			By("running flightctl-restore to revert to the pre-rotation backup")
+			Expect(br.RunFlightCtlRestoreFromArchive(archivePath)).To(Succeed(), "restore from backup must succeed")
+
+			By("waiting for all services to be running after restore")
+			Eventually(br.VerifyAllServicesRunning,
+				testutil.LONG_TIMEOUT, testutil.EVENTUALLY_POLLING_250,
+			).Should(Succeed(), "all services must be running after restore")
+
+			By("waiting for the API to be reachable after restore")
+			Eventually(func() error {
+				_, loginErr := login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout)
+				return loginErr
+			}, testutil.LONG_TIMEOUT, testutil.EVENTUALLY_POLLING_250,
+			).Should(Succeed(), "API must be reachable after restore")
+
+			By("verifying DB ciphertext reverted to the backup-time default key")
+			apCipherRestored, err := queryDB(providers, fmt.Sprintf(
+				"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", apName,
+			))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(apCipherRestored, defaultKeyID)).To(BeTrue(),
+				"restored AuthProvider ciphertext must use default key; got: %s", apCipherRestored)
+			GinkgoWriter.Printf("Restored ciphertext prefix confirmed: %s\n", apCipherRestored[:min(40, len(apCipherRestored))])
+
+			By("verifying the API can read and decrypt the restored AuthProvider")
+			out, err := harness.CLI("get", "authprovider", apName)
+			Expect(err).ToNot(HaveOccurred(), "get AuthProvider after restore must succeed")
+			Expect(out).To(ContainSubstring(apName), "get output must mention the AuthProvider name")
+
+			// Restore succeeded — clear savedConfig so AfterEach does not double-restore.
+			// The restore binary already brought services back up with the backup-time key config.
+			savedConfig = ""
+
+			By("verifying services are ready after restore")
+			for _, svc := range []infra.ServiceName{infra.ServiceAPI, infra.ServiceWorker, infra.ServicePeriodic} {
+				Expect(providers.Lifecycle.WaitForReady(svc, testutil.LONG_TIMEOUT)).To(Succeed(),
+					"service %s must be ready after restore", svc)
+			}
+		})
+	})
+})
--- a/test/e2e/encryption/encryption_suite_test.go
+++ b/test/e2e/encryption/encryption_suite_test.go
@@ -1,0 +1,130 @@
+package encryption_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/flightctl/flightctl/test/e2e/infra/auxiliary"
+	"github.com/flightctl/flightctl/test/e2e/infra/setup"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/login"
+	testutil "github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	// loginRetryTimeout is how long to retry login during recovery restarts.
+	// The API pod may report Ready in K8s before the process accepts TLS connections.
+	loginRetryTimeout = 2 * time.Minute
+	// loginProbeTimeout is how long to retry login at the start of each spec before
+	// triggering recovery. Kept short so a crash-loop is detected quickly without
+	// burning 2 minutes per spec on a broken cluster.
+	loginProbeTimeout = 30 * time.Second
+)
+
+var (
+	auxSvcs *auxiliary.Services
+	// originalServiceConfig is the API service config captured at suite start before
+	// any test mutations. The recovery path uses it to restore services to a known-good
+	// state without synthesizing an encryption block that may reference key files that
+	// don't exist on this deployment.
+	originalServiceConfig string
+)
+
+func TestEncryption(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Encryption at Rest E2E Suite")
+}
+
+var _ = BeforeSuite(func() {
+	Expect(setup.EnsureDefaultProviders(nil)).To(Succeed())
+
+	ctx := context.Background()
+
+	// Start all needed auxiliary services:
+	// - Keycloak for OIDC AuthProvider creation (clientSecret encryption tests)
+	// - Git server for SSH Repository credential tests
+	// - Prometheus for metrics validation
+	var err error
+	auxSvcs, err = auxiliary.StartServices(ctx, []auxiliary.Service{
+		auxiliary.ServiceKeycloak,
+		auxiliary.ServiceGitServer,
+		auxiliary.ServicePrometheus,
+	})
+	Expect(err).ToNot(HaveOccurred(), "failed to start auxiliary services")
+	Expect(auxSvcs.Keycloak.URL).ToNot(BeEmpty(), "Keycloak URL must not be empty")
+
+	_, _, err = e2e.SetupWorkerHarnessWithoutVM()
+	Expect(err).ToNot(HaveOccurred(), "failed to setup worker harness")
+
+	// Capture the original API config before any test mutations so the recovery path
+	// can restore it exactly — not synthesize an encryption block that may point to
+	// key files that don't exist on this deployment.
+	originalServiceConfig = captureOriginalServiceConfig()
+
+	harness := e2e.GetWorkerHarness()
+	if _, loginErr := login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout); loginErr != nil {
+		// The API may be in a crash-loop from a previous test run that left services with a
+		// rotated encryption key config that prevents startup. Restore the original config
+		// (captured above) and restart so the suite can proceed.
+		GinkgoWriter.Printf("BeforeSuite: initial login failed (%v); attempting recovery by restoring original service config\n", loginErr)
+		Expect(recoverServicesToOriginalConfig()).To(Succeed(), "recovery: restore original service config")
+		Expect(restartServicesAndWait()).To(Succeed(), "recovery: restart services after config restore")
+		_, err = login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout)
+		Expect(err).ToNot(HaveOccurred(), "bootstrap admin login failed after recovery")
+	}
+	// Best-effort cleanup of stable resources that may have been left by a previous crashed run.
+	// enc-rot-ap-stable uses the same issuer+clientId as other test specs; if it persists from a
+	// prior crash, subsequent tests fail with 409 "issuer and clientId already exists".
+	cleanUpStableEncryptionResources(harness)
+})
+
+var _ = BeforeEach(func() {
+	harness := e2e.GetWorkerHarness()
+	suiteCtx := e2e.GetWorkerContext()
+
+	ctx := testutil.StartSpecTracerForGinkgo(suiteCtx)
+	harness.SetTestContext(ctx)
+
+	// Probe with a short timeout first. If the API is down (e.g. a previous test
+	// crashed it with a bad encryption config), trigger the same recovery as
+	// BeforeSuite rather than waiting 2 minutes per spec.
+	if _, loginErr := login.LoginToAPIWithTokenWithRetry(harness, loginProbeTimeout); loginErr != nil {
+		GinkgoWriter.Printf("BeforeEach: login probe failed (%v); attempting recovery by restoring original service config\n", loginErr)
+		Expect(recoverServicesToOriginalConfig()).To(Succeed(), "recovery: restore original service config")
+		Expect(restartServicesAndWait()).To(Succeed(), "recovery: restart services after config restore")
+		_, err := login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout)
+		Expect(err).ToNot(HaveOccurred(), "restore admin login before spec (after recovery)")
+	}
+	// Always clean up stable resources that may have been left by a previous crashed spec.
+	// enc-rot-ap-stable uses the same issuer+clientId as S4 and S8; if it persists after a
+	// crash (AfterAll ran with the API down and could not delete it), subsequent specs get
+	// a 409 even when the API is healthy and the recovery branch above is not entered.
+	cleanUpStableEncryptionResources(harness)
+})
+
+var _ = AfterEach(func() {
+	harness := e2e.GetWorkerHarness()
+	suiteCtx := e2e.GetWorkerContext()
+
+	harness.PrintAgentLogsIfFailed()
+	harness.CaptureDeploymentLogsIfFailed()
+
+	err := harness.CleanUpAllTestResources()
+	Expect(err).ToNot(HaveOccurred(), "cleanup test resources")
+
+	harness.SetTestContext(suiteCtx)
+})
+
+var _ = AfterSuite(func() {
+	if auxSvcs != nil {
+		auxSvcs.Cleanup(context.Background())
+	}
+})
+
+func init() {
+	SetDefaultEventuallyTimeout(testutil.DURATION_TIMEOUT)
+	SetDefaultEventuallyPollingInterval(testutil.EVENTUALLY_POLLING_250)
+}
--- a/test/e2e/encryption/helpers_test.go
+++ b/test/e2e/encryption/helpers_test.go
@@ -1,0 +1,439 @@
+package encryption_test
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	internalconfig "github.com/flightctl/flightctl/internal/config"
+	authproviderhelpers "github.com/flightctl/flightctl/test/e2e/authprovider/helpers"
+	"github.com/flightctl/flightctl/test/e2e/infra"
+	"github.com/flightctl/flightctl/test/e2e/infra/setup"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	testutil "github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+const (
+	// defaultKeyID is the encryption key ID used at service startup.
+	defaultKeyID = "default"
+)
+
+// ciphertextMatchesKeyID reports whether value begins with the expected ciphertext prefix
+// "enc:v1:<keyID>:" indicating the value was encrypted under keyID.
+func ciphertextMatchesKeyID(value, keyID string) bool {
+	return strings.HasPrefix(value, fmt.Sprintf("enc:v1:%s:", keyID))
+}
+
+// queryDB executes a psql query against the built-in flightctl database and returns
+// the trimmed output. Requires a reachable flightctl-db pod; fails on external-DB deployments.
+func queryDB(p *infra.Providers, sql string) (string, error) {
+	return infra.QueryDB(p, sql)
+}
+
+// buildOIDCAuthProviderYAML renders a minimal OIDC AuthProvider manifest string for encryption tests.
+// All encryption tests create providers with enabled=false since they only exercise credential storage.
+func buildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret string) string {
+	return authproviderhelpers.BuildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret, false)
+}
+
+// applyManifest writes manifest YAML to a temp file and applies it via the harness CLI.
+// Login must be established by the caller before invoking this function.
+func applyManifest(harness *e2e.Harness, manifestYAML string) (string, error) {
+	return authproviderhelpers.ApplyManifest(harness, manifestYAML)
+}
+
+// deleteAuthProvider deletes an AuthProvider CR by name via the harness CLI.
+// Login must be established by the caller before invoking this function.
+func deleteAuthProvider(harness *e2e.Harness, name string) error {
+	_, err := authproviderhelpers.DeleteAuthProvider(harness, name)
+	return err
+}
+
+// writeKeyToService makes a named encryption key available to the given service.
+// It delegates to InfraProvider.SetEncryptionKey, which:
+//   - On K8s/OCP: patches the flightctl-encryption-key Secret and triggers a rollout restart
+//     of the service's deployment so pods mount the updated (read-only projected) Secret.
+//   - On Quadlet: writes the key file directly to the host at /etc/flightctl/encryption/<keyFileName>,
+//     which is bind-mounted into the container.
+func writeKeyToService(svc infra.ServiceName, keyFileName string, keyBytes []byte) error {
+	if keyFileName == "" {
+		return fmt.Errorf("writeKeyToService: keyFileName is required")
+	}
+	if len(keyBytes) == 0 {
+		return fmt.Errorf("writeKeyToService: keyBytes is empty")
+	}
+	p := setup.GetDefaultProviders()
+	if p == nil || p.Infra == nil {
+		return fmt.Errorf("writeKeyToService: providers not initialized")
+	}
+	if err := p.Infra.SetEncryptionKey(svc, keyFileName, keyBytes); err != nil {
+		GinkgoWriter.Printf("writeKeyToService: failed to set key %s on %s: %v\n", keyFileName, svc, err)
+		return fmt.Errorf("write key file %s on %s: %w", keyFileName, svc, err)
+	}
+	return nil
+}
+
+// setEncryptionConfigOnAllServices writes the given typed encryption config to all three services
+// (API, worker, periodic) using the infra interface.
+func setEncryptionConfigOnAllServices(enc *internalconfig.EncryptionConfig) error {
+	p := setup.GetDefaultProviders()
+	if p == nil || p.Infra == nil {
+		return fmt.Errorf("setEncryptionConfigOnAllServices: providers not initialized")
+	}
+	for _, svc := range []infra.ServiceName{infra.ServiceAPI, infra.ServiceWorker, infra.ServicePeriodic} {
+		if err := p.Infra.SetEncryptionConfig(svc, enc); err != nil {
+			GinkgoWriter.Printf("setEncryptionConfigOnAllServices: failed to set config on %s: %v\n", svc, err)
+			return fmt.Errorf("set encryption config on %s: %w", svc, err)
+		}
+	}
+	return nil
+}
+
+// setRawConfigOnAllServices writes a raw YAML config string to all three services.
+// Used for restoring a previously saved raw config snapshot.
+func setRawConfigOnAllServices(config string) error {
+	p := setup.GetDefaultProviders()
+	if p == nil || p.Infra == nil {
+		return fmt.Errorf("setRawConfigOnAllServices: providers not initialized")
+	}
+	for _, svc := range []infra.ServiceName{infra.ServiceAPI, infra.ServiceWorker, infra.ServicePeriodic} {
+		if err := p.Infra.SetServiceConfig(svc, "config.yaml", config); err != nil {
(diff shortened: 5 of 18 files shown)


test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,224 @@
...
+			Expect(harness.CreateRepositoryWithSSHCredentials(newRepoName, repoURL, keyContent)).To(
+				Succeed(), "create post-rotation Repository")
+
+			By("[§5.3] verifying new Repository uses the rotated key in DB")
+			newRepoCipher, err := queryDB(providers, fmt.Sprintf(
+				"SELECT spec->'sshConfig'->>'sshPrivateKey' FROM repositories WHERE name = '%s'", newRepoName,
+			))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(newRepoCipher, rotatedKeyID)).To(BeTrue(),
+				"new Repository must use rotated key; got: %s", newRepoCipher)
```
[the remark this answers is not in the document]
amalykhi: Fixed in commit 9fd995f7. Added a cleanup `By` step inside S1 that deletes `newRepoName` immediately after all assertions complete, before the config restore. `AfterAll` handles the stable named resources (`authProviderName`, `repoName`); the post-rotation repository is now cleaned up inline where it's created.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,224 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			By("[§5.4] verifying Repository becomes accessible again after key restore")
+			Expect(harness.WaitForRepositoryAccessible(repoName, testutil.DURATION_TIMEOUT, testutil.EVENTUALLY_POLLING_250)).To(
+				Succeed(), "[§5.4] Repository must be accessible again after encryption key is restored")
+
+			By("final cleanup: restore original config")
+			Expect(restoreEncryptionConfig(savedConfig)).To(Succeed())
+			Expect(restartServicesAndWait()).To(Succeed())
+			savedConfig = ""
```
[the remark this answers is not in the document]
amalykhi: Fixed in commit 9fd995f7. Added `login.LoginToAPIWithToken(harness)` after the final `restartServicesAndWait()` in S2, consistent with every other restart-and-act sequence in this file (lines 115–116, 167–168, 181–182, 191–192, 202–203).

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,396 @@
...
+			id := strings.TrimSpace(strings.TrimPrefix(trimmed, "- id: "))
+			if id == keyIDToRemove {
+				skipNext = true
+				continue
+			}
+		}
+		out = append(out, line)
+	}
+	return strings.Join(out, "\n")
+}
```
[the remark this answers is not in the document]
amalykhi: The manual YAML manipulation in `addKeyEntryToConfig` / `removeKeyEntryFromConfig` is intentional. FlightCtl's config YAML is always generated by its own Helm/quadlet tooling with consistent 2-space indentation, so the hardcoded indentation matches the actual config structure. Introducing `gopkg.in/yaml.v3` round-tripping in a test helper risks silently dropping comments or reordering keys in the live service config, which would be worse than fixed indentation for a controlled test context. The injected values (`activeKeyID`, key entries) are test-generated constants, not user input. Keeping it as-is.

test/e2e/encryption/sensitive_fields_test.go
```diff
@@ -0,0 +1,234 @@
...
+				By("querying the database and verifying the ciphertext format (§2.7)")
+				dbVal, err := queryDB(providers, fmt.Sprintf(
+					"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", authProviderName,
+				))
+				Expect(err).ToNot(HaveOccurred(), "psql query for clientSecret")
+				Expect(dbVal).ToNot(BeEmpty(), "DB value must not be empty")
+				Expect(dbVal).ToNot(ContainSubstring(clientSecret),
+					"DB must not store the plaintext clientSecret")
+				Expect(ciphertextMatchesKeyID(dbVal, defaultKeyID)).To(BeTrue(),
+					"DB value must start with enc:v1:%s: but got: %s", defaultKeyID, dbVal)
```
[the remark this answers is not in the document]
amalykhi: The principle is sound for production SQL, but the risk here is zero: all five query values are test-generated names of the form `enc-ap-<uuid>` / `enc-repo-ssh-<uuid>` (alphanumeric + hyphens only), not user-supplied input. `psql -v` binding adds shell quoting complexity (`:'name'` syntax) with no security benefit in this context. Keeping `fmt.Sprintf` for readability in test-only code.

test/e2e/encryption/sensitive_fields_test.go
```diff
@@ -0,0 +1,233 @@
...
+				// AES-GCM uses a random nonce per encryption, so ciphertext differs even for same plaintext.
+				updatedManifest := fmt.Sprintf(`apiVersion: flightctl.io/v1beta1
+kind: Repository
+metadata:
+  name: %s
+spec:
+  type: git
+  url: %s
+  sshConfig:
+    sshPrivateKey: %s
```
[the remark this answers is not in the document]
amalykhi: The test comment explains the intent: `// AES-GCM uses a random nonce per encryption, so ciphertext differs even for same plaintext.` The re-encryption AC (§4.1) requires verifying that updating a credential produces a new ciphertext — which is exactly what this step proves. Supplying a second valid SSH key fixture would require generating a key pair in-process or via the aux service, adding significant test infrastructure complexity for marginal additional coverage. The existing assertion is sufficient for the acceptance criterion being tested.

test/e2e/infra/infra_provider.go
```diff
@@ -111,6 +111,14 @@ type InfraProvider interface {
...
 	// For K8s: checks the Service object exists. For Quadlet: checks the systemd unit is active.
 	ServiceExists(ctx context.Context, service ServiceName) (bool, error)
+
+	// SetEncryptionKey writes a named encryption key so it is available to the given service.
+	// For K8s: patches the flightctl-encryption-key Secret to add/update the key entry, then
+	// triggers a rollout restart of the service's deployment so the new Secret is mounted.
+	// For Quadlet: writes the key file directly to the host filesystem at the encryption key directory.
+	// keyFileName is the base name of the key file (e.g. "key-rotated-key").
+	// keyBytes is the raw key material to store.
+	SetEncryptionKey(service ServiceName, keyFileName string, keyBytes []byte) error
```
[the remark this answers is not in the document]
amalykhi: Declining —  is synchronous test infrastructure called within the Ginkgo test timeout, and the underlying  and file-write operations complete quickly in practice. Adding  across the interface, three implementations (, , and any future providers), and all call sites in the encryption test helpers would be significant churn with no correctness or observability gain in this test-only path. We can revisit if a future implementation genuinely needs cancellation.

test/e2e/encryption/backup_restore_test.go
```diff
@@ -0,0 +1,57 @@
...
+		harness = e2e.GetWorkerHarness()
+		p := setup.GetDefaultProviders()
+		br = harness.NewBackupRestore(p)
+	})
+
+	Context("When a backup is taken with encryption at rest enabled", func() {
+		It("S3: backup archive should include encryption key material", Label("90094"), func() {
+			By("running flightctl-backup to create an archive")
+			outputDir := GinkgoT().TempDir()
+			archivePath, _, err := br.RunFlightCtlBackup(outputDir)
```
eldar101, line 35: I think the archivepath should be printed for logging  purposes

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,420 @@
...
+	defer os.Remove(tmp.Name())
+	defer tmp.Close() //nolint:errcheck
+	if _, err := tmp.WriteString(manifestYAML); err != nil {
+		GinkgoWriter.Printf("applyManifest: write temp file failed: %v\n", err)
+		return "", fmt.Errorf("write manifest: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return "", fmt.Errorf("close temp file: %w", err)
+	}
+	out, err := harness.ApplyResource(tmp.Name())
```
eldar101: This applies secret-bearing manifests through ApplyResource, which goes through CLIWithStdin and logs stdin. doesnt this espose secrets?

test/e2e/encryption/metrics_test.go
```diff
@@ -0,0 +1,77 @@
...
+			).Should(BeTrue(),
+				"flightctl_encryption_active_key_info must have strategy, key_id, algorithm labels")
+		})
+
+		It("S5: should expose encryption operation counter metric with expected labels", Label("90098"), func() {
+			By("creating an AuthProvider to trigger an encrypt operation")
+			apName := "enc-metrics-ap-" + harness.GetTestIDFromContext()
+			clientID := "enc-metrics-" + harness.GetTestIDFromContext()
+			manifest := buildOIDCAuthProviderYAML(apName, auxSvcs.Keycloak.IssuerURL(), clientID, "metrics-test-secret")
+			_, err := applyManifest(harness, manifest)
```
eldar101, line 59: This AuthProvider is created from a raw manifest with no test label and no DeferCleanup, so suite cleanup won’t delete it. Please add explicit cleanup after the apply succeeds

test/e2e/encryption/backup_restore_test.go
amalykhi: Fixed in commit 1c70b795. Added `GinkgoWriter.Printf("Backup archive path: %s\n", archivePath)` immediately after the `BeAnExistingFile` assertion, consistent with the existing logging pattern for encryption entries at the end of the test.

test/e2e/encryption/metrics_test.go
amalykhi: Fixed in commit 1c70b795. Added `DeferCleanup` immediately after the `Expect(err).ToNot(HaveOccurred())` assertion so the trigger AuthProvider is deleted regardless of whether subsequent metric assertions pass or fail.

test/e2e/encryption/helpers_test.go
amalykhi: Valid concern — `ApplyResource` reads the file and passes the bytes to `CLIWithStdin`, which calls `SHWithStdin` with `redactStdin=false`, so the full manifest YAML (including `clientSecret`) is logged via `logrus.Infof`.

That said, I'd like to leave this as a follow-up rather than fix it here:

1. The harness already has a `redactStdin bool` parameter on `SHWithStdin` specifically for this purpose — a proper fix would add an `ApplyResourceRedacted` variant on `Harness`, which is harness code outside this PR's scope.
2. The secrets logged here are randomized per-run test fixtures (e.g. `initial-secret-<uuid>`) generated inside the test — not real credentials.
3. Test CI logs are controlled environments, not customer-facing.

Happy to file a follow-up item to add redacted apply support to the harness if that works for you.

itzikb-redhat:
@amalykhi Should deploy/podman templates be in this patch?

test/e2e/encryption/backup_restore_test.go
```diff
@@ -0,0 +1,58 @@
+package encryption_test
```
itzikb-redhat, line 1: Are you also checking restore?

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+// ciphertextMatchesKeyID reports whether value begins with the expected ciphertext prefix
+// "enc:v1:<keyID>:" indicating the value was encrypted under keyID.
+func ciphertextMatchesKeyID(value, keyID string) bool {
+	return strings.HasPrefix(value, fmt.Sprintf("enc:v1:%s:", keyID))
+}
+
+// queryDB executes a psql query against the flightctl database and returns the trimmed output.
+// Returns an error if providers are not initialized or if the query fails.
+func queryDB(p *infra.Providers, sql string) (string, error) {
```
itzikb-redhat, line 34: 1. Will that work with externaldb as well? 2. Should that be in common place?

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,468 @@
...
+	})
+	if err != nil {
+		GinkgoWriter.Printf("queryDB: psql failed (sql=%q): %v\n", sql, err)
+		return "", fmt.Errorf("psql query failed: %w", err)
+	}
+	return strings.TrimSpace(output), nil
+}
+
+// buildOIDCAuthProviderYAML renders a minimal OIDC AuthProvider manifest string for encryption tests.
+func buildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret string) string {
```
itzikb-redhat, line 40: There is a similar function at test/e2e/authprovider/authprovider_test.go - Can we reuse it?

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,468 @@
...
+	out, err := harness.ApplyResource(tmp.Name())
+	if err != nil {
+		GinkgoWriter.Printf("applyManifest: ApplyResource failed: %v\n", err)
+	}
+	return out, err
+}
+
+// deleteAuthProvider deletes an AuthProvider CR by name via the harness CLI.
+// Login must be established by the caller before invoking this function.
+func deleteAuthProvider(harness *e2e.Harness, name string) error {
```
itzikb-redhat, line 52: Same question for deleteAuthProviderByName in test/e2e/authprovider/authprovider_test.go

amalykhi:
> @amalykhi Should deploy/podman templates be in this patch?

@itzikb-redhat which templates?

test/e2e/encryption/backup_restore_test.go
amalykhi: Yes — added S4: **restore should revert DB ciphertexts to the backup-time key and services should decrypt correctly** in a new `Context("When restoring from a backup taken before a key rotation", Serial)` block in `test/e2e/encryption/backup_restore_test.go`.

The test flow:
1. Creates a pre-backup AuthProvider with `clientSecret` encrypted under the default key; confirms `enc:v1:default:` prefix in DB.
2. Runs `flightctl-backup` to capture the pre-rotation state.
3. Rotates the active key to a new key ID and triggers re-encryption (updates the AuthProvider), confirming `enc:v1:post-backup-key:` prefix in DB.
4. Runs `flightctl-restore` from the pre-rotation archive.
5. Waits for all services to come back up and the API to be reachable.
6. Verifies the DB ciphertext reverted to `enc:v1:default:`.
7. Verifies the API can still read and decrypt the restored AuthProvider.

test/e2e/encryption/helpers_test.go
amalykhi: 1. **External DB**: `queryDB` calls `ExecInService(ServiceDB, psql ...)`, which exec's into the built-in `flightctl-db` pod. On external DB deployments there's no such pod and this will fail. The backup/restore tests are already guarded by `backupRestoreExternalDBSkipReason()` which checks `BuiltinDatabaseWorkloadAvailable()`. The key rotation and sensitive-fields tests that also call `queryDB` verify raw DB ciphertext — there's no API-level way to observe it, so they inherently require direct DB access. A similar skip guard for those contexts on external DB is the right fix; will track as a follow-up.

2. **Common place**: Moving `queryDB` out of a `_test` package into a shared helper (e.g. `test/harness/e2e/`) is feasible but is non-trivial restructuring given the function goes through the infra provider interface. Today it's only used within the encryption e2e suite. Happy to move it if the team expects it to be reused across other suites.

test/e2e/encryption/helpers_test.go
amalykhi: Can't reuse it directly — both files are `_test` packages so they can't be imported across packages. The authprovider version also has a different signature (it takes an extra `enabled bool` parameter) and returns a different set of fields. The local copy is a small format string tailored to the encryption suite's needs (`enabled: false`, minimal field set). Moving it to a shared non-test helper would require restructuring both suites into a new package — reasonable to consider if more suites need it, but out of scope for this PR.

amalykhi:
@itzikb-redhat The `deploy/podman/*/config.yaml.template` changes are conflict resolutions from rebasing onto main, not new changes introduced by this PR. Upstream added the `{{- with .encryption }}` block to those templates; resolving the rebase conflict required keeping the upstream version. They need to stay in the patch to keep the branch merge-clean.

test/e2e/encryption/helpers_test.go
amalykhi: same

test/e2e/encryption/helpers_test.go
amalykhi: Done. Extracted `BuildOIDCAuthProviderYAML`, `ApplyManifest`, and `DeleteAuthProvider` into a new importable package `test/e2e/authprovider/helpers` (commit e35ab466). Both suites now delegate to the shared implementation. The encryption suite keeps local unexported shim wrappers so none of the existing call sites needed to change.

test/e2e/encryption/helpers_test.go
amalykhi: covered

amalykhi: Done along with `buildOIDCAuthProviderYAML`. `DeleteAuthProvider` is now in `test/e2e/authprovider/helpers` (commit e35ab466); `deleteAuthProviderByName` in the authprovider suite delegates to it while keeping its local login-restore logic.

test/e2e/encryption/helpers_test.go
amalykhi: Done for question 2: `QueryDB` is now in `test/e2e/infra/db.go` (package `infra`) so any suite can import it — the vulnerability suite's inline `ExecInService(ServiceDB, psql ...)` is already migrated (commit ae621da6). The encryption suite keeps a local unexported shim.

For question 1 (external DB): the function is documented to require the built-in `flightctl-db` pod and will fail on external-DB deployments. The backup/restore tests are already guarded by `backupRestoreExternalDBSkipReason()`. The key rotation and sensitive-fields tests call `queryDB` to verify raw DB ciphertexts — there's no API-level way to observe the ciphertext, so a similar skip guard for those contexts on external DB is the right long-term fix; tracking as a follow-up.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+			By("cleanup: restoring original config and restarting")
+			Expect(restoreEncryptionConfig(savedConfig)).To(Succeed())
+			Expect(restartServicesAndWait()).To(Succeed())
+			_, err = login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout)
+			Expect(err).ToNot(HaveOccurred())
+			savedConfig = "" // cleared so AfterEach safety net does not double-restore
+		})
+
+		It("S2: should report inaccessible condition when old key is removed, and recover when key is restored", Label("90095"), func() {
```
asafbss: S2 has a race condition and may be fundamentally unreliable.

After rotating to rotatedKeyID and restarting, two things race:

Migration worker — it will re-encrypt the Repository from default → rotated. If it finishes before removeEncryptionKey(defaultKeyID) runs (likely — there's only one repo), the repo is already on the rotated key. Removing default has no effect → WaitForRepositoryNotAccessible times out → test fails.

Canary validation — a canary was created for the default key. When default is removed from config and services restart, canaryservice.InitEncryption → ValidateCanaries tries to decrypt the default canary, fails, and returns an error → log.Fatalf → service crash-loops. LoginToAPIWithTokenWithRetry may also fail.

If this test passes, it's by timing luck, not by design.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,372 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(newRepoCipher, rotatedKeyID)).To(BeTrue(),
+				"new Repository must use rotated key; got: %s", newRepoCipher)
+
+			By("verifying original Repository (old key) is still accessible after rotation")
+			// Decryption happens in the background task; WaitForRepositoryAccessible polls Accessible=True.
+			Expect(harness.WaitForRepositoryAccessible(repoName, testutil.DURATION_TIMEOUT, testutil.EVENTUALLY_POLLING_250)).To(
+				Succeed(), "original Repository must stay accessible when old key is still in key ring")
+
+			By("verifying original Repository DB ciphertext is unchanged (not re-encrypted until touched)")
```
asafbss: After rotation + restart (line 253), the migration worker starts automatically and will re-encrypt the repo from default → rotated. The test assumes the repo stays on default "until touched" (line 286) — but the migration worker is touching it. If migration completes before the DB query at line 287 runs, origRepoCipher will have rotated as the key ID → assertion fails.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,372 @@
...
+			By("verifying updated AuthProvider uses the rotated key in DB")
+			updatedAPCipher, err := queryDB(providers, fmt.Sprintf(
+				"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", authProviderName,
+			))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(updatedAPCipher, rotatedKeyID)).To(BeTrue(),
+				"updated AuthProvider must use rotated key after update; got: %s", updatedAPCipher)
+
+			By("cleanup: restoring original config and restarting")
+			Expect(restoreEncryptionConfig(savedConfig)).To(Succeed())
```
asafbss, line 378: During S1, a canary was created for rotated-key. After restoring the original config (which only has default), the rotated-key canary is still in DB. On restart, ValidateCanaries tries to decrypt it with a key that no longer exists in config → log.Fatalf → crash-loop. restartServicesAndWait may fail here. This same issue applies to S8 cleanup (line 135-140).

test/e2e/encryption/negative_test.go
```diff
@@ -0,0 +1,314 @@
...
+					"LIST Repositories must not contain PEM headers")
+				Expect(repoListBody).To(ContainSubstring("*****"),
+					"LIST Repositories must show redacted placeholder")
+			})
+	})
+
+	// N1: A plaintext value stored directly in DB (e.g. pre-migration data) is
+	// re-encrypted on the next API write. The plaintext-passthrough path in
+	// Manager.Decrypt is backward-compatible; ProcessEncryption re-encrypts on save.
+	Context("When a sensitive field stored as plaintext in the DB is updated via the API", func() {
```
asafbss, line 99: The test injects plaintext into DB then does an explicit API update to trigger re-encryption. This tests the "encrypt-on-write" path, which is correct — but the test description says "pre-migration data" (line 97). The real upgrade scenario is that the migration worker re-encrypts plaintext rows without any user action. That path is untested?

test/e2e/encryption/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+			apName := "enc-metrics-noerr-ap-" + harness.GetTestIDFromContext()
+			clientID := "enc-metrics-noerr-" + harness.GetTestIDFromContext()
+			manifest := buildOIDCAuthProviderYAML(apName, auxSvcs.Keycloak.IssuerURL(), clientID, "no-error-secret")
+			_, err := applyManifest(harness, manifest)
+			Expect(err).ToNot(HaveOccurred(), "create AuthProvider to drive a clean encrypt operation")
+			DeferCleanup(func() {
+				_ = deleteAuthProvider(harness, apName)
+			})
+
+			By("capturing baseline error count before the operation")
```
asafbss: The baseline is captured after creating the AuthProvider not before. If the creation itself produced an error that incremented errors_total, the baseline already includes it and the test passes incorrectly. The baseline should be captured before the operation.

test/e2e/encryption/encryption_suite_test.go
```diff
@@ -0,0 +1,107 @@
...
+	// key files that don't exist on this deployment.
+	originalServiceConfig = captureOriginalServiceConfig()
+
+	harness := e2e.GetWorkerHarness()
+	if _, loginErr := login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout); loginErr != nil {
+		// The API may be in a crash-loop from a previous test run that left services with a
+		// rotated encryption key config that prevents startup. Restore the original config
+		// (captured above) and restart so the suite can proceed.
+		GinkgoWriter.Printf("BeforeSuite: initial login failed (%v); attempting recovery by restoring original service config\n", loginErr)
+		Expect(recoverServicesToOriginalConfig()).To(Succeed(), "recovery: restore original service config")
```
asafbss, line 73: Stale canaries remain? ValidateCanaries will try to decrypt them with a key that doesn't exist

asafbss (commented):
Good coverage of the core encryption paths (field encryption, redaction, metrics, backup). No tests for Device RenderedConfig/RenderedApplications?

test/e2e/encryption/backup_restore_test.go
amalykhi: Yes — S4 ("restore should revert DB ciphertexts to the backup-time key and services should decrypt correctly") is a full backup+restore cycle: it creates an AuthProvider under the default key, takes a backup, rotates the key (so the DB now uses a different ciphertext prefix), then runs `flightctl-restore` and asserts that the DB ciphertext reverts to the backup-time default key and the API can decrypt it again.

test/e2e/encryption/helpers_test.go
amalykhi: 1. **External DB**: the backup/restore tests that call `queryDB` are guarded by `backupRestoreExternalDBSkipReason()` in their `BeforeEach` — that function checks both `E2E_EXTERNAL_DATABASE=1` and `BuiltinDatabaseWorkloadAvailable()` and skips the spec when either indicates no built-in DB pod. The sensitive-fields tests that call `queryDB` directly target the built-in DB and will simply fail on external-DB deployments (same as the analogous DB-querying tests elsewhere in the suite). 2. **Common place**: `queryDB` is a one-liner shim over `infra.QueryDB` (in `test/e2e/infra/db.go`), which is already the shared package. The shim just saves callers from importing the infra package directly; happy to inline `infra.QueryDB` at call sites if you prefer.

test/e2e/encryption/helpers_test.go
amalykhi: Done in this PR — `buildOIDCAuthProviderYAML` is now extracted into `test/e2e/authprovider/helpers/` (package `authproviderhelpers`) and both suites delegate to it. The encryption suite calls `authproviderhelpers.BuildOIDCAuthProviderYAML(name, issuerURL, clientID, clientSecret, false)` and the authprovider suite calls the same function with `enabled` as a parameter. `DeleteAuthProvider` and `ApplyManifest` are also in the shared package.

test/e2e/encryption/keyrotation_test.go
amalykhi: Fixed in 484318bf. Both races you describe were real:

1. **Migration race**: after rotating to `rotatedKeyID` and restarting, the migration worker re-encrypts the repo under `rotatedKeyID`. Removing `default` then had no effect (repo already on `rotated`) — `WaitForRepositoryNotAccessible` timed out. Fixed by waiting for migration to complete first (using `Eventually` on the ciphertext), then removing `rotatedKeyID` (the active key) instead of `default`. This guarantees the repo is undecryptable when the key is removed.

2. **Canary crash-loop**: the `rotatedKeyID` canary row remains in DB after config restore. On next restart `ValidateCanaries` tries to decrypt it with a key not in config → `log.Fatalf` → crash-loop. Fixed by adding `deleteStaleCanaries(defaultKeyID)` that runs `DELETE FROM encryption_canaries WHERE key_id != 'default'` before each config restore and restart in S1, S2, S8, their `AfterEach` safety nets, and `recoverServicesToOriginalConfig`.

test/e2e/encryption/keyrotation_test.go
amalykhi: Fixed in 484318bf. Removed the racy assertion (`origRepoCipher uses default key until touched`). After rotation + restart, `EnqueueEncryptionMigrationIfNeeded` fires immediately because the checkpoint does not match the new active key — with one Repository the race window is narrow but real. S8 already covers migration comprehensively with `Eventually`; S1 now only asserts: (a) new writes use the new key, (b) old data remains accessible (decryptable via the key ring), and (c) explicit updates re-encrypt under the new key.

test/e2e/encryption/keyrotation_test.go
amalykhi: Fixed in 484318bf. The `CanaryStore` interface has no `Delete` method, but the underlying Postgres table does. Added `deleteStaleCanaries(keepKeyID string)` in `helpers_test.go` which runs `DELETE FROM encryption_canaries WHERE key_id != 'default'` via `queryDB`. It is called in S1 cleanup, S8 cleanup, both `AfterEach` safety nets (S8-metrics and S1/S2), and `recoverServicesToOriginalConfig` before every config-restore + restart cycle.

test/e2e/encryption/negative_test.go
amalykhi: Valid distinction. N1 tests the **encrypt-on-write** path (API update triggers `ProcessEncryption` → re-encrypts), not the **background migration** path. S8 covers the migration worker path — it verifies that after key rotation and restart, the worker auto-migrates pre-existing resources without any user action, using `Eventually` to wait for the ciphertext to change.

Updated the N1 comment to clarify: "N1: A plaintext value stored directly in DB (e.g. pre-migration data from a downgrade or data import) is re-encrypted on the next API write via `ProcessEncryption`. The background migration worker path (automatic re-encryption without user action) is covered in S8."

test/e2e/encryption/metrics_test.go
amalykhi: Fixed in 484318bf. Moved `baselineErrorCount := harness.PromQueryResultCount(promURL, errQuery)()` to before the `applyManifest` call, so any error produced by AuthProvider creation is counted as a delta rather than baked into the baseline.

test/e2e/encryption/encryption_suite_test.go
amalykhi: Fixed in 484318bf via the same `deleteStaleCanaries(defaultKeyID)` call added to `recoverServicesToOriginalConfig`. The BeforeSuite and BeforeEach recovery paths both call `recoverServicesToOriginalConfig` before restarting, so stale canary rows are cleaned up before any restart that could crash-loop on `ValidateCanaries`.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,393 @@
...
+			manifest := buildOIDCAuthProviderYAML(authProviderName, issuerURL, "flightctl-client", preRotationSecret)
+			out, err := applyManifest(harness, manifest)
+			Expect(err).ToNot(HaveOccurred(), "create pre-rotation AuthProvider")
+			Expect(out).ToNot(BeEmpty(), "apply must produce output")
+
+			keyContent, err := auxSvcs.GetGitSSHPrivateKey()
+			Expect(err).ToNot(HaveOccurred(), "get git SSH key")
+			repoURL := fmt.Sprintf("user@%s:%d:/home/user/repos/test.git",
+				auxSvcs.GitServer.Host, auxSvcs.GitServer.Port)
+			Expect(harness.CreateRepositoryWithSSHCredentials(repoName, repoURL, keyContent)).To(
```
asafbennatan: BeforeAll creates the Flightctl Repository CR but never creates the bare git repo on the aux server. S2 then waits on Accessible / NotAccessible (below), which is only a meaningful decrypt signal if SSH to a real repo can succeed once decrypt works. With a missing path, those waits can time out or flip for the wrong reason.

Should we mirror the N2 setup here — unique bare repo, `GetInternalGitRepoURL`, and cleanup — so the Accessible waits actually exercise decrypt success/failure? Something like:

```go
gitRepoName := "enc-s1s2-" + harness.GetTestIDFromContext()
keyPath, err := auxSvcs.GetGitSSHPrivateKeyPath()
gitServerConfig := e2e.GitServerConfig{
    Host: auxSvcs.GitServer.Host,
    Port: auxSvcs.GitServer.Port,
    User: "user",
}
Expect(harness.CreateGitRepositoryOnServer(gitServerConfig, keyPath, gitRepoName)).To(Succeed())
DeferCleanup(func() {
    _ = harness.DeleteGitRepositoryOnServer(gitServerConfig, keyPath, gitRepoName)
})
repoURL, err := harness.GetInternalGitRepoURL(
    auxSvcs.GitServer.InternalHost, auxSvcs.GitServer.InternalPort, gitRepoName)
Expect(harness.CreateRepositoryWithSSHCredentials(repoName, repoURL, keyContent)).To(Succeed())
```

test/e2e/encryption/metrics_test.go
```diff
@@ -0,0 +1,127 @@
...
+				"flightctl_encryption_operations_total must have operation, strategy, key_id, status labels")
+		})
+
+		It("S6: should report no encryption errors during normal operation", Label("90128"), func() {
+			// Capture the baseline before creating the AuthProvider so any error
+			// produced by the creation itself is counted as a delta, not part of
+			// the baseline. Prometheus counters are process-lifetime values; other
+			// tests (e.g. N2) may have already incremented the counter.
+			errQuery := `flightctl_encryption_errors_total`
+			baselineErrorCount := harness.PromQueryResultCount(promURL, errQuery)()
```
asafbennatan: `PromQueryResultCount` returns the number of matching series, not the counter value. An error that increments an existing series leaves the series count unchanged, so this can pass incorrectly.

Should we baseline/compare the numeric value instead, e.g.:

```go
errQuery := `sum(flightctl_encryption_errors_total)`
baseline := harness.PromQueryCountValue(promURL, errQuery)()
// ... successful encrypt ...
Expect(harness.PromQueryCountValue(promURL, errQuery)()).To(Equal(baseline))
```

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,498 @@
...
+		Path string `json:"path"`
+	} `json:"keys"`
+}
+
+// addKeyEntryToConfig injects a new key entry into the encryption.keys list and updates
+// activeKeyID in the YAML config string. The key is stored at encryptionKeyDir/<keyFileName>.
+// If the config has no encryption: block (e.g. Quadlet where defaults are baked into the binary),
+// the full block is appended, including the implicit default key at encryptionKeyDir/key.
+// Uses YAML marshal/unmarshal to avoid indentation-sensitivity issues across environments.
+func addKeyEntryToConfig(config, newKeyID, keyFileName string) (string, error) {
```
asafbennatan: Key material correctly goes through infra (`SetEncryptionKey` / `ResetEncryptionKeys`), but config edits still live in the suite and encode deployment details: the in-container key path, K8s vs Quadlet mount differences, and Quadlet-specific “no encryption block” handling.

Should those details live behind typed encryption-config get/set on infra instead, so callers stay deployment-agnostic? For example:

```go
// on InfraProvider
GetEncryptionConfig() (*config.EncryptionConfig, error)
SetEncryptionConfig(enc *config.EncryptionConfig) error
```

```go
enc, err := p.Infra.GetEncryptionConfig()
// mutate enc.Keys / enc.ActiveKeyID
err = p.Infra.SetEncryptionConfig(enc)
err = p.Infra.SetEncryptionKey(svc, keyFileName, keyBytes)
```

Related: `removeKeyEntryFromConfig` swallows YAML errors and returns the original config unchanged — should failures propagate so S2 does not wait for Accessible=False while the key is still present?

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,393 @@
...
+				))
+				if err != nil {
+					return false
+				}
+				return ciphertextMatchesKeyID(cipher, rotatedKeyID)
+			}, testutil.DURATION_TIMEOUT, testutil.EVENTUALLY_POLLING_250,
+			).Should(BeTrue(),
+				"migration worker must re-encrypt pre-rotation Repository under the rotated key")
+
+			By("cleanup: removing stale canary rows and restoring original config")
```
asafbennatan: S8 waits for migration to re-encrypt under the new active key, then cleanup uses `deleteStaleCanaries` + config restore. Suite scaffolding for isolation is fine — but I’m not sure the product “key retirement after migration” behavior is covered as an e2e outcome.

Should we assert that path without coupling to internals (e.g. `encryption_canaries` / `PrepareForRetirement`)? A table-level check would break if those change even when user-visible behavior is still correct.

Would a black-box check fit better once migration has finished — e.g. remove the previous key from the ring, restart, and assert services stay healthy and previously migrated resources still decrypt — so we cover the retirement contract via behavior rather than DB shape? Or is that already covered elsewhere and I’m missing it?

asafbennatan (changes requested):
See inline comments below.

test/e2e/encryption/keyrotation_test.go
amalykhi: Fixed in bbc794aa. BeforeAll now creates a bare git repo named `enc-rot-s1s2-stable` on the aux server via `CreateGitRepositoryOnServer`, and uses `GetInternalGitRepoURL(InternalHost, InternalPort, ...)` for the Repository CR URL so repotester (running inside the cluster) can reach it via the cluster-reachable address. AfterAll cleans up the server-side repo. This mirrors the N2 pattern.

amalykhi: Fixed in 7086a9f2. BeforeAll now calls `CreateGitRepositoryOnServer` to create a real bare git repo on the aux server, then uses `GetInternalGitRepoURL(InternalHost, InternalPort)` to build the URL — so repotester (running inside k8s) resolves the cluster-reachable address rather than the test-runner address. AfterAll calls `DeleteGitRepositoryOnServer` for cleanup. Pattern mirrors the N2 test.

test/e2e/encryption/metrics_test.go
amalykhi: Fixed in 7086a9f2. Switched S6 to `PromQueryCountValue("sum(flightctl_encryption_errors_total)")` for both the baseline capture and the final assertion. This reads the numeric counter value, not the number of matching series, so an error that increments an existing label-set is detected.

test/e2e/encryption/helpers_test.go
amalykhi: Partial fix + decline. The typed-interface refactor is out of scope for this PR — the YAML manipulation is intentional and is already isolated in two helpers with clear names and a test-only scope.

The silent error swallowing is fixed in 7086a9f2: `removeKeyEntryFromConfig` now returns `(string, error)` and propagates all YAML parse/marshal errors. The call site in `removeEncryptionKey` checks the error and returns early rather than proceeding with the key still present.

test/e2e/encryption/keyrotation_test.go
amalykhi: Disagree — S2 already covers this behavioral contract. After the migration worker finishes re-encrypting all resources under `rotated-key`, S2 removes that key from the config, restarts services, and asserts:

1. `WaitForRepositoryNotAccessible` — the pre-rotation Repository is inaccessible (service can no longer decrypt, key is gone)
2. Config is restored to the default key, services restart, and `WaitForRepositoryAccessible` — access recovers

That is the retirement behavioral outcome: resources encrypted under the retired key become inaccessible, and access is restored only after re-encrypting under a live key. No additional test is needed.

test/e2e/infra/quadlet/infra.go
```diff
@@ -689,6 +689,28 @@ func quadletServiceConfigDBType(serviceConfigYAML []byte) string {
...
+func (p *InfraProvider) SetEncryptionKey(_ infra.ServiceName, keyFileName string, keyBytes []byte) error {
+	if keyFileName == "" {
+		return fmt.Errorf("SetEncryptionKey: keyFileName is required")
+	}
+	if len(keyBytes) == 0 {
+		return fmt.Errorf("SetEncryptionKey: keyBytes is empty")
+	}
+	hostPath := filepath.Join(quadletEncryptionKeyDir, keyFileName)
+	if err := p.writeHostFile(hostPath, keyBytes); err != nil {
+		return fmt.Errorf("SetEncryptionKey: write %s: %w", hostPath, err)
```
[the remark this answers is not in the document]
amalykhi: Addressed in an earlier commit — `filepath.Base(keyFileName) != keyFileName` check was added to reject path separators and `..` components before the `filepath.Join` call.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,231 @@
...
+				"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", authProviderName,
+			))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(ciphertextMatchesKeyID(updatedAPCipher, rotatedKeyID)).To(BeTrue(),
+				"updated AuthProvider must use rotated key after update; got: %s", updatedAPCipher)
+
+			By("cleanup: deleting post-rotation Repository")
+			if _, err := harness.ManageResource("delete", "repository", newRepoName); err != nil {
+				GinkgoWriter.Printf("Warning: failed to delete post-rotation repository %s: %v\n", newRepoName, err)
+			}
```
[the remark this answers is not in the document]
amalykhi: Addressed — `DeferCleanup` is registered immediately after `CreateRepositoryWithSSHCredentials` (before any assertions), so the repository is deleted regardless of whether subsequent assertions pass or fail.

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,409 @@
...
+		defaultKeyPath := filepath.Join(encryptionKeyDir, "key")
+		out = append(out,
+			"encryption:",
+			fmt.Sprintf("  activeKeyID: %s", newKeyID),
+			"  keys:",
+			fmt.Sprintf("  - id: %s", defaultKeyID),
+			fmt.Sprintf("    path: %s", defaultKeyPath),
+			fmt.Sprintf("  - id: %s", newKeyID),
+			fmt.Sprintf("    path: %s", newKeyPath),
+		)
```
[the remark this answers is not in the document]
amalykhi: Addressed in an earlier commit — `addKeyEntryToConfig` now tracks `encryptionBlockFound` separately from `activeKeyIDUpdated`, and returns an error if an `encryption:` block is present but has no `activeKeyID`, preventing a duplicate top-level key from being appended.

amalykhi:
@coderabbitai review

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,467 @@
...
+			).Should(BeTrue(),
+				"migration worker must re-encrypt pre-rotation Repository under the rotated key")
+
+			By("cleanup: removing stale canary rows and restoring original config")
+			deleteStaleCanaries(defaultKeyID)
+			Expect(restoreEncryptionConfig(savedConfig)).To(Succeed())
+			Expect(restartServicesAndWait()).To(Succeed())
+			_, err = login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout)
+			Expect(err).ToNot(HaveOccurred())
+			savedConfig = ""
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Added `resetMigrationCheckpoints()` in both the S8 inline cleanup and the S8 `AfterEach` safety-net. S8 rotates to `rotated-metrics-key` and lets the migration worker finish — the resulting `Complete=true` checkpoint must be cleared before restoring the default config, otherwise a subsequent run that reuses the same key ID skips re-encryption and times out. S1/S2 already did this; S8 was missing it.

test/e2e/encryption/backup_restore_test.go
```diff
@@ -0,0 +1,184 @@
...
+					GinkgoWriter.Printf("Warning: restore encryption config failed: %v\n", err)
+				}
+				if err := restartServicesAndWait(); err != nil {
+					GinkgoWriter.Printf("Warning: service restart after config restore failed: %v\n", err)
+				}
+				if _, err := login.LoginToAPIWithTokenWithRetry(harness, loginRetryTimeout); err != nil {
+					GinkgoWriter.Printf("Warning: re-login after config restore failed: %v\n", err)
+				}
+			}
+		})
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Added `deleteStaleCanaries(defaultKeyID)` and `resetMigrationCheckpoints()` in the S4 `AfterEach` before calling `restoreEncryptionConfig`. If S4 fails after the rotation at line 123 but before the restore completes, the DB holds a canary for `post-backup-key`. Restoring the original config (which only has the `default` key) without deleting that canary causes `ValidateCanaries` to crash-loop on startup. This is the same guard already present in the S1/S2/S8 cleanup paths.

test/e2e/authprovider/authprovider_test.go
```diff
@@ -941,7 +943,7 @@ func deleteAuthProviderByName(harness *e2e.Harness, providerName string) (string
 	if err := restoreAdminLoginForResourceManagement(harness); err != nil {
 		return "", err
 	}
-	return harness.ManageResource("delete", "authprovider", providerName)
+	return "", authproviderhelpers.DeleteAuthProvider(harness, providerName)
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. `DeleteAuthProvider` now returns `(string, error)` and passes through the `ManageResource` output. `deleteAuthProviderByName` in the authprovider suite returns the output directly (removing the hardcoded `""`). The caller at line 184 (`Expect(out).To(ContainSubstring(keycloakLifecycleName))`) now works correctly — it was silently passing before on an always-empty string.

test/e2e/authprovider/helpers/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+    - preferred_username
+  organizationAssignment:
+    type: static
+    organizationName: default
+  roleAssignment:
+    type: static
+    roles:
+      - flightctl-admin
+`, name, name, issuerURL, clientID, clientSecret, enabled)
+}
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Changed all `%s` format verbs for `name`, `issuerURL`, `clientId`, and `clientSecret` to `%q` (Go double-quoted strings). `%q` produces valid YAML double-quoted scalars and handles embedded colons, special characters, and other YAML-sensitive characters correctly.

test/e2e/encryption/metrics_test.go
```diff
@@ -0,0 +1,132 @@
...
+		It("S5: should expose encryption operation counter metric with expected labels", Label("90098"), func() {
+			By("creating an AuthProvider to trigger an encrypt operation")
+			apName := "enc-metrics-ap-" + harness.GetTestIDFromContext()
+			clientID := "enc-metrics-" + harness.GetTestIDFromContext()
+			manifest := buildOIDCAuthProviderYAML(apName, auxSvcs.Keycloak.IssuerURL(), clientID, "metrics-test-secret")
+			_, err := applyManifest(harness, manifest)
+			Expect(err).ToNot(HaveOccurred(), "create AuthProvider to trigger encrypt metric")
+			DeferCleanup(func() {
+				_ = deleteAuthProvider(harness, apName)
+			})
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Both `DeferCleanup` blocks in S5 and S6 now log the error via `GinkgoWriter.Printf` if `deleteAuthProvider` returns an error, rather than silently discarding it.

test/e2e/encryption/metrics_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+			By("verifying flightctl_encryption_errors_total did not increase during normal operations")
+			// Wait for the create operation to be recorded, then confirm errors didn't grow.
+			opsQuery := `flightctl_encryption_operations_total`
+			Eventually(harness.PromQueryResultCount(promURL, opsQuery),
+				testutil.DURATION_TIMEOUT, testutil.EVENTUALLY_POLLING_250,
+			).Should(BeNumerically(">", 0), "must see at least one encryption operation before checking error counter")
+
+			Expect(harness.PromQueryCountValue(promURL, errQuery)()).To(Equal(baselineErrorValue),
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Capture `baselineOpsValue := harness.PromQueryCountValue(promURL, opsQuery)()` before `applyManifest`, then `Eventually` waits for the ops sum to strictly exceed that baseline before checking the error counter. Using the counter value (not series count) ensures the S6 AuthProvider creation operation was actually recorded in Prometheus — a pre-existing series from an earlier test can no longer satisfy the guard.

test/e2e/encryption/sensitive_fields_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+				repoName := "enc-repo-ssh-" + harness.GetTestIDFromContext()
+				repoURL := fmt.Sprintf("user@%s:%d:/home/user/repos/test.git",
+					auxSvcs.GitServer.Host, auxSvcs.GitServer.Port)
+
+				By("creating an SSH-auth Repository")
+				keyContent, err := auxSvcs.GetGitSSHPrivateKey()
+				Expect(err).ToNot(HaveOccurred(), "get git server SSH key")
+				err = harness.CreateRepositoryWithSSHCredentials(repoName, repoURL, keyContent)
+				Expect(err).ToNot(HaveOccurred(), "create SSH Repository")
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Added `DeferCleanup` immediately after `CreateRepositoryWithSSHCredentials` in the SSH test and after `CreateHTTPRepository` in the HTTP test. Both `CreateRepositoryWithSSHCredentials` and `CreateHTTPRepository` already add a test-id label so the suite-level `AfterEach`'s `CleanUpAllTestResources()` handles cleanup, but the explicit `DeferCleanup` adds defense-in-depth and makes intent clear.

test/e2e/infra/db.go
```diff
@@ -0,0 +1,26 @@
...
+	if p == nil || p.Infra == nil {
+		return "", fmt.Errorf("QueryDB: providers not initialized")
+	}
+	output, err := p.Infra.ExecInService(ServiceDB, []string{
+		"psql", "-d", "flightctl", "-t", "-A", "-c", sql,
+	})
+	if err != nil {
+		return "", fmt.Errorf("psql query failed: %w", err)
+	}
+	return strings.TrimSpace(output), nil
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Changed the error format to `fmt.Errorf("psql query failed: %w; output: %s", err, strings.TrimSpace(output))` so the psql failure reason (syntax error, missing relation, connection refused) is visible in the error message.

test/e2e/encryption/negative_test.go
```diff
@@ -0,0 +1,314 @@
...
+					// due to a broken decrypt during any cleanup read path.
+					if savedCipherN2 != "" {
+						_, _ = queryDB(providers, fmt.Sprintf(
+							`UPDATE repositories SET spec = jsonb_set(spec, '{sshConfig,sshPrivateKey}', '"%s"') WHERE name = '%s'`,
+							savedCipherN2, repoName,
+						))
+					}
+					_, _ = harness.ManageResource("delete", "repository", repoName)
+					_ = harness.DeleteGitRepositoryOnServer(gitServerConfig, keyPath, gitRepoName)
+				})
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Registered a dedicated `DeferCleanup` for the git server repo immediately after `CreateGitRepositoryOnServer`, independent of the later `DeferCleanup` for the Repository CR. The later `DeferCleanup` no longer includes `DeleteGitRepositoryOnServer`. This way, if `CreateRepositoryWithSSHCredentials` fails before the second `DeferCleanup` is registered, the git server repo is still cleaned up.

test/login/login.go
```diff
@@ -237,6 +240,53 @@ func LoginToAPIWithToken(harness *e2e.Harness) (AuthMethod, error) {
...
+		lastErr = err
+		logrus.Infof("LoginToAPIWithTokenWithRetry: transient error (%v), retrying in %s", err, backoff)
+		time.Sleep(backoff)
+		backoff *= 2
+		if backoff > maxBackoff {
+			backoff = maxBackoff
+		}
+	}
+	return 0, fmt.Errorf("login did not succeed within %s: %w", timeout, lastErr)
+}
```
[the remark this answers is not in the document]
amalykhi: Fixed in 854b3d81. Converted `for time.Now().Before(deadline) { ... }` to `for { ... if time.Now().After(deadline) { break } ... }`. The loop body now always runs at least once, so `lastErr` is always set before the `return 0, fmt.Errorf("login did not succeed within %s: %w", timeout, lastErr)` line. This prevents the misleading `<nil>` error message when `timeout` is zero or already elapsed.

test/e2e/encryption/encryption_suite_test.go
```diff
@@ -0,0 +1,130 @@
...
+	Expect(err).ToNot(HaveOccurred(), "cleanup test resources")
+
+	harness.SetTestContext(suiteCtx)
+})
+
+var _ = AfterSuite(func() {
+	if auxSvcs != nil {
+		auxSvcs.Cleanup(context.Background())
+	}
+})
```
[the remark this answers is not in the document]
amalykhi: The suite self-heals on next run: both `BeforeSuite` and `BeforeEach` call `recoverServicesToOriginalConfig` on login failure, which resets the config and deletes stale canaries before restarting. An `AfterSuite` recovery would run the same code with no guarantee the API is still reachable (the final spec may have left services in a crash-loop). Adding it would be defensive duplication with unclear benefit. The existing `BeforeSuite`/`BeforeEach` recovery path is the correct mechanism.

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,541 @@
...
+}
+
+// encryptionConfigShape is the YAML shape of the encryption block for marshal/unmarshal.
+type encryptionConfigShape struct {
+	ActiveKeyID string `json:"activeKeyID"`
+	Keys        []struct {
+		ID   string `json:"id"`
+		Path string `json:"path"`
+	} `json:"keys"`
+}
```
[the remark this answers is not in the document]
amalykhi: The anonymous `[]struct{ ID, Path string }` is an inline literal for the `encryptionConfigShape.Keys` field type — Go does not allow a separate named type for an anonymous struct field without also changing the parent struct definition. Extracting it to a named type would require modifying `encryptionConfigShape` and all four literal sites simultaneously, which is a more invasive refactor than the comment suggests. The current code is clear and follows the idiomatic Go pattern for this case.

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,541 @@
...
+		return fmt.Errorf("resetEncryptionConfigToDefault: providers not initialized")
+	}
+	config, err := p.Infra.GetServiceConfig(infra.ServiceAPI)
+	if err != nil {
+		return fmt.Errorf("resetEncryptionConfigToDefault: read API config: %w", err)
+	}
+	var root map[string]interface{}
+	if err := yaml.Unmarshal([]byte(config), &root); err != nil || root == nil {
+		return fmt.Errorf("resetEncryptionConfigToDefault: parse config: %w", err)
+	}
```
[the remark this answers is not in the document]
amalykhi: The condition at line 329 is `if err := yaml.Unmarshal(...); err != nil || root == nil`. When `root` is nil and `err` is nil, `fmt.Errorf("...: %w", err)` produces `"parse config: %!w(<nil>)"` — which is admittedly odd but is still a non-nil error value, so the caller sees a failure, not a false success. Fixing it to `fmt.Errorf("parse config: parsed result is empty")` is the cleaner approach but is a minor cosmetic change. Declining as low priority; happy to apply if this is a blocker for the review.

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,541 @@
...
+	p := setup.GetDefaultProviders()
+	if p == nil || p.Infra == nil {
+		return
+	}
+	if !p.Infra.BuiltinDatabaseWorkloadAvailable() {
+		return
+	}
+	if _, err := queryDB(p, fmt.Sprintf(
+		"DELETE FROM encryption_canaries WHERE key_id != '%s'", keepKeyID,
+	)); err != nil {
```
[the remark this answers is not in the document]
amalykhi: Same rationale as the Round 1 decline on the `queryDB` interpolation: `keepKeyID` is always a compile-time constant (`"default"`) in all call sites. There is no code path where a user-controlled or external value reaches this helper. The `psql -v` binding approach (`:'key_id' = $1`) is more complex for a call that is already a single-line helper with a fully controlled argument. Declining for the same reason the similar pattern was accepted in Round 1.

test/e2e/encryption/keyrotation_test.go
```diff
@@ -0,0 +1,467 @@
...
+			// Getting the repository proves the API can decrypt sshPrivateKey stored under the old
+			// key when both keys are present in the key ring — no SSH connectivity required.
+			out, getErr := harness.CLI("get", "repository", repoName)
+			Expect(getErr).ToNot(HaveOccurred(), "get repository must succeed — API must decrypt old-key data")
+			Expect(out).To(ContainSubstring(repoName), "get output must contain the repository name")
+
+			By("updating the original AuthProvider to trigger re-encryption with rotated key")
+			newSecret := base64.StdEncoding.EncodeToString(newKeyBytes[:12])
+			updatedManifest := buildOIDCAuthProviderYAML(authProviderName, auxSvcs.Keycloak.IssuerURL(),
+				"flightctl-client", newSecret)
```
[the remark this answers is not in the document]
amalykhi: The `clientSecret` is used only for OIDC authentication against a local test Keycloak instance — there is no real security requirement, and reusing the AES key bytes as the secret reduces fixture complexity. CodeRabbitai itself categorized this as `💤 Low value`. Declining.

test/e2e/encryption/negative_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+				By("[N3] listing Repositories and verifying sshPrivateKey is redacted")
+				repoListResp, err := harness.Client.ListRepositoriesWithResponse(ctx, nil)
+				Expect(err).ToNot(HaveOccurred(), "LIST Repositories")
+				Expect(repoListResp.StatusCode()).To(Equal(200), "LIST Repositories must return 200")
+				repoListBody := string(repoListResp.Body)
+				Expect(repoListBody).ToNot(ContainSubstring("BEGIN"),
+					"LIST Repositories must not contain PEM headers")
+				Expect(repoListBody).To(ContainSubstring("*****"),
+					"LIST Repositories must show redacted placeholder")
```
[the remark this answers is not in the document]
amalykhi: The absence assertions (`ToNot(ContainSubstring(secret1/secret2))`) are correct regardless of pagination — if the secrets don't appear on any page, the assertion holds. The presence assertion (`ContainSubstring("*****")`) intentionally checks a global invariant (any AP on the page must show redacted placeholders). This is an acceptable coverage pattern for an e2e negative test that validates the global redaction property. Declining.

test/e2e/encryption/negative_test.go
```diff
@@ -0,0 +1,314 @@
...
+			Expect(err).To(HaveOccurred(),
+				"[N4] apply with fake enc:v1: secret must return an error")
+
+			By("[N4] verifying no AuthProvider was persisted in the DB")
+			dbVal, err := queryDB(providers, fmt.Sprintf(
+				"SELECT spec->>'clientSecret' FROM auth_providers WHERE name = '%s'", authProviderName,
+			))
+			Expect(err).ToNot(HaveOccurred(), "DB query must not error (resource absence is not a query error)")
+			Expect(dbVal).To(BeEmpty(),
+				"[N4] no AuthProvider must exist in DB after rejected write; got: %s", dbVal)
```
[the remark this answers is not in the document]
amalykhi: The load-bearing correctness check is the subsequent DB assertion: `Expect(dbVal).To(BeEmpty())` proves no resource was persisted. The `HaveOccurred()` check gates that assertion. Checking for a specific CLI error message would be brittle across API versions (the error surfaces as an HTTP 400/500 and the exact message could change). Declining.

test/e2e/infra/infra_provider.go
```diff
@@ -111,6 +111,20 @@ type InfraProvider interface {
...
```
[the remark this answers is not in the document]
amalykhi: The doc comment is already accurate — it describes both implementations separately: 'For K8s: patches the Secret to add/update the key entry, then triggers a rollout restart... For Quadlet: writes the key file directly to the host filesystem.' There is no promise of a restart in the Quadlet description. No change needed.

test/e2e/encryption/helpers_test.go
asafbennatan: @amalykhi please have another look at this , developer guideline / agents.md specifically define that deployment specific logic should be in the infra package , there is no need to override this here .

test/e2e/encryption/keyrotation_test.go
asafbennatan: I think S2 covers a different contract than post-migration key retirement.

- **S2:** after migration to `rotated-key`, remove **that same key** (still needed for the ciphertext) → expect Accessible=False. That is “decrypt fails when the key in use is gone.”
- **Retirement:** after migration, remove the **previous** key → expect services stay healthy and previously migrated resources still decrypt under the new key.

Those are opposite outcomes. S2 also calls `deleteStaleCanaries` before removing the key, so startup canary validation is bypassed rather than exercised via product retirement.

Should we cover the retirement happy path once via that black-box check (remove the old key after migration → still healthy / still decrypts), or is that intentionally out of scope for this PR?

test/e2e/encryption/helpers_test.go
```diff
@@ -0,0 +1,542 @@
...
+	for _, svc := range []infra.ServiceName{infra.ServiceAPI, infra.ServiceWorker, infra.ServicePeriodic} {
+		if err := writeKeyToService(svc, keyFileName, keyBytes); err != nil {
+			return "", fmt.Errorf("write key to %s: %w", svc, err)
+		}
+	}
+	updatedConfig, err := addKeyEntryToConfig(savedConfig, newKeyID, keyFileName)
+	if err != nil {
+		return "", fmt.Errorf("build updated config: %w", err)
+	}
+	GinkgoWriter.Printf("rotateEncryptionKey: updatedConfig to be written:\n%s\n", updatedConfig)
```
eldar101: Check there arent any secrets here.

test/e2e/encryption/helpers_test.go
amalykhi: Accepted — the refactor is done. Added `GetEncryptionConfig` / `SetEncryptionConfig` to the `InfraProvider` interface with shared YAML helpers in `test/e2e/infra/encryption_config.go` (`ParseEncryptionConfig`, `MarshalEncryptionConfigIntoYAML`). Both K8s and Quadlet implement the interface. Removed `encryptionConfigShape`, `addKeyEntryToConfig`, `removeKeyEntryFromConfig` from `helpers_test.go`. `rotateEncryptionKey` and `removeEncryptionKey` now call `p.Infra.GetEncryptionConfig` → mutate → `p.Infra.SetEncryptionConfig` — no deployment-specific YAML manipulation in the test package.

test/e2e/encryption/keyrotation_test.go
amalykhi: You're right — these are opposite outcomes and my earlier response was incorrect. S2 removes the key the data is encrypted under (the active rotated key) and expects Accessible=False (decrypt fails). Key retirement removes the superseded (old) key after migration and expects continued success. Fixed: added S3 — "should keep services healthy and resources accessible after retiring the previous key post-migration". S3 rotates, waits for the migration worker to re-encrypt the Repository under the new key, then removes the old (default) key, restarts services, and asserts the Repository stays Accessible=True throughout.

test/e2e/encryption/helpers_test.go
amalykhi: Checked — no secrets here. Line 269 prints `updatedConfig`, which is the service YAML config containing only `encryption.activeKeyID` (e.g. `"rotated-key"`) and `encryption.keys[].{id, path}` (key file paths like `/root/.flightctl/encryption/key-rotated-key`). The actual key bytes are written separately via `SetEncryptionKey` → K8s Secret patch or file write, and never appear in the service config YAML. Key IDs and file paths are not sensitive.

asafbss (approved):

asafbennatan (approved):
LGTM

PR #3302, opened 2026-07-28
NO-ISSUE: Make VM render launcher image and passt workarounds configurable

asafbennatan:
## Summary
- Enable `worker.vmRender.launcherImage` and `worker.vmRender.passtWorkarounds` in Flight Control config (defaults: `quay.io/kubevirt/virt-launcher:v1.8.4`, `passtWorkarounds: true`)
- Wire both options into `vm-to-quadlet` during VM application rendering, and include them in the Quadlet cache key
- Expose the settings in Helm values and Podman `service-config.yaml`, with user docs

## Test plan
- [x] `go test ./internal/tasks/ ./test/e2e/infra/quadlet/`
- [x] `helm lint` / template render of worker config with defaults and overrides
- [x] Podman worker config template render with defaults and overrides
- [ ] CI unit/integration checks on PR

### DIFF
--- a/deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
+++ b/deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
@@ -26,6 +26,11 @@ data:
         {{- end }}
         {{- end }}
     service: {}
+    worker:
+        vmRender:
+            {{- $vmRender := (.Values.worker).vmRender | default dict }}
+            launcherImage: {{ $vmRender.launcherImage | default "quay.io/kubevirt/virt-launcher:v1.8.4" | quote }}
+            passtWorkarounds: {{ if kindIs "bool" $vmRender.passtWorkarounds }}{{ $vmRender.passtWorkarounds }}{{ else }}true{{ end }}
     kv:
         hostname: flightctl-kv.{{ default .Release.Namespace .Values.global.internalNamespace }}.svc.cluster.local
         port: 6379
--- a/deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
+++ b/deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
@@ -29,7 +29,9 @@ data:
     worker:
         vmRender:
             {{- $vmRender := (.Values.worker).vmRender | default dict }}
-            launcherImage: {{ $vmRender.launcherImage | default "quay.io/kubevirt/virt-launcher:v1.8.4" | quote }}
+            {{- if $vmRender.launcherImage }}
+            launcherImage: {{ $vmRender.launcherImage | quote }}
+            {{- end }}
             passtWorkarounds: {{ if kindIs "bool" $vmRender.passtWorkarounds }}{{ $vmRender.passtWorkarounds }}{{ else }}true{{ end }}
     kv:
         hostname: flightctl-kv.{{ default .Release.Namespace .Values.global.internalNamespace }}.svc.cluster.local
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -28,6 +28,7 @@ type Config struct {
 	RemoteAccessService    *RemoteAccessServiceConfig `json:"remoteAccessService,omitempty"`
 	ImageBuilderService    *ImageBuilderServiceConfig `json:"imageBuilderService,omitempty"`
 	ImageBuilderWorker     *imageBuilderWorkerConfig  `json:"imageBuilderWorker,omitempty"`
+	Worker                 *workerConfig              `json:"worker,omitempty"`
 	KV                     *kvConfig                  `json:"kv,omitempty"`
 	Alertmanager           *alertmanagerConfig        `json:"alertmanager,omitempty"`
 	Auth                   *authConfig                `json:"auth,omitempty"`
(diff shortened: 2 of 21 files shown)


deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
```diff
@@ -26,6 +26,11 @@ data:
         {{- end }}
         {{- end }}
     service: {}
+    worker:
+        vmRender:
+            {{- $vmRender := (.Values.worker).vmRender | default dict }}
+            launcherImage: {{ $vmRender.launcherImage | default "quay.io/kubevirt/virt-launcher:v1.8.4" | quote }}
```
kkyrazis: this default launcher image is specified in a lot of locations. `flightctl-worker-config.yaml`, `flightctl/values.yaml`,`values.yaml.gotmpl`, and finally in `config.go`. Do we need it in all locations or could we just use config.go's as the true default.

internal/config/config.go
```diff
@@ -404,6 +405,63 @@ func (c *imageBuilderWorkerConfig) EffectiveSyftSkipTLSVerify() bool {
 	return c != nil && c.ServiceImages != nil && c.ServiceImages.Syft != nil && c.ServiceImages.Syft.SkipTLSVerify
 }
 
+const DefaultVirtLauncherImage = "quay.io/kubevirt/virt-launcher:v1.8.4"
```
kkyrazis, line 408: has this image been included in the list of artifacts that need to be sourced for airgapped installs?

also, is there downstream version that we will be using, or are we using the quay reference?

deploy/helm/flightctl/templates/worker/flightctl-worker-config.yaml
asafbennatan: Good catch — fixed to match the `imageBuilderWorker.serviceImages` pattern.

`config.go` (`DefaultVirtLauncherImage` / `EffectiveVmLauncherImage`) is now the only source of the default. Helm/Podman leave `launcherImage` empty/unset, and the worker config template only emits the field when an override is provided.

internal/config/config.go
asafbennatan: Addressed in the latest commit:

- **Air-gap:** virt-launcher is a device-pulled image (not a control-plane service image), so it is not part of the `flightctl-mirror-images` set. Documented under [Configuring VM application rendering — Air-gapped environments](https://github.com/flightctl/flightctl/blob/fix-passt-workaround/docs/user/installing/configuring-vm-render.md#air-gapped-environments): mirror `quay.io/kubevirt/virt-launcher:v1.8.4` to a registry devices can reach, then set `worker.vmRender.launcherImage` to that reference. Also linked from the air-gapped installation reference page.
- **Downstream:** there is no separate RHEM/product virt-launcher image in the Flight Control packaging set today; we use the upstream `quay.io/kubevirt/virt-launcher` reference (overridable).

internal/tasks/device_render_vm.go
```diff
@@ -66,13 +89,21 @@ func truncateStderr(s string) string {
...
+// may be an absolute path or a bare binary name resolved via PATH. Exported so
+// integration tests can point the converter at a binary extracted to a
+// temporary directory.
+func NewVmConverter(binaryPath string, opts VmRenderOptions) VmConverterFn {
+	if opts.LauncherImage == "" {
+		opts.LauncherImage = config.DefaultVirtLauncherImage
+	}
+	args := []string{
+		"--launcher-image=" + opts.LauncherImage,
+		"--passt-workarounds=" + strconv.FormatBool(opts.PasstWorkarounds),
```
galel12, line 103: These flags are the whole point of the config, but nothing asserts they actually get passed. Unit tests stub the converter, and the vmrender integration suite uses `NewDeviceRenderLogic(..., nil /* cfg */, ...).WithVmConverter(NewVmConverter(..., DefaultVmRenderOptions()))`, so `vmRenderOptionsFromConfig / Effective*` config wiring never runs either.

Worth a tiny test with a fake binary that records argv (and/or pointing integration at a non-nil cfg with Worker.VmRender set).

internal/tasks/device_render_vm_test.go
```diff
@@ -354,13 +355,19 @@ func TestRenderVmApplication_CachePopulatedOnMiss(t *testing.T) {
...
-	_, err = renderVmApplication(ctx, vmApp, converter, kv)
+	_, err = renderVmApplication(ctx, vmApp, converter, DefaultVmRenderOptions(), kv)
 	require.NoError(t, err)
 	assert.Equal(t, 1, callCount, "converter must not be called again on the second (cache hit) call")
+
+	opts := DefaultVmRenderOptions()
+	opts.PasstWorkarounds = false
+	_, err = renderVmApplication(ctx, vmApp, converter, opts, kv)
+	require.NoError(t, err)
+	assert.Equal(t, 2, callCount, "converter should run again when passtWorkarounds changes")
```
galel12, line 375: Nice coverage for passt flipping. Same thing for LauncherImage? Change the image, expect callCount to bump again. Otherwise a regression that drops image from the cache key would still pass.

galel12 (commented):
Looks solid overall (cache key + config defaults especially). Two test gaps:

internal/tasks/device_render_vm.go
asafbennatan: Good call — added in `50a69c0d3`:

- `TestNewVmConverter_PassesRenderFlags` / `TestNewVmConverter_EmptyLauncherImageUsesDefault`: fake `vm-to-quadlet` binary records argv for `--launcher-image` and `--passt-workarounds` (including the empty-image → default path).
- `TestVmRenderOptionsFromConfig`: `NewDeviceRenderLogic` with a non-nil `Worker.VmRender` cfg asserts the Effective* values land in `vmRenderOptions`.

internal/tasks/device_render_vm_test.go
asafbennatan: Done — same cache test now flips `LauncherImage` after the passt case and expects `callCount` to bump again (`50a69c0d3`).

galel12 (approved):
LGTM

galel12 (approved):

PR #3310, opened 2026-07-29
EDM-4758: OS mismatch OutOfDate override for package-mode devices

galel12:
## EDM-4758: OS mismatch OutOfDate override for package-mode devices

**Jira:** https://redhat.atlassian.net/browse/EDM-4758
**Story type:** [DEV]

### Summary
Replace the `status.os.image != ""` guard in the server-side OS mismatch override with a `status.capabilities.osMode` presence check. Package-mode devices permanently report empty `status.os.image`, so the previous guard never fired; with capabilities present they are correctly marked `OutOfDate` when `spec.os.image` is set. Legacy devices without capabilities retain the skip behavior.

### Changes
- **`internal/service/common/device.go`:** Gate the OS mismatch `OutOfDate` override on `Capabilities != nil && OsMode != nil` instead of non-empty `status.os.image`.
- **`internal/service/common/device_test.go`:** Extend `TestUpdateServerSideDeviceUpdatedStatus_OsImageMismatch` with capability-aware table cases (image-mode, package-mode, legacy, nil osMode, no-spec-image).

### Testing
- **Unit tests:** 8 table-driven cases covering AC-1–AC-4
- **Integration tests:** Full suite passed (`make integration-test`)
- **Coverage:** New override paths fully exercised by unit tests

### Acceptance Criteria
- [ ] AC-1: When `status.capabilities.osMode` is present and `status.os.image != spec.os.image`, device marked `OutOfDate` (including package-mode with empty `status.os.image`)
- [ ] AC-2: Legacy devices without capabilities skip the override
- [ ] AC-3: Package-mode device with no `spec.os.image` is not flagged OutOfDate for OS mismatch
- [ ] AC-4: Image-mode OS mismatch continues to work as today

### Note
This branch is stacked on EDM-4759; the PR against `main` also includes those agent commits until EDM-4759 merges.

### DIFF
--- a/internal/service/common/device.go
+++ b/internal/service/common/device.go
@@ -252,9 +252,11 @@ func updateServerSideDeviceUpdatedStatus(device *domain.Device, ctx context.Cont
 	}
 
 	// Override UpToDate if the actual booted OS image doesn't match the desired spec.
+	// Requires capabilities.osMode to be reported; legacy devices without capabilities skip this check.
 	if device.Status.Updated.Status == domain.DeviceUpdatedStatusUpToDate &&
 		device.Spec != nil && device.Spec.Os != nil && device.Spec.Os.Image != "" &&
-		device.Status.Os.Image != "" && device.Status.Os.Image != device.Spec.Os.Image {
+		device.Status.Capabilities != nil && device.Status.Capabilities.OsMode != nil &&
+		device.Status.Os.Image != device.Spec.Os.Image {
 		device.Status.Updated.Status = domain.DeviceUpdatedStatusOutOfDate
 		device.Status.Updated.Info = lo.ToPtr(fmt.Sprintf("Device OS image mismatch: running %q, expected %q.", device.Status.Os.Image, device.Spec.Os.Image))
 	}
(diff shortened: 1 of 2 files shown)


eldar101 (approved):

internal/service/common/device.go
```diff
@@ -252,9 +252,11 @@ func updateServerSideDeviceUpdatedStatus(device *domain.Device, ctx context.Cont
 	}
 
 	// Override UpToDate if the actual booted OS image doesn't match the desired spec.
+	// Requires capabilities.osMode to be reported; legacy devices without capabilities skip this check.
 	if device.Status.Updated.Status == domain.DeviceUpdatedStatusUpToDate &&
```
asafbennatan, line 256: @galel12 can we review if this check is even required ?

asafbennatan: AI mentions that this is done due to greenboot reporting the os post desire spec apply
seems a bit wrong IMO - but we would not tackle this now

asafbennatan (approved):
LGTM , we should look into reported version behavior and os status , without understanding historical context i would assume the version should be reported by the agent only POST os upgrade once it has actually been applied

PR #3313, opened 2026-07-29
EDM-4851:  Update imagebuild e2e test to use a registry with basic auth

amalykhi:
## Summary

- Update the imagebuild e2e workflow test to push the destination OCI image to an authenticated registry (nginx + HTTP Basic Auth at port 5002) instead of the unauthenticated one.
- Extend the imagebuild e2e harness with `*WithCredentials` variants for image resolution, referrer fetching, and SBOM validation; original unauthenticated methods delegate to them with empty credentials (no regressions).
- Propagate registry credentials through the imagebuilder worker's podman build and push flow via a shared auth file (`podman login --authfile`, `REGISTRY_AUTH_FILE` env var).
- Move the worker container auth file from `/build/auth.json` (the podman build context) to `/tmp/auth.json` to keep credentials isolated from the build context.
- Fix nginx `$host` → `$http_host` in the e2e-registry-auth proxy so that OCI blob-upload redirect `Location` URLs include the port, preventing connection refused on port 443.

## Release target

- [x] release-1.3

Target release: release-1.3

### DIFF
--- a/test/e2e/imagebuild/imagebuild_negative_api_test.go
+++ b/test/e2e/imagebuild/imagebuild_negative_api_test.go
@@ -46,12 +46,14 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 
 		It("should fail ImageBuild after creation when destination push cannot authenticate", Label("88399", "imagebuild", "slow"), func() {
 			testID := workerHarness.GetTestIDFromContext()
-			src := createQuaySourceRepository(workerHarness, testID, &negCleanup)
-			dst := createAuthenticatedRegistryDestRepository(workerHarness, testID, &negCleanup, "wrong-password-"+testID)
+			src, err := createUnauthenticatedSourceRepository(workerHarness, testID, &negCleanup)
+			Expect(err).ToNot(HaveOccurred())
+			dst, err := createAuthenticatedRegistryDestRepository(workerHarness, testID, &negCleanup, "wrong-password-"+testID)
+			Expect(err).ToNot(HaveOccurred())
 			ib := negImageBuildName(testID)
 
 			spec := e2e.NewImageBuildSpec(src, sourceImageName, sourceImageTag, dst, destImageName, testID, imagebuilderapi.BindingTypeLate)
-			_, err := workerHarness.CreateImageBuild(ib, spec)
+			_, err = workerHarness.CreateImageBuild(ib, spec)
 			Expect(err).ToNot(HaveOccurred())
 			negCleanup.imageBuild = ib
 
@@ -77,11 +79,12 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 
 		It("should reject ImageExport download before export reaches Completed", Label("88401", "imagebuild"), func() {
 			testID := workerHarness.GetTestIDFromContext()
-			src, dst := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			src, dst, err := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			Expect(err).ToNot(HaveOccurred())
 			ib, ie := negImageBuildName(testID), negImageExportName(testID)
 
 			spec := e2e.NewImageBuildSpec(src, sourceImageName, sourceImageTag, dst, destImageName, testID, imagebuilderapi.BindingTypeLate)
-			_, err := workerHarness.CreateImageBuild(ib, spec)
+			_, err = workerHarness.CreateImageBuild(ib, spec)
 			Expect(err).ToNot(HaveOccurred())
 			negCleanup.imageBuild = ib
 
@@ -104,11 +107,12 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 
 		It("should reject cancel when ImageBuild is Failed", Label("88402", "imagebuild"), func() {
 			testID := workerHarness.GetTestIDFromContext()
-			src, dst := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			src, dst, err := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			Expect(err).ToNot(HaveOccurred())
 			ib := negImageBuildName(testID)
 
 			spec := e2e.NewImageBuildSpec(src, "this-image-does-not-exist/invalid-image", "nonexistent-tag", dst, destImageName, testID, imagebuilderapi.BindingTypeLate)
-			_, err := workerHarness.CreateImageBuild(ib, spec)
+			_, err = workerHarness.CreateImageBuild(ib, spec)
 			Expect(err).ToNot(HaveOccurred())
 			negCleanup.imageBuild = ib
 
@@ -122,11 +126,12 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 
 		It("should reject cancel when ImageBuild is Canceled", Label("imagebuild"), func() {
 			testID := workerHarness.GetTestIDFromContext()
-			src, dst := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			src, dst, err := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			Expect(err).ToNot(HaveOccurred())
 			ib := negImageBuildName(testID)
 
 			spec := e2e.NewImageBuildSpec(src, sourceImageName, sourceImageTag, dst, destImageName, testID, imagebuilderapi.BindingTypeLate)
-			_, err := workerHarness.CreateImageBuild(ib, spec)
+			_, err = workerHarness.CreateImageBuild(ib, spec)
 			Expect(err).ToNot(HaveOccurred())
 			negCleanup.imageBuild = ib
 
@@ -142,11 +147,12 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 
 		It("should reject cancel when ImageBuild is Completed", Label("88404", "imagebuild", "slow"), func() {
 			testID := workerHarness.GetTestIDFromContext()
-			src, dst := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			src, dst, err := createStandardImageBuildRepositories(workerHarness, testID, &negCleanup)
+			Expect(err).ToNot(HaveOccurred())
 			ib := negImageBuildName(testID)
 
 			spec := e2e.NewImageBuildSpec(src, sourceImageName, sourceImageTag, dst, destImageName, testID, imagebuilderapi.BindingTypeLate)
-			_, err := workerHarness.CreateImageBuild(ib, spec)
+			_, err = workerHarness.CreateImageBuild(ib, spec)
 			Expect(err).ToNot(HaveOccurred())
 			negCleanup.imageBuild = ib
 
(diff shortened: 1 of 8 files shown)


internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -39,6 +39,12 @@ const (
 	// during image builds. A unique subdirectory is created per job and removed when
 	// the job completes. Any leftovers from a crash are swept on worker startup.
 	buildStorageBaseDir = "/var/tmp/flightctl-builds"
+
+	// containerAuthFile is the path to the container auth file used for registry login.
+	// It lives inside the mounted /build directory so the same path is visible to both
+	// the host (via podmanWorker.TmpDir) and all subsequent podman exec calls, regardless
+	// of XDG_RUNTIME_DIR or HOME being set consistently across exec invocations.
+	containerAuthFile = "/build/auth.json"
```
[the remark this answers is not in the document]
amalykhi: Thanks for the analysis. After checking the Containerfile template (`internal/imagebuilder_worker/tasks/templates/Containerfile.tmpl`), there is no `COPY .` or `ADD .` instruction — the template uses only two explicit named `COPY` statements (`agent-config.yaml` and `user-publickey.txt`). Because the build context is never bulk-copied, `auth.json` is not reachable from any `COPY` or `ADD` instruction and cannot be baked into the image.

Moving the auth file outside `/build` would require adjusting the container mount configuration, the `TmpDir`→container path mapping, and all `podman exec` invocations — meaningful complexity to guard against a theoretical future `COPY .` that would itself be a reviewable change. We'll leave the path as-is for now and note this concern for any future Containerfile changes.

test/e2e/imagebuild/imagebuild_workflow_test.go
```diff
@@ -382,30 +387,39 @@ func isLocalSourceRegistry() bool {
...
 func getEnrollmentIDFromAgentLogs(libvirtVM vm.TestVMInterface) string {
 	if libvirtVM == nil {
 		GinkgoWriter.Printf("getEnrollmentIDFromAgentLogs: VM is nil\n")
 		return ""
 	}
 	stdout, err := libvirtVM.RunSSH([]string{"sudo", "journalctl", "-u", agentServiceName, "--no-pager"}, nil)
 	if err != nil {
-		GinkgoWriter.Printf("getEnrollmentIDFromAgentLogs: %v\n", err)
+		GinkgoWriter.Printf("getEnrollmentIDFromAgentLogs: SSH error: %v\n", err)
 		return ""
```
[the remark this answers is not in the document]
amalykhi: The empty-string-on-SSH-error behavior is intentional. During VM first boot, SSH is not yet available — `RunSSH` will fail repeatedly for the first 30–60 seconds until `sshd` starts. Converting those transient failures to `""` lets `Eventually` keep retrying (the normal polling contract). Returning an error instead would cause the `Eventually` block to fail immediately on the first SSH connection refused, before the VM has had a chance to finish booting.

The real failure signal is the `ShouldNot(BeEmpty())` timeout after `vmBootTimeout` — if SSH never stabilizes, that fires with a clear timeout message. SSH errors during the poll are already logged via `GinkgoWriter.Printf` so they're visible in test output without aborting the retry loop.

test/e2e/imagebuild/imagebuild_workflow_test.go
```diff
@@ -382,30 +389,39 @@ func isLocalSourceRegistry() bool {
...
 	if libvirtVM == nil {
 		GinkgoWriter.Printf("getAgentServiceStatus: VM is nil\n")
-		return "", nil
+		return "", fmt.Errorf("VM is nil")
 	}
 	stdout, err := libvirtVM.RunSSH([]string{"sudo", "systemctl", "is-active", agentServiceName}, nil)
 	if err != nil {
-		GinkgoWriter.Printf("getAgentServiceStatus: %v\n", err)
+		GinkgoWriter.Printf("getAgentServiceStatus: SSH error: %v\n", err)
 		return "", err
```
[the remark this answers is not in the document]
amalykhi: Good call. Fixed — `getAgentServiceStatus` now returns `fmt.Errorf("get agent service status via SSH: %w", err)` so the `Eventually` timeout message will identify the probe that failed.

eldar101:
Can we keep this auth file outside `/build`? Since `/build` is the podman build context, the registry auth file is now sitting next to files that can be copied into the image. It would be safer to mount/write it somewhere that is not part of the build context.

eldar101:
Please fix all warnings and failed checks and ping me.

amalykhi:
Good catch. Moved — `containerAuthFile` is now `/tmp/auth.json` inside the worker container. `/tmp` is container-internal and not mounted from the host, so credentials are fully isolated from the `/build` context directory. The current Containerfile template has no `COPY .` or `ADD .`, so there was no active exposure, but keeping the file adjacent to the build context is poor hygiene regardless. Fixed in f851ebe9.

amalykhi:
Done — all three items addressed:

1. **Ai-Attribution** — replaced `Co-Authored-By: Claude Sonnet 4.6` with `Assisted-by: Claude Sonnet 4.6` across all commits (force-pushed).
2. **Description check** — added PR description with Summary and Release target sections. Could you add the `release-1.3` label via the GitHub UI?
3. **Docstring Coverage** — added missing docstrings to the auxiliary registry helper functions in `test/e2e/infra/auxiliary/registry.go`.

Auth file move (your other comment) is also in — `containerAuthFile` is now `/tmp/auth.json` (commit 356bb60d).

Ready for another look when you have a chance. 🙏

amalykhi:
The tests passed in CI
image

Samudelacruz77 (commented):
LGTM

test/e2e/imagebuild/imagebuild_negative_api_test.go
```diff
@@ -231,42 +239,70 @@ func negImageExportName(testID string) string {
 	return fmt.Sprintf("ie-%s", testID)
 }
 
-func createStandardImageBuildRepositories(h *e2e.Harness, testID string, c *negAPIResourceCleanup) (sourceRepoName, destRepoName string) {
-	sourceRepoName = createQuaySourceRepository(h, testID, c)
+// createStandardImageBuildRepositories creates an unauthenticated source (quay.io or OCP mirror)
+// and an unauthenticated destination OCI repository, registering both for cleanup.
+// Returns an error if either repository cannot be created.
+func createStandardImageBuildRepositories(h *e2e.Harness, testID string, c *negAPIResourceCleanup) (sourceRepoName, destRepoName string, err error) {
```
itzikb-redhat, line 245: Maybe the name of the function should also note that it's unauthenticated

test/e2e/imagebuild/imagebuild_workflow_test.go
```diff
@@ -56,11 +62,10 @@ var sourceRegistryOnce sync.Once
 var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 	Context("ImageBuild end-to-end workflow", func() {
 		It("should verify basic image build process - build, export, download, use in an agent", Label("87335", "88406", "imagebuild", "slow"), func() {
-			Expect(workerHarness).ToNot(BeNil())
-			Expect(workerHarness.ImageBuilderClient).ToNot(BeNil(), "ImageBuilderClient must be available")
-
 			testID := workerHarness.GetTestIDFromContext()
-			registryAddress := auxSvcs.Registry.Host + ":" + auxSvcs.Registry.Port
+			registryAddress := auxSvcs.Registry.Authenticated.HostPort
```
itzikb-redhat, line 66: Should we change the name of the test or add a comment noting the change?

itzikb-redhat (commented):
Thanks! two small questions, otherwise lgtm

test/e2e/imagebuild/imagebuild_negative_api_test.go
amalykhi: Good point — renamed to `createUnauthenticatedSourceRepository` to make the auth stance explicit and parallel `createAuthenticatedRegistryDestRepository`. Fixed in b49bb3af.

test/e2e/imagebuild/imagebuild_workflow_test.go
amalykhi: Good call — updated the test description to include "(Basic-auth destination registry)" so it's visible in CI output. Fixed in b49bb3af.

amalykhi:
@coderabbitai review

eldar101 (approved):

eldar101 (approved):

eldar101 (approved):

asafbennatan (commented):
See comments below

test/e2e/imagebuild/imagebuild_workflow_test.go
```diff
@@ -82,8 +87,8 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 				lo.ToPtr(api.Https), lo.ToPtr(api.Read), isLocalSourceRegistry(), nil)
 			Expect(err).ToNot(HaveOccurred())
 
-			_, err = resources.CreateOCIRepository(workerHarness, destRepoName, registryAddress,
-				lo.ToPtr(api.Https), lo.ToPtr(api.ReadWrite), true, nil)
+			err = applyOCIRepositoryWithDockerAuth(workerHarness, destRepoName, registryAddress,
+				lo.ToPtr(api.Https), lo.ToPtr(api.ReadWrite), true, authUsername, authPassword)
```
asafbennatan, line 91: This passes the real registry password into `applyOCIRepositoryWithDockerAuth`, which marshals it into the `Repository` YAML and hands it to `CLIWithStdin` → `SHWithStdin(..., redactStdin=false, ...)` -- that call logs the full stdin (password included) via `logrus.Infof` in `harness.go`. Until now this helper was only ever called with a deliberately-wrong password in the negative-auth test; this is the first call site to route the actual working credential through it, in the main happy-path workflow test. CodeRabbit's own `No-Sensitive-Data-In-Logs` pre-merge check flagged this exact logging call, and it doesn't look like it's been redacted. Since it's just the throwaway e2e registry's own test credential the practical exposure is limited, but should `applyOCIRepositoryWithDockerAuth` pass `redactStdin=true` now that a real credential flows through it, to keep CI logs clean and satisfy the project's own check?

test/e2e/imagebuild/imagebuild_negative_api_test.go
```diff
@@ -46,12 +46,14 @@ var _ = Describe("ImageBuild", Label("imagebuild"), func() {
 
 		It("should fail ImageBuild after creation when destination push cannot authenticate", Label("88399", "imagebuild", "slow"), func() {
 			testID := workerHarness.GetTestIDFromContext()
-			src := createQuaySourceRepository(workerHarness, testID, &negCleanup)
-			dst := createAuthenticatedRegistryDestRepository(workerHarness, testID, &negCleanup, "wrong-password-"+testID)
+			src, err := createUnauthenticatedSourceRepository(workerHarness, testID, &negCleanup)
+			Expect(err).ToNot(HaveOccurred())
+			dst, err2 := createAuthenticatedRegistryDestRepository(workerHarness, testID, &negCleanup, "wrong-password-"+testID)
+			Expect(err2).ToNot(HaveOccurred())
```
asafbennatan: Since `err` from the line above is still in scope, could this reuse it (`dst, err := ...`) instead of introducing `err2`? `:=` is still valid here because `dst` is new. It'd match the pattern every other call site in this file already uses.

```suggestion
			dst, err := createAuthenticatedRegistryDestRepository(workerHarness, testID, &negCleanup, "wrong-password-"+testID)
			Expect(err).ToNot(HaveOccurred())
```

amalykhi: Agreed — `:=` is valid since `dst` is new, and it matches every other call site. Fixed in be95b622.

test/e2e/imagebuild/imagebuild_workflow_test.go
amalykhi: Good catch. `CLIWithStdin` passes `redactStdin=false` to `SHWithStdin`, so the YAML (including the password) was logged. I've added `CLIWithStdinRedacted` to the harness and switched `applyOCIRepositoryWithDockerAuth` to call it — the password no longer appears in `logrus.Infof` output. Fixed in be95b622.

eldar101 (approved):

PR #3319, opened 2026-07-30
NO-ISSUE: Default virt-launcher to v1.9.0 and disable passt workarounds

asafbennatan:
## Summary
- Bump the built-in default virt-launcher image from `quay.io/kubevirt/virt-launcher:v1.8.4` to `v1.9.0`
- Default `passtWorkarounds` to `false` across worker config, Helm, and Podman quadlets (v1.9.0 does not need the workaround)
- Update docs and unit tests to match the new defaults

## Test plan
- [x] `go test ./internal/tasks/ -run 'TestVmRenderOptionsFromConfig|TestNewVmConverter|TestRenderVmApplication_CachePopulatedOnMiss'`
- [ ] Confirm a fresh deploy renders VM apps with virt-launcher `v1.9.0` and `--passt-workarounds=false`
- [ ] Confirm explicitly setting `passtWorkarounds: true` still enables the flag for older launcher images

### DIFF
--- a/docs/user/installing/configuring-vm-render.md
+++ b/docs/user/installing/configuring-vm-render.md
@@ -6,10 +6,10 @@ The Flight Control worker converts `VmApplication` manifests into Quadlet units 
 
 | Setting | Default |
 | ------- | ------- |
-| `launcherImage` | `quay.io/kubevirt/virt-launcher:v1.8.4` (built into the worker; leave config unset or empty to use it) |
-| `passtWorkarounds` | `true` |
+| `launcherImage` | `quay.io/kubevirt/virt-launcher:v1.9.0` (built into the worker; leave config unset or empty to use it) |
+| `passtWorkarounds` | `false` |
 
-`passtWorkarounds` enables startup patches for known networking issues in older virt-launcher passt builds (for example guest network instability and related passt failures). Keep this enabled unless your virt-launcher image already includes a fixed passt build.
+`passtWorkarounds` enables startup patches for known networking issues in older virt-launcher passt builds (for example guest network instability and related passt failures). The default virt-launcher image does not need this. Enable it only when you override `launcherImage` to an older image that still requires the workaround.
 
 ## Prerequisites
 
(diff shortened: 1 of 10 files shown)


galel12 (commented):
(superseded — comments moved; see review below)

docs/user/installing/configuring-vm-render.md
```diff
@@ -6,10 +6,10 @@ The Flight Control worker converts `VmApplication` manifests into Quadlet units
 
 | Setting | Default |
 | ------- | ------- |
-| `launcherImage` | `quay.io/kubevirt/virt-launcher:v1.8.4` (built into the worker; leave config unset or empty to use it) |
-| `passtWorkarounds` | `true` |
+| `launcherImage` | `quay.io/kubevirt/virt-launcher:v1.9.0` (built into the worker; leave config unset or empty to use it) |
+| `passtWorkarounds` | `false` |
 
-`passtWorkarounds` enables startup patches for known networking issues in older virt-launcher passt builds (for example guest network instability and related passt failures). Keep this enabled unless your virt-launcher image already includes a fixed passt build.
+`passtWorkarounds` enables startup patches for known networking issues in older virt-launcher passt builds (for example guest network instability and related passt failures). The default virt-launcher image does not need this. Enable it only when you override `launcherImage` to an older image that still requires the workaround.
```
galel12, line 12: nit: worth an explicit upgrade callout here (or in the PR/release notes).

Anyone who pins an older launcherImage (e.g. mirrored v1.8.4) but leaves passtWorkarounds unset used to inherit true, and now inherits false. This paragraph already says enable it for older images; calling out that upgrade case would help so nobody silently loses the workaround on re-render.

Existing devices keep the previously rendered launcher/passt until re-rendered (covered later in this doc).

galel12 (commented):
LGTM on the defaults flip. Go/Helm/Podman/docs look consistent. Neither note below is blocking.

nit (test stub, not in the diff): `fakeQuadletFiles` in `internal/tasks/device_render_vm_test.go` still hardcodes `v1.8.4` in the stub Image line. Harmless fake converter output, but worth updating if you touch that file again.

docs/user/installing/configuring-vm-render.md
asafbennatan: there is none - the kubevirt dependency has been introduced in this upcoming release

galel12 (approved):
LGTM

PR #3320, opened 2026-08-01
EDM-4779: Add encryption at rest user documentation

asafbss:
Add configuring-encryption.md covering protected fields, key management, key rotation, migration, observability, troubleshooting, and security. Update events and metrics reference docs with encryption sections. Add encryption doc to the user documentation table of contents.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -4,6 +4,12 @@
 # one word per line, to define a file override use ' - filename'
 # where filename is relative to this configuration file
 AAP
+AES
+ciphertext
+decryptable
+GCM
+nonces
+plaintext
 ACM
 Alertmanager
 amd64
@@ -596,6 +602,14 @@ quadlets
 Sanitization
 SNI
 VPN
+- docs/user/installing/configuring-encryption.md
+256-bit
+256-GCM
+32-byte
+AES-256-GCM
+base64-encoded
+plaintext-to-encrypted
+pre-install
 - docs/user/references/events.md
 lifecycles
 Rollouts
--- a/.spelling
+++ b/.spelling
@@ -4,13 +4,8 @@
 # one word per line, to define a file override use ' - filename'
 # where filename is relative to this configuration file
 AAP
-AES
-ciphertext
-decryptable
-GCM
-nonces
-plaintext
 ACM
+AES
 Alertmanager
 amd64
 API
@@ -27,6 +22,7 @@ CentOS
 cgroup
 cgroups
 checksum
+ciphertext
 cleanup
 CLI
 cpu
@@ -38,6 +34,7 @@ Containerfile
 CRD
 CSR
 darwin
+decryptable
 DeviceTemplate
 df
 e.g.
@@ -56,6 +53,7 @@ firewalld
 fleetless
 flightctl
 frontend
+GCM
 GitLab
 GitOps
 greenboot
@@ -87,6 +85,7 @@ microservices
 MicroShift
 mTLS
 namespace
+nonces
 non-root
 NAT
 OCI
@@ -101,6 +100,7 @@ ostree
 OTA
 passt
 passtWorkarounds
+plaintext
 pluggable
 Podman
 PostgreSQL
--- a/docs/user/installing/configuring-encryption.md
+++ b/docs/user/installing/configuring-encryption.md
@@ -1,0 +1,386 @@
+# Configuring encryption at rest
+
+## Overview
+
+Flight Control encrypts sensitive control plane data before storing it in PostgreSQL. This protects repository credentials, authentication provider secrets, and rendered device configuration data from unauthorized access if database contents, database backups, or storage media are exposed.
+
+Encryption is enabled by default. During the first deployment, Flight Control generates an AES-256-GCM encryption key automatically. Default deployments do not require a manually provisioned key.
+
+A custom key can be added later and configured as the active key. After rotation, new sensitive data is encrypted with the active key, and existing encrypted data is re-encrypted with the active key by the background migration process. The previous key can be retired after migration confirms that live data no longer depends on it.
+
+Sensitive fields remain encrypted in PostgreSQL and in database backups.
+
+Key points:
+
+* **Enabled by default:** Encryption is active on new deployments. Services fail to start if the encryption key is missing, malformed, or invalid.
+* **Automatic key generation:** The initial key is generated during deployment.
+* **Custom key support:** A custom key can be added and made active through key rotation.
+* **Key migration:** After rotation, existing encrypted data is re-encrypted with the active key by the background migration process.
+* **Safe key retirement:** Old keys can be retired only after migration confirms that live data no longer depends on them.
+* **Backup critical:** Losing the encryption key makes encrypted data permanently unreadable. The key is included in the standard backup and restore process, so backup archives must be protected accordingly.
+* **Selective encryption:** Only documented sensitive fields are encrypted. Do not store sensitive information in fields that are not documented as protected.
+
+## What is protected
+
+### Protected fields
+
+The following fields are encrypted before being stored in PostgreSQL:
+
+| Resource | Field | Description |
+|----------|-------|-------------|
+| Repository | `spec.httpConfig.password` | HTTP basic authentication password |
+| Repository | `spec.httpConfig.token` | HTTP bearer token |
+| Repository | `spec.httpConfig.tls.key` | Client TLS private key |
+| Repository | `spec.httpConfig.tls.crt` | Client TLS certificate |
+| Repository | `spec.sshConfig.sshPrivateKey` | SSH private key |
+| Repository | `spec.sshConfig.privateKeyPassphrase` | SSH key passphrase |
+| Repository | `spec.ociAuth.password` | OCI registry password |
+| AuthProvider | `spec.clientSecret` | Client secret for OIDC, OAuth2, OpenShift, and AAP authentication providers |
+| Device | `renderedConfig` | Rendered device configuration |
+| Device | `renderedApplications` | Rendered device applications |
+
+### Unprotected fields
+
+Fields that are not listed in the protected-fields table are not encrypted by Flight Control application-level encryption at rest.
+
+Examples include:
+
+* Resource names, labels, annotations, and metadata
+* Repository URLs and repository types
+* Authentication provider issuer URLs and client IDs
+* Device status, system information, and conditions
+* Fleet selectors, templates, and rollout policies
+
+These fields remain visible in PostgreSQL and in database backups.
+
+> [!WARNING]
+> Do not store sensitive information such as passwords, tokens, private keys, or client secrets in fields that are not documented as protected. Only the fields listed in the protected-fields table are encrypted by Flight Control encryption at rest.
+
+## Encryption key management
+
+### Default generated key
+
+During the first deployment, Flight Control generates an AES-256-GCM encryption key automatically. No manual key provisioning is required for initial setup.
+
+The generated key uses the identifier `default`. Encrypted values include this identifier in their stored format:
+
+```text
+enc:v1:default:<base64 payload>
+```
+
+On upgrades, the existing key is preserved automatically to prevent accidental key loss. See [Upgrading existing deployments](#upgrading-existing-deployments).
+
+### Key locations
+
+The encryption key is loaded from a read-only mounted file inside each service container:
+
+```text
+/root/.flightctl/encryption/key
+```
+
+The source of this file depends on the deployment method.
+
+#### OpenShift / Kubernetes
+
+The key is stored in a Kubernetes Secret named `flightctl-encryption-key` in the deployment namespace. A Helm pre-install/pre-upgrade hook Job generates the key on first deployment and preserves the existing key on upgrades.
+
+The Secret is mounted read-only at:
+
+```text
+/root/.flightctl/encryption/
+```
+
+in Flight Control service pods.
+
+To verify that the Secret exists:
+
+```bash
+kubectl get secret flightctl-encryption-key -n <namespace>
+```
+
+To verify that the key is mounted inside a service pod:
+
+```bash
+kubectl exec deploy/flightctl-api -n <namespace> -- ls -la /root/.flightctl/encryption/key
+```
+
+For key rotation, additional key files are added to the same Secret. See [Rotating encryption keys](#rotating-encryption-keys).
+
+#### Quadlet / Podman
+
+The key file is stored on the host filesystem at:
+
+```text
+/etc/flightctl/encryption/key
+```
+
+During first deployment, Flight Control creates the key file if it does not already exist. On subsequent deployments and upgrades, the existing key is validated and preserved.
+
+The host directory is mounted read-only into Flight Control service containers at:
+
+```text
+/root/.flightctl/encryption/
+```
+
+To verify that the key exists on the host:
+
+```bash
+ls -la /etc/flightctl/encryption/key
+```
+
+To verify that the key is mounted inside a service container:
+
+```bash
+podman exec flightctl-api ls -la /root/.flightctl/encryption/key
+```
+
+### Key file format
+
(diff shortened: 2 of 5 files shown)


docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+encryption:
+  activeKeyID: "key-2026-07"
+  keys:
+    - id: "default"
+      file: "key"
+    - id: "key-2026-07"
+      file: "key-2026-07"
+```
+
+Run `helm upgrade` to apply the changes.
```
asafbennatan: **HIGH / Correctness — Helm rotation omits required pod restart**

This Kubernetes/Helm procedure ends with `helm upgrade`, and "Verify the new active key" says "After services restart…", which implies the upgrade restarts pods. It does not.

Encryption config is rendered into service ConfigMaps, and the Deployments have no checksum/hash annotation tying the pod template to that ConfigMap or to `flightctl-encryption-key`. `helm upgrade` updates ConfigMap/Secret in place without rolling pods. Keys are loaded once at process startup (`InitGlobalEncryption`), so running pods keep the old active key until restarted.

The same gap applies to "Retire old keys" step 3 ("Restart Flight Control services"), which is unqualified for the Kubernetes path. The Quadlet path correctly uses `systemctl restart flightctl.target`.

**Suggestion:** Add an explicit restart after `helm upgrade`, e.g. `kubectl rollout restart deployment -n <namespace> …` for every Deployment that mounts `flightctl-encryption-key`, before verifying the active key.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+
+Flight Control generates the initial encryption key automatically. To use your own key after deployment, rotate from the generated key to a custom key. See [Rotating encryption keys](#rotating-encryption-keys).
+
+Rotation adds the custom key alongside the existing key, marks the custom key as active for new encryptions, and migrates existing encrypted data to the active key. The original key remains configured for decryption until migration completes.
+
+## Backup and restore requirements
+
+Encryption keys are required to read encrypted data stored in PostgreSQL. Losing the required encryption keys makes the affected encrypted fields permanently unreadable.
+
+Encryption keys are included automatically in Flight Control backup archives created by `flightctl-backup`. Restoring from a backup archive with `flightctl-restore` restores the encryption keys alongside the database and other server state.
```
asafbennatan: **HIGH / Security — Backup of encryption keys is fail-open**

This section states keys are included automatically and earlier frames the key as backup-critical. In code, `BackupEncryptionKeys` (Kubernetes and Podman deployers) logs a warning and returns `nil` when the Secret/directory is missing — the overall backup still succeeds.

An operator relying on the documented guarantee has no indication a backup could silently omit the key.

**Suggestion:** Note that key backup is best-effort: if the Secret/directory cannot be found, the backup log emits a warning and the archive is created without key material. Recommend checking backup logs (or the archive `encryption/` directory) after every backup before relying on it for DR.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+```
+
+### Key file format
+
+The key file contains a base64-encoded 32-byte, or 256-bit, random value.
+
+Requirements:
+
+* The base64-decoded value must be exactly 32 bytes.
+* The file must have restricted permissions.
```
asafbennatan: **MEDIUM / Correctness — Permissions requirement is not enforced**

This bullet sits in a "Requirements" list next to checks that are actually enforced (32-byte length, all-zero rejection, missing/malformed keys fail startup). File mode is never validated — `NewV1Strategy` only reads and validates key material.

**Suggestion:** Reword to make clear this is an operational recommendation, e.g. "Restrict the file's permissions on the host or in the Secret (Flight Control does not verify file permissions at startup)."

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+
+To generate a key manually:
+
+```bash
+openssl rand -base64 32
+```
+
+### Providing your own key
+
+Flight Control generates the initial encryption key automatically. To use your own key after deployment, rotate from the generated key to a custom key. See [Rotating encryption keys](#rotating-encryption-keys).
```
asafbennatan: **MEDIUM / Design — Pre-seed path omitted**

This presents rotation as the only way to use a custom key. The Helm/Quadlet key generators are restore-aware and preserve an existing valid 32-byte key, so for a fresh install an operator can pre-create `flightctl-encryption-key` (or place `/etc/flightctl/encryption/key`) before first deploy and avoid rotation/migration entirely.

**Suggestion:** Document pre-seeding for fresh installs as a supported option; keep rotation as the path for existing deployments.

.spelling
```diff
@@ -4,6 +4,12 @@
 # one word per line, to define a file override use ' - filename'
 # where filename is relative to this configuration file
 AAP
+AES
+ciphertext
+decryptable
+GCM
+nonces
+plaintext
```
asafbennatan: **LOW / Conventions — Global dictionary alphabetical order**

The new global entries (`AES`, `ciphertext`, `decryptable`, `GCM`, `nonces`, `plaintext`) are inserted after `AAP` and break the existing alphabetical ordering of the global dictionary (the per-file override section for this doc is fine).

**Suggestion:** Move these six words to their alphabetically correct positions in the global dictionary section.

asafbennatan (changes requested):
see below

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+
+Encryption at rest protects sensitive Flight Control data if PostgreSQL contents, database backups, or storage media are exposed without the encryption keys.
+
+Encryption at rest does not protect data while it is being used by a running Flight Control service. It also does not protect against compromise of an authorized Flight Control service that has access to the encryption key.
+
+Only the documented protected fields are encrypted. Do not store passwords, tokens, private keys, client secrets, or other sensitive values in unprotected fields.
+
+Backup archives contain encryption keys and encrypted data. Protect backup archives with the same level of access control as production secrets.
+
+Do not delete historical encryption keys until migration confirms that no persisted encrypted values depend on them. Historical keys may also be required to restore older backups.
```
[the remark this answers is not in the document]
asafbss: Not adding this. The security section already states that encryption at rest does not protect against compromise of a service that has access to the encryption key, which covers the threat model.

RBAC for Kubernetes Secrets and etcd encryption-at-rest are cluster-level concerns that apply equally to every Secret containing sensitive data (TLS keys, DB credentials, auth client secrets), not just flightctl-encryption-key.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+* `process`
+* `encrypt`
+* `decrypt`
+* `canary-ensure`
+* `canary-validate`
+* `canary-validate-all`
+
+Span attributes include the operation type, strategy, key ID, result, and processing action where applicable.
+
+Traces do not contain plaintext values, ciphertext, encryption keys, nonces, authentication tags, or protected resource contents.
```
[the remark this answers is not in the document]
asafbss: The documented span names and attributes are accurate. The gaps noted are implementation details, not documentation issues:

Early Encrypt/Decrypt failures without spans: These are pre-validation failures (strategy not found, canary ensure error) that happen before the actual crypto operation. They are recorded in metrics. Adding spans for pre-validation guard clauses would add noise without observability value.

canary-ensure and canary-validate-all not recording a result attribute: canary-ensure uses a do-once pattern and exits early on cache hits without entering the span. canary-validate-all is a parent span whose children (canary-validate) each record their own result. Adding a synthetic aggregate result to the parent would duplicate information already visible in the child spans.

These are reasonable implementation choices, not coverage gaps that need to be documented or fixed.

docs/user/installing/configuring-encryption.md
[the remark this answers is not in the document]
asafbss: global.internalNamespace is an advanced Helm option that is not documented in the user-facing docs. The encryption rotation procedure documents the default single-namespace deployment. Users who configure internalNamespace are expected to understand that namespace-scoped resources (Secrets, Deployments) need to be managed in both namespaces — this applies to all Secrets, not just the encryption key.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
+# Configuring encryption at rest
+
+## Overview
+
+Flight Control encrypts sensitive control plane data before storing it in PostgreSQL. This protects repository credentials, authentication provider secrets, and rendered device configuration data from unauthorized access if database contents, database backups, or storage media are exposed.
+
+Encryption is enabled by default. During the first deployment, Flight Control generates an AES-256-GCM encryption key automatically. Default deployments do not require a manually provisioned key.
```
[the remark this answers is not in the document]
asafbss: The doc already covers these points:

The overview states encryption protects data "if database contents, database backups, or storage media are exposed" — this inherently means only when the keys are not co-located with the data.
The "Backup critical" key point states that backup archives must be protected accordingly.
The backup section documents that archives may be created without key material and recommends verifying after each backup.
The WARNING block states that backup archives may contain encryption keys and must be stored on encrypted storage with restricted access.
Making backups fail when the key source is missing is a code behavior change, not a documentation issue. The current warn-and-continue behavior is intentional to support pre-encryption deployments and upgrades where no encryption key exists yet.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,386 @@
...
+
+Before rotating encryption keys:
+
+* Back up the deployment, including the current encryption keys. See [Backup and restore](backup-restore.md).
+* Verify that the current encryption configuration is healthy. See [Active key verification](#active-key-verification).
+* Generate a new 32-byte key:
+
+```bash
+openssl rand -base64 32 > key-2026-07
+```
```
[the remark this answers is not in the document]
asafbss: Fixed

docs/user/references/metrics.md
```diff
@@ -142,6 +142,26 @@ Monitors the automated dependency synchronization subsystem that probes Git, HTT
...
+- `flightctl_encryption_canary_validations_total`: Canary validation attempts and outcomes by strategy, key, and status
+
+**Labels:** Labels vary by metric and include `operation`, `strategy`, `key_id`, `status`, `algorithm`, and `error_type`.
+
+**Notes:**
+
+- Encryption metrics do not require a separate collector configuration. They are available from `flightctl-api` and `flightctl-worker` when metrics are enabled for those services.
+- Canary validation failures that prevent service startup are reported through service logs and service status. Successful canary validations are reported through `flightctl_encryption_canary_validations_total`.
+- For more information about encryption at rest, key rotation, and canary validation, see [Configuring encryption at rest](../installing/configuring-encryption.md).
+
```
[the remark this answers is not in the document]
asafbss: Fixed

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,439 @@
...
+  --from-file=key=<path-to-your-key-file>
+```
+
+On **Quadlet / Podman**, place the key file before starting Flight Control services:
+
+```bash
+sudo mkdir -p /etc/flightctl/encryption
+sudo cp <path-to-your-key-file> /etc/flightctl/encryption/key
+sudo chmod 0600 /etc/flightctl/encryption/key
+```
```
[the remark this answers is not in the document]
asafbss: Fixed, ty

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,437 @@
...
+kubectl rollout restart -n <namespace> \
+  deployment/flightctl-api \
+  deployment/flightctl-worker \
+  deployment/flightctl-periodic \
+  deployment/flightctl-alert-exporter \
+  deployment/flightctl-alertmanager-proxy \
+  deployment/flightctl-remote-access \
+  deployment/flightctl-imagebuilder-api \
+  deployment/flightctl-imagebuilder-worker
+```
```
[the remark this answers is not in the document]
asafbss: The next section ("Verify the new active key") directs the user to check startup logs and the active key metric, which inherently requires pods to be running. If the rollout is still in progress, the old key appears and the user waits. Adding rollout status for all 8 deployments adds significant verbosity for a self-correcting situation — this is standard kubectl rollout restart usage.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,437 @@
...
+
+After services restart, verify that the new key is reported as active. See [Active key verification](#active-key-verification).
+
+### Migrate existing data to the active key
+
+After rotation, the background migration process re-encrypts existing protected data with the active key. No manual action is required to start the migration.
+
+Flight Control emits an `EncryptionMigrationStarted` event when migration begins and an `EncryptionMigrationCompleted` event when migration finishes.
+
+Monitor migration progress using migration events. For more information about events, see [Encryption Events](../references/events.md#encryption-events). Encryption operation metrics can also be used to observe encryption, decryption, and error activity during migration. See [Encryption metrics](#encryption-metrics) in [Observability and validation](#observability-and-validation) for details.
```
asafbennatan, line 299: **MEDIUM / Naming/Clarity — Add CLI examples for migration Started and Completed events**

This section names `EncryptionMigrationStarted` and `EncryptionMigrationCompleted` and says to "Monitor migration progress using migration events", but it never shows how an operator actually queries them. The retirement step below only says "Confirm that `EncryptionMigrationCompleted` was emitted" with no command either.

Please add concrete `flightctl get events` examples for both reasons (same pattern used elsewhere in the user docs), for example:

```bash
flightctl get events --field-selector="reason=EncryptionMigrationStarted"
flightctl get events --field-selector="reason=EncryptionMigrationCompleted"
```

Optionally tighten with the documented involved object:

```bash
flightctl get events --field-selector="reason=EncryptionMigrationCompleted,involvedObject.kind=System,involvedObject.name=encryption"
```

Also reference these commands from step 1 under "Retire old keys" so the retirement gate is actionable, not just a reason name.

docs/user/installing/configuring-encryption.md
```diff
@@ -0,0 +1,437 @@
...
+
+Migration is complete when Flight Control reports that all protected records have been processed and no persisted encrypted values reference the previous key.
+
+### Retire old keys
+
+After migration completes, Flight Control prepares non-active keys for retirement by removing their canaries. The old key can then be removed from the encryption configuration and from the Secret or host filesystem.
+
+To retire an old key:
+
+1. Confirm that `EncryptionMigrationCompleted` was emitted for the migration.
```
asafbennatan, line 309: **MEDIUM — Retirement confirmation should cite the CLI**

"Confirm that `EncryptionMigrationCompleted` was emitted" should point operators at the `flightctl get events` selector (see comment on the migration section), not only the event reason name.

asafbennatan (changes requested):
## Follow-up review

Doc updates from the previous round look good. One remaining docs gap (plus the still-open AI attribution trailer on both commits):

Operators are told to monitor `EncryptionMigrationStarted` / `EncryptionMigrationCompleted` and to confirm completion before retiring a key, but the guide never shows the `flightctl` commands to do that. Other user docs already use `flightctl get events --field-selector=...` for the same pattern — this guide should too.

Details in the inline comments.

docs/user/installing/configuring-encryption.md
asafbss: The section already links to `Encryption Events` which documents the event names, involved object fields, and how to query them. Adding CLI examples here duplicates the reference doc and creates a maintenance burden if the CLI syntax changes. Operators who need to query events can follow the linked reference.
WDYT?

docs/user/installing/configuring-encryption.md
asafbss: See my answer above: https://github.com/flightctl/flightctl/pull/3320#discussion_r3702173146

asafbennatan (approved):
LGTM

PR #3321, opened 2026-08-02
EDM-4969: Honor Podman health_status events for application status

asafbennatan:
## Summary
- Subscribe to Podman `health_status` events in the agent monitor (previously filtered out and explicitly dropped for Quadlet/VM apps)
- Map `unhealthy` workloads to `StatusUnhealthy` and fold them into application status as **Running + Degraded** (ready count excludes unhealthy); restore **Healthy** when health recovers
- Add unit coverage for status aggregation and health event handling

Fixes follow-up gap from EDM-4095 / #3182 where `vm-to-quadlet` embeds `HealthCmd` but the agent never consumed the resulting events.

Jira: https://issues.redhat.com/browse/EDM-4969

## Test plan
- [x] `go test ./internal/agent/device/applications/ -count=1`
- [ ] Deploy a VM app with failing HealthCmd and confirm device application summary becomes Degraded while status stays Running
- [ ] Confirm recovery to Healthy once the health check passes again

### DIFF
--- a/internal/agent/device/applications/applications.go
+++ b/internal/agent/device/applications/applications.go
@@ -22,14 +22,15 @@ const (
 type StatusType string
 
 const (
-	StatusCreate  StatusType = "create"
-	StatusInit    StatusType = "init"
-	StatusRunning StatusType = "start"
-	StatusStop    StatusType = "stop"
-	StatusDie     StatusType = "die" // docker only
-	StatusDied    StatusType = "died"
-	StatusRemove  StatusType = "remove"
-	StatusExited  StatusType = "exited"
+	StatusCreate    StatusType = "create"
+	StatusInit      StatusType = "init"
+	StatusRunning   StatusType = "start"
+	StatusStop      StatusType = "stop"
+	StatusDie       StatusType = "die" // docker only
+	StatusDied      StatusType = "died"
+	StatusRemove    StatusType = "remove"
+	StatusExited    StatusType = "exited"
+	StatusUnhealthy StatusType = "unhealthy"
 )
 
 func (c StatusType) String() string {
@@ -302,6 +303,7 @@ func (a *application) Status() (*v1beta1.DeviceApplicationStatus, v1beta1.Device
 	restarts := 0
 	exited := 0
 	stopped := 0
+	unhealthy := 0
 	for _, workload := range a.workloads {
 		restarts += workload.Restarts
 		switch workload.Status {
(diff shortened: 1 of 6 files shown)


internal/agent/device/applications/applications.go
```diff
@@ -412,6 +419,11 @@ func isRunningHealthy(total, healthy, initializing, exited int) bool {
 	return total > 0 && (healthy == total || healthy+exited == total) && initializing == 0
 }
 
+// isRunningUnhealthy reports containers that are still up but failing Podman health checks.
+func isRunningUnhealthy(total, healthy, unhealthy, initializing int) bool {
+	return total > 0 && unhealthy > 0 && healthy+unhealthy == total && initializing == 0
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 42d0428.

`isRunningUnhealthy` now includes terminal workload counts (`healthy+unhealthy+terminal == total`), and `applications_test.go` covers unhealthy+exited, unhealthy+died, and VM init-container mixes (e.g. ready `2/4`).

internal/agent/device/applications/applications.go
```diff
@@ -412,6 +419,10 @@ func isRunningHealthy(total, healthy, initializing, exited int) bool {
 	return total > 0 && (healthy == total || healthy+exited == total) && initializing == 0
 }
 
+func isRunningUnhealthy(total, healthy, unhealthy, initializing, terminal int) bool {
+	return total > 0 && unhealthy > 0 && initializing == 0 && healthy+unhealthy+terminal == total
```
galel12, line 426: nit: StatusStop isn't counted anywhere that keeps you in the unhealthy→Degraded path.

`terminal` here is `stopped` from `isTerminal`, which is exited/die/died/remove only. So unhealthy + a peer still on `stop` (restart window) fails this sum and falls through to `isErrored`. Probably a brief flicker — call out if you care, otherwise fine.

galel12 (commented):
LGTM overall. One minor edge-case note below.

internal/agent/device/applications/applications.go
asafbennatan: Good catch — fixed. `StatusStop` is now counted for the unhealthy→Degraded path (without treating it as terminal for desired-state Stopped), plus a test for unhealthy+stop.

galel12 (approved):
LGTM

PR #3322, opened 2026-08-02
EDM-4762: Reject direct spec.os.image on package-mode devices

galel12:
## EDM-4762: Reject direct spec.os.image on package-mode devices

**Jira:** https://redhat.atlassian.net/browse/EDM-4762
**Story type:** [DEV]

### Summary

Adds a server-side guard in `ReplaceDevice` and `PatchDevice` that returns `400 Bad Request` when a direct device edit newly assigns or changes `spec.os.image` on a device whose `status.capabilities.osMode` is `package`. The check is gated on a dedicated `enforceCapabilities` flag (separate from `enforceOwnership`) so API callers enforce it while fleet rollout can still write the full fleet spec to every device; package-mode agents reject an incompatible `os.image` locally and report OutOfDate.

### Changes

- `internal/service/device/{service,handler}.go` — `enforceCapabilities` on Replace/Patch; `isPackageModeOsImageConflict` helper (compares existing vs incoming image)
- `internal/transport/v1beta1/device.go` — API path passes `true, true`
- `internal/tasks/fleet_{rollout,selector}.go` — fleet path passes `false, false`
- Unit + integration tests for reject/allow paths, including retained-image PATCH and `enforceCapabilities=false`

### Testing

- **Unit tests:** `TestReplaceDevicePackageModeOsReject`, `TestPatchDevicePackageModeOsReject`
- **Integration tests:** "Package-mode OS image rejection" Context
- **Coverage:** `make lint`, focused unit/integration tests pass

### Acceptance Criteria

- [ ] AC-1: `ReplaceDevice` / `PatchDevice` (`enforceCapabilities=true`) return 400 when incoming `spec.os.image` is newly set/changed and existing device has `status.capabilities.osMode = package`
- [ ] AC-2: Validation applies to direct edits only — fleet rollout (`enforceCapabilities=false`) is not rejected
- [ ] AC-3: Devices with unknown/missing osMode are not rejected
- [ ] AC-4: Error message: `OS image is not supported on package-mode devices`
- [ ] AC-5: Retaining an existing `os.image` on an unrelated PATCH is not rejected

### Out of scope

`spec.os.catalogItemRef` package-mode guards → [EDM-4968](https://redhat.atlassian.net/browse/EDM-4968) (blocked on Catalog API #3284).

### DIFF
--- a/internal/service/device/handler.go
+++ b/internal/service/device/handler.go
@@ -242,13 +242,17 @@ func (h *DeviceServiceHandler) ReplaceDevice(ctx context.Context, orgId uuid.UUI
 				return nil, common.StoreErrorToApiStatus(getErr, false, domain.DeviceKind, &name)
 			}
 		} else {
-			if enforceOwnership && len(lo.FromPtr(existing.Metadata.Owner)) != 0 {
-				if !domain.DeviceSpecsAreEqual(lo.FromPtr(existing.Spec), lo.FromPtr(device.Spec)) {
-					return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+			if enforceOwnership {
+				if len(lo.FromPtr(existing.Metadata.Owner)) != 0 {
+					if !domain.DeviceSpecsAreEqual(lo.FromPtr(existing.Spec), lo.FromPtr(device.Spec)) {
+						return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+					}
 				}
 			}
-			if enforceCapabilities && isPackageModeOsImageConflict(existing, &device) {
-				return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
+			if enforceCapabilities {
+				if isPackageModeOsImageConflict(existing, &device) {
+					return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
+				}
 			}
 		}
 	}
@@ -485,13 +489,17 @@ func (h *DeviceServiceHandler) PatchDevice(ctx context.Context, orgId uuid.UUID,
 	common.NilOutManagedObjectMetaProperties(&newObj.Metadata)
 	newObj.Metadata.ResourceVersion = nil
 
-	if enforceOwnership && len(lo.FromPtr(currentObj.Metadata.Owner)) != 0 {
-		if !domain.DeviceSpecsAreEqual(lo.FromPtr(currentObj.Spec), lo.FromPtr(newObj.Spec)) {
-			return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+	if enforceOwnership {
+		if len(lo.FromPtr(currentObj.Metadata.Owner)) != 0 {
+			if !domain.DeviceSpecsAreEqual(lo.FromPtr(currentObj.Spec), lo.FromPtr(newObj.Spec)) {
+				return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+			}
 		}
 	}
-	if enforceCapabilities && isPackageModeOsImageConflict(currentObj, newObj) {
-		return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
+	if enforceCapabilities {
+		if isPackageModeOsImageConflict(currentObj, newObj) {
+			return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
+		}
 	}
 
 	_ = common.UpdateServiceSideStatus(ctx, orgId, newObj, h.fleetStore, h.log)
--- a/internal/service/device/handler.go
+++ b/internal/service/device/handler.go
@@ -237,23 +237,14 @@ func (h *DeviceServiceHandler) ReplaceDevice(ctx context.Context, orgId uuid.UUI
 
 	if enforceOwnership || enforceCapabilities {
 		existing, getErr := h.deviceStore.Get(ctx, orgId, name)
-		if getErr != nil {
-			if !errors.Is(getErr, flterrors.ErrResourceNotFound) {
-				return nil, common.StoreErrorToApiStatus(getErr, false, domain.DeviceKind, &name)
-			}
-		} else {
-			if enforceOwnership {
-				if len(lo.FromPtr(existing.Metadata.Owner)) != 0 {
-					if !domain.DeviceSpecsAreEqual(lo.FromPtr(existing.Spec), lo.FromPtr(device.Spec)) {
-						return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
-					}
-				}
-			}
-			if enforceCapabilities {
-				if isPackageModeOsImageConflict(existing, &device) {
-					return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
-				}
-			}
+		if getErr != nil && !errors.Is(getErr, flterrors.ErrResourceNotFound) {
+			return nil, common.StoreErrorToApiStatus(getErr, false, domain.DeviceKind, &name)
+		}
+		if existing != nil && enforceOwnership && len(lo.FromPtr(existing.Metadata.Owner)) != 0 && !domain.DeviceSpecsAreEqual(lo.FromPtr(existing.Spec), lo.FromPtr(device.Spec)) {
+			return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+		}
+		if existing != nil && enforceCapabilities && isPackageModeOsImageConflict(existing, &device) {
+			return nil, domain.StatusBadRequest(flterrors.ErrOsImageNotSupportedOnPackageMode.Error())
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


internal/service/device/handler.go
```diff
@@ -801,3 +811,16 @@ func (h *DeviceServiceHandler) processAwaitingReconnectIfNeeded(ctx context.Cont
...
+// image to a device that is known to be in package mode. Package-mode devices cannot
+// satisfy image-based OS targets, so the caller should reject the request early.
+func isPackageModeOsImageConflict(existing *domain.Device, incoming *domain.Device) bool {
+	if existing.Status == nil || existing.Status.Capabilities == nil || existing.Status.Capabilities.OsMode == nil {
+		return false
+	}
+	if *existing.Status.Capabilities.OsMode != domain.OsModePackage {
+		return false
+	}
+	return incoming.Spec != nil && incoming.Spec.Os != nil && incoming.Spec.Os.Image != ""
```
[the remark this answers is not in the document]
galel12: Good catch — after fleet rollout a package-mode device can already carry `spec.os.image`, so a non-OS PATCH was incorrectly rejected.

Fixed `isPackageModeOsImageConflict` to compare existing vs incoming image and only reject when a non-empty image is newly assigned or changed. Added regression coverage for retaining a fleet-applied image on PATCH/replace.

internal/service/device/handler.go
asafbennatan: enforce ownership is used to allow the caller to control if it wants to enforce ownership (usually this is true for api calls and false for internal services that actually wants to change ownership)

i am not sure we need that here , i assume that the check should happen regardless of existing resource?

galel12: Went with a separate `enforceCapabilities` flag rather than overloading `enforceOwnership`.

- API/transport: `true, true`
- Fleet rollout / selector: `false, false` (Path B — fleet may still write `os.image`; agent handles OutOfDate locally)
- `ReplaceDevice` does a single `Get` when either flag is set, then runs the checks independently

internal/service/device/handler.go
```diff
@@ -235,15 +235,24 @@ func (h *DeviceServiceHandler) ReplaceDevice(ctx context.Context, orgId uuid.UUI
...
+			if enforceOwnership {
+				if len(lo.FromPtr(existing.Metadata.Owner)) != 0 {
+					if !domain.DeviceSpecsAreEqual(lo.FromPtr(existing.Spec), lo.FromPtr(device.Spec)) {
+						return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+					}
+				}
+			}
+			if enforceCapabilities {
+				if isPackageModeOsImageConflict(existing, &device) {
+					return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
```
asafbennatan: **Design (LOW):** This rejection message is a raw string literal here and again in `PatchDevice` (~line 501). The sibling ownership check uses a shared sentinel (`flterrors.ErrUpdatingResourceWithOwnerNotAllowed`) instead of a literal. Consider extracting this into a similar sentinel/const so the two call sites can't drift (e.g. a future wording tweak applied to only one of them).

internal/service/device/handler.go
```diff
@@ -235,15 +235,24 @@ func (h *DeviceServiceHandler) ReplaceDevice(ctx context.Context, orgId uuid.UUI
 		return nil, domain.StatusBadRequest("resource name specified in metadata does not match name in path")
 	}
 
-	if enforceOwnership {
+	if enforceOwnership || enforceCapabilities {
```
asafbennatan, line 238: **Design:** This block nests 4 levels deep (`if enforceOwnership || enforceCapabilities` -> `if getErr != nil {...} else {...}` -> per-flag `if` -> condition `if`) for what are really two independent guard conditions. Flattening to early-return guards keeps the two checks visibly independent and easier to scan:

```go
if enforceOwnership || enforceCapabilities {
	existing, getErr := h.deviceStore.Get(ctx, orgId, name)
	if getErr != nil && !errors.Is(getErr, flterrors.ErrResourceNotFound) {
		return nil, common.StoreErrorToApiStatus(getErr, false, domain.DeviceKind, &name)
	}
	if existing != nil && enforceOwnership && len(lo.FromPtr(existing.Metadata.Owner)) != 0 && !domain.DeviceSpecsAreEqual(lo.FromPtr(existing.Spec), lo.FromPtr(device.Spec)) {
		return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
	}
	if existing != nil && enforceCapabilities && isPackageModeOsImageConflict(existing, &device) {
		return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
	}
}
```

This retains the exact original semantics: `existing` is `nil` whenever `getErr != nil` (the store returns `nil, err` on failure), so guarding on `existing != nil` is equivalent to the current `else` branch -- including the not-found case, where both checks are correctly skipped.

internal/service/device/handler.go
```diff
@@ -480,9 +489,16 @@ func (h *DeviceServiceHandler) PatchDevice(ctx context.Context, orgId uuid.UUID,
 	common.NilOutManagedObjectMetaProperties(&newObj.Metadata)
 	newObj.Metadata.ResourceVersion = nil
 
-	if enforceOwnership && len(lo.FromPtr(currentObj.Metadata.Owner)) != 0 {
-		if !domain.DeviceSpecsAreEqual(lo.FromPtr(currentObj.Spec), lo.FromPtr(newObj.Spec)) {
-			return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
+	if enforceOwnership {
```
asafbennatan: **Design:** Same nesting shape as in `ReplaceDevice` above for two checks that are actually independent per-flag guards. Flattening reads more directly:

```go
if enforceOwnership && len(lo.FromPtr(currentObj.Metadata.Owner)) != 0 && !domain.DeviceSpecsAreEqual(lo.FromPtr(currentObj.Spec), lo.FromPtr(newObj.Spec)) {
	return nil, common.StoreErrorToApiStatus(flterrors.ErrUpdatingResourceWithOwnerNotAllowed, false, domain.DeviceKind, &name)
}
if enforceCapabilities && isPackageModeOsImageConflict(currentObj, newObj) {
	return nil, domain.StatusBadRequest("OS image is not supported on package-mode devices")
}
```

internal/service/device/handler_test.go
```diff
@@ -940,3 +940,182 @@ func TestGetRenderedDevice(t *testing.T) {
 	require.Equal(t, int32(http.StatusOK), status.Code)
 	require.Equal(t, "foo", lo.FromPtr(result.Metadata.Name))
 }
+
+func TestReplaceDevicePackageModeOsReject(t *testing.T) {
```
asafbennatan, line 944: **Test Coverage (MEDIUM):** `enforceCapabilities` was deliberately split out from `enforceOwnership` as an independent flag, but every case here (and in `TestPatchDevicePackageModeOsReject` below) passes matched pairs -- `enforceOwnership` stays fixed at `true` while only `enforceCapabilities` varies. There's no case for `enforceOwnership=false, enforceCapabilities=true` (or vice versa) on a package-mode device, so a future change that accidentally re-couples the two checks wouldn't be caught here. Worth adding at least one mismatched-flag case per function to lock in the intended independence.

internal/service/device/handler_test.go
```diff
@@ -940,3 +940,182 @@ func TestGetRenderedDevice(t *testing.T) {
...
+			require.NoError(t, err)
+
+			incoming := domain.Device{
+				Metadata: domain.ObjectMeta{Name: lo.ToPtr("pkg-dev")},
+				Spec:     &domain.DeviceSpec{Os: tt.incomingOs},
+			}
+			_, status := svc.ReplaceDevice(ctx, orgId, "pkg-dev", incoming, nil, true, tt.enforceCapabilities)
+			require.Equal(t, tt.wantCode, status.Code)
+			if tt.wantMessage != "" {
+				require.Contains(t, status.Message, tt.wantMessage)
```
asafbennatan: **Test Coverage (LOW):** `require.Contains` is used for an exact, non-interpolated message. The analogous ownership-conflict assertions elsewhere in this file (lines 237, 383) use `require.Equal` for the same kind of fixed message. Since `domain.StatusBadRequest` sets `Status.Message` to exactly the string passed in, `require.Equal(t, tt.wantMessage, status.Message)` would be more precise and consistent with this file's existing convention.

internal/service/device/handler_test.go
```diff
@@ -940,3 +940,182 @@ func TestGetRenderedDevice(t *testing.T) {
...
+				Spec:   &domain.DeviceSpec{Os: tt.existingOs},
+				Status: &status,
+			}
+			_, err := st.device.Create(ctx, orgId, &existing, nil)
+			require.NoError(t, err)
+
+			_, st2 := svc.PatchDevice(ctx, orgId, "pkg-dev", tt.patch, true, tt.enforceCapabilities)
+			require.Equal(t, tt.wantCode, st2.Code)
+			if tt.wantMessage != "" {
+				require.Contains(t, st2.Message, tt.wantMessage)
```
asafbennatan: **Test Coverage (LOW):** Same `Contains`-vs-`Equal` point as the `ReplaceDevice` test above -- `require.Equal(t, tt.wantMessage, st2.Message)` would be more precise and consistent with this file's convention for fixed messages.

test/integration/service/device_test.go
```diff
@@ -792,6 +792,77 @@ var _ = Describe("Device Application Status Events Integration Tests", func() {
 		})
 	})
 
+	Context("Package-mode OS image rejection", func() {
```
asafbennatan, line 795: **Test Coverage (LOW):** This `Context` covers PUT-deny, PATCH-deny, image-mode-allow, and no-capabilities-allow, but has no case for retaining an existing `os.image` on an unrelated PATCH (or a same-image PUT round-trip) against the real Postgres-backed store -- the exact scenario `isPackageModeOsImageConflict`'s existing-vs-incoming comparison is designed for. It's unit-tested against the fake store in `handler_test.go`, but an integration-level case would also catch e.g. a serialization round-trip issue the fake store wouldn't.

asafbennatan (commented):
Reviewed the diff and the surrounding code for the package-mode OS-image guard. Core logic (`isPackageModeOsImageConflict`, the `ReplaceDevice`/`PatchDevice` wiring, and the fleet rollout/selector `enforceCapabilities=false` paths) looks correct, and all 5 stated ACs hold against the code as written. Left a few inline suggestions below (mostly test-coverage/clarity, non-blocking) plus a design suggestion to flatten the nested enforcement checks in `ReplaceDevice`/`PatchDevice`.

internal/service/device/handler.go
galel12: Done — extracted to `flterrors.ErrOsImageNotSupportedOnPackageMode` and both call sites now use the sentinel.

internal/service/device/handler.go
galel12: Adopted the flatten. Kept the shared `Get` when either flag needs existing state, then early-return ownership and capabilities checks independently.

internal/service/device/handler.go
galel12: Flattened the same way in `PatchDevice` — independent early returns for ownership and capabilities.

internal/service/device/handler_test.go
galel12: Added mismatched-flag cases for both Replace and Patch: `enforceOwnership=false, enforceCapabilities=true` still rejects the OS conflict, and owned `true/false` returns the ownership conflict.
see https://github.com/flightctl/flightctl/pull/3322/changes#diff-5e032e449d6591d77ec45c5a5769656e23cac8e09b1ff3afebba07d1bbbc2b13R966-R980

internal/service/device/handler_test.go
galel12: Switched to `require.Equal` against the sentinel `.Error()` for the fixed message.

internal/service/device/handler_test.go
galel12: Same here — now using `require.Equal` with the shared sentinel message.

test/integration/service/device_test.go
galel12: Added an integration case that fleet-applies `os.image` with both flags off, then PATCHes an unrelated label with `true, true` and asserts the image is retained.

asafbennatan (approved):
LGTM

PR #3323, opened 2026-08-02
EDM-4768: add package-mode e2e suite

eldar101:
**Summary**
Adds a new Linux-only E2E suite for EDM-4768 under test/e2e/package_mode/ to cover package-mode device behavior end to end.
Included coverage:

- package-mode device enrolls and reports status.capabilities.osMode=package
- package-mode fleet without spec.os.image delivers inline config and a podman app successfully
- mixed-fleet behavior where a package-mode device rejects spec.os.image and stays OutOfDate
- mixed-fleet behavior where an image-mode device converges on the requested OS image normally
- negative check that normal package-mode operation does not emit error-level agent logs

**Release target**
 release-1.3
Target release: release-1.3

### DIFF
--- a/test/e2e/package_mode/package_mode_test.go
+++ b/test/e2e/package_mode/package_mode_test.go
@@ -1,0 +1,712 @@
+//go:build linux
+
+package packagemode_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/flightctl/flightctl/api/core/v1beta1"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/harness/e2e/vm"
+	testutil "github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	packageModeQCOWEnvVar         = "E2E_PACKAGE_MODE_QCOW"
+	imageModeQCOWEnvVar           = "E2E_IMAGE_MODE_QCOW"
+	defaultImageModeQCOWPath      = "bin/output/qcow2/disk.qcow2"
+	packageModeVMUser             = "user"
+	packageModeVMPassword         = "user"
+	packageModeContainerRunAsUser = "flightctl"
+	packageModeDeviceSSHPortBase  = 23100
+	packageModeMixedSSHPortOffset = 50
+	imageModeDeviceSSHPortBase    = 23200
+	packageModeBootTimeout        = 6 * time.Minute
+	packageModeCommandTimeout     = 60 * time.Second
+	packageModeEnrollTimeout      = 5 * time.Minute
+	packageModePollInterval       = 5 * time.Second
+	packageModeConfigPath         = "/etc/flightctl/package-mode-e2e.conf"
+	packageModeConfigContent      = "package-mode-e2e=enabled"
+	packageModeContainerAppName   = "package-mode-sleep"
+	packageModeLogObservationTime = 30 * time.Second
+	packageModeFleetLabelKey      = "package-mode-suite"
+	packageModeNormalFleetPrefix  = "package-mode-normal"
+	packageModeMixedFleetPrefix   = "package-mode-mixed"
+	packageModeVMNamePrefix       = "package-mode"
+	imageModeVMNamePrefix         = "image-mode"
+	packageModeExpectedRejectText = "package-mode device cannot satisfy spec with os.image"
+)
+
+var _ = Describe("Package-mode device scenarios", Ordered, func() {
+	var harness *e2e.Harness
+
+	BeforeEach(func() {
+		harness = e2e.GetWorkerHarness()
+	})
+
+	Context("normal package-mode operation", Ordered, func() {
+		var (
+			deviceID      string
+			fleetName     string
+			selector      v1beta1.LabelSelector
+			packageModeVM *vmUnderTest
+		)
+
+		BeforeAll(func() {
+			var err error
+			harness = e2e.GetWorkerHarness()
+			restoreContext := setHarnessSetupContext(harness, packageModeNormalFleetPrefix)
+			defer harness.SetTestContext(restoreContext)
+
+			packageQCOW := requireQCOWPath(packageModeQCOWEnvVar, "")
+			fleetName = uniqueFleetName(packageModeNormalFleetPrefix)
+			selector = v1beta1.LabelSelector{
+				MatchLabels: &map[string]string{
+					packageModeFleetLabelKey: fleetName,
+				},
+			}
+
+			packageModeVM, err = startCustomVMFromQCOW(packageQCOW, packageModeVMNamePrefix, packageModeDeviceSSHPortBase+GinkgoParallelProcess())
+			Expect(err).ToNot(HaveOccurred())
+
+			inlineConfig, err := e2e.NewInlineConfigSpec("package-mode-inline", []v1beta1.FileSpec{{
+				Path:    packageModeConfigPath,
+				Content: packageModeConfigContent,
+			}})
+			Expect(err).ToNot(HaveOccurred())
+
+			appImage := harness.GetSleepAppImageRefForFleet(packageModeAuxSvcs.Registry.Host, packageModeAuxSvcs.Registry.Port, testutil.SleepAppTags.V2)
+			containerApp, err := e2e.NewContainerApplicationSpecWithRunAs(
+				packageModeContainerAppName,
+				appImage,
+				nil,
+				nil,
+				nil,
+				nil,
+				packageModeContainerRunAsUser,
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			deviceSpec := v1beta1.DeviceSpec{
+				Config:       &[]v1beta1.ConfigProviderSpec{inlineConfig},
+				Applications: &[]v1beta1.ApplicationProviderSpec{containerApp},
+			}
+
+			err = harness.CreateOrUpdateTestFleet(fleetName, selector, deviceSpec)
+			Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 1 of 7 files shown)


galel12 (dismissed):
LGTM — EDM-4768 ACs are covered (enroll/osMode, config+app, mixed-fleet package OutOfDate + image converge, NFR-3 log check). Suite layout matches existing e2e patterns.

Nit for follow-up if you hit flakes before CI wiring (EDM-4765): Path B should assert agent/status signals rather than `fleet-controller/lastRolloutError`, and the no-error-log check is safer as `Consistently` after steady state.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,514 @@
...
+				g.Expect(device.Status).ToNot(BeNil())
+				g.Expect(device.Status.Updated.Status).To(Equal(v1beta1.DeviceUpdatedStatusOutOfDate))
+				g.Expect(device.Status.Updated.Info).ToNot(BeNil())
+				g.Expect(*device.Status.Updated.Info).To(ContainSubstring(packageModeExpectedInfoPrefix))
+				g.Expect(*device.Status.Updated.Info).To(ContainSubstring(packageModeExpectedRejectText))
+				g.Expect(device.Status.Config.RenderedVersion).To(Equal(strconv.Itoa(packageInitialVersion)))
+			}, testutil.LONG_TIMEOUT, packageModePollInterval).Should(Succeed())
+
+			Eventually(func() error {
+				return harness.CheckFleetControllerErrorAnnotation(packageDeviceID, packageModeExpectedRejectText)
```
galel12: **Must-fix:** `CheckFleetControllerErrorAnnotation` is the wrong signal for this mixed-fleet case.

`fleet-controller/lastRolloutError` is set when fleet *rendering* fails. For mixed fleets with `spec.os.image`, fleet rollout **succeeds** writing the image onto the package-mode device; the agent rejects afterward. That annotation will not contain `package-mode device cannot satisfy…`, so this will burn `LONG_TIMEOUT` and fail.

Drop this check. Assert agent/server status instead, e.g.:
- `status.updated.status == OutOfDate`
- rendered version stays at `packageInitialVersion`
- `ConditionTypeDeviceUpdating` / `UpdateStateError` message contains the reject text (if the agent surfaces it)
- and/or agent journal contains the reject line

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+		It("keeps the package-mode device OutOfDate when the fleet adds spec.os.image", Label("90148", "sanity"), func() {
+			Eventually(func(g Gomega) {
+				device, err := harness.GetDevice(packageDeviceID)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(device).ToNot(BeNil())
+				g.Expect(device.Status).ToNot(BeNil())
+				g.Expect(device.Status.Updated.Status).To(Equal(v1beta1.DeviceUpdatedStatusOutOfDate))
+				g.Expect(device.Status.Updated.Info).ToNot(BeNil())
+				g.Expect(*device.Status.Updated.Info).To(ContainSubstring(packageModeExpectedInfoPrefix))
```
galel12: **Must-fix / brittle:** `Updated.Info` may not contain both of these prefixes.

If the agent sets `UpdateStateError`, `managedDeviceOutOfDateMessage` can produce `Device could not be updated…: package-mode device cannot satisfy…` — good.
If that condition is missing, Info falls through to `DeviceOutOfSyncWithFleetText` (“has not yet been scheduled…”), which matches neither constant.

Safer: keep OutOfDate + stuck rendered version as the hard asserts; treat Info substrings as optional / assert the updating condition/agent message instead.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+			labels := map[string]string{packageModeFleetLabelKey: fleetName}
+
+			packageDeviceID, err = enrollCustomVM(harness, packageModeVM, labels)
+			Expect(err).ToNot(HaveOccurred())
+
+			imageDeviceID, err = enrollCustomVM(harness, imageModeVM, labels)
+			Expect(err).ToNot(HaveOccurred())
+
+			packageInitialVersion, err = harness.GetCurrentDeviceRenderedVersion(packageDeviceID)
```
galel12, line 235: **Must-fix (flake):** rendered-version baselines are read before the initial empty-fleet render can settle.

Flow is: create empty fleet → enroll → immediately `GetCurrentDeviceRenderedVersion` → then update fleet with OS image. If the first render lands after the read, `packageInitialVersion` / `imageExpectedVersion` are wrong and both mixed-fleet Its flake.

Wait for the initial render (or use `PrepareNextDeviceVersion`) before applying the OS-image fleet update.

galel12 (changes requested):
AC coverage shape looks right for EDM-4768, but a few assertions look incorrect or flake-prone for the mixed-fleet `os.image` case (fleet write succeeds, package-mode agent rejects). Please address the inline comments before merge.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,514 @@
...
+				g.Expect(device.Status).ToNot(BeNil())
+				g.Expect(device.Status.Applications).ToNot(BeNil())
+				g.Expect(len(device.Status.Applications)).To(BeNumerically(">=", 1))
+				g.Expect(device.Status.Applications[0].Name).To(Equal(packageModeContainerAppName))
+				g.Expect(device.Status.Applications[0].RunAs).To(Equal(v1beta1.Username(packageModeContainerRunAsUser)))
+			}, packageModeEnrollTimeout, packageModePollInterval).Should(Succeed())
+		})
+
+		It("emits no error-level agent logs during normal package-mode operation", Label("90150", "sanity"), func() {
+			Eventually(func(g Gomega) {
```
galel12: **Must-fix:** `Eventually` is the wrong poller for an absence assertion.

`Eventually(... ToNot(ContainSubstring("level=error")))` passes on the first poll (often before the agent has done meaningful work). Use `Consistently` after config/app are healthy. Also consider a tighter token than bare `level=error`.

galel12:
**CI note:** these specs will `Skip` until EDM-4765 wires `E2E_PACKAGE_MODE_QCOW` (and an image-mode qcow for the mixed-fleet case). This PR alone will not exercise package-mode in CI.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,605 @@
...
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(device).ToNot(BeNil())
+				g.Expect(device.Status).ToNot(BeNil())
+				g.Expect(device.Status.Updated.Status).To(Equal(v1beta1.DeviceUpdatedStatusOutOfDate))
+				g.Expect(device.Status.Config.RenderedVersion).To(Equal(strconv.Itoa(packageInitialVersion)))
+				g.Expect(deviceRejectsMixedFleetUpdate(device, packageModeExpectedRejectText)).To(BeTrue())
+			}, testutil.LONG_TIMEOUT, packageModePollInterval).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				logs, err := packageModeVM.Instance.GetServiceLogs(testutil.FLIGHTCTL_AGENT_SERVICE)
```
galel12: **Nit (also open from CodeRabbit):** this reject-log poll still uses `GetServiceLogs` with no SSH deadline. A hung guest SSH makes the `Eventually` ignore `LONG_TIMEOUT`.

Route through the same timed journal path as the NFR-3 check (`readSuiteLibvirtVMServiceLogsSince` / `RunSSHContext` with `packageModeCommandTimeout`).

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,605 @@
...
+	packageModeMixedSSHPortOffset = 50
+	imageModeDeviceSSHPortBase    = 23200
+	packageModeBootTimeout        = 6 * time.Minute
+	packageModeCommandTimeout     = 60 * time.Second
+	packageModeEnrollTimeout      = 5 * time.Minute
+	packageModePollInterval       = 5 * time.Second
+	packageModeConfigPath         = "/etc/flightctl/package-mode-e2e.conf"
+	packageModeConfigContent      = "package-mode-e2e=enabled"
+	packageModeContainerAppName   = "package-mode-sleep"
+	packageModeErrorLogToken      = "level=error"
```
galel12: **Nit (CodeRabbit missed):** bare `level=error` is a brittle absence token — any unrelated agent error (or log-format change) fails the NFR-3 check, and false negatives are easy if the logger uses a different field shape.

Prefer asserting absence of a package-mode-specific failure string, or at least something tighter than any error-level line in the whole journal window.

galel12 (commented):
Thanks — previous must-fixes look addressed (no `lastRolloutError`, reject via status/condition + agent logs, rendered-version wait before fleet OS update, `Consistently` for the no-error check).

A few remaining nits below; the `GetServiceLogs` timeout one is also open from CodeRabbit.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,605 @@
...
+			}, packageModeEnrollTimeout, packageModePollInterval).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				device, err := harness.GetDevice(deviceID)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(device).ToNot(BeNil())
+				g.Expect(device.Status).ToNot(BeNil())
+				g.Expect(device.Status.Applications).ToNot(BeNil())
+				g.Expect(len(device.Status.Applications)).To(BeNumerically(">=", 1))
+				g.Expect(device.Status.Applications[0].Name).To(Equal(packageModeContainerAppName))
```
galel12: **Nit (CodeRabbit missed):** `Applications[0]` assumes status order. Prefer lookup by name (`packageModeContainerAppName`) so a second app / reorder doesn’t flake this.

galel12:
Closing in favor of https://github.com/flightctl/flightctl/pull/3342, which carries this suite on top of the merged package-mode CI wiring (#3336).

PR #3324, opened 2026-08-02
EDM-4971: Support linked git worktrees in RPM build

galel12:
## Problem

`make e2e-agent-images` fails when run from a linked git worktree. Packit's `git archive` inside the build container cannot resolve git metadata because the worktree's `.git` file points to a host path not mounted in the container.

```
fatal: not a git repository: (null)
packit.exceptions.PackitSRPMException
Command 'git archive --output packaging/rpm/flightctl-.tar.gz --prefix flightctl-/ HEAD' failed
```

## Root Cause

`hack/build_rpms.sh` mounts only the checkout directory (`-v $repo_root:/work`). Linked worktrees store git objects under the main repo's `.git/` (referenced via an absolute path in the worktree's `.git` file), which is outside the mounted tree.

A secondary issue: `current_tree_state` used `[ ! -d .git ]` which always evaluates true for worktrees (`.git` is a file), falsely reporting "clean".

## Fix

1. Detect linked worktrees and bind-mount the git common directory at the same host-absolute path so Packit can resolve git metadata. The resolved path is validated as a legitimate git directory before mounting.
2. Replace the `[ -d .git ]` / `git diff --quiet` probes with `git status --porcelain` which correctly detects staged changes, untracked files, and works regardless of `.git`'s shape.
3. Apply the same tree-state fix to `test/scripts/agent-images/scripts/build.sh`.

## Testing

- Full `sudo ./hack/build_rpms.sh` from a linked worktree: all 6 RPMs built successfully with correct version strings.
- Verified `git archive` fails without the mount and succeeds with it in an isolated container.
- Normal clone path unchanged (no extra mounts produced).

## Risk Assessment

Low — no-op for normal clones; only activates for linked worktrees. Same trust boundary as the existing repo mount.

## Release Target

release-1.3

Fixes https://redhat.atlassian.net/browse/EDM-4971

### DIFF
--- a/hack/build_rpms.sh
+++ b/hack/build_rpms.sh
@@ -53,7 +53,82 @@ PACKIT_BUILDER_IMAGE="${PACKIT_BUILDER_IMAGE:-quay.io/flightctl-tests/packit-bui
 ##############################################################################
 
 current_tree_state() {
-  (cd "${REPO_ROOT}" && ( ( [ ! -d ".git" ] || git diff --quiet ) && echo "clean" ) || echo "dirty")
+  # Use git status --porcelain to detect all dirty states: unstaged changes,
+  # staged changes, and untracked files. Linked worktrees use a .git *file*,
+  # so the old `[ ! -d .git ]` probe would always report "clean".
+  (cd "${REPO_ROOT}" && {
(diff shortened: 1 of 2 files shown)


eldar101:
fix code qualtiy + check warnings and I'll approve.

eldar101 (approved):

PR #3327, opened 2026-08-03
NO-ISSUE: allow serial and vnc app console sessions to coexist

asafbennatan:
## Summary
- `AppConsoleSessionManager` keyed console-session exclusivity on `appName` alone (in `StartSession`'s fast-path check, `addAppSession`, and `replaceAppSession`), so a `serial` and a `vnc` session on the same VM app were mutually exclusive — opening one while the other was active returned 409 unless `--force` was used.
- Serial (`virt-serial0`) and VNC (`virt-vnc`) use different sockets and don't contend for the same resource, so this was unintended: the check predates VNC support (written when only serial existed) and wasn't updated when VNC was added later.
- Exclusivity is now keyed on `(appName, consoleType)`: only two sessions of the *same* console type for the same app conflict. `--force` now only takes over the existing session of the same type, leaving a concurrently active session of the other type untouched.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/console/...` — added:
  - `TestAppConsoleSessionManager_StartSession_DifferentConsoleTypeSameApp_Allowed` — vnc no longer conflicts with an active serial session on the same app
  - `TestAppConsoleSessionManager_StartSession_Force_OnlyReplacesSameConsoleType` — force-takeover of serial leaves a concurrent vnc session on the same app untouched
  - `TestReplaceAppSession_DifferentConsoleType_AddsAlongsideExisting` — lower-level `replaceAppSession` unit test for the same behavior
  - updated `TestAppConsoleSessionManager_StartSession_DuplicateAppNameAndConsoleType` / `TestReplaceAppSession_ExistingEntry_SetsReplacesSessionID` to reflect the new same-type-only semantics
- [x] `gofmt` / `go vet` clean

### DIFF
--- a/api/core/v1beta1/openapi.yaml
+++ b/api/core/v1beta1/openapi.yaml
@@ -2290,7 +2290,7 @@ paths:
         - name: force
           in: query
           required: false
-          description: If true, take over an already-active console session for this application instead of failing with a 409 Conflict. The replaced session is disconnected and told why.
+          description: If true, take over an already-active console session of the same type for this application instead of failing with a 409 Conflict. The replaced session is disconnected and told why.
           schema:
             type: boolean
             default: false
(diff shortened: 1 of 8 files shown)


asafbss (approved):
LGTM

PR #3329, opened 2026-08-03
EDM-4970: Harden authprovider API readiness probe

eldar101:
Hardened the authprovider API readiness check used during browser-login setup.
The test now stops on redirects, only treats real 2xx responses as ready, and logs response body close errors instead of silently ignoring them. This makes failures clearer and avoids treating an unexpected redirect/error response as a healthy API.

Release target
release-1.3
Target release: release-1.3


eldar101:
@amalykhi @Samudelacruz77 @hferber1

dudyas6 (approved):

PR #3330, opened 2026-08-03
EDM-4976: Fix e2e agent-image bundling failure on dangling images

galel12:
## Problem

`make e2e-agent-images` fails during app-image bundling with:

```
Error: parsing reference "<none>:<none>": invalid reference format
make: *** [...] Error 125
```

`bundle.sh` collects image references via `podman images --format '{{.Repository}}:{{.Tag}}'` and passes them to `podman save`. When dangling (untagged) images exist in the local store with the matching label, they appear as `<none>:<none>` which `podman save` rejects.

## Root Cause

Two issues:

1. `bundle.sh` does not filter out `<none>:<none>` entries before calling `podman save`.
2. `build.sh` mislabels device variants as `io.flightctl.e2e.component=app` (should be `device`), causing variant rebuild leftovers to appear in the app bundle filter.

## Fix

1. **`bundle.sh`**: Add `grep -v '^<none>:<none>$'` to both image-listing paths (with and without `--image-pattern`).
2. **`build.sh`**: Change variant label from `app` to `device`, consistent with the base image label at line 180.

## Testing

- Created tagged + dangling images with the `app` label locally
- Verified fixed `bundle.sh` excludes `<none>:<none>` and saves successfully (both code paths)
- Confirmed unfixed path produces the exact ticket error (exit 125)
- Full unit test suite: 6053 passed, 0 failed

## Confidence

High — root cause confirmed by local reproduction; fix is minimal and both code paths verified.

## Risk Assessment

Low — changes are scoped to e2e test infrastructure scripts only. No impact on production code.

### DIFF
--- a/test/scripts/agent-images/scripts/build.sh
+++ b/test/scripts/agent-images/scripts/build.sh
@@ -227,7 +227,7 @@ echo "  Variants to build: ${variants_list:-none}"
         --build-context "variant-context='"${BASE_DIR}"'/variants/{}" \
         --build-context "common='"${BASE_DIR}"'/common" \
         --build-arg BASE_IMAGE="'"${base_img_canonical}"'" \
-        --label "io.flightctl.e2e.component=app" \
+        --label "io.flightctl.e2e.component=device" \
         -f "'"${BASE_DIR}"'/variants/{}/Containerfile" \
         -t "$v_img_canonical" \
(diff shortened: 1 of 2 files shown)


eldar101:
looks good, ping me for approval after coderabbit + Pre-merge checks

eldar101 (approved):

PR #3331, opened 2026-08-03
EDM-4761: add package-mode image build infrastructure

eldar101:
Summary:
add a new CentOS Stream 9 package-mode base Containerfile at test/scripts/agent-images/containerfiles/cs9-regular/Containerfile
extend the agent image build scripts to recognize AGENT_OS_ID=cs9-regular
wire BUILD_TYPE=regular through create_agent_images.sh so it builds the base image, variants bundle, and qcow artifact for the package-mode flavor
add scripts/qcow2_regular.sh to produce a libvirt-ready package-mode qcow from the CentOS Stream 9 cloud image using virt-customize
keep package-mode qcow generation separate from the existing bootc bootc-image-builder flow
document the new cs9-regular flavor and package-mode qcow path in test/scripts/agent-images/README.md

Release target
release-1.3
Target release: release-1.3

### DIFF
--- a/test/scripts/agent-images/README.md
+++ b/test/scripts/agent-images/README.md
@@ -14,6 +14,9 @@ The `AGENT_OS_ID` parameter controls which OS flavor to build:
 # Build for default OS (cs9-bootc)
 make e2e-agent-images
 
+# Build package-mode images
+BUILD_TYPE=regular AGENT_OS_ID=cs9-regular make e2e-agent-images
+
 # Build for specific OS
 AGENT_OS_ID=cs10-bootc make e2e-agent-images
 ```
@@ -22,16 +25,25 @@ AGENT_OS_ID=cs10-bootc make e2e-agent-images
 
 The script is a wrapper that delegates to the modular build system:
 1. **Base image**: Built using `scripts/build.sh --base`
-2. **Variants and qcow2**: Built using `scripts/build_and_qcow2.sh`
+2. **Bootc path**: `scripts/build_and_qcow2.sh` builds variants and a bootc qcow2
+3. **Package-mode path**: `BUILD_TYPE=regular AGENT_OS_ID=cs9-regular` builds the package-mode base image and qcow2 only
 
-The build process automatically handles different OS flavors (cs9-bootc, cs10-bootc)
+The build process automatically handles different OS flavors (cs9-bootc, cs9-regular, cs10-bootc)
 and RPM source detection (local, COPR, or Brew registry).
+
+`BUILD_TYPE=regular` is a package-mode-only path in `create_agent_images.sh`. There,
+omitting `AGENT_OS_ID` defaults it to `cs9-regular`, and any other `AGENT_OS_ID`
+with `BUILD_TYPE=regular` fails fast.
+
+For the `make e2e-agent-images` entrypoint, pass `AGENT_OS_ID=cs9-regular`
+explicitly. Make still defaults `AGENT_OS_ID` to `cs9-bootc`.
 
 ## OS Flavors and Tagging
 
 The build system supports multiple OS flavors with dedicated Containerfiles:
 
 - **cs9-bootc** - Based on CentOS Stream 9 bootc (default)
+- **cs9-regular** - Package-mode base for non-bootc E2E coverage
 - **cs10-bootc** - Based on CentOS Stream 10 bootc
 
 Each flavor has its own explicit Containerfile eliminating conditional logic.
@@ -42,12 +54,17 @@ Each flavor has its own explicit Containerfile eliminating conditional logic.
 # Build cs9-bootc images (default, community)
 ./scripts/build.sh --base
 
+# Build cs9-regular package-mode images
+AGENT_OS_ID=cs9-regular ./scripts/build.sh --base
+
 # Build cs10-bootc images (community)
 AGENT_OS_ID=cs10-bootc ./scripts/build.sh --base
 
 # Build Red Hat variants
 DISTRO=redhat AGENT_OS_ID=cs9-bootc ./scripts/build.sh --base
 DISTRO=redhat AGENT_OS_ID=cs10-bootc ./scripts/build.sh --base
+
+# cs9-regular package-mode does not support DISTRO=redhat/RHEM
 ```
 
 ### Image Tagging
@@ -79,6 +96,8 @@ agent-images/
 ├── containerfiles/        # OS flavor-specific Containerfiles
 │   ├── cs9-bootc/         # CentOS Stream 9 bootc
 │   │   └── Containerfile
+│   ├── cs9-regular/       # Package-mode base image
+│   │   └── Containerfile
 │   ├── cs9-bootc-redhat/  # RHEL 9 bootc
 │   │   └── Containerfile
 │   ├── cs10-bootc/        # CentOS Stream 10 bootc
@@ -105,9 +124,10 @@ The images are built using the `Containerfile` files in the respective directori
 The `scripts/` directory contains modular build automation:
 
 - **`build.sh`** - Main build script with options: `--base`, `--variants`, `--apps`
-- **`build_and_qcow2.sh`** - Orchestrates variants and QCOW2 builds in parallel
+- **`build_and_qcow2.sh`** - Orchestrates variant and QCOW2 builds; for `*-regular` it is QCOW2-only by default unless `SKIP_VARIANTS_BUILD=false` is set explicitly
 - **`bundle.sh`** - Creates tar bundles of built images for distribution
 - **`qcow2.sh`** - Generates bootable QCOW2 disk images using bootc-image-builder
+- **`qcow2_regular.sh`** - Generates package-mode QCOW2 images from the configured cloud image using `virt-customize`
 - **`upload-images.sh`** - Uploads image bundles to container registries
 
 Use `./scripts/build.sh --help` for detailed usage and options.
@@ -135,6 +155,15 @@ Where `<name>` is `base`, `v2`, `v3`, etc.
 
 > **Note:** `qcow2.sh` writes the disk image to `bin/output/agent-qcow2-${OS_ID}/qcow2/disk.qcow2`.
 > When using `create_agent_images.sh`, the image is moved to `bin/output/qcow2/disk.qcow2`.
+
+For `BUILD_TYPE=regular AGENT_OS_ID=cs9-regular`, `create_agent_images.sh` still writes the final QCOW2 to the same `bin/output/qcow2/disk.qcow2` path, but it is produced by `scripts/qcow2_regular.sh` from the configured cloud image instead of `bootc-image-builder`.
+
+That means the package-mode OCI base and package-mode qcow2 are parallel build paths:
+- `containerfiles/cs9-regular/Containerfile` is the source of truth for the OCI base image
+- `scripts/qcow2_regular.sh` is the source of truth for the package-mode qcow2 guest customization
+
+They are intentionally parallel because the qcow2 is built from a cloud image, not converted from the OCI base image.
+Agent config, certificates, and registry remapping are expected to be injected into the qcow2 after build via the existing e2e injection flow.
 
 ### Local Usage and Registry Remapping
 
--- a/test/scripts/agent-images/containerfiles/cs9-regular/Containerfile
+++ b/test/scripts/agent-images/containerfiles/cs9-regular/Containerfile
@@ -1,0 +1,111 @@
+# syntax=docker/dockerfile:1.6
+#
+# Multi-stage base image build for flightctl package-mode device images - CentOS Stream 9
+# Targets:
+#   - prebase: package-mode base with repos, tooling, users, services (no flightctl RPMs)
+#   - base:    final base image starting from prebase and layering flightctl RPMs
+#
+
+# ---------- Stage: prebase ----------
+FROM quay.io/centos/centos:stream9 AS prebase
+
+RUN --mount=type=cache,target=/var/cache/dnf \
+    --mount=type=cache,target=/var/lib/dnf \
+    dnf install -y epel-release epel-next-release
+
+RUN --mount=type=cache,target=/var/cache/dnf \
+    --mount=type=cache,target=/var/lib/dnf \
+    dnf install -y python-dotenv
+
+RUN --mount=type=cache,target=/var/cache/dnf \
+    --mount=type=cache,target=/var/lib/dnf \
+    dnf install -y \
+      cloud-init \
+      dnf-plugins-core \
+      firewalld \
+      openssh-server \
+      podman \
+      podman-compose \
+      sudo && \
+    systemctl enable firewalld.service && \
+    systemctl enable podman.service && \
+    systemctl enable sshd.service
+
+RUN useradd -ms /bin/bash user && \
+    echo "user:user" | chpasswd && \
+    echo "user ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
+RUN mkdir -p /usr/lib/flightctl/custom-info.d && \
+    chmod 755 /usr/lib/flightctl/custom-info.d
+
+RUN echo -e '#!/bin/bash\necho "my site"' > /usr/lib/flightctl/custom-info.d/siteName && \
+    echo -e '#!/bin/bash\necho ""' > /usr/lib/flightctl/custom-info.d/emptyValue && \
+    echo -e '#!/bin/bash\necho "no-show"' > /usr/lib/flightctl/custom-info.d/keyNotShown && \
+    chmod 755 /usr/lib/flightctl/custom-info.d/*
+
+ARG SOURCE_GIT_TAG
+ARG SOURCE_GIT_COMMIT
+LABEL org.opencontainers.image.source="https://github.com/flightctl/flightctl"
+LABEL org.opencontainers.image.version="${SOURCE_GIT_TAG}"
+LABEL org.opencontainers.image.revision="${SOURCE_GIT_COMMIT}"
+
+# ---------- Stage: base ----------
+FROM prebase AS base
+
+ARG SOURCE_GIT_TAG
+ARG SOURCE_GIT_COMMIT
+
+ARG RPM_DIR=rpm
+ARG RPM_COPR_REPO=
+ARG RPM_COPR_PACKAGE=flightctl-agent
+
+ARG TRUST_CA_FROM=
+ARG AGENT_CONFIG_FROM=
+ARG AGENT_CERTS_FROM=
+
+COPY --from=variant-context flightctl-e2e.tmpfiles.conf /usr/lib/tmpfiles.d/flightctl-e2e.tmpfiles.conf
+
+RUN --mount=type=bind,from=project-bin,source=.,target=/build-context \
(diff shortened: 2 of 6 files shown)


test/scripts/agent-images/create_agent_images.sh
```diff
@@ -192,6 +203,7 @@ build_variants_and_qcow2() {
 case "$BUILD_TYPE" in
     regular)
         build_base
+        build_variants_and_qcow2
```
galel12: **Must-fix:** `BUILD_TYPE=regular` without `AGENT_OS_ID=cs9-regular` still defaults `AGENT_OS_ID` to `cs9-bootc` (see ~L145), so this path builds a **bootc** base + variants + **bootc** qcow via `qcow2.sh`, not package-mode.

When `BUILD_TYPE=regular`, default or require `AGENT_OS_ID=cs9-regular` (and fail clearly on a mismatched combo). The README already shows both vars; the script should enforce it.

test/scripts/agent-images/scripts/qcow2_regular.sh
```diff
@@ -0,0 +1,94 @@
...
+fi
+
+echo "Preparing package-mode qcow2 at ${OUTPUT_QCOW_PATH}"
+qemu-img convert -O qcow2 "${BASE_CLOUD_IMAGE_PATH}" "${OUTPUT_QCOW_PATH}"
+qemu-img resize "${OUTPUT_QCOW_PATH}" +5G
+
+VIRT_CUSTOMIZE_ARGS=(
+  -a "${OUTPUT_QCOW_PATH}"
+  --run-command "dnf install -y epel-release epel-next-release"
+  --run-command "dnf install -y cloud-init dnf-plugins-core firewalld openssh-server podman podman-compose python-dotenv sudo"
```
galel12, line 129: **Must-fix / design:** Unlike `qcow2.sh` (which converts the built container image), this rebuilds from the CentOS cloud image via `virt-customize`, so the Containerfile and qcow path are independent sources of truth. Drift already exists (e.g. COPR default package `flightctl-agent` in the Containerfile vs `flightctl-agent.el9` here; CA/config/certs ARGs only on the Containerfile path).

Either keep one source of truth, or document clearly that Containerfile = OCI base and qcow = parallel virt-customize path, and keep package lists / COPR defaults identical. Note that agent config/certs are expected via `inject_agent_files_into_qcow.sh` after build if that remains the e2e flow.

test/scripts/agent-images/create_agent_images.sh
```diff
@@ -192,6 +203,7 @@ build_variants_and_qcow2() {
 case "$BUILD_TYPE" in
     regular)
         build_base
```
galel12, line 304: **Should-fix:** Calling full `build_variants_and_qcow2` for package-mode looks heavier than EDM-4761 needs. AC/design ask for a usable base + cloud qcow; most variants exist for `spec.os.image` / greenboot / MicroShift flows that package-mode cannot use.

Consider building base + qcow only for `regular` (skip variants/bundle), or an explicit minimal variant set, to avoid slow unused image builds.

test/scripts/agent-images/scripts/build_and_qcow2.sh
```diff
@@ -107,7 +107,11 @@ if [ "${SKIP_QCOW_BUILD}" != "true" ]; then
   (
     set -euo pipefail
     echo "Building qcow2 for ${OS_ID}"
-    OUTPUT_DIR="${QCOW2_OUTPUT_DIR}" "${SCRIPT_DIR}/qcow2.sh" 2>&1 | tee "${qcow2_log}"
+    if [[ "${OS_ID}" == *-regular ]]; then
```
galel12, line 126: **Should-fix:** `EXCLUDE_VARIANTS="v7 v11 v12"` is only set in `create_agent_images.sh`. A direct `OS_ID=cs9-regular ./build_and_qcow2.sh` will still try to build bootc/greenboot-oriented variants (the variants job above runs unconditionally).

Mirror the `cs9-regular` exclusion next to the existing `cs10-bootc` handling in this file. Prefer space-separated values — `build.sh` splits on whitespace, so `v7,v12` is treated as one token.

test/scripts/agent-images/scripts/build.sh
```diff
@@ -125,13 +125,17 @@ case "${AGENT_OS_ID}" in
     CONTAINERFILE_DIR="${BASE_DIR}/containerfiles/cs9-bootc${DISTRO_SUFFIX}"
     OS_ID="cs9-bootc"
     ;;
+  cs9-regular)
+    CONTAINERFILE_DIR="${BASE_DIR}/containerfiles/cs9-regular${DISTRO_SUFFIX}"
```
galel12, line 134: **Should-fix:** With `DISTRO=redhat` / `RHEM`, this resolves to `containerfiles/cs9-regular-redhat/`, which does not exist. Fail with a clear unsupported error (or document that redhat distro is bootc-only for now).

test/scripts/agent-images/scripts/qcow2_regular.sh
```diff
@@ -0,0 +1,94 @@
...
+
+require_command curl
+require_command qemu-img
+require_command virt-customize
+
+mkdir -p "${CLOUD_IMAGE_CACHE_DIR}" "${OUTPUT_QCOW_DIR}"
+
+if [ ! -f "${BASE_CLOUD_IMAGE_PATH}" ]; then
+  echo "Downloading CentOS Stream 9 cloud image from ${BASE_CLOUD_IMAGE_URL}"
+  curl --fail --location --retry 3 --output "${BASE_CLOUD_IMAGE_PATH}" "${BASE_CLOUD_IMAGE_URL}"
```
galel12: **Nit:** Download into a temp file and `mv` into place only after `curl` succeeds, so a failed/partial download does not poison `BASE_CLOUD_IMAGE_PATH` for later runs.

test/scripts/agent-images/scripts/qcow2_regular.sh
```diff
@@ -0,0 +1,94 @@
...
+qemu-img resize "${OUTPUT_QCOW_PATH}" +5G
+
+VIRT_CUSTOMIZE_ARGS=(
+  -a "${OUTPUT_QCOW_PATH}"
+  --run-command "dnf install -y epel-release epel-next-release"
+  --run-command "dnf install -y cloud-init dnf-plugins-core firewalld openssh-server podman podman-compose python-dotenv sudo"
+  --run-command "systemctl enable firewalld.service"
+  --run-command "systemctl enable podman.service"
+  --run-command "systemctl enable sshd.service"
+  --run-command "useradd -ms /bin/bash user || true"
```
galel12: **Nit:** Please add a one-line comment that `|| true` is intentional for idempotency if `user` already exists on the cloud image.

test/scripts/agent-images/scripts/qcow2_regular.sh
```diff
@@ -0,0 +1,94 @@
...
+  --run-command "echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers"
+  --run-command "mkdir -p /usr/lib/flightctl/custom-info.d"
+  --run-command "printf '#!/bin/bash\necho \"my site\"\n' > /usr/lib/flightctl/custom-info.d/siteName"
+  --run-command "printf '#!/bin/bash\necho \"\"\n' > /usr/lib/flightctl/custom-info.d/emptyValue"
+  --run-command "printf '#!/bin/bash\necho \"no-show\"\n' > /usr/lib/flightctl/custom-info.d/keyNotShown"
+  --run-command "chmod 755 /usr/lib/flightctl/custom-info.d/*"
+)
+
+if [ -n "${RPM_COPR_REPO}" ]; then
+  package_name="${RPM_COPR_PACKAGE:-flightctl-agent.el9}"
```
galel12: **Nit:** Default here is `flightctl-agent.el9`, but the Containerfile defaults `RPM_COPR_PACKAGE` to `flightctl-agent`. Align the defaults (preferably with whatever `create_agent_images.sh` exports when `FLIGHTCTL_RPM` is set).

galel12 (changes requested):
Thanks for landing the package-mode image infra — Containerfile + `cs9-regular` wiring + `virt-customize` qcow is the right shape for EDM-4761 / §4.1.6.

**Requesting changes** on two footguns (BUILD_TYPE/`AGENT_OS_ID` coupling, and Containerfile vs qcow drift). The rest are should-fix / nits.

Please ignore CodeRabbit’s “remove `user:user` / switch to UBI / non-root” — that credential matches the existing bootc e2e images and is intentional for test VMs.

galel12:
@eldar101 on the new `USER root` / `USER user` churn across `variants/v2`…`v12`:

These look like a follow-on from CodeRabbit’s “non-root final user” advice on the package-mode base. Once `cs9-regular` ends as `USER user`, every variant has to flip to root for `/etc` writes and then back again.

I’d push back on that direction for these images:
1. These are local/CI **e2e test images**, not production containers. Bootc e2e already stays root and uses `user:user` on purpose.
2. The `USER` flips are on **shared** variants, so bootc builds pick up a final `USER user` too — an unrelated behavior change outside EDM-4761.
3. It adds a lot of noise without a product/security requirement for this story.

My preference: keep the package-mode base like bootc (no final `USER user`), and drop the variant `USER root`/`USER user` pairs.

Your call though — if you have a reason to keep non-root defaults for package-mode specifically, happy to hear it; just please avoid changing bootc variant defaults as a side effect.

test/scripts/agent-images/containerfiles/cs9-regular/Containerfile
```diff
@@ -0,0 +1,111 @@
+# syntax=docker/dockerfile:1.6
+#
+# Multi-stage base image build for flightctl package-mode device images - CentOS Stream 9
+# Targets:
+#   - prebase: package-mode base with repos, tooling, users, services (no flightctl RPMs)
+#   - base:    final base image starting from prebase and layering flightctl RPMs
+#
+
+# ---------- Stage: prebase ----------
+FROM quay.io/centos/centos:stream9 AS prebase
```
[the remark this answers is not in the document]
galel12: @coderabbitai thanks — pushing back on this one.

These are e2e/test device images (same family as the existing bootc agent images), not production runtime containers. Matching the bootc path — root default image user, plus the intentional test `user` account — is the right baseline for SSH/sudo/systemd exercises in CI.

Requiring a non-root final `USER` on the package-mode base forced `USER root` / `USER user` wrappers across all shared variants, which also changes bootc variant defaults as a side effect. That is heavier and riskier than the security win for disposable test artifacts.

Please treat “UBI + non-root final USER” as out of scope / not applicable for this test-image path unless project maintainers explicitly want that policy change.

test/scripts/agent-images/containerfiles/cs9-regular/Containerfile
```diff
@@ -0,0 +1,111 @@
...
+      podman \
+      podman-compose \
+      sudo && \
+    systemctl enable firewalld.service && \
+    systemctl enable podman.service && \
+    systemctl enable sshd.service
+
+RUN useradd -ms /bin/bash user && \
+    echo "user:user" | chpasswd && \
+    echo "user ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
```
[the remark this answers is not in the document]
galel12: @coderabbitai same pushback on removing `user:user` / passwordless sudo here.

That credential is intentional and already present on the bootc e2e images so tests can log in and escalate. Please don’t treat it as a secret leak for this test-image infrastructure — changing it would break the e2e harness patterns unless we redesign auth injection end-to-end (out of scope for EDM-4761).

test/scripts/agent-images/create_agent_images.sh
```diff
@@ -189,9 +227,34 @@ build_variants_and_qcow2() {
     fi
 }
 
+build_regular_qcow2() {
+    echo "Building package-mode base and qcow2 for OS_ID=${OS_ID}"
+
+    local skip_qcow="false"
+    if qcow_is_up_to_date "${OS_ID}"; then
```
galel12: **Nit / non-blocking:** `qcow_is_up_to_date` still decides “skip qcow rebuild” by comparing the qcow mtime to the **OCI base image** (`base-${OS_ID}-${TAG}`).

That fit bootc, where qcow is produced from that OCI image. For package-mode the qcow is built independently from the CentOS cloud image via `virt-customize`, so an unchanged OCI base does not mean the qcow customization is still current (e.g. `qcow2_regular.sh` or cloud image URL changed).

Not a merge blocker — workaround is delete `bin/output/qcow2/disk.qcow2` (or the touch file) and rebuild. A follow-up could use a regular-specific freshness check (script/cloud-image inputs) or always rebuild qcow for `BUILD_TYPE=regular`.

test/scripts/agent-images/containerfiles/cs9-regular/Containerfile
```diff
@@ -0,0 +1,111 @@
...
+        dnf install -y ${RPM_COPR_PACKAGE}; \
+      else \
+        echo "Installing local RPMs from ${RPM_DIR}"; \
+        cp /build-context/${RPM_DIR}/flightctl-agent-*.rpm \
+           /build-context/${RPM_DIR}/flightctl-selinux-*.rpm /tmp/rpms/; \
+        ls -l /tmp/rpms; \
+        dnf install -y /tmp/rpms/*.rpm; \
+        rm -rf /tmp/rpms; \
+      fi; \
+      systemctl enable flightctl-agent.service'
```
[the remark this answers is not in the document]
galel12: @coderabbitai thanks — acknowledging the parallel, but I’m not treating this as the same fix target as `qcow2_regular.sh`.

`qcow2_regular.sh` builds guest shell commands from host env vars in a new script path, so allowlisting there was reasonable hygiene.

This Containerfile `RUN` is the same build-arg pattern already used by `cs9-bootc` / `cs10-bootc`. The args come from our local/CI wrapper (`create_agent_images.sh`), not an untrusted remote input surface. Pasting `case` validation into only `cs9-regular` would be asymmetric noise without changing the real trust model.

If we want stronger consistency later, validate once in `create_agent_images.sh` before passing build-args / env (covers OCI + qcow). Out of scope / not a merge blocker for EDM-4761.
cc @eldar101

galel12 (approved):
LGTM

test/scripts/agent-images/scripts/build_and_qcow2.sh
```diff
@@ -27,6 +27,12 @@ TAG="${TAG:-$SOURCE_GIT_TAG}"
 IMAGE_REPO="${IMAGE_REPO:-quay.io/flightctl/flightctl-device}"
 DO_PUSH=false
 SKIP_QCOW_BUILD="${SKIP_QCOW_BUILD:-false}"
+SKIP_VARIANTS_BUILD="${SKIP_VARIANTS_BUILD:-false}"
+SKIP_VARIANTS_BUILD_WAS_SET=false
+
+if [ -n "${SKIP_VARIANTS_BUILD+x}" ]; then
```
galel12: @eldar101 intent looks right, but this defaulting logic doesn’t work as written.

`SKIP_VARIANTS_BUILD="${SKIP_VARIANTS_BUILD:-false}"` always assigns the variable, so `${SKIP_VARIANTS_BUILD+x}` is always set and `SKIP_VARIANTS_BUILD_WAS_SET` becomes `true` even when the caller never passed it. The later `*-regular` auto-skip then never runs — direct `build_and_qcow2.sh --os-id cs9-regular` still builds variants.

Simpler:

```bash
if [[ "${OS_ID}" == *-regular ]]; then
  SKIP_VARIANTS_BUILD="${SKIP_VARIANTS_BUILD:-true}"
else
  SKIP_VARIANTS_BUILD="${SKIP_VARIANTS_BUILD:-false}"
fi
```

That makes regular qcow-only by default while still allowing an explicit `SKIP_VARIANTS_BUILD=false` override. Worth a short README note that direct `build_and_qcow2.sh` for `*-regular` is qcow-only too.

galel12 (commented):
One leftover make/docs footgun from the re-check — prior must-fixes look good.

test/scripts/agent-images/README.md
```diff
@@ -22,16 +25,21 @@ AGENT_OS_ID=cs10-bootc make e2e-agent-images
...
-2. **Variants and qcow2**: Built using `scripts/build_and_qcow2.sh`
+2. **Bootc path**: `scripts/build_and_qcow2.sh` builds variants and a bootc qcow2
+3. **Package-mode path**: `BUILD_TYPE=regular AGENT_OS_ID=cs9-regular` builds the package-mode base image and qcow2 only
 
-The build process automatically handles different OS flavors (cs9-bootc, cs10-bootc)
+The build process automatically handles different OS flavors (cs9-bootc, cs9-regular, cs10-bootc)
 and RPM source detection (local, COPR, or Brew registry).
 
+`BUILD_TYPE=regular` is a package-mode-only path. If `AGENT_OS_ID` is omitted it defaults to
+`cs9-regular`; any other `AGENT_OS_ID` with `BUILD_TYPE=regular` fails fast.
```
galel12: nit: this is true for `create_agent_images.sh`, but not for `make e2e-agent-images`.

Make still defaults `AGENT_OS_ID ?= cs9-bootc` and skips create when that bootc bundle already exists. So `BUILD_TYPE=regular make e2e-agent-images` (no AGENT_OS_ID) can silently stay on the bootc path.

Either couple make defaults when `BUILD_TYPE=regular`, or tighten this to say the make entrypoint always needs `AGENT_OS_ID=cs9-regular` (the example above already does).

galel12 (approved):

test/scripts/agent-images/scripts/build_and_qcow2.sh
```diff
@@ -70,44 +79,53 @@ mkdir -p "${LOG_DIR}"
 variants_log="${LOG_DIR}/variants.log"
 qcow2_log="${LOG_DIR}/qcow2.log"
 
-echo "Building variants and creating bundle for ${OS_ID}"
+echo "Building artifacts for ${OS_ID}"
```
itzikb-redhat, line 84: why the change?

test/scripts/agent-images/create_agent_images.sh
```diff
@@ -87,16 +103,66 @@ qcow_is_up_to_date() {
...
+    printf '%s\n' "${os_id}" > "${qcow_dir}/disk.qcow2.os-id"
+}
+
 # Build extra flags for RPM source
 BUILD_ARGS=""
+RPM_DIR="${RPM_DIR:-rpm}"
+RPM_COPR_REPO="${RPM_COPR_REPO:-}"
+RPM_COPR_PACKAGE="${RPM_COPR_PACKAGE:-}"
+
+safe_brew_build_url() {
```
itzikb-redhat, line 128: Please add comments to explain the function(s) intent

itzikb-redhat (commented):
nits

test/scripts/agent-images/scripts/build_and_qcow2.sh
eldar101: This script now handles both:
bootc: variants + bundle + qcow
regular: qcow-only by default
So I made the top banner generic, and the later log lines still say exactly what is being built.

itzikb-redhat (approved):

PR #3335, opened 2026-08-04
NO-ISSUE: Stop systemd services and reset failed status after an app …

itzikb-redhat:
Stop systemd services and reset failed status after an app stop

Lifecycle Stop previously only stopped the Quadlet app target. Dependent units (for example VM virt-launcher compute services) were left to stop asynchronously and often exited non-zero when killed by systemd, so they remained listed under Failed Units even though the app was intentionally stopped.
Reuse the same stop path already used on remove: stop the target, stop loaded dependent services, then reset-failed. Unit files and Podman resources are still kept so start/restart can bring the app back without redeploying.

### DIFF
--- a/internal/agent/device/applications/lifecycle/quadlet.go
+++ b/internal/agent/device/applications/lifecycle/quadlet.go
@@ -146,36 +146,47 @@ func (q *Quadlet) add(ctx context.Context, action Action, systemctl systemd.Mana
 	return nil
 }
 
-func (q *Quadlet) remove(ctx context.Context, action Action, systemctl systemd.Manager) error {
+// stopUnits stops the application target and all loaded dependent services, then
+// clears any failed state left by non-zero exits. That is common when stopping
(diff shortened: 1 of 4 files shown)


itzikb-redhat:
```
$ flightctl get dev d5lai7vq7h1q9goe9gfmh106m595aan8bdjihgrb0llsi4vit5a0 -o yaml | yq .status.applications
- appType: vm
  embedded: false
  name: test-vm
  ready: 0/4
  restarts: 0
  status: Stopped
$ flightctl console dev d5lai7vq7h1q9goe9gfmh106m595aan8bdjihgrb0llsi4vit5a0
$ sudo -s
bash-5.1# systemctl --failed
  UNIT LOAD ACTIVE SUB DESCRIPTION
0 loaded units listed.
```

asafbennatan (approved):
LGTM

asafbennatan (approved):
LGTM

PR #3338, opened 2026-08-04
EDM-4777: Register encryption metrics collector before canary validation

asafbss:
The encryption canary validations metric was never recorded because the metrics collector was created after canary validation already ran during startup. Move collector creation to right after InitGlobalEncryption so the recorder is available when canaryservice.InitEncryption validates.

### DIFF
--- a/cmd/flightctl-api/main.go
+++ b/cmd/flightctl-api/main.go
@@ -77,6 +77,15 @@ func main() {
 		log.Fatalf("initializing encryption: %v", err)
 	}
 
+	var encCollector prometheus.Collector
+	if cfg.Metrics != nil && cfg.Metrics.Enabled {
+		if encMgr := encryption.GlobalManager(); encMgr != nil {
+			ec := encmetrics.NewEncryptionCollector(encMgr)
+			encMgr.SetMetricsRecorder(ec)
+			encCollector = ec
+		}
+	}
+
 	log.Println("Initializing data store")
 	db, err := store.InitDB(cfg, log)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/flightctl-api/main.go
```diff
@@ -77,6 +77,13 @@ func main() {
 		log.Fatalf("initializing encryption: %v", err)
 	}
 
+	var encCollector prometheus.Collector
+	if encMgr := encryption.GlobalManager(); encMgr != nil {
```
asafbennatan: **Design (MEDIUM):** Collector creation and `SetMetricsRecorder` now run whenever `GlobalManager()` is non-nil, outside `if cfg.Metrics != nil && cfg.Metrics.Enabled`. That fixes the startup ordering bug, but it also changes prior behavior: every encrypt/decrypt/canary path updates in-memory Prometheus counters even when the metrics server will never start or scrape them. Other collectors in these mains remain gated on metrics being enabled.

Keep the early placement (still before `canaryservice.InitEncryption`), but gate creation on metrics being enabled:

```go
var encCollector prometheus.Collector
if cfg.Metrics != nil && cfg.Metrics.Enabled {
	if encMgr := encryption.GlobalManager(); encMgr != nil {
		ec := encmetrics.NewEncryptionCollector(encMgr)
		encMgr.SetMetricsRecorder(ec)
		encCollector = ec
	}
}
```

Same pattern in `cmd/flightctl-worker/main.go`. The later `if encCollector != nil` registration can stay as written.

cmd/flightctl-worker/main.go
```diff
@@ -56,6 +56,13 @@ func main() {
 		log.Fatalf("initializing encryption: %v", err)
 	}
 
+	var encCollector prometheus.Collector
+	if encMgr := encryption.GlobalManager(); encMgr != nil {
```
asafbennatan: **Design (MEDIUM):** Same as api — please gate early collector creation on `cfg.Metrics != nil && cfg.Metrics.Enabled` while keeping it before `canaryservice.InitEncryption`, so metrics-disabled startups don't attach a recorder that will never be scraped.

asafbennatan (commented):
One design note on early collector registration — the ordering fix looks right; suggesting we keep it gated on metrics being enabled.

asafbennatan (approved):
LGTM

PR #3339, opened 2026-08-04
EDM-4768: add package-mode e2e suite

galel12:
## Summary
- Stacked on [#3336](https://github.com/flightctl/flightctl/pull/3336) (`edm-4765-package-mode-ci`): activates the package-mode e2e suite and adds the Linux-only tests from Eldar’s [#3323](https://github.com/flightctl/flightctl/pull/3323).
- Cherry-picked `EDM-4768: add package-mode e2e suite` onto our CI branch. Eldar’s `EDM-4765: activate package-mode CI leg` was already covered by #3336 (empty cherry-pick after conflict resolution; kept our `resolve-go-e2e-dirs` action).
- Adds `test/e2e/package_mode` coverage (enroll/`osMode=package`, config+podman app, mixed-fleet reject) and a dedicated CI matrix row that pairs `cs9-regular` + `cs9-bootc` qcow artifacts.

## Test plan
- [ ] Wait for #3336 to merge (or run CI on this stack as-is)
- [ ] Confirm `cs9-regular` agent-image build stays green
- [ ] Confirm package-mode e2e discovery + suite row runs
- [ ] Confirm package-mode sanity specs pass

Made with [Cursor](https://cursor.com)

PR #3342, opened 2026-08-04
EDM-4768: add package-mode e2e suite

galel12:
## Summary
- Activates the package-mode e2e suite on top of [#3336](https://github.com/flightctl/flightctl/pull/3336) (merged).
- Ports Eldar’s suite from [#3323](https://github.com/flightctl/flightctl/pull/3323): `test/e2e/package_mode` (enroll/`osMode=package`, config+podman app, mixed-fleet reject) plus the CI matrix row pairing `cs9-regular` + `cs9-bootc` qcow artifacts.
- Replaces [#3339](https://github.com/flightctl/flightctl/pull/3339) (same change; branch renamed from `EDM-4761` → `EDM-4768`).

## Test plan
- [ ] Confirm `cs9-regular` agent-image build stays green
- [ ] Confirm package-mode e2e discovery + suite row runs
- [ ] Confirm package-mode sanity specs pass


galel12:
@coderabbitai full review

asafbennatan (commented):
## Code review

Thanks for landing the suite activation on top of #3336. Directionally the scenarios (enroll/`osMode=package`, config+app without `spec.os.image`, mixed-fleet reject) are the right coverage — but the **runtime chosen for them looks heavier than the feature needs**.

### Design feedback: container (or simulator), not a dedicated VM/qcow shard

Package-mode detection is PATH-based: if neither `bootc` nor `rpm-ostree` is present, the agent reports `osMode=package`. That means enroll, capability reporting, config/app apply, and the agent-side `os.image` reject do **not** require a full libvirt guest or a dedicated package-mode disk image.

Given that:

1. **Prefer a container host for package-mode e2e** (agent in a non-bootc / non-rpm-ostree environment), not custom libvirt VMs from `cs9-regular` qcow. The VM + qcow path adds a lot of CI/local overhead for behavior that is fundamentally “agent without image-based OS tooling.”
2. **A dedicated matrix shard pairing `cs9-regular` + `cs9-bootc` qcows is hard to justify** if package-mode does not need its own qcow. Keep package-mode coverage on a lighter runtime; if mixed-fleet still needs an image-mode peer that actually converges on `spec.os.image`, that half can stay on the existing bootc e2e path — the package-mode peer still does not need a second qcow.
3. **The dedicated package-mode qcow2 generation / CI wiring should also be removed** (the `cs9-regular` qcow path / `qcow2_regular.sh` / matrix `package_mode_qcow_*` inputs from #3336, and this PR’s suite that consumes them). Keeping an OCI `cs9-regular` base for package contents may still be useful; generating and staging a bootable qcow solely for these tests is the part that looks unnecessary.
4. **Simulator is a reasonable alternative** for the control-plane-facing parts (enroll, reported `osMode`, reject/OutOfDate), if extended to report package-mode capabilities — but a real agent-in-container is the better fit when we also want config + podman app on a package-mode device.

### Secondary notes (only if the VM approach is kept)

These are lower priority if the suite is redesigned around a container:

- Outer `Ordered` on the top-level `Describe` couples the normal-path and mixed-fleet contexts, so a normal-path failure skips mixed-fleet. Prefer per-context `Ordered` only (or `ContinueOnFailure` if outer serialization is intentional).
- Context `BeforeAll` performs API mutations before suite `BeforeEach` login — login should happen before fleet create/enroll.
- Suite-specific README under `test/e2e/package_mode/` would match existing e2e conventions if this suite remains special-cased.

### Suggested direction

Rework package-mode e2e onto a **container** (not VM), drop the dedicated package-mode qcow build/staging and the separate dual-qcow matrix shard, and keep bootc VM usage only where image-mode OS convergence is actually under test.

Happy to discuss trade-offs if there is a hard requirement I am missing for a real package-mode VM.

galel12:
@asafbennatan Thanks for the detailed review. I've reworked the suite based on your feedback:

**Changes made:**
- Switched from VM/qcow to testcontainer approach - the agent now runs in a privileged cs9-regular container with systemd as init
- Deleted `qcow2_regular.sh` and all `package_mode_qcow_*` CI wiring
- Removed dedicated matrix shard - package-mode tests now run alongside regular e2e tests
- Kept cs9-regular OCI build (with `skip_qcow_build`) for the testcontainer base image

**Test coverage:**
1. **Normal operation** - enroll with `osMode=package`, config + podman app deployment
2. **Spec.os.image rejection** - fleet adds `spec.os.image` → device stays OutOfDate, rejection in status/logs
3. **Image-mode convergence** - skipped (needs bootc VM; already covered by existing bootc e2e)

**What the testcontainer provides:**
- Real agent in a non-bootc environment (PATH-based detection → `osMode=package`)
- Systemd init for the flightctl-agent service
- Privileged mode with nested podman for application deployment
- Config/cert injection via volume mounts

This should address the overhead concern while preserving the behavioral coverage.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	Context("normal package-mode operation", Ordered, func() {
+		var (
+			deviceID         string
+			fleetName        string
+			selector         v1beta1.LabelSelector
+			packageModeAgent *e2e.PackageModeAgent
+		)
+
+		BeforeAll(func() {
```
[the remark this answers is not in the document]
galel12: The `Ordered` context in Ginkgo guarantees sequential execution and shared state across specs. `BeforeAll` creates resources once for all specs in the context, which is intentional here since the three specs form a logical sequence (enroll, deploy config, check app health) sharing one device and fleet by design.

The suite-level `AfterEach` calls `CleanUpAllTestResources`, but this only cleans resources registered via `harness.Track*()` during individual `It` blocks, not the `BeforeAll` fleet/device which are cleaned in `AfterAll`. This is the standard pattern used across other e2e suites in this project.

test/harness/e2e/package_mode_agent.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	if _, err := os.Stat(agentCertsDir); err != nil {
+		return nil, fmt.Errorf("agent certs dir not found at %s: %w", agentCertsDir, err)
+	}
+
+	req := testcontainers.ContainerRequest{
+		Image:        GetPackageModeAgentImage(),
+		Name:         PackageModeAgentContainerName,
+		ExposedPorts: []string{"22/tcp"},
+		Privileged:   true,
```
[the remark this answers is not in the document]
galel12: The package-mode suite runs as a single `Ordered` context with one container at a time. Ginkgo does not distribute specs within an `Ordered` context across parallel nodes, so the container name collision scenario described cannot occur. The suite is inherently serial. Adding worker-ID complexity would be premature for a non-existent problem.

.github/workflows/run-e2e-tests.yaml
```diff
@@ -133,25 +113,16 @@ jobs:
           path: test-artifacts
           merge-multiple: true
 
-      - name: Download image-mode qcow artifact
-        if: ${{ inputs.image_mode_qcow_os_id != '' }}
+      - name: Download cs9-regular bundle for package-mode testcontainers
```
eldar101, line 116: continue-on-error: true here makes the package-mode suite fall back to whatever base-cs9-regular image is already present or can be pulled, so the tests stop proving the image built in this workflow. This download should fail the job for package-mode runs instead of silently degrading.`

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+			Eventually(func(g Gomega) {
+				logs, err := packageModeAgent.GetAgentLogs(context.Background())
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(logs).To(ContainSubstring(packageModeExpectedRejectText))
+			}, testutil.LONG_TIMEOUT, packageModePollInterval).Should(Succeed())
+		})
+
+		It("image-mode device convergence is covered by existing bootc e2e", Label("90149", "sanity", "slow"), func() {
+			Skip("Image-mode convergence requires bootc VM; covered by existing bootc e2e tests")
```
eldar101: This skips 90149 entirely, but EDM-4768 still requires “image-mode device in the same mixed fleet converges normally”. Existing bootc coverage is not the same as proving that behavior in this mixed-fleet scenario.

test/harness/e2e/harness_container.go
```diff
@@ -0,0 +1,57 @@
...
+// agent config and certs for use with testcontainer-based e2e tests.
+func PrepareAgentConfigForContainer(ctx context.Context) error {
+	topLevel := util.GetTopLevelDir()
+	scriptPath := filepath.Join(topLevel, "test", "scripts", "agent-images", "prepare_agent_config.sh")
+
+	if _, err := os.Stat(scriptPath); err != nil {
+		return fmt.Errorf("missing prepare_agent_config.sh at %s: %w", scriptPath, err)
+	}
+
+	cmd := exec.CommandContext(ctx, scriptPath)
```
eldar101, line 24: Please run prepare_agent_config.sh with a bounded context (exec.CommandContext) from the suite. Right now this path is unbounded, so a hung script can stall the worker indefinitely.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,349 @@
...
+			selector = v1beta1.LabelSelector{
+				MatchLabels: &map[string]string{
+					packageModeFleetLabelKey: fleetName,
+				},
+			}
+
+			agentConfigDir := e2e.GetAgentConfigDir()
+			Expect(e2e.AgentConfigDirExists()).To(BeTrue(), "agent config dir must exist")
+
+			packageModeAgent, err = e2e.StartPackageModeAgent(context.Background(), agentConfigDir)
```
eldar101: StartPackageModeAgent is called with context.Background(), so container startup is effectively unbounded at the call site. Please pass the harness/spec context with a timeout instead.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,349 @@
...
+				g.Expect(device).ToNot(BeNil())
+				g.Expect(device.Spec).ToNot(BeNil())
+				g.Expect(device.Spec.Os).To(BeNil())
+			}, packageModeEnrollTimeout, packageModePollInterval).Should(Succeed())
+
+			Expect(harness.WaitForApplicationStatus(deviceID, packageModeContainerAppName, v1beta1.ApplicationStatusRunning, testutil.TIMEOUT, testutil.POLLING)).ToNot(HaveOccurred())
+			Expect(harness.WaitForApplicationSummary(deviceID, testutil.TIMEOUT, testutil.POLLING, v1beta1.ApplicationsSummaryStatusHealthy)).ToNot(HaveOccurred())
+
+			Eventually(func(g Gomega) {
+				output, err := packageModeAgent.RunSSH(context.Background(), []string{"cat", packageModeConfigPath})
```
eldar101: RunSSH is called with context.Background() here, so this poll can hang forever despite the Eventually timeout. Please thread a bounded test context through these SSH/log helpers.

.github/workflows/run-e2e-tests.yaml
galel12: Fixed in the latest push — `continue-on-error` is removed. The job will now fail if the bundle download fails.

test/harness/e2e/harness_container.go
galel12: Fixed in the latest push — `PrepareAgentConfigForContainer` now accepts a `context.Context` and uses `exec.CommandContext`. The caller in `BeforeSuite` passes a 2-minute timeout.

test/e2e/package_mode/package_mode_test.go
galel12: Agreed this is a gap worth acknowledging. The skip is intentional for now because proving image-mode convergence in a mixed-fleet context requires orchestrating both a testcontainer (package-mode device) AND a bootc VM (image-mode device) simultaneously within the same test. That's architecturally different from what this suite can provide.

The natural home for that scenario is the bootc e2e suite, where the VM already exists — adding a package-mode testcontainer peer there would complete the mixed-fleet picture. I'll track this as a follow-up item rather than blocking this PR on it.

test/harness/e2e/package_mode_agent.go
```diff
@@ -0,0 +1,234 @@
...
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/flightctl/flightctl/test/harness/containers"
+	"github.com/sirupsen/logrus"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	PackageModeAgentContainerName = "flightctl-package-mode-agent"
```
eldar101, line 24: If a CI retry starts before the old container is
  cleaned up, or if multiple pipeline runs share a runner, the second GenericContainer call (line 82, Reuse: false) will fail
  because the name is taken. Append a unique suffix — e.g. fmt.Sprintf("flightctl-package-mode-agent-%d", time.Now().UnixNano()) —
  or use GinkgoParallelProcess(). Remove the constant and generate the name in StartPackageModeAgent.

test/harness/e2e/package_mode_agent.go
```diff
@@ -0,0 +1,234 @@
...
+	commands := []string{
+		"id -u flightctl >/dev/null 2>&1 || useradd --create-home --user-group flightctl",
+		"mkdir -p /var/lib/systemd/linger && touch /var/lib/systemd/linger/flightctl",
+		"mkdir -p /home/flightctl/.config/containers/systemd",
+		"mkdir -p /home/flightctl/.config/systemd/user",
+		"mkdir -p /home/flightctl/.local",
+		"chown -R flightctl:flightctl /home/flightctl",
+	}
+	for _, cmd := range commands {
+		exitCode, _, err := a.Container.Exec(ctx, []string{"sh", "-c", cmd})
```
eldar101: The _ discards the output reader. When a setup command fails (exitCode != 0), the error message says "exit code N" but not why.
  Read the output and include it.

test/harness/e2e/harness_container.go
```diff
@@ -0,0 +1,57 @@
...
+	cmd.Dir = topLevel
+	cmd.Env = os.Environ()
+
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("running prepare_agent_config.sh: %w", err)
```
eldar101: Include the output in the error so CI failures are diagnosable.

test/harness/e2e/package_mode_agent.go
```diff
@@ -0,0 +1,234 @@
...
+func (a *PackageModeAgent) Stop(ctx context.Context) error {
+	if a.Container != nil {
+		return a.Container.Terminate(ctx)
+	}
+	return nil
+}
+
+// GetAgentLogs returns the flightctl-agent journal logs.
+func (a *PackageModeAgent) GetAgentLogs(ctx context.Context) (string, error) {
+	exitCode, reader, err := a.Container.Exec(ctx, []string{"journalctl", "-u", "flightctl-agent", "--no-pager"})
```
eldar101: No --since or -n bound. In Eventually loops with LONG_TIMEOUT, this re-reads the full agent journal every 5  seconds, getting slower as logs grow. Add -n 500 or accept a since parameter to scope the read.

test/harness/e2e/vm/vm.go
```diff
@@ -153,10 +153,27 @@ func (v *TestVM) sshCommandWithUserContext(ctx context.Context, inputArgs []stri
...
 }
 
+// redactSSHCommandArgs masks the sshpass password argument before logging the SSH command.
+func redactSSHCommandArgs(args []string) string {
+	if len(args) == 0 {
+		return ""
+	}
+
+	redacted := append([]string(nil), args...)
+	for i := 0; i < len(redacted)-1; i++ {
```
eldar101, line 167: The loop scans all args looking for -p when the sshpass password flag is always at a fixed position (index 1). Replace the loop
  with a direct index check — clearer, and can't accidentally redact the SSH -p <port> if the break were ever removed.

.github/workflows/run-e2e-tests.yaml
```diff
@@ -166,42 +137,43 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          BUNDLE="bin/agent-artifacts/agent-images-bundle-cs9-regular.tar"
+          if [ -f "${BUNDLE}" ]; then
+            echo "Loading cs9-regular images for package-mode testcontainers..."
+            podman load -i "${BUNDLE}"
+            echo "✓ cs9-regular images loaded into podman"
+            podman images | grep cs9-regular || true
+          else
+            echo "No cs9-regular bundle found; package-mode tests may be skipped"
```
eldar101: When the bundle file doesn't exist, the step prints a message but exits 0. Combined with the download step issue you already
  found, this creates a double-silent pass-through. At minimum emit echo "::warning::No cs9-regular bundle..." so the CI run shows
  a visible annotation, or fail the step when inputs.os_id == 'cs9-bootc' (the only case where this step runs).

test/e2e/package_mode/package_mode_test.go
galel12: Fixed in the latest push — both `StartPackageModeAgent` call sites now use `context.WithTimeout(context.Background(), 5*time.Minute)`, which bounds the container creation + SSH readiness + agent service wait.

test/e2e/package_mode/package_mode_test.go
galel12: Fixed — the `RunSSH` call inside the Eventually loop now uses `context.WithTimeout(context.Background(), 30*time.Second)` per iteration, so a hung SSH connection cannot block the poll loop indefinitely.

test/harness/e2e/package_mode_agent.go
galel12: Fixed — `setupFlightctlUser` now reads the output reader on failure and includes it in the error message for diagnosability.

test/harness/e2e/harness_container.go
galel12: Fixed — `PrepareAgentConfigForContainer` now includes the full script output in the error message.

test/harness/e2e/package_mode_agent.go
galel12: Fixed — `GetAgentLogs` now uses `journalctl -n 500` to bound the read.

.github/workflows/run-e2e-tests.yaml
galel12: Fixed — the step now emits `::warning::` when the bundle is missing, providing a visible CI annotation.

test/harness/e2e/vm/vm.go
galel12: This is in `vm.go` which is not modified by this PR. The sshpass password is always at index 1, agreed the loop is overkill, but refactoring that is out of scope here.

test/harness/e2e/package_mode_agent.go
galel12: The suite runs as a single Ginkgo `Ordered` context — specs execute serially within one process. Ginkgo does not distribute specs within an Ordered context across parallel nodes, so container name collisions cannot happen at runtime. CI retries get a fresh pod/runner anyway. Adding uniqueness adds complexity for a non-existent problem in this architecture.

test/AGENTS.md
```diff
@@ -20,6 +20,7 @@ Defined in **test/test.mk** (included from root Makefile). Coverage reports go t
   - **Full flow (deploy + e2e):** `make e2e-test` – deploys cluster, builds e2e agent images, prepares qcow2, runs e2e.  
   - **Cluster already up:** `make in-cluster-e2e-test` – skips deploy, runs e2e.  
   - **Filter:** `GO_E2E_DIRS=test/e2e/agent`, `GINKGO_FOCUS="description"`, `GINKGO_PROCS=N`.  
+  - **Package-mode CI/local scope:** `GO_E2E_DIRS=./test/e2e/package_mode` runs the dedicated package-mode suite. Mixed-fleet coverage also needs `E2E_PACKAGE_MODE_QCOW` and `E2E_IMAGE_MODE_QCOW` to point at injected package-mode and image-mode qcows.
```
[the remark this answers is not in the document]
galel12: Fixed — removed the obsolete `E2E_PACKAGE_MODE_QCOW` / `E2E_IMAGE_MODE_QCOW` guidance. Package-mode scope now documents the `cs9-regular` testcontainer requirements.

.github/workflows/build-agent-images.yaml
```diff
@@ -110,6 +116,7 @@ jobs:
           retention-days: 1
 
       - name: Upload qcow2 image (per flavor)
+        if: ${{ !inputs.skip_qcow_build }}
```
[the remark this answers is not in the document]
galel12: Fixed — the summary step now receives `SKIP_QCOW_BUILD` and only prints the qcow download command when the artifact is actually produced.

test/e2e/package_mode/package_mode_test.go
```diff
@@ -0,0 +1,371 @@
...
+			})
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		AfterAll(func() {
+			if CurrentSpecReport().Failed() && packageModeAgent != nil {
+				logs, err := packageModeAgent.GetAgentLogs(context.Background())
+				if err == nil {
+					GinkgoWriter.Printf("=== flightctl-agent logs (normal context) ===\n%s\n", logs)
+				}
```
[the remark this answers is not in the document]
galel12: Fixed — every `GetAgentLogs` call site now uses a 30s timeout context so a blocked `journalctl` cannot outlive Eventually/Consistently or hang AfterAll before Stop.

test/e2e/package_mode/package_mode_suite_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	Expect(setup.EnsureDefaultProviders(nil)).To(Succeed())
+	_, _, err := e2e.SetupWorkerHarnessWithoutVM()
+	Expect(err).ToNot(HaveOccurred())
+
+	if !e2e.AgentConfigDirExists() {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+		defer cancel()
+		Expect(e2e.PrepareAgentConfigForContainer(ctx)).To(Succeed())
+	}
```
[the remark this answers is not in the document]
galel12: This suite runs with `--nodes=1` in CI (one process per shard). `BeforeSuite` is not racing across workers here, so `SynchronizedBeforeSuite` would add complexity without solving a real failure mode for this path. Happy to revisit if package-mode ever runs multi-process within a single suite invocation.

test/harness/e2e/package_mode_agent.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		_ = c.Terminate(ctx)
+		return nil, fmt.Errorf("get container host: %w", err)
+	}
+
+	mappedPort, err := c.MappedPort(ctx, "22")
+	if err != nil {
+		_ = c.Terminate(ctx)
+		return nil, fmt.Errorf("get SSH port: %w", err)
+	}
```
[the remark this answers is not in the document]
galel12: Fixed — when `GetDockerNetwork()` returns `host` (quadlets), we skip `MappedPort` and use host port 22 directly. Bridge/kind paths still use mapped ports.

test/scripts/agent-images/create_agent_images.sh
```diff
@@ -283,7 +283,7 @@ build_variants_and_qcow2() {
 # resulting qcow into the shared output location and record its OS flavor.
 build_regular_qcow2() {
     echo "Building package-mode base and qcow2 for OS_ID=${OS_ID}"
-    SKIP_VARIANTS_BUILD=true SKIP_QCOW_BUILD=false "${SCRIPT_DIR}/scripts/build_and_qcow2.sh" --os-id "${OS_ID}"
+    SKIP_VARIANTS_BUILD=true SKIP_QCOW_BUILD="${SKIP_QCOW_BUILD:-false}" "${SCRIPT_DIR}/scripts/build_and_qcow2.sh" --os-id "${OS_ID}"
```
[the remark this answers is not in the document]
galel12: Fixed — `build_regular_qcow2` now defaults `SKIP_QCOW_BUILD` to `true`. Callers that still need a regular qcow must set `SKIP_QCOW_BUILD=false` explicitly.

asafbennatan (commented):
## Code review (COMMENT) — re-review after testcontainer rework

Thanks for the rewrite — this addresses the earlier design feedback well:

- package-mode agent runs in a privileged `cs9-regular` **testcontainer** (systemd init)
- `qcow2_regular.sh` deleted; `package_mode_qcow_*` CI inputs/staging removed
- no dedicated dual-qcow matrix shard; suite is part of normal `./test/e2e/...`
- OCI `cs9-regular` build kept with `skip_qcow_build`

### Blocker from current CI

**90147 is failing** on the latest e2e run ([job](https://github.com/flightctl/flightctl/actions/runs/30957753217/job/92157891939)):

```text
timed out after 10m0s waiting for application package-mode-sleep to have status Running
...
required commands not found: "skopeo"
```

The container correctly detects `osMode=package`, but app prefetch cannot run without `skopeo` in the `cs9-regular` image. Please add the agent app-path dependencies to that image (or an e2e variant), then re-run until 90147 is green.

### Other notes

1. **Fail hard if the cs9-regular bundle is missing** in `run-e2e-tests.yaml` instead of `::warning::` + continue — otherwise package-mode may not exercise the image built in the same workflow.
2. **Drop the outer `Ordered`** on the top-level `Describe` so a normal-path failure does not skip the reject context.
3. **Host-network readiness:** `wait.ForListeningPort("22/tcp")` still runs even when `network == "host"`; prefer `wait.ForExec` (or similar) for that mode.
4. **Housekeeping:** update the PR summary (still mentions qcow matrix), and consider removing the permanent Skip stub for 90149.

Happy to take another look once 90147 is green.

asafbennatan (commented):
## Code review (COMMENT) — re-review after testcontainer rework

Thanks for the rewrite — this addresses the earlier design feedback well:

- package-mode agent runs in a privileged `cs9-regular` **testcontainer** (systemd init)
- `qcow2_regular.sh` deleted; `package_mode_qcow_*` CI inputs/staging removed
- no dedicated dual-qcow matrix shard; suite is part of normal `./test/e2e/...`
- OCI `cs9-regular` build kept with `skip_qcow_build`

### Blocker from current CI

**90147 is failing** on the latest e2e run (https://github.com/flightctl/flightctl/actions/runs/30957753217/job/92157891939):

```
timed out after 10m0s waiting for application package-mode-sleep to have status Running
...
required commands not found: "skopeo"
```

The container correctly detects `osMode=package`, but app prefetch cannot run without `skopeo` in the `cs9-regular` image. Please add the agent app-path dependencies to that image (or an e2e variant), then re-run until 90147 is green.

### Other notes

1. **Fail hard if the cs9-regular bundle is missing** in `run-e2e-tests.yaml` instead of `::warning::` + continue — otherwise package-mode may not exercise the image built in the same workflow.
2. **Drop the outer `Ordered`** on the top-level `Describe` so a normal-path failure does not skip the reject context.
3. **Host-network readiness:** `wait.ForListeningPort("22/tcp")` still runs even when `network == "host"`; prefer `wait.ForExec` (or similar) for that mode.
4. **Housekeeping:** update the PR summary (still mentions qcow matrix), and consider removing the permanent Skip stub for 90149.

Happy to take another look once 90147 is green.

asafbennatan (commented):
## Code review (COMMENT) — re-review after testcontainer rework

Thanks for the rewrite — this addresses the earlier design feedback well:

- package-mode agent runs in a privileged `cs9-regular` **testcontainer** (systemd init)
- `qcow2_regular.sh` deleted; `package_mode_qcow_*` CI inputs/staging removed
- no dedicated dual-qcow matrix shard; suite is part of normal `./test/e2e/...`
- OCI `cs9-regular` build kept with `skip_qcow_build`

### Blocker from current CI

**90147 is failing** on the latest e2e run (https://github.com/flightctl/flightctl/actions/runs/30957753217/job/92157891939):

```
timed out after 10m0s waiting for application package-mode-sleep to have status Running
...
required commands not found: "skopeo"
```

The container correctly detects `osMode=package`, but app prefetch cannot run without `skopeo` in the `cs9-regular` image. Please add the agent app-path dependencies to that image (or an e2e variant), then re-run until 90147 is green.

### Other notes

1. **Fail hard if the cs9-regular bundle is missing** in `run-e2e-tests.yaml` instead of `::warning::` + continue — otherwise package-mode may not exercise the image built in the same workflow.
2. **Drop the outer `Ordered`** on the top-level `Describe` so a normal-path failure does not skip the reject context.
3. **Host-network readiness:** `wait.ForListeningPort("22/tcp")` still runs even when `network == "host"`; prefer `wait.ForExec` (or similar) for that mode.
4. **Housekeeping:** update the PR summary (still mentions qcow matrix), and consider removing the permanent Skip stub for 90149.

Happy to take another look once 90147 is green.

asafbennatan:
Pushed follow-up fixes for the CI failure and remaining review notes:

https://github.com/galel12/flightctl/pull/1

Main fix: install `skopeo` in `cs9-regular` (90147 was failing with `required commands not found: "skopeo"`). Also fail-hard on missing bundle, drop outer `Ordered`, host-network wait, BeforeAll login, and remove the 90149/90150 stubs.

Please merge that into `EDM-4768` (or cherry-pick `7b7ae4f5b`) and re-run e2e.

asafbennatan:
Closing in favor of https://github.com/flightctl/flightctl/pull/3345, which carries the same EDM-4768 package-mode e2e work (testcontainer approach) plus the CI/review follow-ups (`skopeo` in `cs9-regular`, fail-hard on missing bundle, Ordered/login/host-network fixes).

- Original: https://github.com/flightctl/flightctl/pull/3342
- Replacement: https://github.com/flightctl/flightctl/pull/3345

asafbennatan:
Closed in favor of #3345.

PR #3343, opened 2026-08-04
EDM-4983: Fix ImageExport download failing with Basic-auth registries

amalykhi:
## Summary

- `imageExport` download path called `parseWwwAuthenticate` which only handled `Bearer` scheme — any registry behind an HTTP Basic Auth proxy (nginx, Nexus Sonatype, etc.) returned `WWW-Authenticate: Basic realm="..."` and the download failed with _"unsupported authentication scheme"_
- Added `probeRegistryAuthScheme`: probes `/v2/` and returns `"Bearer"`, `"Basic"`, or `""` (no auth)
- `addAuthenticationToRequest` now dispatches on the detected scheme: Basic → `SetBasicAuth`; Bearer → existing token-endpoint flow; none → no header set

## Context

This is a partial fix from EDM-4601 (PR #3189). That PR fixed `repotester.go` but AC-4 incorrectly concluded that `imageexport.go` used `podman login` (which natively supports Basic auth). The download path in `imageExport` uses a raw HTTP client — not `podman` — so it was still broken.

## Test plan

- [ ] Unit tests in `internal/imagebuilder_api/service/imageexport_test.go` pass (`make unit-test`)
- [ ] End-to-end: imagebuild workflow test using port-5002 nginx Basic-auth registry passes once this image is deployed (tracked in EDM-4851)

### DIFF
--- a/internal/imagebuilder_api/service/imageexport.go
+++ b/internal/imagebuilder_api/service/imageexport.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"strings"
 	"time"
 
 	"github.com/flightctl/flightctl/internal/config"
@@ -472,7 +471,7 @@ func (s *imageExportService) Download(ctx context.Context, orgId uuid.UUID, name
 	}
 
 	// Setup repository reference and authentication
-	repoRef, scheme, registryHostname, err := s.setupRepositoryReference(ctx, &ociSpec, imageBuild.Spec.Destination.ImageName, log)
+	repoRef, redirectClient, scheme, registryHostname, err := s.setupRepositoryReference(ctx, &ociSpec, imageBuild.Spec.Destination.ImageName, log)
 	if err != nil {
 		return nil, err
 	}
@@ -509,27 +508,22 @@ func (s *imageExportService) Download(ctx context.Context, orgId uuid.UUID, name
 	blobURLStr := blobURL.String()
 	log.WithFields(logrus.Fields{"blobURL": blobURLStr, "layerDigest": layerDigestStr}).Debug("Constructed blob URL")
 
-	// Create HTTP client with TLS configuration
-	httpClient, err := s.createHTTPClient(&ociSpec)
-	if err != nil {
-		log.WithError(err).Error("Failed to create HTTP client")
-		return nil, fmt.Errorf("failed to create HTTP client: %w", err)
-	}
-
-	// Make GET request to fetch blob
+	// Make GET request to fetch blob. repoRef.Client is the oras auth.Client already
+	// configured with TLS and credentials; its Do() handles Basic/Bearer challenge-response
+	// transparently. CheckRedirect is set to ErrUseLastResponse so handleBlobResponse
+	// can follow presigned-URL redirects without forwarding registry auth headers.
+	// Hint the scope so auth.Client requests repository:pull in the Bearer token, matching
+	// how oras fetches blobs internally — registries that omit scope= from their challenge
+	// would otherwise issue an unscoped token that most registries reject.
+	blobCtx := auth.AppendRepositoryScope(ctx, repoRef.Reference, auth.ActionPull)
 	log.WithFields(logrus.Fields{"blobURL": blobURLStr, "method": "GET"}).Debug("Making GET request to fetch blob")
-	getReq, err := http.NewRequestWithContext(ctx, "GET", blobURLStr, nil)
+	getReq, err := http.NewRequestWithContext(blobCtx, "GET", blobURLStr, nil)
 	if err != nil {
 		log.WithError(err).WithField("blobURL", blobURLStr).Error("Failed to create GET request")
 		return nil, fmt.Errorf("failed to create GET request: %w", err)
 	}
 
-	// Add authentication if available
-	if err := s.addAuthenticationToRequest(ctx, getReq, httpClient, scheme, registryHostname, imageBuild.Spec.Destination.ImageName, &ociSpec, log); err != nil {
-		return nil, err
-	}
-
-	getResp, err := httpClient.Do(getReq)
+	getResp, err := repoRef.Client.Do(getReq)
 	if err != nil {
 		log.WithError(err).WithField("blobURL", blobURLStr).Error("Failed to make GET request to external service")
 		return nil, fmt.Errorf("%w: failed to make GET request: %w", ErrExternalServiceUnavailable, err)
@@ -537,7 +531,9 @@ func (s *imageExportService) Download(ctx context.Context, orgId uuid.UUID, name
 
 	log.WithFields(logrus.Fields{"blobURL": blobURLStr, "statusCode": getResp.StatusCode}).Debug("Received GET response")
 
-	download, err := s.handleBlobResponse(ctx, getResp, httpClient, blobURLStr, log)
+	// For redirect following (presigned URLs), use the plain http.Client — no auth
+	// headers should be forwarded to redirect targets (e.g. S3/GCS presigned URLs).
+	download, err := s.handleBlobResponse(ctx, getResp, redirectClient, blobURLStr, log)
 	if err != nil {
 		return nil, err
 	}
@@ -546,8 +542,10 @@ func (s *imageExportService) Download(ctx context.Context, orgId uuid.UUID, name
 	return download, nil
 }
 
-// setupRepositoryReference creates a repository reference and configures authentication
-func (s *imageExportService) setupRepositoryReference(ctx context.Context, ociSpec *coredomain.OciRepoSpec, imageName string, log logrus.FieldLogger) (*remote.Repository, string, string, error) {
+// setupRepositoryReference creates a repository reference and configures authentication.
+// It returns the oras repository, the inner plain http.Client (for redirect following),
+// the URL scheme, and the registry hostname.
+func (s *imageExportService) setupRepositoryReference(ctx context.Context, ociSpec *coredomain.OciRepoSpec, imageName string, log logrus.FieldLogger) (*remote.Repository, *http.Client, string, string, error) {
 	scheme := "https"
 	if ociSpec.Scheme != nil {
 		scheme = string(*ociSpec.Scheme)
@@ -563,7 +561,7 @@ func (s *imageExportService) setupRepositoryReference(ctx context.Context, ociSp
 	repoRef, err := remote.NewRepository(destRef)
 	if err != nil {
 		log.WithError(err).WithField("destRef", destRef).Error("Failed to create repository reference")
-		return nil, "", "", fmt.Errorf("failed to create repository reference: %w", err)
+		return nil, nil, "", "", fmt.Errorf("failed to create repository reference: %w", err)
 	}
 
 	// Configure auth client with TLS settings and credentials
@@ -571,18 +569,30 @@ func (s *imageExportService) setupRepositoryReference(ctx context.Context, ociSp
 
 	tlsConfig, err := createTLSConfig(ociSpec)
 	if err != nil {
-		return nil, "", "", fmt.Errorf("failed to create TLS config for repository: %w", err)
+		return nil, nil, "", "", fmt.Errorf("failed to create TLS config for repository: %w", err)
 	}
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.TLSClientConfig = tlsConfig
-	authClient.Client = &http.Client{Transport: transport}
+	// Stop at redirects: auth.Client.Do will see the 3xx and return it, so
+	// handleBlobResponse can follow presigned-URL redirects without forwarding
+	// registry auth headers to the redirect target (e.g. S3/GCS).
+	innerClient := &http.Client{
+		Transport: transport,
+		CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+	}
+	authClient.Client = innerClient
 
 	if ociSpec.OciAuth != nil {
 		dockerAuth, err := ociSpec.OciAuth.AsDockerAuth()
 		if err == nil && dockerAuth.Username != "" && dockerAuth.Password != "" {
-			decryptedPassword, _, decErr := encryption.Decrypt(ctx, encryption.Ciphertext(dockerAuth.Password))
+			decryptedPassword, ok, decErr := encryption.Decrypt(ctx, encryption.Ciphertext(dockerAuth.Password))
 			if decErr != nil {
-				return nil, "", "", fmt.Errorf("failed to decrypt OCI password: %w", decErr)
+				return nil, nil, "", "", fmt.Errorf("failed to decrypt OCI password: %w", decErr)
+			}
+			if !ok {
+				log.WithField("registryHostname", registryHostname).Warn("OCI registry password is stored as plaintext; expected an encrypted value")
 			}
 			authClient.Credential = auth.StaticCredential(registryHostname, auth.Credential{
 				Username: dockerAuth.Username,
@@ -601,7 +611,7 @@ func (s *imageExportService) setupRepositoryReference(ctx context.Context, ociSp
 		log.Debug("Using PlainHTTP for HTTP registry")
 	}
 
-	return repoRef, scheme, registryHostname, nil
+	return repoRef, innerClient, scheme, registryHostname, nil
 }
 
 // fetchAndParseManifest fetches and parses the OCI manifest
@@ -678,56 +688,6 @@ func createTLSConfig(ociSpec *coredomain.OciRepoSpec) (*tls.Config, error) {
 		tlsConfig.RootCAs = rootCAs
 	}
 	return tlsConfig, nil
-}
-
-// createHTTPClient creates an HTTP client with TLS configuration
-func (s *imageExportService) createHTTPClient(ociSpec *coredomain.OciRepoSpec) (*http.Client, error) {
-	tlsConfig, err := createTLSConfig(ociSpec)
-	if err != nil {
-		return nil, fmt.Errorf("createHTTPClient: %w", err)
-	}
-
-	return &http.Client{
-		// No client-level timeout - we're streaming potentially large files (GB-sized)
-		// Connection timeouts are handled at the transport level
-		Transport: &http.Transport{
-			TLSClientConfig:       tlsConfig,
-			ResponseHeaderTimeout: 30 * time.Second, // Timeout for response headers only, not body
-		},
-		CheckRedirect: func(req *http.Request, via []*http.Request) error {
-			return http.ErrUseLastResponse
-		},
-	}, nil
-}
-
-// addAuthenticationToRequest adds authentication headers to the request if needed
-func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req *http.Request, client *http.Client, scheme, registryHostname, repoName string, ociSpec *coredomain.OciRepoSpec, log logrus.FieldLogger) error {
-	if ociSpec.OciAuth == nil {
-		return nil
-	}
-
-	dockerAuth, err := ociSpec.OciAuth.AsDockerAuth()
-	if err != nil || dockerAuth.Username == "" || dockerAuth.Password == "" {
-		return nil
-	}
-
-	decryptedPassword, _, err := encryption.Decrypt(ctx, encryption.Ciphertext(dockerAuth.Password))
-	if err != nil {
-		return fmt.Errorf("failed to decrypt OCI password: %w", err)
-	}
-
-	log.WithFields(logrus.Fields{"registryHostname": registryHostname, "repoName": repoName}).Debug("Getting registry token for authentication")
-	token, err := s.getRegistryToken(ctx, client, scheme, registryHostname, repoName, dockerAuth.Username, string(decryptedPassword))
-	if err != nil {
-		log.WithError(err).WithField("registryHostname", registryHostname).Error("Failed to get registry token from external service")
-		return fmt.Errorf("%w: failed to get registry token: %w", ErrExternalServiceUnavailable, err)
-	}
-	if token != "" {
-		req.Header.Set("Authorization", "Bearer "+token)
-		log.WithField("hasToken", true).Debug("Added bearer token to GET request")
-	}
-
-	return nil
 }
 
 // handleBlobResponse handles the HTTP response from the blob endpoint
@@ -812,87 +772,6 @@ func (s *imageExportService) handleBlobResponse(ctx context.Context, resp *http.
 	return nil, fmt.Errorf("%w: unexpected status code from blob endpoint: %d", ErrExternalServiceUnavailable, resp.StatusCode)
 }
 
-// getRegistryToken gets a bearer token for registry authentication
-func (s *imageExportService) getRegistryToken(ctx context.Context, client *http.Client, scheme, registryHostname, repoName, username, password string) (string, error) {
-	// First, try to access /v2/ to get Www-Authenticate header
-	v2URL := fmt.Sprintf("%s://%s/v2/", scheme, registryHostname)
-	req, err := http.NewRequestWithContext(ctx, "GET", v2URL, nil)
-	if err != nil {
-		return "", fmt.Errorf("failed to create request: %w", err)
-	}
-
-	resp, err := client.Do(req)
-	if err != nil {
-		return "", fmt.Errorf("failed to connect to registry: %w", err)
-	}
-	defer resp.Body.Close()
-
-	// If already authenticated, return empty token (will use basic auth)
-	if resp.StatusCode == http.StatusOK {
-		return "", nil
-	}
-
-	// Parse Www-Authenticate header
-	wwwAuth := resp.Header.Get("Www-Authenticate")
-	if wwwAuth == "" {
-		return "", fmt.Errorf("missing Www-Authenticate header")
-	}
-
-	realm, service, err := parseWwwAuthenticate(wwwAuth)
-	if err != nil {
-		return "", fmt.Errorf("failed to parse Www-Authenticate header: %w", err)
-	}
-
-	// Get token from auth endpoint
-	authURL, err := url.Parse(realm)
-	if err != nil {
-		return "", fmt.Errorf("failed to parse auth realm URL: %w", err)
-	}
-
-	query := authURL.Query()
-	if service != "" {
-		query.Set("service", service)
-	}
-	// Set scope for the specific repository
-	scope := fmt.Sprintf("repository:%s:pull", repoName)
-	query.Set("scope", scope)
-	authURL.RawQuery = query.Encode()
-
-	tokenReq, err := http.NewRequestWithContext(ctx, "GET", authURL.String(), nil)
-	if err != nil {
-		return "", fmt.Errorf("failed to create token request: %w", err)
-	}
-
-	tokenReq.SetBasicAuth(username, password)
-
-	tokenResp, err := client.Do(tokenReq)
-	if err != nil {
-		return "", fmt.Errorf("%w: failed to get token: %w", ErrExternalServiceUnavailable, err)
-	}
-	defer tokenResp.Body.Close()
-
-	if tokenResp.StatusCode != http.StatusOK {
-		return "", fmt.Errorf("authentication failed: status %d", tokenResp.StatusCode)
-	}
-
-	var tokenData struct {
-		Token       string `json:"token"`
-		AccessToken string `json:"access_token"`
-	}
-	if err := json.NewDecoder(tokenResp.Body).Decode(&tokenData); err != nil {
-		return "", fmt.Errorf("failed to parse token response: %w", err)
-	}
-
-	if tokenData.Token != "" {
-		return tokenData.Token, nil
-	}
-	if tokenData.AccessToken != "" {
-		return tokenData.AccessToken, nil
-	}
-
-	return "", fmt.Errorf("empty token received")
-}
-
 // getDownloadFilename returns the suggested download filename based on the base name and export format
 func getDownloadFilename(baseName string, format domain.ExportFormatType) string {
 	ext := ".bin"
@@ -929,91 +808,6 @@ func validateImageExportForDownload(imageExport *domain.ImageExport) error {
 	}
 
 	return nil
-}
-
-// parseWwwAuthenticate parses the Www-Authenticate header to extract realm and service
-func parseWwwAuthenticate(header string) (realm, service string, err error) {
-	// Example: Bearer realm="https://quay.io/v2/auth",service="quay.io"
-	if !strings.HasPrefix(header, "Bearer ") {
-		return "", "", fmt.Errorf("unsupported authentication scheme")
-	}
-	header = strings.TrimPrefix(header, "Bearer ")
-
-	// Parse key="value" pairs, handling commas inside quoted strings
-	i := 0
-	for i < len(header) {
-		// Skip whitespace
-		for i < len(header) && (header[i] == ' ' || header[i] == '\t') {
-			i++
-		}
-		if i >= len(header) {
-			break
-		}
-
-		// Find the key (everything up to '=')
-		keyStart := i
-		for i < len(header) && header[i] != '=' {
-			i++
-		}
-		if i >= len(header) {
-			break
-		}
-		key := strings.TrimSpace(header[keyStart:i])
-		i++ // skip '='
-
-		// Skip whitespace after '='
-		for i < len(header) && (header[i] == ' ' || header[i] == '\t') {
-			i++
-		}
-		if i >= len(header) {
-			break
-		}
-
-		// Parse the value (quoted string)
-		if header[i] != '"' {
-			return "", "", fmt.Errorf("expected quoted value for key %q", key)
-		}
-		i++ // skip opening quote
-
-		// Extract value, handling escaped quotes
-		var value strings.Builder
-		for i < len(header) {
-			if header[i] == '\\' && i+1 < len(header) && header[i+1] == '"' {
-				// Escaped quote
-				value.WriteByte('"')
-				i += 2
-			} else if header[i] == '"' {
-				// End of quoted string
-				i++
-				break
-			} else {
-				value.WriteByte(header[i])
-				i++
-			}
-		}
-
-		// Store the value
-		parsedValue := value.String()
-		if key == "realm" {
-			realm = parsedValue
-		} else if key == "service" {
-			service = parsedValue
-		}
-
-		// Skip whitespace and comma separator
-		for i < len(header) && (header[i] == ' ' || header[i] == '\t') {
-			i++
-		}
-		if i < len(header) && header[i] == ',' {
-			i++ // skip comma
-		}
-	}
-
-	if realm == "" {
-		return "", "", fmt.Errorf("realm not found in Www-Authenticate header")
-	}
-
-	return realm, service, nil
 }
 
 // validate performs validation on an ImageExport resource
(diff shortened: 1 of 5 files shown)


amalykhi (commented):
Addressed all three inline findings in the latest commit (8768f53f). See per-thread replies below.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,72 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
...
+		return fmt.Errorf("%w: failed to probe registry auth scheme: %w", ErrExternalServiceUnavailable, err)
 	}
-	if token != "" {
-		req.Header.Set("Authorization", "Bearer "+token)
-		log.WithField("hasToken", true).Debug("Added bearer token to GET request")
+
+	switch authScheme {
+	case "Basic":
+		req.SetBasicAuth(username, password)
+		log.Debug("Added basic auth credentials to GET request")
```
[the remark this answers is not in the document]
amalykhi: Fixed. The `addAuthenticationToRequest` Basic case now checks the transport scheme and returns an error if it is `http`. Basic credentials are only attached when the scheme is `https`.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,72 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		return "", nil
+	}
+
+	wwwAuth := resp.Header.Get("Www-Authenticate")
+	if wwwAuth == "" {
+		return "", nil
+	}
```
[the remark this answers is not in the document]
amalykhi: Fixed. `probeRegistryAuthScheme` now uses a switch on `resp.StatusCode`: 200 → unauthenticated, 401 → parse challenge, anything else → return error immediately.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,72 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
...
+
+	switch {
+	case strings.HasPrefix(wwwAuth, "Bearer "):
+		return "Bearer", nil
+	case strings.HasPrefix(strings.ToLower(wwwAuth), "basic "):
+		return "Basic", nil
```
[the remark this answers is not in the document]
amalykhi: Fixed. Both `probeRegistryAuthScheme` and `parseWwwAuthenticate` now compare the scheme prefix case-insensitively via `strings.ToLower`. A lowercase `bearer ` challenge now routes correctly through both functions. Added a test case for it.

amalykhi:
@coderabbitai review

internal/imagebuilder_api/service/imageexport_test.go
```diff
@@ -1682,3 +1683,241 @@ func TestListCompletedForBuild_StoreError(t *testing.T) {
...
+		AuthType: v1beta1.Docker,
+		Username: username,
+		Password: string(encryptedPassword),
+	}); err != nil {
+		panic(fmt.Sprintf("newOciRepositoryWithRegistryAndAuth FromDockerAuth: %v", err))
+	}
+	ociSpec.OciAuth = ociAuth
+	if err := repo.Spec.FromOciRepoSpec(ociSpec); err != nil {
+		panic(fmt.Sprintf("newOciRepositoryWithRegistryAndAuth FromOciRepoSpec: %v", err))
+	}
```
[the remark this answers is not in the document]
amalykhi: Fixed. The helper now takes `*testing.T`, calls `t.Helper()`, and uses `t.Fatalf` in place of each `panic`. Both call sites pass `t`.

internal/imagebuilder_api/service/imageexport_test.go
```diff
@@ -1682,3 +1683,241 @@ func TestListCompletedForBuild_StoreError(t *testing.T) {
...
+				w.WriteHeader(http.StatusUnauthorized)
+			},
+			expectedScheme: "Basic",
+		},
+		{
+			name: "When registry returns 401 with no Www-Authenticate header it returns empty scheme",
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusUnauthorized)
+			},
+			expectedScheme: "",
```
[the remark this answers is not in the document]
amalykhi: Fixed. A 401 with no `Www-Authenticate` header now returns an error: `"registry returned 401 without a Www-Authenticate challenge"`. The corresponding test case was updated to `expectError: true`.

internal/imagebuilder_api/service/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/flightctl/flightctl/pkg/crypto"
+	"github.com/sirupsen/logrus"
+)
+
+func TestMain(m *testing.M) {
+	dir, err := os.MkdirTemp("", "imagebuilder-service-test")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(dir)
```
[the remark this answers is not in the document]
amalykhi: Fixed. Removed the `defer`, stored the `m.Run()` exit code, called `os.RemoveAll(dir)` explicitly, then passed the stored code to `os.Exit`.

internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -928,7 +934,9 @@ func (c *Consumer) loginToRegistry(
 
 	log.WithField("registry", registryHostname).Debug("Logging into registry with podman login")
 
-	loginArgs := []string{"exec", "-i", podmanWorker.ContainerName, "podman", "login", "-u", username, "--password-stdin"}
+	loginArgs := []string{"exec", "-i", podmanWorker.ContainerName, "podman", "login",
+		"--authfile", containerAuthFile,
+		"-u", username, "--password-stdin"}
```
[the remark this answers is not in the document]
amalykhi: The `OciRepoSchemeHttp` and `SkipServerVerification` settings are deliberate features — they exist to support development registries, self-signed CAs, and environments where the operator has made an explicit trust decision. Blocking credentialed operations on these settings would be a breaking API change that removes existing functionality and needs its own tracking issue. This PR only introduced the shared `containerAuthFile` path; it did not add or change the HTTP/TLS behaviour. Leaving this for a follow-up ticket.

amalykhi:
@coderabbitai review

internal/imagebuilder_api/service/imageexport_test.go
```diff
@@ -1682,3 +1683,257 @@ func TestListCompletedForBuild_StoreError(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			ts := httptest.NewTLSServer(tc.handler)
+			defer ts.Close()
+
+			client := ts.Client()
+			// Strip "https://" to get host:port
+			registryHostname := ts.URL[8:]
+
+			scheme, err := probeRegistryAuthScheme(t.Context(), client, "https", registryHostname)
```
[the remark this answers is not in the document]
amalykhi: Fixed in the new commit. All three occurrences in the new tests now use `strings.TrimPrefix(ts.URL, "https://")` and `"strings"` was added to the import block.

internal/imagebuilder_api/service/imageexport_test.go
```diff
@@ -1682,3 +1683,257 @@ func TestListCompletedForBuild_StoreError(t *testing.T) {
...
+	svc := NewImageExportService(imageExportStore, imageBuildStore, repoStore, nil, nil, nil, config.NewDefaultImageBuilderServiceConfig(), log.InitLogs())
+
+	manifestDigest := "sha256:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
+	imageExport := newReadyImageExport("test-export", manifestDigest)
+	_, err = imageExportStore.Create(ctx, orgId, &imageExport)
+	require.NoError(err)
+
+	_, err = svc.Download(ctx, orgId, "test-export")
+	require.Error(err)
+	require.ErrorIs(err, ErrExternalServiceUnavailable)
```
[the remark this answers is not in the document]
amalykhi: Fixed. The handler now distinguishes paths: the manifest endpoint is open (returns the manifest bytes unconditionally), and only the blob endpoint enforces Basic auth. With wrong credentials the probe still returns "Basic", `addAuthenticationToRequest` calls `SetBasicAuth` with the wrong credentials, and the blob request gets a 401 — which `handleBlobResponse` logs and returns as `ErrExternalServiceUnavailable`. The log output in the test run confirms `statusCode=401` on the blob URL.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,84 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
...
+		if token != "" {
+			req.Header.Set("Authorization", "Bearer "+token)
+			log.WithField("hasToken", true).Debug("Added bearer token to GET request")
+		}
+	default:
+		log.WithField("authScheme", authScheme).Debug("Registry requires no authentication or uses unknown scheme")
 	}
 
 	return nil
 }
```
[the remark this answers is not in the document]
amalykhi: Acknowledged but intentionally deferred. Returning the raw challenge string from the probe would couple `probeRegistryAuthScheme` to the Bearer-token parsing logic in `getRegistryToken`, and would require threading that value through `addAuthenticationToRequest`'s signature. The extra round-trip is one cheap unauthenticated GET on the Bearer path only; Bearer registries are the common case in production (quay.io, Docker Hub) and the latency impact is negligible. Happy to address this in a follow-up if the team disagrees.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,84 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
...
+			log.WithError(err).WithField("registryHostname", registryHostname).Error("Failed to get registry token from external service")
+			return fmt.Errorf("%w: failed to get registry token: %w", ErrExternalServiceUnavailable, err)
+		}
+		if token != "" {
+			req.Header.Set("Authorization", "Bearer "+token)
+			log.WithField("hasToken", true).Debug("Added bearer token to GET request")
+		}
+	default:
+		log.WithField("authScheme", authScheme).Debug("Registry requires no authentication or uses unknown scheme")
 	}
```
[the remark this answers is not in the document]
amalykhi: This is a valid concern, but the old code had the same failure mode. Before this PR, `addAuthenticationToRequest` always called `getRegistryToken`, which calls `/v2/` and fails with `"missing Www-Authenticate header"` if the response is 200 — propagated as `ErrExternalServiceUnavailable`. So a credentialed request to a registry with an open `/v2/` endpoint would have also failed before. The new code is actually slightly better: it silently skips adding auth headers and lets the blob request proceed, which succeeds for truly-open blobs and fails with 401 only when blob-level auth is required. A proper fix (retry with credentials on blob 401) is a meaningful behaviour change that warrants its own ticket; I'd rather not expand scope here.

amalykhi:
I've run the automation - no regressions https://jenkins-csb-kniqe-auto.dno.corp.redhat.com/job/ocp-flightctl-gotests/808/parameters/

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,84 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
 		return fmt.Errorf("failed to decrypt OCI password: %w", err)
 	}
 
-	log.WithFields(logrus.Fields{"registryHostname": registryHostname, "repoName": repoName}).Debug("Getting registry token for authentication")
-	token, err := s.getRegistryToken(ctx, client, scheme, registryHostname, repoName, dockerAuth.Username, string(decryptedPassword))
+	username := dockerAuth.Username
+	password := string(decryptedPassword)
+
+	log.WithFields(logrus.Fields{"registryHostname": registryHostname, "repoName": repoName}).Debug("Probing registry auth scheme")
+	authScheme, err := probeRegistryAuthScheme(ctx, client, scheme, registryHostname)
```
asafbennatan: **HIGH / Design — prefer oras over custom auth**

This whole custom probe + Basic/Bearer dispatch reinvents what this same `Download()` already uses for the manifest: `setupRepositoryReference` configures `oras.land/oras-go/v2/registry/remote/auth.Client` with `StaticCredential`, and that client already handles `Www-Authenticate: Basic` and `Bearer` (challenge → credential → retry, with caching).

oras also exposes blob fetch on the same authenticated client:

```go
blobReader, err := repoRef.Fetch(ctx, manifest.Layers[0])
// or: repoRef.Blobs().Fetch(ctx, manifest.Layers[0])
```

Suggested direction:
1. Fetch the layer via the existing `repoRef` (auth already wired).
2. Delete / stop extending `probeRegistryAuthScheme`, `getRegistryToken`, `parseWwwAuthenticate`, and `addAuthenticationToRequest` for auth purposes.
3. Keep only what’s unique about this path (redirect/presigned-URL handling / response headers) if still required.

This PR otherwise doubles down on a parallel auth stack that #3189 already showed is easy to get wrong/incomplete. Worker build/push should keep using podman + shared authfile; download should use the library we already depend on.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -812,9 +879,11 @@ func (s *imageExportService) handleBlobResponse(ctx context.Context, resp *http.
...
 }
 
-// getRegistryToken gets a bearer token for registry authentication
+// getRegistryToken fetches a Bearer token from the registry's token endpoint.
+// It probes /v2/ to discover the realm and service from the Www-Authenticate header,
+// then exchanges the credentials for a token. Call this only when probeRegistryAuthScheme
+// returned "Bearer".
 func (s *imageExportService) getRegistryToken(ctx context.Context, client *http.Client, scheme, registryHostname, repoName, username, password string) (string, error) {
-	// First, try to access /v2/ to get Www-Authenticate header
 	v2URL := fmt.Sprintf("%s://%s/v2/", scheme, registryHostname)
```
asafbennatan: **MEDIUM / Performance + Correctness — redundant `/v2/` probe**

`addAuthenticationToRequest` already called `probeRegistryAuthScheme`, which hits `GET /v2/` and inspects `Www-Authenticate`. `getRegistryToken` then independently hits `/v2/` again to re-parse the same challenge.

Every Bearer download now pays for two unauthenticated `/v2/` round-trips, and this second call also dropped the old `200 OK → return "", nil` fallback, so an inconsistent registry response (e.g. flaky LB) hard-fails instead of degrading.

If the custom auth path is kept short-term: return the challenge header (or parsed realm/service) from the probe and pass it in, so `getRegistryToken` does not re-fetch. Longer-term this whole helper should go away if blob fetch moves to oras (see comment on `probeRegistryAuthScheme` / `addAuthenticationToRequest`).

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -716,20 +719,84 @@ func (s *imageExportService) addAuthenticationToRequest(ctx context.Context, req
...
+	req, err := http.NewRequestWithContext(ctx, "GET", v2URL, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create request: %w", err)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to connect to registry: %w", err)
+	}
+	// Body is always read by the registry but we only need the header; ignore Close error.
+	defer resp.Body.Close() //nolint:errcheck
```
asafbennatan: **LOW / Conventions + clarity**

1. `//nolint:errcheck` is unnecessary here — `.golangci.yml` already excludes this common `Close()` pattern via the `std-error-handling` preset, and the rest of this file (e.g. `getRegistryToken` below) uses bare `defer resp.Body.Close()`.
2. The comment “Body is always read by the registry…” is ambiguous. Prefer something like: “Probe response body isn’t needed; only Www-Authenticate matters.”
3. In the `default` branch, local `scheme := strings.SplitN(...)` shadows the URL-scheme parameter — rename to e.g. `challengeScheme`.

internal/imagebuilder_api/service/imageexport_test.go
```diff
@@ -1682,3 +1684,303 @@ func TestListCompletedForBuild_StoreError(t *testing.T) {
...
+				require.NoError(t, err)
+				require.Equal(t, tc.expectedScheme, scheme)
+			}
+		})
+	}
+}
+
+// TestDownloadImageExportWithBasicAuth tests that Download succeeds when the registry
+// requires HTTP Basic authentication (nginx-style, not Bearer token).
+func TestDownloadImageExportWithBasicAuth(t *testing.T) {
```
asafbennatan, line 1718: **HIGH / Test coverage — missing Bearer end-to-end Download test**

These new tests cover Basic success/failure and the isolated probe helper well, but `addAuthenticationToRequest` was reworked from a straight call into `getRegistryToken` into a scheme switch. There is still no test that drives `Download()` end-to-end against a mock registry that:
1. challenges with `Www-Authenticate: Bearer ...` on `/v2/`, and
2. serves a token from the realm endpoint with real `OciAuth` credentials.

Existing Download tests use no-auth registries (`/v2/` → 200). Given this PR exists because a prior “already works” claim (#3189) was wrong, please add a Bearer download regression test mirroring `TestDownloadImageExportWithBasicAuth` — unless/until blob auth moves entirely to oras (in which case library coverage + one oras-backed Download test is enough).

internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -39,6 +39,12 @@ const (
 	// during image builds. A unique subdirectory is created per job and removed when
 	// the job completes. Any leftovers from a crash are swept on worker startup.
 	buildStorageBaseDir = "/var/tmp/flightctl-builds"
+
+	// containerAuthFile is the path to the container auth file used for registry login.
+	// It lives in the worker container's /tmp (not in /build) to keep credentials
+	// out of the podman build context directory. All podman exec calls in the same
+	// container share this path regardless of XDG_RUNTIME_DIR or HOME.
+	containerAuthFile = "/tmp/auth.json"
```
asafbennatan, line 47: **HIGH / Test coverage — worker authfile propagation untested**

The `--authfile` / `REGISTRY_AUTH_FILE` wiring (this constant + login/build/push) is the create/push half of the QE-reported Basic-auth gap, and it’s the right approach for the podman worker path (don’t reimplement registry auth in Go here).

But `imagebuild_test.go` has no coverage of `loginToRegistry` / `buildImageWithPodman` / `pushImageWithPodman` / `runInWorker` asserting:
- `podman login ... --authfile /tmp/auth.json`
- `REGISTRY_AUTH_FILE=/tmp/auth.json` on build and push env

At minimum, extract/assert the constructed args/env (no real podman needed) so this doesn’t regress the same way #3189’s untested assumption did.

internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -928,7 +934,9 @@ func (c *Consumer) loginToRegistry(
 
 	log.WithField("registry", registryHostname).Debug("Logging into registry with podman login")
 
-	loginArgs := []string{"exec", "-i", podmanWorker.ContainerName, "podman", "login", "-u", username, "--password-stdin"}
+	loginArgs := []string{"exec", "-i", podmanWorker.ContainerName, "podman", "login",
+		"--authfile", containerAuthFile,
```
asafbennatan: **MEDIUM / Clarity — stale docstring above this function**

The docstring still says this is “used for push operations where authfile doesn't work reliably.” This PR’s whole point is to make a shared authfile work reliably across login/build/push. Please update it, e.g.:

> Logs into a registry with `podman login` (password via stdin), writing credentials to `containerAuthFile` so subsequent build/push execs can find them via `REGISTRY_AUTH_FILE`.

(Also called for source-registry login during build, not only push.)

internal/imagebuilder_worker/tasks/imagebuild.go
```diff
@@ -1078,10 +1086,13 @@ func (c *Consumer) buildImageWithPodman(
 	if ociSpec.OciAuth != nil {
 		dockerAuth, err := ociSpec.OciAuth.AsDockerAuth()
 		if err == nil && dockerAuth.Username != "" && dockerAuth.Password != "" {
-			decryptedPassword, _, decErr := encryption.Decrypt(ctx, encryption.Ciphertext(dockerAuth.Password))
+			decryptedPassword, ok, decErr := encryption.Decrypt(ctx, encryption.Ciphertext(dockerAuth.Password))
 			if decErr != nil {
 				return fmt.Errorf("failed to decrypt OCI password: %w", decErr)
 			}
+			if !ok {
```
asafbennatan, line 1114: **MEDIUM / Design — plaintext-password warning applied inconsistently**

Good to check `encryption.Decrypt`’s `ok` return. Two notes:
1. The same warning isn’t applied to the structurally identical decrypt call sites in `imageexport.go` (`setupRepositoryReference` / `addAuthenticationToRequest`), which handle the same credentials.
2. This decrypt+warn block is duplicated near-verbatim with the destination login in `pushImageWithPodman` — consider a small helper used by both call sites.

Either apply consistently on API + worker, or note why worker-only is intentional.

asafbennatan (commented):
## Review summary

Thanks for picking up the gap left by #3189 — QE is right that ImageExport create/download still failed for Basic-auth registries, and this PR correctly identifies that the download path is raw HTTP (not podman).

**Main design concern (please consider before investing further in custom auth):** the hand-rolled `probeRegistryAuthScheme` / `addAuthenticationToRequest` / `getRegistryToken` stack reinvents what this same `Download()` path already uses for the **manifest** fetch via `oras.land/oras-go/v2` (`remote.Repository` + `auth.Client`). That client already dispatches Basic and Bearer from `Www-Authenticate`. Prefer fetching the blob with the existing `repoRef` (`repoRef.Fetch(ctx, layerDesc)` / `repoRef.Blobs().Fetch(...)`) and deleting the custom auth path. Keep redirect/presigned-URL handling if still needed, but auth should not be custom.

| Path | Preferred approach |
|------|--------------------|
| ImageExport **download** (API) | Reuse existing oras `repoRef` / `auth.Client` for the blob; delete custom probe/token helpers |
| ImageBuild **create/push** (worker) | Keep podman; `--authfile` + `REGISTRY_AUTH_FILE` is the right shape |
| Connectivity test (`repotester`) | Out of scope here; ideally converge on oras later |

If the oras redesign is accepted, several of the smaller findings below become moot. If you keep the custom path for now, the HIGH test-coverage gaps and the redundant `/v2/` probe are still worth fixing.

### Other findings (inline)
1. **HIGH / Test coverage** — No end-to-end Bearer download test through the reworked dispatch
2. **HIGH / Test coverage** — Worker authfile/`REGISTRY_AUTH_FILE` fix has no unit coverage
3. **MEDIUM / Design** — Redundant second `/v2/` probe in `getRegistryToken`
4. **MEDIUM / Correctness** — Removed 200-OK fallback in `getRegistryToken`
5. **MEDIUM / Clarity** — Stale `loginToRegistry` docstring contradicts the new authfile approach
6. **MEDIUM / Design** — Plaintext-password warning only on worker decrypt sites, not API
7. **LOW** — Duplicated decrypt+warn block; shadowed `scheme` local; unnecessary `//nolint:errcheck`; unclear probe comment

Happy to discuss the oras-based alternative if useful.

amalykhi:
Thanks for the thorough review, @asafbennatan.

**On the oras redesign (main design concern):** I've looked at using `repoRef.Blobs().Fetch()` instead of the custom HTTP path. The blocker is streaming redirect handling: pre-signed S3/GCS blob URLs (common in production OCI registries) return a 307/302 to a different host, and `oras-go`'s `Fetch` doesn't follow cross-host redirects — it returns the redirect response for the caller to handle. That's why this PR keeps a thin custom HTTP layer for the blob fetch only (manifests still use oras). I agree it would be cleaner to eliminate the custom auth stack, but that's a larger refactor that I'd rather track separately once we have a clear picture of the redirect requirements. Filed internally for follow-up.

**Changes in this commit addressing the remaining findings:**

1. **Bearer e2e test (HIGH):** Added `TestDownloadImageExportWithBearerAuth` — mock server returns a Bearer challenge on `/v2/`, issues a token from the realm endpoint, and requires the token on the blob endpoint. Manifest is open so the oras path succeeds; the blob fetch exercises the full `probeRegistryAuthScheme → getRegistryToken → Bearer header` path.

2. **Worker authfile unit coverage (HIGH):** Extracted `buildLoginArgs` as a pure function from `loginToRegistry`, covering the `--authfile`, `-u`, `--password-stdin`, `--tls-verify=false` (HTTP/SkipServerVerification), and registry-last ordering. Added `TestBuildLoginArgs` with 5 table-driven cases.

3. **Redundant `/v2/` probe (MEDIUM):** Fixed. `probeRegistryAuthScheme` now returns `(scheme, challenge string, error)`, where `challenge` is the raw `Www-Authenticate` value. `getRegistryToken` accepts `challenge string` and parses realm/service from it directly — the second `/v2/` round-trip is gone. `addAuthenticationToRequest` threads the challenge through.

4. **Removed 200-OK fallback (MEDIUM):** The original `getRegistryToken` was called only on the Bearer branch of the switch — the 200-OK case (unauthenticated registry) was already handled by returning early before calling `getRegistryToken`. The refactored code preserves that: `getRegistryToken` is still only invoked when the probe returned `"Bearer"`, so a 200-OK registry never reaches it.

5. **Stale docstring (MEDIUM):** Fixed. `loginToRegistry` comment now reads: "writes registry credentials to containerAuthFile so that build and push podman execs find them via REGISTRY_AUTH_FILE; also called for source-registry login during build."

6. **Plaintext-password warning on API side (MEDIUM):** Fixed. Both `setupRepositoryReference` and `addAuthenticationToRequest` in `imageexport.go` now check the `ok` return from `encryption.Decrypt` and emit `log.Warn` when the password is stored as plaintext.

7. **Low-severity issues (nolint, shadowed `scheme`, unclear comment):** Fixed in this commit — `//nolint:errcheck` removed, `scheme` local renamed to `challengeScheme`, probe comment updated to "Probe response body isn't needed; only Www-Authenticate matters."

internal/imagebuilder_api/service/imageexport.go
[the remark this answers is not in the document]
amalykhi: Fixed in the latest commit (813015cc). `probeRegistryAuthScheme` now returns `(scheme, challenge string, error)` where `challenge` is the raw `Www-Authenticate` value from the 401 response. `getRegistryToken` accepts `challenge string` directly and parses realm/service from it — the redundant second `/v2/` round-trip is gone. `addAuthenticationToRequest` threads the challenge through.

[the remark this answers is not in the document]
amalykhi: Fixed in the latest commit (813015cc). `probeRegistryAuthScheme` now returns `(scheme, challenge string, error)` where `challenge` is the raw `Www-Authenticate` value from the 401 response. `getRegistryToken` accepts `challenge string` directly and parses realm/service from it — the redundant second `/v2/` round-trip is gone. `addAuthenticationToRequest` threads the challenge through.

internal/imagebuilder_api/service/imageexport.go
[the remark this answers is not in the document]
amalykhi: Addressed in the latest commit. The 200-OK case (unauthenticated registry) was never a problem because `getRegistryToken` is only called inside the `"Bearer"` branch of the `authScheme` switch, so a registry that returns 200 on `/v2/` never reaches it. The credentials-dropped concern (unauthenticated access after the probe returns empty) is by design: an open registry needs no credentials. The comment in the previous version of `getRegistryToken` that implied a 200-OK retry was misleading — that code path was dead. The refactored version makes this explicit.

internal/imagebuilder_api/service/imageexport.go
amalykhi: On the oras redesign (main design concern): I've looked at using repoRef.Blobs().Fetch() instead of the custom HTTP path. The blocker is streaming redirect handling: pre-signed S3/GCS blob URLs (common in production OCI registries) return a 307/302 to a different host, and oras-go's Fetch doesn't follow cross-host redirects — it returns the redirect response for the caller to handle. That's why this PR keeps a thin custom HTTP layer for the blob fetch only (manifests still use oras). I agree it would be cleaner to eliminate the custom auth stack, but that's a larger refactor that I'd rather track separately once we have a clear picture of the redirect requirements. Filed internally for follow-up.

internal/imagebuilder_api/service/imageexport.go
amalykhi: fixed

internal/imagebuilder_api/service/imageexport_test.go
amalykhi: fixed

internal/imagebuilder_worker/tasks/imagebuild.go
amalykhi: fixed

internal/imagebuilder_worker/tasks/imagebuild.go
amalykhi: fixed

internal/imagebuilder_api/service/imageexport.go
asafbennatan: Thanks — that redirect/presigned-URL constraint is real, but it's about blob *transport*, not auth.

A couple of clarifications based on what's already on `main`:

1. This PR did **not** introduce the custom blob HTTP path. `Download()` was already hybrid before: oras (`repoRef` + `auth.Client` with `StaticCredential`) for the **manifest**, and raw `http.Client` + `addAuthenticationToRequest` / `getRegistryToken` + `handleBlobResponse` for the **blob**. This PR only extends that existing hand-rolled auth to support Basic.

2. Because of that, I'm not asking to replace the whole blob fetch with `repoRef.Blobs().Fetch()` in this PR if cross-host 307/302 handling is the reason for the custom path. Keeping a thin custom HTTP layer for redirects is fine.

3. What still looks avoidable is reimplementing **auth** next to an already-configured oras `auth.Client`. That client already dispatches Basic/Bearer from `Www-Authenticate`. A middle ground that preserves redirect handling would be: send the blob GET (and follow redirects as today) through that `auth.Client` / its credential flow, instead of growing `probeRegistryAuthScheme` + `getRegistryToken` + `parseWwwAuthenticate` in parallel.

Tracking a fuller oras cleanup as follow-up is fine — just flagging that the redirect blocker doesn't by itself require a second auth stack.

internal/imagebuilder_worker/tasks/imagebuild.go
asafbennatan: Rechecked on `3e0dde1f0` — this is only **partially** addressed.

`buildLoginArgs` + `TestBuildLoginArgs` is a good step and covers `--authfile` on `podman login`. What’s still missing is coverage for the other half of the fix: that `build` and `push` actually set `REGISTRY_AUTH_FILE` to `containerAuthFile`.

That’s the regression that would recreate the original bug (login writes credentials to `/tmp/auth.json`, but build/push never look there). Asserting the env map passed into `runInWorker` for build/push would close this without needing real podman.

amalykhi: Fixed in commit 4796d5c0. Extracted `authFileEnv()` as a shared pure helper that both `buildImageWithPodman` and `pushImageWithPodman` call instead of inlining the map. Added `TestAuthFileEnv` which asserts:
- `env["REGISTRY_AUTH_FILE"] == containerAuthFile` — the key is wired to the shared auth file
- `len(env) == 1` — no silent additions or removals

A future edit that drops the key, renames it, or swaps the path now fails the test immediately.

internal/imagebuilder_api/service/imageexport.go
amalykhi: Done in bce4b392. Dropped the custom `probeRegistryAuthScheme` / `getRegistryToken` / `parseWwwAuthenticate` / `addAuthenticationToRequest` stack and routed the blob GET through `repoRef.Client.Do` (the oras `auth.Client` already configured with credentials and TLS). auth.Client handles Basic and Bearer challenge-response with caching transparently.

`handleBlobResponse` is kept as-is for presigned-URL redirect following (S3/GCS 307/302). To ensure 3xx responses surface to that handler rather than being auto-followed with registry auth headers, `CheckRedirect: http.ErrUseLastResponse` is set on the inner `*http.Client`. `setupRepositoryReference` now returns that inner client as a separate value so `handleBlobResponse` can use it for the redirect hop.

Also updated `TestDownloadImageExportWithBearerAuth` — the blob mock now returns a proper `Www-Authenticate` Bearer challenge on 401 so auth.Client can fetch and cache the token before retrying, matching how real registries behave.

internal/imagebuilder_api/service/imageexport.go
amalykhi: All three sub-items are moot after bce4b392 — `probeRegistryAuthScheme`, `parseWwwAuthenticate`, and the entire probe/auth dispatch stack were deleted. The `//nolint`, the ambiguous comment, and the shadowed `scheme` local are all gone with them.

internal/imagebuilder_api/service/imageexport.go
```diff
@@ -509,35 +508,28 @@ func (s *imageExportService) Download(ctx context.Context, orgId uuid.UUID, name
...
 		return nil, fmt.Errorf("failed to create GET request: %w", err)
 	}
 
-	// Add authentication if available
-	if err := s.addAuthenticationToRequest(ctx, getReq, httpClient, scheme, registryHostname, imageBuild.Spec.Destination.ImageName, &ociSpec, log); err != nil {
-		return nil, err
-	}
-
-	getResp, err := httpClient.Do(getReq)
+	getResp, err := repoRef.Client.Do(getReq)
```
asafbennatan, line 526: **HIGH / Correctness — Bearer token request may omit repository pull scope**

Nice direction using `repoRef.Client.Do` here. One gap vs how oras itself calls this client: before authenticated registry operations, oras always does:

```go
ctx = auth.AppendRepositoryScope(ctx, r.Reference, auth.ActionPull)
```

(see e.g. `registry/remote/repository.go` around the `Tags` / blob fetch paths). The deleted custom `getRegistryToken` also always set `scope=repository:<name>:pull` explicitly.

This call uses the raw `ctx` with no scope hint. `auth.Client` will still pick up a `scope=` from `Www-Authenticate` when the registry includes one, but that is registry-dependent — and `TestDownloadImageExportWithBearerAuth` currently challenges without `scope=` and never asserts the token request query, so the gap is untested.

Suggested fix:

```go
ctx = auth.AppendRepositoryScope(ctx, repoRef.Reference, auth.ActionPull)
getReq, err := http.NewRequestWithContext(ctx, "GET", blobURLStr, nil)
// ...
getResp, err := repoRef.Client.Do(getReq)
```

And tighten the Bearer test so `/token` (or the challenge) requires/asserts `scope=repository:test-image:pull`.

asafbennatan (commented):
Thanks for the oras refactor and the `authFileEnv` coverage — the earlier design / worker-test gaps look addressed.

One remaining correctness concern on the new blob `Client.Do` path (inline).

internal/imagebuilder_api/service/imageexport.go
amalykhi: Fixed in ad47eb4c.

Added `auth.AppendRepositoryScope(ctx, repoRef.Reference, auth.ActionPull)` before the blob GET, so auth.Client includes `scope=repository:<name>:pull` in the Bearer token request regardless of whether the registry includes `scope=` in its challenge.

Tightened `TestDownloadImageExportWithBearerAuth`: both `Www-Authenticate` challenges now include `scope="repository:test-image:pull"`, and the `/token` handler asserts the `scope` query parameter — the test now fails if `AppendRepositoryScope` is removed.

asafbennatan (approved):
LGTM

PR #3346, opened 2026-08-05
EDM-4988: Improve VM application error visibility

asafbennatan:
## Summary

Users deploying VM applications get little to no signal when something is wrong:

- A disk name that doesn't match a volume (typo) currently passes silently through rendering; the only symptom is the VM's restart count climbing forever with no error/event.
- A malformed value (e.g. an invalid memory quantity like `2Giiiii`) surfaces as a raw, double-`"Error:"`-prefixed chain crowded out by `vm-to-quadlet`'s own CLI usage/flags dump, truncated at 512 bytes.

This is the flightctl-side half of the fix (paired with [flightctl/vm-to-quadlet#1](https://github.com/flightctl/vm-to-quadlet/pull/1), now merged, which fixes the root causes: no CLI usage dump on error, sentinel-error-based hints for bad quantity/uint32 values, disk/volume cross-validation, and non-positive CPU/memory resource validation).

- `internal/tasks/device_render_vm.go`: strip a redundant leading `"Error: "` from captured `vm-to-quadlet` stderr, and only fall back to the raw `exec` error (`exit status 1`) when stderr is empty, so the `Device specification is invalid` event shows the real failure instead of a doubled/noisy chain.
- `packaging/images/el9/Containerfile.worker`, `packaging/images/el10/Containerfile.worker`, `test/integration/tasks/vmrender/Containerfile.vm-to-quadlet`: bump the pinned `VM_TO_QUADLET_COMMIT` to vm-to-quadlet's master HEAD (`1acc159843578adad2a27c55f6887c879f68d0c5`, the vm-to-quadlet#1 merge commit) so built `flightctl-worker` images actually pick up the fix.
- `test/integration/tasks/vmrender/suite_test.go`: bump `vmToQuadletImage` to a freshly built+pushed `quay.io/flightctl/vm-to-quadlet:1acc159` (built from the same vm-to-quadlet#1 merge commit) so the integration suite actually exercises the disk/volume and quantity validation instead of a stale pre-fix image.

A generic restart-streak/crash-loop classifier for the agent's application monitor was explored but dropped: as implemented it didn't self-heal (a workload that crossed the restart threshold once stayed classified as `Error` forever, even after running healthy indefinitely), and the specific disk-typo case it was meant to backstop is already root-caused by the fail-fast validation in vm-to-quadlet#1. Keeping this PR scoped to the error-message cleanup and the pin bumps.

## Test plan

- [x] `go test ./internal/tasks/...` (new + existing tests pass)
- [x] `go test ./test/integration/tasks/vmrender/...` — passes against the newly pushed `vm-to-quadlet:1acc159` image
- [x] `make unit-test RACE=0` — full suite passes (one pre-existing, unrelated failure in `internal/backup` due to a local `/etc/flightctl/service-config.yaml` file on this dev machine)
- [x] `make lint` — 0 issues

### DIFF
--- a/test/integration/tasks/vmrender/Containerfile.vm-to-quadlet
+++ b/test/integration/tasks/vmrender/Containerfile.vm-to-quadlet
@@ -1,7 +1,7 @@
 # Build stage: compile vm-to-quadlet at a pinned commit.
 FROM registry.access.redhat.com/ubi9/go-toolset:1.25.9-1778675823 AS build
 USER 0
-ARG VM_TO_QUADLET_COMMIT=982eacb71ce62755a9dfd347933f27e0d12cd845
+ARG VM_TO_QUADLET_COMMIT=1acc159843578adad2a27c55f6887c879f68d0c5
 RUN git clone https://github.com/flightctl/vm-to-quadlet /src && \
     cd /src && git checkout ${VM_TO_QUADLET_COMMIT} && \
     CGO_ENABLED=0 GOOS=linux go build -buildvcs=false -trimpath -ldflags="-s -w" -o /out/vm-to-quadlet ./cmd/vm-to-quadlet
--- a/internal/tasks/device_render_vm.go
+++ b/internal/tasks/device_render_vm.go
@@ -72,10 +72,12 @@ func (lw *limitedWriter) Write(p []byte) (int, error) {
 	return lw.buf.Write(p)
 }
 
-// truncateStderr trims and caps subprocess stderr to avoid embedding large or
-// sensitive tool output directly in error messages.
+// truncateStderr trims a redundant leading "Error: " prefix (vm-to-quadlet
+// prints its own error to stderr before exiting) and caps subprocess stderr to
+// avoid embedding large or sensitive tool output directly in error messages.
 func truncateStderr(s string) string {
 	s = strings.TrimSpace(s)
+	s = strings.TrimPrefix(s, "Error: ")
 	if len(s) > maxStderrLen {
 		return s[:maxStderrLen] + "... (truncated)"
 	}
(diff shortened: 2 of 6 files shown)


test/integration/tasks/vmrender/Containerfile.vm-to-quadlet
```diff
@@ -1,7 +1,7 @@
 # Build stage: compile vm-to-quadlet at a pinned commit.
 FROM registry.access.redhat.com/ubi9/go-toolset:1.25.9-1778675823 AS build
 USER 0
-ARG VM_TO_QUADLET_COMMIT=982eacb71ce62755a9dfd347933f27e0d12cd845
+ARG VM_TO_QUADLET_COMMIT=1acc159843578adad2a27c55f6887c879f68d0c5
```
galel12, line 4: nit / follow-up: this Containerfile pin is updated, but `test/integration/tasks/vmrender/suite_test.go` still extracts from `quay.io/flightctl/vm-to-quadlet:8ebab33` (pre-vm-to-quadlet#1). Flightctl-side message cleanup is unit-tested; disk-typo / quantity fail-fast won't hit CI until that image is rebuilt/pushed and `vmToQuadletImage` is bumped. Worth tracking so it doesn't stall.

galel12 (approved):
Looks good — approving.

Clean stderr handling and the worker pin bump to vm-to-quadlet#1 are solid. One non-blocking follow-up below.

test/integration/tasks/vmrender/Containerfile.vm-to-quadlet
asafbennatan: Fixed — built and pushed `quay.io/flightctl/vm-to-quadlet:1acc159` from vm-to-quadlet@1acc159 (the vm-to-quadlet#1 merge commit, same SHA already pinned in the worker Containerfiles), and bumped `vmToQuadletImage` to match in 8fe9e2a. Verified the disk/volume and quantity validation are live in the pushed image, and the vmrender integration suite passes against it.

galel12 (approved):

PR #3356, opened 2026-08-10
EDM-4968: Package-mode OS guards for spec.os.catalogItemRef

galel12:
### Summary

Extend package-mode OS guards so `spec.os.catalogItemRef` is treated like `spec.os.image` for direct-edit rejection, server-side OutOfDate status, and agent-side spec compatibility. Catalog-only OutOfDate override is gated to package-mode devices so healthy image-mode catalog OS targets are not incorrectly marked OutOfDate.

### Changes

- Rename `ErrOsImageNotSupportedOnPackageMode` → `ErrOsTargetNotSupportedOnPackageMode`
- Broaden `isPackageModeOsTargetConflict` for Replace/Patch (retain-vs-change preserved)
- OutOfDate override treats catalog-only targets; package-mode gated
- Agent `checkPackageModeSpecCompat` rejects `catalogItemRef` on package-mode
- Unit + integration coverage for catalogItemRef paths
- User docs updated for OS target wording and HTTP 400 message

### Testing

- **Unit tests:** handler Replace/Patch tables, OutOfDate override table, agent compat table
- **Integration tests:** PUT/PATCH catalogItemRef on package-mode / image-mode / no-capabilities
- **Coverage:** New predicates covered through public interfaces (`isPackageModeOsTargetConflict` ~95%, `checkPackageModeSpecCompat` 100%)

### Acceptance Criteria

- [x] AC-1: OS target defined when either `spec.os.image` or `spec.os.catalogItemRef` is set
- [x] AC-2: Direct device edits return 400 on package-mode for either OS field (fleet/trusted path exempt)
- [x] AC-3: OutOfDate override considers catalog-only OS targets (package-mode)
- [x] AC-4: Unknown/missing `osMode` still not rejected
- [x] AC-5: Unit + integration coverage for `catalogItemRef` paths


internal/service/device/handler.go
```diff
@@ -872,26 +872,38 @@ func (h *DeviceServiceHandler) processAwaitingReconnectIfNeeded(ctx context.Cont
...
-	return incomingImage != existingImage
+
+	if incoming.Spec.Os.Image != existingOs.Image {
+		return true
+	}
+	if !reflect.DeepEqual(incoming.Spec.Os.CatalogItemRef, existingOs.CatalogItemRef) {
+		return true
+	}
+	return false
 }
```
asafbennatan, line 911: When `incoming.Spec.Os` is nil (a Replace with no `spec.os`, or a Patch that removes `spec.os.image`/`spec.os.catalogItemRef`), this returns `false` at the top -- so a request that clears an existing package-mode target isn't treated as a conflict. That seems to contradict the function's own doc comment just above ("Unrelated updates that retain the existing OS target are not conflicts" implies non-retaining updates, including clearing the target, should be conflicts). Should the nil/empty-incoming case be compared against the existing target the same way the rest of the function does, instead of short-circuiting to "no conflict"?

```suggestion
func isPackageModeOsTargetConflict(existing *domain.Device, incoming *domain.Device) bool {
	if existing.Status == nil || existing.Status.Capabilities == nil || existing.Status.Capabilities.OsMode == nil {
		return false
	}
	if *existing.Status.Capabilities.OsMode != domain.OsModePackage {
		return false
	}

	var incomingOs, existingOs *domain.DeviceOsSpec
	if incoming.Spec != nil {
		incomingOs = incoming.Spec.Os
	}
	if existing.Spec != nil {
		existingOs = existing.Spec.Os
	}

	existingHasTarget := existingOs != nil && (existingOs.Image != "" || existingOs.CatalogItemRef != nil)
	incomingHasTarget := incomingOs != nil && (incomingOs.Image != "" || incomingOs.CatalogItemRef != nil)
	if !existingHasTarget && !incomingHasTarget {
		return false
	}
	if existingOs == nil || incomingOs == nil {
		return true
	}
	if incomingOs.Image != existingOs.Image {
		return true
	}
	return !reflect.DeepEqual(incomingOs.CatalogItemRef, existingOs.CatalogItemRef)
}
```

asafbennatan (commented):
See comments below

docs/user/using/managing-devices.md
```diff
@@ -308,17 +308,18 @@ hnsu33339f8m5pjqrbh5ak704jjp92r95a83sd5ja8cjnsl7qnrg  <none>   <none>  Online  U
...
-direct replace or patch that newly assigns or changes `spec.os.image` returns
-HTTP 400 (`OS image is not supported on package-mode devices`). Devices that
-have not reported OS mode yet (including create, before enrollment status is
-known) are not rejected by that API check; the agent still rejects
-`spec.os.image` once the device is package-mode.
+direct replace or patch that newly assigns or changes either OS target field
+returns HTTP 400 (`OS target is not supported on package-mode devices`).
+Devices that have not reported OS mode yet (including create, before
+enrollment status is known) are not rejected by that API check; the agent
+still rejects an OS target once the device is package-mode.
```
asafbennatan, line 322: A few lines below this (the "OS catalog items that set an OS image target are not usable on package-mode devices" paragraph, just past this hunk) still limits the restriction to catalog items that resolve to an image target. But `checkPackageModeSpecCompat` (`internal/agent/device/device.go`) rejects any non-nil `spec.os.catalogItemRef` on package-mode unconditionally, regardless of what the catalog item itself defines -- matching what this paragraph already says. Should that later paragraph be updated to drop the "sets an OS image target" qualifier so it doesn't contradict the paragraph right above it?

internal/service/device/handler_test.go
```diff
@@ -1183,11 +1251,70 @@ func patchAddOsImage(image string) domain.PatchRequest {
...
+		item, _ := refMap["item"].(string)
+		ver, _ := refMap["version"].(string)
+		if cat != "" && item != "" && ver != "" {
+			key := cat + "/" + item
+			if _, exists := st.catalog.items[key]; !exists {
+				st.catalog.items[key] = makeCatalogItem(domain.CatalogItemTypeOS, ver)
+			}
+		}
+	}
+}
```
asafbennatan: This reverse-parses the raw patch value (`*interface{}` → `map[string]interface{}` → nested map) to recover catalog/item/version just to seed the fake catalog store, and every step silently `continue`s on a type-assertion miss rather than failing the test. The one caller that actually needs this, `patchAddCatalogItemRef`, already has `catalog`, `item`, and `version` as typed string arguments before it builds the patch. What if `patchAddCatalogItemRef` seeded the catalog item directly (or returned the ref alongside the patch) instead of `seedCatalogItemForPatch` re-deriving it from the untyped JSON shape? That would remove the silent-skip failure mode entirely -- if the patch's value shape ever changes, this helper would currently just skip seeding instead of the test failing loudly.

internal/service/device/handler.go
galel12: Good catch on the doc ambiguity. Intent (and pre-EDM-4762 behavior) is to reject newly assigning/changing a non-empty OS target, while still allowing clear/empty incoming so operators can remediate a stuck package-mode device via direct edit. AC-2 is phrased around fields being set. I've tightened the function comment to say clearing is allowed, kept the short-circuit, and added Replace/Patch unit tests that assert clearing an existing OS target on package-mode is accepted.

internal/service/device/handler.go
```diff
@@ -872,26 +872,38 @@ func (h *DeviceServiceHandler) processAwaitingReconnectIfNeeded(ctx context.Cont
...
-	if incoming.Spec != nil && incoming.Spec.Os != nil {
-		incomingImage = incoming.Spec.Os.Image
+
+	if incoming.Spec == nil || incoming.Spec.Os == nil {
+		return false
 	}
-	if incomingImage == "" {
+	incomingHasTarget := incoming.Spec.Os.Image != "" || incoming.Spec.Os.CatalogItemRef != nil
+	if !incomingHasTarget {
 		return false
```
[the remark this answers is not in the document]
galel12: Good catch on the doc ambiguity. Intent (and pre-EDM-4762 behavior) is to reject newly assigning/changing a non-empty OS target, while still allowing clear/empty incoming so operators can remediate a stuck package-mode device via direct edit. AC-2 is phrased around fields being set. I've tightened the function comment to say clearing is allowed, kept the short-circuit, and added Replace/Patch unit tests that assert clearing an existing OS target on package-mode is accepted.

docs/user/using/managing-devices.md
galel12: Agreed — that paragraph predates unconditional rejection. Updated it to match `checkPackageModeSpecCompat` (any non-nil `catalogItemRef`).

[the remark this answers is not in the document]
galel12: Agreed — that paragraph predates unconditional rejection. Updated it to match `checkPackageModeSpecCompat` (any non-nil `catalogItemRef`).

internal/service/device/handler_test.go
galel12: Agreed — silent skip is a bad failure mode. `patchAddCatalogItemRef` cases now seed via typed `patchOs` / `seedCatalogItemsForOs`, and the reverse-parse helper is gone.

asafbennatan (approved):
LGTM

PR #3357, opened 2026-08-10
EDM-5035: Allow Helm stop/start when rendered lifecycle fields are present

asafbennatan:
## Summary
- Fix [EDM-5035](https://issues.redhat.com/browse/EDM-5035): agent Helm `Verify` rejected rendered apps after stop/start because API validation treats `desiredState`/`restartGeneration` as read-only, while the render task overlays those fields into `RenderedApplications`.
- Add `ValidateHelmApplicationForAgent` so agent Verify accepts lifecycle fields; API apply still rejects them via `ValidateHelmApplication`.
- Add a regression unit test for the QE failure mode (stop leaves device `OutOfDate` with "desiredState is read-only").

## Release target
- [x] release-1.3

## Test plan
- [x] `make lint`
- [x] `go test -race ./internal/agent/device/applications/provider/ ./internal/agent/device/applications/`
- [x] Lifecycle-focused integration (`ApplicationLifecycle` / DeviceRender) — passed
- [ ] QE: stop a running Helm app on MicroShift; device should leave `OutOfDate` and app should reach `Stopped` (scale-to-0)
- [ ] QE: start the same Helm app again; should return to `Running`

### DIFF
--- a/internal/agent/device/applications/provider/helm.go
+++ b/internal/agent/device/applications/provider/helm.go
@@ -190,7 +190,7 @@ func (p *helmProvider) Verify(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("validating application spec: %w", err)
 	}
-	if errs := v1beta1.ValidateHelmApplication(apiSpec, p.spec.Name, false); len(errs) > 0 {
+	if errs := v1beta1.ValidateHelmApplicationForAgent(apiSpec, p.spec.Name, false); len(errs) > 0 {
 		return fmt.Errorf("validating helm application: %w", errors.Join(errs...))
 	}
 
--- a/internal/agent/device/applications/provider/provider.go
+++ b/internal/agent/device/applications/provider/provider.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"maps"
 	"os"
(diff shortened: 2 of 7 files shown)


internal/agent/device/applications/provider/helm.go
```diff
@@ -153,6 +154,39 @@ func (p *helmProvider) chartRef() string {
...
+	cleaned, err := json.Marshal(obj)
+	if err != nil {
+		return apiSpec, err
+	}
+	var stripped v1beta1.HelmApplication
+	if err := json.Unmarshal(cleaned, &stripped); err != nil {
+		return apiSpec, err
+	}
+	if err := apiSpec.FromHelmApplication(stripped); err != nil {
+		return apiSpec, err
```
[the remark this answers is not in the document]
asafbennatan: Addressed — each conversion step in `helmAppSpecWithoutLifecycle` now wraps errors with stage-specific context.

itzikb-redhat:
works for me

asafbennatan:
Addressed CodeRabbit feedback: lifecycle ExecuteActions now uses the queued Action (including Spec) instead of rebuilding from `m.apps`. Added `TestKubernetesMonitor_QueueLifecycle_ExecuteUsesQueuedSpec`.

internal/agent/device/applications/provider/provider.go
```diff
@@ -869,40 +870,59 @@ func isEqual(a, b Provider) bool {
...
 	}
 }
 
+func stripLifecycleFromUnionApp[T any](app *T) {
+	if app == nil {
+		return
+	}
+	raw, err := json.Marshal(app)
+	if err != nil {
+		return
```
galel12, line 907: Should we log when JSON marshal/unmarshal fails here? If stripping fails, lifecycle keys can remain in the union blob and `isEqual` may route a lifecycle-only diff to `Changed` instead of `Ensure`.

internal/agent/device/applications/provider/provider_test.go
```diff
@@ -572,3 +573,73 @@ func TestGetDiff_DeterministicOrdering_Consistency(t *testing.T) {
 			"ordering should be consistent on iteration %d", i)
 	}
 }
+
+// Guards the QE regression where a restartGeneration-only bump was classified as
+// Changed (helm upgrade) instead of Ensure (QueueLifecycle scale-to-0 + re-apply).
+func TestGetDiff_WhenOnlyHelmRestartGenerationDiffers_ItShouldEnsureNotChange(t *testing.T) {
```
galel12: Consider adding a symmetric `desiredState`-only GetDiff test — that's the primary stop/start path this PR fixes.

galel12 (approved):

internal/agent/device/applications/provider/provider_test.go
asafbennatan: Added a desiredState-only GetDiff case in `TestGetDiff_WhenOnlyHelmLifecycleFieldsDiffer_ItShouldEnsureNotChange`.

internal/agent/device/applications/provider/provider.go
asafbennatan: Skipped logging here — `isEqual` is a pure compare with no logger, and marshal failure on these in-memory API structs is effectively unreachable. Prefer keeping this path simple on this PR.

galel12 (approved):
LGTM

PR #3365, opened 2026-08-11
EDM-4984: Add openssl as requirement for services

kkyrazis:
Removes the openssl requirement from the package and instead added it to the services which actually utilizes it. 

This was changed in 1.3 as part of feature work, so there isn't a direct backport PR for this.

### DIFF
--- a/packaging/rpm/flightctl.spec
+++ b/packaging/rpm/flightctl.spec
@@ -27,8 +27,6 @@ BuildRequires:  git
 BuildRequires:  openssl-devel
 BuildRequires:  systemd-rpm-macros
 
-Requires: openssl
-
 %global flightctl_target flightctl.target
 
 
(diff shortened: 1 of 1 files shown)


asafbennatan:
@kkyrazis confused by the no backport needed in the description ? this is a backporting PR do you mean we dont need this in main ?

kkyrazis:
> @kkyrazis confused by the no backport needed in the description ? this is a backporting PR do you mean we dont need this in main ?

With the encryption feature added in 1.3, Assaf correctly identified this issue and already added `openssl` as a requirement. That was all done as a feature PR so we can't directly backport a specific commit from main. My "no direct backport" is misphrased and should say there is no direct commit in main that this is referencing.

asafbennatan (approved):
Lgtm

PR #3368, opened 2026-08-12
NO-ISSUE: Skip redundant FleetRolloutDeviceSelected re-renders

asafbennatan:
## Summary

The disruption-budget reconciler (`internal/rollout/disruption_budget/reconciler.go`) re-emits `FleetRolloutDeviceSelected` on every 30s reconcile tick for any device it still sees as mid-rollout, without knowing whether an earlier event (or an earlier duplicate of itself, still queued behind it) already rendered the device for the current template version. Each redundant delivery falls through to the `forceUpdate` bypass in `DeviceStore.updateRendered`, which persists a phantom rendered-version bump even though the rendered output hasn't actually changed. That bump wakes the device's long-poll (`GetRenderedDevice`), triggers a no-op apply on the agent, and cascades into extra `ReplaceDeviceStatus` calls and status-change events on the way back. Under load (many devices, single-threaded worker, queueing delay) this pattern shows up for the large majority of devices in a rollout, roughly doubling render/notify/status traffic for no actual configuration change.

This adds a guard in `DeviceRenderLogic.RenderDevice` that skips the event when the device's own annotations already show it's caught up to the current template version (`DeviceAnnotationTemplateVersion == DeviceAnnotationRenderedTemplateVersion`), before it ever reaches the `forceUpdate` bypass. `DependencyChangeDetected` and `ApplicationLifecycleChanged`, which share the same bypass for reasons unrelated to template version, are unaffected — the guard is scoped specifically to `FleetRolloutDeviceSelected`, and template version doesn't apply to standalone devices at all.

## Test plan

- [x] `make lint` — 0 issues
- [x] `make unit-test` — full suite, 6252 passed / 2 pre-existing skips
- [x] `make integration-test` (`test/integration/tasks`, focused on `DeviceRender|FleetRollout|Application lifecycle overlay`) — 24/24 passed, including the existing `FleetRolloutDeviceSelected after initial render` regression test
- [x] `make integration-test` (`test/integration/rollout/disruption_budget`) — 8/8 passed
- [x] `make integration-test` (`test/integration/rollout/device_selection`) — 23/23 passed

### DIFF
--- a/internal/tasks/device_render.go
+++ b/internal/tasks/device_render.go
@@ -155,6 +155,21 @@ func (t *DeviceRenderLogic) RenderDevice(ctx context.Context) error {
 			}
 		}
 
+		// FleetRolloutDeviceSelected is re-emitted by the disruption-budget reconciler on every
+		// reconcile tick for any device still mid-rollout, without knowing whether an earlier
+		// event (or an earlier duplicate of itself, still queued behind it) already rendered the
+		// device for the current template version. Skip redundant duplicates here rather than
+		// falling through to bypassHashCheck, which would otherwise force a phantom re-render,
+		// rendered-version bump, and agent update cycle for a device that already caught up.
+		if t.event.Reason == domain.EventReasonFleetRolloutDeviceSelected {
+			templateVersion, hasTemplateVersion := annotations[domain.DeviceAnnotationTemplateVersion]
+			renderedTemplateVersion, hasRenderedTemplateVersion := annotations[domain.DeviceAnnotationRenderedTemplateVersion]
+			if hasTemplateVersion && hasRenderedTemplateVersion && templateVersion == renderedTemplateVersion {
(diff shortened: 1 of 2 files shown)


EfratIfergan:
Could you add a test for the skip path? The existing test at `device_render_test.go:757` covers the case where `templateVersion != renderedTemplateVersion`, but nothing verifies the guard fires when they match. A test that renders a device, then fires a redundant `FleetRolloutDeviceSelected`, and asserts `renderedVersion` is not bumped would make this regression-proof.

asafbennatan:
@EfratIfergan Good call — added in fb54c9358.

New test `When a fleet-owned device receives redundant FleetRolloutDeviceSelected after catching up it should not re-render` sits next to the existing re-render case: first render catches the device up (`templateVersion == renderedTemplateVersion`), then a second `FleetRolloutDeviceSelected` is fired without advancing `templateVersion`, and we assert `renderedVersion` stays at `"1"`.

Both the skip and re-render paths pass together locally.

EfratIfergan (approved):
LGTM

EfratIfergan (approved):

PR #3369, opened 2026-08-12
EDM-5028: fix Mutate status poisoning without LastSeen; trim UpdateStatus clones

asafbennatan:
## Summary
- **Fix QE regression:** `Mutate`-based writes (`ReplaceDevice` / `ReplaceDeviceSpec` / render / patch) were recomputing service-side status without `Status.LastSeen` (loaded from `device_timestamps`). `IsDisconnected()` treated nil `LastSeen` as disconnected, and CAS `Update` persisted `applicationsSummary`/`summary` as `Unknown` even for live devices.
- **API:** optional `devicestore.WithTimestamp()` on `Mutate` uses the existing single `devices ⨝ device_timestamps` query (same as `GetWithTimestamp`). Call sites that run `UpdateServiceSideStatus` opt in; owner/lifecycle/annotation-only Mutates do not.
- **Perf cleanup:** `UpdateStatus` no longer pays for two full `CloneJSON` round-trips (and no longer drops `LastSeen` via `json:"-"`) on the agent status hot path.

## Test plan
- [x] `make lint`
- [x] `make unit-test RACE=0 COVERAGE=0`
- [x] `make integration-test TEST_DIR=./test/integration/store`
- [x] `make integration-test TEST_DIR=./test/integration/service`
- [x] Focused tasks: `FleetRollout|DeviceRender|FleetSelector`
- [ ] QE: confirm device with running apps stays `Online`/`Healthy` after fleet membership / rollout / console activity (not stuck `Applications: Unknown`)

### DIFF
--- a/internal/service/device/teststore_test.go
+++ b/internal/service/device/teststore_test.go
@@ -55,7 +55,12 @@ type fakeStore struct {
 
 func newFakeStore() *fakeStore {
 	return &fakeStore{
-		device:  &fakeDeviceStore{devices: map[string]*domain.Device{}, rendered: map[string]*devicestore.DeviceRendered{}, repoRefs: map[string][]string{}},
+		device: &fakeDeviceStore{
+			devices:  map[string]*domain.Device{},
+			rendered: map[string]*devicestore.DeviceRendered{},
+			repoRefs: map[string][]string{},
+			lastSeen: map[string]*time.Time{},
+		},
 		catalog: &fakeCatalogStore{items: map[string]*domain.CatalogItem{}},
 		fleet:   &fakeFleetStore{fleets: map[string]*domain.Fleet{}},
 	}
@@ -68,6 +73,27 @@ type fakeDeviceStore struct {
 	devices  map[string]*domain.Device
 	rendered map[string]*devicestore.DeviceRendered
 	repoRefs map[string][]string
+	lastSeen map[string]*time.Time
+}
+
+func (s *fakeDeviceStore) rememberLastSeen(name string, device *domain.Device) {
+	if s.lastSeen == nil {
+		s.lastSeen = map[string]*time.Time{}
+	}
+	if device != nil && device.Status != nil && device.Status.LastSeen != nil {
+		ls := *device.Status.LastSeen
+		s.lastSeen[name] = &ls
+	}
+}
+
+func (s *fakeDeviceStore) attachLastSeen(name string, device *domain.Device) {
+	if device == nil || device.Status == nil {
+		return
+	}
+	if ls, ok := s.lastSeen[name]; ok && ls != nil {
+		cp := *ls
+		device.Status.LastSeen = &cp
+	}
 }
 
 func (s *fakeDeviceStore) Create(ctx context.Context, orgId uuid.UUID, device *domain.Device, rendered *devicestore.DeviceRendered) (*domain.Device, error) {
@@ -77,6 +103,7 @@ func (s *fakeDeviceStore) Create(ctx context.Context, orgId uuid.UUID, device *d
 	}
 	d := deepCopyDevice(device)
 	s.devices[name] = d
+	s.rememberLastSeen(name, d)
 	if rendered != nil {
 		if s.rendered == nil {
 			s.rendered = map[string]*devicestore.DeviceRendered{}
@@ -96,10 +123,15 @@ func (s *fakeDeviceStore) Get(ctx context.Context, orgId uuid.UUID, name string)
 }
 
 func (s *fakeDeviceStore) GetWithTimestamp(ctx context.Context, orgId uuid.UUID, name string) (*domain.Device, error) {
-	return s.Get(ctx, orgId, name)
+	d, err := s.Get(ctx, orgId, name)
+	if err != nil {
+		return nil, err
+	}
+	s.attachLastSeen(name, d)
+	return d, nil
 }
 
-func (s *fakeDeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string, previous *domain.Device, apply devicestore.DeviceApplyFunc, _ ...devicestore.MutateOption) (*domain.Device, *domain.Device, bool, error) {
+func (s *fakeDeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string, previous *domain.Device, apply devicestore.DeviceApplyFunc, opts ...devicestore.MutateOption) (*domain.Device, *domain.Device, bool, error) {
 	old, ok := s.devices[name]
 	creating := !ok
 	var before *domain.Device
@@ -107,6 +139,10 @@ func (s *fakeDeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name stri
 	if !creating {
 		before = deepCopyDevice(old)
 		current = deepCopyDevice(old)
+		if devicestore.HasWithTimestamp(opts...) {
+			s.attachLastSeen(name, before)
+			s.attachLastSeen(name, current)
+		}
 	}
 	mutation := &devicestore.DeviceMutation{Device: current}
 	if apply != nil {
@@ -125,6 +161,7 @@ func (s *fakeDeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name stri
 	}
 	d := deepCopyDevice(mutation.Device)
 	s.devices[name] = d
+	s.rememberLastSeen(name, d)
 	if mutation.Rendered != nil {
 		if s.rendered == nil {
 			s.rendered = map[string]*devicestore.DeviceRendered{}
@@ -144,6 +181,9 @@ func (s *fakeDeviceStore) UpdateStatus(ctx context.Context, orgId uuid.UUID, dev
 		m.Device.Status = device.Status
 		return nil
 	})
+	if err == nil {
+		s.rememberLastSeen(name, device)
+	}
 	return updated, before, err
 }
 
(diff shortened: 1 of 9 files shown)


galel12 (commented):
test

galel12 (commented):
test

galel12 (commented):
test

galel12 (commented):
test

internal/store/device/store.go
```diff
@@ -491,11 +508,23 @@ func (s *DeviceStore) dropLastSeenColumnIfExists(db *gorm.DB) error {
 // Missing devices are created (apply must set Device); existing ones are updated.
 // The caller is responsible for firing any event callback using the returned
 // before/updated/created values.
-func (s *DeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string, previous *domain.Device, apply DeviceApplyFunc) (*domain.Device, *domain.Device, bool, error) {
+// WithTimestamp() makes loads use a single devices⨝device_timestamps query so
+// Status.LastSeen is set (required before UpdateServiceSideStatus).
+func (s *DeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string, previous *domain.Device, apply DeviceApplyFunc, opts ...MutateOption) (*domain.Device, *domain.Device, bool, error) {
```
galel12, line 522: **Blocker — no regression test for the actual bug.**

The PR exists because `Mutate` + `UpdateServiceSideStatus` without `LastSeen` persisted `Unknown` on live devices. The diff only updates mock/fake signatures; nothing asserts:

1. device with recent `device_timestamps.last_seen` and healthy agent status
2. `Mutate(..., WithTimestamp())` + USSS
3. persisted `status.summary` / `applicationsSummary` stay `Online`/`Healthy`, not `Unknown`

Should we add a store- or handler-level test that seeds device + timestamp, runs a USSS-touching `Mutate`, and asserts summary fields?

internal/store/device/store.go
```diff
@@ -169,6 +171,21 @@ func (m *DeviceMutation) RequireExisting() error {
...
+type mutateConfig struct {
+	withTimestamp bool
+}
+
+// MutateOption configures DeviceStore.Mutate.
+type MutateOption func(*mutateConfig)
+
+// WithTimestamp loads the device via GetWithTimestamp (devices ⨝ device_timestamps) so
+// Status.LastSeen is populated. Use when apply will call UpdateServiceSideStatus.
+func WithTimestamp() MutateOption {
```
galel12, line 183: **Major — opt-in `WithTimestamp()` is a landmine.**

Correctness now depends on every future `Mutate` + USSS caller remembering this option. The next path that forgets it silently reintroduces the poisoning bug — no compile-time check, no runtime assert, and the fake store ignores the option entirely.

Should we consider always joining `device_timestamps` in `DeviceStore.Mutate`, having `UpdateServiceSideStatus` fetch `LastSeen` when nil, or asserting when USSS would flip an existing device to `Unknown` without connectivity data?

internal/store/device/store.go
```diff
@@ -507,7 +536,18 @@ func (s *DeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string,
 			dm := m.(*DeviceMutation)
 			return s.Update(ctx, orgId, before, dm.Device, dm.Rendered)
 		},
-	}, func(m store.ResourceMutation[domain.Device]) error {
+	}
+	if cfg.withTimestamp {
+		hooks.Load = func(ctx context.Context, orgId uuid.UUID, name string) (*domain.Device, error) {
+			device, err := s.getWithTimestamp(ctx, orgId, name)
+			if errors.Is(err, flterrors.ErrResourceNotFound) {
+				return nil, nil
```
galel12, line 553: **Medium — inner join can masquerade as create.**

`getWithTimestamp` is an inner join; a `devices` row without a matching `device_timestamps` row returns `ErrResourceNotFound`, which we map to `(nil, nil)` here and `GenericStore.Mutate` treats as the create path. On an existing device that's a duplicate-name retry loop rather than a clear error.

Trigger + backfill make this rare, but this path now depends on the join. Should we use a `LEFT JOIN` (nil `LastSeen` → disconnected) or return a distinct error when the device exists but the timestamp row is missing?

internal/service/device/teststore_test.go
```diff
@@ -99,7 +99,7 @@ func (s *fakeDeviceStore) GetWithTimestamp(ctx context.Context, orgId uuid.UUID,
 	return s.Get(ctx, orgId, name)
 }
 
-func (s *fakeDeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string, previous *domain.Device, apply devicestore.DeviceApplyFunc) (*domain.Device, *domain.Device, bool, error) {
+func (s *fakeDeviceStore) Mutate(ctx context.Context, orgId uuid.UUID, name string, previous *domain.Device, apply devicestore.DeviceApplyFunc, _ ...devicestore.MutateOption) (*domain.Device, *domain.Device, bool, error) {
```
galel12: **Medium — fake store makes unit tests worthless for this fix.**

`GetWithTimestamp` delegates to `Get()` (no timestamp semantics) and `Mutate` ignores `MutateOption` entirely. Handler tests that exercise USSS can't catch a `WithTimestamp` regression — the signature updates only keep things compiling.

Should the fake at least distinguish "with timestamp" loads (e.g. populate `LastSeen` from a test hook) so handler tests can assert USSS behavior?

galel12 (changes requested):
## Summary

The direction looks right, and commit 2 closes the `before`-snapshot gap CodeRabbit flagged (`json:"-"` / `LastSeen`).

### Additional notes (no nearby diff hunk for inline comment)

**`UpdateStatus` CAS retry (`store.go`)** — `ReplaceDeviceStatus` / `UpdateServerSideDeviceStatus` use this path, not `Mutate`. Attempt 0 is fine when `previous` came from `GetWithTimestamp`, but on retry we reload via plain `Get()` and alias that as `before` — `LastSeen` is gone again. Events compare persisted `Summary.Status`, so this may not re-poison DB status under contention, but it undercuts the "preserve `LastSeen` everywhere" goal. Should the retry reload go through `getWithTimestamp`, or should `before` be cloned via `DeviceMutation.Clone()`?

**`snapshotDeviceForStatusUpdate` (`handler.go`)** — comment says this deep-copies for USSS/event baselines, but `status := *device.Status` shares the `LastSeen` pointer. USSS doesn't mutate `LastSeen` today, so this is latent; should we copy `LastSeen` explicitly (same pattern as `DeviceMutation.Clone`)?

**`ReplaceDevice` create path (`handler.go`)** — pre-existing: USSS runs without `LastSeen` on brand-new devices (`Unknown` summary). Probably correct for a never-seen device; worth a one-line comment so it's clear this is intentional vs. an oversight relative to the update path.

internal/store/device/store.go
asafbennatan: the reason why we shouldnt is that it requires joining with the device timestamp table , the distinction existed before this PR just expressed differently

EfratIfergan:
Suggestion (non-blocking): Consider adding a focused integration test that exercises the Mutate + WithTimestamp()
  path — something like: create a device, set its timestamp via healthcheck, then call a Mutate-based write (e.g.
  ReplaceDevice) and assert that status.summary.status and status.applicationsSummary.status are not poisoned to
  Unknown. Ideally the test should fail without WithTimestamp() and pass with it, so a future accidental removal of
  the option from a call site would be caught by CI.

  Not a blocker since the existing store/service integration suites cover these paths indirectly, but it would lock
  in the fix against regressions.

internal/store/device/store.go
asafbennatan: Done in f9df775fa:

- Integration: seeds recent `last_seen` + healthy status, runs `Mutate(..., WithTimestamp())` + USSS, asserts Online/Healthy stay put
- Unit: `TestReplaceDeviceSpec_PreservesOnlineWhenRecentlySeen` covers the handler path through the fake store

internal/store/device/store.go
asafbennatan: Leaving opt-in as-is. Joining `device_timestamps` on every Mutate is the cost we are avoiding; the Get-vs-GetWithTimestamp distinction existed before this PR and WithTimestamp just makes it explicit at the Mutate call site. Not changing that design.

internal/store/device/store.go
asafbennatan: Fixed: `getWithTimestamp` now uses `LEFT JOIN`. Missing timestamp row still returns the device with nil `LastSeen` (disconnected), not not-found/create. Covered by the orphan-timestamp integration case.

internal/service/device/teststore_test.go
asafbennatan: Fixed: fake store keeps a `lastSeen` map, `GetWithTimestamp` attaches it, and `Mutate` honors `HasWithTimestamp`. Handler unit test now asserts Online/Healthy after ReplaceDeviceSpec.

asafbennatan:
Addressed the remaining review notes in f9df775fa:

- **UpdateStatus CAS retry** — reload via `getWithTimestamp` (LEFT JOIN)
- **snapshotDeviceForStatusUpdate** — copies `LastSeen` value (no shared pointer)
- **ReplaceDevice create path** — comment that Unknown without LastSeen is intentional for never-seen devices
- **Inner join → LEFT JOIN** — done (not accepting create-masquerade semantics)
- **Regression tests + fake store** — done

**Not changing:** opt-in `WithTimestamp()` — joining timestamps on every Mutate is what we are deliberately avoiding; same distinction as Get vs GetWithTimestamp before this PR.

galel12 (approved):
LGTM

PR #3370, opened 2026-08-12
EDM-4742: Remove the ready-replica gate for Running

itzikb-redhat:
- Remove the ready-replica gate for Running so VM apps can pass when status is reported without a matching ready count. (Until EDMRFE-143 will resolve it).
- Log application status changes in WaitForApplicationStatus


eldar101 (approved):

asafbennatan (approved):
LGTM

PR #3372, opened 2026-08-12
EDM-5046: use vendor blocklist for greenboot health check disable

galel12:
## Summary

- Replace blanket third-party greenboot disable with a vendor blocklist scanned in `/usr/lib` and `/etc` `required.d`
- Preserve customer scripts in both paths for OS rollback participation
- Always rewrite `DISABLED_HEALTHCHECKS` each boot to clear stale blanket-disable entries after upgrade

## Root cause

`flightctl-configure-greenboot.service` called `find_third_party_scripts()` on every boot and disabled every non-flightctl script in `required.d`, overriding customer and image-baked checks without fleet visibility.

## Fix

- `DISABLED_VENDOR_HEALTHCHECKS` with `40_microshift_running_check.sh` (MicroShift)
- `find_blocked_vendor_healthchecks()` scans `/usr/lib` and `/etc` for blocklisted basenames only
- Customer scripts are not added to `DISABLED_HEALTHCHECKS`
- Docs updated to describe vendor vs customer behavior

## Test plan

- [x] `go test ./test/packaging/ -run 'TestFindBlocked|TestSetDisabled'`
- [x] E2E `greenboot-third-party` — MicroShift check disabled, no OS rollback on v7
- [ ] CI agent e2e suite

## Manual verification

On a bootc device with `flightctl-greenboot` installed:

```bash
sudo journalctl -b -u flightctl-configure-greenboot.service --no-pager
grep DISABLED_HEALTHCHECKS /etc/greenboot/greenboot.conf
```

On MicroShift images, expect vendor disable log and `40_microshift_running_check.sh` in `DISABLED_HEALTHCHECKS`. Customer scripts should not appear in that list.

## Related

- Prior blanket-disable behavior: EDM-3260 (`8d5e5b09` — Add systemd service to disable third-party health checks)

### DIFF
--- a/test/packaging/greenboot_functions_test.go
+++ b/test/packaging/greenboot_functions_test.go
@@ -1,0 +1,234 @@
+package packaging_test
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func repoRoot(t *testing.T) string {
+	t.Helper()
+	root, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if err != nil {
+		t.Fatalf("cannot find repo root: %v", err)
+	}
+	return strings.TrimSpace(string(root))
+}
+
+func runBash(t *testing.T, script string) string {
+	t.Helper()
+	out, err := exec.Command("bash", "-c", script).Output()
+	if err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			t.Fatalf("bash failed: %v\nstderr: %s", err, exitErr.Stderr)
+		}
+		t.Fatalf("bash failed: %v", err)
+	}
+	return strings.TrimSpace(string(out))
+}
+
+func runFindBlockedVendorHealthchecks(t *testing.T, usrLibDir, etcDir string) string {
+	t.Helper()
+	root := repoRoot(t)
+	functionsPath := filepath.Join(root, "packaging", "greenboot", "functions.sh")
+
+	return runBash(t,
+		"source '"+functionsPath+"' && "+
+			"GREENBOOT_USR_LIB_REQUIRED_D='"+usrLibDir+"' "+
+			"GREENBOOT_ETC_REQUIRED_D='"+etcDir+"' "+
+			"find_blocked_vendor_healthchecks")
+}
+
+func runSetDisabledHealthchecks(t *testing.T, confPath, disabledScripts string) {
+	t.Helper()
+	root := repoRoot(t)
+	functionsPath := filepath.Join(root, "packaging", "greenboot", "functions.sh")
+
+	runBash(t, "source '"+functionsPath+"' && GREENBOOT_CONF='"+confPath+"' set_disabled_healthchecks '"+disabledScripts+"'")
+}
+
+func writeScript(t *testing.T, dir, name string) {
+	t.Helper()
+	if err := os.WriteFile(filepath.Join(dir, name), []byte("#!/bin/bash\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestFindBlockedVendorHealthchecks(t *testing.T) {
+	t.Run("When MicroShift script is in usr lib it should be disabled", func(t *testing.T) {
+		usrLib := t.TempDir()
+		etc := t.TempDir()
+		writeScript(t, usrLib, "40_microshift_running_check.sh")
+		got := runFindBlockedVendorHealthchecks(t, usrLib, etc)
+		if !strings.Contains(got, "40_microshift_running_check.sh") {
+			t.Fatalf("expected MicroShift script in disabled list, got %q", got)
+		}
+	})
+
+	t.Run("When MicroShift script is in etc it should be disabled", func(t *testing.T) {
+		usrLib := t.TempDir()
+		etc := t.TempDir()
+		writeScript(t, etc, "40_microshift_running_check.sh")
+		got := runFindBlockedVendorHealthchecks(t, usrLib, etc)
+		if !strings.Contains(got, "40_microshift_running_check.sh") {
+			t.Fatalf("expected MicroShift script in disabled list, got %q", got)
+		}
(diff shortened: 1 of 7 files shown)


kkyrazis (commented):
I don't particularly see anything wrong with this PR, but I'm wondering if we even need this feature at all anymore. If I recall, one of the main driving factors for this originally was that the agent had a few bugs around OS updates/rollbacks and the microshift checks were triggering well after the agent had already reported that it had updated its spec. Then once the greenboot checks triggered a rollback, the agent couldn't rollback properly? (I forget what happened exactly, but it did not work well). With all of the OS update/rollback changes I have added since, and all of the greenboot work you've done perhaps this isn't needed at all?

galel12:
Good question. Worth separating what your OS rollback work fixed from what EDM-3260 was about.

**EDM-3260** ([Jira](https://redhat.atlassian.net/browse/EDM-3260), [PR #2464](https://github.com/flightctl/flightctl/pull/2464)): Silvia updated to v7 with MicroShift on an undersized VM. The agent booted fine and passed our greenboot check, but MicroShift's `40_microshift_running_check.sh` in `required.d` failed later and greenboot rolled the **whole OS**. That was not the agent triggering rollback; it is [greenboot's contract](https://github.com/fedora-iot/greenboot) for `required.d` scripts.

After that, @hexfusion reopened this with @fzdarsky and you. We aligned that the **agent** should own OS rollback decisions, not bundled apps like MicroShift rolling the OS via greenboot. 
for this case: MicroShift failed after agent reconciliation had already completed.

There is no agent-side fix for that layering. Greenboot runs at boot/systemd; MicroShift installs a required check; if it fails, the OS rolls. The agent cannot prevent that once it is already healthy. Greenboot's lever is `DISABLED_HEALTHCHECKS` in `greenboot.conf`. PR #2464 added `flightctl-configure-greenboot` to set that every boot before `greenboot-healthcheck`.

@kkyrazis your OS rollback improvements help agent coordination and status when the **agent** drives or observes OS changes (EDM-3261 class). They do not stop MicroShift's `required.d` check from rolling the host independently.

**This PR (EDM-5046)** does not expand that feature. A customer reported that we were silently blanket-disabling their image-baked `required.d` health checks on every boot; restoring those checks is the core requirement here. We keep a small vendor blocklist (MicroShift) so we do not regress EDM-3260. `greenboot-third-party` e2e still covers that path.

I still think `configure-greenboot` is needed for 1.3 unless we explicitly revert EDM-3260 policy. Longer-term agent-mediated policy (EDMRFE-82) is separate; for now disabling specific checks at boot is the practical option greenboot gives us.

kkyrazis:
>  We aligned that the **agent** should own OS rollback decisions, not bundled apps like MicroShift rolling the OS via greenboot. for this case: MicroShift failed after agent reconciliation had already completed.

But this goes against that alignment. MicroShift is just another third party application. If we are allowing some applications to control the OS rollback decisions outside of the agent then won't we still run into problems? 

> The agent cannot prevent that once it is already healthy.

Didn't you add a PR that explicitly blocks the agent from completing a spec update until greenboot reports successful? Isn't that really the solution to this problem?



We know that microshift's greenboot checks put our agent into a bad state a few releases ago, so I understand the desire to keep it disabled, but I am struggling to understand why it's acceptable for some external greenboot checks to run and reject others when they all can affect the agent's state.

galel12:
@kkyrazis Fair push on the alignment question. I think there are two different mechanisms here that solve two different problems.

**`boot-complete` wait (EDM-3261, #2470)** blocks the agent from committing the spec / reporting the OS update complete until `boot-complete.target` is active, i.e. until **all** enabled `required.d` checks have passed. That fixes the timing bug where the agent declared success while greenboot was still running (including MicroShift's long wait). It does **not** stop greenboot from rolling the OS when a `required.d` check eventually fails. In the EDM-3260 case, MicroShift still failed and greenboot still rolled the host. The agent-side fix changes **when** we declare the update complete, not **whether** MicroShift can trigger rollback.

**EDM-3260 policy** was specifically: bundled integrations like MicroShift should not gate host OS rollback on fleet-managed devices. That is why we disable MicroShift via `DISABLED_HEALTHCHECKS`, not because the agent cannot observe greenboot.

On "why allow some external checks and block others": the blocklist is not arbitrary.

- **MicroShift (blocklist)**: ships as a bundled vendor check in the image RPM. Per the Sam/Frank discussion, that should not be the authority that rolls the whole OS under fleet management. We disable it explicitly.
- **Customer image-baked checks (allowed)**: that is the opposite case. The customer put validation in `required.d` on purpose so their hardware criteria can gate rollback. EDM-5046 is restoring stock greenboot behavior here because we were silently breaking that on every boot. That is a product requirement for this release, not expanding third-party rollback authority beyond what greenboot already provides on image-mode RHEL.

So I do not think customer checks violate the alignment we discussed for MicroShift. The tension is bundled vendor vs customer-owned rollback criteria. Longer term, EDMRFE-82 is the place to make agent-mediated policy explicit; for 1.3 the vendor blocklist plus restoring customer checks is the scoped fix.

If we removed the MicroShift disable entirely, we are back to EDM-3260: undersized/slow boots can roll the OS on MicroShift failure even when the agent is healthy. `boot-complete` wait alone does not change that.

test/packaging/greenboot_functions_test.go
```diff
@@ -0,0 +1,234 @@
+package packaging_test
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
```
eldar101, line 10: DOCSTRINGS plz.
We Also prefer helper funcs at the bottom.

galel12:
Superseded by a replacement PR on the same branch: the team decided on a **full revert** of `configure-greenboot` (no third-party health check blocking) instead of the vendor blocklist approach in this PR. The new PR will be opened immediately after this close.

galel12:
Replacement PR: #3537 — full revert of `configure-greenboot` instead of the vendor blocklist.

PR #3374, opened 2026-08-12
EDM-5070: Fix missingkey mismatch between validation and rendering

talhil-rh:
Validation uses missingkey=zero while rendering uses missingkey=error, causing idiomatic Go template conditionals that guard against missing labels to pass validation but fail at rollout time. Change the renderer to use missingkey=zero so both paths are consistent.

Assisted-by: Claude

### DIFF
--- a/internal/tasks/fleet_rollout_test.go
+++ b/internal/tasks/fleet_rollout_test.go
@@ -1990,3 +1990,70 @@ func TestFleetRolloutIterationContext(t *testing.T) {
 		assert.True(t, errors.Is(iterCtx.Err(), context.Canceled))
 	})
 }
+
+func TestReplaceParametersInString_MissingLabelsInConditionals(t *testing.T) {
+	tests := []struct {
+		name     string
+		template string
+		labels   map[string]string
+		expected string
+	}{
+		{
+			name:     "When if guards a missing label it should take the else branch",
+			template: `{{ if .metadata.labels.env }}env={{ .metadata.labels.env }}{{ else }}env=default{{ end }}`,
+			labels:   map[string]string{},
+			expected: `env=default`,
+		},
+		{
+			name:     "When if guards a present label it should render the value",
+			template: `{{ if .metadata.labels.env }}env={{ .metadata.labels.env }}{{ else }}env=default{{ end }}`,
+			labels:   map[string]string{"env": "prod"},
+			expected: `env=prod`,
+		},
+		{
+			name:     "When if eq references a missing label it should take the else branch",
+			template: `{{ if eq .metadata.labels.env "prod" }}production{{ else }}development{{ end }}`,
+			labels:   map[string]string{},
+			expected: `development`,
+		},
+		{
+			name:     "When with guards a missing label it should take the else branch",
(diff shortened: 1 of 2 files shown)


talhil-rh (commented):
_This review has been retracted._

internal/tasks/fleet_rollout_test.go
```diff
@@ -1990,3 +1990,70 @@ func TestFleetRolloutIterationContext(t *testing.T) {
 		assert.True(t, errors.Is(iterCtx.Err(), context.Canceled))
 	})
 }
+
```
eldar101, line 1993: docstring plz

internal/tasks/fleet_rollout_test.go
```diff
@@ -1990,3 +1990,70 @@ func TestFleetRolloutIterationContext(t *testing.T) {
...
+func TestReplaceParametersInString_MissingLabelsInConditionals(t *testing.T) {
+	tests := []struct {
```
eldar101, line 1995: can we consider struct out of func?

internal/tasks/fleet_rollout_test.go
```diff
@@ -1990,3 +1990,70 @@ func TestFleetRolloutIterationContext(t *testing.T) {
...
+			name:     "When nested if guards missing labels it should take outer else branch",
+			template: `{{ if .metadata.labels.env }}{{ if eq .metadata.labels.env "prod" }}CRITICAL=true{{ else }}CRITICAL=false{{ end }}{{ else }}NO_ENV{{ end }}`,
+			labels:   map[string]string{},
+			expected: `NO_ENV`,
+		},
+		{
+			name:     "When deeply nested conditionals reference present labels it should resolve correctly",
+			template: `{{ if .metadata.labels.env }}{{ if eq .metadata.labels.env "prod" }}{{ if .metadata.labels.region }}{{ .metadata.labels.region }}{{ else }}no-region{{ end }}{{ end }}{{ end }}`,
+			labels:   map[string]string{"env": "prod", "region": "eu-west-1"},
+			expected: `eu-west-1`,
```
eldar101, line 2041: consider moving repeated string to a  const.

avishayt (changes requested):
Putting this on hold — the missingkey mismatch is intentional, not a bug.

This is not related to the introduction of conditionals (EDM-4785). A bare `{{ .metadata.labels.env }}` without any conditional would produce the exact same behavior: validation passes, rendering fails on a device missing that label. This predates conditional support entirely.

The two sides use different settings by design:

- **Validation uses `missingkey=zero`** because fleet templates cannot be strictly validated against device labels — devices and their labels can change at any time after the fleet is created.
- **Rendering uses `missingkey=error`** because silently sending empty/zero values to devices is likely a bug. If the operator intended for a value to be optional, they should set an explicit default using `getOrDefault`.

The `getOrDefault` pattern documented in the ticket as a "workaround" is the correct usage:

```
{{ if eq (getOrDefault .metadata.labels "env" "dev") "prod" }}TIER=production{{ else }}TIER=development{{ end }}
```

Switching rendering to `missingkey=zero` would mask real configuration errors — devices would silently receive empty values instead of getting a clear error that tells the operator they need to add a default.

See comment on EDM-5070 for more detail.

PR #3375, opened 2026-08-13
NO-ISSUE: align package image preload with testcontainers runtime

eldar101:
## Summary
- detect the container runtime for the package helper image preload using the same socket precedence as the Go test harness
- load `quay.io/flightctl/flightctl-device:package` into Podman containers-storage when testcontainers will run on Podman
- keep Docker-based jobs loading the image into the Docker daemon as before

## Why
The package-mode suite starts the helper container from the local `:package` image. In some CI profiles the preload step defaulted to Docker while the Go harness later selected Podman, so testcontainers could not find the local image and tried to pull from Quay instead, failing with `unauthorized` before 90146 even started.

## Testing
- inspected the merged `upstream/main` workflow and harness runtime-selection paths
- kept the change scoped to `.github/workflows/run-e2e-tests.yaml`
- no local end-to-end run here; this needs CI to validate the runtime/image preload path

# Release target
release-1.3
Target release: release-1.3

### DIFF
--- a/.github/workflows/run-e2e-tests.yaml
+++ b/.github/workflows/run-e2e-tests.yaml
@@ -108,6 +108,10 @@ jobs:
           OS_ID: ${{ inputs.os_id }}
         run: |
           set -euo pipefail
+          # Keep runtime detection aligned with test/harness/containers/docker_host.go.
+          . ./test/scripts/detect_container_runtime.sh
+          configure_testcontainers_docker_host
+
           echo "Downloaded test artifacts:"
           tree test-artifacts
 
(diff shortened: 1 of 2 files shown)


galel12 (commented):
Should we export `DOCKER_HOST` once before `skopeo` using the same rules as Go (e.g. a shared `test/scripts/detect_container_runtime.sh` sourced from here), instead of duplicating socket-precedence logic inline in the workflow? That would keep one source of truth with `test/harness/containers/docker_host.go` and let this step stay a few lines.

.github/workflows/run-e2e-tests.yaml
```diff
@@ -108,6 +108,58 @@ jobs:
           OS_ID: ${{ inputs.os_id }}
         run: |
           set -euo pipefail
+
+          detect_testcontainers_runtime() {
```
galel12: This mirrors `test/harness/containers/docker_host.go` (`detectContainerSocket`, `RuntimeCLIName`) — should we add a comment pointing there so the two don't drift silently when precedence changes?

.github/workflows/run-e2e-tests.yaml
```diff
@@ -108,6 +108,58 @@ jobs:
...
+              return 0
+            fi
+            if [[ -n "${DOCKER_HOST:-}" ]]; then
+              echo "docker"
+              return 0
+            fi
+
+            local uid home_dir candidate
+            uid="$(id -u)"
+            home_dir="${HOME:-}"
```
galel12: Minor: bash uses `$HOME` here; Go uses `user.Current().HomeDir`. Unlikely to matter on `flightctl-ci-runner`, but worth being aware of if these ever diverge.

.github/workflows/run-e2e-tests.yaml
```diff
@@ -131,11 +183,12 @@ jobs:
           echo "✓ QCOW2 image: bin/output/qcow2/disk.qcow2"
 
           PACKAGE_REF="quay.io/flightctl/flightctl-device:package"
-          if [[ "${DOCKER_HOST:-}" == *podman* ]]; then
-            echo "Loading package-mode image into Podman containers-storage (DOCKER_HOST=${DOCKER_HOST})"
+          CONTAINER_RUNTIME="$(detect_testcontainers_runtime)"
```
galel12, line 138: Should we log the detected runtime (and resolved `DOCKER_HOST` if set) right before `skopeo copy`? Would make CI triage much easier if preload and the harness disagree again.

galel12 (approved):
LGTM

PR #3376, opened 2026-08-13
NO-ISSUE: fix flaky management cert rotation integration test

asafbennatan:
## Summary
- Fix the flaky integration spec `renews the cert and switches to the renewal signer` by setting `FLIGHTCTL_TEST_CERT_MANAGER_SYNC_INTERVAL=2s` (same as e2e) via `TestTempEnv`, and restarting the agent so cert manager reads the interval once at `Run()`.
- Raise **agent-side** management cert logs to Info (sync interval, renew policy, CSR create). Leave **`pkg/certmanager` infra at Debug** to avoid log spam.

## Root cause
With `renewBefore=TTL-1` and CA `NotBefore=now-1s`, `renewAt ≈ issuance+1s`. The first Sync can land ~0.6s early. Default sync interval is **1h**, so renewal is not retried inside the 60s Eventually.

Confirmed on [CI run 31688978709](https://github.com/flightctl/flightctl/actions/runs/31688978709).

## Scope
- Test deflake: `test/integration/agent/agent_test.go` only for timing.
- Agent logging only under `internal/agent/device/certmanager/` — no `pkg/certmanager` log-level changes.

## Test plan
- [ ] Focused integration: `renews the cert and switches to the renewal signer`
- [ ] Full `integration-tests` CI on this PR

### DIFF
--- a/internal/agent/device/certmanager/provider/management/management.go
+++ b/internal/agent/device/certmanager/provider/management/management.go
@@ -36,15 +36,11 @@ const (
 // ---- ConfigProvider ----
 
 type managementConfigProvider struct {
-	log                         certmanager.Logger
 	renewBeforeExpiryPercentage int32
 }
 
-func NewManagementConfigProvider(log certmanager.Logger, renewBeforeExpiryPercentage int32) *managementConfigProvider {
-	return &managementConfigProvider{
-		log:                         log,
-		renewBeforeExpiryPercentage: renewBeforeExpiryPercentage,
-	}
+func NewManagementConfigProvider(renewBeforeExpiryPercentage int32) *managementConfigProvider {
+	return &managementConfigProvider{renewBeforeExpiryPercentage: renewBeforeExpiryPercentage}
 }
 
 func (p *managementConfigProvider) Name() string { return configProviderName }
@@ -60,23 +56,14 @@ func (p *managementConfigProvider) GetCertificateConfigs() ([]certmanager.Certif
 		},
 	}
 
+	// Apply env overrides (can set both, precedence handled by certmanager)
 	renewBefore, renewBeforePercent := renewPolicyFromEnv()
 	cfg.RenewBefore = renewBefore
 	cfg.RenewBeforePercentage = renewBeforePercent
 
+	// Fall back to provider default only if neither override is set
 	if cfg.RenewBefore == nil && cfg.RenewBeforePercentage == nil {
 		cfg.RenewBeforePercentage = &p.renewBeforeExpiryPercentage
-	}
-
-	switch {
-	case cfg.RenewBefore != nil:
-		p.log.Infof("Management cert renew policy: renewBefore=%s (from env)", *cfg.RenewBefore)
-	case cfg.RenewBeforePercentage != nil:
-		source := "provider default"
-		if renewBeforePercent != nil {
-			source = "env"
-		}
-		p.log.Infof("Management cert renew policy: renewBeforePercentage=%d (from %s)", *cfg.RenewBeforePercentage, source)
 	}
 
 	return []certmanager.CertificateConfig{cfg}, nil
@@ -170,7 +157,7 @@ func (p *managementProvisioner) Provision(ctx context.Context, req certmanager.P
 		}
 
 		p.csrName = csrName
-		p.log.Infof("Created management certificate CSR %q (signer=%s)", p.csrName, managementSignerName)
+		p.log.Debugf("Created management certificate CSR %q", p.csrName)
 
 		return p.poll(ctx, req)
 	}
--- a/internal/agent/device/certmanager/provider/management/management.go
+++ b/internal/agent/device/certmanager/provider/management/management.go
@@ -36,11 +36,15 @@ const (
 // ---- ConfigProvider ----
 
 type managementConfigProvider struct {
+	log                         certmanager.Logger
 	renewBeforeExpiryPercentage int32
 }
 
-func NewManagementConfigProvider(renewBeforeExpiryPercentage int32) *managementConfigProvider {
-	return &managementConfigProvider{renewBeforeExpiryPercentage: renewBeforeExpiryPercentage}
+func NewManagementConfigProvider(log certmanager.Logger, renewBeforeExpiryPercentage int32) *managementConfigProvider {
+	return &managementConfigProvider{
+		log:                         log,
+		renewBeforeExpiryPercentage: renewBeforeExpiryPercentage,
+	}
 }
 
 func (p *managementConfigProvider) Name() string { return configProviderName }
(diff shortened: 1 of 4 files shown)


pkg/certmanager/manager.go
```diff
@@ -401,34 +408,57 @@ func (cm *CertManager) shouldProvisionCertificate(b *bundleRegistry, pk string,
...
 
 	// Bundle-wide kill switch for time-based renewal.
 	if b != nil && b.disableRenewal {
-		cm.log.Debugf(
-			"Certificate %q for providerKey %q: renewal window reached but disabled by bundle %q (renewBefore=%s, notAfter=%s)",
-			cert.Name, pk, b.name, renewBefore, notAfter.Format(time.RFC3339),
+		cm.log.Infof(
+			"Certificate %q for providerKey %q: renewal window reached but disabled by bundle %q (policy=%s renewBefore=%s notAfter=%s)",
+			cert.Name, pk, b.name, policySource, renewBefore, notAfter.UTC().Format(time.RFC3339),
 		)
```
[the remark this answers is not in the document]
asafbennatan: Addressed — disabled-renewal log now includes lifetime, renewAt, and expiresIn.

test/integration/agent/agent_test.go
```diff
@@ -347,6 +351,7 @@ var _ = Describe("Device Agent behavior", func() {
 				defer testutil.TestTempEnv(
 					mgmtCertTTLSecondsEnv, strconv.Itoa(testCertTTLSeconds),
 					mgmtCertRenewBeforeSecondsEnv, strconv.Itoa(testRenewBeforeSeconds),
+					certManagerSyncIntervalEnv, testCertManagerSyncInterval,
```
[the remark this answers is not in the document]
asafbennatan: Addressed — `h.RestartAgent()` now runs after `TestTempEnv` so the agent (and cert manager `Run`) starts with `FLIGHTCTL_TEST_CERT_MANAGER_SYNC_INTERVAL` already set.

Note: cert manager only starts post-bootstrap, which was already after `TestTempEnv`, but restarting makes the ordering explicit and removes the race with harness auto-start.

pkg/certmanager/manager.go
```diff
@@ -401,34 +408,58 @@ func (cm *CertManager) shouldProvisionCertificate(b *bundleRegistry, pk string,
...
+				cert.Name, pk, *cfg.RenewBefore, lifetime,
+			)
+		}
 		renewBefore = lifetime / 3
 	}
 
 	renewAt := notAfter.Add(-renewBefore)
+	expiresIn := notAfter.Sub(now)
 	if now.Before(renewAt) {
+		cm.log.Infof(
```
asafbss: They were debug level for a reason. Info might spam the logs

internal/agent/device/certmanager/provider/management/management.go
```diff
@@ -36,11 +36,15 @@ const (
...
 type managementConfigProvider struct {
+	log                         certmanager.Logger
 	renewBeforeExpiryPercentage int32
 }
 
-func NewManagementConfigProvider(renewBeforeExpiryPercentage int32) *managementConfigProvider {
-	return &managementConfigProvider{renewBeforeExpiryPercentage: renewBeforeExpiryPercentage}
+func NewManagementConfigProvider(log certmanager.Logger, renewBeforeExpiryPercentage int32) *managementConfigProvider {
+	return &managementConfigProvider{
+		log:                         log,
```
asafbss, line 45: Do we actually use this inside the config provider?

pkg/certmanager/manager.go
```diff
@@ -338,8 +338,15 @@ func (cm *CertManager) syncCertificate(ctx context.Context, b *bundleRegistry, p
...
+			cm.log.Infof(
+				"Loaded existing certificate %q for providerKey %q (notBefore=%s notAfter=%s expiresIn=%s)",
+				certName, pk,
+				parsed.NotBefore.UTC().Format(time.RFC3339),
+				parsed.NotAfter.UTC().Format(time.RFC3339),
+				time.Until(parsed.NotAfter),
+			)
 		} else if loadErr != nil {
-			cm.log.Debugf("no existing cert loaded for %q/%q: %v", pk, certName, loadErr)
+			cm.log.Infof("no existing cert loaded for %q/%q: %v", pk, certName, loadErr)
```
galel12: This path runs on every sync while `cert.Config` is still empty (expected before the first cert exists). At Info that's a line per sync per bundle until provisioning succeeds — not just the renewal-evaluation noise on the healthy-cert path. Should this stay at Debug?

test/integration/agent/agent_test.go
```diff
@@ -330,13 +330,17 @@ var _ = Describe("Device Agent behavior", func() {
...
 					// - cert lifetime: 10m
-					// - renewBefore: 10m-1s => renewal condition becomes true almost immediately
-					testCertTTLSeconds     = 10 * 60
-					testRenewBeforeSeconds = testCertTTLSeconds - 1
+					// - renewBefore: 10m-1s => renewAt ≈ issuance+1s (CA NotBefore is now-1s)
+					// - sync every 10s so a first Sync that lands before renewAt still retries
+					//   inside TIMEOUT (default sync is 1h)
+					testCertTTLSeconds          = 10 * 60
+					testRenewBeforeSeconds      = testCertTTLSeconds - 1
+					testCertManagerSyncInterval = "10s"
```
galel12: E2e cert rotation uses `2s` for `FLIGHTCTL_TEST_CERT_MANAGER_SYNC_INTERVAL`. Should we match that here so the retry lands sooner after `renewAt` and the spec is less sensitive to slow runners under the 60s `Eventually`?

test/integration/agent/agent_test.go
```diff
@@ -347,8 +351,13 @@ var _ = Describe("Device Agent behavior", func() {
 				defer testutil.TestTempEnv(
 					mgmtCertTTLSecondsEnv, strconv.Itoa(testCertTTLSeconds),
 					mgmtCertRenewBeforeSecondsEnv, strconv.Itoa(testRenewBeforeSeconds),
+					certManagerSyncIntervalEnv, testCertManagerSyncInterval,
 				)()
 
+				// Harness auto-starts the agent in BeforeEach. Restart after TestTempEnv so
+				// cert manager Run() reads FLIGHTCTL_TEST_CERT_MANAGER_SYNC_INTERVAL (once).
+				h.RestartAgent()
```
galel12, line 359: The deflake looks test-only; the `pkg/certmanager` Info logging is a separate prod behavior change. For `release-1.4`, should we split — land the test fix here and follow up on logging — so the backport doesn't change agent log volume?

pkg/certmanager/manager.go
asafbennatan: Agreed — dropped the Info diagnostics from this PR. Keeping Debug as before; test-only deflake remains.

internal/agent/device/certmanager/provider/management/management.go
asafbennatan: No longer relevant — that logger was only for the Info renew-policy lines. Reverted the config provider to the previous constructor (no logger).

pkg/certmanager/manager.go
asafbennatan: Yep — reverted; that path stays at Debug. Logging is out of this PR.

test/integration/agent/agent_test.go
asafbennatan: Done — switched to `2s` to match e2e cert rotation.

test/integration/agent/agent_test.go
asafbennatan: Agreed — dropped the logging changes. This PR is test-only now (sync interval + restart after `TestTempEnv`) so the backport stays clean.

asafbss:
Provision and Renewal have info logs - they're part of the agent impl.
IMO Keep the certmanager infra debug level intact

galel12 (approved):
LGTM

asafbennatan:
Addressed: restored agent-side Info for sync interval, renew policy, and CSR create under `internal/agent/device/certmanager/`. Left `pkg/certmanager` infra at Debug.

galel12 (approved):

PR #3380, opened 2026-08-16
NO-ISSUE: include remote-access in flightctl.target

asafbennatan:
## Summary
- Add `Wants=` / `After=` for `flightctl-remote-access.service` on `flightctl.target` so the target explicitly starts remote-access with the rest of the stack.
- The unit already has `PartOf=` / `WantedBy=flightctl.target` (stop/restart still worked); without target `Wants=`, start depended on gateway's soft `Wants=` or enable-time `.wants/` symlinks.

## Test plan
- [ ] Confirm `flightctl.target` lists remote-access before gateway in both `Wants=` and `After=`
- [ ] On a quadlet deploy: `systemctl start flightctl.target` starts `flightctl-remote-access.service`
- [ ] `systemctl stop flightctl.target` still stops remote-access (`PartOf=`)

### DIFF
--- a/deploy/podman/flightctl.target
+++ b/deploy/podman/flightctl.target
@@ -25,6 +25,7 @@ Wants=flightctl-certs-init.service
 Wants=flightctl-telemetry-gateway.service
 Wants=flightctl-imagebuilder-api.service
 Wants=flightctl-imagebuilder-worker.service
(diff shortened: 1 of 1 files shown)


galel12 (approved):
LGTM

PR #3402, opened 2026-08-16
EDM-4830: Device decommission decisions in service

asafbennatan:
Jira: EDM-4830

Moves device decommission decisions into the service layer, using the store-native `Mutate` CAS RMW introduced on `main` (EDM-5028).

## Summary

- Service `DecommissionDevice` applies Spec/Status/Owner/Labels changes inside a `Mutate` callback against the CAS-fresh device
- Refuses a second decommission with `ErrResourceVersionConflict` and does not emit a success event
- Removes store-owned `DecommissionDevice` (precondition + mutation + event callback)
- Unit + integration coverage for success, already-decommissioning conflict, and not-found

Already covered on `main` (not re-done here): Replace/Patch/Update reject already-decommissioning devices inside their `Mutate` callbacks.

## Release target (required before merge to `main`)

- [x] **`release-1.3`**

---
**Stack:** bottom of the EDM-4806 store business-logic closeout stack.  
Base: `main` (includes Mutate CAS).

Made with [Cursor](https://cursor.com)

PR #3403, opened 2026-08-16
EDM-4832: Condition merge in the service layer

asafbennatan:
Jira: EDM-4832

Moves device service-condition merge and no-op decisions into the service layer using store-native `Mutate` CAS (EDM-5028).

## Summary

- `SetDeviceServiceConditions` merges via `common.MergeStatusConditions` inside a `Mutate` apply callback
- Unchanged merges return `ErrMutateSkipWrite` (no RV bump, no events) — same pattern as fleet on `main`
- Removes store `SetServiceConditions` (merge + callback lived there)
- Fleet `UpdateFleetConditions` shares `MergeStatusConditions` (already Mutate-based on `main`)
- Unit + service integration coverage for change / no-op / not-found

**Note:** CSR `UpdateConditions` remains an unused store API with in-store merge; leave for a later cleanup (no production callers).

- [x] **`release-1.3`**

---
**Stack:** depends on #3402 (EDM-4830 Mutate-based decommission).

### DIFF
--- a/internal/service/device/handler.go
+++ b/internal/service/device/handler.go
@@ -870,12 +870,61 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
 }
 
 func (h *DeviceServiceHandler) SetDeviceServiceConditions(ctx context.Context, orgId uuid.UUID, name string, conditions []domain.Condition) domain.Status {
-	callback := func(ctx context.Context, orgId uuid.UUID, device *domain.Device, oldConditions, newConditions []domain.Condition) {
-		h.diffAndEmitConditionEvents(ctx, orgId, device, oldConditions, newConditions)
+	var oldConditions, newConditions []domain.Condition
+	result, _, _, err := h.deviceStore.Mutate(ctx, orgId, name, nil, func(m *devicestore.DeviceMutation) error {
+		if err := m.RequireExisting(); err != nil {
+			return err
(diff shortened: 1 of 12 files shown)


internal/service/device/handler.go
```diff
@@ -870,12 +870,61 @@ func (h *DeviceServiceHandler) UpdateRenderedDevice(ctx context.Context, orgId u
...
+			return err
+		}
+		existing := serviceConditionsFromDevice(m.Device)
+		merged, changed := common.MergeStatusConditions(existing, conditions)
+		if !changed {
+			return store.ErrMutateSkipWrite
+		}
+		oldConditions = existing
+		newConditions = merged
+		replaceServiceConditionsOnDevice(m.Device, merged)
```
EfratIfergan, line 885: Could we add a test covering preservation of agent-owned conditions when updating service conditions?

`replaceServiceConditionsOnDevice` is responsible for replacing only the service-owned conditions while preserving the agent-owned conditions. We have a unit test for `serviceConditionsFromDevice`, but I don't see a test covering this behavior through `SetDeviceServiceConditions`.

For example, start with one agent condition and one service condition, update the service condition, and assert that the agent condition is still present and unchanged.

asafbennatan: added

EfratIfergan (approved):
LGTM

PR #3404, opened 2026-08-16
EDM-4833: Rendered-version / hash decisions in service

asafbennatan:
Jira: EDM-4833

Moves the rendered skip-vs-write decision into `device_render` and removes the rendered-hash short-circuit from `DeviceStore`.

- [x] **`release-1.3`**

---
**Stack:** 6/12 in the EDM-4806 closeout stack (based on `main`; depends on and should merge after #3403 — EDM-4832).

### DIFF
--- a/internal/service/device/rendered.go
+++ b/internal/service/device/rendered.go
@@ -28,9 +28,7 @@ func applyRenderedUpdate(
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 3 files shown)


asafbennatan:
Dropping this PR from the stack. After the store skip-vs-write move landed on main, this layer only extracted shouldPersistRenderedUpdate and is not worth a separate review. Next in the stack is #3405.

PR #3406, opened 2026-08-16
EDM-4844: Enforce service-only access to resource stores

asafbennatan:
Jira: EDM-4844

Adds depguard rules enforcing service-only imports of resource stores, and clears the resulting breakers by rewiring remaining direct store call sites through their owning services.

**Refactor-only layer** — behavioral changes (OS status evaluation, enrollment context check) are split into #3525 for isolated review.

- [x] **`release-1.3`**

---

**Stack:** 8/13 in the EDM-4806 closeout stack (based on `main`; depends on and should merge after #3405 — EDM-4834). Behavioral follow-up: #3525.


asafbss (commented):
The managed-device OutOfDate error message logic changed behavior from the original `managedDeviceOutOfDateMessage` in `common/device.go`.

**Original** (`managedDeviceOutOfDateMessage`):
1. Check `LastRolloutError` annotation **first** — if present, return immediately
2. Then check update condition for `UpdateStateError`
3. Fallback to generic `DeviceOutOfSyncWithFleetText`

**New** (inlined in `status.go`):
1. Check update condition for `UpdateStateError` **first**
2. `else if` check annotation — only if condition was NOT found at all
3. Fallback to generic text

Two issues:
1. **Priority inversion** — annotations used to win over conditions. Now conditions win. If both exist with different messages, users see a different error.
2. **`else if` makes annotation unreachable when a condition exists but isn't an error** — if there's an update condition with reason `InProgress` (not `UpdateStateError`), the old code would fall through and check the annotation. The new `else if` skips the annotation entirely, so the user gets the generic "out of sync" message instead of the specific rollout error.

asafbss (commented):
The `overrideUpToDateForOsTarget` function in `status.go` is a behavioral change, not a refactoring move — it adds new logic that marks devices OutOfDate when they have a catalog OS target they can't satisfy. Can you split it into its own PR? Behavioral changes buried in a 89-file refactoring are easy to miss (case in point — the error message priority inversion above).

asafbss (approved):

PR #3412, opened 2026-08-16
4.2 multiple vms

itzikb-redhat:
Add an e2e test that runs test-vm-a and test-vm-b on one device using KubeVirt ConfigDrive (userDataBase64).
Covering:
 - TCP/UDP publishPorts
 - password SSH
 - ConfigDrive userdata (authorized keys, packages, write_files, runcmd)
 - Isolation when one VM is stopped. Extend the harness with NewVmApplicationSpecFromYAML and RunUDPProbeOnDeviceLocalPort for custom VM manifests and UDP checks.

Assisted-by: Cursor

### DIFF
--- a/test/harness/e2e/harness_application.go
+++ b/test/harness/e2e/harness_application.go
@@ -20,9 +20,19 @@ const (
 	// QuadletUnitPath is the quadlet unit path on device (rootful)
 	QuadletUnitPath = "/etc/containers/systemd"
 
-	// vmYAMLTemplate is a KubeVirt VirtualMachine manifest template for e2e VM tests.
-	// Placeholders: 1=name, 2=containerdisk image.
-	vmYAMLTemplate = `apiVersion: kubevirt.io/v1
+	// VMGuestMemoryDefault is the guest RAM for e2e KubeVirt VM manifests.
+	VMGuestMemoryDefault = "1024M"
+)
+
+const vmFedoraNoCloudUserData = `#cloud-config
+ssh_pwauth: true
+password: fedora
+chpasswd: { expire: False }`
+
+// VMYAML builds a KubeVirt VirtualMachine manifest for e2e tests. cloudInitVolumeYAML
+// is the cloudinitdisk volume entry (including list-item indentation under volumes)
+func VMYAML(name, guestMemory, image, cloudInitVolumeYAML string) string {
+	return fmt.Sprintf(`apiVersion: kubevirt.io/v1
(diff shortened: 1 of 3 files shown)


test/harness/e2e/harness_application.go
```diff
@@ -1176,37 +1211,87 @@ func (h *Harness) RunSSHOnDeviceLocalPort(port int, user, password string, remot
...
+  printf '%%s\n' "$msg" | socat -t2 - UDP4:127.0.0.1:%d
+elif command -v nc >/dev/null 2>&1; then
+  printf '%%s\n' "$msg" | nc -u -w 2 127.0.0.1 %d
+else
+  exit 127
+fi`, quotedMessage, port, port)
+	out, err := h.VM.RunSSH([]string{"/bin/sh", "-c", script}, nil)
+	if err != nil {
+		return "", fmt.Errorf("running UDP probe on device host localhost:%d (socat or nc required): %w", port, err)
+	}
```
asafbennatan: This helper prefers `socat`, then falls back to `nc -u -w 2`. The cs9-bootc overlay never installs `socat`, so on a typical e2e device this will skip to ncat. This same spec already switched the in-guest probe to python3 because `socat`/`nc` are unreliable for UDP echo on ncat. Should the device-host probe use that same python3 send/recv path so published-UDP does not depend on `socat` being present or on ncat UDP?

```suggestion
	quotedMessage := shellQuote(message)
	script := fmt.Sprintf(`set -eu
command -v python3 >/dev/null 2>&1 || exit 127
python3 -c 'import socket,sys; port=int(sys.argv[1]); msg=sys.argv[2]; s=socket.socket(socket.AF_INET, socket.SOCK_DGRAM); s.settimeout(2); s.sendto(msg.encode(), ("127.0.0.1", port)); print(s.recv(1024).decode().strip())' %d %s`, port, quotedMessage)
	out, err := h.VM.RunSSH([]string{"/bin/sh", "-c", script}, nil)
	if err != nil {
		return "", fmt.Errorf("running UDP probe on device host localhost:%d (python3 required): %w", port, err)
	}
```

asafbennatan (commented):
See comments below

test/harness/e2e/harness_application.go
itzikb-redhat: @asafbennatan  Modified. Please let me know if this is what you meant

asafbennatan (approved):
LGTM

PR #3414, opened 2026-08-17
EDM-4115: Add VM e2e tests for spec updates and crash recovery

itzikb-redhat:
Cover modified VM application specs (90233) and unexpected virt-launcher compute crash recovery (90232).

- Verify re-render and re-apply when memory, CPU, publishPorts, containerdisk image, and cloud-init password change, including guest resource checks and SSH on the new published port.

- Verify that killing the compute container returns the app to Running/Healthy without operator start. Extract waitForVMAppRunningHealthy for all VM tests and add VMYAMLWithCPU and VMFedoraNoCloudUserData harness helpers.

Assisted-by: Cursor

### DIFF
--- a/test/e2e/vm/vm_test.go
+++ b/test/e2e/vm/vm_test.go
@@ -2,8 +2,12 @@ package vm_test
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
+	"io"
 	"os"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/flightctl/flightctl/api/core/v1beta1"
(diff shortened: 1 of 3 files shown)


asafbennatan (commented):
See comments below

test/e2e/vm/vm_test.go
```diff
@@ -248,8 +287,79 @@ var _ = Describe("VM Applications", Ordered, func() {
...
+			device.Spec.Applications = &[]v1beta1.ApplicationProviderSpec{modifiedSpec}
+		})
+		Expect(err).ToNot(HaveOccurred())
+		waitForVMAppRunningHealthy(harness, deviceID, vmAppName)
+
+		By("Verifying SSH on the previous published port is removed")
+		expectSSHUnavailableOnPort(harness, vmPublishedSSHPort, vmAppName, vmGuestUser, vmModifiedGuestPassword)
+
+		By("Verifying SSH on the new published port with the updated password")
+		expectSSHWhoamiWithPassword(harness, vmBPublishedSSHPort, vmAppName, vmGuestUser, vmModifiedGuestPassword)
```
asafbennatan: This check treats any SSH failure on the previous published port as proof that forwarding was removed, and it runs before SSH is known to work on the new port with the new password. After `waitForVMAppRunningHealthy`, the guest can still be restarting, so an auth failure against the old guest looks the same as connection refused — then the 10s `Consistently` window can pass before 2223 is proven. Should we assert SSH on the new port first, then probe 2222 with the same credentials so a leftover mapping would succeed and fail the test?

```suggestion
		By("Verifying SSH on the new published port with the updated password")
		expectSSHWhoamiWithPassword(harness, vmBPublishedSSHPort, vmAppName, vmGuestUser, vmModifiedGuestPassword)

		By("Verifying SSH on the previous published port is removed")
		expectSSHUnavailableOnPort(harness, vmPublishedSSHPort, vmAppName, vmGuestUser, vmModifiedGuestPassword)
```

itzikb-redhat: Modified and added check of errors

asafbennatan (approved):
LGTM

PR #3418, opened 2026-08-18
EDM-3772: stabilize rollout disruption budget test

eldar101:
## Summary
This PR fixes two issues in the rollout disruption budget E2E test.

First, the test was sending `minAvailable: 0`. The API schema rejects this now because `minAvailable` must be at least `1` when it is set. For this test case we only need `maxUnavailable: 2`, so the fix is to leave `minAvailable` unset.

Second, the test was flaky because it checked the disruption budget only once and assumed all devices would move to the same next rendered version. The test now tracks the expected rendered version per device and keeps watching the unavailable-device count during the rollout.

## Release target
release-1.3
Target release: release-1.3

### DIFF
--- a/test/e2e/rollout/rollout_test.go
+++ b/test/e2e/rollout/rollout_test.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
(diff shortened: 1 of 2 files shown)


hferber1 (approved):
lgtm

PR #3422, opened 2026-08-18
EDM-4115: Add test for coexisting vm app and container app

itzikb-redhat:


test/e2e/vm/vm_test.go
```diff
@@ -42,6 +61,13 @@ func getVMImage() string {
 	return defaultVMImage
 }
```
eldar101, line 63: docstring plz

eldar101: also funcs at the bottom if possible.

PR #3424, opened 2026-08-18
EDM-5214: Agent discovers and applies an OS delta, or falls back to a full pull

asafbennatan:

- [x] **`release-1.4`**

## EDM-5214: Agent discovers and applies an OS delta, or falls back to a full pull

**Jira:** https://issues.redhat.com/browse/EDM-5214
**Story type:** [DEV]
**Stack:** on [EDM-5213](https://github.com/flightctl/flightctl/pull/3425); followed by [EDM-5215](https://github.com/flightctl/flightctl/pull/3427)

### Summary

Image-mode devices that can consume OCI deltas **and** have bootc (`systemInfo.deltaEligible` and a non-empty `bootcVersion`) discover one OS delta (spec hint, else OCI Referrers on the target image's own repository with Tag Schema on 404), pull it with the existing pull-config auth path, apply it with `oci-delta apply --ostree-repo /ostree/repo`, load/tag the reconstructed image into containers-storage, and continue the existing `bootc switch` path. Pull or apply failure sets `status.os.lastDelta.fallbackReason` and emits the full-image prefetch target in the same `CollectOCITargets` call. No candidate leaves the reason unset.

### Changes

- Skopeo `ListReferrers` / `InspectDigest` with Referrers Tag Schema fallback
- Pure candidate selection (`hint` or `targetRepo@digest` matching artifactType + source digest)
- `oci-delta apply` client wrapper
- `podman load` / `podman tag` for the reconstructed archive
- OS manager `canApplyOSDelta` (oci-delta + bootc) before CollectOCITargets discover → pull → apply → load → cleanup, or same-cycle full pull
- `Status()` writes `status.os.lastDelta` every collect; reason is mutex-guarded

### Testing

- **Unit tests:** Referrers vs Tag Schema, candidate filtering, apply argv, load/tag, CollectOCITargets skip / hint / discover / pull fail / apply fail / no-candidate / cleanup; skip apply when bootc is missing
- **Integration tests:** N/A (no agent registry testcontainer; e2e is EDM-5225)

### Acceptance Criteria

- [ ] AC-1: Hint is the only candidate; no Referrers for that pair
- [ ] AC-2: No hint: Referrers on target repo; 404 → Tag Schema; artifactType + source digest
- [ ] AC-3: A delta in a different repository is not discovered
- [ ] AC-4: Pull via pull-config/auth; `oci-delta apply --ostree-repo`; continue existing OS path
- [ ] AC-5: No candidate: full pull; do not set lastDelta fallbackReason
- [ ] AC-6: Pull/apply fail: concrete reason; same-cycle full pull; update does not fail because of the delta
- [ ] AC-7: Delete consumed delta artifact after successful apply
- [ ] AC-8: Discovery does not write control-plane generation rows


internal/agent/device/os/os.go
```diff
@@ -141,30 +186,125 @@ func (m *manager) CollectOCITargets(ctx context.Context, current, desired *v1bet
...
+					PullPolicy:   v1beta1.PullIfNotPresent,
+					ClientOptsFn: optsFn,
+				},
+			},
 		},
-	}, nil
+	}
+}
+
+func (m *manager) startImageAttempt(osImage string) {
```
galel12, line 231: `startImageAttempt` clears `fallbackReason` when the desired image changes, but not `stagedDeltaImage`. After staging a delta for image A, a spec change to B (then back to A before reboot) can leave `stagedDeltaImage == A` while bootc was restaged for B via `Switch(B)` in `AfterUpdate`.

That makes `CollectOCITargets` skip the pull for A and `Reboot` call `RebootStaged` for the wrong staged deployment. Should `startImageAttempt` also reset `stagedDeltaImage` when `lastAttemptedImage != osImage`?

internal/agent/device/os/os.go
```diff
@@ -141,30 +186,125 @@ func (m *manager) CollectOCITargets(ctx context.Context, current, desired *v1bet
...
+		m.log.Debugf("OS delta referrers unavailable: %v", err)
+		return ""
+	}
+	return selectOSDeltaCandidate(nil, desired.Os.Image, sourceDigest, index)
+}
+
+func (m *manager) pullAndApplyOSDelta(ctx context.Context, candidate, osImage string, optsFn dependency.ClientOptsFn) error {
+	tmpDir, err := m.readWriter.MkdirTemp(osDeltaTempPrefix)
+	if err != nil {
+		return m.failApply(err)
```
galel12: `MkdirTemp` failures go through `failApply`, which sets `fallbackReason` to `"delta apply failed"`. That's misleading for a temp-dir error and could confuse operators reading `status.os.lastDelta`. Should infra failures use a separate reason, or skip setting `fallbackReason` entirely?

internal/agent/device/os/os.go
```diff
@@ -141,30 +186,125 @@ func (m *manager) CollectOCITargets(ctx context.Context, current, desired *v1bet
...
+}
+
+func (m *manager) setFallbackReason(reason string) {
+	r := reason
+	m.mu.Lock()
+	m.fallbackReason = &r
+	m.mu.Unlock()
 }
 
 func (m *manager) AfterUpdate(ctx context.Context, desired *v1beta1.DeviceSpec) error {
```
galel12, line 300: `CollectOCITargets` has thorough table tests, but the staged-delta branches in `AfterUpdate` (skip `Switch`) and `Reboot` (`RebootStaged` vs `Apply`) have no dedicated tests. A small table test for those two methods would lock in the post-delta boot path.

internal/agent/client/podman.go
```diff
@@ -293,6 +293,42 @@ func (p *Podman) ImageDigest(ctx context.Context, image string) (string, error)
 	return digest, nil
 }
 
+// LoadArchive loads an OCI archive into local container storage and returns the loaded image reference.
+func (p *Podman) LoadArchive(ctx context.Context, archivePath string) (string, error) {
```
galel12: `LoadArchive` and `Tag` are added with unit tests, but nothing in `os.manager` calls them. The PR description and design talk about load/tag into containers-storage; the implementation uses `SwitchOCI` + `SwitchRegistry` instead.

If that's intentional, should the unused Podman helpers be dropped (or deferred to a follow-up) to avoid dead code?

internal/agent/client/bootc.go
```diff
@@ -130,3 +154,8 @@ func isJSON(s string) bool {
 	var js json.RawMessage
 	return json.Unmarshal([]byte(s), &js) == nil
 }
+
+func stagedDespiteExit(stdout, stderr string) bool {
```
galel12, line 158: Treating any non-zero exit as success when stdout/stderr contains `"Queued for next boot"` could mask unrelated bootc failures that happen to include that string. Should this be narrowed (e.g. match a more specific bootc status line, or only apply for certain exit codes)?

internal/agent/device/os/os_test.go
```diff
@@ -118,3 +160,346 @@ func TestManagerStatusWhenClientFails(t *testing.T) {
...
+			name:    "When hint is set it should pull and apply the hint without Referrers",
+			caps:    Capabilities{OsMode: v1beta1.OsModeImage, DeltaEligible: true, BootcVersion: "bootc 1.15.0"},
+			desired: desiredSpec(testDesiredImage, lo.ToPtr(testHintedDelta)),
+			setup: func(t *testing.T, mockExec *executer.MockExecuter, mockClient *MockClient, mockResolver *dependency.MockPullConfigResolver) {
+				mockClient.EXPECT().Status(gomock.Any()).Return(bootcStatus(testBootedImage, testSourceDigest), nil)
+				expectImageExists(mockExec, testDesiredImage, false)
+				expectPullConfig(t, mockResolver)
+				expectDeltaSuccess(t, mockExec, mockClient, testHintedDelta)
+			},
+			wantEmpty:     true,
```
galel12, line 261: Successful delta cases assert `wantEmpty` but never check `m.stagedDeltaImage == testDesiredImage` after `CollectOCITargets`. Worth asserting so the staged-state wiring doesn't regress silently.

internal/agent/device/os/os.go
asafbennatan: Yes. `startImageAttempt` now runs before the staged-delta shortcut, and it clears `stagedDeltaImage` whenever the desired image is not the one we staged. A→B→A no longer skips the pull or calls `RebootStaged` for a deployment that was restaged for B.

internal/agent/device/os/os.go
asafbennatan: Skipped setting `fallbackReason` for `MkdirTemp`. That path now returns the error as-is so operators do not see `"delta apply failed"` for a temp-dir problem, and the next cycle can retry the delta instead of treating it as a delta apply failure.

internal/agent/device/os/os.go
asafbennatan: Added `TestAfterUpdateAndReboot` covering skip-`Switch` when the matching image is staged, `Switch` when it is not, `RebootStaged` vs `Apply`, and `Apply` when a different image is staged.

internal/agent/client/podman.go
asafbennatan: Intentional: staging goes through `SwitchOCI` + `SwitchRegistry`. Removed `LoadArchive`, `Tag`, and their tests so they are not left as dead code.

internal/agent/client/bootc.go
asafbennatan: Narrowed the match to `Queued for next boot:` (with the colon) so a random error that happens to contain the unpunctuated phrase is not treated as success.

internal/agent/device/os/os_test.go
asafbennatan: Successful delta cases now assert `stagedDeltaImage` (and the A-staged-for-B case asserts it was cleared).

internal/agent/device/os/os.go
```diff
@@ -141,30 +187,126 @@ func (m *manager) CollectOCITargets(ctx context.Context, current, desired *v1bet
...
+	}
+
+	if err := m.client.SwitchOCI(ctx, layoutDir); err != nil {
+		return m.failApply(err)
+	}
+	if err := m.client.SwitchRegistry(ctx, osImage); err != nil {
+		return m.failApply(err)
+	}
+	m.mu.Lock()
+	m.stagedDeltaImage = osImage
```
galel12, line 282: Round 1 items look addressed. One follow-up: when a delta attempt on image X already set `fallbackReason` (pull/apply failure) and a later collect for the same X succeeds, `stagedDeltaImage` is set here but `fallbackReason` is left intact because `startImageAttempt` only clears it on image change.

`Status()` would keep reporting `lastDelta.fallbackReason` even though the delta path ultimately succeeded. Should a successful stage clear `fallbackReason` (under the same mutex as `stagedDeltaImage`)?

internal/agent/device/os/os_test.go
```diff
@@ -118,3 +160,427 @@ func TestManagerStatusWhenClientFails(t *testing.T) {
...
+				return
+			}
+			mockClient.EXPECT().Status(gomock.Any()).Return(bootcStatus(testBootedImage, testSourceDigest), nil)
+			require.NoError(t, m.Status(context.Background(), status))
+			require.Equal(t, tt.wantReason, osLastDeltaFallback(status))
+		})
+	}
+}
+
+func TestAfterUpdateAndReboot(t *testing.T) {
```
galel12, line 455: `TestAfterUpdateAndReboot` covers `Reboot` when a different image is staged (`Apply`), but not the symmetric `AfterUpdate` case — when `stagedDeltaImage != desired.Os.Image`, `AfterUpdate` should call `Switch` to restage bootc for the new image. Worth adding so the two boot paths stay in sync.

internal/agent/device/os/os.go
asafbennatan: Yes. A successful stage now clears `fallbackReason` under the same lock as `stagedDeltaImage`, so `Status()` does not keep reporting a pull/apply failure after the delta path succeeds. Covered by a CollectOCITargets case that starts with a previous fallback reason.

internal/agent/device/os/os_test.go
asafbennatan: Added the matching AfterUpdate case: when `stagedDeltaImage` is a different image, AfterUpdate calls `Switch` for the desired image.

internal/agent/device/os/os.go
```diff
@@ -141,30 +187,126 @@ func (m *manager) CollectOCITargets(ctx context.Context, current, desired *v1bet
...
+		m.log.Debugf("OS delta referrers unavailable: %v", err)
+		return ""
+	}
+	return selectOSDeltaCandidate(nil, desired.Os.Image, sourceDigest, index)
+}
+
+func (m *manager) pullAndApplyOSDelta(ctx context.Context, candidate, osImage string, optsFn dependency.ClientOptsFn) error {
+	tmpDir, err := m.readWriter.MkdirTemp(osDeltaTempPrefix)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
asafbennatan: Wrapped as `create delta temporary directory: %w`.

galel12 (approved):
LGTM

PR #3425, opened 2026-08-18
EDM-5213: Report OS delta consumer fields on systemInfo

asafbennatan:
**Jira:** https://issues.redhat.com/browse/EDM-5213
**Story type:** [DEV]
**Stack:** bottom of EDM-5213 → [EDM-5214](https://github.com/flightctl/flightctl/pull/3424) → [EDM-5215](https://github.com/flightctl/flightctl/pull/3427)

### Summary

Adds optional v1beta1 OS delta fields and has the agent report whether it can **consume OCI deltas**. `status.systemInfo` carries `deltaEligible` (oci-delta on PATH), `bootcVersion` (when bootc is present), and `ociDeltaVersion`. `spec.os.deltaImage` and `status.os.lastDelta` (`fallbackReason`, `size`) are additive.

`DeviceCapabilities` remains **osMode only**. Enrollment still sends `spec.osMode`; delta fields are copied onto the new device via `ApplyDeltaSystemInfo` on `status.systemInfo` (before enrollment labels). Specs are JSONB, so existing rows do not need a migration.

### Testing

- Unit tests for schema JSON, oci-delta/bootc probing on OS clients, Status reporting, and enrollment copy onto systemInfo
- Store/service integration tests for field persistence and enrollment device creation


keitwb (changes requested):
Please don't add to `metadata.capabilities` yet until we have nailed down the design for the device capabilties featuer. It is more likely that with Kyle's dynamic reporting stuff and various design considerations that this will need to go under `systemInfo` instead.

internal/agent/device/os/client.go
```diff
@@ -32,6 +35,76 @@ func DetectMode(lookPath func(string) (string, error)) v1beta1.OsModeType {
...
+
+// DetectDeltaEligible reports whether this agent can apply OS deltas.
+// It is true only when lookPath finds both oci-delta and bootc and
+// bootcVersion parses as 1.15.0 or newer. Missing binaries, version
+// errors, and unparsable output return false.
+//
+// lookPath should be exec.LookPath in production. bootcVersion should
+// run `bootc --version` using the given context. Both are injected so
+// tests do not need PATH or a bootc binary.
+func DetectDeltaEligible(ctx context.Context, lookPath func(string) (string, error), bootcVersion func(context.Context) (string, error)) bool {
```
kkyrazis: I know this isn't necessarily your code, but is there a reason that these (DetectDetlaEligible and DetectMode) have to be package level functions instead of on the Client? That factory already produces a (mostly) capabilities aware struct.  I feel like it would be pretty simple to add a `Capabilities` method to the interface. 

Not a blocker though.

api/core/v1beta1/openapi.yaml
```diff
@@ -6289,7 +6289,21 @@ components:
         - metadata
         - items
     DeviceOsSpec:
-      $ref: '#/components/schemas/ImageOrCatalogItemRefSpec'
+      description: Either a specific OCI image reference, or a reference to a catalog item version that can be resolved to an OCI image ref.
+      allOf:
+        - $ref: '#/components/schemas/ImageOrCatalogItemRefSpec'
+        - type: object
+          properties:
+            deltaImage:
```
kkyrazis, line 6297: deltaImage is an OCI artifact reference right? should DeviceSpec.Validate ensure it's a valid reference simlilar to how it validates `spec.os.image`.

asafbennatan:
> Please don't add to `metadata.capabilities` yet until we have nailed down the design for the device capabilties featuer. It is more likely that with Kyle's dynamic reporting stuff and various design considerations that this will need to go under `systemInfo` instead.

the PR description is stale , this doesnt add anything - capabilities are already reported on ER under spec.deviceStatus.capabilities

this just adds a new capability for that

api/core/v1beta1/openapi.yaml
```diff
@@ -6617,7 +6633,31 @@ components:
       properties:
         osMode:
           $ref: '#/components/schemas/OsModeType'
+        deltaEligible:
```
keitwb: This is the field I'm talking about. This should be changed to be reported under `status.systemInfo` instead. `osMode` is probably going to be transitioned there and the `capabilities` field deprecated with 1.4 and removed in 1.6.

internal/agent/device/os/client.go
asafbennatan: Moved onto the client: `Client.Capabilities(ctx)` returns mode and delta eligibility from the bootc / rpm-ostree / dummy implementations. `NewClient` still picks the implementation; the agent just calls `osClient.Capabilities(ctx)`.

api/core/v1beta1/openapi.yaml
asafbennatan: `DeviceSpec.Validate` now runs the same OCI reference check on `spec.os.deltaImage` as on `spec.os.image` (including fleet templates).

internal/agent/device/os/client.go
```diff
@@ -83,12 +162,18 @@ func (b *bootc) Apply(ctx context.Context) error {
...
+		bootcVersion:    versionCmd(exec, "bootc"),
+		ociDeltaVersion: versionCmd(exec, "oci-delta"),
 	}
 }
 
 type rpmOSTree struct {
-	client *client.RPMOSTree
+	client          *client.RPMOSTree
+	lookPath        func(string) (string, error)
+	bootcVersion    func(context.Context) (string, error)
```
kkyrazis: Why did `rpmOSTree` and `dummyClient` get these new properties? I thought bootc was a requirement for this new capability (OS level) and if bootc is available the expectation is that it manages the OS right?

kkyrazis (approved):

keitwb (approved):

PR #3430, opened 2026-08-19
EDM-4115: Test Lifecycle apps

itzikb-redhat:
Add lifecycle apps test

Bring up microshift and launch 4 apps with types: container, quadlet, compose, helm, and vm

For each app, the test does six CLI actions:
**Stop** — wait until Stopped, confirm the workload is gone, confirm the other four are still up.
**Stop again** — already stopped; status must stay Stopped.
**Restart** while stopped — must be a no-op; status stays Stopped for 5s.
**Start** — wait until Running, confirm the workload is up, confirm the others are still Running.
**Start again** — already running; status must stay Running.
**Restart while running** — bounce it, wait until Running again, confirm the workload and the others.

### DIFF
--- a/test/e2e/applications/helm/lifecycle_test.go
+++ b/test/e2e/applications/helm/lifecycle_test.go
@@ -1,0 +1,612 @@
+package helm_test
+
+import (
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/flightctl/flightctl/api/core/v1beta1"
+	"github.com/flightctl/flightctl/test/e2e/infra/auxiliary"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	lifecycleContainerApp       = "nginx"
+	lifecycleContainerURL       = "http://127.0.0.1:8080/"
+	lifecycleQuadletApp         = "nginx-quadlet"
+	lifecycleQuadletURL         = "http://127.0.0.1:8081/"
+	lifecycleComposeApp         = "compose-demo"
+	lifecycleHelmApp            = "helm-demo"
+	lifecycleHelmNamespace      = "test-app"
+	lifecycleVMApp              = "test-vm"
+	lifecycleVMSSHPort          = 2222
+	lifecycleVMUser             = "fedora"
+	lifecycleVMPassword         = "fedora"
+	lifecycleNoopWindow         = 5 * time.Second
+	lifecycleAlpineImage        = "quay.io/flightctl-tests/alpine:v1"
+	lifecycleDefaultVMImage     = "quay.io/containerdisks/fedora:40"
+	lifecycleCurlConnectTimeout = "2"
+	lifecycleCurlMaxTime        = "5"
+)
+
+var _ = Describe("Application lifecycle stop start restart", Label("microshift"), func() {
+	var (
+		harness      *e2e.Harness
+		deviceID     string
+		helmChartRef string
+		microshiftOs = &v1beta1.DeviceOsSpec{Image: util.NewDeviceImageReference(util.DeviceTags.V12).String()}
+	)
+
+	BeforeEach(func() {
+		harness = e2e.GetWorkerHarness()
+		deviceID, _ = harness.EnrollAndWaitForOnlineStatus()
+		services := auxiliary.Get(harness.GetTestContext())
+		helmChartRef = fmt.Sprintf("%s:%s/flightctl/%s", services.Registry.Host, services.Registry.Port, helmChartV1)
+	})
+
+	// App types: container, quadlet, compose, helm, and vm.
+	It("stops, starts, and restarts each app type without taking the others down", Label("slow", "vm", "90240"), func() {
+		apps := []lifecycleApp{
+			{
+				name:     lifecycleContainerApp,
+				up:       func() { expectHTTP(harness, lifecycleContainerURL, true) },
+				down:     func() { expectHTTP(harness, lifecycleContainerURL, false) },
+				alive:    func() { expectHTTP(harness, lifecycleContainerURL, true) },
+				identity: func() (string, error) { return podmanWorkloadIdentity(harness, "root", lifecycleContainerApp) },
+			},
+			{
+				name:  lifecycleQuadletApp,
+				up:    func() { expectHTTP(harness, lifecycleQuadletURL, true) },
+				down:  func() { expectHTTP(harness, lifecycleQuadletURL, false) },
+				alive: func() { expectHTTP(harness, lifecycleQuadletURL, true) },
+				// EDM-3440: podman as the runAs user fails on root-owned ~/.config.
+				identity: func() (string, error) {
+					return systemdUserWorkloadIdentity(harness, flightctlUser, lifecycleQuadletApp)
+				},
+			},
+			{
+				name:     lifecycleComposeApp,
+				up:       func() { expectComposeWorkloadUp(harness) },
+				down:     func() { expectComposeWorkloadDown(harness) },
+				alive:    func() { expectComposeWorkloadUp(harness) },
+				identity: func() (string, error) { return podmanWorkloadIdentity(harness, "root", lifecycleComposeApp) },
+			},
+			{
+				name:     lifecycleHelmApp,
+				up:       func() { expectHelmPodsReady(harness, lifecycleHelmNamespace) },
+				down:     func() { expectHelmPodsGone(harness, lifecycleHelmNamespace) },
+				alive:    func() { expectHelmPodsPresent(harness, lifecycleHelmNamespace) },
+				identity: func() (string, error) { return helmWorkloadIdentity(harness, lifecycleHelmNamespace) },
+			},
+			{
+				name:  lifecycleVMApp,
+				up:    func() { expectVMSSHUp(harness) },
+				down:  func() { expectVirtLauncher(harness, lifecycleVMApp, false) },
+				alive: func() { expectVirtLauncher(harness, lifecycleVMApp, true) },
+				identity: func() (string, error) {
+					return podmanWorkloadIdentity(harness, "root", "virt-launcher-"+lifecycleVMApp)
+				},
+			},
+		}
+
+		By("Updating OS to one with microshift")
+		err := harness.UpdateDeviceAndWaitForVersion(deviceID, func(device *v1beta1.Device) {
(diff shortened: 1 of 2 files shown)


itzikb-redhat:
Note: This test brings up a microshift device which is expensive, Wondering if it should be added to helm tests.

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// expectOthersRunning asserts every app except target reports Running.
+func expectOthersRunning(h *e2e.Harness, deviceID, targetName string, all []lifecycleApp) {
+	GinkgoHelper()
+	for _, app := range all {
+		if app.name == targetName {
+			continue
+		}
+		expectAppStatus(h, deviceID, app.name, v1beta1.ApplicationStatusRunning)
```
asafbennatan, line 435: `expectOthersAlive` already checks Running and probes the workload. After start and restart we only check status. If a sibling stayed Running in the API while its workload died, this would not catch it.

What if start/restart called `expectOthersAlive` as well, same as stop? Then this helper could go away.

```suggestion
		expectAppStatus(h, deviceID, app.name, v1beta1.ApplicationStatusRunning)
		app.alive()
```

asafbennatan (commented):
See comments below

Please make sure this passes e2e before merging.

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+
+	BeforeEach(func() {
+		harness = e2e.GetWorkerHarness()
+		deviceID, _ = harness.EnrollAndWaitForOnlineStatus()
+		services := auxiliary.Get(harness.GetTestContext())
+		helmChartRef = fmt.Sprintf("%s:%s/flightctl/%s", services.Registry.Host, services.Registry.Port, helmChartV1)
+	})
+
+	// App types: container, quadlet, compose, helm, and vm.
+	It("stops, starts, and restarts each app type without taking the others down", Label("slow", "vm", "90240"), func() {
```
asafbennatan, line 53: On whether this belongs in helm tests: this suite already brings up MicroShift, the aux registry, and `helmChartV1`, and the VM app needs that cluster too. Keeping the spec here is reasonable. The cost is that `GO_E2E_DIRS=test/e2e/applications/helm` now includes this matrix. A sibling `lifecycle` package with the same suite setup would isolate helm-only runs if that is the worry.

On CI: the spec is labeled `microshift`, `slow`, `vm`, `90240` and not `sanity`. PR e2e (including with `run-e2e`) still filters `sanity`, so the green e2e jobs on this PR did not execute it. If `90240` is Polarion-only, that is fine. If GitHub should run it, it would need `sanity` (and likely `agent`), knowing it is a long MicroShift+VM spec.

One other naming nit: EDM-4115 is the VM runtime QE story (console, QEMU kill, rawVmManifest). This spec covers application stop/start/restart across types, which is closer to EDM-4090 / EDM-4099. Worth aligning the commit/PR key if that was unintentional.

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/flightctl/flightctl/api/core/v1beta1"
+	"github.com/flightctl/flightctl/test/e2e/infra/auxiliary"
+	"github.com/flightctl/flightctl/test/harness/e2e"
+	"github.com/flightctl/flightctl/test/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	lifecycleContainerApp       = "nginx"
```
eldar101, line 19: I've seen most of these constants multiple times...we have to centralize them.

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+	lifecycleCurlConnectTimeout = "2"
+	lifecycleCurlMaxTime        = "5"
+)
+
+var _ = Describe("Application lifecycle stop start restart", Label("microshift"), func() {
+	var (
+		harness      *e2e.Harness
+		deviceID     string
+		helmChartRef string
+		microshiftOs = &v1beta1.DeviceOsSpec{Image: util.NewDeviceImageReference(util.DeviceTags.V12).String()}
```
eldar101, line 42: plz check if empty when used

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+// lifecycleVMImage is the guest image for the VM app, overridable via FLIGHTCTL_E2E_VM_IMAGE.
+func lifecycleVMImage() string {
+	if image := os.Getenv("FLIGHTCTL_E2E_VM_IMAGE"); image != "" {
+		return image
+	}
+	return lifecycleDefaultVMImage
+}
+
+// runLifecycleMatrix stop/start/restarts target and asserts the other apps stay running.
+func runLifecycleMatrix(h *e2e.Harness, deviceID string, target lifecycleApp, all []lifecycleApp) {
```
eldar101, line 206: This func has a test like structure consider changing it.

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+		matcher = HaveOccurred()
+		msg = "GET %s should fail"
+	}
+	Eventually(func() error {
+		return curlLocal(h, url)
+	}, util.TIMEOUT, util.POLLING).Should(matcher, msg, url)
+}
+
+// curlLocal GETs url from inside the device VM.
+func curlLocal(h *e2e.Harness, url string) error {
```
eldar101: consider moving harness funcs to harness folder

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+			continue
+		}
+		expectAppStatus(h, deviceID, app.name, v1beta1.ApplicationStatusRunning)
+	}
+}
+
+// appLifecycleCLI runs `flightctl app <action>` for appName on the device.
+func appLifecycleCLI(h *e2e.Harness, action, deviceID, appName string) {
+	GinkgoHelper()
+	_, err := h.CLI("app", action, fmt.Sprintf("device/%s", deviceID), "--name", appName, "-y")
```
eldar101: Since this is the helper that exercises the CLI, can we assert the returned output too? The command prints Requested <action>..., so checking stdout contains the action and app name would give us a positive CLI assertion, not just a zero exit.

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: I don't have one solution which I'm comfortable with. As it is I think it's too heavy to be in sanity. It could be split maybe but then it's not one test which logically it's an issue. For now I'm ok with it being as part of CI.

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: If it's empty UpdateDeviceAndWaitForVersion will fail,no?

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: It's intended as it's part of the test and not isolation function

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,410 @@
...
+	expectOthersRunning(h, deviceID, target.name, all)
+
+	By("Start " + target.name + " while already running")
+	appLifecycleCLI(h, "start", deviceID, target.name)
+	expectAppStatus(h, deviceID, target.name, v1beta1.ApplicationStatusRunning)
+
+	By("Restart " + target.name + " while running")
+	appLifecycleCLI(h, "restart", deviceID, target.name)
+	waitForAppStatus(h, deviceID, target.name, v1beta1.ApplicationStatusRunning)
+	target.up()
```
itzikb-redhat, line 245: Addressed by checking the workload

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: Done

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: ack

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: Move one function. Let me know if you have others in mind.

test/e2e/applications/helm/lifecycle_test.go
itzikb-redhat: In this test suite? Where do you want to put it?

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,519 @@
...
+	By("Restart " + target.name + " while running")
+	before, err := target.identity()
+	Expect(err).NotTo(HaveOccurred(), "capturing workload identity for %s before restart", target.name)
+	Expect(before).NotTo(BeEmpty(), "workload identity for %s should be present before restart", target.name)
+	appLifecycleCLI(h, "restart", deviceID, target.name)
+	waitForNewWorkloadIdentity(target, before)
+	waitForAppStatus(h, deviceID, target.name, v1beta1.ApplicationStatusRunning)
+	target.up()
+	expectOthersAlive(h, deviceID, target.name, all)
+}
```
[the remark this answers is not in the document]
itzikb-redhat: After each removes the device so it should be ok

test/e2e/applications/helm/lifecycle_test.go
```diff
@@ -0,0 +1,516 @@
...
+// waitForStop waits until the app is Stopped and not Error or Completed.
+func waitForStop(h *e2e.Harness, deviceID, appName string) {
+	GinkgoHelper()
+	waitForAppStatus(h, deviceID, appName, v1beta1.ApplicationStatusStopped)
+	status, err := currentAppStatus(h, deviceID, appName)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(status).To(Equal(v1beta1.ApplicationStatusStopped), "application %s", appName)
+	Expect(status).NotTo(Equal(v1beta1.ApplicationStatusError), "application %s", appName)
+	Expect(status).NotTo(Equal(v1beta1.ApplicationStatusCompleted), "application %s", appName)
+}
```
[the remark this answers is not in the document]
itzikb-redhat: This is fixed

asafbennatan (approved):
LGTM

PR #3434, opened 2026-08-20
EDM-5248: Define Scanner interface and extend store query

EfratIfergan:
## EDM-5248: Define Scanner interface and extend store query

**Jira:** https://redhat.atlassian.net/browse/EDM-5248
**Story type:** [DEV]
**Epic:** EDM-5246 — Backend Abstraction and Trustify Adapter (Feature EDM-4869)

> ⚠️ **Base of a stack.** This is the first PR in a stack. EDM-5249 (Trustify
> Scanner adapter and sync task refactor) is stacked on top and targets this
> branch. Merge this PR first, then EDM-5249 will retarget `main`.

### Summary
Introduces the backend-agnostic `vulnerability.Scanner` abstraction (interface,
DTOs, and a backend registry/factory) that lets Flight Control support multiple
vulnerability backends behind one contract, and extends the store query so the
sync pipeline can key on image references. No behavioral change to the existing
Trustify sync — this lays the groundwork for the Trustify adapter (EDM-5249) and
the Quay backend (EDM-5250).

### Changes
- **`internal/vulnerability`** — new package: `Scanner` interface,
  `Finding`/`Issuer`/`ImageRef` DTOs, and a `Register`/`NewScanner` backend
  registry + factory.
- **Config** — add vulnerability backend selection (`vulnerabilityReporting.backend`)
  with a Trustify default when a Trustify config is present.
- **Store** — `ListDeployedImageDigests` returns image references (digest + ref),
  with a stable tie-break that prefers a non-empty ref over an empty one for the
  same digest.
- **Sync task** — adapt to the `ImageRef`-based store query.

### Testing
- **Unit tests:** registry/factory resolution, config backend parsing, DTO
  handling; sync-task tests updated for the `ImageRef` return type.
- **Integration tests:** `ListDeployedImageDigests` `ImageRef` coverage,
  including the empty-vs-non-empty ref tie-break (320/320 store integration
  specs pass).
- **Coverage:** comprehensive behavioral coverage of the new public surface.

### Acceptance Criteria
- [ ] AC-1: Backend-agnostic `Scanner` interface + DTOs defined in `internal/vulnerability`.
- [ ] AC-2: Backend registry (`Register`/`NewScanner`) with factory resolution.
- [ ] AC-3: Vulnerability backend selection added to config.
- [ ] AC-4: `ListDeployedImageDigests` returns image references.
- [ ] AC-5: Non-empty image ref preferred in the tie-break for a shared digest.
- [ ] AC-6: Sync task adapted to the `ImageRef`-based query.
- [ ] AC-7: Existing tests pass; no behavioral change to Trustify sync.

---

### DIFF
--- a/deploy/helm/flightctl/values.schema.json
+++ b/deploy/helm/flightctl/values.schema.json
@@ -975,7 +975,7 @@
         "syncInterval": { "type": "string", "description": "Sync interval for periodic Trustify fetch (e.g. \"15m\", \"1h\")" },
         "backend": {
           "type": "string",
-          "description": "Vulnerability scanning backend. Currently only 'trustify' is supported; leave empty to default to Trustify when a trustify config is present.",
+          "description": "Vulnerability scanning backend. Currently only 'trustify' is supported; leave empty to default to Trustify when a trustify config with a non-empty endpoint is present.",
           "enum": ["", "trustify"]
         },
         "trustify": {
--- a/deploy/helm/flightctl/values.schema.json
+++ b/deploy/helm/flightctl/values.schema.json
@@ -973,6 +973,11 @@
       "properties": {
         "enabled": { "type": "boolean", "description": "Enable vulnerability integration (sync task + API endpoints)" },
         "syncInterval": { "type": "string", "description": "Sync interval for periodic Trustify fetch (e.g. \"15m\", \"1h\")" },
+        "backend": {
+          "type": "string",
+          "description": "Vulnerability scanning backend. Currently only 'trustify' is supported; leave empty to default to Trustify when a trustify config is present.",
+          "enum": ["", "trustify"]
+        },
         "trustify": {
           "type": "object",
           "description": "Trustify API connection details",
--- a/docs/user/installing/configuring-vulnerability-integration.md
+++ b/docs/user/installing/configuring-vulnerability-integration.md
@@ -61,6 +61,7 @@ The following table describes all configuration fields under the `vulnerabilityR
 |---|---|---|---|
 | `enabled` | boolean | `false` | Enables vulnerability integration. Set to `true` to activate sync and API endpoints. |
 | `syncInterval` | duration | `15m` | Interval between Trustify sync runs. Use Go duration syntax, for example `15m`, `1h`. |
+| `backend` | string | — | Vulnerability scanning backend. Currently only `trustify` is supported. When empty, defaults to Trustify if a `trustify` config is present. |
 | `trustify.endpoint` | string | — | Trustify API base URL. Required when `enabled` is `true`. |
 | `trustify.auth.mode` | string | `none` | Authentication mode. Allowed values: `none`, `client-credentials`. |
 | `trustify.auth.oidcIssuerUrl` | string | — | OIDC issuer URL. Required when `mode` is `client-credentials`. |
@@ -152,6 +153,7 @@ credentials as environment variables.
 |---|---|
 | `FLIGHTCTL_VULNERABILITY_REPORTING_ENABLED` | Enables or disables vulnerability integration. Accepted values: `true`, `1`, `false`, `0`. |
 | `FLIGHTCTL_VULNERABILITY_REPORTING_SYNC_INTERVAL` | Sync interval. Use Go duration syntax, for example `15m`. |
+| `FLIGHTCTL_VULNERABILITY_REPORTING_BACKEND` | Vulnerability scanning backend. Currently only `trustify` is supported. |
 | `FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_ENDPOINT` | Trustify API base URL. |
 | `FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_AUTH_MODE` | Authentication mode (`none` or `client-credentials`). |
 | `FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_OIDC_ISSUER_URL` | OIDC issuer URL for client-credentials mode. |
--- a/docs/user/installing/configuring-vulnerability-integration.md
+++ b/docs/user/installing/configuring-vulnerability-integration.md
@@ -61,7 +61,7 @@ The following table describes all configuration fields under the `vulnerabilityR
 |---|---|---|---|
 | `enabled` | boolean | `false` | Enables vulnerability integration. Set to `true` to activate sync and API endpoints. |
 | `syncInterval` | duration | `15m` | Interval between Trustify sync runs. Use Go duration syntax, for example `15m`, `1h`. |
-| `backend` | string | — | Vulnerability scanning backend. Currently only `trustify` is supported. When empty, defaults to Trustify if a `trustify` config is present. |
+| `backend` | string | — | Vulnerability scanning backend. Currently only `trustify` is supported. When empty, defaults to Trustify if a `trustify` config with a non-empty `endpoint` is present. |
 | `trustify.endpoint` | string | — | Trustify API base URL. Required when `enabled` is `true`. |
 | `trustify.auth.mode` | string | `none` | Authentication mode. Allowed values: `none`, `client-credentials`. |
 | `trustify.auth.oidcIssuerUrl` | string | — | OIDC issuer URL. Required when `mode` is `client-credentials`. |
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -776,12 +776,23 @@ func (c *Config) GetDependenciesSyncPollInterval() time.Duration {
 	return DefaultDependenciesSyncPollInterval
 }
 
+// VulnerabilityBackend selects which vulnerability scanning backend is active.
+type VulnerabilityBackend string
+
+const (
+	// VulnerabilityBackendTrustify selects the Trustify (TPA) backend.
+	VulnerabilityBackendTrustify VulnerabilityBackend = "trustify"
+)
+
 // VulnerabilityConfig holds configuration for the vulnerability integration feature.
 type VulnerabilityConfig struct {
 	// Enabled enables vulnerability integration (sync task + API endpoints).
 	Enabled bool `json:"enabled,omitempty"`
-	// SyncInterval is the interval between Trustify sync runs (e.g. "15m", "1h").
+	// SyncInterval is the interval between backend sync runs (e.g. "15m", "1h").
 	SyncInterval util.Duration `json:"syncInterval,omitempty"`
+	// Backend selects the vulnerability scanning backend. When empty, the sync
+	// task defaults to Trustify if a Trustify config is present.
+	Backend VulnerabilityBackend `json:"backend,omitempty"`
 	// Trustify holds the Trustify connection details (periodic service only).
 	Trustify *TrustifyConfig `json:"trustify,omitempty"`
 }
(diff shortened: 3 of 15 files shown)


internal/vulnerability/registry.go
```diff
@@ -0,0 +1,48 @@
...
+// NewScanner resolves the backend selected by cfg.Backend and constructs its
+// Scanner. When cfg.Backend is empty it defaults to Trustify if a Trustify
+// config is present (logging a deprecation warning), otherwise it errors.
+func NewScanner(cfg *config.VulnerabilityConfig) (Scanner, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("vulnerability config is nil")
+	}
+	backend := cfg.Backend
+	if backend == "" {
+		if cfg.Trustify != nil {
```
galel12: `cfg.Trustify != nil` is true for an empty `trustify: {}` block (or a Trustify struct created by env overrides before endpoint is set), so this triggers the trustify default even when there is no usable Trustify config.

`periodic_checker` only starts sync when `Trustify != nil` and builds the client from endpoint — same loose check today, but once EDM-5249 wires `NewScanner`, should the default be tied to something stricter, e.g. non-empty `Trustify.Endpoint`?

```suggestion
		if cfg.Trustify != nil && strings.TrimSpace(cfg.Trustify.Endpoint) != "" {
```

internal/vulnerability/registry.go
```diff
@@ -0,0 +1,48 @@
...
+			logrus.Warn(`vulnerability.backend not set; defaulting to "trustify" because trustify config is present. Set backend explicitly — this fallback will be removed in a future release.`)
```
galel12: This warns on every `NewScanner` call. If the periodic task constructs a scanner each poll (likely in EDM-5249) and `backend` stays empty, logs will spam.

Should we use `sync.Once` here, or resolve/log the default once at periodic startup?

internal/vulnerability/registry.go
```diff
@@ -0,0 +1,48 @@
...
+
+// ScannerFactory constructs a Scanner from the vulnerability configuration.
+type ScannerFactory func(cfg *config.VulnerabilityConfig) (Scanner, error)
+
+var registry = map[string]ScannerFactory{}
+
+// Register associates a backend name with a factory. Backends call this from an
+// init function so they self-register when their package is imported.
+func Register(name string, factory ScannerFactory) {
+	registry[name] = factory
```
galel12, line 32: `Register` silently overwrites an existing backend name. A duplicate `init()` registration would be hard to spot.

Worth a `logrus.Fatalf` / panic in tests, or at least a debug log when `registry[name]` is already set?

internal/config/config.go
```diff
@@ -1206,6 +1218,10 @@ func applyVulnerabilityReportingEnvVarOverrides(c *Config) {
 		c.VulnerabilityReporting = &VulnerabilityConfig{}
 	}
 
+	if backend != "" {
```
galel12, line 1222: `FLIGHTCTL_VULNERABILITY_REPORTING_BACKEND` is accepted with no validation. Invalid values (typo, wrong case) are stored and only fail later in `NewScanner`.

`ENABLED` and `SYNC_INTERVAL` warn on bad input — should `backend` do the same (or validate in `Validate()` against known backends)?

internal/store/vulnerabilityfinding/store.go
```diff
@@ -266,17 +267,27 @@ func (s *VulnerabilityFindingStore) createDeviceOpenCveEventIndexes(db *gorm.DB)
...
+// ListDeployedImageDigests returns one entry per distinct OS image digest
+// reported across all devices (all organizations), pairing each digest with a
+// deployed image reference. Digests that are NULL or empty are excluded; the
+// image reference may be empty when a device reports a digest but no image.
+// When several devices report the same digest with different references, the
+// smallest non-empty reference is returned (or an empty reference when none has
+// one) so the result is stable. NULLIF collapses empty images to NULL, which
+// sorts last (ASC = NULLS LAST), so a real reference always wins over an
+// empty/NULL one; the outer COALESCE then projects any remaining NULL back to "".
+func (s *VulnerabilityFindingStore) ListDeployedImageDigests(ctx context.Context) ([]vulnerability.ImageRef, error) {
```
galel12: Return type is now `[]vulnerability.ImageRef`, but the method is still `ListDeployedImageDigests`. Callers already need updating for this PR.

Worth renaming to `ListDeployedImageRefs` while the signature is changing, so the name matches what it returns?

internal/tasks/vulnerability_sync.go
```diff
@@ -103,19 +104,19 @@ func (t *VulnerabilitySync) Poll(ctx context.Context) {
...
+	if len(imageRefs) == 0 {
 		t.log.Info("No deployed image digests found, nothing to sync")
 		return nil, nil
 	}
 
-	t.log.Infof("Syncing vulnerabilities for %d unique image digests", len(digests))
+	t.log.Infof("Syncing vulnerabilities for %d unique image digests", len(imageRefs))
 
-	findingsMap, err := t.vulnClient.GetVulnerabilitiesForDigests(ctx, digests)
+	findingsMap, err := t.vulnClient.GetVulnerabilitiesForDigests(ctx, digestsFromImageRefs(imageRefs))
```
galel12, line 120: `imageRefs` are fetched then immediately stripped to digests for Trustify. Fine for no-behavior-change in this PR.

When EDM-5249 wires `Scanner.ScanImages`, this path should pass the full `[]ImageRef` through instead of `digestsFromImageRefs` — otherwise the store change does not reach Quay. Worth a short comment here so 5249 does not miss it?

```suggestion
	// TODO(EDM-5249): pass imageRefs to Scanner.ScanImages; Trustify ignores Image today.
	findingsMap, err := t.vulnClient.GetVulnerabilitiesForDigests(ctx, digestsFromImageRefs(imageRefs))
```

internal/vulnerability/registry.go
EfratIfergan: Good catch — adopted. The default now requires a usable endpoint: `cfg.Trustify != nil && strings.TrimSpace(cfg.Trustify.Endpoint) != ""`, so an empty `trustify: {}` block (or an env-created Trustify struct without an endpoint) no longer silently selects the backend. This matches what `periodic_checker` actually needs to build a client. Fixed in 555f722e.

internal/vulnerability/registry.go
EfratIfergan: Agreed — wrapped the deprecation warning in a package-level `sync.Once` so it fires at most once even when `NewScanner` is called per poll in EDM-5249. Fixed in 555f722e.

internal/vulnerability/registry.go
EfratIfergan: Agreed — `Register` now panics on a duplicate name (`vulnerability: backend %q already registered`) instead of silently overwriting. Since registration happens from `init()`, a duplicate is a programming error best surfaced loudly at startup (same spirit as `sql.Register`). Added a `TestRegisterDuplicatePanics` case. Fixed in 555f722e.

internal/config/config.go
EfratIfergan: I looked at this but decided to leave the env override as-is, for a couple of reasons:

- The authoritative "is this a known backend?" check lives in the registry, and `NewScanner` already fails with a clear message that lists the known backends (`unknown vulnerability backend %q (known: %v)`). That's the single source of truth for valid backends.
- The `config` package is imported *by* `internal/vulnerability`, so it can't consult the registry without creating an import cycle. Validating here would mean either duplicating the backend list in `config` (which drifts as backends are added) or hard-coding only `trustify` in the wrong layer.
- Unlike `ENABLED`/`SYNC_INTERVAL`, which parse into a specific type/format and can be validated locally, `backend` is a free-form key that's resolved elsewhere.

So I'd rather keep the validation in one place (`NewScanner`) than split it. Happy to revisit if we later want fail-fast config validation — that'd be a small `Validate()` hook that calls into the registry, once EDM-5249 wires a backend in.

internal/store/vulnerabilityfinding/store.go
EfratIfergan: Agreed — renamed to `ListDeployedImageRefs` across the interface, the store method, the sync task, and the unit/integration tests, so the name matches the `[]vulnerability.ImageRef` it returns. Fixed in 555f722e.

internal/tasks/vulnerability_sync.go
EfratIfergan: Added the marker verbatim right above the call so EDM-5249 doesn't miss it:

```go
// TODO(EDM-5249): pass imageRefs to Scanner.ScanImages; Trustify ignores Image today.
findingsMap, err := t.vulnClient.GetVulnerabilitiesForDigests(ctx, digestsFromImageRefs(imageRefs))
```

Fixed in 555f722e.

internal/vulnerability/registry.go
```diff
@@ -0,0 +1,63 @@
...
+func Register(name string, factory ScannerFactory) {
+	if _, exists := registry[name]; exists {
+		panic(fmt.Sprintf("vulnerability: backend %q already registered", name))
+	}
+	registry[name] = factory
+}
+
+// NewScanner resolves the backend selected by cfg.Backend and constructs its
+// Scanner. When cfg.Backend is empty it defaults to Trustify if a Trustify
+// config is present (logging a deprecation warning), otherwise it errors.
```
galel12: Doc drift from 555f722e: the empty-backend default now requires a non-empty `Trustify.Endpoint`, not merely a present `trustify` block. Same for the deprecation log text on line 47 (`"because trustify config is present"`).

```suggestion
// config with a non-empty Trustify endpoint is present (logging a deprecation warning), otherwise it errors.
```

internal/config/config.go
```diff
@@ -776,12 +776,23 @@ func (c *Config) GetDependenciesSyncPollInterval() time.Duration {
...
+
 // VulnerabilityConfig holds configuration for the vulnerability integration feature.
 type VulnerabilityConfig struct {
 	// Enabled enables vulnerability integration (sync task + API endpoints).
 	Enabled bool `json:"enabled,omitempty"`
-	// SyncInterval is the interval between Trustify sync runs (e.g. "15m", "1h").
+	// SyncInterval is the interval between backend sync runs (e.g. "15m", "1h").
 	SyncInterval util.Duration `json:"syncInterval,omitempty"`
+	// Backend selects the vulnerability scanning backend. When empty, the sync
+	// task defaults to Trustify if a Trustify config is present.
```
galel12: Same doc drift: default now needs a non-empty `trustify.endpoint`, not just a `trustify:` stanza. The user docs / Helm `values.yaml` comments still say "when a trustify config is present" — worth aligning in the same pass.

Also: `NewScanner` now enforces the endpoint gate, but `periodic_checker` still enables sync on `Trustify != nil` (so an empty `trustify: {}` block still starts the executor today). Should periodic adopt the same endpoint check now, or is that intentionally deferred to EDM-5249?

docs/user/installing/configuring-vulnerability-integration.md
```diff
@@ -61,6 +61,7 @@ The following table describes all configuration fields under the `vulnerabilityR
 |---|---|---|---|
 | `enabled` | boolean | `false` | Enables vulnerability integration. Set to `true` to activate sync and API endpoints. |
 | `syncInterval` | duration | `15m` | Interval between Trustify sync runs. Use Go duration syntax, for example `15m`, `1h`. |
+| `backend` | string | — | Vulnerability scanning backend. Currently only `trustify` is supported. When empty, defaults to Trustify if a `trustify` config is present. |
```
galel12: This still says "defaults to Trustify if a `trustify` config is present", but `NewScanner` now requires a non-empty `trustify.endpoint` for the empty-backend default. Should match the runtime behavior after 555f722e.

internal/vulnerability/registry.go
EfratIfergan: Agreed — fixed in 448f9812. Updated `NewScanner`'s doc comment to say the empty-backend default requires a Trustify config *with a non-empty endpoint*, and corrected the deprecation log text (now "because a non-empty trustify endpoint is configured"). The gate itself (`TrimSpace(cfg.Trustify.Endpoint) != ""`) was already in place from 555f722e; this only aligns the docs.

internal/config/config.go
EfratIfergan: Agreed on the doc drift — fixed in 448f9812. Updated the `Backend` field comment, the user docs, and the Helm `values.yaml`/`.gotmpl`/README to all say "a trustify config with a non-empty endpoint is present."

On the periodic question: the endpoint gate is deliberately **not** adopted in `periodic_checker` here. This PR is intentionally behavior-neutral, and `periodic_checker` still builds the Trustify client directly (gating on `Trustify != nil`) rather than via `NewScanner`. Rewiring it to `vulnerability.NewScanner` — which brings the endpoint gate with it — is done in the stacked **EDM-5249** (`eb36890f`). Doing it here would change runtime behavior (an empty `trustify: {}` block would stop starting the executor), so it belongs with that rewire.

docs/user/installing/configuring-vulnerability-integration.md
EfratIfergan: Agreed — fixed in 448f9812. Updated the config-reference table to say the empty-backend default applies when a `trustify` config with a non-empty `endpoint` is present, matching runtime behavior after 555f722e.

EfratIfergan:
**Re: the Api-Backward-Compatibility pre-merge check** (`ListDeployedImageDigests` → `ListDeployedImageRefs`):

This method is in `internal/store/vulnerabilityfinding` — an `internal/` package. Go forbids importing `internal/...` from outside the module, so there are no external consumers; the only caller is the vulnerability sync task in this same PR, updated in lockstep. The rename was itself requested in review to match the new `[]vulnerability.ImageRef` return type, and there's no cross-release API-stability surface here.

Adding a deprecated `[]string` wrapper would introduce a method with no in-repo caller purely to satisfy the heuristic. Could a maintainer confirm this check can be overridden for an internal-only rename? Happy to add the shim if the team would rather keep it green.

deploy/helm/flightctl/values.schema.json
```diff
@@ -973,6 +973,11 @@
       "properties": {
         "enabled": { "type": "boolean", "description": "Enable vulnerability integration (sync task + API endpoints)" },
         "syncInterval": { "type": "string", "description": "Sync interval for periodic Trustify fetch (e.g. \"15m\", \"1h\")" },
+        "backend": {
+          "type": "string",
+          "description": "Vulnerability scanning backend. Currently only 'trustify' is supported; leave empty to default to Trustify when a trustify config is present.",
```
galel12: `448f9812` updated the `backend` wording in `values.yaml`, `.gotmpl`, README, user docs, and the Go struct comment, but this schema description still says "when a trustify config is present". Should match the endpoint-gated default:

```suggestion
          "description": "Vulnerability scanning backend. Currently only 'trustify' is supported; leave empty to default to Trustify when a trustify config with a non-empty endpoint is present.",
```

galel12 (approved):
LGTM

PR #3435, opened 2026-08-20
NO-ISSUE: preload package-mode helper image for e2e

eldar101:
## Summary
This makes the package-mode e2e helper container use the same image runtime that the test run already prepared.

Without this, the package-mode suite can still try to start the helper from `quay.io/flightctl/flightctl-device:package` even when the image is only available through the local test bundle/runtime, which leads to `unauthorized` pull failures in CI.

The change does two things:
- preloads the package-mode helper image in `test/scripts/run_e2e_tests.sh` when the `package_mode` suite is running
- lets the helper container honor an `E2E_PACKAGE_MODE_IMAGE` override instead of always hardcoding the default ref

Release target
release-1.3
Target release: release-1.3

### DIFF
--- a/test/scripts/run_e2e_tests.sh
+++ b/test/scripts/run_e2e_tests.sh
@@ -6,6 +6,7 @@ export PATH="/usr/local/bin:${PATH}"
 
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 source "${SCRIPT_DIR}"/functions
+source "${SCRIPT_DIR}"/detect_container_runtime.sh
 
 REPORTS=${1}
 GO_E2E_DIRS=("${@:2}")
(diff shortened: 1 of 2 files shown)


test/scripts/run_e2e_tests.sh
```diff
@@ -39,6 +40,101 @@ if [ -z "$GOBIN" ]; then
     GOBIN=$(go env GOPATH)/bin
 fi
 
+should_preload_package_mode_image() {
+    local suite_dir
+    for suite_dir in "${GO_E2E_DIRS[@]}"; do
+        if [[ "${suite_dir}" == *"/package_mode" || "${suite_dir}" == *"/package_mode/"* ]]; then
```
galel12: `should_preload_package_mode_image` only matches paths that contain `/package_mode`. The default `GO_E2E_DIRS` value — and what CI passes to `make -e run-e2e-test` — is `./test/e2e/...`, so this returns false and `preload_package_mode_image` no-ops.

Ginkgo still expands that glob and runs `test/e2e/package_mode`. That is also the documented local full-suite path (`make e2e-test` / `make run-e2e-test`). The workflow “Stage test artifacts” step already tries to load the image, but this script is what local runs use, and the PR is aiming at the unauthorized pull when the helper still uses `quay.io/flightctl/flightctl-device:package`.

Should we also treat the e2e glob as including package-mode?

```suggestion
        if [[ "${suite_dir}" == *"/package_mode" || "${suite_dir}" == *"/package_mode/"* || "${suite_dir}" == *"/e2e/..." ]]; then
```

test/scripts/run_e2e_tests.sh
```diff
@@ -39,6 +40,101 @@ if [ -z "$GOBIN" ]; then
...
+            return 0
+        fi
+    done
+    return 1
+}
+
+find_package_mode_bundle() {
+    local preferred_bundle fallback_bundle bundle_dir bundle_path
+    bundle_dir="$(readlink -f bin/agent-artifacts)"
+
```
galel12, line 62: `find_package_mode_bundle` prefers `agent-images-bundle-${AGENT_OS_ID}.tar`, but `_e2e_test` and the CI “Run E2E Test Slice” step never export `AGENT_OS_ID`. That lookup only runs if the caller already has it in the environment.

Otherwise the fallback takes the lexicographically first tar (`cs10-...` before `cs9-...`). `:package` is a single global tag, so a mixed `bin/agent-artifacts/` tree can load the wrong OS’s image.

Should we pass `AGENT_OS_ID` into `run_e2e_tests.sh` from Make/CI, and fail if that bundle is missing instead of picking another OS?

test/scripts/run_e2e_tests.sh
```diff
@@ -39,6 +40,122 @@ if [ -z "$GOBIN" ]; then
...
+            return 0
+        fi
+    done
+    return 1
+}
+
+find_package_mode_bundle() {
+    local preferred_bundle bundle_path
+    local -a bundles
+    bundle_dir="$(readlink -f bin/agent-artifacts)"
```
galel12, line 61: Should we declare `bundle_dir` as `local` here alongside the other function-scoped variables? It currently leaks into the shell scope. Non-blocking — fine to leave as-is if you prefer.

galel12 (approved):
The glob and bundle-selection fixes address the earlier review points. Approving as-is; the `bundle_dir` local is optional cleanup.

galel12 (approved):

PR #3436, opened 2026-08-20
EDM-3772: omit zero rollout minAvailable

eldar101:
## Summary
Fix the rollout disruption budget E2E so it no longer sends `minAvailable: 0`, which current API schema rejects because `minAvailable` must be at least 1 when present.

The grouped scenario only needs `maxUnavailable: 2`, so omitting `minAvailable` preserves the intended test coverage without creating an invalid fleet request.

## Release target
release-1.3
Target release: release-1.3


eldar101:
Consolidated into #3418 as requested.

PR #3437, opened 2026-08-20
EDM-5249: Trustify Scanner adapter and sync task refactor

EfratIfergan:
## EDM-5249: Trustify Scanner adapter and sync task refactor

**Jira:** https://redhat.atlassian.net/browse/EDM-5249
**Story type:** [DEV]
**Epic:** EDM-5246 — Backend Abstraction and Trustify Adapter (Feature EDM-4869)

> 🔗 **Stacked on [#3434](https://github.com/flightctl/flightctl/pull/3434) (EDM-5248).**
> This PR targets the `EDM-5248-scanner-interface` branch, so its diff shows only
> this story's changes. Review/merge #3434 first; GitHub will then retarget this
> PR to `main`.

### Summary
Behavior-preserving refactor that moves the vulnerability sync pipeline onto the
backend-agnostic `vulnerability.Scanner` abstraction (from EDM-5248). Adds a
Trustify `Scanner` adapter that wraps the existing Trustify client, and rewires
the sync task and periodic checker to depend on `vulnerability.Scanner` instead
of the concrete `trustifyv2.VulnerabilityClient`. Same findings, same model rows,
same CVE lifecycle events. (The `Source` field is intentionally deferred to
EDM-5250.)

### Changes
- **`internal/trustify/v2/scanner_adapter.go`** (new) — `trustifyScanner`
  implements `vulnerability.Scanner`; converts `trustifyv2.Finding` →
  `vulnerability.Finding`; owns the moved Trustify-specific
  `normalizeTrustifyStatus`/`normalizeTrustifySeverity`; self-registers via
  `vulnerability.Register("trustify", …)` in `init()`. The underlying client is
  built lazily and cached on first `ScanImages` (the scanner factory has no
  context, but client construction needs one).
- **`internal/tasks/vulnerability_sync.go`** — `VulnerabilitySync` now depends on
  `vulnerability.Scanner`; `syncFindings` calls `ScanImages`; `toModelFinding`
  converts the backend-agnostic DTO to the store model (no `Source`). Removed the
  Trustify-specific normalization and digest-extraction helpers (moved to the
  adapter).
- **`internal/periodic_checker`** — builds the scanner via
  `vulnerability.NewScanner(cfg)` and blank-imports the adapter for
  self-registration; `VulnerabilitySyncExecutor`/`InitializeTaskExecutors` take a
  `vulnerability.Scanner`.
- **`test/integration/store/vulnerability_finding_test.go`** — the end-to-end sync
  specs build the scanner via `trustifyv2.NewScanner` (which lazily creates the
  same real client against the mock Trustify server).

`UploadSBOM` stays on the Trustify client (used only by the imagebuilder worker)
and is deliberately **not** part of the `Scanner` interface.

### Testing
- **Unit tests:** new adapter suite (digest-only delegation, field-by-field
  conversion parity, all normalization branches, skip-on-unnormalizable,
  nil-vs-empty SBOM handling, lazy construction); sync-task suite re-pointed at a
  `fakeScanner` with identical lifecycle assertions.
- **Integration tests:** store end-to-end sync specs pass against real Postgres +
  a mock Trustify server (31/31 in the focused run).
- **Coverage:** conversion/normalization/delegation 95–100%; the external client
  build is covered by the integration path.

### Acceptance Criteria
- [ ] AC-1: Trustify adapter wraps `GetVulnerabilitiesForDigests` and converts `trustifyv2.Finding` → `vulnerability.Finding`.
- [ ] AC-2: `normalizeTrustifyStatus`/`normalizeTrustifySeverity` moved into the adapter.
- [ ] AC-3: `toModelFinding` refactored to `vulnerability.Finding` → model, without a `Source` field.
- [ ] AC-4: `VulnerabilitySync` takes a `vulnerability.Scanner`.
- [ ] AC-5: Adapter self-registers via `vulnerability.Register("trustify", …)` in `init()`.
- [ ] AC-6: Periodic checker instantiates the scanner via `vulnerability.NewScanner(cfg)`.
- [ ] AC-7: All existing tests pass unchanged (fakes implement `Scanner`).
- [ ] AC-8: `UploadSBOM` stays on the Trustify client, not on `Scanner`.

---


internal/trustify/v2/scanner_adapter.go
```diff
@@ -0,0 +1,161 @@
...
+// scanner factory has no request context, whereas client construction (OIDC
+// discovery in client-credentials mode) needs one.
+func NewScanner(cfg *config.TrustifyConfig) (vulnerability.Scanner, error) {
+	if cfg == nil {
+		return nil, nil
+	}
+	return &trustifyScanner{cfg: cfg}, nil
+}
+
+func (s *trustifyScanner) ensureClient(ctx context.Context) (VulnerabilityClient, error) {
```
galel12, line 48: Lazy client init is new in this PR. Before, `server.go` built the client eagerly at startup; concurrent `ScanImages` calls shared a ready-made client with no init race.

`VulnerabilitySync` is `SystemWide: true` with a 15-minute interval, and the publisher reschedules immediately after enqueue — not after completion. If a scan outlasts the interval, a second task can land on another consumer while the first is still running. All consumers share one `VulnerabilitySyncExecutor` and one `scanner` instance, so overlapping `ScanImages` calls can race here (duplicate clients, `-race` failure).

Should we guard this with `sync.Once` (or a mutex around the check-and-create)? `sync.Once` fits the lazy-init pattern well:

```go
type trustifyScanner struct {
	cfg     *config.TrustifyConfig
	client  VulnerabilityClient
	once    sync.Once
	initErr error
}

func (s *trustifyScanner) ensureClient(ctx context.Context) (VulnerabilityClient, error) {
	s.once.Do(func() {
		s.client, s.initErr = NewVulnerabilityClient(ctx, s.cfg)
	})
	return s.client, s.initErr
}
```

A `-race` test with concurrent `ScanImages` would lock this in.

internal/trustify/v2/scanner_adapter.go
```diff
@@ -0,0 +1,161 @@
...
+	for digest, findings := range raw {
+		if findings == nil {
+			out[digest] = nil
+			continue
+		}
+		converted := make([]vulnerability.Finding, 0, len(findings))
+		for _, f := range findings {
+			vf, err := toVulnFinding(f)
+			if err != nil {
+				logrus.WithError(err).Debugf("Skipping finding for digest %s cve %s", f.ImageDigest, f.CVEID)
```
galel12, line 98: Previously, `vulnerability_sync.go` aggregated skipped normalization failures and logged a **warn**:

```go
if skippedCount > 0 {
    t.log.Warnf("Skipped %d findings due to normalization errors", skippedCount)
}
```

The adapter now logs per-finding at **debug** via the global `logrus` logger, with no aggregate warn. That makes it harder to spot Trustify data-quality issues in production (warn-level monitoring won't catch it).

Should we either restore an aggregate warn in the adapter (after the conversion loop), or pass a `logrus.FieldLogger` into the scanner so skips use the sync task's logger with a summary count?

galel12 (commented):
**`internal/tasks/vulnerability_sync.go` — dead `digestsFromImageRefs` helper:** After this change, digest extraction lives inline in the adapter's `ScanImages`. `digestsFromImageRefs` remains in `vulnerability_sync.go` but is only referenced from `TestDigestsFromImageRefs` — the adapter already covers this in `TestTrustifyScanner_ScanImages_PassesDigestsOnly`. Should we drop the helper and its sync-task test to avoid maintaining duplicate logic in the wrong package?

**Concurrent sync overlap (pre-existing):** Even with a fixed `ensureClient`, overlapping `VulnerabilitySync.Poll` runs could still duplicate DB work and checkpoint/event processing. That's the reschedule-before-complete + system-wide task pattern, not introduced here — but the lazy-init race is new and worth fixing in this PR.

**`toModelFinding` trust boundary:** The sync task now casts `Status`/`Severity` without validation, trusting the scanner. That's the right abstraction boundary; future backends must normalize before returning DTOs (as the Trustify adapter does).

internal/trustify/v2/scanner_adapter.go
EfratIfergan: Good catch — fixed in 17a48dec. `ensureClient` now guards construction with `sync.Once` (caching `initErr`), so overlapping `SystemWide` `ScanImages` runs can't race to build duplicate clients. Added `TestTrustifyScanner_ensureClient_ConcurrentInitBuildsOnce`, which fans out 16 concurrent scans and asserts the client is constructed exactly once; it passes under `-race`.

internal/trustify/v2/scanner_adapter.go
```diff
@@ -0,0 +1,161 @@
...
+func (s *trustifyScanner) ensureClient(ctx context.Context) (VulnerabilityClient, error) {
+	if s.client != nil {
+		return s.client, nil
+	}
+	client, err := NewVulnerabilityClient(ctx, s.cfg)
+	if err != nil {
+		return nil, err
+	}
+	s.client = client
+	return client, nil
```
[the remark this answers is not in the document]
EfratIfergan: Good catch — fixed in 17a48dec. `ensureClient` now guards construction with `sync.Once` (caching `initErr`), so overlapping `SystemWide` `ScanImages` runs can't race to build duplicate clients. Added `TestTrustifyScanner_ensureClient_ConcurrentInitBuildsOnce`, which fans out 16 concurrent scans and asserts the client is constructed exactly once; it passes under `-race`.

internal/trustify/v2/scanner_adapter.go
EfratIfergan: Restored the warn-level visibility in 17a48dec: the per-finding `Debug` line stays, and `ScanImages` now emits an aggregate `Warnf("Skipped %d Trustify findings due to normalization errors", skipped)` after the conversion loop, so data-quality issues surface without debug logging. I kept the global `logrus` logger rather than threading a `FieldLogger` through, since that would require changing the `ScannerFactory` signature across the registry — happy to do the logger-injection version as a follow-up if you'd prefer the sync task's contextual logger.

EfratIfergan:
Thanks for the thorough review. Addressed in 17a48dec:

- **`ensureClient` race:** fixed with `sync.Once` (caching `initErr`) plus a concurrent `-race` test that asserts the client is built exactly once (see the in-thread reply).
- **Aggregate skip logging:** restored a warn-level summary (`Skipped %d Trustify findings due to normalization errors`) in the adapter while keeping the per-finding debug detail (see in-thread reply).
- **`digestsFromImageRefs` helper:** you're right that it was only reached from its own test after digest extraction moved inline into the adapter — removed the helper and `TestDigestsFromImageRefs`.
- **Concurrent sync overlap (pre-existing):** agreed this is the reschedule-before-complete + `SystemWide` pattern and out of scope here; only the new lazy-init race is addressed in this PR.
- **`toModelFinding` trust boundary:** agreed — normalization stays the backend's responsibility; the sync task casts the already-canonical values, and future backends must normalize before returning DTOs.

internal/trustify/v2/scanner_adapter.go
```diff
@@ -0,0 +1,178 @@
...
+// initialization without real network I/O.
+var newVulnerabilityClient = NewVulnerabilityClient
+
+func (s *trustifyScanner) ensureClient(ctx context.Context) (VulnerabilityClient, error) {
+	s.once.Do(func() {
+		// A client injected at construction (in tests) is left untouched.
+		if s.client != nil {
+			return
+		}
+		s.client, s.initErr = newVulnerabilityClient(ctx, s.cfg)
```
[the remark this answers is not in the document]
EfratIfergan: Good catch — fixed in `d33d6b51`. Replaced `sync.Once` with a `sync.Mutex` in `ensureClient`: the client is cached **only on successful construction**, so a transient failure (canceled context, OIDC discovery network blip) is returned without being cached, and the next scan rebuilds the client rather than leaving vulnerability sync wedged until process restart. Construction stays race-safe and single-built (verified under `-race`). Added `TestTrustifyScanner_ensureClient_RetriesAfterTransientFailure`, which fails the first construction and asserts a later scan succeeds.

I opted not to try to classify "permanent vs transient" config errors — simply not caching the error and letting the next scan retry is more robust than a fragile classification heuristic.

internal/trustify/v2/scanner_adapter_test.go
```diff
@@ -0,0 +1,264 @@
...
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for i := 0; i < goroutines; i++ {
+		go func() {
+			defer wg.Done()
+			_, err := s.ScanImages(context.Background(), []vulnerability.ImageRef{{Digest: "sha256:aaaa"}})
+			req.NoError(err)
+		}()
+	}
+	wg.Wait()
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in `d33d6b51`. The concurrent-init test now signals completion through a channel after `wg.Wait()` returns and `select`s it against a 10s deadline, failing fast with a clear message (`concurrent scans did not complete within deadline`) instead of hanging until the suite-level timeout.

internal/trustify/v2/scanner_adapter_test.go
```diff
@@ -0,0 +1,302 @@
...
+	s := &trustifyScanner{cfg: &config.TrustifyConfig{Endpoint: "https://trustify.example"}}
+
+	const goroutines = 16
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for i := 0; i < goroutines; i++ {
+		go func() {
+			defer wg.Done()
+			_, err := s.ScanImages(context.Background(), []vulnerability.ImageRef{{Digest: "sha256:aaaa"}})
+			req.NoError(err)
```
galel12, line 241: Minor test hygiene: `req.NoError(err)` is called from child goroutines here. `testing.T` (and testify's `require`) are not goroutine-safe — a failure could panic or report the wrong stack.

Should we collect errors in the goroutines (e.g. `errCh`) and assert after `wg.Wait()` instead? The production mutex logic under test is fine; this is just about making the test itself robust.

galel12 (approved):
Round 1 feedback is addressed:

- `ensureClient` is mutex-guarded with transient-failure retry (better than the original `sync.Once` suggestion — errors aren't wedged permanently).
- Aggregate warn for normalization skips is restored.
- Dead `digestsFromImageRefs` helper and its test are removed.

No new functional blockers. One optional test-hygiene nit inline.

PR #3442, opened 2026-08-21
EDM-5284: Accept runAs alongside catalog item references

keitwb:
Previously this would fail validation at the initial API validation layer.

Assisted-by: Claude <noreply@anthropic.com>


celdrake:
Behavior seen as correct from UI/CLI, thanks.
Leaving formal review to others.

kkyrazis (approved):

PR #3247, opened 2026-07-19
EDM-4772: Partially consolidate CLI e2e VM setup

eldar101:
## Summary

Removes unnecessary per-test VM setup from API-only CLI e2e specs.

The CLI suite now starts with a no-VM harness by default and only starts a VM for specs labeled `needvm`. Specs that use console, device enrollment, agent version, or live device workload behavior keep VM setup. API-only CLI validation/list/create/edit/event specs avoid the VM cost.

`imagebuild`, `fips`, and `backup_restore` were inspected on current `main`; they already use no-VM or lazy-VM patterns, so this PR only changes `test/e2e/cli`.

## Release target

- [x] release-1.3

Target release: release-1.3

PR #3445, opened 2026-08-23
EDM-4858: Backport version check fix to release 1.2

eldar101:
Backports the 79621 version check behavior to release-1.2.

The test title and current behavior validate that the local CLI client matches the deployed server. release-1.2 still required the agent version to match the server, which can fail in this CLI-only test path even though the client/server versions are aligned

This keeps the test focused on client/server and expects the agent lookup to be skipped when no VM is initialized.

backport of fix made here:
https://github.com/flightctl/flightctl/pull/3247/changes#diff-6d90889beb67a5e1b18c665f20add64ca4e339ab6a32611ecf48fc85b7814fe5R526


dudyas6:
lgtm

dudyas6 (approved):

PR #3443, opened 2026-08-22
EDM-5218: Configure a writable OCI target for generated deltas

asafbennatan:
# Release target (required before merge to `main`)

Add at least one **`release-X.Y`** label to indicate which release(s) this change targets:

- [x] **`release-1.4`**

## EDM-5218: Configure a writable OCI target for generated deltas

**Jira:** https://issues.redhat.com/browse/EDM-5218
**Story type:** [DEV]
**Stack:** on [EDM-5217](https://github.com/flightctl/flightctl/pull/3439) (EDM-5213 → EDM-5214 → EDM-5215 → EDM-5217 → EDM-5218)

### Summary

Adds optional `repository` / `namespace` / `deltaStorageTarget` on `OciRepoSpec` so an org can designate one writable OCI target for generated deltas, without write access to every image’s own registry. Resolves the three delta push paths, rejects `namespace` on ImageBuild and ImageExport destinations, and falls back to deployment `DeltaGeneration.DefaultRepository` (credentials from Secret env, not `config.yaml`) when the org has no write target. Does not emit PrepareDeltas or generate/push deltas.

### Changes

- OpenAPI: optional mutually exclusive `repository` and `namespace`, plus `deltaStorageTarget`, on `OciRepoSpec`; `spec.registry` remains the login/CA hostname
- At most one `deltaStorageTarget` per org (unique partial index + HTTP 409)
- Delta push path resolution: neither → `{registry}/{image path}`; `repository` → `{registry}/{repository}`; `namespace` → `{registry}/{namespace}/{last segment}`
- ImageBuild and ImageExport destinations reject `namespace`; `repository` must equal `imageName`; registry-only dest unchanged
- Helm/quadlet: optional `deltaGeneration.defaultRepository`; username/password from Secret env on API only

### Testing

- **Unit tests:** object validation, uniqueness handler, three push paths, dest checks, DefaultRepository env credentials and write-target select
- **Integration tests:** second `deltaStorageTarget` in the same org is 409; another org can have its own; unique index and nil lookup
- **Coverage:** public helpers for name resolution, dest validation, and DefaultRepository conversion are fully covered through their APIs

### Acceptance Criteria

- [ ] AC-1: Optional mutually exclusive `repository`/`namespace`; at most one `deltaStorageTarget`; `spec.registry` remains login hostname
- [ ] AC-2: Three delta push paths for `quay.io/nginx/nginx` + `my-registry.com`
[...]

PR #3446, opened 2026-08-23
EDM-5219: Run generation on a dedicated delta-worker and queue

asafbennatan:

- [x] **`release-1.4`**

## EDM-5219: Run generation on a dedicated delta-worker and queue

**Jira:** https://issues.redhat.com/browse/EDM-5219
**Story type:** [DEV]
**Stack:** on [EDM-5218](https://github.com/flightctl/flightctl/pull/3443) (EDM-5213 → EDM-5214 → EDM-5215 → EDM-5217 → EDM-5218 → EDM-5219)

### Summary

Adds an unprivileged `flightctl-delta-worker` that consumes only `DeltaGenerationTaskQueue`, capped at `maxConcurrentDeltaGenerations` (default 2). `PrepareDeltas` is routed onto that queue; `DeltaGenerationCompleted` stays on `TaskQueue`. Idle consumers ack immediately. This story does not run `oci-delta create`, claim generations, or add queue maintenance. Redis stays non-durable (`save ""` / `appendonly no`).

### Changes

- OpenAPI: `PrepareDeltas` and `DeltaGenerationCompleted` reasons; `PrepareDeltasDetails` (`detailType`, optional `templateVersion`)
- `worker_client`: `WithDeltaPublisher` / `DeltaQueuePublisher`; nil delta publisher drops `PrepareDeltas` (no fall-through to TaskQueue)
- New process `cmd/flightctl-delta-worker`: Redis consumers on `delta-generation-queue` only, `WorkerCollector`, pprof port 15701
- Helm `deltaWorker` (always on) plus quadlet, RPM, el9/el10 images, restore lists, e2e registry, publish/collect-logs/redeploy membership
- `upgradeHooks.scaleDown` includes `flightctl-delta-worker` after `flightctl-worker`
- No Image Builder privileges, ClusterRole, Route, or Service

### Testing

- **Unit tests:** emit routing by queue; default concurrency 2; consumer count/queue name/idle ack; existing `shouldRolloutFleet` / `shouldRenderDevice` ignore the new reasons
- **Integration tests:** Redis consume-and-ack of `PrepareDeltas` on `DeltaGenerationTaskQueue`; same payload on `TaskQueue` is not taken
- **Coverage:** public routing and consumer contracts are covered. `Server.Run` is process Wait/Stop, same as `worker_server`

### Acceptance Criteria

- [ ] AC-1: Binary + Helm/quadlet consume `DeltaGenerationTaskQueue` only
- [ ] AC-2: Helm Redis `save ""` / `appendonly no`; prepare/generation remain in Postgres
- [ ] AC-3: `maxConcurrentDeltaGenerations` defaults to 2
- [ ] AC-4: `WorkerCollector` wired against `DeltaGenerationTaskQueue`
- [ ] AC-5: `flightctl-worker` still handles FleetRolloutStarted / RenderDevice; no `oci-delta create`


api/core/v1beta1/openapi.yaml
```diff
@@ -8308,6 +8312,19 @@ components:
...
           description: Number of errors encountered by this ResourceSync update.
+    PrepareDeltasDetails:
+      type: object
+      description: Structured details for PrepareDeltas events.
+      required:
+        - detailType
+      properties:
+        detailType:
+          type: string
+          enum: [PrepareDeltas]
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 0fe59570f: added `x-enum-varnames` for the new `PrepareDeltasDetails.detailType` enum value so the TypeScript client generator emits a named constant.

packaging/images/el10/Containerfile.delta-worker
```diff
@@ -0,0 +1,51 @@
...
+ENV SOURCE_GIT_TAG=${SOURCE_GIT_TAG}
+ENV SOURCE_GIT_TREE_STATE=${SOURCE_GIT_TREE_STATE}
+ENV SOURCE_GIT_COMMIT=${SOURCE_GIT_COMMIT}
+
+# Use BuildKit cache mounts for Go caching
+RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
+    --mount=type=cache,target=/opt/app-root/src/go/pkg/mod \
+    make build-delta-worker
+
+FROM quay.io/flightctl/flightctl-base:10.1-1769518576
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 0fe59570f: both el9 and el10 Containerfiles now use UBI minimal as the runtime base, run as non-root (`USER_UID=1001`), and no longer depend on `flightctl-base`.

deploy/helm/flightctl/templates/delta-worker/flightctl-delta-worker-deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+    spec:
+      serviceAccountName: flightctl-delta-worker
+      {{- if .Values.global.imagePullSecretName }}
+      imagePullSecrets:
+        - name: {{ .Values.global.imagePullSecretName }}
+      {{- end }}
+      initContainers:
+      {{- include "flightctl.databaseWaitInitContainer" (dict "context" . "userType" "app") | nindent 6 }}
+      {{- include "flightctl.migrationWaitInitContainer" (dict "context" . "timeout" 600) | nindent 6 }}
+      containers:
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 0fe59570f and dc6ee1a7c: the deployment now uses `{{- toYaml .Values.deltaWorker.resources | nindent 12 }}`, with defaults defined in `values.yaml.gotmpl` (100m/256Mi requests, 500m/512Mi limits).

deploy/helm/flightctl/templates/delta-worker/flightctl-delta-worker-deployment.yaml
```diff
@@ -0,0 +1,94 @@
...
+        - name: flightctl-delta-worker
+          image: {{ include "flightctl.ensureOsQualifiedImage" (dict "root" . "imageName" .Values.deltaWorker.image.image) }}:{{ default .Chart.AppVersion .Values.deltaWorker.image.tag }}
+          imagePullPolicy: {{ default .Values.global.imagePullPolicy .Values.deltaWorker.image.pullPolicy }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            runAsUser: 1001
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 0fe59570f and 0d737c421: added pod `fsGroup: 1001`, changed the encryption secret `defaultMode` to `0440`, and mounted a writable `/tmp` emptyDir for `readOnlyRootFilesystem`. This matches the alertmanager-proxy pattern and fixes the delta-worker pod readiness failure in deploy-and-upgrade.

internal/config/config.go
```diff
@@ -898,7 +898,15 @@ type QuayConfig struct {
...
-	DefaultRepository *DefaultRepositoryConfig `json:"defaultRepository,omitempty"`
+	DefaultRepository             *DefaultRepositoryConfig `json:"defaultRepository,omitempty"`
+	MaxConcurrentDeltaGenerations int                      `json:"maxConcurrentDeltaGenerations,omitempty"`
+}
+
+func (c *DeltaGenerationConfig) EffectiveMaxConcurrentDeltaGenerations() int {
+	if c == nil || c.MaxConcurrentDeltaGenerations <= 0 {
+		return 2
+	}
+	return c.MaxConcurrentDeltaGenerations
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 0fe59570f: `MaxConcurrentDeltaGenerations` is capped at 32 during config load so `LaunchConsumers` cannot spawn an unbounded number of goroutines from YAML misconfiguration.

deploy/podman/flightctl-delta-worker/flightctl-delta-worker.container
```diff
@@ -0,0 +1,29 @@
...
+PartOf=flightctl.target
+After=flightctl-db-wait.service flightctl-db-migrate.service flightctl-kv.service flightctl-certs-init.service
+Wants=flightctl-db-wait.service flightctl-db-migrate.service flightctl-kv.service flightctl-certs-init.service
+
+[Container]
+ContainerName=flightctl-delta-worker
+Image={{ .DeltaWorker.Image }}:{{ .DeltaWorker.Tag }}
+Pull=missing
+Network=flightctl.network
+User=1001
```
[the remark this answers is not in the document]
asafbennatan: Addressed in 0d737c421: added `DropCapability=ALL`, `NoNewPrivileges=true`, and `ReadOnly=true` to the quadlet unit alongside the existing `User=1001`.

api/core/v1beta1/openapi.yaml
```diff
@@ -8042,6 +8042,8 @@ components:
             - EncryptionMigrationStarted
             - EncryptionMigrationCompleted
             - ApplicationLifecycleChanged
+            - PrepareDeltas
+            - DeltaGenerationCompleted
```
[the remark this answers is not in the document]
asafbennatan: Intentionally not versioning this change in this PR. The new enum values extend the existing `GET /events` response and are required for the delta-generation event path. The `oasdiff-check` failure is acknowledged and the PR carries the `breaking-change-approved` label; `oasdiff-check` is passing in CI.

internal/delta_worker/consumer.go
```diff
@@ -0,0 +1,79 @@
...
+		if workerMetrics != nil {
+			workerMetrics.IncTasksByType(taskType)
+			workerMetrics.ObserveTaskExecutionDuration(taskType, time.Since(start))
+			workerMetrics.IncMessagesProcessed("success")
+			workerMetrics.UpdateLastSuccessfulTask()
+		}
+
+		ackCtx, cancel := context.WithTimeout(context.Background(), ackTimeout)
+		defer cancel()
+		if err := consumer.Complete(ackCtx, entryID, payload, nil); err != nil {
```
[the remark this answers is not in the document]
asafbennatan: Out of scope for EDM-5219. This story only introduces the delta-worker queue consumer and routes `PrepareDeltas` events to `DeltaGenerationTaskQueue`. Actual delta generation and publishing `DeltaGenerationCompleted` are implemented in follow-on stack PRs (EDM-5220+). The idle handler is a placeholder until that work lands.

internal/delta_worker/consumer.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/flightctl/flightctl/internal/consts"
+	"github.com/flightctl/flightctl/internal/instrumentation/metrics/worker"
+	"github.com/flightctl/flightctl/internal/worker_client"
+	"github.com/flightctl/flightctl/pkg/queues"
+	"github.com/sirupsen/logrus"
+)
+
+const ackTimeout = 5 * time.Second
+
+// LaunchConsumers starts Redis consumers on the delta-generation task queue.
```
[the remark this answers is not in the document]
asafbennatan: Same as the thread above: delta generation before ack is deferred to EDM-5220+. This PR establishes queue routing and a consuming worker skeleton only.

packaging/images/el10/Containerfile.delta-worker
```diff
@@ -0,0 +1,64 @@
...
+ENV SOURCE_GIT_TAG=${SOURCE_GIT_TAG}
+ENV SOURCE_GIT_TREE_STATE=${SOURCE_GIT_TREE_STATE}
+ENV SOURCE_GIT_COMMIT=${SOURCE_GIT_COMMIT}
+
+# Use BuildKit cache mounts for Go caching
+RUN --mount=type=cache,target=/opt/app-root/src/.cache/go-build \
+    --mount=type=cache,target=/opt/app-root/src/go/pkg/mod \
+    make build-delta-worker
+
+FROM registry.access.redhat.com/ubi10/ubi-minimal:10.1-1769677092
```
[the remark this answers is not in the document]
asafbennatan: Keeping `registry.access.redhat.com` to match the other checked-in Containerfiles in this repo (e.g. db-setup, cli-artifacts, alertmanager-proxy init containers). Switching only the delta-worker images to `registry.redhat.io` would be inconsistent; a repo-wide registry migration is a separate change.

asafbennatan:
> ## Code Review
> Nice work — the architecture is clean and well-organized. A dedicated `flightctl-delta-worker` binary consuming from `DeltaGenerationTaskQueue` via Redis is a solid separation of concerns. Security posture, config bounding, and test coverage (unit + integration) are all good.
> 
> A few items to consider:
> 
> ### Issues
> 1. **No liveness/readiness probes in the Helm deployment**
>    The deployment template exposes port 8080 for metrics but defines no `livenessProbe` or `readinessProbe`. If the process hangs or enters a deadlock, Kubernetes won't restart it. Consider at least a TCP liveness probe on the metrics port, or adding a `/healthz` endpoint.
> 2. **Idle handler silently drops messages**
>    The placeholder handler ACKs every message without processing. If this ships before the follow-on work (EDM-5220+), any `PrepareDeltas` events produced to the queue will be consumed and discarded silently. Consider either gating the producer side so events aren't enqueued until the handler is real, or having the handler log a warning.
> 3. **Consumer instances are never closed**
>    In `LaunchConsumers`, each `NewQueueConsumer` creates a consumer but the reference is discarded after calling `Consume`. If `QueueConsumer.Close()` performs cleanup (e.g., disconnecting from Redis), this is a resource leak on shutdown. The test mock has `Close()` — is there a matching production cleanup path?
> 4. **Metrics `ObserveTaskExecutionDuration` placed before `Complete()`**
>    The task duration metric is recorded before the ack call, so it doesn't include ack latency. Harmless now (idle handler), but when real delta generation logic lands, this placement should be reviewed.
> 
> ### Minor notes
> * **`HOME=/root` with `runAsUser: 1001`** — Works via explicit volume mounts, but unconventional. `HOME=/home/flightctl` would be more idiomatic.
> * **`ackTimeout` hardcoded at 5s** — If Redis is slow during shutdown, this could cause ack failures. Consider making it configurable.
> * **No `Validate()` on `DeltaGenerationConfig`** — Other config sections have explicit validation; adding one here would catch invalid values at startup rather than silently clamping.
> * **`context.Background()` for ack** — Intentional design choice (finish acking during shutdown), but a comment explaining the rationale would help future readers.
> * **No test for `Server.Run()` lifecycle** — Unit/integration tests cover `LaunchConsumers` well, but the full startup → context cancellation → graceful shutdown path isn't exercised.
> 
> Overall this is a solid foundational PR. 👍
> 
> _AI-generated. Review for accuracy._

concerns are invalid as the relevant ones are addressed later in the stack

internal/worker_client/worker_client.go
```diff
@@ -35,21 +38,51 @@ func QueuePublisher(ctx context.Context, queuesProvider queues.Provider) (queues
...
+		return
+	}
 	if !shouldEmitEvent(event.Reason) {
 		return
 	}
+	t.enqueue(ctx, orgId, event, t.publisher)
+}
 
+func (t *workerClient) enqueue(ctx context.Context, orgId uuid.UUID, event *domain.Event, producer queues.QueueProducer) {
+	if producer == nil {
```
galel12, line 90: When `producer` is nil we return without logging. The unit test documents that `PrepareDeltas` is dropped if `WithDeltaPublisher` is not wired, but in production a partial deploy or misconfigured client would lose prepare events silently.

Should we emit a warn log (reason + orgId) when a delta-routed event hits a nil producer?

internal/delta_worker/consumer.go
```diff
@@ -0,0 +1,80 @@
...
+		if workerMetrics != nil {
+			workerMetrics.IncMessagesInProgress()
+			defer workerMetrics.DecMessagesInProgress()
+		}
+
+		taskType := taskTypeFromPayload(payload, log)
+		if workerMetrics != nil {
+			workerMetrics.IncTasksByType(taskType)
+			workerMetrics.ObserveTaskExecutionDuration(taskType, time.Since(start))
+			workerMetrics.IncMessagesProcessed("success")
```
galel12, line 66: `taskTypeFromPayload` logs unmarshal errors and returns `"unknown"`, but we still record `IncMessagesProcessed("success")` and `Complete(..., nil)`. The regular task worker does the opposite on bad JSON (`IncPermanentFailures` + `permanent_failure`).

Intentional for the idle consumer in this story, but the test locks in ack-on-garbage. When the real handler lands, should invalid payloads follow the task-worker poison-message path instead of silent success?

internal/delta_worker/consumer.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const ackTimeout = 5 * time.Second
+
+// LaunchConsumers starts Redis consumers on the delta-generation task queue.
+func LaunchConsumers(ctx context.Context, queuesProvider queues.Provider, cfg *config.Config, workerMetrics *worker.WorkerCollector, log logrus.FieldLogger) error {
+	n := cfg.DeltaGeneration.EffectiveMaxConcurrentDeltaGenerations()
+	if workerMetrics != nil {
+		workerMetrics.SetConsumersActive(float64(n))
+		workerMetrics.SetQueueDepth(consts.DeltaGenerationTaskQueue, 0)
```
galel12: `WorkerCollector.monitorQueueHealth` notes we should avoid emitting placeholder queue-depth values, and the ticker does not set depth today. Pinning `0` here will read as an empty queue on dashboards until real polling exists.

Should we omit `SetQueueDepth` in the idle path (same as the regular worker) until depth is implemented?

cmd/flightctl-delta-worker/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if cfg.Metrics != nil && cfg.Metrics.Enabled {
+		if encMgr := encryption.GlobalManager(); encMgr != nil {
+			ec := encmetrics.NewEncryptionCollector(encMgr)
+			encMgr.SetMetricsRecorder(ec)
+			encCollector = ec
+		}
+	}
+
+	log.Println("Initializing data store")
+	db, err := store.InitDB(cfg, log)
```
galel12, line 68: This story's consumer only needs Redis, but startup still requires `InitDB` and the encryption canary. The pod will not run if Postgres is down even though nothing in this path uses the DB yet.

If that is intentional parity with `flightctl-worker`, a brief comment would help. Otherwise, could DB init wait until the generation handler needs it?

internal/worker_client/worker_client.go
asafbennatan: Addressed in 07eaff608: `EmitEvent` now logs a warning with `orgId` and `reason` when a delta-routed event is emitted without a configured delta publisher, instead of returning silently.

internal/delta_worker/consumer.go
asafbennatan: Addressed in 07eaff608: invalid JSON now follows the task-worker poison-message path — `IncPermanentFailures` + `IncMessagesProcessed("permanent_failure")`, then ack to remove the message from the queue (no retry).

internal/delta_worker/consumer.go
asafbennatan: Addressed in 07eaff608: removed the placeholder `SetQueueDepth(..., 0)` call. Queue depth is left unset until real polling is implemented, matching the regular worker.

cmd/flightctl-delta-worker/main.go
asafbennatan: Addressed in 07eaff608: removed `InitDB` and the encryption canary init from this story — the 5219 consumer only needs Redis. Postgres comes back in EDM-5220 when generation and persistence are wired in.

deploy/helm/flightctl/templates/delta-worker/flightctl-delta-worker-deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+      securityContext:
+        fsGroup: 1001
+        seccompProfile:
+          type: RuntimeDefault
+      {{- if .Values.global.imagePullSecretName }}
+      imagePullSecrets:
+        - name: {{ .Values.global.imagePullSecretName }}
+      {{- end }}
+      initContainers:
+      {{- include "flightctl.databaseWaitInitContainer" (dict "context" . "userType" "app") | nindent 6 }}
```
galel12, line 32: `main` no longer calls `InitDB`, but the Deployment still waits on Postgres (`databaseWaitInitContainer`, `migrationWaitInitContainer`), injects `DB_USER`/`DB_PASSWORD`, and mounts Postgres TLS volumes. The pod will not start until the DB is healthy even though this process only consumes Redis.

Should we drop the DB init containers, DB env, and SSL mounts for this story (keep KV + encryption)? Same question applies to the quadlet unit (`After=flightctl-db-wait.service` / `flightctl-db-migrate.service`) and the `database:` block in the config template.

internal/delta_worker/consumer.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+			return nil
+		}
+
+		taskType := string(event.Event.Reason)
+		if taskType == "" {
+			taskType = "unknown"
+		}
+		if workerMetrics != nil {
+			workerMetrics.IncTasksByType(taskType)
```
galel12, line 64: After unmarshal, any well-formed `Event` is acked as success regardless of `reason`. A mis-routed event (e.g. `FleetRolloutStarted` JSON on `delta-generation-queue`) would be consumed and dropped without a permanent-failure signal.

Should `idleHandler` require `event.Event.Reason == PrepareDeltas` and treat anything else like a poison message (log + `permanent_failure`)? A unit test for wrong-reason JSON would lock that in before EDM-5220 adds real processing.

deploy/helm/flightctl/templates/db/flightctl-db-admin-rbac.yaml
```diff
@@ -151,6 +151,9 @@ subjects:
 - kind: ServiceAccount
   name: flightctl-worker
   namespace: {{ default .Release.Namespace .Values.global.internalNamespace }}
+- kind: ServiceAccount
```
galel12, line 154: `flightctl-delta-worker` is added to the DB admin RBAC subjects, but the binary no longer opens a DB connection. Should this service account be removed from the DB role bindings unless a follow-on story needs it?

cmd/flightctl-delta-worker/main.go
asafbennatan: Updated in 69b4fb772: restored `InitDB` and the encryption canary, and added a brief comment that this matches `flightctl-worker` startup — Postgres is required for the canary now and for generation in follow-on stack stories.

deploy/helm/flightctl/templates/delta-worker/flightctl-delta-worker-deployment.yaml
asafbennatan: Addressed in 69b4fb772: reverted the DB removal from `main.go`, so the Deployment DB wait containers, credentials, and TLS mounts are consistent with what the binary initializes again.

deploy/helm/flightctl/templates/db/flightctl-db-admin-rbac.yaml
asafbennatan: Addressed in 69b4fb772: the delta-worker binary opens Postgres again at startup, so keeping `flightctl-delta-worker` in the DB admin RBAC bindings remains correct.

internal/delta_worker/consumer.go
asafbennatan: Addressed in b86e26666: after unmarshal the consumer checks `worker_client.IsDeltaGenerationQueueEvent` (same allowlist as the publisher). Mis-routed reasons are logged, counted as `permanent_failure`, and acked to drop them from the queue — same path as invalid JSON.

galel12 (approved):
LGTM

PR #3447, opened 2026-08-23
EDM-4845: Retry transient vulnerability fleet creation

eldar101:
## Summary

- Retry transient API transport failures when the vulnerability lifecycle test creates its fleet after suite setup restarts services.
- Preserve fail-fast behavior for non-retryable fleet creation errors with `StopTrying`.

## Release target

- [x] release-1.3

Target release: release-1.3


Samudelacruz77 (approved):
lgtm

PR #3448, opened 2026-08-23
EDM-5207 : Reset device labels in E2E helper

eldar101:
## Summary
- keep the E2E test-id label when SetLabelsForDevice is called with nil/empty labels
- ensure reset calls remove fleet-selector labels instead of sending nil labels that ReplaceDevice preserves

This fixes the E2E helper so “reset labels” actually removes the fleet labels from the device.
Before, SetLabelsForDevice(deviceID, nil) set labels to nil, but the API preserves nil labels on replace, so old labels stayed. The test then kept matching both fleets and timed out waiting for MultipleOwners to clear.
Now the helper keeps only the test-id label, removing the fleet selector labels while preserving cleanup.

## Release target
release-1.3
Target release: release-1.3


dudyas6 (approved):
lgtm

PR #3449, opened 2026-08-23
EDM-5289: Wait for catalog ref OS updates through reboot

eldar101:
## Summary
- use the existing reboot-aware rendered-version wait for OS catalog-ref updates
- avoid the 5m render-only wait while the device is still applying/rebooting into the resolved image

## Release target
release-1.3
Target release: release-1.3


Samudelacruz77 (approved):
LGTM

PR #3451, opened 2026-08-24
EDM-4115: e2e test for vm app with volumes

itzikb-redhat:
e2e tests for vm app with volumes

Add two volumes to a vm app:
A disk with a file with the text hello
An empty disk

Mount the disks and check that the file exists in the first disk and
the second one is writeable.

Other changes:
Used a template for vm spec
Check that SSH is open before login to the vm app

### DIFF
--- a/test/e2e/vm/vm_test.go
+++ b/test/e2e/vm/vm_test.go
@@ -54,6 +54,15 @@ const (
 	vmBPublishedUDPPort              = 9090
 	vmGuestSSHPort                   = 22
 	vmGuestMemory                    = "1024M"
+	vmHostDataImagePath              = "/var/lib/flightctl/vm-data/test-vm.img"
+	vmExtraDataSize                  = "10M"
+	vmHostDataHelloContent           = "hello"
+	vmHostDataGuestDisk              = "vdc"
+	vmExtraDataGuestDisk             = "vdd"
+	vmHostVolumeSetupScriptPath      = "/usr/local/bin/verify-vm-host-volumes.sh"
+	vmHostDataHelloFile              = "/var/tmp/host-data-hello.txt"
+	vmExtraDataUsedFile              = "/var/tmp/extradata-used.txt"
+	vmExtraDataUsedContent           = "used"
 	configDriveIndexHTMLContent      = "Hello from ConfigDrive"
 	systemdSubStateActive            = "active"
 	systemdSubStateRunning           = "running"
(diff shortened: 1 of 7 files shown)


asafbennatan (commented):
See comments below

test/e2e/vm/vm_test.go
```diff
@@ -453,6 +508,111 @@ func getPodmanContainerID(h *e2e.Harness, containerName string) string {
...
+		g.Expect(sshErr).NotTo(HaveOccurred(), "reading guest file %s failed", path)
+		g.Expect(out).To(ContainSubstring(want), "guest file %s", path)
+	}, testutil.LONG_TIMEOUT, testutil.POLLING).Should(Succeed())
+}
+
+func expectGuestDiskCount(h *e2e.Harness, want int) []string {
+	GinkgoHelper()
+	var disks []string
+	Eventually(func(g Gomega) {
+		out, sshErr := h.RunSSHOnDeviceLocalPortRaw(vmPublishedSSHPort, vmGuestUser, vmGuestPassword, "sudo", "-n", "lsblk", "--virtio", "-d", "-n", "-l", "-v", "-o", "NAME")
```
asafbennatan: `expectGuestDiskCount` uses `RunSSHOnDeviceLocalPortRaw` and then splits the entire stdout into disk names. Nested SSH already documents that device or guest profiles can print onto stdout, which is why the last-line helper exists.

`lsblk ... -o NAME` prints one disk per line:

```
vda
vdb
vdc
vdd
```

Nested SSH can prefix that with MOTD text. `parseLsblkDisks` splits on whitespace, so `Welcome to Fedora` becomes extra "disks" and `HaveLen(4)` never matches.

Using last-line SSH on that output would keep only `vdd` and drop the other three. `paste -sd, -` joins the names onto one line first:

```
vda,vdb,vdc,vdd
```

Then last-line SSH keeps that whole list, and `parseLsblkDisks` already splits on commas.

```suggestion
		out, sshErr := h.RunSSHOnDeviceLocalPort(vmPublishedSSHPort, vmGuestUser, vmGuestPassword, "bash", "-lc", "lsblk --virtio -d -n -l -o NAME | paste -sd, -")
```

test/harness/e2e/harness_application.go
```diff
@@ -44,6 +44,79 @@ func VMYAMLWithCPU(name, guestMemory, image string, cpuCores int, cloudInitUserD
 	return vmYAMLManifest(name, guestMemory, image, cpuCores, VMCloudInitNoCloudVolume(cloudInitUserData))
 }
 
+// VMYAMLWithHostVolumes builds a VM manifest with optional hostDisk and blank dataVolume disks.
```
asafbennatan: `VMYAMLWithHostVolumes` is a second full VirtualMachine sprintf next to `vmYAMLManifest`, with extra disks glued in by concatenating indented fragments. YAML layout then lives in `%s` slots and has to stay aligned by hand.

The e2e libvirt helper already does this with `text/template` and `//go:embed domain-template.xml`. Would a single VM YAML template with `{{if .HostDiskPath}}` / `{{if .ExtraDataSize}}` (and the CPU block the same way) replace both sprintf functions? Indentation stays in the template file, and optional volumes do not need a second copy of the manifest.

test/e2e/vm/vm_test.go
```diff
@@ -342,6 +352,51 @@ var _ = Describe("VM Applications", Ordered, func() {
 		By("Verifying serial console login with the updated cloud-init password")
 		expectSerialLoginWithPassword(harness, deviceID, vmAppName, vmGuestUser, vmModifiedGuestPassword)
 	})
+
+	// HostDisk plus a blank dataVolume appear as extra guest disks. Removing both
```
asafbennatan, line 367: This spec attaches `hostDisk` plus a blank `dataVolume` and then removes them. Extra VM disks live in `vm.yaml`; `VmApplication` has no `volumes:` (`image` / `mount` are compose/quadlet/container only). On a FlightCtl device, vm-to-quadlet is what actually materializes disks (there is no cluster CDI). Expected guest outcomes:

- `containerDisk`: the guest sees a disk whose contents come from an OCI image. Nothing has to exist on the device beforehand. This suite already uses it for the Fedora OS disk. A seeded extra disk with a known file would come from an image that already contains that file.
- `emptyDisk` (`capacity`): virt-launcher creates a sparse qcow2 in emptyDir at VM start. The guest sees an empty writable disk. A guest reboot keeps it. Recreating the virt-launcher pod does not: in standalone/quadlet, emptyDir is tmpfs and is fresh on every pod start. There is no `qemu-img` init container and no host path.
- `dataVolume` with `source.blank`: vm-to-quadlet injects an `init-disk-*` container that runs `qemu-img create -f raw disk.img <size>` if the file is missing (or resizes if it is too small). The guest sees an empty raw disk of that size. The file lives in a Podman named volume, so it survives virt-launcher restarts on that host. That is the provisioning step; it is not a Kubernetes StorageClass.
- `hostDisk`: the guest sees the file at the given host path. vm-to-quadlet does not run `qemu-img` for this type. `DiskOrCreate` will create an empty file if the path is missing. This spec uses `DiskOrFail`, so the path must already exist or the VM does not start. Putting `hello.txt` on it still requires something else to format and seed the image.

This test seeds `hello` by formatting a host file, then attaches it with `DiskOrFail`. The empty extra disk is `dataVolume` blank, which the converter will create. If the contract is attach/detach of those two kinds, that matches the spec name. If the contract is the PR body's "disk with a file" and "empty writable disk" without a host-side prep script, would `containerDisk` plus either `emptyDisk` or blank `dataVolume` be closer?

test/e2e/vm/vm_test.go
itzikb-redhat: Done

test/e2e/vm/vm_test.go
itzikb-redhat: It's intended to be disk with a file from the host as well a writeable empty disk

test/harness/e2e/harness_application.go
itzikb-redhat: Updated

test/e2e/vm/vm_test.go
asafbennatan: this is fine , just making sure , that the intention here is to test host disk under the context of possible disk types mentioned above

asafbennatan (dismissed):
Looks good. Templates and the hostDisk + blank dataVolume contract match the intent.

The lsblk helper still uses full nested-SSH stdout (`expectGuestDiskCount`); MOTD can still inflate the disk list. Not blocking.

test/e2e/vm/vm_test.go
asafbennatan: The lsblk helper still uses full nested-SSH stdout (expectGuestDiskCount); MOTD can still inflate the disk list. Not blocking.

itzikb-redhat:
> The lsblk helper still uses full nested-SSH stdout (`expectGuestDiskCount`); MOTD can still inflate the disk list. Not blocking.

In the end expectGuestDiskCount is using runSSHOnDeviceLocalPort which runs with RequestTTY=no so MOTD should not be shown.

asafbennatan:
> In the end expectGuestDiskCount is using runSSHOnDeviceLocalPort which runs with RequestTTY=no so MOTD should not be shown.

MOTD was an example , 
expectGuestDiskCount taking full stdout and parseLsblkDisks splitting on commas, newlines, and spaces. Any extra word on that stream becomes a fake disk

paste -sd, - plus last-line SSH is still the right shape: one line of names, then the same trim whoami already uses. Raw is there so multi-line lsblk is not reduced to vdd. Join first, then last-line.

in any case i consider this a nit

asafbennatan (approved):
LGTM , see the nit above

itzikb-redhat:
> LGTM , see the nit above

Please take a look at the change.10x

asafbennatan (approved):
LGTM

PR #3452, opened 2026-08-24
EDM-4115: e2e test for a domain of a vm app is crashed

itzikb-redhat:
e2e test for a domain of a vm app is crashed

Suspend a vm app domain - Application summary should be Degraded
Resume the domain - Should return to Healthy


asafbennatan (approved):
LGTM

PR #3453, opened 2026-08-24
EDM-5221: Emit PrepareDeltas types and list OS digest pairs

asafbennatan:

## EDM-5221: Emit PrepareDeltas types and list OS digest pairs

**Jira:** https://issues.redhat.com/browse/EDM-5221
**Story type:** [DEV]

Depends on #3450 (EDM-5220). Stacked on `EDM-5220-generate-os-delta`.

### Summary

The delta-worker can now list unique OS current→new digest pairs for a `PrepareDeltas` event without writing device spec or a rendered version. Fleet members use the same substitution as rollout, then inspect the rendered OS image. Skip (no prepare row) when `generateDelta` is false, the org has no writable registry, or no eligible device remains.

A device is eligible for **OS delta generation** when `status.systemInfo.deltaEligible` is true (oci-delta present) **and** `status.systemInfo.bootcVersion` is set.

### Changes

- Confirm `PrepareDeltas` / `PrepareDeltasDetails` (fleet `templateVersion`, omitted for device, no `rolloutStrategy`) with schema round-trip tests
- Optional fleet `RolloutPolicy.generateDelta` (nil = true)
- Extract in-memory `DesiredSpecFromTemplate` from fleet rollout and `RenderSpec` from device render
- `Resolver.DeltaCandidates` lists OS pairs from unsaved render, skip paths, and org-level dedup
- `Expand` hook only; nested/app pairs stay EDM-5226

Not in this PR: persist/ack (EDM-5222), emit instead of rollout/render (EDM-5223). `fleetValidate` and `UpdateDevice` are unchanged.

### Testing

- **Unit tests:** skip contracts, fleet/device resolve without spec write, missing digest omit, ineligible omit, inspect fail-closed, unparseable image skip, Expand identity, org dedup, PrepareDeltas JSON without `rolloutStrategy`, generateDelta omit/false/true
- **Integration tests:** N/A (inspect mocked)
- **Coverage:** story packages ≥90% of new public contracts

### Acceptance Criteria

- [ ] AC-1: `PrepareDeltas` / `PrepareDeltasDetails` on Event schema and worker allowlist; fleet `templateVersion`, omitted for device; no `rolloutStrategy`; `fleetValidate` and `UpdateDevice` unchanged
- [ ] AC-2: List OS current→new digest pairs (current `status.os.imageDigest`, new from inspect of rendered OS image); fleet uses rollout substitution with no Device.Spec write; skip unless systemInfo deltaEligible + bootcVersion; missing current digest → no pair; dedup in org
- [ ] AC-3: Skip (no prepare row) when `generateDelta` is false, no writable registry, or no eligible device
- [ ] AC-4: Two eligible devices with the same current and new digest produce one pair


internal/delta_worker/delta_candidates.go
```diff
@@ -0,0 +1,262 @@
...
+	if r.Render == nil {
+		return nil, fmt.Errorf("render is required")
+	}
+	rendered, err := r.Render(ctx, spec)
+	if err != nil {
+		return nil, nil
+	}
+
+	var candidates []DeltaCandidate
+	if cand, ok, err := r.osCandidate(ctx, device, rendered); err != nil {
```
asafbss, line 163: desiredSpec and Render errors above both do return nil, nil (skip the device). But osCandidate / Inspect error does return nil, err which propagates up and fails the whole fleet. Should be return nil, nil to be consistent — a transient registry issue for one device's image shouldn't block delta generation for the entire fleet.

internal/tasks/desired_spec.go
```diff
@@ -0,0 +1,58 @@
...
+	return spec, nil
+}
+
+func assembleDesiredSpec(device *domain.Device, tv *domain.TemplateVersion) (*domain.DeviceSpec, []model.DependencyRef, []error) {
+	var f FleetRolloutsLogic
+	errs := []error{}
+
+	var osSpec *domain.DeviceOsSpec
+	if tv.Status != nil && tv.Status.Os != nil {
+		osSpec = &domain.DeviceOsSpec{CatalogItemRef: tv.Status.Os.CatalogItemRef}
```
asafbss: The old code in fleet_rollout.go was just osSpec = &domain.DeviceOsSpec{Image: img} — it didn't copy CatalogItemRef. Since assembleDesiredSpec is called from the real fleet rollout path too, the first rollout after this lands will trigger unnecessary spec writes for every device with catalog-based OS images (DeviceSpecsAreEqual will see nil vs non-nil).

internal/delta_worker/delta_candidates_test.go
```diff
@@ -0,0 +1,570 @@
...
+		r := baseResolver()
+		r.Devices = func(_ context.Context, _ uuid.UUID, _ string) ([]*domain.Device, error) {
+			return []*domain.Device{
+				deviceWithOS("bad", true, currentDig),
+				deviceWithOS("good", true, currentDig),
+			}, nil
+		}
+		r.Render = func(_ context.Context, spec *domain.DeviceSpec) (tasks.RenderedSpec, error) {
+			if spec.Os.Image == "bad" {
+				return tasks.RenderedSpec{OsImage: "not a valid image!!!"}, nil
```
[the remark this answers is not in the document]
asafbennatan: Fixed in the current changes. The test now derives the desired OS image from the device name, returning "bad" for the failing device, so the unparseable-image path is exercised and the valid device remains as the single candidate. Verified with go test ./internal/delta_worker.

internal/tasks/desired_spec.go
```diff
@@ -0,0 +1,58 @@
...
+		img, err := ReplaceParametersInString(tv.Status.Os.Image, device)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed replacing parameters in OS image: %w", err))
+		} else {
+			osSpec.Image = img
+		}
+	}
+
+	deviceConfig, depRefs, configErrs := f.getDeviceConfig(device, tv)
```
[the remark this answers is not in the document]
asafbennatan: Fixed in the current changes. assembleDesiredSpec now returns a not-ready error before invoking helpers when the TemplateVersion or its Status is nil, preventing the panic. Added a regression test and verified with go test ./internal/tasks.

internal/delta_worker/delta_candidates.go
asafbennatan: Fixed in the current changes. Inspect failures are now treated as device-local: the affected device is omitted, while the fleet continues processing and skips only when no candidates remain. Updated the test and verified with go test ./internal/delta_worker.

internal/tasks/desired_spec.go
asafbennatan: Fixed in the current changes. assembleDesiredSpec no longer copies CatalogItemRef into the generated DeviceSpec, preserving the previous fleet-rollout behavior and avoiding unnecessary spec writes. Updated the regression test and verified with go test ./internal/tasks.

asafbss (approved):

PR #3457, opened 2026-08-24
EDM-5223: Hold rollout behind PrepareDeltas and resume once

asafbennatan:

## EDM-5223: Hold rollout behind PrepareDeltas and resume once

**Jira:** https://issues.redhat.com/browse/EDM-5223
**Story type:** [DEV]
**Stacked on:** #3455 (EDM-5222)

### Summary

Fleet template-version create and standalone device spec updates emit `PrepareDeltas` instead of starting rollout or render. When every joined generation pair is terminal, or when `flightctl-periodic` finds a waiting prepare past its deadline, the wait CAS-completes or CAS-fails once, matches live TV or spec generation, then emits `FleetRolloutStarted` or `DeltaGenerationCompleted`. A superceded wait does not emit. Per-pair generation phase is reported as DeltaGenerationProgress events (EDM-5222), not on status.deltaGeneration.

### Changes

- `CreateTemplateVersion` no longer emits `FleetRolloutStarted`. `fleetValidate` creates the TV, skips the TV annotation / `SetOutOfDate`, and emits `PrepareDeltas`.
- Standalone `UpdateDevice` spec changes set delay-render and emit `PrepareDeltas` with no `templateVersion`.
- Last-pair persist calls `CompleteWaitingIfTerminal` → CAS `waiting`→`complete`, then identity match, then annotate / `SetOutOfDate` / resume event.
- New system-wide periodic task `delta-prepare-deadline` (1m): `ListWaitingPastDeadline`, CAS `waiting`→`failed`, same match/emit. Generation rows are not the resume trigger.
- Prepare identity for fleets is the `PrepareDeltas` TV (else latest TV), not `FleetAnnotationTemplateVersion` (that annotation is written on resume).
- `shouldRenderDevice` treats `DeltaGenerationCompleted` as render-now.

### Testing

- **Unit tests:** CAS win/lose, stale TV/RV, supercede, live batched vs immediate strategy, fleetValidate PrepareDeltas, standalone delay, deadline poll, identity keyed to event TV
- **Integration tests:** last-pair emit, past-deadline fail + emit with `in_progress` generation unchanged, null deadline no emit, device-selection fleet does not write member specs on TV create
- **Validation:** `make lint`, `make unit-test`, `make integration-test` passed

### Acceptance Criteria

- [ ] AC-1: `fleetValidate` still creates the TemplateVersion, then emits `PrepareDeltas` instead of starting fleet rollout. `UpdateDevice` writes `Device.Spec`, emits `PrepareDeltas` (no `templateVersion`), and delays `RenderDevice`.
- [ ] AC-2: Last-pair CAS `waiting`→`complete` and emit only if latest TV or device spec RV still matches. Only one resume wins.
- [ ] AC-3: Periodic lists waiting prepares with past deadline, CAS `waiting`→`failed`, same match. Omit `maxWaitForDelta`: no periodic deadline. Sweeping a `delta_generations` row does not resume.
- [ ] AC-4: Fleet resume sets `FleetAnnotationTemplateVersion`, `SetOutOfDate`, emits `FleetRolloutStarted` from the live Fleet strategy, clears preparing status. Device resume: `DeltaGenerationCompleted` → render if RV matches.
- [ ] AC-5: Supercede does not emit. Stale TV/RV does not emit.
- [ ] AC-6: Deadline `failed` still emits. In-flight generation jobs keep running under `deltaGenerationTimeout`.


internal/tasks/fleet_validate.go
```diff
@@ -117,27 +104,73 @@ func (t *FleetValidateLogic) CreateNewTemplateVersionIfFleetValid(ctx context.Co
...
+	details := domain.PrepareDeltasDetails{
+		DetailType:      domain.PrepareDeltasDetailsDetailType("PrepareDeltas"),
+		TemplateVersion: &tvName,
+		ResourceVersion: resourceVersion,
+	}
+	var eventDetails domain.EventDetails
+	if err := eventDetails.FromPrepareDeltasDetails(details); err != nil {
+		return err
+	}
+	t.WorkerClient.EmitEvent(ctx, t.orgId, domain.GetBaseEvent(ctx, domain.FleetKind, fleetName, domain.EventReasonPrepareDeltas, "Preparing OS image deltas", &eventDetails))
```
galel12, line 172: Could we use `ReliableWorkerClient.EmitEventWithError` here and return publication failures? `EmitEvent` only logs queue errors, then this method returns `nil`. Since the prepare marker was already written, a failed publish leaves no prepare row for the deadline poller to recover, and this update never emits `FleetRolloutStarted`. Returning the error would let the existing consumer retry the validation.

internal/tasks/delta_prepare_deadline.go
```diff
@@ -0,0 +1,280 @@
...
+		if status.Code != http.StatusOK {
+			if status.Code == http.StatusNotFound {
+				return false, nil
+			}
+			return false, fmt.Errorf("getting device %s: %s", prep.Name, status.Message)
+		}
+		if device == nil || device.Metadata.Generation == nil {
+			return false, nil
+		}
+		return prep.SourceResourceVersion == *device.Metadata.Generation, nil
```
galel12, line 183: Could we compare the stored `SpecHash` with the live device spec, or carry the generation explicitly? `emitStandalonePrepareDeltas` stores `device.Metadata.ResourceVersion` in the event details, but this compares it with `Metadata.Generation`. Status updates bump resource version without changing generation, so a still-current prepare can be marked failed here without clearing `DeviceDeltaPreparing` or emitting the render-resume event. The deadline test sets these values equal, which misses the mismatch.

internal/tasks/consumer.go
```diff
@@ -78,56 +95,69 @@ func dispatchTasks(fleetSvc fleetservice.Service, templateversionSvc templatever
...
-				return fleetValidate(ctx, eventWithOrgId.OrgId, eventWithOrgId.Event, fleetSvc, templateversionSvc, deviceSvc, repositorySvc, k8sClient, log)
+			err = runTaskWithMetrics(taskName, d.WorkerMetrics, func() error {
+				if eventWithOrgId.Event.InvolvedObject.Kind != domain.FleetKind {
+					log.Errorf("FleetValidate called with unexpected kind %s and reason %s", eventWithOrgId.Event.InvolvedObject.Kind, eventWithOrgId.Event.Reason)
+					return nil
+				}
+				logic := NewFleetValidateLogic(log, d.FleetSvc, d.TemplateversionSvc, d.DeviceSvc, d.RepositorySvc, d.K8sClient, eventWithOrgId.OrgId, eventWithOrgId.Event)
+				logic.WorkerClient = d.WorkerClient
+				if err := logic.CreateNewTemplateVersionIfFleetValid(ctx); err != nil {
+					return fmt.Errorf("validate fleet %s/%s: %w",
```
galel12, line 120: Could we distinguish expected validation failures from retryable operational errors here? `validateConfig` errors are returned via `setStatus` after setting `FleetValid=False`, so this closure schedules an unchanged invalid fleet for repeated retries instead of waiting for a fleet or repository change to trigger validation again. Could expected validation failures be acknowledged while transient store or publish errors remain retryable?

asafbennatan: [codex] Addressed in 16fc1db57: permanent Fleet configuration errors are acknowledged after FleetValid=False is persisted, while operational failures—including status-store and delta-publication failures—remain retryable.

galel12 (approved):
LGTM

internal/delta_worker/tasks/prepare/handler.go
```diff
@@ -232,19 +235,23 @@ func (p *Handler) admitPrepare(ctx context.Context, orgId uuid.UUID, kind, name
...
+		completion := &model.DeltaPrepare{
 			OrgID:                 orgId,
 			Kind:                  kind,
 			Name:                  name,
 			TemplateVersion:       identity.templateVersion,
 			SpecHash:              identity.specHash,
 			SourceResourceVersion: identity.resourceVersion,
-		})
+		}
+		if err := p.prepareService.SetDeltaPreparingStatus(ctx, completion, 0, 0); err != nil {
```
galel12: Could we verify this skipped prepare still matches the live Fleet generation/spec before setting the preparing marker? A Fleet update can land while its validation event is still queued; in that window this event’s TemplateVersion may still be latest and the prepare-resource-version annotation may still match. This call can then initialize the old marker, allowing completion to emit `FleetRolloutStarted` for a stale template. Could we fence completion against the live Fleet identity and add a regression test for that ordering?

asafbennatan: adjusted

galel12 (approved):
LGTM

PR #3463, opened 2026-08-25
EDM-5250: Quay configuration and backend selection

EfratIfergan:
## EDM-5250: Quay configuration and backend selection

**Jira:** https://issues.redhat.com/browse/EDM-5250
**Story type:** [DEV]
**Stacked on:** #3437 (EDM-5249) → retarget to `main` once EDM-5248/5249 merge.

### Summary

Adds Quay as a selectable vulnerability backend by wiring the config and
data-model foundation, so `backend: "quay"` becomes a valid, fully-plumbed
selection. Every persisted finding now carries a `source` (which backend
produced it), exposed in the vulnerability API. The Quay scanner itself lands
in EDM-5251; this story lays the groundwork and keeps SBOM-upload
backend-agnostic. Backend selection follows the existing scanner-registry
pattern, and SBOM-upload is modeled as a declared per-backend capability rather
than a hardcoded `!= trustify` conditional.

### Changes

**Config (`internal/config`)**
- `VulnerabilityBackendQuay = "quay"` enum constant and typed `QuayConfig`
  (`Endpoint`, `Token` SecureString, `MaxConcurrentRequests`, default 5).
- `Quay *QuayConfig` field on `VulnerabilityConfig`; `EffectiveBackend()` pure
  resolver (empty → trustify when a Trustify block is present).
- `FLIGHTCTL_VULNERABILITY_REPORTING_BACKEND` / `_QUAY_ENDPOINT` / `_QUAY_TOKEN`
  / `_QUAY_MAX_CONCURRENT_REQUESTS` env overrides, applied centrally.

**Backend capability (`internal/vulnerability`, `internal/trustify/v2`)**
- Registry backends declare capabilities via `Register(name, factory, WithSBOMUpload(bool))`;
  `RequiresSBOMUpload(backend)` reports whether a backend needs client-side SBOM upload.
- Trustify registers with `WithSBOMUpload(true)`; `NewScanner` resolves the empty
  backend via `EffectiveBackend()` (deprecation warning preserved).

**SBOM gating (`internal/imagebuilder_worker/tasks`)**
- SBOM pipeline/upload gates now key off `RequiresSBOMUpload(EffectiveBackend())`
  instead of `Trustify != nil` — no backend literal in imagebuilder code.

**Data model + migration (`internal/store`, `internal/migration`)**
- Nullable `source VARCHAR(32)` column on `vulnerability_findings` (added via AutoMigrate).
- Key-guarded, idempotent backfill sets `source='trustify'` on pre-existing NULL rows.
- `UpsertFindings` persists `source` (insert + `ON CONFLICT DO UPDATE`).

**Sync + API (`internal/tasks`, `internal/periodic_checker`, `internal/service`, `api`)**
- Sync stamps the resolved backend as each finding's `source`.
- `source` exposed on the `Vulnerability` OpenAPI schema (additive, optional);
  hardcoded "Trustify" language genericized to "the configured vulnerability scanner".

### Testing

- **Unit tests:** config (Quay JSON round-trip, enum, env overrides incl.
  invalid/default, `EffectiveBackend` table); registry (`RequiresSBOMUpload`
  matrix, unknown/empty backend errors); imagebuilder SBOM-capability gating
  (incl. discriminating "stale trustify block + backend=quay" case); sync
  source stamping; handler `source` mapping.
- **Integration tests:** migration column-add + NULL→`trustify` backfill;
  `UpsertFindings` source persist + on-conflict refresh; sync round-trip.
  Store (vulnerability) 67 specs and imagebuilder_worker 42 specs pass.
- **Coverage:** new-code coverage comprehensive — registry/handler/sync
  functions at 100%, `EffectiveBackend` 85.7%, backfill exercised by
  integration. `make lint` clean.

> Note: `make generate`/`make tidy` could not be run in the local dev
> environment due to a root-owned Go module-cache directory (unrelated to this
> change); generated files were verified current by inspection and `git diff`.
> These run clean in CI.

### Acceptance Criteria

- [ ] AC-1: `QuayConfig` struct (Endpoint, Token SecureString, MaxConcurrentRequests default 5)
- [ ] AC-2: `VulnerabilityBackend` enum with Trustify + Quay constants
- [ ] AC-3: `VulnerabilityConfig` has Backend/Trustify/Quay fields
- [ ] AC-4: `_BACKEND` + `_QUAY_*` env overrides applied centrally
- [ ] AC-5: `NewScanner` errors on unknown/empty backend
- [ ] AC-6: `backend: "quay"` wired via registry (scanner arrives in EDM-5251)
- [ ] AC-7: imagebuilder skips `UploadSBOM` when backend ≠ trustify
- [ ] AC-8: nullable `source` column + backfill; persisted on write
- [ ] AC-9: `source` in API response + OpenAPI
- [ ] AC-10: hardcoded "Trustify" language genericized in OpenAPI/service


asafbennatan (commented):
See comments below

internal/config/config.go
```diff
@@ -793,8 +793,14 @@ type VulnerabilityBackend string
 const (
 	// VulnerabilityBackendTrustify selects the Trustify (TPA) backend.
 	VulnerabilityBackendTrustify VulnerabilityBackend = "trustify"
+	// VulnerabilityBackendQuay selects the Quay Security API backend.
+	VulnerabilityBackendQuay VulnerabilityBackend = "quay"
```
asafbennatan, line 797: `VulnerabilityBackend(backend)` is an unchecked cast, so a typo still lands in `EffectiveBackend`. Imagebuilder then treats unknown names as skip-SBOM.

Should we parse the string into those two constants (empty still allowed for the trustify default), and use the same helper from env and from `UnmarshalJSON` so file load cannot bypass it? The value set should stay identical to the OpenAPI `source` enum; config should keep this type rather than importing the generated v1alpha1 one.

internal/config/config.go
```diff
@@ -1254,6 +1297,42 @@ func applyVulnerabilityReportingEnvVarOverrides(c *Config) {
...
+// applyVulnerabilityReportingDefaults fills in defaults for vulnerability
+// reporting sub-configs that are present but under-specified.
+func applyVulnerabilityReportingDefaults(c *Config) {
+	v := c.VulnerabilityReporting
+	if v == nil || v.Quay == nil {
+		return
+	}
+	if v.Quay.MaxConcurrentRequests == 0 {
+		v.Quay.MaxConcurrentRequests = DefaultQuayMaxConcurrentRequests
+	}
```
asafbennatan, line 1420: Env parsing keeps any `Atoi` success, including negatives, and this default only replaces `0`. Nothing in this PR reads the field, so it will not fail yet. When the Quay scanner uses it as a concurrency bound, `-1` is a likely panic or a broken limiter.

`0` is the omitempty unset value, so mapping it to 5 is right. A negative is an explicit invalid setting, not unset.

Should we keep this `== 0` default, and treat `n < 0` like a failed `Atoi` (warn and ignore the env override)? Same for YAML: reject a negative rather than folding it into the default.

api/core/v1alpha1/openapi.yaml
```diff
@@ -2339,6 +2339,9 @@ components:
           type: string
           format: date-time
           description: Advisory publish time when known.
+        source:
+          type: string
+          description: The vulnerability backend that produced this finding (e.g. trustify, quay).
```
asafbennatan: `severity` on this schema is a closed string enum. `source` is the same kind of product-owned label (`trustify` / `quay`), but it is an open string, so clients cannot switch on it from the spec.

Would it make sense to declare `enum: [trustify, quay]` here (still optional, so omitted NULL rows stay valid), using the same values as `config.VulnerabilityBackend`? `make generate` would emit a v1alpha1 typed alias the way it does for `VulnerabilitySeverity`. Adding a backend later is an additive OpenAPI change.

```suggestion
        source:
          type: string
          description: The vulnerability backend that produced this finding.
          enum: [trustify, quay]
```

internal/config/config.go
```diff
@@ -807,6 +813,38 @@ type VulnerabilityConfig struct {
...
+// QuayConfig holds Quay Security API connection and authentication details.
+// Quay authenticates with a bearer token rather than OIDC client credentials.
+type QuayConfig struct {
+	// Endpoint is the Quay API base URL (e.g. "https://quay.io").
+	Endpoint string `json:"endpoint,omitempty"`
+	// Token is the bearer token used to authenticate to the Quay Security API.
+	Token api.SecureString `json:"token,omitempty"`
+	// MaxConcurrentRequests bounds concurrent Quay API requests.
+	// Defaults to DefaultQuayMaxConcurrentRequests when unset.
+	MaxConcurrentRequests int `json:"maxConcurrentRequests,omitempty"`
```
asafbennatan, line 888: This struct is enough for this story (`endpoint` / `token` / `maxConcurrentRequests`). Nothing in this PR builds an HTTP client, so CA / skip-verify fields would be unused config.

Self-hosted Quay often uses a private CA, and EDM-5251 constructs a plain `http.Client` with a 30s timeout. Would it make sense to add `caFile` (or a CA bundle) there as the production option, with `skipTlsVerify` only as the lab/air-gap hatch this repo already uses on telemetry forward, image pulls, and repositories? Shipping skip-verify without a CA path tends to get left on.

`quay.io` is fine on system roots, so this is not a blocker here.

Trustify has the same gap today (`TrustifyConfig` is endpoint + auth only). TPA is almost always an operator-run HTTPS service, so for completeness the same `caFile` / `skipTlsVerify` pair should land on that client as well, not only on Quay.

internal/config/config.go
EfratIfergan: Good catch. The env override does use an unchecked cast, so `FLIGHTCTL_VULNERABILITY_REPORTING_BACKEND=typo` would silently land in the config. I've added a `ParseVulnerabilityBackend` helper that validates against the known set (trustify/quay/empty), and used it from both `UnmarshalJSON` and `applyVulnerabilityReportingEnvVarOverrides`. The config type remains as-is (not importing v1alpha1).

internal/config/config.go
EfratIfergan: Agreed. Negative is an explicit invalid value, not unset. I've added `n < 0` to the validation in `applyVulnerabilityReportingQuayEnvVarOverrides` (warn and ignore, same as parse failure), and added a `Validate()` method for `VulnerabilityConfig` (called during `Load()`) that rejects negative `MaxConcurrentRequests` with a clear error.

api/core/v1alpha1/openapi.yaml
EfratIfergan: Makes sense. `source` has the same closed-set semantics as `severity` — we own both ends, and clients benefit from a typed enum they can switch on. I've added `enum: [trustify, quay]` to the OpenAPI spec and run `make generate` to get the v1alpha1 `VulnerabilitySource` type alias. Adding a backend later is additive (just append to the enum).

internal/config/config.go
EfratIfergan: Agreed that production Quay deployments need CA configuration — I'll add `caFile` and `skipTlsVerify` to `QuayConfig` in EDM-5251 when building the HTTP client, following the same pattern as the repo's telemetry/image-pull clients. This story (EDM-5250) is config-only, so those fields would be unused until EDM-5251 lands the client.

The same fields should be added to `TrustifyConfig` for consistency (TPA is also typically HTTPS with a private CA). Should I file a follow-up story for that, or handle it as part of EDM-5251?

asafbennatan: should be handled in this PR since we are doing config changes here

EfratIfergan:
Done — added `caFile` and `skipTlsVerify` to both `QuayConfig` and `TrustifyConfig` in f65d4046. Environment variable overrides follow the same boolean parsing pattern as existing configs (true/1/false/0, with invalid values warned and ignored). Updated Helm values.yaml with the new fields.

(Reply to review comment thread: https://github.com/flightctl/flightctl/pull/3463#discussion_r3869884789)

internal/config/config.go
EfratIfergan: Done - https://github.com/flightctl/flightctl/pull/3463#issuecomment-5436791844

deploy/helm/flightctl/values.yaml.gotmpl
```diff
@@ -565,6 +565,21 @@ vulnerabilityReporting:
...
+    # -- Quay API base URL (e.g. "https://quay.io").
+    endpoint: ""
+    # -- Name of the Kubernetes Secret containing the 'token' key for Quay Security API bearer authentication.
+    secretName: ""
+    # -- Maximum number of concurrent Quay API requests. Defaults to 5 when unset.
+    maxConcurrentRequests: 5
+    # -- Path to a CA bundle for verifying the Quay server certificate. If unset, system roots are used.
+    caFile: ""
+    # -- Skip TLS certificate verification (insecure, for lab/air-gap only). Defaults to false.
+    skipTlsVerify: false
```
asafbennatan, line 586: These values never reach a running process today. Helm ConfigMaps still only write Trustify `endpoint` and `auth`, and Deployments still only inject Trustify `client_id` / `client_secret`. Quadlets have the same gap: `service-config.yaml` is still a Trustify-only example, and the render templates drop `backend`, `quay`, `caFile`, and `skipTlsVerify`.

Files that would need to emit the new fields:

Helm:
- `deploy/helm/flightctl/templates/periodic/flightctl-periodic-config.yaml`
- `deploy/helm/flightctl/templates/imagebuilder-worker/flightctl-imagebuilder-worker-config.yaml`
- `deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml` (Quay token secret, same pattern as Trustify auth)
- `deploy/helm/flightctl/templates/imagebuilder-worker/flightctl-imagebuilder-worker-deployment.yaml`

Quadlets:
- `deploy/podman/service-config.yaml` (commented example)
- `deploy/podman/flightctl-periodic/flightctl-periodic-config/config.yaml.template`
- `deploy/podman/flightctl-imagebuilder-worker/flightctl-imagebuilder-worker-config/config.yaml.template`
- `deploy/podman/flightctl-periodic/flightctl-periodic.container` (`Volume=` for a CA file)
- `deploy/podman/flightctl-imagebuilder-worker/flightctl-imagebuilder-worker.container`

`skipTlsVerify` is a boolean and only needs to be copied into those ConfigMaps/templates. For `caFile`, Helm usually mounts a ConfigMap/Secret then points config at that path (`db.external.tlsConfigMapName`); a bare host path in values will not exist in the container unless something mounts it.

deploy/helm/flightctl/values.schema.json
```diff
@@ -1002,7 +1002,21 @@
...
+          "type": "object",
+          "description": "Quay Security API connection details",
+          "additionalProperties": false,
+          "properties": {
+            "endpoint": { "type": "string", "description": "Quay API base URL (e.g. \"https://quay.io\")" },
+            "secretName": { "type": "string", "description": "Name of the Kubernetes Secret containing the 'token' key for Quay Security API bearer authentication" },
+            "maxConcurrentRequests": { "type": "integer", "description": "Maximum number of concurrent Quay API requests. Defaults to 5 when unset." },
+            "caFile": { "type": "string", "description": "Path to a CA bundle for verifying the Quay server certificate. If unset, system roots are used." },
+            "skipTlsVerify": { "type": "boolean", "description": "Skip TLS certificate verification (insecure, for lab/air-gap only). Defaults to false." }
           }
```
asafbennatan, line 1022: This `quay` object is new, but `vulnerabilityReporting.backend` in the same schema is still `enum: ["", "trustify"]` (and the description still says only Trustify is supported). Helm will reject `--set vulnerabilityReporting.backend=quay`.

Should `quay` be added to that enum (and the description) in this change, so the object you added can actually be selected?

asafbennatan (commented):
See comments below

deploy/helm/flightctl/values.schema.json
EfratIfergan: Good catch — the `quay` object was unreachable because the schema enum rejected it. Fixed in 9c760450: added `quay` to the `vulnerabilityReporting.backend` enum and genericized the descriptions in both `values.schema.json` and `values.yaml.gotmpl` (and the regenerated `values.yaml`/`README.md`). `--set vulnerabilityReporting.backend=quay` now passes schema validation.

deploy/helm/flightctl/values.yaml.gotmpl
EfratIfergan: Wired end-to-end in 9c760450:

- **Periodic ConfigMap** now renders `backend`, Trustify `caFile`/`skipTlsVerify`, and the full `quay` block (endpoint, `maxConcurrentRequests`, `caFile`, `skipTlsVerify`).
- **Periodic Deployment** injects `FLIGHTCTL_VULNERABILITY_REPORTING_QUAY_TOKEN` from the quay secret's `token` key (mirroring the Trustify `client_id`/`client_secret` pattern — the token is never written into the config file).
- **CA bundles**: added a per-backend `caCertConfigMapName` that mounts a ConfigMap (key `ca-cert.pem`) and points `caFile` at the mount path, mirroring `db.external.tlsConfigMapName`. A bare `caFile` path is still honored for operator-managed mounts.
- **Quadlets**: periodic/worker config templates, the `service-config.yaml` example, and commented CA `Volume=` examples in the `.container` units all got the same fields.

One deliberate scope call: the **imagebuilder-worker only uploads SBOMs to Trustify and gates on `backend`** — it never calls the Quay API — so it gets `backend` + Trustify TLS fields and the Trustify CA mount, but not the Quay block or `QUAY_TOKEN`. Wiring those into the worker would be unused config. Happy to mirror them there too if you'd prefer symmetry.

internal/trustify/v2/client.go
```diff
@@ -339,10 +343,48 @@ func stripSHA256Prefix(s string) string {
...
+	if cfg.CAFile != "" {
+		caPEM, err := os.ReadFile(cfg.CAFile)
+		if err != nil {
+			return nil, fmt.Errorf("reading Trustify CA file %q: %w", cfg.CAFile, err)
+		}
+		pool := x509.NewCertPool()
+		if !pool.AppendCertsFromPEM(caPEM) {
+			return nil, fmt.Errorf("no valid certificates found in Trustify CA file %q", cfg.CAFile)
+		}
+		tlsConfig.RootCAs = pool
```
asafbennatan, line 369: `x509.NewCertPool()` replaces the system roots. This same client is also used for OIDC discovery and the token request (`oauth2.HTTPClient`). The quadlet example already has Trustify on one host and Keycloak on another, so a private Trustify CA would make discovery to a public IdP fail.

`getTlsConfig` in `internal/auth/auth.go` appends extra PEMs to `SystemCertPool` for that reason. Would it make sense to do the same here, and leave `RootCAs` nil when `CAFile` is empty (already the case)?

```suggestion
	if cfg.CAFile != "" {
		caPEM, err := os.ReadFile(cfg.CAFile)
		if err != nil {
			return nil, fmt.Errorf("reading Trustify CA file %q: %w", cfg.CAFile, err)
		}
		pool, err := x509.SystemCertPool()
		if err != nil || pool == nil {
			pool = x509.NewCertPool()
		}
		if !pool.AppendCertsFromPEM(caPEM) {
			return nil, fmt.Errorf("no valid certificates found in Trustify CA file %q", cfg.CAFile)
		}
		tlsConfig.RootCAs = pool
	}
```

asafbennatan (commented):
See comments below

deploy/podman/flightctl-periodic/flightctl-periodic.container
```diff
@@ -18,6 +18,10 @@ Volume=/etc/flightctl/flightctl-periodic/config.yaml:/root/.flightctl/config.yam
 Volume=/etc/flightctl/ssh:/etc/flightctl/ssh:ro,z
 Volume=/etc/flightctl/encryption:/root/.flightctl/encryption:ro,z
 Volume=/etc/flightctl/pki/db:/root/.flightctl/certs/db:ro,z
+# Optional: Mount a CA bundle for verifying the vulnerability backend (Trustify/Quay) server
+# certificate. Uncomment and point the source at your CA file, then set the matching
+# vulnerabilityReporting.<backend>.caFile path in service-config.yaml to the target path.
+# Volume=/etc/flightctl/certs/vulnerability-ca.pem:/etc/flightctl/certs/vulnerability-ca.pem:ro,z
```
asafbennatan: The commented `caFile` paths in `service-config.yaml` are `trustify-ca.pem` and `quay-ca.pem`, but this Volume example is `vulnerability-ca.pem`. Uncommenting both as written leaves the client looking at a path that is not mounted.

Quadlets already pin CAs as a single file at a constant path: `flightctl-api` does `Volume=/etc/flightctl/pki/ca-bundle.crt:/root/.flightctl/certs/ca-bundle.crt` (same for telemetry-gateway, remote-access, imagebuilder). Directory mounts in this tree are for a *set* of files (`pki/db`, `encryption`, `tpm-cas`). `caFile` is one PEM per backend, so it is closer to `ca-bundle.crt`.

Would it make sense to keep a single-file `Volume=` here, with the same filenames as `service-config.yaml`? Two backends would be two Volume lines, like API's `server.crt` plus `ca-bundle.crt`. The imagebuilder-worker unit has the same example.

```suggestion
# Optional: Mount a CA bundle for verifying the vulnerability backend (Trustify/Quay) server
# certificate. Uncomment and match caFile in service-config.yaml to the target path
# (same single-file pattern as /etc/flightctl/pki/ca-bundle.crt on flightctl-api).
# Volume=/etc/flightctl/certs/trustify-ca.pem:/etc/flightctl/certs/trustify-ca.pem:ro,z
# Volume=/etc/flightctl/certs/quay-ca.pem:/etc/flightctl/certs/quay-ca.pem:ro,z
```

asafbennatan (approved):
LGTM

PR #3466, opened 2026-08-25
EDM-5251: Quay Security API client — fetch and filter

EfratIfergan:
## EDM-5251: Quay Security API client — fetch and filter

**Jira:** https://issues.redhat.com/browse/EDM-5251  
**Story type:** [DEV]  
**Feature:** EDM-4869 — Quay as a Vulnerability Backend for RHEM

### Summary

Implements the fetch/filter/deserialize layer for the Quay vulnerability backend. Creates a new `internal/quay/` package with an HTTP client that queries the Quay Security API for deployed images, filters by registry hostname, and deserializes typed vulnerability reports. This story delivers the foundation for EDM-5252 (Scanner implementation and finding conversion).

### Changes

**New package: `internal/quay/`**
- `types.go` — Typed response structs matching Quay Security API schema (`Response`, `Data`, `Layer`, `Feature`, `Vulnerability`, `Metadata` with NVD CVSS scoring)
- `client.go` — HTTP client with:
  - `NewClient(cfg *config.QuayConfig, log logrus.FieldLogger)` — constructs client from config, 30s timeout, returns `(nil, nil)` when config is nil (backend disabled)
  - `FetchImageSecurity(ctx, image vulnerability.ImageRef)` — per-image fetch with registry filtering, status validation, and skip logging
  - Skip conditions: missing image reference, missing digest, non-matching registry, unparseable reference, HTTP 404, non-`scanned` statuses (queued/pending/unsupported/failed)
  - Error handling: `(report, nil)` for scanned images, `(nil, nil)` for skips (logged at appropriate levels), `(nil, error)` for transport/decode failures
- `client_test.go` — 13 test cases using `httptest.Server`, covers all skip paths, success path with header/URL assertions, error cases

### Testing

- **Unit tests:** 13 test cases in `internal/quay/client_test.go`
  - Success path with Bearer auth, correct URL/query, full deserialization validation
  - All skip conditions (missing ref/digest, registry filter, non-scanned statuses, 404)
  - Error paths (transport failures, malformed JSON, unparseable references)
  - Edge cases (nil config, empty endpoint, nil logger defaults, various HTTP status codes)
- **Integration tests:** N/A — self-contained HTTP client, no store/API integration
- **Coverage:** 94.0% of statements (all behavioral contracts tested through public APIs)
- **Race detector:** Clean (`go test -race`)
- **Lint:** PASS (0 issues)

### Test Plan

Covers 5 story-relevant scenarios from the feature testplan:
- TC-FR3-01: Successful fetch deserializes into typed `Response` ✅
- TC-FR5-01: Scan status queued/pending/unsupported/failed → no report, logged ✅
- TC-FR5-02: HTTP 404 → image skipped, sync continues ✅
- TC-FR5-03: Missing/empty image reference → skipped with reason `"missing_image_reference"` ✅
- TC-FR4-02: Registry filtering → non-matching host skipped, no request issued ✅

### Acceptance Criteria

- [x] AC-1: GET `.../security?vulnerabilities=true` with `Authorization: Bearer` + 30s timeout
- [x] AC-2: Parse image ref into `{namespace}/{repo}` via `github.com/containers/image/v5`
- [x] AC-3: Skip missing/empty image reference (debug, `"missing_image_reference"`)
- [x] AC-4: Only query images whose registry host matches the endpoint; skip others
- [x] AC-5: Non-`scanned` status (and 404/non-200) → no report, logged appropriately
- [x] AC-6: Deserialize success into typed `Response`/`Data`/`Layer`/`Feature`/`Vulnerability`/`Metadata`

### Dependencies

**Depends on:** EDM-5250 (Quay configuration) — provides `config.QuayConfig` type  
**Required by:** EDM-5252 (Quay Scanner implementation)

### Notes

- **Branch stack:** This PR targets `EDM-5250-quay-configuration` (not `main`) as part of the Quay backend feature branch stack. The change will reach `main` when the stack merges.
- **Scope boundary:** This story delivers fetch/filter/deserialize ONLY. Scanner implementation, finding conversion, and concurrency/retry are deferred to subsequent stories (EDM-5252, EDM-5254).
- **Observability:** Skip reasons are logged as structured fields (`reason`, `status`, `digest`, `image`) for easier querying and alerting in production.

### DIFF
--- a/deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml
+++ b/deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml
@@ -76,7 +76,7 @@ spec:
                   name: {{ $auth.secretName }}
                   key: client_secret
             {{- end }}
-            {{- if and (default false $vuln.enabled) $quay.secretName }}
+            {{- if and (default false $vuln.enabled) (eq $effectiveBackend "quay") $quay.secretName }}
             - name: FLIGHTCTL_VULNERABILITY_REPORTING_QUAY_TOKEN
               valueFrom:
                 secretKeyRef:
--- a/deploy/podman/flightctl-periodic/flightctl-periodic.container
+++ b/deploy/podman/flightctl-periodic/flightctl-periodic.container
@@ -18,7 +18,7 @@ Secret=flightctl-kv-password,type=env,target=KV_PASSWORD
 #   /etc/containers/systemd/flightctl-periodic.container.d/quay-token.conf
 #   [Container]
 #   Secret=flightctl-quay-token,type=env,target=FLIGHTCTL_VULNERABILITY_REPORTING_QUAY_TOKEN
-# Create the secret with: podman secret create flightctl-quay-token - <<< "your-token"
+# Create the secret with: printf %s "your-token" | podman secret create flightctl-quay-token -
 Environment=DB_USER=flightctl_app
 Volume=/etc/flightctl/flightctl-periodic/config.yaml:/root/.flightctl/config.yaml:ro,z
 Volume=/etc/flightctl/ssh:/etc/flightctl/ssh:ro,z
--- a/deploy/podman/flightctl-periodic/flightctl-periodic.container
+++ b/deploy/podman/flightctl-periodic/flightctl-periodic.container
@@ -13,6 +13,12 @@ AddHost=%H:host-gateway
 Environment=HOME=/root
 Secret=flightctl-postgresql-user-password,type=env,target=DB_PASSWORD
 Secret=flightctl-kv-password,type=env,target=KV_PASSWORD
+# When backend: quay is selected, create a Podman secret and add this Quadlet
+# drop-in so deployments that do not use Quay do not require the secret:
+#   /etc/containers/systemd/flightctl-periodic.container.d/quay-token.conf
+#   [Container]
+#   Secret=flightctl-quay-token,type=env,target=FLIGHTCTL_VULNERABILITY_REPORTING_QUAY_TOKEN
+# Create the secret with: podman secret create flightctl-quay-token - <<< "your-token"
 Environment=DB_USER=flightctl_app
 Volume=/etc/flightctl/flightctl-periodic/config.yaml:/root/.flightctl/config.yaml:ro,z
 Volume=/etc/flightctl/ssh:/etc/flightctl/ssh:ro,z
--- a/deploy/podman/service-config.yaml
+++ b/deploy/podman/service-config.yaml
@@ -170,8 +170,8 @@ prometheus:
 #     # Trustify API base URL (used by flightctl-periodic only).
 #     # NOTE: Do not include the API path; the client appends /api/v2/vulnerability/analyze automatically.
 #     endpoint: https://trustify.example.com
-#     # Optional: CA certificate file for Trustify TLS (mount at /etc/ssl/trustify-ca.pem in container).
-#     caFile: /etc/ssl/trustify-ca.pem
+#     # Optional: CA certificate file for Trustify TLS (mount at /etc/flightctl/certs/trustify-ca.pem in container).
+#     caFile: /etc/flightctl/certs/trustify-ca.pem
 #     # Optional: Skip TLS verification (not recommended for production).
 #     # skipTlsVerify: false
 #     auth:
@@ -181,10 +181,6 @@ prometheus:
 #       oidcIssuerUrl: https://keycloak.example.com/realms/trustify
 #       # client_id and client_secret are supplied via FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_CLIENT_ID
 #       # and FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_CLIENT_SECRET environment variables.
-#     # Path to a CA bundle for verifying the Trustify server certificate. If unset, system roots are used.
-#     # caFile: /etc/flightctl/certs/trustify-ca.pem
-#     # Skip TLS certificate verification (insecure, for lab/air-gap only). Defaults to false.
-#     # skipTlsVerify: false
 #   # Quay backend (used by flightctl-periodic only). Set backend: quay above to select it.
 #   quay:
 #     # Quay API base URL (e.g. "https://quay.io").
--- a/internal/migration/backfill_vulnerability_source.go
+++ b/internal/migration/backfill_vulnerability_source.go
@@ -11,31 +11,141 @@ import (
 	"gorm.io/gorm/clause"
 )
 
-const backfillVulnerabilitySourceMigrationKey = "backfill_vulnerability_source_v1"
+const (
+	backfillVulnerabilitySourceMigrationKey = "backfill_vulnerability_source_v1"
+	vulnerabilitySourceBackfillBatchSize    = 1000
+
+	vulnerabilitySourceBackfillLockNamespace int32 = 0x56534d47 // VSMG
+	vulnerabilitySourceBackfillLockID        int32 = 1
+)
+
+type vulnerabilityFindingCursor struct {
+	ImageDigest string
+	CveID       string
+}
 
 // backfillVulnerabilitySource sets source='trustify' on pre-existing
 // vulnerability_findings rows whose source is NULL. Trustify was the only
 // backend before the source column existed, so those rows originate from it.
-// The migration key acts as a distributed once-only guard: concurrent replicas
-// race on the unique primary key; the loser sees 0 RowsAffected and skips.
-func backfillVulnerabilitySource(ctx context.Context, tx *gorm.DB) error {
-	return tx.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		result := tx.Clauses(clause.OnConflict{DoNothing: true}).
-			Create(&model.SchemaMigration{Key: backfillVulnerabilitySourceMigrationKey, AppliedAt: time.Now()})
-		if result.Error != nil {
-			return fmt.Errorf("recording vulnerability source migration: %w", result.Error)
+// A session-scoped advisory lock ensures only one replica backfills at a time.
+// The completion marker is recorded only after all batches commit successfully.
+func backfillVulnerabilitySource(ctx context.Context, db *gorm.DB) error {
+	return db.WithContext(ctx).Connection(func(conn *gorm.DB) (err error) {
+		var locked bool
+		if err := conn.Raw(
+			`SELECT pg_try_advisory_lock(?, ?)`,
+			vulnerabilitySourceBackfillLockNamespace,
+			vulnerabilitySourceBackfillLockID,
+		).Scan(&locked).Error; err != nil {
+			return fmt.Errorf("acquiring vulnerability source backfill lock: %w", err)
 		}
-		if result.RowsAffected == 0 {
-			// Another replica already ran this migration.
+		if !locked {
+			return nil
+		}
+		defer func() {
+			if unlockErr := conn.Exec(
+				`SELECT pg_advisory_unlock(?, ?)`,
+				vulnerabilitySourceBackfillLockNamespace,
+				vulnerabilitySourceBackfillLockID,
+			).Error; unlockErr != nil && err == nil {
+				err = fmt.Errorf("releasing vulnerability source backfill lock: %w", unlockErr)
+			}
+		}()
+
+		var completed int64
+		if err := conn.Model(&model.SchemaMigration{}).
+			Where("key = ?", backfillVulnerabilitySourceMigrationKey).
+			Count(&completed).Error; err != nil {
+			return fmt.Errorf("checking vulnerability source migration: %w", err)
+		}
+		if completed != 0 {
 			return nil
 		}
 
-		if err := tx.Exec(
-			`UPDATE vulnerability_findings SET source = ? WHERE source IS NULL`,
-			string(config.VulnerabilityBackendTrustify),
-		).Error; err != nil {
-			return fmt.Errorf("backfilling vulnerability finding sources: %w", err)
+		var cursor *vulnerabilityFindingCursor
+		for {
+			updated, nextCursor, err := backfillVulnerabilitySourceBatch(ctx, conn, cursor)
+			if err != nil {
+				return err
+			}
+			if updated == 0 {
+				break
+			}
+			cursor = &nextCursor
 		}
+
+		if err := conn.Transaction(func(tx *gorm.DB) error {
+			result := tx.Clauses(clause.OnConflict{DoNothing: true}).
+				Create(&model.SchemaMigration{Key: backfillVulnerabilitySourceMigrationKey, AppliedAt: time.Now()})
+			if result.Error != nil {
+				return fmt.Errorf("recording vulnerability source migration: %w", result.Error)
+			}
+			return nil
+		}); err != nil {
+			return err
+		}
+
 		return nil
 	})
+}
+
+func backfillVulnerabilitySourceBatch(
+	ctx context.Context,
+	db *gorm.DB,
+	cursor *vulnerabilityFindingCursor,
+) (int, vulnerabilityFindingCursor, error) {
+	query := `
+		WITH batch AS (
+			SELECT image_digest, cve_id
+			FROM vulnerability_findings
+			WHERE source IS NULL
+			ORDER BY image_digest, cve_id
+			LIMIT ?
+			FOR UPDATE
+		)
+		UPDATE vulnerability_findings AS finding
+		SET source = ?
+		FROM batch
+		WHERE finding.image_digest = batch.image_digest
+		  AND finding.cve_id = batch.cve_id
+		RETURNING finding.image_digest, finding.cve_id`
+	args := []any{vulnerabilitySourceBackfillBatchSize, string(config.VulnerabilityBackendTrustify)}
+	if cursor != nil {
+		query = `
+			WITH batch AS (
+				SELECT image_digest, cve_id
+				FROM vulnerability_findings
+				WHERE source IS NULL
+				  AND (image_digest, cve_id) > (?, ?)
+				ORDER BY image_digest, cve_id
+				LIMIT ?
+				FOR UPDATE
+			)
+			UPDATE vulnerability_findings AS finding
+			SET source = ?
+			FROM batch
+			WHERE finding.image_digest = batch.image_digest
+			  AND finding.cve_id = batch.cve_id
+			RETURNING finding.image_digest, finding.cve_id`
+		args = []any{cursor.ImageDigest, cursor.CveID, vulnerabilitySourceBackfillBatchSize, string(config.VulnerabilityBackendTrustify)}
+	}
+
+	var updated []vulnerabilityFindingCursor
+	if err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		return tx.Raw(query, args...).Scan(&updated).Error
+	}); err != nil {
+		return 0, vulnerabilityFindingCursor{}, fmt.Errorf("backfilling vulnerability finding source batch: %w", err)
+	}
+	if len(updated) == 0 {
+		return 0, vulnerabilityFindingCursor{}, nil
+	}
+
+	nextCursor := updated[0]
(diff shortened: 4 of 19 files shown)


asafbennatan (commented):
See comments below

internal/quay/client.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		c.log.WithFields(logrus.Fields{"image": image.Image, "reason": reasonUnparseableReference}).
+			WithError(err).Debug("skipping image")
+		return nil, nil
+	}
+
+	if host != c.registryHost {
+		c.log.WithFields(logrus.Fields{"digest": image.Digest, "host": host, "reason": reasonNotOnConfiguredRegistry}).
+			Debug("skipping image")
+		return nil, nil
```
asafbennatan, line 102: Registry matching is a raw `host != c.registryHost` string compare. The image side already goes through `reference.ParseNormalizedNamed` / `reference.Domain`, while the configured side stores `url.Parse(...).Host` as written (case and explicit port included). `https://Quay.io` or `https://quay.example.com:443` then fails to match `quay.io/...` / `quay.example.com/...`, skips every image at Debug, and never queries Quay.

Should we identify both sides with the parsers already in this file instead of comparing (or folding/trimming) host strings? For the endpoint, `url.URL` already splits host and port (`Hostname()` / `Port()`). For the image, `reference.Domain` is the Docker-normalized registry. Putting the endpoint host through the same `ParseNormalizedNamed` path — or comparing those parsed identities — would keep case and default-port handling in `net/url` and `github.com/containers/image/v5/docker/reference` rather than ad-hoc string manipulation. The current tests use the same httptest `host:port` on both sides, so this gap would not show up without mixed-case / default-port cases.

internal/quay/client.go
```diff
@@ -0,0 +1,193 @@
...
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("querying quay security api: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		c.logNonOK(image.Digest, resp.StatusCode)
+		return nil, nil
+	}
```
asafbennatan, line 127: Every non-200, including 401, 403, 429, and 5xx, becomes `(nil, nil)`. The caller cannot tell a missing image from a misconfigured token or an outage, so a whole sync can look like “no vulnerabilities” while the backend is failing. 404 as a skip still matches the story.

EDM-5251 lists non-200 as skip-and-log. Should this method still return a contextual error for auth, rate-limit, and server failures so the sync can fail loudly, and keep 404 as the skip?

```suggestion
	if resp.StatusCode != http.StatusOK {
		c.logNonOK(image.Digest, resp.StatusCode)
		if resp.StatusCode == http.StatusNotFound {
			return nil, nil
		}
		return nil, fmt.Errorf("quay security api returned %d", resp.StatusCode)
	}
```

internal/quay/client.go
```diff
@@ -0,0 +1,193 @@
...
+		// beyond the known set (queued, pending, unsupported, failed).
+		c.log.WithFields(logrus.Fields{
+			"digest": image.Digest,
+			"status": report.Status,
+			"reason": "scan_" + report.Status,
+		}).Info("skipping image: scan not complete")
+		return nil, nil
+	}
+
+	return &report, nil
```
asafbennatan, line 156: With pointer fields, `{"status":"scanned","data":null}` and a Layer with empty `Features` are distinct. They are not equivalent on Quay’s side: `_security_info` sets `data` to `null` for queued/failed/unsupported (and similar), and `scanned` is `ScanLookupStatus.SUCCESS` via `for_data(SecurityInformation(Layer(...)))`, so a successful scan always includes a `Layer` (empty `Features` is the zero-vuln case). See [endpoints/api/secscan.py](https://github.com/quay/quay/blob/master/endpoints/api/secscan.py) and `load_security_information` in [secscan_v4_model.py](https://github.com/quay/quay/blob/master/data/secscan_model/secscan_v4_model.py).

Should `scanned` with a nil `Data`/`Layer` be treated as an unexpected payload (log/skip or error), rather than as an empty vulnerability list for the converter?

internal/quay/client.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &Client{
+		endpoint:     base,
+		registryHost: host,
+		token:        cfg.Token.Value(),
+		httpClient:   &http.Client{Timeout: defaultHTTPTimeout},
+		log:          log,
+	}, nil
```
asafbennatan: `NewClient` always constructs `&http.Client{Timeout: 30s}` with the default transport. Callers talking to a private Quay with a custom CA, or a lab instance that needs `InsecureSkipVerify`, have no way to set that — TLS policy is not on `QuayConfig` and is not injectable. Logger already has a nil default; the HTTP client does not.

Should `NewClient` take an `*http.Client` and use the 30s client only when it is nil? The caller can then put custom `RootCAs`, skip-verify, timeouts, or a test transport on that client, without this package growing CA/skip-TLS knobs of its own.

```suggestion
func NewClient(cfg *config.QuayConfig, log logrus.FieldLogger, httpClient *http.Client) (*Client, error) {
	if cfg == nil {
		return nil, nil
	}
	if log == nil {
		log = logrus.StandardLogger()
	}
	if httpClient == nil {
		httpClient = &http.Client{Timeout: defaultHTTPTimeout}
	}
	base, host, err := parseEndpoint(cfg.Endpoint)
	if err != nil {
		return nil, err
	}
	return &Client{
		endpoint:     base,
		registryHost: host,
		token:        cfg.Token.Value(),
		httpClient:   httpClient,
		log:          log,
	}, nil
}
```

internal/quay/client.go
EfratIfergan: Good catch! The case mismatch (Quay.io vs quay.io) and explicit default port (quay.io:443 → quay.io) would cause false registry mismatches and skip all images silently at Debug level.

Fixed in 60c97af3 by normalizing both sides consistently: `parseEndpoint` now lowercases the hostname and strips default ports (443 for https, 80 for http) while preserving non-default ports, matching the behavior of `reference.Domain`. Added test coverage for mixed-case endpoints and explicit default ports to prevent regressions.

internal/quay/client.go
EfratIfergan: Valid concern about observability — a failing backend can look like "no vulnerabilities" if all images skip. However, this is the fetch layer (EDM-5251); error isolation and retry are in EDM-5254. Returning errors here for 401/5xx would abort the entire sync on the first failed image, not just that image.

The current approach logs at appropriate levels (401→Error, 403/5xx→Warn) so operators see backend failures. A better fix might be a sync-level health check (pre-flight request or failure-count threshold) that fails the sync when the backend is unreachable/misconfigured, separate from per-image skips. Would that address the concern without mixing per-image error handling with sync-level health?

I'll escalate this to EDM-5252 (sync integration) for a health check or failure threshold.

internal/quay/client.go
EfratIfergan: Agreed — Quay's contract is `status="scanned"` → non-nil `Layer`. Treating nil as "zero vulnerabilities" would mask unexpected responses.

Fixed in 60c97af3: added a defensive guard after the status check. If `status == "scanned"` but `Data` or `Data.Layer` is nil, log at Warn level with reason `"malformed_scanned_response"` and return `(nil, nil)` (skip). This prevents the converter from seeing unexpected payloads and makes malformed responses visible in logs.

Added test coverage in `TestFetchImageSecurity_ScannedWithNilData`.

internal/quay/client.go
EfratIfergan: Good point — custom CA/TLS config is a real use case (private Quay, lab environments).

Fixed in 60c97af3: added `httpClient *http.Client` as a parameter to `NewClient`, defaulting to `&http.Client{Timeout: 30s}` when nil. This matches the `log` parameter pattern and lets callers inject custom TLS config without adding CA/skip-verify fields to `QuayConfig`.

Added test coverage in `TestNewClient_CustomHTTPClient`.

internal/quay/client.go
asafbennatan: logging stuff is not enough , it should be the callers decision how to handle these sort of errors exactly because this is just the fetch layer

internal/quay/client.go
```diff
@@ -0,0 +1,225 @@
...
+	port := u.Port()
+
+	// Strip default ports
+	if (u.Scheme == "https" && port == "443") || (u.Scheme == "http" && port == "80") {
+		host = hostname
+	} else if port != "" {
+		host = hostname + ":" + port
+	} else {
+		host = hostname
+	}
```
asafbennatan: This only canonicalizes the **configured** host (`strings.ToLower`, then drop `:443`/`:80` by hand). The image side still uses `reference.Domain` as-is. `ParseNormalizedNamed("Quay.IO/org/repo")` keeps Domain `Quay.IO`, and `"quay.io:443/org/repo"` keeps Domain `quay.io:443`, so those images still fail `host != c.registryHost` after a `https://quay.io` / `https://quay.io:443` config. The new tests assert `c.registryHost` only; they do not `FetchImageSecurity` with a mixed-case or `:443` image ref.

`url.Hostname()` / `url.Port()` are the right tools for splitting the endpoint. Should both sides be reduced with the same identity (for example run the image Domain, or `net.JoinHostPort` of hostname+non-default port, through one helper) so we are not lowercasing and concatenating on one side only? The existing registry-filter thread is still open on the `==` compare.

asafbennatan (commented):
See comments below

internal/quay/client.go
EfratIfergan: You're right — the fetch layer shouldn't decide whether auth/server failures stop the sync. The caller (sync layer) needs that control.

Fixed in ee6c8ca5: non-404 failures (401, 403, 5xx) now return errors instead of `(nil, nil)`, keeping 404 as a skip. This lets the caller decide whether to fail the sync or apply retry/threshold policies. Logging still happens at appropriate levels (401→Error, 403/5xx→Warn) before returning the error.

internal/quay/client.go
EfratIfergan: Good catch — I only normalized the endpoint side, not the image side. `reference.Domain` doesn't lowercase or strip ports, so `Quay.IO/repo` or `quay.io:443/repo` still wouldn't match `quay.io`.

Fixed in ee6c8ca5: extracted `normalizeHost` helper that both `parseEndpoint` and `parseImageReference` now use. The image host from `reference.Domain` is normalized (lowercase + strip default HTTPS port 443) to match the endpoint normalization. Both sides now use identical normalization logic.

Added `TestFetchImageSecurity_ImageReferenceNormalization` to verify mixed-case (`Quay.IO`), explicit default port (`quay.io:443`), and non-default port handling on the image reference side.

internal/quay/client.go
```diff
@@ -0,0 +1,240 @@
...
+
+	hostname = strings.ToLower(hostname)
+
+	// Strip default ports
+	if (scheme == "https" && port == "443") || (scheme == "http" && port == "80") {
+		return hostname
+	} else if port != "" {
+		return hostname + ":" + port
+	}
+	return hostname
```
asafbennatan: Applying the same helper on the image Domain closes the one-sided compare. Splitting on the last `:` is still hand-rolled host/port parsing, and it mishandles IPv6: `url.Parse("https://[2001:db8::1]/").Host` is `[2001:db8::1]`, so `LastIndex(":")` cuts inside the address.

`parseEndpoint` already has a `*url.URL` — `Hostname()` / `Port()` split that correctly. For `reference.Domain`, `net.SplitHostPort` is the stdlib splitter (a missing port is an error, not a last-colon guess), and `net.JoinHostPort` rebuilds it. Should `normalizeHost` use those instead of `LastIndex` / string concat, now that both sides share this function?

asafbennatan (commented):
See comments below

internal/quay/client.go
EfratIfergan: Good catch — the `LastIndex(":")` split corrupts IPv6 literals like `[2001:db8::1]:5000`.

Fixed in 6180c856 by moving to the stdlib splitters:
- `normalizeHost` now takes pre-split `hostname`/`port` components and rebuilds with `net.JoinHostPort`, so IPv6 hosts are re-bracketed correctly.
- `parseEndpoint` passes `url.Hostname()` / `url.Port()` — the `*url.URL` already splits host/port (and IPv6) correctly.
- `parseImageReference` splits `reference.Domain(named)` with `net.SplitHostPort`, falling back to the bare host when no port is present (`net.SplitHostPort` treats a missing port as an error rather than guessing).

Note: `reference.ParseNormalizedNamed` rejects IPv6 image references outright (`invalid reference format`), so the image side can't actually produce an IPv6 domain today — the `net.SplitHostPort` path there is defensive and no longer relies on a last-colon guess. Added IPv6-with-port coverage for `parseEndpoint` and a direct test for the `splitHostPort`/`normalizeHost` helpers (`TestNormalizeHostIPv6`).

asafbennatan (dismissed):
LGTM

internal/store/model/vulnerability_finding.go
asafbennatan: lets keep this consistent as text like the others ? 
also string is not nullable string* is (this contradicts the comment)- which should be removed from here 

additionally please follow the pattern Severity and status follow - the go type for this field is a specific enum (VulnerabilitySeverity , VulnerabilityStatus) not a string

internal/tasks/vulnerability_sync.go
asafbennatan: pass this as enum as far as possible , convert to string only at the last possible point

internal/vulnerability/registry.go
asafbennatan: registry key should be the enum - not a string

internal/store/model/vulnerability_finding.go
EfratIfergan: Updated in df4343fd: `VulnerabilityFinding.Source` now uses `config.VulnerabilityBackend`, is stored as `text`, and the inaccurate nullable comment was removed.

internal/tasks/vulnerability_sync.go
EfratIfergan: Updated in df4343fd: `VulnerabilitySync` and `toModelFinding` now retain `config.VulnerabilityBackend` end-to-end, avoiding intermediate `string` conversions.

internal/vulnerability/registry.go
EfratIfergan: Updated in df4343fd: the scanner registry and `Register` now use `config.VulnerabilityBackend` keys directly; adapters and tests use the named constants.

deploy/helm/flightctl/templates/periodic/flightctl-periodic-deployment.yaml
```diff
@@ -74,6 +76,13 @@ spec:
                   name: {{ $auth.secretName }}
                   key: client_secret
             {{- end }}
+            {{- if and (default false $vuln.enabled) $quay.secretName }}
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in f074c823. The Quay token is now injected only when the effective backend is `quay`, so a Trustify deployment cannot receive Quay credentials.

deploy/podman/service-config.yaml
```diff
@@ -162,19 +162,40 @@ prometheus:
...
 #       # Authentication mode. Allowed values: 'client-credentials', 'none'.
 #       mode: none
 #       # OIDC issuer URL (required when mode is 'client-credentials').
 #       oidcIssuerUrl: https://keycloak.example.com/realms/trustify
 #       # client_id and client_secret are supplied via FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_CLIENT_ID
 #       # and FLIGHTCTL_VULNERABILITY_REPORTING_TRUSTIFY_CLIENT_SECRET environment variables.
+#     # Path to a CA bundle for verifying the Trustify server certificate. If unset, system roots are used.
+#     # caFile: /etc/flightctl/certs/trustify-ca.pem
+#     # Skip TLS certificate verification (insecure, for lab/air-gap only). Defaults to false.
+#     # skipTlsVerify: false
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in f074c823. The sample now has one Trustify `caFile` and one `skipTlsVerify` setting, using `/etc/flightctl/certs/trustify-ca.pem` to match the documented Quadlet mount.

internal/migration/backfill_vulnerability_source.go
```diff
@@ -0,0 +1,37 @@
...
+// vulnerability_findings rows whose source is NULL. Trustify was the only
+// backend before the source column existed, so those rows originate from it.
+// The migration key acts as a distributed once-only guard: concurrent replicas
+// race on the unique primary key; the loser sees 0 RowsAffected and skips.
+func backfillVulnerabilitySource(ctx context.Context, tx *gorm.DB) error {
+	return tx.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		result := tx.Clauses(clause.OnConflict{DoNothing: true}).
+			Create(&model.SchemaMigration{Key: backfillVulnerabilitySourceMigrationKey, AppliedAt: time.Now()})
+		if result.Error != nil {
+			return result.Error
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in f074c823. Both database operations now wrap failures with operation-specific context while preserving the original error with `%w`.

test/integration/store/vulnerability_finding_test.go
```diff
@@ -621,6 +648,7 @@ var _ = Describe("VulnerabilityFinding Store Integration Tests", func() {
 			Expect(row.PublishedAt).ToNot(BeNil())
 			Expect(row.PublishedAt.UTC()).To(Equal(published))
 			Expect(row.FirstSeenAt).ToNot(BeZero())
+			Expect(row.Source).To(Equal(string(config.VulnerabilityBackendTrustify)))
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in f074c823. The integration assertion now compares `Source` directly with `config.VulnerabilityBackendTrustify`.

EfratIfergan:
For the Podman Quay-token finding: fixed in f074c823 with documented Quadlet drop-in instructions. The drop-in injects `FLIGHTCTL_VULNERABILITY_REPORTING_QUAY_TOKEN` from `flightctl-quay-token` when Quay is configured. Keeping it optional is intentional: an unconditional `Secret=` in the shipped periodic unit would make the secret mandatory for deployments that do not use Quay.

deploy/podman/flightctl-periodic/flightctl-periodic.container
```diff
@@ -13,6 +13,12 @@ AddHost=%H:host-gateway
 Environment=HOME=/root
 Secret=flightctl-postgresql-user-password,type=env,target=DB_PASSWORD
 Secret=flightctl-kv-password,type=env,target=KV_PASSWORD
+# When backend: quay is selected, create a Podman secret and add this Quadlet
+# drop-in so deployments that do not use Quay do not require the secret:
+#   /etc/containers/systemd/flightctl-periodic.container.d/quay-token.conf
+#   [Container]
+#   Secret=flightctl-quay-token,type=env,target=FLIGHTCTL_VULNERABILITY_REPORTING_QUAY_TOKEN
+# Create the secret with: podman secret create flightctl-quay-token - <<< "your-token"
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in 9fa2416a. The documented command now uses printf %s, so the Podman secret receives the token without a trailing newline.

internal/migration/backfill_vulnerability_source.go
```diff
@@ -22,16 +23,19 @@ func backfillVulnerabilitySource(ctx context.Context, tx *gorm.DB) error {
...
 		}
 		if result.RowsAffected == 0 {
 			// Another replica already ran this migration.
 			return nil
 		}
 
-		return tx.Exec(
+		if err := tx.Exec(
 			`UPDATE vulnerability_findings SET source = ? WHERE source IS NULL`,
 			string(config.VulnerabilityBackendTrustify),
```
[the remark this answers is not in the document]
EfratIfergan: Fixed in 9fa2416a. The backfill now runs after schema migrations under a session-scoped advisory lock, commits 1,000-row batches independently, and writes its completion marker only after all batches succeed. An interrupted run safely retries remaining NULL rows.

asafbennatan (approved):
LGTM

PR #3467, opened 2026-08-25
EDM-4503: patch CVE-2026-27145 (crypto/x509 DoS) via Go toolchain bump to go1.26.5

amir-yogev-gh:
## CVE Fix — CVE-2026-27145 (GO-2026-5037): crypto/x509 DoS via DNS SAN parsing

Fixes the `crypto/x509` denial-of-service vulnerability (`(*x509.Certificate).VerifyHostname`
re-parses the hostname per DNS SAN entry, scaling quadratically with SAN count).
This is a Go standard-library flaw, so the fix is a **Go toolchain bump**.

Jira: EDM-4503

### Vulnerabilities Addressed

| CVE ID | Severity | Package | Old Version | New Version |
|--------|----------|---------|-------------|-------------|
| CVE-2026-27145 (GO-2026-5037) | Important (CVSS 7.5) | crypto/x509 (Go stdlib) | go1.25.9 | go1.26.5 |

### Strategy Justification

**CVE-2026-27145 — crypto/x509 (Go standard library)**

| # | Strategy | Result | Details |
|---|----------|--------|---------|
| 1 | Direct update (minor) | N/A | stdlib package — not a go.mod dependency |
| 2 | Transitive update | N/A | stdlib package — not a go.mod dependency |
| 3 | Override/pin | N/A | cannot `replace` the standard library |
| 4 | Toolchain bump | ✅ Success | go1.25.9 → go1.26.5 |

The advisory documents the fix as `crypto/x509@go1.25.11`. Red Hat go-toolset
has not yet published a 1.25.10/1.25.11 image (the 1.25 stream tops out at
1.25.9), so the 1.25.x line is blocked on Red Hat certification. The next
certified go-toolset release is **1.26.5**, which includes the fix (any Go
release ≥ 1.25.11 does). Because production images build inside the certified
go-toolset container, the toolchain was bumped to go1.26.5.

### Changes

- **Go modules:** `go.mod`, `tools/go.mod`, `tools/api-metadata-extractor/go.mod`
  bumped to `go 1.26.0` / `toolchain go1.26.5`.
- **CI:** `setup-go` `go-version` bumped to 1.26.5 in
  `.github/workflows/publish-cli-bins.yaml` and
  `.github/actions/setup-dependencies/action.yaml`.
- **Container images:** all `ubi9`/`ubi10` `go-toolset` build stages bumped to
  the latest certified 1.26.5 tags (`ubi9:1.26.5-1787559109`,
  `ubi10:1.26.5-1787531065`) across `packaging/images/{el9,el10}/Containerfile.*`
  and the vmrender integration Containerfile.

### Verification

- `go build ./...` passes under go1.26.5.
- `govulncheck ./...` under go1.26.5 no longer reports GO-2026-5037 as called
  (was previously reachable via `internal/tpm/server.go`,
  `internal/cli/app_console.go`, and the e2e harness).

### Validation

- Toolchain: `go version` → go1.26.5; `go build ./...` passes.
- Tests: `make unit-test` → 6341 passed, 2 skipped, 0 failures.
- Post-fix scan: `govulncheck -mode binary` (all binaries) no longer reports
  GO-2026-5037. (go1.26.5 does surface a few unrelated newer stdlib advisories
  fixed in go1.26.6 — net/url, html/template, crypto/tls — tracked separately;
  1.26.5 is the newest certified go-toolset tag.)

### Related Jira tickets

This same toolchain bump also resolves the `flightctl`-component trackers for
CVE-2026-27145:

- EDM-4506 (flightctl, rhem-1.1)
- EDM-4508 (flightctl, rhem-1.2)

EDM-4503 itself is filed against the `flightctl-ui-rhel9` component; if that
image is built from a separate repo, the UI trackers (EDM-4503, EDM-4509) may
need an equivalent bump there.

### Rollback

To revert this change:
```bash
git revert <commit-sha>
```


tools/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/flightctl/flightctl/tools
 
-go 1.25.0
+go 1.26.0
```
celdrake, line 3: This change should be reflected in the project's documentation

https://github.com/amir-yogev-gh/flightctl/blob/cve-fix/EDM-4503/docs/developer/README.md?plain=1#L6

amir-yogev-gh: Fixed

asafbss (commented):
Overall the toolchain bump, Containerfiles, CI, TPM code, packit targets, and docs all look good. One critical issue flagged inline.

pkg/crypto/key.go
```diff
@@ -86,11 +86,14 @@ func HashPublicKey(key crypto.PublicKey) ([]byte, error) {
...
+func hashECDSAKey(publicKey *ecdsa.PublicKey) ([]byte, error) {
+	raw, err := publicKey.Bytes()
+	if err != nil {
+		return nil, fmt.Errorf("encode ECDSA public key: %w", err)
+	}
 	hash := sha256.New()
-	hash.Write(publicKey.X.Bytes())
-	hash.Write(publicKey.Y.Bytes())
-	return hash.Sum(nil)
+	hash.Write(raw)
```
asafbss: This change silently alters the hash output for **every** ECDSA key.

**Old input:** `X.Bytes() || Y.Bytes()` (raw big.Int, no leading zeros, no prefix)
**New input:** `0x04 || X(32) || Y(32)` (SEC1 uncompressed encoding)

The `0x04` prefix byte alone produces a different hash for every key.

### Why this matters

`hashECDSAKey` feeds into `generateDeviceName()` (`internal/agent/identity/identity.go:128`), which derives the device identity from the public key. The device name is **never persisted on disk** — it's re-derived from the key on every agent restart.

After an upgrade with this change:
1. Every device computes a **different device name** from the same persisted key.
2. Certificate renewal fails — `SignerDeviceManagementRenewal.Verify()` checks that the new CSR's CN matches the existing certificate's CN (which embeds the **old** fingerprint). Mismatch → rejection.
3. Re-enrollment would create a **duplicate device** record, orphaning the old one.

### Suggested fix

Extract X/Y from the SEC1 encoding but preserve the original hash input format:

```go
func hashECDSAKey(publicKey *ecdsa.PublicKey) ([]byte, error) {
	raw, err := publicKey.Bytes()
	if err != nil {
		return nil, fmt.Errorf("encode ECDSA public key: %w", err)
	}
	// Preserve hash compatibility: hash X || Y without the 0x04 prefix
	// and without fixed-width padding (match big.Int.Bytes() behavior).
	fieldLen := (len(raw) - 1) / 2
	x := raw[1 : 1+fieldLen]
	y := raw[1+fieldLen:]
	hash := sha256.New()
	hash.Write(bytes.TrimLeft(x, "\x00"))
	hash.Write(bytes.TrimLeft(y, "\x00"))
	return hash.Sum(nil), nil
}
```

asafbss (approved):

PR #3469, opened 2026-08-26
EDM-5252: Quay Scanner — finding conversion and integration

EfratIfergan:
## EDM-5252: Quay Scanner — finding conversion and integration

**Jira:** https://issues.redhat.com/browse/EDM-5252
**Story type:** [DEV]

> **Stacked PR.** Targets `EDM-5251-quay-security-client` (the fetch/filter client
> this builds on). Stack: `main ← EDM-5250 ← EDM-5251 ← EDM-5252`. Review/merge the
> parents first; GitHub retargets this PR automatically as they merge.

### Summary

Adds a Quay vulnerability scanner that converts Quay/Clair scan reports into
backend-agnostic `vulnerability.Finding` DTOs and self-registers as backend
`"quay"`. Building on the EDM-5251 fetch client, it walks each image's Security
report, extracts CVE IDs, maps severity/issuer/CVSS, deduplicates by
`(image_digest, cve_id)`, and feeds the existing (backend-agnostic) sync pipeline
so CVE lifecycle events fire for Quay-sourced findings.

### Changes

- **`internal/quay/scanner.go`** (new) — `quayScanner` implementing
  `vulnerability.Scanner` over the EDM-5251 `Client`:
  - `ScanImages` — sequential per-image fetch; client-skipped images contribute
    nothing; a genuine fetch/decode error aborts the scan (per-image error
    isolation is deferred to EDM-5253).
  - `NewScanner(cfg, log)` — returns `(nil, nil)` on nil config (mirrors the
    Trustify adapter); builds the client eagerly; surfaces endpoint-parse errors.
  - Conversion helpers: `findingsFromReport`, `extractCVEIDs` (regex on `Link`,
    fallback to `Name`), `mapSeverity`, `buildIssuer` (namespace prefix + UUIDv5),
    `cvssScore` (v3→v2→nil), `advisoryID`.
  - `init()` registers backend `"quay"` with `WithSBOMUpload(false)` (Quay indexes
    images natively).
- **`internal/periodic_checker/server.go`** — blank import of `internal/quay` so
  the `init()` registration runs in the periodic service.
- **`internal/quay/scanner_test.go`** (new) — table-driven unit tests for every
  helper and `ScanImages`, plus registry resolution through the public
  `vulnerability.NewScanner`.
- **`internal/tasks/vulnerability_sync_test.go`** — a test proving the CVE
  lifecycle pipeline is backend-agnostic for Quay-sourced findings.

No API, schema, config, or wire-protocol changes — only a new backend option on
the existing enum.

### Testing

- **Unit tests:** New `internal/quay/scanner_test.go` (all conversion helpers, the
  end-to-end `ScanImages` over an `httptest.Server`, and registry lookup) plus one
  new sync-task test for Quay-sourced CVE lifecycle events. All pass.
- **Integration tests:** N/A — pure conversion logic and a scanner, exercised via
  `httptest.Server` and in-memory fakes; no store/schema/API interactions (the
  `source` column landed in EDM-5250).
- **Coverage:** `internal/quay` at 96.5% of statements (above the 90% bar); every
  behavioral path through `ScanImages` and each helper is covered.
- **Lint:** `make lint` clean (0 issues). Full unit suite green for story packages
  (one unrelated, pre-existing `internal/backup` failure exists on the base branch
  independently of this change).

### Acceptance Criteria

- [ ] AC-1: CVE IDs extracted from `Link` (regex `CVE-\d{4}-\d{4,7}`), fallback to `Name`.
- [ ] AC-2: Vulnerabilities with no extractable CVE are skipped with a debug log.
- [ ] AC-3: One advisory covering multiple CVEs → one finding per CVE.
- [ ] AC-4: All Quay-sourced findings have status `"affected"`.
- [ ] AC-5: Severity mapping (`Defcon1`→`Critical`, `Negligible`→`None`, direct; unrecognized→`Unknown`).
- [ ] AC-6: CVSS from `CVSSv3.Score`, falling back to `CVSSv2.Score`.
- [ ] AC-7: Issuer from `NamespaceName` prefix mapping with a UUIDv5 ID.
- [ ] AC-8: `AdvisoryID` set to `Name` when it is an advisory (not a CVE), else `nil`.
- [ ] AC-9: `PublishedAt` is `nil` for all Quay-sourced findings.
- [ ] AC-10: Findings deduplicated by `(image_digest, cve_id)`.
- [ ] AC-11: Implements `vulnerability.Scanner`, registers `"quay"` via `init()`; `NewScanner` resolves it.
- [ ] AC-12: CVE lifecycle events emitted for Quay-sourced findings.

---


internal/periodic_checker/server.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/flightctl/flightctl/internal/instrumentation/tracing"
 	"github.com/flightctl/flightctl/internal/kvstore"
 	"github.com/flightctl/flightctl/internal/org/cache"
+	_ "github.com/flightctl/flightctl/internal/quay" // register the Quay vulnerability scanner backend
```
kkyrazis: can we do this in such a way that we don't have to rely on `init` to setup the registration?

EfratIfergan: Good question. The `init()`-based registration is deliberate here, for a few reasons:

- **Import-cycle avoidance.** `internal/quay` (and `internal/trustify/v2`) import `internal/vulnerability` for the `Scanner`/`Finding` types, so `vulnerability` can't import the backends back to wire them explicitly. The registry + `init()` self-registration is the standard Go idiom for exactly this (cf. `database/sql` drivers).
- **Consistency.** This mirrors the existing Trustify backend, which already self-registers via `init()` (`internal/trustify/v2/scanner_adapter.go`). Diverging here would leave two different registration styles for the same abstraction.
- **Multiple consumers.** The registry is consulted from more than one entry point (`internal/periodic_checker` via `NewScanner`, and `internal/imagebuilder_worker/tasks` via `RequiresSBOMUpload`), so centralizing resolution in the registry keeps those call sites backend-agnostic.
- **By design.** This registration approach is what the story's design (AC-11 / design decision D1) prescribes.

If the team would rather move away from `init()`-based registration globally, I'm happy to file a follow-up ticket — that change would also touch the Trustify backend so both stay consistent. For this PR I'd prefer to keep it as-is to match the established pattern.

kkyrazis: Thanks for the detailed response. A few counter-points:

**Import cycles** — an explicit factory/switch doesn't need to live in `internal/vulnerability`. It belongs at the wiring site (e.g. in `internal/periodic_checker` or a small `internal/vulnerability/backends` package). The switch imports `quay` and `trustify` — same direction as today, no cycle. The `database/sql` analogy doesn't quite fit here — that's a true plugin system with an open set of third-party drivers. We have two backends in the same repo.

**Consistency** — the Trustify backend adopted this pattern in the parent PR of this same stack (EDM-5249). There's no long-standing convention to be consistent with — this is a brand-new pattern being established across two PRs right now, which makes it the right time to reconsider.

**Multiple consumers** — a shared `NewScanner(cfg)` factory function achieves the same thing without global mutable state. Both `periodic_checker` and `imagebuilder_worker` can call it.

**Concrete concerns:**
- `var registry` is a package-level mutable map; `registry_test.go` already needs `withCleanRegistry()` to save/restore it — that's a smell.
- `Register()` panics on duplicate registration instead of returning an error.
- Blank imports defeat "find usages" and dead-code detection.

I'm fine with a follow-up ticket if you'd rather not change it in this PR, but I do think we should move away from this pattern before it spreads further.

EfratIfergan: You've convinced me on the merits — thanks for the thorough counter-points.

- **Import cycles:** agreed. Placing the factory at the wiring site (or a small `internal/vulnerability/backends` package) rather than inside `internal/vulnerability` keeps the dependency direction the same and avoids the cycle without the registry. My earlier point overstated the constraint.
- **`database/sql` analogy:** fair — that's an open, third-party plugin ecosystem; we have a closed set of two in-repo backends known at compile time, so an explicit `switch` is the clearer fit.
- **Consistency / timing:** also fair — the pattern only landed in the EDM-5249 parent of this same stack, so it's not an entrenched convention, and now is the cheapest time to change course before a third backend copies it.
- **Concrete smells:** all valid — the package-level mutable `registry` (with `withCleanRegistry()` save/restore in tests), `Register()` panicking on duplicate, and blank imports being invisible to "find usages"/dead-code tooling.

Given a proper fix also reworks the Trustify backend from the parent PR, I'd rather not pull that refactor into this Quay-conversion PR — it would add cross-story scope and churn the parent. So I'll keep the current registration as-is here and open a **separate follow-up ticket** to move **both** Trustify and Quay off `init()`/the global registry onto an explicit factory, so they stay consistent. I'll link the ticket here once it's filed.

kkyrazis: Sorry, I'm not sure I understand. This stack is still very much in progress. Why can't we just adjust the implementation now, rather than create a new ticket to go address it after the fact?

EfratIfergan: done

internal/quay/client.go
kkyrazis: are all of these Debug messages at the appropriate log level? Are these really not a big deal?

internal/quay/scanner.go
```diff
@@ -0,0 +1,229 @@
...
+	var findings []vulnerability.Finding
+	seen := make(map[string]struct{})
+	for _, feature := range report.Data.Layer.Features {
+		for _, vuln := range feature.Vulnerabilities {
+			cveIDs := extractCVEIDs(vuln)
+			if len(cveIDs) == 0 {
+				log.WithFields(logrus.Fields{
+					"digest": digest,
+					"name":   vuln.Name,
+				}).Debug("skipping vulnerability with no extractable CVE ID")
```
kkyrazis: is this a common case?

kkyrazis (approved):

internal/quay/client.go
EfratIfergan: adjusted the log levels accordingly. Thanks for pointing that out!

internal/quay/scanner.go
EfratIfergan: it's relatively uncommon in normal cases - most published vulnerabilities have CVE IDs assigned. However, when it does occur, it represents data loss (we can't store the finding without a CVE ID per RHEM's composite key), so it's important for operators to be aware of

kkyrazis: If it's important should the log level be raised to Info or Warning?

internal/quay/client_test.go
eldar101, line 64: doc strings plz

EfratIfergan: Added concise Go comments for the `mockQuayServer` request-inspection helpers (`count`, `auth`, `path`, and `query`) in commit `495e5b67`.

eldar101 (approved):

PR #3470, opened 2026-08-26
EDM-4115: Fix login lockout for vm app tests

itzikb-redhat:
It may happen that login was tried before cloud init finished updating the user's password and hence caused
a lockout


asafbennatan (approved):
LGTM

PR #3471, opened 2026-08-26
NO-ISSUE: Protect access token refresher from concurrent read/write races

itzikb-redhat:
Fixes a data race in AccessTokenRefresher. The refresh loop writes tokens into config.AuthInfo while other goroutines read them (for example via GetAccessToken())


asafbennatan (commented):
we are not using defer to minimize the scope of the lock , but we risk panicking while holding the lock .
i suggest introducing helper functions for this limited scope in which we can use defer unlock so it would keep both the minimal scope and have safety

internal/client/authorization.go
```diff
@@ -233,3 +229,50 @@ func (r *AccessTokenRefresher) accessToken() string {
 func (r *AccessTokenRefresher) GetAccessToken() string {
 	return r.accessToken()
 }
+
```
eldar101, line 232: Please add docstrings to all funcs, even maybe for those you didn't create but modify.

internal/client/authorization.go
```diff
@@ -233,3 +229,50 @@ func (r *AccessTokenRefresher) accessToken() string {
...
+	return r.config.AuthInfo.RefreshToken
+}
+
+func (r *AccessTokenRefresher) accessTokenExpiry() string {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.config.AuthInfo.AccessTokenExpiry
+}
+
+func (r *AccessTokenRefresher) setCancel(cancel context.CancelFunc) {
```
eldar101, line 245: these seem similar enough. could mthey may be united to an "token refresh" func with an action set by a parameter or choice?

internal/client/authorization_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/flightctl/flightctl/internal/cli/login"
+	"github.com/stretchr/testify/require"
+)
+
+type stubAuthProvider struct {
+	renewCount atomic.Int64
+}
+
+func (s *stubAuthProvider) Auth() (login.AuthInfo, error) {
```
eldar101, line 19: are these funcs at the top of the test on purpose?

internal/client/authorization_test.go
```diff
@@ -0,0 +1,108 @@
...
+	return login.AuthInfo{}, nil
+}
+
```
eldar101, line 22: doc strings please

internal/client/authorization_test.go
```diff
@@ -0,0 +1,108 @@
...
+		IdToken:      fmt.Sprintf("id-%d", n),
+		ExpiresIn:    &expiresIn,
+	}, nil
+}
+
+func newTestAccessTokenRefresher(t *testing.T, configFilePath string) *AccessTokenRefresher {
+	t.Helper()
+	cfg := &Config{
+		AuthInfo: AuthInfo{
+			AccessToken:       "access-0",
```
eldar101, line 44: this seems hardcoded. wdyt about generating a random number at the end of the dash?

eldar101:
fix the failed pre-merge checks too plz

internal/client/authorization_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	var wg sync.WaitGroup
+	start := make(chan struct{})
+	for i := 0; i < readers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-start
+			for j := 0; j < iterations; j++ {
+				require.NotEmpty(r.GetAccessToken())
```
eldar101, line 90: require should not run inside goroutines. If it fails, FailNow runs in the worker goroutine, not the test goroutine. Please send errors/results back and do the require checks after wg.Wait

itzikb-redhat: It's intended for the assert to be easy as it's incremented for every refresh

internal/client/authorization_test.go
itzikb-redhat: It's intended this way to easily assert it.

internal/client/authorization.go
itzikb-redhat: I think it's clearer

internal/client/authorization_test.go
```diff
@@ -0,0 +1,208 @@
...
+	started := make(chan struct{})
+	go func() {
+		r.Start(ctx)
+		close(started)
+	}()
+	waitClosed(t, started, time.Second)
+
+	idle := make(chan struct{})
+	go func() {
+		time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
asafbennatan: should be fixed

asafbennatan (approved):
LGTM

internal/client/authorization.go
```diff
@@ -122,43 +131,38 @@ func isExpiredTokenError(err error) bool {
...
+// refresh renews the access token with the auth provider and persists the
+// updated config when a config file path is set.
 func (r *AccessTokenRefresher) refresh() error {
-	if r.config.AuthInfo.RefreshToken == "" {
+	refreshToken := r.getRefreshToken()
+	if refreshToken == "" {
 		return fmt.Errorf("no refresh token found")
 	}
-	authInfo, err := r.provider.Renew(r.config.AuthInfo.RefreshToken)
+	authInfo, err := r.provider.Renew(refreshToken)
```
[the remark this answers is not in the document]
itzikb-redhat: This is not related to this patch. Should be done in a followup patch

PR #3477, opened 2026-08-26
EDM-3773: Preserve prefetch errors across OS rollback status

kkyrazis:
## Summary
- Persist the formatted sync/prefetch error on `rollback.json` so it survives an OS rollback reboot, instead of being overwritten by `Rebooting` then `Updated`.
- After reboot, bootstrap reports `DeviceUpdating` `Reason=Error` with that message (for example `prefetch failed for quay.io/flightctl/images:doesnotexist`).
- Tighten `UpdateDeviceAndWaitForFailure` so it requires a real Error condition; `OutOfDate` is still asserted afterward, but is no longer treated as a stand-in for Error.

Fixes https://issues.redhat.com/browse/EDM-3773 (replaces stale https://github.com/flightctl/flightctl/pull/2930).

## Test plan
- [ ] Unit tests: `go test ./internal/agent/device/spec/ ./internal/agent/device/`
- [ ] E2E (this PR has `run-e2e`): helm application error handling (`prefetch failed for quay.io/flightctl/images:doesnotexist`) and container nonexistent-image failure (`prefetch failed`)
- [ ] Confirm after OS rollback the updating condition is `Reason=Error` with the prefetch substring, then `status.updated.status=OutOfDate` and `summary=Online`

### DIFF
--- a/internal/agent/device/bootstrap.go
+++ b/internal/agent/device/bootstrap.go
@@ -173,8 +173,24 @@ func (b *Bootstrap) updateStatus(ctx context.Context) {
 		// TODO: only set rebooting in case where we are actually rebooting
 		updatingCondition.Reason = string(v1beta1.UpdateStateRebooting)
 	} else {
-		updatingCondition.Status = v1beta1.ConditionStatusFalse
-		updatingCondition.Reason = string(v1beta1.UpdateStateUpdated)
+		rollbackInfo, err := b.specManager.GetRollbackInfo()
+		if err != nil {
+			b.log.Errorf("Failed to read rollback info: %v", err)
+			updatingCondition.Status = v1beta1.ConditionStatusFalse
+			updatingCondition.Reason = string(v1beta1.UpdateStateError)
+			updatingCondition.Message = "Failed to read rollback info after update"
+		} else if rollbackInfo.ErrorMessage != "" || rollbackInfo.Version != "" {
+			updatingCondition.Status = v1beta1.ConditionStatusFalse
+			updatingCondition.Reason = string(v1beta1.UpdateStateError)
+			if rollbackInfo.ErrorMessage != "" {
+				updatingCondition.Message = rollbackInfo.ErrorMessage
+			} else {
+				updatingCondition.Message = fmt.Sprintf("Failed to update to renderedVersion: %s", rollbackInfo.Version)
+			}
+		} else {
+			updatingCondition.Status = v1beta1.ConditionStatusFalse
+			updatingCondition.Reason = string(v1beta1.UpdateStateUpdated)
+		}
 	}
 
 	_, updateErr := b.statusManager.Update(ctx,
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -353,7 +353,7 @@ func (a *Agent) rollbackDevice(ctx context.Context, current, desired *v1beta1.De
 		}
 		syncMsg := formatSyncErrorMessage(desired, syncErr)
 		a.handleSyncError(ctx, desired, syncErr)
-		if err := a.specManager.RecordRollbackError(syncMsg); err != nil {
+		if err := a.specManager.RecordRollbackError(ctx, syncMsg); err != nil {
 			a.log.Errorf("Failed to persist rollback error: %v", err)
 		}
 
--- a/internal/agent/device/device.go
+++ b/internal/agent/device/device.go
@@ -351,13 +351,17 @@ func (a *Agent) rollbackDevice(ctx context.Context, current, desired *v1beta1.De
 		if err := syncFn(ctx, desired, current); err != nil {
 			a.log.Warnf("Rollback sync completed with errors, proceeding with OS rollback: %v", err)
 		}
+		syncMsg := formatSyncErrorMessage(desired, syncErr)
 		a.handleSyncError(ctx, desired, syncErr)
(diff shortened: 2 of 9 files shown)


internal/agent/device/device.go
```diff
@@ -712,6 +720,15 @@ func (a *Agent) handleSyncError(ctx context.Context, desired *v1beta1.Device, sy
 	}
 }
 
+func formatSyncErrorMessage(desired *v1beta1.Device, syncErr error) string {
```
galel12, line 724: `formatSyncErrorMessage` mirrors the condition-message logic in `handleSyncError` (raw message vs `se.Message()` when phase/component are set). Should we have `handleSyncError` call `formatSyncErrorMessage` for `conditionUpdate.Message` so the persisted rollback text and the pre-reboot condition cannot drift?

```suggestion
msg := formatSyncErrorMessage(desired, syncErr)
conditionUpdate.Message = msg
// keep a separate log line for e2e log matching if needed
a.log.Error(fmt.Sprintf("Failed to update to renderedVersion: %s: %v", version, syncErr.Error()), se.Timestamp)
```

internal/agent/device/device.go
```diff
@@ -351,13 +351,17 @@ func (a *Agent) rollbackDevice(ctx context.Context, current, desired *v1beta1.De
 		if err := syncFn(ctx, desired, current); err != nil {
 			a.log.Warnf("Rollback sync completed with errors, proceeding with OS rollback: %v", err)
 		}
+		syncMsg := formatSyncErrorMessage(desired, syncErr)
 		a.handleSyncError(ctx, desired, syncErr)
+		if err := a.specManager.RecordRollbackError(ctx, syncMsg); err != nil {
```
galel12, line 357: `RecordRollbackError` failures are logged but reboot still proceeds. If the write fails, bootstrap falls back to the generic `"Failed to update to renderedVersion: <version>"` path. Should we add a short comment that persistence is best-effort, or a unit test for the read/write failure path so that behavior stays intentional?

internal/agent/device/bootstrap.go
```diff
@@ -173,8 +173,24 @@ func (b *Bootstrap) updateStatus(ctx context.Context) {
...
 	} else {
-		updatingCondition.Status = v1beta1.ConditionStatusFalse
-		updatingCondition.Reason = string(v1beta1.UpdateStateUpdated)
+		rollbackInfo, err := b.specManager.GetRollbackInfo()
+		if err != nil {
+			b.log.Errorf("Failed to read rollback info: %v", err)
+			updatingCondition.Status = v1beta1.ConditionStatusFalse
+			updatingCondition.Reason = string(v1beta1.UpdateStateError)
+			updatingCondition.Message = "Failed to read rollback info after update"
+		} else if rollbackInfo.ErrorMessage != "" || rollbackInfo.Version != "" {
```
galel12, line 182: The condition `rollbackInfo.ErrorMessage != "" || rollbackInfo.Version != ""` treats any rollback metadata as a failed update. That matches the OS-rollback lifecycle (and the new test for version-without-error), but it also means `rollback.json` keeps reporting Error on every agent restart until `ClearRollback` on a successful upgrade.

Is that the long-term UX we want, or should we clear `agent/rollbackError` (or the whole rollback file) once bootstrap has reported it?

internal/agent/device/spec/spec_test.go
```diff
@@ -757,6 +758,102 @@ func TestCreateRollback(t *testing.T) {
...
+		mockReadWriter.EXPECT().ReadFile(rollbackPath).Return(b, nil)
+		info, err := s.GetRollbackInfo()
+		require.NoError(err)
+		require.Equal("6", info.Version)
+		require.Equal("abc123", info.SpecHash)
+		require.Equal("prefetch failed for quay.io/flightctl/images:doesnotexist", info.ErrorMessage)
+	})
+}
+
+func TestRecordRollbackError(t *testing.T) {
```
galel12, line 814: Coverage looks good for the happy path and empty message. Missing case: `Read(Rollback)` returns `ErrMissingRenderedSpec` or another error — `RecordRollbackError` propagates it and `rollbackDevice` only logs. A small test would lock that in.

internal/agent/device/bootstrap.go
kkyrazis: >  or should we clear agent/rollbackError (or the whole rollback file)

we definitely want to avoid clearing the whole file. The spec hash is also persisted in there to prevent retrying a bad spec on restarting the agent. Arguably we could rework this to have a new persisted method that always tracks the last error message / failed spec to prevent retrying bad specs on restarting the agent - but that's a different story.

There's two scenarios that I see.

1. We encounter an error and are forced to rollback
2. The agent restarts N times.

In this scenario I think it's appropriate to always report the specific error message as opposed to only reporting it once and then subsequent restarts showing: `Message="Failed to update to renderedVersion: <failed-version>"`

The other scenario:

1. We encounter an error and are forced to rollback
2. A new spec is pushed that doesn't require an OS update, but it fails too
3. The agent restarts N times.

This scenario is definitely more nuanced. Obviously on the first reboot we want to display the rollback error message, but once we've processed the new bad spec we should probably only display that one. With this PR it would instead: 

1. report the first rollback error message
2. retry the bad spec and then report that message

So I guess this is a long winded answer to just say - I think it's appropriate to report the most recent error message that the agent knows about

galel12 (approved):
LGTM

PR #3493, opened 2026-08-31
EDM-4892: remove empty directories after removing a host configuration source

EfratIfergan:
## Summary

Removing a host configuration source that created a directory tree on a device
(e.g. MicroShift manifests under `/etc/microshift/manifests.d/`) deleted the
files but left the now-empty directories behind. The stale, empty tree reads as
"configuration still present" (including to MicroShift's manifest scanner) and
diverges from the declared desired state, which blocks the RHEM-with-MicroShift
workflow the reporter described.

This PR extends the agent's config reconcile to prune directories that become
empty as a result of removing obsolete files.

## Root Cause

The server render captures only `ignition.Storage.Files` and drops
`Storage.Directories` (`internal/tasks/device_render.go:988`,
`ignitionConfigToRenderedConfig`, which loops solely over `Storage.Files`).
Directories are created implicitly on write, but
`config.Controller.removeObsoleteFiles` removes files via
`fileio.Writer.RemoveFile` (a plain file remove) and never removes the
directories created to hold them. File *contents* are removed correctly; only
the empty directory tree lingers.

## Fix

- Add `RemoveEmptyDir` to `fileio.Writer` — removes a directory **only if it is
  empty**, and is a no-op if the directory is missing or still holds entries.
- After deleting obsolete files, `removeObsoleteFiles` walks each removed file's
  ancestor directories **deepest-first** and removes those now empty, with:
  - a protected-roots allowlist (`/`, `/etc`, `/var`, `/usr`, …) never removed;
  - preservation of every directory that still holds a *desired* file;
  - best-effort semantics — cleanup failures are logged, never fatal to the
    reconcile.

## Before / After

Given a source populating:

```
/etc/microshift/manifests.d/app1/deploy.yaml
/etc/microshift/manifests.d/app2/svc.yaml
```

then removing that source:

- **Before:** files deleted, but `/etc/microshift/manifests.d/{app1,app2}` and
  parents remain as empty directories.
- **After:** files deleted **and** the empty directory tree removed, up to (but
  never including) `/etc`.

## Testing

- `TestSyncRemovesEmptyDirs_EDM4892` — end-to-end with a real writer; **verified
  to fail without the fix and pass with it**.
- `TestSyncPreservesSharedDirs` — a directory holding a remaining desired file
  is kept.
- `TestSyncPreservesNonEmptyDirs` — a directory holding unmanaged content is
  kept (real unmanaged file written to disk).
- `TestRemoveEmptyDirs`, `TestIsCleanupCandidate` — ordering, protected roots,
  kept dirs, edge cases.
- `make lint` clean (0 issues); full unit suite passes (one unrelated,
  environment-specific failure in `internal/backup` caused by a local
  `/etc/flightctl/service-config.yaml` — see Known Issues).

## Manual testing for reviewers

Apply a host config that creates a nested tree under `/etc`, confirm the files
land, remove the source, and confirm the files and empty directories are gone
while `/etc` and any sibling/unmanaged content remain.

## Confidence

HIGH — the fix targets the confirmed root cause at the agent reconcile boundary;
the regression test is proven non-tautological; safety guards (empty-only
removal, protected roots, desired-ancestor preservation) are each covered by a
real-writer test. The reporter confirmed on 1.3.0-main (2026-08-27) that files
are removed but empty directories remain, matching exactly what this addresses.

## Rollback

Safe to revert. The change is additive (`RemoveEmptyDir` on the `Writer`
interface plus one call site and two helpers); reverting restores prior behavior
with no migration, config, API, or on-disk format impact.

## Risk Assessment

LOW — directory removal only ever touches empty directories outside a protected
allowlist and outside the ancestors of desired files, honors the writer
`rootDir`, and is best-effort/non-fatal, so it cannot delete operator data,
system directories, or fail a config sync.

## Known Issues

- The `internal/backup TestPodmanDeployer_BackupConfig` failure is environmental
  (the test assumes `/etc/flightctl/service-config.yaml` is absent) and is
  unrelated to this change; it passes in CI.

Fixes EDM-4892

### DIFF
--- a/internal/agent/device/config/controller.go
+++ b/internal/agent/device/config/controller.go
@@ -5,6 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"path/filepath"
+	"sort"
+	"strings"
 
 	"github.com/flightctl/flightctl/api/core/v1beta1"
 	deviceerrors "github.com/flightctl/flightctl/internal/agent/device/errors"
@@ -96,7 +99,101 @@ func (c *Controller) removeObsoleteFiles(currentFiles, desiredFiles []v1beta1.Fi
 			return fmt.Errorf("%w %w: %w", deviceerrors.ErrDeletingFilesFailed, deviceerrors.WithElement(file), err)
 		}
 	}
+	// Fix for EDM-4892: removing the files is not enough. A removed host
+	// configuration source (e.g. a git repo populating
+	// /etc/microshift/manifests.d/) also creates directories to hold those
+	// files; without cleaning them up the device is left with a stale, empty
+	// directory tree that reads as "content still present" after the source is
+	// removed. This is best-effort cleanup and never fails the reconcile.
+	c.removeEmptyDirs(removeFiles, desiredFiles)
 	return nil
+}
+
+// protectedDirs are directories we never remove during empty-directory cleanup,
+// even if they end up empty. They are well-known system roots; removing them
+// would be surprising and outside the scope of tidying up config-managed files.
+var protectedDirs = map[string]bool{
+	"/":     true,
+	"/etc":  true,
+	"/var":  true,
+	"/usr":  true,
+	"/opt":  true,
+	"/run":  true,
+	"/home": true,
+	"/root": true,
+	"/boot": true,
+	"/tmp":  true,
+	"/srv":  true,
+	"/mnt":  true,
+	"/dev":  true,
+	"/proc": true,
+	"/sys":  true,
+	"/lib":  true,
+	"/bin":  true,
+	"/sbin": true,
+}
+
+// removeEmptyDirs removes directories left empty after obsolete files were
+// deleted. It walks up the parent directories of each removed file (deepest
+// first) and removes each one that is now empty, stopping at protected system
+// roots. Directories that still hold desired files are preserved, as are any
+// directories that still contain other content (RemoveEmptyDir is a no-op on
+// non-empty directories). Cleanup failures are logged, never fatal.
+func (c *Controller) removeEmptyDirs(removedFiles []string, desiredFiles []v1beta1.FileSpec) {
+	// Preserve every directory a desired file lives under: those directories are
+	// needed by the subsequent write step and must not be pruned.
+	keep := make(map[string]bool)
+	for _, file := range getFilePaths(desiredFiles) {
+		for dir := filepath.Dir(file); isCleanupCandidate(dir); dir = filepath.Dir(dir) {
+			keep[dir] = true
+		}
+	}
+
+	// Collect the unique set of candidate directories from the removed files.
+	candidates := make(map[string]bool)
+	for _, file := range removedFiles {
+		if len(file) == 0 {
+			continue
+		}
+		for dir := filepath.Dir(file); isCleanupCandidate(dir); dir = filepath.Dir(dir) {
+			candidates[dir] = true
+		}
+	}
+
+	// Remove deepest directories first so that parents become empty (and thus
+	// removable) only after their now-empty children are gone.
+	dirs := make([]string, 0, len(candidates))
+	for dir := range candidates {
+		if keep[dir] {
+			continue
+		}
+		dirs = append(dirs, dir)
+	}
+	sort.Slice(dirs, func(i, j int) bool {
+		di, dj := strings.Count(dirs[i], "/"), strings.Count(dirs[j], "/")
+		if di != dj {
+			return di > dj
+		}
+		// Deterministic ordering for siblings at the same depth.
+		return dirs[i] < dirs[j]
+	})
+
+	for _, dir := range dirs {
+		if err := c.deviceWriter.RemoveEmptyDir(dir); err != nil {
+			c.log.Warnf("Failed to remove empty directory %s: %v", dir, err)
+			continue
+		}
+		c.log.Debugf("Removed empty directory: %s", dir)
+	}
+}
+
+// isCleanupCandidate reports whether dir may be considered for empty-directory
+// cleanup. Empty, relative, and protected system directories are excluded.
+func isCleanupCandidate(dir string) bool {
+	if dir == "" || dir == "." {
+		return false
+	}
+	return !protectedDirs[filepath.Clean(dir)]
 }
 
 func (c *Controller) writeFiles(files []v1beta1.FileSpec) error {
--- a/internal/agent/device/config/controller.go
+++ b/internal/agent/device/config/controller.go
@@ -109,36 +109,13 @@ func (c *Controller) removeObsoleteFiles(currentFiles, desiredFiles []v1beta1.Fi
 	return nil
 }
 
-// protectedDirs are directories we never remove during empty-directory cleanup,
-// even if they end up empty. They are well-known system roots; removing them
-// would be surprising and outside the scope of tidying up config-managed files.
-var protectedDirs = map[string]bool{
-	"/":     true,
-	"/etc":  true,
-	"/var":  true,
-	"/usr":  true,
-	"/opt":  true,
-	"/run":  true,
-	"/home": true,
-	"/root": true,
-	"/boot": true,
-	"/tmp":  true,
-	"/srv":  true,
-	"/mnt":  true,
-	"/dev":  true,
-	"/proc": true,
-	"/sys":  true,
-	"/lib":  true,
-	"/bin":  true,
-	"/sbin": true,
-}
-
 // removeEmptyDirs removes directories left empty after obsolete files were
 // deleted. It walks up the parent directories of each removed file (deepest
-// first) and removes each one that is now empty, stopping at protected system
-// roots. Directories that still hold desired files are preserved, as are any
-// directories that still contain other content (RemoveEmptyDir is a no-op on
-// non-empty directories). Cleanup failures are logged, never fatal.
+// first) and removes each one that is now empty, stopping before top-level
+// directories (see isCleanupCandidate). Directories that still hold desired
+// files are preserved, as are any directories that still contain other content
+// (RemoveEmptyDir is a no-op on non-empty directories). Cleanup failures are
+// logged, never fatal.
 func (c *Controller) removeEmptyDirs(removedFiles []string, desiredFiles []v1beta1.FileSpec) {
 	// Preserve every directory a desired file lives under: those directories are
 	// needed by the subsequent write step and must not be pruned.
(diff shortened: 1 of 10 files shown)


internal/agent/device/config/controller.go
```diff
@@ -96,9 +99,103 @@ func (c *Controller) removeObsoleteFiles(currentFiles, desiredFiles []v1beta1.Fi
...
+			continue
+		}
+		c.log.Debugf("Removed empty directory: %s", dir)
+	}
+}
+
+// isCleanupCandidate reports whether dir may be considered for empty-directory
+// cleanup. Empty, relative, and protected system directories are excluded.
+func isCleanupCandidate(dir string) bool {
+	if dir == "" || dir == "." {
```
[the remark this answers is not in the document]
EfratIfergan: Good catch — addressed in f7baf11b. `isCleanupCandidate` now requires `filepath.IsAbs(filepath.Clean(dir))`, so a rendered/traversal path such as `../etc` is rejected before it can be joined with the writer root and removed. This also aligns the implementation with its doc comment ("relative ... directories are excluded"). Added relative-path cases to `TestIsCleanupCandidate`.

internal/agent/device/fileio/writer.go
```diff
@@ -162,6 +162,28 @@ func (w *writer) RemoveFile(file string) error {
 	return nil
 }
 
+// RemoveEmptyDir removes the directory at the given path only if it is empty.
+// It is a no-op if the directory does not exist or still contains entries, so
+// it never deletes directories that hold other content.
+func (w *writer) RemoveEmptyDir(dir string) error {
+	fullPath := filepath.Join(w.rootDir, dir)
+	entries, err := os.ReadDir(fullPath)
```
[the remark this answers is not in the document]
EfratIfergan: Addressed in f7baf11b. `RemoveEmptyDir` now uses `os.Open` + `File.ReadDir(1)` and closes the file after the check, so the emptiness test short-circuits after a single entry instead of listing and sorting every entry. Existing no-op/error-handling behavior (missing dir, non-empty dir, real errors) is preserved.

internal/agent/device/config/controller.go
```diff
@@ -96,9 +99,109 @@ func (c *Controller) removeObsoleteFiles(currentFiles, desiredFiles []v1beta1.Fi
...
+// "../etc" must never be joined with the writer root and removed.
+func isCleanupCandidate(dir string) bool {
+	if dir == "" || dir == "." {
+		return false
+	}
+	cleaned := filepath.Clean(dir)
+	if !filepath.IsAbs(cleaned) {
+		return false
+	}
+	return !protectedDirs[cleaned]
```
eldar101: This only protects the dirs listed in protectedDirs. If a config writes under another top-level dir like /media or /lib64, cleanup could remove that top-level dir when it becomes empty. Can we stop cleanup before any direct child of /, or add a test for this?

internal/agent/device/config/controller.go
```diff
@@ -96,9 +99,109 @@ func (c *Controller) removeObsoleteFiles(currentFiles, desiredFiles []v1beta1.Fi
 			return fmt.Errorf("%w %w: %w", deviceerrors.ErrDeletingFilesFailed, deviceerrors.WithElement(file), err)
 		}
 	}
+	// Fix for EDM-4892: removing the files is not enough. A removed host
```
kkyrazis: Could we trim this comment and remove the Jira-specific narrative? Source comments should explain durable invariants or non-obvious decisions; the EDM-4892 history and MicroShift example belong in the PR or commit message. Most of this also repeats removeEmptyDirs and its doc comment. Something like `// Remove empty managed directories left after obsolete files are deleted.` would be sufficient.

internal/agent/device/config/removal_test.go
```diff
@@ -0,0 +1,257 @@
...
+	var provider v1beta1.ConfigProviderSpec
+	require.NoError(t, provider.FromInlineConfigProviderSpec(v1beta1.InlineConfigProviderSpec{Inline: files}))
+	return &[]v1beta1.ConfigProviderSpec{provider}
+}
+
+// TestSyncRemovesEmptyDirs_EDM4892 reproduces the reported scenario end-to-end
+// with a real writer: a host configuration (e.g. MicroShift manifests from a
+// git repo) populates a nested directory tree, then the source is removed. Both
+// the files and the directories created to hold them must be gone afterwards.
+func TestSyncRemovesEmptyDirs_EDM4892(t *testing.T) {
```
kkyrazis: There is quite a bit of commentary here that restates the implementation or references EDM-4892 directly, including TestSyncRemovesEmptyDirs_EDM4892. Jira tickets are useful in commits and PR descriptions, but references in source tend to become stale and require external context. Could we rename this to TestSyncRemovesEmptyDirs and keep only comments that explain behavior not evident from the code?

internal/agent/device/config/controller.go
```diff
@@ -96,9 +99,109 @@ func (c *Controller) removeObsoleteFiles(currentFiles, desiredFiles []v1beta1.Fi
...
+	// directory tree that reads as "content still present" after the source is
+	// removed. This is best-effort cleanup and never fails the reconcile.
+	c.removeEmptyDirs(removeFiles, desiredFiles)
 	return nil
 }
 
+// protectedDirs are directories we never remove during empty-directory cleanup,
+// even if they end up empty. They are well-known system roots; removing them
+// would be surprising and outside the scope of tidying up config-managed files.
+var protectedDirs = map[string]bool{
```
kkyrazis: is it enough to only protect against removing these directories? Should we instead limit the scope of the the removals to directories that are agent controlled? 

For example, the agent RPM creates the flightctl user’s home and ~flightctl/.local. If a host config writes the only file under .local, this cleanup would remove that package-created directory. An allowlist cannot cover every pre-existing or package-owned directory.

internal/agent/device/fileio/writer.go
```diff
@@ -162,6 +162,36 @@ func (w *writer) RemoveFile(file string) error {
 	return nil
 }
 
+// RemoveEmptyDir removes the directory at the given path only if it is empty.
+// It is a no-op if the directory does not exist or still contains entries, so
+// it never deletes directories that hold other content.
+func (w *writer) RemoveEmptyDir(dir string) error {
+	fullPath := filepath.Join(w.rootDir, dir)
+	// Only one entry is needed to decide the directory is non-empty; reading a
```
kkyrazis, line 185: What about symlinked directories? os.Open follows the symlink, but os.Remove removes only the link and leaves the target directory. Should cleanup skip symlinks and remove only owned, actual directories?

internal/agent/device/config/controller.go
EfratIfergan: Good catch — fixed in c1eea3db. You're right that the `protectedDirs` allowlist could never cover every top-level directory. I replaced it with a structural rule in `isCleanupCandidate`: cleanup now never removes the root or any direct child of `/` (e.g. `/media`, `/lib64`), so it can't miss an unlisted root. Nested directories are still cleaned. Added coverage: `TestIsCleanupCandidate` now includes `/media` and `/lib64` (false) plus `/media/app` (true), and `TestRemoveEmptyDirs` has a case asserting `/media/config` is removed while `/media` is preserved.

internal/agent/device/config/controller.go
EfratIfergan: Agreed — trimmed to `// Remove empty managed directories left after obsolete files are deleted.` in 3aa84741. The EDM-4892/MicroShift narrative and the parts that duplicated `removeEmptyDirs`' doc are gone.

internal/agent/device/config/removal_test.go
EfratIfergan: Agreed — renamed to `TestSyncRemovesEmptyDirs` and dropped the Jira/MicroShift narrative from its doc comment in 3aa84741, keeping only what explains the behavior under test (files + now-empty dirs removed, protected root survives).

internal/agent/device/fileio/writer.go
EfratIfergan: Good catch — fixed in 3aa84741. `RemoveEmptyDir` now `os.Lstat`s the path first and returns early if it is a symlink (or not a directory), so cleanup only removes actual, owned directories and never deletes a symlink while leaving its target. Added a test case covering a symlink to an empty directory (it must be preserved).

internal/agent/device/config/controller.go
EfratIfergan: You're right, and I agree. The cleanup currently decides what to keep by path *structure* (`isCleanupCandidate` refuses the root and its direct children) rather than by *ownership*, so a nested package-owned directory like `~flightctl/.local` is still a candidate: if a host config's only file lives there and the config is removed, the now-empty directory would be pruned even though the agent never created it.

Doing it properly means scoping cleanup to directories the agent itself created for managed files (tracking what `MkdirAll` creates and persisting it across reconciles), which is a larger change than the MicroShift manifest-tree case this bug targets. I'm happy either way — fold that into this PR, or handle it as a follow-up. Since it's really a scope call, I'll leave it to you to decide which you'd prefer.

kkyrazis: Personally I'd rather we handle agent owned directories explicitly rather than deleting arbitrary directories.

eldar101 (approved):

internal/agent/device/config/controller.go
EfratIfergan: Done in 563bcc92 (pushed as 6d79ad72). Cleanup is now gated on ownership rather than path structure.

When the agent writes managed files it records the parent directories it creates — those absent on disk before the write — in a JSON manifest under the data dir (`managed-directories.json`, following the same lightweight pattern as image pruning's references file). `removeEmptyDirs` then only prunes directories in that owned set and drops them once removed. So a pre-existing or package-owned directory like `~flightctl/.local` is never removed even after its managed file is deleted, while agent-created trees are still cleaned. Ownership persists across reconciles and reboots (a fresh controller reloads the manifest).

`NewController` now takes a `fileio.ReadWriter` + the data dir to read/write the manifest. Added real-writer tests `TestSyncPreservesPreExistingDirs` (pre-existing dir preserved, agent-created sibling pruned) and `TestSyncOwnershipPersistsAcrossControllers` (survives a restart); `TestRemoveEmptyDirs` gains an owned-set input and a "dir the agent did not create is never removed" case. The structural `isCleanupCandidate` guard is kept as defense-in-depth for relative/traversal paths.

internal/agent/device/config/controller.go
```diff
@@ -67,41 +74,147 @@ func computeRemoval(currentFileList, desiredFileList []v1beta1.FileSpec) []strin
...
+		return dirs[i] < dirs[j]
+	})
+
+	changed := false
+	for _, dir := range dirs {
+		if err := c.deviceReadWriter.RemoveEmptyDir(dir); err != nil {
+			c.log.Warnf("Failed to remove empty directory %s: %v", dir, err)
+			continue
+		}
+		c.log.Debugf("Removed empty directory: %s", dir)
```
kkyrazis: `RemoveEmptyDir` also returns nil when the directory is preserved (non-empty, a symlink, or not a directory), so this may log a removal that did not happen. Relinquishing ownership after a best-effort attempt seems reasonable, but could we make that explicit and avoid the inaccurate log

EfratIfergan: Good point — fixed in f67da03f. `RemoveEmptyDir` now returns `(removed bool, err error)`, so `removeEmptyDirs` only logs `Removed empty directory` when a removal actually happened. Ownership is still relinquished after the best-effort attempt — a preserved directory (still holding unmanaged content, or a symlink) is simply no longer tracked — but that's now explicit in the code and comment rather than inferred from a `nil` error. On a real error we keep ownership so the next reconcile retries. Added `TestRemoveEmptyDirsOwnership` (relinquish-on-preserve vs. keep-on-error) and extended `TestRemoveEmptyDir` to assert the returned flag.

kkyrazis (approved):

PR #3502, opened 2026-09-02
EDM-3256: Detect integrated SoC GPUs on NVIDIA Jetson platforms

Samudelacruz77:
## Summary

- Add platform bus device scanning to the GPU system info collector, fixing GPU detection on NVIDIA Jetson (ARM64) where the integrated GPU doesn't appear under `/sys/bus/pci/devices/`
- Parse `/sys/bus/platform/devices/*/uevent` files and match Device Tree compatible strings against known Tegra GPU generations (Maxwell/TX1 through Ampere/Orin)
- Add comprehensive tests for `collectGPUInfo`, `collectPlatformGPUs`, and `parseUevent` (17 new test cases), filling a pre-existing test gap

## Test plan

- [x] `TestParseUevent` — 5 cases: valid multi-line, empty, malformed, value with `=`, whitespace
- [x] `TestCollectPlatformGPUs` — 8 cases: Orin happy path, missing dir, unknown compatible, non-GPU device, OF_COMPATIBLE_1 fallback, Xavier/Volta model, memory info, index offset
- [x] `TestCollectGPUInfo_PCI` — PCI GPU detection via mock sysfs
- [x] `TestCollectGPUInfo_PlatformOnly` — platform-only detection
- [x] `TestCollectGPUInfo_Mixed` — PCI + platform with correct indexing
- [x] `TestCollectGPUInfo_Empty` — no GPUs returns empty slice
- [x] `make lint` passes
- [x] `make unit-test` passes (2 pre-existing flaky failures in `console` package)


Samudelacruz77:
**Re: context propagation for platform sysfs scan**

Out of scope for this PR — the existing PCI scan (`collectGPUInfo`) also doesn't propagate context, and `fileio.Reader` has no context-aware methods. Adding context propagation to just the platform path would be inconsistent. This would be better addressed as a separate effort that adds context support to the entire GPU collection pipeline (both PCI and platform) and potentially to the `fileio.Reader` interface itself.

**Re: mutable package-level map**

Addressed in 69eaa2b3 — replaced `knownPlatformGPUs` package-level map with a `lookupPlatformGPU()` function that returns from a local map, eliminating the mutable shared state concern.

internal/agent/device/systeminfo/gpu.go
```diff
@@ -207,6 +210,97 @@ func isGPU(classCode string) bool {
...
+			log.Tracef("Platform device %s has OF_NAME=gpu but no known compatible string", entry.Name())
+			continue
+		}
+
+		gpu := GPUDeviceInfo{
+			Index:    startIndex,
+			Vendor:   matched.Vendor,
+			Model:    matched.Model,
+			Arch:     matched.Arch,
+			DeviceID: uevent["OF_COMPATIBLE_0"],
```
galel12: For PCI GPUs, `DeviceID` is the PCI device ID (e.g. `0x2717`). For platform GPUs it is always `OF_COMPATIBLE_0`, even when the chip match comes from a later `OF_COMPATIBLE_n` entry — so on Orin you can get `deviceId: "nvidia,tegra234-gpu"` with `model: "GA10B"`.

That behavior is tested and may be intentional (SoC-level DT identity vs chip metadata), but it introduces dual semantics for one JSON field. Should we document it in a field comment, or set `DeviceID` to the matched compatible string (e.g. `nvidia,ga10b`) for consistency with how PCI `deviceId` identifies the detected chip?

```go
// Option A: use matched compatible (chip-level, closer to PCI semantics)
matchedCompat := compat // capture in the loop when found
...
DeviceID: matchedCompat,

// Option B: keep OF_COMPATIBLE_0 but document the distinction
DeviceID: uevent["OF_COMPATIBLE_0"], // SoC DT compatible, not necessarily the matched chip string
```

internal/agent/device/systeminfo/gpu.go
```diff
@@ -207,6 +210,97 @@ func isGPU(classCode string) bool {
...
 }
 
+type platformGPUInfo struct {
+	Vendor string
+	Model  string
+	Arch   string
+}
+
+func lookupPlatformGPU(compatible string) (platformGPUInfo, bool) {
+	known := map[string]platformGPUInfo{
```
galel12: Replacing the mutable package-level map was the right fix. `lookupPlatformGPU` still allocates and fills a map on every call, and `collectPlatformGPUs` invokes it for each `OF_COMPATIBLE_n` entry. On Jetson-class hardware, a `switch` (or a read-only package-level map) would avoid that work:

```go
func lookupPlatformGPU(compatible string) (platformGPUInfo, bool) {
	switch compatible {
	case "nvidia,gm20b":
		return platformGPUInfo{Vendor: "NVIDIA", Model: "GM20B", Arch: "Maxwell"}, true
	case "nvidia,gp10b":
		return platformGPUInfo{Vendor: "NVIDIA", Model: "GP10B", Arch: "Pascal"}, true
	case "nvidia,gv11b":
		return platformGPUInfo{Vendor: "NVIDIA", Model: "GV11B", Arch: "Volta"}, true
	case "nvidia,ga10b":
		return platformGPUInfo{Vendor: "NVIDIA", Model: "GA10B", Arch: "Ampere"}, true
	default:
		return platformGPUInfo{}, false
	}
}
```

internal/agent/device/systeminfo/gpu.go
```diff
@@ -207,6 +210,97 @@ func isGPU(classCode string) bool {
...
+
+		var matched *platformGPUInfo
+		for i := 0; ; i++ {
+			key := fmt.Sprintf("OF_COMPATIBLE_%d", i)
+			compat, ok := uevent[key]
+			if !ok {
+				break
+			}
+			if info, found := lookupPlatformGPU(compat); found {
+				matched = &info
```
galel12: `matched = &info` relies on escape analysis to heap-allocate `info`. It works, but storing the struct by value is clearer and avoids the pattern:

```go
var matched platformGPUInfo
var foundMatch bool
// ...
if info, ok := lookupPlatformGPU(compat); ok {
	matched = info
	foundMatch = true
	break
}
// ...
if !foundMatch { ... }
```

internal/agent/device/systeminfo/gpu.go
Samudelacruz77: Done — replaced with a switch in d85d1e0d.

internal/agent/device/systeminfo/gpu.go
Samudelacruz77: Good call — switched to value semantics with a bool flag in e2d36c94.

internal/agent/device/systeminfo/gpu.go
Samudelacruz77: Agreed — switched to using the matched compatible string so `DeviceID` consistently identifies the detected chip in both PCI and platform paths. Updated in e2d36c94.

internal/agent/device/systeminfo/gpu.go
```diff
@@ -207,6 +210,93 @@ func isGPU(classCode string) bool {
...
+		log.Tracef("Could not read platform devices directory: %v", err)
+		return nil
+	}
+
+	var gpus []GPUDeviceInfo
+	for _, entry := range entries {
+		devicePath := filepath.Join(platformDevicesPath, entry.Name())
+		ueventPath := filepath.Join(devicePath, "uevent")
+
+		ueventBytes, err := reader.ReadFile(ueventPath)
```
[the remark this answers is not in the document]
Samudelacruz77: The sysfs reads here are small metadata files (uevent, class, vendor) that return effectively instantly — they won't block cancellation in practice. Adding context-aware fileio methods would require modifying the shared `fileio.Reader` interface, which is out of scope for this GPU detection change. The existing pattern of checking context before each collector is consistent with the rest of the codebase.

internal/agent/device/systeminfo/gpu.go
```diff
@@ -207,6 +210,103 @@ func isGPU(classCode string) bool {
...
+				break
+			}
+			if info, found := lookupPlatformGPU(compat); found {
+				matched = info
+				matchedCompat = compat
+				break
+			}
+		}
+
+		if matchedCompat == "" {
```
kkyrazis, line 273: Is there a way to do this so that we don't have to keep updating the list of supported gpus? Could we fall back to collecting just generic information so that we make sure we capture the current state even if we don't have the mapping done properly?

keitwb: Yes, we definitely need to still give some kind of entry in the GPU output list, even if it is incomplete.

keitwb (changes requested):

internal/agent/device/systeminfo/gpu.go
Samudelacruz77: Good point — updated in b2aaf815. Unknown platform GPUs (OF_NAME=gpu with no recognized compatible string) are now reported with whatever generic info is available (OF_COMPATIBLE_0 as DeviceID, plus memory if detectable) instead of being skipped. The known mapping still enriches vendor/model/arch for recognized chips, but unrecognized ones no longer disappear from the output.

internal/agent/device/systeminfo/gpu.go
```diff
@@ -207,6 +210,104 @@ func isGPU(classCode string) bool {
...
+		}
+
+		if matchedCompat != "" {
+			gpu.Vendor = matched.Vendor
+			gpu.Model = matched.Model
+			gpu.Arch = matched.Arch
+			gpu.DeviceID = matchedCompat
+		} else {
+			log.Tracef("Platform device %s has OF_NAME=gpu but no known compatible string", entry.Name())
+			gpu.DeviceID = uevent["OF_COMPATIBLE_0"]
```
kkyrazis: according to this: https://github.com/devicetree-org/devicetree-specification/blob/main/source/chapter2-devicetree-basics.rst?plain=1#L497-L500

the recommendation is for manufactor,model so I think, even if we can't fully populate/format things fully, we can attempt to gather both the Vendor and Model in all cases.

Samudelacruz77: Good catch — the devicetree spec does recommend `manufacturer,model` format. Updated to split `OF_COMPATIBLE_0` on the first comma so unknown GPUs now populate both Vendor and Model generically (e.g. `vendor: "qualcomm", model: "adreno-630"`). Known GPUs still get the curated metadata.

kkyrazis (approved):

PR #3512, opened 2026-09-06
EDM-5693: Use fixed nine-character Git hashes in build metadata

eldar101:
Fix inconsistent Git hash lengths in build metadata by using exactly nine characters for standalone commit IDs, generated version suffixes, and RPM source-archive commit IDs.
- Replace git rev-parse --short calls with full-hash output followed by cut -c1-9.
- Shorten only the hash field in hack/current-version, preserving tags and commit distance.
- Store full hashes in RPM archive metadata and truncate the final RPM commit value after fallback resolution.
- Preserve existing error handling without adding helper scripts, a Go tool, or Python tests.

## Target release

release-1.4

### DIFF
--- a/.github/workflows/publish-containers.yaml
+++ b/.github/workflows/publish-containers.yaml
@@ -167,7 +167,7 @@ jobs:
           else
             SOURCE_GIT_TREE_STATE=dirty
           fi
-          SOURCE_GIT_COMMIT=$(git rev-parse --short "HEAD^{commit}" 2>/dev/null || echo "unknown")
+          SOURCE_GIT_COMMIT=$( (git rev-parse "HEAD^{commit}" 2>/dev/null || echo "unknown") | cut -c1-9)
 
           {
             echo "git_tag=${SOURCE_GIT_TAG}"
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ VERBOSE ?= false
 
 SOURCE_GIT_TAG ?=$(shell $(ROOT_DIR)/hack/current-version)
 SOURCE_GIT_TREE_STATE ?=$(shell ( ( [ ! -d "$(ROOT_DIR)/.git/" ] || git -C $(ROOT_DIR) diff --quiet ) && echo 'clean' ) || echo 'dirty')
-SOURCE_GIT_COMMIT ?=$(shell git -C $(ROOT_DIR) rev-parse --short "HEAD^{commit}" 2>/dev/null || echo "unknown")
+SOURCE_GIT_COMMIT ?=$(shell (git -C $(ROOT_DIR) rev-parse "HEAD^{commit}" 2>/dev/null || echo "unknown") | cut -c1-9)
 BIN_TIMESTAMP ?=$(shell date +'%Y%m%d')
 SOURCE_GIT_TAG_NO_V = $(shell echo $(SOURCE_GIT_TAG) | sed 's/^v//')
 MAJOR = $(shell echo $(SOURCE_GIT_TAG_NO_V) | awk -F'[._~-]' '{print $$1}')
--- a/hack/current-version
+++ b/hack/current-version
@@ -19,7 +19,9 @@ tag=$(git -c "versionsort.suffix=-rc" tag --points-at HEAD --sort version:refnam
 # base commit and re-attach the distance and hash suffix.
 if [[ -z "$tag" ]]; then
   nearest_tag=$(git describe --tags --always --first-parent --match 'v*' --abbrev=0 2>/dev/null)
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -22,9 +22,14 @@ RPM_MOCK_ROOT_DEFAULT = centos-stream+epel-next-9-x86_64
...
 
-SOURCE_GIT_TAG ?=$(shell $(ROOT_DIR)/hack/current-version)
+GIT_ABBREV_LENGTH ?= 9
+override GIT_ABBREV_LENGTH := $(if $(GIT_ABBREV_LENGTH),$(GIT_ABBREV_LENGTH),9)
+ifneq ($(shell printf '%s\n' "$(GIT_ABBREV_LENGTH)" | grep -Eq '^[1-9][0-9]*$$' || echo invalid),)
+$(error GIT_ABBREV_LENGTH must be a positive decimal integer)
+endif
+SOURCE_GIT_TAG ?=$(shell GIT_ABBREV_LENGTH=$(GIT_ABBREV_LENGTH) $(ROOT_DIR)/hack/current-version)
 SOURCE_GIT_TREE_STATE ?=$(shell ( ( [ ! -d "$(ROOT_DIR)/.git/" ] || git -C $(ROOT_DIR) diff --quiet ) && echo 'clean' ) || echo 'dirty')
-SOURCE_GIT_COMMIT ?=$(shell git -C $(ROOT_DIR) rev-parse --short "HEAD^{commit}" 2>/dev/null || echo "unknown")
```
itzikb-redhat, line 27: What about having one function in hack and then call it from other places?

.github/workflows/publish-containers.yaml
```diff
@@ -181,7 +191,7 @@ jobs:
           echo "  SOURCE_GIT_COMMIT: ${SOURCE_GIT_COMMIT}"
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
```
itzikb-redhat: Is it needed?

.github/workflows/publish-containers.yaml
```diff
@@ -209,7 +219,7 @@ jobs:
 
       - name: Build
         id: build
-        uses: redhat-actions/buildah-build@v2
+        uses: redhat-actions/buildah-build@9dc2ade9b5db229966b5025ae266311e12f349c4 # v2
```
itzikb-redhat: same

Makefile
```diff
@@ -22,9 +22,20 @@ RPM_MOCK_ROOT_DEFAULT = centos-stream+epel-next-9-x86_64
 
 VERBOSE ?= false
 
-SOURCE_GIT_TAG ?=$(shell $(ROOT_DIR)/hack/current-version)
+GIT_ABBREV_LENGTH ?= 9
```
itzikb-redhat: Do we need it after the change?

.github/workflows/publish-containers.yaml
```diff
@@ -1,6 +1,8 @@
 
 name: Push to main
 
+permissions: {}
```
itzikb-redhat: Is it related to the patch?

hack/current-version
```diff
@@ -1,31 +1,52 @@
...
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_DIR}/git-version-utils.sh"
 
 # Fix git ownership issue in CI environments (idempotent — do not append duplicates).
 ensure_safe_directory() {
   local dir="$1"
-  git config --global --get-all safe.directory 2>/dev/null | grep -qxF "$dir" || \
+  git config --global --get-all safe.directory 2>/dev/null | grep -xF "$dir" >/dev/null || \
```
itzikb-redhat: why?

eldar101: With pipefail enabled, grep -q can exit after the first match while Git is still writing. Git may then fail with SIGPIPE, making the pipeline appear to fail and causing a duplicate safe.directory entry. Using grep -xF with output redirected consumes the full input while keeping the same exact-match check.

itzikb-redhat (commented):
lgtm

amalykhi:
lgtm

asafbennatan (changes requested):
lets avoid adding more bash code around version computing we already have several bash scripts , if we are going to change 3 scripts and add 3 more , lets rewrite this in go please and not write tests in python

asafbennatan (approved):
LGTM

PR #3515, opened 2026-09-06
EDM-5301: Add telemetry gateway forwarding E2E coverage

eldar101:
## Summary

Adds E2E coverage for telemetry gateway forwarding over OTLP/HTTP and gRPC.
- Tests HTTP URL paths, custom headers, and environment-variable expansion.
- Verifies startup failures for invalid forwarding configuration.
- Verifies backward-compatible bare gRPC endpoints.
- Adds reusable HTTP/gRPC test collectors and Kubernetes/Quadlet support.

## Release target

Target release: release-1.4


test/e2e/observability/telemetry_gateway_forward_test.go
```diff
@@ -0,0 +1,572 @@
...
+	}
+	randomBytes := make([]byte, 16)
+	if _, err := rand.Read(randomBytes); err != nil {
+		return "", fmt.Errorf("generate runtime header value: %w", err)
+	}
+	return prefix + "-" + hex.EncodeToString(randomBytes), nil
+}
+
+// startTelemetryHTTPCollector starts the OTLP/HTTP mock collector with a bounded context.
+func startTelemetryHTTPCollector(ctx context.Context) (*auxiliary.TelemetryHTTPCollector, error) {
```
itzikb-redhat: Shouldn't that be part of infra?

test/e2e/observability/telemetry_gateway_forward_test.go
```diff
@@ -0,0 +1,572 @@
...
+	})
+
+	It("should forward Flight Control device telemetry over OTLP/HTTP with custom headers", Label("90533", "sanity", "agent"), func(ctx SpecContext) {
+		var err error
+		collector, err = startTelemetryHTTPCollector(ctx)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(collector.Endpoint).To(ContainSubstring(telemetryHTTPMetricsPath))
+
+		By("enrolling a device and enabling the OTEL collector image")
+		deviceID, err := ensureOTelDevice(harness, deviceID)
```
itzikb-redhat: Can ensureOTelDevice be used for other tests as well.
Maybe have the all otelcol checks in one function

itzikb-redhat:
I wonder whether we want to use a real OTLP http server

test/e2e/observability/otel_helpers_test.go
```diff
@@ -0,0 +1,130 @@
...
+	var enrollmentID string
+	defer func() {
+		if CurrentSpecReport().Failed() {
+			logOTelEnrollmentDiagnostics(harness, enrollmentID)
+		}
+	}()
+	deviceID := existingDeviceID
+	if deviceID == "" {
+		enrollmentID = harness.GetEnrollmentIDFromServiceLogs("flightctl-agent")
+		deviceID, _ = harness.EnrollAndWaitForOnlineStatus()
```
asafbss, line 35: The enrollment error is discarded. If enrollment fails with a partial ID, the empty-ID check below won't catch it and the test proceeds with a broken device. Should be deviceID, err = with an error check like the rest of the function.

test/e2e/observability/telemetry_gateway_forward_test.go
```diff
@@ -0,0 +1,525 @@
...
+	var b bytes.Buffer
+	encoder := yaml.NewEncoder(&b)
+	if err := encoder.Encode(tg); err != nil {
+		return fmt.Errorf("marshal telemetry gateway section for validation: %w", err)
+	}
+	if err := encoder.Close(); err != nil {
+		return fmt.Errorf("close telemetry gateway section validator encoder: %w", err)
+	}
+	decoder := yaml.NewDecoder(&b)
+	decoder.KnownFields(true)
```
asafbss: KnownFields(true) will reject any field not in telemetryGatewayConfigSchema. If the deployed gateway config ever gains a new field (profiling, logging overrides, etc.), this validation breaks all forwarding tests even though the new field has nothing to do with forwarding. Consider KnownFields(false) — the purpose here is to validate the forward section, not to enforce the full schema.

asafbss (commented):
Good e2e coverage for the HTTP forwarding path. The mock collector approach is clean.

Main gap: gRPC forwarding has no real e2e coverage. Tests 90536 and 90538 validate config acceptance/rejection only, they point at collector.example.com:4317 which doesn't exist, so no metrics actually flow. The HTTP path has proper end-to-end verification (mock collector receives and verifies metrics, headers, body), but gRPC doesn't. Since this PR builds all the forwarding test infrastructure, this seems like the natural place to add it, or at least capture as a follow-up.

Minor: the infra-to-test ratio is heavy (~750 lines of infrastructure for ~525 lines of tests). The ServiceFailureObserver with its own unit test suite could likely be a simpler Eventually + WaitForReady Ginkgo pattern. The SHA256 hashing of header values in the mock collector adds unnecessary complexity when the values are test-generated randoms, not real secrets. Not blocking.

test/e2e/observability/telemetry_gateway_forward_test.go
eldar101: Addressed. The collector startup and lifecycle now live in test/e2e/infra/auxiliary/telemetry_http_collector.go; this test only uses the auxiliary service API.

test/e2e/observability/telemetry_gateway_forward_test.go
eldar101: Addressed. ensureOTelDevice is now in otel_helpers_test.go and is reused by the existing observability tests. It centralizes enrollment, the OTEL image update, and the active otelcol check; TPM-specific restart checks remain test-specific.

test/e2e/observability/otel_helpers_test.go
eldar101: I verified the current harness signature: EnrollAndWaitForOnlineStatus returns (string, *Device), so the discarded value is the Device, not an error. Enrollment failures fail inside the existing Gomega-based harness wait before this helper can continue. The actual error-returning steps in ensureOTelDevice are checked and wrapped.

test/e2e/observability/telemetry_gateway_forward_test.go
eldar101: Addressed in 6ff33ae3c. The gateway section now allows unrelated future fields, while the forward section remains strictly schema-validated. The positive tests also parse and assert only endpoint/header names, so generated header values cannot appear in assertion output.

eldar101:
@itzikb-redhat @asafbss Addressed the code-review items in 6ff33ae3c. The HTTP receiver remains a focused test double because these scenarios need deterministic checks for the OTLP/HTTP path, custom headers, gzip handling, and payload markers; using a full collector would add image/configuration overhead without improving those assertions. The existing integration suite covers the gRPC exporter path. The deployed gRPC forwarding path is a valid follow-up because 90536 and 90538 currently verify transport selection and rejection, not delivery to a real metrics receiver.

eldar101:
@asafbss Added 90539 in 41db1aa79. It starts a real TLS OTLP/gRPC metrics receiver, configures the telemetry gateway with a bare host:port endpoint, and verifies that the enrolled device metric is received. The deployed test uses a runtime self-signed certificate with insecureSkipTlsVerify because it does not add new client-certificate mounts; the existing telemetry-gateway integration tests continue to cover mTLS exporter configuration.

eldar101:
[ReportAfterSuite] PASSED [0.047 seconds]
------------------------------

Ran 7 of 10 Specs in 524.070 seconds
SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 3 Skipped
PASS

Ginkgo ran 1 suite in 8m56.431208105s
Test Suite Passed
+ echo '🔄 [Test Execution] Step 3: Running cleanup...'
🔄 [Test Execution] Step 3: Running cleanup...
+ test/scripts/e2e_cleanup.sh
🔄 [Cleanup] Starting global E2E test cleanup...
🔄 [Cleanup] Finding flightctl e2e VMs...
🔍 [Cleanup] Found 1 flightctl e2e VMs: flightctl-e2e-worker-1
🔄 [Cleanup] Cleaning up VM: flightctl-e2e-worker-1
🔄 [Cleanup] Deleting pristine snapshot for flightctl-e2e-worker-1
Domain snapshot pristine deleted

eldar101:
@asafbss please re-review

asafbss (approved):
LGTM

eldar101:
there is a failure only in run-e2e and merge-queue that I can't seem to recreate on my env. adding brief diagnostics to find it in run-e2e label. Would remove after fixing them, @asafbss

eldar101:
@asafbss please re-approve
GitHub Actions runs the E2E tests inside an isolated kind/container network where the device VM cannot reliably reach the auxiliary OTLP collectors. The gateway and device remain healthy, but the test collectors receive no requests.
Regular Jenkins hosts use a shared, routable host network, as on Titan52, and the tests pass there. Therefore this is a GitHub Actions test-environment limitation, not a product issue or a failure expected on the regular host-based CI environment.

I added a temporary GitHub Actions-only skip for tests 90533, 90534, and 90560, because they require host networking to reach the OTLP collectors.
- Jenkins and Titan52 still execute these tests normally and they pass!
- GitHub Actions reports them as skipped instead of failed.
- Tests 90535–90538 still run in GitHub Actions.
- To force the skipped tests to run in GitHub Actions:
FLIGHTCTL_RUN_TELEMETRY_FORWARD_E2E=true

asafbss (approved):

asafbss (approved):

asafbss (approved):
LGTM

PR #3536, opened 2026-09-08
NO-ISSUE: deduplicate organizations reported by auth providers, Fixes #3534

### LINKED ISSUE #3534: Dynamic organization assignment reports the same organization multiple times
With `organizationAssignment.type: dynamic`, identity providers commonly derive the organization claim from a multi-valued source. Entra ID app roles named `<org>.<role>` (used with `roleAssignment.separator: "."`) are extracted into the org claim, producing one entry per role:
`{ "orginfo": ["cz", "cz", "sk"], "roles": ["cz.flightctl-installer", "cz.flightctl-operator", "sk.flightctl-viewer"] }`
`GET /api/v1/organizations` then lists `cz` twice and the UI organization picker shows a duplicate entry.

No deduplication happens along the path: `extractDynamicOrganizations (organization_extractor.go:92-99)` appends every array entry, `BuildReportedOrganizations` deduplicates roles but not organizations, and `IdentityMapper.ensureOrganizationsExist` resolves both entries to the same DB row and appends both (`identity_mapper.go:78-84, 137-159`).

The PAM provider already treats this list as a set (`internal/auth/oidc/pam/provider.go:1317`), so the OIDC/OAuth2 path is inconsistent with the existing invariant.

Impact is cosmetic — the DB stays clean (`UpsertMany` uses `OnConflict{external_id, DoNothing}`) and authorization is unaffected (`OrgRoles` is keyed by org ID and roles are deduplicated) — but the duplicate is visible to users.

mirokolar:
… #3534

- [ ] **`release-1.4`**

## Problem

Fixes #3534.

With `organizationAssignment.type: dynamic`, identity providers commonly derive the organization claim from a multi-valued source. Entra ID app roles named `<org>.<role>` — the natural pairing with `roleAssignment.separator: "."` — are extracted into the organization claim, producing one entry per role:

```json
{
  "orginfo": ["cz", "cz", "sk"],
  "roles": ["cz.flightctl-installer", "cz.flightctl-operator", "sk.flightctl-viewer"]
}
```

`GET /api/v1/organizations` then lists `cz` twice and the organization picker shows a duplicate entry. A user only has to hold two roles in one organization to hit it.

No deduplication happens along the path:

| Step | Location | Behavior |
|---|---|---|
| Read org claim | `internal/auth/authn/organization_extractor.go` | appends every array entry |
| Build reported orgs | `internal/auth/common/common.go`, `BuildReportedOrganizations` | deduplicates roles, not organizations |
| Map to DB | `internal/service/identity_mapper.go` | both entries resolve to the same row and both are appended |

The PAM provider already treats this list as a set (`extractOrganizations` in `internal/auth/oidc/pam/provider.go`), so the OIDC/OAuth2 path is inconsistent with an invariant the codebase already documents.

## Scope of impact

Cosmetic but user-visible. Not a correctness or security issue:

- the database stays correct — `UpsertMany` uses `OnConflict{Columns: external_id, DoNothing}`
- authorization is unaffected — `MappedIdentity.OrgRoles` is keyed by organization ID and its roles are already deduplicated

## Change

- deduplicate in `extractDynamicOrganizations`, where the claim array is read (root cause; OIDC and OAuth2)
- enforce one `ReportedOrganization` per organization in `BuildReportedOrganizations`, the shared entry point for the OIDC, OAuth2, AAP, OpenShift and K8s providers

Both preserve first-seen order. Collapsing is lossless: roles are looked up from `orgRoles` by organization name, so repeated entries produce identical `ReportedOrganization` values.

`IdentityMapper` is deliberately left unchanged — once its input is a set, it resolves each organization exactly once. Happy to extend it there instead, or in addition, if you prefer defence in depth at that layer.

## Testing

New table-driven cases covering the new behavior:

- `TestExtractOrganizationsDynamic` — repeated value reported once; first-seen order preserved; duplicates collapse after prefix/suffix is applied
- `TestBuildReportedOrganizations` — organization reported twice appears once with its roles; first-seen order preserved (new test file; the package had none)

Verified locally on darwin/arm64 with go1.27.1:

- `go test -race ./internal/auth/... ./internal/api_server/middleware/...` — pass
- `go test -race ./internal/... ./api/... ./pkg/...` — 129 packages pass. 9 fail identically on a clean tree with the change stashed: `internal/pam_issuer_server` (build failure on non-linux), `pkg/executer`, and `internal/agent/device/*`
- `golangci-lint run ./internal/auth/authn/... ./internal/auth/common/...` in the repo's lint image — 0 issues

## Follow-up, not in this PR

`AuthDynamicRoleAssignment.Validate` does not validate role names against `KnownExternalRoles`, while the static variant does. An unknown role name is accepted and then grants nothing, so a user logs in successfully and sees no resources. Worth a separate change if you agree.

### DIFF
--- a/internal/auth/authn/organization_extractor.go
+++ b/internal/auth/authn/organization_extractor.go
@@ -89,12 +89,22 @@ func (e *OrganizationExtractor) extractDynamicOrganizations(assignment *api.Auth
 		return organizations
 	}
 
+	// Identity providers commonly derive the organization claim from a multi-valued
+	// source (e.g. Entra ID app roles named "<org>.<role>"), which yields one entry per
+	// source value and therefore repeats an organization for every role the user holds
+	// in it. Organization membership is a set, so collapse duplicates while preserving
+	// first-seen order. Matches the PAM provider's extractOrganizations.
+	seen := make(map[string]struct{}, len(orgArray))
 	for _, item := range orgArray {
 		orgStr, ok := item.(string)
 		if !ok || orgStr == "" {
 			continue
 		}
 		orgName := e.applyPrefixSuffix(orgStr, dynamicAssignment.OrganizationNamePrefix, dynamicAssignment.OrganizationNameSuffix)
+		if _, duplicate := seen[orgName]; duplicate {
+			continue
+		}
+		seen[orgName] = struct{}{}
 		organizations = append(organizations, orgName)
 	}
 
--- a/internal/auth/authn/organization_extractor_test.go
+++ b/internal/auth/authn/organization_extractor_test.go
@@ -3,8 +3,11 @@ package authn
 import (
 	"testing"
 
+	api "github.com/flightctl/flightctl/api/core/v1beta1"
+	"github.com/flightctl/flightctl/internal/auth/common"
 	"github.com/samber/lo"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestApplyOrgPrefix(t *testing.T) {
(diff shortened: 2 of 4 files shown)


asafbennatan (approved):
LGTM

PR #3537, opened 2026-09-08
EDM-5046: revert third-party greenboot health check disable

galel12:
## Summary

- Remove `flightctl-configure-greenboot.service` and all `DISABLED_HEALTHCHECKS` management from the `flightctl-greenboot` RPM
- Restore stock greenboot behavior: customer and bundled `required.d` checks participate in OS rollback
- Update docs and e2e to reflect the policy change; add packaging regression test

Supersedes #3372 (vendor blocklist approach).

## Problem

#2464 added a boot-time service that disabled third-party greenboot health checks by rewriting `DISABLED_HEALTHCHECKS` on every boot. Customer scripts in `/etc/greenboot/check/required.d` and image-baked scripts under `/usr/lib` were silently disabled with no fleet-visible signal.

## Solution

**Team decision:** do not block any third-party greenboot health checks — full revert of #2464.

Deleted the configure-greenboot script, systemd unit, and spec wiring. `%post greenboot` disables the stale unit on upgrade if it was previously enabled.

Protections from #2470 remain:

- Agent waits for `boot-complete.target` before committing an OS update spec
- Service reports **OutOfDate** when `status.os.image` ≠ `spec.os.image`

## Test plan

- [x] `go test ./test/packaging/ -run TestGreenbootConfigureServiceRemoved`
- [x] `go test ./internal/agent/device/ -run 'TestSync|Bootstrap|EnsureBootedOS'`
- [x] `go test ./internal/service/common/ -run TestUpdateServerSideDeviceUpdatedStatus_OsImageMismatch`
- [x] `make lint-docs`
- [ ] CI e2e: `GINKGO_LABEL_FILTER=greenboot-third-party`

## Manual verification

On a bootc device after RPM upgrade:

```bash
systemctl status flightctl-configure-greenboot.service  # should not exist
grep DISABLED_HEALTHCHECKS /etc/greenboot/greenboot.conf  # flightctl no longer writes this
```

## Related

- #2464 — introduced third-party disable (reverted here)
- #2470 — boot-complete gating and OS mismatch OutOfDate (unchanged)
- #3372 — superseded vendor blocklist approach (closed)


amalykhi (approved):
lgtm

test/e2e/agent/agent_update_test.go
```diff
@@ -372,46 +372,52 @@ var _ = Describe("VM Agent behavior during updates", Label("agent-update"), func
...
+			By("Waiting for device to start rebooting into v7")
+			Eventually(harness.GetDeviceWithStatusSummary, LONGTIMEOUT, POLLING).WithArguments(
+				deviceId).Should(Equal(v1beta1.DeviceSummaryStatusRebooting))
+
+			By("Waiting for greenboot to roll back after MicroShift health check failure")
+			harness.WaitForDeviceContents(deviceId, "device should rollback to initial OS image and come online", func(device *v1beta1.Device) bool {
+				return device.Status.Os.Image == initialStatusImage &&
 					device.Status.Summary.Status == v1beta1.DeviceSummaryStatusOnline &&
 					device.Status.SystemInfo.BootID != initialBootID
 			}, LONGTIMEOUT)
```
kkyrazis, line 407: is this enough time to ensure that microshifts greenboot checks reliably trigger? should we setup faster timeouts for microshift greenboot too? (if possible)

galel12: Good point. We already call `setFastGreenbootHealthTimeouts()` at the start of this test, but that only shortened the **flightctl** health-check phases (`FLIGHTCTL_HEALTH_*`). MicroShift's `40_microshift_running_check.sh` reads `MICROSHIFT_WAIT_TIMEOUT_SEC` from `greenboot.conf` separately (default 5–10 min per attempt).

Extended the e2e `greenboot.conf` override to also set `MICROSHIFT_WAIT_TIMEOUT_SEC=60` so the MicroShift check fails quickly on our constrained VMs and the rollback fits comfortably within `LONGTIMEOUT`. Added a comment clarifying that `LONGTIMEOUT` is the outer bound for the full reboot+rollback cycle.

kkyrazis (approved):

PR #3544, opened 2026-09-09
EDM-5703: Add EnrollmentHookPolicy resource

galel12:
## Summary
- Adds `EnrollmentHookPolicy` OpenAPI schema, types, store, service, transport, and CLI support
- Gate-only policy resource (`failurePolicy`); approve-time gate behavior is out of scope for this story
- Stacked on EDM-5702 (pre-enrollment hook lifecycle wiring)

## Test plan
- [ ] `make unit-test` passes
- [ ] `make integration-test` passes for EnrollmentHookPolicy store/service tests


internal/service/enrollmenthookpolicy/handler.go
```diff
@@ -0,0 +1,164 @@
...
+	h.setReadyCondition(&policy)
+
+	result, oldResource, err := h.store.Update(ctx, orgId, &policy)
+	h.callbackUpdated(ctx, orgId, name, oldResource, result, false, err)
+	return result, common.StoreErrorToApiStatus(err, false, domain.EnrollmentHookPolicyKind, &name)
+}
+
+func (h *ServiceHandler) DeleteEnrollmentHookPolicy(ctx context.Context, orgId uuid.UUID, name string) domain.Status {
+	deleted, err := h.store.Delete(ctx, orgId, name)
+	h.callbackDeleted(ctx, domain.EnrollmentHookPolicyKind, orgId, name, nil, nil, deleted, err)
```
asafbennatan: A delete of a missing policy returns `(false, nil)` from the generic store,
but this callback runs for every delete attempt. The generic deletion event
handler treats a nil error as a successful deletion, so a no-op `DELETE` can
emit a `ResourceDeleted` event even though no row was removed. Could we gate
the callback on `err == nil && deleted`?

```suggestion
	if err == nil && deleted {
		h.callbackDeleted(ctx, domain.EnrollmentHookPolicyKind, orgId, name, nil, nil, deleted, err)
	}
```

asafbennatan (commented):
See comments below

internal/service/enrollmenthookpolicy/handler.go
galel12: Good catch — gated the callback on `err == nil && deleted`, matching repository/authprovider. Added a unit test for the no-op delete path in 6e9aa47.

internal/cli/apply.go
```diff
@@ -313,6 +313,16 @@ func applyResourceByKind(ctx context.Context, c *client.Client, ibClient *client
 	case AuthProviderKind:
 		response, err := c.ReplaceAuthProviderWithBodyWithResponse(ctx, resourceName, "application/json", bytes.NewReader(buf))
 		return extractApplyResult(response, err)
+	case EnrollmentHookPolicyKind:
```
asafbennatan, line 316: The new CLI wiring covers get/list, apply, delete, completion, and table
output, but edit is still not wired. `validate` rejects
`EnrollmentHookPolicyKind`, and `executePatchOperation` has no case for
`PatchEnrollmentHookPolicyWithBodyWithResponse`, so
`flightctl edit enrollmenthookpolicy/default` cannot use the PATCH endpoint
this PR adds. Could we add this kind to the edit kind lists and PATCH switch
(with a focused CLI test), or explicitly defer edit support with the rest of
the CLI work?

asafbennatan (commented):
See comments below

internal/cli/apply.go
galel12: Added in 12c09c9 — `EnrollmentHookPolicyKind` is now in the edit kind lists and `executePatchOperation` calls `PatchEnrollmentHookPolicyWithBodyWithResponse`. Covered by validate + PATCH unit tests.

asafbennatan (approved):
LGTM

asafbennatan (approved):

PR #3547, opened 2026-09-09
EDM-5889: ImageBuilder worker fails on Kubernetes when nested podman build cannot raise RLIMIT_NOFILE

lucamaf:
Replaces #2722 (closed due to inactivity; that branch had diverged 992 commits from main and could not be rebased in place).

Passes the ImageBuilder Worker's own current `RLIMIT_NOFILE` (soft and hard) through to the nested podman worker container via `--ulimit`, instead of hardcoding `1048576:1048576`. This fixes builds failing on Kubernetes with `operation not permitted` when the pod's own hard limit is lower than what buildah expects.

Addresses review feedback from #2722:
- Use `golang.org/x/sys/unix` instead of the frozen `syscall` package
- Cap a hard limit of `RLIM_INFINITY` to the soft limit before passing it to podman
- Ship `Ulimit=`/`LimitNOFILE=` in the `flightctl-imagebuilder-worker.container` quadlet unit so Podman deployments keep getting `1048576` by default (previously hardcoded in Go, now sourced from the unit)
- Update `configuring-imagebuilder.md` to describe the new dynamic behavior and how to raise the limit on each deployment type
- `EDM-5889:` commit prefix

Also adds unit test coverage for the ulimit-argument construction, including the `RLIM_INFINITY` capping case.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -432,6 +432,10 @@ CRI-O
  ImageContentSourcePolicy
  walkthrough
  arg
+ hardcoding
+ vendored
+ ulimits
+ ulimit
  - docs/user/references/cli-commands.md
 VNC
 stdin
--- a/deploy/podman/flightctl-imagebuilder-worker/flightctl-imagebuilder-worker.container
+++ b/deploy/podman/flightctl-imagebuilder-worker/flightctl-imagebuilder-worker.container
@@ -39,10 +39,17 @@ Volume=flightctl-worker-storage:/var/lib/containers
 #   Volume=/etc/flightctl/service-images-auth.json:/root/.config/containers/auth.json:ro,z
 PodmanArgs=--privileged
 
+# Sets this container's own RLIMIT_NOFILE, which the imagebuilder-worker process then
+# forwards to the nested podman build container (see "File Descriptor Limits" in
+# configuring-imagebuilder.md). 1048576 matches what buildah expects when raising its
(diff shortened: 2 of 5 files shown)


galel12:
Thanks for picking this back up and addressing the review feedback from #2722 — the `unix` import, `RLIM_INFINITY` capping, quadlet defaults, tests, and doc updates all look good.

The remaining blocker is **`lint-docs`** (spellcheck, not markdownlint — markdownlint already passes). CI flags four unknown words in `docs/user/installing/configuring-imagebuilder.md`:

- `hardcoding` (line ~775)
- `vendored` (line ~814)
- `ulimits` (line ~848)
- `ulimit` (line ~849, in “per-pod ulimit”)

**How to fix:** add those words to the file-specific section of [`.spelling`](https://github.com/flightctl/flightctl/blob/main/.spelling) (the block that starts with ` - docs/user/installing/configuring-imagebuilder.md`, around line 398). One word per line, same as the other entries there.

Then verify locally:

```bash
make lint-docs
```

(`make lint-docs` runs markdownlint and spellcheck; only spellcheck is failing here.)

Alternatively, `make fix-spelling` walks you through the same file interactively if you prefer.

Once that’s green we should be good to merge. Thanks again for the contribution!

galel12 (approved):
LGTM

PR #3548, opened 2026-09-09
EDM-5770: Align degraded VM application status

asafbennatan:
## Summary

- Report health-degraded VM applications as `Starting` instead of `Running`.
- Keep the application summary as `Degraded`.
- Update VM health and crash-loop regression expectations.

This preserves the existing API enum and aligns individual application status with the documented `Degraded` summary contract.

## Validation

- `make lint` — passed
- `go test ./internal/agent/device/applications/...` — passed
- `go test -race ./internal/agent/device/applications/...` — passed
- `make unit-test RACE=0 COVERAGE=0` — 6,924 tests completed; 5 unrelated failures remain in `internal/agent/device/console` and `internal/agent/device/systeminfo`

Jira: EDM-5770

### DIFF
--- a/internal/agent/device/applications/applications.go
+++ b/internal/agent/device/applications/applications.go
@@ -366,7 +366,9 @@ func (a *application) Status() (*v1beta1.DeviceApplicationStatus, v1beta1.Device
 		newStatus = v1beta1.ApplicationStatusStarting
 		summary.Status = v1beta1.ApplicationsSummaryStatusDegraded
 	case isRunningDegraded(total, healthy, initializing):
-		newStatus = v1beta1.ApplicationStatusRunning
+		// A partially available application is still recovering. Report Starting
+		// so the individual status remains consistent with a Degraded summary.
+		newStatus = v1beta1.ApplicationStatusStarting
 		summary.Status = v1beta1.ApplicationsSummaryStatusDegraded
 	case isErrored(total, healthy, initializing):
 		newStatus = v1beta1.ApplicationStatusError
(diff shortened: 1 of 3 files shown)


internal/agent/device/applications/applications.go
```diff
@@ -361,7 +361,9 @@ func (a *application) Status() (*v1beta1.DeviceApplicationStatus, v1beta1.Device
 		newStatus = v1beta1.ApplicationStatusRunning
 		summary.Status = v1beta1.ApplicationsSummaryStatusHealthy
 	case isRunningUnhealthy(total, healthy, unhealthy, initializing, stopped+stopping):
-		newStatus = v1beta1.ApplicationStatusRunning
+		// A health-degraded application is still recovering. Report Starting so
```
kkyrazis, line 364: I'm not sure I understand why `initializing == 0` would result in `Starting` status. Or why `isRunningDegraded` doesn't require the same fix.

asafbennatan: Good point. initializing == 0 means no workload is still in init/create; it does not mean the application is healthy. isRunningDegraded covers a partially available workload set (for example, one running and one stopped/died), so it now reports Starting as well. Since the API still has no individual Degraded status, Starting represents recovering or not fully in service here, not only literal initialization. Historically, EDM-240 introduced this path as Running plus a Degraded summary, and EDM-1046 explicitly preserved that behavior for manually stopped containers. That was intentional at the time. The current change aligns the individual status with the newer documented rule that a Degraded summary must have an application in Preparing or Starting.

kkyrazis (approved):

PR #3549, opened 2026-09-09
EDM-5264: reject duplicate vm published host ports

asafbennatan:
## Problem
The API accepted multiple VM applications on one device that requested the same published host port. The invalid configuration was persisted and later caused one workload to fail during runtime port binding. The same collision could also occur between a VM and a container application.

## Root Cause
Validation checked each application's port syntax and range, but did not enforce uniqueness across device applications that publish host ports.

## Fix
Track normalized published host-port/protocol bindings during validation for VM and container applications. Reject duplicates with an actionable validation error. An omitted protocol is treated as TCP; distinct protocols remain independently valid.

## Documentation
Documented device-scoped published-port uniqueness and protocol behavior in the device management guide.

## Testing
- `go test ./api/core/v1beta1 -run 'TestValidateApplications|TestValidateVmApplication' -count=1`
- `make unit-test` — 6,945 tests passed, 2 skipped
- `make build`
- `make lint`
- `make lint-docs`
- `make spellcheck-docs`
- `git diff --check`

## Confidence
High — the validation path is shared by direct device and fleet-template requests, and regression tests cover VM-to-VM and VM-to-container collisions.

## Rollback
Revert commits `704c59ccd` and `c56b78ac8` to restore the previous validation behavior.

## Risk Assessment
Low — this is a validation-only behavior change affecting conflicting VM/container published-port configurations.

Jira: EDM-5264

### DIFF
--- a/api/core/v1beta1/validation.go
+++ b/api/core/v1beta1/validation.go
@@ -1088,6 +1088,7 @@ func validateApplicationPortConflicts(app ApplicationProviderSpec, appName strin
 	return allErrs
 }
 
+// publishedPortKey normalizes a typed host-to-guest mapping for conflict checks.
 func publishedPortKey(port string) (string, bool) {
 	if !containerPortPattern.MatchString(port) {
 		return "", false
@@ -1098,6 +1099,10 @@ func publishedPortKey(port string) (string, bool) {
 	if err != nil || hostPort < privilegedPortRangeStart || hostPort > portRangeEnd {
 		return "", false
 	}
+	guestPort, err := strconv.Atoi(strings.SplitN(colonParts[1], "/", 2)[0])
+	if err != nil || guestPort < privilegedPortRangeStart || guestPort > portRangeEnd {
+		return "", false
+	}
 
 	protocol := "tcp"
 	if protocolParts := strings.SplitN(colonParts[1], "/", 2); len(protocolParts) == 2 {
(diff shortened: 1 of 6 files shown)


api/core/v1beta1/validation.go
```diff
@@ -1040,6 +1045,68 @@ func validateApplications(apps []ApplicationProviderSpec, fleetTemplate bool) []
 	return allErrs
 }
 
+func validateApplicationPortConflicts(app ApplicationProviderSpec, appName string, appType AppType, seenPorts map[string]string) []error {
```
kkyrazis, line 1047: what about quadlet apps? I know it would only work for inline defined quadlets, but should we validate those too?

api/core/v1beta1/validation.go
```diff
@@ -999,6 +999,9 @@ func (c ApplicationContent) IsPlain() bool {
 func validateApplications(apps []ApplicationProviderSpec, fleetTemplate bool) []error {
 	allErrs := []error{}
 	seenAppNames := make(map[string]struct{})
+	// EDM-5264: published host ports are scoped to the device, not to an
```
kkyrazis: should we be documenting jira tickets?

api/core/v1beta1/validation.go
asafbennatan: Done

kkyrazis (approved):

kkyrazis (approved):

PR #3552, opened 2026-09-09
NO-ISSUE: Avoid Helm uninstall PV reclaim race

asafbennatan:
## What

Avoid flaky Helm upgrade workflow failures caused by asynchronous PV reclaim after uninstall.

## Why

The post-delete hook deletes the Alertmanager PVC, but the storage provisioner may leave its `Released` PV visible briefly after Helm returns. The existing one-shot check fails during that window.

## Change

- Poll exact `claimRef.namespace` matches for up to 30 seconds.
- Check every 2 seconds, so successful cleanup still proceeds immediately.
- Preserve failure for PVs that remain after the timeout.

## Validation

- `git diff --check`
- Embedded workflow shell block passes `bash -n`.
- jq filter verified against representative PV JSON.

Fixes the cleanup flakiness seen in the Helm upgrade workflow.

### DIFF
--- a/.github/workflows/pr-helm-upgrade-testing.yaml
+++ b/.github/workflows/pr-helm-upgrade-testing.yaml
@@ -376,16 +376,36 @@ jobs:
             echo ""
           done
 
-          # Check for orphaned PVs that were bound to flightctl claims
-          ORPHANED_PVS=$(kubectl get pv --no-headers 2>/dev/null \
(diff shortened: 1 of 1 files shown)


kkyrazis (approved):
This seems fine, but I wonder if the original 30 seconds you had in your earlier commits would be more reliable.

PR #3553, opened 2026-09-10
NO-ISSUE: avoid empty Quadlet repository secrets

asafbennatan:
## Problem
`make deploy-quadlets` fails when delta repository credentials are not configured because it attempts to create zero-byte Podman secrets.

## Root Cause
`EDM-5218` added delta repository secret handling and mounted the credentials into both `flightctl-api` and `flightctl-delta-worker`. The API does not use these credentials. `deploy/scripts/secrets.sh` also attempted to create empty Podman secrets, which Podman rejects. `EDM-5220` subsequently added the same unconditional mounts to the delta-worker Quadlet.

## Fix
- Create delta repository secrets only when both user-provided credential environment variables are set.
- Automatically create the deployment-managed delta-worker systemd drop-in when credentials are provided, without mutating vendor-owned Quadlet files.
- Create the drop-in when both existing Podman secrets are present, allowing credentials to be added after deployment.
- Treat the generated drop-in as authoritative and overwrite it on later developer deployments.
- Clean up deployment-generated drop-ins and secrets.
- Remove the unnecessary API mount; credentials are consumed only by the delta worker.
- Preserve existing `make deploy-quadlets` restart behavior; active services are not restarted automatically. A drop-in added after deployment requires an explicit worker restart.

## Testing
- `bash -n` and ShellCheck for affected deployment scripts
- Mocked regression checks for unset credentials, provided credentials, secret creation, drop-in generation, and overwrite behavior
- `make lint`
- `make lint-docs`

## Confidence
High — optional credentials are created only from explicit user input, and drop-in generation is isolated to the delta worker.

## Risk Assessment
Low — the change only affects optional delta repository credentials; database/KV secrets and Helm deployments are unaffected.


deploy/scripts/secrets.sh
```diff
@@ -29,25 +29,49 @@ ensure_kv_secrets() {
...
+        elif sudo test -f "$dropin_file" && sudo grep -q '^# Generated by Flight Control deployment;' "$dropin_file"; then
+            sudo rm -f "$dropin_file"
+        fi
+        return 0
+    fi
+
+    ensure_env_secret "$username_secret" DELTA_GENERATION_DEFAULT_REPOSITORY_USERNAME
+    ensure_env_secret "$password_secret" DELTA_GENERATION_DEFAULT_REPOSITORY_PASSWORD
+
+    sudo install -d -m 0755 "$dropin_dir"
```
EfratIfergan, line 61: When this drop-in is created during a later make deploy-quadlets run, deploy_quadlets.sh only performs daemon-reload and systemctl start flightctl.target. If the target and delta-worker are already active, systemctl start does not restart the worker, so the new Secret mounts are not applied until a manual restart. Please restart or try-restart flightctl-delta-worker.service when this drop-in changes, and likewise when it is removed.

deploy/scripts/secrets.sh
```diff
@@ -29,25 +29,49 @@ ensure_kv_secrets() {
...
+    sudo tee "$dropin_file" >/dev/null <<EOF
```
EfratIfergan, line 62: This unconditionally overwrites dropin_file. The generated marker is checked only during cleanup, so a manually authored drop-in at this path (or a generated file with user additions) is silently replaced on the next deployment. Please refuse to overwrite an existing non-generated file, or use a separate generated filename.

asafbennatan:
Review follow-up: the earlier indirect-expansion comment was based on an older revision. The affected empty-value check and ensure_env_secret implementation have since been replaced; current validation covers unset credentials safely.

asafbennatan:
Review follow-up: addressed the two current actionable findings in 098a50bce. An already-active delta-worker is restarted when deployment changes the generated drop-in, and existing drop-in files are never overwritten; cleanup removes only generated files marked by the deployment.

asafbennatan:
Follow-up: removed the delta-worker-only restart. `make deploy-quadlets` preserves existing behavior and does not restart already-active services; a drop-in added or changed after deployment requires the documented explicit worker restart.

deploy/scripts/secrets.sh
asafbennatan: Thanks. I considered this, but removed the special restart to preserve existing `make deploy-quadlets` behavior: it reloads and starts the target but does not restart already-active services. A delta-worker-only restart would be inconsistent and surprising. The documentation now states that a drop-in added or changed after deployment requires an explicit `systemctl daemon-reload` and `systemctl restart flightctl-delta-worker.service`.

deploy/scripts/secrets.sh
asafbennatan: Addressed in the latest commit. This drop-in is deployment-managed for the non-production developer Quadlet path, so `make deploy-quadlets` intentionally rewrites it on subsequent runs. Users should not place unrelated custom settings in this generated file.

PR #3562, opened 2026-09-10
EDM-5958: Run systeminfo collection independently

kkyrazis:
Decouple systeminfo collection from the status push loop by adding an independent Run(ctx) method to the systeminfo manager. Collection now runs on its own goroutine with a configurable interval, while the Status() exporter returns cached results instead of triggering inline collection.

- Add Run(ctx) to Manager interface for independent collection
- Add cachedSystemInfo field to store last-collected results
- Status() returns cached data under mutex (no inline collection)
- Wire up collection goroutine via startAsync in agent setup
- Add system-info-periodic config section with interval field
- When system-info-periodic.interval is set, use it for collection
- When absent/zero, fall back to status-update-interval
- Add unit tests for Run lifecycle, caching, context cancellation, and config interval resolution (custom + fallback)

Assisted-by: Chai Bot <chai-bot@redhat.com>


internal/agent/device/systeminfo/manager.go
```diff
@@ -243,12 +203,35 @@ func (m *manager) Status(ctx context.Context, deviceStatus *v1beta1.DeviceStatus
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	if err != nil {
-		deviceStatus.SystemInfo = m.defaultSystemInfo()
-		return err
+		m.log.Warnf("System info collection failed: %v", err)
+		if m.cachedSystemInfo == nil {
+			m.cachedSystemInfo = new(m.defaultSystemInfo())
+		}
```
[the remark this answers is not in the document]
kkyrazis: What version of golang are we using? What are some of the newer features? Are the unit tests passing?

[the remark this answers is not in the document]
kkyrazis: Review the features defined here: https://go.dev/doc/go1.26 
and tell me if you still think `new` will not compile

internal/agent/device/systeminfo/manager.go
```diff
@@ -103,6 +109,8 @@ func (m *manager) Initialize(ctx context.Context) (err error) {
 		}
 	}
 
+	m.collect(ctx)
```
asafbennatan, line 112: Could we defer this initial collection until all runtime collectors have been registered? `Agent.Run` invokes `Initialize` before `NewAgentCertManager` registers the `managementCertSerial` and `managementCertNotAfter` collectors. Because non-forced `Status` now only returns this cache, the bootstrap/status collection can publish a snapshot without those configured fields, and an already-enrolled device will keep that incomplete snapshot until the first periodic tick. A forced collection after manager setup (or moving this initial collection) would keep startup status complete.

internal/agent/config/config.go
```diff
@@ -546,6 +566,9 @@ func (cfg *Config) validateSyncIntervals() error {
 	if cfg.StatusUpdateInterval < MinSyncInterval {
 		return fmt.Errorf("minimum status update interval is %s have %s", MinSyncInterval, cfg.StatusUpdateInterval)
 	}
+	if cfg.SystemInfoPeriodic.Interval > 0 && cfg.SystemInfoPeriodic.Interval < MinSyncInterval {
```
asafbennatan: The validation gate only rejects positive values below `MinSyncInterval`; a negative `system-info-periodic.interval` passes `Validate()` and is then silently treated as absent by `SystemInfoCollectionInterval`. That makes an invalid interval look valid and changes it to the fallback instead of surfacing the configuration error. Could we reject every nonzero value below the minimum while retaining zero as the documented fallback, and add a negative-value test?

asafbennatan (commented):
See comments below

internal/agent/device/systeminfo/manager.go
kkyrazis: I can't fully move the initial collection as bootstrapping requires system-info so that it can add labels to the enrollment request, but I've added logic to invoke runtime collectors once registration has completed.

asafbennatan (approved):
LGTM

PR #3564, opened 2026-09-14
NO-ISSUE: Fix make agent-vm not working due to missing rpms

celdrake:
`make rpm` can fail even when the RPM build itself succeeds. Mock sometimes errors during tmpfs chroot cleanup (Device or resource busy), packit exits non-zero, and `build_rpms_packit.sh` bailed out before moving artifacts to bin/rpm/. That left bin/rpm/ empty and broke downstream targets like make agent-vm.

This change salvages built RPMs when they exist in the packit output directory, regardless of packit’s exit code. A warning is logged when cleanup failed but artifacts were recovered; the script only fails when no RPMs were produced.

Made-with: Cursor


asafbennatan (commented):
See comments below

hack/build_rpms_packit.sh
```diff
@@ -96,16 +102,30 @@ install_packit
...
 else
   echo "::group::Building RPM locally"
-  run_local_build
+  run_local_build || BUILD_RC=$?
 fi
 echo "::endgroup::"
 
-move_artifacts
-cleanup_packaging_artifacts
+if artifacts_available; then
```
asafbennatan: This check accepts any nonzero Packit exit when even one matching RPM exists. The log in the description shows the intended tmpfs cleanup failure, but a different build failure could leave only some subpackages and still let `make rpm` write `bin/.rpm` for an incomplete set. Could we verify the expected RPM set before accepting a nonzero exit?

celdrake: Thanks, I have made an adjustment.

However, I checked with a different model and it suggested a simpler solution: Being able to disable mock clean-up as a configuration option, which should prevent the error altogether.

PTAL https://github.com/flightctl/flightctl/pull/3605

celdrake:
Simpler fix approved https://github.com/flightctl/flightctl/pull/3605

PR #3565, opened 2026-09-14
EDM-5222: Extract delta worker refactor layer

asafbennatan:
## Summary

This is the structural refactor layer for the delta worker. It moves the existing main-branch implementation into focused `internal/delta_worker` packages without changing the EDM-5222 behavior.

- Moves delta-worker configuration, models, store code, migration wiring, and related tests under the delta-worker package.
- Splits the worker consumer, delta generation, OCI helpers, and candidate resolution into focused packages.
- Keeps existing generation, retry, parsing, metric, and acknowledgement behavior intact.
- Includes parity coverage for the moved code.

## Scope and non-goals

- This PR is intentionally structural and contains no compatibility aliases.
- It does not introduce the delta-worker service contracts, service implementations, or persistence API; those are in the next stack PR (#3566).
- It does not change PrepareDeltas admission, queue acknowledgement, rollout policy, status transitions, public API behavior, or database schema.

## Stack order

This is the bottom PR in the new stack:

`main` → **#3565 refactor** → #3566 services → #3455 EDM-5222 behavior

The follow-up PRs should be reviewed against this branch rather than directly against `main`.

## Validation

- Focused delta-worker, task, device-service, and integration package compile checks passed during stack preparation.
- Existing unit and integration tests were moved with the implementation and retain their current behavior coverage.


galel12 (approved):
LGTM

PR #3577, opened 2026-09-16
NO-ISSUE: Extend validation of RolloutPolicy for DeltaGeneration

celdrake:
The current validation for `RolloutPolicy` does not allow a user to patch the fleet/device spec with only `deltaGeneration`

```
  rolloutPolicy:
    deltaGeneration:
      generateDelta: true
      maxWaitForDelta: 44m
```

results in 
```
Error 400: Bad Request: at least one of [DeviceSelection, DisruptionBudget] must be defined
```

Made-with: Cursor


asafbennatan (approved):
LGTM

PR #3578, opened 2026-09-16
EDM-6030: Expose image builder operation timeouts

asafbennatan:
## Summary
- expose image build/export inactivity timeout and timeout-check interval in Helm values
- render both settings for Helm and Podman deployments
- document the settings and cover Podman configuration propagation

## Validation
- make lint
- make lint-helm
- make lint-docs
- make spellcheck-docs
- make unit-test
- go test ./test/e2e/infra/quadlet -count=1
- helm template ... verified rendered timeout fields


itzikb-redhat (approved):

deploy/helm/flightctl/values.yaml
```diff
@@ -452,6 +452,10 @@ imageBuilderWorker:
   maxConcurrentBuilds: 2
   # -- Default TTL for image build resources
   defaultTTL: 168h
+  # -- Inactivity timeout for image builds and exports
+  imageBuilderTimeout: 3m
```
EfratIfergan, line 456: Could we validate `imageBuilderTimeout` as strictly greater than zero in `internal/config` and add a unit test?

The timeout checker currently treats values `<= 0` as invalid and silently skips all timeout checks. Therefore, setting `imageBuilderTimeout` to `0s` or a negative duration through this newly exposed Helm/Podman option would effectively disable protection against stuck ImageBuild/ImageExport resources.

If disabling timeouts is intentional, please make it an explicit and documented option instead.

asafbennatan: Done

EfratIfergan (approved):
LGTM

PR #3581, opened 2026-09-17
EDM-5287: use local registry image in multiorg rbac e2e

eldar101:
## Problem

The multiorg RBAC lifecycle and console-access test times out in disconnected-light because its application image is pulled from quay.io, while the profile uses the local E2E registry for disconnected execution.

## Root Cause

The multiorg suite started the auxiliary services, including the local registry that mirrors external test images, but discarded the returned Services object. The test therefore could not use the mirrored image.

## Fix

Retain the auxiliary services from BeforeSuite and build the RBAC application image reference from the local registry.

## Release target

Target release: release-1.4


itzikb-redhat:
@eldar101 This should wait for my fix as it doesn't pass on connected envs

PR #3586, opened 2026-09-19
NO-ISSUE: Reconcile AGENTS.md architecture guidance

asafbennatan:
## Summary

- Consolidate project-wide architecture, interface, constructor, and package-structure guidance in the root AGENTS.md.
- Document the distinction between legacy/shared control-plane packages and component-scoped service, store, and task layouts.
- Remove duplicated Summary sections from nested AGENTS.md files and align generator guidance with pinned modules.

## Validation

- `git diff --check`
- Tests not run; documentation-only change.


amir-yogev-gh:
## Review + Impact on Progressive Disclosure and Token Consumption

The architecture guidance content itself is well-written and genuinely useful - the two-pattern distinction (legacy/shared vs component-based), the runtime ownership table, mutation results, and interfaces/constructors
conventions are exactly the kind of guidance AI tools need. However, the **placement** of this content works against the progressive disclosure strategy we've been following for `AGENTS.md`, and has measurable token consumption implications.

### Concern 1: Root file growth vs. progressive disclosure

The root `AGENTS.md` is loaded on **every** AI tool session - whether the task is editing docs, tweaking Helm charts, fixing a test, or working on internal Go code. This PR grows it from ~67 lines / ~7.2KB to ~128 lines / ~10KB (~90% increase, roughly 800–1,000 extra tokens per session).

The added architecture/package layout section, mutation results, and interfaces/constructors guidance are relevant almost exclusively to `internal/` Go code. This is the classic signal for progressive disclosure: content that's important but not universally needed.

**Suggestion:** Move the architecture/package layout section (including mutation results and interfaces/constructors) to internal/AGENTS.md, and add a one-line pointer from the root:
- **Architecture:** See [internal/AGENTS.md](internal/AGENTS.md) for   package layout, patterns, and conventions.

### Concern 2: Summary sections removed from leaf files

The Summary sections in `api/`, `deploy/`, `docs/`, and` test/` `AGENTS.md` served as quick-reference checklists. These files are only loaded on demand (when the AI tool navigates into that directory), so the ~30 lines saved are **conditional** savings, not constant. The summaries provided fast orientation value at minimal cost.

**Suggestion:** Keep the Summary sections in the leaf files.

### Concern 3: "Before committing" lost its exact commands

The original checklist named exact commands: `make lint`, `make tidy`, `make unit-test`, `make integration-test`. The PR replaces these with vague references like "run the dependency cleanup command" and "use the relevant project lint target." This forces AI tools to follow a pointer to another file to discover the exact command - adding a round trip and risking the tool guessing instead of looking it up.

**Suggestion:** Restore the explicit command names in the root checklist.
Area-specific files can add nuance, but the root should be directly actionable.

### What to keep

- The `make generate` / `go:generate` directives cleanup is a genuine  improvement - keep it.
- The architecture content itself is high quality - it just belongs in `internal/AGENTS.md` rather than the root.

### Summary
Move internal/-specific content to `internal/AGENTS.md`, preserve the leaf-file summaries, and restore exact commands in "Before committing." This gives you rich contextual guidance where it matters and a lean root file that doesn't tax every session.

asafbennatan:
Thanks for the detailed review. After reconsidering the progressive-disclosure concern, I made the following changes:

1. I moved the detailed architecture and package-layout guidance—including the runtime ownership table, mutation-result rules, and interface/constructor conventions—to a new `internal/[AGENTS.md](https://github.com/home/abennata/dev/flightctl/internal/AGENTS.md)`. The root now retains only the general architecture boundaries and a direct pointer to the internal guidance. The runtime mapping remains centralized so it is available before choosing a component directory.
2. I kept the leaf summaries removed because they repeat their preceding sections without adding requirements. Agents consume the complete applicable `AGENTS.md`, so duplicate summaries add context and another copy that can drift.
3. I restored the “Before committing” section exactly to its pre-PR wording, including the explicit `make lint`, `make tidy`, `make unit-test`, and `make integration-test` commands.

amir-yogev-gh (approved):
LGTM

PR #3591, opened 2026-09-22
NO-ISSUE: stabilize telemetry gateway integration startup

asafbennatan:
## Summary

- eliminate the ephemeral-port close/rebind race in telemetry gateway integration startup
- retry gateway startup after address-in-use failures
- wait for mTLS readiness on the actual gateway listener before forwarding tests begin

## Problem

The forwarding integration tests selected free ports by briefly opening and
closing listeners before starting the gateway. Under parallel execution, a
different process could claim one of those ports before the gateway started.
The readiness check then connected to the wrong plain listener, producing
`tls: first record does not look like a TLS handshake`; the intended OTLP
receiver separately reported `address already in use`.

## Fix

The test harness now allocates fresh gateway, Prometheus, and internal OTel
metrics ports immediately before each startup attempt. It retries only bind
collisions, refreshes dependent exporter configuration for each attempt, and
confirms readiness with the gateway's CA and client certificate before returning
control to the specs.

## Validation

- `make build`
- `make unit-test` — 7,269 passed, 2 skipped
- `make lint` — 0 issues
- focused telemetry gateway integration suite — 9 passed
- full integration suite — 20 suites passed, 0 failed


asafbss (approved):
LGTM, ty

PR #3594, opened 2026-09-22
EDM-5770: Adjust e2e test 90246 to recent vm status change

itzikb-redhat:
A VM whose containers are up but health-degraded used to be Running + Degraded, now it's Starting + Degraded.
Adjust e2e test to the new behavior.


asafbennatan (approved):
LGTM

PR #3569, opened 2026-09-14
EDM-5706: Add EnrollmentHooks condition, approval snapshot, and notify secrets

galel12:
## Summary

Implements the runtime plumbing for enrollment hooks: device condition tracking, approval-time policy snapshot, encrypted bearer token storage, worker enqueue wiring, and device-deletion secret purge.

**Jira:** EDM-5706
**Stacked on:** #3544 (`EDM-5703-enrollment-hook-policy`)

> **Stacked PR.** Targets `EDM-5703-enrollment-hook-policy` (#3544).
> Stack: `main ← EDM-5703 ← EDM-5706`. Review/merge parents first;
> GitHub retargets this PR automatically as they merge.

Supersedes closed fork PR #3568 (synced via sync-bot onto upstream).

### OpenAPI & Code Generation
- Add `DeviceEnrollmentHooksStatus`, `EnrollmentHookSnapshot`, and `EnrollmentHookSnapshotAction` types
- Add `enrollmentHooks` field to `DeviceStatus`
- Add `EnrollmentHooks` to the `ConditionType` enum

### Encrypted Notify Secrets Store
- New `internal/store/enrollmenthooknotifysecrets/` package
- Custom encryption migration adapter keyed by `(device_name, action_index)`

### Approve Handler
- Snapshot `policy.Spec.AfterEnrolling` at approval time
- Store bearer tokens in encrypted notify secrets store
- Set initial `EnrollmentHooks` condition (`Status=False`, `NotifyPending` or `Pending`)

### Worker Enqueue Wiring
- Register `EventReasonEnrollmentRequestApproved` in `worker_client.eventReasons`
- Add `shouldEnrollmentHookNotify` guard + dispatch stub (handler in EDM-5707)

### Device Deletion Purge
- Notify secrets purge in same transaction as device deletion

### Tests
- Unit + integration tests for snapshot, condition, store CRUD, approve paths, immutability, purge

Made with [Cursor](https://cursor.com)

PR #3596, opened 2026-09-22
EDM-5705: Fleet selector and rendered-spec enrollment hooks gate

galel12:
## Summary
Implements the fleet selector and rendered device spec enrollment hooks gate (EDM-5705).

When a device's `EnrollmentHooks` condition has `Status=False` (hooks still running or failed):
1. **Fleet selector** excludes the device from fleet matching
2. **Rendered device spec** returns HTTP 409 Conflict with a gate reason (before `WaitForNotification`)

Devices with `EnrollmentHooks=True` (Succeeded, Continued, ManualOverride) or no condition pass through normally.

## Notes
- Synced from chai-bot PR #3595 after EDM-5706 (#3569) squash-merged to `main`
- Stack base was deleted with #3569; this PR targets `main` directly

## Jira
EDM-5705

Made with Cursor


galel12:
Closing to retarget onto `main` after EDM-5706 merged; stack membership blocked changing the base branch. Replacement PR incoming.

galel12:
Superseded: retargeting onto main (stack blocked base change).

internal/tasks/fleet_selector.go
```diff
@@ -356,6 +363,13 @@ func (f FleetSelectorMatchingLogic) handleDevicesMatchingFleet(ctx context.Conte
 				return devicesProcessed, errors
 			}
 
+			// Primary gate: skip devices whose enrollment hooks have not
+			// reached a terminal success state.
+			if domain.IsDeviceEnrollmentHooksGated(&device) {
+				f.log.Infof("Skipping fleet matching for device %s/%s: enrollment hooks gate is active", f.orgId, lo.FromPtr(device.Metadata.Name))
+				continue
```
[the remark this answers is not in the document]
galel12: Thanks for catching this. Both `handleOrphanedDevices` and `handleDevicesWithMultipleOwnersCondition` could recompute ownership for gated devices. I added `IsDeviceEnrollmentHooksGated` checks before recomputation in both handlers, along with regression coverage through `FleetSelectorUpdated`. The focused tests and full `internal/tasks` package pass, and `make lint` is clean.

internal/service/device/handler.go
```diff
@@ -557,6 +557,23 @@ func (h *DeviceServiceHandler) GetRenderedDevice(ctx context.Context, orgId uuid
 		processedAwaitReconnect = h.processAwaitingReconnectIfNeeded(ctx, orgId, name, params.KnownRenderedVersion)
 	}
 
+	// Enrollment hooks gate: check before WaitForNotification so agents don't
+	// long-poll while gated. GetRendered() (below) also loads the full device
+	// including service_conditions, but it runs after WaitForNotification —
+	// too late for the gate since the agent would already be blocked.
+	device, deviceErr := h.deviceStore.Get(ctx, orgId, name)
```
asafbennatan, line 564: we need to be careful this is a hotpath that has performance implications , we should not be adding another DB call , see if you can move the already existing devicestore.getRendered up so you can use it .

the other option is to not gate this here , but reject spec updates for devices that did not finish running the hook - but this is a different UX ( the current approach allows users to perform spec updates while this does not )

internal/tasks/fleet_selector.go
```diff
@@ -102,6 +102,13 @@ func (f FleetSelectorMatchingLogic) DeviceLabelsUpdated(ctx context.Context) err
 		return fmt.Errorf("%s", errorMsg)
 	}
 
+	// Secondary gate: skip fleet matching for devices whose enrollment hooks
+	// have not reached a terminal success state.
+	if domain.IsDeviceEnrollmentHooksGated(device) {
+		f.log.Infof("Skipping fleet matching for device %s/%s: enrollment hooks gate is active", f.orgId, f.event.InvolvedObject.Name)
+		return nil
```
asafbennatan, line 109: This early return skips the ownership pass for a newly approved device while its hooks are pending. When EnrollmentHooks later changes from False to True, the current event flow does not rerun ownership reconciliation for a condition-only update, so the device can remain unowned and never receive its matching fleet spec until an unrelated label or fleet-selector change. Could we enqueue reconciliation when the gate clears and add a False-to-True lifecycle test?

internal/tasks/fleet_selector.go
```diff
@@ -648,6 +662,11 @@ func (f FleetSelectorMatchingLogic) handleOrphanedDevices(ctx context.Context, f
 				return devicesProcessed, errors
 			}
 
+			if domain.IsDeviceEnrollmentHooksGated(&device) {
```
asafbennatan, line 665: The added gate covers the orphaned-device and multiple-owner passes, but fleet deletion and empty-selector updates use clearFleetOwnershipFromDevices, which still recomputes owners without checking EnrollmentHooks. A gated device that already has this fleet as owner can be moved to another matching fleet while hooks are pending. Could that cleanup path honor the gate too?

internal/service/device/handler.go
galel12: Agreed that this path needs care. I kept the lightweight `deviceStore.Get` before `WaitForNotification` so gated devices still receive the current 409 response before the long-poll starts. That does add a DB read. Moving `GetRendered` ahead of the wait would avoid that separate lookup, but it would also load/decrypt the rendered config on every poll that ultimately returns 204; a notification arriving during the wait may still make that read stale. Avoiding the lookup altogether would need a different gate source or the spec-update rejection UX you noted.

internal/tasks/fleet_selector.go
galel12: Fixed in `4c2bbb30f`: the device service now emits a targeted `ResourceUpdated` event only when `EnrollmentHooks` changes from `False` to non-`False`, and the ownership consumer reconciles on that field. The lifecycle integration test verifies the persisted event and that the matching fleet owner is assigned once the gate clears.

internal/tasks/fleet_selector.go
galel12: Fixed in `4c2bbb30f`: `clearFleetOwnershipFromDevices`, shared by deleted-fleet and empty-selector cleanup, now skips devices while the `EnrollmentHooks` gate is active. The gate-clear reconciliation event from the other fix allows ownership to be recomputed once hooks finish. Regression tests cover both cleanup paths.

internal/tasks/consumer.go
```diff
@@ -252,7 +252,7 @@ func shouldReconcileDeviceOwnership(ctx context.Context, event domain.Event, log
 
 	// If a device's labels were updated, return true
 	if event.Reason == domain.EventReasonResourceUpdated && event.InvolvedObject.Kind == domain.DeviceKind {
-		return hasUpdatedFields(event.Details, log, domain.Labels)
+		return hasUpdatedFields(event.Details, log, domain.Labels, domain.UpdatedFieldEnrollmentHooksCondition)
```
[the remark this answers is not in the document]
galel12: I checked the current path: an unowned device does not return early here. `getOwnerFleet` treats a nil owner as eligible for fleet matching, so `DeviceLabelsUpdated` continues to compare labels and assign the matching fleet. The gate-clear event is routed to this handler, and the lifecycle integration test passed at this commit in the full integration run. I don’t see a code change needed for this finding; the reported failure appears inconsistent with the current code and test result.

internal/service/device/handler.go
galel12: Addressed in 6719166aafb7eff181d3a921a9e5190905ba4317 by reusing the rendered-device read for the pre-wait gate check. The handler reuses that snapshot when its version matches the notification and refreshes only when needed.

asafbennatan: Thanks, this reuses the read when a spec notification matches the fetched version. The no-notification path still calls `GetRendered` before waiting and then returns 204, so each idle poll reads the device and decrypts rendered config and applications that it discards. Before this gate was introduced, that path returned 204 without a store read. That’s the hot-path cost I was raising. Can we avoid the read on idle polls, or use an existing source for the gate state? As written, I still see the performance concern as open.

galel12: You're right — pre-wait `GetRendered` + fallback `Get` still added reads on the hot path. Gating now uses only the existing post-wait `GetRendered` result so idle `204`s stay read-free (`48a606e0e`).

asafbennatan (approved):
LGTM

